PR #14169, opened 2017-05-25
iOS: Introduce API for moving screen reader's focus

rigdern:
This change introduces an API, `setAccessibilityFocus`, which moves the screen reader's focus to the passed in element. This causes VoiceOver to announce the element and draw a focus rectangle around it.

Similar functionality is already available in RN Android through the `sendAccessibilityEvent` method. Here's an example of what exists today in RN Android:

```
RCTUIManager.sendAccessibilityEvent(
  node,
  8 /* TYPE_VIEW_FOCUSED */);
```

## Test Plan (required)

Called `setAccessibilityFocus` on a couple of elements to verify that focus does indeed move when VoiceOver is enabled. Additionally, my team is using this change in our app.

Adam Comella
Microsoft Corp.

### DIFF
--- a/React/Modules/RCTAccessibilityManager.m
+++ b/React/Modules/RCTAccessibilityManager.m
@@ -9,6 +9,7 @@
 
 #import "RCTAccessibilityManager.h"
 
+#import "RCTUIManager.h"
 #import "RCTBridge.h"
 #import "RCTConvert.h"
 #import "RCTEventDispatcher.h"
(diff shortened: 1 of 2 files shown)


React/Modules/RCTAccessibilityManager.m
```diff
@@ -162,6 +163,16 @@ - (void)setMultipliers:(NSDictionary<NSString *, NSNumber *> *)multipliers
   self.multipliers = multipliers;
 }
 
+RCT_EXPORT_METHOD(setAccessibilityFocus:(nonnull NSNumber *)reactTag)
+{
+  dispatch_async(RCTGetUIManagerQueue(), ^{
+    [self.bridge.uiManager addUIBlock:^(__unused RCTUIManager *uiManager, NSDictionary<NSNumber *, UIView *> *viewRegistry) {
```
javache: This will make it so that we'll wait for the next UI batch to complete. If you don't want that, dispatch directly to the main queue and use RCTUIManager's viewForTag method.

rigdern: @javache Sweet, I wasn't aware of `viewForReactTag:`.

I addressed your feedback.

PR #14145, opened 2017-05-24
fix memory leak

yk3372:
Signed-off-by: yk3372 <yk3372@gmail.com>

## Motivation (required)

What existing problem does the pull request solve?

ViewManagersPropertyCache, ViewManagerPropertyUpdater, MessageQueueThreadRegistry static field not release when ReactInstanceManager called destroy.

## Test Plan (required)

I use this url to integrate RN:
[http://facebook.github.io/react-native/docs/integration-with-existing-apps.html](http://facebook.github.io/react-native/docs/integration-with-existing-apps.html)
and in Activity's onDestroy function add the follow code to release RN:
```java
mReactInstanceManager.destroy();
mReactInstanceManager = null;
```
and then when I exit activity, find the static field not release.
the follow is screen shot:
before:

after:

PR #14172, opened 2017-05-25
fix memory leak

yk3372:
Signed-off-by: yk3372 <yk3372@gmail.com>

## Motivation (required)

ViewManagersPropertyCache, ViewManagerPropertyUpdater, MessageQueueThreadRegistry static field not release when ReactInstanceManager called destroy.

## Test Plan (required)

I use this url to integrate RN:
[http://facebook.github.io/react-native/docs/integration-with-existing-apps.html](http://facebook.github.io/react-native/docs/integration-with-existing-apps.html)
and in Activity's onDestroy function add the follow code to release RN:
```java
mReactInstanceManager.destroy();
mReactInstanceManager = null;
```
and then when I exit activity, find the static field not release.
the follow is screen shot:
before:

after:


yk3372:
@javache  I has made some change, and has made the test run success. Thanks

https://github.com/facebook/react-native/pull/14145

PR #14176, opened 2017-05-25
[docs] Update 0.45 docs with new Getting Started material

dikaiosune:
@hramos -- I *think* I've got everything in this cherry pick that came up in the last few docs PRs. Let me know if you'd like me to include any other docs commits that should be backported?


hramos:
LGTM

PR #14190, opened 2017-05-25
Fix Incorrect Build Script Bash False-Matching

hnryjms:
When attempting to inverse a Bash boolean, the `!` character converts the meaning to become `[[ ! -n false ]]` (checking that the string "false" is empty) which never matches, and therefore this condition can never occur.

See https://stackoverflow.com/a/2953673


jeanregisser:
Nice! I stumbled upon this bug a couple of months ago and forgot about it.

PR #14196, opened 2017-05-25
WIP: Remove Haste from Libraries/ and convert to relative requires

skevy:
Was chatting with @cpojer, and we were discussing how if we remove Haste requires from RN, we can actually rid ourselves of the awful and unstable `providesModuleNodeModules` option in the packager and in jest-haste-map (at least in OSS). Now that React is use in RN with a flat bundle (thanks @bvaughn!), and `fbjs` Haste requires were removed long ago, this seems like it might be a good move.

This would change Facebook's internal RN development workflow for anyone that works on RN core, but wouldn't change it for product engineers at Facebook, theoretically. The @providesModule pragmas are still here -- so if a FB engineer says in product code, `require('View')`, things should still work. But in OSS, we can turn off the providesModuleNodeModules option and just treat RN like a regular npm dependency.

Marking this as WIP -- as I haven't vetted this super well yet. Curious what others thoughts are.


skevy:
Oh, also, here's the script I used to do this. Maybe not the prettiest thing ever, but it gets the job done. 

https://gist.github.com/skevy/c323242673fb2c124269fd53b607d878

Libraries/Alert/Alert.js
```diff
@@ -11,9 +11,9 @@
...
 'use strict';
 
-const AlertIOS = require('AlertIOS');
-const NativeModules = require('NativeModules');
-const Platform = require('Platform');
+const AlertIOS = require('./AlertIOS');
+const NativeModules = require('../BatchedBridge/NativeModules');
+const Platform = require('../Utilities/Platform');
 
 import type { AlertType, AlertButtonStyle } from 'AlertIOS';
```
skevy: Just realized that I didn't modify the `import type` statements. Not sure if this is necessary, since flow in this repo is working in `haste` mode, but maybe I should change them for both consistency as well as so that that module mode isn't required for people consuming RN with Flow.

ericvicenti:
This approach looks great to me!

Clearly the lint bot needs to be reconfigured as well

davidaurelio:
Please sync up with @rozele how this is going to work with `react-native-windows`. They are currently overriding core modules using haste, e.g. [`Text.windows.js`](https://github.com/Microsoft/react-native-windows/blob/master/Libraries/Text/Text.windows.js)

rozele:
Thanks @davidaurelio. @skevy for third-party plugin platforms like [Microsoft/react-native-windows](http://github.com/Microsoft/react-native-windows), we need the ability to override core modules. Throughout the react-native codebase, there are conditionals that allow differentiated native behaviors for iOS and Android, e.g. https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollView.js#L583-L593.  We need to be able to also add in conditions for native behaviors of plugin platforms.

I'd be as happy as anyone to see `providesModuleNodeModule` go away, but we need to replace it with something equally flexible.

skevy:
@rozele seems reasonable! I'll give it some thought. Thanks for letting me know!

skevy:
Gonna close this for now.

The path forward is:

- Figure out a sane way to handle new React Native platforms without relying on Haste (this is a big project that requires some careful thought -- I can open up a new issue on the metro-bundler repo when that exists and we can discuss this further)

- Perhaps in the meantime, we can add a pre-publish step that strips haste from the JS in libraries, and publish two folders in the NPM release -- `Libraries` and `Libraries-Haste`. I'll open up a new PR that outlines my proposal here.

Thanks for the feedback everyone -- we'll do this eventually!

vovkasm:
But now... platforms easily handled without Haste...

```javascript
    import { Platform } from 'react-native'

    const Button = Platform.select({
        ios: () => require('./button-ios'),
        android: () => require('./button-android')
    })

    export default Button
```

PR #14200, opened 2017-05-26
iOS: Improve accessibility of multiline TextInputs

### LINKED ISSUE #13691: TextInput with multiline={true} does not support accessibility on iOS
### Description

A multiline `<TextInput>` does not support any sort of accessibility on iOS.

1. Add the following text input:
```
<TextInput multiline={true} />
```
2. Turn on VoiceOver
3. Navigate to the element
4. Observe that no text is spoken for the element.

### Solution

When multiline={false} is set, accessibility works as expected on iOS. `multiline={true}` should behave in the same way.

### Additional Information

* React Native version: latest
* Platform: iOS
* Development Operating System: MacOS
* Dev tools: Xcode, VSCode

rigdern:
Fixes #13691.

Prior to this change, VoiceOver generally didn't say anything when selecting a multiline TextInput. The only exception was VoiceOver would announce the TextInput's placeholder if it had one set.

After this change, VoiceOver announces the following for multiline TextInputs:
  - The TextInput's content.
  - The TextInput's placeholder if the TextInput doesn't have any content in it.
  - The fact that the TextInput is a textfield.
  - The cursor position if the TextInput is being edited.

This is similar to the behavior of single line TextInputs.

This change achieves this by disabling `RCTTextView` as an accessibility element. `RCTTextView` is a subclass of `RCTView` so VoiceOver doesn't recognize this as a textfield. Instead, VoiceOver now sees the child `RCTUITextView` which is a subclass of `UITextView` so VoiceOver does recognize it as a textfield. Additionally, an `accessibilityLabel` implementation was added to `RCTUITextView` in order to take the value of the placeholder into account.

## Test Plan (required)

Verified the announcements of TextInputs with various props:
  - No placeholder and no content
  - Just a placeholder
  - Just content
  - Both a placeholder and content

Did this for both singe line inputs and multiline inputs. For setting content in multiline inputs, I tested both using the `value` prop and passing children. All other props being equal, these configurations resulted in similar announcements.

I verified that the following accessibility props work the same on singleline and multiline TextInputs:
  - `accessible`
  - `accessibilityLabel`
  - `accessibilityTraits`
  - `accessibilityViewIsModal`
  - `onAccessibilityTap`
  - `onMagicTap`

Additionally, my team has been using this change in our app.

Adam Comella
Microsoft Corp.


Libraries/Text/RCTUITextView.m
```diff
@@ -51,6 +51,25 @@ - (void)dealloc
   [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+- (NSString *)accessibilityLabel
+{
+  NSMutableString *accLabel = [NSMutableString stringWithString:@""];
```
shergin: Should be `NSMutableString *accessibilityLabel = [NSMutableString new];`.
(Please, use `new` and, please, avoid abbreviation.)

shergin:
(Just for reference: This PR also fixes https://github.com/facebook/react-native/issues/13691 and internal task t17745139.)

shergin:
And seems that this PR also breaks common support for `accessibilityLabel` and co.

rigdern:
@shergin Can you elaborate on what this PR breaks?

rigdern:
@shergin I see what you mean. It was confusing because `accessibilityLabel` is both the name of a method I implemented and the name of a prop. Initially, I didn't realize you were talking about the prop.

I addressed your feedback and ensured that the accessibility props work the same on singleline and multiline TextInputs:
  - `accessible`
  - `accessibilityLabel`
  - `accessibilityTraits`
  - `accessibilityViewIsModal`
  - `onAccessibilityTap`
  - `onMagicTap`

shergin:
@rigdern Sorry, I was not clear! 😞 
I am really appreciate your effort. Accessibility stuff is really important!
Let me think about it couple of days and we will definitely merge this next week.

shergin:
@rigdern I just landed 072d2709df5ca866a300f58ab4d2934aa0fdde66 which removes necessity to proxy all accessibility props to subview. So, we have to revert all changes in `RCTTextView.m`. Sorry about that.

shergin:
I just did it. I believe we are good to go! I will merge it on Monday.

PR #14210, opened 2017-05-26
Update ListView Basics to use FlatList, SectionList.

hramos:
## Motivation (required)

The new list views, FlatList and SectionList, are recommended over ListView.

## Test Plan (required)

Built website on localhost and verified the guide is rendered correctly.

screencapture-localhost-8079-react-native-docs-using-a-listview-html-1495834607096


website/core/SnackPlayer.js
```diff
@@ -51,33 +53,44 @@ var SnackPlayer = React.createClass({
...
     var code = encodeURIComponent(this.props.children);
     var params = this.parseParams(this.props.params);
-    var platform = params.platform ? params.platform : 'ios';
-    var name = params.name ? decodeURIComponent(params.name) : 'Example';
+    var platform = params.platform
+      ? params.platform
+      : 'ios';
+    var name = params.name
+      ? decodeURIComponent(params.name)
+      : 'Example';
```
hramos: Blame Prettier for this.

website/core/SnackPlayer.js
```diff
@@ -15,8 +15,10 @@ var React = require('React');
 
 const PropTypes = require('prop-types');
 
-const LatestSDKVersion = '15.0.0';
+const LatestSDKVersion = '16.0.0';
```
hramos: Not using 17 here because the Snack player doesn't seem to support it.

PR #14231, opened 2017-05-27
replace Float.compare(value, UNDEFINED) == 0 with Float.isNaN to impr…

ransj:
Float computations are costly on mobile platform. The replacement of "Float.compare(value, UNDEFINED) == 0" can help improve performance.

Test case are like this:
    public static void main(String[] args) {
        // wait 5s for cpu idle
        for (long start = System.currentTimeMillis(); start + 5000 >= System.currentTimeMillis(); ) {
            Thread.yield();
        }
        int count = 10000;
        long start = System.currentTimeMillis();
        for (int i = 0; i < count; i++) {
            Float.isNaN(Float.NaN);
            Float.isNaN(100f);
        }
        System.out.println(" Float.isNaN cost " + (System.currentTimeMillis() - start));
        start = System.currentTimeMillis();
        for (int i = 0; i < count; i++) {
            Float.compare(Float.NaN, Float.NaN);
            Float.compare(Float.NaN, 100f);
        }
        System.out.println(" Float.Float.compare cost " + (System.currentTimeMillis() - start));
    }

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/yoga/YogaConstants.java
+++ b/ReactAndroid/src/main/java/com/facebook/yoga/YogaConstants.java
@@ -14,7 +14,7 @@ public class YogaConstants {
   public static final float UNDEFINED = Float.NaN;
 
   public static boolean isUndefined(float value) {
-    return Float.compare(value, UNDEFINED) == 0;
+    return Float.isNaN(value);
   }
 
   public static boolean isUndefined(YogaValue value) {


javache:
Can you submit this pull-request to https://github.com/facebook/yoga instead?

It would also be helpful to include the results of your benchmark, and not just the benchmark code.

ransj:
@javache is it means that the commit to https://github.com/facebook/yoga will finally affect RN project?

javache:
Yup, all changes are automatically synced.

PR #14238, opened 2017-05-29
upgrade image-size version fix caculate webp image size bugs

Lxxyx:
When I try to use webp for app. But some webp images get an error message `unsupported file type` by image-size@0.3.5.

And I in image-size repo I found a issue https://github.com/image-size/image-size/issues/65.
This is the same problem.

And try image-size^0.5.1 the error fixed.

image


Lxxyx:
@shergin The new pull request is here.
Thank you.

Lxxyx:
@shergin import failed but I don't know what makes test failed. Would you help me?

shergin:
@Lxxyx I believe we also have to update `DependencyGraph-test.js`. Could you please do this?

shergin:
It can not be landed because this diff breaks a packages consistency test.

Issue #14244, opened 2017-05-29
OkHttpClientProvider for Android 4 does not support TLSv1


PaulVanStaden:
#### Correctly functioning behaviour.
In Android v5 and upwards when using the standard OkHttpClientProvider it uses OkHttpClient.Builder() to determine the connectionSpecs. This returns a default
```
DEFAULT_CONNECTION_SPECS = Util.immutableList(
      ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS, ConnectionSpec.CLEARTEXT);
```
#### Inconsistent behaviour 
For Android SDK `Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN && Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT` it goes into a custom `enableTls12OnPreLollipop` codeblock that specifically only implements `ConnectionSpec.MODERN_TLS`

This results that connections using TLSv1 will not work on Android 4. 

### Reproduction Steps and Sample Code

0. Run an emulator on API 16. 
0. Try a SSL network request to a SSL server only accepting TLSv1
0. Fail network request due to the following error
```
javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb9561090: Failure in SSL library, usually a protocol error
error:14077102:SSL routines:SSL23_GET_SERVER_HELLO:unsupported protocol (external/openssl/ssl/s23_clnt.c:714 0x9e0fd925:0x00000000)
```

### Solution

Will Add pull request. 
Added #14245

in com.facebook.react.modules.network.TLSSocketFactory replace 
```
private Socket enableTLSOnSocket(Socket socket) {
        if(socket != null && (socket instanceof SSLSocket)) {
            ((SSLSocket)socket).setEnabledProtocols(new String[] {"TLSv1.1", "TLSv1.2"});
        }
        return socket;
    }
```
with 
```
private Socket enableTLSOnSocket(Socket socket) {
        if(socket != null && (socket instanceof SSLSocket)) {
            ((SSLSocket)socket).setEnabledProtocols(new String[] {"TLSv1","TLSv1.1", "TLSv1.2"});
        }
        return socket;
    }
```
### Additional Information

* React Native version: v0.45.0-rc.2
* Platform: Android
* Development Operating System: OSX and Windows
* Dev tools: Android Studio version, Android SDK 25

hramos:
[...]

PR #9840, opened 2016-09-10
Enable TLS 1.1 and TLS 1.2 on Android 4.1-4.4

fkoester:
This is a proposed patch for issue #7192.

Android 4.1-4.4 has support for TLS 1.1 and 1.2 but it is disabled by default. Because of the known security issues and more and more servers switching to TLS 1.2 only, it would be nice for react-native to enable this support.

I demonstrated a demo application which showcases the problem and can be used to test this patch. All sources and documentation for it can be found here:

https://github.com/bringnow/react-native-tls-test

Credits to Alex Gotev (@gotev) for the nice implementation.

PR #14245, opened 2017-05-29
Adding TLSv1 support for custom SSL socket factory

PaulVanStaden:
## Motivation (required)
This problem tries to solve issue #14244 

## Test Plan (required)

Implemented a fetch request to a TLSv1 server. It now successfully resolves TLS handshake. 
Tested same fetch to TLSv1.2 server and still successfully resolves TLS handshake.


shergin:
I believe this is okay because original revision #9840 (which introduced this) does not have an intent to disable TLSv1.

PR #14255, opened 2017-05-30
Fixes wrong scalesPageToFit property (WebView component) documentation.

haraldreingruber:
## Motivation

The scalesPageToFit doesn't seem to be `true` by default. Here is another documentation of this observation: https://react-native.canny.io/feature-requests/p/webview-zoom

## Test Plan

This is probably not required, as this is only a small change in the documentation.


haraldreingruber:
Not sure if I applied the fix to the right location. Looking forward to your feedback :)

peterp:
I've created [a pull-request](https://github.com/facebook/react-native/pull/14363) to fix the default props for WebView on iOS because I believe the documentation is the expected behaviour.

PR #14261, opened 2017-05-30
Android: Set WebViewManager methods/fields as protected instead of private

cbrevik:
**Motivation**

See discussion in #10946. The motivation is to make `ReactWebViewManager` more extensible.

Re-using logic from the ReactWebViewManager when implementing your own custom WebView is a pain since so much of the logic is set as `private`.

This PR makes for easier extension/overriding of behavior, and less duplication of code, since most of the methods/fields are set as `protected` instead.

I've also made some "create" methods for the `WebView` and `WebViewBridge` so they can more easily be overridden.

**Test plan**
The test plan is the same as the other PR (#10946). I've made an simple test app which extends `RCTWebViewManager`: https://github.com/cbrevik/overrideWebview

See [CustomWebViewManager.java](https://github.com/cbrevik/overrideWebview/blob/master/android/app/src/main/java/com/overridewebview/CustomWebViewManager.java) for a simple implementation.

CC @shergin (https://github.com/facebook/react-native/pull/10946#issuecomment-304763562)


cbrevik:
@shergin mind taking another look at this now that the iOS ViewManager PR has landed? (Thanks btw!)

Would love to get the ball rolling on #10946 again 😄

shergin:
@facebook-github-bot shipit

PR #14267, opened 2017-05-31
Fix configure glog script when building from xcodebuild

janicduplessis:
I encountered an issue when building with fastlane gym / xcodebuild where glog would not build because of missing config.h header file. I tracked it down to the ios-configure-glog.sh script that ended up error-ing because of missing valid c compiler. I guess it didn't enter the if to set c compiler env in xcodebuild and that env doesn't have proper values set like it does in xcode so just removing this check fixed it. Also tested that it still works properly in xcode.

### DIFF
--- a/scripts/ios-configure-glog.sh
+++ b/scripts/ios-configure-glog.sh
@@ -1,11 +1,8 @@
 #!/bin/bash
 set -e
 
-# Only set when not running in an Xcode context
-if [ -z "$ACTION" ] || [ -z "$BUILD_DIR" ]; then
-  export CC="$(xcrun -find -sdk iphoneos cc) -arch armv7 -isysroot $(xcrun -sdk iphoneos --show-sdk-path)"
-  export CXX="$CC"
-fi
+export CC="$(xcrun -find -sdk iphoneos cc) -arch armv7 -isysroot $(xcrun -sdk iphoneos --show-sdk-path)"
+export CXX="$CC"
 
 ./configure --host arm-apple-darwin
 
--- a/scripts/ios-configure-glog.sh
+++ b/scripts/ios-configure-glog.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


janicduplessis:
cc @javache

javache:
I added this because it was causing issues with the tvos build (see the Travis failure on your diff). Maybe a better approach would be to get the right `-sdk iphoneos` flag from the environment?

janicduplessis:
@javache I see, I think getting -sdk flag from the env is the right solution I'll figure out how to get that.

janicduplessis:
@javache Ok I've figured the problem we need to get both the sdk and the arch from the build env variables (tvos is arm64 probably why it didn't work).

Tested that it build for simulator and archive on iOS and tvOS.

javache:
Awesome! Maybe we should provide defaults for these variables as well? In case people end up running the script outside of Xcode.

janicduplessis:
@javache Added default values that match what the old code did

rborn:
@janicduplessis I still have this issue. I have some pods installed and `fastlane gym` will complain with the below error while an xcode build (from the workspace of course) will go smooth

Any ideas? 😊

~~~
Undefined symbols for architecture arm64:
  "facebook::react::parseTypeFromHeader(facebook::react::BundleHeader const&)", referenced from:
      +[RCTJavaScriptLoader attemptSynchronousLoadOfBundleAtURL:runtimeBCVersion:sourceLength:error:] in libReact.a(RCTJavaScriptLoader.o)
  "_JSNoBytecodeFileFormatVersion", referenced from:
      +[RCTJavaScriptLoader loadBundleAtURL:onProgress:onComplete:] in libReact.a(RCTJavaScriptLoader.o)
      +[RCTJavaScriptLoader attemptSynchronousLoadOfBundleAtURL:runtimeBCVersion:sourceLength:error:] in libReact.a(RCTJavaScriptLoader.o)
ld: symbol(s) not found for architecture arm64

~~~

PR #14286, opened 2017-06-01
Fix Travis CI runs

javache:


javache:
These tests are still faling:

Failing tests:
	-[RNTesterIntegrationTests testAsyncStorageTest]
	-[RNTesterSnapshotTests testSliderExample]
	-[RNTesterSnapshotTests testTabBarExample]

@shergin?

shergin:
😞 Seems we have strange issue with color profiles which are different across images and machines. May be our screenshots have embedded profile, may be source images have the profile.

javache:
I've pulled in some Android fixes as well. I'm going to ship for this now, since it does fix the tvOS test.

Issue #7463, opened 2016-05-09
[XMLHttpRequest] `HEAD` request doesn't work


grabbou:
The same request works fine on iOS, on Android it fails with `TypeError: Network request failed`. It used to work few version ago and no changes were done on the API or app side. Replacing the method from HEAD to GET and also making an extra endpoint on the API side temporarily resolves the issue (worth noting that the API handler itself stays unchanged, to the logic is the same)

AFAIK it has something to do with OkHTTP since:
a) it works flawlessly on iOS
b) it never reaches the API

I am not sure if that was fixed recently in OkHttp, but I guess it's kind of a blocker until we upgrade to 3.x ?

9mm:
Yes... spent quite a while figuring out what broke. +1

DJAndries:
Any update on this issue?

satya164:
@grabbou Since we've now updated to OkHttp3, is this still an issue?

satya164:
I tried and it seems to be broken.
cc @AndrewJack

satya164:
So I did some tests, and seems the issue is on JS side,

```js
const url = 'https://hey-neighbor.s3.amazonaws.com/uploaded/content/satya164/1e03a291-acda-47d9-ba09-6898837bc48b/1/boy2.png';

fetch(url, { method: 'HEAD' }).then(() => console.warn('fetch: success')).catch(() => console.warn('fetch: error'));

const req1 = new XMLHttpRequest();
req1.open('HEAD', 'https://hey-neighbor.s3.amazonaws.com/uploaded/content/satya164/1e03a291-acda-47d9-ba09-6898837bc48b/1/boy2.png', true);
req1.onreadystatechange = () => {
  if (req1.readyState === req1.DONE) {
    if (req1.status === 200) {
      console.warn('onreadystatechange: success');
    } else {
      console.warn('onreadystatechange: error');
    }
  }
};
req1.send();

const req2 = new XMLHttpRequest();
req2.open('HEAD', url, true);
req2.onload = () => {
  if (req2.status === 300) {
    console.warn('onload: success');
  } else {
    console.warn('onload: error');
  }
};
req2.onerror = () => {
  console.warn('onload: error');
};
req2.send();
```

output:

```
fetch: error
onreadystatechange: success
onload: error
```

So it actually works when using `onreadystatechange`. Haven't investigated why, but these are my findings.

jkimbo:
So, out of curiosity I investigated this bug and think I've found the issue.

[...]

PR #14289, opened 2017-06-01
Fix `HEAD` http requests in Android

SandroMachado:
Closes https://github.com/facebook/react-native/issues/7463.

This PR fixes the `HEAD` http requests in Android. 
In Android all the HEAD http requests will fail, even if the request succeeds in the native layer due to an issue in the communication with the `OkHttp`.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -369,7 +369,19 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
               // Otherwise send the data in one big chunk, in the format that JS requested.
               String responseString = "";
               if (responseType.equals("text")) {
-                responseString = responseBody.string();
+                try {
+                  responseString = responseBody.string();
+                } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
```diff
@@ -369,7 +369,14 @@ public void onResponse(Call call, Response response) throws IOException {
               // Otherwise send the data in one big chunk, in the format that JS requested.
               String responseString = "";
               if (responseType.equals("text")) {
-                responseString = responseBody.string();
+                try {
+                  responseString = responseBody.string();
```
satya164: We should check the request method instead of making an assumption. Probably `response.request()` has the information. If not, we can store the request method according to request id in a map somewhere and check that.

SandroMachado: I think `response.request()` has that information. I will try to do that and amend the PR.

SandroMachado: @satya164 done, can you please review it again.

satya164:
@facebook-github-bot shipit

PR #14291, opened 2017-06-01
Add instructions to first start React packager

Gi11i4m:
It seems the React Packager needs to be running before the application installed by the run-android command can actually load your JS bundle. This might also be true for run-ios.

## Motivation
Incomplete documentation

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -554,6 +554,12 @@ The above command will automatically run your app on the iOS Simulator by defaul
 
 ## Running your React Native application
 
+Before running your application, the React packager must be running:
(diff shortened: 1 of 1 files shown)


hramos:
This should not be necessary. If the packager does not start automatically, then it is a bug. I'm only aware of this happening on linux systems, and it should have been fixed by #14040.

Gi11i4m:
You're right, it's a bug. I'm on Linux Mint 18.1, React-Native CLI 2.0.1. The run-android command terminates after installing the apk. It has output that says it's starting the packager but this doesn't keep running apparently.

PR #14316, opened 2017-06-04
Document Android native communication

cdlewis:
## Motivation

Native communication is documented for iOS but not Android. Typically pages in Guides (IOS) and Guides (Android) share a common structure/content but provide language specific examples.

I followed this pattern with the Android native communication documentation. It is missing some content that the iOS documentation contains, such as the layout computation flow. But provides a solid base for future efforts.

## Test Plan

Ran local instance of React Native website. Went to http://localhost:8079 and read through the new page (react-native/docs/communication-android.html). Also tested links two/from the page.

As a result of this link testing, I updated the Native Modules Android documentation to link to the Apple TV documentation, as it is now the previous page on the navigation sidebar.


hramos:
Resolved some conflicts. Can you pull these in, then update Contributing.md to that its previous link points to communication-android ?

cdlewis:
Hi @hramos sorry I hadn't noticed you commented in August!

PR #14336, opened 2017-06-06
UITapGestureRecognizer added to RCTSlider

NikolaBorislavovHristov:
Tap gesture recognizer added to iOS slider.

## Motivation

iOS slider's value can be changed by tap somewhere on the slider.
By adding a gesture recognizer for tap, it is achieved similarly user experience on both platforms. 

## Test Plan

Demo video:

ios

### DIFF
--- a/React/Views/RCTSliderManager.m
+++ b/React/Views/RCTSliderManager.m
@@ -27,6 +27,12 @@ RCT_EXPORT_MODULE()
    forControlEvents:(UIControlEventTouchUpInside |
                      UIControlEventTouchUpOutside |
                      UIControlEventTouchCancel)];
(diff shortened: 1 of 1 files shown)


NikolaBorislavovHristov:
Can you help me with the integration tests?

javache:
Hi @NikolaBorislavovHristov, thank you for this PR! I think with the default RN components we want to stay as close as possible to the default UIKit behaviour. I'd recommend releasing this as a standalone component instead.

NikolaBorislavovHristov:
"default UIKit behaviour" ???
Check the slider in the sound and display settings in any iOS device. That is a "default UIKit behaviour".

javache:
Tapping the Brightness slider in Settings.app does not move the slider for me. If it's default UIKit behaviour, it wouldn't require you to set up your own UITapGestureRecognizer to make it work.

PR #14339, opened 2017-06-06
Update default blue on iOS

xavier-villelegier:
Hey,

## Motivation

The `defaultBlue` variable is currently `#0C42FD`. We can see that it's not the same as iOS. According to the official Apple website it's supposed to be `#007AFF`.
See the `Blue` color on [this page](https://developer.apple.com/ios/human-interface-guidelines/visual-design/color/)

## Test Plan

No code added.

### DIFF
--- a/Libraries/Components/Button.js
+++ b/Libraries/Components/Button.js
@@ -136,7 +136,7 @@ class Button extends React.Component {
 let defaultBlue = '#2196F3';
 if (Platform.OS === 'ios') {
   // Measured default tintColor from iOS 10
-  defaultBlue = '#0C42FD';
+  defaultBlue = '#007AFF';
 }
 
 const styles = StyleSheet.create({


hramos:
Is it possible that the hue changed between iOS 10 and 11? The guidelines you linked to were updated yesterday with the reveal of the iOS 11 beta.

If the hue is the same as on iOS 10 I'd be happy to merge this in.

shergin:
Seems this (`#007AFF`, new one) is the default blue color since iOS 7.
https://stackoverflow.com/questions/19032940/how-can-i-get-the-ios-7-default-blue-color-programmatically

xavier-villelegier:
@shergin Yes you're right, iOS 7 was the first to introduce this blue color (`#007AFF`), since it was a complete different design in iOS 6.

| iOS version  | Style |
| ------------- | ------------- |
| 6  | image|
| 7  | image|

So yes, @hramos the hue is the same as iOS 10 👍

Issue #14348, opened 2017-06-06
v0.45.0+ - VirtualizedList slow performance and rendering


niftylettuce:
The rendering of VirtualizedList (namely `<SectionList>`) has extremely poor performance.  I will be submitting a PR that seems fixes the issue introduced by this line <https://github.com/facebook/react-native/blame/v0.45.0-rc.3/Libraries/Lists/VirtualizedList.js#L537> per this commit <https://github.com/facebook/react-native/commit/52e50af56d601e63a0c83118c194149d3aed38ea#diff-25477c6bad509193f026645fd928dec6R515>.  There are a few more instances of the change from `this._updateCellsToRenderBatcher.schedule();` to `this._scheduleCellsToRenderUpdate();`.  Therefore I believe that there is more work to be done in order to fully fix this issue.

To reproduce, simply use RN v0.45.x+ and create a `<SectionList>` with a lot of records.  The performance is horrible compared to v0.44.x.  However I need v0.45.x since it brings the scrollTo* methods on `<SectionList>` component.

cc @sahrens

niftylettuce:
So the issue is actually this function:

```js
  _scheduleCellsToRenderUpdate() {
    this._updateCellsToRenderBatcher.schedule();
    return;
    const {first, last} = this.state;
    const {offset, visibleLength, velocity} = this._scrollMetrics;
    const itemCount = this.props.getItemCount(this.props.data);
    let hiPri = false;
    if (first > 0 || last < itemCount - 1) {
      const distTop = offset - this._getFrameMetricsApprox(first).offset;
      const distBottom = this._getFrameMetricsApprox(last).offset - (offset + visibleLength);
      const scrollingThreshold = this.props.onEndReachedThreshold * visibleLength / 2;
      hiPri = (
        Math.min(distTop, distBottom) < 0 ||
        (velocity < -2 && distTop < scrollingThreshold) ||
        (velocity > 2 && distBottom < scrollingThreshold)
      );
    }
    if (hiPri) {
      // Don't worry about interactions when scrolling quickly; focus on filling content as fast
      // as possible.
      this._updateCellsToRenderBatcher.dispose({abort: true});
      this._updateCellsToRender();
      return;
    } else {
      this._updateCellsToRenderBatcher.schedule();
    }
  }
```

niftylettuce:
Specifically on first load, this function has `hiPri` true, which delays the rendering until it finishes iterating over all records.  If we have 1000+ records, it's incredibly slow.

```
[...]

PR #14349, opened 2017-06-06
Revert schedule() function for rendering cells per #14348

niftylettuce:
See #14348

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -240,11 +240,11 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
 
   /**
    * Scroll to a specific content pixel offset in the list.
-   * 
+   *
    * Param `offset` expects the offset to scroll to.
    * In case of `horizontal` is true, the offset is the x-value,
    * in any other case the offset is the y-value.
-   * 
+   *
    * Param `animated` (`true` by default) defines whether the list
    * should do an animation while scrolling.
    */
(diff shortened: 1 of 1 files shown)


niftylettuce:
This isn't valid

sahrens:
The only extra work it does is a little bit of math to determine if it should be a hi-pri update or not. Or are you seeing that the updates are hi-pri more often than they should be?

niftylettuce:
@sahrens hey I sent you a message on Slack! (Expo)

sahrens:
I'm not on the expo channel - can you invite me or use messenger or something?

sahrens:
Or just discuss here?

niftylettuce:
@sahrens So ignore the PR - the poor performance was due to my list having components that were not PureComponent's.  I was using the [NativeBase library](http://nativebase.io) and it was causing rendering issues/slow performance.  I switched to rendering with pure <View> and <Text> and it works great now.

sahrens:
Awesome, glad to hear it.

tjmadsen:
@niftylettuce I'm facing the same issue (very slow initial load).  Can you explain a bit more on what you had to do to fix this, or point me int he right direction?  It's slightly faster once I changed my renderItem component to be a PureComponent, but still slower to load than ListView was.  Do you know what is causing the slow load performance?

niftylettuce:
@tjmadsen I think the only approach for fast scrolling lists on React Native right now is to use FlatList in combination with React.PureComponent, and to not really have many views or components going on at once.  In general though, React Native does not really support fast and long scrolling lists.  It's very slow and limited with what you can do.  I have considered writing a pure iOS and Android component instead.

tjmadsen:
@niftylettuce FYI - just implemented this fix and it made it WAY faster for me (though still not perfect)
https://github.com/facebook/react-native/commit/ebcf5fd24173c2628a5dcc65871fa7a8b219f96e

Issue #14275, opened 2017-05-31
keyboardWillShow does not fire on Android


brentvatne:
As per https://github.com/facebook/react-native/issues/10613#issuecomment-300121457

This is easy to verify by searching the repo and verifying that the `keyboardWillShow` string does not exist in the React Native Android codebase: https://github.com/facebook/react-native/search?utf8=%E2%9C%93&q=keyboardWillShow&type=

satya164:
AFAIK we meed to listen to root view size changes to detect when keyboard is shown/hidden. So I am not sure if `willX` events can even be implemented.

ptomasroos:
I can only add to whats @satya164 saying. As with my knowledge in this as well there is no way to actually discover when a soft keyboard will happen. May I ask in what use-case this would be needed? I know people are using will* to make transitions "faster" on IOS.

brentvatne:
@ptomasroos - I think it's less about making them "faster" and instead making them more responsive, so it can animate while the keyboard is opening rather than waiting until it has opened. I don't have a personal use case but just copied this over from #10613.

If anyone reads this and knows if it's possible to hook into a notification for the soft keyboard beginning to open on Android, let us know here!

janicduplessis:
Yea keyboard handling is really bad on Android, there is no event triggered so the way used to detect when it opens is to use a layout listener on the window and check if the height changed. Doing this means there is only one event triggered, the only thing we could do is trigger both `keyboardWillShow` and `keyboardDidShow` at the same time so the behavior is closer to iOS but that may not be really useful...

brentvatne:
It sounds like the best thing to do is clearly document this limitation then. Any takers?

DukeNukem1990:
Just opened a PR for this. Its my very first so if I did anything wrong please let me know.

reznord:
If this is still open, I'd like to work on this.

evanjmg:
I've added 'adjustNothing' to AndroidManifest with the latest react native - 0.51.0-rc.2-  this does not work -

PR #14350, opened 2017-06-06
Keyboard will show documentation

DukeNukem1990:
Clarified documentation for keyboardWillShow on Android as seen in issue #14275 

## Motivation (required)

Clarifies documentation of keyboardWillShow

## Test Plan (required)

None as it is just a comment that changed.


Libraries/Components/Keyboard/Keyboard.js
```diff
@@ -101,6 +101,10 @@ let Keyboard = {
    *
    * Note that if you set `android:windowSoftInputMode` to `adjustResize`  or `adjustNothing`,
    * only `keyboardDidShow` and `keyboardDidHide` events will available on Android.
+   * 
+   * Limitation for keyboardWillShow: On Android no event is fired when the keyboard is about
+   * to be opened. Instead you can use a layout listener to check if the height of the layout
```
brentvatne: I think we can delete this new text and add clarification on to the end of the previous sentence, something like this but maybe better wording:

> only `keyboardDidShow` and `keyboardDidHide` events will available on Android because it is not possible to listen for `keyboardWillShow` or `keyboardWillHide` events in the underlying Android APIs.

shergin: Can we workaround it on native side (use another event)? Or it is just conceptually impossible/meaningless.

DukeNukem1990: [This StackOverflow page](https://stackoverflow.com/questions/25216749/softkeyboard-open-and-close-listener-in-an-activity-in-android)  and several others suggest that there is indeed no way on Android to do this.

Libraries/Components/Keyboard/Keyboard.js
```diff
@@ -100,7 +100,9 @@ let Keyboard = {
    * - `keyboardDidChangeFrame`
    *
    * Note that if you set `android:windowSoftInputMode` to `adjustResize`  or `adjustNothing`,
-   * only `keyboardDidShow` and `keyboardDidHide` events will available on Android.
+   * only `keyboardDidShow` and `keyboardDidHide` events will be available on Android and
+   * `keyboardWillShow` as well as `keyboardWillHide` are generally not available on Android
+   * since there is no corresponding event.
```
shergin: The relationship between `android:windowSoftInputMode`, [`keyboardDidShow` + `keyboardDidHide`], and  [`keyboardWillShow` + `keyboardWillHide`] is still unclear to me. Does  `android:windowSoftInputMode` affect  [`keyboardWillShow` + `keyboardWillHide`], or this pair of events generally unavailable on Android?

DukeNukem1990: `keyboardWillShow` and `keyboardWillHide` are not available on Android. There is just no event for that. (As of now, might of course change in the future (hopefully)).

DukeNukem1990: Is there anything more you like to know?

PR #14355, opened 2017-06-07
add breaking test, fix edge case for replacing browser "main" field

mvayngrib:
## Motivation (required)

fixes edge case for replacing browser "main" field, when "main" field value is a file name rather than a file path

## Test Plan (required)

added breaking test in 1st commit, fix in 2nd commit


skevy:
The packager no longer lives in this repo. You should send a PR to https://github.com/facebook/metro-bundler

mvayngrib:
@skevy thanks, will do

PR #14358, opened 2017-06-07
Android - Update Image.getSize to use encoded image

orenklein:
Regarding issue #14357

Manually tested it on remote URLs of png, jpeg and bmp formats.

Additional notes:
1. It is possible to add support for static images in the future.
2. parseMetaData function is also extracting image format which is unnecessary - thus, it can be replaced with the code from parseMetaData that only gets width & height.

(Might be low-level breaking change (?) - Changing from fetchDecodedImage to fetchEncodedImage probably means that the decoded image will not be cached as it used to be [apparently, regarding fresco documentation, the encoded image will be cached])

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
@@ -21,7 +21,8 @@ import com.facebook.datasource.DataSource;
 import com.facebook.datasource.DataSubscriber;
 import com.facebook.drawee.backends.pipeline.Fresco;
 import com.facebook.imagepipeline.core.ImagePipeline;
-import com.facebook.imagepipeline.image.CloseableImage;
+import com.facebook.imagepipeline.image.EncodedImage;
+import com.facebook.imagepipeline.memory.PooledByteBuffer;
 import com.facebook.imagepipeline.request.ImageRequest;
 import com.facebook.imagepipeline.request.ImageRequestBuilder;
 import com.facebook.react.bridge.Arguments;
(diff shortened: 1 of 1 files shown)


javache:
cc @foghina, could you help review this? :)

ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
```diff
@@ -81,22 +82,23 @@ public void getSize(
...
         protected void onNewResultImpl(
-            DataSource<CloseableReference<CloseableImage>> dataSource) {
+            DataSource<CloseableReference<PooledByteBuffer>> dataSource) {
           if (!dataSource.isFinished()) {
             return;
           }
-          CloseableReference<CloseableImage> ref = dataSource.getResult();
+          CloseableReference<PooledByteBuffer> ref = dataSource.getResult();
           if (ref != null) {
+            EncodedImage image = ref.get();
```
foghina: Hmm, how come this doesn't warn about trying to assign a `PooledByteBuffer` to an `EncodedImage`? Shouldn't `ref` and `dataSource` use `EncodedImage` as the type arg?

orenklein: My bad, you're right.

orenklein:
@foghina I pushed a commit to fix that error :)

ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
```diff
@@ -81,22 +82,23 @@ public void getSize(
     Uri uri = Uri.parse(uriString);
     ImageRequest request = ImageRequestBuilder.newBuilderWithSource(uri).build();
 
-    DataSource<CloseableReference<CloseableImage>> dataSource =
-      Fresco.getImagePipeline().fetchDecodedImage(request, mCallerContext);
+    DataSource<CloseableReference<PooledByteBuffer>> dataSource =
+      Fresco.getImagePipeline().fetchEncodedImage(request, mCallerContext);
```
bood: Fresco only supports limited source types for encoded image:
https://github.com/facebook/fresco/blob/v1.3.0/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java#L169

For example: `SOURCE_TYPE_LOCAL_CONTENT` (content://) which is returned from `RollCamera` is not supported

Use `ContentResolver.openInputStream` together with  [BitmapUtil.decodeDimensions(...)](https://github.com/facebook/fresco/blob/df2854a76786f58e044a71617309cc3c0a0c358c/imagepipeline-base/src/main/java/com/facebook/imageutils/BitmapUtil.java#L91) for those not supported types.

Actually `BitmapUtil.decodeDimensions` probably has better performance than loading the fully encoded image since it only needs to decode the metadata. So I suggest use that when possible.

bood: @orenklein @foghina 
See my approach at https://github.com/bood/react-native/commit/ea4b4dd8b6dbcf725ef34ae75c2ce36861229bfe

I'll make another PR if it looks good.

orenklein:
@bood You should open a new PR and reference this one.

bood:
@orenklein PR #15887 created

PR #14370, opened 2017-06-07
Refactor PropTypes to use `prop-types` package

JonnyBurger:
This PR switches out all occurences of `React.PropTypes` with `PropTypes` of `prop-types`.

## Motivation (required)

To remove the warnings caused by React 16.
`React.PropTypes` is deprecated and should be migrated to the `prop-types` package.

## Test Plan (required)

No more tests are failing than on the master branch.
The prop types package should have an equal API with `React.PropTypes`, with only two exceptions:
- Validator functions cannot be called, which there seem to be none.
- `React.checkPropTypes`  is now `PropTypes.checkPropTypes`. Two occurences were updated.


Libraries/CameraRoll/CameraRoll.js
```diff
@@ -12,7 +12,7 @@
  */
 'use strict';
 
-const {PropTypes, checkPropTypes} = require('React');
+const PropTypes = require('prop-types');
```
chirag04: nit: maybe `const { checkPropTypes } = require('prop-types');`

chirag04:
looks good. @hramos

Libraries/CameraRoll/CameraRoll.js
JonnyBurger: `PropTypes` is also used in the file.

This would be possible:
```js
const PropTypes = require('React');
const {checkPropTypes} = PropTypes;
```

chirag04: nevermind then 👍

javache:
I think all of the remaining callsites have been fixed in https://github.com/facebook/react-native/commit/50b11aa09b479e5ce8b06de98e02ba1dd6a8cca3

PR #14377, opened 2017-06-07
Tvos cocoapods support

navratill:

## Motivation (required)

Making work **cocoapods** react-native dependecy on tvOS. This patch excludes files, which are not supported by tvOS, adding support of deployment platform and make it possible to use cocoapods integration into tvOS project. 

## Test Plan (required)

No tests are needed, there are no changes in code. Changes are made just in podpsecs (React.podspec and Yoga.podspec).

Working of this code can be test by adding cocoapods integration into tvOS project, where tvOS subspec has to be added as well. After "pod install" it is possible to build react-native in tvOS project.

Everything works and was tested on 0.45-stable

### DIFF
--- a/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h
+++ b/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h
@@ -9,9 +9,10 @@
 
 #import <Foundation/Foundation.h>
 
-#import <RCTAnimation/RCTValueAnimatedNode.h>
 #import <React/RCTBridgeModule.h>
 #import <React/RCTUIManager.h>
+
+ #import "RCTValueAnimatedNode.h"
 
 @interface RCTNativeAnimatedNodesManager : NSObject
 
--- a/React.podspec
+++ b/React.podspec
@@ -36,7 +36,8 @@ Pod::Spec.new do |s|
   s.source                  = source
   s.default_subspec         = "Core"
   s.requires_arc            = true
-  s.platform                = :ios, "8.0"
+  s.ios.deployment_target   = "8.0"
+  s.tvos.deployment_target  = "9.0"
   s.pod_target_xcconfig     = { "CLANG_CXX_LANGUAGE_STANDARD" => "c++14" }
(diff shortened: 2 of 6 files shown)


shergin:
Ooops, looks like your PR contains additional unnecessary changes.

navratill:
I rebase commits and update import of RCTValueAnimatedNode to local one, that it can be integrated into cocoapods as well.

shergin:
Right now your diff contains changes in 81 (!) files. You have to rebase your `navratill:tvos-cocoapods-support` on `facebook:master`.
Please do this, and then your PR have good changes to be merged.

navratill:
@shergin Done, thank you for your help, I missed that I went from stable branch.

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h
```diff
@@ -9,10 +9,11 @@
 
 #import <Foundation/Foundation.h>
 
-#import <RCTAnimation/RCTValueAnimatedNode.h>
 #import <React/RCTBridgeModule.h>
 #import <React/RCTUIManager.h>
 
+ #import "RCTValueAnimatedNode.h"
```
shergin: Why do we need this?

navratill: It was crashing cocoapods build, because only module visible is <React/...>. 

Also it is local dependency and it should be imported like that. The best solution would be completely get rid of this import and replace it by forward declaration of RCTValueAnimatedNodeObserver (which is actually need it in this file). Should I do it?

navratill: @protocol RCTValueAnimatedNodeObserver;

shergin:
@navratill 
Let's try to land the change in podspec file only for now?

shergin:
@dlowder-salesforce Are you okay with these podspec changes?

dlowder-salesforce:
Some of the podspec changes look like ones that I already committed in #15065

PR #14393, opened 2017-06-08
Fix deltatime calculation.

mikelambert:
This is necessary to fix an extraneous warning when a VirtualizedList is constructed on the first frame.

On the first frame, the dt is huge (ie, time since epoch).
On the second frame (which may legitimately be slow as a result of a lot to render), it will then assume there were two consecutive slow frames, and print a warning:
  "VirtualizedList: You have a large list that is slow to update..."

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -734,7 +734,9 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
     const visibleLength = this._selectLength(e.nativeEvent.layoutMeasurement);
     const contentLength = this._selectLength(e.nativeEvent.contentSize);
     const offset = this._selectOffset(e.nativeEvent.contentOffset);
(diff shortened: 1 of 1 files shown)


Libraries/Lists/VirtualizedList.js
```diff
@@ -734,7 +734,9 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
     const visibleLength = this._selectLength(e.nativeEvent.layoutMeasurement);
     const contentLength = this._selectLength(e.nativeEvent.contentSize);
     const offset = this._selectOffset(e.nativeEvent.contentOffset);
-    const dt = Math.max(1, timestamp - this._scrollMetrics.timestamp);
+    const dt = this._scrollMetrics.timestamp
+      ? Math.max(1, timestamp - this._scrollMetrics.timestamp)
+      : 0;
```
sahrens: Let's do 1 here as well so we don't risk division by zero.

mikelambert: Fixed and repushed.

PR #14408, opened 2017-06-08
Add a type cast to RCT_ENUM_CONVERTER for C++ compatibility

jballer:
## Motivation

C++ doesn't provide an implicit cast to an enum value from the enum's backing type. When a `.mm` file calls `RCT_ENUM_CONVERTER`, we end up with the following compiler error:
> Error: cannot initialize return object of type `<TypeName>` with an rvalue of type `NSInteger`

Since `RCT_ENUM_CONVERTER` is a macro, this error will appear whenever we try to expand the macro in a C++ context.

## Test Plan

The project compiles and runs as expected when this additional cast is added 😃

### DIFF
--- a/React/Base/RCTConvert.h
+++ b/React/Base/RCTConvert.h
@@ -218,7 +218,7 @@ RCT_CUSTOM_CONVERTER(type, type, [RCT_DEBUG ? [self NSNumber:json] : json getter
   dispatch_once(&onceToken, ^{                            \
     mapping = values;                                     \
   });                                                     \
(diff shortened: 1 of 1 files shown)


javache:
What type are you trying to convert? [NSNumber getter] should already be giving you the right type.

jballer:
The C++ compiler expects the enum type ("TypeName" in `typedef NS_ENUM(NSInteger, TypeName)…`).

The getter (in this case, `integerValue`) is properly returning the enum's `NSInteger` backing type. In standard ObjC this will compile without issue, but C++ requires an explicit cast from `NSInteger` back to the enum type.

Issue #14246, opened 2017-05-29
Cannot find entry file index.ios.js [index.android.js] in any of the roots


MyTotoro:
[edit: [scroll down for additional details](https://github.com/facebook/react-native/issues/14246#issuecomment-307205124) - @hramos]

### Description

Recently upgrade to RN 0.45 from 0.42 (I was hit with  #14225 and I figured I might as well spend the time migrating) 

Error in Bundler:

```
error: bundling: Error
    at DependencyGraph._getAbsolutePath (/Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/node-haste/DependencyGraph.js:280:11)
    at DependencyGraph.getDependencies (/Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/node-haste/DependencyGraph.js:218:26)
    at Resolver.getDependencies (/Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/Resolver/index.js:107:27)
    at /Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/Bundler/index.js:591:37
    at Generator.next (<anonymous>)
    at step (/Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/Bundler/index.js:12:445)
    at /Users/mytotoro/Desktop/app/node_modules/react-native/packager/src/Bundler/index.js:12:605
    at process._tickCallback (internal/process/next_tick.js:109:7)
Bundling `index.android.js`  0.0% (0/1), failed.
```
and red screen showing

`Response error 404, Cannot find entry file index.android.js in any of the roots: [\"/Users/mytotoro/desktop/app/node_modules/react-native/packager\"]`

With NotFoundError

```
Looking for JS files in
   /Users/mytotoro/Desktop/app/node_modules/react-native/packager 


React packager ready.
```

### Reproduction Steps and Sample Code

```
"react": "^16.0.0-alpha.12",
"react-native": "^0.45.0-rc.2",

Mac: 10.12.4
node: v7.10.0
npm: 4.6.1
```

I run `react-native run-android` from the root directory of my project as I have always been doing:

```
total 456
-rw-r--r--    1 mytotoro  staff     175 Apr  7 23:36 index.ios.js
-rw-r--r--    1 mytotoro  staff     175 Apr  7 23:36 index.android.js
drwxr-xr-x    7 mytotoro  staff     238 Apr  7 23:36 Tests
[...]

PR #14412, opened 2017-06-09
[packager] Fix broken default getProjectRoots

skevy:
In <= 0.44, the default implementation of getProjectRoots() came from `local-cli/core/default.config.js`. With changes happening in the CLI and the packager over the course of the last two months, various pieces of this logic (specifically `local-cli/utils/Config.js`) were rewritten, and though default.config.js was still being imported and used in `local-cli/core/index.js`, the default `getProjectRoots()` was being overriden by the defaults specified in `local-cli/utils/Config.js`.

This PR moves the logic from default.config.js into Config.js and index.js, as appropriate. Specifically:

- The `getProjectCommands()`, `getProjectConfig()`, and `getDependencyConfig()` methods, which have traditionally not been part of the rn-cli.config.js spec, are now defined in `local-cli/core/index.js`.
- The `getProjectRoots()` method, which contained logic for properly resolving the _actual_ project root as well as resolving symlinks within that root, has been moved to `local-cli/utils/Config.js`, to match the fact that other default  rn-cli.config.js definitions live there.


skevy:
Fixes https://github.com/facebook/react-native/issues/14246 in master.

See https://github.com/facebook/react-native/pull/14413 for 0.45 backport.

cc @hramos

davidaurelio:
Thanks for fixing, @skevy!

PR #14422, opened 2017-06-09
add node/npm version to ISSUE_TEMPLATE.md

tabrindle:
## Motivation
A huge number of Issues are opened with some key information about node/npm versions missing. Adding this information as a required part of the ISSUE_TEMPLATE may help issues get closed faster.   

## Test Plan
Once merged in, check ISSUE_TEMPLATE by opening new issue, verify new line is there.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -29,24 +29,40 @@
 ### Environment
 
 <!--
-  Please fill in all the relevant fields by running these commands in terminal.
+  If you are on v0.48.0 or greater, please run `react-native info` and paste its contents here then specify target platform, like this:
+
+  Environment:
+    OS:  macOS Sierra 10.12.6
+    Node:  8.4.0
+    Yarn:  0.27.5
+    npm:  5.4.0
+    Watchman:  4.7.0
+    Xcode:  Xcode 8.3.3 Build version 8E3004b
+    Android Studio:  2.3 AI-162.4069837
+
+  Packages: (wanted => installed)
+    react-native: 0.48.1 => 0.48.1
+    react: 16.0.0-alpha.12 => 16.0.0-alpha.12
+
+  Target Platform: iOS (10.3)
+
+  Otherwise, fill in the following relevant fields
 -->
 
+<!-- Run these commands in terminal. -->
 1. `react-native -v`:
 2. `node -v`:
 3. `npm -v`:
 4. `yarn --version`<!-- (if you use Yarn) -->:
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -39,6 +39,7 @@ What needs to be done to address this issue? Ideally, provide a pull request wit
 ### Additional Information
 
 * React Native version: [FILL THIS OUT: Be specific, filling out "latest" here is not enough.]
+* Node/npm version: [FILL THIS OUT: `node -v` and `npm -v`]
```
hramos: Can you add yarn here? If yarn is installed, it will be used instead of npm.

tabrindle: Can do - also this may depend on https://github.com/facebook/react-native/pull/14428

We may want to wait until this one is finished.

tabrindle: Added yarn next to npm, gave node its own line. Also reordered and added hint about `react-native info` to coincide with https://github.com/facebook/react-native/pull/14428

hramos:
We can merge this once `react-native info` is available in the CLI. I recently updated the issue template with specific commands you have to run, but it would be nice to just have people run `info` instead!

tabrindle:
@hramos considering the changes in the mean time, might it be better to close this issue, rebase my other PR and include this in that one?

tabrindle:
Hey @hramos I can either update this PR or submit a completely new one based on the changes I've made with react native info and your template. Which do you prefer?

hramos:
Updating this one seems fine.

tabrindle:
@hramos - updated, ready for review.

.github/ISSUE_TEMPLATE.md
```diff
@@ -29,24 +29,40 @@
 ### Environment
 
 <!--
-  Please fill in all the relevant fields by running these commands in terminal.
+  If you are on v0.48.0 or greater, please run `react-native info` and paste its contents here then specify target platform, like this:
```
hramos: We don't want to encourage people to open issues for old releases, and 0.48 is already the current stable. Can you rewrite this a bit so that it assumes people are using 0.48 or newer? Just have them run `react-native info`, and maybe add a note saying that if they get (insert whatever error is shown if they're not using 0.48+) then they should make sure they're using a recent version of React Native.

That way we can remove the whole next section that asks people to run several commands.

tabrindle: Sounds good to me. Will update.

PR #14443, opened 2017-06-10
Fixed crash because of spread operator

moppymopperson:
The spread operator was not playing nicely with `StyleSheet` objects and causing a crash. This fixes the issue. Figured out the fix after reading [this issue](https://github.com/react-native-community/react-native-blur/issues/43).

## Motivation 

Makes the code in the Animation tutorial work properly

## Test Plan

This is a trivial change to the documentation. Testing likely not required.

### DIFF
--- a/docs/Animations.md
+++ b/docs/Animations.md
@@ -49,11 +49,13 @@ class FadeInView extends React.Component {
     let { fadeAnim } = this.state;
 
     return (
-      <Animated.View                 // Special animatable View
-        style={{
-          ...this.props.style,
-          opacity: fadeAnim,         // Bind opacity to animated value
-        }}
+      <Animated.View // Special animatable View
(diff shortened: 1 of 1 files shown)


docs/Animations.md
```diff
@@ -49,11 +49,13 @@ class FadeInView extends React.Component {
...
-        style={{
-          ...this.props.style,
-          opacity: fadeAnim,         // Bind opacity to animated value
-        }}
+      <Animated.View // Special animatable View
+        style={[
+          {
+            opacity: fadeAnim // Bind opacity to animated value
+          },
+          ...this.props.style
```
chirag04: you can remove the spread if you are passing an array to style. 

this.props.style can be a number so spreading can still crash. passing array takes care of this.

PR #14455, opened 2017-06-11
Clearer code in Button.js

dpgao:

## Motivation (required)

* Combined repeating `if` checks for clearer logic.
* `defaultBlue` is inlined because it is only used once for each OS.

## Test Plan (required)

- Updated the `Button.js` file in an existing project and all buttons behave exactly the same. Here is a simple demo:
  

- There is no change in the underlying logic of the code.


dpgao:
@ericvicenti Thanks for the review. My personal opinion is that if we are to export the color of the button we should do it using the existing `color` prop instead of creating a new variable. And we can work on that later.

PR #14470, opened 2017-06-12
Use <ImageBackground> instead of <Image>

sunnylqm:
https://github.com/facebook/react-native/commit/9637dd4a1b78d5f6da510b1b4ad10d45d67badbf


shergin:
@sunnylqm Thank you for pointing on this! 
This corrections are legit but I would move the point a little bit. I would suggest to implement this behaviour manually for each particular case and highlight that fact that actual implementation of `<ImageBackground>` is super simple. And I would advice to use `<ImageBackground>` only after taking a look of the the source code to make the best choice suited each particular case.

hramos:

PR #14477, opened 2017-06-13
fix init

### LINKED ISSUE #14476: [0.46.0-rc.2] react-native init fails
Running `react-native init rn46 --version "0.46.0-rc.2"` fails to create a new project, with exception:

```
npm WARN react-native@0.46.0-rc.2 requires a peer of react@16.0.0-alpha.12 but none was installed.
module.js:472
    throw err;
    ^

Error: Cannot find module 'metro-bundler/build/lib/TerminalClass'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/rn46/node_modules/react-native/local-cli/server/runServer.js:18:18)
    at Module._compile (module.js:571:32)
    at loader (/rn46/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/rn46/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
```

The generated package.json does not include react:
```
  "dependencies": {
    "react-native": "0.46.0-rc.2"
  }
```

I tried this with node version `v7.9`, `v7.10`, `v8.1`, and yarn `v0.24.6`, all giving the same exception.

nihgwu:
Close #14476 
`TerminalClass` was renamed to `Terminal` by https://github.com/facebook/metro-bundler/commit/c1ff8b5b81f53d359abee0696ba9468d4fae9b85, this sync the changes

### DIFF
--- a/local-cli/server/runServer.js
+++ b/local-cli/server/runServer.js
@@ -15,7 +15,7 @@
 require('../../setupBabel')();
 const InspectorProxy = require('./util/inspectorProxy.js');
 const ReactPackager = require('metro-bundler');
-const Terminal = require('metro-bundler/build/lib/TerminalClass');
+const Terminal = require('metro-bundler/build/lib/Terminal');
 
 const attachHMRServer = require('./util/attachHMRServer');
 const connect = require('connect');
--- a/local-cli/bundle/buildBundle.js
+++ b/local-cli/bundle/buildBundle.js
@@ -13,7 +13,7 @@
 
 const log = require('../util/log').out('bundle');
 const Server = require('metro-bundler/build/Server');
-const Terminal = require('metro-bundler/build/lib/TerminalClass');
+const Terminal = require('metro-bundler/build/lib/Terminal');
 const TerminalReporter = require('metro-bundler/build/lib/TerminalReporter');
 const TransformCaching = require('metro-bundler/build/lib/TransformCaching');
 


nihgwu:
ping @jeanlauliac , and I think this should be cherry picked into 0.46-stable as `metro-bundler` is separated to it's own repo in this branch

local-cli/server/runServer.js
```diff
@@ -15,7 +15,7 @@
 require('../../setupBabel')();
 const InspectorProxy = require('./util/inspectorProxy.js');
 const ReactPackager = require('metro-bundler');
-const Terminal = require('metro-bundler/build/lib/TerminalClass');
+const Terminal = require('metro-bundler/build/lib/Terminal');
```
jeanlauliac: How come Flow is complaining here? Doesn't that mean the project is still using the previous version of `metro-bundler` where it is still named `TerminalClass`? In which case we cannot switch it over just yet?

jeanlauliac:
Alright, I understand what happens now. cc @cpojer  The problem is that we released a new version of `metro-bundler` as a revision bump, 0.7.5. Unfortunately this contained breaking change, so it should have been a minor bump, 0.8.0. Without it, `react-native` is now picking up the new revision automatically because it uses a caret specifier for `metro-bundler`'s version. To add to the problem, we cannot land this PR, because I believe it'd break everyone using `react-native` with a locked dependency tree (ex. `yarn.lock`), because they'd pick up the `TerminalClass` version of `metro-bundler`, in which case this changeset would break react native for them.

The solution, I believe, is to unpublish 0.7.5 and republish it as 0.8.0. If unpublishing is prohibited per npm rules, we need to do a branch reverting back to the `TerminalClass` name, and publish as 0.7.6, or override 0.7.5 if possible.

nihgwu:
@jeanlauliac yes that's a issue introduced by a wrong semver change, I think the best way is to release a new minor version *0.7.6*

nihgwu:
as for the `yarn.lock` stuff, I think it's not a big problem as it only effects those who are using RN0.46beta, which is assumed with breaking changes

jeanlauliac:
Yeah, publishing a new version.

nihgwu:
close as it has been resolved by the packager update

PR #14486, opened 2017-06-13
ImageBackground: use cached styles

doochik:
Replace object with cached styles from StyleSheet

## Motivation (required)

Create new style object may cause performance problems

## Test Plan (required)

Check that ImageBackground position works the same way.


shergin:
Привет, @doochik! 😄 
That's sweet! Could you please resolve conflicts?

doochik:
@shergin i've fixed it

PR #14495, opened 2017-06-13
Fixed navigation template

### LINKED ISSUE #14313: navigation template is not working anymore
App generated by cli 'react-native init --template navigation' is not working anymore. The template should be upgraded.

agrcrobles:

## Motivation (required)

It fixes #14313

### DIFF
--- a/local-cli/templates/HelloNavigation/App.js
+++ b/local-cli/templates/HelloNavigation/App.js
@@ -1,5 +1,3 @@
-'use strict';
-
 /**
  * This is an example React Native app demonstrates ListViews, text input and
  * navigation between a few screens.
--- a/local-cli/templates/HelloNavigation/README.md
+++ b/local-cli/templates/HelloNavigation/README.md
@@ -1,6 +1,6 @@
 # App template for new React Native apps
 
-This is a simple React Native app template which demonstrates a few basics concepts such as navigation between a few screens, ListViews, and handling text input.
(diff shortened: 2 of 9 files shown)


agrcrobles:
cc @skevy

agrcrobles:
This PR fixes navigation template that it works as an example of how using a stack navigator and tab navigator nested.

In ios the tab navigator goes to the bottom by default but in android it places to the top.

In android there is need to hide stack header for scenes `ChatListScreen.js` and `WelcomeScreen.js` in order to avoid having the tab pager + the header stack together.

But I couldn't hide it by defining it in the routes because when navigating to the chat scene, stack header should be shown.

I am not sure about which is the best way to do it, [undefined vs null](https://github.com/facebook/react-native/pull/14495/files#diff-fba519b215ed8fe87c76726f0c7ad7d6R16) works, but I appreciate any thoughts or suggest to make it better.

Also removed 'use strict' since ES6 modules are always in strict mode, and moved from ListView to FlatList after reading the blog post:
https://facebook.github.io/react-native/blog/2017/03/13/better-list-views.html

agrcrobles:
hey guys, navigation template is not working as it is now and I think this PR fixes it ( but needs revision )

do you think does it worth to maintain it or should this template be deprecated?

@hramos @mkonicek

hramos:
@agrcrobles I think it's worth fixing up. I hope to write a blog post on templates in the next month or so, and we'll need some sample templates for people to use, including this one. This reminds me I need to update https://github.com/hramos/react-native-template-master-detail prior to that.

PR #14524, opened 2017-06-15
Allow applicationId to be specified on build

shockdesign:
Added the ability to define a specific applicationId (in case it's different to the package name)

<details>

</details>

## Motivation (required)

We have a specific requirement that our package name differs from the applicationId but not in a way that allows us to add it as a suffix to our package name. I wanted to add the option to specify a specific applicationId.

## Test Plan (required)

`react-native run-android --appId <Specific Application Id>`


shockdesign:
@facebook-github-bot It looks like Travis CI is failing on the iOS tests which are timing out. Note, this change has nothing to do, nor should it affect the iOS builds.

henrikra:
So basically what you mean is that when running `react-native run-android` react native assumes that you `applicationId` inside `build.gradle` is the same as `package` in `AndroidManifest.xml` right?

So with this pr I could run `react-native run-android --appId com.company.example` as an example?

If yes this pr is great :)

One suggestion is that would it make sense that the script would read the `applicationId` straight from `build.gradle` since that is what it has to match :)

shockdesign:
@henrikra Correct, only really useful if your applicationId is different from your package name.

I did look into that but it would require a lot more smarts especially if you're running different flavors as it would have to know which one to pick. This seemed like a quick and easy approach to solve the problem.

henrikra:
I think this should be merged then :)

local-cli/runAndroid/runAndroid.js
```diff
@@ -90,7 +90,7 @@ function buildAndRun(args) {
       'utf8'
     ).match(/package="(.+?)"/)[1];
 
-  const packageNameWithSuffix = args.appIdSuffix ? packageName + '.' + args.appIdSuffix : packageName;
+  const packageNameWithSuffix = args.appId ? args.appId : args.appIdSuffix ? packageName + '.' + args.appIdSuffix : packageName;
```
shergin: Could you please make it more clear?

henrikra: Yeah maybe function like `getPackageNameWithSuffix({appId, appIdSuffix, packageName})`

henrikra: And maybe use template literals liks `´${packageName}.${args.appIdSuffix}´`

shergin: I am okay with anything which makes the code looks obvious. Even adding parentheses and indentations for original solution is enough.

shockdesign:
@shergin Just let me know if the changes are sufficient?

shockdesign:
@henrikra / @shergin - anything else we can do to progress this along?

henrikra:
For me it is okay 👍

PR #14559, opened 2017-06-16
[VirtualList] Fix off-by-one errors when rendering window-sized items.

mikelambert:
I'm using a `<FlatList horizontal windowSize={1} initialNumToRender={1} initialScrollIndex={1} pagingEnabled />` (where each element is exactly the width of the screen).

The existing calculation would always detect a window size of 2(!). Namely because checking [375, 375, 750, 750] would cause it to match the first element (of 0, 375) and second element (of 375, 750).

This causes the elements to be sized as (0, 374) and (375, 749), and in turn causes our search window to be [375, 375, 749, 749], which in turn would give a window size of exactly one, as expected.

In the case of natural scroll lists, this is essentially irrelevant. In the case of pagingEnabled lists, this can result in an extra frame being rendered, and since I'm dealing with performance issues, I'm trying to minimize over-draw (at least on initial render), and want to be sure it *only* prioritizes the one element on the screen.

I've verified (with my FlatList designed as above), that the window size is now only one-element large (as expected), and only has two elements when I am dragging the FlatList and it is showing two elements on screen simultaneously. I'm not sure if that's sufficient of a test plan here?


mikelambert:
Ahh yay, there's a jest test file for this one. Added a test for what I consider a new common case (ie 199), (and "fixed" the broken one to reflect the new behavior.)

mikelambert:
Merged to head. @sahrens should I be tagging someone else on this review?

sahrens:
@facebook-github-bot shipit

mikelambert:
Ping @sahrens  ?

mikelambert:
...ping? :(

hramos:
Jest is failing -- might be a bad conflict resolution fix on my part.

hramos:
It seems like I never got around to looking into the Jest failure here. If someone wants to take a look at it, I'd be happy to import once it passes CI.

cpojer:
@mikelambert it seems that all this time the reason this wasn't landed is because it actually has a legitimate test failure. If you rebase and run Jest on the react-native repo (yarn test), you'll see that one test is actually failing after your change.

If you can fix that, we are happy to land it.

cpojer:
I'm gonna close this PR as unfortunately we didn't do a good job iterating on it in the past, and the original author isn't responsive any longer (sorry about that :( ). If anyone is coming across this PR, please patch it, rebase and make sure the tests are passing.

Issue #13522, opened 2017-04-15
Android, Animated, useNativeDriver, and rotateY transforms - Animation not appearing


mini-eggs:
Want to start of by saying love the React Native project and have used it on a number of apps so far. Great dev experience.

### Description
On Android while using the Animated API with useNativeDriver set to true, i.e.
```
Animated.timing(this.state.rotateY, { useNativeDriver: true, toValue: 1, duration: 500 }).start();
```
the animation will not appear. Say we rotateY a `View` that is initially at `0deg` to `360deg`, the `View` will disappear when the animation begins then reappear after the animation's duration has ended. If you're rotating from `0deg` to a degree, X, s.t. `X % 360 !== 0` the `View` will disappear entirely until its rotateY degree, Y, fulfills `Y % 360 === 0` (hope that helps understand the issue 🤞 ). All other animations I have tried on Android with `useNativeDriver` option support has worked great.

Currently in prod on the mobile application I've been experience this issue (code linked below), the Android version of the app has useNativeDriver set to false (with a noticeable decline in performance) while iOS has it set to true and is working beautifully.

### Reproduction Steps and Sample Code
You can find the project I've been experience this issue [here](https://github.com/mini-eggs/Flippour). The important code to look at starts at line 26 [here](https://github.com/mini-eggs/Flippour/blob/master/src/scenes/game/animatedSquare.utilities.js). This is a small wrapper function for creating `Animated.View`s and is put to use in another important file that displays this problem occurring, [here](https://github.com/mini-eggs/Flippour/blob/master/src/scenes/game/animatedSquare.js).

Notice on line 32 of [animatedSquare.utilities.js](https://github.com/mini-eggs/Flippour/blob/master/src/scenes/game/animatedSquare.utilities.js) I have `useNativeDriver: Platform.OS === "ios",`. To reproduce this set this line to `useNativeDriver: true,`.

You should be able to git clone the repo linked above
```
git clone https://github.com/mini-eggs/Flippour.git
```
 and have it up and running on Android with `yarn android` after a simulator has been started. I have not included keystores for obvious reasons. 😜

### Solution
Ensure the transform rotateY is working on Android.

### Additional Information
* React Native version: ^0.43.3
* React version: 16.0.0-alpha.6
* Platform: Android
* Development Operating System: MacOS
* Dev tools: compileSdkVersion 25, buildToolsVersion 25.0.2, minSdkVersion 16, targetSdkVersion 23

[...]

Issue #14462, opened 2017-06-12
negative value of property transform doesn't work on some android 7.0 devices.


Feng999:
RN version:  "0.41.2"
devices:
FRD-AL10（honor 8）  EMUI:5.0      android: 7.0
MHA-AL00(Mate9)    EMUI:5.0      android:7.0

On some android device mentioned before , negative value of transform(e.g.. scaleX  scaleY ) doesn't work.  i found this problem when i'm using  the package of [react-native-invertible-scroll-view](https://github.com/expo/react-native-invertible-scroll-view).  It write like this:
```
let styles = StyleSheet.create({
  verticallyInverted: {
    flex: 1,
    transform: [
      { scaleY: -1 },
    ],
  },
  horizontallyInverted: {
    flex: 1,
    transform: [
      { scaleX: -1 },
    ],
  },
});
```
but when i use scaleY which is deprecated directly, like this:
```
let styles = StyleSheet.create({
  verticallyInverted: Platform.select({
    ios: {transform: [{ scaleY: -1 }]},
    android: {scaleY: -1},
  }),
  horizontallyInverted: Platform.select({
    ios: {transform: [{ scaleX: -1 }]},
    android: {scaleX: -1},
  }),
});
```
 it worked!   
Guess there is some thing wrong when get property from transform in native code,  Any ideas how does this fix?

yury:
Confirm at RN 0.43.2 Huawei OS7.0 P10. VTR-L09/VTR-L29

Doesn't work with transform, works with deprecated.

ptomasroos:
This is a real case reported by various user by us as well.

Feng999:
@ptomasroos  how do you guys fix this?

janicduplessis:
Looks like this could be a potential fix https://github.com/facebook/react-native/pull/14560

Feng999:
@janicduplessis  add` {"perspective": 1} `  is a better way to fix this  if it works.  i will try it latter.

yury:
No, `{"perspective": 1}` doesn't help :(

hramos:
Hi there! This issue is being closed because it has been inactive for a while. Maybe the issue has been fixed in a recent release, or perhaps it is not affecting a lot of people. Either way, we're automatically closing issues after a period of inactivity. Please do not take it personally!

If you think this issue should definitely remain open, please let us know. The following information is helpful when it comes to determining if the issue should be re-opened:

[...]

PR #14560, opened 2017-06-16
Fix inverted FlatList for Android: add perspective transform

bartolkaruza:
On one of our test phones, a Huawei P10 VTR-L09, running Android 7.0 any rotate or scale animations through transforms are not working, unless `{"perspective": 1}` is added in the transforms array. I found the perspective solution here https://github.com/facebook/react-native/issues/13522 .

The new `inverted` property for `FlatList` relies on a scaleX/scaleY transform. Because this transform does not work on some Android devices, the FlatList is not displayed at all on those devices. This fix just adds the extra perspective array element. The result is a correctly displayed FlatList when `inverted` is set to `true`.

## Test Plan

Tested this on inverted `FlatList` on Android and iOS, on real devices. iOS is unaffected by the addition. Android now displays the `FlatList` correctly.

Minimal gist for testing: https://gist.github.com/bartolkaruza/87ef49d1b2d0227782739584b110722b

Minimal gist showing the View transform with scaleY issue:
https://gist.github.com/bartolkaruza/42bb9f9d63baaaebb48fc493c8127e1a

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -22,6 +22,7 @@ const ScrollView = require('ScrollView');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
 const ViewabilityHelper = require('ViewabilityHelper');
+const Platform = require('Platform');
 
 const flattenStyle = require('flattenStyle');
 const infoLog = require('infoLog');
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1204,7 +1204,7 @@ class CellRenderer extends React.Component {
 const verticalTransform = [{scaleY: -1}];
 const horizontalTransform = [{scaleX: -1}];
 
-if(Platform.OS === 'android') {
+if (Platform.OS === 'android') {
   verticalTransform.push({perspective: 1});
   horizontalTransform.push({perspective: 1});
 }
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1200,12 +1200,20 @@ class CellRenderer extends React.Component {
   }
 }
 
+const verticalTransform = [{scaleY: -1}];
+const horizontalTransform = [{scaleX: -1}];
+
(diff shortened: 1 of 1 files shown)


janicduplessis:
@bartolkaruza Thanks for figuring out a fix for this issue. I think a better fix would be to always set perspective: 1 for all transforms in the native code, this way it would completely fix the bug instead of just for FlatList.

I think we could always call view.setCameraDistance with whatever camera distance value `perspective: 1` gives if no perspective transform is specified. I was thinking just adding an else branch in that code https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L195

bartolkaruza:
I agree that would be a much better fix. I can look into this today if you are not implementing it @janicduplessis ?

Feng999:
Seems to be same problem with #14462

janicduplessis:
@bartolkaruza I don't have a Huawei device so if you could look into it it would be great.

bartolkaruza:
I opened a new PR for the fix in native code https://github.com/facebook/react-native/pull/14646

yury:
We have Huawei device and `{"perspective": 1}` doesn't help :(
Works only with deprecated scaleX/Y.

bartolkaruza:
@yury Have you tried `"perspective": 1280`, which I used in the proposed fix at https://github.com/facebook/react-native/pull/14646? Both are working on a P9 Huawei device for me.

yury:
@bartolkaruza I didn't try 1280, but I think it is not good solution either - it is dependant on screen density.
> The distance is expressed in "depth pixels." The default distance depends on the screen density. For instance, on a medium density display, the default distance is 1280. On a high density display, the default distance is 1920.
[from docs](https://developer.android.com/reference/android/view/View.html#setCameraDistance(float))

bartolkaruza:
The perspective value is multiplied in native code. So 1280 * density is the actual applied value. By setting 1280 you set it to the default value for all densities respectively.

yury:
@bartolkaruza will try 1280 later today and let you know about results.

yury:
@bartolkaruza `1280` works

PR #14562, opened 2017-06-16
[VirtualList] Fix problems if _updateCellsToRender gets called too early

mikelambert:
(prior to the scroll metrics being set up, or prior to the view being scrolled to the right location).

I'm using a StackNavigator to push a FlatList secondary screen, and it was causing extraneous rendering (item X ->item 0 -> item X)

I've logged inside the render functions of my `renderItem` function, and verified that it no longer attempts to render item 0 anymore. And I've verified from the `VirtualizedList.state` no longer renders a `{first: 0, last: 0}`, but persistently stays at `{first: X, last: X}` from start until things stabilize.

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -815,9 +815,25 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
     this.setState((state) => {
       let newState;
       if (!disableVirtualization) {
-        newState = computeWindowedRenderLimits(
-          this.props, state, this._getFrameMetricsApprox, this._scrollMetrics,
-        );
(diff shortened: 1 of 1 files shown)


Libraries/Lists/VirtualizedList.js
```diff
@@ -815,9 +815,25 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
     this.setState((state) => {
       let newState;
       if (!disableVirtualization) {
-        newState = computeWindowedRenderLimits(
-          this.props, state, this._getFrameMetricsApprox, this._scrollMetrics,
-        );
+        newState = state;
```
sahrens: We should do this with the declaration on line 816.

PR #14563, opened 2017-06-16
[VirtualList] Fix initial-render triggering a render of all items.

mikelambert:
When rendering a large list without a getItemLayout (ie SectionList, where it's hard to compute), it freaks out and attempt to render them all, starving the rendering/layout engine from computing the size (and more accurately figurnge out how few it actually should render.)

Before this change, with <FlatList maxToRenderPerBatch={5} data={objectOfLengthN} />, I essentially saw it doing:
```
{first: 0, last: 0}
{first: 0, last: 5}
{first: 0, last: 10}
....
{first: 0, last: N}
```
(no more hiPri renders since all elements have been rendered)
(actually renders and lays out all N objects, and computes their sizes)
(realizes that it doesn't need to show all N for my current screen size and row size)
```{first: 0, last: 55}```

After this change, that whole first part where it keeps incrementing `last` to try to "keep up with the scrolling" disappears.

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -775,7 +775,12 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
         (velocity > 2 && distBottom < scrollingThreshold)
       );
     }
-    if (hiPri) {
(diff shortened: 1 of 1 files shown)


sahrens:
Sweet, thanks for the fix!

PR #14568, opened 2017-06-16
Add navigator.languages, navigator.language and Languages module

zoontek:
## Motivation

Polyfill [NavigatorLanguage.languages](https://developer.mozilla.org/fr/docs/Web/API/NavigatorLanguage/languages) and [NavigatorLanguage.language](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language). Provides a solution on Android because the system doesn't reload the application after a language change.

## Test Plan

You can access `navigator.languages` and `navigator.languages`.
I added two screens in the RNTester app too, and a bit of documentation.

screen shot 2017-06-16 at 21 11 57

screen shot 2017-06-18 at 19 31 09


zoontek:
I just realised we do not need to squash commits anymore (with the squash and merge GitHub function). Sorry about that! 😅

zoontek:
@brentvatne I cleanup the code, the documentation and switch to an event listener based solution on Android (similar as https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onlanguagechange)

RNTester/js/LanguagesExample.js
```diff
@@ -0,0 +1,58 @@
...
+exports.description = 'Device languages';
+
+const examples = [
+  {
+    title: 'navigator.language',
+    render: function(): React.Element<any> {
+      return <Text>{navigator.language}</Text>;
+    },
+  },
+  {
```
zoontek: This has been fixed here: https://github.com/facebook/react-native/pull/14568/commits/b6e8b91fb5efd18a4789751a8687985fe2d375ad#diff-0468751a1a9708b31a37cde71e096450R20

javache:
Can you motivate why this needs to be in the React Native core instead of being a 3rd party package?

zoontek:
@javache I guess for the same reasons navigator.geolocation is in the core :)

brentvatne:
@javache 

> Can you motivate why this needs to be in the React Native core instead of being a 3rd party package?

It's really hard to draw the lines between what should or shouldn't go in core at this point, maybe we need to have a written policy for this. We have modules like `AdSupportIOS` which almost certainly do not belong, but it's a little less clear to me when it comes to APIs that are common to a majority or a large minority of apps, like getting the device language. In my opinion, if `AccessibilityInfo` belongs in core, then perhaps `navigator.language` does too -- both deal with exposing APIs that provide the developer with critical information about how to present the app to the user.

By adding `navigator.language(s)` we're adding support for a web standard that library authors can then depend on for cross-platform internationalization libraries. I believe this will lead to less fragmentation in the ecosystem, where different internationalization libraries provide their own ad-hoc implementation. Adding this to core isn't the only way to achieve this -- we may want to pull out web polyfills that aren't deeply integrated into react-native (eg deeply integrated: timers, eg not integrated: location) into a react-native-web-polyfills library then. I am in support of this if it is something that is explicitly called out in the react-native documentation.

fkgozali:
I think there are a few things to consider:

#### What is the use-case of JS (product code) knowing the locale(s)?
I imagine the use case is to:
1. show translated texts
2. have different UI treatment based on the nature of the locale, e.g. RTL (right-to-left)
3. fetching data differently from server based on locale

For (1) it's safer to ask native to provide the translated texts (see the next 2 sections) because native knows what locale is actually supported by the app. We don't want to be in the situation where JS thinks we support language B (hence displaying language B content from somewhere) but the app doesn't think it supports language B.

For (2), native module can provide those capability information and can be kept OS-specific. For example: https://github.com/facebook/react-native/blob/9ee815f6b52e0c2417c04e5a05e1e31df26daed2/React/Modules/RCTI18nManager.m#L27-32 was built exactly for that.

It's also more robust to built flexible UI with flexbox + onLayout() to handle dynamically sized component based on text context length (e.g. for language with longer translation).

For (3), the native networking library that backs the fetching API should be able to send the locale from native via User-Agent

So I'm not sure if we really need to know the locale code in JS.

#### Differentiating device locale from locale that the app supports
An app decides what it wants to support by including the translation assets in the app. This is independent of what locale the device is actually set to, and the user's language system preference. If an app only supports en_US, what locale value needs to be exposed to the navigator.language? Exposing non-en_US locale may not be very helpful because the JS code can't really depend on it if the app doesn't support that locale.

#### Locale codes consistency
I don't have a handy document describing the difference between Apple's locale code set from Android set, but I imagine there may be some differences. In that case, perhaps the values exposed in navigator.languages need to be normalized to a standard set.

Given these, I think i18n setup should be specific to an app given all the setup an app needs to do in both platforms:
* iOS: https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html
* Android: https://developer.android.com/guide/topics/resources/localization.html

A 3rd party module model could help with the i18n boilerplate set up, but I'm not sure it belongs to core. But things like core language capability like RTL is kinda tied to the core because it affects how we do layout in Yoga (hence why it's in RCTI18nManager linked above).

> Adding this to core isn't the only way to achieve this -- we may want to pull out web polyfills that aren't deeply integrated into react-native (eg deeply integrated: timers, eg not integrated: location) into a react-native-web-polyfills library then

I think this is probably a good idea going forward. I think not everything in web standard is directly applicable to native world 1-to-1 so I think we can be selective on which ones make sense.

zoontek:
@fkgozali Maybe a Library (like Linking, Geolocation, in `/Libraries`)  would be a better idea ?
Can I make a new proposal with this changes?

ide:
One thing @zoontek's approach (or react-native-web-polyfills) achieves is that it keeps "system" code in native and application code in JS. This is something that doesn't really matter inside Facebook but helps a lot everywhere else.

zoontek:
@javache @fkgozali @ide @brentvatne I moved `Languages` to an external module, like `Geolocation`. Also, it's not specific to Android anymore (`LanguagesAndroid` => `Languages`)

I don't update the default project template for now, only the `RNTester` project. Please give it a try.
If it won't do it, I'll create `react-native-languages` (not really a polyfill, without `InitializeCore` it doesn't seems logic to me to add properties to the `navigator` global object.

zoontek:
Any news?

fkgozali:
Thanks for making it more generic, and sorry for the late response. I think my preference is to put it as a separate module like "react-native-languages" so that apps that don't need it don't have to incur the cost of loading the module.

zoontek:
The project repository lives here: https://github.com/react-community/react-native-languages

PR #14584, opened 2017-06-18
Expose content-available APS key for iOS silent push

blankg:
<details>

  - [ ] Explain the **motivation** for making this change.
  - [ ] Match the **code formatting** of the rest of the codebase.
  - [ ] Target the `master` branch, NOT a "stable" branch.

</details>

## Motivation (required)

_What existing problem does the pull request solve?
In iOS when sending a silent push notification you need to configure the 'content-available' APS key to the value of 1 (When this key is present, the system wakes up your app in the background and delivers the notification to its app delegate, see [apple docs](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html#//apple_ref/doc/uid/TP40008194-CH17-SW1)).

This PR exposes this property to the notification event handler so app code can handle silent push scenario specifically. Currently this property is not available.

## Test Plan (required)

I've updated the PushNotificationIOSExample in the RNTester.

1. Open RNTester in xcode
2. Enable the  push notifications capability
3. run on device
4. Go to PushNotificationIOS
5. click on "send fake notification"
6. verify alert message contains 'content-available' with a value of 1.


RNTester/js/PushNotificationIOSExample.js
```diff
@@ -72,11 +72,13 @@ class NotificationExample extends React.Component {
 
   _sendNotification() {
     require('RCTDeviceEventEmitter').emit('remoteNotificationReceived', {
+      remote: true,
```
shergin: Why do we need this?

RNTester/js/PushNotificationIOSExample.js
```diff
@@ -115,9 +117,15 @@ class NotificationExample extends React.Component {
   }
 
   _onRemoteNotification(notification) {
+    const result = `Alert message: ${notification.getMessage()}
```
shergin: nit: Indention.

Libraries/PushNotificationIOS/PushNotificationIOS.js
```diff
@@ -485,6 +487,13 @@ class PushNotificationIOS {
   }
 
   /**
+   * Gets the content-available number from the `aps` object
+   */
+  getContentAvailable(): ?number {
```
shergin: Why the returning type is different from `_contentAvailable`.

shergin: Looks like the only possible values are `1` and `null`, right?

RNTester/js/PushNotificationIOSExample.js
blankg: This property is added by the iOS native code for remote notification (handleRemoteNotificationReceived) and the PushNotificationIOS constructor looks for it otherwise it will not look for the aps object.

Libraries/PushNotificationIOS/PushNotificationIOS.js
blankg: yes, it can be 1 or null/undefined so this is why it is optional. I aligned _contentAvailable member with the other private members in this class which add the optional flag only in the getter.

RNTester/js/PushNotificationIOSExample.js
blankg: will fix

Libraries/PushNotificationIOS/PushNotificationIOS.js
shergin: Can we declare special type `type ContentAvailable = 1;` (https://flow.org/en/docs/types/unions/) and then use it as `?ContentAvailable`? (I am not sure.) Or maybe we can create special already optional type?

blankg: Create a ContentAvailable type which is already optional

Issue #14606, opened 2017-06-19
Android React Text Component includeFontPadding:false incorrect height


mishagreenberg:
On Android, when includeFontPadding is set to false, the React Text component does not adjust its height. This makes it difficult to lay out other components at a precise spacing relative to a Text component.

For the native TextView component, the height is different depending on the value of includeFontPadding-

When includeFontPadding is set to true, the height of the TextView is based on abs(FontMetrics.top) + abs(FontMetrics.bottom). 

When includeFontPadding to set to false, the height of the TextView is based on abs(FontMetrics.ascent) + abs(FontMetrics.descent).

This is happening because in ReactTextShadowNode.java, includeFontPadding is hardcoded to true when passing parameters to StaticLayout and BoringLayout. This seems like an easy fix.

An example from the measure function in ReactTextShadowNode.java -

private final YogaMeasureFunction mTextMeasureFunction = new YogaMeasureFunction() {
```
BoringLayout.make(
                text,
                textPaint,
                boring.width,
                Layout.Alignment.ALIGN_NORMAL,
                1.f,
                0.f,
                boring,
                true); // this parameter is includeFontPadding and should be based on the includeFontPadding value style of the component
...
};
```

dmueller39:
@hramos Can you provide a reason for closing the issue? Does it need more information?

lelandrichardson:
I'd like to reopen this issue as I believe font rendering remains to be one of the biggest reasons for cross-platform differences across iOS and Android and this is an important thing to explore.

msand:
We're tracking similar issues in react-native-svg here: https://github.com/react-native-community/react-native-svg/issues/362 and here https://github.com/react-native-community/react-native-svg/issues/377

I've fixed most of the issues, but it depends on magic constants in two places. And kerning is still off (if the letter-spacing is not set to a length value). Font rendering and kerning has a surprisingly big effect on the user / reading experience, especially if there are large amounts of text or larger font-sizes on headers and labels etc.

mishagreenberg:
#14609 is a PR to fix this issue

mishagreenberg:
#14609 was merged

PR #14609, opened 2017-06-19
Size height of Android Text component based on includeFontPadding style

mishagreenberg:
Overview - 

This PR resolves the issue described in #14606. This PR makes Text components take into account the includeFontPadding property when calculating their size.

Background - 

Currently, on Android, when includeFontPadding is set to false, the React Text component does not adjust its height. This makes it difficult to lay out other components at a precise spacing relative to a Text component.

iOS calculates the height of a UILabel based on the font's descent + ascent.

Android lets you choose whether to calculate the height of a TextView based on the font's top + bottom (includeFontPadding=true) or ascent + descent (includeFontPadding=false).

In order for a text component to be the same size on iOS and Android (relative to the rest of the layout in points and dips), one should set includeFontPadding=false on Android - but the React Text component needs to take this property into account when sizing itself for this to work.

Please see this stack overflow post for a visual explanation of the difference between a font's ascent/descent and top/bottom - https://stackoverflow.com/questions/27631736/meaning-of-top-ascent-baseline-descent-bottom-and-leading-in-androids-font

Testing -

Please see the attached screenshots to see the height difference of a Text component with this PR when includeFontPadding is true vs false.

The font I am using has an ascent + descent = em-size so that the height of the Text component will be equal to the font-size for a single line of text. This is to clearly show the additional height that includeFontPadding=true adds to the Text component.

For Text components that are styled in the same way,

When includeFontPadding=true, height = ~29.714 dips
When includeFontPadding=false, height= 24 dips

includefontpaddingtrue

includefontpaddingfalse

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/flat/RCTText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/flat/RCTText.java
@@ -57,6 +57,7 @@ import com.facebook.react.uimanager.annotations.ReactProp;
   private float mSpacingAdd = 0.0f;
   private int mNumberOfLines = Integer.MAX_VALUE;
   private int mAlignment = Gravity.NO_GRAVITY;
+  private boolean mIncludeFontPadding = true;
 
   public RCTText() {
     setMeasureFunction(this);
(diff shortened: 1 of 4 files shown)


hramos:
Thanks for submitting this. I'll see if I can get someone to review this.

PR #14618, opened 2017-06-20
DEPRECATION: Make NetInfo API cross platform and expose whether connection is 2g/3g/4g

rigdern:
This change intends to fix 2 issues with the NetInfo API:
  - The NetInfo API is currently platform-specific. It returns completely different values on iOS and Android.
  - The NetInfo API currently doesn't expose a way to determine whether the connection is 2g, 3g, or 4g.

The NetInfo API currently just exposes a string-based enum representing the connectivity type. The string values are different between iOS and Andorid. Because of this design, it's not obvious how to achieve the goals of this change without making a breaking change. Consequently, this change deprecates the old NetInfo APIs and introduces new ones. Specifically, these are the API changes:
  - The `fetch` method is deprecated in favor of `getConnection`
  - The `change` event is deprecated in favor of the `connectionchange` event.
  - `getConnection`/`connectionchange` use a new set of enum values compared to `fetch`/`change`. See the documentation for the new values.
    - On iOS, `cell` is now known as `cellular`. It's worth pointing out this one in particular because the old and new names are so similar. The rest of the iOS values have remained the same.
    - Some of the Android enum values have been removed without a replacement (e.g. `DUMMY`, `MOBILE_DUN`, `MOBILE_HIPRI`, `MOBILE_MMS`, `MOBILE_SUPL`, `VPN`). If desirable, we could find a way to expose these in the new API. For example, we could have a `platformValue` key that exposes the platform's enum values directly (like the old `fetch` API did).

`getConnection` and `connectionchange` each expose an object which has 2 keys conveying a `ConnectionType` (e.g. wifi, cellular) and an `EffectiveConnectionType` (e.g. 2g, 3g). These enums and their values are taken directly from the W3C's Network Information API spec (https://wicg.github.io/netinfo/). Copying the W3C's API will make it easy to expose a `navigation.connection` polyfill, if we want, in the future. Additionally, because the new APIs expose an object instead of a string, it's easier to extend the APIs in the future by adding keys to the object without causing a breaking change.

Note that the W3C's spec doesn't have an "unknown" value for `EffectiveConnectionType`. I chose to introduce this non-standard value because it's possible for the current implementation to not have an `effectiveConnectionType` and I figured it was worth representing this possibility explicitly with "unknown" instead of implicitly with `null`.

**Test Plan (required)**

Verified that the methods (`fetch` and `getConnection`) and the events (`change` and `connectionchange`) return the correct data on iOS and Android when connected to a wifi network and a 4G cellular network. Verified that switching networks causes the event to fire with the correct information. Verified that the old APIs (`fetch' and 'change') emit a deprecation warning when used. My team is using a similar patch in our app.

Adam Comella
Microsoft Corp.


Libraries/Network/NetInfo.js
```diff
@@ -79,56 +80,51 @@ const _isConnectedSubscriptions = new Map();
  * NetInfo exposes info about online/offline status
  *
  * ```
- * NetInfo.fetch().then((reach) => {
- *   console.log('Initial: ' + reach);
+ * NetInfo.getConnection().then((connection) => {
```
shergin: This name is quite confusing (because it is not connection), can/should we rename it to something like `getConnectionInfo` or `getConnectionStatus`?

Libraries/Network/NetInfo.js
```diff
@@ -167,22 +163,77 @@ const _isConnectedSubscriptions = new Map();
...
     handler: Function
   ): {remove: () => void} {
-    const listener = NetInfoEventEmitter.addListener(
-      DEVICE_CONNECTIVITY_EVENT,
-      (appStateData) => {
-        handler(appStateData.network_info);
-      }
-    );
+    let listener;
+    if (eventName === 'connectionchange') {
```
shergin: Why the event name is all lower case? And

Libraries/Network/NetInfo.js
rigdern: @shergin I modeled it after the W3C Network Information API which offers this information off of a `connection` property (https://wicg.github.io/netinfo/#-dfn-connection-dfn-attribute). I can switch it to `getConnectionInfo` if you'd prefer.

Libraries/Network/NetInfo.js
rigdern: I'll switch it to `connectionChange`

rigdern:
@shergin

> Why not just add new values to existing event, deprecate old values, and delete old values eventually?

If we were to reuse the existing event which fires with a string, how would we convey both the old value (so it's not a breaking change) and the new value? For example, on iOS we used to have `cell`. Now we have `cellular`. What string would the event fire with?

An additional thing to note is that the new event fires with an object of 2 keys (`type` and `effectiveType`) whereas the old event fired with a string.

Libraries/Network/NetInfo.js
shergin: I see. `The connection attribute, when getting, returns an object that implements the NetworkInformation interface.` Yeah, I think using `getConnectionInfo` or `getConnectionInformation` would be much better.

shergin:
@rigdern I didn't noticed that the old event has just string parameter.
So, 👍!

rigdern:
@shergin Thanks for your feedback. I pushed a commit to rename those 2 APIs:
  - `connectionchange` event is now `connectionChange`
  - `getConnection` method is now `getConnectionInfo`

shergin:
@rigdern Could you please verify that RNTester can be built with this patch?

I am experiencing this error (seems we have to add CoreTelefony framework to the project file):
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_CTTelephonyNetworkInfo", referenced from:
      objc-class-ref in libRCTNetwork.a(RCTNetInfo.m.o)
  "_CTRadioAccessTechnologyWCDMA", referenced from:
      _RCTReachabilityCallback in libRCTNetwork.a(RCTNetInfo.m.o)
  "_CTRadioAccessTechnologyHSDPA", referenced from:
  "_CTRadioAccessTechnologyHSUPA", referenced from:
  "_CTRadioAccessTechnologyEdge", referenced from:
  "_CTRadioAccessTechnologyCDMAEVDORevA", referenced from:
  "_CTRadioAccessTechnologyGPRS", referenced from:
  "_CTRadioAccessTechnologyCDMAEVDORevB", referenced from:
  "_CTRadioAccessTechnologyeHRPD", referenced from:
  "_CTRadioAccessTechnologyCDMA1x", referenced from:
  "_CTRadioAccessTechnologyCDMAEVDORev0", referenced from:
  "_CTRadioAccessTechnologyLTE", referenced from:
```

PR #14632, opened 2017-06-20
Update of react gradle to increase node heap

kazak1377:
Updated gradle file, so that packager will not fail on assembling projects with big amount of assets

## Motivation (required)

When building the apps for android with more then 200mb of assets the packager will fail on JavaScript heap end. This will solve that problem.

## Test Plan (required)

Try to build application with 150 images 4-20mb each. Before now it will fail on JavaScript heap end. After this commit that error will be solved (for now only for android).

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -80,10 +80,10 @@ gradle.projectsEvaluated {
                 def devEnabled = !(config."devDisabledIn${targetName}"
                     || targetName.toLowerCase().contains("release"))
                 if (Os.isFamily(Os.FAMILY_WINDOWS)) {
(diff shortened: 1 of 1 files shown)


chirag04:
i don't think this is required anymore after this: https://github.com/facebook/react-native/commit/58fb91e62bbab1db648725367f5b1af3350a1954

@kazak1377 can you confirm?

kazak1377:
Yep, after https://github.com/facebook/react-native/commit/58fb91e62bbab1db648725367f5b1af3350a1954 my pull request is not required anymore.

PR #14637, opened 2017-06-20
Allow apps to manually control when the dev support tools are enabled

### LINKED ISSUE #13135: DevOptionsSupportManager doesn't null dialog in reload()
If reload() gets called in DevOptionsSupportManagerImpl while the dialog is showing, it will call dismiss() on the dialog but not null it (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManagerImpl.java#L917).
The next time (and any time after that) showDevOptionsdDialog() is called, the dialog will be non-null but dismissed (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManagerImpl.java#L349), a state which can't be recovered from.

Solution:
reload() should null out mDevOptionsDialog when dismissing it.

* React Native version: 0.38.103
* Platform: Android
* Development Operating System: N/A
* Dev tools: N/A

gpeal:
## Motivation
For multi-activity apps, apps that want fine grained control over when the dev support tools are enabled, or apps that frequently switch between native and react native, it isn't ideal to automatically enable and disable the dev tools in onHostPause and onHostResume. This allows apps to manually control when the dev tools are enabled. For example, in the Airbnb app, we enable and disable dev tools in our base Activity so the menu is available in native screens as well as React Native screens.

## Test Plan
This has been running on the Airbnb fork for several months (https://github.com/airbnb/react-native/commit/50325eae895112eef903fde5b92c566c8d392efa and https://github.com/airbnb/react-native/commit/404b1030ead595e60a1d4c79e1732a8bfc8aeb1a)

Fixes #13135

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java
@@ -151,6 +151,7 @@ public class ReactInstanceManager {
   private final boolean mSetupReactContextInBackgroundEnabled;
   private final boolean mUseSeparateUIBackgroundThread;
   private final int mMinNumShakes;
+  private final boolean mManuallyEnableDevSupport;
 
   private final ReactInstanceDevCommandsHandler mDevInterface =
       new ReactInstanceDevCommandsHandler() {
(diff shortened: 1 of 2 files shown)


gpeal:
@lelandrichardson

Issue #14548, opened 2017-06-16
[0.45.1] Android release build failed


lslvxy:
just like #14417
```

FAILURE: Build failed with an exception.

* What went wrong:
Could not list contents of '/Users/lise/WorkSpace/idlebox/node_modules/react-native/third-party/glog-0.3.4/test-driver'. Couldn't follow symbolic link.

* Try:
Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.GradleException: Could not list contents of '/Users/lise/WorkSpace/idlebox/node_modules/react-native/third-party/glog-0.3.4/test-driver'. Couldn't follow symbolic link.
        at org.gradle.api.internal.file.collections.jdk7.Jdk7DirectoryWalker$1.visitFile(Jdk7DirectoryWalker.java:79)
        at org.gradle.api.internal.file.collections.jdk7.Jdk7DirectoryWalker$1.visitFile(Jdk7DirectoryWalker.java:54)
        at org.gradle.api.internal.file.collections.jdk7.Jdk7DirectoryWalker.walkDir(Jdk7DirectoryWalker.java:54)
        at org.gradle.api.internal.file.collections.DirectoryFileTree.walkDir(DirectoryFileTree.java:146)
        at org.gradle.api.internal.file.collections.DirectoryFileTree.visitFrom(DirectoryFileTree.java:130)
        at org.gradle.api.internal.file.collections.DirectoryFileTree.visit(DirectoryFileTree.java:115)
        at org.gradle.api.internal.file.collections.DirectoryFileTree.visitTreeOrBackingFile(DirectoryFileTree.java:111)
        at org.gradle.api.internal.file.collections.FileTreeAdapter.visitTreeOrBackingFile(FileTreeAdapter.java:114)
        at org.gradle.api.internal.changedetection.state.CachingTreeVisitor.doVisitTree(CachingTreeVisitor.java:154)
        at org.gradle.api.internal.changedetection.state.CachingTreeVisitor.visitTreeForSnapshotting(CachingTreeVisitor.java:92)
        at org.gradle.api.internal.changedetection.state.DefaultFileCollectionSnapshotter.visitFiles(DefaultFileCollectionSnapshotter.java:56)
        at org.gradle.api.internal.changedetection.state.AbstractFileCollectionSnapshotter$DefaultFileCollectionSnapshotPreCheck.<init>(AbstractFileCollectionSnapshotter.java:104)
[...]

Issue #14464, opened 2017-06-12
Could not list contents of '/third-party/glog-0.3.4/test-driver' - Android Release


ben-snaize:
**Scenario**
When creating an Android release APK, the error `Could not list contents of '/Users/user/project/node_modules/react-native/third-party/glog-0.3.4/test-driver'` is encountered which causes the build to fail.

**Steps to Reproduce**
`npm i react-native@0.45 --save`
`cd android && ./gradlew assembleRelease && ./gradlew installRelease`

There is a folder named `third-party-podspecs` that contains the `GLog.podspec`. This probably gets expanded into another folder, `third-party` on build. Inside of this, there is a file under `glog-0.3.4/test-driver` which appears to be an alias that does not resolve.

<img width="983" alt="screen shot 2017-06-12 at 10 03 32" src="https://user-images.githubusercontent.com/13312361/27026659-6f2acbb6-4f56-11e7-9fea-f95c52c1d336.png">

There is a temporary fix which is to delete the `third-party` folder. Please can we look at getting this one corrected?

ms88privat:
https://github.com/facebook/react-native/issues/14417

nehvaleem:
also having this issue

g4laura:
I'm also having this issue.

itinance:
This worked for me: to delete the broken test-driver-link as follows

` unlink node_modules/react-native/third-party/glog-0.3.4/test-driver `

kelset:
Same issue over here, fix by @itinance works 🔥

tapz:
In addition to unlinking, I also had to remove `third-party/glog-0.3.4` in the project root.

ujwal-setlur:
With RN 0.46.3, @itinance solution worked for me.

ujwal-setlur:
This issue occurs in 0.46.3 only if you do an android build **after** you do an iOS build. That link gets created during the iOS build. If you do an android build after a fresh npm install, it's fine.

It's ironic that I ran into this issue after I upgraded after https://github.com/facebook/react-native/issues/14382#issuecomment-315233853 was fixed in 0.46.3

ujwal-setlur:
Please see https://github.com/facebook/react-native/commit/5c53f89dd86160301feee024bce4ce0c89e8c187#commitcomment-23107918

ujwal-setlur:
This is fixed in master branch. Guess we need to wait for 0.47.0

hramos:
[...]

Issue #14417, opened 2017-06-09
[0.45.0] Android release build failed


vomchik:
After updating to new version release build failed. Any ideas how does this fix?

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not list contents of '/Users/vomchik/Projects/LITS/toolook/toolook-app-react-native/node_modules/react-native/third-party/glog-0.3.4/test-driver'. Couldn't follow symbolic link.
```

olivierkaisin:
Same problem here, it happens to me when I try to build a release APK for Android at the `:app:bundleReleaseJsAndAssets` step.

As a temporary solution deleting the whole `node_modules` folder and re-installing dependencies fixes it.

tabrindle:
I use a Makefile that points its clean target to a full clean script. I have found most issues like this can be solved by killing all simulators, packagers, etc, running clean, then reinstalling node_modules:

clean.sh
```
watchman watch-del-all
rm -rf node_modules
rm -rf $TMPDIR/react-*
```

hramos:
Hey, thanks for reporting this issue!

It looks like your description is missing some necessary information. Can you please add all the details specified in the [template](https://github.com/facebook/react-native/blob/master/.github/ISSUE_TEMPLATE.md)? This is necessary for people to be able to understand and reproduce the issue being reported.

I am going to close this, but feel free to open a new issue with the additional information provided. Thanks!

cinder92:
same issue here! whats going on with latest react native release? is fully os issues in both platforms :(

jdeibe:
Same error for me but after a cleanup all works.

benoitvallon:
I have the same error. Cleaning the `node_modules` works for me but the error reappears after every Xcode build. Any idea why?

nehvaleem:
Also having exact same issue. Solution provided by @olivierkaisin seems to work, but it is indeed frustrating having to delete node_modules at each build.

attilaaronnagy:
after every node_modules reinstall i have to update the build tools versions.. so the solution provided by @olivierkaisin works, but please fix this bug, every android build takes 10 times longer than before :(

hramos:
This issue is closed. If this is still affecting you, please open a new issue and make sure to fill out everything in the template as documented in the contribution guidelines. Thanks!

lslvxy:
Removing node_modules/.bin/ worked

itinance:
For me helped to delete the broken test-driver-link:

[...]

PR #14638, opened 2017-06-20
Removing test-driver symlink in ios third party install script

ArturKlajnerok:
## Motivation (required)
After execution of `scripts/ios-install-third-party.sh` a symlink is created : 
`<YOUR-APP-PATH>/node_modules/react-native/third-party/glog-0.3.4/test-driver`
that is pointing to `test-driver -> /usr/share/automake-1.14/test-driver`
This can be executed indirectly by `react-native run-ios`.

This breaks the bundle process if the system don't contain a given file under the link and having this strict dependency on the system setup is not a good practice.

Once the `test-driver` symlink is created android app release is failing, for :
`./gradlew assembleRelease`
the `:app:bundleReleaseJsAndAssets` returns :
```

* What went wrong:
Could not list contents of '<YOUR-APP-PATH>/node_modules/react-native/third-party/glog-0.3.4/test-driver'. Couldn't follow symbolic link.
```
Related issues:
https://github.com/facebook/react-native/issues/14417
https://github.com/facebook/react-native/issues/14464
https://github.com/facebook/react-native/issues/14548

## Test Plan (required)

1. Create new project with `react-native init <YOUR-APP>`
2. cd `<YOUR-APP>/`
3. Run app on iOS `react-native run-ios` so `scripts/ios-install-third-party.sh` is executed.
4. cd `android/`
5. Run android app release `./gradlew assembleRelease` (it will work properly after this fix and fail if the `test-driver` symlink exists)

## Summary
IMHO we should resolve the issue with this quick fix and apply the proper fix later after the new version of `google/glog` will be released.
The proper cleanup of files generated by autotools was already applied : https://github.com/google/glog/pull/188

Please let me know if I should provide more details : @javache, @mhorowitz, @hramos

### DIFF
--- a/scripts/ios-configure-glog.sh
+++ b/scripts/ios-configure-glog.sh
@@ -7,6 +7,11 @@ if [ -z "$ACTION" ] || [ -z "$BUILD_DIR" ]; then
   export CXX="$CC"
 fi
 
+# Remove automake symlink if it exists
+if [ -h "test-driver" ]; then
+    rm test-driver
+fi
+
 ./configure --host arm-apple-darwin
 
 # Fix build for tvOS
--- a/scripts/ios-install-third-party.sh
+++ b/scripts/ios-install-third-party.sh
@@ -25,6 +25,9 @@ function fetch_and_unpack () {
              exit 1
          fi
          cd "$dir"
+         if [ -h "test-driver" ]; then
+             rm test-driver
+         fi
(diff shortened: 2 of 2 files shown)


javache:
Can you do this as part of `ios-configure-glog.sh`?

ArturKlajnerok:
@javache done in : 75651f7

PR #14640, opened 2017-06-20
[0.45] react@16 did away with PropTypes; require prop-types instead

mojodna:
`react@16` (a `peerDependency`) did away with the `PropTypes` export in favor of the `prop-types` module.

This updates all of the remaining references to `ReactPropTypes` (and yes, targets the `0.45-stable` branch; I'll submit a PR that targets `master` as well).

### DIFF
--- a/Libraries/Components/AppleTV/TVViewPropTypes.js
+++ b/Libraries/Components/AppleTV/TVViewPropTypes.js
@@ -10,7 +10,7 @@
  * @flow
  */
 'use strict';
-var PropTypes = require('React').PropTypes;
+var PropTypes = require('prop-types');
 
 /**
  * Additional View properties for Apple TV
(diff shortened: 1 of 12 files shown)


hramos:
Let's wait for #14641 to land first, then we can cherry-pick the commit from master.

mojodna:
FWIW, this commit includes other fixes that got taken care of somewhere on the way to the current master, so cherry picking won't hit all of the spots.

Jonovono:
What's the status on this? Currently React Native 0.45.1 breaks with React 16.0.0 a 12 because of the prop types.

hramos:
@mojodna which commits need to be cherry picked? If I can get a list of those commits already on master, that we can cherry pick onto 0.45 and thus cut a 0.45.2 release, I can get that moving.

hramos:
We won't be doing this via PR, so let's move the discussion to the following issue: #14712

PR #14646, opened 2017-06-20
always set camera distance on transforms, to default 1280 if 0

bartolkaruza:
This is the better fix for the same issue as mentioned in PR https://github.com/facebook/react-native/pull/14560

Certain rotateX, rotateY, scaleX and scaleY animations do not work correctly on some phones in Android 7.0.0, causing issues such as https://github.com/facebook/react-native/issues/14462 and https://github.com/facebook/react-native/issues/13522.

The issue can be fixed on JS side by setting an additional transform for perspective, eg. `{perspective: 1}` which triggers a `setCameraDistance` call in native code.

The fix in this PR always sets the camera distance on transforms, even when no perspective transform was specified. The default camera distance is set before the scale multiplication, to make sure that the value is appropriate for the phones density. The value calculates to an Android 'default' camera distance of 1280 * scale multiplier; https://developer.android.com/reference/android/view/View.html#setCameraDistance(float)

If a perspective transform is specified, this value will be used correctly still.

This fix was tested on the RNTester. Before the fix, on some devices, the FlatList example, with inverted turned on, will not display the list.

Devices that have been confirmed to have this issue:
FRD-AL10（honor 8） EMUI:5.0 android: 7.0
MHA-AL00(Mate9) EMUI:5.0 android:7.0
Huawei P10 VTR-L09, running Android 7.0

After the fix, the inverted FlatList displays correctly.


bartolkaruza:
@janicduplessis Could you help getting this looked at by someone? The pull request has been sitting in 'waiting for review' for more than a month

janicduplessis:
@bartolkaruza Sorry about the delay, I forgot about this.

@hramos Could you get someone @ fb to look at this?

hramos:
cc @shergin

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -184,16 +184,18 @@ private static void setTransformProperty(View view, ReadableArray transforms) {
 
     if (perspectiveArray.length > PERSPECTIVE_ARRAY_INVERTED_CAMERA_DISTANCE_INDEX) {
       float invertedCameraDistance = (float) perspectiveArray[PERSPECTIVE_ARRAY_INVERTED_CAMERA_DISTANCE_INDEX];
-      if (invertedCameraDistance < 0) {
-        float cameraDistance = -1 / invertedCameraDistance;
-        float scale = DisplayMetricsHolder.getScreenDisplayMetrics().density;
+      if (invertedCameraDistance == 0) {
+        // Default camera distance, before scale multiplier (1280)
+        invertedCameraDistance = 0.00078125f;
```
shergin: Should It be negative?

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -184,16 +184,18 @@ private static void setTransformProperty(View view, ReadableArray transforms) {
 
     if (perspectiveArray.length > PERSPECTIVE_ARRAY_INVERTED_CAMERA_DISTANCE_INDEX) {
       float invertedCameraDistance = (float) perspectiveArray[PERSPECTIVE_ARRAY_INVERTED_CAMERA_DISTANCE_INDEX];
-      if (invertedCameraDistance < 0) {
-        float cameraDistance = -1 / invertedCameraDistance;
-        float scale = DisplayMetricsHolder.getScreenDisplayMetrics().density;
+      if (invertedCameraDistance == 0) {
```
shergin: In which cases `invertedCameraDistance` is `0` and why?
Does `0` make any logical sense  as `invertedCameraDistance` value?

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
bartolkaruza: This is the exact value that comes through if you pass a `{perspective: 1280}` value in the transform array on the JS side. When omitting a perspective transform, the value is 0. With this change, the value is forced from 0 to the default 1280 value, forcing the `. setCameraDistance` call.

Not calling `.setCameraDistance` is causing the rendering issue, which is caused by a 0 `invertedCameraDistance`.

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
bartolkaruza: 0 is the value when no perspective transform is passed in the array in javascript. The deleted line `if (invertedCameraDistance < 0) {` was preventing that case from causing a division by 0 and a crash, but it was also preventing a call to `.setCameraDistance` which is causing the rendering issues. This change affects the 0 case, by forcing it to a sensible default other than 0.

shergin:
@bartolkaruza  Seems reasonable to me. Thank you!

PR #14652, opened 2017-06-21
Fix Apple TV compile errors

dlowder-salesforce:
## Motivation

Apple TV compilation is currently broken in master -- this fixes it.

## Test Plan

Existing tests will pass after this change.

### DIFF
--- a/React/Modules/RCTDeviceInfo.m
+++ b/React/Modules/RCTDeviceInfo.m
@@ -106,6 +106,7 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
 
 - (void)_interfaceOrientationDidChange
 {
+#if !TARGET_OS_TV
   UIInterfaceOrientation nextOrientation = [RCTSharedApplication() statusBarOrientation];
 
   // Update when we go from portrait to landscape, or landscape to portrait
@@ -121,6 +122,7 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
       }
 
   _currentInterfaceOrientation = nextOrientation;
+#endif
 }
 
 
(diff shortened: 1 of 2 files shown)


React/Modules/RCTUIManager.h
```diff
@@ -148,8 +148,8 @@ RCT_EXTERN NSString *const RCTUIManagerRootViewKey;
 - (void)rootViewForReactTag:(NSNumber *)reactTag withCompletion:(void (^)(UIView *view))completion;
 
 /**
- * Finds a view that is tagged with {@param nativeId} as its nativeID prop
```
javache: Any reason to change these?

React/Modules/RCTDeviceInfo.m
```diff
@@ -106,6 +106,7 @@ - (void)interfaceOrientationDidChange
 
 - (void)_interfaceOrientationDidChange
 {
+#if !TARGET_OS_TV
```
javache: Can you instead hide the methods completely if `!TARGET_OS_TV`? So put them from line 96 to line 124.

React/Modules/RCTUIManager.h
dlowder-salesforce: The original is giving a "documentation error"  when compiling and running the unit and integration tests.

React/Modules/RCTDeviceInfo.m
dlowder-salesforce: Done.

React/Modules/RCTUIManager.h
dlowder-salesforce: compile error image

PR #13785, opened 2017-05-04
React.podspec fix for RCTAnimation header issue

vjeranc:
Fixes issue described #13198

Simpler, compared to #13217 

Build no longer fails on not being able to find `RCTAnimation/RCTValueAnimatedNode.h`.

PR #13217, opened 2017-03-30
Enable full podspec CI build and make RCTAnimation subspec green.

alloy:
Fixes #13198.

This enables all subspecs defined in the podspec to be built individually, which does take a lot longer but ensures build failures in the smaller subspecs are caught too.

Also fixes a build failure in RCTAnimation by making the import consistent with other imports in the same area of the codebase.

/cc @janicduplessis @mkonicek

Issue #13198, opened 2017-03-29
[iOS][cocoapods] 'RCTAnimation/RCTValueAnimatedNode.h' file not found


skv-headless:
### Description

`'RCTAnimation/RCTValueAnimatedNode.h' file not found.`
in current master https://github.com/facebook/react-native/tree/c233191485ef733aa7a8d4ea5c758b9e6a06964c

1) related to 28 of march changes. @janicduplessis might help
2) podspec test are passed. @alloy do you have ideas how tests could be improved?
Maybe we need to run `xcbuild` to test?

### Reproduction Steps and Sample Code

https://github.com/skv-headless/AwesomeProject - demo project based on this page https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#cocoapods

### Solution

It works if change `#import <RCTAnimation/RCTValueAnimatedNode.h>` to `#import "RCTValueAnimatedNode.h"`

### Additional Information

* Platform: [iOS]
* Development Operating System: [MacOS]

alloy:
> podspec test are passed. @alloy do you have ideas how tests could be improved?
> Maybe we need to run xcbuild to test?

Nah, we should ideally build all the subspecs, I’ll take a look at how long those take.

alloy:
@skv-headless Your example fails to build from a fresh clone with different issues than the one this ticket is about. Can you make sure it fails with the exact issue you mention from a fresh clone (so no `pod install`, no `npm install`) ?

alloy:
While enabling full podspec testing I did encounter this failure, which is probably the one you’re referring to:

```
In file included from /Users/eloy/Code/React/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m:10:
/Users/eloy/Code/React/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h:12:9: fatal error: 'RCTAnimation/RCTValueAnimatedNode.h' file not found
#import <RCTAnimation/RCTValueAnimatedNode.h>
```

skv-headless:
Yep. This error. I've updated repo in case it still needed. Are you going to update test script?

alloy:
Yeah I’ll do that, and then we’ll see how much time it takes to complete on CI.

alloy:
Pushed #13217

shrutic:
[...]

PR #14664, opened 2017-06-21
[iOS][cocoapods] Fixes import error in "RCTNativeAnimatedNodesManager.h" 

ceyhuno:
Fixes import error in "RCTNativeAnimatedNodesManager.h" mentioned at #13198. 
This same fix was included in #13217 but it was not merged.
Unfortunately #13785 didn't fix this problem.


javache:
The import in the header needs to be namespaced, this is not a correct solution.

guysegal:
@javache when using nameapace it's impossible to compile the code without a script that changes the import.
I know the non namespaced imports are deprecated but this is a long time blocker for anyone who uses cocoapods and wants to upgrade

vjeranc:
This definitely worked on 0.44.0 after #13785

There was no need for namespace change.

Igor1201:
@vjeranc Could you please tell us in what device configuration (OSX/Xcode version) are you able to build it?
Also, we're using cocoapods 1.2.1.

guysegal:
I just doubled check, it's not working:

react native - 0.44.1
xcode - 8.3.2
cocoapods -  1.2.1 

We are using swift pods in the project so the podfile contains `use_frameworks!` declaration

shergin:
Looks like we have to add this header to "Copy header files" build step inside for RCTAnimation project.

guysegal:
if this will fix it that would be awesome.. :)

ceyhuno:
`RCTValueAnimatedNode.h` is already under `Copy Files` build phase in `RCTAnimation` project

shergin:
¯\_(ツ)_/¯

esam091:
Any conclusion to this PR? can't get my project to build without hacks 😕

timwangdev:
Please merge. Thanks!

rodriguesbreno:
@javache 

What about
instead of 
```import "RCTValueAnimatedNode.h"```
use 
```import  <React/RCTValueAnimatedNode.h>```

it works (but tested only with the lib at cocoa pods)

**Comment by shergin** [comment]
Okay, does anyone have idea which exact problem we are trying to solve?
Why exactly is the issue #13198 happening?

**Comment by vjeranc** [comment]
I must admit to having no idea why the [previous fix](https://github.com/facebook/react-native/pull/13785/files) does not work. We've been building 0.45.1 and newer updates without any issues.

There might be some conflicts in `project.pbxproj` and the modified version after applying `*.podspec` definitions. Our `project.pbxproj` was heavily cleaned (removed references to all unneeded projects like tests and similar). Although I do remember trying out the fix on a clean 0.44 project and it worked.

**Comment by ceyhuno** [comment]
@rodriguesbreno Changed `import "RCTValueAnimatedNode.h"` to `import <React/RCTValueAnimatedNode.h>` and it works now 👍   Updated the PR, thank you!

**Comment by msand** [comment]

**Comment by msand** [comment]
@shergin The root issue is that when using cocoapods the React Native modules/frameworks are all built as one and thus can't reference/import from each other using the namespaced global/system imports (Header Search Paths), as they're not separate pods but just sub-specs of one pod. Instead, in this case, the files need to be imported locally from the User Header Search Paths. https://stackoverflow.com/a/1044399/1925631 Importing locally breaks the non-cocoapods builds, thus requiring the if statement to handle both cases. https://clang.llvm.org/docs/LanguageExtensions.html#include-file-checking-macros
```
#if __has_include(<React/RCTValueAnimatedNode.h>)
#import <React/RCTValueAnimatedNode.h>
#else
#import "RCTValueAnimatedNode.h"
#endif
```

PR #14684, opened 2017-06-22
Support for preventing swipe left or swipe right

lprhodes:
The primary use case for this is disabling swipe right when using react-navigation's swipe to go back feature.

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableRow.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableRow.js
@@ -70,6 +70,8 @@ const SwipeableRow = React.createClass({
   propTypes: {
     children: PropTypes.any,
     isOpen: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


lprhodes:
I'd appreciate suggestions on how to provide a solid test plan for this...it's a pretty minor change.

skevy:
cc @sahrens

migueloller:
There is already a way to prevent the "swipe to go back" functionality in React Navigation.

See [`gesturesEnabled`](https://reactnavigation.org/docs/navigators/stack#gesturesEnabled).

lprhodes:
@migueloller That isn't the intention. Swipe to go back should still work when used in conjunction with the experimental swipe-able cells. That's not currently the case. Hold on approval though please. I'd like to double check no further amends were made to this.

lprhodes:
Too late - @sahrens can you please look at and approve #14900. Apologies for the mess up.

PR #14688, opened 2017-06-22
Export `Animated.Node` for ease of type checking

jamesreggio:
## Motivation

At present, there's no uniform way of determining whether you've received an Animated value node which can be attached to an animated prop.

You can attempt to use `instanceof Animated.Value`, but this fails for interpolated values. You can use `instanceof Animated.Interpolation`, but this fails for the values that are returned from the math functions (e.g., `Animated.Add`).

This commit exposes the base type from which all of these value nodes derive. As such, it is now possible to do the following:

```js
import React, { Component, PropTypes } from 'react';
import { Animated } from 'react-native';

class Widget extends Component {
  static propTypes = {
    progress: PropTypes.instanceOf(Animated.Node).isRequired,
  }
  // ...
}
```

## Test Plan

Unnecessary for cosmetic change.


javache:
The code currently tries to google-chrome, which your changed would override for linux users to chromium. Both of them are valid, can you instead change the open logic to try the two variants somehow?

jamesreggio:
The tests are extremely flakey. Any advice on how to get them to pass? (I can guarantee that my change is not responsible for these failures.)

shergin:
Should we use `export type ...` construction for this?
https://flow.org/blog/2015/02/18/Import-Types/

jamesreggio:
@shergin, sorry for the delay.

`export type` gets stripped away during compilation. The reason I'm exposing this type is to enable type checking at runtime, via `instanceof`.

Any other hesitations about this?

PR #14703, opened 2017-06-23
Adds plugin install feature to `install` and `uninstall`

rozele:
<details>

</details>

## Motivation (required)

_What existing problem does the pull request solve?_

For react-native-windows (as well as other CLI plugins for react-native), the user should not be expected to know whether they need to use yarn or NPM to install a CLI plugin. We were running into issues where users installed react-native using yarn, but when following our tutorial for adding `rnpm-plugin-windows`, the CLI would break as the instructions currently say to use NPM. These plugins will allow us to have a simple tutorial for how to install the `rnpm-plugin-windows`, as well as any other RN CLI plugin. The plugins use the existing CLI utilities to check if the project is yarn or NPM based, and then install the specified plugin accordingly.

## Test Plan (required)

_A good test plan has the exact commands you ran and their output, provides screenshots or videos if the pull request changes UI or updates the website._

I tested this CLI plugin using sinopia. I made the changes, published react-native to sinopia, generated new projects from the package in sinopia (with yarn and NPM), and verified that the correct tool was used to install plugins in each case.

cc @grabbou

### DIFF
--- a/local-cli/install/install.js
+++ b/local-cli/install/install.js
@@ -18,26 +18,41 @@ const spawnOpts = {
 
 log.heading = 'rnpm-install';
 
-function install(args, config) {
+function install(args, config, options) {
   const name = args[0];
 
-  let res = PackageManager.add(name);
+  let res;
+  if (options.plugin) {
+    res = PackageManager.addDev(name);
+  } else {
+    res = PackageManager.add(name);
+  }
 
   if (res.status) {
     process.exit(res.status);
   }
 
-  res = spawnSync('react-native', ['link', name], spawnOpts);
-
-  if (res.status) {
-    process.exit(res.status);
+  if (!options.plugin) {
+    res = spawnSync('react-native', ['link', name], spawnOpts);
+    if (res.status) {
+      process.exit(res.status);
+    }
   }
 
-  log.info(`Module ${name} has been successfully installed & linked`);
+  if (options.plugin) {
+    log.info(`Plugin ${name} has been successfully installed`);
+  } else {
+    log.info(`Module ${name} has been successfully installed & linked`);
+  }
 }
 
 module.exports = {
   func: install,
   description: 'install and link native dependencies',
   name: 'install <packageName>',
+  options: [{
+    command: '--plugin',
+    description: 'signals that the target is a plugin',
+    default: false,
+  }],
 };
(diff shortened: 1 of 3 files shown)


rozele:
Recommendation from @grabbou is to make this an extension of the existing `install` command.

grabbou:
Thanks @rozele!

Yeah, I think this is a better fit for `install` - assuming we agree to ship it. 

Although I see the problem here, I am not sure if it's not a documentation issue and we should educate community that one doesn't have to use `npm` if he uses `yarn` already.

Could you elaborate a bit on what exactly was breaking? All in all, I assume all they did was:
```
yarn add react-native
npm install rnpm-plugin-xxx
```
right?

That shouldn't break though.

axemclion:
Would `add-plugin` also do `react native link` ? For other plugins, looks like the user is now required to do `react-native add-plugin` and `react-native link`

rozele:
@grabbou updated to use `install` and `uninstall`.
@axemclion the point of the `--plugin` extensions to the `install` and `uninstall` command is to bypass the link command and save it to dev.

In fact, maybe instead of adding the `--plugin` flag, we could add a `--dev` flag and `--skip-link` feature... although that's quickly approaching what @axemclion said we should avoid (using react-native to resolve issues between yarn and NPM).

Of course, if we come up with a single command for downloading platform or CLI plugins (similar to Cordova's `cordova platform add <platformName>`), then we wouldn't need this, but I think this is useful in the near term.

rozele:
@grabbou They actually ran `react-native init myproj` and then `npm install rnpm-plugin-windows --save-dev`. The user had `yarn` installed on their system, so the `react-native-cli` chose to use `yarn`. On the subsequent call to `npm install ...`, a bunch of packages get uninstalled, breaking the `react-native-cli`.

So, the repro is basically
```
# install yarn
react-native init foo
cd foo
npm install --save-dev rnpm-plugin-windows
```
The output of the call to `npm install rnpm-plugin-windows` is:
```
+ rnpm-plugin-windows@0.2.5
added 28 packages, removed 621 packages and updated 66 packages in 40.262s
```
For whatever reason, 621 packages get removed just for trying to install an NPM package. Once that occurs, all commands on the CLI disappear, which unfortunately, people attribute to installing `rnpm-plugin-windows`, not to the incompatibility of yarn and NPM :-1:

axemclion:
@rozele Looks like this is true for any `npm install`, if there is yarn already installed on the system.

axemclion:
@rozele - Is the main cause of the problem, the fact that there is no npm package lock file ? Would simply adding a default npm package lock file in the project generated with `react-native init` do the job ?

Issue #14707, opened 2017-06-23
Infinite loop when passing {} as data for FlatList


JulioC:
### Description

If you pass a non-array object to `FlatList` `data` property, ReactNative javascript thread gets stuck in a infinite loop, due to `FlatList._getItemCount` returning `NaN`.

### Reproduction Steps

Pass any object that doesn't has `length` property to `FlatList`. 

#### Sample Code

https://snack.expo.io/BJrFCR9QZ

Simply: `<FlatList data={{}} />`

### Solution

Change `FlatList._getItemCount` to test for the `data.length`:

```js
_getItemCount = (data: ?Array<ItemT>): number => {
  return (data && data.length) ? Math.ceil(data.length / this.props.numColumns) : 0;
};
```

https://github.com/facebook/react-native/blob/master/Libraries/Lists/FlatList.js#L470

### Additional Information

* React Native version: "0.44.0",
* Platform: any

hramos:
@facebook-github-bot label icebox

strindhaug:
This but is certainly not fixed yet in 0.52.0

I noticed this bug because the unreliable API our customer had built randomly returned objects instead of arrays for some lists (probably because in PHP arrays are really what we'd call objects in javascript). And this froze the app completely each time a flatlist was fed an object.

Either the flatlist should tolerate objects and behave as if fed an empty array possibly logging a warning; or it should throw an exception so that the app can crash properly. Making the app unresponsive without an error message is not good.

And it's really annoying to have to safeguard every usage of a FlatList with a length check something like this:
```jsx
  {uglyApiResponse.listOfSomething.length > 0 ? (
    <FlatList
      data={uglyApiResponse.listOfSomething}
      keyExtractor={item => item.id}
      renderItem={({ item }) => (
        <SomethingLine data={item} />
      )}
    />
  ) : (
    <Text>
      ERROR: Invalid format of uglyApiResponse.listOfSomething expected array
    </Text>
  )}
```
just to avoid an unresponsive app when the API serves garbage

strindhaug:
Weird, i tried to make a synthetic test to force the problem now that the API is fixed, and it seems there is some kind of check when you try to supply a literal object like in the description here or a variable assigned to an object. (Mabe because it's somehow optimized away by the js engine when written like this?):
```jsx
// This seems to work ok:
const ExampleScreen = props => {
  return <FlatList data={{}} />;
};
```

```jsx
// This seems to work ok as well:
const ExampleScreen = props => {
[...]

PR #14708, opened 2017-06-23
Fixed FlatList._getItemCount returning NaN

JulioC:
If FlatList data prop is set to a non-array object, this method return `NaN` causing a infinite loop down the road

<details>

  - [ ] Explain the **motivation** for making this change.
  - [ ] Match the **code formatting** of the rest of the codebase.
  - [ ] Target the `master` branch, NOT a "stable" branch.

</details>

## Motivation (required)

https://github.com/facebook/react-native/issues/14707

## Test Plan (required)

See the Issue above.

### DIFF
--- a/Libraries/Lists/FlatList.js
+++ b/Libraries/Lists/FlatList.js
@@ -467,7 +467,7 @@ class FlatList<ItemT>
   };
 
   _getItemCount = (data: ?Array<ItemT>): number => {
-    return data ? Math.ceil(data.length / this.props.numColumns) : 0;
+    return (data && data.length) ? Math.ceil(data.length / this.props.numColumns) : 0;
   };
 
   _keyExtractor = (items: ItemT | Array<ItemT>, index: number) => {


sahrens:
It's required to be an array, why would it be otherwise? An invariant to require it is an array would be better for the dev experience.

JulioC:
Although I agree with you on the required type, the possibility that this function returns a NaN will trigger a infinite loop on the List rendering, which freezes the javascript thread, with no errors displayed.

The change is not that big and can help other developers.

JulioC:
Oh, sorry about that! I didn't know what you meant by `invariant`. Would it be better to open another PR or add another commit to this one?

smallpath:
when `this.props.numColumns`  went to 0, `_getItemCount` returns Infinity which crash app too. Maybe I can send another PR?

JulioC:
@smallpath if you want address both issues on your PR, just mention it here and I'll close this one

Issue #1882, opened 2015-07-06
[TextInput] Get keydown event to handle "Enter"


JonasJonny:
Hello.
I tried to catch keydown/keypress events for several hours with no luck.
I can get whole .text string (onChange={}) but what if I need to do an action when "Enter/Backspace" are pressed?
Is there any solution how to know which keyboard character was just pressed in TextInput keyboard?

Thank you.

chirag04:
Not sure if ``controlled={true}`` can help you here.

JonasJonny:
No luck with ``controlled={true}`` but thank you @chirag04.
Attributes onKeyDown/Up/Pressed didn't work but I know they are not supported. 
Event for onChange have just target, text, timestamp :(.

brentvatne:
I would like to see `onKey*` implemented, @sahrens do you have any suggestions for where to get started on implementing this? Maybe @JonasJonny or another interested contributor would be able to take that and turn it into a PR

JonasJonny:
Unfortunately I don't have enough skill to help.

grabbou:
@brentvatne I think the way you emit events for your `video` component should work quite nicely in this case.

dsibiski:
@JonasJonny I have a working implementation of `onKeyPress` that successfully detects `return` and `delete`.

Right now it is returning the character of the key that was pressed and for `return` and `delete`, it's returning ASCII codes `CR` and `DEL` respectively.

Would this take care of your use case? @brentvatne any other ideas as to how the `return` and `delete` codes should be returned to JS? Would it be better to use ASCII code numbers?

I'd like to figure these things out before submitting a PR.

dsibiski:
Ah, it looks like there really isn't a `delete` key in iOS, it's more like a `backspace`, so it should send back `BS` or `08`

JonasJonny:
Thank you @dsibiski for following. I really appreciate that.

In my opinion you should return "keyCode" http://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes

brentvatne:
Agreed with @JonasJonny - I think we should emulate the modern browser version of this: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key - so it would return {key: "Backspace"} and [some other properties too](https://developer.mozilla.org/en-US/docs/Web/Events/keypress) (probably for now we can just do key and implement others on a per-use-case basis)

dsibiski:
[...]

PR #12301, opened 2017-02-09
[TextInput] Implements 'onKeyPress' for Android - PR fixes

adasiewiczr:
Link to original PR : https://github.com/facebook/react-native/pull/10665

It seems that author of the pull request from the link above has finished work on his PR, so I did a fork of his repository and I applied requested changes.

An attempt at the Android implementation. Attempted to mirror the iOS implementation by supporting 'Enter' and 'Backspace'.

PR #14720, opened 2017-06-25
Implement onKeyPress Android

joshyhargreaves:
## Motivation

This implements onKeyPress for Android on TextInputs and addresses https://github.com/facebook/react-native/issues/1882. 
**N.B. that this PR has not yet addressed hardware keyboard inputs**, but doing will be fairly trivial. The main challenge was doing this for soft keyboard inputs. 

I've tried to match the style as much as I could. Will happily make any suggested edits be they architectural or stylistic design (edit: and of course implementation), but hopefully this is a good first pass :). 
I think important to test this on the most popular keyboard types; maybe different languages too.
I have not yet added tests to test implementation, but will be happy to do that also. 

## Test Plan
- Build & run RNTester project for Android and open TextInput. 
- Enter keys into 'event handling' TextInput. 
- Verify that keys you enter appear in onKeyPress below the text input
- Test with autocorrect off, on same input and validate that results are the same. 

Below is a gif of PR in action. 
onkeypressandroid


ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputInputConnection.java
```diff
@@ -0,0 +1,155 @@
...
+ * and then commits the user input of space or return afterwards, as a secondary commit within the batch edit. We of course
+ * want the keyPress entered by the user, so we take the second of these two commits as the keyPress.
+ *
+ * A final case is the case whereby a user has committed some text, and their cursor comes straight
+ * after the word they have just committed with no trailing space, as is the default behavior. If a user
+ * is to input a character as to start a new word, the stock IME will first commit a space to the
+ * input, and then set the composing text to be the character the user entered. In this case we
+ * choose our onKeyPress to be the new composing character.
+ */
+ class ReactTextInputInputConnection extends InputConnectionWrapper {
```
shergin: space character before `class ...`

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -81,14 +83,15 @@
...
+  private final InputConnectionWrapper mInputConnectionWrapper;
   private final InternalKeyListener mKeyListener;
   private boolean mDetectScrollMovement = false;
 
   private ReactViewBackgroundDrawable mReactBackgroundDrawable;
 
   private static final KeyListener sKeyListener = QwertyKeyListener.getInstanceForFullKeyboard();
 
-  public ReactEditText(Context context) {
+  public ReactEditText(Context context, InputConnectionWrapper inputConnection) {
```
joshyhargreaves: 'totally n00b in Android': that makes two of us then ;). Thank-you yes, good point!

joshyhargreaves:
@shergin, thank-you for taking a look at this, very much appreciated :). 

I actually have only just seem this [ticket](https://github.com/facebook/react-native/pull/12301) on the `Better <TextInput>' board, already triaged. They have taken the approach of detecting the key by changes to the text itself, just like the iOS [implementation](https://github.com/facebook/react-native/commit/6c7c845145e1da44ac5d965d49097eff1c11cc49). This was the first route I went down, but found that TextWatching really didn't give you much in the way of helping you distinguish between different types of input, be they autocorrection's, cut/paste etc. Although it's hard to say at this point if key detection as done in this PR is too hard/a good idea with the potential wide variety of behavior from different IMEs. 

I will say, you do at least need to use the InputConnection override for detecting the 'empty delete' case I mentioned, for detecting deletes which result in no text changes. Maybe even this could be used in combination with a `TextWatcher`, to reason about if text changes were results of changes from the soft keyboard.

I do think this is a really fiddly problem, and I'd say that the implementation handles pretty much every case (with stock Android keyboard). But perhaps we've ended up with a solution that's overfitted, and we'll have to go more generic solution.

joshyhargreaves:
**Status update**
I think it can be quite easy to get bogged down in implementation details sometimes, instead of thinking bigger picture!

So I think it's worth asking, 'what is the real value add' with implementing `onKeyPress` on Android for users? The thing that users cannot currently do currently on RN Android is detect 'empty deletes' or distinguish between key presses and cut and/or pastes. This we can do natively, and this is what I think we should expose through this API. We should support the same features as `onKeyPress` on iOS: 'Backspace' & 'Enter' & any other key press events should only be triggered by keyboard input and not cut/copy & we should support the empty delete case.

From further investigation with SwiftKey, what we cannot do natively is know what key was pressed to 'complete' a word; we do not get that granularity. So given we cannot do this natively, (or at least very easily at all, see the distinct lack of stack overflow answers on the topic), we should not worry about this use case; RN on iOS also has these same issues.

PardeepK:
@joshyhargreaves , with this implementation, we won't get onKeyPress event for keys like up/down/left/right or hardware keyboard's ENTER key.
So why don't we use native Android control EditText.onKeyup() (or KeyListener.onKeyUp()) to propagate key event to JS? That way we will also get this event if user press non-printable keys like up/down/left/right arrow keys?

PardeepK:
@joshyhargreaves also is there any timeline by when we will get this merged to master?

joshyhargreaves:
@pardeepk, onKeyUp events only work with hardware keyaboards as of ~Android 4.1, so not possible. We can of course still forward the events for hardware keyboards. This would be unlike the iOS inplementstion which works by watching the text.

I also think I’ve made too many assumptions about how different IMEs work based on the behaviour of the Android built in keyboard. For one, I don’t think you can assume beginBatchEdit is going to be called before any input events, and it’s up to the IME to call this, but not a requirement.

I have no timeline on when this will be merged; I think it needs to be integrated upon a few more times. Ultimately I think the behaviour RN users want but do not currently have is th ability to detect delete key presses on an empty text field, and perhaps this is the real value add of implementing this API and is what we should mainly be focussing upon as I don’t think this API is really possible to implement in the same way as the web.

PardeepK:
Thanks @joshyhargreaves for quick reply!
I need onKeyPress event for h/w keyboard up/down arrow key. I was investigating and found that in vanilla (non-react) Android app, I get EditText::onKeyDown/onKeyUp() callback for all keys for both s/w and h/w keyboard. But when I am using react-native TextInput  in Android app, I am not getting these onkeyDown() onKeyUp() methods invoked. any reason?
Also onKeyDown/onKeyUp() methods in InternalKeyListener (inner class in ReactEditText) are not getting invoked when user presses left/right/up/down key on h/w keyboard. I am not sure where these keyEvents are getting consumed. In fact I added these onKeyUp/onKeyDown methods in ReactRootView class, those also not getting invoked when I used h/w keyboard.

Appreciate any help!

joshyhargreaves:
Copied from stackoveflow answer: from http://developer.android.com/reference/android/view/KeyEvent.html

*As soft input methods can use multiple and inventive ways of inputting text, there is no guarantee that any key press on a soft keyboard will generate a key event: this is left to the IME's discretion, and in fact sending such events is discouraged. You should never rely on receiving KeyEvents for any key on a soft input method. In particular, the default software keyboard will never send any key event to any application targetting Jelly Bean or later, and will only send events for some presses of the delete and return keys to applications targetting Ice Cream Sandwich or earlier. Be aware that other software input methods may never send key events regardless of the version. Consider using editor actions like IME_ACTION_DONE if you need specific interaction with the software keyboard, as it gives more visibility to the user as to how your application will react to key presses.*

I have set up a vanilla project to test this and have observed that onKeyPress etc so perhaps the above is why you’re not seeing the events? I didn’t try arrow keys actually so they might still be sent by the IME and maybe that’s what you’re seeing? Hmm, not sure why you’d be seeing differences actually I haven’t observed different bevahour in the different cases. I have noticed subtle differences when autocorrect is on and off, so that might be worth checking.

No problem, I’d like to see this solved too!

PardeepK:
@joshyhargreaves any recommendation/help how can we capture hardware keyboard key press for TextInput (ReactEditText) RN control on Android?

shergin:
@joshyhargreaves Are you personally satisfied with this?
Could you please resolve conflicts?

joshyhargreaves:
@shergin sorry I missed the email mention!

Yes I can fix up the PR, sorry this fell off my radar with a talk I was doing! But now I'm free to finish this up; thank you for the continued input, it'd be great to get this landed!

So @shergin, I'm not quite happy with this exact approach, it's a solution which is definitely overfit to the stock Android stock keyboard IME lifecycle events and could be a bit fragile with other keyboards as a result. I think the solution is to make the logic a bit more basic (and cover more keyboards) by using a `TextWatcher` implementation similar to the `onKeyPress` iOS implementation. 

The key thing I think we should enable as part of this PR is backspace detection on an empty text-input, as that's something that users need. For example: https://github.com/jwohlfert23/react-native-tag-input/issues/31. Would you agree?

So I think I'll aim for a combined approach of what I have here for backspace on empty text-input detection, along with simplified text-watcher logic.

joshyhargreaves:
@shergin, it looks like the logic of can be a lot simpler by using TextWatcher; but using an the InputConnection to catch the ‘empty deletes’. Hopefully this should be fairly trivial, but looking into detecting when a text input update came from a cut/paste, but fingers crossed that should be easy enough

joshyhargreaves:
Looking to update this PR after the weekend, will keep everyone posted.

joshyhargreaves:
@shergin I implemented the logic using the `TextWatcher` API and I am happy with the resultant functionality! I haven't updated this PR, as I may swing back to fixing up the original method given now I have a greater understand of the finer details of the implementation: https://github.com/facebook/react-native/compare/master...joshyhargreaves:implement_edittext_onkeypress_android_textwatcher_rework?diff=unified&name=implement_edittext_onkeypress_android_textwatcher_rework

Functionality wise, it seems to work as I'd expect with both SwiftKey and stock Android keyboard :).  However it turns out that the logic needed to be a bit more hacky (as you can probably see) to implement the API. We know less about how the input was changed with the `TextWatcher` approach, so we need to have some additional state around whether it was a cut/paste which changed the input. 

Regardless of this, I'm positive that with a bit more refinement & testing, we will have something soon that should be good enough to use!

joshyhargreaves:
@shergin I revised the methodology with the original implementation; code is a lot cleaner than the alternate TextWatcher approach I linked above. 

I have tested on SwiftKey and stock Android keyboard, and it seems to work well :).
keypress_demo

joshyhargreaves:
@lako90, would you also mind having a look?

joshyhargreaves:
@shergin, would you mind having another look if you have a chance? 😊

elliscwc:
@lako90, @shergin really appreciate it if this gets into RN 0.50 release

joshyhargreaves:
@elliscwc, would you mind doing some testing of this PR and see if you find any bugs? :)

joshyhargreaves:
@hramos, could you maybe look into asking somebody to review this? Thank you :).

elliscwc:
@shergin .

joshyhargreaves:
@hramos, sorry to ask again!

elliscwc:
Any updates?

DarrylD:
Any movement on this?

joshyhargreaves:
@DarrylD I'm happy to whatever needs to be done to get this merged. @shergin @hramos could you provide some guidance on this?

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -88,7 +88,9 @@ public String getName() {
 
   @Override
   public ReactEditText createViewInstance(ThemedReactContext context) {
-    ReactEditText editText = new ReactEditText(context);
+    ReactTextInputInputConnection inputConnection = new ReactTextInputInputConnection(null, true, context);
```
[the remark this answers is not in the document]
joshyhargreaves: Thank you, very good point!

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputInputConnection.java
```diff
@@ -0,0 +1,129 @@
...
+
+import android.view.KeyEvent;
+import android.view.inputmethod.InputConnection;
+import android.view.inputmethod.InputConnectionWrapper;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.uimanager.UIManagerModule;
+import com.facebook.react.uimanager.events.EventDispatcher;
+
+class ReactTextInputInputConnection extends InputConnectionWrapper {
+  public static final String NewLineRawValue = "\n";
```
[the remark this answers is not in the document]
joshyhargreaves: Thanks!

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputInputConnection.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.annotation.Nullable;
+
+import android.view.KeyEvent;
+import android.view.inputmethod.InputConnection;
+import android.view.inputmethod.InputConnectionWrapper;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.uimanager.UIManagerModule;
+import com.facebook.react.uimanager.events.EventDispatcher;
+
+class ReactTextInputInputConnection extends InputConnectionWrapper {
```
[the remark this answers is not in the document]
joshyhargreaves: Sounds like a more sensible name, thank you :P.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputInputConnection.java
```diff
@@ -0,0 +1,129 @@
...
+  public static final String NewLineRawValue = "\n";
+  public static final String BackspaceKeyValue = "Backspace";
+  public static final String EnterKeyValue = "Enter";
+
+  private @Nullable ReactEditText mEditText;
```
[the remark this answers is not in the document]
joshyhargreaves: Thanks, with your suggested changes, `@Nullable` no longer necessary!

joshyhargreaves:
@kmagiera made the changes you suggested, let me know what you think.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
```diff
@@ -0,0 +1,164 @@
...
+  private boolean mIsBatchEdit;
+  private @Nullable String mKey = null;
+
+  public ReactEditTextInputConnectionWrapper(
+      InputConnection target,
+      boolean mutable,
+      final ReactContext reactContext,
+      final ReactEditText editText
+  ) {
+    super(target, mutable);
```
kmagiera: As we are no longer updating `target` we can just default to `mutable=false`. Also there seem to be no need to expose `mutable` as an argument of your subclass constructor (ReactEditTextInputConnectionWrapper). I recommend to replace this call with:
```
super(target, false);
```
And remove `boolean mutable` from the list of constructor arguments.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
```diff
@@ -0,0 +1,164 @@
...
+    if(mKey != null) {
+      dispatchKeyEvent(mKey);
+      mKey = null;
+    }
+    return super.endBatchEdit();
+  }
+
+  @Override
+  public boolean setComposingText(CharSequence text, int newCursorPosition) {
+    final int previousSelectionStart = mEditText.getSelectionStart();
```
kmagiera: can you drop all the `final` keywords here. It is not common to use `final` in this context in Java (unlike in JS where you'd use `const`). Usually `final` is only used for scoped vars when they are being referenced in anonymous classes ("lambda classes") – in which case you get a compiler error if you forget to use `final`

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public boolean beginBatchEdit() {
+    mIsBatchEdit = true;
+    return super.beginBatchEdit();
+  }
+
+  @Override
+  public boolean endBatchEdit() {
+    mIsBatchEdit = false;
+    if(mKey != null) {
```
kmagiera: space: `if(mKey` -> `if (mKey`

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
joshyhargreaves: Good spot, thanks.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
joshyhargreaves: Thank you!

joshyhargreaves:
@kmagiera made the changes!

edo1493:
Can we make this happen before Christmas? 🙏

edo1493:
Hey @joshyhargreaves and @shergin, what's the status on this? 🤓

joshyhargreaves:
@kmagiera and I are looking to move this forward after the holiday period.

brentvatne:
@facebook-github-bot shipit

PR #14723, opened 2017-06-26
Update NativeModulesAndroid.md

sunnylqm:


javache:
Fixed in https://github.com/facebook/react-native/pull/14835

PR #14728, opened 2017-06-26
Update docs to use pure components when possible 

pontusab:
## Motivation
Update the docs to follow best practises using pure components when possible where nor state and lifecycle methods are used.

## Test Plan

Go to `website` directory and run `npm start` 

Screenshot:
screen shot 2017-06-26 at 13 40 10

### DIFF
--- a/docs/DirectManipulation.md
+++ b/docs/DirectManipulation.md
@@ -98,25 +98,17 @@ Composite components are not backed by a native view, so you cannot call
 import React from 'react';
 import { Text, TouchableOpacity, View } from 'react-native';
 
-class MyButton extends React.Component {
-  render() {
-    return (
-      <View>
-        <Text>{this.props.label}</Text>
-      </View>
-    )
-  }
-}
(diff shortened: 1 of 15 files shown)


hramos:
Thanks for the PR! I think back then we decided against following this pattern in order to avoid confusion early on (not everyone reading the docs may be familiar with React). That said I am open to using these in the docs in some way. I want to make sure people understand why it's possible to write components in this way before we show examples using this pattern.

ms88privat:
These are named `functional components` or? Because there is also `React.PureComponent`

charpeni:
I think you've mixed pure and stateless components, because these are stateless.

Stateless components are re-rendered every time they receive props, so that's not really a "best practice", just a shortcut to get something more readable. However, stateless components should be optimized in the future.

Pure components are components that extends `PureComponent` class, they've a mixin performing a shallow comparison of previous and next props in `shouldComponentUpdate`. There's the real performance gain and "best practice".

Issue #14712, opened 2017-06-23
[0.45.2] Commits to cherry-pick and use prop-types with React 16 instead


hramos:
`react@16` (a `peerDependency`) did away with the `PropTypes` export in favor of the `prop-types` module.

The issue has been fixed in master. @mojodna has indicated interest in getting this cherry-picked into the 0.45 release. Several issues have been opened that seem to be related to this, enough that I think it's worth cherry picking into 0.45.2.

hramos:
@mojodna do you have a list of the commits in master that need to be cherry picked into 0.45 in order to fix the issue?

I've also opened #14713 as we'll also need to cherry pick them into the 0.46 RC in order to make sure the stable release doesn't suffer from this next month.

mojodna:
I don't, but I'll go hunting now.

mojodna:
* 50b11aa09b479e5ce8b06de98e02ba1dd6a8cca3
* 1f3140c4962fbbf9f3a00f977daa64d2940c0c5b
* 8e9322c65e751dc8e7512874e03e7c4349a28b93

Ashoat:
One extant issue in master is that `Text` is [still implemented](https://github.com/facebook/react-native/blob/master/Libraries/Text/Text.js) with `React.createClass` (for the mixins functionality, I assume). Without this resolved React Native will error out on any attempt to use `Text`, so it should probably be considered blocking for 0.45.2.

It seems that 0.45.1 is pretty unusable in the meantime. Is there a recommended workaround, or should everybody downgrade to earlier versions? If we should downgrade, is it worth considering withdrawing the release, considering that it's unusable?

Ashoat:
Curiously, there have been folks reporting that they've been experiencing this issue intermittently, and I've noticed that two weeks ago I was able to use react-native@0.45.0-rc.2/react@16.0.0-alpha.12 without issue and now am not. It would seem that `react@16.0.0-alpha.12` should never have been usable with any version of React Native (including master which still has extant `React.createClass` references), but there are folks for whom this has worked. Perhaps it has something to do with how React Native or the packager resolves the 'react' package internally?

hramos:
@Ashoat generally we recommend that people use the release candidate, in order to surface issues like these before a stable release is cut. Since the createClass issue is still present in master, we just need someone to send a PR to fix it. We're less than a week away from a stable release cut, so ideally this would be done within the next 24 hours so that we can get it into the RC early enough for people to test.

[...]

PR #14729, opened 2017-06-26
Replace React.createClass with create-react-class

### LINKED ISSUE #14620: Warning: React.createClass is no longer supported.
When try ```react-native run-ios```, by using ```Debug JS Remotely```,  it showed those warning in Google Chrome developer tools.

```
ExceptionsManager.js:71 Warning: React.createClass is no longer supported.
 Use a plain JavaScript class instead. 
If you're not yet ready to migrate, 
create-react-class is available on npm as a drop-in replacement. (https://fb.me/migrating-from-react-create-class)
```

> https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.createclass

By searching ```React.createClass``` in ```react-native``` folder, there are many parts used it.

For example, in ```Image.android.js```, line 79:

```
var Image = React.createClass({
  propTypes: {
    ...ViewPropTypes,
    style: StyleSheetPropType(ImageStylePropTypes),
```

* React Native version: 0.45.1
* Platform: iOS
* Development Operating System: macOS 10.12.5
* Build tools: Xcode 8.3.2

mojodna:
This replaces all uses of `React.createClass` with `createReactClass` from the `create-react-class` package, attempting to match use of `var` and `const` according to local style.

Fixes #14620
Refs #14712


mojodna:
Done in fb5a63e

hramos:
Having trouble landing this internally. Won't this require `create-react-class` to be added to `package.json`?

mojodna:
`create-react-class` is already present: https://github.com/facebook/react-native/blob/master/package.json#L165

I just rebased against master, removing the changes in `ReactNativeFiber-dev.js` and `ReactNativeStack-dev.js` (there are warning messages that reference `React.createClass` but only in strings).

hramos:
Alright, let me dig in a bit more into these failures.

grabbou:
@hramos were you able to indicate why it never landed?

grabbou:
Tip: bump `react-docgen` in `website` to `3.0.0-beta5` so it supports `create-react-class` package. I believe this is source of errors. I had to do the same for 0.46-stable.

hramos:
Apologies for the delay, just came back from a long holiday weekend. Taking a look again. I've been kicking the internal land every couple of days and finally got past some flaky tests. The only thing holding it back right now is a failure during the generation of the website which I haven't been able to reproduce on a local check out of this PR. I'm importing the latest changes (post merge with master) to try again.

grabbou:
Sounds good, thanks @hramos for taking care of that. In the meantime, I have sent 0.47.0-rc.0 build and planning to release rc.1 with that fix.

0.46.0-stable already has it cherry-picked.

hramos:
This is close to landing. While create-react-class was already present in package.json, it wasn't yet in use in React Native proper, so I had to make some changes internally to make sure the new dep was loaded correctly by other projects at Facebook. There's only one more issue to resolve and I'll be able to land this.

hramos:
This should be landing shortly. Props to @skevy for pointing me in the right direction to resolve the failed Jest snapshot tests that were holding this PR back.

When `createReactClass` is used, a `displayName` property must be defined in order for Jest to generate the snapshots correctly. I've opened https://github.com/facebook/react/pull/10122 to fix the React docs.

PR #14731, opened 2017-06-26
Add FORCE/SKIP_BUNDLING flags for iOS builds

cooperka:
### Motivation

*See discussion below for updated motivation.*

Anything running in Debug should use the packager anyway; no need to bundle. This saves a **huge amount of time** during development when testing things like push notifications that require a real device.

The code being modified was originally moved here in https://github.com/facebook/react-native/commit/9ae3714f4bebdd2bcab4d7fdbf23acebdc5ed2ba to make sure bundles are always created in `Release`, but the change can be applied to real devices, too. Ideally there should be very little difference in how a simulator is treated compared to a physical device.

### Test Plan

Run a Debug build in Xcode targeting a physical device before and after this commit.

You can use the `FORCE_BUNDLING` and `SKIP_BUNDLING` flags to manually change the default behavior. For example, under **Build Phases** > **Bundle React Native code and images**:

```bash
export SKIP_BUNDLING=true
../node_modules/react-native/packager/react-native-xcode.sh
```

### DIFF
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -17,11 +17,15 @@ fi
 
 case "$CONFIGURATION" in
   *Debug*)
-    if [[ "$FORCE_BUNDLING" ]]; then
-      echo "FORCE_BUNDLING enabled; bundling even in Debug."
+    if [[ "$PLATFORM_NAME" == *simulator ]]; then
+      if [[ "$FORCE_BUNDLING" ]]; then
+        echo "FORCE_BUNDLING enabled; continuing to bundle."
+      else
+        echo "Skipping bundling in Debug for the Simulator (since the packager bundles for you). Use the FORCE_BUNDLING flag to change this behavior."
+        exit 0;
+      fi
     else
-      echo "Skipping bundling in Debug (since the packager bundles for you). Use the FORCE_BUNDLING flag if you want to change this behavior."
-      exit 0;
+      echo "Bundling for physical device. Use the SKIP_BUNDLING flag to change this behavior."
     fi
 
     DEV=true
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -10,11 +10,21 @@
 # This script is supposed to be invoked as part of Xcode build process
 # and relies on environment variables (including PWD) set by Xcode
 
+if [[ "$SKIP_BUNDLING" ]]; then
+  echo "SKIP_BUNDLING enabled; skipping."
+  exit 0;
+fi
+
 case "$CONFIGURATION" in
   *Debug*)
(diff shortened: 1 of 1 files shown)


javache:
Everyone has different workflows for this. Can you instead make this script look for an environment variable that can be set by your xcodeproj, e.g. SKIP_BUNDLING?

cooperka:
Thanks for the feedback @javache, I've rebased and added the flag you mentioned. I like this even better now.

javache:
Cool! Let's keep the old behaviour too when you don't specify either of these options.

cooperka:
@javache is there a particular reason you want to keep the old behavior? I'm happy to do so, but to me it seems buggy that it would need to bundle every time just for the device. By default (e.g. after using CRNA) there doesn't seem to be any reason why a physical device would need special treatment.

[The original commit](https://github.com/facebook/react-native/commit/9ae3714f4bebdd2bcab4d7fdbf23acebdc5ed2ba) that caused the code to be like this doesn't seem to care whether it's a simulator or physical device; what's relevant is `Debug` mode. Curious to hear your thoughts. Thanks!

javache:
Because we shouldn't assume that people will see this commit and enable FORCE_BUNDLING. The current version is a good set of defaults: e.g. you want to install a quick debug build on your phone so you can play it when disconnected from your computer, so let's keep them.

cooperka:
@javache updated. That reasoning makes sense.

emusgrave:
@cooperka Have you tested the SKIP_BUNDLING flag after a clean of your build folder? I don't think it will work unless you have hard-coded the IP of your development workstation somewhere else.

If running on a physical device, the bundling step is providing an [`ip.txt`](https://github.com/facebook/react-native/blob/d07279c978994db07d229b8d19700c83d80db4fa/scripts/react-native-xcode.sh#L103) file in the bundle which is required in order to figure out what IP address to use to connect to the packager.
If that file isn't found, then it erroneously attempts to connect to `localhost`, which in this case would be the iOS device.

I think this means that even if you are loading the bundle from the packager, you still need a bundle to be included in your ipa so that it can find the packager's IP address.

I was able to get around this by still running the section of code in `react-native-xcode.sh` to create the `ip.txt` even if SKIP_BUNDLING was true. I wanted to run this concept by you before I make a pull request.

cooperka:
@emusgrave thanks for the detailed context. I didn't dig that deep because it seemed to work for me, but if that file is removed during cleaning, I can see how that would cause problems. Feel free to submit a PR 💯

antoinerousseau:
why is this not included by default in the HelloWorld template of the `.pbxproj` file?
```
if [ "${CONFIGURATION}" == "Debug" ]; then
  export SKIP_BUNDLING=true
 fi
```
I mean, why would a new project create the bundle when in dev mode?

PR #14773, opened 2017-06-29
Update copyToClipBoard.js

saigrandhi:
Since copy to clipboard functionality is now available in Linux, the comment above `copyToClipboard` function has been updated.


hramos:
Can we remove the entire line instead? We probably don't need to call this out if all three platforms we expect people to use are supported.

saigrandhi:
I have removed the comment specifying different platforms.

PR #14775, opened 2017-06-29
Support native ViewManager inheritance on iOS

cbrevik:
**Motivation**
This is a re-worked version of #14260, by @shergin's suggestion.

For iOS, if you want to inherit from a native ViewManagers, your custom ViewManager will not automatically export the parents' props. So the only way to do this today, is to basically copy/paste the parent ViewManager-file, and add your own custom logic.

With this PR, this is made more extensible by exporting the `baseModuleName` (i.e. the iOS `superclass` of the ViewManager), and then using that value to re-establish the inheritance relationship in `requireNativeComponent`.

**Test plan**
I've run this with a test project, and it works fine there. But needs more testing.

Opened this PR as [per @shergin's suggestion](https://github.com/facebook/react-native/pull/10946#issuecomment-311860545) though, so we can discuss approach.

**Discussion**
* Android already supports inheritance, so this change should be compatible with that. But, not every prop available on `UIManager.RCTView.NativeProps` is actually exported by every ViewManager. So should `UIManager.RCTView.NativeProps` still be merged with `viewConfig.NativeProps`, even if the individual ViewManager does not export/use them to begin with?
* Does this break other platforms? [UWP](https://github.com/Microsoft/react-native-windows)?

### DIFF
--- a/Libraries/ReactNative/requireNativeComponent.js
+++ b/Libraries/ReactNative/requireNativeComponent.js
@@ -70,13 +70,19 @@ function requireNativeComponent(
     viewConfig.propTypes = null;
   }
 
-  // The ViewConfig doesn't contain any props inherited from the view manager's
-  // superclass, so we manually merge in the RCTView ones. Other inheritance
-  // patterns are currenty not supported.
-  const nativeProps = {
-    ...UIManager.RCTView.NativeProps,
-    ...viewConfig.NativeProps,
-  };
+  let baseModuleName = viewConfig.baseModuleName;
+  let nativeProps = { ...viewConfig.NativeProps };
+  while (baseModuleName) {
+    const baseModule = UIManager[baseModuleName];
+    if (!baseModule) {
+      warning(false, 'Base module "%s" does not exist', baseModuleName);
+      baseModuleName = null;
+    } else {
+      nativeProps = { ...nativeProps, ...baseModule.NativeProps };
+      baseModuleName = baseModule.baseModuleName;
+    }
+  }
+
   for (const key in nativeProps) {
     let useAttribute = false;
     const attribute = {};
--- a/React/Views/RCTComponentData.m
+++ b/React/Views/RCTComponentData.m
@@ -430,7 +430,7 @@ static RCTPropBlock createNSInvocationSetter(NSMethodSignature *typeSignature, S
     @"propTypes": propTypes,
     @"directEvents": directEvents,
     @"bubblingEvents": bubblingEvents,
-    @"baseModuleName": superClass == [NSObject class] ? (id)kCFNull : moduleNameForClass(superClass)
+    @"baseModuleName": superClass == [NSObject class] ? [NSNull null] : moduleNameForClass(superClass)
   };
 }
 
--- a/React/Views/RCTComponentData.m
+++ b/React/Views/RCTComponentData.m
@@ -41,7 +41,7 @@ typedef NSMutableDictionary<NSString *, RCTPropBlock> RCTPropBlockDictionary;
     _viewPropBlocks = [NSMutableDictionary new];
     _shadowPropBlocks = [NSMutableDictionary new];
 
-    _name = [self moduleNameForClass:managerClass];
+    _name = moduleNameForClass(managerClass);
 
     _implementsUIBlockToAmendWithShadowViewRegistry = NO;
     Class cls = _managerClass;
@@ -430,7 +430,7 @@ static RCTPropBlock createNSInvocationSetter(NSMethodSignature *typeSignature, S
     @"propTypes": propTypes,
     @"directEvents": directEvents,
     @"bubblingEvents": bubblingEvents,
-    @"baseModuleName": superClass != [NSObject class] ? [self moduleNameForClass:superClass] : [NSNull null]
+    @"baseModuleName": superClass == [NSObject class] ? (id)kCFNull : moduleNameForClass(superClass)
   };
 }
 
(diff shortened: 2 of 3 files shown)


React/Views/RCTComponentData.m
```diff
@@ -439,11 +423,14 @@ - (void)setProps:(NSDictionary<NSString *, id> *)props forShadowView:(RCTShadowV
...
 #endif
-
+  
+  Class superClass = [_managerClass superclass];
+  
   return @{
     @"propTypes": propTypes,
     @"directEvents": directEvents,
     @"bubblingEvents": bubblingEvents,
+    @"baseModuleName": superClass != [NSObject class] ? [self moduleNameForClass:superClass] : [NSNull null]
```
shergin: nit: Please avoid "negative logic": Changing `!=` to `==` will make this code easier to read.

Libraries/ReactNative/requireNativeComponent.js
```diff
@@ -70,13 +70,19 @@ function requireNativeComponent(
...
   }
 
-  // The ViewConfig doesn't contain any props inherited from the view manager's
-  // superclass, so we manually merge in the RCTView ones. Other inheritance
-  // patterns are currenty not supported.
-  const nativeProps = {
-    ...UIManager.RCTView.NativeProps,
-    ...viewConfig.NativeProps,
-  };
+  let baseModuleName = viewConfig.baseModuleName;
```
shergin: Does using `do { } while ();` loop make this code a bit cleaner/easier-to-read?

cbrevik: `do { } while ();` would make it execute the statement at least once, which we don't want to do if there is no `baseModuleName` exported. So not sure if that is a good idea.

cbrevik:
@shergin Android does not expose a `baseModuleName`, but since it already supports inheritance this code will not directly break because of that. That is to say, Android already exports most of the props through `viewConfig.NativeProps`, without having to merge with `UIManager.RCTView.NativeProps`. 

The question is of course if it should still merge with `UIManager.RCTView.NativeProps`, when each ViewManager does not explicitly implement/export all the ones available on `RCTView`.

shergin:
It does support but how? Does Android use the same approach which we reject for iOS?
Okay, but current/old code does merge `UIManager.RCTView.NativeProps` and `viewConfig.NativeProps` for Android as well, right (line 76)? And seems the new code breaks it. No?

cbrevik:
Yeah, the current logic does merge for Android as well. My question is if it should do that though?

For Android, using [JSONDiff](http://www.jsondiff.com/) to compare what for example
[`viewConfig.NativeProps for AndroidTextInput`](https://gist.github.com/cbrevik/d6c63a480420351ea75e0cd5c96ff69d) contains vs. [`Androids UIManager.RCTView.NativeProps`](https://gist.github.com/cbrevik/4abe42bec48455f6f61ac42f5595072a). There you see it is missing a few props from `RCTView` on `AndroidTextInput`, but most are already there. This is because the inheritance chain from the native side (Java) already exports the inherited props (from parent ViewManagers) to `viewConfig.NativeProps`.

Whilst if, for iOS, you compare [`viewConfig.NativeProps for RCTTextView`](https://gist.github.com/cbrevik/c2872341d424696b09ad61191b2f670f) vs [`iOS' UIManager.RCTView.NativeProps`](https://gist.github.com/cbrevik/7603ef2d61868ead38a019f38a9e93f1), none of the props from `RCTView` already exists on `viewConfig.NativeProps`, so there we have to merge the two.

So for Android it is missing a few from `RCTView`, but should we still merge the two if for example `AndroidTextInput`'s ViewManager (or its parents) does not implement the missing props?

cbrevik:
For example the `accessible` prop is missing from `AndroidTextInput`. That prop is implemented on [`ReactViewManager.java`](https://github.com/facebook/react-native/blob/9a51fa8e15acc052b5d271c369dfa4b227d657c2/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java#L53). 

But [`ReactTextInputManager.java` does not extend `ReactViewManager`](https://github.com/facebook/react-native/blob/f954f3d9b674b13977f722bc3b8dc6c1b99fe6c7/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L66). So should `accessible` still be merged onto `AndroidTextInput`'s NativeProps? 

I grant that I might have misunderstood the logic here. 🤔

shergin:
In this case, I think we should do this:
 1. Split the logic here between iOS and other platforms. (And implement new logic only for iOS.)
 2. Land it. And continue investigate benefits from this solution on iOS.
 3. (In separate diff/PR.) We have to try remove merging only for Android (another split). And see will it break something or not and how it will affect performance.
 4. Try to implement same (as on iOS) logic on Android (`baseClassName` and co). And see how it will affect performance. (It should improve TTI because it should significantly reduce amount of transferred data across the bridge. @javache Pieter, what do you think?)
 5. Encourage WP/VR guys to do same thing for another platforms.
 6. Celebrate wins. 🎉

@javache We need your opinion and blessing here. :)

React/Views/RCTComponentData.m
```diff
@@ -455,4 +442,26 @@ - (RCTViewManagerUIBlock)uiBlockToAmendWithShadowViewRegistry:(NSDictionary<NSNu
   return nil;
 }
 
+- (NSString *)moduleNameForClass:(Class)managerClass
```
javache: This method doesn't depend on any state, make it a static C-function.

React/Views/RCTComponentData.m
javache: Use `(id)kCFNull` instead of `[NSNull null]`

javache:
Code looks good to me, thanks for providing those JSON diffs! I don't understand yet why Android behaves differently here, we should probably understand that before we land this.

PR #14778, opened 2017-06-29
[Text] allow wordWrapping on iOS and tvOS

JulienKode:
The goal of this pull request is to add wordWrapping on iOS and tvOS 

I have test it on iOS and tvOS like that: 

```
<Text ellipsizeMode="wordWrapping">
    {this.props.text}
</Text>
```

Word wrapping is available in [RCTConvert.m](https://github.com/facebook/react-native/blob/master/React/Base/RCTConvert.m#L276)

### DIFF
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -111,10 +111,11 @@ const Text = React.createClass({
      * - `tail` - The line is displayed so that the beginning fits in the container and the
      * missing text at the end of the line is indicated by an ellipsis glyph. e.g., "abcd..."
      * - `clip` - Lines are not drawn past the edge of the text container.
+     * - `wordWrapping` - Wrapping occurs at word boundaries, unless the word itself doesn’t fit on a single line.
      *
      * The default is `tail`.
      *
-     * > `clip` is working only for iOS
+     * > `clip` and `wordWrapping` works only for iOS and tvOS
      */
     ellipsizeMode: PropTypes.oneOf(['head', 'middle', 'tail', 'clip', 'wordWrapping']),
     /**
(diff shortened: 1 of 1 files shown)


hramos:
Can you add screenshots? Just want to make sure this works on iOS.

JulienKode:
Hello,

This is some screenshot with and without wordWrapping:

WordWrapping

Default mode

JulienKode:
Yes, let met show you a little example:

If we have this text: 
`let test = "Hello I am a test\nHello I am a test REACTNATIVEPOWER";`

With wordWrapping, you can see the last word was cut (this is the behavior of [NSLineBreakByWordWrapping](https://developer.apple.com/documentation/uikit/nslinebreakmode/nslinebreakbywordwrapping?language=objc)):
WordWrapping

With tail this is the bahavior: 
Tail


You can see the code of the component: 

```
	<Text
		ellipsizeMode={"wordWrapping"}
		numberOfLines={2}
		style={styles.instructions}
	>
		{test}
	</Text>
</View>
```

shergin:
Yes, `ellipsizeMode={'wordWrapping'}` is supported, but the problem here is that `wordWrapping` is not `ellipsizeMode` conceptually; it is something like `wrapMode`.
Initially conversion from `ellipsizeMode={'wordWrapping'}` to NSLineBreakByWordWrapping was not implemented intensionally, but lately someone add this.

So, we have to implement a new prop for this functionality.

JulienKode:
Hello,

Thanks for your answer, how many wrapMode there are ?

shergin:
I can imagine another values, but we can start from this.

PR #14780, opened 2017-06-29
Identify keyboardDismissMode platform-specific options

jkdf2:
Similar to `TextInput`'s `returnKeyType`, comments allow to see at a glance which options are cross-platform and which are not.

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -199,16 +199,21 @@ const ScrollView = React.createClass({
     /**
      * Determines whether the keyboard gets dismissed in response to a drag.
      *
+     * *Cross platform*
+     *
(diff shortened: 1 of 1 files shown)


chirag04:
@hramos

hramos:
Inline comments like these are not visible in the autodocs. Perhaps edit the text above where the parameters are documented as well?

chirag04:
i think the intent here is to help someone reading the code. 

For documentation i think it's documented above already. `On android this is not supported and it will have the same behavior as 'none'`.

hramos:
I see. I still think the documentation can be improved here. The older docs are a bit vague - they don't clearly state that interactive is iOS only, for example.

jkdf2:
> Inline comments like these are not visible in the autodocs. Perhaps edit the text above where the parameters are documented as well?

Updated to clarify in both code & docs now @hramos

PR #14789, opened 2017-06-30
Prevent a crash when no cursor drawable is set

gpeal:
## Summary
React Native 0.43 added additional functionality to setSelectionColor that also tints the cursor drawable of the View. However, some views may not have a cursor drawable set in which case, the code will crash when attempting to load a drawable with resource id 0.

We encountered this in our RN 0.45 upgrade in the Airbnb app.

@lelandrichardson


shergin:
Why should not 'catch' (try-catch) this exception as we catch others at the end of the functions?

gpeal:
@shergin It's not really an exception. It's a valid state for it to be 0 so it would be counter intuitive to let it throw.

gpeal:
@shergin For the record, the exception that would be thrown is [NotFoundException](https://developer.android.com/reference/android/content/res/Resources.NotFoundException.html)

gpeal:
@shergin bump

PR #14799, opened 2017-07-01
Wrong location of sdk tools on OSX in documentation ('Getting Started')? 

pixelpax:
On OSX, the path to the cli tools seem to be in `$ANDROID_HOME/tools/bin`, not `$ANDROID_HOME/tools`

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -400,7 +400,7 @@ Add the following lines to your `$HOME/.bash_profile` config file:
 
 ```
 export ANDROID_HOME=$HOME/Library/Android/sdk
-export PATH=$PATH:$ANDROID_HOME/tools
+export PATH=$PATH:$ANDROID_HOME/tools/bin
 export PATH=$PATH:$ANDROID_HOME/platform-tools
 ```
 


javache:
The main tool you want is `android` which is at `$ANDROID_HOME/tools/android`

pixelpax:
@javache : 

I hear what you're saying, and I wish you were correct because the `android` command is my preferred way of interacting with the sdk. However, this is no longer the supported way; the `android` is command is deprecated. Resources which are linked to on the react-native documentation now recommend that developers use `$ANDROID_HOME/tools/bin/avdmanager` and `$ANDROID_HOME/tools/bin/sdkmanager`

Download the latest sdk and run `android` from the path you describe, you will see the following output:

```
*************************************************************************
The "android" command is deprecated.
For manual SDK, AVD, and project management, please use Android Studio.
For command-line tools, use tools/bin/sdkmanager and tools/bin/avdmanager
*************************************************************************
Invalid or unsupported command ""

Supported commands are:
android list target
android list avd
android list device
android create avd
android move avd
android delete avd
android list sdk
android update sdk
```

I should've communicated this in my original PR, my apologies.

javache:
Given that people may have older and newer versions of the SDK installed, let's just put both on the path for now?

PR #14800, opened 2017-07-02
Add Socky app to showcase.json

paalex:
Adding to the showcase Socky - A React Native messaging app for Kids with Autism


paalex:
@ericnakagawa

hramos:
Nice app! As it happens, we're not adding any new apps here for now, sorry.

PR #14811, opened 2017-07-03
Fixes hasProperty returning true even when there's an empty getter.

SudoPlz:
Fixes: https://github.com/facebook/react-native/issues/14810

Problem described in more detail here: https://stackoverflow.com/questions/43967489/could-not-get-unknown-property-repositoryurl-for-project


SudoPlz:
Closing because no reviewers have been found is not a good reason to close this.

PR #14826, opened 2017-07-04
Better doc for 'data:' uri scheme & mandatory size

harry-g:
It is quite confusing that the 'data:' uri scheme is not documented, but working. 
It is very useful when getting e.g. an icon from a REST call.

Also, the mandatory size style for network & data images should be mentioned here, not only in the image guide.


harry-g:
This and other PR builds currently seem to be failing because of a file system issue:
    lsof: WARNING: can't stat() vmhgfs file system /Volumes/VMware Shared Folders

PR #14834, opened 2017-07-05
default value for javaScriptEnabled is false

koenpunt:
This did cost me quite some time, but it turns out the default value for `javaScriptEnabled` isn't `true`, but `false`. So without it, any JavaScript injected through `injectJavaScript`/`injectedJavaScript` won't ever make it to the webview.

Ref: https://developer.android.com/reference/android/webkit/WebSettings.html#setJavaScriptEnabled(boolean)

### DIFF
--- a/Libraries/Components/WebView/WebView.ios.js
+++ b/Libraries/Components/WebView/WebView.ios.js
@@ -280,7 +280,7 @@ class WebView extends React.Component {
 
     /**
      * Boolean value to enable JavaScript in the `WebView`. Used on Android only
-     * as JavaScript is enabled by default on iOS. The default value is `true`.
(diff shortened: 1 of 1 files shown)


charpeni:
I guess that the default value was referring to `defaultProps`, not Android one.

```javascript
static defaultProps = {
  javaScriptEnabled : true,
  scalesPageToFit: true,
};
```

https://github.com/koenpunt/react-native/blob/master/Libraries/Components/WebView/WebView.android.js#L188

koenpunt:
@charpeni probably, but the default prop apparently isn't passed to the native view?

koenpunt:
So my change might not be applicable, but there's an issue with it nonetheless.

PR #14844, opened 2017-07-05
Clarity in pagingEnabled description for ScrollView

seanyusa:
Current description is misleading (without looking at implementation) to believe that both horizontal and vertical pagination are supported on both platforms. This comment clarifies that vertical pagination is not supported on Android.


skevy:
@facebook-github-bot  shipit

hramos:
We couldn't find a particular reason why shipit failed here. Importing manually.

PR #14847, opened 2017-07-06
Enable eslint during Circle CI

TheSavior:
Running this here to verify that circle-ci will pass.

From:
```
✖ 8074 problems (64 errors, 8010 warnings)
```

To:
```
✖ 80 problems (0 errors, 80 warnings)
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,4 +3,5 @@
 **/staticBundle.js
 **/main.js
 Libraries/vendor/**/*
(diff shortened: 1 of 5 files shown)


hramos:
This seems fine, but I'll let others who have more experience with Flow take a look.

TheSavior:
Please review the internal diff instead. D5374369. This was only meant to be a test to ensure that circle ci passed successfully with the change.

PR #14860, opened 2017-07-06
[WebView] Update docs: comments in injected JS cause injection to fail

koenpunt:
This caused some headaches...

Might not be the correct wording, so happy to get some feedback.


hramos:
Does this only happen on Android for sure?

hramos:
We've moved docs out of the JS files.

PR #14880, opened 2017-07-07
[website] Remove unused npm dependencies and add missing ones

ide:
A couple of npm dependencies are unused (grepped for them or their variants since some are Babel presets/plugins) and there were also some missing that are listed in react-native's package.json but are directly required by the website.

Test plan: follow the steps in the website README, open the website, browse a few different types of pages

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -6,11 +6,7 @@
   },
   "dependencies": {
     "babel-core": "^6.6.4",
-    "babel-plugin-external-helpers": "^6.5.0",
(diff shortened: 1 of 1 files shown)


hramos:
This requires manual import as we'll need to update the offline node modules cache in the internal repo. I've already made the necessary changes and am waiting for tests to run. Ping me if it's been a few days and I have not landed this.

hramos:
Landing.

PR #13517, opened 2017-04-15
[npm] Remove unused npm packages and move CI dependency to devDependencies

ide:
There are several npm packages that are no longer used and do not show up when grepping the code base. Also the "async" dependency is used only in CI and should be moved to devDependencies.

Test Plan: Load the UIExplorer. Run Travis CI, Circle CI, and FB internal tests.

PR #14881, opened 2017-07-07
[cleanup] Remove remnants of the packager

ide:
There were still some references to "packager/" that are no longer used since the `packager` directory has been deleted after moving to Metro. Cleaned up the ones that were doing nothing and updated the references that are still meaningful.

Test plan: Ran `npm start`, CI tests

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,4 +3,3 @@
 **/staticBundle.js
 **/main.js
 Libraries/vendor/**/*
-packager/src/worker-farm/**/*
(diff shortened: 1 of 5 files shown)


cpojer:
Woohoo, you rock!

`package.json` also still has a bunch of deps that are only needed in Metro. Did you wanna take a look at those and remove the ones that are not needed in RN directly?

cpojer:
(separate PR).

ide:
@cpojer yes I have a PR out for that too, though maybe it ran into merging problems: #13517

PR #14883, opened 2017-07-07
[ci] Upgrade the version of Yarn to 0.27.5 for CI

ide:
The TravisCI tests were using an especially old version of Yarn. Upgrade to 0.27.5 so we're using a version closer to the version that more people (and especially RN maintainers) would be using.

Test plan: Look at the OSS CI tests; make sure there are no Yarn / node_module errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,8 @@ install:
   # Use node 6 because that is what runs on land-blocking tests
   - nvm install 6
   - rm -Rf "${TMPDIR}/jest_preprocess_cache"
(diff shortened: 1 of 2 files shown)


javache:
Any reason we shouldn't use the latest release of yarn always (e.g. `brew install yarn --ignore-dependencies`). It seems like we'll forget to update this anyway.

ide:
@javache Nice suggestion, let's give it a try. I kept the Yarn version in the Docker setup static because I think it's intended for people to run it on their own machines, and erring on the side of having fewer moving parts is more helpful in that context I think.

ide:
Travis E2E tests pass so I think this worked, I don't see any Yarn-related errors.

PR #14884, opened 2017-07-07
Fixed ImageBackground could't be wrapped by Touchable* component

patw0929:
In 0.46, as warning's advice, I use `ImageBackground` to replace `Image`s which used as background image wrapper, and in some cases, I also wrap `ImageBackground` with `TouchableHighlight` to make it clickable.

But here comes an error:

> Touchable child must either be native or forward setNativeProps to a native component

So I pick some code from `Image.ios.js` into `ImageBackground.js` to solve it.

Please help to review, thanks!

### DIFF
--- a/Libraries/Image/ImageBackground.js
+++ b/Libraries/Image/ImageBackground.js
@@ -16,6 +16,7 @@ const Image = require('Image');
 const React = require('React');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
+const ensureComponentIsNative = require('ensureComponentIsNative');
 
 /**
  * Very simple drop-in replacement for <Image> which supports nesting views.
@@ -42,11 +43,24 @@ const View = require('View');
  * ```
  */
 class ImageBackground extends React.Component {
+  setNativeProps(props: Object) {
+    if (this._viewRef) {
+      ensureComponentIsNative(this._viewRef);
+      this._viewRef.setNativeProps(props);
+    }
+  }
+
+  _viewRef: View;
+
+  _captureRef = ref => {
+    this._viewRef = ref;
+  };
+
   render() {
     const {children, style, imageStyle, imageRef, ...props} = this.props;
 
     return (
-      <View style={style}>
+      <View style={style} ref={this._captureRef}>
         <Image
           {...props}
           style={[


javache:
React.createClass is deprecated. Can you instead implement setNativeProps on this class and forward it to the child view?

patw0929:
Got it, I have revised it at https://github.com/facebook/react-native/pull/14884/commits/9383abe9b56102297f87535747e0f44776c94c90.
Thank you.

Libraries/Image/ImageBackground.js
```diff
@@ -42,11 +42,23 @@ const View = require('View');
  * ```
  */
 class ImageBackground extends React.Component {
+  setNativeProps(props: Object) {
+    if (this._viewRef) {
+      this._viewRef.setNativeProps(props);
```
javache: Can you call `ensureComponentIsNative` here?

patw0929: Replaced it at https://github.com/facebook/react-native/pull/14884/commits/f94ff87698781194d90d939933028c23468b6444. Is that right?

Libraries/Image/ImageBackground.js
```diff
@@ -42,11 +43,24 @@ const View = require('View');
  * ```
  */
 class ImageBackground extends React.Component {
+  setNativeProps(props: Object) {
+    if (this._viewRef) {
+      ensureComponentIsNative(this._viewRef);
```
shergin: Just curious, why do we need this? We know that <View> is "native".

Libraries/Image/ImageBackground.js
```diff
@@ -42,11 +43,24 @@ const View = require('View');
...
+  _captureRef = ref => {
+    this._viewRef = ref;
+  };
+
   render() {
     const {children, style, imageStyle, imageRef, ...props} = this.props;
 
     return (
-      <View style={style}>
+      <View style={style} ref={this._captureRef}>
```
shergin: N00b question: Will this conflict with ref prop existed in this.props (<ImageBackground ref={...}>...</ImageBackground>)?

Libraries/Image/ImageBackground.js
```diff
@@ -42,11 +43,24 @@ const View = require('View');
...
+  setNativeProps(props: Object) {
+    if (this._viewRef) {
+      ensureComponentIsNative(this._viewRef);
+      this._viewRef.setNativeProps(props);
+    }
+  }
+
+  _viewRef: View;
+
+  _captureRef = ref => {
```
shergin: I think we harmlessly can inline it. No?

Libraries/Image/ImageBackground.js
javache: Ah good point, we're always rendering View. It doesn't hurt to check.

Libraries/Image/ImageBackground.js
javache: Nope, because those go on Image, not View

Libraries/Image/ImageBackground.js
javache: There's some difference with function binding, which can be expensive. Doing it like this will guarantee it's only bound once I believe.

PR #14885, opened 2017-07-07
troubleshooting docs

formatlos:
I added troubleshooting docs for connection to the development server from device, otherwise it's maybe hard to find out why it's not working.


skevy:
@facebook-github-bot shipit

hramos:
The importer will refuse to land a PR if a commit is added right before a shipit command is made. This is to prevent unreviewed changes from making it in (even though 7e4af4f was made by the same person who ran shipit). I'll try landing again.

hramos:
This won't land due to a disparity in what GitHub's API returns when we attempt to import the PR. A quick workaround would be to open a new PR with the same commits. Otherwise, I can find sometime in the next month to apply the changes manually in the internal repo.

formatlos:
As suggested I opened a new PR https://github.com/facebook/react-native/pull/15304

PR #14886, opened 2017-07-07
[PushNotificationIOS] Rename _remoteNotificationCompleteCalllbackCalled to _remoteNotificationCompleteCallbackCalled

koenpunt:
Just noticed this typo


hramos:
That's unfortunate. To avoid introducing a breaking change, would you consider leaving the misspelled method in place, but with a deprecation warning?

koenpunt:
Is that necessary? since it's an internal method

koenpunt:
Ehm, wait, it isn't even a method, just a property

hramos:
Ah, you're right. Landing.

PR #14891, opened 2017-07-07
FlatList doesn't specify that it accepts ScrollView Props

tomasreimers:
ListView does specify this: https://facebook.github.io/react-native/docs/listview.html#props

The link in ListView is generated by the documentation generator, but because the format of FlatList's props is different, it fails to catch it. Consider finding a way to specify this explicitly to the documentation generator (although this would be a bigger change).

### DIFF
--- a/Libraries/Lists/FlatList.js
+++ b/Libraries/Lists/FlatList.js
@@ -303,6 +303,7 @@ type DefaultProps = typeof defaultProps;
  * - By default, the list looks for a `key` prop on each item and uses that for the React key.
  *   Alternatively, you can provide a custom `keyExtractor` prop.
  *
+ * Also inherets [ScrollView Props](docs/scrollview.html#props).
  */
 class FlatList<ItemT> extends React.PureComponent<
   DefaultProps,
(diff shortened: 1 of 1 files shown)


skevy:
@facebook-github-bot shipit

shergin:
Are you sure that FlatList accepts ScrollView props?

tomasreimers:
I think so, but am not positive. 

Here is where FlatList passes its own props to VirtualizedList: 

https://github.com/facebook/react-native/blob/352c3d97b3017ab91cfca15a8d54fb753edc7ed5/Libraries/Lists/FlatList.js#L557

And here is where VirtualizedList passes its own props to ScrollView: 

https://github.com/facebook/react-native/blob/352c3d97b3017ab91cfca15a8d54fb753edc7ed5/Libraries/Lists/VirtualizedList.js#L662

https://github.com/facebook/react-native/blob/352c3d97b3017ab91cfca15a8d54fb753edc7ed5/Libraries/Lists/VirtualizedList.js#L676

https://github.com/facebook/react-native/blob/352c3d97b3017ab91cfca15a8d54fb753edc7ed5/Libraries/Lists/VirtualizedList.js#L761

(The one exception to this being nested views with same orientation I suppose, so I can include that as an update)

shergin:
Even of FlatList does propagate own props down to ScrollView (technically and/or partially), I don't think that it can be considered as public API. And it is not even Flow typed.
cc @sahrens

tomasreimers:
@shergin totally valid point. I'll leave it up to the maintainers for what they want to do on this on. If we choose not to allow those types then, there are a variety of props/methods we should add (i.e. all the onScroll methods).

PR #14895, opened 2017-07-07
Fix type of StatusBar backgroundColor so it is displayed correctly in docs

brentvatne:
# Before

screen shot 2017-07-07 at 2 54 01 pm

# After

screen shot 2017-07-07 at 2 55 08 pm

# Notes

The change that set this type to $FlowFixMe was a [massive one-year-old old commit](https://github.com/facebook/react-native/commit/a2fb703bbb988038323c55b29b40e8f5ff52966d) and I don't think it is necessary anymore (flow passes with this change). If it does, it must be something internal to Facebook, maybe we should $FlowFixIt somehow?

### DIFF
--- a/Libraries/Components/StatusBar/StatusBar.js
+++ b/Libraries/Components/StatusBar/StatusBar.js
@@ -63,7 +63,10 @@ type DefaultProps = {
 /**
  * Merges the prop stack with the default values.
  */
-function mergePropsStack(propsStack: Array<Object>, defaultValues: Object): Object {
+function mergePropsStack(
+  propsStack: Array<Object>,
+  defaultValues: Object
+): Object {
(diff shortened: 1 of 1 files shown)


Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -141,7 +153,7 @@ class StatusBar extends React.Component {
   props: {
     hidden?: boolean,
     animated?: boolean,
-    backgroundColor?: $FlowFixMe,
+    backgroundColor?: string,
```
chirag04: `ColorPropType` maybe? idk if it's too late. @hramos already imported.

chirag04: ah damn! sorry. flow type for a prop type..so meta!

PR #14898, opened 2017-07-07
MaskedViewIOS -- A way to apply alpha masks to views on iOS

skevy:
## Description

It's very important in complex UIs to be able to apply alpha channel-based masks to arbitrary content. Common use cases include adding gradient masks at the top or bottom of scroll views, creating masked text effects, feathering images, and generally just masking views while still allowing transparency of those views.

The original motivation for creating this component stemmed from work on `react-navigation`. As I tried to mimic behavior in the native iOS header, I needed to be able to achieve the effect pictured here (this is a screenshot from a native iOS application):

iOS native navbar animation

In this image, there are two masks:

- A mask on the back button chevron
- A gradient mask on the right button

In addition, the underlying view in the navigation bar is intended to be a UIBlurView. Thus, alpha masking is the only way to achieve this effect.

Behind the scenes, the `maskView` property on `UIView` is used. This is a shortcut to setting the mask on the CALayer directly.

This gives us the ability to mask any view with any other view. While building this component (and testing in the context of an Expo app), I was able to use a `GLView` (a view that renders an OpenGL context) to mask a `Video` component!

I chose to implement this only on iOS right now, as the Android implementation is a) significantly more complicated and b) will most likely not be as performant (especially when trying to mask more complex views).

## Test Plan (required)

Review the `<MaskedViewIOS>` section in the RNTester app, observe that views are masked appropriately.

example


Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -0,0 +1,170 @@
...
+type Props = ViewProps & {
+  /**
+   * Only a single child is supported.
+   */
+  children: React.Element<*>,
+  /**
+   * Should return a React element to be rendered and applied as the
+   * mask for the child element.
+   */
+  renderMask: () => React.Element<*>,
```
javache: Why not pass this as an Element directly, instead of wrapping it an a function?

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -0,0 +1,170 @@
...
+
+    if (typeof this.props.renderMask !== 'function') {
+      if (!this._hasWarnedInvalidRenderMask) {
+        console.warn(
+          'MaskedView: Invalid `renderMask` prop was passed to MaskedView. ' +
+            'Expected a function. No mask will render.'
+        );
+        this._hasWarnedInvalidRenderMask = true;
+      }
+      return <View style={this.props.style}>{target}</View>;
```
javache: Why wrap this in another view? This way you're not passing all View props through, it would be better for the caller to immediately set all props on `target`

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -0,0 +1,170 @@
...
+      maskElementWithRef = cloneReferencedElement(maskElement, {
+        ref: this._handleMaskViewRef,
+      });
+    }
+
+    return (
+      <RCTMaskedView
+        style={this.props.style}
+        maskRef={this._maskViewNodeHandle}>
+        <View
```
javache: Can you set these properties from the native side, when you render the mask?

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -0,0 +1,170 @@
...
+      </RCTMaskedView>
+    );
+  }
+
+  _handleMaskViewRef = (node: React.Element<*>) => {
+    if (!node) {
+      return;
+    }
+    if (this.state.maskViewNodeRef !== node) {
+      this.setState({ maskViewNodeRef: node });
```
javache: Don't use state for this, it will cause an immediate re-render. Just set an instance variable.

javache:
The native side of this can be much simpler. Instead of passing refs and numbers through, just provide a custom view from your ViewManager that knows that it will have exactly 2 subviews, the first being the target and the second being the mask.

In your custom view, override `didUpdateReactSubviews` to insert the first subview and set the second subview as mask.

skevy:
Awesome @javache. Thanks so much for the review. I love the `didUpdateReactSubviews` solution -- thanks for the tip!

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: The reason I wrapped the mask in a view is so that I didn't have to do anything weird to get Yoga to lay it out appropriately. If I just create the wrapper view on the native side, layout doesn't occur as I want it to.

However, I see the value of simplifying the JS side -- so if you have thoughts on how to make sure the mask is still laid out properly while creating it's container natively, I'm happy to do that. It just wasn't clear to me how to make that happen, and this seemed like the best way.

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: 👍

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: Truth. Will pass as an Element.

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: Refs aren't needed at all anymore given your suggestion below. 👍

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
ide: Are there cases where MaskedView may have information that affects the rendering of the mask view that would make sense to pass in as args to the render function? (This is the rationale for renderScrollComponent() for composing scrollable components.)

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
javache: Do you need the mask to participate in the layout process? If you're going the custom subview route, you may be able to just skip it completely.

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: @ide I think no, for now. I think just passing in an element is fine.

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: I believe we do -- we want arbitrary views to "just work" as mask elements. Putting them inside an absolutely positioned View inside of the parent allows for intuitive styling of the Mask element by the user I think.

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -0,0 +1,110 @@
...
+            'Expected a React Element. No mask will render.'
+        );
+        this._hasWarnedInvalidRenderMask = true;
+      }
+      return <View {...otherViewProps}>{children}</View>;
+    }
+
+    return (
+      <RCTMaskedView {...otherViewProps}>
+        <View pointerEvents="none" style={[StyleSheet.absoluteFill]}>
```
javache: No need to wrap StyleSheet.absoluteFill in an array.

React/Views/RCTMaskedView.h
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <UIKit/UIKit.h>
+#import "RCTView.h"
```
javache: Use qualified imports in headers, `<React/RCTView.h>`

React/Views/RCTMaskedView.m
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#import "RCTMaskedView.h"
+#import "UIView+React.h"
+
+@implementation RCTMaskedView
+
+- (void)didUpdateReactSubviews
+{
+  [super didUpdateReactSubviews];
```
javache: -[UIView didUpdateReactSubviews] is very simple. It may be better to not call super here, avoid the `maskView removeFromSuperview` and just iterate over `reactSubviews` from index 1?

That also avoid a whole set of window events for the maskView subtree.

skevy: We won't be able to use `removeClippedSubviews` on the mask view this way, but I suppose that's probably ok?

React/Views/RCTMaskedView.h
skevy: 👍

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
skevy: Whoops! 👍

skevy:
Addressed the nits. Assuming we want to go forward with the simplified `didUpdateReactSubviews`, this is probably good to land?

skevy:
Thanks @javache. Addressed the nits. Assuming we want to go forward with the simplified `didUpdateReactSubviews`, this is probably good to land?

React/Views/RCTMaskedView.m
javache: Yeah, if you want clipped subviews you can always wrap the children of your MaskedView yourself.

PR #14900, opened 2017-07-08
Fixes unintended side effects caused by #14684

lprhodes:
I had fixed this locally but hadn't updated the original pull request, sorry. This commit is working for us in production.


hramos:
Fix the lint warnings and we can merge this.

PR #14901, opened 2017-07-08
imageSource should not be required

lprhodes:
Making the imageSource required was causing prop type warnings. Using SwipeableRow without an image works fine in production. Ideally we'd additionally not render the <Image /> component. A task for another time.


shergin:
Having a problem in a third party component is a weak motivation for such kind of change.

lprhodes:
What third party component? I'm trying to use this code directly. Swipeable rows shouldn't have to have an image, it's an odd requirement.

shergin:
SwipebleRow is the third party component, we have to fix problem in the component, not in Image.

lprhodes:
This isn't making a change in Image. The component is a part of react-native in Experimental.

lprhodes:
The component being changed is SwipeableQuickActionButton

shergin:
Oops, you are right. 😄 I am sorry. I was mislead by the title of PR.

PR #14902, opened 2017-07-08
No need for divider width

lprhodes:
At the moment there's an odd 4px gap between buttons. This hasn't been necessary for anything to work so I would assume this is to fix an issue that no longer exists. This amend is working in production.


hramos:
Seems like a small change but as I'm not sure why the value was used in the first place, I'm hesitant to approve. Is there anyone who might be able to provide some context?

shergin:
Could you provide a screenshot before and after? And if we don't need the gap anymore we have to remove whole 'divider' style.

PR #14904, opened 2017-07-08
Only update visible length on scroll if greater than 0

tomasreimers:
Observed behavior where if you set `initialScrollIndex` to some number outside the initially rendered frames (i.e. frame 11), the onScroll event (triggered by the internal call of scrollToIndex in componentDidMount) will return a layout size of 0. This causes all the content to be hidden until the point at which the user scrolls again which triggers another update with the correct length.

If you don't do this it causes problems for computeWindowedRenderLimits which believes only one cell will ever be rendered at a time.


tomasreimers:
Cannot reproduce. Lost in the sands of time.

PR #14906, opened 2017-07-08
Replacing `react-native android` with `react-native eject`

Grivel-l:
Just updated the documentation. To create the android directory, `react-native android` doesn't work anymore, it has been replaced by `react-native eject`.

### DIFF
--- a/local-cli/runAndroid/runAndroid.js
+++ b/local-cli/runAndroid/runAndroid.js
@@ -27,7 +27,7 @@ function checkAndroid(root) {
  */
 function runAndroid(argv, config, args) {
   if (!checkAndroid(args.root)) {
-    console.log(chalk.red('Android project not found. Maybe run react-native android first?'));
+    console.log(chalk.red('Android project not found. Maybe run react-native eject first?'));
     return;
   }
 


hramos:
Looks like your PR is missing a description.

Grivel-l:
Thank you, fixed it

local-cli/runAndroid/runAndroid.js
```diff
@@ -27,7 +27,7 @@ function checkAndroid(root) {
  */
 function runAndroid(argv, config, args) {
   if (!checkAndroid(args.root)) {
-    console.log(chalk.red('Android project not found. Maybe run react-native android first?'));
+    console.log(chalk.red('Android project not found. Maybe run react-native eject first?'));
```
hramos: react-native eject only applies when a project has been created using create-react-native-app. We still support using `react-native init` to create a project.

Grivel-l: I just tried and `react-native android` does not work with a project created with `react-native init` but `react-native eject` work well.

mkonicek:
The `android` command has been removed. As @hramos said, `eject` is not designed to work in projects created using `react-native init`.

Thanks for the PR, it's a good catch. It's probably easiest to remove the whole sentence "Maybe run react-native eject first?".

I'm curious, how did it happen that your project only had iOS native code? `react-native init MyApp` creates both iOS and Android native code. Is there a command to create iOS files only? If yes we might want to kill it - creating iOS and Android files keeps the CLI code simple and people can delete the `android` folder if they really don't want it. But there's no harm in always keeping the `android` folder around. An app that starts as iOS only might need an Android version in the future :)

Grivel-l:
Ahah, I just modified a lot of files in my Android directory and I wanted to get a fresh new one, so I simply deleted it.

hramos:
Wouldn't running `react-native start` re-create the folder anyway?

hramos:
Or maybe I'm getting confused with the `build/` folder inside `/android`.

Grivel-l:
Nop the only way I found is to do a `react-native eject`

hramos:
In any case, `react-native init` should build the `android` and `ios` folder for you. Closing the PR as the doc update is not necessary.

PR #14930, opened 2017-07-10

rajsuvariya:
There is conflict in project name between package.json and index.android.js. Because of it I was getting error.


hramos:
Looks like #14940 is a more complete version of this.

Issue #14869, opened 2017-07-06
`withCredentials` flag gets ignored on iOS


JonnyBurger:
### Is this a bug report?

Yes

### Have you read the Bugs section of the Contributing to React Native Guide?

Yes

### Environment

1. `react-native -v`: 0.46

Then, specify:

1. Target Platform: iOS
2. Development Operating System: macOS
3. Build tools: Xcode

### Description

In `react-native@0.44`, a flag `withCredentials` was introduced. In `react-native@0.46`, the behaviour was unintentionally changed by this commit: 047961fbf77cb012b53978184102e8ca3d00c7ec

`withCredentials = true` translates to `HTTPShouldHandleCookies = YES` in native code. After the commit that I linked above, the `Cookie` header gets set explicitly, which overrides any behaviour set by `HTTPShouldHandleCookies`.

### Steps to Reproduce

Make a `fetch` request with the flag `withCredentials: true`.

### Expected Behavior

The flag gets respected.
The cookie header only gets set explicitly if `withCredentials = false`

### Actual Behavior

The flag gets overridden.

### Reproducible Demo

Hard to make a demo, but the bug can be proven with the [Apple documentation](https://developer.apple.com/documentation/foundation/nsmutableurlrequest/1415485-httpshouldhandlecookies?preferredLanguage=occ):

> If your app sets the Cookie header on an NSMutableURLRequest object, then this method has no effect, and the cookie data you set in the header overrides all cookies from the cookie store.

[Here](https://github.com/facebook/react-native/blob/master/Libraries/Network/RCTNetworking.mm#L245) you can see that `HTTPShouldHandleCookies` is being set based on the `withCredentials` flag.

[Here](https://github.com/facebook/react-native/blob/master/Libraries/Network/RCTNetworking.mm#L233:L234) you can see that the `Cookie` header is being set explicitly, which takes precedence over `HTTPShouldHandleCookies`.

### Related issues
- withCredentials flag in XHRs should default to "true" #14063

### cc
@clozr

jozan:
We've hit this as well. Nice catch! :)

JonnyBurger:
Made a PR: #14931 

Reviews appreciated as Objective-C and especially Networking in it is not my strength.

cbjs:
0.46.2, still not working

clozr:
Sorry for late reply. back from vacation. It seems that withCredential flag and the fix to send the cookie by default conflicts each other. Let me analyze a bit more and will update.

[...]

PR #10575, opened 2016-10-27
Added Cookie Header to XML and Websocket request

clozr:
Continuation of Pull Request #7167

https://github.com/facebook/react-native/pull/7167

Needed to clean my repository. So created this Pull Request

PR #14931, opened 2017-07-10
Fix `withCredentials` flag for fetch requests

JonnyBurger:
This PR changes `RCTNetworking.mm` so that the cookies are managed using `HTTPShouldHandleCookies ` if the `withCredentials` flag is set to true.

This was the behavior in react-native@0.45, but was broken in react-native@0.46 where the cookies would be overwritten no matter what the `withCredentials` flag is set to.

This PR restores the behavior of the `withCredentials` flag from react-native@0.45, but still uses the new cookie management from react-native@0.46 if that flag is not set. 

I have documented the issue and explained why this code change is necessary here: #14869


JonnyBurger:
Can this be considered for 0.47? @grabbou

This fixes a bug introduced in 0.46.

chirag04:
cc @rigdern

jamesreggio:
Given the relative lack of interest, I'm going to propose a PR to revert #10575. Maybe that will spark up the conversation we need to have about these breakages.

JonnyBurger:
I agree with @jamesreggio actually, my PR is not the right way to do it.
I think either  #10575 should either be reverted or fixed so that cookies are only sent when `withCredentials` is true. Currently the cookies are always being included, but not using `request.HTTPShouldHandleCookies`.

jamesreggio:
@JonnyBurger, I opened a PR to revert the original breaking change in https://github.com/facebook/react-native/pull/16127.

If you agree, do you mind chiming in?

PR #14934, opened 2017-07-10
Be able to scroll to last item in list

maartenschumacher:
The invariant condition doesn't allow scrolling to the last index

Here is a simple reproduction of the problem: https://snack.expo.io/BJ59gzWrZ

Surely, if the last item is rendered, we should be able to scroll to it without an invariant failing?

Tested locally that this change fixes the issue.


shergin:
@facebook-github-bot shipit

PR #14940, opened 2017-07-11

rajsuvariya:
Conflict in the project name in package.json and index.xxx.js


hramos:
Waiting for CLA to be signed.

rajsuvariya:
When will this be merged?

hramos:
If all goes well, not too long from now. I've already accepted the PR and started the import process. GitHub is not the source of truth for React Native, so we need to wait for the patch to be applied internally first. Our systems will then sync out the changes back to GitHub.

PR #14943, opened 2017-07-11
[packager] Change `polyfills` to `getPolyfills` function for more configurability

ide:
Changing the `polyfills` option to `getPolyfills({platform})` will let us return a different set of polyfills for each platform. See https://github.com/facebook/metro-bundler/issues/25 for the motivation.

Test Plan: None so far, since this is an unimplemented property on Metro's side.

### DIFF
--- a/local-cli/server/runServer.js
+++ b/local-cli/server/runServer.js
@@ -157,7 +157,7 @@ function getPackagerServer(args, config) {
     maxWorkers: args.maxWorkers,
     platforms: defaultPlatforms.concat(args.platforms),
     polyfillModuleNames: config.getPolyfillModuleNames(),
-    polyfills: config.polyfills,
+    getPolyfills: config.getPolyfills,
     postMinifyProcess: config.postMinifyProcess,
     postProcessModules: config.postProcessModules,
     projectRoots: args.projectRoots,
--- a/local-cli/util/Config.js
+++ b/local-cli/util/Config.js
@@ -94,7 +94,7 @@ export type ConfigT = {
    * An optional list of polyfills to include in the bundle. The list defaults
    * to a set of common polyfills for Number, String, Array, Object...
    */
-  polyfills: Array<string>,
+  getPolyfills: ({platform: string}) => Array<string>,
 
   /**
    * An optional function that can modify the code and source map of bundle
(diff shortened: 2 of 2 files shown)


mjesun:
Looks like since the `require.resolve` call is now conditional, Flow requires the `$FlowFixMe` again. Could you add it on top of the dynamic require located on `local-cli/util/Config.js`? Then we can merge that. The `$FlowFixMe` looked like this: `//$FlowFixMe: necessary dynamic require`.

Thanks!

local-cli/server/runServer.js
```diff
@@ -157,7 +157,7 @@ function getPackagerServer(args, config) {
     maxWorkers: args.maxWorkers,
     platforms: defaultPlatforms.concat(args.platforms),
     polyfillModuleNames: config.getPolyfillModuleNames(),
-    polyfills: config.polyfills,
+    getPolyfills: config.getPolyfills,
```
mjesun: I'd suggest performing the call at this point; i.e. `polyfills: config.getPolyfills()`. The reason is that the `metro-bundler` machinery is already ready to accept a list of custom polyfills, but as an `Array<string>`, not as a `() => Array<string>`.

hramos:
Thanks for taking a look @mjesun. Feel free to merge once you approve the PR.

local-cli/server/runServer.js
ide: @mjesun The reason I would like to not call it here is so that Metro can call it from `Resolver#_getPolyfillDependencies` and pass the platform (ios or android) into `config.getPolyfills`. (In case it wasn't clear, getPolyfills is `({platform: string}) => Array<string>`.)

ide:
I updated the PR to pass Flow checks and also added the `{platform}` parameter to the default `getPolyfills` implementation since other config functions appeared to list out their parameters too.

mjesun:
Looks good to me; we'll switch the metro code to use the method. Thanks for taking a look at it @ide!

ide:
Thank you!

mjesun:
@ide Just FYI, we just landed some minor refactor, moving the polyfill list function into a separate file, so that it can be required from multiple places (including the RN CLI so that they're used when building) and because we will reuse that file internally :)

(See [here](https://github.com/facebook/react-native/commit/6dd9d1683392d568184f955192d63ecc377088ac#diff-c817eeb56a164e72724e0f2af01c7249)).

mjesun:
Continuation in metro-bundler: https://github.com/facebook/metro-bundler/commit/abf49f0a02c2279dc8bbea8f2c747e4c071a5767

PR #14946, opened 2017-07-11
Preserve new lines in message when sharing to WhatsApp

krizpoon:
When sharing to WhatsApp using Share.share in iOS, new line characters were converted to spaces.

Related SO question:
<https://stackoverflow.com/questions/30255923/uiactivityviewcontroller-not-passing-new-line-characters-to-some-activities>

By using a custom UIActivityItemProvider to convert `\n` to `<br/>` when WhatsApp was selected, we could send multi-line messages to WhatsApp.

** Test
1. Share a multi-line message:
`Share.share({ message: 'Line1\nLine2' });`

2. Select WhatsApp from the action sheet then select a recipient

3. The message should consist of two lines

### DIFF
--- a/Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.h
+++ b/Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.h
@@ -1,10 +1,11 @@
-//
-//  RCTMessageActivityItemProvider.h
-//  RCTActionSheet
-//
-//  Created by Poon Kwok Cheung on 11/7/2017.
-//  Copyright © 2017 Facebook. All rights reserved.
-//
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
 
 #import <UIKit/UIKit.h>
 
--- a/Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.h
+++ b/Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.h
@@ -1,0 +1,15 @@
+//
+//  RCTMessageActivityItemProvider.h
+//  RCTActionSheet
+//
+//  Created by Poon Kwok Cheung on 11/7/2017.
+//  Copyright © 2017 Facebook. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface RCTMessageActivityItemProvider : UIActivityItemProvider
+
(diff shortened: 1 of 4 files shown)


Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.m
```diff
@@ -0,0 +1,28 @@
+//
```
hramos: Please use the standard header here:

```
/**
 * Copyright (c) 2015-present, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 */
```

Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.h
```diff
@@ -0,0 +1,15 @@
+//
```
hramos: Please use the standard header here:

```
/**
 * Copyright (c) 2015-present, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 */
```

Libraries/ActionSheetIOS/RCTMessageActivityItemProvider.m
```diff
@@ -0,0 +1,28 @@
...
+
+#import "RCTMessageActivityItemProvider.h"
+
+@implementation RCTMessageActivityItemProvider
+
+
+- (id) activityViewController:(UIActivityViewController *)activityViewController itemForActivityType:(NSString *)activityType
+{
+  if([activityType isEqualToString:@"net.whatsapp.WhatsApp.ShareExtension"]) {
```
hramos: Can you add a space after `if` here? You can look at [other files](https://github.com/facebook/react-native/blob/master/Libraries/ActionSheetIOS/RCTActionSheetManager.m) in the repo to get an idea of the coding style we use.

javache:
I don't think we should have WhatsApp specific code in the RN core. Can you find a way to make this check more generic?

krizpoon:
Most ideally, the API could provide a callback to allow tailoring the message when an activity type is chosen.

krizpoon:
Instead of having hard-coded activity type, I think we could use a map to provide different message targeting different app.

So the code in the JS side would be like:

```
Share.share({
  message, // the default message
  customMessage: [
    { types: ['type1', 'type2'], message: customizedMessage1 },
    { types: ['type3', 'type*'], message: customziedMessage2 }, // simple prefix match supported
  ]
});
```

PR #14955, opened 2017-07-11
[CI] Migrate to Circle 2.0

hramos:
Following the migration guide. Let's see what happens here.


.circleci/config.yml
```diff
@@ -0,0 +1,116 @@
...
+            npm config set progress=false
+            npm install
+            # for eslint bot
+            npm install github@0.2.4
+            cd website && npm install
+
+      - save_cache:
+          key: react-native
+          paths:
+            - "ReactAndroid/build/downloads"
```
ide: If the list of RN downloads changes we'll want the key to invalidate itself. Perhaps we should include the checksum of one of the Gradle files in a cache key.

.circleci/config.yml
```diff
@@ -0,0 +1,116 @@
...
+          command: |
+            # copy test report for Circle CI to display
+            mkdir -p $CIRCLE_TEST_REPORTS/junit/
+            find . -type f -regex ".*/build/test-results/debug/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+            find . -type f -regex ".*/outputs/androidTest-results/connected/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+            # TODO circle does not understand Buck's report, maybe need to transform xml slightly
+            #find . -type f -regex ".*/buck-out/gen/ReactAndroid/src/test/.*/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+
+      - deploy:
+          branch: [/.*-stable/, /master/]
```
ide: I'm not sure if this syntax is supported anymore since the CircleCI docs say to check the branch name in the `command` section manually.

.circleci/config.yml
```diff
@@ -0,0 +1,116 @@
...
+            buck fetch ReactAndroid/src/test/java/com/facebook/react/modules
+            buck fetch ReactAndroid/src/main/java/com/facebook/react
+            buck fetch ReactAndroid/src/main/java/com/facebook/react/shell
+            buck fetch ReactAndroid/src/test/...
+            buck fetch ReactAndroid/src/androidTest/...
+            ./gradlew :ReactAndroid:downloadBoost :ReactAndroid:downloadDoubleConversion :ReactAndroid:downloadFolly :ReactAndroid:downloadGlog :ReactAndroid:downloadJSCHeaders
+            # CIRCLE_NPM_TOKEN is in React Native project settings in Circle CI.
+            # It was generated for bestander user, easy to replace with anyone's else
+            echo "//registry.npmjs.org/:_authToken=${CIRCLE_NPM_TOKEN}" > ~/.npmrc
+            npm config set spin=false
```
ide: I think we'll need to install Node before this, probably need to get nvm first

.circleci/config.yml
```diff
@@ -0,0 +1,116 @@
+version: 2
+jobs:
+  build:
+    machine: true
```
grabbou: We've been using workflows with separate images (in particular, testing on different Node versions, just like Travis does by default) and it's been working well. Not sure about performance disadvantages though.

grabbou:
Not sure it's worth it, but one small tweak that I've applied when migrating was to take advantage of `.circleci` folder and extract all commands into separate scripts. That way, I've reduced config size and made it more analysable at a glance (quite handy, when workflows are in place)

serima:
I think that CI's speed-up will contribute to improving the progress of the project and I update my favorite projects to CircleCI 2.0. I want to work on this issue together. How do I proceed?

hramos:
@serima I don't think I'll get back to this PR for some time as this month is pretty busy for me, but any PR opened against this repo should trigger a CI run. You can start fresh or use this PR as a base, and open a new PR with your changes. There's several things running on Circle, so a successful PR will need to ensure every test is covered, including website deployment and publishing to npm.

serima:
Hi, @hramos, I see. Thank you.

hramos:
Had some extra time between tasks today. Instead of migrating all of the old tests over, I'll handle them one by one.

Hopefully I'll have more time to dig into this tomorrow. My plate is quite full at the moment and I was not planning on working on this just yet, but it may be necessary in order to test multiple Node versions concurrently.

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+version: 2
+jobs:
+  test-node-8:
+      working_directory: ~/react-native
+      docker:
+        - image: circleci/node:8.1.4
+      steps:
+        - checkout
+        - restore-cache: *restore-cache
+        - npm install
```
ide: Need these to be behind `run:` keys

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+     - "/home/ubuntu/buck"
+    key: buck-dependencies-{{ .Branch }}
+
+
+version: 2
+jobs:
+  test-node-8:
+      working_directory: ~/react-native
+      docker:
+        - image: circleci/node:8.1.4
```
ide: What do you think of just making this be `circleci/node:8` so we naturally test the latest version of 8.x (but give up some reproduceability)?

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+      - checkout
+      - restore-cache: *restore-cache
+      - npm install
+      - save-cache: *save-cache
+      - run: |
+          npm run lint
+          npm run flow -- check
+          npm test -- --maxWorkers=1
+  test-node-4:
+    working_directory: ~/jest
```
ide: Should this be ~/react-native like the others?

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
+aliases:
+  - &restore-cache
+    keys:
+      - dependencies-{{ .Branch }}-{{ checksum "package.json" }}
```
ide: Can you prefix all the keys with `v1-` so we can easily manually bust the cache if we need to?

I'm also concerned about sharing node_modules caches across npm versions. In particular, Node 4 comes with npm 2 which doesn't have the deduping functionality of npm 3+.

What if we didn't cache node_modules for react-native but continued to cache them for the website and danger (where we use just a single Node + npm version)?

serima:
Thank you very much. The point of @ide's code review seems very appropriate. After that, it takes time to repeat trial and error.
Anyway, I appreciate your thoughtfulness.

.circleci/config.yml
hramos: It should! I followed [Jest's](https://github.com/facebook/jest/blob/master/.circleci/config.yml) as a starting point, and I overlooked this line.

hramos:
Yup, this has been helpful. I had to push out the previous commit earlier before I had to run. Taking another stab at doing some basic JS tests tonight.

hramos:
Lints and basic tests are working on Node 6 and 8. npm install failed on 4 as expected. Having all the individual workflow results listed here on GitHub is very useful, and should help us get better signal even when some tests are broken.

.circleci/config.yml
```diff
@@ -0,0 +1,201 @@
...
+      - attach_workspace:
+          at: ReactAndroid/src/androidTest/assets/
+      - run: |
+          if [[ ! -e ReactAndroid/src/androidTest/assets/AndroidTestBundle.js ]]; then
+            echo "JS bundle missing, verify build-js-bundle step"; exit 1
+          else
+            echo "JS bundle found.";
+          fi
+# dependencies
+      - run: echo 'export PATH=~/$CIRCLE_PROJECT_REPONAME/gradle-2.9/bin:/home/circleci/buck/bin:$PATH'
```
ide: Wouldn't PATH need to be defined under the `environment` section above?

.circleci/config.yml
```diff
@@ -0,0 +1,201 @@
...
+            cd /home/circleci/buck && ant
+            buck --version
+            source scripts/circle-ci-android-setup.sh && getAndroidSDK
+            buck fetch ReactAndroid/src/test/java/com/facebook/react/modules
+            buck fetch ReactAndroid/src/main/java/com/facebook/react
+            buck fetch ReactAndroid/src/main/java/com/facebook/react/shell
+            buck fetch ReactAndroid/src/test/...
+            buck fetch ReactAndroid/src/androidTest/...
+            ./gradlew :ReactAndroid:downloadBoost :ReactAndroid:downloadDoubleConversion :ReactAndroid:downloadFolly :ReactAndroid:downloadGlog :ReactAndroid:downloadJSCHeaders
+      - run:
```
ide: How about splitting the dependencies step into two separate ones (getting the Android SDK and getting Buck and the RN deps), and then prelaunching the emulator as soon as we've fetched the SDK?

I also suspect that caching some of the Android SDK directories (something to do with ANDROID_HOME or ANDROID_SDK_ROOT, probably need to SSH into the box and poke around to figure out what to cache) would speed up Android builds a good amount without making them flakier, just because developers rarely need to clear out ANDROID_HOME due to bad caching bugs. That said right now I do like this more than creating a Docker image with the Android emulator, really nice not to need to update Docker images <3

.circleci/config.yml
```diff
@@ -0,0 +1,201 @@
...
+          if [[ ! -e ReactAndroid/src/androidTest/assets/AndroidTestBundle.js ]]; then
+            echo "JS bundle missing, verify build-js-bundle step"; exit 1
+          else
+            echo "JS bundle found.";
+          fi
+# dependencies
+      - run:
+          name: Setting up Buck and Android dependencies
+          command: |
+            if [[ ! -e /home/circleci/buck ]]; then
```
ide: Small suggestion - how about using "~/buck" just to avoid thinking as much about how CircleCI does homedirs?

.circleci/config.yml
```diff
@@ -0,0 +1,201 @@
...
+      - image: circleci/openjdk:8-jdk
+    environment:
+      - TERM: "dumb"
+      - ADB_INSTALL_TIMEOUT: 10
+      - GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx512m -XX:+HeapDumpOnOutOfMemoryError"'
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ReactAndroid/src/androidTest/assets/
+      - run: |
```
ide: Maybe give this step a name so that CircleCI shows the name instead of this whole command in its web UI

.circleci/config.yml
```diff
@@ -0,0 +1,201 @@
...
+            echo "JS bundle found.";
+          fi
+# dependencies
+      - run:
+          name: Setting up Buck and Android dependencies
+          command: |
+            if [[ ! -e /home/circleci/buck ]]; then
+              git clone https://github.com/facebook/buck.git /home/circleci/buck;
+            fi
+            echo 'export PATH=~/CIRCLE_PROJECT_REPONAME/gradle-2.9/bin:/home/circleci/buck/bin:$PATH'
```
ide: This probably needs to go under the environment section above?

.circleci/config.yml
hramos: Usually, yes, but interpolating environment variables there is not yet supported. This is [a workaround suggested by a Circle CI employee](https://discuss.circleci.com/t/how-to-add-a-path-to-path-in-circle-2-0/11554).

Without the workaround, defining PATH in `environment` would blow away the original value of PATH (because $PATH is not defined), resulting in errors such as `mkdir` not being available in the path.

.circleci/config.yml
hramos: Good idea on splitting (and actually I did this just now before reading this).

I do want to set up caching, but I'll probably get to it once we have a full successful run. I don't want to pollute the cache or introduce side effects until I'm sure everything is working as expected.

I wish GitHub would not hide review comments automatically with any new commit, as I want to get back to this piece of feedback later on.

.circleci/config.yml
hramos: Same as above, cannot do because $PATH is not available to us when `environment` is processed.

.circleci/config.yml
hramos: Hadn't seen this comment but I believe it's done now :)

.circleci/config.yml
hramos: Done.

hramos:
The image I'm using doesn't come with Android pre-installed. Looks like circleci/android:api-23-alpha is based on the same openjdk8 image we're using, but comes with Android pre-installed, so I'll make the switch.

.circleci/config.yml
```diff
@@ -0,0 +1,211 @@
...
+          npm run flow -- check
+          npm test -- --maxWorkers=1
+
+  test-node-6:
+    working_directory: ~/react-native
+    docker:
+      - image: circleci/node:6.11.0
+    steps:
+      - checkout
+      - run: npm install --no-package-lock
```
ide: --no-package-lock applies only to npm 5+, which ships only with Node 8 and newer, think we don't need the flag here

.circleci/config.yml
```diff
@@ -0,0 +1,211 @@
...
+          npm run flow -- check
+          npm test -- --maxWorkers=1
+
+  test-node-4:
+    working_directory: ~/react-native
+    docker:
+      - image: circleci/node:4.8.4
+    steps:
+      - checkout
+      - run: npm install --no-package-lock
```
ide: nor here

hramos:
I'm using 23 here as building projects natively requires the 23.0.1 build tools. I need to check again with the rest of the team to see if anything has changed, but as I understand it we don't officially support building with a newer version of the build tools. I'm not very familiar with how the build tools may be coupled with the SDK, nor if it matters if we use the same API level here.

That said, we probably already have great coverage to ensure we can build against level 23 internally. Making sure projects can be built using the latest SDK sounds like a good idea here.

ide:
Let's start with the v23 build tools to get things working and then later see if upgrading to v26 works, thanks for powering through this...

hramos:
Done for the day. Earlier I ssh'd into the image and was able to get all the uncommented Android tests to run successfully. If the latest commit fails to run successfully, I probably forgot to add something to the config file.

.circleci/config.yml
```diff
@@ -0,0 +1,251 @@
...
+      # Fallback in case checksum fails
+      - v1-android-dependencies-{{ .Branch }}-
+  - &save-cache-android
+    paths:
+     - "ReactAndroid/build/downloads"
+    key: v1-android-dependencies-{{ .Branch }}-{{ checksum "ReactAndroid/build.gradle" }}
+
+  - &restore-cache-buck
+    keys:
+# TODO: Pick file to checksum against
```
ide: I just realized with CircleCI caching we'd stop using the latest copy of Buck. If we want that, one thing we could do is commit a small file to RN that contains just a Git commit hash or Git tag of the Buck repo. We'd check out just that commit and cache it. That way we intentionally don't use the latest copy of Buck and also consistently use the same version, not at the whim of CircleCI's caching behavior.

If we do want to use the latest copy of Buck, we could create a shallow clone of its Git repo (to avoid downloading 500MB) and build it each time.

.circleci/config.yml
```diff
@@ -0,0 +1,251 @@
...
+            done
+# Tests
+      - run:
+          name: Build Android app
+          command: |
+            buck build ReactAndroid/src/main/java/com/facebook/react
+            buck build ReactAndroid/src/main/java/com/facebook/react/shell
+      - run:
+          name: Compile native libs for unit and integration tests
+          command: ./gradlew :ReactAndroid:packageReactNdkLibsForBuck -Pjobs=1 -Pcom.android.build.threadPoolSize=1
```
ide: Could try bumping the `make` jobs to 2 since compiling C++ is slow here

.circleci/config.yml
```diff
@@ -0,0 +1,251 @@
...
+          no_output_timeout: 6m
+      - run:
+          name: Unit Tests
+          command: buck test ReactAndroid/src/test/... --config build.threads=1
```
ide: Might be able to do threads=2

.circleci/config.yml
```diff
@@ -0,0 +1,251 @@
...
+    working_directory: ~/react-native
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - run: npm install --no-package-lock
+      - run: |
+          npm run lint
+          npm run flow -- check
+          npm test -- --maxWorkers=1
```
ide: CircleCI says there are 2 CPUs per build now, could try 2 workers (and for the other Node versions as well)

.circleci/config.yml
```diff
@@ -0,0 +1,251 @@
...
+  build-js-bundle:
+    working_directory: ~/react-native
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - run: npm install --no-package-lock
+      - run:
+          name: Build JavaScript Bundle
+          command: node local-cli/cli.js bundle --max-workers 1 --platform android --dev true --entry-file ReactAndroid/src/androidTest/js/TestBundle.js --bundle-output ReactAndroid/src/androidTest/assets/AndroidTestBundle.js
```
ide: Another place we could try using 2 workers

.circleci/config.yml
```diff
@@ -56,6 +56,14 @@ aliases:
       - /opt/ndk
     key: v1-android-ndk-r10e-{{ .Branch }}
 
+  - &restore-cache-watchman
+    keys:
+      - v1-watchman-v4.9.0-{{ .Branch }}
```
hramos: It'd be useful to have some variables that define which version of watchman, Buck, etc we'd like to use.

hramos:
Side note here: some work has already been done by the ContainerShip folks to configure a custom Docker image for tests (see `ContainerShip/`). While in this PR I'd like to focus on getting Circle 2 working with the provided images, we could switch to using a custom image in the future. Doing so could reduce friction in running local tests, in that contributors would only need to install Docker and download our custom image before getting started.

.circleci/config.yml
```diff
@@ -0,0 +1,290 @@
...
+          command: |
+            if [[ ! -e ~/watchman ]]; then
+              mkdir ~/watchman
+              git clone https://github.com/facebook/watchman.git ~/watchman
+              cd ~/watchman
+              git checkout v4.9.0
+              ./autogen.sh
+              ./configure
+              make
+              sudo make install
```
ide: Since this probably installs files outside of ~/watchman, we probably want to run this step every time regardless of whether watchman was cached

.circleci/config.yml
```diff
@@ -0,0 +1,290 @@
...
+            cd ~/buck && ant
+            buck --version
+      - save-cache: *save-cache-buck
+      - restore-cache: *restore-cache-watchman
+      - run:
+          name: Install Watchman
+          command: |
+            if [[ ! -e ~/watchman ]]; then
+              mkdir ~/watchman
+              git clone https://github.com/facebook/watchman.git ~/watchman
```
ide: `--branch v4.9.0 --depth=1` might speed things up (and similar for the Buck repo clone)

.circleci/config.yml
ide: I made a separate issue to keep track of a couple extra things like Android SDK caching for later: https://github.com/facebook/react-native/issues/15855

hramos:
Hmph, I didn't run into these watchman make issues when I ssh'd into the machine earlier.

ide:
It might make sense to remove the `{{ .Branch }}` part from the NDK cache key since the NDK has been much more unchanged than other parts of the code, the reason being that it takes longer to upload the cache than it takes to download the NDK. Otherwise every PR is going to have to spend 5 minutes reuploading the cache.

Times:
screenshot 2017-09-07 10 42 23

.circleci/config.yml
```diff
@@ -0,0 +1,333 @@
...
+              mkdir ~/ndk
+              cd ~/ndk
+              curl -o ndk.zip --silent https://dl.google.com/android/repository/android-ndk-r10e-linux-x86.zip
+              curl -o ndk_64.zip --silent https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
+              unzip -o ndk.zip
+              unzip -o ndk_64.zip
+              rm ndk.zip
+              rm ndk_64.zip
+            fi
+      - save-cache: *save-cache-ndk
```
hramos: @ide any ideas how we can only update the cache if `~/ndk` didn't exist? From what I can tell, saving a cache can only be done by specifying a `save-cache` key as a step, so this can't be easily moved into the step above.

hramos: Actually, updating the cache if ~/ndk didn't exist is not what we want. Leaving this line in should be OK if Circle skips updating the cache if the key already exists.

hramos:
Unit tests are failing, seems like we're missing some dependencies?

Moved down the attach workspace step which fails when a job is retried with ssh, which should at least allow us to ssh into the machine after all the deps are in place in order to debug this further.

Rest of my day is taken up with meetings, I'll come back to this later.

hramos:
Forgot to mention: caching is in place in various places, which should significantly speed up the build!

hramos:
Failures are in the Animated library, I suspect it's because our AVD is launched with no hw accel.

hramos:
Looks like starting the emulator like so _should_ fix this, but I am getting a segfault:

```
$ LD_LIBRARY_PATH=${ANDROID_HOME}/emulator/lib64:${ANDROID_HOME}/emulator/lib64/qt/lib emulator64-arm -avd testAVD -no-audio -no-window -no-boot-anim -gpu on
Segmentation fault
```
https://stackoverflow.com/questions/45424110/how-to-run-android-ui-tests-properly-on-circleci-2-0

hramos:
I'm afraid Docker images on Circle are unable to take advantage of hardware acceleration. If I understand correctly, the Animated tests require hw accel.

ide:
This is getting out of my depth but FWIW CircleCI 1.0 ran its tests in Docker too, which makes me think that it's possible to run the Animated tests somehow. Mind if I clone this PR and try to poke around?

hramos:
Yup, go for it. If you ssh into the machine, make sure to update your PATH in the same way as we do in https://github.com/facebook/react-native/pull/14955/files#diff-1d37e48f9ceff6d8030570cd36286a61R198 because $BASH_ENV points to a random file when this runs in the image. Once the ssh-enabled run has installed all deps, you should be able to use buck to run the tests manually.

The JS bundle won't be available to you when you re-run with SSH enabled though, as it's built in a different job and the attach_workspace step fails when the build is retried. I'll see if I can provide the bundle as an artifact to aid with debugging.

hramos:
In case it helps, there's a difference in how we manage SDKs and the AVD in Circle 1.0 versus Circle 2.0. With Circle 1.0, we use the `android` CLI which is now deprecated, and we start the AVD using `$ANDROID_HOME/tools/emulator -avd testAVD -no-skin -no-audio -no-window`.

With Circle 2.0, I'm using `sdkmanager` instead to manage all SDK dependencies, and the AVD is started using emulator64-arm. I'm not very familiar with Android tooling, but it's possible that the emulator CLI used in Circle 1.0 supports hw accel. I've attempted to enable the GPU with emulator64-arm, but it only results in a Segmentation fault if the ARM image is used. I looked into using the x86 image, but after looking at the instructions to enable hw accel in linux, it looks like the Docker image is not running in a host that would allow us to use hw accel.

Ideally, I'd set up the Android SDK and AVD in the same manner as we do in Circle 1.0, but it looks like both the android-23 and android-26 images are already set up with recent versions of the Android SDK where `android` is already deprecated. I'll keep prodding here and see if I can get closer to the Circle 1.0 config.

ide:
@hramos I added a couple of commits at the end to fix Android unit tests (ended up being unrelated to GPU) and to comment out Android integration tests that were commented out in the CircleCI 1 config anyway. Tests look green now -- do you want to uncomment the Node 6 and Website jobs and then land this?

hramos:
Uncommented tests. This should be good to go now.

Issue #14872, opened 2017-07-06
FlatList complex example does not explain 'selected' state & prop


jkdf2:
The complex example in [the FlatList documentation](https://facebook.github.io/react-native/docs/flatlist.html) goes through pains to set `this.state.selected` with the correct `MyListItem` but provides no motivation for why we care about the `selected` prop in `MyListItem`.

Is the idea that `MyListItem` will use `selected` to style itself depending on whether or not it's a selected element?

If you are already using a `TouchableHighlight` or other `Touchable` that includes feedback and an `onPress` that responds appropriately, is there necessarily a reason to bother with identifying which one is selected?

hramos:
I'm not sure this belongs here in the issue tracker.

jkdf2:
@hramos The intention here is that the documentation be clarified. Either someone (most likely @sahrens) can clarify on this issue and I will update the documentation with a pull request, or they will create a pull request that updates the documentation directly and closes this issue.

Therefore I believe the issue should be re-opened and closed via a pull request that clarifies the documentation.

sahrens:
The example is just showing how to properly manage state across multiple items, for example in a todo list where you can select multiple items then take an action on all of them (like delete them). Feel free to clarify that in a PR if you think it's important and list me as a reviewer.

PR #14957, opened 2017-07-11
Give complex FlatList example more motivation

jkdf2:
Previously, the example MyListItem never referenced the selected prop, leaving ambiguity about the intention of the demo.

By adding a concrete implementation with coloring based on the selected prop, we can see that this is a demo of a multi-select list for batch actions instead of, say, a click-to-navigate nested list.

Here is a working Snack demo for future reference:
https://snack.expo.io/BkRMRTGB-

References #14872

### DIFF
--- a/Libraries/Lists/FlatList.js
+++ b/Libraries/Lists/FlatList.js
@@ -223,67 +223,70 @@ type DefaultProps = typeof defaultProps;
  *       renderItem={({item}) => <Text>{item.key}</Text>}
  *     />
  *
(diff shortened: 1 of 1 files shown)


jkdf2:
Thank you @sahrens for clarifying the FlatList demo in #14872.

I have made changes in hopes of clarifying the demo and would appreciate your review.

sahrens:
Hey sorry missed your last update - feel free to ping me next time. Hopefully it still merges cleanly :)

Issue #14436, opened 2017-06-09
NativeComponents iOS Documentation - RCT_CUSTOM_VIEW_PROPERTY 


MattFoley:
### Description

The RCT_CUSTOM_VIEW_PROPERTY example for RNTMap no longer functions, and isn't properly explained.

```objc
// RNTMapManager.m
RCT_CUSTOM_VIEW_PROPERTY(region, MKCoordinateRegion, RNTMap)
{
  [view setRegion:json ? [RCTConvert MKCoordinateRegion:json] : defaultView.region animated:YES];
}

```

The native RNTMap class does not exist, and prior to this is not mentioned at all throughout the tutorial. It is however created later on when describing event handling, although the creation of it is not explained.

Side note, why does RCT_CUSTOM_VIEW_PROPERTY supply "json" rather than "name"? Seems very counter intuitive that you're implementing a method that is supposed to make use of arguments that aren't defined any where near the definition.

ie:
```objc
#define RCT_CUSTOM_VIEW_PROPERTY(name, type, viewClass) \
RCT_REMAP_VIEW_PROPERTY(name, __custom__, type)         \
- (void)set_##name:(id)**name** forView:(viewClass *)view withDefaultView:(viewClass *)defaultView
```

### Solution

Review documentation, break into multiple components, explain in more detail.

### Additional Information

* React Native version:0.45
* Platform: iOS
* Development Operating System: N/A
* Dev tools: N/A

hramos:
`MapViewIOS` was removed a couple of versions ago. No one has touched this guide in a while, so unless someone volunteers to get it back up to date, we'll probably end up removing it from the docs.

ShannonChenCHN:
@MattFoley I fixed this issue by using `MKMapView` to replace `RNTMap`, and it worked great. Maybe you can have a try.

teameh:
@ShannonChenCHN Yeah that worked for me as well. I'll create a PR for this. Took me way too long to figure this out.

teameh:
Added another pr, clearing up the categories a bit.

teameh:
I'll stop here for now, but maybe I could rewrite the whole page if I understand how everything works a bit more 😏

hdchan:
I'm still getting a "Warning: Native component for "RNTMap" does not exist"
Has anyone successfully implemented this?

teameh:
[...]

PR #14983, opened 2017-07-12
Docs: ios-native-ui-guide RCTConvert categories edit

teameh:
I was struggling a bit with the [iOS Native UI Components tutorial](https://facebook.github.io/react-native/docs/native-components-ios.html#content).

As noted here it's outdated. https://github.com/facebook/react-native/issues/14436

In my opinion, these steps are more clear when the header files are also included in the example. Besides that, the tutorial can use the already existing `RCTConvert+CoreLocation` category for simplicity.


teameh:
Closing this one, sorry, I'm going to rewrite the whole page (guide). I've encountered a few more things that doesn't work any more.

PR #10427, opened 2016-10-17
Apple TV support 4: support for input (tvOS focus engine)

dlowder-salesforce:
- Motivation

Make Apple TV React Native apps actually work by supporting Siri Remote and keyboard input through the TouchableHighlight and TouchableOpacity components.  Also enable input into TextInput components, and support TabBarIOS.  Add an API for changing which component has the tvOS focus.
- Test plan

Existing Apple TV automation still passes.  Reenabling existing snapshot tests for Apple TV.  Will add snapshot tests later for verifying the tvOS focus API.

PR #14984, opened 2017-07-12
Remove unused focusedOpacity prop and function

levsero:
The `focusedOpacity` prop is only used inside `_opacityFocused` which is not used anywhere. This pr removes this unused code.
The code was added in https://github.com/facebook/react-native/pull/10427 but it does not appear to be used in the final version of the pr.


shergin:
@dlowder-salesforce Do we still need this?

dlowder-salesforce:
@shergin No I don't think we need this any more.

shergin:
@levsero Please add trailing commas back.

Issue #13991, opened 2017-05-16
RN + socket.io - Failed to set the 'responseType' property on 'XMLHttpRequest'


deanvaessen:
### Description

Explain what you did, what you expected to happen, and what actually happens.

Good morning,

I am having an issue combining React Native with socket.io. I do hope I am in the right place here, but considering I am using the same code on React Native and React while only encountering issues in RN, I figured it was worth a shot. It's a bit of an odd issue, and I've tried looking online but I'm having trouble figuring this out. Other people seem to be able to work with socket.io + RN, so I'm a bit stumped.

I am trying to connect to a backend server through socket.io but it seems that as soon as I try to do this, React Native complains to me:
"Failed to set the 'responseType' property on 'XMLHttpRequest': The response type cannot be set after the request has been sent."

I am not actually creating an XMLHttpRequest myself (yet I think with React Native the socket connection might fall back to XMLHttp? I read something about this).

Additionally out of curiosity, if I have a server with socket.io running (I do), could I use React Native's own implementation of sockets to connect to it? Is this compatible?

Here is a stack trace:
Http://imgur.com/a/NKa7V

### Reproduction Steps and Sample Code

Here is a snippet of what I am currently using:
```
/**
* { Dependencies }
*/
	var config = require('./../../../../../package.json').config;
	const PORT = config.backEndExternalPort;
	const BASECONNECTION = config.domain;

	// Vendor functions
		// Socket.IO
		import io from 'socket.io-client';
		//import io from 'socket.io-client/socket.io';
		const socket = io.connect(`${BASECONNECTION}:${PORT}`);

		import { setWhisper, setAnnouncement, connectLobby, updateSessionDetails, startGame, startRound, finishRound, finishGame } from '../../actions/actions'

/**
* { App }
*/

const index = (function() {

	const init = function(view) {

	  socket.on('connectLobby', (data) => {
	  	console.log('Lobby connected!');
	  	console.log(data);

	  	view.props.dispatch(connectLobby(data));
	  });
```

### Additional Information

* React Native version: "0.44.0"
* Platform: Android
* Development Operating System: Windows
* Dev tools: Android Dev Studio, everything configured with versions as per the "Getting started" guide for RN.
( "socket.io-client": "^2.0.1" )

dobrienlzd:
Could you confirm that you are really running React Native and Socket versions you indicated?  

[...]

PR #14992, opened 2017-07-13
Removes the error thrown by setting responseType after a request has …

beau6183:
Resolves Issue #13991. There are clear reproduction steps in that ticket.

Specifically, when using `socket.io-client` within `react-native`, an attempt is made to change the `responseType` after a result is received to match the `content-type`. This works fine with other implementations of `XMLHttpRequest`, but not RN's.

I simply removed the exception being thrown (and test scenario for it as it's no longer needed).

Examples of other standard impls of set responseType:
Chromium: https://chromium.googlesource.com/chromium/blink.git/+/99b8c9800ac123eddc3e199088d22569c5294b22/Source/core/xml/XMLHttpRequest.cpp#217

Mozilla: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType (no indication that an exception will be thrown if changing after response is received)

### DIFF
--- a/Libraries/Network/XMLHttpRequest.js
+++ b/Libraries/Network/XMLHttpRequest.js
@@ -175,12 +175,6 @@ class XMLHttpRequest extends EventTarget(...XHR_EVENTS) {
   }
 
   set responseType(responseType: ResponseType): void {
-    if (this._sent) {
-      throw new Error(
-        'Failed to set the \'responseType\' property on \'XMLHttpRequest\': The ' +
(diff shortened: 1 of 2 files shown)


jussirantala:
This should be fixed. You can't set custom headers if you use websocket transportation mode and if this bug isn't fixed, you can't use polling mode.

I have this bug with:
- react-native 0.47.2
- socket.io-client 2.0.4

Issue #10635, opened 2016-10-31
null is not an object (evaluating 'this._component.getScrollableNode')


jerson:
### Description

http://crashes.to/s/280c411449d

```
Fatal Exception: com.facebook.react.common.JavascriptException: null is not an object (evaluating 'this._component.getScrollableNode'), stack:
value@297:1662
value@297:1031
unmountComponent@144:3508
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountChildren@142:830
unmountChildren@139:2252
unmountComponent@119:361
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponent@144:3555
unmountComponent@134:444
unmountComponentFromNode@234:1542
unmountComponentAtNode@234:1438
unmountComponentAtNodeAndRemoveContainer@234:1196
unmountApplicationComponentAtRootTag@328:1109
value@58:2827
<unknown>@58:1062
d@58:130
value@58:1034

       at com.facebook.react.modules.core.ExceptionsManagerModule.showOrThrowError(ExceptionsManagerModule.java:97)
       at com.facebook.react.modules.core.ExceptionsManagerModule.reportFatalException(ExceptionsManagerModule.java:81)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:525)
       at com.facebook.react.bridge.BaseJavaModule$JavaMethod.invoke(BaseJavaModule.java:318)
[...]

PR #15019, opened 2017-07-14
Verify that the component passed to createAnimatedComponent is not functional

janicduplessis:
Stateless functional components don't support refs and we need that for the component to work, it used to crash with this error message: `undefined is not an object (evaluating 'this._component.getScrollableNode')`. This makes it clear what the issue is.

Fixes some of the errors in #10635, not sure if it fixes all the cases described in the issue though.

**Test plan**
Tested that passing a component with createClass or extends Component works but passing a function causes an error.

## Release Notes

[GENERAL] [ENHANCEMENT] [Animated] - Verify that the component passed to createAnimatedComponent is not functional


Libraries/Animated/src/AnimatedImplementation.js
```diff
@@ -1732,6 +1732,12 @@ class AnimatedProps extends Animated {
 }
 
 function createAnimatedComponent(Component: any): any {
+  invariant(
+    Component.prototype.render,
```
janicduplessis: Is there a better way to check this? It does work but meh...

janicduplessis:
@sahrens Could you have a look at this?

Libraries/Animated/src/createAnimatedComponent.js
```diff
@@ -17,7 +17,15 @@ const AnimatedProps = require('./nodes/AnimatedProps');
 const React = require('React');
 const ViewStylePropTypes = require('ViewStylePropTypes');
 
+const invariant = require('fbjs/lib/invariant');
+
 function createAnimatedComponent(Component: any): any {
```
ide: For Flow checking within the RN codebase, you could add something like this:

```js
createAnimatedComponent<P>(
  Component: Class<React.Component<P>>
): Class<React.Component<*>> {
  ...
}
```

I believe this will filter out functions too.

janicduplessis:
@sahrens Yea I’m not sure if there is a better way, I commented on the diff to ask that but it doesn’t show up since I rebased. Maybe ask someone from the react team.

Libraries/Animated/src/createAnimatedComponent.js
janicduplessis: I remember trying to type this properly but I hit some issue. React components typing changed in flow 0.53 I can try again to see if I can figure out something that works.

janicduplessis:
@bvaughn Is there a public API to check if a component is functional? (needed to know if refs are available).

bvaughn:
FWIW we use this type of duck typing in React too (eg [`ReactFiberBeginWork`](https://github.com/facebook/react/blob/85b59d2fc96a286271f6b4ad126bca72871ac9bc/src/renderers/shared/fiber/ReactFiberBeginWork.js#L478-L480), [`ReactTestUtilsEntry`](https://github.com/facebook/react/blob/85b59d2fc96a286271f6b4ad126bca72871ac9bc/src/renderers/dom/test/ReactTestUtilsEntry.js#L132-L134)).

`ReactBaseClass` (aka `React.Component`, `React.PureComponent`) also defines an [`isReactComponent`](https://github.com/facebook/react/blob/85b59d2fc96a286271f6b4ad126bca72871ac9bc/src/isomorphic/modern/class/ReactBaseClasses.js#L30) prop that we use some places (like in [`ReactFiber`](https://github.com/facebook/react/blob/85b59d2fc96a286271f6b4ad126bca72871ac9bc/src/renderers/shared/fiber/ReactFiber.js#L229-L231)).

janicduplessis:
Thanks @bvaughn ! Update to use `isReactComponent`.

@ide I played with trying to flow type it properly again but there was an issue with default props, it didn't seem to pick it up properly anymore so it caused a bunch of errors where some props that should not be required were.

janicduplessis:
ping @sahrens @bvaughn :)

Libraries/Animated/src/createAnimatedComponent.js
```diff
@@ -17,7 +17,15 @@ const AnimatedProps = require('./nodes/AnimatedProps');
...
+  invariant(
+    Component.prototype && Component.prototype.isReactComponent,
+    '`createAnimatedComponent` does not support stateless functional components, ' +
```
bvaughn: tiny nit: "," should be a ";" I think

janicduplessis: I'll trust you on that one. I never know when `;` should be used, I guess I wasn't listening at school that day :)

ide:
@janicduplessis would you want to rebase this?

janicduplessis:
@ide Looks like tests need to be updated to pass this new check, will try to find some time to update this soon.

janicduplessis:
Updated and re-tested, we actually need to allow string components too since in prod views are just RCTView strings.

PR #15020, opened 2017-07-14
Update Releases.md

hramos:
Cleanup and clarification based on my experience performing some patch releases recently.


hramos:
Not my intention! The original was OK and I was able to follow it to release a few versions without issues.

grabbou:
I know :) Just kidding! I was able to learn a lot by reviewing it. Now
gonna check how my release goes by reading it!

PR #15023, opened 2017-07-14
Add 'contentInsetAdjustmentBehavior' (new in iOS 11) to ScrollView

skevy:
In iOS11, Apple added a new layout feature called "Safe Areas" (this blog post talks a bit about it: https://www.bignerdranch.com/blog/wwdc-2017-large-titles-and-safe-area-layout-guides/).

UIScrollView is one component that is affected by this change in Apple's API. When the `contentInsetAdjustmentBehavior` is set to `automatic`, for example, it will adjust the insets (and override any manually set insets) automatically based on whether or not there's a UINavigationBar, a UITabBar, a visible status bar, etc on the screen. Frustratingly, Apple decided to default to `Automatic` for this behavior, which will cause any apps that set contentInset/contentContainerStyle padding to have their values offset by, at the very least, the size of the status bar, when they compile their app for iOS 11. Here's more information about this behavior: https://developer.apple.com/documentation/uikit/uiscrollview/2902261-contentinsetadjustmentbehavior?language=objc

Mostly, this is a really straightforward change -- it simply adds a new iOS-only prop to ScrollView that allows setting `contentInsetAdjustmentBehavior`. But I did decide to default the behavior to `never`, so that it mimics the behavior we've seen in iOS < 11. I think it's good to keep something as crucial as scrollview content insets non-magical, and also keep it behaving similarly between platforms.


shergin:
I think we have to just default this prop to `never` (and do not expose it to RN), because:
 * RCTScrollView is not ready to randomly changing contentInsets;
 * It probably will not work with RN wrappers around UINavigationBar and UITabBar anyways.
@skevy Do you really use `auto` value in your code?

skevy:
I don’t today, but I might! :p

I think it _will_ actually work with RN wrappers around UINavigationController and UITabBarController, as iOS makes it pretty automagical.

But honestly, I think you’re first point is the killer here. That makes sense. It probably would break in weird cases.

I’m fine with setting it to never. I will probably never use this property. :p

skevy:
I could of course add support for an event when content insets are adjusted – but that would be weird when thinking about cross platform support (cuz android would never randomly change your content insets).

shergin:
@javache Pieter, what do you think, should we expose it?

javache:
Given that may want to set this when integrating with other UIKit hierarchies, I think we should.

shergin:
@skevy Could you please wrap all native usages of new `UIScrollViewContentInsetAdjustment` enum using:
```
#if __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_11_0
  ...
#endif
```
?

skevy:
@shergin np. Sorry for the late reply, am on vacation!

PR #15042, opened 2017-07-17
fix arc drawing when sweep >= 360

nihgwu:
I hit this issue in my own component https://github.com/nihgwu/react-native-pie, and find this https://stackoverflow.com/questions/19383842/weird-behaviour-in-drawing-a-ring-using-path-arcto-in-android, after this change, it looks exactly the same on both Android and iOS


nihgwu:
Sample code:
```js
        <Pie
          radius={100}
          series={[10, 20, 30, -40]}
          colors={['red', 'lime', 'blue', 'yellow']} />
        <Pie
          radius={100}
          innerRadius={60}
          series={[10, 20, 30, 400]}
          colors={['#f00', '#0f0', '#00f', '#ff0']} />
        <View>
          <Pie
            radius={50}
            innerRadius={45}
            series={[100]}
            colors={['#f00']}
            backgroundColor='#ddd' />
          <View style={styles.gauge}>
            <Text style={styles.gaugeText}>100%</Text>
          </View>
        </View>
      </View>
```
After:
image

Before:
image

On iOS:
image

nihgwu:
cc @javache @shergin

shergin:
@nihgwu That's great!
One question: How the demo will look (on iOS) with 50% transparent colors? Could you please create a snack with your demo?

nihgwu:
@shergin https://snack.expo.io/By3_faIPb

nihgwu:
@shergin Any thing should I do to make this be merged?

shergin:
Android documentations said:
> sweepAngle: Sweep angle (in degrees) measured clockwise, treated mod 360.

So, it means that it already does not support arcs "bigger" that 360 deg (but iOS does, I assume), so it means that this should not introduce a regression here.

Let's try to land it on Monday.

nihgwu:
@shergin another release is coming, can we make this in?

PR #15043, opened 2017-07-17
Add delay support to Animated.spring

billoosijok:
Aadding a `delay` option to `Animated.spring` works now 👇:

spring_delay

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -583,7 +583,7 @@ class SpringAnimation extends Animation {
       } else {
         this.onUpdate();
       }
-    }
+    };
 
     //  If this._delay is more than 0, we start after the timeout.
     if (this._delay) {
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -495,6 +495,7 @@ class SpringAnimation extends Animation {
   _tension: number;
   _friction: number;
   _delay: number;
+  _timeout: any;
   _lastTime: number;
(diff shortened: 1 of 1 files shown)


janicduplessis:
Thanks for the PR!

Could you just refactor this to be like TimingAnimation, ie create a start function and call it either directly or in a setTimeout. It would make the it a bit clearer.

billoosijok:
Sure. That makes sense!

Libraries/Animated/src/AnimatedImplementation.js
```diff
@@ -571,10 +576,20 @@ class SpringAnimation extends Animation {
...
+      if (this._useNativeDriver) {
+        this.__startNativeAnimation(animatedValue);
+      } else {
+        this.onUpdate();
+      }
+    }
+
+    //  If this._delay is more than 0, we start after the timeout.
+    if (this._delay) {
+        this._timeout = setTimeout(start, this._delay);
```
janicduplessis: This should be indented 2 spaces less.

Libraries/Animated/src/AnimatedImplementation.js
```diff
@@ -571,10 +576,20 @@ class SpringAnimation extends Animation {
...
-    if (this._useNativeDriver) {
-      this.__startNativeAnimation(animatedValue);
+
+    var start = () => {
+      if (this._useNativeDriver) {
+        this.__startNativeAnimation(animatedValue);
+      } else {
+        this.onUpdate();
+      }
+    }
```
janicduplessis: One last missing semicolon here :)

billoosijok: Oops sorry about that!

PR #15046, opened 2017-07-17
Update WebSocketModule.java

sm2017:
Convert to base64 not utf8

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
@@ -159,7 +159,7 @@ public class WebSocketModule extends ReactContextBaseJavaModule {
 
       @Override
       public void onMessage(WebSocket webSocket, ByteString bytes) {
-        String text = bytes.utf8();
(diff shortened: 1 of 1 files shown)


javache:
Can you please provide a motivation for this change? What bug does this fix? Why was the previous behaviour wrong?

sm2017:
In android (not iOS) enviroment when I get binary message from websocket server , it throw `Invalid string. Length must be a multiple of 4` error , I found that https://github.com/facebook/react-native/blob/master/Libraries/WebSocket/WebSocket.js#L158 uses `base64.toByteArray(ev.data).buffer` and the https://github.com/beatgammit/base64-js/blob/master/index.js#L23 throw error 

I review iOS and Android native codes
If you see iOS version https://github.com/facebook/react-native/blob/master/Libraries/WebSocket/RCTWebSocketModule.m#L119-L120 it use `base64EncodedStringWithOptions` for `binary` messages

In android for `binary` messages , it encode to `utf8` not `base64` https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java#L162 , So in js side we have `Invalid string. Length must be a multiple of 4` error

PR #15054, opened 2017-07-17
Stop native driver animations when value is set.

kmagiera:
This diff changes the behaviour of natively driven animations in case the node that they are being run for has its value changed using `setValue` or as a result of an incoming event.

The reason for changing that is to match the JS implementation of `setValue` which behaves as described above (see relevant code here: https://github.com/facebook/react-native/blob/7cdd4d48c89ad642a1c107e0b40c25eb75682175/Libraries/Animated/src/AnimatedImplementation.js#L743)

**Test Plan:**
Use this sample app: https://snack.expo.io/B1V7RX9r-
Change: `USE_NATIVE_DRIVER` const between `true` and `false`.
See the animation stops regardless of the state of `USE_NATIVE_DRIVER` unlike before when it would stop only when `USE_NATIVE_DRIVER` was set to `false`

### DIFF
--- a/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
+++ b/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
@@ -170,6 +170,7 @@
     RCTLogError(@"Not a value node.");
     return;
   }
+  [self stopAnimationsForNode:node];
 
   RCTValueAnimatedNode *valueNode = (RCTValueAnimatedNode *)node;
   valueNode.value = value.floatValue;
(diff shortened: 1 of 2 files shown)


Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -264,6 +265,20 @@ - (void)stopAnimation:(nonnull NSNumber *)animationId
   }
 }
 
+- (void)stopAnimationsForNode:(nonnull RCTAnimatedNode *)node
+{
```
janicduplessis: Probably won’t make a big difference because active animations count is usually low but we should use a normal for loop like you did on android to avoid the extra array and removeObject (o(n)) calls.

kmagiera: I need to loop through `_activeAnimations` which is of type `NSMutableSet` whereas on android is just an array.
Not sure what is your suggestion on how to use a "normal for loop"? If I call `[_activeAnimations allObjects]` that would allocate new array for all the items anyways.

Complexity wise its O(n) either way unless we have mapping from node to active animations running for that node which IMO is not worth adding considering that the number of running animations should be very low (like literally just a few of them).

janicduplessis: I see, I thought it was a `NSArray` then this is fine.

janicduplessis:
@facebook-github-bot shipit

Issue #3847, opened 2015-11-03
WritableMap fromBundle cannot handle array


Matthew-C:
Arguments.java line 122:
value.getClass().isArray() always return false cause ArrayList is not array! 

Note bundle.putStringArray takes an ArrayList as parameter so this code never works.

Also is there a way to allow user to directly put JSONObject as initial properties?

yychun1217:
You should convert ArrayList to Array

christopherdro:
@Matthew-C Your going to have to convert the array to a `WritableArray`.

christopherdro:
@facebook-github-bot answered

PR #15056, opened 2017-07-17
Android: Added support to arrays in toBundle

Guichaguri:
The support for `ReadableArray` in `toBundle` was never implemented, throwing an `UnsupportedOperationException` when trying to convert an array.

* Created `toList` -- A method that converts a `ReadableArray` to an `ArrayList`
* Modified `toBundle` to support arrays using `toList`
* Created `fromList` -- A method that converts a `List` to a `WritableArray`
* Modified `fromBundle` to also support lists using `fromList`

This PR allows `toBundle` and `fromBundle` (as well as `toList` and `fromList`) to work consistently without loosing information.

**Test Plan**

I've created three different arrays: one full of integers, one full of strings, and one mixed (with a integer, a boolean, a string, null, a map with a string and a boolean array), putting all of them inside a map.

After converting the map to a `Bundle` using `toBundle`, the integer array was retrieved through `Bundle.getIntegerArrayList`, the string array through `Bundle.getStringArrayList` and the mixed array through `Bundle.get` (casting it to an `ArrayList`)

After checking whether each value from the bundle was correct, I converted the bundle back to a map using `fromBundle`, and checked again every value.

The code and results from the test can be found in [this gist](https://gist.github.com/Guichaguri/5c7574b31f9584b6a9a0c182fd940a86).

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/Arguments.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/Arguments.java
@@ -11,6 +11,7 @@ package com.facebook.react.bridge;
 import java.lang.reflect.Array;
 
 import java.util.AbstractList;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


Guichaguri:
This PR fixes #3847

ReactAndroid/src/main/java/com/facebook/react/bridge/Arguments.java
```diff
@@ -277,7 +315,57 @@ public static WritableMap fromBundle(Bundle bundle) {
   }
 
   /**
+   * Convert a {@link WritableArray} to a {@link ArrayList}.
+   *
+   * @param readableArray the {@link WritableArray} to convert.
+   * @return the converted {@link ArrayList}.
+   */
+  @Nullable
+  public static ArrayList toList(@Nullable ReadableArray readableArray) {
```
javache: Can you make the return type List? It's better to return the abstract interface rather than the concrete implementation type.

Guichaguri: The reason why I use `ArrayList` is because it is `Serializable`, while a `List` isn't, so I can use it in `toBundle` with `Bundle.putSerializable`.

Do you think it's better to return `List` and then cast it in `toBundle`?

PR #12249, opened 2017-02-07
ImagePickerIOS take photo with wrong orientation

yujiayinshi:
When using ImagePickerIOS camera take photo, photos only have landscape orientation, which is not matched with fact. 
So I add a function to save orientation info.

PR #15060, opened 2017-07-17
Fix photo orientation from ImagePickerIOS

doochik:
Original PR: https://github.com/facebook/react-native/pull/12249

ImagePickerIOS saves photos to ImageStoreManager without meta information. So photo has wrong orientation.

**Test plan**
1. Take the 2 photos (in landspape and portrait orientation) with this code: 
```
ImagePickerIOS.openCameraDialog(
  {},
  (uri) => CameraRoll.saveToCameraRoll(uri),
  () => {}
);
```
2. Ensure that photos in Photos app have right orientation.


shergin:
@doochik Why should not we fix ImageStoryManager instead?
I am a bit concern that we are resampling completely valid UIImage instance just because another module cannot properly consume this.

doochik:
@shergin Globally you're right, we should fix ImageStoryManager to store image with meta information. But this is a lot of work :)
I'll check if I can do it by myself.

doochik:
@shergin Take a look at my new PR. `RCTImageStoreManager` now converts images to NSData accoring to original image orientation.

Libraries/Image/RCTImageUtils.m
```diff
@@ -314,7 +314,7 @@ BOOL RCTUpscalingRequired(CGSize sourceSize, CGFloat sourceScale,
   return (__bridge_transfer id)imageProperties;
 }
 
-NSData *__nullable RCTGetImageData(CGImageRef image, float quality)
+NSData *__nullable RCTGetImageData(CGImageRef image, float quality, NSInteger orientation)
```
shergin: This function should accept `UIImageOrientation` and use conversion internally.

Libraries/Image/RCTImageUtils.m
```diff
@@ -325,7 +325,10 @@ BOOL RCTUpscalingRequired(CGSize sourceSize, CGFloat sourceScale,
   } else {
     // get jpeg data
     destination = CGImageDestinationCreateWithData(imageData, kUTTypeJPEG, 1, NULL);
-    properties = @{(NSString *)kCGImageDestinationLossyCompressionQuality: @(quality)};
+    properties = @{
+      (NSString *)kCGImageDestinationLossyCompressionQuality: @(quality),
+      (NSString *)kCGImagePropertyOrientation: @(orientation)
```
shergin: Is this make sense *only* for JPEG?

Libraries/Image/RCTImageUtils.m
```diff
@@ -314,7 +314,7 @@ BOOL RCTUpscalingRequired(CGSize sourceSize, CGFloat sourceScale,
   return (__bridge_transfer id)imageProperties;
 }
 
-NSData *__nullable RCTGetImageData(CGImageRef image, float quality)
+NSData *__nullable RCTGetImageData(CGImageRef image, float quality, NSInteger orientation)
 {
   NSDictionary *properties;
```
shergin: If we changing this quite radically (eg switching to using `NSMutableDictionary`), we can rename it to `imageOptions` (actual correct name from `CGImageDestinationCreateWithData` signature).

Libraries/Image/RCTImageStoreManager.m
```diff
@@ -76,13 +76,29 @@ - (void)storeImage:(UIImage *)image withBlock:(void (^)(NSString *imageTag))bloc
...
-    NSString *imageTag = [self _storeImageData:RCTGetImageData(image.CGImage, 0.75)];
+    NSInteger imageOrientation = [self _convertUIImageOrientationToCGImagePropertyOrientation:image.imageOrientation];
+    NSString *imageTag = [self _storeImageData:RCTGetImageData(image.CGImage, 0.75, imageOrientation)];
     dispatch_async(dispatch_get_main_queue(), ^{
       block(imageTag);
     });
   });
 }
 
+- (NSInteger)_convertUIImageOrientationToCGImagePropertyOrientation:(UIImageOrientation)imageOrientation
```
shergin: It should not be instance method because it completely stateless. It should be `static` C function.
And we should name it something like `CGImagePropertyOrientationFromUIImageOrientation`.
And, also, it would be great if we add comment with this link: https://stackoverflow.com/a/6699649/496389 .
And, moreover, if we create additional enum for all CGImagePropertyOrientation values from documentation (something like that: https://github.com/facebook/react-native/blob/master/React/Base/RCTRootView.h#L23), it would be just awesome!! 😄

Libraries/Image/RCTImageStoreManager.m
```diff
@@ -197,7 +213,7 @@ - (NSString *)storeImage:(UIImage *)image
   RCTLogWarn(@"RCTImageStoreManager.storeImage() is deprecated and has poor performance. Use an alternative method instead.");
   __block NSString *imageTag;
   dispatch_sync(_methodQueue, ^{
-    imageTag = [self _storeImageData:RCTGetImageData(image.CGImage, 0.75)];
+    imageTag = [self _storeImageData:RCTGetImageData(image.CGImage, 0.75, 1)];
```
shergin: UIImageOrientationUp

Libraries/Image/RCTImageUtils.m
doochik: This is for `CGImageDestinationAddImage` and `properties` is the correct name.
https://developer.apple.com/documentation/imageio/1464962-cgimagedestinationaddimage

doochik:
I've updated my PR

Libraries/Image/RCTImageUtils.m
```diff
@@ -314,14 +330,18 @@ BOOL RCTUpscalingRequired(CGSize sourceSize, CGFloat sourceScale,
   return (__bridge_transfer id)imageProperties;
 }
 
-NSData *__nullable RCTGetImageData(CGImageRef image, float quality, NSInteger orientation)
+NSData *__nullable RCTGetImageData(CGImageRef image, float quality, UIImageOrientation imageOrientation)
```
shergin: What a second, what if we make `RCTGetImageData` accept `UIImage` instance instead of `CGImage` and `UIImageOrientation`? Looks like it will be much simpler!

Libraries/Image/RCTImageUtils.m
```diff
@@ -314,18 +330,25 @@ BOOL RCTUpscalingRequired(CGSize sourceSize, CGFloat sourceScale,
...
 {
   NSDictionary *properties;
   CGImageDestinationRef destination;
   CFMutableDataRef imageData = CFDataCreateMutable(NULL, 0);
+  CGImagePropertyOrientation orientation = CGImagePropertyOrientationFromUIImageOrientation(imageOrientation);
   if (RCTImageHasAlpha(image)) {
     // get png data
     destination = CGImageDestinationCreateWithData(imageData, kUTTypePNG, 1, NULL);
+    properties = @{
```
shergin: And, we have to move it above to avoid duplication.

Libraries/Image/RCTImageUtils.m
doochik: Good point!

doochik:
I've updated PR and squash commits

doochik:
@shergin Thanks a lot!

PR #15062, opened 2017-07-17
[Geolocation] Use Apple's significant-change API (for iOS 11 UX)

jasongaare:
## Context: 
In the yet-to-be-released iOS 11, Apple has changed the way they notify the user of location services. (You can watch their session from WWDC about all the changes [here](https://developer.apple.com/videos/play/wwdc2017/713/).)

The current implementation of `RCTLocationObserver` uses the standard location services from Apple. When the user has granted `Always` location permission and the application uses the background location service, the user is presented with the *_Blue Bar of Shame_* (for more information check out [this blog post](https://blog.set.gl/ios-11-location-permissions-and-avoiding-the-blue-bar-of-shame-1cee6cd93bbe)):

image

### navigator.geolocation.watchPosition(success, error?, options?):
* Added `useSignificantChanges` boolean to the options passed.

### RCTLocationObserver.m
* Added `_usingSignificantChanges` boolean based on user options. If `true`, then the CLLocationManager will use functions `startMonitoringSignificantLocationChanges`/ `stopMonitoringSignificantLocationChanges` rather than the standard location services.
* Changed method signature of `beginLocationUpdatesWithDesiredAccuracy` to include `useSignificantChanges` flag
* Added check for new `NSLocationAlwaysAndWhenInUseUsageDescription`

## Test Plan:
All unit tests passed.

Tested in simulator and on device, toggling `useSignificantChanges` option when calling `watchPosition`. Results were as expected. **When `TRUE`, the _Blue Bar of Shame_ was not present.**

Changes do not affect Android and location services still work as expected on Android.

## Caveats:
* Change is for iOS only
* Using a different API will have different accuracy results. Adding `useSignificantChanges` as an option was by design so apps that want to have most accurate and most frequent update can still use standard location services.

### DIFF
--- a/Libraries/Geolocation/RCTLocationObserver.m
+++ b/Libraries/Geolocation/RCTLocationObserver.m
@@ -32,6 +32,7 @@ typedef struct {
   double maximumAge;
   double accuracy;
   double distanceFilter;
+  bool useSignificantChanges;
 } RCTLocationOptions;
 
 @implementation RCTConvert (RCTLocationOptions)
@@ -47,7 +48,8 @@ typedef struct {
     .timeout = [RCTConvert NSTimeInterval:options[@"timeout"]] ?: INFINITY,
     .maximumAge = [RCTConvert NSTimeInterval:options[@"maximumAge"]] ?: INFINITY,
     .accuracy = [RCTConvert BOOL:options[@"enableHighAccuracy"]] ? kCLLocationAccuracyBest : RCT_DEFAULT_LOCATION_ACCURACY,
-    .distanceFilter = distanceFilter
+    .distanceFilter = distanceFilter,
+    .useSignificantChanges = [RCTConvert BOOL:options[@"useSignificantChanges"]] ?: NO,
   };
 }
 
@@ -108,6 +110,7 @@ static NSDictionary<NSString *, id> *RCTPositionError(RCTPositionErrorCode code,
   NSDictionary<NSString *, id> *_lastLocationEvent;
   NSMutableArray<RCTLocationRequest *> *_pendingRequests;
   BOOL _observingLocation;
+  BOOL _usingSignificantChanges;
   RCTLocationOptions _observerOptions;
 }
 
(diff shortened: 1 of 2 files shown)


Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -32,6 +32,7 @@ typedef NS_ENUM(NSInteger, RCTPositionErrorCode) {
   double maximumAge;
   double accuracy;
   double distanceFilter;
+  bool useSignificantChanges;
```
javache: Use `BOOL` instead. There's some really subtleties with `bool` so we try to be consistent.

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -117,8 +120,12 @@ @implementation RCTLocationObserver
 
 - (void)dealloc
 {
-  [_locationManager stopUpdatingLocation];
+  _usingSignificantChanges ?
+    [_locationManager stopMonitoringSignificantLocationChanges] :
+    [_locationManager stopUpdatingLocation];
+
   _locationManager.delegate = nil;
+  _usingSignificantChanges = NO;
```
javache: No need to reset this in dealloc.

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -133,14 +140,20 @@ - (dispatch_queue_t)methodQueue
...
 {
   [self requestAuthorization];
 
   _locationManager.distanceFilter  = distanceFilter;
   _locationManager.desiredAccuracy = desiredAccuracy;
   // Start observing location
-  [_locationManager startUpdatingLocation];
+  if (useSignificantChanges) {
+    [_locationManager startMonitoringSignificantLocationChanges];
+    _usingSignificantChanges = YES;
```
javache: Just have `_usingSignificantChanges = useSignificantChanges` somewhere above.

jasongaare:
Changes found in https://github.com/facebook/react-native/pull/15062/commits/98cb37701dec177342327933c46a14e788415f7f.

Setting `_usingSignificantChanges` now on  [line 148](https://github.com/jasongaare/react-native/blob/98cb37701dec177342327933c46a14e788415f7f/Libraries/Geolocation/RCTLocationObserver.m#L148), however I could also refactor to setting `_usingSignificantChanges` prior to calling `beingLocationUpdatesWithDesiredAccuracy` and reverting that method to its original method signature.

javache:
Some flow checks are failing internally, I need to update the internal callers to land this/

jasongaare:
Any update on the status of this?

PR #15065, opened 2017-07-17
Apple TV Cocoapods support

dlowder-salesforce:
**Motivation**

Support Apple TV for people adding React Native to their projects using Cocoapods.

**Test plan**

Working test project at https://github.com/dlowder-salesforce/react-native-tvos-cocoapods-test

### DIFF
--- a/React.podspec
+++ b/React.podspec
@@ -52,6 +52,14 @@ Pod::Spec.new do |s|
                               "React/Cxx*/*",
                               "React/Base/RCTBatchedBridge.mm",
                               "React/Executors/*"
+    ss.ios.exclude_files    = "React/**/RCTTVView.*"
+    ss.tvos.exclude_files   = "React/Modules/RCTClipboard*",
+                              "React/Views/RCTDatePicker*",
+                              "React/Views/RCTPicker*",
+                              "React/Views/RCTRefreshControl*",
+                              "React/Views/RCTSlider*",
+                              "React/Views/RCTSwitch*",
+                              "React/Views/RCTWebView*"
     ss.header_dir           = "React"
     ss.framework            = "JavaScriptCore"
     ss.libraries            = "stdc++"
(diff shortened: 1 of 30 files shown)


dlowder-salesforce:
@javache since you've worked on podfiles recently, can you look at this?

React.podspec
```diff
@@ -48,7 +48,6 @@ Pod::Spec.new do |s|
                               "IntegrationTests/*",
                               "React/DevSupport/*",
                               "React/Inspector/*",
-                              "React/**/RCTTVView.*",
```
javache: Can you mark the TVView files as tvOS only? I believe cocoapods supports specifying this.

React/Modules/RCTClipboard.m
```diff
@@ -23,15 +23,19 @@ - (dispatch_queue_t)methodQueue
 
 RCT_EXPORT_METHOD(setString:(NSString *)content)
 {
+#if !TARGET_OS_TV
```
javache: It would be better to just exclude the whole module. That way you can feature-check it from JS (if `NativeModules.Clipboard`

javache:
Use multi-platform support to override these rules instead: https://guides.cocoapods.org/syntax/podspec.html#group_multi_platform_support

React.podspec
dlowder-salesforce: It's probably simpler to just wrap those files in `if TARGET_OS_TV`

React/Modules/RCTClipboard.m
dlowder-salesforce: OK

dlowder-salesforce:
Thanks for the tip, @javache .... I removed all the source file changes, now only podspecs are changed.

React.podspec
dlowder-salesforce: Never mind -- this is done now.

javache:
you're awesome

PR #15066, opened 2017-07-17
Show bundle download progress on iOS

janicduplessis:
This shows progress for the download of the JS bundle (different from the packager transform progress that we show already). This is useful especially when loading the JS bundle from a remote source or when developing on device (on simulator + localhost it pretty much just downloads instantly). This will be nice for the expo client since all bundles are loaded over the network and can take several seconds to load.

This depends on https://github.com/facebook/metro-bundler/pull/28 to work but won't crash or anything without it, it just won't show the progress percentage.

img_05070155d2cc-1

**Test plan**
Tested that bundle download progress is shown properly in RNTester on both localhost + simulator and on real device with network conditionner to simulate a slow loading bundle.

Tested that it doesn't cause issues if the packager doesn't send the Content-Length header.


React/Base/RCTJavaScriptLoader.mm
```diff
@@ -270,6 +274,16 @@ static void attemptAsynchronousLoadOfBundleAtURL(NSURL *scriptURL, RCTSourceLoad
   return progress;
 }
 
+static RCTLoadingProgress *progressEventFromDownloadProgress(NSNumber *total, NSNumber *done)
+{
+  RCTLoadingProgress *progress = [RCTLoadingProgress new];
+  progress.status = @"Downloading JavaScript bundle";
+  // Progress values are in bytes transform them to kilobytes for smaller numbers.
+  progress.done = done != nil ? @([done integerValue] / 1024) : nil;
```
janicduplessis: Mostly just to make the numbers smaller, we don't really have a way to show units with the current setup so it's no really clear what the numbers are anyway.

shergin: Why we don't make them smaller right before displaying them?

janicduplessis: This is as close as we get to displaying them, this is also used for showing the packager module transform progress so it just takes a progress number and the total then formats it to something like "10% (10/100)".

janicduplessis:
@shergin Could you check why this didn't import

shergin:
@janicduplessis This strange but one of our internal builds failes with this error:
```
Libraries/FBReactKit/js/react-native-github/React/Base/RCTMultipartDataTask.h:14:81: error: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified) [-Werror,-Wnullability-completeness]
```
Do you have any idea? Seems this diff even don't modifies this line.

janicduplessis:
@shergin Fixed the error. I just removed the nullability specifier since we didn't use any in that file to begin with.

Seems like a new warning that can be enabled in xcode 9. Didn't update the projects to enable it for now but we might want to do that in the future when xcode 9 is released and update the projects to the new defaults.

React/Base/RCTMultipartStreamReader.m
```diff
@@ -56,14 +65,34 @@ - (void)emitChunk:(NSData *)data callback:(RCTMultipartCallback)callback done:(B
...
+  CFTimeInterval currentTime = CACurrentMediaTime();
+
+  NSUInteger headersContentLength = headers[@"Content-Length"] != nil ? [headers[@"Content-Length"] unsignedIntValue] : 0;
+  if (callback && (currentTime - _lastDownloadProgress > 0.016 || final)) {
+    _lastDownloadProgress = currentTime;
+    callback(headers, @(headersContentLength), @(contentLength));
+  }
+}
+
+- (BOOL)readAllParts:(RCTMultipartCallback)callback progressCallback:(RCTMultipartProgressCallback)progressCallback
```
shergin: If we changed the signature of this method, may be we can name it a bit better including `callback` word into this? Something like `readAllPartsWithCallback:progressCallback:` or something even better?

ide: Maybe "CompletionCallback" so it's clear how it's different from the progress callback?

janicduplessis: Oups forgot to update the signature in tests, also updated it to `readAllPartsWithCompletionCallback:progressCallback:`. Should land this time!

PR #15067, opened 2017-07-18
[Geolocation] Provide option to request specific location permissions

yatchoi:
# Context
On IOS, the React Native Geolocation library will make system permission requests based on the capability of the app as defined in `Info.plist`.

This PR allows the option for feature developers to specify the exact level of location permission request to make, by passing an argument in as part of the options parameter. This is important in scenarios where user privacy and permission regarding location-based features needs to be specifically scoped.

This will prevent situations where apps will ask users for background location permission, even though certain features may only require location in the foreground.

# Changes
- Add `locationAuthorizationIOS` as an option in `GeoOptions`
- Expose `Geolocation.IOS_AUTHORIZATIONS` as a String enum
- Add `locationAuthorization` as an argument for `beginLocationUpdatesWithDesiredAccuracy`
- Make location requests either based on the option, or fall back to the old Info.plist approach

# Test Plan
Tested in simulator, using an app with Background Location capability, called `beginLocationUpdatesWithDesiredAccuracy`with various authorizations requested.

### DIFF
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -31,6 +31,7 @@ type GeoOptions = {
   maximumAge: number,
   enableHighAccuracy: bool,
   distanceFilter: number,
+  locationAuthorizationIOS: string,
 }
 
 /**
@@ -185,5 +186,9 @@ var Geolocation = {
     }
   }
 };
+
+Geolocation.IOS_AUTHORIZATIONS = {};
+Geolocation.IOS_AUTHORIZATIONS.WHEN_IN_USE = "locationAuthorizationWhenInUse";
+Geolocation.IOS_AUTHORIZATIONS.ALWAYS = "locationAuthorizationAlways";
 
 module.exports = Geolocation;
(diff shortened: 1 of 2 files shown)


rigdern:
There's a second PR for fixing this problem (#15096). We should figure out which one to take. I believe the primary difference between these PRs is:
  - This one tweaks `Geolocation` such that you can tell it which permissions to request.
  - PR #15096 tweaks `Geolocation` such that you can tell it not to request permissions on your behalf. Instead, the app has to request permissions on its own.

I prefer the solution used in my PR (#15096). It's consistent with the Android implementation of `Geolocation` which doesn't request any permissions. Instead, apps are expected to use the `PermissionsAndroid` module to request the needed permissions. I think we should go in a similar direction for iOS. React Native should eventually have a cross-platform `Permissions` module that an app can use to request any permission it wants. I like that this gives apps a lot of flexibility in that they get complete control over their permission requests. My team has used this approach in our app and it has worked well for us.

@yatchoi do you have any thoughts on this?

Adam Comella
Microsoft Corp.

PR #15072, opened 2017-07-18
Add missing file to xcodeproj

janicduplessis:
5701ae2145c7af4fa86b5a82f5883e62af7a61f0 didn't add the new files to xcodeproj, the project is still building fine but is getting rejected by apple app analysis tools because it thinks we are trying to use a private api `rootView`. Just adding the files that define the selector makes it get accepted now.

**Test plan**
Tested that I'm now able to submit a build on testflight using this change.


janicduplessis:
cc @Yu-w @shergin

PR #15077, opened 2017-07-18
Native Animated - Override __makeNative in AnimatedInterpolation

syaau:
Summary:
Fixes the error `Trying to update interpolation node that has not been
attached to the parent` in android which occurs when using multiple
Animated.Values along with interpolation and an animation is run before
another one that uses interpolation. On ios, no error is thrown in such
case but the animation also doesn't work as expected.

You can check the snack code here which works properly without
useNativeDriver: true. But fails on android and skips the first stage
of animation on ios.
  https://snack.expo.io/HyD3zdjSZ

**Test Plan**
The animations worked properly after the __makeNative override made
the parent node native as well.

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -1124,6 +1124,11 @@ class AnimatedInterpolation extends AnimatedWithChildren {
     this._interpolation = Interpolation.create(config);
   }
 
+  __makeNative() {
+    this._parent.__makeNative();
+    super.__makeNative();
+  }
+
   __getValue(): number | string {
     var parentValue: number = this._parent.__getValue();
     invariant(


janicduplessis:
Thanks for the fix! Looks good.

Noitidart:
Wow so timely, I just ran into this bug today. What version does this land in?

syaau:
Most likely 0.48. A quick hack into the `node_modules/react-native/Libraries/Animated/src/AnimatedImplementation.js` file can fix this issue if you are in a hurry.

Noitidart:
@syaau thanks! Yep that's how I'm doing it. Very cool thanks!

Noitidart:
Hey all, I was wondering if I'm right here, this PR landed in v0.49.1? Is this correct?

janicduplessis:
@Noitidart Yes, you can check the tags on the commit to know in which release it is.

image

Noitidart:
Awesome thank you sir!

Issue #15081, opened 2017-07-18
TabBarIOS (Tabs) in AppleTV flickering when switching between tabs


rada:
### Is this a bug report?

Yes

### Have you read the Bugs section of the Contributing to React Native Guide?

Yes

### Environment

1. `react-native -v`: 0.46.1
2. `node -v`: 6.5.0
3. `npm -v`: 3.10.3
4. `yarn --version` (if you use Yarn):

Then, specify:

- Target Platform: AppleTV

- Development Operating System: macOS Sierra (10.12.5)

- Build tools: Xcode (8.3.3)

### Steps to Reproduce

1. Switch to any Tab back and forth
2. Switch to any Tab back and forth

### Expected Behavior

The switch to any tab is smooth and without any flickering or selection falling back to previous tab.

### Actual Behavior

Switching between TabBarIOS.Items is flickering and is not smooth. While selecting the new tab, It seems like first the new tab is selected, then immediately the old one and then again the new tab. This results in tab flickering when switching between the tabs. Worse case is, when the tabs render some more complex composed element. In this case, the new tab is not being selected, as the render is taking longer. Once it is rendered, we can select it but again the flickering occurs.

It looks like by the time the JS processes the received native event (tab switch) and sends an event (prop) back to Native, the tab selection falls back to the previous tab.

I have included very basic TabBar example with plain view and one text where flickering occurs. However, given the simple View is used in tabs here, switching is occasionally smooth.

### Reproducible Demo

```javascript
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 * @flow
 */

import React, { Component } from 'react';
import {
  AppRegistry,
  Text,
  View,
  TabBarIOS,
} from 'react-native';

export default class tvTabBar extends Component {

	_tabItems: Array<{ key: string }> = [
		{ key: 'TAB 1' },
		{ key: 'TAB 2' },
		{ key: 'TAB 3' },
		{ key: 'TAB 4' },
	];

	constructor(props) {
		super(props);
		this.state = { activeTab: 'TAB 1' };
	}

	_getTabItem = (key: string) => {
		switch (key) {
			case 'TAB 1':
				return (
					<View style={{ flex: 1, top: 140, borderColor: 'red', borderWidth: 10 }}>
						<Text style={{ fontSize: 30 }}>TAB 1</Text>
					</View>
				);
			case 'TAB 2':
				return (
[...]

PR #15093, opened 2017-07-19
Fix Apple TV TabBarIOS issue (#15081)

dlowder-salesforce:
**Motivation**

Fix flickering in TabBarIOS on Apple TV... issue #15081 

**Test plan**

Existing CI should still pass.  Issue is resolved when testing the example code in #15081 .


React/Views/RCTTabBar.m
```diff
@@ -175,6 +175,18 @@ - (void)setItemPositioning:(UITabBarItemPositioning)itemPositioning
 
 #pragma mark - UITabBarControllerDelegate
 
+#if TARGET_OS_TV
+
+- (void)tabBarController:(UITabBarController *)tabBarController didSelectViewController:(nonnull UIViewController *)viewController
```
dlowder-salesforce: `didSelectViewController` seems to work fine on iOS too... if desired, we could just use this method for both iOS and tvOS.

hramos:
Removed unrelated Danger warning.

React/Views/RCTTabBar.m
javache: Yes, I'd prefer that :)

dlowder-salesforce: Done

javache:
Hmm, this may cause subtle differences. It seems like before we actually rejected all calls to `shouldSelectViewController`, and then the JS side would actually set the tabbar item it wanted, but in the new model, we would always switch tabs, no longer leaving JS a choice.

dlowder-salesforce:
Wow... so JS could actually select a different tab from the one the user picked?  Interesting!

Since tvOS tab bar works very differently, I'd suggest that we go back to my original commit that just fixes tvOS and leaves iOS behavior unchanged.

javache:
Well the idea is that TabBarItemIOS is a controlled component: all state changes need to come from JS and not from native. I'm not sure why this doesn't work correctly on tvOS.

dlowder-salesforce:
In tvOS testing, I find that the way the tab bar works is very different.  If you use the TV remote to navigate from one tab bar item to the next, the new tab bar item is highlighted BEFORE `shouldSelectViewController:` is called.  So there's no way to avoid changing the state of the tab bar prior to JS being called.  For this reason, I think using `didSelectViewController:` is better for tvOS.

javache:
Ok, makes sense. It does mean it's possible for JS state and native state to go out of sync though. One way classes like RCTTextView solve this is by accepting the native state change, and having the JS side immediately call setNativeProps with the desired state (even if it's the same as the old one).

dlowder-salesforce:
Did additional testing of the issue... this patch is part of the solution, but some additional changes are needed.  Closing this, and will submit the complete fix later.

PR #15094, opened 2017-07-19
Android: Improve getCurrentPosition API

rigdern:
We ran into a couple of problems with the implementation of `getCurrentPosition` on Android:
  - It sometimes returns an inaccurate location
  - It times out when `enableHighAccuracy` is `true` (#7495)

This change improves `getCurrentPosition` for both of the above problems. Instead of calling `requestSingleUpdate` it now calls `requestLocationUpdates` so it can receive multiple locations giving it an opportunity to pick a better one. Unlike `requestSingleUpdate`, this approach doesn't seem to timeout when `enableHighAccuracy` is `true`.

**Test plan (required)**

Verified in a test app that `getCurrentPosition` returns a good location and doesn't timeout when `enableHighAccuracy` is `true`. Also, my team has been using this change in our app in production.

Adam Comella
Microsoft Corp.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/location/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/location/BUCK
@@ -7,6 +7,7 @@ android_library(
         "PUBLIC",
     ],
     deps = [
+        react_native_dep("libraries/fbcore/src/main/java/com/facebook/common/logging:logging"),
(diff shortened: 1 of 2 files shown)


hramos:
Removed unrelated Danger warning.

dabit1:
+1!!!!!!!!!!!!!!!!!!!!!!!!!

PR #15096, opened 2017-07-19
iOS: Geolocation: Allow skipping of permission prompts

rigdern:
This change enables developers to tell the Geolocation module to skip its permissions requests so the app can manage the permissions requests on its own. React Native Android's Geolocation module already requires apps to request permissions on their own.

Currently, the Geolocation module automatically makes permissions requests for you based on what you've specified in your property list file. However, the property list file doesn't always represent what permissions the app wants right now. For example, suppose an app sometimes wants location when "in use" and sometimes wants location "always" depending on what app features the user has engaged with. The Geolocation API will request the "always" location permission even if that's not what the app wants for the current scenario.

This change enables developers to tell the Geolocation module to skip permission requests so that the app can explicitly ask for the appropriate permissions. By default, Geolocation will still ask for permissions so this is not a breaking change.

This change adds a method to Geolocation that is not supported by the web spec for geolocation (`setRNConfiguration`). This method includes `RN` in the name to minimize the chances that the web spec will introduce an API with the same name.

**Test plan (required)**

Verified that Geolocation requests permissions by default and when `skipPermissionRequests` is `false`. Verified the permission requests are skipped when `skipPermissionRequests` is `true`. Also, my team is using this change in our app.

Adam Comella
Microsoft Corp.

### DIFF
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -26,6 +26,10 @@ const PermissionsAndroid = require('PermissionsAndroid');
 var subscriptions = [];
 var updatesEnabled = false;
 
+type GeoConfiguration = {
+  skipPermissionRequests: bool;
+}
+
 type GeoOptions = {
   timeout?: number,
   maximumAge?: number,
@@ -74,6 +78,23 @@ type GeoOptions = {
  *
  */
 var Geolocation = {
+
+  /*
+    * Sets configuration options that will be used in all location requests.
+    *
+    * ### Options
+    *
+    * #### iOS
+    *
+    * - `skipPermissionRequests` - defaults to `false`, if `true` you must request permissions
+    * before using Geolocation APIs.
+    *
+    */
+  setRNConfiguration: function(
+    geo_config: GeoConfiguration
+  ) {
+    RCTLocationObserver.setRNConfiguration(geo_config);
+  },
 
   /*
    * Request suitable Location permission based on the key configured on pList.
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -93,8 +93,8 @@ var Geolocation = {
   setRNConfiguration: function(
     config: GeoConfiguration
   ) {
-    if (RCTLocationObserver.setRNConfiguration) {
-      RCTLocationObserver.setRNConfiguration(config);
+    if (RCTLocationObserver.setConfiguration) {
+      RCTLocationObserver.setConfiguration(config);
     }
   },
 
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -91,9 +91,11 @@ var Geolocation = {
     *
     */
   setRNConfiguration: function(
-    geo_config: GeoConfiguration
+    config: GeoConfiguration
   ) {
-    RCTLocationObserver.setRNConfiguration(geo_config);
+    if (RCTLocationObserver.setRNConfiguration) {
+      RCTLocationObserver.setRNConfiguration(config);
+    }
   },
 
   /*
--- a/Libraries/Geolocation/RCTLocationObserver.m
+++ b/Libraries/Geolocation/RCTLocationObserver.m
@@ -188,7 +188,7 @@ RCT_EXPORT_MODULE()
 
 #pragma mark - Public API
 
-RCT_EXPORT_METHOD(setRNConfiguration:(RCTLocationConfiguration)config)
+RCT_EXPORT_METHOD(setConfiguration:(RCTLocationConfiguration)config)
 {
   _locationConfiguration = config;
 }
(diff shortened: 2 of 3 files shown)


Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -135,7 +149,9 @@ - (dispatch_queue_t)methodQueue
 
 - (void)beginLocationUpdatesWithDesiredAccuracy:(CLLocationAccuracy)desiredAccuracy distanceFilter:(CLLocationDistance)distanceFilter
 {
-  [self requestAuthorization];
+  if (!_locationConfiguration.skipPermissionRequests) {
```
chirag04: wondering if we can just check if permissions are already granted instead of a skip flag

rigdern: @chirag04 That's an interesting suggestion. A downside to that design is it'll open the door to a tricky to catch app bug. Suppose the app wants to request permissions on its own but fails to. With the current implementation, the app will see an exception due to not having the proper permissions. With your suggestion, RN will request permissions on behalf of the app making it difficult for the developer to realize there's a bug in their app.

Libraries/Geolocation/Geolocation.js
```diff
@@ -75,6 +79,23 @@ type GeoOptions = {
...
+    * ### Options
+    *
+    * #### iOS
+    *
+    * - `skipPermissionRequests` - defaults to `false`, if `true` you must request permissions
+    * before using Geolocation APIs.
+    *
+    */
+  setRNConfiguration: function(
+    geo_config: GeoConfiguration
```
javache: Use camelCase. In this case, just name the argument `config`

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -160,6 +176,11 @@ - (void)timeout:(NSTimer *)timer
 
 #pragma mark - Public API
 
+RCT_EXPORT_METHOD(setRNConfiguration:(RCTLocationConfiguration)config)
```
javache: Why the RN in the name? Can this be just `setConfiguration:`

ReactAndroid/src/main/java/com/facebook/react/modules/location/LocationModule.java
```diff
@@ -104,6 +104,11 @@ private static LocationOptions fromReactMap(ReadableMap map) {
     }
   }
 
+  @ReactMethod
```
javache: Remove this completely and instead check on the JS side if the method exists before calling it.

Libraries/Geolocation/RCTLocationObserver.m
rigdern: @javache "This method includes RN in the name to minimize the chances that the web spec will introduce an API with the same name."

hramos:
I've taken a look at the current status of this PR, and based on the comments left by other reviewers as well as the conflicts, I'm marking it as "needs-revision".

rigdern:
@hramos thanks for the ping.

@javache thanks for reviewing this. I pushed an update which addresses your comments. As described in my other comment, I left the method `setRNConfiguration` with the RN prefix. I justified why I used the prefix but if you'd prefer, I can remove the prefix. Let me know.

Libraries/Geolocation/RCTLocationObserver.m
javache: Sorry, missed that in your original description. I think it's a good idea to avoid potential conflicts with the web spec. We can name the native module however we want though, so let's call the native method `setConfiguration` and the API on geolocation `setReactConfiguration`?

rigdern:
PR #15067 also solves this problem. We should figure out which PR to take. See my comment (https://github.com/facebook/react-native/pull/15067#issuecomment-325207098) for a comparison of these PRs.

ngandhy:
@rigdern - when I add the line "navigator.geolocation.setRNConfiguration({'skipPermissionRequests':true});" to my code the location times out every time. 

The second I remove the line of code, everything works again. Any tips / advice?

ngandhy:
@rigdern @javache @hramos - Found the issue. _locationManager is never being set if you skip the permission request using skipPermissionsRequest:true. This is because _locationManager is being set INSIDE the authorization method. 

Adding the following inside beginLocationUpdatesWithDesiredAccuracy fixes the issue.

if (!_locationManager) {
    _locationManager = [CLLocationManager new];
    _locationManager.delegate = self;
  }

rigdern:
@ngandhy That looks like a bug. Good find! Do you want to send a PR?

Adam Comella
Microsoft Corp.

ngandhy:
kk, I'll give it a try :)

Also, I've seen your features, fixes, and all the other stuff you're doing for the community and just wanted to say THANKS!  I'm sure you guys don't hear it as much as you should

rigdern:
> Also, I've seen your features, fixes, and all the other stuff you're doing for the community and just wanted to say THANKS! I'm sure you guys don't hear it as much as you should

@ngandhy, thanks, I'm glad you've found our work to be valuable :)

PR #15111, opened 2017-07-19
[Docs] How to Contribute updates

hramos:
Rename "Testing" guide. Add docs for several new GitHub bot commands. Format commands table.

## Test Plan

Built website and verified everything looks good.


forki:
can you please add a paragraph about what we can do when pull requests with fixes are lying around for long time and nobody from FB is looking on it?

PR #15112, opened 2017-07-20
Validate the content type of the bundle returned by the packager

janicduplessis:
When using the packager behind something like an http tunnel it is possible something else than JS is returned, in that case throw an error instead of trying to parse it.

This is useful for Expo since the packager runs behind ngrok. This allows to intercept that error and show a more meaningful error message based on the ngrok response.

**Test plan**
Tested by changing the packager to return text/html content type and validate that the error shows up properly.
Also tested that it works when multipart response is disabled.

screen shot 2017-07-19 at 8 01 58 pm


shergin:
@facebook-github-bot shipit

PR #13876, opened 2017-05-09
Request toolbar layout after drawable loads

Benjamin-Dobell:
## Motivation

Fixes #11209

Updating action items in a `ToolbarAndroid`, or more specifically the native `ReactToolbar`, after the initial render presently will not always work as expected - typically manifesting itself as new action items not being displayed at all (under certain circumstances).

This seems to be happening because Fresco gets back to us asynchronously and updates the `MenuItem` in a listener. However when a keyboard is displayed the `Toolbar` is in a weird state where updating the `MenuItem` doesn't automatically trigger a layout.

The solution is to trigger one manually.

## Test Plan

This is a bit wacky, so I created a sample project:

https://github.com/Benjamin-Dobell/DynamicToolbar

`master` demonstrates the problem. Run the project, the toolbar action item is scheduled to update every 2 seconds. It works fine _until_ you give the `TextInput` focus. Once you give it focus the action item disappears and it never recovers (despite on-going renders due to state changes).

You can then checkout the `fixed` branch, run `yarn` again, and see that problem is fixed. This branch is using a prebuilt version of React Native with this patch applied. Of course you could (and probably should) also modify `master` to use a version of RN built by the Facebook CI (assuming that's a thing you guys do).

PR #15120, opened 2017-07-20
[WIP] Request toolbar layout after drawable loads

forki:
This applies https://github.com/facebook/react-native/pull/13876 in order to test it on CI against master. This is only temporay PR.


forki:
Ok all green. #18376 looks really good from what I can see

PR #15123, opened 2017-07-20
Fix panResponder nativeEvent.locationX and locationY values on touch move

### LINKED ISSUE #12591: panResponder nativeEvent.locationX and nativeEvent.locationY values do not update on Android
We use GitHub Issues for  

I am referencing #9786 but because of the unspecific title I worry the react-native team is not paying attention to this bug.

### Description
in onPanResponderMove evt.nativeEvent.locationX and evt.nativeEvent.locationY do not update on Android. pageX and pageY do update on Android. @Arkanine 's suggestion to return true at the end of the panresponder does not fix this issue.

[FILL THIS OUT: Explain what you did, what you expected to happen, and what actually happens.]

### Reproduction

Many people are having this issue and no feedback from the team, I am researching this further but not an Android wizard and haven't looked at the android end of the source for react-native too much.

### Additional Information

* React Native version: 0.38.1
* Platform: Android only
* Operating System: MacOS

newyankeecodeshop:
Fixes #12591

The Android JSTouchDispatcher was using `mTargetCoordinates` when creating the TouchEvent for a move. However, these are final values which are set when the touch down is received. When the user's finger moves, it's important to be able to track the coordinates of the touch as it moves. Thus, we need to update the x,y coordinates by calling `TouchTargetHelper` on each move event.

Test Plan:
All unit tests pass.
I tested this by validating that the `locationX` and `locationY` properties of the nativeEvent correspond to the location of the finger as it moves across the screen in our application. We use a PanResponder to move a view as the user's finger moves across the screen.

On iOS, the `locationX` and `locationY` properties already track the coordinates of the touch event.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java
@@ -116,14 +116,21 @@ public class JSTouchDispatcher {
       mGestureStartTime = TouchEvent.UNSET;
     } else if (action == MotionEvent.ACTION_MOVE) {
       // Update pointer position for current gesture
+      float[] moveTargetCoordinates = new float[2];
+      TouchTargetHelper.findTargetTagAndCoordinatesForTouch(
+        ev.getX(),
+        ev.getY(),
+        mRootViewGroup,
+        moveTargetCoordinates,
+        null);
       eventDispatcher.dispatchEvent(
         TouchEvent.obtain(
           mTargetTag,
           TouchEventType.MOVE,
           ev,
           mGestureStartTime,
-          mTargetCoordinates[0],
-          mTargetCoordinates[1],
+          moveTargetCoordinates[0],
+          moveTargetCoordinates[1],
           mTouchEventCoalescingKeyHelper));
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/uimanager/JSTouchDispatcher.java
```diff
@@ -116,14 +116,21 @@ public void handleTouchEvent(MotionEvent ev, EventDispatcher eventDispatcher) {
       mGestureStartTime = TouchEvent.UNSET;
     } else if (action == MotionEvent.ACTION_MOVE) {
       // Update pointer position for current gesture
+      float[] moveTargetCoordinates = new float[2];
```
shergin: Should we update `mTargetCoordinates` instead? If we do so, `UP` event also will have proper coordinates, right?

newyankeecodeshop: I considered that, but I am not that familiar with this code and its history. I'm not sure if there is other code and/or libraries relying on `mTargetCoordinates` being fixed during a touch. I generally agree that it would make sense to update the coordinates on move and up too. Let me know if I should make that change.

PR #15132, opened 2017-07-21
fix font typo

caoer:
This clearly seems to be a  typo. If `didFintFont` is NO, and system `has` font, we will create the font..


caoer:
sorry, just found out that RN don't have fontName so this should be correct for what it means. closing PR.

PR #15136, opened 2017-07-21
[metro] Control whether Metro tells Babel to lookup .babelrc files

ide:
This adds support to RN's configuration file to let people turn off Babel's behavior of looking up .babelrc files. Most of the support for this feature is in Metro (https://github.com/facebook/metro-bundler/pull/31).

Test Plan: See the Metro test plan (tested with RN).


ide:
If this failed Flow checks, it may be because we need the latest copy of Metro (v0.11.0 I believe).

PR #15144, opened 2017-07-21
Add documentation for onMomentumScrollEnd

tomasreimers:

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -236,6 +236,10 @@ const ScrollView = createReactClass({
      */
     minimumZoomScale: PropTypes.number,
     /**
+     * Callback for when the scrolling animation has ended.
+     */
+    onMomentumScrollEnd: PropTypes.func,
+    /**
      * Fires at most once per frame during scrolling. The frequency of the
      * events can be controlled using the `scrollEventThrottle` prop.
      */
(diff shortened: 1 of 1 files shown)


tomasreimers:
Absolutely! Also, FWIW, I don't think that onScrollAnimationEnd is ever called...

https://snack.expo.io/BJyf5OGIW

https://github.com/facebook/react-native/blob/603cc48ceba001827d10231d51b4031c7768eef8/React/Views/RCTScrollView.m#L765

shergin:
@tomasreimers Yeah... You are right. Thank you!
I think, we also have to do:
 * Remove `onScrollAnimationEnd` from documentation and delete all (unused) mentions of it from Android and iOS sources;
 * Implement `onMomentumScrollBegin` on iOS (it already should work on Android) and add this to documentation.

I will add it to my plans. (Or a PR is always welcome. 😄)

tomasreimers:
@shergin 

(1) Done: https://github.com/facebook/react-native/pull/15156 (Nothing at FB is someone else's problem, right? ;) )

(2) Already done: https://github.com/facebook/react-native/blob/968aa56ca38b6ad7f08a08081321e799ee015322/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java#L263

shergin:
Awesome! Now we just have to mention this in the documentation.

tomasreimers:
@shergin  https://github.com/facebook/react-native/pull/15158

PR #15151, opened 2017-07-22
Update Images.md with ImageBackground

Minishlink:
This part of the documentation was outdated since v0.46. :)

### DIFF
--- a/docs/Images.md
+++ b/docs/Images.md
@@ -186,7 +186,9 @@ On the user side, this lets you annotate the object with useful attributes such 
 
 ## Background Image via Nesting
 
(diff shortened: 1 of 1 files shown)


shergin:
@Minishlink 
Thank you!
This is reasonable changes but I would move the point a little bit. I would suggest to implement this behaviour manually for each particular case and highlight that fact that actual implementation of <ImageBackground> is super simple. And I would advice to use <ImageBackground> only after taking a look of the the source code to make the best choice suited each particular case.

Could you please mention this in the documentation?

Minishlink:
@shergin Sure, does this seem good?

PR #15161, opened 2017-07-24
iOS: Fix RCTDeviceInfo Dimensions for iPad multitasking

kipricker:
**Issue:**
On iOS Dimensions `window` and `screen` both get resolved to the UIScreen bounds. This will always give us the dimensions in terms of the current physical screen currently in use. For the `window` value of Dimensions this sh    ould be considered incorrect. With the introduction of multitasking on iOS we now have the ability to have our app rendered on only part of the screen and the `window` value should reflect the dimensions our app is currently rendering in.

**Example:**
When run on iPad this gist will render a red view with a black border using `Dimensions.get('window').width`. When rotating, this view will display correctly. If you instead run this gist while in multitasking mode, the view will render incorrectly as `Dimensions.get('window').width` is still returning the full width of the physical screen.
https://gist.github.com/kipricker/c87b606ba439e7b09fd29303bf324065

**Proposed fix:**
For updating, instead of using the rotation of the screen to determine an update in our `window` dimensions, we can have RCTWindowManager report our `availableSize` when it changes. This will update us properly on window size changes including size changes on rotation.

For values, `screen` will still use the bounds of UIScreen, and `window` will now use the bounds of the current rootViewController.view.

**Test Plan:**
To test you can watch the values of the `didUpdateDimensions` DeviceEvent. The following should be true:
Dimensions properly update on rotation.
Dimensions properly update in multitasking mode when app window resizes.
Dimensions only update once with the same result when doing the above actions.
Dimensions only update when the app is in the foreground.

Current:
simulator screen shot jul 23 2017 1 26 32 pm
simulator screen shot jul 23 2017 1 26 41 pm

Fixed:
simulator screen shot jul 23 2017 1 29 23 pm
simulator screen shot jul 23 2017 1 30 34 pm

### DIFF
--- a/React/Modules/RCTDeviceInfo.m
+++ b/React/Modules/RCTDeviceInfo.m
@@ -12,13 +12,10 @@
 #import "RCTAccessibilityManager.h"
 #import "RCTAssert.h"
 #import "RCTEventDispatcher.h"
+#import "RCTUIManager.h"
 #import "RCTUtils.h"
 
-@implementation RCTDeviceInfo {
-#if !TARGET_OS_TV
-  UIInterfaceOrientation _currentInterfaceOrientation;
-#endif
-}
+@implementation RCTDeviceInfo
 
 @synthesize bridge = _bridge;
 
@@ -34,17 +31,14 @@ RCT_EXPORT_MODULE()
   _bridge = bridge;
 
   [[NSNotificationCenter defaultCenter] addObserver:self
-                                           selector:@selector(didReceiveNewContentSizeMultiplier)
+                                           selector:@selector(didUpdateDimensions)
                                                name:RCTAccessibilityManagerDidUpdateMultiplierNotification
                                              object:_bridge.accessibilityManager];
-#if !TARGET_OS_TV
-  _currentInterfaceOrientation = [RCTSharedApplication() statusBarOrientation];
 
   [[NSNotificationCenter defaultCenter] addObserver:self
-                                           selector:@selector(interfaceOrientationDidChange)
-                                               name:UIApplicationDidChangeStatusBarOrientationNotification
-                                             object:nil];
-#endif
+                                           selector:@selector(didUpdateDimensions)
+                                               name:RCTUIManagerDidUpdateAvailableSizeNotification
+                                             object:_bridge.uiManager];
 }
 
 static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
@@ -53,15 +47,24 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
 
   // Don't use RCTScreenSize since it the interface orientation doesn't apply to it
   CGRect screenSize = [[UIScreen mainScreen] bounds];
-  NSDictionary *dims = @{
+  NSDictionary *screenDims = @{
                          @"width": @(screenSize.size.width),
                          @"height": @(screenSize.size.height),
                          @"scale": @(RCTScreenScale()),
                          @"fontScale": @(bridge.accessibilityManager.multiplier)
                          };
+
+  CGRect window = [[UIApplication sharedApplication] keyWindow].rootViewController.view.bounds;
+  NSDictionary *windowDims = @{
+                         @"width": @(window.size.width),
+                         @"height": @(window.size.height),
+                         @"scale": @(RCTScreenScale()),
+                         @"fontScale": @(bridge.accessibilityManager.multiplier)
+                         };
+
   return @{
-           @"window": dims,
-           @"screen": dims
+           @"window": windowDims,
+           @"screen": screenDims
            };
 }
 
@@ -80,7 +83,7 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
   return constants;
 }
 
-- (void)didReceiveNewContentSizeMultiplier
+- (void)didUpdateDimensions
 {
   RCTBridge *bridge = _bridge;
   RCTExecuteOnMainQueue(^{
@@ -92,38 +95,5 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
 #pragma clang diagnostic pop
   });
 }
-
-#if !TARGET_OS_TV
-
-- (void)interfaceOrientationDidChange
-{
-  __weak typeof(self) weakSelf = self;
-  RCTExecuteOnMainQueue(^{
-    [weakSelf _interfaceOrientationDidChange];
-  });
-}
-
-
-- (void)_interfaceOrientationDidChange
-{
-  UIInterfaceOrientation nextOrientation = [RCTSharedApplication() statusBarOrientation];
-
-  // Update when we go from portrait to landscape, or landscape to portrait
-  if ((UIInterfaceOrientationIsPortrait(_currentInterfaceOrientation) &&
-       !UIInterfaceOrientationIsPortrait(nextOrientation)) ||
-      (UIInterfaceOrientationIsLandscape(_currentInterfaceOrientation) &&
-       !UIInterfaceOrientationIsLandscape(nextOrientation))) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-        [_bridge.eventDispatcher sendDeviceEventWithName:@"didUpdateDimensions"
-                                                    body:RCTExportedDimensions(_bridge)];
-#pragma clang diagnostic pop
-      }
-
-  _currentInterfaceOrientation = nextOrientation;
-}
-
-#endif // TARGET_OS_TV
-
 
 @end
--- a/React/Modules/RCTDeviceInfo.m
+++ b/React/Modules/RCTDeviceInfo.m
@@ -54,7 +54,7 @@ static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
                          @"fontScale": @(bridge.accessibilityManager.multiplier)
                          };
 
-  CGRect window = [[UIApplication sharedApplication] keyWindow].rootViewController.view.bounds;
+  CGRect window = [[UIApplication sharedApplication] keyWindow].bounds;
   NSDictionary *windowDims = @{
                          @"width": @(window.size.width),
                          @"height": @(window.size.height),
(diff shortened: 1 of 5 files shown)


React/Modules/RCTDeviceInfo.m
```diff
@@ -53,15 +47,24 @@ - (void)setBridge:(RCTBridge *)bridge
...
   CGRect screenSize = [[UIScreen mainScreen] bounds];
-  NSDictionary *dims = @{
+  NSDictionary *screenDims = @{
                          @"width": @(screenSize.size.width),
                          @"height": @(screenSize.size.height),
                          @"scale": @(RCTScreenScale()),
                          @"fontScale": @(bridge.accessibilityManager.multiplier)
                          };
+
+  CGRect window = [[UIApplication sharedApplication] keyWindow].rootViewController.view.bounds;
```
shergin: Can we just use `[[UIApplication sharedApplication] keyWindow].bounds`?

Why do we need to subscribe for `RCTUIManagerDidUpdateAvailableSizeNotification`? Is your app hybrid (uses not only RN but also native views and view controllers)?

I hope, should be the better/easier way to subscribe for UIWindow size changing notification.

kipricker: `[[UIApplication sharedApplication] keyWindow].bounds` is better, I can change that.

Our app is pure RN. The two best ways I could find for getting alerted to a UIWindow size change are either using the UIContentContainer protocol method `viewWillTransitionToSize` or catch the change using UIViews `layoutSubviews`. Since there is no RN view controller class I opted to report the size change from `layoutSubviews`. So, `setAvailableSize` seemed like the most appropriate place to add the `RCTUIManagerDidUpdateAvailableSizeNotification` post to update us on a size change.

shergin:
@kipricker I agree that we have to improve/fix this, but I believe that introducing `RCTUIManagerDidUpdateAvailableSizeNotification` and relying on `RCTUIManger` is not a good approach. 😢 

https://developer.apple.com/library/content/documentation/WindowsViews/Conceptual/AdoptingMultitaskingOniPad/QuickStartForSlideOverAndSplitView.html

I briefly read this article and seems iOS posts `UIApplicationWillResignActive` notification right after resizing start... and I think then iOS should post `UIApplicationDidBecomeActive` when resizing ends. So, we can try to rely on `UIApplicationDidBecomeActive`.

kipricker:
@shergin One of the things I was trying to accomplish with the `RCTUIManagerDidUpdateAvailableSizeNotification` notifications was removing the need for the orientation event notifications, this also prevents potentially triggering the `didUpdateDimensions` event multiple times. Another alternative here could be introducing a `RCTRootViewController` which notifies `RCTDeviceInfo` of calls to `willTransitionToSize:`. I think this would be the most accurate solution and would again allow us to capture rotation and multitasking screen size changes with the same code path. Thoughts?

Otherwise, I'll take a look at the `UIApplicationWillResignActive` and `UIApplicationWillBecomeActive` tonight.

shergin:
@kipricker 
Yeah, but:
 * We don't want to introduce mandatory `ReactViewController` because it overcomplicates some use cases (and RN components do not utilize `ViewController` paradigm and infrastructure);
 * `willTransitionToSize:` should not be used to watch for `UIWindow` changes.

React/Modules/RCTDeviceInfo.m
```diff
@@ -53,15 +47,24 @@ - (void)setBridge:(RCTBridge *)bridge
 
   // Don't use RCTScreenSize since it the interface orientation doesn't apply to it
   CGRect screenSize = [[UIScreen mainScreen] bounds];
-  NSDictionary *dims = @{
+  NSDictionary *screenDims = @{
```
shergin: Also, please don't use abbreviation in code.

kipricker:
Ok, I had the same thoughts about the `RCTRootViewController` idea, which is why I took my original approach.

I looked into using `UIApplicationDidBecomeActive` and sadly the OS calls this before updating the window size. The most widely recommended method of detecting a screen resize is either using `willTransitionToTraitCollection` and `willTransitionToSize`, which are both controller protocols that we don't have access to without an `RCTRootViewController`.

So, I feel detecting a size change using `layoutSubviews` is still the best place for us to get the update. It also give us updates to the size change on multitasking and rotation, simplifying the final code. But, I have moved the notification to the `RCTRootView layoutSubviews`, which I believe is a more appropriate place for it than going through `RCTUIManager`.

kipricker:
@shergin Just wanted to see if you had any input on my last change.

shergin:
May be we can rely on `UIApplicationDidChangeStatusBarFrameNotification`?

kipricker:
`UIApplicationDidChangeStatusBarFrameNotification` is only called on rotation, and not when the statusbar frame actually change. Go figure.

As far as using a system notification, there aren't any that actually come at a reliable time to pick up the window size change. Otherwise, the 'correct' way to pick up window size changes is by using the `UIContentContainer` protocol, which is used by `UIViewController`. Even then, Apple discourages reacting to size changes this way and encourages using AutoLayout.

With the current architecture I still think monitoring the `RCTRootView` bounds is the most accurate location to pick up size changes as soon as they happen. Plus, the `RCTRootView` should be considered our window anyway. Also, this solution agnostically handles any window resizing and removes the need to actually react to the device rotation.

Though, one alternative that I think may be better than using `layoutSubviews` would be using a KVO on the `RCTRootView`s bounds. This would allow us to react at the exact time the window size has changed, no matter how it changed.

shergin:
> UIApplicationDidChangeStatusBarFrameNotification is only called on rotation, and not when the statusbar frame actually change.

Have you tried?

I clearly understand your point, but:
 * We cannot consider `RCTRootView` as a window because it is not an actual window, especially for hybrid apps. Using `window size` as `RootView` size in JS code is pretty weak assumption.
 * `RCTRootView` may and may not react on `UIWindow` size change. So, we cannot use it as an event emitter.
 * Using `layoutSubviews` for handling size changing is not correct, we have to override `setFrame:` method for this.
 * We don't have actual agent which can implement `UIContentController` because we (RN) can be placed far away from `UIWindow`.

Probably we can consider:
 * Deprecating this API because it useless now.
 * Inrtoducing new set of API which makes sense for RootView.
 * Using `KVO` to handle `UIWindow` instance size changing.

emusgrave:
Based on this PR, is it correct to assume that no one in the React Native community currently has an app with proper support for iPad Multitasking? (unless they've replicated this fix)

I didn't think that I was _that_ far on the bleeding edge until I tried multitasking and was unable to get the correct dimensions for my app. I've been looking into solutions and stumbled on this PR, and am unclear on what approach I should take to fix the problem.

shergin:
@emusgrave No, absolutely not. React Native is pretty agnostic to window/viewport size, so in most cases you don't need this module/event.
We (or I) will love to merge this PR as soon as it will be good enought. 
Do you already have particular problem with it?

BTW, multitasking on iPad is still quite bleading edge area in product development, unfortunately. 😢 
(This comment was written on iPad, iOS 11.)

emusgrave:
@shergin You mentioned above that you think the Dimensions API could be deprecated. If so, what is your recommended way to respond to device rotation changes and/or Split View changes? I need to adjust the layout of my app based on orientation (and in practice a "Split View" is really just an orientation change, viewing the Portrait mode of the app while the device is in landscape).

There are a number of places that I am using `Dimensions.get('window').width` to setup my components. For a multitasking split view, the width is always the full physical width, so my View is too stretched and is "covered" by the other side of the split window.

As another example, the Shoutem UI toolkit [http://shoutem.github.io/](url) uses `Dimensions` in a lot of their code to determine image size/scale.

I'll keep deep-diving on RCTRootView/etc and the above comments to see if I can come up with any other contributions to this PR. Keep in mind that I searched a while on React Native + iOS / iPad Multitasking and this thread is the only resource I found. 😔

shergin:
@emusgrave I assume that we just have to have a way to handle any particular <View> resize change to implement adoptive interface, right? So, you don't need to know the size of the whole viewport, you just need to know the size of the particular component... right?
So, for instance, on iOS `sizeClass`es make sense only for concrete view or view controller, not for whole app.

kipricker:
@emusgrave I have iPad multitasking working on my app. I had to remove all usage of the Dimensions lib, including third party libraries :( Then, to determine the correct width and height of the app I use a root `flex: 1` view in the app and use it's `onLayout` to determine the current width and height given to us by the renderer. Then you can serve that information to the rest of your app via Redux/Mobx/Relay.

I've been busy lately so I haven't had the time to tend to this PR.

@shergin I had already looked at `UIApplicationDidChangeStatusBarFrameNotification` and it will not work.

I'm a little confused about why we can't assume `RCTRootView` is our window. When setting up the app in the `AppDelegate` we make the `RCTRootView` the view of our key window. Even if the user decides to give RN a smaller space on the screen, `RCTRootView`'s frame will accurately represent the window that we have access to for drawing. Unless there is something I am missing.

The KVO of `RCTRootView`'s frame is still the best option I can think of given the other constraints.

shergin:
@emusgrave Oh, I had to clarify it previously: We cannot assume that `RCTRootView` has same size as `UIWindow` just because a lot of apps are hybrid and that assumption is wrong for them because they can have immediade view structures between `UIWindow` and `RCTRootView` instances. Moreover, an app can have several active `RCTRootView` instances at the same time!
Mostly because of this, I think that have screen/window-specific API is a bit inconseptual or has limited use cases.

So, we have to:
 * Have Component Specific APIs, like `onLayout`, if it is not enought;
 * (for this module) Subscribe to actual `UIWindow` instance, not to `RCTRootView`.

kipricker:
@shergin ah, that makes sense, I was unaware that was possible. Knowing that it's possible to have multiple `RCTRootView`s I would agree that `Dimensions` is not the correct solution. Let me come up with a test case for that and see if I can come up with a good proposal.

emusgrave:
@shergin @kipricker Thanks for both of your help! I did not realize that `onLayout` was giving the correct dimensions of the app when used on the root of my Views. I should be able to do everything I need to with that information, and am already in the progress of refactoring without using `Dimensions` anywhere in my code. Now that I "see the light", it is obvious that Dimensions is an anti-pattern.

kipricker:
@shergin I'm finally getting some time to explore this more. Would it be reasonable to assume that each `RCTRootView` is going to only have one JS context and each JS context is only going to have one mounted instance of `<AppContainer>`? I'm right now thinking that we could use the `onLayout` of the `<AppContainer>` to set and update the `window` key of `Dimensions`. I feel this would allow us to keep `Dimensions` and have it report meaningful information as the dimensions of each `RCTRootView` would be reported through its own JS context's `Dimensions` instance.

shergin:
@kipricker Usually many `RCTRootView`s share same bridge and JS context/VM.

PR #15162, opened 2017-07-24
iOS: Fix font weight resolution

kipricker:
**Issue:**
Some fonts are defined with weights that don't match with the UIFontWeight constants.

**Example:**
UIFontWeightTraits for Roboto font
Light: -0.230
Thin: -0.365

Currently, the UIFontWeightTrait is always used if it != 0.0, and given the UIFontWeight constants for Light and Thin:
UIFontWeightThin -0.6
UIFontWeightLight -0.4

A style font weight of "300" or "200" will both resolve to Roboto-Thin as its weight -0.365 is closer to -0.4 (UIFontWeightLight) and -0.6 (UIFontWeightThin) than -0.230 (Roboto-Light).

**Proposed fix:**
When resolving `getWeightOfFont` try to match the name of weight to the name of the font first, and guess the font with UIFontWeightTrait as the fall back.

**Test Plan:**
Attempt to display Roboto at weights "200" and "300" and Roboto-Thin and Roboto-Light should be displayed correctly.

Current:
simulator screen shot jul 7 2017 11 44 42 am

Fixed:
simulator screen shot jul 7 2017 11 42 25 am

### DIFF
--- a/React/Views/RCTFont.mm
+++ b/React/Views/RCTFont.mm
@@ -54,15 +54,14 @@ static RCTFontWeight weightOfFont(UIFont *font)
     };
   });
 
-  NSDictionary *traits = [font.fontDescriptor objectForKey:UIFontDescriptorTraitsAttribute];
-  RCTFontWeight weight = [traits[UIFontWeightTrait] doubleValue];
-  if (weight == 0.0) {
-    for (NSString *name in nameToWeight) {
-      if ([font.fontName.lowercaseString hasSuffix:name]) {
-        return [nameToWeight[name] doubleValue];
-      }
+  for (NSString *name in nameToWeight) {
+    if ([font.fontName.lowercaseString hasSuffix:name]) {
+      return [nameToWeight[name] doubleValue];
     }
   }
+
+  NSDictionary *traits = [font.fontDescriptor objectForKey:UIFontDescriptorTraitsAttribute];
+  RCTFontWeight weight = [traits[UIFontWeightTrait] doubleValue];
   return weight;
 }
 


nihgwu:
@javache @grabbou  that's a breaking change for my app, I think we should add it to the breaking changes section for the release note

grabbou:
Makes sense - I've been struggling with this issue for a while. Glad to see it's fixed!

nihgwu:
Seems there is a regression with `PingFangSC`, I will paste the diff later

nihgwu:
After: 
simulator screen shot 3 aug 2017 00 58 19
Before:
simulator screen shot 3 aug 2017 00 58 50
Native:
image

It's obviously that the `PingFangSC-Light` is incorrect after this commit
ping @kipricker

kipricker:
@nihgwu good catch. Thats actually an issue with the original name suffix check. It looks like it's finding ultra**light** before **light**.

I'll add a check for a weight prefix, ultrabold and semibold would be affected the same. Actually, the suffix check wont work with italic fonts either. Even with this fix, font selection is still guess work, which I don't like. I'll investigate.

nihgwu:
@kipricker good to know, I know nothing about Objective-C, if you come with the patch, can we also cherry-pick to `0.48-stable` branch @grabbou

grabbou:
Yeah, I'll cherry-pick it as soon as there's a reasonable fix. Let's take
time and try to come up with a working algorithm.

nihgwu:
@kipricker any progress on this?

kipricker:
@nihgwu I've been super busy and haven't had a lot of time. I am still investigating how to accurately detect font weights beyond just string comparison. I'm going to make the time to have a fix for this by next week.

PR #15177, opened 2017-07-24
Fix description of next tutorial on "Handling Text Input" page

stanparker:
The last sentence says that the next tutorial is the ScrollView, when the Next button actually points to "Handling Touches. (ScrollView is after Handling Touches).  Updated the sentence and link to go to Handling Touches instead.


hramos:
Hey @stanparker thanks for the fix! Looks like we just merged this in through https://github.com/facebook/react-native/pull/15176

PR #15181, opened 2017-07-24
Separate window dimensions change from orientation

atsjenk:
**Summary:**
There was a bug with RN.Dimensions returning incorrect window dimensions. In certain cases when device was in portrait, window dimensions reported landscape dimensions and vice versa.

This happened because in certain scenarios, after device orientation changed, dimensions update event from ReactRootView had incorrect dimensions.
Was able to reproduce this when device was rotated during app launch. After rotation global layout listener callback gets invoked. Inside the callback current and previous orientations are compared. When a change is detected, orientation and dimension change events are sent to JS. It is assumed, when orientation changes, new dimensions are available immediately. This is not the case for window dimensions as they are retrieved from resources object which gets updated asynchronously after orientation change. In cases when app is doing a lot of work on the main thread, like app startup, it takes more time to update the resources object. And when orientation change is detected in global layout, resources object is not updated with new dimensions yet. This causes dimensions update to be sent to JS with old window dimensions.
Global layout listener callback does get invoked a second time when resources object is finally updated with new dimensions, but since orientation no longer changes, no event is sent to JS.

Fixed this by separating dimensions update from orientation update. Now RN keeps track of previous window and screen dimension values. When a change is detected, an event is sent to JS with updated dimensions. This ensures that whenever dimensions change, JS gets the updated values.

This has a side effect of sending dimension update twice in some cases.
One example is the case above where window dimensions take time to update, but screen dimensions are updated immediately. This will cause two events to be sent to JS. One for window dimensions and one for screen dimensions update.
Other change is that initial value for both window and screen fields is empty. Which results in first change to trigger an event. Previously initial orientation value was 0 which meant when app started in normal portrait orientation, first layout did not trigger a dimension update event. Now even first layout sends the event. This should not be an issue as it is to make sure dimensions in JS side are correct.

**Testing:**
Verified with a sample app that correct dimensions are available when app launches.
Verified that after orientation dimensions are updated.
Verified that in the scenario described above where window dimensions are updated later, we get correct dimension values in JS.
We have incorporated this fix into our app and have been testing it internally.

Ats Jenk
Microsoft Corp.


ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -404,13 +407,37 @@ private void checkForDeviceOrientationChanges() {
...
-      emitUpdateDimensionsEvent();
       emitOrientationChanged(rotation);
     }
 
+    private void checkForDeviceDimensionsChanges() {
+      // Get current display metrics.
+      DisplayMetricsHolder.initDisplayMetrics(getContext());
+      // Check changes to both window and screen display metrics since they may not update at the same time.
+      if (!areMetricsEqual(mWindowMetrics, DisplayMetricsHolder.getWindowDisplayMetrics())
+          || !areMetricsEqual(mScreenMetrics, DisplayMetricsHolder.getScreenDisplayMetrics())) {
```
shergin: Please make the code style consistent with existing code.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -404,13 +408,37 @@ private void checkForDeviceOrientationChanges() {
...
+    }
+
+    private boolean areMetricsEqual(DisplayMetrics displayMetrics, DisplayMetrics otherMetrics) {
+      if (Build.VERSION.SDK_INT >= 17) {
+        return displayMetrics.equals(otherMetrics);
+      } else {
+        // DisplayMetrics didn't have an equals method before API 17.
+        // Check all public fields manually.
+        return displayMetrics.widthPixels == otherMetrics.widthPixels
+          && displayMetrics.heightPixels == otherMetrics.heightPixels
```
shergin: I meant we have to put `&&` on previous line (seems that is current style). (Sorry.)

atsjenk: I see. Yes that seems to be the pattern. I'll fix that.

Indentation for the if condition did not match the current style either as per line [147](https://github.com/atsjenk/react-native/blob/e99afe23e45d8140dc0c827f833a862d3ae270f7/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java#L147).

shergin:
@atsjenk Great! Thank you!
I will land it on Monday.

atsjenk:
@shergin any updates on this PR?

PR #15184, opened 2017-07-25
Add more information to `__fbBatchedBridge is undefined` error

terribleben:
**Motivation:**  This error can be a symptom of various other issues (see: an issue search for `__fbBatchedBridge is undefined`). I'm hoping to provide slightly more information about what might be going wrong and how to self-help.

**Test Plan:** Run some JS before the bridge has injected itself into the JS context. (sort of copping out here since the change is just to an error string literal.)

### DIFF
--- a/React/Executors/RCTJSCExecutor.mm
+++ b/React/Executors/RCTJSCExecutor.mm
@@ -596,7 +596,11 @@ RCT_EXPORT_METHOD(setContextName:(nonnull NSString *)contextName)
       }
     } else {
       if (!errorJSRef && JSC_JSValueGetType(ctx, batchedBridgeRef) == kJSTypeUndefined) {
-        error = RCTErrorWithMessage(@"Unable to execute JS call: __fbBatchedBridge is undefined");
+        error = RCTErrorWithMessage(@"Unable to execute JS call: __fbBatchedBridge is undefined. This can happen "
+                                    "if you try to execute JS and the bridge has not set up, for example if it encountered "
+                                    "an incomplete bundle or a fatal script execution error during startup. Verify that a "
+                                    "valid JS bundle is included with your app and that it loaded correctly, or try "
+                                    "reinstalling the app.");
       }
     }
 


brianmmorton:
Hey fyi, I happened to see this issue and PR, and im having this exact issue. Looked at the error message and unforunately still couldn't fix my issue! What is the point of reinstalling "this app"? Is that my app or are you referring to react-native? Still a bit confused :/

javache:
>  This can happen if you try to execute JS before the bridge has finished setting up.

Under what scenarios is this possible? We should try and catch those in the infrastructure instead.

The main cause of this error I see is when there are fatal script execution errors during startup, or you're loading an incomplete bundle, such that the BatchedBrdge module is never executed.

terribleben:
Sure thing! I was encountering the scenario I mentioned, but I defer to your judgment about what would be the most common cause here. I updated the message to reflect what you said, does it look any better?

javache:
Unable to execute JS call: __fbBatchedBridge is undefined. This can happen "
 +                                    "if you try to execute JS and the bridge has not set up, for example if it encountered "
 +                                    "an incomplete bundle or a fatal script execution error during startup. You might try "
 +                                    "double checking that the bundle loaded correctly, or resetting the state of React Native "
 +                                    "devtools on your device (e.g. by uninstalling and reinstalling your app from the device "
 +                                    "home screen).

Can you remove the mention of React Native devtools? (We don't want any confusion with react devtools).

Maybe replace the last sentence with "Verify that a valid JS bundle is included with your app and that it loaded correctly, or try reinstalling the app."

terribleben:
Sounds good to me - done

mhorowitz:
This code is not long for this world.  Once the bugs in the CocoaPods C++ build are shaked out, I'll be deleting the old bridge of which this is a part.  It might be worth making the same to JSCExecutor around https://github.com/facebook/react-native/blob/f9808f07c8abbfcb7a18bc3267d7733de21ba735/ReactCommon/cxxreact/JSCExecutor.cpp#L372-L374

MrHubble:
How do we verify that a valid JS bundle is included with our app and that it loaded correctly? Is this the information needed?

screen shot 2017-08-25 at 2 44 27 pm

PR #15192, opened 2017-07-25
Remove hasPressHandler function and it's uses because it's not used…

shadowm2:
…  used in other touchable components also it's not necessary.

By default TouchableHighlight component needs to have a press handler funciton to show and hide underlay color which is not appropriate.

Test Plan : create a TouchableHighlight component and it won't show underlay on press if no press handler function is assigned

### DIFF
--- a/Libraries/Components/Touchable/TouchableHighlight.js
+++ b/Libraries/Components/Touchable/TouchableHighlight.js
@@ -221,7 +221,7 @@ var TouchableHighlight = createReactClass({
   },
 
   _showUnderlay: function() {
-    if (!this._isMounted || !this._hasPressHandler()) {
+    if (!this._isMounted) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


henrikra:
Btw when you are using `TouchableHighlight` vs `TouchableOpacity`?

shadowm2:
@henrikra 
I can't think of a reason why you asked this. please explain more about your question.

henrikra:
I mean I have only used `TouchableOpacity` since I haven't found any use cases for `TouchableHighlight` yet :D So when in your app you use `TouchableHighlight`?

shadowm2:
@henrikra 
I was going through the official tutorial and i faced TouchableHighlight and i was playing around with it when i noticed the difference between these Touchable components. Also the tutorial says :

> Generally, you can use TouchableHighlight anywhere you would use a button or link on web. The view's background will be darkened when the user presses down on the button

henrikra:
@shadowm2 Okay so basically it just darkens the views background lol

shadowm2:
@henrikra 
This conversation is unnecessary and off-topic.

PR #15203, opened 2017-07-25
Fix updating a view z-index on Android

janicduplessis:
If the z-index was updated after the initial mount, changes would not be reflected because we did not recalculate the z-index mapped child views and redraw the view. This adds code to do that and call it whenever we update z-index.

**Test plan**
Tested by reproducing the bug with 2 overlapping views that change z-index every second. Made sure it now works properly and z-index changes are reflected.


Noitidart:
Thank you sir! I will test this tonight! :)

Noitidart:
I never tested something like this, but I tried my hand. I saw the 4 files you changed, and made those files changes in my node_modules folder. Then I re-ran `react-native run-android`, however no change :(

It's a very simple repo, may you please try with this commit - https://github.com/Noitidart/Qaida-Board/tree/3ec2502c3682073fc43c8738a7849d09e867674f/App

janicduplessis:
Testing on Android is a bit more complex because you need to build the project from source, see http://facebook.github.io/react-native/docs/android-building-from-source.html.

janicduplessis:
I can test with the project you linked tomorrow if you don't get it working, I know building from source can be a pain :)

Noitidart:
Oh thanks, I'll try this now and let you know how it goes. :)

paranoidjk:
+1 for this

Noitidart:
My sincerest apologies sir, @janicduplessis - I was not able to figure this testing process out. I'm feeling like a fool right now, it seems the instructions are very good, I just am screwing up somewhere. :(

Noitidart:
I did take a screencast of it being bad though the other day. On Android it works like this for me:

https://youtu.be/2MNBLw382-0

But it should work like this:

Recording of iOS - https://youtu.be/3cwRMAnB45A
Recording of this on Windows (UWP) - https://youtu.be/0x-DUFVk5LA

ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactZIndexedViewGroup.java
```diff
@@ -12,4 +12,10 @@
    * @return The child view index considering z-index
    */
   int getZIndexMappedChildIndex(int index);
+
+  /**
+   * Redraw the view based on updated z-index. This should be called after updating one of it's child
+   * z-index.
+   */
+  void updateZIndex();
```
piotrek1543: I think that name of this method is a bit too short. Maybe ` updateViewParentZIndex()` would be better as I suppose this method belongs to parent not child.

janicduplessis: Yea I had a hard time figuring out a name. `updateViewParentZIndex` would not work in this case because it doesn't update the parent it redraws the view based on the updated z-index of its children.

hramos:
Passing on to Andrew Chen for review.

Noitidart:
@janicduplessis were you able to test with the demo commit I gave? I am very eager to hear if that fixed it please sir. I don't think there is an error on my side because it worked as expected on iOS and even Windows (UWP).

rtman:
@janicduplessis curious how this is going? Doesn't seem resolved yet?

hramos:
@rtman this PR has been merged and the fix should be on master. Please confirm that you're not seeing this resolved on master.

Noitidart:
Thanks @hramos is there some steps for novices to follow? I followed the previous guide and failed miserably.

ide:
I cherry picked this into 0.48 as well. When 0.48-rc.2 gets published (no ETA though) it will include this change.

rtman:
Sorry what version of react-native is this included in?

rtman:
@hramos Oh i understand now, this is in the current master right? I will try this sometime soon and see if it works.

@ide  0.48-rc.2 is the earliest version that contains this fix?

ide:
@rtman, this is not in 0.48-rc.1 but if rc.2 is published it will be there.

rtman:
@ide ok thanks!

Noitidart:
Just tested out0.49 with my zIndex example above and it works, yahoo! Thanks!

rtman:
Yea it works for me on 0.49.3!

_Sent from my Google Nexus 5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Leadrive:
in 0.53, this issue still, and I read the react native code, It already modify to:
https://github.com/facebook/react-native/commit/28c7535a75ceb4edf06a0ecfd8130c448fae9627

PR #15206, opened 2017-07-26
[Docs] Clarify use of images from the local filesystem in the ScrollView tutorial

dbnex14:
The example will not work as is.  Who ever tries this needs to add img directory to AwesomeProject and favicon.png icon in it.  This is simple issue but for absolute beginer is no go.


hramos:
Slight edit. Thanks for the update! We should probably introduce Image properly. It's used earlier in the tutorial, but it's a remote image.

hramos:
We can merge this once the CLA is signed.

PR #15212, opened 2017-07-26
Fix tvOS build issues with UIDeviceOrientation

javache:


javache:
cc @dlowder-salesforce

PR #15220, opened 2017-07-27
Apple TV: RCTTabBar selection controlled by native after render (fix #15081)

### LINKED ISSUE #15081: TabBarIOS (Tabs) in AppleTV flickering when switching between tabs

dlowder-salesforce:
**Motivation**

Fix flickering in TabBarIOS on Apple TV... issue #15081

After this change, on Apple TV, TabBarIOS item selections will be controlled purely from the native side after initial render with the `selected` prop.  This is necessary because the `UITabBar` implementation in tvOS moves the selection before calling `shouldSelectViewController:`; this issue does not occur on iOS.

**Test plan**

Existing CI should still pass. Issue is resolved when testing the example code in #15081 .


dlowder-salesforce:
@javache please let me know what you think of this solution for the Apple TV `TabBarIOS` issues...

dlowder-salesforce:
@shergin please also review this PR when you get a chance

javache:
Ok, let's do this as stop-gap solution, we should see if we could fix this correctly in the future.

rada:
@dlowder-salesforce You wrote "TabBarIOS item selections will be controlled purely from the native side after initial render with the selected prop." Does this mean, that we cannot switch tabs programmatically from JS after initial render?

dlowder-salesforce:
That is correct. Unfortunately, given the differences in implementation of UITabBar on AppleTV versus iOS, this was the only way to solve the flickering and rendering issues we were seeing.

PR #15221, opened 2017-07-27
Apple TV: Enable long presses on TV remote; dev menu on TV device; example code

dlowder-salesforce:
**Motivation**

Properly support long presses on the Apple TV remote, and also enable dev menu functionality on a real Apple TV device (shaking an Apple TV doesn't work 😄 )

**Test plan**

New example added to `RNTester`.


React/Base/RCTTVRemoteHandler.m
```diff
@@ -100,9 +110,16 @@ - (void)selectPressed:(UIGestureRecognizer *)r
   [self sendAppleTVEvent:@"select" toView:r.view];
 }
 
-- (void)longPress:(UIGestureRecognizer *)r
+- (void)longPlayPausePressed:(UIGestureRecognizer *)r
+{
+  [self sendAppleTVEvent:@"longPlayPause" toView:r.view];
+  // If shake to show is enabled on device, use long play/pause event to show dev menu
+  [[NSNotificationCenter defaultCenter] postNotificationName:@"RCTShowDevMenuNotification" object:nil];
```
javache: Instead of using a string here, can you can change the definition of RCTShowDevMenuNotification to extern (put an RCT_EXTERN definition in the header)

dlowder-salesforce: Done

PR #15230, opened 2017-07-27
Navigator is deprecated

0x1ad2:

### DIFF
--- a/Libraries/Components/StatusBar/StatusBar.js
+++ b/Libraries/Components/StatusBar/StatusBar.js
@@ -137,6 +137,11 @@ function createStackEntry(props: any): any {
  *  </View>
  * ```
  *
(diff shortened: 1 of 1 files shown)


Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -137,6 +137,11 @@ function createStackEntry(props: any): any {
  *  </View>
  * ```
  *
+ * ## Navigator is deprecated and has been removed from this package.
```
chirag04: at this point, i'm wondering if it's better to just update the example to be navigator agnostic.

@hramos

hramos: I agree. Lets remove it altogether.

PR #15231, opened 2017-07-27
Update releases script

grabbou:
When releasing old stable versions, e.g. `0.43` for Microsoft, I found it to update the `latest` tag because we are technically on the stable branch.

This PR modifies this logic so that latest is not set if we are on an older than latest branch.

I will test this code when releasing new version of React Native in few days.


hramos:
Looks good. Do you want to hold off until you've used this for the next release before merging?

grabbou:
Yeah, sounds good! I'll experiment few times and report back :)

PR #15234, opened 2017-07-27
Fix SectionList examples in documentation

timwangdev:
Document to change http://facebook.github.io/react-native/releases/next/docs/sectionlist.html

This example is not documented incorrectly. 

According to Flow-type in the [source code](https://github.com/facebook/react-native/blob/master/Libraries/Lists/SectionList.js#L25-L52) the `title` is not required and the `renderItem` method takes wrong `item.title` for rendering in example. See https://github.com/facebook/react-native/pull/15234#issuecomment-318555336

## Test Plan (Screenshot)
screen shot 2017-07-28 at 12 24 03 pm

### DIFF
--- a/Libraries/Lists/SectionList.js
+++ b/Libraries/Lists/SectionList.js
@@ -220,17 +220,17 @@ type DefaultProps = typeof defaultProps;
  *       renderItem={({item}) => <ListItem title={item.title} />}
  *       renderSectionHeader={({section}) => <H1 title={section.title} />}
  *       sections={[ // homogenous rendering between sections
- *         {data: [...], title: ...},
- *         {data: [...], title: ...},
- *         {data: [...], title: ...},
+ *         {data: [...]},
+ *         {data: [...]},
+ *         {data: [...]},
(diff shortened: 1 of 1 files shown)


timwangdev:
Added Test Plan

hramos:
It looks like this example does expect a title to be there (see renderItem).

timwangdev:
@hramos Ok, I see. The `title` is a custom property in section here. The `renderItem` is still not correct because the `item` represents structure in `data` array.

I tested with the code:
```javascript
<SectionList
    renderItem={({item}) => <View><Text>{item}</Text></View>}
    renderSectionHeader={({section}) => <View><Text>{section.title}</Text></View>}
    sections={[
        {data: [1, 2, 3], title: 'Section 1'},
        {data: [1, 2, 3], title: 'Section 2'},
        {data: [1, 2, 3], title: 'Section 3'},
    ]}
/>
```
And this are rendered as expected. Will update the PR commit.

timwangdev:
Updated changes

PR #15237, opened 2017-07-27
Fix mistake acquireWakeLockNow in headless js section

ivanzotov:
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/HeadlessJsTaskService.java#L72


chirag04:
@facebook-github-bot shipit

PR #15240, opened 2017-07-27
Fix misspelling UIKit

ajacquelin:


christopherdro:
@facebook-github-bot shipit

PR #15251, opened 2017-07-28
[IntegrationWithExistingApps.md] Rename HelloWorld to MyReactNativeApp

rajsuvariya:
The name should be "MyReactNativeApp" otherwise it gives error that the HelloWorld Application is not found


rajsuvariya:
@hramos Why these checks failed? This is just documentation change.

PR #15258, opened 2017-07-28
Account for header, footer, and content container padding in VirtualizedList

jamesreggio:
### Overview

`VirtualizedList` leans heavily upon frame estimates to power its heuristics. Despite explicit support for headers and footers (and implicit support for top and bottom padding via `contentContainerStyle`), the estimates fail to account for the heights of these additions to the top and bottom of the scrollable content area.

This commit addresses some of the inconsistencies, though I'm certain that I missed some calculations. Notably, it fails to address horizontal `VirtualizedLists`, and it doesn't address any of the sticky header logic, which may or may not be wrong (it was too difficult to tell).

### Test plan

Like most things with RN, these behaviors are difficult to capture with automated tests. However, I put together a repro that demonstrates the original calculation issues and can prove the fix. Code for the repro is [here](https://gist.github.com/jamesreggio/47904fb20b3a142172ef701113abbc4e).

The repro fills a `FlatList` with 100 items, each of height `100px`. It also adds `150px` of `paddingTop` and `paddingBottom` to the `contentContainerStyle` for the list. In the videos below, you'll see the list load. After a second, I invoke `scrollToEnd()`, and a second after that, I invoke `scrollToIndex({index: 50, viewPosition: 0})`.

Note that in the **before** video, `scrollToEnd` scrolls to the edge of the last item, but not to the end of the content area — it omits the dark gray `paddingBottom`. Likewise, its heuristic miscalculates the offset of the item at index 50 because it fails to account for the `paddingTop`. The **after** video shows the correct behavior after applying this patch.

#### Before

repro-before

#### After

repro-after

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -213,7 +213,8 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
       0,
       frame.offset +
         frame.length +
-        this._footerLength -
+        this._footerLength +
+        this._paddingBottom -
         this._scrollMetrics.visibleLength,
     );
     this._scrollRef.scrollTo(
@@ -503,8 +504,13 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
   }
 
   render() {
+    const flatStyles = flattenStyle(this.props.contentContainerStyle);
+    if (flatStyles) {
+      this._paddingTop = flatStyles.paddingTop || 0;
+      this._paddingBottom = flatStyles.paddingBottom || 0;
+    }
+
     if (__DEV__) {
-      const flatStyles = flattenStyle(this.props.contentContainerStyle);
       warning(
         flatStyles == null || flatStyles.flexWrap !== 'wrap',
         '`flexWrap: `wrap`` is not supported with the `VirtualizedList` components.' +
@@ -703,10 +709,12 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
   _hasWarned = {};
   _highestMeasuredFrameIndex = 0;
   _headerLength = 0;
+  _paddingTop = 0;
   _initialScrollIndexTimeout = 0;
   _fillRateHelper: FillRateHelper;
   _frames = {};
   _footerLength = 0;
+  _paddingBottom = 0;
   _scrollMetrics = {
     contentLength: 0,
     dOffset: 0,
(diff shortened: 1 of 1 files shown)


jamesreggio:
Maintainers: `prettier` is stuck in a loop. It advises deleting the newline on 992, and if you comply and run again, it recommends reinserting it.

Please advise.

jamesreggio:
Ping, maintainers?

ssssssssssss:
I guess, should be @sahrens for review.

jamesreggio:
@sahrens @ssssssssssss — any reason why this PR should not be merged?

Libraries/Lists/VirtualizedList.js
```diff
@@ -977,9 +985,14 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
     const itemCount = this.props.getItemCount(this.props.data);
     let hiPri = false;
     if (first > 0 || last < itemCount - 1) {
-      const distTop = offset - this._getFrameMetricsApprox(first).offset;
+      const distTop =
+        offset +
+        this._headerLength +
+        this._paddingTop -
```
sahrens: Why do you need to account for headerLength and paddingTop here?

Libraries/Lists/VirtualizedList.js
```diff
@@ -977,9 +985,14 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
...
+        this._getFrameMetricsApprox(first).offset;
```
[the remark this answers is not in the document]
sahrens: Please run prettier as well.

Libraries/Lists/VirtualizedList.js
```diff
@@ -503,8 +504,13 @@ class VirtualizedList extends React.PureComponent<OptionalProps, Props, State> {
   }
 
   render() {
+    const flatStyles = flattenStyle(this.props.contentContainerStyle);
+    if (flatStyles) {
+      this._paddingTop = flatStyles.paddingTop || 0;
+      this._paddingBottom = flatStyles.paddingBottom || 0;
```
sahrens: Don't mutate in render - componentDidReceiveProps would be a better place for setting these.

sahrens:
Hey sorry, I totally missed this. There were some pretty substantial changes to VirtualizedList to support nested lists - do you mind rebasing over that?

hramos:
Closing an old PR - it's been in "changes requested" for over a year.

PR #15263, opened 2017-07-28
Add altitudeAccuracy attribute on Position object (Android)

dabit1:
As the standard says, the Position object must have "altitudeAccuracy" property.

https://developer.mozilla.org/en-US/docs/Web/API/Coordinates


dabit1:
Excuse me... The method getVerticalAccuracyMeters() is only available on android 8. Let's leave this for the future...

Issue #14925, opened 2017-07-10
[iOS] Undefined symbols for architecture x86_64


vongohren:
### Is this a bug report?
Yes

### Have you read the Bugs section of the Contributing to React Native Guide?

Yes

### Environment

1. `react-native -v`: react-native-cli: 2.0.1 - react-native: 0.46.1
2. `node -v`: v7.5.0
3. `npm -v`: 4.1.2

Then, specify:

1. Target Platform (e.g. iOS, Android): iOS
2. Development Operating System (e.g. macOS Sierra, Windows 10): macOS Sierra 10.12.5

### Steps to Reproduce

1. Have a new project based on 0.46.1 react native.
2. Use "babel-preset-react-native": "^2.0.0"
3. I don't know if there is a library or what that does this but its really weird

### Expected Behavior
I expect the react-native run-ios to build the app properly! 
Because react-native run-android works fine and lets me build and run the android app!

### Actual Behavior
```
 /Users/snorreedwin/Code/entur/entur-clients/native/ios/build/Build/Products/Debug-iphonesimulator/React/React.framework/React

Undefined symbols for architecture x86_64:
  "_JSNoBytecodeFileFormatVersion", referenced from:
      +[RCTJavaScriptLoader loadBundleAtURL:onProgress:onComplete:] in RCTJavaScriptLoader.o
      +[RCTJavaScriptLoader attemptSynchronousLoadOfBundleAtURL:runtimeBCVersion:sourceLength:error:] in RCTJavaScriptLoader.o
  "facebook::react::parseTypeFromHeader(facebook::react::BundleHeader const&)", referenced from:
      +[RCTJavaScriptLoader attemptSynchronousLoadOfBundleAtURL:runtimeBCVersion:sourceLength:error:] in RCTJavaScriptLoader.o
  "facebook::react::customJSCWrapper()", referenced from:
      -[RCTDevSettings isJSCSamplingProfilerAvailable] in RCTDevSettings.o
      -[RCTDevSettings toggleJSCSamplingProfiler] in RCTDevSettings.o
      _RCTNSErrorFromJSErrorRef in RCTJSCErrorHandling.o
      -[RCTSamplingProfilerPackagerMethod handleRequest:withResponder:] in RCTSamplingProfilerPackagerMethod.o
  "facebook::react::systemJSCWrapper()", referenced from:
      -[RCTDevSettings isJSCSamplingProfilerAvailable] in RCTDevSettings.o
[...]

PR #15264, opened 2017-07-28
Add missing subspec part

GantMan:
Though missing `BatchedBridge` works in dev, the archive will fail with a linking error that is very difficult to understand/research.  Fortunately @javache identifies this in https://github.com/facebook/react-native/issues/14925 and this seems to be working for many of us.


javache:
BatchedBridge is the legacy iOS implementation. Please update the instruction to use CxxBridge instead.

PR #15276, opened 2017-07-30
RCTProfile: Use C atomics instead of OSAtomic

Adlai-Holler:
This will save us precious microseconds and it's prettier to look at.

Test plan: Try the profiler, rely on CI.

### DIFF
--- a/React/Profiler/RCTProfile.m
+++ b/React/Profiler/RCTProfile.m
@@ -10,7 +10,7 @@
 #import "RCTProfile.h"
 
 #import <dlfcn.h>
-#import <libkern/OSAtomic.h>
+#import <stdatomic.h>
 #import <mach/mach.h>
 #import <objc/message.h>
 #import <objc/runtime.h>
@@ -42,9 +42,7 @@ static NSString *const kProfilePrefix = @"rct_profile_";
 
 #pragma mark - Variables
 
-// This is actually a BOOL - but has to be compatible with OSAtomic
-static volatile uint32_t RCTProfileProfiling;
-
+static atomic_bool RCTProfileProfiling;
 static NSDictionary *RCTProfileInfo;
 static NSMutableDictionary *RCTProfileOngoingEvents;
 static NSTimeInterval RCTProfileStartTime;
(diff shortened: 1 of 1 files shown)


shergin:
cc @javache

javache:
Can you share some numbers from profiling this? Or is this based on intuition?

React/Profiler/RCTProfile.m
```diff
@@ -42,8 +42,7 @@
 
 #pragma mark - Variables
 
-// This is actually a BOOL - but has to be compatible with OSAtomic
-static volatile uint32_t RCTProfileProfiling;
+static atomic_bool RCTProfileProfiling = ATOMIC_VAR_INIT(NO);
```
javache: Can you mark this as `constexpr` so we can verify at compile-time this doesn't add a global initalizer?

Adlai-Holler: Hmm `constexpr` isn't available except in C++. We can make this `RCTProfile.mm` but that's a bigger change and we'll lose a tiny bit of efficiency (due to ARC exception safety). What do?

Adlai-Holler:
@javache It's just intuition since `OSAtomic*` are function calls and `atomic_*` are inlined down to one or two instructions. I doubt the performance difference will be measurable for our current uses. But the main benefit here IMO is that `OSAtomic` is deprecated ( https://github.com/facebook/react-native/pull/15278#issuecomment-318938721 ) plus I just love language features.

React/Profiler/RCTProfile.m
javache: Our internal setup is to have `-fno-exceptions` for most objective-c code and opt-in to the codepaths that need it. Maybe we can do that here too?

Adlai-Holler: Hmm turns out we can't use `constexpr` because it implies `const` and we need to mutate this value.

Adlai-Holler:
I don't think `constexpr` is doable, but I've rebased the diff for good measure

React/Profiler/RCTProfile.m
javache: Hmm ok. I think we compile by default with -Wglobal-constructors, so that should catch it too.

PR #15278, opened 2017-07-30
RCTImage: Use C atomics instead of OSAtomic

Adlai-Holler:
The next in my series of :atom: migrations.

### DIFF
--- a/Libraries/Image/RCTImageLoader.m
+++ b/Libraries/Image/RCTImageLoader.m
@@ -7,7 +7,7 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import <libkern/OSAtomic.h>
+#import <stdatomic.h>
 #import <objc/runtime.h>
 
 #import <ImageIO/ImageIO.h>
@@ -324,7 +324,8 @@ static UIImage *RCTResizeImageIfNeeded(UIImage *image,
     BOOL requiresScheduling = [loadHandler respondsToSelector:@selector(requiresScheduling)] ?
     [loadHandler requiresScheduling] : YES;
 
-    __block volatile uint32_t cancelled = 0;
+    __block atomic_bool cancelled = ATOMIC_VAR_INIT(NO);
+    // TODO: Protect this variable shared between threads.
     __block dispatch_block_t cancelLoad = nil;
     void (^completionHandler)(NSError *, id, NSString *) = ^(NSError *error, id imageOrData, NSString *fetchDate) {
         cancelLoad = nil;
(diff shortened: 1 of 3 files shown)


Libraries/Image/RCTImageLoader.m
```diff
@@ -524,20 +528,25 @@ - (RCTImageLoaderCancellationBlock)loadImageWithURLRequest:(NSURLRequest *)image
...
+    // TODO: Protect this variable shared between threads.
     __block dispatch_block_t cancelLoad = nil;
     dispatch_block_t cancellationBlock = ^{
+        BOOL alreadyCancelled = atomic_fetch_or(&cancelled, 1);
+        if (alreadyCancelled) {
+            return;
+        }
         dispatch_block_t cancelLoadLocal = cancelLoad;
-        if (cancelLoadLocal && !cancelled) {
+        cancelLoad = nil;
```
Adlai-Holler: I added this line because it seems to have been missing from previous implementation. The pattern of "read cancel block into local variable, erase the __block variable, run the local variable" was copied from line 400 but it was missing this line, which defeats the purpose.

shergin:
@Adlai-Holler That looks great! But... Could you please clarify the motivation? How `stdatomic` is better than `osatomic`? Which problem do we solve?

Adlai-Holler:
Sure! We get a few benefits:

- OSAtomic is deprecated, so we get off from that: https://opensource.apple.com/source/libplatform/libplatform-125/include/libkern/OSAtomicDeprecated.h.auto.html https://github.com/google/protobuf/issues/2182 https://forums.developer.apple.com/thread/62765
- We remove function calls – stdatomic is fully inlined and ends up as one or two instructions.
- The API is more purpose-built and flexible: `atomic_bool` vs `volatile int32_t`.

It doesn't exactly shatter the Earth but I noticed it when I started exploring the code base.

javache:
Actually, it would be fine to make these Objective-C++ as well, and just use std::atomic everywhere.

Libraries/Image/RCTImageStoreManager.m
```diff
@@ -138,13 +138,14 @@ - (BOOL)canHandleRequest:(NSURLRequest *)request
 - (id)sendRequest:(NSURLRequest *)request withDelegate:(id<RCTURLRequestDelegate>)delegate
 {
   __block volatile uint32_t cancelled = 0;
```
javache: You forgot to remove the old definition here

Adlai-Holler:
@javache I've rebased and removed the leftover def. I'm willing to migrate these to C++ but it'll be easier if there's an xcodeproj I can use that will build the files. If I use `React.xcodeproj` the RCTImage implementation files aren't included, and the `ReactLegacy.xcodeproj` and `RCTImage.xcodeproj` don't seem to want to build on their own. Any advice for a solid dev environment?

javache:
I typically use the RNTester.xcodeproj (https://github.com/facebook/react-native/tree/master/RNTester/RNTester.xcodeproj)

PR #15281, opened 2017-07-31
RCTBorderDrawing: Cache results to save resources

Adlai-Holler:
Hi folks. When profiling RNTester on master (688c74) on my iPhone 6 10.3.3 scrolling up and down the grid view example for 10 seconds, average of 3 runs, this patch reduced `RCTGetBorderImage` from 34ms to 4ms. I didn't look at memory consumption but presumably it lowered that as well.

This is a pretty strong effect but the grid view example is pretty border-heavy so you may get less on other screens. Your call. Love your work!

### DIFF
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -733,8 +733,13 @@ RCT_EXTERN NSString *__nullable RCTTempFilePath(NSString *extension, NSError **e
   return [directory stringByAppendingPathComponent:filename];
 }
 
-static void RCTGetRGBAColorComponents(CGColorRef color, CGFloat rgba[4])
+void RCTGetRGBAColorComponents(CGColorRef color, CGFloat rgba[4])
 {
+  if (!color) {
+    rgba = (CGFloat[4]){0, 0, 0, 0};
+    return;
+  }
+  
   CGColorSpaceModel model = CGColorSpaceGetModel(CGColorGetColorSpace(color));
   const CGFloat *components = CGColorGetComponents(color);
   switch (model)
--- a/React/Views/RCTBorderDrawing.m
+++ b/React/Views/RCTBorderDrawing.m
@@ -573,7 +573,11 @@ UIImage *RCTGetBorderImage(RCTBorderStyle borderStyle,
         // We already checked this above.
         return nil;
     }
-    [borderImageCache setObject:image forKey:key];
+
+    // Handle cases where the above functions return nil.
+    if (image != nil) {
+      [borderImageCache setObject:image forKey:key];
+    }
   }
   
   return image;
--- a/React/Views/RCTBorderDrawing.m
+++ b/React/Views/RCTBorderDrawing.m
@@ -505,14 +505,16 @@ static id RCTGetBorderImageKey(RCTBorderStyle borderStyle,
                                CGColorRef backgroundColor,
                                BOOL drawToEdge)
 {
-  // Important that this struct is aligned on all architectures, else
-  // the compiler can insert garbage into the padding and corrupt our hash.
+  // Important that this struct is not padded on any architecture, else
+  // there can be garbage in the padding that corrupts our hash.
+#pragma clang diagnostic push
+#pragma clang diagnostic error "-Wpadded"
   struct {
     RCTBorderStyle borderStyle;
     CGSize viewSize;
     RCTCornerRadii cornerRadii;
     UIEdgeInsets borderInsets;
-    NSUInteger drawToEdge; // BOOL would be misaligned, not good.
+    NSUInteger drawToEdge; // BOOL would be padded.
     CGFloat topColors[4];
     CGFloat rightColors[4];
     CGFloat bottomColors[4];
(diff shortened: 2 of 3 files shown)


React/Views/RCTBorderDrawing.m
```diff
@@ -484,8 +485,6 @@ static CGContextRef RCTUIGraphicsBeginImageContext(CGSize size, CGColorRef backg
   CGContextSetLineWidth(ctx, lineWidth);
   CGContextSetLineDash(ctx, 0, dashLengths, sizeof(dashLengths) / sizeof(*dashLengths));
 
-  CGContextSetStrokeColorWithColor(ctx, [UIColor yellowColor].CGColor);
```
Adlai-Holler: I took this out because it looks like it was there by mistake. The stroke color gets overwritten two lines down, before the next draw operation.

React/Views/RCTBorderDrawing.m
```diff
@@ -498,6 +497,48 @@ static CGContextRef RCTUIGraphicsBeginImageContext(CGSize size, CGColorRef backg
...
+    {0,0,0,0},
+    {0,0,0,0},
+  };
+  RCTGetRGBAColorComponents(borderColors.top, borderParams.topColors);
+  RCTGetRGBAColorComponents(borderColors.bottom, borderParams.bottomColors);
+  RCTGetRGBAColorComponents(borderColors.left, borderParams.leftColors);
+  RCTGetRGBAColorComponents(borderColors.right, borderParams.rightColors);
+  RCTGetRGBAColorComponents(backgroundColor, borderParams.backgroundColors);
+  
+  return [NSValue valueWithBytes:&borderParams objCType:@encode(typeof(borderParams))];
```
Adlai-Holler: You could use NSData here, and it would actually be slightly faster since the `isEqual:` wouldn't have to check the underlying type. But it also means another heap allocation – NSValue stores the data on the object itself instead of pointing into the heap. Plus it just feels more purpose-built.

React/Views/RCTBorderDrawing.m
```diff
@@ -498,6 +497,48 @@ static CGContextRef RCTUIGraphicsBeginImageContext(CGSize size, CGColorRef backg
...
+static id RCTGetBorderImageKey(RCTBorderStyle borderStyle,
+                               CGSize viewSize,
+                               RCTCornerRadii cornerRadii,
+                               UIEdgeInsets borderInsets,
+                               RCTBorderColors borderColors,
+                               CGColorRef backgroundColor,
+                               BOOL drawToEdge)
+{
+  // Important that this struct is aligned on all architectures, else
+  // the compiler can insert garbage into the padding and corrupt our hash.
```
javache: Can we enforce this using a compiler pragma somehow?

React/Base/RCTUtils.m
```diff
@@ -733,8 +733,13 @@ BOOL RCTIsLocalAssetURL(NSURL *__nullable imageURL)
   return [directory stringByAppendingPathComponent:filename];
 }
 
-static void RCTGetRGBAColorComponents(CGColorRef color, CGFloat rgba[4])
+void RCTGetRGBAColorComponents(CGColorRef color, CGFloat rgba[4])
```
javache: Is this an expensive operation? Should we try to normalize the UIColor to this and then use it across the whole border rendering operation?

Adlai-Holler: It's inexpensive, almost never showed up at 1ms sampling on my iPhone 6 so I'd stick with the CGColor.

React/Views/RCTBorderDrawing.m
Adlai-Holler: Yep we can, I initially chose not to for the sake of visual cleanliness but thanks for the feedback, safety is king. Ill drop it in later today.

javache:
Looks good to me. I'll have @shergin take a look as well once he's back from vacation.

React/Views/RCTBorderDrawing.m
```diff
@@ -498,6 +497,51 @@ static CGContextRef RCTUIGraphicsBeginImageContext(CGSize size, CGColorRef backg
...
+static id RCTGetBorderImageKey(RCTBorderStyle borderStyle,
+                               CGSize viewSize,
+                               RCTCornerRadii cornerRadii,
+                               UIEdgeInsets borderInsets,
+                               RCTBorderColors borderColors,
+                               CGColorRef backgroundColor,
+                               BOOL drawToEdge)
+{
+  // Important that this struct is not padded on any architecture, else
+  // there can be garbage in the padding that corrupts our hash.
```
shergin: This is not technically `hash`, so we should not call it `hash`.

React/Views/RCTBorderDrawing.m
```diff
@@ -506,16 +550,38 @@ static CGContextRef RCTUIGraphicsBeginImageContext(CGSize size, CGColorRef backg
...
-      return RCTGetDashedOrDottedBorderImage(borderStyle, cornerRadii, viewSize, borderInsets, borderColors, backgroundColor, drawToEdge);
-    case RCTBorderStyleUnset:
-      break;
+  static NSCache<id, UIImage *> *borderImageCache;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    borderImageCache = [NSCache new];
+  });
+  
+  if (borderStyle == RCTBorderStyleUnset) {
```
shergin: I would split this function to two: `RCTGetBorderImage` and `RCTGetCachedBorderImage`.

chirag04:
seems like something we want to ship. better not let this go stale. @Adlai-Holler can you update as per @shergin's review?

PR #15292, opened 2017-07-31
Fix broken Buck installation link

fson:
Running `./scripts/run-android-local-unit-tests.sh` without having
installed Buck displays this link. Make it point to a URL that exists.

### DIFF
--- a/scripts/validate-android-test-env.sh
+++ b/scripts/validate-android-test-env.sh
@@ -10,7 +10,7 @@
 # Check that Buck is working.
 if [ -z "$(which buck)" ]; then
   echo "You need to install Buck."
-  echo "See https://buckbuild.com/setup/install.htm for instructions."
+  echo "See https://buckbuild.com/setup/install.html for instructions."
   exit 1
 fi
 


scripts/validate-android-test-env.sh
```diff
@@ -10,7 +10,7 @@
 # Check that Buck is working.
 if [ -z "$(which buck)" ]; then
   echo "You need to install Buck."
-  echo "See https://buckbuild.com/setup/install.htm for instructions."
+  echo "See https://buckbuild.com/setup/install.html for instructions."
```
chirag04: should we link directly to `https://buckbuild.com/setup/getting_started.html`?

fson: There are [multiple Buck installation links](https://github.com/facebook/react-native/search?utf8=✓&q=buckbuild.com%2Fsetup&type=) in the docs and they all point to https://buckbuild.com/setup/install.html. Maybe it's easier to make sure the link doesn't break in the future, if they're all the same? We could also change them all to https://buckbuild.com/setup/getting_started.html, which is the URL `/setup/install.html` currently redirects to, but I'm not sure if that's worth the trouble?

chirag04: I just thought that redirect might break in future. if you think it's not worth the trouble then sure, let's run with it.

chirag04: cc @hramos

Issue #10756, opened 2016-11-04
Android: "FATAL EXCEPTION: mqt_js" / Error calling function: RCTDeviceEventEmitter:emit / - malformed JSON parsing in RN's android native code


GeoffreyPlitt:
### Description

Calling fetch() and expecting JSON results (i.e. using ```response.json()```) will crash if server sends malformed JSON. Error/logging system prints "Failed to create String from JSON:" and then does NOT print the JSON (the problem this ticket IS about).

https://github.com/facebook/react-native/issues/9117 promised to fix this, but did not.

### Reproduction

Call fetch() and then response.json() with a server that returns bad JSON.

### Additional Information

* React Native version: 0.36
* Platform: Android
* Operating System: MacOS

Here's an example log:

```
11-04 13:39:46.611  3509  3509 D EPDG -- [EPDGService]: onDataActivity: direction=3
11-04 13:39:46.611  7541  7836 D NetworkController.MobileSignalController(0/2): onDataActivity: direction=3
11-04 13:39:46.781 17917 17917 V MediaPlayer-JNI: getCurrentPosition: 617 (msec)
11-04 13:39:47.031 17917 17917 V MediaPlayer-JNI: getCurrentPosition: 617 (msec)
11-04 13:39:47.281 17917 17917 V MediaPlayer-JNI: getCurrentPosition: 617 (msec)
11-04 13:39:47.541 17917 17917 V MediaPlayer-JNI: getCurrentPosition: 617 (msec)
11-04 13:39:47.641  3509  3509 D EPDG -- [EPDGService]: onDataActivity: direction=1
11-04 13:39:47.641  7541  7836 D NetworkController.MobileSignalController(0/2): onDataActivity: direction=1
11-04 13:39:47.791 17917 17917 V MediaPlayer-JNI: getCurrentPosition: 617 (msec)
11-04 13:39:47.821 17917 17976 I System.out: (HTTPLog)-Static: isSBSettingEnabled false
11-04 13:39:47.821 17917 17976 I System.out: (HTTPLog)-Static: isSBSettingEnabled false
11-04 13:39:47.821  3079  3883 D EnterpriseController: netId is 0
11-04 13:39:47.821  3079  3883 D Netd    : getNetworkForDns: using netid 507 for uid 10726
11-04 13:39:47.821 17917 17968 E AndroidRuntime: FATAL EXCEPTION: mqt_js
[...]

PR #15295, opened 2017-07-31
Fix crash due to mishandling of UTF-8 in progressive download.

dryganets:
Fixes:
```
Fatal Exception: java.lang.RuntimeException: Failed to create String from JSON
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
       at android.os.Looper.loop(Looper.java:234)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:193)
       at java.lang.Thread.run(Thread.java:818)
```
JavaScriptCore is very strict about invalid UTF symbols.
So if you pass an invalid UTF-8 string to it the string will be decoded as an empty string.

The current implementation of progressive downloading for Android blindly cuts the response in 8KB chunks.
That could cause a problem in case the last symbol in the chunk is multi-byte.

To prevent it I added a class which determines if this is the case and cut the string in the appropriate place.
A remainder is prepended to the next chunk of data.

This should fix the root cause of this issue:
https://github.com/facebook/react-native/issues/10756

Test plan:
fetch for this http://www.00ksw.org/html/1/1391/614905.html
shouldn't crash android app anymore.

Release Notes:


dryganets:
continuous-integration/travis-ci/pr seems to be false-negative.
At least it looks like a bug in the CI from the logs.
I didn't change any c++/ObjC or javascript code to fail xcode tests.

javache:
I believe we've had a similar issue elsewhere in the Android codebase. It should be possible to merge/reuse your solution with that.

dryganets:
@javache, could you point me to another place?

I grep the code locally and it looks like NetworkModule is the only place where we split user data to chunks.

I was looking for the String creation with constructors:
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String(byte[])
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#String(byte[],%20java.nio.charset.Charset)

etc.

javache:
Hmm, you're right. I think we just avoided crashing in the C++ layer when receiving invalid input.

An equivalent fix to this was made on iOS a while ago: https://github.com/facebook/react-native/commit/3ac3749ac37bf5f202eab535528f324461029366

dryganets:
Tell me if I could do something to push it forward.
It seems to be a pretty frequent crash.
At least at Skype, it was one of the top crashes.

To make this fool proof I fixed the code causing the crash as well + have a folly slightly patched.
But I think that should be part of the separate commit.

javache:
I'll try to have a look at the code, but in general I'm not comfortable with using our own utf8/utf16 decoding solution. I much prefer the iOS solution where we use the system decoding and then try to fall back / trim the end of the string in case it ends in an invalid character.

dryganets:
@javache I don't have a custom encoding, just a validation.

But I got what you mean. The validation could be done in O(1) by analyzing the last 4 symbols.
Didn't notice that initially. But anyway there is a https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html
API which allows doing what is needed with a support of multiple encodings.

I'll update the algorithm to support other encodings as well.

Thank you for feedback.

dryganets:
Is there anything I could do to help to move it to master?
The patch fixes the root cause of one of the most frequent react-native crashes.

The code I wrote is covered by UnitTests and supports multiple encodings.
It works the same way as iOS implementation does.

dryganets:
Rebased.

dryganets:
@hramos, did you find anyone to take a look at this PR?

It's quite a bad crash.

PR #15302, opened 2017-08-01
Bug fixes

xiehui999:
When scrollToIndex's text is not a number or number bigger than data length,an exception  occurs.


guilharj:
@xiehui999 What about check the lint comments and also, include a check if listRef is undefined? :)

PR #15304, opened 2017-08-01
troubleshooting docs (clone #14885)

formatlos:
I added troubleshooting docs for connection to the development server from device, otherwise it's maybe hard to find out why it's not working.

Due to a disparity in what GitHub's API return the original PR (https://github.com/facebook/react-native/pull/14885) couldn't get imported, so I created a new PR as @hramos suggested.


formatlos:
still not working?

formatlos:
@hramos do you have any clue why the build fails? 

As far as I can see the build fails because of a flow version mismatch.  
> Wrong version of Flow. The config specifies version ^0.49.1 but this is version 0.51.1 

`.flowconfig` and `package.json`both point to 0.51.1. the only place I could find ^0.49.1 was `local-cli/templates/HelloWorld/_flowconfig`

This seems to be a general issue with your CI-setup and should also occur on other PRs. As I only touched docs the tests should still run as expected. Just to be sure I also ran all the test locally which went fine.

formatlos:
just looked at some other recent PRs they also fail with the same error

hramos:
It's timing out. It's not due to any problem with the PR itself.

hramos:
@formatlos the time out is due to a linter check we perform on images prior to import. The proper long term fix is for us to address this issue on our side; short term, we could try replacing the screengrab with a smaller file size-wise.

formatlos:
@hramos sure no problem, what's the maximum file-size allowed?

hramos:
There's no definite file size limit. The import process as one of the lints, checks the PNG to make sure it is optimized. This appears to be taking long enough to make the whole process time out.

hramos:
Don't worry about updating the PR just yet. If this latest import fails, I can still get your commits in manually. I'm really sorry this has been so painful! The issue that blocked your original PR is unrelated to this, you just happened to unluckily hit two issues in a row. I'll get this merged soon.

PR #15315, opened 2017-08-01
Update Linking.js

capdilla:


javache:
Imports should be done using `<React/...>`. Alternative solutions are not supported by RN.

PR #15317, opened 2017-08-01
Improve support for unbundle feature

dmueller39:
unbundle is a useful feature, and it should be exposed. In order to get the most use out of
we expose it as an option at build time in the Build Phase on XCode and the project.ext.react
config in the build.gradle.

Because it is best used with inline requires we add a section under performance that describes
how inline requires can be implemented and how to use with the unbundling feature.

Testing:

- Added a section of the doc which explains how the feature can be enabled
- Use the instructions, build a build on iOS + android (using release so that the bundle is created) and confirm that the bundle has the binary header information.


dmueller39:
@ericvicenti @brentvatne  Should I be recommending something like this instead? https://github.com/zertosh/babel-plugin-transform-inline-imports-commonjs

chirag04:
this is great! cc @hramos @javache

ericvicenti:
@davidaurelio, can you help review this, and perhaps provide clarity on which use cases are important for performance?

javache:
Thanks for writing this up! We should definitely mention something about the startup section too (the initial code segment). We've found this significantly impacts start time as can we load all core modules at once instead of going back and forth over the bridge. (see https://github.com/facebook/metro-bundler/blob/ce0da03a0576fad0b01a7a2e3ee9df0760e35f5d/packages/metro-bundler/src/Bundler/index.js#L52). Ideally you also disable inline requires for the modules inside the preloadedModules.

dmueller39:
@javache OK, I'll add that. I think that means that I shouldn't be suggesting https://github.com/zertosh/babel-plugin-transform-inline-imports-commonjs, because it is applied globally and the method I suggested is more explicit about what is required inline.

javache:
I believe the packager has a default inline requires transform ([this one I believe](https://github.com/facebook/fbjs/blob/master/packages/babel-preset-fbjs/plugins/inline-requires.js)). We configure it through [transform.inlineRequires](https://github.com/facebook/fbjs/blob/master/packages/babel-preset-fbjs/plugins/inline-requires.js)

dmueller39:
@javache  I'll keep the introduction to the concept, but I may remove the higher order component example unless you think its necessary. I will take a stab at describe configuring inlineRequires.

Can you confirm that my understanding is right here?

I believe the transform is based on code paths. the inline require happens when the symbol is first referenced. There is a common pattern in navigation packages where you would define your navigation config upfront like this:

```
import Screen1 from './Screen1';
import Screen2 from './Screen2';
import Screen3 from './Screen3';
import Screen4 from './Screen4';
...

export default function navigationConfig() {
  return {
    screen1: Screen1,
    screen2: Screen2,
    screen3: Screen3,
    screen4: Screen4,
    ...
  };
}
```

I believe that in code like this, the transform would require all the Screens as soon as navigationConfig() is called. Can you confirm?

dmueller39:
@javache Also, are there built in tools for getting the list of preloaded modules? I'm just watching require and then logging all the required modules when the app is done starting up.

dmueller39:
@javache https://github.com/facebook/metro-bundler/pull/37

dmueller39:
This now depends on https://github.com/facebook/metro-bundler/pull/37, because `getModules()` is required to build that moduleName list.

dmueller39:
@javache @hramos @davidaurelio What can I do to move this PR along? I am happy to modify/excise portions. I do feel this is something that a lot of devs have asked for.

PR #15321, opened 2017-08-01
Fixed link command to return a promise

axemclion:
Update to use the promises based API that inquirer uses

As a part of #https://github.com/facebook/react-native/commit/bada25d158e070fd683a10b8ff9cdcb8d2498b14 , a newer version of inquirer was added.

The signature for inquirer@0.12 used to be inquirer.prompt(questions, callback). This has changed to a promised based signature. However the link commands in this line has not been updated to pick up this change.

This PR fixes that.

### DIFF
--- a/local-cli/link/pollParams.js
+++ b/local-cli/link/pollParams.js
@@ -5,5 +5,5 @@ module.exports = (questions) => new Promise((resolve, reject) => {
     return resolve({});
   }
 
-  inquirer.prompt(questions, resolve);
+  inquirer.prompt(questions).then(resolve, reject);
 });


ide:
@facebook-github-bot shipit

javache:
As far as I can tell, you already fixed this a month ago: https://github.com/facebook/react-native/blob/master/local-cli/link/pollParams.js

sergey-akhalkov:
Hi @javache, the fix has been applied for RN 0.46.x, but currently it is absent in RN 0.47.0, please take a look at https://github.com/facebook/react-native/blob/v0.47.0/local-cli/link/pollParams.js#L8

grabbou:
Fixing it now.

GantMan:
Question of a different subject.   Is rn-link going to have more user questions built in?  I didn't notice when this feature was added officially.   Also, interested in if we can circumvent the questions/flag as answers?

guperrot:
Hi, do you have a timeline for releasing this in a stable version?

grabbou:
It's already released.

@Gant - not that I am aware of. Theres interface for asking questions (we
might want to document a bit?) but there's no plan to include any builtins.

grabbou:
No probs! Let's follow up on Slack and come back here once we got a PR :)

PR #13529, opened 2017-04-16
[FlatList] Add setNativeProps

reneweb:
## Motivation (required)

Curently FlatList does not implement setting native props directly like the old ListView did. This pr introduce the `setNativeProps` function which delegates to MetroListView or VirtualizedList. Thos don't have `setNativeProps` handling either, so, I delegated further to the respective ListView and Scroll components, which do have handling for it, thus, allowing to set the native props through FlatList.

## Test Plan (required)

Create a project with a FlatList and change a native property using `setNativeProps`:

```javascript
  componentDidMount() {
        setInterval(() => {
            this.list.setNativeProps({ style: {backgroundColor:"white"} })
        }, 1000)
    }

  render() {
    return (
      <View style={styles.container}>
        <FlatList ref={component => this.list = component}
                  style={{backgroundColor:"black"}}
                  data={[{key: 'a'}, {key: 'b'}]}
                  renderItem={({item}) => <Text>{item.key}</Text>} />
      </View>
    )
  }
```

Fixes #13501

PR #15328, opened 2017-08-02
add setNativeProps to SectionList

octopitus:
# Motivation

#13529 added `setNativeProps` to `FlatList`. So it would be nice to support it in `SectionList` too. Because sometime, for example, I need to disable scroll functionality when using with `SwipeableRow`.

# Test plan

```js
componentDidMount() {
  this._list.setNativeProps({scrollEnabled: false})
}

render() {
  const sections = [
    {key: 'foo', data: [{name: 'Jone Doe', key: 'a'}, {name: 'Susan Briz', key: 'b'}],
    {key: 'bar', data: [{name: 'David Mark', key: 'c'}, {name: 'Daniel Campbell', key: 'd'}]
  ]

  return (
    <SectionList
      ref={c => (this._list = c)}
      sections={sections}
      renderItem={({item}) => <Text>{item.name}<Text>}
    />
  )
}
```

### DIFF
--- a/Libraries/Lists/SectionList.js
+++ b/Libraries/Lists/SectionList.js
@@ -317,6 +317,13 @@ class SectionList<SectionT: SectionBase<any>> extends React.PureComponent<
     }
   }
 
+  setNativeProps(props: Object) {
+    const listRef = this._wrapperListRef && this._wrapperListRef.getListRef();
+    if (listRef) {
+      listRef.setNativeProps(props);
+    }
+  }
+
   render() {
     const List = this.props.legacyImplementation
       ? MetroListView


Libraries/Lists/SectionList.js
```diff
@@ -317,6 +317,13 @@ class SectionList<SectionT: SectionBase<any>> extends React.PureComponent<
     }
   }
 
+  setNativeProps(props: Object) {
+    const listRef = this._wrapperListRef && this._wrapperListRef.getListRef();
```
javache: Instead of accessing `getListRef()` here, why not do

```
if (this._wrapperListRef) {
  this._wrapperListRef.setNativeProps(props);
}
```

octopitus: It will not work if `legacyImplementation` prop is true, because `VirtualizedSectionList` doesn't have `setNativeProps`.

octopitus: Should I add `setNativeProps` method to `VirtualizedSectionList`?

PR #15333, opened 2017-08-02
Remove old Flow modules

ovr:
Hey!

cC @acdlite @bvaughn 

https://github.com/flowtype/flow-typed/pull/773 was merged

P.S I didnt find where `flow-typed` was installed

Test-plan:

1. `npm run flow` (I run it inside root directory of `react-native`, possible I am wrong)

Thank


janicduplessis:
I don't think we are using flow-typed, not sure if there is a reason why. This means these files are still needed unless we start using flow-typed, which would be nice but might cause a bunch of new flow errors that would need to be fixed.

bvaughn:
If that's the case, we should probably just remove the TODO comments then b'c they are a bit confusing 😅

PR #15334, opened 2017-08-02
WebSocket API change to make room for other connection options (SSL pinning)

rh389:
This is a simple groundwork PR to allow options to be passed to the `WebSocket` constructor. It represents a minor change to an undocumented part of the API, moving `headers` to within `options`.

This will be a BC for anyone manually specifying headers other than `origin` but a) that's not a common use case with WebSockets and b) it's not documented even in code and wouldn't currently pass a flow check.

NB: The third argument to the WebSocket constructor isn't part of the W3C spec, so I think this is a good place for RN-specific named parameters, better than adding a fourth argument. `protocols` needs to stay where it is, in line with the spec.

### Motivation
If this goes through I'd like to build on it by adding an additional connection option for SSL certificate pinning, as already supported by the underlying `okhttp` and `RCTSRWebSocket`. It could later be expanded for various other uses.

Currently, there's no way for a `WebSocket` user to specify any connection options other than url, protocol and headers. The fact that `WebSocket` connects in its constructor means any options have to go in there.

### Test Plan
Connect to a websocket server using iOS and Android, observe the connection headers:
1. Without specifying `origin`, the default header should be set
2. Specifying it in the old way `new WebSocket(url, protocols, { origin: 'customorigin.com' })`
3. Specifying it in the new way `new WebSocket(url, protocols, { headers: { origin: 'customorigin.com' }})`.

I've tested myself using the test app with iOS and Android.

### DIFF
--- a/Libraries/WebSocket/WebSocket.js
+++ b/Libraries/WebSocket/WebSocket.js
@@ -99,13 +99,18 @@ class WebSocket extends EventTarget(...WEBSOCKET_EVENTS) {
       protocols = [protocols];
     }
 
-    // Backwards compatibility
-    if (options && typeof options.origin === 'string') {
-      options = {
-        headers: {
-          origin: options.origin
-        }
-      };
+    const { headers = {}, origin, ...unrecognized } = options || {};
+
+    // Preserve deprecated backwards compatibility for 'options.origin'
+    if (origin && typeof origin === 'string') {
+      console.warn(`Specifying "origin" as a WebSocket connection option is deprecated. Include it under "headers" instead.`);
+      headers.origin = origin;
+    }
+
+    // Warn about and discard anything else
+    if (Object.keys(unrecognized).length > 0) {
+      console.warn(`Unrecognized WebSocket connection option(s) "${Object.keys(unrecognized).join(`", "`)}". `
+        + `Did you mean to put these under "headers"?`);
     }
 
     if (!Array.isArray(protocols)) {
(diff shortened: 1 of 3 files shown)


javache:
Can you log a warning in the backwards compatibility codepath? That way, in a couple of release, we can remove it altogether.

Libraries/WebSocket/WebSocket.js
```diff
@@ -93,12 +93,26 @@ class WebSocket extends EventTarget(...WEBSOCKET_EVENTS) {
...
   static isAvailable: boolean = !!WebSocketModule;
 
-  constructor(url: string, protocols: ?string | ?Array<string>, options: ?{origin?: string}) {
+  constructor(url: string, protocols: ?string | ?Array<string>, options: ?{ headers?: { origin?: string}}) {
     super();
     if (typeof protocols === 'string') {
       protocols = [protocols];
     }
 
+    const { headers = {}, origin, ...unrecognized } = options || {};
```
rh389: `npm run lint` doesn't have any problem with this.

rh389:
@javache Done (after a brief fight with eslint!) - now 'origin' gets a deprecation warning, any other unrecognised option gets a warning and discarded.

Eg with: 
```js
new WebSocket(this.state.url, null, { 
  origin: 'test', 
  headers: { 
    'X-Header': 'test' 
  }, 
  'X-Header-2': 'Test' 
});
```
`options` is resolved to 
```js 
{ 
  headers: { 
    'X-Header': 'test', 
    origin: 'test' 
  } 
}
``` 
and we get warnings:

simulator screen shot 3 aug 2017 13 26 32

Issue #15313, opened 2017-08-01
ComponentWillUnmout not called for Modal in AppleTV


rada:
### Is this a bug report?

Yes

### Have you read the Bugs section of the How to Contribute guide?

Yes

### Environment

1. `react-native -v`: 0.46.4
2. `node -v`: 6.5.0
3. `npm -v`: 3.10.3

Then, specify:

- Target Platform: AppleTV

- Development Operating System: macOS Sierra (10.12.5)

- Build tools: Xcode (8.3.3)

### Steps to Reproduce

Open Modal in Tab 1 by clicking on 'Open Modal' touchable.

(Write your steps here:)

1. Focus touchable 'Open Modal' (in simulator press key Down)
2. Press touchable 'Open Modal'
3. Press 'Menu' button on Apple TV Remote (in simulator, press CMD + SHIFT + R to display TV Remote)

### Expected Behavior

After the 'Menu' button is pressed on Apple TV Remote, Modal is closed and the method componentWillUnmout is called for component used in Modal.

### Actual Behavior

Modal is closed but the componentWillUnmout method is not called if the modal is closed by pressing the 'Menu' button in the Apple TV Remote. If the modal is closed by pressing the touchable 'Close Modal', the componentWillUnmout method is called as expected.

### Reproducible Demo

https://gitlab.com/radamich/tvTabBar/tree/tvosModal

index
```javascript
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 * @flow
 */

import React, { Component } from 'react';
import {
  AppRegistry,
  Text,
  View,
  TabBarIOS,
  ListView,
  Modal,
  TouchableHighlight,
} from 'react-native';

import List from './List';

export default class tvTabBar extends Component {

	_tabItems: Array<{ key: string }> = [
		{ key: 'TAB 1' },
		{ key: 'TAB 2' },
		{ key: 'TAB 3' },
		{ key: 'TAB 4' },
	];

	constructor(props) {
		super(props);
		this.state = {
			activeTab: 'TAB 1',
			showModal: false,
		};
	}

	_getTabItem = (key: string) => {
		switch (key) {
			case 'TAB 1':
				return (
					<View style={{ flex: 1, top: 140, borderColor: 'red', borderWidth: 10, padding: 200 }}>
						<Modal 
							visible={this.state.showModal}
							transparent={true}
						>
							<List closeModal={this.closeModal} />

						</Modal>
						<TouchableHighlight onPress={() => {
[...]

PR #15341, opened 2017-08-02
Fix for Modal behavior when menu button pressed on Apple TV (Issue #15313)

dlowder-salesforce:
**Motivation**

On Apple TV, pressing the menu button destroys the native view that backs the `Modal` component, causing an app using this component to get into a broken state.  This fix implements `onRequestClose` for tvOS to have the same behavior as it does for the Android back button.

**Test plan**

Manually tested this with the `ModalExample` in the `RNTester` app.  See also the test code in issue #15313.


React/Views/RCTModalHostView.m
```diff
@@ -89,6 +101,13 @@ - (void)insertReactSubview:(UIView *)subview atIndex:(NSInteger)atIndex
   RCTAssert(_reactSubview == nil, @"Modal view can only have one subview");
   [super insertReactSubview:subview atIndex:atIndex];
   [_touchHandler attachToView:subview];
+#if TARGET_OS_TV
+  if (_onRequestClose) {
```
shergin: `onRequestClose` can be set later, right? So, probably we don't need this check here.

React/Views/RCTModalHostView.m
```diff
@@ -54,6 +57,15 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
   return self;
 }
 
+#if TARGET_OS_TV
+- (void)menuButtonPressed:(UIGestureRecognizer *)r
```
shergin: `- (void)menuButtonPressed:(__unused UIGestureRecognizer *)gestureRecognizer`

React/Views/RCTModalHostView.m
```diff
@@ -54,6 +57,15 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
...
+{
+    if(_onRequestClose) {
+        _onRequestClose(@{});
```
shergin: Can/should we pass `nil` here?

React/Views/RCTModalHostView.m
shergin: And I would prefer to create `_tapGestureRecognized` inside `init` method.

React/Views/RCTModalHostView.m
dlowder-salesforce: I'll try that.

React/Views/RCTModalHostView.m
dlowder-salesforce: In my testing, `onRequestClose` gets set before `insertReactSubview`.  We definitely don't want the gesture recognizer to be added unless `onRequestClose` is implemented, otherwise the default menu button action won't happen.

React/Views/RCTModalHostView.m
dlowder-salesforce: Ok.

React/Views/RCTModalHostView.m
dlowder-salesforce: @shergin I've made the changes you requested

shergin: Why we don't want to? Just because of performance reasons or because it changes default dehaviour?

I think there is no logical connections between adding subviews and setting `onRequest` property. Let's imagine that we apply this property inside `render` function only if some condition is true, so it can happene after the view three is established. Does this make sense?

So, I believe we have to override setter `setOnRequsetClose` and establish gesture recognizer there. And we also have to reinfoce it in `insertReactSubview:`.

dlowder-salesforce: My concern is that if the gesture recognizer is there, but there is no `onRequestClose`, then the application will be stuck with the modal showing.  That said, your statements about needing to override `setOnRequestClose` are correct, and I'll make that change.

dlowder-salesforce: Requested changes are made.

React/Views/RCTModalHostView.m
```diff
@@ -54,6 +61,27 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
...
+        _onRequestClose(nil);
+    }
+}
+
+- (void)setOnRequestClose:(RCTDirectEventBlock)onRequestClose
+{
+  _onRequestClose = onRequestClose;
+  if([self.subviews count] > 0) {
+    if(_onRequestClose && _tapGestureRecognizer) {
+      [self.subviews[0] addGestureRecognizer:_tapGestureRecognizer];
```
shergin: Using `subviews[]` here is technically correct, but it is extremly fragile abstraction. In better world we would have dedicated property, something like `contentView`, which smart getter and setter would manage gesture recognizers and add/remove it from view hierarchy.

React/Views/RCTModalHostView.m
```diff
@@ -101,6 +134,11 @@ - (void)removeReactSubview:(UIView *)subview
   RCTAssert(subview == _reactSubview, @"Cannot remove view other than modal view");
   [super removeReactSubview:subview];
   [_touchHandler detachFromView:subview];
+#if TARGET_OS_TV
+  if(_tapGestureRecognizer) {
+    [subview removeGestureRecognizer:_tapGestureRecognizer];
```
shergin: Space, please.

React/Views/RCTModalHostView.m
shergin: Oh, we even already have ivar for it with strange name `_reactSubview`.

shergin: Oh, actually, `subviews` is always empty. So, it is exactly what I mean by fragile. 😞

React/Views/RCTModalHostView.m
```diff
@@ -101,6 +134,11 @@ - (void)removeReactSubview:(UIView *)subview
   RCTAssert(subview == _reactSubview, @"Cannot remove view other than modal view");
   [super removeReactSubview:subview];
```
shergin: We have to add a comment here about that fact that super method actually removes `subview` from `_modalViewController.view` because it absolutely not obvious.

React/Views/RCTModalHostView.m
dlowder-salesforce: This class seems to be set up so that `subviews` has either 0 or 1 views in it.  I can use `_reactSubview` instead, I suppose....

dlowder-salesforce:
@shergin no I don't yet have production tvOS apps of my own.  Other people have shipped them already though (e.g. Unsplash for Apple TV), or they are in the process of building them, and as a result are filing RN issues for me to fix :)

dlowder-salesforce:
@shergin I've made the changes you requested on this PR.

Issue #5293, opened 2016-01-13
ToolbarAndroid not rendering


ndarilek:
Trying what I'd thought would be a simple view, a toolbar in a DrawerLayout. Eventually plan to add styling and actions but I can't even get this to render:

```
import React, {
  AppRegistry,
  Component,
  DrawerLayoutAndroid,
  StyleSheet,
  Text,
  ToolbarAndroid,
  View
} from 'react-native';

class Perceptron extends Component {

  renderNavigationView() {
    return <View>
      <Text>Home</Text>
    </View>
  }

  onActionSelected(position) {
  }

  render() {
    return <DrawerLayoutAndroid
      drawerWidth={300}
      drawerPosition={DrawerLayoutAndroid.positions.Left}
      renderNavigationView={this.renderNavigationView}
      >
      <ToolbarAndroid
        title="Perceptron"
        navIcon={require("./icons/empty.png")}
        onActionSelected={this.onActionSelected}
        actions = {[
          {title: "Log out", show: "never"}
        ]}
        />
      <Text>Hello, world from the toolbar!</Text>
    </DrawerLayoutAndroid>
  }

}

AppRegistry.registerComponent("Perceptron", () => Perceptron)
```

What am I missing? The only item that renders is the "Hello, world from the toolbar!" text.

Thanks.

ndarilek:
Happening on React Native 0.18.0rc.

Ehesp:
Try adding a height to the toolbar?

ndarilek:
That did it. Thanks! Any reason this doesn't just get a default height?

Ehesp:
Haven't a clue to be honest, I found it pretty strange first time. I guess there should be, @christopherdro worth a PR?

Pungsnigel:
I'm getting the same error, even when I do specify the height of the toolbar. Any ideas? Running react-native 0.21.0

damusnet:
@facebook-github-bot stack-overflow

facebook-github-bot-0:
Hey @ndarilek and thanks for posting this! @damusnet tells me this issue looks like a question that would be best asked on [StackOverflow](http://stackoverflow.com/questions/tagged/react-native). StackOverflow is amazing for Q&A: it has a reputation system, voting, the ability to mark a question as answered. Because of the reputation system it is likely the community will see and answer your question there. This also helps us use the GitHub bug tracker for bugs only. Will close this as this is really a question that should be asked on SO.

ndarilek:
[...]

Issue #7915, opened 2016-06-03
Some components (ProgressBarAndroid and ToolbarAndroid) is not Mounting


holoto:
# Some components useless 
## Genymotion setting
![Genymotion setting](http://7xv1u3.com1.z0.glb.clouddn.com/Screen%20Shot%202016-06-03%20at%2011.13.01%20PM.png)
## Screenshot,ps:Has Reload Js
![Screenshot](http://7xv1u3.com1.z0.glb.clouddn.com/Screen%20Shot%202016-06-03%20at%2011.10.36%20PM.png)
![Screenshot](http://7xv1u3.com1.z0.glb.clouddn.com/Screen%20Shot%202016-06-03%20at%2011.10.53%20PM.png)
### index.android.js
``` 
/*
* @Author: holoto
* @Date:   2016-06-03 22:19:53
* @Last Modified by:   holoto
* @Last Modified time: 2016-06-03 22:20:30
*/
// import React,{Component} from 'react';

import {
  AppRegistry
} from 'react-native';

import {ProgressBarAndroidDemo} from './MyCp/ProgressBarAndroidDemo';
import {ToolbarAndroidDemo} from './MyCp/ToolbarAndroidDemo';


AppRegistry.registerComponent('demo',()=>ToolbarAndroidDemo);
```
### in MyCp; ToolbarAndroidDemo.js
```
/*
* @Author: holoto
* @Date:   2016-06-03 21:10:07
* @Last Modified by:   holoto
* @Last Modified time: 2016-06-03 23:10:27
*/
import React,{Component} from 'react';
import {
	View,
	ProgressBarAndroid
} from 'react-native';
var ToolbarAndroid=require('ToolbarAndroid');
var ProgressBar=require('ProgressBarAndroid');
class ToolbarAndroidDemo extends Component{
	constructor(props) {
	  super(props);

	  this.state = {};
	}
	render(){
		return(
			<View>
			<ToolbarAndroid/>
			<ProgressBarAndroid
			styleAttr="Large"
			/>
			<ProgressBar/>
			</View>
			)
	}
};



export {ToolbarAndroidDemo};



```
## Other components can use

damusnet:
@facebook-github-bot duplicate #5293

Try setting a height to your `<ToolbarAndroid/>`. Also set styles to your view to fill the screen.

damusnet:
@facebook-github-bot duplicate #5293

facebook-github-bot-9:
[...]

PR #15347, opened 2017-08-02
Include a style in the example for AndroidToolbar

TiraO:
- The toolbar does not render unless height and width are set. It may not be a priority to fix, but we should at least update the docs so the example works.
see: https://github.com/facebook/react-native/issues/5293

### DIFF
--- a/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
+++ b/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
@@ -51,6 +51,7 @@ var optionalImageSource = PropTypes.oneOfType([
  * render: function() {
  *   return (
  *     <ToolbarAndroid
+ *       style={{height: 56, alignSelf: 'stretch'}}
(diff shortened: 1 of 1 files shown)


jkantr:
So this is both still affecting people and undocumented. Can we revisit this? Just a note or style example in the docs is enough.

sudarshanc:
https://facebook.github.io/react-native/docs/toolbarandroid

render: function() {
  return (
    <ToolbarAndroid
      logo={require('./app_logo.png')}
      title="AwesomeApp"
      actions={[{title: 'Settings', icon: require('./icon_settings.png'), show: 'always'}]}
      onActionSelected={this.onActionSelected} />
  )
},
onActionSelected: function(position) {
  if (position === 0) { // index of 'Settings'
    showSettings();
  }
}

No height attribute present in docs. So definitely, this is a bug.

#5293 #7915

The above example does not work without height attribute at least in my case.
RN v0.57

PR #15352, opened 2017-08-03
[Docs] Explain navigating between screens in Navigation Guide

psingl:
Elaborated navigating between screen a little bit. I am sure this will help new react-native developers a lot.

### DIFF
--- a/docs/Navigation.md
+++ b/docs/Navigation.md
@@ -30,6 +30,7 @@ npm install --save react-navigation
 Then you can quickly create an app with a home screen and a profile screen:
 
 ```
+/* from App.js */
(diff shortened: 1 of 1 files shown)


charpeni:
Maybe navigation should be aware of navigation libraries, otherwise we'll have to maintain this doc to match with react-native and react-navigation version.

Thought @hramos ?

psingl:
But react-native documentation must provide this much information that a new user can easily use it. And by the way @charpeni the react-navigation getting started guide on http://reactnavigation.org itself is not up to date. They are still hosting the documentation of previous version check that out.

Issue #14699, opened 2017-06-23
RCTPushNotificationManager does not set alertTitle in UILocalNotification


stage88:
### Description

Setting the alertTitle property on PushNotificationIOS.scheduleLocalNotification does not display the title in the notification.

With alertTitle set:
![image](https://user-images.githubusercontent.com/9906800/27482779-1ca92900-5866-11e7-8578-1afed9a5843d.png)

Without:
![image](https://user-images.githubusercontent.com/9906800/27482878-8083cc28-5866-11e7-8d69-6f4455fc4351.png)

### Reproduction Steps

Use the below code to reproduce.

#### Sample Code

`
PushNotificationIOS.scheduleLocalNotification({
      fireDate: new Date(Date.now() + (30 * 1000)).toISOString(),
      alertTitle: 'Incoming Message',
			alertBody: 'This is my message, please read it for good fortune.',
			soundName: 'default',
			applicationIconBadgeNumber: 1,
			userInfo: {
        id: 11
      }
    })
`

### Solution

Adding the following line in RCTPushNotificationManager.m (54) fixes the issue:

`notification.alertTitle = [RCTConvert NSString:details[@"alertTitle"]];`

### Additional Information

* React Native version: 0.45
* Platform: iOS
* Development Operating System: macOS Sierra 10.12.15
* Build tools: XCode 8.3.3

hramos:
Can you send a PR?

hramos:
Tagging as a Good First Task. A solution is provided in the original post above. People can get familiar with the contribution process by applying the fix in a PR.

tinajohnson:
@hramos I'm ready to create a PR as I'm new to the contribution process and is essential for me to get to know the process.

stage88:
Great work on this guys, thanks!

cpojer:
This issue has been moved to react-native-community/react-native-push-notification-ios#2.

PR #15381, opened 2017-08-05
Set alertTitle in UILocalNotification

tinajohnson:
Currently, since the alertTitle is not set in UILocalNotification, the notification displays just the notification body with no title. This commit sets the alertTitle for local notifications.

Issue: #14699 

## Test Plan

- In a sample RN app, created a component and added the following code to componentDidMount():
`PushNotificationIOS.scheduleLocalNotification({
     fireDate: new Date(Date.now() + (30 * 1000)).toISOString(),
     alertTitle: 'Incoming Message',
     alertBody: 'Test message',
     soundName: 'default'
 })`

- On running the app, the notification appears with both body and title once the component is mounted.

settitle_fix

### DIFF
--- a/Libraries/PushNotificationIOS/PushNotificationIOS.js
+++ b/Libraries/PushNotificationIOS/PushNotificationIOS.js
@@ -159,6 +159,7 @@ class PushNotificationIOS {
    * details is an object containing:
    *
    * - `fireDate` : The date and time when the system should deliver the notification.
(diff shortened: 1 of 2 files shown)


shergin:
@tinajohnson Should we also update documentation?

tinajohnson:
@shergin Yeah, we should. I will add it.

PR #15392, opened 2017-08-07
Fix $FlowFixMe in docs

### LINKED ISSUE #15218: '$FlowFixMe' Appearances throughout Docs
### Is this a bug report?

Yes

### Have you read the Bugs section of the How to Contribute guide?

Yes

### Environment

- Target Platform: Web (tested on Chrome, Safari)

- Development Operating System: MacOS

### Steps to Reproduce

Visit the following sites on the React Native docs.
https://facebook.github.io/react-native/docs/tabbarios.html
https://facebook.github.io/react-native/docs/picker.html
https://facebook.github.io/react-native/docs/statusbar.html

### Expected Behavior

It's important to keep the state of the docs updated, as the docs are the main way that most users look to when figuring how to get their application working. A parameter was expected (such as a boolean or color field).

### Actual Behavior

The various appearances of $FlowFixMe happen because of the way the docs are generated. While the generation works most of the time, these errors prevent any information from being passed and make the components much harder to use, as an assumption must be made by looking at the codebase itself to figure out what components to pass.

rsml:
Previously, the type $FlowFixMe appeared on several pages in the docs (https://facebook.github.io/react-native/docs/tabbarios.html, https://facebook.github.io/react-native/docs/picker.html, https://facebook.github.io/react-native/docs/statusbar.html)

This fixes those types while maintaining consistency with the rest of the docs.

However, it would still be unclear to a user reading the docs what 'pickerStyleType' and 'stylePropType' in the following example means.

For example, this is how http://localhost:8079/react-native/docs/picker.html will look now:
image

Fixes #15218

### DIFF
--- a/Libraries/Components/Picker/Picker.js
+++ b/Libraries/Components/Picker/Picker.js
@@ -26,6 +26,8 @@ var ViewStylePropTypes = require('ViewStylePropTypes');
 
 var itemStylePropType = StyleSheetPropType(TextStylePropTypes);
 
+const stylePropType = StyleSheetPropType(ViewStylePropTypes);
+
 var pickerStyleType = StyleSheetPropType({
   ...ViewStylePropTypes,
   color: ColorPropType,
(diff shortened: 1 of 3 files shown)


shergin:
@rsml Could you please resolve conflicts?

hramos:
This is a good one to fix. Can you fix the conflicts, so we can check if Flow succeeds on Circle?

hramos:

PR #15399, opened 2017-08-07
Adding Callback type import

adhambadr:
the example for javascript callbacks doesn't import the Callback from the com.facebook.bridge package leaving one with the interpretation is might be a java api native type.

(Write your motivation here.)

## Test Plan


javache:
We're not listing out the other imports in this sample, so this feels inconsistent.

PR #15400, opened 2017-08-07
Add pertinent info for fresh flatlisters

GantMan:
I just worked with a fellow dev who was switching to flatlist and ended up pretty surprised at the performance drop.  The measurement had an exponential bridge saturation, causing flatlist to lose to a scrollview!

We knew something was up, but a little note in the docs would have helped.

### DIFF
--- a/Libraries/Lists/FlatList.js
+++ b/Libraries/Lists/FlatList.js
@@ -107,6 +107,7 @@ type OptionalProps<ItemT> = {
    *       {length: ITEM_HEIGHT, offset: ITEM_HEIGHT * index, index}
    *     )}
    *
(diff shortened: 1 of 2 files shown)


brentvatne:
did all of the rows in the list implement `shouldComponentUpdate`? I think ideally this should be done before resorting to `getItemLayout`, but I'm not 100% sure on that cc @sahrens

GantMan:
Hey @brentvatne - yes they were good.  Even tried moving them to PureComponent/Functional.  `getItemLayout` was the magical straw that made it work!  If you'd like I can try to put together an expo snack to be a PoC, but I figured the context is an optimization regardless.

sahrens:
Sure, thanks for the PR :)

PR #15402, opened 2017-08-08
Fix format specifier used in RCTCxxMethod

kastiglione:
[`CxxModule::Method::callbacks`](https://github.com/facebook/react-native/blob/2161f92aaf37e126c6906e7ae6202d196b72648c/ReactCommon/cxxreact/CxxModule.h#L66) is `size_t` which uses `%zu` as the format specifier. Newer versions of clang can warn or error on this. This change prevents that from happening.

## Test Plan

Local builds with clang 5.


kastiglione:
I added a test plan. The Travis CI failure looks unrelated to this change.

PR #15414, opened 2017-08-08
Fix format specifier used in RCTUIManager

kastiglione:
Here, `reactTag` is an `NSNumber *` which uses `%@` as the format specifier. Newer versions of clang can warn or error on this. This change prevents that from happening.

See also #15402

## Test Plan

Local builds with clang 5.


javache:
Thanks! FYI, you can just submit these internally too.

shergin:
@kastiglione Uh, really? I believe `%@` notation indicates here that we want to use `description` method of Objective-C object to generate string representation of the number, whereas `%zd` will simply print the pointer address to the object (which is actually useless information).

kastiglione:
@shergin sounds like you agree with this change from `%zd` (useless) to `%@`?

kastiglione:
@javache why did this fail?

javache:
Just some internal flakiness. It's been merged now.

shergin:
@kastiglione Yeah, I misread the diff... Sorry for the mess.

Issue #13578, opened 2017-04-19
FlatList does not accept (number) zero or empty string items


josvos:
When using plain numbers or strings as items (in the "data" list property), rendering FlatList results in an error when encountering number 0 or string ''.  The error is "No item for index X" where X is the index of the 0 or '' in the list.  Examples where the error is "No item for index 2" are data={[-2, -1, 0, 1, 2]} and data={['a', 'b', '']}.

Lizhooh:
I have encountered this situation
Now the solution is to turn Number to String

```js
<Text>{`${count}`}</Text>
```

hramos:
Dupe of #12905. Try the workaround above.

rightaboveit:
@Lizhooh's solution didn't work for me as I had a data source of empty URIs for Images, and received the 'No item for index X'

How I fixed it was for my images to just pass in a string 'blank' and check like so.
```
source={(imgSource != 'blank') ? {uri: imgSource} : undefined}
```

PR #15419, opened 2017-08-09
Allow data lists to include 0 or '' (falsey items)

tomasreimers:
Fixing https://github.com/facebook/react-native/issues/13578

(Write your motivation here.)

## Test Plan


ide:
`false` should be ok, isn't it null or undefined you're looking for?

tomasreimers:
To be quite honest, I'm not actually sure why this invariant exists. Was hoping for someone with more domain knowledge to chime in.

sahrens:
Yeah actually we can probably just remove the invariant completely.

PR #15437, opened 2017-08-09
[website] Fix button spacing and layout on small screens

roya3000:
Fix buttons issue in small screens

Before:

image

After:

image


TheSavior:
Thanks for contributing! 

I think I'd prefer to just stack the buttons if they don't fit horizontally instead of shrinking the text size. Shrinking the text size seems fragile. The buttons on the homepage currently stack once the screen gets small enough, so it would at least be consistent with that. 
screen shot 2017-08-16 at 3 41 08 pm

Could you make this update?

roya3000:
@TheSavior sure i can do that. I see buttons in home page are stuck, I think it would be better to give some space. And how can I fix version number in the header?

website/src/react-native/css/react-native.css
```diff
@@ -1013,10 +1013,24 @@ h2 {
...
+  content: " ";
+  display: table;
+}
+
+.docs-prevnext:after {
+  clear: both;
+}
+
+.docs-prevnext {
+  *zoom: 1;
```
TheSavior: It seems like this is only for really old versions of IE, do we need this?

roya3000:
Apply new styles.

image

image

website/src/react-native/css/react-native.css
```diff
@@ -1013,10 +1013,24 @@ h2 {
...
 .docs-prevnext {
-  min-width: 320px;
   max-width: 640px;
   margin: 40px auto;
-  padding-bottom: 20px; }
+  padding-bottom: 20px;
+}
+
+.docs-prevnext:before,
+.docs-prevnext:after {
```
TheSavior: From my small experimenting in the chrome console, it seems like your media query addition is the only part of the PR needed to fix the previous/next button of the docs. Why do we need this display: table stuff?

website/src/react-native/css/react-native.css
```diff
@@ -1013,10 +1013,24 @@ h2 {
   text-rendering: optimizelegibility; }
 
 .docs-prevnext {
-  min-width: 320px;
   max-width: 640px;
   margin: 40px auto;
-  padding-bottom: 20px; }
+  padding-bottom: 20px;
+}
```
TheSavior: From looking through the rest of this file it looks like we prefer having the ending curly on the previous line

roya3000:
@TheSavior, I removed clearfix (clearfix don't affect buttons issue)

website/src/react-native/css/react-native.css
```diff
@@ -2030,6 +2030,14 @@ article li {
 .btn:hover {
   text-decoration: none !important; }
 
+@media screen and (max-width: 373px) {
+  .btn {
```
TheSavior: Should this be applying to *all* buttons on small screens or should this rule be `.btn.docs-prev, .btn.docs-next`? I haven't checked this PR locally to verify buttons across the site look fine on small screens.

Also, you add margin-bottom in this rule for the stacked buttons, do we need your addition of margin-bottom (of a different amount) for `.buttons-unit .button` above?

roya3000: Should I be more specific and add styles to only doc section? (I checked other pages  and they look good). I don't understand that wrong with margin-bottom in   `.buttons-unit .button`. Should I add `margin-bottom` only for the first button?

TheSavior: In terms of specificity, this *might* be fine, I just have less confidence in this css change than I would if it was limited to just the buttons we are trying to fix rather than *all* buttons on small screens.

My question about the bottom margin is that if we are just trying to fix these for small screens, do we really need to be modifying things for large screens too?

roya3000:
@TheSavior added more specific targets for styles. I have problem with merge and git create new commit. How can I solve this issue or that's ok?

website/src/react-native/css/react-native.css
roya3000: added more specific targets for styles.

TheSavior:
I have updated the specificity of the selectors from how you had them, but tested the changes locally to ensure they still solved what you were aiming to solve. Here is a new set of screenshots with the before and after this PR.

Before this PR:
screen shot 2017-08-19 at 2 07 14 pm
screen shot 2017-08-19 at 2 07 07 pm

After:
screen shot 2017-08-19 at 2 06 18 pm
screen shot 2017-08-19 at 2 06 30 pm

TheSavior:
Thanks for contributing!

Issue #10470, opened 2016-10-20
[Android] Fail to render a large image


woniesong92:
### Issue Description

When I try to render a big image (5312 x 2988), the image container just shows an empty box. If the image is reasonably sized (~1024x1024) or comes directly from device instead from a remote url, it renders fine. Not sure if it's a memory or networking issue. There's another open issue(https://github.com/facebook/react-native/issues/7408) that seems to discuss a similar topic.
### Steps to Reproduce / Code Snippets

Render a large image in a ListView. In a dev mode, the app will crash in a few seconds. In a production mode, the app just shows a white background in place of an image.

Inside a `ListView` I render an image like below:

```
<Image
  source={{uri: imageURI}}
  style={styles.canvas}
  resizeMode="cover"
/> : null

StyleSheet.create({
  canvas: {
    width: null,
    height: 300,
  },
})

<ListView
  dataSource={this.state.dataSource}
  renderRow={this.onRenderRow}
  enableEmptySections={true}
  initialListSize={5}
  pageSize={10}
/>
```
#### Expected Results

Image is rendered successfully.
### Additional Information
- React Native version: 0.35.0
- Platform(s) (iOS, Android, or both?): Android 6
- Operating System (macOS, Linux, or Windows?): MacOS
### Update

Unfortunately, trying to resize a large picture using `ImageEditor` also fails because of an `OutOfMemory` exception. I used the photos that are captured by device's built-in camera.

```
10-20 16:38:47.511 19110 19154 E AndroidRuntime: FATAL EXCEPTION: AsyncTask #4
10-20 16:38:47.511 19110 19154 E AndroidRuntime: Process: com.kpopio, PID: 19110
10-20 16:38:47.511 19110 19154 E AndroidRuntime: java.lang.RuntimeException: An error occurred while executing doInBackground()
10-20 16:38:47.511 19110 19154 E AndroidRuntime:    at android.os.AsyncTask$3.done(AsyncTask.java:309)
10-20 16:38:47.511 19110 19154 E AndroidRuntime:    at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
10-20 16:38:47.511 19110 19154 E AndroidRuntime:    at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
10-20 16:38:47.511 19110 19154 E AndroidRuntime:    at java.util.concurrent.FutureTask.run(FutureTask.java:242)
[...]

PR #15439, opened 2017-08-09
Use BitmapRegionDecoder to efficiently crop images on Android

faifai21:
The Android ImageEditingManager is inefficient and slow when cropping images. It loads the full resolution image into memory and then crops it. This leads to slow performance and occasional OutOfMemory Exceptions. 
[BitmapRegionDecoder](https://developer.android.com/reference/android/graphics/BitmapRegionDecoder.html) can be used to crop without needing to load the full resolution image into memory. Using it is much more efficient and much faster.

Relevant issue: https://github.com/facebook/react-native/issues/10470

## Test Plan

Attempt to crop a very large image (2000x2000) on Android. With this change, the crop should happen almost instantly. On the master branch, it should take 2-3 seconds (and might run out of memory).

Please let me know if there's anything else I can provide.


rahamin1:
@faifai21 This is probably a stupid question, but fear is stronger than shame :)
Does this mean that I can use BitmapRegionDecoder in React Native? Is there an example for how to use it?

gladiko:
rahamin1@, look ta this code [link](https://github.com/facebook/react-native/commit/64ffe4516ec755505339a0a107efa83c3d4c9771) . So this work's automatically for you when you use <Image> in your code. But now I am testing code with many images from api with size >2500px and problems(black empty block instead image, app crashing) occurs same like before that commit with BitmapRegionDecoder. I think image resizing is the only solution if you want to get rid of memory leak with fetching large images.

rahamin1:
Thanks @noth8 I cannot resize since I am displaying previews of web articles in a ListView, and the images are taken from the articles. Hopefully @faifai21 has other insights.

PR #15452, opened 2017-08-10
ClassCastException fix: getText() returns CharSequence not Spanned.

dryganets:
ClassCastException fix: getText() returns CharSequence not Spanned.
From the other hand, EditTexts getText method returns Editable which extends Spanned.
This commit fixes two similar bugs one in flat TextView and another in standard TextView.
Also, it removes redundant checks in the ReactEditText.

##Test Plan
Application without this change sporadically crashes with the following stack trace:
```
java.lang.ClassCastException: java.lang.String cannot be cast to android.text.Spanned
	at com.facebook.react.views.text.ReactTextView.reactTagForTouch(ReactTextView.java:195)
	at com.facebook.react.uimanager.TouchTargetHelper.getTouchTargetForView(TouchTargetHelper.java:269)
	at com.facebook.react.uimanager.TouchTargetHelper.findTargetTagAndCoordinatesForTouch$58866680(TouchTargetHelper.java:101)
	at com.facebook.react.uimanager.JSTouchDispatcher.handleTouchEvent(JSTouchDispatcher.java:77)
	at com.facebook.react.ReactRootView.dispatchJSTouchEvent(ReactRootView.java:151)
	at com.facebook.react.ReactRootView.onInterceptTouchEvent(ReactRootView.java:127)
	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2110)

```


ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextView.java
```diff
@@ -79,7 +79,7 @@ public void setText(ReactTextUpdate update) {
 
   @Override
   public int reactTagForTouch(float touchX, float touchY) {
-    Spanned text = (Spanned) getText();
```
dryganets: The crash we encountered happened here.

ReactAndroid/src/main/java/com/facebook/react/flat/TextNodeRegion.java
```diff
@@ -70,11 +70,14 @@ boolean matchesTag(int tag) {
     }
 
     if (mLayout != null) {
-      Spanned text = (Spanned) mLayout.getText();
-      RCTRawText[] spans = text.getSpans(0, text.length(), RCTRawText.class);
-      for (RCTRawText span : spans) {
-        if (span.getReactTag() == tag) {
-          return true;
+      CharSequence text = mLayout.getText();
+      if (text instanceof Spanned) {
```
dryganets: I know if layout exists it should be spanned.
But shouldn't hurt anyway.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -488,8 +488,8 @@ private void updateImeOptions() {
 
   @Override
   protected boolean verifyDrawable(Drawable drawable) {
-    if (mContainsImages && getText() instanceof Spanned) {
-      Spanned text = (Spanned) getText();
+    if (mContainsImages) {
```
dryganets: As for this class, you could see that checks are not required as TextEdit always returns Editable which is an instance of Spanned.

maxkomarychev:
We're getting this crash too. Mostly on Lenovo tablets.

PR #15460, opened 2017-08-11
Make `react-native link` play nicely with CocoaPods-based iOS projects.

mironiasty:
The core React Native codebase already has full support for CocoaPods. However, `react-native link` doesn’t play nicely with CocoaPods, so installing third-party libs from the RN ecosystem is really hard.

This change will allow to link projects that contains its own `.podspec` file to CocoaPods-based projects. In case `link` detect `Podfile` in `iOS` directory, it will look for related `.podspec` file in linked project directory, and add it to `Podfile`. If `Podfile` and `.podspec` files are not present, it will fall back to previous implementation.

**Test Plan**
1. Build a React Native project where the iOS part uses CocoaPods to manage its dependencies. The most common scenario here is to have React Native be a Pod dependency, among others.
2. Install a RN-related library, that contains `.podspec` file, with `react-native link` (as an example it could be: [react-native-maps](https://github.com/airbnb/react-native-maps)
3. Building the resulting iOS workspace should succeed (and there should be new entry in `Podfile`)


Kureev:
@mironiasty once again, great job 🙌  Really enjoy reviewing high-quality PRs

mironiasty:
@Kureev Thanks :)

Also another way to test it, is to create Expo project, and then use `exp detach` on it. This should create iOS project with `Podfile`. And such project could be updated with modified implementation of `react-native link`.

grabbou:
This looks really solid. Let me get back to you on that tomorrow latest. Pretty sure will be flawless :)

local-cli/link/pods/unregisterNativeModule.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const fs = require('fs');
+const removePodEntry = require('./removePodEntry');
+
+/**
+ * Unregister native module IOS with CocoaPods
+ */
+module.exports = function unregisterNativeModule(dependencyConfig, iOSProject) {
+
```
grabbou: I believe empty line here and on line 14 is not intended? (nitpick)

local-cli/link/pods/unregisterNativeModule.js
```diff
@@ -0,0 +1,15 @@
...
+  const podContent = fs.readFileSync(iOSProject.podfile, 'utf8');
+  const removed = removePodEntry(podContent, dependencyConfig.podspec);
```
grabbou: After removing, shall we also call `pod install` to sync the project? What happens right now after you unlink?

local-cli/link/pods/findLineToAddPod.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = function findLineToAddPod(podLines, firstTargetLine){
+
+  // match line with new targe: target 'project_name' do (most likely target inside podfile main target)
+  const nextTarget = /target (\'|\")\w+(\'|\") do/g;
+  // match line that has only 'end' (if we don't catch new target or function, this would mean this is end of current target)
+  const endOfCurrentTarget = /^\s*end\s*$/g;
+  // match function defeinition, like: post_install do |installer| (some Podfiles have function defined inside main target
```
grabbou: nit: typo in `definition`

local-cli/link/pods/findLineToAddPod.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = function findLineToAddPod(podLines, firstTargetLine){
+
+  // match line with new targe: target 'project_name' do (most likely target inside podfile main target)
```
grabbou: nit: typo in `target`

local-cli/link/pods/findLineToAddPod.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = function findLineToAddPod(podLines, firstTargetLine){
+
```
grabbou: Same here: no need to include an empty line :)

local-cli/link/pods/findPodTargetLine.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = function findPodTargetLine(podLines, projectName) {
```
grabbou: Shouldn't this return an array? What if user has few targets in a single `pod` file, one e.g. for iOS and second for e.g. tvOS? Same issue is happening right now in standard `link` - we just link to first target rather than to all that include React. I am currently rewriting this approach.

Here, I believe we should find all targets that have `pod React` inside (they consume React Native via CocoaPods) and assume that if they already use React Native inside that target, they want to use linked module as well.

andrejunges: @grabbou I can't agree more with you regarding our `link` issue where we just link the first target. Besides being able to have a target for different platforms (tvOS..) I think there are more use cases that'd use more than one target - a white label product (app) is a good example. My team face this issue every time we add a new library with native deps.. Let me know if I can help you in any way.

local-cli/link/pods/unregisterNativeModule.js
mironiasty: I don't call `pod install` anywhere. It is up to user to call it later on (it could be also run automatically in build process/Xcode). Right now link/unlink processes only modify configuration files (and with this change only modify `Podfile` file).

local-cli/link/pods/findPodTargetLine.js
mironiasty: This might be tricky to do. Current implementation is kind of naive, and works only for simple cases (that was assumptions from the beginning). So, I'm just looking for target that has same name as project, and then look for good place to add new line.
But to make it work in most cases, I would have to write some simple Ruby parser to look for all non-nested targets (possible only ones, that already had React as dependency?), and then add new line to them. And I'm sure, that there always will be some project that somehow abuse Podfile syntax, and my approach would fail.

Anyway, if you could send me some real life Podfiles, I might try to check few ways to make it better. Also if you have some ideas how to improve it, let me know.

andrejunges: hmm, that's a good point, I don't think we'll be able to make it work in all podfiles.. (my comments above were more related to not cocoa-pod links..)
I have an example where we customized the Podfile so it'd be possible to add new dep for all targets at once.

```
[
  "Target1",
  "Target2",
  "Target3",
  "Target4",
  "Target5",
  "Target6",
].each do |target_name|
  target target_name do
    pod "IQKeyboardManager"
    pod "Firebase/Messaging"
    pod "Firebase/Crash"
    pod "Firebase/Core"
    pod "FLAnimatedImage"
  end
end
```

mironiasty: Ok, so my implementation would not work for such file - I assumed that target name would be same as project name. Which might be true in this case, but regexp's are not smart enough to find it out :(

grabbou: You can check what Mobile Azure team has done - when you attemp to link one of their native libraries, they ask you to include special comment in Podfile. It's a marker after which pods are inserted. That comment of course can be placed multiple times so that they link each target. 

Usecase proposed here is a bit too complex for us, but I guess checking for `pod React` and inserting after it is a good starting point. Just like we insert projects to gradle files.

CC: @Kureev

Kureev: I was also thinking about rewriting the way we inject dependencies to gradle files (it is quite corky now). Although, this implies writing a parser for gradle files which is a bit too much for me. If I'd work on this full time, I'd say it'll take me a couple of weeks to make it properly, so I doubt if it will be worth it. At the same moment of time I completely agree about multiple targets. 

Are you aware of any projects that are capable of constructing / traversing an AST for a Podfile?

mironiasty: @Kureev  I did some research when I started work on this. But I couldn't find any library that could parse Podfile (or any Ruby file) in JS.
For me, the whole idea of  `react-native link` is to update project configuration, if it is quite simple. If user starts modifying configuration files by hand, there will be always a case when auto-updating will fail. Or code that update configuration would be quite complex.

@grabbou I might add search for these special comment(s) in file, then adding new pod under it. And if there is no such comment, I might use current approach (as the original idea, is to make it work with ejected expo projects).

mironiasty:
Based on your comments, I've updated code to support new marker in Podfile. If there will be comment(s) `# Add new pods below this line` somewhere in Podfile, new pod will be added below such line(s). If there is no such line, code will follow old path - guessing where pod could be added. I think it is easier way to support multiple targets or case described by @andrejunges . And if user has just simple Podfile, generated from `expo detach`, this should still work for him.

I've also added variable indentations - code is trying to guess how many spaces should be inserted before `pod ...`.

mironiasty:
@grabbou Can you take a look on improved version of this feature?

grabbou:
Hey, looks good - I am planning to land it today. Only thing is that I removed `mockFs` dependency in the meantime on master in this commit: https://github.com/facebook/react-native/commit/1f498010e8a10ab8b7f240988cbf892ef71c3154

Would you have time to perform such update? If not, I am happy to do it.

Anyway, this should've been within 0.50.0-rc.0 and I am going to cherry-pick it as the delay has been because of me :)

mironiasty:
I will try to find some time to make this update later today or tomorrow.

grabbou:
Thanks, I will be keeping an eye on it.

mironiasty:
@grabbou I've updated tests to remove `mock-fs` dependency.

local-cli/link/pods/addPodEntry.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = function addPodEntry(podLines, linesToAddEntry, podName, nodePath) {
+	const newEntry = `pod '${podName}', :path => '../node_modules/${nodePath}'\n`;
+
+	if (!linesToAddEntry) {
+		return;
```
grabbou: This file has wrong indentation. Try running `prettier` for all the files in your PR (not sure if you have it integrated already)

mironiasty:
Sorry for these tabs, I hope it is fixed now. (I've been working on other project with tabs, and priettier used that config on react-native files)

grabbou:
@facebook-github-bot shipit

chrisbianca:
One of the core maintainers of https://github.com/invertase/react-native-firebase here.

We've spent the last year battling with RN's lack of integration with Cocoapods and the pain this causes our users.  Having given up on the Cocoapods approach and switching to usage of `react-native link` as it used to be, we are in the position where we have a `.podspec` for legacy support, but do not wish this to be used for new projects due to the complications this brings with needing to also add the appropriate React pod and subspecs.

Since this change, our `.podspec` is picked up automatically, but it seems this PR provides no way for us to opt out of that decision and override the presence of the podspec.

```
return {
    sourceDir: path.dirname(projectPath),
    folder: folder,
    pbxprojPath: path.join(projectPath, 'project.pbxproj'),
    podfile: findPodfilePath(projectPath),
    podspec: findPodspecName(folder),
    projectPath: projectPath,
    projectName: path.basename(projectPath),
    libraryFolder: userConfig.libraryFolder || 'Libraries',
    sharedLibraries: mapSharedLibaries(userConfig.sharedLibraries || []),
    plist: userConfig.plist || [],
  };
```

As such, this has fundamentally broken our on-boarding flow for new users and leaves us with only two options:

1) To remove the .podspec completely which removes backwards compatibility for those devs who have been using the .podspec previously
2) To switch back to the .podspec approach and introduce the added step of having the users need to set up React and subspecs as pod - something which we know causes issues based on our past experiences

Neither of these are particularly viable alternatives and both involve compromises.

As somebody that has tried to submit PRs to improve the build process customisability in the past (on the Android side), only for the PR to be completely ignored, it's incredibly frustrating to now be affected by a change that hasn't taken into consideration what this might mean for some libraries.

mironiasty:
One thing that might be worth to mention here: this change would only work for users that have Pods configured for their iOS project (so users without `Podfile` should not be affected). But from `react-native-firebase` installation documentation, I see that configuring pods is recommended step for users.

I don't know what is exact problem with linking your project using CocoaPods. If problem is related to missing react entries in Podfile, maybe you can recommend creating project with `create-react-native-app` and detaching project from Expo (this way, user should have fully working Podfile).

As for adding option to opt-out from linking podspecs: I can think of two ways of doing it:
- Adding new parameter to `link` (`react-native link react-native-firebase --ignore-pods`)
- Checking for predefined comment in project .podspec file (like `#REACT_NATIVE_IGNORE_FOR_LINKING`)

I'm not sure which one would be better. Also adding any of them, would again increase complexity of linking process.

chrisbianca:
@mironiasty thanks for your response.

Our users need to configure pods in order to use the Firebase dependencies, as this is the approach that Google recommend and we don't want to deviate from it.

We also categorically don't want to recommend people creating projects using Expo as a way of getting started.  This has also caused us numerous problems down the line and is not a viable alternative given the large dependency that `react-native-firebase` has on native code.

Regarding your options above:

1) This would only work when linking just this library, anybody running `react-native link` without the library would encounter problems
2) This would work, though feels somewhat hacky

A much better option (and one that keeps in sync with the rest of the rnpm settings) would be to allow this to be overridden as part of the package.json.  For example, setting an `ignorePodSpec` as so:

```
"rnpm": {
    "ios" {
      "ignorePodSpec": true,
    }
  },
```

PR #15466, opened 2017-08-11
update flowconfig

chirag04:
Travis e2e is failing right now because if mismatched flow-bin version in flow config.

## Test Plan

Travis e2e test should not fail on mismatched flow version.


chirag04:
damn @javache beats me to it. https://github.com/facebook/react-native/pull/15466

PR #15472, opened 2017-08-12
Corrected types of `page_info` return

blargity:
`page_info` has `start_cursor` and `end_cursor` which are currently marked in the docs as `boolean` type, but they're actually strings.

I was confused by the current documentation, as it was requesting me to pass the previously received `end_cursor` into the next fetch but it was declared as a boolean. I checked in the debugger and it's just mislabeled in the docs, so this PR fixes the docs.

## Test Plan

This is a change to documentation only.


chirag04:
@blargity sorry i messed up the rebase. Can you rebase real quick and i will land it?

blargity:
@chirag04, I wasn't able to undo the other changes except by manually going through the file, and creating a new PR from a freshly synced fork made more sense to me, so I've opened #16830. I'll close this one.

PR #15475, opened 2017-08-12
Generalize ActionSheetManager items URL 

fxfactorial:
Not ready but want to start conversation.  (i.e. not sure if to change key name to `urls` from `url`) 

I want to be able to share more than 1 image at a time, the current code only lets one use one url, but the API allows for much more. 

Some places already make this an issue: 

https://github.com/EstebanFuentealba/react-native-share/issues/85

And I also need to be able to email/send on iMessage, etc with multiple URL resources.

### DIFF
--- a/Libraries/ActionSheetIOS/RCTActionSheetManager.m
+++ b/Libraries/ActionSheetIOS/RCTActionSheetManager.m
@@ -133,11 +133,17 @@ RCT_EXPORT_METHOD(showShareActionSheetWithOptions:(NSDictionary *)options
   if (message) {
     [items addObject:message];
   }
-  NSURL *URL = [RCTConvert NSURL:options[@"url"]];
-  if (URL) {
-    if ([URL.scheme.lowercaseString isEqualToString:@"data"]) {
+  NSArray<NSURL *> *URLS = [RCTConvert NSURLArray:options[@"url"]];
+
+  if (URLS.count == 0) {
+    RCTLogError(@"No `url` or `message` to share");
+    return;
+  }
+
+  for (NSURL *url in URLS) {
+    if ([url.scheme.lowercaseString isEqualToString:@"data"]) {
       NSError *error;
-      NSData *data = [NSData dataWithContentsOfURL:URL
+      NSData *data = [NSData dataWithContentsOfURL:url
                                            options:(NSDataReadingOptions)0
                                              error:&error];
       if (!data) {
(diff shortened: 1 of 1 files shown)


fxfactorial:
And I've just tested this locally in my project and its works correctly.

fxfactorial:
@hramos Ping, can someone take a look at this please?

hramos:
Sorry @fxfactorial, we do have quite the backlog of PRs. Can you think of anyone that has worked on this file before that may be able to review the changes?

fxfactorial:
@hramos Sure, I understand that RN is a big project, and probably people are overworked, thank you for looking at all. 

I think this would be like a couple of moments of @nicklockwood 's time to glance over.

fxfactorial:
@hramos another friendly ping

hramos:
@fxfactorial hey, it looks like this has introduced an issue. Here's a simple repro:

```
require('ActionSheetIOS').showShareActionSheetWithOptions(
  {message: 'TEST!'},
  // or {url: 'http://google.com'},
  (error) => {},
  () => {},
);
```

Results in a red screen.

Libraries/ActionSheetIOS/RCTActionSheetManager.m
```diff
@@ -133,11 +133,17 @@ - (CGRect)sourceRectInView:(UIView *)sourceView
   if (message) {
     [items addObject:message];
   }
-  NSURL *URL = [RCTConvert NSURL:options[@"url"]];
-  if (URL) {
-    if ([URL.scheme.lowercaseString isEqualToString:@"data"]) {
+  NSArray<NSURL *> *URLS = [RCTConvert NSURLArray:options[@"url"]];
+
+  if (URLS.count == 0) {
```
olegbl: This means that passing in "message" will always crash.

hramos:
We're going to back out this breaking change.

Libraries/ActionSheetIOS/RCTActionSheetManager.m
```diff
@@ -133,11 +133,17 @@ - (CGRect)sourceRectInView:(UIView *)sourceView
   if (message) {
     [items addObject:message];
   }
-  NSURL *URL = [RCTConvert NSURL:options[@"url"]];
-  if (URL) {
-    if ([URL.scheme.lowercaseString isEqualToString:@"data"]) {
+  NSArray<NSURL *> *URLS = [RCTConvert NSURLArray:options[@"url"]];
```
olegbl: This is a breaking API change that introduces crashes into all existing apps using it. Highly suggest introducing a "urls" argument instead.

Issue #15253, opened 2017-07-28
Expose notification handler for new photo


fxfactorial:
I want my application to be notified when there is a new photo added to the CameraRoll. I don't see a way to get that done currently, besides polling. iOS provides this ability with `ALAssetsLibraryChangedNotification`, but I don't see a single hit for that in the RN codebase, which verifies to me that this is currently not exposed. 

I think I can do this after I finish some contract work, assuming that it is indeed the right way....but as I understand `ALAssetsLibrary` is deprecated anyway and we ought to be using `PHPPhotoLibrary`, so I don't want to implement something using deprecated frameworks and refactoring RN to use PHPPhotoLibrary is a much larger task than just exposing `ALAssetsLibraryChangedNotification`. 

Thoughts? 

(I could also take on refactoring to make RN use `PHPPhotoLibrary`, a rather large task)

I assume that it could be maybe another event exposed via DeviceEventEmitter?

hramos:
This issue looks like a question that would be best asked on [StackOverflow](http://stackoverflow.com/questions/tagged/react-native).

StackOverflow is amazing for Q&A: it has a reputation system, voting, the ability to mark a question as answered. Because of the reputation system it is likely the community will see and answer your question there. This also helps us use the GitHub bug tracker for bugs only.

Will close this as this is really a question that should be asked on StackOverflow.

fxfactorial:
@hramos Are you serious? Did you even read this? This has nothing to do with StackOverflow. 
1) If ReactNative EXPOSES NATIVE APIS
2) If it is worthwhile to refactor the current code in RN to stop using deprecated APIs.

hramos:
@fxfactorial if you believe the issue was closed prematurely, please read the [issue template](https://github.com/facebook/react-native/blob/master/.github/ISSUE_TEMPLATE.md). I also highly advise reading through the [Code of Conduct](https://code.facebook.com/codeofconduct), as your comment is a bit more aggressive than we'd like. Thanks for your understanding.

fxfactorial:
The issue template has nothing to do with what I'm talking about. I am talking about exposing APIs that aren't currently exposed.

hramos:
It does not look like it's exposed. I'm re-opening the issue as it looks like this would fall under "Proposing a change" per the contributor guidelines at http://facebook.github.io/react-native/docs/contributing.html#proposing-a-change. We need to come up with a better way of tagging such issues, as in the future we'll auto-close issues that don't follow the template.

[...]

PR #15481, opened 2017-08-13
Implement deletion of asset-library files

fxfactorial:
WIP: Starting point insofar as I'm not sure if there should be an Android equivalent. 

My application needs me to delete photos from the photos in iOS. 

I have tested this in my application and it works, have a screenshot from iOS asking as well and it does successfully delete the photos from the global Photos app. 

Related to https://github.com/facebook/react-native/issues/15253, it kind of continues using the deprecated assets-library framework, but that does need (and will be an bigger and bigger issue coming up) of assets-library URLs being used. 

I also assume RN prefers error handling at the JS level? Are the URLs starting with `asset-library`, etc.

### DIFF
--- a/Libraries/CameraRoll/CameraRoll.js
+++ b/Libraries/CameraRoll/CameraRoll.js
@@ -135,6 +135,10 @@ class CameraRoll {
     return this.saveToCameraRoll(tag, 'photo');
   }
 
+  static deletePhotos(photos: Array<string>) {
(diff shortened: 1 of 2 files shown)


fxfactorial:
@hramos Ping, can someone take a look at this please?

fxfactorial:
@hramos another friendly ping

PR #15485, opened 2017-08-14
Split up AnimatedImplementation.js

janicduplessis:
AnimatedImplementation.js is getting pretty hard to navigate and reason about so I split it up into different modules. Also took the opportunity to run prettier on that code and do some minor const/let refactorings. This doesn't change any logic, mostly just moves code around and add proper imports / exports.

This opens the door for further cleanup and flow type improvements but want to keep this already big PR as small as possible.

Discussion points:
- Should I use haste for this? Animated is pretty much a standalone module, it still uses a few haste imports but for new modules I used commonjs imports to avoid polluting the haste global namespace too much. The new modules are all internal to Animated and should not be imported externally.
- We're using `requestAnimationFrame` from fbjs instead of the one available globally in RN and browsers is there a reason for that?
- Should we even support web in this implementation? There is a standalone repo that exist for Animated web. Is this implementation of Animated web used internally at facebook?
- Probably still related to web, we used some weird Set polyfill is that still needed?

Notes:
There is a small regression for docs where the type of some classes that are exported like AnimatedValue show up as CallExpression instead if Class.

screen shot 2017-08-14 at 3 19 18 am

**Test plan**
Tested that all Animated related examples still work in RNTester on iOS and Android.
Tested that the doc is still working
Ran unit tests

### DIFF
--- a/Libraries/Animated/src/AnimatedEvent.js
+++ b/Libraries/Animated/src/AnimatedEvent.js
@@ -1,0 +1,195 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ * @format
+ */
+'use strict';
+
+const AnimatedValue = require('./nodes/AnimatedValue');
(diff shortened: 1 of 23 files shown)


sahrens:
Are you planning on doing more cleanup on top of this? e.g. more flow types, const/let, fix docs regession?

janicduplessis:
@sahrens Yea, I wanted to keep this diff minimal to make further changes easy to review.

janicduplessis:
@sahrens Rebased

PR #15490, opened 2017-08-14
Fix flowtype errors for PushNotificationIOS

radko93:
Flow syntax is wrong, instead of using `Array` author used syntax for [tuples](https://flow.org/en/docs/types/tuples/). Tested with flow 0.49.1.
## Test Plan
1. Import `removeDeliveredNotifications` in your code and pass an array with more than one element.
2. Test with flow.


ide:
@facebook-github-bot shipit

Issue #15492, opened 2017-08-15
Local cached images loaded failed occasionally


CodingForMoney:
### Is this a bug report?

Yes.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

YES

### Environment

1. `react-native -v`:0.47.1
2. `node -v`:v6.2.0
3. `npm -v`:3.9.0

Then, specify:

- Target Platform:
iOS

- Development Operating System:
macOS

- Build tools:
Xcode

### Steps to Reproduce

(Write your steps here:)

1. Download [the sample code](https://github.com/CodingForMoney/ReactNative-ImageLoadBugSample.git)
2. npm install,npm start , Run sample code
3. `cmd + R`, try reload the app for several times(Not Appearing-Certain).

### Expected Behavior

All images loaded successful.

![](http://resource.luoxianming.cn/rnimagenormal.png)

### Actual Behavior

Some images failed to load.

![](http://resource.luoxianming.cn/rnimgefailed.png)

### Reproducible Demo

https://github.com/CodingForMoney/ReactNative-ImageLoadBugSample.git

karegaCOPART:
I am facing this issue in multiple versions of RN.

CodingForMoney:
PR  #15493

CodingForMoney:
@karegaCOPART 

hi,guy . If you can't modify the react-native source , you can set this property:

`rctRootView.bridge.imageLoader.maxConcurrentLoadingTasks = 100;
`

Increase the maxTasks setting would make more local image loaded successfully .

PR #15493, opened 2017-08-15
Fix local cached file loaded error

CodingForMoney:
#15492 

When using `RCTURLRequestHandler` to load images from local cached files, unexpected loading error may happened. 
When using local cached image, we will use `dispatch_async` for `RCTURLRequestDelegate` callbacks,like this:

```
- (id)sendRequest:(NSURLRequest *)request
     withDelegate:(id<RCTURLRequestDelegate>)delegate {
  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
    NSHTTPURLResponse *response = [[NSHTTPURLResponse alloc] initWithURL:request.URL
                                                             statusCode:200
                                                            HTTPVersion:@"HTTP/1.1"
                                                           headerFields:nil];
    [delegate URLRequest:request didReceiveResponse:response];
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
      cachedData = [NSData dataWithContentsOfFile:[[NSBundle mainBundle] pathForResource:@"react" ofType:@"png"]];
    });
    [delegate URLRequest:request didReceiveData:cachedData];
    [delegate URLRequest:request didCompleteWithError:nil];
  });
  return request;
}

```

But sometimes the `dispatch_async` function was excuted much faster than current thread. Then the `URLRequest: didCompleteWithError` callback  will be called before `_status = RCTNetworkTaskInProgress` excution in `[RCTNetworkTask start]`. So this `RCTNetworkTask` will be RCTNetworkTaskInProgress forever. After this kind  of error happened more than the `maxConcurrentLoadingTasks` of `RCTImageLoader` , no more `RCTNetworkTask` will start in `dequeueTasks`. Then the last images will never be loaded.

So I just add a simple judge to avoid this problem.

### DIFF
--- a/Libraries/Network/RCTNetworkTask.m
+++ b/Libraries/Network/RCTNetworkTask.m
@@ -75,7 +75,9 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
     id token = [_handler sendRequest:_request withDelegate:self];
     if ([self validateRequestToken:token]) {
       _selfReference = self;
-      _status = RCTNetworkTaskInProgress;
+      if (_status == RCTNetworkTaskPending) {
+        _status = RCTNetworkTaskInProgress;
+      }
     }
   }
 }


CodingForMoney:
somebody help ?

PR #15502, opened 2017-08-15
Use offset adjusted value in ModulusAnimatedNode

rozele:
Modulus animation computation should use the value adjusted for offset, not the raw value.

See the JS implementation here:
https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/AnimatedImplementation.js#L1338

cc @ryangomba


janicduplessis:
@shergin Could you ship this one too

PR #15503, opened 2017-08-16
Switch from Request to phin

Ethanent:
Request is a super large library with **way too many dependencies** for what react-native uses it for. (Way more than needed.)

image
_^ Why so many dependencies, [request](https://www.npmjs.com/package/request)?_ 🤔

[phin](https://github.com/Ethanent/phin) is super lightweight and has **no dependencies**. It has all of the actually useful features built into request and reduces dependency count and install time for react-native noticeably. (Improved by over 15 seconds in my trial!)

## Test Plan

All related tests passed!

image

Also, the code change was super simple. Have a look for yourself! ;)

image


janicduplessis:
@Ethanent Thanks for the PR! Actually request seems to be used only for the website which has it's own package.json so I think we can just remove the request dependency and add it to website/package.json instead.

hramos:
What @janicduplessis said sounds good to me.

Ethanent:
@janicduplessis @hramos That makes sense! I'll submit a PR for the improvement.

PR #15505, opened 2017-08-16
Update documentation for PermissionsAndroid

conanak99:
Current documentation: [https://facebook.github.io/react-native/docs/permissionsandroid.html](https://facebook.github.io/react-native/docs/permissionsandroid.html)

PermissionsAndroid must be imported for the example to work, otherwise the error "Reference Error: can't find variable PermissionsAndroid" will be shown.
https://stackoverflow.com/questions/42434124/why-permissionandroid-doesnt-work

## Test Plan

None - Just documentation changes.


TheSavior:
Thanks for contributing!

PR #12502, opened 2017-02-21
Use contentOffset to set the starting scroll offset in ReactHorizontalScrollView

xqwzts:
Implements `contentOffset` for horizontal scrollviews in Android [already implemented for ios]. This will allow specifying the position the `ScrollView` should start at.

A common use case is having a horizontal `ListView` with pagingEnabled, and having it open to a specified page, without having to wait for it to load and then call `scrollTo`.

Also see #6849 

### Usage:
Add `contentOffset={{ x: someInitialXValue, y: someInitialYValue }}` to your `ScrollView` props.

### Example:

```javascript

import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Text,
  View,
  ListView,
  Dimensions,
} from 'react-native';


export default class offsettest extends Component {
  constructor() {
    super();
    const ds = new ListView.DataSource({ rowHasChanged: (r1, r2) => r1 !== r2 });
    const { width } = Dimensions.get('window');
    this.state = {
      dataSource: ds.cloneWithRows(['row 1', 'row 2', 'row 3', 'row 4', 'row 5']),
      initialPage: width * 3,
    };
  }

  render() {
    return (
      <View style={styles.container}>
        <Text style={styles.welcome}>
          Android Horizontal ListView ContentOffset Example
        </Text>
        <ListView
          dataSource={this.state.dataSource}
          renderRow={(rowData) => (
            <View style={{flex: 1, backgroundColor: 'red', width: Dimensions.get('window').width, justifyContent: 'center', alignItems: 'center'}}>
              <Text style={{color: 'white'}}>{rowData}</Text>
            </View>
          )}
          contentOffset={{ x: this.state.initialPage, y: 0 }}
          horizontal={true}
          pageSize={1}
          pagingEnabled={true}
        />
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  welcome: {
    fontSize: 16,
    textAlign: 'center',
    margin: 10,
  },
});

AppRegistry.registerComponent('offsettest', () => offsettest);
```

Issue #6849, opened 2016-04-06
[Android] ScrollView is missing initial scroll position for Android


bcalik:
`ScrollView` has a `contentOffset` prop for iOS, which sets the initial scroll offset.
But it lacks on Android, so you can not set the initial scroll position.

For anyone who can make this, these are the attributes in Android's native ScrollView.

```
android:scrollX     The initial horizontal scroll offset, in pixels. 
android:scrollY     The initial vertical scroll offset, in pixels. 
```

hufeng:
```js
render() {
  return (
    <ScrollView ref={(view) => this._scrollView = view}>
    </ScrollView>
  );
}

this._scrollView.getScrollResponder().scrollTo({x: 0, y: 0, animated: true})
```

bcalik:
@hufeng That will cause flickering after render, first you will see the 0,0 position, then it will scroll.

Also if you are using a ListView, then you will have to render all the data until the scroll position you want, then you can scroll there..
But in iOS, you can just set `initialListSize={1}` and it can just start rendering where-ever you want by setting `contentOffset`.

mangogogos:
https://productpains.com/post/react-native/content-offsetinset-for-scrollview

I started a product pains post for this a while back, but it hasn't gained any traction. While @hufeng's approach does the correct thing, it's very janky and even if you set animated: false, it will be noticeable to the end-user that the scrollview is jumping to its initial position. I've wanted to take a crack at a PR for this for a while but haven't had the bandwidth outside of work but for me this would be one of the most welcomed features for Android/iOS parity

madox2:
When I try to call `scrollTo` from `componentDidMount` method it does not scroll. I have to use workaround with `setTimeout` to make it work:

```javascript
componentDidMount() {
  setTimeout(() => this.refs._scrollView.scrollTo({ x: 100, y: 0 }) , 0);
}
```

satya164:
It'll be awesome if someone could send a pull request for it.

bcalik:
@madox2 Because the scroll position you want to scroll is not rendered yet, you have to wait for all rows to be rendered before you scroll to that position. So setTimeout is not a solution, which may not work.

iOS ListView does't have this issue, because it has a initial scroll position property, so it can start rendering from that position, not from zero.

mkonicek:
Hi there! This issue is being closed because it has been inactive for a while.

[...]

PR #15511, opened 2017-08-16
Add vertical ScrollView content offset on Android

jonaslu:
This is the (evil) twin commit to:
https://github.com/facebook/react-native/pull/12502

Fixes that contentOffset is not available on
vertical scroll-views on Android.

## Test Plan
* Create an Android component with a ScrollView. Add enough
  items to fill beyond the visible screen. Set the content
  offset to some value that makes the content scroll. Verify
  that it snaps to the y-axis as expected.
* Remove items so screen is no longer filled, keep content
  offset parameter. Scroll view still works but does not scroll.
* Set only one parameter to content-offset (x || y) -
  parameters are now ignored since they must be a pair.
* Remove contentOffset, scrollView still works as before.


shergin:
I am sorry, but we are thinking about deprecating `contentOffset`.
https://github.com/facebook/react-native/issues/6849#issuecomment-323551327

henrikra:
@jonaslu Looks good! Mind fixing conflicting files? :D

jonaslu:
Well, sure, but before I go ahead -do you know if it's going to be merged? The comments in #6849 seems omnious and not so conclusive..

henrikra:
I noticed them too @jonaslu But I think this is the best solution so far. Someone can refactor better solution if they now better. Your solution is better than no solution on Android :)

Jacse:
@shergin have you guys arrived at a decision? If no decision for a better implementation could be made can we get this merged so we'll at least have something working?

I don't think "the old way" should be deprecated before a new and better way is implemented.

fbartho:
This would be a really valuable feature. We need this for consistency between our behavior on both platforms.

cpojer:
I'll close this based on @shergin's comment that we may want to deprecate `contentOffset` altogether.

Issue #15513, opened 2017-08-16
[android / ios] bundling failed: "TransformError


esganzerla:
### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment
1. `react-native -v`:
react-native-cli: 2.0.0
react-native: n/a - not inside a React Native project directory
(I was inside the folder, version is: 0.47.1)
2. `node -v`: v7.10.0
3. `npm -v`: 4.2.0
4. `yarn --version`: 0.17.10

Then, specify:
- Target Platform: Android, iOS
- Development Operating System: macOS Sierra 10.12.6
- Build tools: Using Xcode 8.3.3 and Android Studio 2.3

### Steps to Reproduce
1. `react-native init rn_reproduce_transform_error`
2. `cd rn_reproduce_transform_error`
3. `react-native run-ios` or `react-native run-android`

### Expected Behavior
Should show the default "Welcome to React Native!" screen.

### Actual Behavior
![bundling failed: "TransformError: /Users/eduardo/Desktop/rn_reproduce_transform_error/index.ios.js: Unexpected token ) (While processing preset: \"/Users/eduardo/Desktop/rn_reproduce_transform_error/node_modules/babel-preset-react-native/index.js\")"](https://github.com/esganzerla/rn_reproduce_transform_error/blob/master/screenshots/app-error.png)

Same error occurs when I tried to update RN version to 0.48.0-rc.1
![SyntaxError: Unexpected token )](https://github.com/esganzerla/rn_reproduce_transform_error/blob/master/screenshots/upgrade-error.png)

### Reproducible Demo
https://github.com/esganzerla/rn_reproduce_transform_error

### Notes
- This happens both on android and iOS.
- I saw some people having the same problem on issue #15496.

tmartres:
Had the same issue and after restarting my computer the issue went away for me. You might have to just close and restart your terminal and simulator too.

PhillipWinder:
+1 

experiencing the same issue on `run-android` aswell as `run-ios`

amerej:
@tmartres Restarting don't work for me

esganzerla:
Restarting didn't help here too.

esganzerla:
I've just found a workaround:

`yarn remove babel-preset-react-native`
`yarn add babel-preset-react-native@2.1.0`

The hole is probably deeper, but that is works for me.

jarecsni:
[...]

PR #15516, opened 2017-08-16
Fix TransformError issue for babel-preset-react-native

ivanzotov:
I don't know why, but it fixes issue #15513


charpeni:
Trailing commas in function parameter lists is not supported.

https://github.com/facebook/react-native/blob/6ad7e8281b37ee2ce6425363c0b17420d056807a/babel-preset/configs/main.js#L40

Issue #15271, opened 2017-07-29
Checking for a version match between JS and native


ide:
Many of the issues people run into - with quite a wide variety of symptoms - are caused by a mismatch between the JS and native code. There isn't a single reason for this mismatch, ex: it could be caused by stale Xcode/Android build data, forgetting to rebuild the native app after upgrading RN in node_modules, Watchman/Metro caching bugs, bundle caching on the client, and so on. Some kind of simple version check might help people discover these bugs more quickly with a more accurate error message.

## Proposal

We add three new files: ReactNativeVersion.java, RCTVersion.h, and ReactNativeVersion.js. These files are marked as `@generated` and are written only by the release scripts, which already modify package.json, the Podspec, and a Gradle file.

```java
// ReactNativeVersion.java
// @generated by scripts/bump-oss-version.js
public class ReactNativeVersion {
  public static final String VERSION = "0.47.0-rc.1";
  public static final int MAJOR = 0;
  public static final int MINOR = 47;
  public static final int PATCH = 0;
  public static final int PRERELEASE = "rc.1";
}
```

```objc
// RCTVersion.h
// @generated by scripts/bump-oss-version.js
#define REACT_NATIVE_VERSION @"0.47.0-rc.1"
#define REACT_NATIVE_VERSION_MAJOR 0
#define REACT_NATIVE_VERSION_MINOR 47
#define REACT_NATIVE_VERSION_PATCH 0
#define REACT_NATIVE_VERSION_PRERELEASE @"rc.1"
```

```js
// ReactNativeVersion.js
// @generated by scripts/bump-oss-version.js
exports.version = '0.47.0-rc.1';
exports.major = 0;
exports.minor = 47;
exports.patch = 0;
exports.prerelease = 'rc.1'
```

These files are checked into the repo for simplicity. In addition to these autogenerated files, there would be small native modules that export these versions from native to JS (using constantsToExport, it should be fast since there's not much real work being done). JS would then compare ReactNativeVersion.version to NativeModules.Version.version and call console.error if they don't match (for some definition of match -- maybe it's OK if 0.46.1 JS runs on 0.46.0 native code, don't want this warning to be annoying).

dlowder-salesforce:
I like this idea a lot!  Maybe version should be an object instead of a string... what about

```javascript
// ReactNativeVersion.js
// @generated by scripts/bump-oss-version.js
exports.version = {
  'major' : '0.47',
  'minor' : '0',
  'build' : ''
```

[...]

PR #15518, opened 2017-08-16
Validate that JS and Native code versions match for RN releases

janicduplessis:
Basic implementation of the proposal in #15271

Note that this should not affect facebook internally since they are not using OSS releases.

Points to consider:
- How strict should the version match be, right now I just match exact versions.
- Wasn't able to use haste for ReactNativeVersion because I was getting duplicate module provider caused by the template file in scripts/versiontemplates. I tried adding the scripts folder to modulePathIgnorePatterns in package.json but that didn't help.
- Redscreen vs. warning, I think warning is useless because if the app crashes you won't have time to see the warning.
- Should the check and native modules be __DEV__ only?

**Test plan**
Tested that it works when version match and that it redscreens when versions don't before getting other errors on Android and iOS.


janicduplessis:
cc @ide @grabbou

ide:
- Version match: I'd really like to make this check be configurable and have a default implementation that checks for an exact match but also let people override it and check only the major and minor versions if they wish. The difference between patch releases usually doesn't include breaking changes (just new native code or new JS that doesn't require the JS and native to match) and it would be nice to let people choose if they're OK with 0.46.1 JS running on 0.46.0 native.
- Haste: I wonder if the issue is that `modulePathIgnorePatterns` applies only to Jest but not Metro. I think your workaround is simple and fine, otherwise we could make the templates not contain `@providesModule` and make the template generator add it in.
- Throwing vs warning: let's try throwing for now so we fail fast.
- About `__DEV__`: let's keep the check in production. It's cheap in several ways (bytes & runtime) and if there's a version mismatch in prod that results in errors like "X is not a callable module", that's way harder to diagnose than "You're using the wrong version of RN." If we make it a `__DEV__`-only check, I ask that it does console.error instead of throwing an error so we don't change code paths in production.

javache:
I'm not sure I agree with this direction. It could be useful at development time, but we're actually trying to keep the JS compatible with older native code versions for as long as possible, and this would actually prevent that.

ide:
@javache what is the mechanism you use to keep the JS compatible with the native code? Could we use that instead of semver? Also this code doesn't materially affect FB (or anyone else who runs from master since the compared version is always "master") in case that affects your thinking around this.

javache:
True, since we don't use the release branch, this wouldn't necessarily affect us. I would still prefer an additional native module, so perhaps this can go in `PlatformConstants `?

Internally, we're mainly looking at leveraging flow for typing native interfaces and end-to-end testing to ensure compatibility.

ide:
@javache Sounds good, let's move this into PlatformConstants. Native -> Flow defs sounds useful, I think doing the version check will also catch a different class of issues caused by old packager processes, cached bundles, etc...

janicduplessis:
@ide What kind of API were you thinking of to customize that check? It might be a bit tricky since I'm not sure it's even possible to run code before the InitializeCore module.

janicduplessis:
@ide Just tested and InitializeCore does get executed before the app index.js, so I don't see a way to customize this behavior at runtime. I'd also like to keep the check in InitializeCore since otherwise other random errors could happen before the version check.

ide:
OK, that makes sense. This looks good to me then.

@javache How does this look to you?

scripts/bump-oss-version.js
```diff
@@ -46,10 +46,34 @@ let versionMajor = branch.slice(0, branch.indexOf(`-stable`));
 // e.g. 0.33.1 or 0.33.0-rc4
 let version = argv._[0];
 if (!version || version.indexOf(versionMajor) !== 0) {
-  echo(`You must pass a tag like ${versionMajor}.[X]-rc[Y] to bump a version`);
+  echo(`You must pass a tag like 0.${versionMajor}.[X]-rc[Y] to bump a version`);
   exit(1);
 }
 
+// Generate version files to detect mismatches between JS and native.
+let [, major, minor, patch, prerelease] = version.match(/^(\d+)\.(\d+)\.(\d+)(?:-rc\.(\d+))?$/);
```
ide: Prerelease should be a string (not number) that includes `-beta.1`, etc... can you change the regex to this?
```js
/^(\d+)\.(\d+)\.(\d+)(?:-(.+))?$/
```

Also for clearer error messages:
```js
let match = version.match(...);
if (!match) {
  echo(`You must pass a correctly formatted version; couldn't parse ${version}`);
  exit(1);
}
```

scripts/versiontemplates/ReactNativeVersion.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.facebook.react;
+
+import com.facebook.react.common.MapBuilder;
+
+import java.util.Map;
+
+public class ReactNativeVersion {
+  public static final Map<String, Integer> VERSION = MapBuilder.of(
```
ide: `<String, @Nullable Object>` since the prerelease value may be a String

Libraries/Core/InitializeCore.js
```diff
@@ -116,6 +116,24 @@ if (!global.__fbDisableExceptionsManager) {
   ErrorUtils.setGlobalHandler(handleError);
 }
 
+const formatVersion = version =>
+  `${version.major}.${version.minor}.${version.patch}` +
+  (version.prerelease !== null ? `-rc.${version.prerelease}` : '');
```
ide: Just concatenate the prerelease (instead of assuming -rc, see comments later in this review)

janicduplessis:
@ide Updated

ide:
Waiting for CI to pass.

ide:
@janicduplessis it looks like the Jest setup script needs to be updated (see CircleCI output).

janicduplessis:
Updated, tests now pass locally, let's wait for CI again. Ended up just providing an explicit mock for the InitializeCore module since jest automock feature does run the module which is when the error happened.

Libraries/Core/InitializeCore.js
```diff
@@ -116,6 +116,23 @@ if (!global.__fbDisableExceptionsManager) {
   ErrorUtils.setGlobalHandler(handleError);
 }
 
+const formatVersion = version =>
+  `${version.major}.${version.minor}.${version.patch}` +
+  (version.prerelease !== null ? `-${version.prerelease}` : '');
+
+const ReactNativeVersion = require('./ReactNativeVersion');
+const nativeVersion = require('NativeModules').PlatformConstants.reactNativeVersion;
+if (ReactNativeVersion.version.major !== nativeVersion.major ||
```
chirag04: should be keep this behind the __DEV__ flag?

janicduplessis: @ide prefered not to, I don't really have a preference

ide:
CI failed again, probably need to tell Buck about the new Java file.

Libraries/Core/InitializeCore.js
ide: The main thing is to keep dev and prod similar, particularly around edge cases, error handling, initialization, cleanup. If we move this to a `__DEV__`-only check, we should change this to a console.error call instead of throwing an error.

janicduplessis:
Moved the file to the same package to avoid circular deps which buck doesn't support.

janicduplessis:
Ok finally got CI passing :)

ide:
This looks good to me. 

@facebook-github-bot shipit

Libraries/Core/InitializeCore.js
chirag04: i would rather let it continue in prod than spectacularly failing at startup. idk

ide:
@hramos could you or someone else at fb look into why this failed to land? The OSS CI tests look good for it.

scripts/versiontemplates/ReactNativeVersion.js
```diff
@@ -0,0 +1,19 @@
...
+ * @generated by scripts/bump-oss-version.js
+ *
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
```
hramos: This PR is not importing due to the following error:

```
Flow check failed:
1 Flow error and 0 Flow warnings detected!
=====Flow error #0====
Error: scripts/versiontemplates/ReactNativeVersion.js line 15 col 11-11: Unexpected token {
```

PR #15522, opened 2017-08-17
iOS: Fix registration for UIContentSizeCategoryDidChangeNotification

rigdern:
On an iPhone running iOS 10.3.3, I'm not seeing the UIContentSizeCategoryDidChangeNotification fire when changing the system's font scale setting. Registering for the event on `nil` instead of on the application object fixes the issue.

**Test Plan**

Verified that text in an RN app changes size when you change the system's font scale setting (Settings -> General -> Accessibility -> Larger Text). By default, RN text is supposed to scale with the system setting.

Adam Comella
Microsoft Corp.

### DIFF
--- a/React/Modules/RCTAccessibilityManager.m
+++ b/React/Modules/RCTAccessibilityManager.m
@@ -65,7 +65,7 @@ RCT_EXPORT_MODULE()
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(didReceiveNewContentSizeCategory:)
                                                  name:UIContentSizeCategoryDidChangeNotification
-                                               object:RCTSharedApplication()];
+                                               object:nil];
 
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(didReceiveNewVoiceOverStatus:)


shergin:
Is this issue discussed somewhere like SO?

rigdern:
@shergin I don't know. My team encountered and fixed this issue on our own.

shergin:
@rigdern Thank you!
I am asking because this change does not look quite correct. Specifing additional object in `addObserver:...` should **only** filter messages.

https://developer.apple.com/documentation/foundation/nsnotificationcenter/1415360-addobserver?language=objc

> `anObject`
> The object whose notifications the observer wants to receive; that is, only notifications sent by this sender are delivered to the observer.
> If you pass nil, the notification center doesn’t use a notification’s sender to decide whether to deliver it to the observer.

So, if this really solves described problem, it should be a bug in iOS (`NSNotificationCenter`), so I think this should be discussed somewhere else... before we can merge this.

rigdern:
@shergin Thanks for reviewing this.

We used to pass `RCTSharedApplication()` for the `object` parameter so we only received events from the application object. Now we pass `nil` so we receive events from anybody that posts the notification.

It is strange that iOS 10's behavior changed and broke React Native. I'm not sure why iOS did that.

As best as I can tell, this change is okay. Here's my reasoning:
  - `UIContentSizeCategoryDidChangeNotification` sounds like a global notification so I don't think multiple objects will post it and cause us to receive too many notifications. We were listening to `UIApplication` before which is a singleton.
  - The iOS documentation doesn't indicate what object posts the [`UIContentSizeCategoryDidChangeNotification`](https://developer.apple.com/documentation/uikit/uicontentsizecategorydidchangenotification) notification. However, in other cases such as [`UIApplicationBackgroundRefreshStatusDidChangeNotification`](https://developer.apple.com/documentation/uikit/uiapplicationbackgroundrefreshstatusdidchangenotification?language=objc), the docs do describe this: "The object of the notification is the UIApplication object.". So I'm wondering if the omission of this information in the docs indicates that you should use `nil`. Other notifications (e.g. `UIAccessibilityVoiceOverStatusChanged` and `UIAccessibilityAnnouncementDidFinishNotification`) are in a similar state where the docs don't say what object posts them. React Native registers for these notifications on `nil`.

shergin:
Oh, my bad! 😞  I misread the diff. Sorry.
The fix is absolutely legit.
Thank you!

Issue #11, opened 2015-01-30
Exception in render() propagates to Xcode


gaearon:
If I throw inside `componentDidMount`, I see RSOD.
If I throw in `render`, I'm being thrown into Xcode and I have to restart (not reload) the app.

brentvatne:
This is no longer an issue

PR #15529, opened 2017-08-17
Remove 'box-only' as a layoutOnlyProp

tomasreimers:
Currently, having the property `pointerEvents={'box-only'}` marks a view as layoutOnly. For optimization reasons, layoutOnly nodes are [not added to the node hierarchy](https://github.com/facebook/react-native/blob/b103903ec869bc48dfcaf001dc926957d0b5200a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java#L99) The problem is that should this box ever need to be [transitioned to not-layout-only](https://github.com/facebook/react-native/blob/b103903ec869bc48dfcaf001dc926957d0b5200a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java#L394) (for example, because you add the a CSS transform property),  it must be added to the hierarchy. To add it to the hierarchy the [React Styles Diff Map](https://github.com/facebook/react-native/blob/b103903ec869bc48dfcaf001dc926957d0b5200a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java#L396) is passed along with the backing shadowNode to create the node. The problem is that at no point were the original `pointerEvents` and so the new node will be created with `pointerEvents =  'auto'`.

A more correct fix to this problem might be to have shadowNodes carry around their pointerEvents, although this is likely a greater design decision which is why am I proposing the quick fix now.

Will also resolve: https://github.com/react-native-community/react-native-modal/issues/11

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewProps.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewProps.java
@@ -168,7 +168,7 @@ public class ViewProps {
       return true;
     } else if (POINTER_EVENTS.equals(prop)) {
       String value = map.getString(prop);
-      return "auto".equals(value) || "box-none".equals(value);
+      return "auto".equals(value);
     } else {
       return false;
     }


shergin:
@tomasreimers This in fact affects only Android, right?

tomasreimers:
Yes, checkout the issue I pointed to on react-native-modal that describes
the issue more in depth.

mmazzarolo:
+1 I can confirm this caused a lot of headaches in the #11 `react-native-modal` issue 😄

shergin:
I even don't understand how `pointerEvents` style logically can be connected to `layout only` nodes. 😞

ide:
I suspect pointerEvents are related because one of the reasons we'd create a real view (instead of only a shadow view) is so it can receive touch events. It seems to me there are two bugs in the logic.

We should check for some properties that always require a real view (ex: transform) and make sure they force layout-only to be false. We also should have the shadow views carry more information as mentioned in the first post above so that we can turn a layout-only view into a real view correctly.

janicduplessis:
Maybe just ship this as is and add a todo comment + issue to improve.

PR #15539, opened 2017-08-17
Remove request dependency from core directory

Ethanent:
request is an unnecessary dependency of the core package as it is only used in the /website directory as a required primary dependency.

## Test Plan

request is only used in the /website directory. Searched entire package with regex.

`/require\(('|")request+\('|"))/gi`

All applicable local tests passed!

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,7 +178,6 @@
     "react-devtools-core": "^2.5.0",
     "react-timer-mixin": "^0.13.2",
     "regenerator-runtime": "^0.9.5",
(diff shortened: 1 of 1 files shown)


ide:
@hramos This change looks good to me -- website/package.json already lists `request` under its dependencies.

hramos:
Sounds good to me. No need to manually import this as the dependency should already be present in our offline cache.

ide:
@facebook-github-bot shipit

PR #15542, opened 2017-08-18
iOS: Warn about slow main thread React methods

rigdern:
When code blocks the UI thread for too long, it's a bad sign because this can prevent the app from remaining responsive. This change helps detect such responsiveness issues by warning when a React method executes on the UI thread longer than some threshold.

**Test Plan**

Changed AppState's getCurrentAppState method to sleep for 500 ms and verified that a warning was emitted:

```
2017-08-17 19:45:29.479 [warn][tid:main][RCTModuleMethod.mm:527] mainThreadWatchdog: invocation of [RCTAppState getCurrentAppState:error:] took 501ms
```

Adam Comella
Microsoft Corp.


shergin:
I very much like the idea.

Small considerations:
 * We have to wrap this with `#ifdef RCT_DEV`;
 * I don't think that `NSDate` is the most efficient way to get a time stamp; usually the function from CoreAnimation is a good choice;
 * The message can be more clear and human oriented;
 * The duration should be a named constant and (ideally) configurable some way.

rigdern:
@shergin Thanks for reviewing this so quickly. I have some questions/comments:

> We have to wrap this with `#ifdef RCT_DEV`;

My team measures performance on release builds not debug builds so I don't think it'll work for our use case if we wrap it with `#ifdef RCT_DEV`. Thoughts?

> I don't think that `NSDate` is the most efficient way to get a time stamp; usually the function from CoreAnimation is a good choice;

Specifically which time API should I use? I spent a few minutes searching but couldn't figure it out.

> The message can be more clear and human oriented;

Okay, what do you think of this?

```
mainThreadWatchdog: Invocation of [RCTAppState getCurrentAppState:error:] blocked the main thread for 501ms. Consider spending less time on the main thread to keep the app's UI responsive.
```

> The duration should be a named constant and (ideally) configurable some way.

Do you have a recommendation for an easy way to make it configurable? If not, I'll just hardcode a value in a named constant for the purposes of this PR.

janicduplessis:
> Specifically which time API should I use? I spent a few minutes searching but couldn't figure it out.

CACurrentMediaTime from <QuartzCore/CAAnimation.h> probably

rigdern:
Thanks, @janicduplessis. For reference, I found this detailed article about benchmarking on iOS: http://nshipster.com/benchmarking/. It mentions some relevant details about why `CACurrentMediaTime` is a good fit for our problem:

> Unlike NSDate or CFAbsoluteTimeGetCurrent() offsets, mach_absolute_time() and CACurrentMediaTime() are based on the internal host clock, a precise, monatomic measure, and not subject to changes in the external time reference, such as those caused by time zones, daylight savings, or leap seconds

shergin:
@rigdern 
If we will not wrap it in RCT_DEV, my only consern is performance. Can we (conceptually) cache `[NSThread isMainThread]` value somewhere in this class... or this method (`invokeWithBridge:`) can naturally be called from several threads?

Proposed warning text is awesome! And just using named constant is good enough for now.

shergin:
cc @javache I feel we need your blessing here anyways. 😄

rigdern:
@javache can you comment on whether it's safe to cache `isMainThread` and what level would be a good place to cache it (e.g. at the module or method level). Also, do I need to extend this PR to `RCTCxxMethod.mm` in order to support the new iOS bridge?

React/Base/RCTModuleMethod.mm
```diff
@@ -516,7 +516,19 @@ - (id)invokeWithBridge:(RCTBridge *)bridge
   }
 
   // Invoke method
-  [_invocation invokeWithTarget:module];
+  if (NSThread.isMainThread) {
```
javache: Use `RCTIsMainQueue` for consistency with other code.

React/Base/RCTModuleMethod.mm
```diff
@@ -516,7 +516,19 @@ - (id)invokeWithBridge:(RCTBridge *)bridge
...
+    NSTimeInterval start = [NSDate timeIntervalSinceReferenceDate];
+    [_invocation invokeWithTarget:module];
+    NSTimeInterval duration = [NSDate timeIntervalSinceReferenceDate] - start;
+    if (duration > 0.020) {
+      RCTLogWarn(@"mainThreadWatchdog: invocation of [%@ %@] took %dms",
```
javache: Use self as the argument and it will use the existing `description` method below.

javache:
I agree with @shergin that this needs to be behind a flag. You could maybe do something like

```
#ifndef ENABLE_MAIN_THREAD_WATCHDOG
#def ENABLE_MAIN_THREAD_WATCHDOG RCT_DEV
#endif
```

That way, it can be enabled for your builds by overriding the flag but kept RCT_DEV-only for everyone else.

Also +1 on the `CACurrentMediaTime`.

shergin:
To have a way to tweet threshold duration via `#define` flag is also good idea.

I like this approach in general though, I believe it is helpful. We used something similar heavily at one of my previous job. (But we crashes the app in such cases. 😆 )

rigdern:
@javache, @shergin I addressed your PR feedback:
  - Introduced constant `RCT_MAIN_THREAD_WATCH_DOG_THRESHOLD`. Defaults to `20 ms` in debug builds and to undefined in release builds. Undefined means the watchdog is disabled.
  - Switched from `NSDate` to `CACurrentMediaTime`.
  - Switched from `NSThread.isMainThread` to `RCTIsMainThread`.
  - Switched to `[self methodName]` for printing the module and method name.
  - Made the warning message friendlier.

PR #15544, opened 2017-08-18
add stack-trace for error RawText "" must be wrapped in an explicit <Text> component.

rohozhnikoff:
add stack trace information for common error `RawText "" must be wrapped in an explicit <Text> component.`

currently all we can do, is play bubble-game in our code and dependencies
trying to find "where somebody forget to check string variable for emptiness"
it's common pattern to write `{someStr && <Text>{someStr}<Text>}`, but in case of empty string, this cause the bug ('' is Falsy, and interpreter return it (last statement) from expression, JSX send it to View, like a children and gotcha!)
it's really difficult to debug this kind of bug every time


adamski:
This PR just saved me a lot of time. Please consider pulling this in.

Issue #15126, opened 2017-07-20
Devtools should talk to the development server instead of localhost


ckknight:
### Is this a bug report?

Yes

### Have you read the Bugs section of the How to Contribute guide?

Yes

### Environment

1. `react-native -V`: 0.46.3
2. `node -v`: v6.11.0
3. `npm -v`: 3.10.10
4. `yarn --version`: not used

Then, specify:

- Target Platform: iOS

- Development Operating System: macOS

- Build tools: using XCode to successfully deploy to an iPhone 7

### Steps to Reproduce

(Write your steps here:)

1. Deploy a debug build of a react-native app to a real device on WiFi
2. Use a remote debugging tool such as Nuclide or react-native-debugger

### Expected Behavior

React DevTools should work just as it does when running in the simulator.

### Actual Behavior

In the debugger on my computer, the message "Waiting for React to connect..." appears and never goes away.

Instead of communicating to my computer, React DevTools is attempting to communicate via `localhost` which is the phone.

The offending line is https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/setupDevtools.js#L33

A working replacement for `'localhost'` is `require('NativeModules').SourceCode.scriptURL.split('://')[1].split('/')[0].split(':')[0]`, which works on-device and in-simulator, but there is likely a much better solution.

### Reproducible Demo

All react-native projects are affected.

ckknight:
The sibling file of `getDevServer.js` seems ideal to solve this - it's not obvious why it wasn't used.

nahuelhds:
This would be nice when working with Expo as well!

PR #15547, opened 2017-08-18
Set host of development server for setupDevtools

jhen0409:
Related to #15126, and this would be useful for use React DevTools on real device without modify `setupDevtools.js`.

In Android emulator, the host of `SourceCode.scriptURL` is same with `PlatformConstants.ServerHost` so we can just replace it.

## Test Plan

* Tested on iOS device with [react-devtools](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) package.
* Tested on Android emulator, the `getDevServer` module got the correctly hostname so that don't need `adb reverse`.

## Release Notes

[ENHANCEMENT] [setupDevtools] Set host of development server for setupDevtools


macrozone:
This would be awesome to have. In particular with https://github.com/jhen0409/react-native-debugger. Any update on this?

gaearon:
@javache is not working on RN anymore so we'll need to find somebody else to review this.

vlkosinov:
Can someone merge it? 
Seems like there is no way to use react-devtools for real devices without changing host manually in `Libraries/Core/Devtools/setupDevtools.js` :(

@gaearon

macrozone:
It‘s extremly sad that this pr has been closed :(

This is a very useful change, developing on a feal device is not unusual and sometimes the only way

jhen0409:
Actually this change have been landed in master, it should be released as [RN 0.53](https://github.com/facebook/react-native/commit/fa574c60920588e29d7b642e547e240ac8655e66).

Thanks @javache come back to deal with this. 😃

gaearon:
The PR appears as closed because RN is set up to sync to our internal repository. So all PRs are “closed” on GitHub when they’re imported, but the commits appear on master.

PR #15553, opened 2017-08-18
FlatList - scrollToOffsetByCellKey

humbertojuliao3:
A 'scrollTo' function that  scrolls to a expecified key, defined at 'keyExtractor' prop

A 'scrollTo' function that  scrolls to a expecified key, defined at 'keyExtractor' prop.

In my project, I have a List with variable height for each cell, and i need to scroll to a certain cell and I already know what is the key value of this cell.

## Test Plan

I had made a logic to bypass this problem, so my test was see if it works in the same way, but a pratical way to test it is to create a list with a unique id to 'keyExtractor' prop, and after render navigate to a expecified id


FranciscoMSM:
Is this working? :) Don't you have the problem of the wanted render item has not been renderer yet?

PR #15565, opened 2017-08-19
Update GettingStarted.md

amin0256:
fix a typo on linux keyboark ( Use `ctrl` instead of `command` )


TheSavior:
Thanks for contributing!

PR #15548, opened 2017-08-18
Network Inspector - convert list view to flat list

AlanFoster:
ListView is marked as deprecated in the latest react-native code:

> DEPRECATED - use one of the new list components, such as [`FlatList`](docs/flatlist.html)
> or [`SectionList`](docs/sectionlist.html) for bounded memory use, fewer bugs,
> better performance, an easier to use API, and more features.

I wanted to help out by migrating Network Inspector's existing ListView component to use the new FlatList component

## Test Plan

Ensured that the same functionality worked for both XHR, and WS scenarios, and compared before and after:

### Before

flatlist

### After

The only visual difference is the chosen row colors, and replacing the border highlight with a different background color.

flatlist

PR #15566, opened 2017-08-20
Support flash scroll indicators for android

AlanFoster:
There is missing support for flash scroll indicators on Android. This PR adds this functionality.

## Test Plan

Ensured that the functionality works now within iOS _and_ Android

### Existing iOS Functionality

flashindicators-ios

### New Android Functionality

flashindicators-android


Libraries/Components/ScrollResponder.js
```diff
@@ -471,6 +472,15 @@ var ScrollResponderMixin = {
    * @platform ios
    */
   scrollResponderFlashScrollIndicators: function() {
+    if (Platform.OS === 'android') {
```
janicduplessis: We could just get rid of the iOS code path and always use `UIManager.dispatchViewManagerCommand`, afaik it also works on iOS.

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
```diff
@@ -92,6 +92,10 @@ public void setPagingEnabled(boolean pagingEnabled) {
     mPagingEnabled = pagingEnabled;
   }
 
+  public boolean flashScrollIndicators() {
```
janicduplessis: What is the boolean returned by this? Since we're not using it might as well not return it.

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java
```diff
@@ -121,6 +121,10 @@ public void setScrollEnabled(boolean scrollEnabled) {
     mScrollEnabled = scrollEnabled;
   }
 
+  public boolean flashScrollIndicators() {
```
janicduplessis: Same for the boolean here

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
```diff
@@ -125,6 +125,11 @@ public void receiveCommand(
   }
 
   @Override
+  public void flashScrollIndicators(ReactHorizontalScrollView scrollView) {
+    scrollView.flashScrollIndicators();
```
AlanFoster: Unfortunately not as it's protected :)

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
AlanFoster: I've nuked it now! :)

Just for visibility though; The android API is:

```
true if the animation is played, false otherwise
```

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
```diff
@@ -92,6 +92,10 @@ public void setPagingEnabled(boolean pagingEnabled) {
...
+    return super.awakenScrollBars();
```
AlanFoster: Just a small left-over piece of code; I was initially overriding to just widen the scope:

```java
@Override
public boolean awakenScrollBars() {
  return super.awakeScrolBars();
}
```

But I decided to align the names with the existing API; And now the `super` was out of place. Deleted now, and the PR has been updated :)

Libraries/Components/ScrollResponder.js
AlanFoster: Hmm! In honesty, I wasn't quite sure why on iOS there was an explicit ScrollViewManager as a native module, but on android there isn't one

Unfortunately I'm lacking the iOS-foo to test this change out, and I can't see any equivalent docs for running a modified of react-native with iOS - similar to the the [Android Building React Native from source](https://facebook.github.io/react-native/docs/android-building-from-source.html) docs. I'm going to have to punt on this one unfortunately!

janicduplessis: I think ViewManagers as modules were just never implemented on Android. I was mostly basing this comment on other methods like scrollTo uses only dispatchViewManagerCommand even on iOS. On iOS react nativr is already built from source so you could test by either running the RNTester xcodeproj or changing code under node_modules in an app.

janicduplessis: But there should not be any native changes required on iOS, it should just work when using dispatchViewManagerCommand

janicduplessis: But there should not be any native changes required on iOS, it should just work when using dispatchViewManagerCommand

AlanFoster: Ah; There's some magic inside the `UIManager.js` to handle iOS. Gotcha now :+1:

Updated :)

AlanFoster:
@janicduplessis All feedback applied now; Is the end result good to ship now? :)

janicduplessis:
Thanks @AlanFoster, looks good to me,

@shergin I think this is good to ship.

Libraries/Components/ScrollResponder.js
```diff
@@ -471,8 +471,11 @@ var ScrollResponderMixin = {
    * @platform ios
```
shergin: We have to remove it now. 🎉

AlanFoster:
Thanks for the PR review guys! :+1:

Could I possibly chance my arm for a review of migrating the network inspector from ListView->FlatList? https://github.com/facebook/react-native/pull/15548 😇 

I'm hoping to put up an additional PR once that's approved to make the Network inspector's scrollToBottom mechanism a lot nicer for developers, but I was hoping to get these pre-requisite PRs in first :)

shergin:
Seems I have to fix some internal stuff before it can be landed. I hope I will do it soon.

PR #15576, opened 2017-08-20
explicitly show how to add the optional listener

geirman:
The previous example only showed where to add the optional "listener" but didn't show how to make use of it.

(Write your motivation here.)

## Test Plan


Libraries/Animated/src/AnimatedImplementation.js
```diff
@@ -2857,7 +2857,7 @@ module.exports = {
    *```javascript
    *  onScroll={Animated.event(
    *    [{nativeEvent: {contentOffset: {x: this._scrollX}}}],
-   *    {listener},          // Optional async listener
+   *    {listener: (event, gestureState) => console.log(gestureState)}, // Optional async listener
```
janicduplessis: Since this looks more an onScroll event from ScrollView I think we should not add the gestureState param, might be more confusing than anything else. Maybe just `(event) => console.log(event)`

geirman: I trust your opinion, but gestureState is passed along as the second param... so isn't it better to document it?

janicduplessis: Even for ScrollView onScroll? I thought it was only for PanResponder events.

geirman: Good point, maybe I should move the example to onPanResponderMove... or add another?

janicduplessis: Sure lets just change to onPanResponderMove instead

geirman: Ok, done! I left the example for onScroll too. Can only help, right?

PR #15582, opened 2017-08-21
Fixed ordering in reactContext initialization.

simonracz:
### Description

ReactContext initialization is a bit messy in ReactInstanceManager.java.

There are 3 important events in `setupReactContext`.

- setting the `mCurrentReactContext` member
- copying the `ReactInstanceEventListener`s
- calling the `ReactInstanceEventListener`s

These three are called on two different threads which makes the client code extremely problematic.
Issues:
- `mCurrentReactContext` is set before it is actually initialized
- the standard client code, like the one in `HeadlessJSTaskService.java`, which first checks `getCurrentReactContext`, then sets up a listener if it's null might result in endless wait for the listener, which will never be called

This PR ought to fix these issues, by calling all 3 events in the correct order in one Runnable on the UI thread.

### Test Plan

No special test. It is a refactor.
HeadlessJSTaskService had a rare bug, where `startTask` was never called. That should be fixed by this, but it's impossible to test.


shergin:
Mmm... I am not an Android expert but the original code looks absolutely reasonable for me.
 * I think, It is clear why we create `finalListeners` *not* on UI Thread.
 * What do you mean by uninitialized `mCurrentReactContext`?

simonracz:
Hey @shergin ,

Thanks for your time taking a look at the PR.

By uninitialized `mCurrentReactContext`, I mean that none of the following lines ran, before a "client" could get a hold of the `mCurrentReactContext` instance:
[ReactInstanceManager#L874](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java#L874) to [ReactInstanceManager#L893](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java#L893)

The `ReactInstanceManager` code is okay, but it makes the "client" code brittle, and ugly. Under "client", I mean the one that wants to get a hold of a reactContext instance.

One such example is [HeadlessJsTaskService#L101](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/HeadlessJsTaskService.java#L101)
Client calls:
1. `getCurrentReactContext();`
2. if it's null, then set up a listener and wait for it to be called

Sadly, this simple client code is broken.

Between 1, and 2, many things could happen on a different thread, e.g. reactContext got initialized, the listeners are already copied and your own listener will never be called.
I am working on a tool for RN and I had a similar client code. I encountered this bug in 1:5000 cases.

Another client example is [ReactAppTestActivity](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/androidTest/java/com/facebook/react/testing/ReactAppTestActivity.java#L221). I am mentioning this because of its comment.

Without this PR, the simplest correct client code I could come up is:
1. call `getCurrentReactContext()`
2. if previous call returns null
2.1. set up a `CountDownLatch(1)`
2.2. set up listener that decreses the `CountDownLatch`
3. inside `while(true)` loop
3.1. call `countDownLatch.await(some time);`
3.2. if the wait is unsuccessful, check `getCurrentReactContext()` again

PR #15601, opened 2017-08-22
Android Oreo (8.0) Support

AndrewJack:
Apps targeting Android 8.0 (API level 26) cannot use `TYPE_SYSTEM_OVERLAY ` and `TYPE_SYSTEM_OVERLAY `. Targeting 26 will cause the app to crash when in `DEV_MODE`.

https://developer.android.com/about/versions/oreo/android-8.0-changes.html#cwt

This PR replaces uses of these overlay flags with the new `TYPE_APPLICATION_OVERLAY` when running on a device with Android 8.0 or later.

When using `TYPE_APPLICATION_OVERLAY` it still requires the `SYSTEM_ALERT_WINDOW` permission, just like previous android versions.
https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#TYPE_APPLICATION_OVERLAY

## Test Plan

https://github.com/AndrewJack/react-native-android-oreo tested here

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/WindowOverlayCompat.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/WindowOverlayCompat.java
@@ -1,0 +1,18 @@
+package com.facebook.react.devsupport;
+
+import android.os.Build;
+import android.view.WindowManager;
+
+/**
+ * Compatibility wrapper for apps targeting API level 26 or later.
+ * See https://developer.android.com/about/versions/oreo/android-8.0-changes.html#cwt
+ */
+/* package */ class WindowOverlayCompat {
+
+  private static final int ANDROID_OREO = 26;
+  private static final int TYPE_APPLICATION_OVERLAY = 2038;
+
+  static final int TYPE_SYSTEM_ALERT = Build.VERSION.SDK_INT < ANDROID_OREO ? WindowManager.LayoutParams.TYPE_SYSTEM_ALERT : TYPE_APPLICATION_OVERLAY;
+  static final int TYPE_SYSTEM_OVERLAY = Build.VERSION.SDK_INT < ANDROID_OREO ? WindowManager.LayoutParams.TYPE_SYSTEM_OVERLAY : TYPE_APPLICATION_OVERLAY;
+
+}
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
@@ -40,7 +40,7 @@ import com.facebook.react.bridge.ReactContext;
       WindowManager.LayoutParams params = new WindowManager.LayoutParams(
           WindowManager.LayoutParams.MATCH_PARENT,
           WindowManager.LayoutParams.MATCH_PARENT,
-          WindowManager.LayoutParams.TYPE_SYSTEM_OVERLAY,
+          WindowOverlayCompat.TYPE_SYSTEM_OVERLAY,
           WindowManager.LayoutParams.FLAG_NOT_FOCUSABLE
               | WindowManager.LayoutParams.FLAG_NOT_TOUCHABLE,
           PixelFormat.TRANSLUCENT);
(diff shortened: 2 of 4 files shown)


AndrewJack:
Without the changes in this PR you will see the following crash:
```
08-22 12:30:22.707 25262-25262/com.androido E/AndroidRuntime: FATAL EXCEPTION: main
                                                              Process: com.androido, PID: 25262
                                                              android.view.WindowManager$BadTokenException: Unable to add window android.view.ViewRootImpl$W@4e96541 -- permission denied for window type 2006
                                                                  at android.view.ViewRootImpl.setView(ViewRootImpl.java:789)
                                                                  at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:356)
                                                                  at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:92)
                                                                  at com.facebook.react.devsupport.DevLoadingViewController.setVisible(DevLoadingViewController.java:150)
                                                                  at com.facebook.react.devsupport.DevLoadingViewController.access$100(DevLoadingViewController.java:39)
                                                                  at com.facebook.react.devsupport.DevLoadingViewController$1.run(DevLoadingViewController.java:71)
                                                                  at android.os.Handler.handleCallback(Handler.java:789)
                                                                  at android.os.Handler.dispatchMessage(Handler.java:98)
                                                                  at android.os.Looper.loop(Looper.java:164)
                                                                  at android.app.ActivityThread.main(ActivityThread.java:6541)
                                                                  at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
                                                                  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

ReactAndroid/src/main/java/com/facebook/react/devsupport/WindowOverlayCompat.java
```diff
@@ -0,0 +1,18 @@
...
+import android.view.WindowManager;
+
+/**
+ * Compatibility wrapper for apps targeting API level 26 or later.
+ * See https://developer.android.com/about/versions/o/android-8.0-changes.html#cwt
+ */
+class WindowOverlayCompat {
+
+  private static final int ANDROID_OREO = 26;
+  private static final int TYPE_APPLICATION_OVERLAY = 2038;
```
AndrewJack: Cannot use the Android constants because react native doesn't compile to v26 yet (RN currently uses v23)

ReactAndroid/src/main/java/com/facebook/react/devsupport/WindowOverlayCompat.java
```diff
@@ -0,0 +1,18 @@
+package com.facebook.react.devsupport;
+
+import android.os.Build;
+import android.view.WindowManager;
+
+/**
+ * Compatibility wrapper for apps targeting API level 26 or later.
+ * See https://developer.android.com/about/versions/o/android-8.0-changes.html#cwt
+ */
+class WindowOverlayCompat {
```
janicduplessis: I think it could be useful to make this public, if you prefer not to let's just add a /* package */ comment like here for example https://github.com/AndrewJack/react-native/blob/f1cf378cf4ef3e5473fe2ace6d564be897e56a0e/ReactAndroid/src/main/java/com/facebook/react/devsupport/RedBoxDialog.java#L51

AndrewJack: I don't think it needs to be public. It's only needed in the `devsupport` package and I wouldn't like to encourage developers to use this either.

I'll add `/* package */`

janicduplessis:
@shergin Looks good to me if you want to ship.

arpit:
Any update on this? Definitely seeing this crash and the patch looks good. The CI failure seems to be unrelated to this code unless I am missing something

janicduplessis:
Thanks @AndrewJack for the PR and @shergin for shipping 🎉

rekha110254:
@AndrewJack @janicduplessis   Any update on this?? I'm facing this issue.

AndrewJack:
@rekha110254 What version are you using? This was only shipped in RN 0.50.

temitope:
@AndrewJack thanks for this PR, this bug is kicking my project in the shins. I am using `RN 0.41` and wondering if there is a recommended way to incorporate this fix outside of 1) upgrading to `RN 0.51` and possibly upsetting my other dependencies and 2) manually updating the source files for RN creating an unsupported unique version of the RN codebase just for me :/? Basically I am hoping and wondering if there is a way to just incorporate these changes related to the overlay or do I just need to take the upgrade plunge?
Thanks

zaplo00:
is navigator.geolocation.getCurrentPosition working in Android 8?  Could someone with actual device test it?

rekha110254:
@AndrewJack  RN 0.51.1

PR #15603, opened 2017-08-22
Return meaningful error when no devices available

vladimir-kotikov:
`react-native run-ios --device` should report 'No iOS devices connected.' in case when no devices attached. However due to logic bug when empty array is being treatened as a boolean value, cli reports odd message, saying `Could not find device with the name: "true".`

## Test Plan

Disconnect all iOS devices from developer machine and run `react-native run-ios --device` in any project - console message should state that there are no devices available.


vladimir-kotikov:
Failures doesn't seem to be relevant to the change

PR #15610, opened 2017-08-22
Android ScrollView: Add scrollBy method

rigdern:
iOS PR: facebook/react-native#15609

Currently, React Native allows you to scroll a ScrollView to a particular position using `scrollTo`. If instead you want to scroll by a delta, you could try to do this using `scrollTo` (add the delta to the current scroll position) but this doesn't always work out as intended due to the async nature of React Native (what you think is the current scroll position might be a stale value). This change introduces a `scrollBy` API to solve this problem. `scrollBy` takes a delta to scroll by and native takes care of doing the arithmetic because it knows the latest scroll position.

**Test Plan**

In a test app, verified that `scrollBy` works properly with and without animation and for both horizontal and vertical `ScrollViews`. Also, my team is using this change in our app.

Adam Comella
Microsoft Corp.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewCommandHelper.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewCommandHelper.java
@@ -25,11 +25,13 @@ public class ReactScrollViewCommandHelper {
   public static final int COMMAND_SCROLL_TO = 1;
   public static final int COMMAND_SCROLL_TO_END = 2;
   public static final int COMMAND_FLASH_SCROLL_INDICATORS = 3;
+  public static final int COMMAND_SCROLL_BY = 4;
 
   public interface ScrollCommandHandler<T> {
     void scrollTo(T scrollView, ScrollToCommandData data);
     void scrollToEnd(T scrollView, ScrollToEndCommandData data);
     void flashScrollIndicators(T scrollView);
+    void scrollBy(T scrollView, ScrollByCommandData data);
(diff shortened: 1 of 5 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewCommandHelper.java
```diff
@@ -26,10 +26,12 @@
 
   public static final int COMMAND_SCROLL_TO = 1;
   public static final int COMMAND_SCROLL_TO_END = 2;
+  public static final int COMMAND_SCROLL_BY = 3;
```
AlanFoster: Just for visibility; I've also changed this file: https://github.com/facebook/react-native/pull/15566/files#diff-0e688b19a3a45a0d17cd8d922cc0580aR29

If that makes its way into the code-base, there may be a small rebase required :+1:

rigdern: @AlanFoster thanks for the heads up.

rigdern:
@AaaChiuuu can you review this PR or find an appropriate person to do so?

shergin:
I have to mention here that same consideration described in #15609 also applicable here.

rigdern:
Thanks for the discussion. We decided to go with an alternative solution. We built a native module which combines a number of operations our virtual list view implementation requires (e.g. scrolling & repositioning items) into a single method so all of the operations are completed in the same frame.

PR #15611, opened 2017-08-23
Native Animated - Support events using RCT{Direct|Bubbling}EventBlock on iOS

janicduplessis:
When calling a prop of type `RCTDirectEventBlock` or `RCTBubblingEventBlock` it uses a completely different code path than events using `[RCTEventDispatcher sendEvent:]` and those were not dispatched to the `RCTEventDispatcherListener`s. We also do some event name normalization which caused issues between the JS and native event names. To fix that I simply remove the parts we normalize from the event key.

## Changelog:

[iOS] [Fixed] - Support events using RCT{Direct|Bubbling}EventBlock

## Test plan:

Added a Slider (it used RCTBubblingEventBlock for it's onValueChange event) that can control a native animated value in RNTester to reproduce the bug and made sure this diff fixes it.

### DIFF
--- a/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
+++ b/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
@@ -28,6 +28,21 @@
 #import "RCTValueAnimatedNode.h"
 #import "RCTTrackingAnimatedNode.h"
 
+// We de some normalizing of the event names in RCTEventDispatcher#RCTNormalizeInputEventName.
+// To make things simpler just get rid of the parts we change in the event names we use here.
+// This is a lot easier than trying to denormalize because there would be multiple possible
+// denormalized forms for a single input.
+NSString *RCTNormalizeAnimatedEventName(NSString *eventName)
+{
+  if ([eventName hasPrefix:@"on"]) {
+    return [eventName substringFromIndex:2];
+  }
+  if ([eventName hasPrefix:@"top"]) {
+    return [eventName substringFromIndex:3];
+  }
+  return eventName;
+}
+
 @implementation RCTNativeAnimatedNodesManager
 {
   __weak RCTBridge *_bridge;
@@ -321,7 +336,7 @@
   RCTEventAnimation *driver =
     [[RCTEventAnimation alloc] initWithEventPath:eventPath valueNode:(RCTValueAnimatedNode *)node];
 
-  NSString *key = [NSString stringWithFormat:@"%@%@", viewTag, eventName];
+  NSString *key = [NSString stringWithFormat:@"%@%@", viewTag, RCTNormalizeAnimatedEventName(eventName)];
   if (_eventDrivers[key] != nil) {
     [_eventDrivers[key] addObject:driver];
   } else {
@@ -335,7 +350,7 @@
                           eventName:(nonnull NSString *)eventName
                     animatedNodeTag:(nonnull NSNumber *)animatedNodeTag
 {
-  NSString *key = [NSString stringWithFormat:@"%@%@", viewTag, eventName];
+  NSString *key = [NSString stringWithFormat:@"%@%@", viewTag, RCTNormalizeAnimatedEventName(eventName)];
   if (_eventDrivers[key] != nil) {
     if (_eventDrivers[key].count == 1) {
       [_eventDrivers removeObjectForKey:key];
@@ -357,7 +372,7 @@
     return;
   }
 
-  NSString *key = [NSString stringWithFormat:@"%@%@", event.viewTag, event.eventName];
+  NSString *key = [NSString stringWithFormat:@"%@%@", event.viewTag, RCTNormalizeAnimatedEventName(event.eventName)];
   NSMutableArray<RCTEventAnimation *> *driversForKey = _eventDrivers[key];
   if (driversForKey) {
     for (RCTEventAnimation *driver in driversForKey) {
--- a/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
+++ b/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
@@ -28,7 +28,7 @@
 #import "RCTValueAnimatedNode.h"
 #import "RCTTrackingAnimatedNode.h"
 
-// We de some normalizing of the event names in RCTEventDispatcher#RCTNormalizeInputEventName.
+// We do some normalizing of the event names in RCTEventDispatcher#RCTNormalizeInputEventName.
 // To make things simpler just get rid of the parts we change in the event names we use here.
 // This is a lot easier than trying to denormalize because there would be multiple possible
 // denormalized forms for a single input.
--- a/RNTester/js/examples/NativeAnimation/NativeAnimationsExample.js
+++ b/RNTester/js/examples/NativeAnimation/NativeAnimationsExample.js
@@ -221,7 +221,7 @@ class InternalSettings extends React.Component<
 
 class EventExample extends React.Component<{}, $FlowFixMeState> {
   state = {
-    scrollX: new Animated.Value(0),
+    anim: new Animated.Value(0),
   };
 
   render() {
@@ -233,7 +233,7 @@ class EventExample extends React.Component<{}, $FlowFixMeState> {
             {
               transform: [
                 {
-                  rotate: this.state.scrollX.interpolate({
+                  rotate: this.state.anim.interpolate({
                     inputRange: [0, 1],
                     outputRange: ['0deg', '1deg'],
                   }),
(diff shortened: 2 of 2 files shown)


React/Base/RCTEventDispatcher.m
```diff
@@ -37,6 +37,50 @@
...
+}
+
+@synthesize eventName = _eventName;
+@synthesize viewTag = _viewTag;
+@synthesize coalescingKey = _coalescingKey;
+
+- (instancetype)initWithName:(NSString *)name viewTag:(NSNumber *)viewTag arguments:(NSArray *)arguments
+{
+  self = [super init];
+  if (self) {
```
shergin: One line, please.

React/Base/RCTEventDispatcher.m
```diff
@@ -37,6 +37,50 @@
   );
 }
 
+/**
```
shergin: Separate file, please.

janicduplessis:
@shergin Updated

nerdmed:
@janicduplessis this is awesome! Looking forward for this!

shergin:
@janicduplessis What does `RCTInputEvent` and `Animated` have in common?

janicduplessis:
@shergin Ok I understand why you are confused :)

Not sure why it is called INPUT event because afaik it has nothing to do with inputs, I just kept the current naming.

So when using RCTDirectEventBlock to create a view manager prop it ends up here https://github.com/facebook/react-native/blob/master/React/Views/RCTComponentData.m#L242 then createEventSetter will create a block that calls into `[sendInputEventWithName:]` (https://github.com/facebook/react-native/blob/master/React/Views/RCTComponentData.m#L121) to dispatch the event to JS. We want to call the listeners there too like we do in `[sendEvent:]` so native animated can do its thing.

janicduplessis:
And the other part of the PR is to deal with the event name normalization that is done in sendInputEventWithName (https://github.com/facebook/react-native/blob/master/React/Base/RCTEventDispatcher.m#L91)

shergin:
OMG, now I know how RCTDirectEvent works. Almost. Thanks!
I will try to figure it out. We have to split this anyways though.

janicduplessis:
@shergin 👍 Let me know if you need more context on anything else.

shergin:
@janicduplessis Could you please help me figure out why this was named `sendInputEvent`? Is it just because it was used for TextInput only?
Should we generalize this approach? And rename it?

janicduplessis:
@shergin From doing a quick search looks like it is only used in `createEventSetter` and `sendTextEventWithType`. It has been deprecated for a while (https://github.com/facebook/react-native/blob/e1577df1fd70049ce7f288f91f6e2b18d512ff4d/React/Base/RCTEventDispatcher.h#L87) so changing / removing it should be fine.

There is also `sendEvent` which takes a `RCTEvent` and is similar to `sendInputEvent` but more complex, It uses an event queue and does optional coalescing. Ideally we'd always use `sendEvent` but I don't know if the scheduling would end up being exactly the same.

shergin:
@janicduplessis I just want to know that we know what we are doing. And do the right things. 😄 
I would love to adopt `sendEvent`... everywhere!

janicduplessis:
@shergin Ok, I had a better look at that code and it should be possible to switch to using only `sendEvent`.  There's a bit more overhead but it should not make a noticeable difference performance wise. Some of our more frequent events like onScroll already use `sendEvent` anyway.

I can put up a PR tomorrow to remove `sendInputEvent` and move everything to `sendEvent`.

shergin:
@janicduplessis Great! Do you mean remove `sendInputEvent` from Animated of remove completely?

janicduplessis:
I think we can get rid of it completely and always use sendEvent

janicduplessis:
@shergin Removed `sendInputEvent` and created a new RCTEvent class that can be used with `sendEvent` and that replicates the behaviour of `sendInputEvent`. Named it `RCTComponentEvent` since `sendInputEvent` was used to dispatch events to components (including but not limited to text input events). That name should make it a lot clearer what it is used for.

Also simplified dealing with non-coalescable events by automatically generating a coalescingKey is `canCoalesce` returns NO. Also made a few coalescing related methods/props optional so they can be omitted if `canCoalesce` is false. Otherwise every RCTEvent class would have to implement some unique key generator even if the event is never coalescable.

janicduplessis:
Also this should probably be split in 2 PRs, just wanted to keep it here for now but once we're happy with a solution I'll make a new PR for the removal of `sendInputEvent`.

nerdmed:
@janicduplessis @shergin any news here? What is needed to proceed with this pr?

React/Base/RCTEventDispatcher.m
```diff
@@ -28,12 +29,15 @@
...
+// Key generator for events that cannot be coalesced.
+static uint16_t RCTNonCoalescableEventKey = 0;
+
 static NSNumber *RCTGetEventID(id<RCTEvent> event)
 {
   return @(
     event.viewTag.intValue |
     (((uint64_t)event.eventName.hash & 0xFFFF) << 32) |
-    (((uint64_t)event.coalescingKey) << 48)
+    (((uint64_t)([event canCoalesce] ? event.coalescingKey : RCTNonCoalescableEventKey++)) << 48)
```
shergin: We access optional `canCoalesce` and `coalescingKey` here.
So, we have two options here:
 1. Do `responseToSelector` check here.
 2. Make that props required for all implementations.

I strongly prefer 2 because we should maintain this function as perfromant as possible.

Right?

shergin: Oh, no, I see, I am wrong.

shergin:
@janicduplessis  Okay, we will be back to this PR after landing of #15894. Right?

janicduplessis:
@shergin Yep, then I will rebase this PR on too of the event improvements

nerdmed:
@janicduplessis thanks a lot for your previous effort on this PR. It would enable really nice native animations with RN. This become stale for more a couple of month now. Any change to get this going?

nerdmed:
@shergin @janicduplessis i know you guys have a lot of stuff to do and I can understand that this PR is not easy and needs focus. I would really like to help if my ObjC skills where better but it took me to much time to even find the problem. Please see this just as a friendly reminder for this issues. This change makes react native in a very special way unique as it brings more native power to the users. If there is something else i can help you with just let me know!

janicduplessis:
@nerdmed I'll try to simplify this diff and remove the refactor to event dispatching. This way it should be simple to land.

janicduplessis:
Ok so I re-read the thread to get some context. What I'll need to do is address some comments by @shergin and split this PR in 2 parts. One that refactors `RCTDirectEventBlock` to use `sendEvent` instead of `sendInputEvent` and one to fix the native animated implementation to strip event name prefixes.

janicduplessis:
Going to close this for now, if anyone has time feel free to pick this up.

janicduplessis:
Re-opening this now that #15894 is landed, this just needs a rebase.

janicduplessis:
@shergin Rebased and tested it still works :)

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -28,6 +28,21 @@
 #import "RCTValueAnimatedNode.h"
 #import "RCTTrackingAnimatedNode.h"
 
+// We do some normalizing of the event names in RCTEventDispatcher#RCTNormalizeInputEventName.
+// To make things simpler just get rid of the parts we change in the event names we use here.
+// This is a lot easier than trying to denormalize because there would be multiple possible
+// denormalized forms for a single input.
+NSString *RCTNormalizeAnimatedEventName(NSString *eventName)
```
shergin: We already have same function defined in `Base/RCTEventDispatcher.m$21`.
Are they special reasons why we should duplicate this code here?

janicduplessis: They are actually slightly different, this one just removes the prefix and the one in `Base/RCTEventDispatcher.m$21` replaces `on` with `top`. Basically when happens is that when events are registered / unregistered it is done with the `on` prefix and when they are dispatched it is with the `top` prefix. This just ignore prefixes when creating they key for the event mapping.

The only case this could be problematic is if a component has a `onFoo` and a `topFoo` prop but I think it is fine. The price to pay for the mess we currently have with the event name prefixes...

cpojer:
What's the status of this PR?

janicduplessis:
@cpojer @shergin Rebased this and made sure the fix is still needed and works.

cpojer:
Had to revert this because of:

```
react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m:35:11: error: no previous prototype for function 'RCTNormalizeAnimatedEventName' [-Werror,-Wmissing-prototypes]
NSString *RCTNormalizeAnimatedEventName(NSString *eventName)
          ^
```

Could you send it again with this fixed?

janicduplessis:
@cpojer Fixed https://github.com/facebook/react-native/pull/25317

PR #15614, opened 2017-08-23
fix and enrich Vibration docs

ivan-ha:
The docs for [Vibration](https://facebook.github.io/react-native/docs/vibration.html) is broken and not informative now. Making people difficult to understand the usage without digging into the source code.

screen shot 2017-08-23 at 12 44 50

## Test Plan
Tested the page locally.

fireshot capture 4 - vibration - http___localhost_8079_react-native_docs_vibration html vibrate


hramos:
Wonderful, thanks!

PR #15615, opened 2017-08-23
use `uniqBy` for iteratee support

arahansen:
I noticed `uniq` does not take a second parameter the way it's being used here. I think the function what was intended is [uniqBy](https://lodash.com/docs/4.17.4#uniqBy)


grabbou:
Good catch! This error was introduced before we even merged this to react-native.

CC: @Kureev when replacing `self-written` helpers (in this case - ./uniq.js that I made) with loadsh version - we forgot to use `uniqBy` and as a result - stopped respecting the function.

grabbou:
Thanks for this! I have updated the branch to see if the tests stop failing.

Kureev:
@facebook-github-bot shipit

deecewan:
@Kureev any idea when this will make it into a release?

PR #15626, opened 2017-08-24
Android: Fix sizing of TextInput when font scale != 1

rigdern:
Sometimes, TextInputs on Android have the wrong size. Specifically, this occurs when the user has changed the system's font scale setting, the TextInput has `allowFontScaling={false}`, and the `fontSize` prop isn't set.

To fix this, TextInput's measure function needs to take into account the value of `allowFontScaling`.

**Test Plan**

Created a test app with a multiline `TextInput` which had `numberOfLines` set to `3`. Changed the system's font scale setting and verified that the `TextInput` was tall enough to show exactly 3 lines when `allowFontScaling` was both `true` and `false`.

Here are screenshots for `allowFontScaling={false}`:

**Before**

Notice that the `TextInput` is too short and is clipping the third line.

image

**After**

Notice that the `TextInput` is tall enough to render exactly three lines of text as expected.

image

Adam Comella
Microsoft Corp.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextShadowNode.java
@@ -331,11 +331,11 @@ public class ReactTextShadowNode extends LayoutShadowNode {
 
   private float mLineHeight = Float.NaN;
   private boolean mIsColorSet = false;
-  private boolean mAllowFontScaling = true;
   private int mColor;
   private boolean mIsBackgroundColorSet = false;
   private int mBackgroundColor;
 
+  protected boolean mAllowFontScaling = true;
   protected int mNumberOfLines = UNSET;
   protected int mFontSize = UNSET;
   protected float mFontSizeInput = UNSET;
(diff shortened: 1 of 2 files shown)


shergin:
@rigdern We have to move this logic to `ReactTextInputLocalData.java` now.

shergin:
@@rigdern Or... Wait, it should just work now without any changes. Right?

rigdern:
@shergin I think you are right that this bug might already be fixed in master.

It looks like the shadow node used to receive props from JavaScript and then apply those props to a `ReactEditText` for measuring. We had a bug here because the calculations it did on those props during measuring weren't quite the same as the calculations done on the UI thread during rendering.

Now it looks like the UI thread reads properties directly out of its `ReactEditText`, sends those properties to the shadow node, and then the shadow node puts them into its own `ReactEditText` for measurement purposes. Consequently, the measuring and rendering results should be identical because they're using identical `ReactEditText` properties.

Is that correct? If so, I suspect this bug is already fixed but I'll double check with some testing.

By the way, it seems like the shadow node's measuring code relies on the `onContentSizeChange` prop being non-null. If it's null, `mLocalData` will be null and the measurement function will return a size of 0. Fortunately, Android's TextInput JavaScript implementation never passes null for that prop: `onContentSizeChange={this._onContentSizeChange}`. I wanted to point this implementation detail out because it seems fragile.

shergin:
@rigdern Yeah, that's correct, it is how the new auto resizable TextInput works.
So, why do you think that this implementation relyes on `onContentSizeChange`? I assumed that `localData` is changing on any `onLayout` even or so... ¯\_(ツ)_/¯

PR #15631, opened 2017-08-24
Flow type: saveToCameraRoll returns a string

maartenschumacher:
Updates the flow typing to return Promise\<string\> instead of Promise\<Object\>. To validate that it actually does return a string, see https://github.com/facebook/react-native/blob/6493a857542c0985470311511c1be80982fcd474/Libraries/CameraRoll/RCTCameraRollManager.m#L98 and https://github.com/facebook/react-native/blob/6493a857542c0985470311511c1be80982fcd474/Libraries/CameraRoll/RCTCameraRollManager.m#L116


maartenschumacher:
test plan: i'm getting the same 133 flow errors before and after this change...

hramos:
re: unrelated Flow type errors, a fix should be landing soon

TheSavior:
@shergin We should try to reland this?

Issue #12405, opened 2017-02-15
Trying to add 24 hour format to react-native DatePickerIOS


SudoPlz:
This is a long requested feature, but no one seems to be doing anything about it.

I'm trying to add a boolean value to our DateTimePickerIOS so that the js part of it looks like this:

```javascript
const RCTDatePickerIOS = requireNativeComponent('RCTDatePicker', {
  propTypes: {
    ...DatePickerIOS.propTypes,
    date: PropTypes.number,
    uses24hourFormat: PropTypes.bool, // <---
    minimumDate: PropTypes.number,
    maximumDate: PropTypes.number,
    onDateChange: () => null,
    onChange: PropTypes.func,
  }
});
```

Then I modified `React > Views > RCTDateTimePickerManager.m`. Specifically I added this code:

```Objective-C
RCT_CUSTOM_VIEW_PROPERTY(uses24hourFormat, BOOL, UIDatePicker)
{
  BOOL is24hFormat = json ? [RCTConvert BOOL:json] : NO;
  if (is24hFormat) {
    [view setLocale:[NSLocale localeWithLocaleIdentifier:@"NL"]];
  } else {
    [view setLocale:defaultView.locale];
  }
}
```

That solution is based on [this SO answer](http://stackoverflow.com/questions/21429841/uidatepicker-24hour-formate-in-ios).

I've tried to do this, but with no luck at all.. Even-though this code compiles with no errors, I can't seem to do what I want.

On the other hand, I know that solution works, because when I modify this code on the same react-native file:

```Objective-C
- (UIView *)view
{
  UIDatePicker* datePicker = [RCTDatePicker new];
//  [datePicker setLocale:[NSLocale localeWithLocaleIdentifier:@"NL"]];  // If I uncomment this line
  return datePicker;
}
```
I do get a 24hour time picker, but I want to be able to do this from js.

Any ideas would be appreciated.
Also if someone from the fb team pointed me to the right direction I could open a pull request.

Thanks.

antonshevchenko:
The UIDatePicker displays time based on the user's locale. It doesn't seem like you can override that behavior in terms of UI.

SudoPlz:
You can if you initialise it for a different locale, that's what `    [view setLocale:[NSLocale localeWithLocaleIdentifier:@"NL"]];` does.

I tested it and it does work.

hramos:
[...]

PR #15635, opened 2017-08-24
Added option to manually set DatePickerIOS locale.

Gongreg:
## Motivation

Sometimes there is a need to set DatePickerIOS locale manually. For that I added optional prop locale. If user pass locale prop the DatePickerIOS will use it, otherwise it will use current phone locale.

Issues: #12405, https://github.com/mmazzarolo/react-native-modal-datetime-picker/issues/31

## Test Plan

Added location to RNTester DatePickerIOS screen. You can type locale and it will change to that one.  For checking default locale I change phone language/region to other and it used the new locale.


Gongreg:
Hello,
Should I do something next about this PR? I see that CI is broken for all PR's.

Gongreg:
@javache , @shergin, so thanks for amazing experience with open source possibilities in this project. I see that this functionality was added to react native already. Great support! Keep it up!

PR #15636, opened 2017-08-24
Add a way to dismiss PopupMenu elements

miguelsm:
In native Android apps, like the YouTube app, context menus are closed when the device orientation changes.

In React Native apps instead, when having a [PopupMenu](https://developer.android.com/reference/android/widget/PopupMenu.html) open and rotating the device, the PopupMenu is not dismissed and appears in a wrong position on the screen.

This PR exposes a `dismissPopupMenu` method to allow the application to dismiss any open PopupMenu:

```(javascript)
UIManager.dismissPopupMenu()
```

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -74,6 +74,7 @@ public class NativeViewHierarchyManager {
   private final LayoutAnimationController mLayoutAnimator = new LayoutAnimationController();
 
   private boolean mLayoutAnimationEnabled;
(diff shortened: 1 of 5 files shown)


chirag04:
@miguelsm should we rather aim to dismiss the popup when the screen is rotated or very least fix the position on rotation?

miguelsm:
@chirag04 That would also work for this particular use case. I've tried dismissing the popup from the ```NativeViewHierarchyManager.updateLayout``` method and it works fine. Doing it this way would require fewer changes in the code than my original commit (just that one file). The only drawback I see in dismissing the popup menus this way is that it wouldn't allow for dismissing them programatically from the app.

satya164:
cc @hramos

miguelsm:
Rebased, thank you.

PR #10084, opened 2016-09-24
Improve init and runtime performance of native modules

ransj:
During react context Initialization, exported methods in native module are found by call getAnnotation with declared methods in native modules, this procesure contains a large amount of
reflections. More native modules, more methods in each module, that will be nightmare!
So ReactNativeModuleProcessor handle ReactMethod and ReactSyncHook annotations, assemble exported methods together and create method descriptors during compile time. Since Ids of exported methods are generated, we can use these ids to call target method of native modules directly.
Also this commit will enable ReactNativeModuleProcessor and ReactPropertyProcessor in gradle, anyone work with the source can see the change.

PR #15648, opened 2017-08-25
replace JavaModuleWrapper with AcJavaModuleWrapper

ransj:
Replace JavaModuleWrapper with AcJavaModuleWrapper. This pr aim to enable further optimization of native modules. CoreModuleProvider provides optimization code for official framework, CustomModuleProvider provides optimization code for third-party views and modules. When there
is no optimization,  it will fallback to use methods of parent(JavaModuleWrapper).
The plan is : collect MethodDescriptors during compile time, avoid runtime search. Call method directly, avoid use reflection. Both official and third-party modules should be optimized.

## Test Plan

When there is no optimization,  it will fallback to use methods of parent (JavaModuleWrapper).


ransj:
This pr is one of a series prs derive from old [PR 10084](https://github.com/facebook/react-native/pull/10084), which had proved to accelerate getMethodDescriptors and invoke methods
of native modules by nearly 60x, checkout and run benchmark: [BenchMark](https://github.com/ransj/react-native/blob/master/RNTester/android/app/src/androidTest/java/com/facebook/react/uiapp/NativeModuleBenchMarkTest.java).
The plan as follows:
1. Move MethodDescriptor out from JavaModuleWrapper
2. Add AcJavaModuleWrapper and replace JavaModuleWrapper with AcJavaModuleWrapper
3. Add ReactNativeModuleProcessor to handle ReactMethod annotations
4. Add library project to enable ReactNativeModuleProcessor and ReactPropertyProcessor
5. Update build.gradle file to enable annotations processors during build time
6. Provide compiler to optimize code write by app developers.

javache:
Where can I find the code that generates getModuleHelper? The performance improvements sound really good but it seems like there's some pieces missing here.

javache:
Nevermind, found your other PR's. I'll play around with this a bit, but codegen'ing a lot of metadata like this does can be very expensive from an app size perspective. It may still be useful to have it as an opt-in mechanism for specific modules.

javache:
This doesn't seem to work with multi-dex apps as it generates Lcom/facebook/react/bridge/JavaModuleWrapper$0 multiple times.

Your current benchmark is also quite synthetic. `getMethodDescriptors` is not a large part of the total startup cost of React Native, so even if this is an improvement, it may be a small one relative to the entire runtime, or even negative due to increased class-loading cost.

ransj:
@javache Yes, the benchmark is simply about getMethodDescriptors method, but the optimization is more further, it is more memory friendly, avoid creating hundreds of JavaMethodWrapper. Recently i have not measure the impact on startup of React Native, but i have received feedback like below:
qq20170826-163958 2x

PR #15651, opened 2017-08-25
CallInfo API

ivanzotov:
CallInfo is API to listen a call state like 'ringing', 'offhook', 'idle', it's supported only on Android.

```
CallInfo.addEventListener('change', state => {
   ...
})
```

Also I'm going to add 'incoming/outgoing' and 'phone number' information.

How do you think it is a good idea to continue working on this PR? Or it's better to extract it into a separate repo? I would be glad to maintain this feature in react-native.

### DIFF
--- a/Libraries/Utilities/CallInfo.js
+++ b/Libraries/Utilities/CallInfo.js
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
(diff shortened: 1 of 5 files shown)


javache:
Thanks for this PR! Given that it's completely isolated, I think it would actually make a lot of sense as a 3rd party extension.

It seems like there's already https://www.npmjs.com/package/react-native-call-receiver# but the github page is gone.

ivanzotov:
@javache ok thanks, will do an extension

PR #15657, opened 2017-08-25
Fix Yoga podspec casing

javache:
Fix incorrect filename case

Test plan:

Run `scripts/process-podspecs.sh`


rigdern:
@javache I just want to double check about this. I think we'll need to do this as two separate commits. Changing the casing of a file is tricky with git. In one commit, you completely change the name of the file (i.e. change more than just the casing). In the second commit, you rename the file to the final name you want.

javache:
It seems to have landed correctly, could you try pulling and verifying you have the commit with the right case?

rigdern:
@javache, @alloy This change looks good to me. I thought more about my past experiences. Here are the problems I've run into with changing the casing of a file in git but it looks like you've avoided them:
  - Somebody tries to rename a file but ends up checking in two files: one with the old casing and one with the new casing.
  - Somebody tries to change and rename a file on a case insensitive file system. Git doesn't notice that the file has been renamed so the file has the old casing after the commit. To force git to notice the rename, you have to do `git mv <old-casing> <new-casing>`.

I didn't understand the problems well enough until now to realize they could be avoided in a single commit.

Issue #1831, opened 2015-07-01
[FlatList] Rows aren't rendered until scroll


michalraska:
Hi,
I have list view which displays partially local and partially remote data. Local data are initial for ListView.DataSource. This datasource is set in state of my custom component which wraps ListView and passed in render method to ListView. When are remote data received, new datasource is cloned by cloneWithRowsAndSections method and set to state of custom component. Problem is that are re-rendered only already displayed rows and new rows are rendered after scroll.

![renderbug](https://cloud.githubusercontent.com/assets/827753/8455400/9fa2f94a-2004-11e5-8a76-fec4dd9db164.gif)

Is it bug or how I should to force rendering of ListView? With react-native 0.5 it worked but after upgrade to 0.6 it behaves as described above.

Furizaa:
+1
Encountering exactly the same problem. Worked with 0.5 but broke with 0.6.

ide:
This should be fixed in 0.7

yamill:
I was experience the same exact problem. 

cloneWithRows works fine. but clonesWithRowsandSections doesn't.

sahrens:
Are you seeing cloneWithRowsandSections not working with 0.7?

Also, you might want to try setting initialListSize to a larger number - that might help as a workaround if things aren't fixed in 0.7 for you.

michalraska:
Bigger initialListSize does not help and 0.7 I have not tried yet because vendors dependency. As it will be possible I let you know about 0.7 if somebody will not be quicker :)

coderdave:
Having the same issue with cloneWithRows.

yamill:
Haven't tried it in 0.7. Will give it a try. I know that I also had tried setting the initialListSize to a higher number previously and didn't help.

Shimaa-Hassan:
worked in 0.7

coderdave:
Also not working in 0.7.1

ghislaindj:
Having the same issue with cloneWithRows in version 0.7.1

michalraska:
For me it works well in 0.7.1.

yamill:
Does not work in 0.8.0.

Still using `cloneWithRows` instead of `cloneWithRowsAndSections`

coderdave:
`cloneWithRows` works for you yamill? Doesn't for me in 0.8.0.

yamill:
@coderdave actually `cloneWithRows` doesn't work for me either. my mistake.

yamill:
[...]

Issue #13316, opened 2017-04-05
[FlatList] Sometimes FlatList rows are rendered, but not displayed until scroll. RN 0.43


joonhocho:
### Description

When using FlatList, rows are not displayed though `renderItem` is called.
Rows appear immediately when scroll is triggered on the list.

### Reproduction Steps and Sample Code

```javascript
export default class List extends Component {
  renderItem({item}) {
    return (
      <Text>{item}</Text>
    );
  }

  render() {
    return (
      <FlatList
        style={{
          flex: 1,
          backgroundColor: 'white',
        }}
        data={[1, 2]}
        renderItem={this.renderItem}
      />
    );
  }
}
```

### Solution

FlatList should always display its rendered items.

### Additional Information

* React Native version: 0.43.0
* Platform: iOS
* Development Operating System: MacOS
* Dev tools: Xcode

joonhocho:
Related https://github.com/facebook/react-native/issues/13202

joshyhargreaves:
@joonhocho could you create a gif of this in action? I can't reproduce with your example.

joshyhargreaves:
Are you experiencing this issue in a standalone project and/or are you able to reproduce in a standalone project if not?

Thomas101:
We're also seeing this issue. Looks like it could be something to do with the FlatList being in a navigator on iOS. Managed to reproduce it in a standalone project. You'll have to excuse the code it's a bit rough.

react-native init and add the following code as your index.ios.js. Once launched, tap on A. Tap remove. Tap back. Then you'll need to scroll to get FlatList to re-render.

We found this on RN0.43.1 iOS simulator, iPad but not android. Seems to work as expected on android


```js
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 * @flow
 */

import React, { Component } from 'react';
import { AppRegistry, StyleSheet, Text, View, FlatList, TouchableOpacity, Navigator } from 'react-native';

let data = [ { id: 'a' }, { id: 'b' }, { id: 'c' } ]
let listeners = []

class List extends Component {
  constructor (props) {
    super(props)
    this.state = { data: Array.from(data) }
  }

  componentDidMount () {
    listeners.push(() => { this.setState({ data: data }) })
  }

  render () {
    return (
      <FlatList
        data={this.state.data}
        keyExtractor={(item) => item.id}
        renderItem={({item}) => (
[...]

Issue #8607, opened 2016-07-06
ListView (or perhaps ScrollView) disappears after running in background for several minutes


nihgwu:
I'm using the ScrollView as a `ViewPager` and placing ListView in it just like the F8 app does, and since I upgrade the RN to 0.29 and now 0.30RC0, I'm persecuted by this problem: 

navigate to the detail view from the list view ,then press the home button to make the app running in background, and then wait for several minutes to wake up the app, navigate back to the list view, the ListView (or perhaps or outer ScrollView) just disappears until I scroll it vertically or horizontally ( tap has no effect ) 

I have no idea what happened as it's OK before RN0.29, I've tested on all the 0.29 versions and the newest 0.30RC version they are all the same.

here is the demo:
![listview](https://cloud.githubusercontent.com/assets/2595058/16626480/a1204f88-43db-11e6-91bc-5fcfb69aaeb9.gif)

nihgwu:
I tried to revert the commit https://github.com/facebook/react-native/commit/329c7168975a508de535ea7809ddd1d8c6ab5455 manually, and now the issue occurs every time, no need for background waiting
![listview1](https://cloud.githubusercontent.com/assets/2595058/16642929/28a2ec7a-4442-11e6-9f7f-fc6f7586e793.gif)

@nicklockwood will you have any time to check it? I'm trying to solve this issue to summit the new version to App Store, thanks

 I'll try to revert another commit https://github.com/facebook/react-native/commit/1048e5d3445094393298d4e818ff04c41f4e56a7

nihgwu:
still the save after turn off `removeClippedSubviews`, the `ScrollView` went to white until scroll

satya164:
cc @nicklockwood

winterbe:
I'm encountering a similar problem after migrating RN from 0.28 to 0.29. I have a ListView off-screen which gets updated (render is called) due to an action occurring in another view. When navigating back to the ListView all items are invisible until I scroll the View up or down. The items seem to re-appear from the top of the screen.

mroswald:
We had a similar problem with a ListView on a inactive route which did not render correctly. `removeClippedSubviews={false}` did the trick. But I didn't find out what was changed between 0.25.1 and 0.29.0

mroswald:
[...]

PR #15669, opened 2017-08-27
Fixes removeClipSubviews check for offscreen rendering of ListViews

MattFoley:
This issue has been open for a really long time, but I'm pretty sure this is the line that needed to change:
https://github.com/facebook/react-native/issues/1831

What was happening here is that `CGRectIsEmpty` returns true when either height or width is zero. With the current logic, one of those would always be zero when the parent was rendered off screen. This ensures that there the intersection be of CGSizeZero for the view to actually be clipped. 

That being said, there seems to be something more complex going on here that I'm not understanding? I would think that you'd simply want to check if the child view's frame is within the bounds of the parent at all. If it was, then don't clip it. If I'm in the wrong, could someone explain this a bit more? If so, I'll fix this issue.

## Test Plan

Using this [repository](https://github.com/jcharlet/react_native_listview_bug), this one line change fixes the issue and still clips cells as they are scrolled off screen.


shergin:
@MattFoley Do you have a test case for this?

MattFoley:
@shergin I don't and I'm honestly not sure where to start with that. It looks like there aren't any tests around clipping yet. Do you think you could give me a hand? What would be expected?

shergin:
Starting just from a demo app which did not work properly before would be a good start.

MattFoley:
@shergin That I do have. If you view [this repository](https://github.com/jcharlet/react_native_listview_bug) with no changes, you'll notice that after launch/refresh and tap to the second tab, the cells there do not display on iOS until you scroll them. Once you add my fix, that no longer happens, and it still removes the subviews properly when scrolled off screen. 

The current workaround that I've been using (mentioned in [this issue](https://github.com/facebook/react-native/issues/1831)) is to render the ListView with removeClippedSubviews set to `true`, and then changing this to `false` after it moves into view. Apparently people have found similar workarounds for FlatList and other ScrollView based components, there's a ton of information in that thread. (Hard for me to follow all of it actually)

Does that work?

shergin:
Here the thing, both `ListView` and `removeClipSubviews` are deprecated, and this mean that it is unlikely that someone from core team at Facebook will investigate this pretty magical problem. 😞 
But if someone with good karma assures me that this is a right approach, I will merge this.
I am really sorry, but at this moment we are more affraid to break something stable/old than want to make other old and deprecated thing better.
Anyone? 😄 
cc @janicduplessis

MattFoley:
@shergin Are you sure [removeClippedSubviews](http://facebook.github.io/react-native/docs/view.html#removeclippedsubviews) is deprecated? I don't see this mentioned anywhere in the documentation or code. 

Also, this isn't just a problem with ListView, that issue I linked has reports of people having the same problem with `FlatList`. I would guess this issue effects `ScrollView`, and other Views in some circumstances. (Although, I could totally be wrong)

shergin:
@MattFoley 
>  Are you sure removeClippedSubviews is deprecated? I don't see this mentioned anywhere in the documentation or code.
Technically, it is not deprecated. Yet. But we are thinking about this. We see a lot of issues and negative side effects caused by this feature, and at the same time the positive possible benefits from this is very limited.
This feature will not be removed until it provides some significant benefits though.
We don't use it for <FlatList>, AFAIK. So, it should not be related. (Right?)

MattFoley:
@shergin FlatList is a subclass of VirtualizedList, which is a subclass of ScrollView, which is in turn implemented by RCTScrollView on the native ObjC side, which is then a subclass of RCTView. I'm running a slightly older version of RN, but I think that means that even FlatList would be impacted by issues with the removeClippedSubviews calculation.

Do you guys have a back up in place for removing `removeClippedSubviews`?

Any Table/Cell component needs the ability to pop offscreen views, and reuse them. It's a major performance bottleneck if not. RecyclerView and UITabletView both do this almost automatically for the developer, so it would be a shame if this was removed without a replacement.

chirag04:
related: https://github.com/facebook/react-native/issues/13316

kozillla:
I am experiencing the same issue with FlatList.
 As https://facebook.github.io/react-native/docs/flatlist.html states "Note: may have bugs (missing content) in some circumstances - use at your own risk.This may improve scroll performance for large lists.". 

Is it in any roadmap to when we can expect fix for it ?

nihgwu:
that's really an aged issue, #8607 #1831

shergin:
@MattFoley I am affraid `removingClippingSubviews` feature cannot help with recycling (it cannot reduce memory footprint), it works differently.
Yes, as I mentioned earlier, this feature will not be removed until we have "back up plan".
AFAIK, enabling `removingClippingSubviews` for FlatList does not makes much sense since FlatList does recycle views (so there is no need to remove them from view hierarchy).

MattFoley:
@shergin The issue I linked above still has people running into problems. I get a couple emails a week with replys to that issue. If my PR isn’t sufficient, can a fix for this be prioritised? It seems like a real problem that many people are having with both ListView and FlatList.

lprhodes:
It's a significant issue for every one of our apps. I'm unsure how the issue has gone under the radar for so long.

Using `removeClipSubviews` certainly has an impact on app performance, especially with older devices so I hope it won't be the case that it's removed.

narychen:
@MattFoley It indeed fixes my issue. I'm using the old version 0.27.2 of react-native and cannot update since I changed too much native code. But I have to turn the removeClipSubviews on because if turn it off the app will soon crash due to the memory issue as my app has too many GIF images. 
But after I merged your code and get removeClipSubviews still be true. The ListView rows never will be blank again.
You are amazing. Thank you really very much.

narychen:
@MattFoley It indeed fixes my issue. I'm using the old version 0.27.2 of react-native and cannot update since I changed too much native code. But I have to turn the removeClipSubviews on because if turn it off the app will soon crash due to the memory issue as my app has too many GIF images. 
But after I merged your code and get removeClipSubviews still be true. The ListView rows never will be blank again.
You are amazing. Thank you really very much.

kelset:
I understand that `removingClippingSubviews` will be removed/deprecated soon - but this fix has quite a small footprint; it's more like an hotfix to me, I'd love to see it merged & cherry-picked in 0.49 :3

kozillla:
@MattFoley Thanks a lot for fixing this up. Im using the latest react-native ver 0.48.1 with your fix. It works fine in majority of cases, how ever Im still running into scenario when ListView becomes blank.

sahrens:
Sounds like we should merge this.

Note FlatList doesn't set removeClippedSubviews by default, but developers may set it manually. More importantly, other libraries like navigation my use removeClippedSubviews which causes the proble.

MattFoley:
@sahrens I had assumed it was defaulted to `true` everywhere. I just dug around and found that actually only ListView and WindowedListView set it to `true` by default. Perhaps FlatList has a different issue (or maybe they're just setting it to true for the performance?).

I think I've started understanding the more complex logic in this function. I think it's using the extra intersection logic to compensate for if a parent subview is only partially on screen, but the subview it's checking is still off screen. That must be why it's not just checking if it's within the bounds of it's parent. I'm not sure if that's a worthwhile optimization or not.

What do you think? Would it be better to simply check if a view is within the bounds of it's parent and clip it if it's not?

shergin:
Okay, let's do it!
If someone has troubles because of this change, please let us know.

janiokq:
If it's a performance problem caused by listview
You can try this library
[react-native-nlist](https://github.com/janiokq/react-native-nlist)

PR #15680, opened 2017-08-28
typo alertIOS Docs

adamclark64:
'All your data belongs to us.' instead of 'All your data are belong to us.'

no code was changed


hramos:
Ah, this is just a silly reference to an [old meme](https://en.wikipedia.org/wiki/All_your_base_are_belong_to_us). I believe it's intentional.

PR #15682, opened 2017-08-28
Native Animated - Coalesce onValueChanged events on iOS

janicduplessis:
## Summary

While playing around with the event code I figured we could leverage RCTEvent to easily coalesce onValueChanged events for Native Animated. I meant to do that for a while but didn't knew about a simple solution. This makes it so if JS lags behind we won't spam a bunch on onValueChanged events after.

## Test plan

Tested this fix by logging the number of listener calls every 16ms and make sure it is never more than one when running an animation with a listener in RNTester native animated example. Before the fix when adding some lag to JS the listener would get called 7-8 times after the JS lag ended and after this fix only once.

## Changelog

[iOS] [Changed] - Native Animated - Coalesce onValueChanged events on iOS

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -9,13 +9,13 @@
  */
 'use strict';
 
-import NativeEventEmitter from '../../EventEmitter/NativeEventEmitter';
 import type {
   EventMapping,
   AnimatedNodeConfig,
   AnimatingNodeConfig,
 } from './NativeAnimatedModule';
 import NativeAnimatedModule from './NativeAnimatedModule';
+import RCTDeviceEventEmitter from '../../EventEmitter/RCTDeviceEventEmitter';
 import invariant from 'invariant';
 
 import type {AnimationConfig, EndCallback} from './animations/Animation';
@@ -24,8 +24,6 @@ import type {EventConfig} from './AnimatedEvent';
 
 let __nativeAnimatedNodeTagCount = 1; /* used for animated nodes */
 let __nativeAnimationIdCount = 1; /* used for started animations */
-
-let nativeEventEmitter;
 
 let queueConnections = false;
 let queue = [];
@@ -312,9 +310,6 @@ module.exports = {
   transformDataType,
   // $FlowExpectedError - unsafe getter lint suppresion
   get nativeEventEmitter() {
-    if (!nativeEventEmitter) {
-      nativeEventEmitter = new NativeEventEmitter(NativeAnimatedModule);
-    }
-    return nativeEventEmitter;
+    return RCTDeviceEventEmitter;
   },
 };
--- a/Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
+++ b/Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
@@ -1,10 +1,8 @@
 /**
- * Copyright (c) 2015-present, Facebook, Inc.
- * All rights reserved.
+ * Copyright (c) Facebook, Inc. and its affiliates.
  *
- * This source code is licensed under the BSD-style license found in the
- * LICENSE file in the root directory of this source tree. An additional grant
- * of patent rights can be found in the PATENTS file in the same directory.
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
  */
 
 #import <React/RCTEventDispatcher.h>
--- a/Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
+++ b/Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
@@ -1,0 +1,17 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <React/RCTEventDispatcher.h>
+
+@interface RCTAnimatedValueChangeEvent : NSObject <RCTEvent>
+
+- (instancetype)initWithNodeTag:(NSNumber *)nodeTag
+                          value:(CGFloat)value NS_DESIGNATED_INITIALIZER;
+
+@end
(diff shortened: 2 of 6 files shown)


Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.m
```diff
@@ -0,0 +1,55 @@
...
+@synthesize viewTag = _viewTag;
+@synthesize eventName = _eventName;
+@synthesize coalescingKey = _coalescingKey;
+
+- (instancetype)initWithNodeTag:(NSNumber *)nodeTag
+                          value:(CGFloat)value
+{
+  if ((self = [super init])) {
+    _eventName = @"onAnimatedValueUpdate";
+    _viewTag = nodeTag;
```
janicduplessis: RCTEvent protocol is a bit specific to views events but the coalescing logic can be used for any kind of event. Ideally this would not be called viewTag but for now we use this to coalesce events by node tag. It would have also been possible to use coalescingKey instead.

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -245,9 +245,6 @@ module.exports = {
   assertNativeAnimatedModule,
   shouldUseNativeDriver,
   get nativeEventEmitter() {
-    if (!nativeEventEmitter) {
```
janicduplessis: Sadly we can't use NativeEventEmitter anymore.

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
```diff
@@ -0,0 +1,17 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <React/RCTEventDispatcher.h>
+
+@interface RCTAnimatedValueChangeEvent : NSObject <RCTEvent>
+
+- (instancetype)initWithNodeTag:(NSNumber *)nodeTag
```
shergin: If it is not a `viewTag`, should we don't implement `viewTag` property at all and mark `viewTag` as optional in the protocol.

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.m
```diff
@@ -0,0 +1,55 @@
...
+}
+
+@synthesize viewTag = _viewTag;
+@synthesize eventName = _eventName;
+@synthesize coalescingKey = _coalescingKey;
+
+- (instancetype)initWithNodeTag:(NSNumber *)nodeTag
+                          value:(CGFloat)value
+{
+  if ((self = [super init])) {
```
shergin: We don't need `((` anymore.

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.m
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#import "RCTAnimatedValueChangeEvent.h"
+
+@implementation RCTAnimatedValueChangeEvent
+{
+  CGFloat _value;
+}
+
+@synthesize viewTag = _viewTag;
```
shergin: Why do not declare this as regular `@`-properties?

Libraries/NativeAnimation/RCTNativeAnimatedModule.m
```diff
@@ -22,6 +27,8 @@ @implementation RCTNativeAnimatedModule
   NSMutableArray<AnimatedOperation> *_preOperations;
 }
 
+@synthesize bridge = _bridge;
```
shergin: Just serious, is this a special reason to use `syntesize` here? Why should not use just a ivar?

shergin: Or why don't add this as a prop to class interface inside *this* (.m) file?
To me, `@syntesize` always looks like or something redundant/old-fasioned or something very specific.

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.m
shergin: How can we make it less specific or simply better? Let's do this. 😄

shergin:
Looks awesome, actually.

Libraries/NativeAnimation/RCTNativeAnimatedModule.m
janicduplessis: From what I  understand @synthesize is used to implement properties from a protocol without having to redefine them (https://github.com/facebook/react-native/blob/master/React/Base/RCTBridgeModule.h#L87). It's pretty much used for every native module that needs to access the bridge.

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.m
janicduplessis: Same as the _bridge for native modules, it implements properties from the RCTEvent protocol. Based this on other events (https://github.com/facebook/react-native/blob/master/React/Views/RCTScrollView.m#L51)

Libraries/NativeAnimation/RCTAnimatedValueChangeEvent.h
janicduplessis: This should be possible, we only use viewTag to generate an id for the event based on viewTag + eventName + coalescingKey. We could make view tag optional for events that are not related to views and only use coalescingKey to identify events. (https://github.com/facebook/react-native/blob/master/React/Base/RCTEventDispatcher.m#L34)

hramos:
Sorry it's taken so long - do you still want to get this in?

hramos:
Should we continue working on this PR?

cpojer:
Hey @janicduplessis, just asking again whether we still want to do this? If you aren't interested in moving further on this PR I think we should close it soon. Let us know!

janicduplessis:
Going to close this for now, if anyone has time feel free to pick this up.

hramos:
It looks like this PR was blocked on getting https://github.com/facebook/react-native/pull/15894 merged. https://github.com/facebook/react-native/pull/15894 just got approved, let us know if you want to revive this one.

janicduplessis:
This PR is actually pretty land-able, I'll have some time to work on it soon.

cpojer:
Ping?

cpojer:
@janicduplessis any chance you could do a pass on this PR so we can finally get it in? :)

janicduplessis:
@cpojer But I like having the oldest PR in the repo

Actually I still do after this lands 😄 

Rebased and tested again, surprisingly this still works 2 years later.

cpojer:
As discussed on Discord, we need to make sure that this is backwards compatible over the JS<->native boundary.

janicduplessis:
Ok I don't have time to work on this so closing for now.

PR #15685, opened 2017-08-29
Alert - allow for hiding alert title on iOS

sterlingwes:
## Why
Wanted a "message-only" alert on iOS that doesn't lead to the message being large & bolded via the title field (current state).

Before:

screen shot 2017-08-28 at 9 22 04 pm

After:

screen shot 2017-08-28 at 9 26 56 pm

It also aligns iOS with Android's current behaviour.

The above screenhots were generated from the `RNTester` example added herein.

## What Changed
Allowed for passing an empty string through to `UIAlertController.alertControllerWithTitle` so that the message could be rendered in its expected place on iOS (not as the title).

## Test Plan
* Ran RNTester & compared example alerts before & after (below default with titles, for example)
* Ran end-to-end manual test suite (it's the only one I could get through without unrelated failures)

Before

screen shot 2017-08-28 at 9 21 53 pm

After

screen shot 2017-08-28 at 9 26 40 pm

### DIFF
--- a/Libraries/Alert/AlertIOS.js
+++ b/Libraries/Alert/AlertIOS.js
@@ -209,7 +209,7 @@ class AlertIOS {
       var callback = type;
       var defaultValue = message;
       RCTAlertManager.alertWithArgs({
-        title: title || undefined,
+        title: title,
         type: 'plain-text',
         defaultValue,
       }, (id, value) => {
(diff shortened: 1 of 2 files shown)


11nc:
Unfortunately the flow types suggest that passing a null title to Alert, AlertIOS, and AlertAndroid is valid and there are probably a number of cases in the wild where that is done.  If you do pass null this will crash in RCTAlertManager because it will have an NSNull in the args NSDictionary instead of a nil.

sterlingwes:
Thanks @11nc - I imagine for backwards compatibility we should probably explicitly check for nulls and instead pass in undefined rather than amend the flow types. I can update if that seems reasonable

Issue #15659, opened 2017-08-25
[Website] Add "- React Native" to page titles


sophiebits:
If you search on Google for accessibility, the React Native page just says "Accessibility - Facebook Code":

![image](https://user-images.githubusercontent.com/6820/29734502-6aff128c-89a7-11e7-990f-779398292b92.png)

This is because the document title is just "Accessibility". Let's add "- React Native" to all our page titles (except the homepage itself, which already has it) which will make the results easier to distinguish.

Gabrielglvr:
I want to work on this but i thing that i'm not understand as well, is just about change the doc .md file title or .html pages?  I'm not found the html pages in react-native repository.

sophiebits:
This is just about changing the .html files, which are autogenerated from .md. Try looking in website/layout/ for clues.

Gabrielglvr:
Thank you for help !

TheSavior:
The fix landed. Thanks!

PR #15714, opened 2017-08-30
Add "- React Native" to all pages title in doc section

Gabrielglvr:
Changing all pages in doc to have title " - React Native", which will make the results easier to distinguish like suggested in issue #15659.

## Test Plan

I'm not shure how to write a test to this but if you open any page in doc you can see the page title has changed when the title is passed to Site component.


TheSavior:
Thanks for contributing!

PR #15733, opened 2017-09-01
Added in missing ReactProp in TextInput for Android

cheriejw:
The missing autoComplete is the reason why some Samsung and HTC can't turn off autoSuggestion/autoCompletion successfully. 

While testing a mobile application that prioritizes security, it was noted that on some Samsung and HTC devices the TextInput component (username field) would still auto-suggest even with autoCorrect and autoComplete props set to false.

[Here is a link to a SO question that had the same issue.](https://stackoverflow.com/questions/37001070/how-to-avoid-the-suggestions-of-keyboard-for-android-in-react-native)

## Test Plan

This solution was tested on a Samsung phone that was previously auto-suggesting with props set.

Planning to test this on more HTC and Samsung phones.


cheriejw:
Didn't fix all keyboards.

PR #15755, opened 2017-09-01
Fix unrecognized selector error

giladno:
Fixes "[NSTaggedPointerString unsignedIntValue]: unrecognized selector" error

### DIFF
--- a/React/Base/RCTMultipartStreamReader.m
+++ b/React/Base/RCTMultipartStreamReader.m
@@ -76,7 +76,7 @@
   // Throttle progress events so we don't send more that around 60 per second.
   CFTimeInterval currentTime = CACurrentMediaTime();
 
-  NSUInteger headersContentLength = headers[@"Content-Length"] != nil ? (NSUInteger)[headers[@"Content-Length"] intValue] : 0;
+  NSInteger headersContentLength = headers[@"Content-Length"] != nil ? [headers[@"Content-Length"] intValue] : 0;
   if (callback && (currentTime - _lastDownloadProgress > 0.016 || final)) {
     _lastDownloadProgress = currentTime;
     callback(headers, @(headersContentLength), @(contentLength));
--- a/React/Base/RCTMultipartStreamReader.m
+++ b/React/Base/RCTMultipartStreamReader.m
@@ -76,7 +76,7 @@
   // Throttle progress events so we don't send more that around 60 per second.
   CFTimeInterval currentTime = CACurrentMediaTime();
 
-  NSUInteger headersContentLength = headers[@"Content-Length"] != nil ? [headers[@"Content-Length"] unsignedIntValue] : 0;
+  NSUInteger headersContentLength = headers[@"Content-Length"] != nil ? (NSUInteger)[headers[@"Content-Length"] intValue] : 0;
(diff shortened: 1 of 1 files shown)


chirag04:
cc @gre @janicduplessis

React/Base/RCTMultipartStreamReader.m
```diff
@@ -76,7 +76,7 @@ - (void)emitProgress:(NSDictionary *)headers
   // Throttle progress events so we don't send more that around 60 per second.
   CFTimeInterval currentTime = CACurrentMediaTime();
 
-  NSUInteger headersContentLength = headers[@"Content-Length"] != nil ? [headers[@"Content-Length"] unsignedIntValue] : 0;
```
chirag04: `unsignedIntegerValue` maybe?

gre:
It works! I don't think you need the extra cast tho. just `[headers[@"Content-Length"] intValue]` works ;)

gre:
`unsignedIntegerValue` seems to break too. weird that XCode seems to say it's fine but it breaks at runtime.

giladno:
@gre the extra casting is only to avoid compilation warning by xcode

giladno:
@gre unsignedIntegerValue is a selector of NSNumber, not NSString

gre:
@giladno , mmh look again the code:

```
(NSUInteger)[headers[@"Content-Length"] intValue]
```

you cast the result of the whole thing, not of the header.

intValue is on NSString too

gre:
I don't think the cast is helping anything because it's a `NSUInteger headersContentLength` variable anyway ?

React/Base/RCTMultipartStreamReader.m
gre: unfortunately no, both `unsignedIntValue` and `unsignedIntegerValue` breaks (at runtime)

giladno:
@gre you assign an int to unsigned int - xcode will complain unless you cast. This is only to avoid the warning, it won't break the compilation. The casting is optional...

gre:
oops sorry, I didn't follow the conversation clearly :) ok for the warns ;)

usakc:
Hi,
here is quick fix for this error:
`"postinstall": "sed -i '' 's/\\[headers\\[@\"Content-Length\"\\] unsignedIntValue\\]/\\[headers\\[@\"Content-Length\"\\] intValue\\]/' ./node_modules/react-native/React/Base/RCTMultipartStreamReader.m"`
Add it to "scripts" in package.json in your project.

Hope this help someone :).

React/Base/RCTMultipartStreamReader.m
usakc: I think that the statement can be changed to
NSInteger headersContentLength = headers[@"Content-Length"] != nil ? [headers[@"Content-Length"] intValue] : 0; so the value didn't have to be cast.

janicduplessis:
I'm confused about how this worked for me... This being should be just fine.

ide:
For some reason the value in the dictionary is an NSString, and NSStrings don't define unsignedIntValue.

BTW this should be either `NSInteger = value.integerValue` or `int = value.intValue` to make the types line up.

shergin:
@ide Yeah, you are right. PR is welcome. :)

PR #15760, opened 2017-09-01
Fixed issue with path with spaces

gabrielbull:
@shergin This is a fix I [previously submitted](https://github.com/facebook/react-native/commit/95d5d112acb452c6cacf23d2e2b3d56110f4fefd#diff-883359f85083d00b7266ec2acebcca9fL3494) but was somehow overwritten?


shergin:
Looks like that previous commit fixed different script/configuration.

PR #15761, opened 2017-09-01
Fix scroll events getting skipped on Android

janicduplessis:
This code related to velocity would cause some scroll events to be skipped and caused some jitter for sticky headers. Not sure if there is a better fix but removing this does fix missing events and the velocity calculation still seems to be working.

**Test plan**
Tested that sticky headers now work properly on Android, tested that velocity calculation still seem to work.


janicduplessis:
cc @sahrens

nihgwu:
@grabbou plz cherry-pick this into v0.47 & 0.48 & 0.49rc0, thanks

maxim-c:
hm seems nothing change. i'm using 0.48.1 with this change included. sticky headers still jitter.

janicduplessis:
It's in 0.48.2

PR #15783, opened 2017-09-04
Update RCTTVRemoteHandler.m

iegik:
Want to fix build on 0.48.1

According by comment #13198 (comment) - will fix building project.

To make React Native play nicely with our internal build infrastructure we need to properly namespace all of our header includes.

Where previously you could do #import "RCTBridge.h", you must now write this as #import <React/RCTBridge.h>. If your xcode project still has a custom header include path, both variants will likely continue to work, but for new projects, we're defaulting the header include path to $(BUILT_PRODUCTS_DIR)/usr/local/include, where the React and CSSLayout targets will copy a subset of headers too. To make Xcode copy headers phase work properly, you may need to add React as an explicit dependency to your app's scheme and disable "parallelize build".
 
## Test Plan

Build test app, which uses this part. Build should pass.

### DIFF
--- a/React/Base/RCTTVRemoteHandler.m
+++ b/React/Base/RCTTVRemoteHandler.m
@@ -7,24 +7,22 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import "RCTTVRemoteHandler.h"
+#import <React/RCTTVRemoteHandler.h>
 
 #import <UIKit/UIGestureRecognizerSubclass.h>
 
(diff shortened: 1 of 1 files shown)


hramos:
Do you have a test plan?

iegik:
Did You ever seen Braking Changes at 0.40.0 ? https://github.com/facebook/react-native/releases/tag/v0.40.0
Well done. You did it - You broke it.

Issue #11725, opened 2017-01-04
Duplicate interface definition for class 'RCTBridge'


brunolemos:
### Description

Can't build project after upgrading to 0.40.

PS1: Upgraded from 0.39 using react-native-git-upgrade.
PS2: I compared the diff with this [0.39..0.40 diff](https://github.com/ncuillery/rn-diff/compare/rn-0.39.2...rn-0.40.0) and they are identical, so the upgrade went fine.

### Additional Information

* React Native version: 0.40
* Platform: iOS
* Operating System: macOS 10.12.13

### Screenshot

[![screenshot](https://s30.postimg.org/4rgmzr0a9/Screen_Shot_2017_01_04_at_19_15_33.png)](https://postimg.org/image/sv7eo1ir1/)

DesignMonkey:
Same happens with new installation with all the RN packages imported with react-native link

brunolemos:
This was probably introduced by the [breaking change on 0.40](https://github.com/facebook/react-native/releases/tag/v0.40.0) that requires a "React/" namespace on header paths. Apparently all native modules will have to upgrade themselves to support RN >= 0.40, so... we better send them PRs!

ericvicenti:
Yeah it seems that `react-native-webrct` hasn't upgraded for v0.40 yet. Sorry about the inconvenience

See here: https://github.com/oney/react-native-webrtc/search?utf8=%E2%9C%93&q=import+%22

A bunch of these need to be converted. If it says `#import "RCTConvert.h"`, it needs to change to `#import <React/RCTConvert.h>`

xuechinahb:
same problem to me!

Amurmurmur:
Probem with fbsdk, its the same :(

SMJ93:
I tried what @ericvicenti said above and changed all `#import "RCTConvert.h"`  to `#import <React/RCTConvert.h>`, but I am still getting the error:
`/node_modules/react-native/React/Base/RCTBridge.h:65:1: Duplicate interface definition for class 'RCTBridge'`
 This is for https://github.com/wix/react-native-navigation.

nihgwu:
@SMJ93  remove all the React Native related `Header Search Paths` in 3rd party libraries

SMJ93:
@nihgwu my `header search paths` under `Build Settings` is empty, but I am still getting `Duplicate interface definition for class 'RCTBridge'` error.

SMJ93:
[...]

PR #15784, opened 2017-09-04
Update RCTUIManager.m

iegik:
To make React Native play nicely with our internal build infrastructure we need to properly namespace all of our header includes.

Where previously you could do `#import "RCTBridge.h"`, you must now write this as `#import <React/RCTBridge.h>`. If your xcode project still has a custom header include path, both variants will likely continue to work, but for new projects, we're defaulting the header include path to `$(BUILT_PRODUCTS_DIR)/usr/local/include`, where the React and CSSLayout targets will copy a subset of headers too. To make Xcode copy headers phase work properly, you may need to add React as an explicit dependency to your app's scheme and disable "parallelize build".

## Test Plan

### DIFF
--- a/React/Modules/RCTUIManager.m
+++ b/React/Modules/RCTUIManager.m
@@ -7,39 +7,39 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import "RCTUIManager.h"
+#import <React/RCTUIManager.h>
 
 #import <AVFoundation/AVFoundation.h>
 
 #import <yoga/Yoga.h>
 
-#import "RCTAccessibilityManager.h"
-#import "RCTAnimationType.h"
-#import "RCTAssert.h"
-#import "RCTBridge+Private.h"
-#import "RCTBridge.h"
-#import "RCTComponent.h"
-#import "RCTComponentData.h"
-#import "RCTConvert.h"
-#import "RCTDefines.h"
-#import "RCTEventDispatcher.h"
-#import "RCTLayoutAnimation.h"
-#import "RCTLayoutAnimationGroup.h"
-#import "RCTLog.h"
-#import "RCTModuleData.h"
-#import "RCTModuleMethod.h"
-#import "RCTProfile.h"
-#import "RCTRootContentView.h"
-#import "RCTRootShadowView.h"
-#import "RCTRootViewInternal.h"
-#import "RCTScrollableProtocol.h"
-#import "RCTShadowView+Internal.h"
-#import "RCTShadowView.h"
-#import "RCTUIManagerObserverCoordinator.h"
-#import "RCTUtils.h"
-#import "RCTView.h"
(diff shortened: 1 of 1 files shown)


hramos:
Do you have a test plan?

shergin:
This is a very radical and breaking change which conflicts with current established guidelines and practices.
If you are strongly believe that we should reconsider it, please open an issue and provide detailed motivation and strong argumentation there.

iegik:
Sorry, @shergin, but it was not my opinion. https://github.com/facebook/react-native/releases/tag/v0.40.0

> iOS native headers moved
>
> This change affects all native modules and application code on iOS that refer to react native .h files
> After e1577df, Native code on iOS must refer to headers out of the react namespace. Previously the following would work:
>
> #import "RCTUtils.h"
> But now all headers have been moved:
>
> #import <React/RCTUtils.h>
> This means that all iOS native libraries need a major version bump for RN 0.40. We attempt to minimize changes of this magnitude, and we apologize for any inconvenience caused.

shergin:
That proposal and work is about .h files, not .m or .mm files.
I just want to say that if you want to fight for this, you have to prepare really good arguments and deep and comprehensive analysis of all possible approaches.

iegik:
I notice, if import file once in with `<React/` prefix, next import without prefix cause duplication of declaration error, like now with `RCTBridge.h` https://github.com/facebook/react-native/issues/11725
So, if `RCTBridge.h` was already imported before in some other module with prefix, and here, in *.m file start to importing without prefix, the compilation cause "duplication error".

Solution:
- Compile React files first
- Create list of all React files and describe how they should be imported
- Revert to 0.39.x version of react and use imports without prefix.
- Fix all imports to imports with `<React/` in **all** files

iegik:
@javache , please judge

React/Modules/RCTUIManager.m
```diff
@@ -7,39 +7,39 @@
...
-#import "RCTModuleMethod.h"
-#import "RCTProfile.h"
-#import "RCTRootContentView.h"
-#import "RCTRootShadowView.h"
-#import "RCTRootViewInternal.h"
-#import "RCTScrollableProtocol.h"
-#import "RCTShadowView+Internal.h"
-#import "RCTShadowView.h"
-#import "RCTUIManagerObserverCoordinator.h"
-#import "RCTUtils.h"
```
iegik: https://github.com/facebook/react-native/releases/tag/v0.40.0
> After e1577df, Native code on iOS must refer to headers out of the react namespace. Previously the following would work:
```#import "RCTUtils.h"```
But now all headers have been moved:
```#import <React/RCTUtils.h>```

iegik:
@shergin not only *.h files, but all, that "**refer to** react native .h files":
> all native modules and application code on iOS that **refer to** react native .h files

PR #15788, opened 2017-09-04
Make a forward declaration of RCTMethodInfo

iegik:
> declares an anonymous structure and creates a typedef for it. Thus, with this construct, it doesn't have a name in the tag namespace, only a name in the typedef namespace. This means it also cannot be forward-declared. If you want to make a forward declaration, you have to give it a name in the tag namespace.

https://stackoverflow.com/questions/612328/difference-between-struct-and-typedef-struct-in-c

(Write your motivation here.)

## Test Plan

### DIFF
--- a/React/Base/RCTBridgeModule.h
+++ b/React/Base/RCTBridgeModule.h
@@ -51,11 +51,12 @@ RCT_EXTERN dispatch_queue_t RCTJSThread;
 
 RCT_EXTERN_C_BEGIN
 
-typedef struct RCTMethodInfo {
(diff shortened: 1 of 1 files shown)


hramos:
As with the others, this PR needs a test plan.

shergin:
So, why do we need this? Any real benefits?

iegik:
I have no idea. Please, read carefully Stackoverflow answer. If somebody see the difference - you can apply PR, otherwise - close it w/o merge. Thanks.

Issue #15791, opened 2017-09-04
Content-Type does not respect optional charset encoding for iOS RCTJavascriptLoader


justinr1234:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

1. `react-native -v`: 0.42.0
2. `node -v`: v6.11.2
3. `npm -v`: 4.6.1
4. `yarn --version`: 0.27.5

- Target Platform: iOS
- Development Operating System: macOS
- Build tools: XCode

### Steps to Reproduce

1. Try to run a react native app and download a bundle from a server which returns a `Content-Type` header that includes both the MIME Type and an optional character encoding as specified here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
2. The React Native code for iOS will fail on this line of code as it strictly checks the `Content-Type` header as if the optional character encoding doesn't exist: https://github.com/facebook/react-native/blame/master/React/Base/RCTJavaScriptLoader.mm#L277

### Expected Behavior

The code should follow the specification of `Content-Type` and respect only MIMEType checking instead of strictly checking the `Content-Type` header for an exact match on MIMEType.

Per Apple documentation, `NSURLResponse` contains the `MIMEType` field that should be checked for `application/javascript` instead of checking the raw `Content-Type` header: https://developer.apple.com/documentation/foundation/nsurlresponse

### Actual Behavior

Though this was encountered while using Expo, the error is nonetheless coming from the `RCTJavaScriptLoader`

<img width="384" alt="screen shot 2017-09-04 at 7 52 51 am" src="https://user-images.githubusercontent.com/7017500/30034896-19385ff6-916a-11e7-9a80-d171adf17717.png">

Expected JavaScript, but got content type 'application/javascript; chartset=UTF-8'. Check that your internet connection is working.

### Reproducible Demo

https://github.com/sysgears/apollo-universal-starter-kit

Follow starter instructions and checkout this commit: https://github.com/sysgears/apollo-universal-starter-kit/commit/bbcd52a934e24b4ce873e802d085af2b8fe04ec2

The problem is worked around after this commit in the project.

joearasin:
I just tripped over this when attempting to get Haul to make a build for react-native 0.48.2: https://github.com/callstack-io/haul/issues/228

dikaiosune:
cc'ing @janicduplessis as he was cc'd on an Expo bug report.

[...]

PR #15792, opened 2017-09-04
Fix Content-Type header checking of React/RCTJavascriptLoader.mm #15791

justinr1234:
To fix this issue: https://github.com/facebook/react-native/issues/15791


justinr1234:
Not sure why the build failed, looks to be transient?

justinr1234:
@shergin I can take a look, but can you elaborate on the motivation? It appears your suggestion would be equivalent except requiring additional temporary storage and processing?

janicduplessis:
@shergin @justinr1234 We could also just use a regex, something like `/^(?:application|text)\/javascript/` should do the trick.

justinr1234:
@janicduplessis @shergin I can make changes, I'm just confused what the purpose is and what's gained by changing from the current form. Thanks, it helps me understand going forward to learn the preferences of the project.

ide:
The check in this PR doesn't look at the structure of a MIME type string. It just checks the prefix -- what if you have `text/javascript2`? We don't want to match that unintentionally.

justinr1234:
@ide thanks, good point, I'll update it

shergin:
BTW, I think using a regexp is overkill here.

janicduplessis:
In that case @shergin suggestion to create a substring up to optional ; would work well.

React/Base/RCTJavaScriptLoader.mm
```diff
@@ -274,9 +274,11 @@ static void attemptAsynchronousLoadOfBundleAtURL(NSURL *scriptURL, RCTSourceLoad
 
     // Validate that the packager actually returned javascript.
     NSString *contentType = headers[@"Content-Type"];
-    if (![contentType isEqualToString:@"application/javascript"] &&
-        ![contentType isEqualToString:@"text/javascript"]) {
-      NSString *description = [NSString stringWithFormat:@"Expected Content-Type to be 'application/javascript' or 'text/javascript', but got '%@'.", contentType];
+    NSArray<NSString *> *contentTypeArr = [contentType componentsSeparatedByString:@";"];
```
shergin: Please don't use abbreviations.
`NSString *mimeType = [[contentType componentSeparatedByString:@";"] firstObject]` should be enough, right?

janicduplessis:
@shergin Does this looks good now?

justinr1234:
@shergin did this get merged somewhere?

ide:
The commit above shows where it landed - https://github.com/facebook/react-native/commit/3ae45d5cb8b9aaca568227ca00cc304e053bf073

PR #15806, opened 2017-09-05
[ios] Change intValue -> integerValue in RCTMultipartStreamReader

ide:
We assign the value to an NSInteger so we should convert the NSString to an NSInteger instead of an int.

## Test Plan

Build the app in Xcode and run it.


shergin:
👍

PR #15813, opened 2017-09-05
Update Android Guide for Integration With Existing Apps

GingerBear:
I was trying to follow the instruction on http://facebook.github.io/react-native/docs/integration-with-existing-apps.html to integrate react native into existing android app, but failed. After hours of google/stackoverflow, I finally get it running. I am submitting this pr the list the diffs in the guide so that other developer won't fell into the same issues. 

## Test Plan

- create an example android project
- follow the guide in the pr to integrate react-native into the example android project
- make sure it runs without error

### DIFF
--- a/docs/IntegrationWithExistingApps.md
+++ b/docs/IntegrationWithExistingApps.md
@@ -157,8 +157,10 @@ Go to the root directory for your project and create a new `package.json` file w
 Next, you will install the `react` and `react-native` packages. Open a terminal or command prompt, then navigate to the root directory for your project and type the following commands:
 
 ```
-$ npm install --save react@16.0.0-alpha.12 react-native
+$ npm install --save react@16.0.0-beta.5 react-native
 ```
+
+> Make sure you use the same React version as specified in the [React Native dependencies](https://github.com/facebook/react-native/blob/master/package.json). As of this writing, that would be `16.0.0-beta.5`. This will only be necessary as long as React Native depends on a beta version of React.
 
 This will create a new `/node_modules` folder in your project's root directory. This folder stores all the JavaScript dependencies required to build your project.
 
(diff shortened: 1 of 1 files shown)


docs/IntegrationWithExistingApps.md
```diff
@@ -157,9 +157,11 @@ Go to the root directory for your project and create a new `package.json` file w
 Next, you will install the `react` and `react-native` packages. Open a terminal or command prompt, then navigate to the root directory for your project and type the following commands:
 
 ```
-$ npm install --save react@16.0.0-alpha.12 react-native
+$ npm install --save react@16.0.0-beta.5 react-native
```
GingerBear: @hramos when i use `react-native init` to create a new project, `alpha.12` is the version that actually get installed.

hramos: Gotcha, it's because 0.48 (the latest stable) is still on alpha.12: https://github.com/facebook/react-native/blob/0.48-stable/package.json

I'll need to update the wording here.

hramos: Using beta.5 here is fine, because these are the docs for master and won't be displayed in the current 0.48 docs. We'd need to edit the markdown file in the 0.48 branch directly in this case.

GingerBear: Gotcha, it make sense.

hramos: Fixed in 6488842e1a02501c34c4be0454e6df15009090a3 (0.48, current stable) and 031feb694257d6ffb8aa8fd7b4019d96dac7155a (0.49, next RC).

hramos: Thanks for figuring this out BTW.

PR #15821, opened 2017-09-05
[CI] Settle on Node 4.8.4 for both Circle and Travis.

hramos:
React Native should have no dependencies that require anything newer than Node 4. Let's make sure we test against Node v4 in CI then.

This PR updates both Travis and Circle to use the same Node release: 4.8.4, which is the most recent release of Node 4.x, which is currently in LTS maintenance mode.

# Test Plan

Wait for Travis and Circle to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: objective-c
 osx_image: xcode8.3
 
 install:
-  - nvm install 7
+  - nvm install 4.8.4
   - rm -Rf "${TMPDIR}/jest_preprocess_cache"
   - brew install yarn --ignore-dependencies
   - brew install watchman
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ general:
       - gh-pages # list of branches to ignore
 machine:
   node:
-    version: 6.2.0
+    version: 4.8.4
   environment:
     PATH: "~/$CIRCLE_PROJECT_REPONAME/gradle-2.9/bin:/home/ubuntu/buck/bin:$PATH"
     TERM: "dumb"


ide:
Can we have some CI coverage on the latest version of Node? I'm concerned about breaking things on modern Node too.

Is this also the version of Node that Metro has committed to supporting?

hramos:
Looks like we'll need to work on moving to Circle 2 earlier (assuming that would allow us to test multiple Node versions).

hramos:
As for this PR, it's currently failing on Circle due to the following:

```
148294 error peerinvalid The package eslint-plugin-babel@4.1.2 does not satisfy its siblings' peerDependencies requirements!
148294 error peerinvalid Peer eslint-config-fbjs@1.1.1 wants eslint-plugin-babel@^3.3.0
```

I _think_ that's a npm issue, as I've only been able to reproduce it by running `npm install` using node v4.8.4 and npm v2.15.11. I cannot reproduce after switching to node v6.2.0, which ships with npm v3.8.9.

janicduplessis:
maybe we should drop support for node 4 / npm2, node 6 is the latest LTS

TheSavior:
These errors exist on node 8 but they just warnings now. I think it would be good to understand whether we should just be updating these versions or if there are technical reasons that we have these incompatibilities before dropping support for node 4 based on the responses from the twitter poll a few months ago (and the natural bias that people on twitter are using newer stuff).

janicduplessis:
@hramos Is this PR still valid?

hramos:
Yeah, this can be closed.

We still need to enable Node 4 tests on Circle 2, once the deps issue is resolved.

PR #15825, opened 2017-09-06
fix fontWeight regression

nihgwu:
fix the regression I mentioned in https://github.com/facebook/react-native/pull/15162#issuecomment-319696706

as no one is working on this, I take the step, although I know nothing about Objective C

I find the key point is that the keys in `NSDictionary` are not ordered as presented, it's a hash table, so no grantee on keys order, so I create a new array to do that, then it will check `ultralight` before `light` and `semibold` before `bold`

### DIFF
--- a/React/Views/RCTFont.mm
+++ b/React/Views/RCTFont.mm
@@ -33,42 +33,38 @@
 
 #endif
 
-static NSArray *fontNameOrdering = [NSArray arrayWithObjects:
-    @"normal",
-    @"ultralight",
-    @"thin",
-    @"light",
-    @"regular",
-    @"medium",
-    @"semibold",
-    @"bold",
-    @"heavy",
-    @"black",
-    nil];
-
 typedef CGFloat RCTFontWeight;
 static RCTFontWeight weightOfFont(UIFont *font)
 {
-  static NSDictionary *nameToWeight;
+  static NSArray *fontNames;
+  static NSArray *fontWeights;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-    nameToWeight = @{
-       @"normal": @(UIFontWeightRegular),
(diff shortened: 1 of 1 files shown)


React/Views/RCTFont.mm
```diff
@@ -33,6 +33,19 @@
 
 #endif
 
+static NSArray *fontNameOrdering = [NSArray arrayWithObjects:
```
janicduplessis: Let's initialize this array inside the dispatch once block, just like the `nameToWeight` dictionary.

janicduplessis: Use `@[a, b, c]` shorthand to initialize the array instead of `arrayWithObjects`.

janicduplessis:
cc @shergin 

What do you think about this?

janicduplessis:
@nihgwu Thanks for working on this even if you are not used to working in objc :)

nihgwu:
@janicduplessis thanks for your advice, I intended to ask you to fix it but I don't want to bother you again and again 😄

henrikra:
Isn't map better than two arrays o.O?

janicduplessis:
@henrikra Well it’s either a map + an array or 2 arrays. The 2 arrays avoid having to have the name key twice (one in the array and one as the map key)

shergin:
👍 
@nihgwu Thank you for fixing this.

Why can't we always rely on `UIFontDescriptorTraitsAttribute`?

nihgwu:
That was modified by the mentioned commit, I have no idea

On 7 Sep 2017, at 08:50, Valentin Shergin <notifications@github.com<mailto:notifications@github.com>> wrote:


👍
@nihgwu<https://github.com/nihgwu> Thank you for fixing this.

Why can't we always rely on UIFontDescriptorTraitsAttribute?

PR #15856, opened 2017-09-07
(iOS) Support for DemiBold alias of SemiBold (font-weight 600)

wildseansy:
For more information on font weight naming see [this post](https://www.quora.com/What-is-the-difference-between-Medium-Demi-and-Semibold-fonts).

DemiBold is fairly common in font naming. For example, iOS React Native lacks support of "AvNext-DemiBold".

Also removed warning about `NSUInteger` <=> `NSInteger` comparison, by making `i` an `NSUInteger`

Before and after screenshots:

#### Before fix
before-fix

#### After
after-fix

## Test Plan

Add any DemiBold font to an iOS react native project.  Set fontWeight to `'600'` on a `<Text />` component. The font weight should be applied appropriately.

### DIFF
--- a/React/Views/RCTFont.mm
+++ b/React/Views/RCTFont.mm
@@ -50,6 +50,7 @@ static RCTFontWeight weightOfFont(UIFont *font)
       @"regular",
       @"medium",
       @"semibold",
+      @"demibold",
       @"bold",
       @"heavy",
       @"black"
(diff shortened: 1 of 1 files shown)


ide:
From what I can tell, these are valid aliases. Have you looked into the Android side as well?

wildseansy:
@ide - from what I have read, using numeric font weights ('500', '600', '700'...etc) doesn't work in android. Is this not the case?  In Android, I just explicitly specify `fontFamily` for each weight

wildseansy:
@ide - [here's a post that outlines the discrepancy](https://stackoverflow.com/questions/42318777/font-weight-bug-react-native) between iOS and android when it comes to font weights.  Because there is no font-weight-matching logic on android from a numeric value to a font name, this problem doesn't (yet) apply.  On Android you have to be discrete about each font name, which means demibold will already work (it does for me testing both).

React/Views/RCTFont.mm
```diff
@@ -62,13 +63,14 @@ static RCTFontWeight weightOfFont(UIFont *font)
...
       @(UIFontWeightSemibold),
+      @(UIFontWeightSemibold),
       @(UIFontWeightBold),
       @(UIFontWeightHeavy),
       @(UIFontWeightBlack)
     ];
   });
 
-  for (NSInteger i = 0; i < fontNames.count; i++) {
+  for (NSUInteger i = 0; i < fontNames.count; i++) {
```
shergin: The modern Apple convention is "always use `NSInteger` unless you really need to use `NSUInteger`", we are following it in RN.

PR #15864, opened 2017-09-08
Description about style theme in DatePickerAndroid component

MrVisitor:
I think this is important information


hramos:
@facebook-github-bot no-cla

PR #15874, opened 2017-09-08
 add release notes/changelog requirement to PR Template

tabrindle:
## Motivation
@grabbou has to write release notes for every release, and that's getting ridonc. Have you seen the changelog from 0.48?

PR writers need to do their own, hopefully in a standardized format so it can be scripted. 

## Test Plan

Just a PR template change, just proofreading needed. Please mention if have missed or added any extra Categories, Types, or Where columns. 

Here's a sample:

## Release Notes

[GENERAL] [ENHANCEMENT] [PULL_REQUEST_TEMPLATE.md] - added release notes/changelog requirement to PR Template

### DIFF
--- a/danger/dangerfile.js
+++ b/danger/dangerfile.js
@@ -93,6 +93,35 @@ if (!includesTestPlan && !editsDocs) {
   markdown('@facebook-github-bot label Needs more information');
 }
 
+// Regex looks for given categories, types, a file/framework/component, and a message - broken into 4 capture groups
+const releaseNotesRegex = /\[(ANDROID|CLI|DOCS|GENERAL|IOS|TVOS|WINDOWS)\]\s*?\[(BREAKING|FEATURE|ENHANCEMENT|BUGFIX)\]\s*?\[(.*)\]\s*?\-\s*?(.*)/ig;
+const includesReleaseNotes = danger.github.pr.body.toLowerCase().includes('release notes');
+const correctlyFormattedReleaseNotes = releaseNotesRegex.test(danger.github.pr.body);
+const releaseNotesCaptureGroups = releaseNotesRegex.exec(danger.github.pr.body);
+
+if (!includesReleaseNotes) {
+  const title = ':clipboard: Release Notes';
+  const idea = 'This PR appears to be missing Release Notes.';
+  warn(`${title} - <i>${idea}</i>`);
+  markdown('@facebook-github-bot label Needs more information');
+} else if (!correctlyFormattedReleaseNotes) {
+  const title = ':clipboard: Release Notes';
+  const idea = 'This PR may have incorrectly formatted Release Notes.';
+  warn(`${title} - <i>${idea}</i>`);
+  markdown('@facebook-github-bot label Needs more information');
+} else if (releaseNotesCaptureGroups) {
+  const category = releaseNotesCaptureGroups[1].toLowerCase();
+
+  // Use Release Notes to Tag PRs appropriately
+  if (category === 'ios' ){
+    markdown('@facebook-github-bot label iOS');
+  }
+
+  if (category === 'android' ){
+    markdown('@facebook-github-bot label Android');
+  }
+}
+
 // Tags PRs that have been submitted by a core contributor.
 // TODO: Switch to using an actual MAINTAINERS file.
 const taskforce = fs.readFileSync('../bots/IssueCommands.txt', 'utf8').split('\n')[0].split(':')[1];
--- a/danger/dangerfile.js
+++ b/danger/dangerfile.js
@@ -94,7 +94,7 @@ if (!includesTestPlan && !editsDocs) {
 }
 
 // Regex looks for given categories, types, a file/framework/component, and a message - broken into 4 capture groups
-const releaseNotesRegex = /\[(ANDROID|CLI|DOCS|GENERAL|IOS|TVOS|WINDOWS)\]\s*?\[(BREAKING|FEATURE|ENHANCEMENT|BUGFIX)\]\s*?\[(.*)\]\s*?\-\s*?(.*)/ig;
+const releaseNotesRegex = /\[(ANDROID|CLI|DOCS|GENERAL|INTERNAL|IOS|TVOS|WINDOWS)\]\s*?\[(BREAKING|BUGFIX|ENHANCEMENT|FEATURE|MINOR)\]\s*?\[(.*)\]\s*?\-\s*?(.*)/ig;
 const includesReleaseNotes = danger.github.pr.body.toLowerCase().includes('release notes');
 const correctlyFormattedReleaseNotes = releaseNotesRegex.test(danger.github.pr.body);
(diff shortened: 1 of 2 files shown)


hramos:
This is awesome. @grabbou, do you have any feedback?

@tabrindle do you think you can update the Danger rules in the dangerfile so that any PR that does not include release notes gets flagged automatically? We already have a rule in place that flags PRs without test plans, so it should be straightforward to add a similar check for release notes.

tabrindle:
Never done anything with danger but I'm sure I can figure it out. Will update.

tabrindle:
Added some verification of existence and format. Format might be overboard, but if we want to script the release notes in the future, it's probably important. While I was at it and capturing the info from the regex,  I added tags for iOS and Android when appropriate.

janicduplessis:
I think there should be a way to opt out of release notes, I feel like some commits are not worth mentioning and just add noise to release notes. Things like internal refactorings, minor optimisations. People will be interested in new features, bug fixes and changes to the public api. Other internal changes are not very useful.

janicduplessis:
@hramos It would be nice to also try to enforce this for fb employees, at least the RN team otherwise release notes will miss a lot of stuff.

danger/dangerfile.js
```diff
@@ -93,6 +93,35 @@ if (!includesTestPlan && !editsDocs) {
...
+  const title = ':clipboard: Release Notes';
+  const idea = 'This PR may have incorrectly formatted Release Notes.';
+  warn(`${title} - <i>${idea}</i>`);
+  markdown('@facebook-github-bot label Needs more information');
+} else if (releaseNotesCaptureGroups) {
+  const category = releaseNotesCaptureGroups[1].toLowerCase();
+
+  // Use Release Notes to Tag PRs appropriately
+  if (category === 'ios' ){
+    markdown('@facebook-github-bot label iOS');
```
hramos: Good idea!

hramos:
Good point. We can make it clear in the template that release notes are only encouraged when the change warrants it. We don't need a long explanation in the template itself, we can update the contribution guide with more details if we need to.

I can work on doing the same for the internal template, though we probably follow a different convention. I'll follow up on that.

danger/dangerfile.js
```diff
@@ -93,6 +93,35 @@ if (!includesTestPlan && !editsDocs) {
...
+// Regex looks for given categories, types, a file/framework/component, and a message - broken into 4 capture groups
+const releaseNotesRegex = /\[(ANDROID|CLI|DOCS|GENERAL|IOS|TVOS|WINDOWS)\]\s*?\[(BREAKING|FEATURE|ENHANCEMENT|BUGFIX)\]\s*?\[(.*)\]\s*?\-\s*?(.*)/ig;
+const includesReleaseNotes = danger.github.pr.body.toLowerCase().includes('release notes');
+const correctlyFormattedReleaseNotes = releaseNotesRegex.test(danger.github.pr.body);
+const releaseNotesCaptureGroups = releaseNotesRegex.exec(danger.github.pr.body);
+
+if (!includesReleaseNotes) {
+  const title = ':clipboard: Release Notes';
+  const idea = 'This PR appears to be missing Release Notes.';
+  warn(`${title} - <i>${idea}</i>`);
```
hramos: Maybe use message here and below, since a PR may not necessarily _require_ release notes.

tabrindle:
We could come up with some criteria on when release notes would be required.

- When changes add/remove X number of lines
- When X number of files are added or removed
- When certain important files are changed

However, I would suggest we go a different route; simply add categories and types that would be ignored for the final Release notes. For example, we could add [INTERNAL] and [MINOR]. This way all PRs will have notes, but not all will be shown, reducing noise. 

Letting people opt out of this kind of convention may cause more problems than it solves. If a PR writer incorrectly assigns a category, correctly assigning it will then be part of the PR process, moving all this sorting work to per PR rather than at release time. 

The only drawback on the second option to me would be that we will need to be very attentive to the PR process and make sure the rules we set up work for everyone.

janicduplessis:
@tabrindle Opt out of release notes should definetly be explicit, not providing the release notes section should still flag the PR as missing release notes. I like your example of [INTERNAL] category.

grabbou:
Most of the commits that are troublesome are ones coming directly from Facebook sync script. That said, I am not sure if any sort of open source guidelines we implement here can affect their internal process. 

In general, I love the idea of this PR and I've been trying to categorise commits manually so far. However, we need to work closely with @hramos on that to investigate whether it's possible to come up with an unified naming technology that works across Facebook and OSS.

Another idea was to improve existing releases script and check every commit for files touched. That way, you could easily group by type of files affected (e.g. `js` only - that's "all", `java` - that's Android etc.)

tabrindle:
> I am not sure if any sort of open source guidelines we implement here can affect their internal process.

Well, we can make this awesome and they might adopt it or some facsimile of it in the future. In mean time, it should help the process here at least some. 

I will add the INTERNAL and MINOR tags for the release notes section in the PR template, to denote an explicit opt out of adding noise to the final notes. Ill also add a blurb in the PR template showing thats what they do. 

@grabbou Thats sounds like a good idea - is that a script that lives in this repo or would this be something I do in the dangerfile?

danger/dangerfile.js
tabrindle: I think we will require them in the PR template, but opt out of adding them to the final notes via keyword as discussed below. That ok?

tabrindle:
@hramos @grabbou @janicduplessis Is there anything else I need to add before we merge this? I believe I have addressed everything still relevant.

hramos:
I think the main blocker here is that most commits come from within Facebook, and asking the community to follow these guidelines may not do as much to help with the release note process.

Has anyone looked at the commits coming from Facebook? Is there a pattern already in place that we can adopt?

If there is none, I can work on the Facebook side and see if we can find some common ground. I have to say out front that I probably will not be able to look into this for a few more weeks due to other priorities I have at the moment.

Copying @shergin @TheSavior from the React Native team as FYI.

TheSavior:
AFAIK, there is nothing on the internal side right now around this.

tabrindle:
@hramos @grabbou @janicduplessis 

I don't see any harm in merging this and getting facebook internal onboard afterwards. Partial adherence is better than zero. 

If any of the internal team wants to chime in with changes, I am certainly open to that as well.

grabbou:
I am okay with merging, leaving this up to @hramos 

> Has anyone looked at the commits coming from Facebook? Is there a pattern already in place that we can adopt?

There's indeed one pattern that Metro Bundler and FlatList team were utilising, e.g.:

```
Packager: Server: URLRequest: Implement new feature
```

or

```
FlatList: Row: Fix index
```

It obviously misses a lot of important informations, but it's a good starting point. The biggest problem for me, that takes most of time, is to group pull request by platform (common, android, ios). That requires opening each PR and checking diff. Once that's done, it's generally easier.

grabbou:
Great. I'll go ahead and adopt my changelog generator to take this into
account. I'll actually merge it to this repo as a PR so it's useful for
others in the future.

PR #15869, opened 2017-09-08
Add Housing.com to showcase

f0rr0:
https://medium.com/engineering-housing/how-we-built-our-react-native-app-3380a33811ac

PR #15876, opened 2017-09-09
add JD Finance（京东金融） to featured list in showcase

zhonshu:


acurat:
See @hramos 's response on a similar PR here: https://github.com/facebook/react-native/pull/15869#issuecomment-328140929

PR #15881, opened 2017-09-10
Fix error in flow types

anta-semenov:
Previously type Options had excludeActivityTypes property, but in RCTActionSheetManager.m and docs this property is named excludedActivityTypes.

Want to fix error in flow types

## Test Plan

Actually, it seems there aren't any need for test plan for this PR. Just fix a mistyping error.


ide:
@facebook-github-bot shipit

PR #15887, opened 2017-09-11
Android: Fix Image.getSize to be memory friendly

bood:
Original implementation will need to load the full image into memory,
which will cause out of memory error if the image is large in size.

This is a better approach for PR #14358 

(Write your motivation here.)

## Test Plan

1. Call `Image.getSize` on following type of images (large in width and height):
    * Remote image on web
    * Local resource image
    * Local asset image
    * Local `file://` image
    * Local `content://` image
2. Watch whether there is a peek on memory usage (should be no peek at all)

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
@@ -9,31 +9,39 @@
 
 package com.facebook.react.modules.image;
 
-import javax.annotation.Nullable;
-
+import android.content.Context;
 import android.net.Uri;
+import android.os.AsyncTask;
+import android.util.Pair;
 import android.util.SparseArray;
 
 import com.facebook.common.executors.CallerThreadExecutor;
-import com.facebook.common.references.CloseableReference;
+import com.facebook.common.util.UriUtil;
 import com.facebook.datasource.BaseDataSubscriber;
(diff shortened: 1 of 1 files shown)


janicduplessis:
@bood Thanks for the PR! Have you tried using fetchEncodedImage http://frescolib.org/docs/using-image-pipeline.html (see Skipping the decode section). It would be nice not to have to reimplement the InputStream logic to fetch the image.

bood:
@janicduplessis `fetchEncodedImage` has limitations, see my comments in PR #14358

ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
```diff
@@ -79,45 +87,60 @@ public void getSize(
...
+    }
+
+    @Override
+    protected void doInBackgroundGuarded(Void... params) {
+      InputStream inputStream = null;
+      try {
+        if (UriUtil.isLocalFileUri(mUri) || UriUtil.isLocalContentUri(mUri)) {
+          inputStream = mContext.getContentResolver().openInputStream(mUri);
+        } else if (UriUtil.isNetworkUri(mUri)) {
+          inputStream = new URL(mUri.toString()).openStream();
```
foghina: This is not nice. This is bypassing Fresco's network layer (e.g. OkHttp).

PR #15891, opened 2017-09-11
Add missing delete to requestIdleCallback

esprehn:
This line accidentally became a no-op when the Map was converted to an Object.


janicduplessis:
@facebook-github-bot shipit

PR #15894, opened 2017-09-11
BREAKING - RCTEvent improvements, remove deprecated [sendInputEventWithName:body:]

janicduplessis:
This makes the RCTEvent protocol more generic to make it easier to use the event coalescing feature for type of events other than components. This does a few other improvements that will be useful in follow up PRs.

- Add `RCTComponentEvent` which is used instead of deprecated `[sendInputEventWithName:body:]` and remove that method completely (was only used at 2 places).
- Make `coalescingKey` optional for events that return NO from `canCoalesce`.
- Make `viewTag` optional for events that are not related to views.
- Fast path for events that return NO from `canCoalesce`.
- Add a missing test for event coalescing with different view tags.

Ended up making only one PR for all this since the changes are related and hard to separate.

**Migration**
Use a custom RCTEvent subclass with `[sendEvent:]` (preferred way to allow type safe events) or `RCTComponentEvent`.

## Changelog:

[iOS] [Changed] - Remove deprecated RCTEvent method, sendInputEventWithName:body:

## Test plan:

- Ran RCTEventDispatcher unit tests
- Tested manually in RNTester


React/Base/RCTComponentEvent.m
```diff
@@ -0,0 +1,57 @@
...
+
+@synthesize eventName = _eventName;
+@synthesize viewTag = _viewTag;
+
+- (instancetype)initWithName:(NSString *)name body:(NSDictionary *)body
+{
+  if (self = [super init]) {
+    NSNumber *target = body[@"target"];
+    name = RCTNormalizeInputEventName(name);
+
```
shergin: `#ifdef` ?

React/Base/RCTComponentEvent.h
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <React/RCTEventDispatcher.h>
+
+/**
+ * Event that is dispatched to React components.
```
shergin: Should we discribe here that it is general purpose event? Should/can it be subclasses?

React/Base/RCTComponentEvent.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+@interface RCTComponentEvent : NSObject<RCTEvent>
```
shergin: Should we add a comment here that this type of Event does not have `reactTag`?

React/Base/RCTEventDispatcher.h
```diff
@@ -37,18 +37,25 @@ RCT_EXTERN NSString *RCTNormalizeInputEventName(NSString *eventName);
...
 + (NSString *)moduleDotMethod;
 // must contain only JSON compatible values
 - (NSArray *)arguments;
 
+@optional
+
+// Can be implemented for view based events that need to be coalesced
+// by it's viewTag.
+@property (nonatomic, strong, readonly) NSNumber *viewTag;
+// Coalescing related methods must only be implemented if canCoalesce
```
shergin: New line.

shergin: And `/* **/`.

React/Base/RCTEventDispatcher.h
```diff
@@ -84,12 +91,6 @@ __deprecated_msg("Subclass RCTEventEmitter instead");
 __deprecated_msg("Subclass RCTEventEmitter instead");
 
 /**
- * Deprecated, do not use.
```
shergin: We have two usage of this in internal code base, which is cool and  manageable!

React/Base/RCTComponentEvent.m
```diff
@@ -0,0 +1,57 @@
...
+  NSArray *_arguments;
+}
+
+@synthesize eventName = _eventName;
+@synthesize viewTag = _viewTag;
+
+- (instancetype)initWithName:(NSString *)name body:(NSDictionary *)body
+{
+  if (self = [super init]) {
+    NSNumber *target = body[@"target"];
```
shergin: Why should we extract target from body? Should we pass it directly?

React/Base/RCTComponentEvent.h
```diff
@@ -0,0 +1,19 @@
...
+
+- (instancetype)initWithName:(NSString *)name body:(NSDictionary *)body;
```
shergin: Hmm, why we initialize it with untyped NSDictionary? I thought whole point of having RCTEvent objects is typesafty and explicitly.

How do you see the perpose of this class? Something simple which should replace all current untyped events? Should we have it?

React/Base/RCTComponentEvent.h
shergin: Should we emphasize somehow in the names that it is untyped, simple, event which is legit but... we discourage to use it because it is untyped?

React/Base/RCTComponentEvent.m
janicduplessis: I think it is because target needs to be in the event body for JS so instead of putting it both in the body and passing it as an arg we just extract it from the body.

React/Base/RCTComponentEvent.m
janicduplessis: Yea that would be better (just copied that code :))

React/Base/RCTComponentEvent.h
janicduplessis: I think the main use of RCTEvent is to support event coalescing. RCTComponentEvent is a generic event that is sent to a component so it could be something like a touch or scroll event and those can all have different bodies. It is also used by event blocks created by view managers (RCTDirectEventBlock, RCTBubblingEventBlock). If someone want to make a type safe event I think the best way is to create a new class that implements RCTEvent or subclass RCTComponentEvent. An example of this is https://github.com/facebook/react-native/blob/master/React/Views/RCTScrollView.m#L26

React/Base/RCTComponentEvent.h
janicduplessis: It is a generic untyped event for Components because it DOES use view tags (target that is extracted from the body). We could say that it is preferred to create a custom event class that implements RCTEvent for type safety. The main use case of this class is it's usage internally in RCTDirectEventBlock and RCTBubblingEventBlock. It can also be used to migrate code that used sendInputEvent directly without creating a custom class.

React/Base/RCTComponentEvent.h
janicduplessis: It could be subclassed but just reimplementing RCTEvent is probably better especially now that a bunch of methods are optional.

React/Base/RCTComponentEvent.m
shergin: Mmm... So, it means that our API reflects details of internal implementation? Does not sound cool. No?
Do we modify body somewhere down the road before it does to serialization process? If yes, that coupling `target` and `body` cannot be counted as optimization and we have to fix our API.

React/Base/RCTComponentEvent.h
shergin: Yeah! We have to mention this in comment and probably reflect somehow in the name.

React/Base/RCTComponentEvent.h
shergin: Yeah!
(How can it be used by `RCT*EventBlock` if you just added this?)
So, in my mind we have to advice to product developers use `RCT*EventBlock` or Custom implementation of `RCTEvent`. We have to discourage them to use this untyped and generic RCTComponentEvent. But, it is okay to use it as a temporary solution for migration from `sendEventWithName`.

* `RCT*EventBlock` Is easy to use.
* Custom RCTEvent implementation is powerful and semantically explicit.

This should be the most important point of this change. Seems that Nick Lockwood's original idea when he started this migration awhile ago.

Do you agree with this vision? Is this reasonable?

shergin:
@janicduplessis Thank you so much for this work! Sorry for asking so many questions, I just wont to make it right finally.

React/Base/RCTComponentEvent.h
shergin: (Oh, I understood where we use it as a bearer for `RCTEvent*Block`, that's great! Fantastic!)

shergin:
I have to also mention that is it breaking change for many external libraries, se we have to be super expressive and clear with motivation of this change and give a great advices how to migrate the old code to the reality of the world without `sendEventWithName:`.

React/Base/RCTComponentEvent.m
janicduplessis: We could probably pass target as a parameter and add it to the body in the init.

React/Base/RCTComponentEvent.h
janicduplessis: Yea this class is used internally and could also be used as an easy migration from sendEventWithName. Other events should just implement RCTEvent

janicduplessis:
@shergin I think I addressed all your feedback, let me know if anything else isn't clear :)

React/Base/RCTComponentEvent.m
shergin: Yes, I think so. We have to know for sure should we do this or not. If it is natural performancewise, I would do it.

janicduplessis: I thought the extra copy to make the NSDictionary mutable would be bad but we actually did it already here https://github.com/facebook/react-native/pull/15894/files#diff-b9fc78351b7b976a36356ef6f08c1f79L117. So the main difference performance wise is the extra allocation of RCTEvent

React/Base/RCTComponentEvent.m
```diff
@@ -0,0 +1,52 @@
...
+  NSArray *_arguments;
+}
+
+@synthesize eventName = _eventName;
+@synthesize viewTag = _viewTag;
+
+- (instancetype)initWithName:(NSString *)name viewTag:(NSNumber *)viewTag body:(NSDictionary *)body
+{
+  if (self = [super init]) {
+    NSMutableDictionary *mutableBody = [NSMutableDictionary dictionaryWithDictionary:body];
```
shergin: Wait, that does not look right. Why do we need put tag inside body? That was needed for old poorly designed `sendEventBlahBlahWithoutTagButWithBody:`, right?

janicduplessis: I'm pretty sure JS needs the tag in the body that's why we put it here. I could be wrong though I'd have to look at the RCTEventEmitter.receiveEvent code.

janicduplessis: Yea it is definitely used https://github.com/facebook/react/blob/7f78749ee3023608a614b2f595739bcd36acc128/src/renderers/native/ReactNativeEventEmitter.js#L112. I'd rather not touch this :)

shergin: Yeah... But... I am trying understand that code... and I failed.
Looks like we are sending `tag` as `rootNodeID` (which is not even tag!) as a first argument... Why? 🤔

janicduplessis: Yea I'm not sure why we are sending the tag as both the first arg and in the event body as target. I just think we should keep it like this to avoid breaking react.

shergin: First of all, I am perfectly okay with applying our/your change iteratively. We don't need to modify half of react event subsystem to remove just one deprecated method, we don't need to it in single diff. However, we have to have good reasoning behind our changes and good plan for future changes.
So, let's play with debugger and verify that evething works as we think it works at least at same not-so-deep levels of abstractions, so verify that `tag` really became `rootNodeID`. (I also will investigate this part.) Then we have to explicitly mention that controversy in code and mark that as temporary legacy leftovers. Land this. And continue to improve the architecture of RN.

I am very exited about this diff!

janicduplessis: @bvaughn Could you clarify why we need to pass both `rootNodeID` and `nativeEvent.target` here https://github.com/facebook/react/blob/7f78749ee3023608a614b2f595739bcd36acc128/src/renderers/native/ReactNativeEventEmitter.js#L126? AFAIK those 2 values are always the same.

janicduplessis: @shergin Let's see if someone from the react team can help here. I agree we should have a clear plan here. We also have to consider android if we make any change to the native / react bridging api.

bvaughn: Hi! I'm only somewhat familiar with our event code. (I don't think anyone on the core team presently is deeply familiar with it, to be honest.)

The event params like `target` _seem_ to be optional given the [Flow typing](https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeEventEmitter.js#L103) and the fact that we [fallback to undefined values](https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeEventEmitter.js#L105).

If the 2 values _are_ always the same (for Android too) then it seems silly to always pass them both and I could update the `ReactNativeEventEmitter` to just create this wrapper object (or not, if we determine it's not actually being used for anything).

bvaughn: FWIW it looks like the Java implementation doesn't necessarily pass a `target` but it does pass other attributes in some cases (eg [on content size change](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/ContentSizeChangeEvent.java#L31-L36)). The Java interface also [defines the map as a `@Nullable` type](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/RCTEventEmitter.java#L19) and in [at least some cases](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/views/modal/RequestCloseEvent.java#L33) explicitly passes `null`.

janicduplessis: Thanks @bvaughn, I'm pretty sure we need to keep target in the event payload otherwise it would be a breaking change for apps that rely on this value. Maybe we could avoir passing the tag as the first argument if it's always the same as body.target. Needs more investigation.

@shergin What do you think about landing this as is since it will unblock a few other improvements I wanted to make?

bvaughn: > I'm pretty sure we need to keep target in the event payload otherwise it would be a breaking change for apps that rely on this value. 

Sure! To be clear: I wasn't suggesting we remove `event.target` on the JS side, just that we could remove the duplicate parameter and (on the RN JavaScript side) just always set target equal to the tag number.

janicduplessis:
@shergin Could you have a look at this soon? :)

shergin:
@janicduplessis Definitely.

lnikkila:
Is this PR still being considered?

I’m interested since it seems this makes it possible to use the native animation driver with onLayout and Animated.event. Currently the native driver never sees the event since it’s being sent by an event block which uses `sendInputEventWithName` and doesn’t notify observers.

janicduplessis:
@lnikkila Yes, the initial goal was to fix the case you mentioned, but it required additional changes to be able to do it in a clean way.

janicduplessis:
ping @shergin if you can find some time to revisit this it would be nice

nerdmed:
@janicduplessis  Any chance to rebase this one?

hramos:
Sorry it's taken so long - do you still want to get this in?

janicduplessis:
@hramos Did you have any issues importing this?

hramos:
Woot! Thanks @shergin for taking a look at the diff and approving it! I looked through the internal codebase and only found references to the deprecated method in react-native-fbsdk. I filed an issue to track this elsewhere: https://github.com/facebook/react-native-fbsdk/issues/479

hramos:
I'm looking into whether I can patch `react-native-fbsdk` as part of this diff. Once that's done, I'll land this.

React/Base/RCTComponentEvent.h
```diff
@@ -0,0 +1,19 @@
...
+#import <React/RCTEventDispatcher.h>
+
+/**
+ * Generic untyped event for Components. Used internally by RCTDirectEventBlock and
+ * RCTBubblingEventBlock, for other use cases prefer using a class that implements
+ * RCTEvent to have a type safe way to initialize it.
+ */
+@interface RCTComponentEvent : NSObject<RCTEvent>
+
+- (instancetype)initWithName:(NSString *)name viewTag:(NSNumber *)viewTag body:(NSDictionary *)body;
```
zhongwuzw: We need to add some nonnull attributes to prevent crash?

React/Base/RCTEventDispatcher.m
```diff
@@ -149,34 +135,39 @@ - (void)sendEvent:(id<RCTEvent>)event
...
-    event = [previousEvent coalesceWithEvent:event];
+  
+  if (event.canCoalesce) {
+    [_eventQueueLock lock];
+    
+    NSNumber *eventID = RCTGetEventID(event);
+    
+    id<RCTEvent> previousEvent = _events[eventID];
+    if (previousEvent) {
+      event = [previousEvent coalesceWithEvent:event];
```
zhongwuzw: `canCoalesce` is required, but `coalesceWithEvent` is optional, we'd better do some check? or can assume if `canCoalesce` returns `YES`, it must also implements `coalesceWithEvent`.

shergin:
🎉

React/Base/RCTEventDispatcher.m
janicduplessis: Yes, not ideal but it is mentioned in the comment here https://github.com/facebook/react-native/pull/15894/files#diff-eee95b3297ff73b389d2eab419631d6dR57

React/Base/RCTComponentEvent.h
janicduplessis: This file could use `NS_ASSUME_NONNULL_BEGIN`, feel free to open a PR :)

sahrens:
Heads up I think this (still working on confirming if this PR is the cause) might be causing some crashes with missing touches and some bugs with Animations where some scroll events seem to be getting dropped, causing screens to get out of sync. Easiest way to repro this something like this:
```
<Animated.ScrollView
          onScroll={Animated.event(
            [{nativeEvent: {contentOffset: {y: scrollOffset}}}],
            {useNativeDriver: true},
          )}>
  <Animated.View
    style={{
      height: 100,
      backgroundColor: 'cyan',
      transform: [{translateY: scrollOffset}]}}>
      <Text>FOO</Text>
  </Animated.View>
  <View style={{height: 1000, backgroundColor: 'red'}} />
</Animated.ScrollView>
```

If you scroll aggressively you'll see some jumping around of the cyan view/text when it should stay fixed in place.

shergin:
Hey Janic,

I really do love this PR but we have serious suspicion that this PR can cause a significant crash spike that we are observing (T42979947).

We are seriously considering reverting it but I hope we can figure out the issue without that (with your help). Seems the issue is related to touch event dispatching process and seems that some events are getting lost in the process. Can it be related to changes in coalescing mechanism?

Details:
```
 "assert_message": "<level:warning> QuickReport generated: Unhandled JS Exception: <force_category:console.error: \"Cannot record touch move without a touch start.\nTouch Move: %s\n\", \"Touch Bank: %s\", \"{\"identifier\":2,\"pageX\":206.5,\"pageY\":361,\"timestamp\":79597115.8785}\", \"[null,{\"touchActive\":true,\"startPageX\":168.5,\"startPageY\":516.5,\"startTimeStamp\":79588038.00445834,\"currentPageX\":170.5,\"currentPageY\":510.5,\"currentTimeStamp\":79590372.55025001,\"previousPageX\":168.5,\"previousPageY\":516.5,\"previousTimeStamp\":79588038.00445834}]\", js engine: hermes:force_category>, stack:\npe@1:411668\nrecordTouchTrack@1:455342\nextractEvents@1:455483\nanonymous@1:412884\nze@1:459961\nAe@1:412461\nMe@1:412794\nreceiveTouches@1:457566\nvalue@1:289603\nanonymous@1:288156\nvalue@1:289192\nvalue@1:288114\n",
 "symbolicated_js_trace": "<level:warning> QuickReport generated: Unhandled JS Exception: <force_category:console.error: \"Cannot record touch move without a touch start. Touch Move: %s \", \"Touch Bank: %s\", \"{\"identifier\":2,\"pageX\":206.5,\"pageY\":361,\"timestamp\":79597115.8785}\", \"[null,{\"touchActive\":true,\"startPageX\":168.5,\"startPageY\":516.5,\"startTimeStamp\":79588038.00445834,\"currentPageX\":170.5,\"currentPageY\":510.5,\"currentTimeStamp\":79590372.55025001,\"previousPageX\":168.5,\"previousPageY\":516.5,\"previousTimeStamp\":79588038.00445834}]\", js engine: hermes:force_category>, stack: /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:533:recordTouchMove /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:575:ResponderTouchHistoryStore.recordTouchTrack /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:710:ResponderEventPlugin.extractEvents /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:1049:batchedUpdates$argument_0 /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:7078:_batchedUpdatesImpl /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:1024:batchedUpdates /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:1044:_receiveRootNodeIDEvent /js/react-native-github/Libraries/Renderer/fb/ReactNativeRenderer-prod.js:1109:RCTEventEmitter.register$argument_0.receiveTouches /js/react-native-github/Libraries/BatchedBridge/MessageQueue.js:395:MessageQueue#__callFunction /js/react-native-github/Libraries/BatchedBridge/MessageQueue.js:106:__guard$argument_0 /js/react-native-github/Libraries/BatchedBridge/MessageQueue.js:343:MessageQueue#__guard /js/react-native-github/Libraries/BatchedBridge/MessageQueue.js:105:MessageQueue#callFunctionReturnFlushedQueue",
 "exception_message": "<level:warning> QuickReport generated: Unhandled JS Exception: <force_category:console.error: \"Cannot record touch move without a touch start.\nTouch Move: %s\n\", \"Touch Bank: %s\", \"{\"identifier\":2,\"pageX\":206.5,\"pageY\":361,\"timestamp\":79597115.8785}\", \"[null,{\"touchActive\":true,\"startPageX\":168.5,\"startPageY\":516.5,\"startTimeStamp\":79588038.00445834,\"currentPageX\":170.5,\"currentPageY\":510.5,\"currentTimeStamp\":79590372.55025001,\"previousPageX\":168.5,\"previousPageY\":516.5,\"previousTimeStamp\":79588038.00445834}]\", js engine: hermes:force_category>"
```

cc @hramos

janicduplessis:
@shergin I reviewed the code and it seems like when the event is not coalescable we are no longer adding it the the event queue and just dispatching it directly. I think this can cause events to arrive in the wrong order when combining coalescable and non-coalescable events.

Are you able to reproduce the crash consistently? I can put up a PR with a potential fix and you can see if it works.

PR #15897, opened 2017-09-12
[ImgBot] optimizes images

ImgBotApp:
Hey there. I've just optimized all of your images. They are the same dimensions and quality as before, only they take up less space now.

There were a lot of images in there. You can see the details of the space saving in the commit message body.


TheSavior:
This is pretty neat.

We probably don't want to compress any of the ReferenceImages since I believe those came from screenshot tests.

dabutvin:
thanks @TheSavior 

We are planning to add a `.imgbotconfig` to enable these kind of whitelist/blacklist of images directories
https://github.com/dabutvin/ImgBot/issues/2

for now, the service will try to optimize all of your images it can find in the repo.

TheSavior:
Got it. This seems like it will be a neat bot for the open source community. 

Unfortunately, we won't be able to merge this as is because of the screenshot test images. I think the directories that need to be not compressed are more than just the one I mentioned before. @sahrens would have a better idea on the folders.

I'm going to close this PR since it seems unlikely that someone will go through the images in the commit and revert some of them manually. Maybe we can revisit this once the config is supported.

dabutvin:
@TheSavior I released an update for ImgBot and opened a new PR with the configuration https://github.com/facebook/react-native/pull/16591

PR #15905, opened 2017-09-12
[cli] Add missing node-notifier npm dependency

ide:
local-cli now depends on the node-notifier package but it wasn't listed in package.json.

Test Plan: Run TravisCI e2e-objc test and see that it is able to launch the CLI successfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,6 +167,7 @@
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
     "node-fetch": "^1.3.3",
+    "node-notifier": "^5.1.2",
     "npmlog": "^2.0.4",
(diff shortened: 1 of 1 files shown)


TheSavior:
@ide, thanks for the fix.

@hramos, how was this problem able to land and pass our internal build?

hramos:
I'm not sure I follow. When was the dependency added? Was it in an internal commit or through an external PR?

rigdern:
@hramos `git log -S node-notifier` shows that commit c88b2b2c659ec00148eea13a9f9ff235ba8c9cf1 introduced code which uses `node-notifier`. I don't see a PR associated with this commit so I'm guessing it was an internal commit.

PR #15911, opened 2017-09-12
Add Example for TouchableOpacity.js

ubatin:
### Motivation
Adding example with [react-native-web-player](https://github.com/dabbott/react-native-web-player)

## Test Plan

Screenshot on http://localhost:8079/react-native/docs/touchableopacity.html
react-native-touchableopacity


TheSavior:
Thanks for contributing!

Issue #15915, opened 2017-09-12
HeadlessJsTaskService is expected to run on UI thread


oriharel:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

1. `react-native -v`: react-native-cli: 2.0.1 react-native: 0.48.1
2. `node -v`: v7.5.0
3. `npm -v`: 5.3.0

- Target Platform: Android

- Development Operating System: macOS

- Build tools:

### Steps to Reproduce

1. Received a push notification
2. Starting a Service that extends HeadlessJsTaskService.java
3. App throws exception `com.facebook.react.bridge.AssertionException: Expected to run on UI thread!`

### Expected Behavior

I expect no UI thread requirement for a background service (same as Android native SDK)

### Actual Behavior

App throws an exception

oriharel:
@chirag04 thanks for the note. However, as I understand, one of the common uses of this is performing tasks triggered by a push notification (or a job schedule). These are clearly not required to run in a UI thread as these are background activities.

cmdshepard:
@oriharel according to the documentation, HeadlessJS is meant for tasks that are needed to be ran when your app in the *background*. In which case there are no operations on the main thread.

oriharel:
Exactly. This is why I don't understand the "startTask must be triggered from the UI thread" requirement. Unless there is some limitation forcing the React context to be spawned from the UI thread.

D1no:
That means you can not trigger a headless task from another headless task (threads of threads).

oriharel:
In the documentation, it is clearly stated - 

> Now, whenever you start your service, e.g. as a periodic task or in response to some system event / broadcast, JS will spin up, run your task, then spin down.

Which means, even if a BroadcastReceiver is triggered, say from a push notification, the Headless task should be able to perform. 

But this is not the case, because there is an underline requirement that the onReactContextInitialized event handler will be called from the main thread (AKA UI thread). A requirement that most of the time not fulfilled (for example, when a user swiped out the app from the recent apps menu). 

So in fact - I DO consider this a bug and updated the form.

haimyyy:
+1

ycarmel:
+1

blankg:
+1

foghina:
[...]

PR #15940, opened 2017-09-14
re #15915 HeadlessJsTaskService is expected to run on UI thread

oriharel:
Headless tasks are required to run in the main thread, however due to the nature of the React context creation flow, the handler may be returned outside of the main thread, causing the HeadlessJsTaskContext to throw an exception.

## Test Plan

Swipe out the app. send push notification from a server that starts a HeadlessJsTaskService


chirag04:
cc @foghina

foghina:
Looks good, thanks!

PR #15953, opened 2017-09-14
Native Animated - Allow events that are dispatched from any thread

janicduplessis:
Instead of preventing events from working when not on the UI Thread we can just dispatch to it instead.

**Test plan**
Tested manually that animated events still work in RNTester


ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
```diff
@@ -363,12 +363,22 @@ public void removeAnimatedEventFromView(int viewTag, String eventName, int anima
...
 
   @Override
-  public void onEventDispatch(Event event) {
-    // Only support events dispatched from the UI thread.
-    if (!UiThreadUtil.isOnUiThread()) {
-      return;
+  public void onEventDispatch(final Event event) {
+    // Events can be dispatched from any thread so we have to make sure handleEvent is run from the
+    // UI thread.
+    if (UiThreadUtil.isOnUiThread()) {
```
janicduplessis: On Android we have to explicitly check the thread to avoid dispatching if we are already on the main thread while on iOS this is already handled in RCTExecuteOnMainQueue

PR #15965, opened 2017-09-15
Provide `options` argument for `global.registerDevtoolsPlugin`

jhen0409:
Currently if `__REACT_DEVTOOLS_PORT__` are set, the [relay-runtime](https://github.com/facebook/relay/blob/1ce348a5bc472a58607279c08e2236b8459b5877/packages/relay-runtime/store/RelayModernEnvironment.js#L95-L97) register will use the same port to connect devtools server. I think need to provide an another global variable (like `__RELAY_DEBUGGER_PORT__`), and it should provide an options argument here.

## Test Plan

Tested on RNTester, doesn't affect current `react-devtools` usage.

### DIFF
--- a/Libraries/Core/Devtools/setupDevtools.js
+++ b/Libraries/Core/Devtools/setupDevtools.js
@@ -34,7 +34,7 @@ if (__DEV__) {
    * comment and run Flow. */
   const reactDevTools = require('react-devtools-core');
 
-  register = function (plugin: DevToolsPlugin) {
+  register = function (plugin: DevToolsPlugin, options: DevToolsPluginConnection) {
     // Initialize dev tools only if the native module for WebSocket is available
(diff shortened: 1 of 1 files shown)


jhen0409:
@Slava could you look this? Thanks! I found this problem since https://github.com/jhen0409/react-native-debugger/issues/145, although Electron app of the new Relay debugger is not public yet. 😂

Slava:
Looks good to me! I don't have commit bits, so you would need to ping someone who does.

jhen0409:
@javache could you review this? Thanks!

jhen0409:
The method have been removed by https://github.com/facebook/react-native/commit/bd70f3ab3b2807ce3648548d3e70577d09c4e3c2.

PR #15979, opened 2017-09-18
`uniq` with an iteratee is `uniqBy`

deecewan:
The wrong version of the function was being used which was resulting in flow errors.

See https://lodash.com/docs/4.17.4#uniqBy for reference.

## Test Plan

There should probably be *a* test for this. According to [the lodash source](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L8407), calling `uniq` with a second argument is not supported.

Flow correctly picks up on this, and is causing errors in my project.

### DIFF
--- a/local-cli/link/link.js
+++ b/local-cli/link/link.js
@@ -14,7 +14,7 @@
  * run Flow. */
 const log = require('npmlog');
 const path = require('path');
-const uniq = require('lodash').uniq;
+const uniqBy = require('lodash').uniqBy;
 const flatten = require('lodash').flatten;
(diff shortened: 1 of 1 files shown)


Kureev:
I believe we merged similar PR a couple of weeks ago 🤔

Kureev:
Just found it. Seems we already have a PR that has been opened for a while (26 days). I don't want to discourage you, but I would prefer merging it that PR as far as it was the first one and we already reviewed it. Thanks for pointing out that we didn't merge it yet (in my mind we did haha)!

Hope to see more contributions from you in the future!

deecewan:
That's all good, just want to get it merged :)

PR #15984, opened 2017-09-18
Updated support for WebStorm

esbenvb:
Added the executable path to the `COMMON_EDITORS`
Added the executable name to the arguments selections

Need to be able to jump to specific files and line numbers in WebStorm and autodetect WebStorm as editor, without setting up any environment vars.

## Test Plan

Start a new RN project in WebStorm, make some JS errors and click the errors to confirm that it will jump to the right file and line number in WebStorm.

Make sure that none of the other `COMMON_EDITORS` are running, unless you set the `REACT_EDITOR` environment variable.


local-cli/server/util/launchEditor.js
```diff
@@ -37,6 +37,7 @@ var COMMON_EDITORS = {
   '/Applications/Sublime Text 2.app/Contents/MacOS/Sublime Text 2':
     '/Applications/Sublime Text 2.app/Contents/SharedSupport/bin/subl',
   '/Applications/Visual Studio Code.app/Contents/MacOS/Electron': 'code',
+  '/Applications/WebStorm.app/Contents/MacOS/webstorm': 'webstorm',
```
AlanFoster: Cool; I didn't realize that's how it's wired up in React-Native. Just out of curiosity, do we need to add support for other IntelliJ products, like RubyMine + IntelliJ? 🤔 

```
/Applications/RubyMine.app/Contents/MacOS/rubymine
```

esbenvb: Of course, if people use RubyMine for RN development and it has a command line tool for opening files. I know that webstorm and idea have, but appcode doesn't seem to have one...

PR #15989, opened 2017-09-18
add programmatic variant for permission check (Android)

koenpunt:
It can happen that location access is denied by the user because of them checking the "don't show again" checkbox in the permission dialog. When this is the case, every subsequent request of accessing the location will make the app crash, even when the permission is added to the manifest.

By updating the exception message I would like it to be more clear to developers where else to look than in the `AndroidManifest.xml`. But it might be necessary to make it even more descriptive/to elaborate on the specific case described before.


koenpunt:
@janicduplessis I just noticed you updated the PR. I have now updated again, but using rebase.

Issue #15863, opened 2017-09-08
[RN 0.48.2 TextInput] onSubmitEditing not called if blurOnSubmit set to false


xxsnakerxx:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

1. `react-native -v`: 0.48.2
2. `node -v`: 8.4.0
3. `npm -v`: 4.6.1
4. `yarn --version`:

- Target Platform: iOS, Android

- Development Operating System: macOS Sierra 10.12.6

- Build tools:
`react-native run-ios`
`react-native run-android`

### Steps to Reproduce

1. 
```js
<TextInput 
  blurOnSubmit={false} 
  onSubmitEditing={() => alert('!!!')} 
  returnKeyType={'done'}
/>
```
2. If I press the __done__ button __onSubmitEditing__ callback not called

### Expected Behavior

`onSubmitEditing` should be called for  `blurOnSubmit={false}` text inputs

### Reproducible Demo

https://snack.expo.io/S1tMt619b
but I don't know how to turn on the last RN version there

raphaelhovsepyan:
+1

raphaelhovsepyan:
This is works on 0.47

xxsnakerxx:
related with [this](https://github.com/facebook/react-native/commit/8f93ce680d79fac4b58e518fde5f019fb9eb7e89#diff-1fc9493f9f850337aca70b21009a5ee2R89) commit

I found a workaround

Change in `node_modules/react-native/Libraries/Text/RCTTextInput.m`
```obj-c
- (BOOL)textInputShouldReturn
{
  return _blurOnSubmit;
}
```
to
```obj-c
- (BOOL)textInputShouldReturn
{
  // We send `submit` event here, in `textInputShouldReturn`
  // (not in `textInputDidReturn)`, because of semantic of the event:
  // `onSubmitEditing` is called when "Submit" button
  // (the blue key on onscreen keyboard) did pressed
  // (no connection to any specific "submitting" process).
  [_eventDispatcher sendTextEventWithType:RCTTextEventTypeSubmit
                                 reactTag:self.reactTag
                                     text:self.backedTextInputView.text
                                      key:nil
                               eventCount:_nativeEventCount];
  
  return _blurOnSubmit;
}
```

shergin:
It is fixed in master.

raphaelhovsepyan:
[...]

PR #16040, opened 2017-09-20
Fix onSubmitEditing dispatch behavior with blurOnSubmit/multiline combinations.

Adam0000:
Fixes https://github.com/facebook/react-native/issues/15863 on master. Behavior of `onSubmitEditing` is now consistent between iOS and Android. Tapping the submit button in a TextInput dispatches the event precisely when doing so does not make a newline (when blurOnSubmit is true or multiline is false).

## Test Plan

1. Run this app on iOS and Android:
```
// @flow

import {
  StyleSheet,
  TextInput,
  View
} from 'react-native';

type State = {
  toggled: boolean
};

type Props = {
  blurOnSubmit: boolean,
  multiline: boolean
};

class ToggleColorInput extends Component<Props, State> {
  state: State = {
    toggled: false
  };

  props: Props;

  toggle = () => {
    this.setState({
      toggled: !this.state.toggled
    });
  }

  render() {
    return (
      <TextInput
        blurOnSubmit={this.props.blurOnSubmit}
        multiline={this.props.multiline}
        onSubmitEditing={this.toggle}
        style={[styles.textInput, {backgroundColor: this.state.toggled ? 'blue' : 'azure'}]}
        underlineColorAndroid='transparent'
      />
    )
  }
}

export default class App extends Component<{}> {
  render() {
    return (
      <View>
        <ToggleColorInput blurOnSubmit={true} multiline={true} />
        <ToggleColorInput blurOnSubmit={true} multiline={false} />
        <ToggleColorInput blurOnSubmit={false} multiline={true} />
        <ToggleColorInput blurOnSubmit={false} multiline={false} />
      </View>
    );
  }
}

  textInput: {
    height: 75,
    borderWidth: 1,
    borderColor: 'black'
  }
});
```
2. You see four TextInputs, with each combination of the `blurOnSubmit` and `multiline` properties. For each TextInput, type some text and tap the submit button.
3. The TextInputs in this test will toggle background color when they emit an `onSubmitEditing` event. Verify the following behavior on each platform:
 * blurOnSubmit && isMultiline => Submit event emitted, blurred, no newline inserted
 * blurOnSubmit && !isMultiline => Submit event emitted, blurred
 * !blurOnSubmit && isMultiline => Submit event emitted, newline inserted
 * !blurOnSubmit && !isMultiline => Submit event emitted


shergin:
@Adam0000 Oh, I am sorry, I just remember that there are cases when product developers want to receive onSubmit event even in `!blurOnSubmit && isMultiline` case, so iOS implementation is correct from this prospective.

Sorry again, this is really tricky, I already spent many-many days improving this on both platforms and... it is still imperfect. :(

Adam0000:
@shergin I've made the appropriate changes now and updated the test plan.

PR #16046, opened 2017-09-21
Direct assignment of the ref to avoid warning when mocking.

### LINKED ISSUE #16045: PickerIOS and DatePickerIOs generate warnings about refs when mocked by jest
### Is this a bug report?

Yes

Yes

### Environment

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  7.7.2
  Yarn:  Not Found
  npm:  4.1.2
  Watchman:  4.7.0

  react-native: 0.48.3 => 0.48.3

### Steps to Reproduce

1. Run the attached test with jest

### Expected Behavior

No warning should be generated.

### Actual Behavior

```
Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
    
    Check the render method of `PickerIOS`.
        in Unknown (created by PickerIOS)
        in View (created by View)
        in View (created by PickerIOS)
        in PickerIOS
```

### Reproducible Demo

```
import React from 'react';
import { PickerIOS } from 'react-native';
import renderer from 'react-test-renderer';

describe('StateLessMocks', () => {
  it('generates a warning', () => {
    const tree = renderer.create(<PickerIOS />).toJSON();
    expect(tree).toMatchSnapshot();
  });
});
```

jmoutte:
When Jest is mocking native components that are assigned a ref with an arrow function, a warning is generated about stateless components which can not have a ref. This does not happen when the ref is assigned directly.

Fixes #16045.

## Test Plan

The test provided in the related issue is used to validate the fix. I don't know how to detect console warnings in a Jest test.


breogangf:
This avoids multiple warnings, and I can confirm This fixes issue [#16045](https://github.com/facebook/react-native/issues/16045).

janicduplessis:
Hey @jmoutte thanks for sending a PR! I just had a look and I think a better fix would be to make the native component mock use a class component instead of an arrow function this way refs won't trigger this warning. The mock is here https://github.com/facebook/react-native/blob/master/jest/setup.js#L316.

janicduplessis:
Also updated your description for use Fixes #issuenumber syntax so it closes your issue when this lands :)

jmoutte:
Hey @janicduplessis !
Indeed, that was my first idea so that I could have that mock in my own setup.js but I am afraid that's beyond my current skills with React. I have tried generating a ReactClass dynamically with the same displayName and dummy render method but obviously that breaks all my snapshots. I've looked into the code of react-test-renderer to try to figure out how a class based component could look the same in the snapshots with no success.

janicduplessis:
@jmoutte Can you try this, replace at line 316 of setup.js:
```js
return viewName => class extends React.Component {
    render() {
      return React.createElement(
        viewName,
        this.props,
        this.props.children,
      );
    }
  };
```

Did something like this code https://github.com/facebook/react-native/blob/master/jest/mockComponent.js#L19.

I ran RN tests and that didn't break any snapshots so it might work.

jmoutte:
Darn, I was close. I was trying to pass the original props when the element was created and those contained the ref as well so the warning stayed. Thanks @janicduplessis !

Libraries/Components/DatePicker/DatePickerIOS.ios.js
```diff
@@ -125,7 +125,7 @@ const DatePickerIOS = createReactClass({
     return (
       <View style={props.style}>
         <RCTDatePickerIOS
-          ref={ picker => { this._picker = picker; } }
```
janicduplessis: Can you just revert these changes now

jmoutte:
@janicduplessis done ! I thought you might be interested in trying to get a common pattern for setting the ref but I understand that this can be off topic for that PR. Thanks again for your support.

janicduplessis:
Yea I'd rather not change things we don't have to. Thanks for your work!

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
@facebook-github-bot shipit

PR #16063, opened 2017-09-22
Update PermissionsAndroid.js

jdcc2:
The `request` method does not return `true` on Android version 5.1 and lower devices when the permissions are granted through the manifest file, as was implied by the documentation.


shergin:
@jdcc2 Do you have any proof or reference?

jdcc2:
@shergin No, I observed this behavior while testing my app on Android devices versions  5 through 7.1. Looking at the source, I think this happened because `request()` returns  a `"granted"` string when the permission is already granted on Android 6>, but for 5.1 and lower `request()` returns a boolean `true` when the permission has already been granted (through the manifest). So I think the docs are still incorrect about the behavior of `request()`, only not in the way I thought.

hramos:

PR #16075, opened 2017-09-23
call out that /img/favicon.png must be added

monkeywithacupcake:
Users who start with Create React Native App will not have an img folder or favicon.png in their app folder structure. These users may not immediately understand how to overcome the error that they will get. Added text clarifies that an image file is needed.

(Write your motivation here.)

## Test Plan


hramos:
Can we instead change the example to use the full URL for the website's favicon? That way, it will work within the Snack player, as well as locally.

monkeywithacupcake:
I don't really know what some of the things you said mean.- rebase and context? 

So, I will say that I think adding the full URL  or changing the readme will work.

PR #7959, opened 2016-06-06
Fix LayoutAnimation Android bug when adding and removing views at the same time

janicduplessis:
Similar to #7942 but for android

Fix a bug that happens when views are added during a delete layout animation. What happens is that the inserted view is not inserted at the proper index if the deleted view index is smaller than the inserted one. This is because the view is not immediately removed from the subviews array so we need to offset the insert index for each view that is going to be deleted with an animation and is before the inserted view.

Test plan (required)
Tested that this fixed the bug in an app where I noticed it, also tested the UIExplorer example to make sure LayoutAnimations still worked properly.

PR #16091, opened 2017-09-26
Fix layout animations when adding and removing views at the same time on Android

janicduplessis:
Actual fix for #7959 that was abandoned because of some issues with that implementation.

This fixes some view hierarchy inconsistencies caused by layout delete animations. The fix is similar to what was done on iOS recently. To be able to do so properly we introduce "React" views in `ViewGroupManager` to allow the native view hierarchy to diverge from what React knows about.

This also fixes another bug that is caused by animations being interrupted. To fix this we make sure to `clearAnimations` before starting a new one so that `onAnimationEnd` is called and we update the opacity to the final value there in case the animation didn't finish. Sadly Android seems to have really bad support for running multiple animations at the same time so just updating the value to the final value was the best fix for now.

**Caveats**
This won't work if the parent uses `removeClippedSubviews`. In that case it doesn't use `ViewGroupManager` methods because it implements its own way of having a different native and react view hierarchies. Since we're moving away from `removeClippedSubviews` (I think so haha), I didn't bother trying to make it work. This might be a bad idea since it will crash with some out of bounds errors and I'm not sure how we could detect that and show a useful warning. Another alternative would be to use the new React views api to reimplement subview clipping but I'd rather not mess with that code tbh :).

**Perf**
This should have a negligible impact since we're only adding a single map lookup to view operations for views that don't diverge from their native view hierarchy. It might be worth trying to add a check to get rid of the additional mapping if we detect that both native and react children becomes the same again after diverging.

**Test plan**
Used this gist https://gist.github.com/janicduplessis/211d78c5408ab17dee7ad2e381648a48 to reproduce the bugs and made sure there was no longer any inconsistencies when spamming a bunch of add and remove views. Also made sure no views were leaked because of the extra mapping we keep in `mReactChildrenMap`.


janicduplessis:
cc @shergin This is similar to the fix you did on iOS

jeanregisser:
@janicduplessis @shergin any way to move this forward? 

One app I'm working on is impacted by this bug. Would love to see this merged.

janicduplessis:
@jeanregisser After trying this in a production app I found an issue where sometimes touch events are blocked (probably by a view with 0 opacity that doesn't get removed). I need to investigate this before we can move forward.

I would really like to also add tests to this somehow since it is somewhat complex and easy to break.

jeanregisser:
Ok! Thanks for the update @janicduplessis! Let me know if I can help in any way.

PR #16102, opened 2017-09-26
Add one application for showcase

yolosaucisson:
Added one application in the showcase list.

## Test Plan

I ran the application and the json is still in a correct format, that's why the app still work.


hramos:
As documented in the showcase source code, we are not considering additions to the application list at this time.

PR #16115, opened 2017-09-27
Add finally method to Promise flow definition

Bhullnatik:
The `finally` method definition was missing, while it really exists causing a flow error when using it.

## Test Plan

Added the method definition and make sure flow doesn't throw any error anymore on a real-life example.

### DIFF
--- a/flow/Promise.js
+++ b/flow/Promise.js
@@ -33,6 +33,12 @@ declare class Promise<+R> {
   static race<T>(promises: Array<Promise<T>>): Promise<T>;
 
   // Non-standard APIs
+
+  // See https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/__forks__/Promise.native.js#L21
+  finally<U>(
+    onFinally?: ?(value: any) => Promise<U> | U
+  ): Promise<U>;
+
   done<U>(
     onFulfill?: ?(value: R) => mixed,
     onReject?: ?(error: any) => mixed


flow/Promise.js
```diff
@@ -33,6 +33,10 @@ declare class Promise<+R> {
   static race<T>(promises: Array<Promise<T>>): Promise<T>;
 
   // Non-standard APIs
+  finally<U>(
+    onSettled?: ?(value: any) => Promise<U> | U
```
ide: Name this onFinally, like the spec. Also remove the parameter and set the return type to void.

flow/Promise.js
```diff
@@ -33,6 +33,10 @@ declare class Promise<+R> {
   static race<T>(promises: Array<Promise<T>>): Promise<T>;
 
   // Non-standard APIs
```
ide: May want to make a new section for TC-39 proposals, finally() is in Stage 3

Bhullnatik:
@ide Ok so funny thing. 
I was wondering about the changes you requested because I was sure it was working according to my test plan, so I went down into the rabbit hole, read the spec and everything, and discovered that [fbjs is adding its own `finally` implementation, not compliant with the spec](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/__forks__/Promise.native.js#L21) (required [here](https://github.com/facebook/react-native/blob/master/Libraries/Promise.js#L17)).

I actually pasted the [definition from Flow](https://github.com/facebook/flow/blob/master/tests/type_param_variance2/libs/Promise.js#L34) so I assume it's what expected accross Facebook?

Anyway given all that, I modified the function name and added a comment to look back on the issue, not sure I can do much more.

ide:
OK that sounds reasonable. Probably makes sense for FB to later update the Promise polyfill with finally (instead of adding it in fbjs).

@facebook-github-bot shipit

PR #16123, opened 2017-09-28
Set view id before addEventEmitters, fix #16122

### LINKED ISSUE #16122: Custom JS component in some case can't receive event from Android native UI component.
### Is this a bug report?

Yes.

Yes.

### Environment

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.1.2
  Yarn:  0.27.5
  npm:  5.4.2
  Watchman:  4.9.0
  Android Studio:  3.0 Beta6 AI-171.4333198

  react-native: 0.48.1 => 0.48.1

### Steps to Reproduce

(Write your steps here:)

1.  I built a map native UI component and want to send map loaded event to JS
2.  override addEventEmitters method and set up map loaded listener
3.  inside listener use RCTEventEmitter's receiveEvent method to send event

### Expected Behavior

In my custom map JS component will receive map loaded event.

### Actual Behavior

Custom map JS component never received the event

### Reproducible Demo
The followings are the codes in my demo. 

```Java
@Override
    protected void addEventEmitters(final ThemedReactContext reactContext, final MapView view) {

        view.getMap().setOnMapLoadedListener(new OnMapLoadedListener(){
            @Override
            public void onMapLoaded() {
                reactContext.getJSModule(RCTEventEmitter.class)
                        .receiveEvent(view.getId(), "onMapLoaded", Arguments.createMap());
                Log.d(TAG, "onMapLoaded View id : " + mView.getId());
            }
        });

        view.getMap().setOnClickListener(new View.OnClickListener() {

            @Override
            public void onClick(View view) {
                WritableMap argu = Arguments.createMap();
                argu.putString("method", "View Click");
                reactContext.getJSModule(RCTEventEmitter.class)
                        .receiveEvent(mView.getId(), "onClick", argu);
                Log.d(TAG, "view click View id : " + mView.getId());
            }
        });

        Log.d(TAG, "addEventEmitters View id : " + mView.getId());
    }
```
JS component can receive the onClick event but never received onMapLoaded event.

And the log showed the view id in onMapedListener & addEventEmitters are the same. onClickListener has another view id. 

When I looked into the ViewManager.java source code. I found the issue
```Java
// ViewManager.java
  public final T createView(
      ThemedReactContext reactContext,
      JSResponderHandler jsResponderHandler) {
    T view = createViewInstance(reactContext);
    addEventEmitters(reactContext, view);
    
    ...
    return view;
  }
```
```Java
public synchronized void createView(
      ThemedReactContext themedContext,
      int tag,
      String className,
      @Nullable ReactStylesDiffMap initialProps) {
    ...
    try {
      ViewManager viewManager = mViewManagers.get(className);

      View view = viewManager.createView(themedContext, mJSResponderHandler);
     
     ...

      // Use android View id field to store React tag. This is possible since we don't inflate
      // React views from layout xmls. Thus it is easier to just reuse that field instead of
      // creating another (potentially much more expensive) mapping from view to React tag
      view.setId(tag);
     ...
    
  }
```
`view.setId(tag)` is after `addEventEmitters`. 

I think that's may cause the event can't be sent to JS component issue in some case.

zhgqthomas:
(Write your motivation here.)

1.  I built a map native UI component and want to send map loaded event to JS
2.  override addEventEmitters method and set up map loaded listener
3.  inside listener use RCTEventEmitter's receiveEvent method to send event

### Expected Behavior

In my custom map JS component will receive map loaded event.

### Actual Behavior

Custom map JS component never received the event

## Test Plan
I built a demo to test the fix. And it works like I expected. 
https://github.com/zhgqthomas/build-react-native

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -229,6 +229,9 @@ public class NativeViewHierarchyManager {
       // React views from layout xmls. Thus it is easier to just reuse that field instead of
       // creating another (potentially much more expensive) mapping from view to React tag
       view.setId(tag);
+
+      // setId after event emitter may cause bugs
+      viewManager.addEventEmitters(themedContext, view);
       if (initialProps != null) {
         viewManager.updateProperties(view, initialProps);
       }
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManager.java
@@ -42,7 +42,6 @@ public abstract class ViewManager<T extends View, C extends ReactShadowNode>
       ThemedReactContext reactContext,
       JSResponderHandler jsResponderHandler) {
     T view = createViewInstance(reactContext);
-    addEventEmitters(reactContext, view);
     if (view instanceof ReactInterceptingViewGroup) {
       ((ReactInterceptingViewGroup) view).setOnInterceptTouchEventListener(jsResponderHandler);
     }


janicduplessis:
Thanks for the PR!

Looks like the issue is that the event is triggered synchronously immediately when you add your listener. I'm not 100% sure we want to support this. An simple workaround would be to dispatch a Runnable on the ui thread to add the listener asynchronously.

@AaaChiuuu What do you think about this?

bolan9999:
@zhgqthomas do you know how to fix it on iOS?

Issue #16039, opened 2017-09-20
[RN 0.49-rc.5] Fishhook error


mobinni:
### Is this a bug report?
No.

I tried to manually upgrade from 0.48.3 to 0.49-rc.5 due to the `BLOB_URL` error that happens in sed version. 

https://github.com/facebook/jest/issues/4433
https://github.com/facebook/react-native/issues/15810

When upgrading everything went well until the following error occurred:
![](https://i.imgur.com/PLmFz7Y.png)

I tried adding:
```  
pod "React", :path => react_native_path, :subspecs => ['DevSupport', 'fishhook']
```

However this did nothing for me, there is currently no documentation on how to integrate fishhook when manually upgrading. Can anyone help?

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment

  Environment:
    OS:  macOS Sierra 10.12.4
    Node:  7.8.0
    Yarn:  0.24.6
    npm:  5.4.0
    Xcode:  Xcode 8.3.3 Build version 8E3004b

  Packages: (installed)
    react-native: 0.49-rc.5 
    react: 16.0.0-alpha.12

  Target Platform: iOS (10.3)

### Expected Behavior

Build succeeds

### Actual Behavior
Build fails because it cannot find fishhook files

mobinni:
@here update I figured it out to be a mis-import in reconnecting websocket, it links to
```
#import "<fishhook/fishhook.h>"
```
while it should link to 
```
#import "fishhook.h"
```

johannth:
This is a similar issue to https://github.com/facebook/react-native/issues/13198

I think this stems from Cocoapods not defining subspecs as modules/frameworks so there is no `fishhook` module and we can't do `<fishhook/fishhook.h>`. I've asked on the [Cocoapods mailing list](https://groups.google.com/forum/#!topic/cocoapods/65WdPX8OH6g) to confirm but I agree that the import is probably wrong and should either be:

`#import <React/fishhook.h>`

or

`#import "fishhook.h"`

An alternative in this case would be to depend on the `fishhook` Cocoapod instead of the embedded subspec. Then the import could stay as it is.

ide:
I believe this is addressed in https://github.com/facebook/react-native/pull/16192 so I'm going to close this. That commit should go out with 0.49.1 in an hour or so.

rob-keepsafe:
@johannth You can add a postinstall script to fix that issue into your package.json like

[...]

PR #16130, opened 2017-09-28
Fix reference to fishhook.h in RCTReconnectingWebSocket.m

mobinni:
## Motivation

When upgrading to react-native 0.49 you get an error due to a reference to `fishhook.h` in `RCTReconnectingWebSocket.m`. When I read the fishhook docs it specified that when the `fishhook.a` binary is linked references should be imported as:
```
#import "fishhook.h"
```

Link to issue: https://github.com/facebook/react-native/issues/16039

https://github.com/facebook/react-native/blob/master/Libraries/fishhook/README.md

## Test Plan

Upgrade from a clean react-native 0.48 to 0.49, upgrade your podspec to include fishhook.
Run the project and it should fail on this import.

### DIFF
--- a/Libraries/WebSocket/RCTReconnectingWebSocket.m
+++ b/Libraries/WebSocket/RCTReconnectingWebSocket.m
@@ -11,7 +11,7 @@
 
 #import <React/RCTConvert.h>
 #import <React/RCTDefines.h>
-#import <fishhook/fishhook.h>
+#import "fishhook.h"
 
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 /* __IPHONE_11_0 */
 #import <os/log.h>


ide:
Another commit that fixes this via the podspec has landed (https://github.com/facebook/react-native/pull/16192) and will be going out with 0.49.1 in an hour or so if the build pipeline succeeds. I'm going to close this out since that commit should fix your issue and please tell us if it doesn't.

VinceBurn:
Hi @ide I'm using react-native 0.49.3 and this issue is still valid, my project is not compiling with Xcode 9 if I don't change the source in my Pod version of 'RCTReconnectingWebSocket.m'. 
(I'm on a new project, fresh install of the Pod to integrate React Native to an existing project)

This merge request need to be reopen since the problem is clearly not fixe.

ide:
OK, it's clear this is still an issue but I don't think it is right to modify the source code though I could be wrong about that, I don't have visibility into Facebook's internal build systems. The right fix I think is to figure out how to get Xcode to treat fishhook as a header directory.

oskarsliukis:
Any updates? Can't upgrade to latest RN version due to this and changing import manually is not an option.

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -11,7 +11,7 @@
 
 #import <React/RCTConvert.h>
 #import <React/RCTDefines.h>
-#import <fishhook/fishhook.h>
+#import "fishhook.h"
```
hramos: I don't think this change is correct. The `fishhook.h` file is not in the same directory, so importing it needs to be done through the fishhook library, e.g. `#import <fishhook/fishhook.h>`.

It might be the case that it's actually included under `React`, instead (see https://github.com/facebook/react-native/pull/16192#issuecomment-337428712).

msand:

PR #16132, opened 2017-09-28
Improve YellowBox output format

janicduplessis:
YellowBox currently assumes the first arg is a printf like format string, this adds support for any arguments so it works more like console in the browser. This also adds `stringifySafe` to format arguments when using printf style.

The main annoyance that this fixes is when trying to log a single object it will currently print [object Object] instead of the fully stringified version.

**Test plan**

Tested a bunch of different log combinations.

```js
console.warn({test: 'a'}); // {"test":"a"} (was [object Object] before this patch)
console.warn('test %s %s', 1, {}); // test 1 {}
console.warn('test %s', 1, {}); // test 1 {}
console.warn({}, {}, {}, {}); // {} {} {} {}
```

### DIFF
--- a/Libraries/ReactNative/YellowBox.js
+++ b/Libraries/ReactNative/YellowBox.js
@@ -106,17 +106,22 @@ function sprintf(format, ...args) {
   return format.replace(/%s/g, match => args[index++]);
 }
 
-function updateWarningMap(format, ...args): void {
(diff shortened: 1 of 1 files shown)


yungsters:
This is great. Thanks (and sorry for the slow turnaround)!

PR #16150, opened 2017-09-30
missed check for Null, on duck-typing check for Date

rohozhnikoff:
(Write your motivation here.)

## Test Plan


shergin:
Flow enforces non-nullability here.

Issue #13697, opened 2017-04-28
Image headers unsupported when an array of sources is provided


npomfret:
I'm trying to use the [new](https://github.com/facebook/react-native/commit/8c0e6ecfc0e8b8ceb0471739ac2db53772d538cb) `headers` prop on images in Android.  I've managed to get it working on Android but it didn't behave as expected (and I believe its undocumented so was guessing on how to use it).

For a simple image source it works fine, something like this:

    <Image source={uri: foo, headers: {Authorization: bar}} ...

But in my case I want to pass an array of different sized images.  I assumed this would be the correct approach:

    <Image source={[
        {uri: foo, headers: {Authorization: fooHeader}, 
        {uri: bar, headers: {Authorization: barHeader} ...

But that doesn't work (and the error message that comes back when my auth fails does seem to have any information in it).  Looking at [this code](https://github.com/facebook/react-native/blob/master/Libraries/Image/Image.android.js#L314), I noticed that it always expects the `headers` prop to be at the top level, so one headers prop for all images (unlike the `cache` prop which is one per uri). As the source object is an array, so I ended up with something like this:

    const sourceArray = [[uri: foo], [uri: bar]];
    sourceArray.headers = headers;

Is this by design? It doesn't feel quite right to me - I was expecting the `headers` prop to be treated more the the `cache` prop.  Having said that, I can't see any evidence of the `cache` header being implemented in android.

cbjs:
0.45 Android still not working

hramos:
As far as I can tell, this is not currently supported. You may want to submit a PR that adds support for per-source headers.

phillip055:
I am working on this, any guidance would be gladly welcome

hramos:
Looks like #16155 was merged, thanks @phillip055

PR #16155, opened 2017-10-01
[Image] Add headers prop in source array type prop

phillip055:
#13697 
My first contribution 👍 

## Test Plan


phillip055:
If more tests are required for this, will add it as well 😄

Issue #15801, opened 2017-09-05
App crashes if changing the data of FlatList with renderItem TextInput quickly


JenningsWu:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

1. `react-native -v`: 0.48.1
2. `node -v`: 8.4.0
3. `npm -v`: 5.3.0
4. `yarn --version`:

Then, specify:

- Target Platform: iOS

- Development Operating System: macOS

- Build tools: Xcode 9.0 beta (9M136h)

### Steps to Reproduce

(Write your steps here:)

1. Create a FlatList with renderItem = TextInput, and a Button to control the content of the FlatList
2. Click the Button quickly ("quckly" means you need to click the button before previous render is finished)
3. The app will crash.

### Expected Behavior

The app shouldn't crash.
If the render item of FlatList is another component like Text, it won't crash.

### Actual Behavior

The app crash with log
```
FlatListOfTextInput[16471:6155941] *** Assertion failure in -[RCTUIManager setIntrinsicContentSize:forView:]_block_invoke(), /Users/jinningwu/Github/FlatListOfTextInput/node_modules/react-native/React/Modules/RCTUIManager.m:389
2017-09-04 19:40:06.314030-0700 FlatListOfTextInput[16471:6155941] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Could not locate view with tag #4245'
*** First throw call stack:
(0x18a752fe0 0x1891b4538 0x18a752eb4 0x18b1eb7cc 0x100117da4 0x10068da50 0x10068da10 0x10069b2e8 0x100691634 0x10069d630 0x10069d39c 0x189813100 0x189812cac)
libc++abi.dylib: terminating with uncaught exception of type NSException
```
### Reproducible Demo

Could use this repo [JenningsWu/FlatListOfTextInput](https://github.com/JenningsWu/FlatListOfTextInput) to reproduce it.

It is interesting that I failed to reproduce the crash with expo Snack.

To reproduce, just run this repo with `react-native run-ios` or build with Xcode.

JenningsWu:
It seems that it will only crash with configuration `Debug`.

simonracz:
We also ran into this issue last week. Pretty annoying bug. I am glad we are not the only ones. :)

PR #16170, opened 2017-10-02
Removed a wrong assert.

simonracz:
This fixes [#15801](https://github.com/facebook/react-native/issues/15801)

We ran into a strange crash on iOS (debug only). After removing the clutter I was able to reproduce it in a tiny app. You can check it out [here.](https://github.com/simonracz/textinput_stress)

The UI in JS and native are not always in sync (which is okay). Due to this, a native view might call back into JS, which is no longer present in the shadow view hierarchy there. I think this should be also okay.

TextInput in some cases calls into [setIntrinsicContentView](https://github.com/facebook/react-native/blob/6d67e2dbbcd658b7f845ebb0d0156bd64dc68226/React/Modules/RCTUIManager.m#L382), where it triggers an overly enthusiastic `NSAssert` and crashes the app.

## Test Plan

Check out [textinput_stress](https://github.com/simonracz/textinput_stress)
Rotate the simulator a few times to see the crash or the lack of crash.

### DIFF
--- a/React/Modules/RCTUIManager.m
+++ b/React/Modules/RCTUIManager.m
@@ -376,7 +376,9 @@ static NSDictionary *deviceOrientationEventBody(UIDeviceOrientation orientation)
   NSNumber *reactTag = view.reactTag;
   dispatch_async(RCTGetUIManagerQueue(), ^{
     RCTShadowView *shadowView = self->_shadowViewRegistry[reactTag];
-    RCTAssert(shadowView != nil, @"Could not locate view with tag #%@", reactTag);
+    if (shadowView == nil) {
+      return;
+    }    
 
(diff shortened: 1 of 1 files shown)


simonracz:
Hey @shergin ,

I've updated the PR.

I checked that I see the log message (and not a crash) in my sample app.

FYI, there are several places in RCTUIManager.m with the same RCTAssert : `RCTAssert(shadowView != nil, @"Could not locate view with tag #%@", reactTag);`

I haven't touched them now, because I am not sure whether those are also bugs. My gut feeling is that most of those should be also only warnings.

shergin:
I agree, it should be special internal method (something like `executeBlockWithShadowView:forView:`) which have to check shadowView for null and produce warning.

PR #16176, opened 2017-10-03
Add prettier config, format script and reformat files

janicduplessis:
Prettier 1.7.0 has config file + pragma support so this works really well with the current prettier usage. Also added a prettier script to run it (taken mostly from the relay repo) and ran it which caused 1 files to change (probably wasn't updated when upgrading prettier to 1.7.0).

**Test plan**
Made sure flow still checked
Run 'yarn prettier', should format only files with `@format` using the config in package.json.


shergin:
@janicduplessis Conflicts?

janicduplessis:
@shergin Fixed

janicduplessis:
ping @shergin :)

PR #16185, opened 2017-10-03
Fixing test by updating snapshot for TouchableHighlight

matejstrasek:
Tests were failing due to not updated snapshot about TouchableHighlight.

## Test Plan

Run `npm test`.


janicduplessis:
Thanks for the PR, I had a look at why the snapshot changed and it is caused by https://github.com/matejstrasek/react-native/commit/e69169982021261e138add21506dea11dd91b7e4 so this change is fine.

@javache Could you land this? Also does this test not run on fb infra? I'm surprised it got broken.

janicduplessis:
@facebook-github-bot shipit

Issue #5037, opened 2015-12-30
is ~8mb the smallest size RN android apk can get?


alvinwoon:
I have a simple app. Ran proguard and it went down to 7.5mb. I just want to know whether there's other way to reduce the apk size that I am not aware of.

MuruganDurai:
Hi @alvinwoon, How did you ran the pro-guard, can you please share the steps.

satya164:
@alvinwoon React Native app APKs include JSCore binaries for x86 and ARM. If you don't need x86 you could reduce the size even more.

alvinwoon:
@satya164 Thanks! That reduce it down to 5.2mb but crashes the app upon opening. Need to debug more and report back.

MuruganDurai:
@satya164 After proguard enabled I could not get the build finished, got unable to compute hash error at ":app:packageRelease". Also If I disabled the proguard minify, able to generate the apk but the app crashes always for first navigation with error "Unfortunattely <>...", after that first crash app works perfectly. Do you know how to get the crash reports for this error and fix for the previous error.

brentvatne:
cc @mkonicek

MuruganDurai:
Thanks @brentvatne, after re-generating key everything works fine, but getting crash reports would be very helpful.

satya164:
@MuruganDurai You have 2 options for crash reports,

1. Use `adb logcat`, or the logs panel in Android studio to see logs (filter only errors to see crash reports).
2. Use something like crashlytics in your productions apps to track crashes

hosainnet:
@MuruganDurai How did you solve the 'got unable to compute hash error at ":app:packageRelease"' issue exactly? I tried regenerating the key but no luck

MuruganDurai:
@Hosain As I said before I copied all the contents suggested from stackoverflow for "proguard-rules.pro" and re-generated the key.


-keep class * extends com.facebook.react.bridge.JavaScriptModule { *; }
-keep class * extends com.facebook.react.bridge.NativeModule { *; }
-keepclassmembers class *  { @com.facebook.react.uimanager.UIProp <fields>; }
-keepclassmembers class *  { @com.facebook.react.uimanager.ReactProp <methods>; }
-keepclassmembers class *  { @com.facebook.react.uimanager.ReactPropGroup <methods>; }

# okhttp

-keepattributes Signature
-keepattributes *Annotation*
-keep class com.squareup.okhttp.** { *; }
[...]

PR #16197, opened 2017-10-04
Documentation about ABI split in Android Release

SuhairZain:
It leads to a reduction of 5-7MB for an app, which is a considerable saving when it comes to emerging markets. Also, I had written a [comment](https://github.com/facebook/react-native/issues/5037#issuecomment-223758335) on an issue which seems to have helped at least 26 people.


TheSavior:
Thanks for contributing!

PR #16204, opened 2017-10-04
Correct logo URL

gusgard:
Update logo URL
https://facebook.github.io/react/img/logo_og.png (old) to https://facebook.github.io/react/logo-og.png (new)

## Test Plan

Check that the old URL shows a "Page Not Found" and the new URL the correct image.


bvaughn:
FWIW this should probably point to https://reactjs.org/logo-og.png 😄

TheSavior:
Thanks for contributing! 

Feel free to send another PR updating to @bvaughn's recommended URL.

gusgard:
@TheSavior Sure 👍

gusgard:
@bvaughn It would be possible to move this files (`logo_small.png` &
`logo_small_2x.png`) to the static folder

https://reactjs.org/logo-small.png is not working.

Because is used:
https://github.com/facebook/react-native/blob/2d2dfa26bcef2e11b22c089974180ad8a6865158/RNTester/js/ImageExample.js#L199 
https://github.com/facebook/react-native/blob/2d2dfa26bcef2e11b22c089974180ad8a6865158/RNTester/js/ImageExample.js#L198



@TheSavior The integration tests fails because of this URL
https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/IntegrationTests/ImageCachePolicyTest.js#L68  we could use https://reactjs.org/logo-og.png, unless the size of the image is important for the test.


Let me know so I can make the PR 😄

bvaughn:
Why are `react-native` integration tests reliant on the `react` website 😅 

Yes, it would be possible to move the image logos to the `static` folder to be accessible at the URL you describe, but I'm not sure continuing to support that dependency is a good idea to be honest.

gusgard:
I agree with you, the best could be to have the static files in the web of react-native, like this one (http://facebook.github.io/react-native/img/header_logo.png). Do we have a fancy URL, like https://reactjs.org ?

bvaughn:
They don't need to be hosted with a fancy URL. They could be loaded:

1. from the github.io website, eg [facebook.github.io/react-native/img/header_logo.png](http://facebook.github.io/react-native/img/header_logo.png), or better yet...
1. directly from GitHub (in a folder beside the test so their use is clear), eg [raw.githubusercontent.com/facebook/react-native/master/RNTester/js/hawk.png](https://raw.githubusercontent.com/facebook/react-native/master/RNTester/js/hawk.png)

gusgard:
Yes, could be a good idea to have all the images used (with URLs) in a folder inside RNTester or IntegrationTests.

PR #16207, opened 2017-10-04
Correct logo URL

gusgard:
Added static files to remove dependency with `react` website, related PR https://github.com/facebook/react-native/pull/16204

## Test Plan

After is merged to master, check if all the URLs are working correctly.

@TheSavior


hramos:
Per discussion on #16204 I think we'll want to use a static asset hosted on the React Native site itself.

gusgard:
Yes, makes more sense. I will update the PR to use a URL like this one https://raw.githubusercontent.com/facebook/react-native/master/RNTester/js/hawk.png

Libraries/Image/Image.ios.js
```diff
@@ -53,7 +53,7 @@ const ImageViewManager = NativeModules.ImageViewManager;
  *         />
  *         <Image
  *           style={{width: 50, height: 50}}
- *           source={{uri: 'https://facebook.github.io/react/logo-og.png'}}
+ *           source={{uri: 'https://facebook.github.io/react-native/img/logo-og.png'}}
```
hramos: This image is not loading for me. The opengraph image for RN is https://facebook.github.io/react-native/img/opengraph.png

IntegrationTests/ImageCachePolicyTest.js
```diff
@@ -65,7 +65,7 @@ class ImageCachePolicyTest extends React.Component<Props, $FlowFixMeState> {
         <Text>Hello</Text>
       <Image
         source={{
-              uri: 'https://facebook.github.io/react/img/logo_small_2x.png?cacheBust=notinCache' + Date.now(),
+              uri: 'https://facebook.github.io/react-native/img/logo-small-2x.png?cacheBust=notinCache' + Date.now(),
```
hramos: This image is not loading for me, either.

shergin:
I think that we should not have unnecessary ***external*** image references. It's hard to manage and it increases page loading time.

IntegrationTests/ImageCachePolicyTest.js
gusgard: These URLs won't work until are added to the website page (after merged the PR)

Libraries/Image/Image.ios.js
gusgard: I prefer `logo-og.png` since the logo is in the center and has a dark background and all the examples in RNTester fit better, what do you think? 

image

hramos: Sounds good to me.

IntegrationTests/ImageCachePolicyTest.js
hramos: If this were a simple website change, it would be alright as the website is versioned and the references to the new images will only go live once the images are deployed.

In this case, I think the images won't be visible on https://facebook.github.io/react-native/ for up to two months due to the way the website is deployed.

Let's first use existing images in the repo itself. Maybe update existing images instead, if that does not break anything unrelated?

gusgard: We could use this URL raw.githubusercontent.com/*gusgard*/react-native/*correct-url-logo*/website/src/react-native/img/logo-small.png that don't depend on the website. (changing user and branch for facebook and master). 

Otherwise, I could use the images that are already in the img folder and test it well.

gusgard:
@hramos Updated to use images with this URL format `https://facebook.github.io/react-native`

PR #16211, opened 2017-10-05
Use UnimplementedView for CheckBox on iOS

timwangdev:
`CheckBox` component was introduced in v0.49.0 and not implemented on iOS.

Users who are trying to use `CheckBox` on iOS will get a warning that 
> Native component for "AndroidCheckBox" does not exist 

We should declare in the document that this component is Android only and use `UnimplementedView` for iOS.

## Test Plan

- Use `react-native init` new project
- Apply pull request changes
- Add `<Checkbox />` after welcome text in `App.js`
- Run the app in iOS simulator

### DIFF
--- a/Libraries/Components/CheckBox/CheckBox.android.js
+++ b/Libraries/Components/CheckBox/CheckBox.android.js
@@ -27,7 +27,7 @@ type DefaultProps = {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


janicduplessis:
@facebook-github-bot shipit

PR #16259, opened 2017-10-09
Fix `eslint` in RNTester

grabbou:
CC: @hramos @tomasreimers - please check if prettier is running on your setup properly :)


janicduplessis:
@grabbou My PR landed so we can close this.

Issue #16255, opened 2017-10-09
[DOCS] Links not working (tiny.cc)


gusgard:
### Is this a bug report?

Yes, in docs.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Steps to Reproduce

https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/Easing.js#L168

### Reproducible Demo

http://tiny.cc/elastic_b_1 (bounciness = 1, default)
http://tiny.cc/elastic_b_3 (bounciness = 3)

Since the example (http://easings.net/#easeInElastic) is clear, I think we should delete the links.

hramos:
Sounds good. I'd approve a PR.

gusgard:
Done 👍

PR #16262, opened 2017-10-09
Remove broken links

gusgard:
Issue https://github.com/facebook/react-native/issues/16255

## Test Plan

Remove broken links


shergin:
@janicduplessis Seems we have to update links. 😄

PR #16192, opened 2017-10-03
[Pods] Set fishhook's header_dir to "fishhook"

ide:
This lets us import fishhook like `<fishhook/fishhook.h>`.

Test Plan: Used this in a CocoaPods project with RN and was able to compile without needing to edit the existing fishhook import.

Fixes: #16039
Should address #16130 without needing to modify source that affects FB

PR #16271, opened 2017-10-10
[IOS] [BUGFIX] [Fishhook] - Correct fishhook import in RCTReconnectingWebSocket Fixes #16039

### LINKED ISSUE #16039: [RN 0.49-rc.5] Fishhook error

macdoum1:
RCTReconnectingWebSocket is not compiling correctly because of an incorrect import (https://github.com/facebook/react-native/issues/16039).

## Test Plan:
Everything build and run as usual.

## Release Notes:
[IOS] [BUGFIX] [Fishhook] - Correct fishhook import in RCTReconnectingWebSocket Fixes #16039

### DIFF
--- a/Libraries/WebSocket/RCTReconnectingWebSocket.m
+++ b/Libraries/WebSocket/RCTReconnectingWebSocket.m
@@ -9,7 +9,12 @@
 
 #import <React/RCTConvert.h>
 #import <React/RCTDefines.h>
+
+#if __has_include(<React/fishhook.h>)
+#import <React/fishhook.h>
+#else
 #import <fishhook/fishhook.h>
+#endif
 
 #if __has_include(<os/log.h>) && defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 100300 /* __IPHONE_10_3 */
 #import <os/log.h>


Zyphrax:
Please merge this one in, it is causing a lot of confusion with CocoaPods users.

bithavoc:
please merge

ide:
I suspect this isn't the desired fix since this hasn't been an issue within FB's codebase or else they would have fixed the compiler error. The Xcode project needs to be set up to treat `fishhook` as a header directory.

DylanNWatt:
What's the nuance between this and `#import <React/fishhook.h>`

macdoum1:
I've moved on from our implementation of react-native for now, but this likely would only be an issue for the cocoapod installation method. A `#import <React/fishhook.h>` is a module import, whereas a `#import "fishhook.h` is a file import.
Either one would work I believe, just not `#import <fishhook/fishhook.h>` because Cocoapod subspecs are not built as modules.

macdoum1:
@ide FB's codebase might not use react-native as a cocoapod so they wouldn't see this issue. Not sure.

bithavoc:
@ide header_path doesn't work, there is a PR for it https://github.com/facebook/react-native/pull/16192

ide:
> FB's codebase might not use react-native as a cocoapod so they wouldn't see this issue. Not sure.

Yes, and their builds haven't been failing because of this so I suspect the issue is with how the pods are configured.  To be on the same page I would like us to figure out how to fix compilation issues (fwiw I've tested with pods and don't run into this) without modifying the import to not impact FB's build.

macdoum1:
@ide The issue here is that cocoapod subspecs are not built as modules so the existing import will never work with cocoapods. From the commenters above and on the related issue it appears this happens with new cocoapod installations, so i'd imagine existing installations would work purely from some leftover DerivedData magic. 

It appears none of the travis jobs/tests cover a cocoapod installation which is why this issue has probably gone undetected.

@jgsamudio mentioned in the issue thread that removing `!use_frameworks` resolved this issue for anyone following along: https://github.com/facebook/react-native/issues/16039#issuecomment-341144015 . However, if your project has any swift cocoapods this is not an option.

tirrorex:
@macdoum1 the real issue here is people working in objective-c not taking into account the swift ecosystem, we see that with google projects everyday already

makovkastar:
The issue is still there in 0.51.0.

barbieri:
still happening with 0.53.3

msand:
And in 0.54.0

msand:

msand:

justColbs:
This works for me, but now I can't `import React` without getting the "Could not build Objective-C module React" error. Any ideas? Is this related?

cobarx:
I'm also seeing the issue that @justColbs described. Any idea how to work around or fix this?

keshavkaul:
@macdoum1 Can you change the import to use `<React/fishhook.h>` because FB has implemented `#import <React/RCTDefines.h>` just a line above.

macdoum1:
@keshavkaul Done, adding release notes & test plan now

macdoum1:
@keshavkaul Done, adding release notes & test plan now

ikesyo:
This should brake non-CocoaPods installation so is not a correct fix. The technique described in https://github.com/facebook/react-native/issues/13198#issuecomment-371749528 would be the right direction:

```objc
#if __has_include(<React/fishhook.h>)
#import <React/fishhook.h>
#else
#import <fishhook/fishhook.h>
#endif
```

macdoum1:
@ikesyo Good point, making that change now

macdoum1:
@ikesyo Done!

ikesyo:
You should also rebase onto current `master` to make CI run correctly.

macdoum1:
@ikesyo Rebased this morning, should be good to go

ikesyo:
@macdoum1 It seems that the parent commit of this PR is https://github.com/facebook/react-native/commit/780938170b0add35e6a4151892154fa0aa5f6b57 on 10 Nov 2017, not the today's HEAD of `master`.

macdoum1:
Alright, now we should be up to date

hramos:
0.55.4 was released several months ago - as you can see above, the fix was merged to master 4 days ago. You can expect this in a future release - 0.57 most likely (0.56 was released last month).

ikesyo:
Is the cherry-pick request to 0.56 rejected?

https://github.com/react-native-community/react-native-releases/issues/14#issuecomment-401224900

hramos:
@ikesyo best to keep that discussion in the release thread.

PR #16272, opened 2017-10-10
Added secureTextEntry does not work with multiline

LAITONEN:
The title speaks for itself. Docs regarding secureTextEntry of TextInput were not descriptive enough. Owing to that, it took me more than an hour of debugging to find the issue of why the TextInput in my app was not hiding the input with secureTextEntry.


janicduplessis:
Thanks for the clarification!

@facebook-github-bot shipit

PR #16293, opened 2017-10-11
Android event documentation - Adding part about mapping event names.

tdekoning:
I was quite lost when the documentation told me to go figure the event mapping out myself. It took me quite a while to figure out that i needed to register the names in the `ViewManager`.

This code snippet just makes it way easier to figure out what you need to do to add events.


janicduplessis:
Awesome! Thanks for the PR.

@facebook-github-bot shipit

PR #16301, opened 2017-10-11
Remove mockFs dependency

grabbou:
This is first PR from the series I am going to be sending as a result of fixing 0.50-stable test suite. This one removes `mockFS` dependency that has been causing failures on Node 6.x container.

Here's build before this change: https://circleci.com/gh/facebook/react-native/22529
Here's build after this change: https://circleci.com/gh/facebook/react-native/22538 (green)

Note that the CI may be still red as there are other PRs to be addressed. You can see this in the wild on 0.50.


grabbou:
CC: @skevy

PR #16302, opened 2017-10-11
Fix Analyze step on CI

grabbou:
Recent changes that introduced Circle 2 (thanks @ide and @hramos for work on this) include special step for analyzing code. It takes PR number and processes the build.

Unfortunately, that breaks all non-PR builds (including ones scheduled by me as a part of release step). 

This PR simply checks if such env variable is set and stops executing in case it's undefined.

Also, I have updated the order of the tests so that most important (unit tests) are no longer shadowed by temporary eslint and flow breakage. The reason for this change is that flow has been broken for ~20 days which shadowed breakage in unit tests (addressed in my other PR).

You can see build breaking before this change: https://circleci.com/gh/facebook/react-native/22391
And being green after: https://circleci.com/gh/facebook/react-native/22530

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,15 +92,19 @@ jobs:
       - checkout
       - run: npm install --no-package-lock
       - run: |
+          npm test -- --maxWorkers=2
           npm run lint
           npm run flow -- check
-          npm test -- --maxWorkers=2
-# eslint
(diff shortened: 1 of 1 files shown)


grabbou:
You can see this running in the wild on 0.50-stable branch.

PR #16303, opened 2017-10-11
Fix broken releases script

grabbou:
Fails on my machine due to fact that `replace` returns an instance of a String, rather than an instance of ShellString (that includes `to` on its prototype). 

Solution is to use an explicit `writeFileSync`. You can see that change in the wild on 0.50-stable branch.

CC @janicduplessis (edit by @hramos)

### DIFF
--- a/scripts/bump-oss-version.js
+++ b/scripts/bump-oss-version.js
@@ -15,7 +15,7 @@
  * After changing the files it makes a commit and tags it.
  * All you have to do is push changes to remote and CI will make a new build.
  */
-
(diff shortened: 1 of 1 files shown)


hramos:
This was stuck because of the CC:.


We have an issue where any PR that has CC: or Test Plan: in the description is likely to fail during the import phase, because Phabricator tries to handle those fields. In this case, it failed because it could not find an employee with Janic's handle. I need to address this via Dangerfile because it's unlikely we'll fix it on our side soon.

PR #16311, opened 2017-10-11
[Android] Allow PackagerConnectionSettings to be provided by ReactNativeHost

jamesreggio:
This PR enables a `ReactNativeHost` to customize the `PackagerConnectionSettings` supplied to the bundle loader and dev tools.

I've been working to productionize an implementation of the WebWorkers API for React Native and I've found these changes necessary to support remote debugging of each worker instance. (Each instance requires a custom bundler port, which can now be plumbed through via the `PackagerConnectionSettings`.)

This change should be a no-op for existing apps.

### Test Plan

This code is working in my local, WebWorker-enabled apps, and does not regress the ability to start and remotely debug the `react-native init` boilerplate app on Android.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
@@ -12,8 +12,8 @@ public class AndroidInfoHelpers {
   public static final String GENYMOTION_LOCALHOST = "10.0.3.2";
   public static final String DEVICE_LOCALHOST = "localhost";
 
-  private static final int DEBUG_SERVER_HOST_PORT = 8081;
-  private static final int INSPECTOR_PROXY_PORT = 8082;
+  public static final int DEBUG_SERVER_HOST_PORT = 8081;
+  public static final int INSPECTOR_PROXY_PORT = 8082;
 
   private static boolean isRunningOnGenymotion() {
     return Build.FINGERPRINT.contains("vbox");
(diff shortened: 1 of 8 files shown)


ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
```diff
@@ -12,8 +12,8 @@
   public static final String GENYMOTION_LOCALHOST = "10.0.3.2";
   public static final String DEVICE_LOCALHOST = "localhost";
 
-  private static final int DEBUG_SERVER_HOST_PORT = 8081;
-  private static final int INSPECTOR_PROXY_PORT = 8082;
+  public static final int DEBUG_SERVER_HOST_PORT = 8081;
+  public static final int INSPECTOR_PROXY_PORT = 8082;
```
jamesreggio: There was no compelling reason to keep these values private. My WebWorker library checks for conflicts against these defaults (and utilizes the `getServerIpAddress` below to avoid duplicating logic to find the loopback interface).

jamesreggio:
@janicduplessis @shergin — this goes hand in hand with #16312. Please consider.

janicduplessis:
If I understand correctly the main goal of this is to be able to change the packager address the app tries to connect to in DEV mode?

jamesreggio:
@janicduplessis, that is correct. The packager is only ever used in development.

jamesreggio:
Hey @janicduplessis, answered your question above. This is dev only.

PR #16312, opened 2017-10-11
[Android] Randomize the downloaded bundle filename

jamesreggio:
This PR randomizes the cached filename for JS bundle downloads. It's necessary to avoid collisions if multiple bundles are in use by the same app.

I've been working to productionize an implementation of the WebWorkers API for React Native and I've found these changes necessary to prevent a race condition in the local bundle cache.

### Test Plan

This code is working in my local, WebWorker-enabled apps, and does not regress the ability to start and remotely debug the react-native init boilerplate app on Android.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManagerImpl.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManagerImpl.java
@@ -101,7 +101,7 @@ public class DevSupportManagerImpl implements
 
   private static final int JAVA_ERROR_COOKIE = -1;
   private static final int JSEXCEPTION_ERROR_COOKIE = -1;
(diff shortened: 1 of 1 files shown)


janicduplessis:
Thanks for the PR, looks good.

@shergin Do you think this could break anything internally at fb?

jamesreggio:
@janicduplessis @shergin — is there anything preventing you from merging this PR?

janicduplessis:
Actually this change only affects dev mode so it should be safe.

@facebook-github-bot shipit

shergin:
Can this cause a polltion the app's data directory by many outdated files?

janicduplessis:
@shergin Good point, I think it will.

@jamesreggio Would it work if we add `mJSAppBundleName` instead of a random number?

PR #16317, opened 2017-10-11
Update AppState.js

diegorodriguesvieira:
Add constructor method.


javache:
The existing example is valid, it uses the [stage 3 class fields](https://github.com/tc39/proposal-class-fields) feature.

PR #16326, opened 2017-10-12
Restore Danger on Circle 2

hramos:
This was omitted during the migration to Circle 2 because Danger support landed after the Circle 2 PR was originally opened.

Test Plan

We'll see how Circle 2 handles this config, hopefully the bot will comment here.


grabbou:
Let's wait for master to go green and test. I also need to make sure this doesn't break on release branch.

hramos:
Circle isn't running here, so I'll have to create a new PR.

PR #16348, opened 2017-10-12
[CI] Publish to npm on stable and master branches

hramos:


.circleci/config.yml
```diff
@@ -346,6 +365,72 @@ jobs:
...
+# Run Danger      
+      - run:
+          name: Analyze Pull Request
+          command: |
+            if [ -n "$CIRCLE_PULL_REQUEST" ]; then
+              cd danger && DANGER_GITHUB_API_TOKEN="e622517d9f1136ea8900""07c6373666312cdfaa69" npm run danger
+            else
+              echo "Skipping pull request analysis."            
+            fi
+          when: always
```
grabbou: Wish there was a way to run on pull request builds only in a more declarative way

.circleci/config.yml
```diff
@@ -366,6 +451,21 @@ workflows:
               only:
                 - /.*-stable/
                 - master
+  deploy:
+    jobs:
+      - publish-npm
+    branches:
+      only:
+        - /.*-stable/
+        - master
```
grabbou: This will publish ALWAYS from `stable` regardless of the tests. We should either wait for other workflows programatically to complete (no API yet that I am aware that allows it - you can only queue up jobs within particular workflow) or make this job "manual", meaning that I have to go to CircleCI and manually click "Deploy" . I actually believe that would be better approach. Many times, CI builds fail because of Flow, flakiness or something. Having manual way to approve build regardless of the failure would make it easier to avoid such issues in the future.

hramos: I'm leaning towards taking advantage of the manual approval, for the reasons mentioned. I need to look into who would have the necessary permission to approve the deploy to go out - I wouldn't want to hold you back from pushing out a release if this is only available to FB employees. (It's midnight here, I'll check up on this in the morning)

.circleci/config.yml
hramos: Same.

hramos:
I have another PR open which finally triggered builds. Turns out the way I was filtering out jobs was invalid, but Circle was not showing an error - it was dropping the build silently.

grabbou:
Yeah, I just found it - leaving comments there

.circleci/config.yml
grabbou: Okay. I think write access might be needed. If I need to undergo special security constraints here or something (like use a separate account), happy to comply with it. I think given my past experience, ti would be great to not wait for FB since you guys are pretty busy anyway!

PR #16349, opened 2017-10-12
[Travis] Switch to Xcode 9

hramos:
Test Plan

Run on Travis


hramos:
Master is already failing, and I'd like to wait until master is green. That said, it looks like the tests are successfully starting up on Xcode 9, which is a good sign.

hramos:
We're switching to Circle.

PR #16354, opened 2017-10-13
Migrate Travis over to Circle

grabbou:
This pull request migrates Travis to Circle and pre-starts iOS simulators / tvOS ones as advised in documentation to speed up builds. It also uses Xcode 9.0 to build and test apps.

Note that podspec test is failing and it's been failing for a while on Travis as well (since podspec has been changed to use Cxx bridge by default). I've notified few folks here of that and we are looking to fix this test, but it's not related to the scope of this PR.

Also, previously, podspec tests were only runninng on master (disabled for PR builds) where I think it makes more sense to run them on PR builds as well (all in all, we want to prevent breakage before merging). That said, I've removed `if` check to make it run on all builds.

Other small changes:
- cache `node_modules` properly (previously defined as restore_cache and save_cache but not referenced in following jobs)


grabbou:
All tests are green but `ci/circleci: test-podspec` - I am going to fix its setup. Rather than that, looks to me that everything is green.

grabbou:
CC: @alloy would you mind helping me to figure out what's up with `podspec` suite? It's been failing on Travis for a while and chances are it's actually an error that got hidden.

grabbou:
All the other tests are passing. I believe Travis can be disabled when this is merged.

.circleci/config.yml
```diff
@@ -90,29 +101,19 @@ jobs:
...
+      - run: *install_node_dependencies
       - run: |
           npm test -- --maxWorkers=2
           npm run lint
           npm run flow -- check
-# eslint - doesn't run on non-PR builds
-      - run:
-          name: Analyze Code
-          command: |
-            if [ -n "$CIRCLE_PR_NUMBER" ]; then
```
grabbou: This is re-added in @hramos PR as a separate job.

alloy:
@grabbou I will spend some time tonight to see if an idea I have pans out, afterwards I will be able to give you an estimate of how long it should take to bring it to completion.

hramos:
<3 <3 <3

.circleci/config.yml
```diff
@@ -79,56 +83,116 @@ aliases:
...
+      - run: *run-node-tests
+
+  # Runs unit tests on iOS devices
+  test-objc-ios:
+    <<: *defaults
+    macos:
+      xcode: "9.0"
+    dependencies:
+      pre:
+        - xcrun instruments -w "iPhone 5s (10.3.1)" || true
```
hramos: We can do this separately from this PR, but at some point we should switch over to iOS 11.0 here, and tvOS 11.0 in `test-objc-tvos`. This would match what we're testing internally.

.circleci/config.yml
```diff
@@ -79,56 +83,116 @@ aliases:
       - ~/watchman
     key: v1-watchman-v4.9.0
 
+  - &install-node-dependencies
+    |
+      npm config set spin=false
+      npm config set progress=false
+      npm install --no-package-lock
```
hramos: We can consolidate this into a single command:

`npm install --no-package-lock --no-spin --no-progress`

(Using the separate `npm config` commands is good when changing the preferences globally, but we're mostly running `npm install` on separate images where the preference won't persist)

.circleci/config.yml
```diff
@@ -79,56 +83,116 @@ aliases:
...
       - image: circleci/node:6.11.0
     steps:
       - checkout
-      - run: npm install
-      - run: |
-          npm test -- --maxWorkers=2
-          npm run lint
-          npm run flow -- check
+      - restore-cache: *restore-node-cache
+      - run: *install-node-dependencies
```
hramos: `--no-package-lock` is new to npm@5, which is not used in the Node 6 and Node 4 tests. It's why I initially didn't set up a npm install alias to use everywhere. I do think it's alright to use it here, because earlier npm versions will silently ignore the flag.

janicduplessis:
Glad we're moving away from 2 different CI providers!!

PR #16365, opened 2017-10-14
Fix $FlowFixMes from TabBarIOS

cdlewis:
## Motivation

There are a number of $FlowFixMe statements in TarBarIOS.ios.js as a result of recent Flow upgrades introducing new errors/warnings. I had a stab at removing these statements and introducing what are hopefully sensible types.

## Test Plan

Only types were changed so `yarn flow` should be sufficient.

## Release Notes

[INTERNAL] [MINOR] [TarBarIOS] - |Fix $FlowFixMes|


Libraries/Components/TabBarIOS/TabBarIOS.ios.js
```diff
@@ -20,19 +20,22 @@ const ViewPropTypes = require('ViewPropTypes');
...
+class TabBarIOS extends React.Component<ViewProps & {
+  style?: StyleObj,
+  unselectedTintColor?: string,
+  tintColor?: string,
+  unselectedItemTintColor?: string,
+  barTintColor?: string,
   barStyle?: 'default' | 'black',
   translucent?: boolean,
   itemPositioning?: 'fill' | 'center' | 'auto',
+  children: React.Element<any> | React.Element<any>[],
```
TheSavior: This should probably be [`React.Node`](https://flow.org/en/docs/react/children/)

cdlewis:
@TheSavior I updated this PR to use React.Node as suggested.

TheSavior:
Thank you very much for contributing this. We love PRs improving our flow types!

cdlewis:
Thanks @TheSavior <3

PR #16367, opened 2017-10-14
Return request id in jest mock for requestAnimationFrame

rumax:
## Motivation

`Jest` allows to use fake timers and according to https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame the return value should be `A long integer value, the request id, that uniquely identifies the entry in the callback list`. This allows to use `cancelAnimationFrame`.

In current implementation of `jest/setup.js`, the return value is undefined. Therefore it's not possible to cancel the animation frame request.

### Some implementation with requestAnimationFrame
```
let id = null;
const registerCallback = (callback) => {
  clearCallback();
  id = requestAnimationFrame(() => {
    id = null;
    callback();
  });
};
const clearCallback = () => {
  if (null !== id) {
    cancelAnimationFrame(id);
    id = null;
  }
};
```

### Test
```
jest.useFakeTimers();

const callback = jest.fn();
registerCallback(callback);
clearCallback();

jest.runAllTimers();
expect(callback).toHaveBeenCalledTimes(0); // Will be error in current implementation, since nothing is cleared. And test will pass, after MR is merged
```

This is fake example, but the real usage is when the animation frame request should be removed on `ComponentWillUnmount`.

## Release Notes
 [JEST] [BUGFIX] [requestAnimationFrame] - return request id


janicduplessis:
Nice, thanks for the fix!

@facebook-github-bot shipit

PR #16380, opened 2017-10-15
[WIP] Upgrade connect

yuya-takeyama:
## Motivation

To suppress below warning while installation

```
warning connect@2.30.2: connect 2.x series is deprecated
```

## Test Plan

Have no idea as of now.
CI only?

## Release Notes

 [INTERNAL] [MINOR] [package.json] - Upgrade connect to v3.6.5

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "base64-js": "^1.1.2",
     "chalk": "^1.1.1",
     "commander": "^2.9.0",
-    "connect": "^2.8.3",
+    "connect": "^3.6.5",
     "create-react-class": "^15.5.2",
     "debug": "^2.2.0",
     "denodeify": "^1.2.1",


janicduplessis:
Thanks for the PR!

Could you just update the test plan to explain how you tested that it still works after updating connect, it's a major version bumb so there could be breaking changes. I'm not sure where we use this dependency but you could grep the codebase and make sure to test those features rapidly.

yuya-takeyama:
Thanks for your comment!
Okay, then I'll dig more deeply where it's used and update the test plan.

kelset:
Have you tried it before submitting the PR?

Because using ´yarn resolutions´ I was able to try to use `connect@3.6.5` and the packager closed itself with the error: `connect.compress is not a function`

lxe:
What's preventing this from being landed?

yuya-takeyama:
Let me close this for now.

PR #16403, opened 2017-10-16
[general][enhancement] Move JS-native version check to its own module + unit tests + prefix Obj-C macro w/RCT

ide:
- The version check that ensures the JS and native versions match is now in its own module for two reasons: it is easier to test and it allows react-native-windows to override just this module to implement its own version check (ex: more advanced checks for RNW-specific code).
- Added unit tests for the version checking to specify its behavior more clearly, including parity between dev and prod to avoid prod-only behavior and mitigate SEVs.
- Prefixed the Obj-C `#define` with `RCT_` to conform with other RN globals.

Test Plan: Ran `yarn jest` to run all tests. Compiled iOS app and launched it, which runs the version check.

### DIFF
--- a/Libraries/Core/ReactNativeVersionCheck.js
+++ b/Libraries/Core/ReactNativeVersionCheck.js
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule ReactNativeVersionCheck
+ * @flow
+ * @format
+ */
+'use strict';
+
+const {PlatformConstants} = require('NativeModules');
+const ReactNativeVersion = require('ReactNativeVersion');
+
+/**
+ * Checks that the version of this React Native JS is compatible with the native
+ * code, throwing an error if it isn't.
+ *
+ * The existence of this module is part of the public interface of React Native
+ * even though it is used only internally within React Native. React Native
+ * implementations for other platforms (ex: Windows) may override this module
+ * and rely on its existence as a separate module.
+ */
+exports.checkVersions = function checkVersions(): void {
+  if (!PlatformConstants) {
+    return;
+  }
+
+  const nativeVersion = PlatformConstants.reactNativeVersion;
+  if (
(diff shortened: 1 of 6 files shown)


Libraries/Core/ReactNativeVersionCheck.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule ReactNativeVersionCheck
+ * @flow
```
janicduplessis: You can add `@format` pragma to tell tools files use prettier.

Libraries/Core/__tests__/ReactNativeVersionCheck-test.js
```diff
@@ -0,0 +1,105 @@
...
+
+  it('passes when the minor matches when the major is zero', () => {
+    _mockJsVersion(0, 1, 0);
+    _mockNativeVersion(0, 1, 0);
+
+    const ReactNativeVersionCheck = require('ReactNativeVersionCheck');
+    expect(() => ReactNativeVersionCheck.checkVersions()).not.toThrow();
+  });
+
+  it("throws when the minor doesn't match when the major is zero", () => {
```
janicduplessis: Can you add a test for major don't match too so we cover all cases

Libraries/Core/__tests__/ReactNativeVersionCheck-test.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
```
janicduplessis: Same `@format` here

Libraries/Core/__tests__/ReactNativeVersionCheck-test.js
ide: Yep, will do. We can't add a test for checking that the major matches since we don't have that logic right now, we'll need to add it later if/when we shift RN versions over by one point (RN 51, RN 52, etc...).

janicduplessis:
@facebook-github-bot shipit

PR #16410, opened 2017-10-16
[DOCS] [ ENHANCEMENT][Navigation.md] - Package manager conflict on install instruction

geronimo794:
The instruction will produce error and delete some package if the user first installation of react native using YARN, and the instruction to install react-navigation library use NPM.

## Motivation

I start to learn react with zero knowledge of es6, package manager. I follow all the basic documentation. Until i come to Navigating Between Screens section. I got error after install the react-navigation package using NPM

## Test Plan
This is what happen after i use npm install --save react-navigation
error again
After that i cant run my react native on android
error again 2
After some googling i realize the problem come from the package manager. I shouldnt use npm because my react native projects installed using yarn.
success

I hope my change can save the millisecond of time for the developers who are still learning like me

## Release Notes

  CATEGORY

[DOCS] [ ENHANCEMENT] [Navigation.md] - Add package installation instruction option

### DIFF
--- a/docs/Navigation.md
+++ b/docs/Navigation.md
@@ -27,6 +27,13 @@ The first step is to install in your project:
 npm install --save react-navigation
 ```
 
+Or you can use yarn to install the package in your project :
+
+```
+yarn add react-navigation
+```
+Make sure to use the same package manager, when you first install the react native project to prevent lost package.
+
 Then you can quickly create an app with a home screen and a profile screen:
 
 ```


hramos:
We've been purposefully avoiding listing Yarn in the docs, to avoid issues like these. The docs assume you're using npm everywhere to install the packages. We make the assumption that if you're using Yarn, you'll know which commands to use when our examples list npm.

I wouldn't merge this PR with this single change, but I'd take a look at a PR that updates all of the onboarding docs to include Yarn as an option. This is no small change, as npm is mentioned in several places and we want to avoid confusing newcomers who don't know the difference between npm and yarn.

PR #16419, opened 2017-10-16
Relax error conditions for InitializeCore

AaronFriel:
## Motivation

This patch addresses an issue with non-native platforms using `expo` and `react-native-windows` with React Native 0.49 and 0.50, and conveniently also ensures `react-native-web` does not error either, if that has not been updated.

## Test Plan

I tested using an Android build, `expo`, `react-native-windows`, and `react-native-web`. For the supported platforms (`android`, `ios`), the control flow should be unchanged. This should only affect third party platforms, e.g.: `expo` on Android/iOS, Windows, and web.

## Release Notes

[INTERNAL] [ENHANCEMENT] [InitializeCore] Relax version constraints for 3rd party platforms.

## Did I do this right?

I don't know if I made this pull request correctly but I sure tried to.

If this PR is accepted, I would like it to be backported to 0.49.

### DIFF
--- a/Libraries/Core/InitializeCore.js
+++ b/Libraries/Core/InitializeCore.js
@@ -150,7 +150,7 @@ if (PlatformConstants) {
   console.warn(
     'React Native `PlatformConstants` not defined.\n\n' +
     `${PlatformMissingModuleMessage} ${PlatformProblemPersistsMessage}`
-  )
+  );
 }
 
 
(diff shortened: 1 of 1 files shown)


janicduplessis:
Hey @AaronFriel ! We landed https://github.com/facebook/react-native/commit/7733d4023728f32a940e9e4a31f79e70455871a8 recently to address this issue, it is now possible for other platforms to override the version check module. Will that work for your use cases?

AaronFriel:
I'm not sure, but it looks like the new module returns immediately when `PlatformConstants` doesn't exist, I think that would resolve the issue.

AaronFriel:
@janicduplessis Not all of my dependencies target ~0.50 yet. I do believe this resolves the issue, however.

PR #16435, opened 2017-10-17
Delegate to ProgressBarAndroid from ActivityIndicator on Android, instead of the other way around

brentvatne:
## Motivation

`ProgressBarAndroid` regressed after fixing a bug in https://github.com/facebook/react-native/commit/ccddbf82d79833702f19d9f6ed1ca23755655187

## Test Plan

Run this gist on a new project with this code: https://gist.github.com/brentvatne/a0b57e5bbae1bd2cf76765ea27f077af

Notice that you will see:

screen shot 2017-10-17 at 11 06 03 am

hmmm... doesn't seem right 🤔 

With the patch in this PR applied, you will see:

screen shot 2017-10-17 at 11 01 38 am

oh! there we go 😄 

## Release Notes

[ANDROID] [BUGFIX] [ProgressBarAndroid] - Fix regression in ProgressBarAndroid which limited `styleAttr` to only `Regular`.

### DIFF
--- a/Libraries/Components/ActivityIndicator/ActivityIndicator.js
+++ b/Libraries/Components/ActivityIndicator/ActivityIndicator.js
@@ -14,8 +14,9 @@
 const ColorPropType = require('ColorPropType');
 const NativeMethodsMixin = require('NativeMethodsMixin');
 const Platform = require('Platform');
-const React = require('React');
+const ProgressBarAndroid = require('ProgressBarAndroid');
 const PropTypes = require('prop-types');
+const React = require('React');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
 const ViewPropTypes = require('ViewPropTypes');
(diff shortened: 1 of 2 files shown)


ide:
@facebook-github-bot shipit

bvaughn:
I will follow up on this internally.

Just a quick note though that for the time being, this change has been reverted (within Facebook) because it caused a bunch of yellow box deprecation warnings (the one I pointed out above) to be shown. I _think_ the correct fix at this point is to re-land this change along with removing that deprecation warning, but I want to run this by others first to make sure there's no concerns with it.

ide:
Brent has a commit up to fix the warning too. The longer-term fix we were thinking about is to split into three components, two that are public (ActivityIndicator and ProgressBar) that use a private third component underneath.

bvaughn:
Might be easiest to just let me do that internally along with the rollback of the rollback, if you don't mind. (I've already created the diff.)

bvaughn:
Okay. The internal diff (D6092598) has landed re-adding this change and removing the deprecation message. It should sync to GitHub soon.

PR #16437, opened 2017-10-18
Fix Flow errors at declaration of most major library components

Ashoat:
The relevant changes in the PR are to Libraries/StyleSheet/EdgeInsetsPropType.js; the rest are just removals of FlowIgnores.

The definition of the relevant types is [here](https://github.com/facebook/flow/blob/master/lib/react.js#L262-L271).

The long and short of it is that for whatever reason, Flow is unable to realize that `ReactPropsChainableTypeChecker` is a subtype of `ReactPropsCheckType` unless we assert it. Once we explicitly hint this to the typechecker, it realizes that `EdgeInsetsPropType` is indeed a valid React PropType, and stops complaining that it isn't.


sahrens:
Sweet, let's see if it lands :)

Issue #10712, opened 2016-11-03
Lineheight behaves differently on ios and android


MarcoPolo:
### Description

On ios: 
![screen shot 2016-11-02 at 4 58 11 pm](https://cloud.githubusercontent.com/assets/594035/19951795/a78bac2e-a11d-11e6-90da-4b0b1e5c5728.png)

On Android:
![screen shot 2016-11-02 at 4 57 40 pm](https://cloud.githubusercontent.com/assets/594035/19951806/b0f2b88e-a11d-11e6-89c8-2dbc7f65cae6.png)

note how the on android it seems like the line height is only applied on top of the text, while on ios it's applied to both top and bottom. This is also an issue when you have lineheights < fontSize. On ios it trims from the top, but on android it'll trim from top and bottom, so the text doesn't line up and get's truncated differently. 

### Reproduction

RNplay demo: https://rnplay.org/apps/OJFGEg

### Additional Information

* React Native version: 0.36
* Platform: iOS & Android
* Operating System: MacOS

ugiacoman:
Just ran into this issue, can confirm still exists

dsimmons:
Ditto, can confirm that I'm still seeing this on RN v0.45.0 (latest).

It's not apparent to me if this is intended behavior based on the corresponding underlying native capabilities, or rather an unintentional discrepancy.

I'd just like to add that it certainly makes achieving a consistent layout between both platforms challenging, in achieving parity with pre-designed screens. I've resorted to disabling `lineHeight` on Android in the interim, as I'm unable to control how exactly the text appears.

Is this on anyone's radar by chance?

henrytao-me:
I did a little test on CustomLineHeightSpan.java. It works on new pure Android project but not on RNTesterApp. I am still investigating what the different between ReactNative TextView and PureAndroid TextView. I would like to pair with ReactNative core team to work on this investigation (if possible). 

The idea is simple: use `descent` and `baseline` as base metrics, then recalculate `bottom`, `ascent` and `top` based on `fontSize` and `lineHeight`. `ascent` and `descent` will be centered. 

```java
// Copyright 2004-present Facebook. All Rights Reserved.

package com.facebook.react.views.text;

import android.graphics.Paint;
import android.text.style.LineHeightSpan;

/**
 * We use a custom {@link LineHeightSpan}, because `lineSpacingExtra` is broken. Details here:
 * https://github.com/facebook/react-native/issues/7546
 */
[...]

PR #16448, opened 2017-10-18
 Inverted descent/ascent Android prioritisation to match iOS lineHeight behaviour

bartolkaruza:
## Motivation

We noticed that on Android the lineHeight behaviour is different from iOS for built in fonts and custom fonts. The problem becomes visible when the lineHeight approaches the fontSize, showing a cut-off on the bottom of the TextView. This issue has been raised before in #10712. There is a mention of a PR with a fix in that issue, which has not been merged yet. This implementation is a less intrusive fix leaving the current lineHeight approach in place and fixing the discrepancy only.

This proposed change prioritises ascent over descent for reduction, making the lineHeight functionality behave identical to iOS.

## Test Plan

There is no existing test covering the lineHeight property and its behaviour in the CustomLineHeightSpan. This PR contains new unit tests that covers the various scenario's for the lineHeight calculations.

The original behaviour, before the change can against these unit tests. The case that fails is `shouldReduceAscentThird`, which can be made to succeed on the old code by changing the asserts to:
```
    assertThat(fm.top).isEqualTo(-5);
    assertThat(fm.ascent).isEqualTo(-5);
    assertThat(fm.descent).isEqualTo(-4);
    assertThat(fm.bottom).isEqualTo(-4);
```
The unit test succeeds for the current implementation, which has the values for ascent and descent inverted.

Below screenshots show before, after and iOS:

BEFORE
screen shot 2017-10-18 at 15 35 41

AFTER
screen shot 2017-10-18 at 15 37 02

iOS
screen shot 2017-10-18 at 15 35 22

## Release Notes

[ANDROID] [BUGFIX] [Text] - Fix the lineHeight behaviour on Android to match iOS

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
@@ -29,34 +29,25 @@ public class CustomLineHeightSpan implements LineHeightSpan {
     // The general solution is that if there's not enough height to show the full line height, we
     // will prioritize in this order: descent, ascent, bottom, top
 
-    int height = mHeight;
-    int currentHeight = -fm.top + fm.bottom;
-    int delta = currentHeight - height;
-    if (delta > 0) {
-      int bottomSpaceAvailable = fm.bottom - fm.descent;
-      int topSpaceAvailable = -fm.top - -fm.ascent;
-      if (topSpaceAvailable > 0 && delta > 0) {
-        int topSpaceTaken = topSpaceAvailable;
-        if (delta < topSpaceTaken) {
-          topSpaceTaken = delta;
-        }
-        delta -= topSpaceTaken;
-        fm.top = fm.ascent - (topSpaceAvailable - topSpaceTaken);
-      }
-      if (bottomSpaceAvailable > 0) {
-        int bottomSpaceTaken = bottomSpaceAvailable;
-        if (delta < bottomSpaceAvailable) {
-          bottomSpaceTaken = delta;
-        }
-        delta -= bottomSpaceTaken;
-        fm.bottom = fm.descent + bottomSpaceAvailable - bottomSpaceTaken;
-      }
+    if (-fm.ascent > mHeight) {
+      // Show as much descent as possible
+      fm.bottom = fm.descent = mHeight;
+      fm.top = fm.ascent = 0;
+    } else if (-fm.ascent + fm.descent > mHeight) {
+      // Show all descent, and as much ascent as possible
+      fm.bottom = fm.descent;
+      fm.top = fm.ascent = -mHeight + fm.descent;
+    } else if (-fm.ascent + fm.bottom > mHeight) {
+      // Show all ascent, descent, as much bottom as possible
+      fm.top = fm.ascent;
+      fm.bottom = fm.ascent + mHeight;
+    } else if (-fm.top + fm.bottom > mHeight) {
+      // Show all ascent, descent, bottom, as much top as possible
+      fm.top = fm.bottom - mHeight;
+    } else {
+      // Show proportionally additional top
+      final int additional = mHeight - (-fm.top + fm.bottom);
+      fm.top -= additional;
     }
-
-    fm.top += delta;
-    if(delta > 0) {
-      fm.ascent = Math.max(fm.top, fm.ascent - delta);
-    }
-
   }
 }
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
@@ -36,12 +36,25 @@ public class CustomLineHeightSpan implements LineHeightSpan {
     // The general solution is that if there's not enough height to show the full line height, we
     // will prioritize in this order: descent, ascent, bottom, top
 
-    // Show proportionally additional ascent / top & descent / bottom
-    final int additional = mHeight - (-fm.top + fm.bottom);
-
-    fm.top -= additional / 2;
-    fm.ascent -= additional / 2;
-    fm.descent += additional / 2;
-    fm.bottom += additional / 2;
+    if (-fm.ascent > mHeight) {
+      // Show as much descent as possible
+      fm.bottom = fm.descent = mHeight;
+      fm.top = fm.ascent = 0;
+    } else if (-fm.ascent + fm.descent > mHeight) {
+      // Show all descent, and as much ascent as possible
+      fm.bottom = fm.descent;
+      fm.top = fm.ascent = -mHeight + fm.descent;
+    } else if (-fm.ascent + fm.bottom > mHeight) {
+      // Show all ascent, descent, as much bottom as possible
+      fm.top = fm.ascent;
+      fm.bottom = fm.ascent + mHeight;
+    } else if (-fm.top + fm.bottom > mHeight) {
+      // Show all ascent, descent, bottom, as much top as possible
+      fm.top = fm.bottom - mHeight;
+    } else {
+      // Show proportionally additional top
+      final int additional = mHeight - (-fm.top + fm.bottom);
+      fm.top -= additional;
+    }
   }
 }
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
@@ -54,7 +54,9 @@ public class CustomLineHeightSpan implements LineHeightSpan {
     }
 
     fm.top += delta;
-    fm.ascent = Math.max(fm.top, fm.ascent - delta);
+    if(delta > 0) {
+      fm.ascent = Math.max(fm.top, fm.ascent - delta);
+    }
 
   }
 }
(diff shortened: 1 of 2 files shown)


bartolkaruza:
@andreicoman11 Would you be able to review this PR as the original author of CustomLineHeightSpan?

bartolkaruza:
@shergin Hi, any chance you could review this PR? Thanks

bartolkaruza:
@hramos This PR is a small consistency fix, with a big positive impact; bringing Android and iOS behaviour in line and fixing broken behaviour on Android.
Before this fix text rendering on Android is just plain broken and there are multiple parties that I know of that are maintaining forks for this fix alone. Try setting the lineHeight to a a value close to 1x the fontSize and you will see the issue right away. 
I'm having trouble getting this reviewed though. Could you help me/us out by getting someone's attention for a review? Thank you :)

bartolkaruza:
@andreicoman11 Thanks for taking a look. I've changed the fix to be in the same style as your original implementation. It should be easier to review. 
The reason I changed the style of implementation was that I wanted an implementation that could easily handle a re-ordering of the priorities. Considering this will not likely happen anytime soon after this PR, let's just keep it simple. 👍

ReactAndroid/src/test/java/com/facebook/react/views/text/CustomLineHeightSpanTest.java
```diff
@@ -0,0 +1,74 @@
+package com.facebook.react.views.text;
```
andreicoman11: This is missing the license header, mind adding this so we can land this change?

ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java
```diff
@@ -45,10 +45,9 @@ public void chooseHeight(
       // Show all ascent, descent, bottom, as much top as possible
       fm.top = fm.bottom - mHeight;
     } else {
-      // Show proportionally additional ascent and top
+      // Show proportionally additional top
       final int additional = mHeight - (-fm.top + fm.bottom);
       fm.top -= additional;
-      fm.ascent -= additional;
```
dickie81: @bartolkaruza & @andreicoman11:
I've been looking at this PR and noticed a possible regression. Line height seems to be broken on wrapped lines of text when line height is greater than font size. It seems to be related to this line of code.

Here is an example of what I'm seeing:
screen shot 2017-11-03 at 11 50 19

JSX:
```
<View style={{backgroundColor: "yellow", paddingTop: 100, paddingBottom: 100}}>
    <Text style={{fontSize: 20, lineHeight: 60, backgroundColor: "blue", color: "white"}}>
        Fontsize 20 Lineheight 60 qwertyuiopasdfghjklzxcvbnm QWERTYUIOPASDFGHJKLZXCVBNM
        Fontsize 20 Lineheight 60 qwertyuiopasdfghjklzxcvbnm QWERTYUIOPASDFGHJKLZXCVBNM
    </Text>
</View>```

Comment (dickie81) [reply ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java]: Restoring `fm.ascent -= additional;` gives me this behaviour:
screen shot 2017-11-03 at 11 56 40

Comment (dickie81) [reply ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java]: Proposed fix to make Android consistent with iOS:
```
      // Show proportionally additional ascent / top & descent / bottom
      final int additional = mHeight - (-fm.top + fm.bottom);

      fm.top -= additional / 2;
      fm.ascent -= additional / 2;
      fm.descent += additional / 2;
      fm.bottom += additional / 2;
```
Giving following behaviour:
screen shot 2017-11-03 at 12 11 40

Comment (bartolkaruza) [reply ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java]: @dickie81 Thank you for looking out, I did not check the full range of test cases on the actual UI when changing the style back to the original style, leaning on the unit tests instead. I'll try out your approach against the cases I have locally for all the lineHeight scenario's and post the test setup so everyone can verify.

Comment (bartolkaruza) [reply ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java]: @dickie81 I've been testing the ui effects and there is definitely a regression in the case of bigger lineHeight. Unfortunately your proposed solution still causes cut-off at the bottom, when the lineHeight is small.
@andreicoman11 I'm looking into this more in depth, this PR is not ready yet as it stands now.

Here is the testing setup that you can use to verify on your end: https://gist.github.com/bartolkaruza/b74f228636bb78fea00aa8bd82e3766d

And these are the 4 different implementation compared to iOS. Not one is quite right yet.

Current PR state after review changes:
---
new-style-bartolkaruza

Original implementation before PR:
---
original-andreicoman11

Implementation before review changes:
---
original-bartolkaruza

Proposed change by @dickie81 :
---
proposed-dickie81

**Comment by dickie81** [comment]
@bartolkaruza: Sorry, I was not very clear when I proposed the change. I only meant for the code in the last else statement to be changed: 

```// Copyright 2004-present Facebook. All Rights Reserved.

package com.facebook.react.views.text;

import android.graphics.Paint;
import android.text.style.LineHeightSpan;

/**
 * We use a custom {@link LineHeightSpan}, because `lineSpacingExtra` is broken. Details here:
 * https://github.com/facebook/react-native/issues/7546
 */
public class CustomLineHeightSpan implements LineHeightSpan {
  private final int mHeight;

  CustomLineHeightSpan(float height) {
    this.mHeight = (int) Math.ceil(height);
  }

  @Override
  public void chooseHeight(
          CharSequence text,
          int start,
          int end,
          int spanstartv,
          int v,
          Paint.FontMetricsInt fm) {
    // This is more complicated that I wanted it to be. You can find a good explanation of what the
    // FontMetrics mean here: http://stackoverflow.com/questions/27631736.
    // The general solution is that if there's not enough height to show the full line height, we
    // will prioritize in this order: descent, ascent, bottom, top

    if (-fm.ascent > mHeight) {
      // Show as much descent as possible
      fm.bottom = fm.descent = mHeight;
      fm.top = fm.ascent = 0;
    } else if (-fm.ascent + fm.descent > mHeight) {
      // Show all descent, and as much ascent as possible
      fm.bottom = fm.descent;
      fm.top = fm.ascent = -mHeight + fm.descent;
    } else if (-fm.ascent + fm.bottom > mHeight) {
      // Show all ascent, descent, as much bottom as possible
      fm.top = fm.ascent;
      fm.bottom = fm.ascent + mHeight;
    } else if (-fm.top + fm.bottom > mHeight) {
      // Show all ascent, descent, bottom, as much top as possible
      fm.top = fm.bottom - mHeight;
    } else {
      // Show proportionally additional ascent / top & descent / bottom
      final int additional = mHeight - (-fm.top + fm.bottom);

      fm.top -= additional / 2;
      fm.ascent -= additional / 2;
      fm.descent += additional / 2;
      fm.bottom += additional / 2;
    }
  }
}```

**Comment by dickie81** [comment]
This is the result I get with the above code:
screen shot 2017-11-04 at 12 47 00

**Comment by dickie81** [comment]
@bartolkaruza: It still seems your implemenation before the review changes handled scenarios when line height is less than font sizes better.

**Comment by bartolkaruza** [comment]
There was a bug still in the low lineHeight case and that fix combined with @dickie81's proposed proportional change leads to the below desired results:

final

@andreicoman11 The PR is ready now, the fixes and the license header have been pushed and all tests are passing.

File: ReactAndroid/src/main/java/com/facebook/react/views/text/CustomLineHeightSpan.java [hunk 3]
```diff
@@ -45,10 +52,13 @@ public void chooseHeight(
...
-      // Show proportionally additional ascent and top
+      // Show proportionally additional ascent / top & descent / bottom
       final int additional = mHeight - (-fm.top + fm.bottom);
-      fm.top -= additional;
-      fm.ascent -= additional;
+
+      fm.top -= additional / 2;
+      fm.ascent -= additional / 2;
+      fm.descent += additional / 2;
+      fm.bottom += additional / 2;
```
strindhaug: Why are you using integers? This will cause rounding errors since aditional is an int and 2 is an int so anything that is not divisible by two will be rounded down. (See: https://github.com/facebook/react-native/issues/10712#issuecomment-359028784 )

bartolkaruza: Hi, for us it has solved the problems in the design completely, but I see that there is still something wrong for your design case. Maybe you can open up a new issue for this and see if you can get the attention of other developers, linking to these issues, to get your specific case fixed. Even better to open a new issue, with a PR that fixes it.

tracinah:
Priorbto prioritize is into the beholder!!

On Feb 9, 2018 4:22 AM, "Stein Strindhaug" <notifications@github.com> wrote:

*@strindhaug* commented on this pull request.
------------------------------

In ReactAndroid/src/main/java/com/facebook/react/views/text/
CustomLineHeightSpan.java
<https://github.com/facebook/react-native/pull/16448#discussion_r167213744>:

>        final int additional = mHeight - (-fm.top + fm.bottom);
-      fm.top -= additional;
-      fm.ascent -= additional;
+
+      fm.top -= additional / 2;
+      fm.ascent -= additional / 2;
+      fm.descent += additional / 2;
+      fm.bottom += additional / 2;

Why are you using integers? This will cause rounding errors since aditional
is an int and 2 is an int so anything that is not divisible by two will be
rounded down. (See: #10712 (comment)
<https://github.com/facebook/react-native/issues/10712#issuecomment-359028784>
)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/facebook/react-native/pull/16448#pullrequestreview-95402440>,
or mute the thread

PR #16464, opened 2017-10-19
Fix test suite when running on a feature branch

grabbou:
Test suite for React Native version check relies on `ReactNativeVersion.js` file that is modified on a release branch (it contains values of a real version, not zeros).

That makes it impossible for the build to pass: https://circleci.com/gh/facebook/react-native/23494

This PR mocks it with zero values, just like we mock it in other suites. I am not sure if that is desired, but works for now.

CC @ide

### DIFF
--- a/Libraries/Core/__tests__/ReactNativeVersionCheck-test.js
+++ b/Libraries/Core/__tests__/ReactNativeVersionCheck-test.js
@@ -41,7 +41,7 @@ function _defineCheckVersionTests() {
   });
 
   it('passes when all the versions are zero', () => {
-    jest.dontMock('ReactNativeVersion');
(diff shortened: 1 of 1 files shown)


ide:
Good catch, I was thinking just about FB's internal use case, thanks!

@facebook-github-bot shipit

grabbou:
CC: @hramos shouldn't be `deploy/hold` step hidden here? Otherwise it will stop status badge from becoming "pass", right?

hramos:
Please avoid using "CC:" in the pull request description! I know it's a silly bug, but it will prevent the import from happening every time.

grabbou:
Ah, yeah, you are right! Sorry, I should've remembered this after last time
👌

PR #16473, opened 2017-10-20
Implement a JSTester example for the Dimensions API

dantman:
## Motivation

JSTester is missing a page for the Dimensions API. This also makes it hard to test out some things in the tester app that affect the app's frame when you need to know how they affect the dimensions.

## Test Plan

screenshot_1508500992
simulator screen shot - iphone 7 plus - 2017-10-20 at 05 03 08

## Release Notes
* [INTERNAL] [ENHANCEMENT] [JSTester] - Dimensions example added to JSTester

### DIFF
--- a/RNTester/js/DimensionsExample.js
+++ b/RNTester/js/DimensionsExample.js
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule DimensionsExample
+ * @flow
+ */
+'use strict';
+
+const React = require('react');
+const ReactNative = require('react-native');
+const {
+  Dimensions,
+  Text,
+  View
+} = ReactNative;
+
+class DimensionsSubscription extends React.Component<{dim: string}, {dims: Object}> {
+  state = {
+    dims: Dimensions.get(this.props.dim),
+  };
+
+  componentDidMount() {
+    Dimensions.addEventListener('change', this._handleDimensionsChange);
+  }
+
+  componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


RNTester/js/DimensionsExample.js
```diff
@@ -0,0 +1,61 @@
...
+
+const React = require('react');
+const ReactNative = require('react-native');
+const {
+  Dimensions,
+  Text,
+  View
+} = ReactNative;
+
+class DimensionsSubscription extends React.Component<$FlowFixMeProps, $FlowFixMeState> {
```
janicduplessis: Can you just type those, you can just use Object for the type of state.dims.

dantman:
Fixed.

janicduplessis:
@dantman I'm not seeing the changes

dantman:
O_o I made the fix, pushed it, then I made another fix for a mistake in `@providesModule` and pushed that.

I see the second fix but not the first, so I guess I messed up and undid it. I'll redo that fix.

dantman:
Done... I **think** I see my work in the PR now. That, or I've gotten so sleepy I'm hallucinating visions of a successful git push.

Adamfoust:
How do I stop these emails

dantman:
@Adamfoust Use the unsubscribe button in the right sidebar.

RNTester/js/RNTesterList.ios.js
```diff
@@ -272,6 +272,10 @@ const APIExamples: Array<RNTesterExample> = [
     supportsTVOS: false,
   },
   {
+    key: 'Dimensions',
+    module: require('./DimensionsExample'),
+  },
```
janicduplessis: Looks like you have to specify `supportsTVOS`, it causes a flow error.

dantman: Whoops, did a blind copy from RNTesterList.android since the example works the exact same on both platform. Didn't notice the ios specific property.

Fixed.

janicduplessis:
@facebook-github-bot shipit

PR #16477, opened 2017-10-20
improve fresco docs so that the version is tied to RN

dmueller39:
## Motivation

Some time ago we integrated fresco:animated-gif into our project at version 0.14.1. 
```
   compile 'com.facebook.fresco:animated-gif:0.14.1'
```
Later we upgraded react native several times, and the core library of fresco was updated automatically under ReactAndroid/build.gradle. This did not pose a problem necessarily until there was a version mismatch and the API was broken between fresco common and fresco animated-gifs. It can be an easy to miss detail, but the version of fresco really ought to be explicitly tied to react-native from the get go. This recommendation in the docs future proofs users gradle files against changes in react-native so that their fresco components match.

Furthermore it should prevent upgrades to the fresco library from having to touch Image.ios.js again.

## Test Plan

Create a test app + add the components using the recommended method. 

## Release Notes

[DOCS][ENHANCEMENT][Image.ios.js] - Recommendation to lock fresco component version to fresco version used by react native

### DIFF
--- a/Libraries/Image/Image.ios.js
+++ b/Libraries/Image/Image.ios.js
@@ -108,19 +108,34 @@ const ImageViewManager = NativeModules.ImageViewManager;
  * You will need to add some optional modules in `android/app/build.gradle`, depending on the needs of your app.
  *
  * ```
+ * // use getFrescoComponent to lock your components to the version used by react-native
+ * def getFrescoVersion() {
+ *    String reactGradle = file('../../node_modules/react-native/ReactAndroid/build.gradle').text
+ *    // use the default of the latest version
+ *    def version = '+'
+ *    reactGradle.eachLine {
+ *        if (it =~ /'com.facebook.fresco:fresco:.*'/) {
+ *            version = it.find(/(?<='com.facebook.fresco:fresco:).*(?=')/)
+ *        }
+ *    }
+ *    return version;
+ * }
+ *
+ * def frescoVersion = getFrescoVersion()
+ *
  * dependencies {
  *   // If your app supports Android versions before Ice Cream Sandwich (API level 14)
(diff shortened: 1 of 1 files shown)


dmueller39:
@foghina Can I ask you to look at this to make sure it looks good from a gradle perspective?

Libraries/Image/Image.ios.js
```diff
@@ -108,19 +108,34 @@ const ImageViewManager = NativeModules.ImageViewManager;
  * You will need to add some optional modules in `android/app/build.gradle`, depending on the needs of your app.
  *
  * ```
+ * // use getFrescoComponent to lock your components to the version used by react-native
```
dmueller39: derp not `getFrescoComponent` but `getFrescoVersion` I can update when we're sure we want to move forward with this.

Libraries/Image/Image.ios.js
```diff
@@ -108,19 +108,34 @@ const ImageViewManager = NativeModules.ImageViewManager;
...
+ * def getFrescoVersion() {
+ *    String reactGradle = file('../../node_modules/react-native/ReactAndroid/build.gradle').text
+ *    // use the default of the latest version
+ *    def version = '+'
```
dmueller39: the alternative here is that it just throws an error when it can't find a correct version. doing this will likely just continue the same problem. any thoughts on best practices?

shergin:
@dmueller39 I am sorry, I have no any idea how fresco or graddle work. Since this change affects only OSS workflow, I can help land it if you find any another OSS contributor to approve this.

dmueller39:
@shergin No worries, it seems like a pretty narrow case.

PR #16478, opened 2017-10-20
Implement a StatusBar frame API

dantman:
## Motivation

Status Bars are far too variable to hardcode, they are different on a variety of devices and can change at runtime:

* We worked around the Android issues temporarily with StatusBar.currentHeight. However we have not handled ios,
* The In-Call Status Bar makes the real height of the status bar 40px instead of 20px, though it pushes the app's root down by 20px.
* In iOS 11 it's now standard for the status bar height to be 0 in landscape mode. (The iOS simulator in Xcode 9 seems to do this occasionally on iOS before 11 as well)
* Apple has a new iPhone X that has a 44px status bar height instead of 20px, and it doesn't change height when the In-Call Status bar is shown.

The solution is a new api to get the Status Bar's frame (the height of the translucent Status Bar overlap and the y offset of that overlap if an In-Call Status bar is used).

- A frame is an object with the StatusBar's height and y offset:
  - height: Is the height of the StatusBar that would overlap the app if it were translucent.
  - y: Is the y offset of the StatusBar. This is used in the case of the iOS In-Call Status Bar because it doesn't increase the size of the overlap, it just pushes the content area down another 20px.
- `getCurrentFrame` returns a Promise for the current frame
- The `statusBarFrameDidChange` event allows you to listen to changes in the frame.

## Breaking

This changes the unofficial `getHeight` and `statusBarFrame{Did,Will}Change` events in `NativeModules.StatusBarManager`, which are used by the 3rd party `react-native-status-bar-size` library. This will break that library, but the new API should be able to replace it.

Notable differences:

* These frame objects do not return an `x` and `width` like the old ones. x is never not `0` and `width` is just the same as the viewport you can get with `Dimensions`. The Android version would be pointlessly complex if I tried to calculate the width of the viewport just to provide this.
* The `height` this returns is the "overlap height" of the area that you need to compensate for with padding in your nav bars, drawers, etc... instead of the "real status bar height" that was previously returned. However users of that library who really want the "real height" can just do `frame.y + frame.height` to get it.
* I had to drop `statusBarFrameWillChange` because I cannot do the offset calculation in that event.

I've started a discussion in that library's issue tracker to see how these differences affect users of that library:
https://github.com/jgkim/react-native-status-bar-size/issues/19

## Test Plan

* Open the RNTester app
* Open the "`<StatusBar>`" examples
* Scroll down to the bottom to see the status bar's dimensions. Do various things like changing the orientation, toggling the In-Call status bar, and trying it out on different devices.

<details>
<summary><b>Screenshots</b></summary>

screenshot_1508534406
screenshot_1508534352
simulator screen shot - iphone x - 2017-10-20 at 14 11 59
simulator screen shot - iphone 8 - 2017-10-20 at 14 08 46
simulator screen shot - iphone 8 - 2017-10-20 at 14 08 33
simulator screen shot - iphone 8 - 2017-10-20 at 14 08 29
simulator screen shot - iphone x - 2017-10-20 at 14 07 20
</details>


Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -185,6 +264,60 @@ class StatusBar extends React.Component<{
...
+   *   - `height`: The current height of the status bar.
+   *   - `y`: The y offset of the status bar, if any. This generally only happens
+   *     when the iOS In-Call Status Bar is visible and pushes the app down
+   *     instead of changing the size.
+   */
+  static addEventListener(
+    eventName: EventName,
+    handler: Function
+  ): {remove: () => void} {
+    if (eventName !== FRAME_DID_CHANGE_EVENT) {
```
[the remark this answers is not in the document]
dantman: This is an error check of the arguments to make sure they actually follow the flow spec that makes this supposedly "unreachable".

Do I have to do something to avoid this incorrect warning?

dantman:
Hmmm, I guess the event at least was exposed. Have to decide what to do with StatusBarIOS.

ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
```diff
@@ -50,17 +53,33 @@ public String getName() {
     return "StatusBarManager";
   }
 
-  @Override
-  public @Nullable Map<String, Object> getConstants() {
+  // TODO(dantman): Detect rare status bar height changing events (such as a Phone+Tablet or docking device that can change
```
shergin: Please remove this todo. Without a task number it will not pass linter.

React/Modules/RCTStatusBarManager.m
```diff
@@ -72,35 +70,29 @@ - (dispatch_queue_t)methodQueue
...
-      @"y": @(frame.origin.y),
-      @"width": @(frame.size.width),
-      @"height": @(frame.size.height),
-    },
-  };
-  [self sendEventWithName:eventName body:event];
+  // The In-Call Status Bar pushes the root view down instead of increasing the size of the StatusBar area
+  // that needs to be accounted for in JS. We can fix that by subtracting the y value of the root view
+  // from the StatusBar's height.
+  CGRect rootFrame = RCTPresentedViewController().view.frame;
```
shergin: We cannot rely on `RCTPresentedViewController` here, it simply will not work.

React/Modules/RCTStatusBarManager.m
```diff
@@ -72,35 +70,29 @@ - (dispatch_queue_t)methodQueue
...
-- (void)applicationDidChangeStatusBarFrame:(NSNotification *)notification
+- (void)applicationDidChangeStatusBarFrame
 {
-  [self emitEvent:@"statusBarFrameDidChange" forNotification:notification];
-}
-
-- (void)applicationWillChangeStatusBarFrame:(NSNotification *)notification
-{
-  [self emitEvent:@"statusBarFrameWillChange" forNotification:notification];
+  dispatch_async(dispatch_get_main_queue(), ^{
```
shergin: Why do we need this?

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -185,6 +264,60 @@ class StatusBar extends React.Component<{
...
+   * Adds an event handler. Supported events:
+   *
+   * - `statusBarFrameDidChange`: Fires when the status bar frame has changed.
+   *   The argument to the event handler is a frame object with the keys:
+   *   - `height`: The current height of the status bar.
+   *   - `y`: The y offset of the status bar, if any. This generally only happens
+   *     when the iOS In-Call Status Bar is visible and pushes the app down
+   *     instead of changing the size.
+   */
+  static addEventListener(
```
shergin: Why do we introduce the new listeners APIs (with `remove` function) here?

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -185,6 +264,60 @@ class StatusBar extends React.Component<{
    */
   static currentHeight = StatusBarManager.HEIGHT;
 
+  /**
+   * Get the current frame (height and y offset) of the Status Bar.
+   *
+   * Returns a Promise that resolves to a frame object with a `y` and `height`.
+   */
+  static getCurrentFrame(): Promise<{y: number, height: number}> {
```
shergin: I think we have to name it `top`, not `x`.

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -148,6 +160,73 @@ function createStackEntry(props: any): any {
...
+ *   "height": 24,
+ * }
+ * // iPhone
+ * {
+ *   "y": 0,
+ *   "height": 20,
+ * }
+ * // iPhone with In-Call Status Bar
+ * {
+ *   "y": 20,
```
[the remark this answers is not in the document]
shergin: Oh, I see.

React/Modules/RCTStatusBarManager.m
```diff
@@ -49,17 +49,15 @@ static BOOL RCTViewControllerBasedStatusBarAppearance()
...
 }
 
 #if !TARGET_OS_TV
 
 - (void)startObserving
 {
   NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
-  [nc addObserver:self selector:@selector(applicationDidChangeStatusBarFrame:) name:UIApplicationDidChangeStatusBarFrameNotification object:nil];
-  [nc addObserver:self selector:@selector(applicationWillChangeStatusBarFrame:) name:UIApplicationWillChangeStatusBarFrameNotification object:nil];
+  [nc addObserver:self selector:@selector(applicationDidChangeStatusBarFrame) name:UIApplicationDidChangeStatusBarFrameNotification object:nil];
```
shergin: Why did we remove `willChange` event?

Libraries/Components/StatusBar/StatusBar.js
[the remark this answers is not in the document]
shergin: We don't need this, because it enforced by Flow.
Just move this condition into `invariant`.

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -148,6 +160,73 @@ function createStackEntry(props: any): any {
  * ### Constants
  *
  * `currentHeight` (Android only) The height of the status bar.
```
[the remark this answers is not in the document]
dantman: Sounds reasonable, but it'll probably be a long one since it's used in a lot of things and it's not a simple transition because a lot of people do the really bad thing of hardcoding the `currentHeight` value in a constant, which means it's a tougher transition to using something non-constant.

Libraries/Components/StatusBar/StatusBar.js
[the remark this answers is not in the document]
dantman: iPhone X, 44px status bar that is always 44px in portrait mode.

In-Call Status bar, dynamic status bar that is 40px in portrait mode on non-iPhone X phones. However the app's root view is pushed down 20px when it expands, so navigators still only have 20px of overlap they need to compensate for.

Libraries/Components/StatusBar/StatusBar.js
dantman: Good idea.

React/Modules/RCTStatusBarManager.m
dantman: We can't subtract the y offset of the root view in the willChange event so we can't calculate the height of the portion of the status bar people actually care about from within that event.

React/Modules/RCTStatusBarManager.m
dantman: I had issues with iOS deciding to return `0` instead of a proper value sometimes for the y/height, even `notification` didn't always behave right IIRC.

React/Modules/RCTStatusBarManager.m
dantman: Do you know of a more reliable way to get the y-offset of the app? We can't rely purely on the `statusBarFrame`'s height because iOS reports 40px for the In-Call Status bar, but if navigators use that value then it'll get pushed down 20px more than it should.

ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
dantman: Shame, StatusBar.js had a similar type of comment in it.

Libraries/Components/StatusBar/StatusBar.js
dantman: This is how NetInfo, AppState, and most of the classes that have events do it, so I just copied that. What is the best practice for new classes with events?

Libraries/Components/StatusBar/StatusBar.js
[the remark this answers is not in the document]
dantman: What is the preferred way to deprecate a property in RN?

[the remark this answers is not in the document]
shergin: I don't remember. 😄 Search for `deprecate` in .js files.
Okay, let's deprecate it in another PR.

React/Modules/RCTStatusBarManager.m
shergin: Oh, I see what you mean, you want to have statusBarFrame to be relative to your RootView position. Right? The problem here is that you cannot assume that:
 * You have only one RootView at any particular the moment;
 * Your RootView is not moving.

You probably should use `<SafeAreaView>` to get same effect.
I would start from simple API where statusBarFrame always return absolute values.

React/Modules/RCTStatusBarManager.m
shergin: I am affraid, that kind of hacks are unacceptable. This notification already fires on the main thread, and this `nextTick` looks very hacky and unreliable.

React/Modules/RCTStatusBarManager.m
```diff
@@ -72,35 +70,29 @@ - (dispatch_queue_t)methodQueue
...
-  NSDictionary *event = @{
-    @"frame": @{
-      @"x": @(frame.origin.x),
-      @"y": @(frame.origin.y),
-      @"width": @(frame.size.width),
-      @"height": @(frame.size.height),
-    },
-  };
-  [self sendEventWithName:eventName body:event];
+  // The In-Call Status Bar pushes the root view down instead of increasing the size of the StatusBar area
```
shergin: We probably need to add `RCTAssertMainQueue()` here.

ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
```diff
@@ -50,17 +53,33 @@ public String getName() {
...
     final Context context = getReactApplicationContext();
     final int heightResId = context.getResources()
       .getIdentifier("status_bar_height", "dimen", "android");
-    final float height = heightResId > 0 ?
+    return heightResId > 0 ?
       PixelUtil.toDIPFromPixel(context.getResources().getDimensionPixelSize(heightResId)) :
       0;
+  }
 
+  @Override
```
shergin: Where do we use this?
Exporting constants from Module turns off some "lazy" performance optimization, so if we don't really need this, we have to remove it.

shergin:
@dantman  I forgot to say, thank you for your effort! I believe, this is really valuable feature. 👍

ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
dantman: This is the source of the `StatusBar.currentHeight` property. All I've done for Android is move most of the code in `getConstants` to a common method that is used by both `getConstants` so the old `currentHeight` still works till we drop it. And the new cross-platform `getHeight`.

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -185,6 +264,55 @@ class StatusBar extends React.Component<{
...
+   *   The argument to the event handler is a frame object with the keys:
+   *   - `height`: The current height of the status bar.
+   *   - `top`: How far the top of the status bar is offset from the top of the
+   *     screen. This generally only happens when the iOS In-Call Status Bar is
+   *     visible and pushes the app down instead of changing the size.
+   */
+  static addEventListener(
+    eventName: EventName,
+    handler: Function
+  ): {remove: () => void} {
```
grabbou: AFAIK we can use type: `EmitterSubscription` and return `StatusBarManagerEventEmitter.addListener` directly? That way, we don't have to hold `_subscriptions` and `removeEventListener` at all. I feel like `add/removeListener` is the old pattern in RN and the newest APIs are leaning towards the subscription oriented approach.

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -185,6 +264,55 @@ class StatusBar extends React.Component<{
...
+    return {
+      remove: () => StatusBar.removeEventListener(eventName, handler)
+    };
+  }
+
+  /**
+   * Removes the event listener. Do this in `componentWillUnmount` to prevent
+   * memory leaks
+   */
+  static removeEventListener(
```
grabbou: Same here, feels like repeated action. These [3 all](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/Libraries/Components/Keyboard/Keyboard.js#L109-L130) are defined anyway, so it might be worth to just directly call them.

grabbou: Also, I would keep names consistent with `Keyboard` module, that is `addListener` instead of `addEventListener` etc.

brentvatne:
I think the only thing this is missing is some way to get synchronously access the status bar height when the app first loads. An example of what I mean is described by @kmagiera in [Living in an async world of React Native at 3:44](https://youtu.be/-Izgons3mec?t=224)

shergin:
Any updates? I like this approach in general, let’s fix all issues and I will land it. :)

dantman:
@shergin I'm mixed on my current feelings.

The issue of finding how much the status bar is not overlapping the content is a difficult one to fix. And the change really isn't useful without it.

react-navigation landed a change that dropped the status bar calculating in favour of a hacky custom SafeAreaView implementation.

And while I don't like the hack of how it was implemented, I started to warm up to and see how SafeAreaView could work if it's implemented right.

The one feature the native SafeAreaView is missing is the ability to control which safe area edges are used (react-navigation's has horizontal/vertical/top/left/bottom/right props that can force it to 'never' apply the inset (so you can horizontally or vertically inset things separately) or 'always' apply the inset (so you can add the inset to the bottom of a scroll view's contents).

So I'm starting to think the better option would be to implement those kinds of behaviours in the native SafeAreaView. And then make an Android implementation of SafeAreaView.

shergin:
@dantman Yeah. I think we need both solution though.
Unfortunately, SafeAreaView has many issues and clearly needs more love. I would love to discuss improvements and help with landing.

PR #16479, opened 2017-10-20
improve docs for KeyboardAvoidingView

gwmccull:
## Motivation

The documentation for `KeyboardAvoidingView` was pretty thin. Tried to fill it out more and corrected a couple words.

## Test Plan

n/a

## Release Notes

[DOCS] [ENHANCEMENT] [KeyboardAvoidingView] - Improve the documentation for the props for KeyboardAvoidingView

### Update names of internal methods on KeyboardAvoidingView

* **Who does this affect**: Users that are manually calling the methods on KeyboardingAvoidingView.
* **How to migrate**: Add an underscore before the name of the method
* **Why make this breaking change**: These methods are not meant to be public. For example, the exposed `onLayout` function is not a prop that accepts a function like is typical of the rest of React Native but is the internal method that is called when the component's onLayout is triggered.
* **Severity (number of people affected x effort)**: Low

### DIFF
--- a/Libraries/Components/Keyboard/KeyboardAvoidingView.js
+++ b/Libraries/Components/Keyboard/KeyboardAvoidingView.js
@@ -54,7 +54,7 @@ const KeyboardAvoidingView = createReactClass({
     ...ViewPropTypes,
     /**
      * Specify how the `KeyboardAvoidingView` will react to the presence of
-     * the keyboard. Will it adjust its height, position or bottom padding?
+     * the keyboard. It can adjust the height, position or bottom padding of the view
      */
     behavior: PropTypes.oneOf(['height', 'position', 'padding']),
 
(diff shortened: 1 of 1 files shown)


gwmccull:
The methods listed in the `KeyboardAvoidingView` docs are not meant to be publicly accessible.  Those are the functions that the component uses itself and are not meant to be called by the user.

I'd like to either hide them from appearing on the docs or add documentation that they should not be used.  What is the preferred approach?

janicduplessis:
@gwmccull Thanks for the PR! To hide the methods you can rename them to start with an underscore (_method).

Libraries/Components/Keyboard/KeyboardAvoidingView.js
```diff
@@ -43,15 +43,19 @@ type KeyboardChangeEvent = {
...
  */
 const KeyboardAvoidingView = createReactClass({
   displayName: 'KeyboardAvoidingView',
   mixins: [TimerMixin],
 
   propTypes: {
     ...ViewPropTypes,
+    /**
+     * Specify how the `KeyboardAvoidingView` will react to the presence of
+     * the keyboard. Will it adjust its height, position or bottom padding?
```
janicduplessis: Can you phrase this not as a question. Something like "Specify if it will adjust ..."

gwmccull:
@janicduplessis thanks for having a look at my PR. I updated the text and renamed those functions.  Technically, this is a breaking change so I updated the PR with the breaking change text but it shouldn't affect too many people unless they're manually calling those functions

Let me know if anything else is needed

Issue #10865, opened 2016-11-10
[WebView] onMessage failing when there are JS warnings and errors on the page


raiderrobert:
### Description

I'm receiving this error (same as screenshot below): 
```
Setting onMessage on a WebView overrides existing values of window.postMessage, 
but a previous value was defined.
```

![screen shot 2016-11-10 at 3 33 31 pm](https://cloud.githubusercontent.com/assets/806571/20193259/3a8a267a-a75b-11e6-9115-2dc4e96f25cf.png)

That specific error message is found in the code base here: https://github.com/facebook/react-native/blob/master/React/Views/RCTWebView.m#L286

### Implementation

I followed the example at this link for onMessage at this link: http://facebook.github.io/react-native/releases/0.37/docs/webview.html#examples

I made a callable `respondToOnMessage` and injected some JavaScript. 

```
class MessagingTest extends React.Component {

    respondToOnMessage = e =>{
        console.log(e);
    };

    render() {
        const jsCode = `window.postMessage('test');`;

        return (
        <WebView
        injectedJavaScript={jsCode}
        source={uri: 'https://www.foo.com/'}}
        onMessage={this.respondToOnMessage}
        />
        )
    }
}
```

### Reproduction

I load the app with this example and I have it pointed to a website (rather not say which one. Sorry.) And the website is emitting 2 errors into the Chrome console when I go there.

```
(index):994 A Parser-blocking, cross-origin script, http://example.com/thing.js, is invoked via document.write. This may be blocked by the browser if the device has poor network connectivity.

widgets.js:8 Uncaught TypeError: Cannot set property '[object Array]' of undefined(…)
```

Other websites like google.com and github.com are just fine. If you want to replicate it, change the `uri` to yahoo.com

### Additional Information

* React Native version: 0.37
* Platform: iOS 10.1, iPhone 6 emulator
* Operating System: Mac OSX Sierra 10.12.1

raiderrobert:
I should also add that I attempted to replicate the error on Android 7.0 and API 24 using a Nexus_5x emulator, and I got NO error. So this seems to be a platform specific error.

sreucherand:
I'm getting the same error. Seems like this exception is thrown when `window.postMessage` has been overwritten. But since `webViewDidFinishLoad` is called multiple times, it tries to overwrite window.postMessage over and over. Not sure though. Can we have you expertise @jacobp100?

jacobp100:
[...]

PR #16502, opened 2017-10-23
RCTWebView bridge redirection error fix

samermurad:
Fixing RCTWebView redirection error 

## Motivation
Inspired by following tickets: [10941](https://github.com/facebook/react-native/pull/10941), [10865](https://github.com/facebook/react-native/issues/10865)
On iOS, when working with RCTWebView and bridging between react native and native JavaScript (i.e: _isMessagingEnabled == YES)
any redirection done on the JavaScript side raises the following error on the dev env: 
`Setting onMessage on a WebView overrides existing values of window.postMessage, but a previous value was defined.`

react-native-bug

Like stated above, this error occurs when the `onMessage` prop in the react native side is set and used.
In order to understand this error and why it happens, one must understand two methods
That The native iOS `UIWebView's` delegate has
#### 1) - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)request navigationType:(UIWebViewNavigationType)navigationType
This method is called each time a redirection occurs on the browsers' JS side, and not just for the main window (i.e: for iframes as well; keep that part in mind).
If the method returns `YES` and only then will the WebView continue to execute the redirection load
#### 2) - (void)webViewDidFinishLoad:(UIWebView *)webView
Each successful load (be it a redirection or a page load) will call this method, pretty simple.

### The Problem
When having a bridge between RN and native JS, the `webViewDidFinishLoad` method
performs an extra check to validate if the `window.postMessage` was left untouched and raises the above mentioned error if it was.
The method then overrides the `window.postMessage` after saving it under`window.originalPostMessage`
i.e:
`window.originalPostMessage = window.postMessage;
window.postMessage = function(data) { /* New code */}
`
While this solution is sufficient for cases where there is a simple one page bridge, it doesn't cover cases where these delegate methods will be called again due to any kind of redirection that doesn't kill the last modified window (i.e: iframes).
This solution only takes care of cases where the main window changes (navigating to a new page).
If these delegate methods will be called again, the first dev check will fail, showing the error; and the `window.originalPostMessage` will be overriding again, deleting the original `window.postMessage` implementation

The applied fix adds two simple checks:
1. Take `window.originalPostMessage` method into consideration and check whether it has the implementation of the `window.postMessage` method (dev env only)
2. Check if the `window.originalPostMessage` exists before reassigning it with a new value

## Test  Plan
1. populate the WebView onMessage prop in react native
2. load the WebView from a custom html
3. add an iframe to the html

behavior before fix: red error screen will be shown
expected behavior/after fix: everything will continue working normally as it should


samermurad:
@shergin 
This fix is a fix to the WebView component in iOS, this code (the WebView implementation) still exists in the latest version of react-native, what can I do to push this pull request further?
If this is stuck due to the android test error, then please take a look a the ci/circleci: test-android, it fails because some sort of a timeout.

Issue #16511, opened 2017-10-23
Change Object.assign polyfill to NOT walk entire prototype chain and to NOT force ideologies


th317erd:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

  Environment:
    OS:  Linux 4.10.0-35-generic #39-Ubuntu SMP Wed Sep 13 07:46:59 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
    Node:  8.1.3
    npm:  5.3.0

  Packages: (wanted => installed)
    react-native: 0.49.3 => 0.49.3
    react: 16.0.0-beta.5 => 16.0.0-beta.5

  Target Platform: Android (7)

### Steps to Reproduce

(Write your steps here:)

1. Create an object with keys `var a = { test: 'derp' }`
2. Create an object with `a` as the prototype `var b = Object.create(a)`
3. Attempt to assign `b` to a new object `var c = Object.assign({}, b)`

### Expected Behavior

react-native shouldn't slowly walk the entire prototype chain or force ideologies

### Actual Behavior

react-native DOES slowly walk the entire prototype chain AND force ideologies

### Reproducible Demo

https://github.com/facebook/react-native/blob/master/Libraries/polyfills/Object.es6.js

Your Polyfill for Object.assign (Object.es6.js) is completely backwards, slow, and forcing an ideology. You are DELIBERATELY walking and copying the entire prototype chain, which is slow, and often not desired. The SUPER INSANE part? You are telling me how to live my life, and throwing an exception if there IS an enumerable property in the prototype chain. Forcing people to do things YOUR way is the wrong way. Please update the polyfill to 1) NOT walk the entire chain, 2) NOT tell me how to code.

By the way, I ran into this simply inheriting from EventEmitter, which is SUPER COMMON by the way.

I am upset that Facebook has written controlling polyfills forcing ideologies and telling me how to live my life. Please stay off my lawn, stop telling me how to code, and stop doing things the slow way just to tell other people how to do things

```javascript
// We don't currently support accessors nor proxies. Therefore this
    // copy cannot throw. If we ever supported this then we must handle
    // exceptions and side-effects.
    
    // No SANE person would ever copy an object with `key in obj`
    // please use `Object.keys(obj)` instead, which is much faster,
    // and doesn't involve forced ideologies
    for (var key in nextSource) {
[...]

PR #16512, opened 2017-10-23
Fix Object.es6.js polyfill being slow and forcing ideologies

th317erd:
Fix for issue #16511


TheSavior:
I replied on the issue so as to keep the majority of the conversation in one place and my comments weren't directly code level.

PR #16533, opened 2017-10-25
Fix SKIP_BUNDLING configuration for iOS

emusgrave:
## Motivation

SKIP_BUNDLING, when enabled, previously exited the script before the ip.txt resource was created, which would cause a runtime failure since the App would not know what address to use to connect to the React Native Bundler. This was hidden by the fact that if you had previously built the app using another configuration, the ip.txt file would exist. But after a 'clean' of the project folder, it would be deleted and the failure would start to occur.

[This is a fix for PR #14731](https://github.com/facebook/react-native/pull/14731)

## Test Plan

Run a Debug build in Xcode targeting a physical device before and after this commit.

## Release Notes
[IOS] [BUGFIX] [./scripts] - Fix SKIP_BUNDLING so that ip.txt resource is still created for debugging on physical devices.

### DIFF
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -10,6 +10,28 @@
 # This script is supposed to be invoked as part of Xcode build process
 # and relies on environment variables (including PWD) set by Xcode
 
+# Print commands before executing them (useful for troubleshooting)
+set -x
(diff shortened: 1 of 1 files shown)


emusgrave:
How do I get this to move forward? The andoid test that failed has nothing to do with this PR. I edited my first post to fix the formatting error in the Release Notes.

emusgrave:
@cooperka @javache This PR has been idle, any chance one of you know who would be a proper reviewer?

PR #16535, opened 2017-10-25
WIP: migrate <WebView> from UiWebView to WKWebView in iOS

raphaeleidus:
# WIP

## Motivation

(Write your motivation here.)

## Test Plan

### DIFF
--- a/React/Views/RCTWebView.h
+++ b/React/Views/RCTWebView.h
@@ -8,6 +8,7 @@
  */
 
 #import <React/RCTView.h>
+#import <WebKit/WebKit.h>
 
 @class RCTWebView;
 
@@ -37,6 +38,10 @@ shouldStartLoadForRequest:(NSMutableDictionary<NSString *, id> *)request
 @property (nonatomic, assign) BOOL messagingEnabled;
 @property (nonatomic, copy) NSString *injectedJavaScript;
 @property (nonatomic, assign) BOOL scalesPageToFit;
+@property (nonatomic, assign) BOOL allowsInlineMediaPlayback;
+@property (nonatomic, assign) BOOL mediaPlaybackRequiresUserAction;
+@property (nonatomic, assign) WKDataDetectorTypes dataDetectorTypes;
+
 
 - (void)goForward;
 - (void)goBack;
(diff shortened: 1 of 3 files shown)


raphaeleidus:
the android test failed but I dont know why, I didn't touch any android files or JS files

joshfriend:
The android build failed because it timed out

hramos:
I'd encourage you to ship a WKWebView-based version of the WebView component as a standalone package instead.

raphaeleidus:
@hramos my goal was to maintain consistent api for both Android and iOS. If I go that approach with a new component should I write a new Android implementation as well?

charpeni:
@hramos  UIWebView/WebView was deprecated by Apple.
> Starting in iOS 8.0 and OS X 10.10, use WKWebView to add web content to your app. Do not use UIWebView or WebView.

I think that moving to WKWebView is the right thing to do, but shall we include this in the core or as standalone? Or as Facebook doesn't use WebView internally, it won't be maintained; maybe we should deprecate and eventually remove all implementations of WebView in the core and rely on separate package.

Here's a standalone version: https://github.com/CRAlpha/react-native-wkwebview

@hramos @shergin your thought?

shergin:
Actually we do recognize this problem internally and we alredy started very similar effort at Facebook to tackle this issue.

There are bunch of issues here though:
 * It is not so trivial to reimplement all existing APIs using WKWebView;
 * Even if we do this perfectly, some apps can still suffer from this because their web surfaces was not tested with WKWebView;
 * We have to tackle some additional related to WKWebView problems, such as better message transport and some security conserns.

So, the current plan is:
 * Create a new iOS-only <WKWebView> component based on WKWebView.
 * Add a special prop to existing <WebView useWebKitEngine={true}> which will delegate all execution to the new component;
 * Test it heavily (while it is opt-in feature).
 * Make it default (after couple of months).
 * Completly replace old one with new one (after additional couple of months).

cc @rsnara

raphaeleidus:
@shergin is the plan for the new ios only wkwebview component to maintain the same or at least mostly overlapping api with the existing webview component?

one reason I am not using the existing https://github.com/CRAlpha/react-native-wkwebview component is that it's API is significantly divergent which means to use a webview I need to write 2 pretty different implementations for android and iOS which I was hoping to avoid. 

Is API consistency a development goal for the new wkwebview component?

shergin:
Yes, API consistency is one of the goals.

bowerman0:
@raphaeleidus Take a look at https://github.com/facebook/react-native/pull/16792, which migrates RCTWebView from UIWebView to WKWebView with the same JavaScript API.

Issue #15724, opened 2017-08-31
[XmlHttpRequest] xhr.upload.onprogress not working with xhr.send({uri: 'file:///...'})


1stman:
I think, it's a bug

I'm trying to upload video with youtube v3 API, looking on [this](https://developers.google.com/youtube/v3/docs/videos/insert) example.

So, I wrote something like this:

  ```
var xhr = new XMLHttpRequest();
  xhr.open('PUT', this.url, true);
  xhr.setRequestHeader('Content-Type', this.contentType); <=== video/mp4
  xhr.setRequestHeader('Content-Range', 'bytes ' + this.offset + '-' + (end - 1) + '/' + this.filesize);
  xhr.setRequestHeader('X-Upload-Content-Type', this.filetype);  <=== video/mp4

  if (xhr.upload) {
    xhr.upload.addEventListener('progress', this.onProgress);
  }
  xhr.onload = this.onContentUploadSuccess_.bind(this);
  xhr.onerror = this.onContentUploadError_.bind(this);
  xhr.send({uri: this.file.path});  <=== file:///.....
```
and it works, i have `xhr.onload` and `xhr.onerror` and video file is uploaded to youtube and played there. 

But i haven't `xhr.upload.onprogress`.

I try `xhr.onprogress`, `xhr.upload.onprogress`, `xhr.upload.addEventListener` - all of this not works.

BUT!
When i use `FormData` 

```
let data = new FormData();
data.append('File', {
      uri:  source.path,
      type: source.type,
      name: source.name
});
var xhr = new XMLHttpRequest();
  xhr.open('PUT', this.url, true);
  //xhr.setRequestHeader('Content-Type', this.contentType);  
  //=== comment this, because it throw error 'multipart != video/mp4'
  //xhr.setRequestHeader('Content-Range', 'bytes ' + this.offset + '-' + (end - 1) + '/' + this.filesize);
  //=== comment this, because sending data size != filesize
  xhr.setRequestHeader('X-Upload-Content-Type', this.filetype);

  if (xhr.upload) {
    xhr.upload.addEventListener('progress', this.onProgress);
  }
  xhr.onload = this.onContentUploadSuccess_.bind(this);
  xhr.onerror = this.onContentUploadError_.bind(this);
  xhr.send(data); <==== FormData
```

`xhr.upload.onprogress` is working. But then youtube api return processing error, because it support only 'video/*' and 'application/octet-stream' and does not support 'multipart/form-data'.

[...]

Issue #11853, opened 2017-01-11
Android xhr.upload.onprogress and addEventListener('progress') not supported?


dguillamot:
### Description

We are doing XMLHttpRequest PUT request (sending base64 encoded image data, setting content type header, etc).

iOS sends progress updates, Android does not. 

We tried setting progress listener the following ways:

xhr.addEventListener('progress')
xhr.upload.addEventListener('progress')
xhr.upload.onprogress = () => {}
xhr.onprogress = () => {}

None of these handlers end up being called on Android, however on iOS they work fine. 

The request completes successfully with a 200. 

### Additional Information

* React Native version: 0.39.2
* Platform: both Android and iOS tested (issue only on Android)
* Operating System: MacOS

dguillamot:
Ended up moving to a native (Obj-c and Java) solution for now which does provide upload progress events on Android. 

Regarding XmlHttpRequest, not sure if the Android interface is different from iOS or not but no luck with XmlHttpRequest updateProgress on Android.

christopherdro:
@dguillamot Looks like there is some sort of API for progress on android.
https://github.com/facebook/react-native/blob/e87e181998bf4f8efba2c7e04bec0929344600b9/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java#L190

dguillamot:
@christopherdro Yes, I did see that last week but xhr.upload.addEventListener('progress')  did not result in any updates on Android - only iOS. For now, we have solved it with a Native solution, thanks though.

bogobogo:
Hey @dguillamot, would you mind sharing your solution? I am facing the same problem, and I currently don't have the knowledge to implement it natively (learning java but that will take me some time...).

GeoffreyPlitt:
I'm on RN=0.42 and using XHR with a Content-Length set before uploading begins. I'm seeing the progress events, but no bytesSent or length or anything else that I can use to compute a progress%. Anybody else have luck with this?

GeoffreyPlitt:
I'm on RN=0.42 and using XHR with a Content-Length set before uploading begins. I'm seeing the progress events, but no bytesSent or length or anything else that I can use to compute a progress%. Anybody else have luck with this?

ranjitpandit:
Any ideas how to get send progress events on Android with bytesSent?

aguzubiaga:
@christopherdro Why is this issue closed? This problem persists.

1stman:
[...]

Issue #10423, opened 2016-10-17
Android crash when network changes from Good to Bad


armadilio3:
### Issue Description

When there are fetch requests made on Android and the network changes from a good connection to a very bad connection, sometimes a crash occurs in RealBufferedSink.java:39.
### Steps to Reproduce / Code Snippets
1. Make constant fetch requests (Once every couple of seconds)
2. Switch between a good network connection and a very bad one (if an iPhone is available it can be used as a hotspot for the Android device and can switch between LTE and very Bad Network)

Switching between the networks should reproduce the problem fairly quickly.
#### Expected Results

The error:

> java.lang.IllegalStateException: closed
>         at okio.RealBufferedSink.write(RealBufferedSink.java:39)
>         at okio.ForwardingSink.write(ForwardingSink.java:35)
>         at com.facebook.react.modules.network.ProgressRequestBody$1.write(ProgressRequestBody.java:58)
>         at okio.RealBufferedSink.flush(RealBufferedSink.java:216)
>         at com.facebook.react.modules.network.ProgressRequestBody.writeTo(ProgressRequestBody.java:48)
>         at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:47)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
>         at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
>         at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:109)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
>         at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
>         at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:124)
>         at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
[...]

PR #16541, opened 2017-10-25
[Android] Added progress updates for all XMLHttpRequest upload types

allengleyzer:
## Motivation

Previously, only form-data request bodies emitted upload progress updates. Now,
other request body types will also emit updates.

Addresses issues:
https://github.com/facebook/react-native/issues/15724
https://github.com/facebook/react-native/issues/11853

This is a bug fix for functionality that's missing on Android. These events are already working correctly on iOS.

## Test Plan

Run the following code on Android, and ensure that events are being sent:
```
const fileUri = 'file:///my_file.dat';
const url = 'http://my_post_url.com/';
const xhr = new XMLHttpRequest();

xhr.upload.onprogress = (event) => {
    console.log('progress: ' + event.loaded + ' / ' + event.total);
}

xhr.onreadystatechange = () => {if (xhr.readyState === 4) console.log('done');} 

console.log('start');

xhr.open('POST', url);

xhr.send({ uri: fileUri }); // sending a file (wasn't sending progress)
xhr.send("some big string"); // sending a string (wasn't sending progress)

const formData = new FormData(); formData.set('test', 'data');
xhr.send(formData); // sending form data (was already working)
```

## Release Notes
[ANDROID] [BUGFIX] [XMLHttpRequest] - Added progress updates for all XMLHttpRequest upload types

Previously, only form-data request bodies emitted upload progress updates. Now,
other request body types will also emit updates.

Addresses issues:
https://github.com/facebook/react-native/issues/15724
https://github.com/facebook/react-native/issues/11853

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -235,8 +235,9 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
     String contentEncoding = requestHeaders.get(CONTENT_ENCODING_HEADER_NAME);
     requestBuilder.headers(requestHeaders);
 
+    RequestBody requestBody;
     if (data == null) {
-      requestBuilder.method(method, RequestBodyUtil.getEmptyBody(method));
+      requestBody = RequestBodyUtil.getEmptyBody(method);
     } else if (data.hasKey(REQUEST_BODY_KEY_STRING)) {
       if (contentType == null) {
         ResponseUtil.onRequestError(
@@ -249,14 +250,13 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
       String body = data.getString(REQUEST_BODY_KEY_STRING);
       MediaType contentMediaType = MediaType.parse(contentType);
       if (RequestBodyUtil.isGzipEncoding(contentEncoding)) {
-        RequestBody requestBody = RequestBodyUtil.createGzip(contentMediaType, body);
+        requestBody = RequestBodyUtil.createGzip(contentMediaType, body);
         if (requestBody == null) {
           ResponseUtil.onRequestError(eventEmitter, requestId, "Failed to gzip request body", null);
           return;
         }
-        requestBuilder.method(method, requestBody);
       } else {
-        requestBuilder.method(method, RequestBody.create(contentMediaType, body));
+        requestBody = RequestBody.create(contentMediaType, body);
       }
     } else if (data.hasKey(REQUEST_BODY_KEY_BASE64)) {
       if (contentType == null) {
@@ -269,9 +269,7 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
       }
       String base64String = data.getString(REQUEST_BODY_KEY_BASE64);
       MediaType contentMediaType = MediaType.parse(contentType);
-      requestBuilder.method(
-        method,
-        RequestBody.create(contentMediaType, ByteString.decodeBase64(base64String)));
+      requestBody = RequestBody.create(contentMediaType, ByteString.decodeBase64(base64String));
     } else if (data.hasKey(REQUEST_BODY_KEY_URI)) {
       if (contentType == null) {
         ResponseUtil.onRequestError(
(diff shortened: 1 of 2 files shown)


fengyibo:
when to shipit @janicduplessis @satya164

janicduplessis:
@facebook-github-bot shipit

iamtekeste:
@janicduplessis Do you know on what version of RN this would be available?

fengyibo:
Who knows on what version of RN this would be available?

rsnara:
@allengleyzer. There was a surge in crashes with a recent release of one of our apps. We traced the crashes down to this diff and consequently reverted it. Once the next release ships, we'll know for sure if this diff was the cause. If not, we'll undo the revert. If so, we'll dig deeper and try to fix the problem.

allengleyzer:
@rsnara If I had to guess, it's probably happening while trying to read an uploaded file's size, since this is an IO operation that wasn't happening before. Looking around, it seems that if an InputStream gets closed (e.g. due to a loss in a user's internet connection), this exception might get thrown.

Do you allow user's to upload certain forms of data in your app?

I'll investigate some more to confirm if this is the cause.

fred2028:
Hi @allengleyzer , yes, in Facebook Ads Manager, a user can upload photos and videos as part of their ad creation/edit process. It is definitely a possibility that a user can lose his or her Internet connection during this process.

allengleyzer:
@rsnara @fred2028 I wasn't able to reproduce the issue. Disconnecting in the middle of an upload didn't raise any exceptions on my device. Tried in both release and debug builds (though, on a debug build, if I was also connected to the JS debugger, then disconnecting at any point usually resulted in an IllegalStateException, though this problem existed before this PR and is due to the JS debugging client, and so wouldn't happen in a release build).

Any chance you have more information, such as a stack trace or file/line number? Also, which RN version does this happen on?

allengleyzer:
@rsnara @fred2028 Well, were you able to confirm whether reverting this change fixed your app? And if it did, do you want to share your crash information so that we can actually fix the issue instead of just reintroducing the RN bug this PR attempted to fix?

I'm going to guess that if this PR was the cause, then most likely the app crashes for the same reason that #10423 was already happening. This PR might cause the issue to happen more frequently due to a `ProgressRequestBody` being created for all uploads, instead of just multipart form-data as before. The fix for this bug would probably just be to insert a try-catch handler on the write operation, to prevent the app from crashing.

LEEY19:
@janicduplessis @allengleyzer guys, are you sure that this commit has fixed the `xhr.upload.progress` issue? I have copied pasted your codes but issue still persisted.

allengleyzer:
@LEEY19 These changes aren't in any release yet, and if you're using the latest master, these changes were recently reverted. I'm not sure what you're RN setup looks like, but that might be the issue?

allengleyzer:
FYI, I figured out a way to reproduce the crash/exception that's in 10423, and was able to confirm that my changes would now cause this same crash to happen more often, since it wouldn't just be form-data bodies that cause it, but all requests that have a request body.

I'm opening a new PR with the changes made here, and an additional commit to fix the crash.

janicduplessis:
👍 Thanks @allengleyzer for being on top of this.

@fred2028 Could you confirm that the crash reported in #10423 was the one you were seeing?

LEEY19:
@allengleyzer I just modified the files `NetworkingModule.java` and `NetworkingModuleTest.java` by copy-pasting your code. Since I built my RN from source changes made to native code will be reflected. 

```
      xhr.upload.onprogress = (evt) => {
        console.log('progress: ' + evt.loaded + ' / ' + evt.total);
      }
```
Nothing appeared in console for the above code on Android. It works as expected in iOS

allengleyzer:
@LEEY19 Sorry about that, the updates don't send if you don't set the Content-Type header on the request. I've updated the test example with setting the Content-Type header.

Another issue I noticed, however, is that the progress updates don't work particularly well for strings/form data, as it usually just sends one progress update at the very end. Unfortunately, this is due in part with the implementation used for sending the updates and/or Okio's implementation for their buffers. There are some possible changes/workarounds, but I think it's out of the scope of this PR. Also, the main use case (IMO) of file updates, is working well.

I made some additional changes in the new PR (referenced above), so you should try that one out, and if you have any issues, let's move the discussion to that one instead of here.

Issue #1402, opened 2015-05-26
[Initialization] Launch screen white flash


alinz:
I just wondering if there is a good practice about LaunchScreen. The reason I'm asking this is, if one adds LaunchScreen, there is a white flash before react-native kicks in and load the app. Is there any way we can prevent this?

brentvatne:
@alinz - this is fixed in the next release: https://github.com/tadeuzagallo/react-native/commit/8d992262ed2dcf0d46e8d22e605764902c33012f

liubko:
@alinz does it look something like https://www.dropbox.com/s/xq522ywqsd16vqe/example.mov?dl=0

alinz:
@liubko Yes, this is exactly what I want to get rid of. I haven't applied the patch yet but I will do it tonight.

oblador:
@brentvatne: I don't see how that patch would fix this problem (if I'm blind/stupid please forgive :-). The problem is that there's a moment between when the app has launched and React is still working. Ideally the LaunchScreen would show until React is finished and fully rendered as it does in a "fully native" app.

brentvatne:
@oblador - ah, this would allow you to transition more smoothly by having the same background colour as the launch screen in the "flicker" moment.

@vjeux - how do you get past this issue in FB apps? I noticed that F8, which if I understand correctly is 100% React Native, transitions directly from the launch screen to the app.

nicklockwood:
There's a (separate) fix for this issue too. Use the new `loadingView` property of `RCTRootView`, and set it to a full-screen `UIImageView` showing your launch image. The code might look something like this:

    UIImageView *launchView = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"MyLaunchImage"];
    rootView.loadingView = launchView;

I've been experimenting with a way to automatically detect the launch image and make this completely automatic, but it's not really ready yet. If you're interested, it looks like this:

   // TODO: support landscape orientation

  // Get launch image
  NSString *launchImageName = nil;
  if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPhone) {
    CGFloat height = MAX(RCTScreenSize().width, RCTScreenSize().height);
    if (height == 480) launchImageName = @"LaunchImage-700@2x.png"; // iPhone 4/4s
    else if (height == 568) launchImageName = @"LaunchImage-700-568h@2x.png"; // iPhone 5/5s
[...]

PR #625, opened 2015-04-02
Add docs about running on device

frantic:
Simple guide about running RN apps on device. Closes #240 

cc @vjeux @spicyj @brentvatne

Issue #240, opened 2015-03-26
Documentation missing instructions for running on device via static bundle


brentvatne:
If anyone wants to improve on these instructions and add them to the docs, feel free, I can do it later if nobody else jumps on it - I just wanted to document it somewhere for now.

- Open `iOS/AppDelegate.m`
- Comment out `jsCodeLocation = [NSURL URLWithString:@"http://localhost:8081/index.ios.bundle"];`
- Uncomment `jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];`
- Start up the packager with `npm start` and then run `curl http://localhost:8081/index.ios.bundle -o main.jsbundle` - if this fails (as it did on my machine) add the `--ipv4` flag to the end of it.
- In XCode, right click on your project directory and click `Add Files to "Project Name Here"` - choose the `main.jsbundle` file that you generated.

You can now run the app on a device (either via usb or TestFlight) without needing to have a server running.

stolemotion:
Thanks man, I was wondering why it would run. For anyone else seeking an answer to "Could not connect to development server" look here.

freshteapot:
Worth noting, (at least unless I am missing something) if you go down this path, you will have to run the curl step, every time. If you want to see your changes.

magus:
I dug around and got all the way to the last step before I had to start sifting through issues, thanks for outlining the steps, now it can actually find the bundle file :+1:

drkibitz:
Just noting here, nothing is requiring you to run a node server, could also just serve a static bundle. Could also make the URL a debug app setting.

AgtLucas:
Thanks!

jonaswindey:
Does this mean that we can host the whole bundle on a remote nodejs server, and always load the bundle (the react part) over http from the client? This would be great to deploy new features without having to rebuild the ios app. Is this realistic or doesn't Apple allow dynamic app functionality like that?

AgtLucas:
@jonaswindey 

As far as I know, this is impossible, Apple doesn't allow this.

jonaswindey:
https://www.meteor.com/blog/2014/10/02/Apple-hot-code-push-mobile

This could be great if we could update all our React logic dynamically without the clients having to be updated.

magus:
[...]

PR #16551, opened 2017-10-26
Update documentation for iOS static bundle

emusgrave:
## Motivation

Include instructions on how to use the static bundle when building an iOS App for production. Also include tips to document the SKIP_BUNDLING flag that was recently added, as well as fix the white background flash when the bundle size is large.

There are a number of issues and past pull requests that tackle this same problem but no one has properly updated the documentation to reflect the conclusions that have been made:
[#240](https://github.com/facebook/react-native/issues/240)
[Updates #625](https://github.com/facebook/react-native/pull/625)
Also Refers to #1402 

## Test Plan

Build an app for production using a Release scheme in Xcode, and see that it does not require the React Native Packager to be running.

## Release Notes
[DOCS][ENHANCEMENT][docs/RunningOnDevice.md] - Document how to use a static bundle for iOS

### DIFF
--- a/docs/RunningOnDevice.md
+++ b/docs/RunningOnDevice.md
@@ -347,7 +347,34 @@ To configure your app to be built using the `Release` scheme, go to **Product** 
 
 ![](img/ConfigureReleaseScheme.png)
 
-### 3. Build app for release
+### 3. Configure app to use static bundle
+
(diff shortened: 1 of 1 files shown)


hramos:

PR #16554, opened 2017-10-26
Code cleanup - Xcode 9 build warning/issue.

nbomberger:
Xcode 9 has compiler settings that are more strict.  This can occur if someone updates there project to use the default settings.  

This patch declares the default type instead of allowing the compiler to determine it.  Instead of `()` we now say `(void)` in a block call.

## Motivation

It was just annoying me, and it has no side effects.  If there are side effects, then we should fix the type and not go with empty to represent void.

## Test Plan

Update project settings in Xcode.  This code doesn't have any known side effects since the compiler assumes the type is void when not declared.

## Release Notes

[DOCS] - Fixed potential compiler build issue on Xcode 9 after updating settings in project.


shergin:
@nbomberger Those two occurrences are all problems of the same type across whole ReactNative repo?

nbomberger:
They are the only ones that I found that cause issues when updating project settings.  I am willing to fix other ones but it seems those are the only spots that won’t let you compile using Xcode.

PR #16569, opened 2017-10-27
Fixed broken image link in example code

arjunnn:
## Motivation
Got frustrated for a minute trying to figure out why the code I copy pasted from the docs wasn't working

## Test Plan
The link I replaced the broken one with is the one used in the iOS device example next to the sample code. It can be found at https://facebook.github.io/react-native/img/favicon.png

## Release Notes
 [DOCS] [MINOR] [Libraries/Image/Image.ios.js] - Broken image link in the example code at


janicduplessis:
Thanks for the fix!

@facebook-github-bot shipit

PR #16572, opened 2017-10-27
【iOS】fix Blob/RCTBlobManager.m dependency RCTWebSocket subspec file RCTWeb…

yygene:
## Motivation
fix bugs

## Test Plan
step 1:
find node_modules/react-native/Libraries/WebSocket -name "*.[h,m]" | xargs grep Blob
find node_modules/react-native/Libraries/Blob -name "*.[h,m]" | xargs grep WebSocket

step:2
find node_modules/react-native/Libraries/WebSocket -name "*.js" | xargs grep Blob
find node_modules/react-native/Libraries/Blob -name "*.js" | xargs grep WebSocket

Blob depend on WebSocket in iOS files
WebSokect depend on Blob in JavaScript files
React.podspec describe iOS dependences

## Release Notes
fix Blob/RCTBlobManager.m dependency RCTWebSocket subspec file RCTWebSocketModule.h

[CATEGORY] [TYPE] [LOCATION] - MESSAGE
 [IOS] [BUGFIX] [React.podspec] - fix Blob/RCTBlobManager.m dependency RCTWebSocket subspec file RCTWebSocketModule.h

### DIFF
--- a/React.podspec
+++ b/React.podspec
@@ -145,6 +145,7 @@ Pod::Spec.new do |s|
 
   s.subspec "RCTBlob" do |ss|
     ss.dependency             "React/Core"
+    ss.dependency             "React/RCTWebSocket"
     ss.source_files         = "Libraries/Blob/*.{h,m}"
     ss.preserve_paths       = "Libraries/Blob/*.js"
   end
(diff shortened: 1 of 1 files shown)


shergin:
Well, RCTBlob depends on RCTWebSocket? Why? Conceptually, that even does not sound right.
Which problem are we solving?

yygene:
@shergin Yes，it does not sound right。But actually，_RCTBlobContentHandler implements RCTWebSocketContentHandler protocol in RCTBlobManager.m

PR #16579, opened 2017-10-28
[cli] Remove built-in templates

mkonicek:
UPDATE: The feature has now been [documented](https://medium.com/@chris.geirman/the-1-2-3s-of-react-native-templates-1f5dda037e11) by Chris Geirman! 💯

## Motivation

I added a [template](https://github.com/facebook/react-native/tree/master/local-cli/templates/HelloNavigation) for react-navigation directly into the React Native repo a long time ago. The feature `react-native init --template foo` supports both:
- Remote templates (`react-native-template-foo` in npm as well as HTTP URLs)
- Local templates that ship with react-native itself

In retrospect, adding the local template wasn't a good idea. Templates should live outside of the React Native repo and be versioned independently. This way templates can be fixed independently of React Native releases, and people can use new templates without having to upgrade React Native.

### Future of the `navigation` template?

As discussed above, removing the template from the repo (merging this PR) is a good idea. The question is whether we should move the template to a separate repo.

Here are a few reasons to delete this template and not move it anywhere else:
- It was undocumented so it's unlikely many people use this template
- `react-navigation` itself contains some [examples](https://github.com/react-community/react-navigation/tree/master/examples)
- Most people now use CRNA to get started so they won't benefit from the `react-native init --template` feature
- Project generators like Ignite are a more popular and maintained alternative
- If we move this template to a separate repo, how to make sure it will stay maintained? If it's unmaintained, it will disappoint people when it stops working.

Here are some reasons to delete the template here but move it to a separate repo ("the template must go on" 💯 ):
- The template might still be useful for people getting started, it works, was recently updated to use `FlatList`([history](https://github.com/facebook/react-native/commits/master/local-cli/templates/HelloNavigation)) and has a nice README.
- If this template lived in the `react-navigation` repo there's a chance the contributors to `react-navigation` would maintain it.

@hramos What do you think? I'm leaning towards not moving this template to a new place unless we can guarantee it stays maintained.

I tested the template using:

`react-native init Test --version facebook/react-native#4c05ede9 --template navigation`

It works:

screenshot 2017-10-28 12 01 18

screenshot 2017-10-28 12 01 22

screenshot 2017-10-28 12 01 28

## Test Plan

CLI e2e tests pass on Cirle CI.

## Release Notes

 [CLI] [BREAKING] [TEMPLATES] - Remove built-in project template for react-navigation. Project templates should live outside the React Native repo and be versioned separately. This used to available via `react-native init --template navigation`.


hramos:
The issue of future maintenance is one reason I have yet to publish a few templates I wrote up earlier this year. I also lean towards removing the template entirely as it's not documented anywhere.

It's still a good example of a project and perhaps it can be shared in some other way.

hramos:
@mkonicek looks like you'll need to sign the CLA before I can land this

mkonicek:
Hi @hramos! I've sent an email and will sing the CLA. Will try to do it this weekend.

local-cli/generator/templates.js
```diff
@@ -13,42 +13,6 @@ const execSync = require('child_process').execSync;
 const fs = require('fs');
 const path = require('path');
 
-/**
- * Templates released as part of react-native in local-cli/templates.
- */
-const builtInTemplates = {
```
hramos: Line 44 in the edited file, or 80 in the original, contains a reference to this `builtInTemplates` const. This was picked up by lint during import.

mkonicek:
Apologies for the long delay @hramos. I rebased and opened a new PR because this one was so old.

https://github.com/facebook/react-native/pull/21155

PR #16580, opened 2017-10-28

simonbengtsson:
Makes sure a recent version of react native is used by specifying that at least version 0.49 should be used for Android. Otherwise an old 0.20.1 version hosted in [jcenter](https://bintray.com/bintray/jcenter/com.facebook.react%3Areact-native) might override the version included from `node_modules`.

## Motivation

Ending up with an ancient version of react native might be confusing to new users trying out react native for the first time.

## Test Plan

NA

## Release Notes
 [DOCS] [MINOR] [IntegrationWithExistingApps.md] - Make sure a `node_modules` version of react native is used

### DIFF
--- a/docs/IntegrationWithExistingApps.md
+++ b/docs/IntegrationWithExistingApps.md
@@ -555,7 +555,7 @@ Add the React Native dependency to your app's `build.gradle` file:
 dependencies {
     compile 'com.android.support:appcompat-v7:23.0.1'
     ...
-    compile "com.facebook.react:react-native:+" // From node_modules.
+    compile "com.facebook.react:react-native:0.49+" // From node_modules.
 }
 ```
 
--- a/docs/IntegrationWithExistingApps.md
+++ b/docs/IntegrationWithExistingApps.md
@@ -555,7 +555,7 @@ Add the React Native dependency to your app's `build.gradle` file:
 dependencies {
     compile 'com.android.support:appcompat-v7:23.0.1'
     ...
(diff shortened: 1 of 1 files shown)


hramos:
Can you tell us a bit more about why such an old version may be used? If we can fix that, it'd be ideal. I don't want to have to update this doc every time we release a new version (0.49 will be considered too old at some point).

simonbengtsson:
If react native is added after `jcenter()` in a gradle file it will use a `0.20.1` version hosted by the Facebook organization there. My guess is that `jcenter()` was used for distribution before but npm was chosen instead in mid july this year?

This gradle file would install version `0.20.1` of react native:

```
    repositories {
        jcenter() // Included in default android projects
        maven {
            url "$rootDir/../node_modules/react-native/android"
        }
```

If you move `jcenter()` after the npm installed react native repository it works however. But I thought it would be helpful for other developers specifying a version later than `0.20.1` which then would automatically skip the `jcenter()` version.

But maybe it is a bad idea to mention a specific version even if it has the greater than tag. How about changing the gradle section to this? I.e specifically mentioning jcenter and moving "..." below the npm react native repository?

```
allprojects {
    repositories {
        maven {
            // All of React Native (JS, Android binaries) is installed from npm
            url "$rootDir/node_modules/react-native/android"
        }
        ... // Other maven repositories such as `jcenter()` goes here
    }
    ...
}
```

hramos:

PR #16585, opened 2017-10-29
Update Blob.js

sm2017:
## Motivation

If you run the following code 

```js
fetch('https://mywebsite.com/path')
      .then((response) => response.arrayBuffer())
```

> TypeError: response.arrayBuffer is not a function

The reason is fetch use whatwg-fetch and we have https://github.com/github/fetch/blob/master/fetch.js#L240-L265 and `support.blob` is false because https://github.com/github/fetch/blob/master/fetch.js#L16

`new Blob()` throw the following error

> TypeError: Cannot read property 'forEach' of undefined
>     at new Blob (node_modules\react-native\Libraries\Blob\Blob.js:95)
>     at node_modules\whatwg-fetch\fetch.js:12
>     at a (node_modules\whatwg-fetch\fetch.js:10)
>     at node_modules\whatwg-fetch\fetch.js:487
>     at loadModuleImplementation (node_modules\metro-bundler\src\Resolver\polyfills\require.js:178)
>     at guardedLoadModule (node_modules\metro-bundler\src\Resolver\polyfills\require.js:130)
>     at _require (node_modules\metro-bundler\src\Resolver\polyfills\require.js:110)
>     at node_modules\react-native\Libraries\Network\fetch.js:17
>     at loadModuleImplementation (node_modules\metro-bundler\src\Resolver\polyfills\require.js:178)
>     at guardedLoadModule (node_modules\metro-bundler\src\Resolver\polyfills\require.js:123)

I change constructor to ` constructor(parts: Array<Blob>=[], options: any)` to fix issue

## Test Plan

N?A

## Release Notes

[GENERAL] [BUGFIX] [Blob] - Fix new Blob() issue

### DIFF
--- a/Libraries/Blob/Blob.js
+++ b/Libraries/Blob/Blob.js
@@ -92,7 +92,7 @@ class Blob {
    * Currently we only support creating Blobs from other Blobs.
    * Reference: https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob
    */
-  constructor(parts: Array<Blob>, options: any) {
+  constructor(parts: Array<Blob>=[], options: any) {
     const blobId = uuid();
     let size = 0;
     parts.forEach((part) => {


shergin:
Okay... but in the specification the first argument is not optional.
https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob

sm2017:
@shergin http://jsfiddle.net/thmqwy8t/ 
It's seems the document of mozilla must be updated

sm2017:
@shergin Can you please accept this PR?

ide:
Reading the W3C spec this looks like it has the right behavior: https://w3c.github.io/FileAPI/#constructorBlob. It'd be nice to have a unit test (just for this case), want to add one?

Libraries/Blob/Blob.js
```diff
@@ -92,7 +92,7 @@ class Blob {
    * Currently we only support creating Blobs from other Blobs.
    * Reference: https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob
    */
-  constructor(parts: Array<Blob>, options: any) {
+  constructor(parts: Array<Blob>=[], options: any) {
```
ide: spaces around `=`

sm2017:
@ide I put spaces , Just tell me where must I create unit test , I don't find correct path

ide:
See other examples in this repo -- make __tests__ directories that are colocated with the files under test.

sm2017:
@ide unit test added https://github.com/facebook/react-native/pull/16585/commits/b4a471b50444ad915f7c7d129fd8ed586052809f

sm2017:
@ide @shergin reply please

Libraries/Blob/__tests__/Blob-test.js
```diff
@@ -0,0 +1,24 @@
...
+  let blob;
+
+  try {
+    new Blob();
+    blob = true;
+  } catch (e) {
+    blob = false;
+  }
+
+  expect(blob).toBe(true);
```
ide: can you write this as `expect(() => new Blob()).not.toThrow()`?

sm2017:
@ide Done

ide:
@facebook-github-bot shipit

hramos:
Here's the test that failed:

```
Libraries/Blob/__tests__/Blob-test.js

Blob construct
    expect(function).not.toThrow()
    
    Expected the function not to throw an error.
    Instead, it threw:
      ReferenceError: Blob is not defined
         \033[90m 12 | 
         \033[90m 13 | test('Blob construct', () => {
        >\033[90m 14 |   expect(() => new Blob()).not.toThrow();
         \033[90m 15 | });
         \033[90m 16 | 
          
          at react-native-github/Libraries/Blob/__tests__/Blob-test.js:14:23
          at Object.<anonymous> (node_modules/expect/build/to_throw_matchers.js:43:7)
          at Object.throwingMatcher [as toThrow] (node_modules/expect/build/index.js:215:24)
          at Object.<anonymous> (react-native-github/Libraries/Blob/__tests__/Blob-test.js:14:48)
     \033[90m 12 | 
     \033[90m 13 | test('Blob construct', () => {
    >\033[90m 14 |   expect(() => new Blob()).not.toThrow();
     \033[90m 15 | });
     \033[90m 16 | 
      
      at Object.<anonymous> (Libraries/Blob
```

PR #16588, opened 2017-10-29
Docs: Improve documentation of Slider (add example)

adrianomelo:
## Motivation

It would be great to have examples in the documentation of all components. I have created a PR to add an example for `CheckBox`, and now I am adding an example for the `Slider` component.

## Test Plan

The PR changes documentation. No further test is required.

## Release Notes

[DOCS][ENHANCEMENT][Slider] - Added example to documentation

### DIFF
--- a/Libraries/Components/Slider/Slider.js
+++ b/Libraries/Components/Slider/Slider.js
@@ -28,6 +28,63 @@ type Event = Object;
 
 /**
  * A component used to select a single value from a range of values.
+ *
+ * ### Usage
+ *
+ * The example below shows how to use `Slider` to change
+ * a value used by `Text`. The value is stored using
+ * the state of the root component (`App`). The same component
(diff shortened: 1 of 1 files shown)


Libraries/Components/Slider/Slider.js
```diff
@@ -28,6 +28,63 @@ type Event = Object;
 
 /**
  * A component used to select a single value from a range of values.
+ *
+ * * ### Usage
```
hramos: Does this render correctly? While this is a docs only change, it's still a good idea to build and test that your change looks good.

adrianomelo: Hi @hramos, thank you for your feedback. It does render correctly. I tested it using Android. I am going to update the commit removing the extra `*`.

And I built `./website` locally, it looks good with the new changes. :)

PR #16591, opened 2017-10-29
[ImgBot] optimizes images

ImgBotApp:
Hey there. I've just optimized all of your images. They are the same dimensions and quality as before, only they take up less space now.

This is a follow up from https://github.com/facebook/react-native/pull/15897 last month.
The PR was not originally accepted since the `ReferenceImages` should be excluded from image compression.

Since then we have rolled out support for `.imgbotconfig`  which I have placed in this PR.
This config file tells ImgBot not to worry about files or folders of your choosing.
You can read about the `.imgbotconfig` [here](https://github.com/dabutvin/imgbot#configuration)

Once installed, ImgBot will respect the `.imgbotconfig` file for every PR
https://github.com/apps/imgbot

:)


janicduplessis:
Not sure how I feed about adding a dotfile just for this, still pretty cool!

cc @TheSavior

TheSavior:
Adding @hramos to this since it impacts the repository. I don't particularly mind adding the config.

hramos:
The dotfile is alright. If the location were configurable, I'd move it to either .github or bots/, but root will do.

dabutvin:
Does the build need to be restarted?

hramos:
Nothing wrong with the changes in the PR itself, the import job timed out due to the amount of assets that need to be pulled down. I can probably get around this by creating an internal diff that applies the same changes as the PR. I'll add it to my list, though it's fairly low-pri.

hramos:
@ImgBotApp the website has been moved to https://github.com/facebook/react-native-website. If you open a PR there with the same changes, I'm sure we can get it merged easily as it won't have to be imported.

PR #16596, opened 2017-10-30
[ANDROID] [ENHANCEMENT] [DevSupport] Use currentActivity to display redbox, loading view and dev menu

kmagiera:
This change aims at replacing SYSTEM_ALERT_WINDOW/OVERLAY API being used for rendering dev support related views on Android (redbox, dev menu, green loading view) with API that does not require any special permission. The permission is still used for displaying perf monitor, although it is no longer requested at app startup but only when perf monitor gets enabled.

This change should not affect the way react native apps work in production environment as in release mode all dev support functionality is disabled.

## Motivation

There are two main reasons why requiring SYSTEM_ALERT/OVERLAY permission for displaying basic dev related windows is problematic:
 1) On Android >=6 devices it is required that overlay permission is granted in device settings for apps being side loaded (not installed via play store which is usually the case for apps being developed). Although this setting is not available on some Android devices including Google's stock Android TV version. On such devices App cannot be granted rights to draw in system alert window which cases the app to crash (instead of showing a redbox or dev menu dialog)
 2) Some Android device vendors have issues with implementation of `Settings.canDrawOverlays` that always return false (I've seen it on Xiaomi Redmi 4A with Android 6.1). This issue because of the following code in [ReactActivityDelegate.java#L90](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java#L90), results in the overlay permission settings screen popping up every time the app is launched even though the permission has been perviously granted which is extremely annoying. Since this change only require overlay permission for displaying perf monitor we no longer ask for it on startup but only when user switches perf monitor ON.

## Test Plan

Test need to be performed on pre Android 6 and post Android 6 devices.

1. Run app with devserver off -> should result in redbox
2. Start packager with --reset-cache flag for the loading bar to be visible for some longer period of time. Then restart the app and see the loading bar show up
3. While the app is running, open dev menu, navigate to "dev settings", test "reload"
4. Modify JS app such that the app crashes, see it display redbox properly. Check if "reload" button works well from the redbox
5. Verify that "Show Perf Monitor" option works as expected. On Android >=6 re-install the app to see it ask for overlay permission at the moment when perf monitor option gets selected.

## Release Notes

 - SYSTEM_ALERT_WINDOW permission will no longer be required on Android to display Redbox

## Breaking
 This change can break things for framework users who provide custom implementation of DevSupportManager interface on Android:

- **Who does this affect**: Owners of apps that use custom implementation of DevSupportManager interface on Android.

- **How to migrate**: Update `create` method of your `DevSupportManager`'s factory to take `ReactInstanceManagerDevHelper` type as a second argument instead of `ReactInstanceDevCommandsHandler`. The interface `ReactInstanceDevCommandsHandler` has been renamed to `ReactInstanceManagerDevHelper` but kept all the methods the same (new method got added). If you were calling one of three methods from `ReactInstanceDevCommandsHandler` interface (`onReloadWithJSDebugger`, `onJSBundleLoadedFromServer` and `toggleElementInspector`) you can call exact same methods directly on `ReactInstanceManagerDevHelper` instance that is being provided in exchange for `ReactInstanceManagerDevHelper `.

- **Why make this breaking change**: 
This PR adds a new method to `ReactInstanceManagerDevHelper` called `getCurrentActivity`. In which case the prev name can no longer be justified. The activity is required for some of the DevSupportManager methods in order to start new dialogs and popups so that overlay permission isn't necessary.

- **Severity (number of people affected x effort)**:
Relatively small (perhaps Fb internally is using DevSupportManager abstraction to provide an alternative implementation but since it isn't documented I doubt anyone else uses it). Effort it very low as it boils down to updating uses of interface `ReactInstanceDevCommandsHandler` with `ReactInstanceManagerDevHelper` (all the methods in `ReactInstanceDevCommandsHandler` stays the same)


ReactAndroid/src/main/java/com/facebook/react/devsupport/ReactInstanceDevCommandsHandler.java
```diff
@@ -31,4 +35,10 @@
    * Request to toggle the react element inspector.
    */
   void toggleElementInspector();
+
+  /**
+   * Return activity react views are mounted in. This will be used by DevSupportManager to launch
+   * dialogs (such as dev settings dialog or a red box)
+   */
+  @Nullable Activity getCurrentActivity();
```
janicduplessis: Would there be another way to pass around the activity? It feels kind of out of place on this interface.

kmagiera:
@janicduplessis we could use this technique, but it makes stuff display in the current activity window. I figured it is fine for redbox and dev menu because when they appear you can't navigate to a different activity. This isn't true for perf overlay that only covers part of the screen and in theory you can open/close activities while you watch the perf overlay. This is a usecase for apps that use some sort of native navigation (homemade or one by wix or airbnb)

ReactAndroid/src/main/java/com/facebook/react/devsupport/ReactInstanceDevCommandsHandler.java
kmagiera: I agree, on the other hand new interface would complicate this part of the codebase even more which I don't think is necessary considering it doesn't do much.

Which of these two options would you prefer:
1) Rename this interface so that this method fits better
2) This interface implementation just passes all the calls to private methods of ReactInstanceManager. I could make those methods public and instead of passing `ReactInstanceDevCommandsHandler` ref to `DevSupportManagerFactory` I could pass ref to `ReactInstanceManager`

PS: I prefer the latter

janicduplessis: Yea if it doesn't require too many changes I think number 2 would be nice

kmagiera: @janicduplessis updated this PR with this change. Note that now it is more likely that this code will break some internal fb stuff where they perhaps use their own implementation of devsupportmanager

kmagiera:
@janicduplessis As my latest changes may break stuff for ppl who provide alternative impl of `DevSupportManager` I added a section to the PRs description with migration guide etc.

ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java
```diff
@@ -589,11 +569,43 @@ public void onHostResume(Activity activity, DefaultHardwareBackBtnHandler defaul
...
+      // b) when activity has just been created
+      // In case of (a) the activity is attached to window and it is ok to add new views to it or
+      // open dialogs. In case of (b) there is often a slight delay before such a thing happens.
+      // As dev support manager can add views or open dialogs immediately after it gets enabled
+      // (e.g. in the case when JS bundle is being fetched in background) we only want to enable
+      // it once we know for sure the current activity is attached.
+
+      // We check if activity is attached to window by checking if decor view is attached
+      final View decorView = mCurrentActivity.getWindow().getDecorView();
+      if (decorView.getWindowToken() == null) {
```
ayc1: Have you tried using https://developer.android.com/reference/android/support/v4/view/ViewCompat.html#isAttachedToWindow(android.view.View) ?

kmagiera: Ah, nice. Didn't know about ViewCompat. Will update that. Thanks for suggesting!

kmagiera:
Hey @achen1 or @ayc1 aby update on this?

ayc1:
hey @kmagiera sorry I didn't have a chance to look at this lately. there has been a few changes to these files recently. do you mind rebasing and fixing the merge conflicts? after that, I'll pull it in to fix the internal build failures due to the removal of ReactInstanceDevCommandsHandler.

kmagiera:
Hey @ayc1. Sorry for the delay. Just rebased this PR on top of the most recent changes from master and have run the test plan once again to make sure everything works as expected. Let me know if I can help with anything more here

ReactAndroid/src/main/java/com/facebook/react/devsupport/DevSupportManagerImpl.java
```diff
@@ -23,13 +23,16 @@
 import android.os.AsyncTask;
 import android.util.Pair;
 import android.widget.Toast;
+
 import com.facebook.common.logging.FLog;
 import com.facebook.debug.holder.PrinterHolder;
 import com.facebook.debug.tags.ReactDebugOverlayTags;
 import com.facebook.infer.annotation.Assertions;
 import com.facebook.react.R;
+import com.facebook.react.ReactInstanceManager;
```
ayc1: this creates a circular dependency 

`//java/com/facebook/catalyst/js/react-native-github/ReactAndroid/src/main/java/com/facebook/react:react -> //java/com/facebook/catalyst/js/react-native-github/ReactAndroid/src/main/java/com/facebook/react/devsupport:devsupport -> //java/com/facebook/catalyst/js/react-native-github/ReactAndroid/src/main/java/com/facebook/react:react`

maybe you should leave the interface in react/devsupport?

kmagiera:
Ah, ok, I’ll update that today

kmagiera:
@ayc1 I restored interface to workaround circular deps issues. I had to rename it though as with `getCurrentActivity` method it no longer sound reasonable to call it `ReactInstanceDevCommandsHandler`, the new name is `ReactInstanceManagerDevHelper`.

I also updated "Breaking" section in PR description to account for that change.

ayc1:
@kmagiera this is causing the green dev loading view to appear behind the activity when you reload JS on old Android versions like API 17 and 19 but not on 23. Can you take a look at this regression?

kmagiera:
Hey @ayc1, sure I will take a look

kmagiera:
@ayc1 I created https://github.com/facebook/react-native/pull/17305 to fix this issue on Kitkat and below

PR #16601, opened 2017-10-30
[iOS] Get rid of AppDelegate.m files and use AppDelegate.swift instead

mcplectrum:
## Motivation

In order to make the iOS part more future proof and easier to follow (most tutorials now are using swift), I converted the AppDelegate files to swift versions. 

## Test Plan
NA

- [ ] ./core/__fixtures__/files/project.pbxproj is still referencing an AppDelegate.m and I do not know, how to update that file

## Release Notes

 [IOS] [CLI] [FEATURE] - Initialize project that make use of AppDelegate.swift instead AppDelegate.h/m


janicduplessis:
@mcplectrum Thanks for the PR! My main concern with this is that a lot of apps won't use any native code so it doesn't make sense to bundle the swift runtime (which adds a few mbs to the binary) just for using it in AppDelegate.

Could we just provide a different swift template instead so people that want to use swift can use that instead? We probably should still default to objc for now until swift hits ABI stability and it's runtime is built into iOS.

If you want to do the same for the RNTester projet you could just create a new xcodeproj file a bit like we have for RNTesterLegacy.

PR #16603, opened 2017-10-31
[IOS] [ENHANCEMENT] [NativeAnimated] Allow shadowOpacity and shadowRadius in NativeAnimated

adbl:
## Motivation

I want to animate `shadowOpacity` style property with Native Animated (`useNativeDriver: true`). This is useful for example in an ios-like navigation transition where the shadow fades in with the view that is sliding in from the side. Code comment for `STYLES_WHITELIST` says:
_In general native animated implementation should support any numeric property that doesn't need to be updated through the shadow view hierarchy (all non-layout properties)._
so I just added `shadowOpacity` (and `shadowRadius` too because why not?).

## Test Plan

Before this change, setting `shadowOpacity` (or `shadowRadius`) to an `AnimatedValue` (with `useNativeDriver: true`) would throw the error: `Style property 'shadowOpacity' is not supported by the native animated module`.

After adding `shadowOpacity` (and `shadowRadius`), there is no error. The animation looks correct so it seems to be working. I also tried setting a ridiculously large `shadowRadius` and could see that working too. 

Please advice on any further testing I should do.

## Release Notes

[IOS] [ENHANCEMENT] [NativeAnimated] - Allow `shadowRadius` and `shadowOpacity` as NativeAnimated style properties.

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -147,6 +147,9 @@ const API = {
 const STYLES_WHITELIST = {
   opacity: true,
   transform: true,
+  /* ios styles */
+  shadowOpacity: true,
+  shadowRadius: true,
   /* legacy android transform properties */
   scaleX: true,
   scaleY: true,


janicduplessis:
Nice that it works without any changes! Do you know if shadowOffset also works? I guess not since it's an object, it might need some tweaks in native animated to support it.

janicduplessis:
@facebook-github-bot shipit

adbl:
Yes it almost seemed too simple... Support for shadowOffset was addressed some time ago I think

adbl:
@janicduplessis circle-ci has failed in test-android with something that I don't understand at all, doesn't seem related to this change. Am I expected to do something about this?

PR #16631, opened 2017-11-01
Add some documentation about cookie handling.

dahjelle:
## Motivation

I'm not sure this is the best place for the note, but I've not be able to find any documentation—on Chromium, on Android's CookieManager class, or elsewhere—that the SameSite cookies are not synced. I expect some further documentation on limitations of syncing cookies would be valuable, but I don't have the expertise to flesh that out at the moment.

## Test Plan

Documentation.

## Release Notes
[DOCS][MINOR][WebView] - Note about cookie syncing in WebView.


dahjelle:
While I expect to have to massage this PR to be approved—I don't think that the failure is related to my change?

PR #16639, opened 2017-11-02
[IOS][ENHANCEMENT] - Adding locale prop to DatePickerIOS

RobertPaul01:
## Motivation

While building a React Native application, I've come across the use case of wanting to set a specific locale for DatePickers irrespective of the users OS region setting. Since this is a feature available to native DatePicker components, I think it would be helpful to expose this in React Native as well.

## Test Plan

Testing can be done by passing a `locale` prop to a DatePickerIOS. Example:

```
<DatePickerIOS
  date={this.state.date}
  mode="date"
  locale="fr_FR"
  onDateChange={date => this.setState({ date: date })}
/>
```

## Release Notes

[IOS][ENHANCEMENT][DatePickerIOS] - Adding a locale prop.


Libraries/Components/DatePicker/DatePickerIOS.ios.js
```diff
@@ -127,6 +132,7 @@ const DatePickerIOS = createReactClass({
           ref={ picker => { this._picker = picker; } }
           style={styles.datePickerIOS}
           date={props.date.getTime()}
+          locale={props.locale ? props.locale : undefined}
```
milindagrawal: Can we add something here which will take user locale by default ?

Note: no need to use ternary here, just use props.locale and add user locale in defaultProps.

RobertPaul01: This is part of React Native and it's what is normally done for these props. If the locale isn't specified, it should use the default device locale.

milindagrawal: Awesome

chirag04:
@RobertPaul01 are you using this to force 24hr format for the date time picker using the `NL` locale? just curious.

RobertPaul01:
No, this is to allow certain date formats to be used such as dd/mm/yyyy or mm/dd/yyyy

janicduplessis:
@RobertPaul01 Thanks for the PR! What happens if you pass an invalid locale string?

RobertPaul01:
If the string is invalid, it will use the default device locale.

janicduplessis:
@RobertPaul01 Looks like your PR changes file permissions (`100644 → 100755`), can you revert this?

RobertPaul01:
@janicduplessis I changed the permissions back. I overwrote the previous commit.

janicduplessis:
@facebook-github-bot shipit

Gongreg:
@RobertPaul01 if you want to add example for locale, you can take it from my pull request. I opened it on August but nobody looked into it.

PR #16658, opened 2017-11-03
Fix launchChrome for chromium on Ubuntu

nfeignon:
## Motivation

On Ubuntu (and maybe other distros), the executable for the chromium browser is 'chromium-browser' instead of 'chromium'.
This commit calls 'chromium-browser' before calling 'chromium' if it exists.

## Test Plan

Start an example app on Android with "react-native init project && react-native run-android". Open DevTools: it opens chromium correctly.

## Release Notes

[CLI] [BUGFIX] [local-cli/server/utils/launchChrome.js] - Fix launchChrome for chromium on Ubuntu


janicduplessis:
Thanks for the fix!

@facebook-github-bot shipit

PR #16665, opened 2017-11-04
[podspec] Define `pod_target_xcconfig` for PrivateDatabase

ide:
We don't strictly need this since all the other subspecs that use this set up the header search path for ReactCommon anyway but do it just for correctness.

Test Plan: install iOS pods

## Release Notes

[podspec] Define `pod_target_xcconfig` for PrivateDatabase


janicduplessis:
👍 @facebook-github-bot shipit

janicduplessis:
👍 

@facebook-github-bot shipit

PR #16675, opened 2017-11-04
fix(docs): fixing wrong module name in the example of "Integration with Existing Apps" page

snipter:
## Motivation

The current example of `Objective-C/Swift` component will not work cos of wrong module name in the example. If change `MyReactNativeApp` to the `RNHighScores` everything starts working.

## Release Notes

[DOCS] [BUGFIX] [IntegrationWithExistingApps.md] - Wrong module name in the example.


janicduplessis:
Thanks for the fix!

@facebook-github-bot shipit

PR #16682, opened 2017-11-05
[Bugfix][SwipeableFlatList] Fix SwipeableFlatList on close

kesha-antonov:
## Motivation

This commit https://github.com/facebook/react-native/commit/d8cc6e3c2b8c2dc25b6b32691779790c15de8d10 introduced ```SwipeableFlatList```

It has bug - doesn't close on swipe other rows.

I fixed it.

## Release Notes

[IOS/ANDROID] [BUGFIX] [SwipeableFlatList] - opened rows did not close on swipe other rows

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
@@ -35,7 +35,9 @@ type SwipableListProps = {
 type Props<ItemT> = SwipableListProps & FlatListProps<ItemT>;
 
 type State = {
-  openRowKey: ?string,
+  extraData: {
+    openRowKey: ?string,
+  }
 };
 
 /**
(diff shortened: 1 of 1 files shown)


eballeste:
What if we don't want the other rows to close automatically?

kesha-antonov:
@eballeste https://github.com/kesha-antonov/react-native/blob/d50cbaae89435be46f91f8c6b62550714fdc4f98/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js#L54

kesha-antonov:
Read source. It's by design

AntonPuko:
@kesha-antonov can you check, I think with your current solution all rows are rerendered on every swipe.

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
```diff
@@ -35,7 +35,9 @@ type SwipableListProps = {
 type Props<ItemT> = SwipableListProps & FlatListProps<ItemT>;
 
 type State = {
-  openRowKey: ?string,
+  extraData: {
```
AntonPuko: why do we need extra object inside state? isn't it simpler keep it as before but put into flat list  `extraData={this.state}` ?

tomasreimers: @kesha-antonov 

Following the example in https://facebook.github.io/react-native/docs/flatlist.html

> By passing extraData={this.state} to FlatList we make sure FlatList itself will re-render when the state.selected changes. Without setting this prop, FlatList would not know it needs to re-render any items because it is also a PureComponent and the prop comparison will not show any changes.

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
```diff
@@ -88,31 +90,37 @@ class SwipeableFlatList<ItemT> extends React.Component<Props<ItemT>, State> {
...
     this.state = {
-      openRowKey: null,
+      extraData: {
+        openRowKey: null,
+      }
     };
 
     this._shouldBounceFirstRowOnMount = this.props.bounceFirstRowOnMount;
+
+    this.onRefFlatList = this.onRefFlatList.bind(this);
```
tomasreimers: Why this change?

kesha-antonov: This code works faster than fat arrow.
In render we should avoid create function every time
@tomasreimers

tomasreimers: @kesha-antonov totally get that -- I meant why .bind(this) in the initializer instead of using class fields (like _onScroll)

vysotsky:
Waiting for this fix to be merged? Any updates on it?

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
hramos: Ping

kesha-antonov:
ping

tomasreimers:
Hi @kesha-antonov , I think this was fixed by https://github.com/facebook/react-native/pull/18001

PR #16697, opened 2017-11-06
Doc description of React Navigation ease of use

johngoren:
The guide describes React Navigation as the most "easy" solution to navigation when it would be more accurate and helpful to instead describe it as the most stable.

## Motivation

(Write your motivation here.)

## Test Plan


hramos:
The docs are no longer stored in this repository. See https://github.com/facebook/react-native-website

PR #16709, opened 2017-11-07
Export YellowBox API

wli:
## Motivation

Allow end users to access the YellowBox API so that warnings can be properly ignored via the API first introduced in a974c140db605ecbdf8d3faa7a079b7e2dcebb09. Based on that API, you should be able to do the following:

```
import { YellowBox } from 'react-native';
YellowBox.ignoreWarnings(['Warning: ...']);
```

However, if you actually try this today, it results in a broken import error.

## Test Plan

Verified using an expo instance. First tried without the YellowBox import, observed a failure to import. Then I added the import statement, and observed no warning on my device.

```
import React from 'react';
import { StyleSheet, Text, View, YellowBox } from 'react-native';


YellowBox.ignoreWarnings(['hey']);


  render() {
    console.warn('hey!');
    return (
        <Text>Open up App.js to start working on your app!</Text>
        <Text>Changes you make will automatically reload.</Text>
        <Text>Shake your phone to open the developer menu.</Text>
      </View>
    );
  }
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```

## Release Notes
[GENERAL] [ENHANCEMENT] [YellowBox] - Export YellowBox API so warnings can be ignored using a non-deprecated API.

### DIFF
--- a/Libraries/react-native/react-native-implementation.js
+++ b/Libraries/react-native/react-native-implementation.js
@@ -103,6 +103,7 @@ const ReactNative = {
   get unstable_batchedUpdates() { return require('ReactNative').unstable_batchedUpdates; },
   get Vibration() { return require('Vibration'); },
   get VibrationIOS() { return require('VibrationIOS'); },
+  get YellowBox() { return require('YellowBox'); },
 
   // Plugins
   get DeviceEventEmitter() { return require('RCTDeviceEventEmitter'); },


ide:
@facebook-github-bot shipit

PR #16757, opened 2017-11-08
Update native-modules-android.md

alexanderkjeldaas:
What arguments can be passed to callbacks is undocumented, both in the code and in this file.

I have no idea what is supported, but I'm guessing this is correct.

## Motivation

Callbacks are not documented well.

## Test Plan

N/A


hramos:

Issue #16764, opened 2017-11-09
Switch component handle on iOS loses drop shadow when `thumbTintColor` is set


shaundon:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

    Environment:
      OS:  macOS Sierra 10.12.6
      Node:  8.9.0
      Yarn:  Not Found
      npm:  5.5.1
      Watchman:  4.7.0
      Xcode:  Xcode 9.1 Build version 9B55
      Android Studio:  2.3 AI-162.4069837

    Packages: (wanted => installed)
      react: 16.0.0-beta.5 => 16.0.0-beta.5
      react-native: ^0.49.3 => 0.49.3

  Target Platform: iOS (11.1)

### Steps to Reproduce

The `Switch` component has a `thumbTintColor` prop that allows the handle colour to be set. I've noticed that if this is set, the handle loses its drop shadow.

Here's some sample code:

```
// The handle has a drop shadow.
<Switch value />

// The handle does not have a drop shadow.
<Switch thumbTintColor="#fff" value />
```

### Expected Behavior

I assumed that the handle's drop shadow would be maintained no matter the colour of the handle.

### Actual Behavior

The handle loses its drop shadow when `thumbTintColor` is set, even when it's set to white.

![img_f525a6d535b5-1](https://user-images.githubusercontent.com/73652/32602364-c8fbac48-c53d-11e7-91c9-29fbc642b3ab.jpeg)

### Reproducible Demo

https://snack.expo.io/B1PwL3-Jz

shaundon:
I've investigated this more and it appears to be part of the underlying iOS implementation. 

I tested this by creating a project in Xcode and adding a switch using storyboards, no JavaScript / React Native involved. I set [thumbTintColor](https://developer.apple.com/documentation/uikit/uiswitch/1623684-thumbtintcolor) on the switch to white, and it loses its drop shadow. 

The [UISwitch documentation](https://developer.apple.com/documentation/uikit/uiswitch) doesn't list any properties available to control the shadow, so I don't think there's anything that can be done in React Native about this, except for possibly updating the RN documentation to make people aware of this behaviour.

joshuawright11:
[...]

PR #16788, opened 2017-11-10
Update Switch.js

shaundon:
## Motivation

When `thumbTintColor` is set on iOS, the switch loses its drop shadow. I propose that we be explicit about this behaviour to remove any confusion.

Here's an issue that explains the motivation further: https://github.com/facebook/react-native/issues/16764.

## Test Plan

Just a change to the copy in docs so I'm omitting this.

## Release Notes

 [DOCS] [ENHANCEMENT] [Switch.js] - Added info about `thumbTintColor` iOS behaviour

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -46,12 +46,12 @@ suppress_type=$FlowFixMeProps
 suppress_type=$FlowFixMeState
 suppress_type=$FixMe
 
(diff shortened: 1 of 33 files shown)


hramos:
Here's the public facing version of the Switch doc: https://github.com/facebook/react-native/blob/master/docs/switch.md

shaundon:
@hramos Thanks! I messed up the rebase (never been good at those..) so I threw out the PR and made a new one: https://github.com/facebook/react-native/pull/16888. Sorry for the confusion!

PR #13877, opened 2017-05-09
Add letter spacing for Android versions >= Lollipop (5.0).

zezhouliu:
Thanks for submitting a PR! Please read these instructions carefully:

- [x] Explain the **motivation** for making this change.
- [x] Provide a **test plan** demonstrating that the code is solid.
- [x] Match the **code formatting** of the rest of the codebase.
- [x] Target the `master` branch, NOT a "stable" branch.

## Motivation (required)
What existing problem does the pull request solve?

This PR makes enables `lineSpacing` on Android.  It is based on https://github.com/facebook/react-native/pull/9420, differing only with respect to minor library import references.

We have merged this into Airbnb's react-native fork: https://github.com/airbnb/react-native/pull/13

### Android fonts on Airbnb app
screen shot 2017-05-09 at 2 23 39 pm

### Android fonts specific use cases:

screen shot 2017-01-23 at 9 25 30 pm

screen shot 2017-01-23 at 9 27 42 pm

## Next Steps

Sign the [CLA][2], if you haven't already.

Small pull requests are much easier to review and more likely to get merged. Make sure the PR does only one thing, otherwise please split it.

Make sure all **tests pass** on both [Travis][3] and [Circle CI][4]. PRs that break tests are unlikely to be merged.

For more info, see the ["Pull Requests"][5] section of our "Contributing" guidelines.

[1]: https://medium.com/@martinkonicek/what-is-a-test-plan-8bfc840ec171#.y9lcuqqi9
[2]: https://code.facebook.com/cla
[3]: https://travis-ci.org/facebook/react-native
[4]: http://circleci.com/gh/facebook/react-native
[5]: https://github.com/facebook/react-native/blob/master/CONTRIBUTING.md#pull-requests

PR #13199, opened 2017-03-29
[Android] Implement letterSpacing option

jerolimov:
## Motivation

Implement letterSpacing option on Android. This was requested multiple times. (See also #3485 #10622) - This PR is inspired by #9420 but also fixes the layout calculation which is required for multiline text with letter spacing.

The problem with letter spacing on Android is, in general, that the API TextView#setLetterSpacing expects the letter spacing value in "EM", where "1 EM" is defined by the width of the letter 'M'. From the documentation: "Typical values for slight expansion will be around 0.05. Negative values will tighten text." The react-native developer expectation was that the spacing is defined in points, similar to width, height, margin, padding, and so on.

This patch implements an "approach" and calculates the "EM" based on the font size.

I tested this with the default font and the layout was now "similar" to iOS, but doesn't match the iOS layout exactly.

## Test Plan

Screenshot on iOS (before and after):

simulator screen shot 29 mar 2017 12 16 17

Screenshot on Android before this patch:

Screenshot on Android after this patch:

Sourcecode for this "beautiful app" is available here: https://gist.github.com/jerolimov/39e4fd29702b6f1ad55768907d92c33d

## Documentation Question

How can I update the documentation and remove the "ios" badge on the `letterSpacing` option?

PR #16801, opened 2017-11-13
[WIP] LetterSpacing support for Android Lollipop and above (21)

shockdesign:
## Motivation

Implement letterSpacing for Android for Android 21 (Lollipop) and above. This is required by a project I'm working on, so what better way then to implement. 

Have realised that this has been attempted twice before:

https://github.com/facebook/react-native/pull/13199 & https://github.com/facebook/react-native/pull/13877

However both of those seem to have just stopped. Happy to remove this and work on getting one of those up to scratch (or get the best bits of both those PRs and have them all here).

## Test Plan

## Release Notes

Still a WIP at the moment

[ANDROID] [ENHANCEMENT] [TextInput] Support LetterSpacing on Android 21 devices and up.

### DIFF
--- a/Libraries/Text/TextStylePropTypes.js
+++ b/Libraries/Text/TextStylePropTypes.js
@@ -49,7 +49,8 @@ const TextStylePropTypes = {
   textShadowRadius: ReactPropTypes.number,
   textShadowColor: ColorPropType,
   /**
-   * @platform ios
+   * Specifies the letter spacing (kerning). Defaults to 0.0 (no kerning) and is supported on iOS and 
+   * only on Android Lollipop (21) and greater.
    */
   letterSpacing: ReactPropTypes.number,
   lineHeight: ReactPropTypes.number,
--- a/RNTester/js/TextInputExample.android.js
+++ b/RNTester/js/TextInputExample.android.js
@@ -636,6 +636,19 @@ exports.examples = [
     }
   },
   {
+    title: 'Letter spacing',
+    platform: 'android',
+    render: function() {
+      return (
+        <View>
+          <TextInput lineSpacing={1.2}
+           placeholder="Line spacing of 1.2"
+          />
+        </View>
+      );
+    }
+  },
+  {
     title: 'Auto-expanding',
     render: function() {
       return (
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
@@ -242,6 +242,8 @@ public abstract class ReactBaseTextShadowNode extends LayoutShadowNode {
   protected int mTextAlign = Gravity.NO_GRAVITY;
   protected int mTextBreakStrategy =
       (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) ? 0 : Layout.BREAK_STRATEGY_HIGH_QUALITY;
+  protected float mLetterSpacing = 0.0f;
+  protected float mPreviousLetterSpacing = Float.NaN;
 
   protected float mTextShadowOffsetDx = 0;
   protected float mTextShadowOffsetDy = 0;
@@ -291,6 +293,14 @@ public abstract class ReactBaseTextShadowNode extends LayoutShadowNode {
     return useInlineViewHeight ? mHeightOfTallestInlineImage : mLineHeight;
   }
 
+  private static float calculateLetterSpacing(float letterSpacing, int fontSize) {
+    if (Float.isNaN(letterSpacing) || letterSpacing == 0.0f) {
+      return 0.0f;
+    }
+
+    return letterSpacing / fontSize * 2.0f;
+  }
+
   // Return text alignment according to LTR or RTL style
   private int getTextAlign() {
     int textAlign = mTextAlign;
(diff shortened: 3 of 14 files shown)


RNTester/js/TextInputExample.android.js
```diff
@@ -635,6 +635,19 @@ exports.examples = [
       );
     }
   },
+  {
+    title: 'Letter spacing',
+    platform: 'android',
+    render: function() {
+      return (
+        <View>
+          <TextInput lineSpacing={1.2}
```
motiz88: You probably mean to use `letterSpacing` here.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -291,6 +293,14 @@ public float getEffectiveLineHeight() {
     return useInlineViewHeight ? mHeightOfTallestInlineImage : mLineHeight;
   }
 
+  private static float calculateLetterSpacing(float letterSpacing, int fontSize) {
+    if (Float.isNaN(letterSpacing) || letterSpacing == 0.0f) {
+      return 0.0f;
+    }
+
+    return letterSpacing / fontSize * 2.0f;
```
motiz88: I believe the [approach](https://github.com/facebook/react-native/pull/13877/files#diff-dde94cc0ccc8d662c9efae752698d1c2R82) taken in #13877 is a more correct one with respect to the unit conversions involved; see also my comment on `setFontSize` below.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -304,6 +314,13 @@ private int getTextAlign() {
     return textAlign;
   }
 
+  @ReactProp(name = ViewProps.LETTER_SPACING, defaultFloat = Float.NaN)
+  public void setLetterSpacing(float letterSpacing) {
+    mPreviousLetterSpacing = letterSpacing;
```
motiz88: Nitpick (though I'm not an authority on this codebase): "previous" implies you're keeping an old state, but in fact you're just converting units for the _same_ state; maybe something like `mLetterSpacingPts` and `mLetterSpacingEms`? (or `mLetterSpacingInput` and `mLetterSpacing`, following the pattern seen elsewhere in this file with e.g. `mFontSize`)

ReactAndroid/src/test/java/com/facebook/react/views/textinput/ReactTextInputPropertyTest.java
```diff
@@ -347,6 +347,20 @@ public void testTextAlign() {
...
   }
 
+  @Test
+  public void testLetterSpacing() {
+    ReactEditText view = mManager.createViewInstance(mThemedContext);
+
+    mManager.updateProperties(view, buildStyles());
+    assertThat(view.getLetterSpacing()).isEqualTo(0.0);
+
+    mManager.updateProperties(view, buildStyles("numberOfLines", 0.5));
```
motiz88: You probably mean `letterSpacing` here and below.

Libraries/Text/TextStylePropTypes.js
```diff
@@ -49,7 +49,8 @@ const TextStylePropTypes = {
   textShadowRadius: ReactPropTypes.number,
   textShadowColor: ColorPropType,
   /**
-   * @platform ios
+   * Specifies the letter spacing (kerning). Defaults to 0.0 (no kerning) and is supported on iOS and
```
motiz88: Nitpick: Kerning seems to be the wrong term per [Wikipedia](https://en.wikipedia.org/wiki/Letter-spacing).

> Letter-spacing should not be confused with kerning. [...] Kerning is a spacing adjustment of one or more specific pairs of adjacent characters [...]

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -363,6 +380,7 @@ public void setFontSize(float fontSize) {
               : (float) Math.ceil(PixelUtil.toPixelFromDIP(fontSize));
     }
     mFontSize = (int) fontSize;
+    mLetterSpacing = calculateLetterSpacing(mPreviousLetterSpacing, mFontSize);
```
motiz88: Still on the subject of unit conversions:

`mFontSize` is in pixels here, but letter spacing is given in _points_ (presumably the same units as `mFontSizeInput` - SP or DIP depending on `mAllowFontScaling`). We need to pass _ems_ to `setLetterSpacing()`; the correct calculation to yield that would seem to be `letterSpacing / mFontSizeInput` (handling `UNSET` gracefully somehow - is there a default font size defined somewhere?), since the font size is by definition the em size.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -419,6 +420,14 @@ public void setTextAlignVertical(ReactEditText view, @Nullable String textAlignV
     }
   }
 
+  @ReactProp(name = ViewProps.LETTER_SPACING, defaultFloat = 0.0f)
+  public void setLetterSpacing(ReactEditText view, float spacing) {
+    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
```
motiz88: Wondering aloud: Is there a precedent for RN enabling an Android-version-dependent feature with neither a fallback nor a warning for older versions?

docs/text.md
```diff
@@ -493,7 +493,7 @@ default is `true`.
 
   - **`fontVariant`**: array of enum('small-caps', 'oldstyle-nums', 'lining-nums', 'tabular-nums', 'proportional-nums') (_iOS_)
 
-  - **`letterSpacing`**: number (_iOS_)
+  - **`letterSpacing`**: number
```
motiz88: It's probably prudent to amend this, perhaps to say something like `(iOS, Android >= 5.0)`.

motiz88:
Also, having had a more in-depth look at #13877, I now think the correct approach would be to apply `letterSpacing` in a Span so nested nodes are supported - see proposed approach in https://github.com/facebook/react-native/pull/13877#issuecomment-354567566

motiz88:
@shockdesign Want to close this to keep the focus on #17398?

hramos:
Closing in favor of newer PR.

Issue #12336, opened 2017-02-10
react-native-git-upgrade fails on git diff --binary HEAD~1 HEAD


macnube:
### Description

Attempting to upgrade react native from 0.38.0 to 0.39.2 using react-native-git-upgrade 0.39.2
### Reproduction

```
react-native-git-upgrade 0.39.2
git-upgrade info Check for updates 
git-upgrade info Read package.json files 
git-upgrade info Check declared version 
git-upgrade info Check matching versions 
git-upgrade info Check React peer dependency 
git-upgrade info Check that Git is installed 
git-upgrade info Get information from NPM registry 
git-upgrade info Upgrading to React Native 0.39.2, React ~15.4.0-rc.4 
git-upgrade info Setup temporary working directory 
git-upgrade info Configure Git environment 
git-upgrade info Init Git repository 
git-upgrade info Add all files to commit 
git-upgrade info Commit current project sources 
git-upgrade info Create a tag before updating sources 
git-upgrade info Generate old version template 
git-upgrade info Add updated files to commit 
git-upgrade info Commit old version template 
git-upgrade info Install the new version 
npm WARN deprecated node-uuid@1.4.7: use uuid module instead
git-upgrade info Generate new version template 
git-upgrade info Add updated files to commit 
git-upgrade info Commit new version template 
git-upgrade info Generate the patch between the 2 versions 
git-upgrade ERR! An error occurred during upgrade: 
git-upgrade ERR! Error: Command 'git diff --binary HEAD~1 HEAD' exited with code null:
stderr: undefined
```

### Solution

No idea!

### Additional Information

* React Native version: 0.38.0
* Platform: N/A
* Operating System: MacOS 10.12.3

lacker:
Does this happen on a fresh 0.38.0 install?

ncuillery:
From https://nodejs.org/api/child_process.html#child_process_event_exit:

>If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null.

It means that the `diff` command's process hasn't exited on its own, it has been terminated by the a signal.

@macnube Do you see a reason to have this process killed by a signal on your machine (Ctrl-C, watchdog or whatever)? Does it happen every time ?

It could be a good idea to improve the debug information in the [exec function](https://github.com/facebook/react-native/blob/master/react-native-git-upgrade/cliEntry.js#L58):
[...]

PR #16822, opened 2017-11-14
react-native-git-upgrade signal termination

mateusz-:
## Motivation

When performing an upgrade using react-native-git-upgrade git can fail with a signal, however, only exit codes are accounted for.

Related issue: #12336

## Test Plan

In my case, Git fails with a `SIGPIPE` signal - have not figured out why yet, but that's a separate issue. Before, since exit code was not zero, the console would default to the error case, but since no exit code was supplied, the error message was meaningless - `exited with code null`. Now, it errors out with `terminated with signal 'SIGPIPE'`, while still taking account of exit codes. Quoting [nodejs docs](https://nodejs.org/api/child_process.html#child_process_event_exit):

> The 'exit' event is emitted after the child process ends. If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null.

I would be happy to write a test case for this but I haven't seen any react-native-git-upgrade ones?

## Release Notes
[CLI] [ENHANCEMENT] [react-native/react-native-git-upgrade] - Git terminated by signal error message

### DIFF
--- a/react-native-git-upgrade/cliEntry.js
+++ b/react-native-git-upgrade/cliEntry.js
@@ -52,12 +52,18 @@ function exec(command, logOutput) {
       process.stderr.write(data);
     });
 
-    child.on('exit', code => {
-      (code === 0)
-        ? resolve(stdout)
-        : reject(new Error(`Command '${command}' exited with code ${code}:
+    child.on('exit', (code, signal) => {
+      if (code === 0) {
+        resolve(stdout);
+      } else if (code) {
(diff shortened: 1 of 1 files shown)


mateusz-:
As a side note, I rebooted my PC  (MacBook Pro with Sierra) and I am no longer getting the SIGPIPE. It happened when trying to diff changes of .flowconfig, coming from 0.47.1.

PR #16847, opened 2017-11-15
Add contextMenuHidden to TextInput

amhinson:
## Motivation

There is currently no way to disable to tooltip that automatically appears over a TextInput. This is especially troublesome if you would like to disable the user from pasting text into certain fields. This PR adds a `tooltipHidden` property to TextInput that will hide it.

## Test Plan

I'm not sure if testing is necessary here. I would be happy to investigate further on how this would be tested, if deemed necessary!

## Release Notes

[FEATURE][TextInput] - Added `tooltipHidden` property


shergin:
Well, this is not totatly correct.

shergin:
My conserns:

 * This thing is called `context menu`, not `tooltip`;
 * I don't think that turning off all context menu actions is a good idea, we have to have more granular control of those features;
 * We have to avoid "Negative logic" in naming things;
 * We also have to implement it for multiline <TextInput>;
 * And for Android as well.

amhinson:
@shergin Thanks for looking it over. Do you have any thoughts on how we can add more granular control? My idea would be to add an array of strings with multiple options, such as `'copy'`, `'paste'`, etc and also an `'all'` option that will remove it completely.

amhinson:
I've updated to with the following adjustments:

- Changed name to `contextMenu` instead of `tooltip`
- Multiline & Android support

I'm not quite sure what would be the best way to add granular control of this feature (i.e. only disable "Paste", disable both "Select All" & "Paste" but still allow "Copy"). It seems the implementation for this cross-platform would be tricky.

Noitidart:
Does iOS also offer a hidden tooltip feature?

amhinson:
@Noitidart Yes, this PR is for both iOS and Android.

chirag04:
throwing an idea out here: what if we allow passing custom contextual menus here. those who want to hide them could then simply pass `null`. Might be broader in scope but could help with multiple use-cases. cc @shergin

amhinson:
@chirag04 I like the idea. One of the issues I saw though was that iOS and Android have slightly different ways they behave for this menu, so mapping them 1 for 1 might be tough. Also, the Android implementation wouldn't be quite as straight-forward as iOS, unless I'm missing something. I'm definitely not an Android expert by any means.

amhinson:
Is there any chance we could get this reviewed?

amhinson:
@facebook-github-bot Yes, this PR is still relevant. Just rebased.

amhinson:
Closing this to create a new clean PR.

AKKIDETROJA:
```
<TextInput
   **contextMenuHidden={true}**
  value={this.state.text}
onChangeText={(text) => this.setState({ text})}
/>
```

PR #16857, opened 2017-11-16
Fixes crash when setting selectionColor on old android devices

jeremistadler:
Setting selectionColor for TextInput crashes on sdk version 19

This just bails out instead of crashing. 

## Test plan

Start an emulator with **android sdk 19**

```jsx
import { TextInput } from 'react-native';
  render() {
    return (
      <TextInput
        value="hello"
        selectionColor="#aabbcc"
        style={{margin: 40}}
      />
    );
  }
}
```

## Release Notes

[ANDROID] [BUGFIX] [TextInput] - setting selectionColor crashed on sdk 19


jeremistadler:
@facebook-open-source-bot run tests

shergin:
Actually, it was fixed in https://github.com/facebook/react-native/pull/14789.

PR #16859, opened 2017-11-16
[CI] Pin Buck to commit which supports source only abi

hramos:
Android tests on CI have been failing since late October due to dd016f334cef26f393b1a3c1e052e89a69502a56 which makes use of a new Buck feature, `required_for_source_only_abi`. Circle was using an older September release. In this PR, we pin the Buck version to match the commit where the feature was enabled.

We're not plainly installing from master in order to take advantage of caching.

## Motivation

Green is good. Red is bad.

## Test Plan

Wait for Circle to run.

## Release Notes

[INTERNAL][BUGFIX][./circleci/config.yml] - Update Buck version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,11 +69,11 @@ aliases:
 
   - &restore-cache-buck
     keys:
-      - v1-buck-{{ arch }}-v2017.09.04.02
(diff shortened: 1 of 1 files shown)


hramos:
Buck is releasing a new version today with this change, so let's wait for that instead.

grabbou:
I did it previously! Right now, I just reverted the commit causing issues - it's green again. I'll keep an eye on the BUCK update.

PR #16882, opened 2017-11-20
Fix $FlowFixMe in Flatlist

cdlewis:
## Motivation

$FlowFixMe suppressed type validation on _listRef. Add appropriate typing to let Flow succeed when $FlowFixMe is removed.

Not sure what this project's philosophy around ref types is. In some places I've seen them annotated as any rather than trying to get a more granular type to match. I'm open to suggestions.

## Test Plan

`yarn flow`

## Release Notes

[INTERNAL] [ENHANCEMENT] [FlatList.js] - Remove $FlowFixMe from FlatList


cdlewis:
@janicduplessis thanks for reviewing my earlier PR. I had a pretty weird merge after incorporating your old feedback so I declared git bankruptcy and re-opened here :)

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
Alright hopefully it lands now, the bot finally listened :)

Libraries/Lists/FlatList.js
```diff
@@ -16,6 +16,7 @@ const MetroListView = require('MetroListView'); // Used as a fallback legacy opt
 const React = require('React');
 const View = require('View');
 const VirtualizedList = require('VirtualizedList');
+const ListView = require('ListView');
```
sahrens: Wut?

Libraries/Lists/FlatList.js
```diff
@@ -457,13 +470,10 @@ class FlatList<ItemT> extends React.PureComponent<Props<ItemT>, void> {
   }
 
   _hasWarnedLegacy = false;
-  _listRef: VirtualizedList;
+  _listRef: null | VirtualizedList | ListView;
```
sahrens: Why would the listRef ever be a ListView?

janicduplessis: Right, it should be MetroListView instead (https://github.com/cdlewis/react-native/blob/22e2686a7ec2bc81656d8309676ddbd74057f0ba/Libraries/Lists/FlatList.js#L624). My bad for not noticing this, probably still passed flow because they have the same props.

cdlewis: Hmmm, when I change it to MetroListView I see a bunch of errors related to that component not having properties such as `getScrollResponder`. Any ideas?

janicduplessis: Hmm if it's just `getScrollResponder` maybe you could just add it to MetroListView and just call the method on the underlying ListView ref.

PR #16888, opened 2017-11-20
Update Switch docs to add more context about iOS behaviour when thumb…

shaundon:
**Note:** I made this PR already but messed up rebasing, so I've thrown it out and made this one. Sorry for the confusion!

## Motivation

When `thumbTintColor` is set on iOS, the switch loses its drop shadow. I propose that we be explicit about this behaviour to remove any confusion.

Here's an issue that explains the motivation further: https://github.com/facebook/react-native/issues/16764.

## Test Plan

Just a change to the copy in docs so I'm omitting this.

## Release Notes

 [DOCS] [ENHANCEMENT] [Switch.js] - Added info about `thumbTintColor` iOS behaviour

### DIFF
--- a/Libraries/Components/Switch/Switch.js
+++ b/Libraries/Components/Switch/Switch.js
@@ -70,7 +70,8 @@ var Switch = createReactClass({
      */
     onTintColor: ColorPropType,
     /**
-     * Color of the foreground switch grip.
+     * Color of the foreground switch grip. If this is set on iOS,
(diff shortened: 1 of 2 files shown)


hramos:
Heya! Do you mind making the change over at https://github.com/facebook/react-native-website? Sorry for the delay getting back to you.

shaundon:
No problem - here you go: https://github.com/facebook/react-native-website/pull/24

Issue #16635, opened 2017-11-02
Unexpected keyboard dismiss in ScrollView when soft keyboard appears.


jiaminglu:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Scanning folders for symlinks in /Users/jiaminglu/test/node_modules (20ms)

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.2.1
  Yarn:  0.27.5
  npm:  5.3.0
  Watchman:  4.9.0
  Xcode:  Xcode 9.1 Build version 9B55
  Android Studio:  3.0 AI-171.4408382

Packages:
  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: 0.49.5 => 0.49.5

### Steps to Reproduce

1. Use a TextInput inside a ScrollView, with keyboardDismissMode set to 'on-drag'
2. Layout the TextInput at the bottom of the screen.
3. Click the text input

### Expected Behavior

The keyboard should appear.

### Actual Behavior

The keyboard appears and then disappears immediately.

### Reproducible Demo

1. react-native init test
2. replace App.js with:
```jsx
import React, { Component } from 'react';
import { View, TextInput, ScrollView, Dimensions } from 'react-native';

export default class App extends Component {
  render() {
    return (
      <ScrollView style={{flex: 1}} keyboardDismissMode='on-drag'>
        <View  style={{flex: 1}}>
          <TextInput defaultValue="Hello world" style={{
            marginTop: Dimensions.get('window').height - 50,
            height: 40,
            borderColor: 'black',
            borderWidth: 1,
          }}/>
        </View>
      </ScrollView>
    );
  }
}
```

### Cause

When the TextInput clicked and soft keyboard appeared, the Android system pushes up the underlying EditText, and then the ScrollView detects a scroll event, dismissed the keyboard according to `keyboardDismissMode="on-drag"`, which I think should be handled on `onScrollBeginDrag` instead.

RonnieRen:
https://github.com/facebook/react-native/pull/16889

darylalexjohnson:
I'm curious about the fix for this issue. ScrollViews on Android with the  keyboardDismissMode set to on-drag, no longer dismiss the keyboard. For example 
```import React, { Component } from 'react';
import { View, Text, FlatList, StyleSheet, TextInput } from 'react-native';

export default class App extends Component {
[...]

PR #16889, opened 2017-11-20
when keyboardDismissMode ==='on-drag', we also make sure we are reall…

RonnieRen:
…y on-drag state and then dismiss the keyboard.

## Motivation

fix issue #16635
the issue is a little serious if you use a scrollview which contains textinput, you will never get a chance to input something since keyboard will always be dismissed because of on-drag event if your scroll-view use the property: keyboardDismissMode ='on-drag'

## Test Plan

verify the issue https://github.com/facebook/react-native/issues/16635 and tested this fix worked


RonnieRen:
https://github.com/facebook/react-native/issues/16635

shergin:
@RonnieRen Could you please sign the CLA?

angly-cat:
I'd like to see this PR merged.

> Is this pull request still relevant? If yes could you please rebase?

@RonnieRen could you rebase it, please?

mistenkt:
Any change of a merge? Having a problem on android that would be fixed with this PR

canastro:
Having this fix merged would be really important for the app I'm currently developing. Any update?

cpojer:
Sorry for the long wait. It seems like @mistenkt already made the same change and as of today it is present in `ScrollView.js`: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollView.js#L835 changed in https://github.com/facebook/react-native/commit/2f1421dec7cd3a35779caceac108e872033c7d72.

PR #16891, opened 2017-11-20
Geolocation: add authorizationStatusChange event

koenpunt:
## Motivation

To be able to listen for authorization changes. For example when a user previously denied authorization to their location, you can direct them to the system settings, and by using this event you can reflect the updated setting in the application.

Mostly putting this out here to see if there's any desire by others too for this functionality.

## Test Plan

Currently not sure, but if someone has an idea on how this should be tested I'm happy to hear.

## Release Notes

[IOS] [FEATURE] [Geolocation] - add `authorizationStatusChange` event

### DIFF
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -92,7 +92,28 @@ type LocationEventName = $Enum<{
  */
 var Geolocation = {
 
-  /*
+  STATUSES: {
+    // User has not yet made a choice with regards to this application
+    NOT_DETERMINED: 'not_determined',
+    // This application is not authorized to use location services.  Due
+    // to active restrictions on location services, the user cannot change
+    // this status, and may not have personally denied authorization
+    RESTRICTED: 'restricted',
+    // User has explicitly denied authorization for this application, or
+    // location services are disabled in Settings.
+    DENIED: 'denied',
+    // User has granted authorization to use their location at any time,
+    // including monitoring for regions, visits, or significant location
+    // changes.
+    AUTHORIZED_ALWAYS: 'authorized_always',
+    // User has granted authorization to use their location only when your app
+    // is visible to them (it will be made visible to them if you continue to
+    // receive location updates while in the background).  Authorization to use
+    // launch APIs has not been granted.
+    AUTHORIZED_WHEN_IN_USE: 'authorized_when_in_use',
+  },
+
+   /*
     * Sets configuration options that will be used in all location requests.
     *
     * ### Options
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -41,6 +41,10 @@ type GeoOptions = {
   useSignificantChanges?: bool,
 }
 
+type LocationEventName = $Enum<{
+  authorizationStatusChange: string,
+}>;
+
 /**
  * The Geolocation API extends the web spec:
  * https://developer.mozilla.org/en-US/docs/Web/API/Geolocation
@@ -215,6 +219,28 @@ var Geolocation = {
         }
       }
       subscriptions = [];
+    }
+  },
+
+  addEventListener: function(eventName: LocationEventName, callback: Function) {
+    if (eventName === 'authorizationStatusChange') {
+      return LocationEventEmitter.addListener(
+        'locationAuthorizationStatusDidChange',
+        callback
+      );
+    } else {
+      warning(false, 'Trying to subscribe to unknown event: ' + eventName);
+    }
+  },
+
+  removeEventListener: function(eventName: LocationEventName, callback?: Function) {
+    if (eventName === 'authorizationStatusChange') {
+      return LocationEventEmitter.removeListener(
+        'locationAuthorizationStatusDidChange',
+        callback
+      );
+    } else {
+      warning(false, 'Trying to unsubscribe from unknown event: ' + eventName);
     }
   }
 };
--- a/Libraries/Geolocation/Geolocation.js
+++ b/Libraries/Geolocation/Geolocation.js
@@ -122,6 +122,17 @@ var Geolocation = {
   },
 
   /*
+   * Invokes the callback once with the current authorization status.
+   */
+  getAuthorizationStatus: function(callback: Function) {
+    invariant(
+      typeof callback === 'function',
+      'Must provide a valid callback.'
+    );
+    RCTLocationObserver.getAuthorizationStatus(callback);
+  },
+
+  /*
    * Invokes the success callback once with the latest location info.  Supported
    * options: timeout (ms), maximumAge (ms), enableHighAccuracy (bool)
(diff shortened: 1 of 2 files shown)


Libraries/Geolocation/Geolocation.js
```diff
@@ -216,6 +220,28 @@ var Geolocation = {
       }
       subscriptions = [];
     }
+  },
+
+  addEventListener: function(eventName: LocationEventName, callback: Function) {
```
shergin: @grabbou Hey! Do you have any concerns related to event-subscribing API here?

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -148,9 +161,34 @@ - (dispatch_queue_t)methodQueue
...
+    _locationManager.delegate = self;
+  }
 }
 
+- (void)stopObserving
+{
+  _hasListeners = NO;
+}
+
+- (NSDictionary *)constantsToExport
```
shergin: Why do we need this? Implementing `constantsToExport` prevents this module to be lazy-loaded which affects performance.

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -68,6 +68,18 @@ + (RCTLocationOptions)RCTLocationOptions:(id)json
 
 @end
 
+@implementation RCTConvert (CoreLocation)
+
+RCT_ENUM_CONVERTER(CLAuthorizationStatus, (@{
```
shergin: Should we unify these names with Android counterpart... somehow?

koenpunt: I actually tried that in the app where I extracted this methods from, and that didn't went well.. That might of course have just be me. 

But for one the status of the authorization is binary on Android, or at least with the currently available API:

```js
PermissionsAndroid.check(permission).then((granted) => {
  if (granted) {
    // ...
  }
});
```

Where on iOS there a lot more values.

Also now that I think of it, I should probably also add `getAuthorizationStatus` to this PR.

Libraries/Geolocation/RCTLocationObserver.m
koenpunt: Well to be able to compare them to the value of the `authorizationStatusChange`. So something like this:

```js
Geolocation.addEventListener('authorizationStatusChange', (status) => {
  if (Geolocation.authorizedAlways === status) {
    // ...
  }
  if (Geolocation.denied === status) {
    // ...
  }
});
```

Or am I missing something and can this be achieved in another way?

Libraries/Geolocation/Geolocation.js
koenpunt: I indeed saw a comment of @grabbou on a different PR where he did put some remarks on the event-subscribing, but I figured that since on other places the `addEventListener` conversion is still being used I should follow that.

To make it compatible with other implementation throughout the source code I should probably also `return { remove: () => null }` for the warning case.

Libraries/Geolocation/RCTLocationObserver.m
shergin: Yeah, I see. So, theoretically, we can use iOS values on Android (something like `always` and `denied`), right?

Libraries/Geolocation/Geolocation.js
shergin: Honestly, I am okay with any approach, I just want to have single one pattern across whole codebase. I also don't want to have it overengineered.

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -148,9 +161,34 @@ - (dispatch_queue_t)methodQueue
...
+    _locationManager.delegate = self;
+  }
+}
+
+- (void)stopObserving
+{
+  _hasListeners = NO;
 }
 
+- (NSDictionary *)constantsToExport
```
shergin: Why do we need this?

Libraries/Geolocation/RCTLocationObserver.m
```diff
@@ -148,9 +161,34 @@ - (dispatch_queue_t)methodQueue
...
-  return @[@"geolocationDidChange", @"geolocationError"];
+  return @[@"geolocationDidChange",
+           @"geolocationError",
+           @"locationAuthorizationStatusDidChange"];
+}
+
+- (void)startObserving
+{
+  _hasListeners = YES;
+  if (!_locationManager) {
```
shergin: Do we need main thread here?

Libraries/Geolocation/RCTLocationObserver.m
koenpunt: My previous comment was hidden under outdated changes, so here again:

--- 

Well to be able to compare them to the value of the `authorizationStatusChange`. So something like this:

```js
Geolocation.addEventListener('authorizationStatusChange', (status) => {
  if (Geolocation.authorizedAlways === status) {
    // ...
  }
  if (Geolocation.denied === status) {
    // ...
  }
});
```

Or am I missing something and can this be achieved in another way?

Libraries/Geolocation/RCTLocationObserver.m
koenpunt: Yes and no, because the API for android will return `false` when the user was never prompted a request in the first place, whereas iOS then returns `notDetermined`.

koenpunt: However, on requesting permissions more details are revealed; 

```js
const permission = PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION;
PermissionsAndroid.request(permission).then((result) => {
  if (result === PermissionsAndroid.RESULTS.NEVER_ASK_AGAIN) {
    //
  }
  if (result === PermissionsAndroid.RESULTS.DENIED) {
    //
  }
  if (result === PermissionsAndroid.RESULTS.GRANTED) {
    //
  }
});

koenpunt: Maybe we can add something [like this](https://github.com/facebook/react-native/blob/75123c614bb5a61c383cdc247230b3a76c76e14d/ReactAndroid/src/main/java/com/facebook/react/modules/permissions/PermissionsModule.java#L117-L124) to the `check` method, although this would be a breaking change on the API..

koenpunt: Or maybe we can implement the `getAuthorizationStatus` also for Android, and leave the `PermissionsAndroid.check` as-is?

Libraries/Geolocation/RCTLocationObserver.m
shergin: Just move those constants to JS file.

koenpunt: Done

cpojer:
Seems like through all this time there have been only few people who were asking for this functionality so I'm gonna close this PR. I'm happy to reopen this PR if people decide we should include this in RN.

PR #16893, opened 2017-11-20
Fix Bug: Set a React Controller whose view is a RCTRootView as a part…

fishmwei:
Problem representation：   
 
rn-tabbar-bug

## Motivation

To fix a bug I met: Set a React Controller whose view is a RCTRootView as a part of UITabbarController, firstly push to a normal controller, then pop back to the React Controller again. Randomly, the height of RCTContentView is 49 pixels smaller than the height of RCTRootView.

## Test Plan

Currently not sure how to test, but if someone has an idea on how this should be tested I'm happy to hear.

## Release Notes
 [IOS] [BUGFIX]   - Fix the error of the height of the content view when pop back to UITabbarController.


fishmwei:
@shergin

default

As the log shows, sometime setBounds(7) does not trigger layoutSubview, when setBounds interval too fast . At last,  the height of contentView is 569.  49 pixels smaller than the height of RCTRootView (which backgroundColor is white). In my opinion, it's unnecessary to set available size every time, especially when the size does not equal to its superview. 

Note:　it happened when iOS version < 11 , in iOS 11 , it won't call setFrame twice, may be a bug of iOS.

React/Base/RCTRootContentView.m
```diff
@@ -86,6 +86,14 @@ - (void)updateAvailableSize
     return;
   }
 
+  //unnecessary to update avaliableSize, it's triggered by setBounds: from RCTUIManager
+  if (self.superview) {
+    CGRect superviewFrame = self.superview.bounds;
+    if (!CGRectEqualToRect(superviewFrame, self.frame)) {
```
shergin: Okay, but what is an idea/reasoning behind this changeset?
Comparing `view.frame` with `superview.bounds` makes no any sense, I am affraid.

fishmwei: I always think that the size of content view should be equal to its super view's size. That's why you set content's frame in the function layoutSubviews of rootView ? 
```
- (void)layoutSubviews
{
  [super layoutSubviews];
  _contentView.frame = self.bounds;  // here
  _loadingView.center = (CGPoint){
    CGRectGetMidX(self.bounds),
    CGRectGetMidY(self.bounds)
  };
}
```

if `view.frame`  != `superview.bounds`, it's invalidate

fishmwei: @shergin what's your  opinion?

hramos:
Have you signed the CLA yet?

hramos:
Closing, CLA remains unsigned.

Issue #14101, opened 2017-05-22
AsyncStorage not fulfilling promise on Android 7+


jsdario:
### Issue

I am using `redux-persist` in a react native project, that runs just fine in a broad number of devices execept Android 7. I am trying to debug the problem on why my AsyncStorage is nor persisting and found this:

The following code executes inside React Native component lifecycle's
```
  componentDidMount() {
    attachObservables(store)
    setInterval(async () => {
      console.log('Inside setInterval')
      const data = await AsyncStorage.getAllKeys()
      console.log('inside the getAllKeys')
      data.forEach(async k => {
        const value = await AsyncStorage.getItem(k)
        console.group(k)
        console.log(value)
        console.groupEnd()
      })
    }, 3000)
  }
```

Code after 'Inside setInterval' is never called. It only runs once if outside the setInterval. If I call once the code outside the setInterval it appears to run just fine. I also tried callback format vs async / await version but it does not seem to matter.

Same problem I had using firebase js library (callbacks never return after the first one). I am now looking for alternatives to workaround the problem.

Works with `setImmediate` and `InteractionManager.runAfterInteractions`. Seems that `redux-persist` uses `setImmediate` or fallsback to `setTimeout`.

Any ideas?

### Additional Information

* React Native version: React Native 0.40+ (last tested on RN@0.44)
* Platform: Only Android 7 and above
* Development Operating System: Mac os, Node 7.4
* Dev tools: 

```
android {
    compileSdkVersion 24
    buildToolsVersion "24.0.2"

    defaultConfig {
        multiDexEnabled true
        applicationId "com.netbeast.yeti"
        minSdkVersion 16
        targetSdkVersion 24
        versionCode 69
        versionName "1.3.91"
        ndk {
            abiFilters "armeabi-v7a", "x86"
        }
    }
```

In case this is a user level question I opened a question on stackoverflow, no luck yet, https://stackoverflow.com/questions/44114061/asyncstorage-is-not-returning-the-callback?noredirect=1#comment75248695_44114061

dinzo:
@jsdario any updates on this?  were you able to work around or resolve the problem?

jsdario:
I was not able to fix it, but I cooked a heavy and chunky workaround. I simply reimplemented the AsyncStarage API around https://github.com/itinance/react-native-fs so we could use it as storage in `redux-persist`.

Here it is the package https://github.com/netbeast/redux-persist-react-native-fs @dinzo 

[...]

PR #16905, opened 2017-11-21
Fix AsyncStorage not resolving its promises

jsdario:
## Motivation

Projects with other background activities may seen its AsyncStorage problems delayed or never resolved. Here it is an old issue explaining the bug conditions for devices with Android 7+ 

https://github.com/facebook/react-native/issues/14101#issuecomment-345972681

## Test Plan
We can reproduce the error following the steps in issues, which I copy paste here:

The following code executes inside React Native component lifecycle's
```
  componentDidMount() {
    setInterval(async () => {
      console.log('Inside setInterval')
      const data = await AsyncStorage.getAllKeys()
      console.log('inside the getAllKeys')
      data.forEach(async k => {
        const value = await AsyncStorage.getItem(k)
        console.group(k)
        console.log(value)
        console.groupEnd()
      })
    }, 3000)
  }
```

Code after 'Inside setInterval' is never called. It only runs once if outside the setInterval. If I call once the code outside the setInterval it appears to run just fine. I also tried callback format vs async / await version but it does not seem to matter.

## Release Notes

[ANDROID] [BUGFIX] [ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java] - Fixed AsyncStorage when not resolving its promises for Android 7+


jsdario:
The same thing happened with other PRs of mine, that have been accepted as well. If I can do something to fix / improve this, let me know!

mbret:
Hey @jsdario is there any chance for you to update your PR ? Do you need help or anything ?

jsdario:
I believe it's ready to merge, I am using this fork on production, no surprises so far > 5000 devices

esprehn:
How does this compile? Did you mean executeOnExecutor? This seems to pass AsyncTask.THREAD_POOL_EXECUTOR into a bunch of stuff that doesn't actually accept an executor (ex SQLiteStatement or AsyncTask#execute).

RickDT:
@esprehn that is why the CircleCI tests are failing. It should be `executeOnExecutor`. I'm also concerned about arbitrarily changing AsyncStorage from being serial, which guarantees order, to parallel, which would have no such guarantee. For instance, if I use AsyncStorage to save a new value for a key 3 times in a row and then immediately read it back out, there's no guarantee I will get the third value. That would probably cause a lot of knock-on effects for developers.

However, AsyncStorage does have potential deadlocks with other uses of AsyncTasks in the app. In my app, we were able to change to parallel tasks to avoid the deadlock.

Ideally, AsyncStorage's use of AsyncTasks would place the work on it's own serial Executor instead of the default one. I'm not familiar enough with Java to competently implement this or I'd jump in :/ (I just wrote my first copy-pasta Java code about a month ago, when I ran across this issue)

esprehn:
@RickDT I did that by copying the SerialExecutor [1] into a forked version of AsyncStorageModule (and friends).

It's alarming that something is blocking the shared SERIAL_EXECUTOR though. It doesn't seem like it's been identified in the bug, does anyone know what it is?

[1] https://android.googlesource.com/platform/frameworks/base.git/+/1488a3a19d4681a41fb45570c15e14d99db1cb66/core/java/android/os/AsyncTask.java#237

RickDT:
I believe this happens due to other Java code in the app, outside of React Native, behaving badly and blocking the shared SERIAL_EXECUTOR.

The OP claims "Projects with **other background activities**...", which is certainly where we saw it in our codebase.

The test case in the OP may be failing due to a bug in the test case. I created a new RN app and pasted the test code and indeed did not get the inner log. However, that's because nothing had been saved in AsyncStorage. The updated test case does work as expected with AsyncStorage:

```
async componentDidMount() {
  const intervalMillis = 3000;

  await AsyncStorage.setItem("test1", "this is a good test");  //save something first

  setInterval(async () => {
    console.log("Inside setInterval");
    const data = await AsyncStorage.getAllKeys();
    console.log("inside the getAllKeys");
    data.forEach(async k => {
      const value = await AsyncStorage.getItem(k);
      console.group(k);
      console.log(value);
      console.groupEnd();
    });
  }, intervalMillis);
}
```

This will run successfully over many iterations, and if you set intervalMillis to 0 in order to really test the theory, it also runs continuously without blocking.

I think that putting AsyncStorage on a non-shared serial executor will solve the problem for most cases. Other cases may be another bug altogether. If not, we need a true failing test case. If needed, I can whip up a test project that demonstrates blocking from outside of RN in a misbehaving AsyncTask.

julestruong:
i believe it is fixed in v51

sunnylqm:
What is the status of this PR now? @shergin

dannycochran:
With your code snippet, @RickDT, and remote debugging running, if I refresh the default react-native app example app fast enough and often enough, eventually the first call to AsyncStorage will hang indefinitely and never resolve. If I turn off remote debugging, the call will never hang.

Are others experiencing this problem with remote debugging turned off?

shergin:
@jsdario Do you have any idea why this error happens?
```
stderr: /fbandroid/java/com/facebook/catalyst/js/react-native-github/ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java:147: error: no suitable method found for execute(java.util.concurrent.Executor)
    }.execute(AsyncTask.THREAD_POOL_EXECUTOR);
     ^
    method android.os.AsyncTask.execute(java.lang.Void...) is not applicable
      (varargs mismatch; java.util.concurrent.Executor cannot be converted to java.lang.Void)
    method android.os.AsyncTask.execute(java.lang.Runnable) is not applicable
      (argument mismatch; java.util.concurrent.Executor cannot be converted to java.lang.Runnable)
```

chirag04:
Is it intentional to change from a serial to parallel executor?

https://developer.android.com/reference/android/os/AsyncTask.html

```
Warning: Allowing multiple tasks to run in parallel from a thread pool is generally not what one wants, because the order of their operation is not defined. For example, if these tasks are used to modify any state in common (such as writing a file due to a button click), there are no guarantees on the order of the modifications. Without careful work it is possible in rare cases for the newer version of the data to be over-written by an older one, leading to obscure data loss and stability issues. Such changes are best executed in serial; to guarantee such work is serialized regardless of platform version you can use this function with SERIAL_EXECUTOR.
```

cc @shergin

jsdario:
@shergin, I am afraid not. Not have that kind of knowledge of java, might be trivial but I won't know.

@chirag04 I understand that can be a concern, however that can be produced also by a misuse of the library. In the end, it is not guaranteed that the js code that we are running access the store in a serial order anyway, so it ends up being a responsibility for the developer to correctly access the store.

It fixes #14101  for a number of developers if you guys disagree or we can find another way to fix that I'll be more than happy to try something else.

lucky110209:
@jsdario I saw your post on upwork which help me to find out this link. Your issue already fixed in newest version. So do you want to fix with your current version.

NomadGraphix:
@lucky110209 what do you mean? This issue still exists in RN 53

lucky110209:
@NomadGraphix No :). I mean it already fixed in RN 53. I've tested and it works well.

NomadGraphix:
@lucky110209 It actually still exists in RN 53 on android. I've tested and it. It has something to do with the app being monitored by LogCat within the Android studio. Very strange. Once I turn off Android Studio, and refresh the app in Android Emulator it starts to resolve AsyncStorage requests.

leohuangyi:
@NomadGraphix 
It still exist on RN 0.54.0, the AsyncStore.getItem(key)'s return Promise Never resovle
Once I turn off Android Studio's Logcat view , and refresh the app in Android Emulator it starts to resolve AsyncStorage requests

HaswinVidanage:
Any update on this?

mvanroon:
In my case, I'm on a physical device (Samsung S5 Neo, Android 6.0.1) running react-native 0.54.1.
Android Studio is not running.
When I freshly compile from the command line, the promises are resolving, but after refreshing the app, they no longer are.

NomadGraphix:
So, I've discovered as well that I have to restart the app to get it to work. Every time it decides to stop working after a refresh I go ahead and quit/start the app and it begins again. Very strange. I wonder if there's some kind of restriction in Android blocking storage?

mvanroon:
Could it be related to #18209?

mvanroon:
**Update**: I've just upgraded to RN 0.54.2 (simple npm/yarn update, no react-native upgrade) and no longer have this issue

AlicanC:
We too have upgraded to 0.54.2 and the issue persists.

Jono20201:
I'm also getting this on 0.54.2 - can this please be pulled in?

C0dekid:
Is there any fix yet? I have spent the whole week looking for the bugs in my code, and now I find out it's an issue with Android 7?

Metroninja:
React native 0.54.2 (just upgraded) and still facing this issue.

dannycochran:
As far as I can tell, the change in this commit does not actually work.

We've been using the following change in production for the past 4 months and it works:
https://github.com/facebook/react-native/pull/18522

RickDT:
The solution from @dannycochran has two advantages: staying with Sync execution and not relying on shared system executors. I would much prefer it get merged.

I think the root of the problem is that the shared system executors are getting out of whack when restarting the React Native application without restarting the underlying Android Main Activity.

And (sorry to repeat myself), arbitrarily swapping to parallel execution could definitely result in unanticipated knock-on effects for some app developers.

slegaitis:
This is still a problem. RN 0.55.1 and very annoying. Testing on real device with Chrome debugging and live reload.

slegaitis:
I figured a sort of reason why this is happening. If You click CTRL + S or CMD + S to save the file twice this will trigure, Live Reload. ( If You click it twice in a row you should be able to replicate the issue). Sometimes if you click save once Live Reload doesn't work, and if you click it second time, compiler possibly doesn't finish or there is some sort of other issue and AsyncStorage does not resolve promises from then.

Workaround that works for me for now is to close the app, go to Simulator/ Real device menu and open it again. Then AsyncStorage works again.

nstanwar:
Issue still persist in react-native: 0.54.2

Pastromhaug:
I still have this problem in 0.55.4

Gennady77:
The same problem. RN 0.55.4, Android 6.0

tntdarken:
I think I have found a possible solution to this problem. I hope it helps:

Component:
componentWillMount(){
        this.storageClass.getStorage().then( retorno => this.setState(retorno));
    }

storageClass:
async getStorage(){
        data = await AsyncStorage.getAllKeys();
        return AsyncStorage.multiGet(data).then( data => {
            let retorno = {};
            data.forEach( async x => {
                key = x[0];
                value = x[1];
                retorno[key] = value;
            });
            return retorno;
        });
    }

leandrohsilveira:
As @slegaitis said, it is something related to Live Reload option.
It works when the app is opened with the Live Reload disabled.

Fortunately, it also works with "Reload" option and the VSCode Extension called [React Native Tools](https://marketplace.visualstudio.com/items?itemName=vsmobile.vscode-react-native) has a shortcut to use this option using `CTRL + Shift + P` > `React-Native: Reload App`.

bstst:
It does not work with remote debugging enabled though, no matter if Live Reload is enabled.

haggholm:
Any news on this? It looks to me like the CI test was failing for reasons unrelated to the PR; maybe it just needs to re-run? It would be great to see this in the next RN release…

ITsolution-git:
Hello.
When I use debugger, asyncstorage promise seems to be resolved.
But it does not work without debugging, and it does not work on testflight deployed.
The asyncstorage promise never resolves, and even the app freezes and abort after about 5-10 seconds.

I placed an loader till the promise resolves but loader is not rendered and seems like the app refused to render anymore.

Please help anyone.
Thanks

jsdario:
Closing as the bug is solved and this PR is over a year old

Waqas-Jani:
i have the same issue .. asynstroage not working on Android version 24+(7+).. any one has solutaion??

ollyfg:
@Waqas-Jani Can you tell us what version of React Native you are running? Are you using the older built-in AsyncStorage or the newer community one (`@react-native-community/async-storage`)? Are you seeing any related errors, either through JS debugging or logCat?

Either way, this bug has been solved ad you might be seeing a different one. I would recommend opening a new issue on https://github.com/react-native-community/async-storage

PR #16913, opened 2017-11-21
Updates CircleCI config

grabbou:
Contains changes needed for the `publish` step to be fixed.


grabbou:
@hramos this is ready to be shipped. I've managed to ship 0.51-stable with this config.

 Note, that I ignored your `BUCK` bump as it doesn't work on master and results in a different issue. On 0.51-stable I simply reverted Instagram commit that breaks it. 

This has to be figured out at some point.

.circleci/config.yml
```diff
@@ -110,9 +132,54 @@ aliases:
...
+android_defaults: &android_defaults
+  <<: *defaults
+  docker:
+    - image: circleci/android:api-26-alpha
+  resource_class: "large"
+  environment:
+    - TERM: "dumb"
+    - ADB_INSTALL_TIMEOUT: 10
+    - _JAVA_OPTIONS: "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"
+    - GRADLE_OPTS: '-Dorg.gradle.daemon=false -Dorg.gradle.jvmargs="-XX:+HeapDumpOnOutOfMemoryError"'
```
AndrewJack: Could try `JVM_OPTS: -Xmx3200m` or larger if you've increased the size?
https://circleci.com/docs/2.0/language-android/

grabbou: Shouldn't be needed with `UseCGroupMemoryLimitForHeap` flag. As per Twitter discussion, we tried different combinations and none of them ended up working.

Decided to increase the container size for Android as an alternative (and free) approach. CircleCI support turned it on for our repository after my request.

hramos:
Amazing, thanks!

PR #16915, opened 2017-11-21
Implement requiresMainQueueSetup in RCTTVNavigationEventEmitter

charpeni:
## Motivation

Fix the following warning:

> Module RCTTVNavigationEventEmitter requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.

## Test Plan

Trivial change.

## Release Notes

[IOS] [MINOR] [RCTTVNavigationEventEmitter] - Implement `requiresMainQueueSetup` in `RCTTVNavigationEventEmitter`

### DIFF
--- a/React/Modules/RCTTVNavigationEventEmitter.m
+++ b/React/Modules/RCTTVNavigationEventEmitter.m
@@ -17,6 +17,11 @@ static NSString *const TVNavigationEventName = @"onTVNavEvent";
 
 RCT_EXPORT_MODULE()
 
++ (BOOL)requiresMainQueueSetup
+{
+  return NO;
+}
+
 - (instancetype)init
 {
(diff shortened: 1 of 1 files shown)


janicduplessis:
I think a better fix would be to use startObserving / stopObserving to manage the subscription instead of init / dealloc. Something like https://github.com/charpeni/react-native/blob/c1198ebedb995664d8856343b240a23069a74c9d/React/Modules/RCTKeyboardObserver.m#L20.

shergin:
cc @dlowder-salesforce

shergin:
Well... While it is not required main threaded initialization, this fix is legit.

Issue #16543, opened 2017-10-26
rn update from 0.44 to 0.49 error: Unable to resolve module 'uuid' from ..


cq920702:
### Is this a bug report?
YES
(write your answer here)
When I upgrade rn from 0.44.0 to 0.49.3. got error:Unable to resolve module 'uuid' from '..node_modules/react-native/Libraries/Blob/Blob.js':Mudule does not exist in the module map

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

(Write your answer here.)

### Environment
Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.5.0
  Yarn:  Not Found
  npm:  4.2.0
  Watchman:  4.7.0
  Xcode:  Xcode 9.0 Build version 9A235
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0-alpha.6 => 16.0.0-alpha.6
  react-native: 0.49.3 => 0.49.3

### Steps to Reproduce

(Write your steps here:)

1.Xcode run
2. Simulator show the error
3.

### Expected Behavior

(Write what you thought would happen.)
Can run correctly.
### Actual Behavior

(Write what happened. Add screenshots!)

### Reproducible Demo

(Paste the link to an example project and exact instructions to reproduce the issue.)

Kureev:
Hi @cq920702! Did you try to re-install `node_modules`?

joefazz:
Hi @cq920702 , I recently went through a similar experience upgrading from 0.44 to 0.49. I strongly recommend following [this articles advice](https://medium.com/react-native-training/updating-your-react-native-app-a724c996a76d). I had a similar issue and it was due to a change in 0.48 where the RCTBlob.xcodeproj couldn't auto merge correctly. The solution was to go inside xcode and add the project to the file manually.

I hope this helps you out, let me know if you have any questions.

KaterGor:
Hi! I am experiencing the same issue.
When I run npm install, uuid dependency is not found. 
Using yarn instead of npm is solving this issue.
My assumption is that yarn places uuid as a shared dependency at the top of the dependency tree and npm does not do it. If uuid module is used in the blob library, it should b explicitly added to React dependencies.

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.7.0
  Yarn:  1.2.1
  npm:  5.5.1
  Watchman:  4.9.0
[...]

PR #11573, opened 2016-12-20
Implement Blob support for XMLHttpRequest

satya164:
This PR is a followup to https://github.com/facebook/react-native/pull/11417 and should be merged after that one is merged.

### What does the PR do?
1. Add support for creating blobs from strings, not just other blobs
2. Add the `File` constructor which is a superset of `Blob`
3. Add the `FileReader` API which can be used to read blobs as strings or data url (base64)
4. Add support for uploading and downloading blobs via `XMLHttpRequest` and `fetch`
5. Add ability to download local files on Android so you can do `fetch(uri).then(res => res.blob())` to get a blob for a local file (iOS already supported this)

### Test plan
1. Clone the repo https://github.com/expo/react-native-blob-test
2. Change the `package.json` and update `react-native` dependency to point to this branch, then run `npm install`
3. Run the `server.js` file with `node server.js`
4. Open the `index.common.js` file and replace `localhost` with your computer's IP address
5. Start the packager with `yarn start` and run the app on your device

If everything went well, all tests should pass, and you should see a screen like this:

screen shot 2017-06-08 at 7 53 08 pm!

Pull to rerun all tests or tap on specific test to re-run it

## Release Notes
 [GENERAL] [FEATURE] [Blob] - Implement blob support for XMLHttpRequest

PR #16936, opened 2017-11-23
Add uuid dependency needed by react-native/Libraries/Blob/Blob.js.

mikelambert:
This should fix https://github.com/facebook/react-native/issues/16543

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,6 +193,7 @@
     "semver": "^5.0.3",
     "shell-quote": "1.6.1",
     "stacktrace-parser": "^0.1.3",
+    "uuid": "^3.1.0",
     "whatwg-fetch": "^1.0.0",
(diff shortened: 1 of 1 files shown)


EricPKerr:
Noticed this as an issue when upgrading from RN 0.51 to 0.52

mikelambert:
@hramos? Not sure who to ping for this.

hramos:
I'm looking at this via #11573 which also adds `uuid`. We cannot pull in `uuid` at the moment due to a conflict it would introduce with an internal uuid library we use. Solving this may require moving RN off using Haste. TBD. To avoid splitting the conversation, I am going to close this PR, but feel free to ping me if for whatever reason `uuid` is not added via #11573.

PR #16937, opened 2017-11-23
Update Style.md

PaulSebi:
Small change in language to better match commonplace jargon

## Motivation

(Write your motivation here.)

## Test Plan


hramos:

PR #16946, opened 2017-11-23
Added Reachability Library Note

fuatsengul:
## Motivation

NetInfo functions are broken because Reachability Library should be installed manually. I've added a note and link for this library.

## Test Plan

No test required, just added information.


hramos:
Can you add this to the markdown docs instead? See https://github.com/facebook/react-native-website/blob/master/docs/netinfo.md

Issue #16774, opened 2017-11-09
"Something went wrong while linking. Error: Cannot read property 'match' of undefined"


fungilation:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

`$ react-native info`:
```
Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.9.1
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.9.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: Not Found

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: 0.50.3 => 0.50.3
```

  Target Platform: iOS 11.1

### Steps to Reproduce

1. `yarn add react-native-blur@3.2.2`
2. `react-native link react-native-blur`

### Expected Behavior

Package linked without error

### Actual Behavior

```
$ react-native link react-native-blur

rnpm-install info Linking react-native-blur android dependency 
rnpm-install info Android module react-native-blur has been successfully linked 
rnpm-install info Linking react-native-blur ios dependency 
rnpm-install ERR! Something went wrong while linking. Error: Cannot read property 'match' of undefined 
Please file an issue here: https://github.com/facebook/react-native/issues 

Cannot read property 'match' of undefined

```

Note that I haven't seen a similar error, before RN 0.50 or react-native-blur <= 3.2.0

Issue also reported at https://github.com/react-native-community/react-native-blur/issues/246

ngankt2:
same bug with "react-native": "^0.50.3" and react-native link ANYTHING => same bug
I did the following and it worked:

step1: remove node_modules
step2: close xcode
step3: cd ios and run pod install (i have use firebase) 
step4: react-native link ANYTHING => and OK
 
I hope it will help you

henrydeclety:
I get the same error with react-native-keychain when trying to link
`rnpm-install info Linking react-native-keychain android dependency 
rnpm-install info Android module react-native-keychain has been successfully linked 
rnpm-install info Linking react-native-keychain ios dependency 
rnpm-install ERR! Something went wrong while linking. Error: Cannot read property 'match' of undefined 
Please file an issue here: https://github.com/facebook/react-native/issues 

Cannot read property 'match' of undefined`

And if I uninstall react-native-keychain, the same happens with react-native-vector-icons:
[...]

PR #16970, opened 2017-11-24
Improved error reporting for react-native link's Cocoapods integration

erlendfh:
## Motivation
Ran into an issue (#16774) with the new cocoapods integration after upgrading to 0.50.3. The error message `Cannot read property 'match' of undefined` is not very helpful, so I've added some more tests and a bit better error reporting.

## Test Plan

I've updated and added to the existing set of specs for the local-cli/link/pods implementation, and added a new example Podspec. I've also manually tested this version against a react native project by running `link` and `unlink`, both with a valid and invalid Podspec. Running with an invalid podspec now yields this:

> rnpm-install info Linking react-native-video android dependency
> rnpm-install info Android module react-native-video has been successfully linked
> rnpm-install info Linking react-native-video ios dependency
> rnpm-install info Adding podspec for react-native-video to testproject
> rnpm-install info Couldn't find "# Add new pods below this line" in Podfile, will try to find target...
> rnpm-install ERR! Something went wrong while linking. Error: Couldn't find "# Add new pods below this line" or "target 'testproject' do" in Podfile, unable to continue
> Please file an issue here: https://github.com/facebook/react-native/issues
> 

## Release Notes

[IOS] [BUGFIX] [cli] - Improved error messages for working with Podspecs that don't match the expected format


erlendfh:
Maybe @Kureev or @grabbou would be willing to take a look at this Cocoapods-related PR?

Kureev:
This PR is definitely a step into right direction, although I'm wondering if `MARKER_TEXT` is required at all. I'd rather detect the last pod in the list and append new one afterwards. Is it something other people would like to consider?

erlendfh:
@Kureev Thanks for taking a look at this! Just to make a couple of clarifications:
- The original PR already implements looking for `target '${projectName}' do` and then adds the pod below that line
- MARKER_TEXT is not something I've introduced, it was already in the original PR for Podspec support. MARKER_TEXT will only be used if the above method fails, i.e. for more complex Podfile configurations. All I've done is add a few tests and ensure that a more informative error message is displayed in the npm log.

erlendfh:
@Kureev any chance of getting this PR merged?

cpojer:
This PR has been open for a long time without resolution. We just moved this code to https://github.com/react-native-community/react-native-cli so please submit your PR there.

PR #16978, opened 2017-11-25
debugger-ui js does not support object spread operator  fixes #16977

### LINKED ISSUE #16977: Remote debugger bug, react-native v 0.50 android
### Is this a bug report?
yes

### Environment

Environment:
  OS: Linux 4.10
  Node: 8.5.0
  Yarn: 1.0.1
  npm: 5.3.0
  Watchman: Not Found
  Xcode: N/A

  Android Studio: 2.3 AI-162.4069837

  react: 16.0.0 => 16.0.0
  react-native: 0.50.4 => 0.50.4

Browser: Chrome Version 59.0.3071.115 (Official Build) (64-bit)
 Target: (Android 8.0)

### Steps to Reproduce

`react-native init AwesomeProject`
`cd AwesomeProject`
`react-native run-android`
enable remote debugger on device

open browser to http://localhost:8081/debugger-ui/

### Expected Behavior

device should connect to remote debugger with no errors

### Actual Behavior

device will not connect to remote debugger. 

console will display - 

 'Uncaught SyntaxError: Unexpected token ...'  localhost/:183

localhost/:183 -         
worker.postMessage({
          ...object,
          url: await getBlobUrl(object.url),
      });

I reverted back to 0.49 and it got rid of the error, and am able to connect properly, so I am sure it is with 0.50

it looks like the error is coming from object spread operator in /react-native/blob/master/local-cli/server/util/debugger-ui/index.html

Object spread is not supported pre version 60 on google chrome. 

### Reproducible Demo

react-native init ${app name}

universalmind303:
## Motivation

debugger-ui was broken on v0.50 + 

## Test Plan

JS does not support object spread operator natively, So I simply rewrote 1 block of code to use Object.assign instead of Object spread. 

## Release Notes

[CLI][BUGFIX][local-cli/server/util/debugger-ui/index.html] - changed from object spread to Object.assign line 182

It looks like the cause of the issue stems */debugger-ui/index.html using Object Spread features

Ideally, async/await should be removed from debugger-ui as well to support further backwards compatibility.

My fix should be backwards compatible as far back as google chrome v55 and Firefox 57

The debugger should not need to rely on the newest browser versions to work properly.

### DIFF
--- a/local-cli/server/util/debugger-ui/index.html
+++ b/local-cli/server/util/debugger-ui/index.html
@@ -179,10 +179,11 @@
         shutdownJSRuntime();
         Page.setState({status: {type: 'disconnected'}});
       } else if (object.method === 'executeApplicationScript') {
-        worker.postMessage({
-          ...object,
-          url: await getBlobUrl(object.url),
-        });
+        worker.postMessage(
+          Object.assign({}, object, {
+              url: await getBlobUrl(object.url),
+            }
+        ));
       } else {
         // Otherwise, pass through to the worker.
         worker.postMessage(object);


universalmind303:
It looks like the cause of the issue stems from google chrome. object spread is not supported until version 60. and async/await at 55. 

Any chrome version before 60 and debugger will not work. 
The object.assign fix will support versions back to 55. 
Anything before that will cause an issue because of the async/await in the same file.

cpojer:

PR #16992, opened 2017-11-28
Catch attempts to import * as x from react-native

rh389:
## Motivation
Motivated by some very misleading redbox errors. When my IDE (Webstorm) slipped up auto-generating an import line, I ended up with this in my code:
`import * as Platform from 'react-native'`

The result:
### Android (with remote debugging enabled)
```
Navigator is deprecated and has been removed 
from this package. It can now be installed and 
imported from `react-native-deprecated-custom-
components` instead of `react-native`. Learn about 
alternative navigation solutions at http://facebook
.github.io/react-native/docs/navigation.html`
```

### iOS
```
Native module cannot be null.
```

This is because `import * ...` triggers the getters in `react-native-implementation.js`, some of which throw errors for various reasons. These error messages had me chasing ghosts until I noticed my bad import line.

(NB: Android without remote debugging didn't throw any errors at all - I guess the getters are called lazily?)

## Test Plan
The situation above now results in the following redbox on iOS and android:
image

## Release Notes
[GENERAL] [ENHANCEMENT] [Libraries/react-native/react-native-implementation.js] - Better error reporting when attempting to import * from react-native


rh389:
I don’t know who best to review a developer experience issue - maybe @hramos ?

hramos:
Can you rebase?

empyrical:
What if these deprecated/erroring modules had the property descriptor `enumerable: false`, so you could still `import *` without calling those getters? 

Here's a quick example against Babel's `_interopRequireWildcard` that works without throwing an error:

```javascript
const fakeModule = { 'foo': 'bar' };

Object.defineProperty(fakeModule, 'invisible', {
  enumerable: false,
  get() {
    throw new Error('Bad getter called');
  }
});

function _interopRequireWildcard(obj) { if (obj && obj.__esModule) { return obj; } else { var newObj = {}; if (obj != null) { for (var key in obj) { if (Object.prototype.hasOwnProperty.call(obj, key)) newObj[key] = obj[key]; } } newObj.default = obj; return newObj; } }

const fakeImport = _interopRequireWildcard(fakeModule);
```

rh389:
Seems like a much better solution to me

empyrical:
I took a stab at having deprecated exports (not just Navigator) be unenumerable, but I discovered a side-effect of how `import` is transpiled that makes that a bit more complicated

Say you wanted to use the wildcard import:

```js
import * as ReactNative from 'react-native';

ReactNative.SomeApi.someThing();
```

Then in a later release `SomeApi` is deprecated. Since it's unenumerable, what I thought would be nice is have it so it could still be manually imported while you transition away:

```js
import * as ReactNative from 'react-native';
import {SomeApi} from 'react-native';

SomeApi.someThing();
```

[Babel 6 in this fiddler](https://babeljs.io/en/repl.html#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAKjgQwM5wEoFNkGMYByyMwAblnAGZQQhwDkUO-AtAHbFlb0DcAUKEiw4AbwDKtLAEEwwAL5UadRsxjtO5Xnz4SQ02QDpUkgCoALYGwDmACgCUPIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Clatest%2Ces2015-loose&prettier=false&targets=&version=6.26.0&envVersion=) transpiles this to:

```js
var _reactNative = require('react-native');

var ReactNative = _interopRequireWildcard(_reactNative);

// snipped helper definition

_reactNative.SomeApi.someThing();
```

But Babel 7 in Metro transpiles it to:

```js
  var ReactNative = babelHelpers.interopRequireWildcard(_$$_REQUIRE(_dependencyMap[0], "react-native"));
  ReactNative.SomeApi.someThing();
```

The big important difference here is in the first example, it's trying to get `SomeApi` right off of react native's export object, but in the second example if you try to wildcard import from the module then it will try to get `SomeApi` off of the wildcard export object. Since it was declared unenumerable, it won't be available.

I'm not sure if this is a babel 7 issue or a metro babel plugin issue. Will try and look into this later.

empyrical:
Actually - it appears as if this stuff isn't needed after all. `_interopRequireWildcard` was updated in Babel 7 so it will copy getter functions over without calling them! So `import * as RN from 'react-native'` will work without error in RN 0.57

cpojer:
Based on @empyrical's comment, I'll close this PR, as it seems like with babel7 this won't be necessary any longer. While I don't personally recommend using `import *`, that's a subjective thing and everyone should be able to use it as much as they'd like to :)

PR #17000, opened 2017-11-28
Add support for scoped npm package linking in local-cli

vekerdyb:
## Motivation

Linking scoped npm packages breaks the android build using the local cli.

I found no documentation for this, but Android Studio correctly replaces the `/` in the npm package name with `_` in the gradle include statement.

## Test Plan

The test included with this PR verifies, that scoped dependencies, are linked with this replacement.
 
## Release Notes
 [CLI] [BUGFIX] [local-cli/link/link.js] - Change link function for scoped npm packages to link as expected on Android

### DIFF
--- a/local-cli/link/link.js
+++ b/local-cli/link/link.js
@@ -51,6 +51,8 @@ const dedupeAssets = (assets) => uniqBy(assets, asset => path.basename(asset));
 
 
 const linkDependencyAndroid = (androidProject, dependency) => {
+  dependency.name = dependency.name.replace(/\//g, '_');
+
   if (!androidProject || !dependency.config.android) {
     return null;
   }
--- a/local-cli/link/__tests__/link.spec.js
+++ b/local-cli/link/__tests__/link.spec.js
@@ -205,4 +205,42 @@ describe('link', () => {
       done();
     });
   });
+
+  it('should replace forward slashes in scoped package names for android', () => {
+    const isInstalledAndroid = jest.fn(() => false);
+    const registerNativeModule = sinon.stub();
+    const registerDependencyAndroid = jest.fn();
+    const config = {
+      getProjectConfig: () => ({ ios: {}, android: {}, assets: [] }),
+      getDependencyConfig: sinon.stub(),
+    };
+
+    const  mockDependencyConfig = [{
+      config: { ios: {}, android: {}, windows: null, assets: [], commands: {}, params: [], },
+      name: '@scope/react-native-package',
+    }];
+
+    jest.setMock('../ios/isInstalled.js', sinon.stub().returns(true));
+
+    jest.mock('../getDependencyConfig.js', () => () => mockDependencyConfig);
+
+    jest.mock(
+      '../android/isInstalled.js',
+      () => isInstalledAndroid,
+    );
+
+    jest.setMock(
+      '../ios/registerNativeModule.js',
+      registerNativeModule,
+    );
+
+    jest.mock('../android/registerNativeModule.js', () => registerDependencyAndroid);
+
+    const link = require('../link').func;
+
+    return link(['react-native-test'], config).then(() => {
+      expect(isInstalledAndroid).toHaveBeenCalledWith({}, '@scope_react-native-package');
+      expect(registerDependencyAndroid).toHaveBeenCalledWith('@scope_react-native-package', {}, {}, {});
+    });
+  });
 });


local-cli/link/link.js
```diff
@@ -51,6 +51,8 @@ const dedupeAssets = (assets) => uniqBy(assets, asset => path.basename(asset));
 
 
 const linkDependencyAndroid = (androidProject, dependency) => {
+  dependency.name = dependency.name.replace(/\//g, '_');
```
grabbou: I don't think this is the right place for this change. First and foremost thing is that it mutates `dependency` object making this function no longer pure. 

Another thing is that I can't tell where the new value with `/` replaced to `_` should be used. In your comment you said that it breaks `include` statement in Gradle. 

Maybe you should do this replacement directly inside the function that puts `include` statement into Gradle file?

grabbou: E.g. here https://github.com/vekerdyb/react-native/blob/e9e841628fc77faa9bd271c86a031d0518d5e7e6/local-cli/link/android/patches/makeSettingsPatch.js#L22, make `const escapedName = name.replace()` and done :) (hopefully)

Kureev: Mmm, that will only work if this `name` isn't re-used anywhere else. IIRC, we use it in many places, so we need to have a consistent name across different functions. As a compromise, we can introduce a new function, something like `sanitizeName` or similar that only transforms a given dependency name into a version that will be understandable by Grade and other parts of the `react-native link`.

grabbou: > Mmm, that will only work if this name isn't re-used anywhere else

Yup, and that way it's explicit where we sanitize the name. Doing it globally on a `dependency` can have side effects as the `name` field is supposed to be the name of the package. 

Function is a great idea that plays well with our design. I think I've seen it already in `generator` subfolder where they do the same for `react-native init` (sanitize the name when creating Android main application). We could potentially reuse that as well :)

vekerdyb: Thanks for the feedback! Unfortunately I am quite busy at the moment, but will get back to this mid February, unless it is picked up by someone else by then.
>  I think I've seen it already in generator subfolder where they do the same for react-native init (sanitize the name when creating Android main application). We could potentially reuse that as well 
I agree, I faintly remember seeing `init` working as expected now.

Titozzz: Hello @vekerdyb , did you get back to this or shall I add it to my march TODO ? :)

vekerdyb: Hey @Titozzz, unfortunately I haven't had the time yet, but will aim to do it this weekend. If not, I will let you know!

vekerdyb: And I didn't have any screen time this weekend, sorry, probably best if it goes on your march TODO!

Titozzz: Okay, I'll tell you if/when I start working on it !

Titozzz: Ok so I spent one hour trying to understand what was happening, but I think I finally got it, there was some regressions due to react-native 0.54.0...

The function makeBuildPatch gets called with an object instead of the name ..., @grabbou , i'll fix it and we should release 0.54.1 because it completely destroy the detection of previous android links and re-links every time

Titozzz: Nevermind i'll just base of https://github.com/facebook/react-native/pull/18131

Titozzz: Okay so this thing https://github.com/facebook/react-native/commit/2c5fbd7 broke so much stuff everywhere, android linking detection is broken (fixed by https://github.com/facebook/react-native/pull/18131) but also unlinking is broken on android AND iOS.

Fixing everything is going to require more time @vekerdyb @grabbou

grabbou: @rozele that looks like a major regression - can you check what's failing and whether we can get this fixed quite easily?

rozele: I published #18207 to fix the issue for Android, I haven't been able to get a repro of the unlink issue for iOS.

Titozzz: Because your branch is being landed, i'll apply my fixes to it 💛

Titozzz: There you go: #18275

Titozzz:
@Kureev You can close this issue now that #18275 has landed

cpojer:

PR #17004, opened 2017-11-28
Update Performance.md

kelset:
## Motivation

Removing the `react-addons-perf` reference since now RN relies on React 16, and it's not [supported anymore](https://reactjs.org/docs/perf.html).

## Test Plan

Nothing to test.

## Release Notes

[DOCS] [BUGFIX] [Performance.md] removing old reference in the doc.


brentvatne:
@facebook-github-bot shipit

PR #17006, opened 2017-11-28
Docs: Add example to Modal documentation

AleksandrChernyavenko:
## Motivation

I find it easier to understand the behavior of a component when there is a simple example showing its usage. I recently used the Modal component and noticed that it doesn't have a code example. This PR adds an example to the Modal documentation.

## Test Plan

N/A

## Release Notes
[DOCS][ENHANCEMENT][Modal] - Added example to documentation


hramos:

PR #17011, opened 2017-11-29
Eliminate openURL deprecated error if deployment target is 10.0 or above

ocleo1:
## Motivation

openURL deprecated is marked as error since iOS 10.0

openurl

## Test Plan

1. Compiling RCTLinking should be no error if iOS deployment target is iOS 10.0 or above.
2. Opening web URL in iOS, its behavior should be same as previous by invoking Linking.openURL().

## Release Notes
 [IOS] [ENHANCEMENT] [Linking] - Eliminate openURL deprecated error if deployment target is 10.0 or above

### DIFF
--- a/Libraries/LinkingIOS/RCTLinkingManager.m
+++ b/Libraries/LinkingIOS/RCTLinkingManager.m
@@ -89,12 +89,13 @@ RCT_EXPORT_METHOD(openURL:(NSURL *)URL
                   resolve:(RCTPromiseResolveBlock)resolve
                   reject:(RCTPromiseRejectBlock)reject)
 {
-  BOOL opened = [RCTSharedApplication() openURL:URL];
-  if (opened) {
-    resolve(nil);
-  } else {
-    reject(RCTErrorUnspecified, [NSString stringWithFormat:@"Unable to open URL: %@", URL], nil);
-  }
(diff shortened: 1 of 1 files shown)


ide:
Does this compile if the target is iOS 8 still?

shergin:
@ocleo1 Could you please make it compatible with iOS 8? You can see some examples in `RCTScrollView.m` (we have to check for it in both compile-time and runtime).

ocleo1:
@ide Sorry for the late reply. Yes, it is still compiled successfully if target to iOS 8.

ocleo1:
@shergin Sorry for the late reply. I tested my change if target to iOS 8. URL can be opened properly. Runtime should be good, no need add something like `#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 100000 /* __IPHONE_10_0 */`. How do you think?

ocleo1:
@cpojer Got it. Let me check and update PR. Thanks

cpojer:
@ocleo1 do you have any updates for us?

hramos:
This can be closed now that https://github.com/facebook/react-native/commit/fd3db034276b28e4769cd8b339280885f2e4d288 has landed.

PR #17015, opened 2017-11-29
Update native-modules-android.md

dsandmark:
`import com.facebook.react.bridge.Callback;` is needed to be able to use `Callback` as a parameter.

## Motivation

Thought it needed to be clearer that you need to import `Callback` as it's not a native Java type.

## Test Plan


chirag04:
@facebook-github-bot shipit

hramos:

You may want to make the same change over there. Sorry for the unnecessary work here, we're still in the process of landing the commits that will remove the duplicate docs from this repo.

chirag04:
Nice! I didn’t realize it was moved otherwise wouldn’t have shipped, 😐

PR #17019, opened 2017-11-29
Add focus() and blur() to TextInput documentation

aghoneim92:
I needed the focus function to handle a submit on a previous TextInput in a form and realized that even though it and blur() exist on TextInputs they are not documented.


hramos:

PR #17020, opened 2017-11-30
Adding support for custom accessibility actions on iOS.

ewdavison:
## Motivation

This feature has been requested by customers.  Our previous (pre-react) application had support for custom accessibility actions.  

This feature allows UI elements to provide a list of custom actions that can be read when VoiceOver is enabled.  UI elements expose one accessibility action by default.  Some UI elements may support multiple actions though other mechanisms like tap and hold.  To expose these actions in an accessible way iOS provides custom accessibility actions.

## Test Plan

Feature was tested in the iOS simulator using the Accessibility Inspector.  Custom actions were added to a button and observed in the tool.  Custom actions were also invoked using the tool and then stepped through in the debugger.

The feature was also tested on an iPhone.  VoiceOver was enabled on the device and custom actions were observed for controls that exposed them.

We have been using this feature in our app for some time as well.

## Release Notes

[IOS] [ENHANCEMENT] [Accessibility] - Added support for custom accessibility actions

Eric Davison
Microsoft Corp.


hramos:
Thanks for adding this! Can you update the website docs as well? https://github.com/facebook/react-native-website/blob/master/docs/view.md

PR #17036, opened 2017-12-01
Fix local-cli's installedGlobally check to work on Windows platforms

sballew:
## Motivation

The react-native local-cli does a check to see if it is being run from a global install or not. If running from a global install, an error is printed and the CLI exits.

This check for a global install does not work on Windows. The check of `process.argv` does not contain the expected `node_modules/.bin/react-native`. It instead contains a direct path to the `wrong-react-native.js` file, as determined by the `node_modules/.bin/react-native.cmd` entry point.

This update will, on Windows platforms, do a global check by instead looking for the existence of a package.json above the node_modules. If not found, we assume a global install and print the error.

## Test Plan

In a react-native project, I originally tried running the local react-native cli:

```
> yarn react-native --version
yarn run v1.3.2
$ E:\myproject\node_modules\.bin\react-native --version
Looks like you installed react-native globally, maybe you meant react-native-cli?
To fix the issue, run:
npm uninstall -g react-native
npm install -g react-native-cli
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I replaced the `wrong-react-native.js` with the modified version and reran the command:

```
> yarn react-native --version
yarn run v1.3.2
$ E:\myproject\node_modules\.bin\react-native --version
Scanning folders for symlinks in E:\myproject\node_modules (93ms)
0.50.3
Done in 1.86s.
```

## Release Notes

[CLI] [BUGFIX] [local-cli/wrong-react-native.js] - Updated local-cli on Windows to check for the absence of a package.json file to determine if being run from a global installation or not.

### DIFF
--- a/local-cli/wrong-react-native.js
+++ b/local-cli/wrong-react-native.js
@@ -16,8 +16,10 @@ const isWindows = process.platform === 'win32';
 
 var installedGlobally;
 if (isWindows) {
+  // On Windows, assume we are installed globally if we can't find a package.json above node_modules.
   installedGlobally = !(fs.existsSync(path.join(__dirname, '../../../package.json')));
 } else {
+  // On non-windows, assume we are installed globally if we are called from outside of the node_mobules/.bin/react-native executable.
   var script = process.argv[1];
   installedGlobally = script.indexOf('node_modules/.bin/react-native') === -1;
 }
--- a/local-cli/wrong-react-native.js
+++ b/local-cli/wrong-react-native.js
@@ -9,13 +9,12 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-const fs = require('fs');
-const path = require('path');
-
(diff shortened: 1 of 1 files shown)


local-cli/wrong-react-native.js
```diff
@@ -9,8 +9,21 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-var script = process.argv[1];
-var installedGlobally = script.indexOf('node_modules/.bin/react-native') === -1;
+const fs = require('fs');
+const path = require('path');
```
GantMan: you can move these requires inside of the windows code, so they won't be required on non-win32.

Not a huge refactor, but I'd do it.

sballew: Agreed, and done.

TheSavior:
Thanks for contributing!

PR #17048, opened 2017-12-01
Fixing bad link to RN documentation.

ryanml:
## Motivation

The link referencing the root of RN documentation was leading to a 404 error. As the docs don't seem to have a true "index" page, it makes sense that it goes to the top level entry in the docs, the "Getting Started Section". The alternative solution may be not linking that phrase at all, this seems helpful though.

## Test Plan

No changes in this pull request require any sort of unit/AB testing, the link directing to the correct place should be apparent.

## Release Notes
[DOCS] [BUGFIX] [README.md] - Fixing doc resource link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Follow the [Getting Started guide](https://facebook.github.io/react-native/docs/
 
 ## Documentation
 
-[The website’s documentation](https://facebook.github.io/react-native/docs/) is divided into multiple sections.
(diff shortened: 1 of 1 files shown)


chirag04:
@facebook-github-bot shipit

hramos:
Thank you for catching this! Prior to today, that link would have worked. I'll see if I can fix this in the updated website.

ryanml:
@hramos great :) Thanks for getting back to me

PR #17059, opened 2017-12-04
fix API>=21 cookieManager not call flush()

xialei92:
fix API>=21 cookieManager not call flush()


radeno:
@xialei92 is this still related? What is your solution different in behaviour than this commit? https://github.com/facebook/react-native/commit/b01ac3bffea0180002439a15e3ecf435e34dd0fe (connected PR https://github.com/facebook/react-native/pull/19770)

Issue #11655, opened 2016-12-28
Newly initialized react native project has flow errors


drewwyatt:
### Description

The long and short of it is, I'm getting flow errors on brand new projects created with the react native cli (version 2.0.1). Do I have a weird version mismatch or something?

### Reproduction

    drew.wyatt$ react-native --version
    react-native-cli: 2.0.1
    react-native: 0.39.2
    
    drew.wyatt$ flow version
    Flow, a static type checker for JavaScript, version 0.35.0

 ```
Drew-Wyatt:Projects drew.wyatt$ react-native init Whatever
    This will walk you through creating a new React Native project in /Users/drew.wyatt/Projects/Whatever
    Using yarn v0.16.1
    Installing react-native...
    yarn add v0.16.1
    info No lockfile found.
    [1/4] 🔍  Resolving packages...
    warning react-native > xcode > node-uuid@1.4.7: use uuid module instead
    [2/4] 🚚  Fetching packages...
    [3/4] 🔗  Linking dependencies...
    warning Unmet peer dependency "react@~15.4.0-rc.4".
    [4/4] 📃  Building fresh packages...
    success Saved lockfile.
    success Saved 466 new dependencies.
    ├─ absolute-path@0.0.0
    ├─ accepts@1.2.13
    │  ├─ mime-db@1.25.0
    │  └─ mime-types@2.1.13
    ├─ align-text@0.1.4
    ├─ ansi-escapes@1.4.0
    ├─ ansi-regex@2.0.0
    ├─ ansi-styles@2.2.1
    ├─ ansi@0.3.1
    ├─ are-we-there-yet@1.1.2
    ├─ array-differ@1.0.0
    ├─ array-filter@0.0.1
    ├─ array-find-index@1.0.2
    ├─ array-map@0.0.0
    ├─ array-reduce@0.0.0
    ├─ array-union@1.0.2
    ├─ array-uniq@1.0.3
    ├─ arrify@1.0.1
    ├─ art@0.10.1
    ├─ asap@2.0.5
    ├─ async@2.1.4
    ├─ babel-code-frame@6.20.0
    ├─ babel-core@6.21.0
    │  └─ json5@0.5.1
    ├─ babel-generator@6.21.0
    ├─ babel-helper-builder-react-jsx@6.21.1
    ├─ babel-helper-call-delegate@6.18.0
    ├─ babel-helper-define-map@6.18.0
    ├─ babel-helper-function-name@6.18.0
[...]

PR #17086, opened 2017-12-06
Fix some flow errors that appear

deecewan:
## Motivation

There are some errors I noticed after upgrading my flow version.  I had them in my project, and they were also reported in #11655.  These errors were ignored on master, but I went through and fixed them so the static analysis will work still.

After these changes, I receive no errors using flow `0.60.1` on latest master, or using `0.59` in my local project (which does not have the requisite ignores included).

## Test Plan

- init a new project
     `react-native init --version <path-to-repo> helloworld`
- add the flow binary
     `yarn add -D flow-bin`
- run flow
     `yarn flow`
- make sure there are no errors in the project.
     ```
     No errors!
     ✨  Done in 23.60s.
     ```

## Release Notes

[GENERAL][BUGFIX][./Libraries] - Fix up some flow definitions

### DIFF
--- a/Libraries/Animated/src/Easing.js
+++ b/Libraries/Animated/src/Easing.js
@@ -62,22 +62,14 @@ class Easing {
   /**
    * A stepping function, returns 1 for any positive value of `n`.
    */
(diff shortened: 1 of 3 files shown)


hramos:
Great work

PR #17125, opened 2017-12-08
Expose SectionList getItem

maartenschumacher:
## Motivation

Implementing `getItemLayout` for a SectionList is tricky, because the `index` provided doesn't directly map to the `data`, which is the sections (see also https://github.com/facebook/react-native/issues/14966). If we had access to the `getItem` used internally, then we could reliably map index to the actual item. Example implementation:

```
import { range } from 'lodash'
import { getItem } from 'SectionList'

const heights = {
  sectionHeader: 20,
  row: 40,
  separator: 5,
}

const getItemLayout = (data, index) => ({
  index,
  length: heights[getItem(data, index).type],
  offset: range(index).reduce((result, i) => result + heights[getItem(data, i).type], 0),
})
```

### DIFF
--- a/Libraries/Lists/VirtualizedSectionList.js
+++ b/Libraries/Lists/VirtualizedSectionList.js
@@ -502,7 +502,7 @@ class ItemWithSeparator extends React.Component<
   }
 }
 
-function getItem(sections: ?$ReadOnlyArray<Item>, index: number): ?Item {
+export function getItem(sections: ?$ReadOnlyArray<Item>, index: number): ?Item {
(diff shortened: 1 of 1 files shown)


maartenschumacher:
Or maybe `getItem` should be the third argument to `getItemLayout`? Could be a better api.

maartenschumacher:
Actually it wouldn't even really work because `getItem` doesn't tell you if the item is a section header or footer, so you couldn't really use it for `getItemLayout`.

PR #17146, opened 2017-12-11
Remove internal `utf8` utility

mathiasbynens:
## Motivation

As requested by @davidaurelio in https://github.com/mathiasbynens/utf8.js/issues/17#issuecomment-312500535, this makes it possible for developers to use the `utf8` package from npm in combination with React Native.

Ref. https://github.com/mathiasbynens/utf8.js/issues/17


EdenGottlieb:
https://github.com/facebook/react-native/blob/master/Libraries/Utilities/utf8.js#L9
` * @providesModule utf8`

This is what causes react-native's utf8 to override node_modules one, not the filename.

EdenGottlieb:
The real question would be: why is it even there? I couldn't find a single require/import of utf8 in the whole repo.

mathiasbynens:
> I couldn't find a single require/import of utf8 in the whole repo.

Outside of the tests, indeed!

janicduplessis:
Yea I think we can delete it unless some internal code @fb depends on it.

cc @davidaurelio

mathiasbynens:
@davidaurelio Can you confirm whether this is used internally at Facebook? If not, I’d be happy to update the patch to remove these files.

janicduplessis:
@mathiasbynens You can go ahead and remove it, it won't pass fb internal CI anyway if it's used.

mathiasbynens:
@janicduplessis Done.

hramos:
Landing, I could not find internal uses but like Janic said, CI should block this from landing if it's indeed used internally.

PR #13166, opened 2017-03-27
Fix pinch crash in touch-responsive views.

gillessed:
## Motivation (required)

Multiple react native developer (including myself) have run into a crash with the react-native-photo-view library (and possibly others). The common solution to this problem lies in the underlying java code, and thus requires a change in the react native source.

The stack trace I am getting is the same as listed here https://github.com/alwx/react-native-photo-view/issues/15.

There was a PR to fix this (#12085) but it was closed. In response to the comments there, in my PR, I do log the exceptions. I don't think we can get any closer to the exception because in the next level of the stack trace, we are in the android sdk code.

Looking at some stack overflow pages and the android bug tracker, it seems that this is the common solution to this bug, and does not cause any impact any functionality.

https://code.google.com/p/android/issues/list?can=1&q=pointerindex+out+of+range&colspec=ID+Status+Priority+Owner+Summary+Stars+Reporter+Opened&cells=tiles

## Test Plan (required)

I have manually tested this by compiling react native android from source and have confirmed the exception still gets hit and logged, but does not cause the app to terminate.

PR #12085, opened 2017-01-26
Issue: java.lang.IllegalArgumentException: pointerIndex out of range

mobinni:
Thanks for submitting a pull request! Please provide enough information so that others can review your pull request:

Explain the **motivation** for making this change. What existing problem does the pull request solve?

Currently when implementing a Carousel in React Native ScrollView when you add pinch-to-zoom or some other type of touch handler, functionality the onTouchEvent can crash sometimes on Android.

see:
https://github.com/alwx/react-native-photo-view/issues/15
https://github.com/alwx/react-native-photo-view/issues/22
https://code.google.com/p/android/issues/detail?id=18990
https://github.com/chrisbanes/PhotoView/issues/293
https://github.com/chrisbanes/PhotoView/issues/81

PR #17167, opened 2017-12-13
Fix pinch crash in touch-responsive views.

tobycox:
Fork and rebase of @gillessed's PR https://github.com/facebook/react-native/pull/13166 which has gotten stale.

From original PR:

Motivation (required)

Multiple react native developer (including myself) have run into a crash with the react-native-photo-view library (and possibly others). The common solution to this problem lies in the underlying java code, and thus requires a change in the react native source.

The stack trace I am getting is the same as listed here alwx/react-native-photo-view#15.

There was a PR to fix this (#12085) but it was closed. In response to the comments there, in my PR, I do log the exceptions. I don't think we can get any closer to the exception because in the next level of the stack trace, we are in the android sdk code.

Looking at some stack overflow pages and the android bug tracker, it seems that this is the common solution to this bug, and does not cause any impact any functionality.

https://code.google.com/p/android/issues/list?can=1&q=pointerindex+out+of+range&colspec=ID+Status+Priority+Owner+Summary+Stars+Reporter+Opened&cells=tiles

Test Plan (required)

I have manually tested this by compiling react native android from source and have confirmed the exception still gets hit and logged, but does not cause the app to terminate.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/drawer/ReactDrawerLayout.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/drawer/ReactDrawerLayout.java
@@ -11,10 +11,12 @@ package com.facebook.react.views.drawer;
 
 import android.support.v4.widget.DrawerLayout;
 import android.view.Gravity;
+import android.util.Log;
 import android.view.MotionEvent;
 import android.view.View;
 
 import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.common.ReactConstants;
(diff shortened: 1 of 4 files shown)


tlenclos:
Hi there,
could someone restart the circle CI job on timeout and review this ? 

This issue is encountered on many android apps with pinch functionnalities and right now the only work around is to maintain a fork of React native to avoid crashing the app. I don't mind helping or fix tests if that is an issue, just tell me.

PR #17186, opened 2017-12-13
Misc. fixes from lgtm.com

esbena:
## Motivation

I looked at the [alerts for this project on lgtm.com](https://lgtm.com/projects/g/facebook/react-native/alerts/?mode=list), and fixed the ones I felt confident in. 

Each commit fixes an alert of one kind, let me know if you want a different structure.

## Test Plan

The changes are localized, and should be verifiable by a quick look.

`npm test` runs successfully after the changes.

## Release Notes

The changes are all very minor and spread out, I do not think they require a change note. Let me know if you think otherwise.

----

(Full disclosure: I work on lgtm's JavaScript analysis.)


cpojer:
Thanks for your PR. It seems like for some reason we were never able to check this one out. It seems like most of these were fixed at this point so I'm going to close this.

PR #17193, opened 2017-12-14
[Circle] Switch to yarn

hramos:
Switch to Yarn in order to more closely match internal Facebook environment.


hramos:
On CI, `test-android` is failing:

```
error jest@22.1.2: The engine "node" is incompatible with this module. Expected version ">= 6".
error Found incompatible module
```

PR #17203, opened 2017-12-14
Set minHeight instead of minWidth for calculating the layout in RCTSurfaceRootShadowView

maicki:
## Motivation

Fix calculating layout in `RCTSurfaceRootShadowView` as the `minWidth` is set doubled in `calculateLayoutWithMinimumSize:maximumSize:`.

cc @shergin


maicki:
@shergin Hey - could you take a look at this PR. If this would go into a public release we would have to manually patch it in this release if we would like to use it. Thanks!

shergin:
@maicki Oh, sorry, I missed this. Sure!

shergin:
@maicki Just curious, where do you use this? 😄

PR #17212, opened 2017-12-14
Switch to `declare module.exports` syntax

samwgoldman:
## Motivation

In [this commit](https://github.com/facebook/flow/commit/1c91714e54893ab610ab02c45220e73376df90f4), Flow added a new syntax to define a CommonJS modules module.exports type. Instead of writing `declare var exports: T`, you can write `declare module.exports: T`.

This new syntax was released in Flow v0.25, nearly 2 years ago. Our intention was to eventually remove the special-case-y support for declare var exports within a few versions, but late is better than never.

## Test Plan

`flow` -- there were some errors before this change, but the errors are unchanged.

## Related PRs

## Release Notes

 [INTERNAL] [MINOR] [Flow] - Switch to `declare module.exports` syntax

### DIFF
--- a/Libraries/react-native/react-native-interface.js
+++ b/Libraries/react-native/react-native-interface.js
@@ -24,5 +24,5 @@ declare var Headers: any;
 declare var Request: any;
 declare var Response: any;
 declare module requestAnimationFrame {
-  declare var exports: (callback: any) => any;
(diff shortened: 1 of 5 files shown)


samwgoldman:
landed in 5f8d8e90c2c43127b8a5d2fc5d18f16185c7a67e

Issue #17209, opened 2017-12-14
[Android] Font Scale not updating correctly


steveccable:
### Is this a bug report?

Yes, this is a bug report.  The bug affects only Android, with iOS behaving as expected.
I am also aware that this may in fact be two distinct bugs.  If it would be better for me to make two separate issues to handle them, I can do that, but for now am putting them up as one together.
PR for my take at how to fix the first issue can be found at https://github.com/facebook/react-native/pull/17227

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

I have read the contributing guidelines.

### Steps to Reproduce

1. Open up an app with some text in it, and attempt to call `PixelRatio.getFontScale()`
2. Go to Settings > Accessibility > Font Size and change the font size
3. Return to the app and trigger another call to `PixelRatio.getFontScale()`

### Expected Behavior

The call to `PixelRatio.getFontScale()` should return an updated value for the new font scaling factor.
In addition, the text should actually scale to the appropriate font size specified in the settings app.

### Actual Behavior

The call to `PixelRatio.getFontScale()` does not return an updated value until the Javascript has been reloaded (which can be accomplished by killing the app).
Furthermore, the text does not actually scale until you change the font size again, and then it lags behind by one update each time.  (see video posted at the bottom)

### Reproducible Demo

Note that the following code snippet will NOT work in Expo, because Expo reloads the javascript on switching in and out of the app.  This code can be rendered in any standalone application to reproduce the issue, however.

```
import React from 'react';
import { PixelRatio, Text, View } from 'react-native';

class App extends React.Component {

  constructor(props) {
    super(props);
    this.fontScale = PixelRatio.getFontScale();
    console.warn('initial scale of', this.fontScale);
  }

  shouldComponentUpdate() {
    if (this.fontScale !== PixelRatio.getFontScale()) {
      this.fontScale = PixelRatio.getFontScale();
      console.warn('adjusting scale to', this.fontScale);
      return true;
    }
    return false;
  }

  render() {
    return (
      <View style={{flex: 1}}>
        <Text>This is a font scale tester</Text>
        <Text onPress={() => console.warn('focring update...') || this.forceUpdate()}>The font scale is {PixelRatio.getFontScale()}</Text>
      </View>
    );
  }
}

export default App;
```

The code above produces the following result in the application.  Note that the font scale being printed out in the console warning never changes, and the font adjustment is lagging behind by one change each time I return to the application.

[...]

PR #17227, opened 2017-12-15
[ANDROID] Fix DeviceInfoModule LifecycleEventListener

steveccable:
## Motivation

Fixes one of the two parts of [issue posted here](https://github.com/facebook/react-native/issues/17209).  Motivation is that we wanted our app to be able to handle user changing font size while the app is running.  Yes, we know that changing font size doesn't inherently trigger a re-render, but at the minimum it should cause `PixelRatio.getFontScale()` to return an updated value.

## Test Plan

1. Render an interface and invoke `PixelRatio.getFontScale()`
2. Change the font scale in Settings > Accessibility > Font Size
3. Return to the React Native app (without reloading Javascript) and invoke `PixelRatio.getFontScale()` again.
4. Verify that the result of `PixelRatio.getFontScale()` has changed to reflect the new font size.

For a video of the problem, see the linked issue under the Motivation heading.  In this fixed version, the number actually does update as expected.

## Related PRs

None.  This should require no documentation change, as it is behavior that the docs seem to indicate should be happening already.
In fact, [the documentation at the bottom of this page](https://facebook.github.io/react-native/docs/native-modules-android.html) appears to already indicate that this should be done for anything implementing LifecycleEventListener.

## Release Notes

[ANDROID] [BUGFIX] [DeviceInfo] - Fix the DeviceInfoModule to properly respond to LifecycleEvents

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/deviceinfo/DeviceInfoModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/deviceinfo/DeviceInfoModule.java
@@ -39,6 +39,7 @@ public class DeviceInfoModule extends BaseJavaModule implements
   public DeviceInfoModule(ReactApplicationContext reactContext) {
     this((Context) reactContext);
     mReactApplicationContext = reactContext;
+    mReactApplicationContext.addLifecycleEventListener(this);
   }
 
   public DeviceInfoModule(Context context) {


steveccable:
Also fwiw I have no idea why circleci is failing, especially on the js and objc tests, since this PR makes effectively a 1-line change to Java code used only on Android...

steveccable:
@hramos Any chance you could have a look at this?  Alternatively, if you're not the right person to tag, do you know who would be?

asowayan:
@joshyhargreaves Any chance you could have a look at this?

asowayan:
@janicduplessis any chance you can review this PR?

ReactAndroid/src/main/java/com/facebook/react/modules/deviceinfo/DeviceInfoModule.java
```diff
@@ -9,11 +9,6 @@
 
 package com.facebook.react.modules.deviceinfo;
 
-import javax.annotation.Nullable;
```
joshyhargreaves: Unless there's any styling rule requirement, moving of the imports is most likely unnecessary.

steveccable: Sure, that was Android Studio moving them around automatically but i can set them back.

steveccable:
@joshyhargreaves Cleaned up the imports, now it should be just the one line of actual fix.

brentvatne:
lgtm

grabbou:
@facebook-github-bot shipit

PR #17229, opened 2017-12-15
Feature/android html text rendering

bartolkaruza:
## Motivation

This change adds html text rendering for a subset of html tags to Android `<Text>` components, by using the native built in [Html.fromHtml()](https://developer.android.com/reference/android/text/Html.html#fromHtml) functionality. This is currently the only way that sub/super script tags can be handled properly. If no HTML is present in the text inside the `<Text>` component, the string is displayed as is.

With this change the below string would be rendered without the HTML tags and with enhanced styling:
```
<a href="url">my link</a>
<sub>subscript</sub>
<sup>superscript</sup>

```
For a full list of supported tags check [this post](https://commonsware.com/blog/Android/2010/05/26/html-tags-supported-by-textview.html)

I'm aware that there are other libraries out there for rendering parts of the HTML spec, but none handle the particular cases of subscript and superscript very well. Additionally, parsing the HTML at the spot of this PR, improves performance as no extra views beyond the RawText are needed. The cost is a small parsing overhead, even when no HTML is present in the string.

We previously made this functionality with custom subclasses of the Text view manager and shadow nodes, but they were breaking with each React Native release and because of the many private inner classes and private fields, large parts of the RN code had to be duplicated. With this PR we hope to relieve ourselves of that maintenance challenge and help the community with a valuable new feature.

Moving forward, we could make this feature for iOS almost as easily as on Android:

```
NSMutableAttributedString *attributedString = [[NSMutableAttributedString alloc] initWithData:[self.text dataUsingEncoding:NSUTF8StringEncoding]
                                                                 options:@{NSDocumentTypeDocumentAttribute: NSHTMLTextDocumentType,
                                                                           NSCharacterEncodingDocumentAttribute: @(NSUTF8StringEncoding)}
                                                      documentAttributes:nil error:nil];
```

But I wanted to gage what the chances are for landing this into RN before spending the time adding the iOS implementation to this PR. 

## Test Plan
I will provide a test project for this change if the outlook of landing the change is good.
```
<Text>{'<sup>Superscript</sup><sub>subscript</sub><b>bold</b><i>italic</i> etc...'}</Text>
``` 

## Release Notes
[ ANDROID  ] [ FEATURE] [Text] - HTML text handling in the Text component


bartolkaruza:
I agree @cpojer . This was a long shot when all other routes seemed closed to implement a subset of html tags by other means. Thanks for getting around to it.

Issue #16989, opened 2017-11-27
[iOS] TextInput clear button is not accessible


haitaoli:
### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment

Environment:
  OS: macOS High Sierra 10.13.1
  Node: 6.12.0
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.7.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 3.0 AI-171.4408382

Packages: (wanted => installed)
  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: ^0.49.5 => 0.49.5

Target Platform: iOS (9)

### Steps to Reproduce
https://snack.expo.io/S1EkFy5gG

1. Input any text for the clear button to appear.
2. Turn on Voice Over
3. Try to select the clear button.

### Expected Behavior
"Clear" button can be selected and activated. iOS native text input control supports it.

### Actual Behavior
Voice Over cannot focus on the "clear" button.

### Reproducible Demo
https://snack.expo.io/S1EkFy5gG

sasasa42:
Similarly, I am trying to automate using Appium to locate the clear button for my app, but I am not able to inspect it individually. I am not getting the control for it.

sahrens:
Thanks for the clear bug - this is still an issue.

Looks like this is because we wrap the `RCTTextInputView` component in a `TouchableWithoutFeedback`. As a workaround, you can use `RCTTextInputView` directly (e.g. `requireNativeComponent('RCTSinglelineTextInputView');`)

This is part of a more general issue with nested accessibility support that will be tricky to solve: https://github.com/facebook/react-native/issues/24515

PR #17230, opened 2017-12-15
Make TextInput "clear text" button accessible

haitaoli:
## Motivation
This PR makes TextInput "clear text" button accessible. UITextField is wrapped inside a RCTUITextField control which has isAccessibleElement set to TRUE. This prevents iOS accessibility interface from enumerating child elements. This change makes the RCTextField not accessible, but allows child elements to be enumerated if the `accessible` prop is set to true.

## Issue
https://github.com/facebook/react-native/issues/16989

## Test Plan
Manually tested with RNTester sample app and accessibility explorer.

## Release Notes
[IOS] [BUGFIX] [TextInput] - Made "clear text" button accessible

### DIFF
--- a/Libraries/Text/TextInput/Singleline/RCTUITextField.m
+++ b/Libraries/Text/TextInput/Singleline/RCTUITextField.m
@@ -87,6 +87,22 @@
   self.enabled = editable;
 }
 
+#pragma mark - Accessibility
+
+- (BOOL)isAccessibilityElement
+{
+  // Always return NO to enumerate child elements. Otherwise the "clear text" button is hidden
+  // from accessibility interface.
+  return NO;
+}
+
+- (BOOL)accessibilityElementsHidden
+{
+  // If "accessible" prop is set to false on TextInput, disable all accessible children.
+  // Without this function the "clear text" UIButton is still exposed.
+  return ![super isAccessibilityElement];
+}
+
 #pragma mark - Caret Manipulation
 
 - (CGRect)caretRectForPosition:(UITextPosition *)position


shergin:
How does this affect the accessibility of UITextView's content?

haitaoli:
The UITextView is exposed to accessibility interface as a child element. `accessibilityLabel` and `accessibilityValue` are not affected on `TextInput` in my tests.

shergin:
@haitaoli Ooops, I meant UITextField.
So, I actually I have two Q:
 * Should we apply same changes for RCTUITextView? Why?
 * Have you tested this? Does it work exactly like it worked especially for text content?

shergin:
Thank you for tackling this, BTW! This is really valuable!

haitaoli:
Is UITextView used when TextInput is multi-line? UITextView doesn't have a clear button so it doesn't need this fix. Here are the screenshots of accessibility explorer output on the TextInput component and it's "clear text" button:

screen shot 2017-12-18 at 3 37 38 pm
screen shot 2017-12-18 at 3 37 50 pm

haitaoli:
I was able to set `accessibilityLabel` on TextInput and it shows up in accessibility explorer correctly.

shergin:
@haitaoli Well, actually RCTUITextField does not wrap it, it subclasses UITextField. So, the text input should behave very well in terms of accessibility. And I am afraid that this changes can interfere with built-in logic. 🤔

Why do you/we need to make "clear" button accessible in the first place?

shergin:
@haitaoli So, how does the current default behaviour of <TextInput> different from default UITextView/UITextField? If it's not, we have to be really really careful because this component is really really important and must be perfect from an accessibility perspective. And it must be not RN specific issue, so we have to try to find how other products address it.

I am not an accessible expert, so I don't know for sure, but... do you people with disabilities really need this button? Maybe there is a special gesture for cleaning the text input so they simply don't need a button?

haitaoli:
@shergin the current behavior of RCTUITextField overrides `isAccessibilityElement` (via one of the base classes) to return a value based on `accessible` prop, which disables enumerating child elements. This is the correct behavior for most elements that don't have accessible child elements. But TextInput is different. This is RN specific. I believe this button is important. Without it the user can only use the "backspace" key, and it takes 2 taps to remove one character. Imagine doing this while you cannot see the screen.

shergin:
Yes, I tested on a real device and in standard apps clear button is accessible. So, if in RN it does not by default, we have to fix it. But first of all, we have to understand why it is happening now.
@haitaoli DId you tried to override `setIsAccessibilityElement` method and test which component calls it?

haitaoli:
I didn't. But it should be called on any component with `accessible` prop set to true.

janicduplessis:
Seems like it is marked accessible because of the native TextInput component is wrapped with TouchableWithoutFeedback (https://github.com/facebook/react-native/blob/2cf50e8fff0dc3621092cef4ce9ddffcb1a20903/Libraries/Components/TextInput/TextInput.js#L825) which is accessible by default (https://github.com/facebook/react-native/blob/2cf50e8fff0dc3621092cef4ce9ddffcb1a20903/Libraries/Components/Touchable/TouchableWithoutFeedback.js#L188). Maybe this could just be fixed in js land?

cc @shergin

haitaoli:
I don't think there is a workaround on JS side. The native view cannot be set to be accessible without hiding that clear button.

cpojer:
Seems like this is an RN in open source issue, and @janicduplessis is right that we can probably fix this with a JS only fix. I'm gonna close this PR for now but happy to reopen if I'm wrong.

PR #14675, opened 2017-06-22
Allow app to init an OkHttpClientProvider.IProvider with custom settings(FrescoModule and NetworkingModule)

leon087:
## Motivation 

This allows app to init an OkHttpClientProvider.IProvider with custom settings(FrescoModule and NetworkingModule).

PR #14068, opened 2017-05-19
Inject configured OkHttpClient to networking module using MainPackageConfig

thotegowda:
## Motivation 
The main purpose of this PR is to enable brownfield applications to inject existing`OkHttpClient` instance to react-native networking sub system.  Most native applications would have their `OkHttpClient` instance configured with headers, interceptors, logging, fallback handling, error handling, etc. It doesn't make much sense to duplicate all of this to use `fetch` in javascript. Currently the only way you could use existing `OkHttpClient` is by custom-bridge to native, but this will stops app from using `fetch` interface in JS. 

With this commit, App can pass `OkHttpClient` instance to networking module using `MainPackageConfig`, while creating `MainReactPackage`. 

## Test Plan
Unit tests.

PR #17237, opened 2017-12-16
Add back ability to customise OkHttp client

cdlewis:
## Motivation

Prior to 0a71f48b1349ed09bcb6e76ba9ff8eb388518b15, users could customise the OkHttp client used by React Native on Android by calling replaceOkHttpClient in OkHttpClientProvider.

This functionality has a variety of legitimate applications from changing connection timeouts or pool size to Stetho integration. The challenge is to add back support for replacing the client without causing a breaking change or reintroducing the problems @olegbl sought to address in his original commit.

Introducing a client factory archives these aims, it adds a new, backwards compatible interface and is called each time a client is requested rather than re-using the same instance (unless you explicitly want this behaviour, in which case you could replicate it using a static class property inside your custom factory). 

A number of PRs have been opened to add this functionality: https://github.com/facebook/react-native/pull/14675, https://github.com/facebook/react-native/pull/14068.

I don't have a lot of Java experience so I'm open to better/more idiomatic ways to achieve this :)

## Test Plan

Create React Native application and set a custom factory in the constructor, e.g.  `OkHttpClientProvider.setOkHttpClientFactory(new CustomNetworkModule());`

Where a custom factory would look like:

```
class CustomNetworkModule implements OkHttpClientFactory {
    public OkHttpClient createNewNetworkModuleClient() {
        return new OkHttpClient.Builder().build();
    }
}
```

## Related PRs

Remove the existing replace client method to prevent accident use and alert existing users that its functionality has changed: https://github.com/facebook/react-native/pull/16972

## Release Notes

[Android] [Minor] [Networking] - | Provide interface for customising the OkHttp client used by React Native |

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientFactory.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientFactory.java
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+package com.facebook.react.modules.network;
+
+import okhttp3.OkHttpClient;
+
+public interface OkHttpClientFactory {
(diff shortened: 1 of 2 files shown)


cdlewis:
@hramos and @mkonicek have been reviewers on similar PRs.

PR #17243, opened 2017-12-16
Update includes of folly shim headers

yfeldblum:
## Motivation

Some headers in folly have moved. Relevant to react-native, `folly/Bits.h` has moved to `folly/lang/Bits.h`. A temporary shim has been left at `folly/Bits.h`, but it is slated for removal.

Also updates folly dep to `v2017.12.11.00`.

## Test Plan

CI.

## Related PRs

N/A.


cpojer:
Thank you for your PR. Unfortunately we didn't get around to looking at this when you originally opened this PR and now it seems like this is really outdated so I'm going to close it.

Issue #17248, opened 2017-12-17
Changing TextInput keyboardType from visible-password does not unset the underlying flags (Android)


BrandonWilliamsCS:
### Is this a bug report?

Yes.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes.

### Environment

Environment:
  OS: macOS High Sierra 10.13.2
  Node: 9.2.0
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: facebook/react-native#master => 1000.0.0
(react-native built from master branch on 2017-12-17)

Target Platform: Android 8.x
Tested specifically on Nexus 5x, api level 27

### Steps to Reproduce

1. Create a TextInput with `secureTextEntry` set and a `keyboardType` of `null`, `undefined`, or `'default'`
2. Change the `keyboardType` to `'visible-password'`
3. Change the `keyboardType` back to `'default'`

### Expected Behavior

The field would initially be an obscured password field, then become unobscured, then become obscured again when no longer of type `'visible-password'`.
Additionally, the soft keyboard variant stays in "password" mode (e.g., no suggestions) in each case due to `secureTextEntry`.

### Actual Behavior

After set to `'visible-password'`, the field remains unobscured even after switching to a different type.
However, when not obscured, the soft keyboard falls returns to a plain text keyboard with suggestions instead of numbers on the top row (may vary by device).

### Reproducible Demo

I believe this branch should exhibit the bug out-of-the-box (after installing packages and building):
https://github.com/BrandonWilliamsCS/react-native-test-harness/tree/fix/android-textfield-keyboardtype-flags

Once running on an Android device or emulator, tap the field so that the soft keyboard appears and enter text into the field. This will trigger a state change that toggles between `'visible-password'` and `'default'`.

hanspoo:
Same problem here with: 

    "react": "16.3.0-alpha.1",
    "react-native": "0.54.0"

The use of a keyboardType email-address disables secureTextEntry.

PR #17249, opened 2017-12-17
Unset phone and visible-password keyboardType flags on Android

BrandonWilliamsCS:
## Motivation

Attempted fix for issue #17248

## Test Plan

Added unit tests that cover the affected function and manually repeated the reproduction steps found in the issue.

Note: I had to apply the attached patch file to actually *run* the tests because they were not enabled. I didn't include this in the PR because it seemed like a secondary problem with possible ramifications (see comment near patched line) beyond this issue. For example, other, unrelated tests break when that line is uncommented.
[textInputTestEnable.patch.txt](https://github.com/facebook/react-native/files/1565807/textInputTestEnable.patch.txt)

If I should apply the patch to this PR or re-enable the tests in some other fashion, please let me know and I can do so.

## Related PRs

(none)

## Release Notes

[ANDROID] [BUGFIX] [TextInput] - Fix an issue when swapping to and from the `'visible-password'` or `'phone-pad'` keyboard types.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -73,6 +73,12 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
   private static final int INPUT_TYPE_KEYBOARD_NUMBERED =
       InputType.TYPE_CLASS_NUMBER | InputType.TYPE_NUMBER_FLAG_DECIMAL |
           InputType.TYPE_NUMBER_FLAG_SIGNED;
+  private static final int PASSWORD_VISIBILITY_FLAG = InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD &
+        ~InputType.TYPE_TEXT_VARIATION_PASSWORD;
+  private static final int KEYBOARD_TYPE_FLAGS = INPUT_TYPE_KEYBOARD_NUMBERED |
+            InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS |
+            InputType.TYPE_CLASS_TEXT | InputType.TYPE_CLASS_PHONE |
+            PASSWORD_VISIBILITY_FLAG;
 
   private static final String KEYBOARD_TYPE_EMAIL_ADDRESS = "email-address";
   private static final String KEYBOARD_TYPE_NUMERIC = "numeric";
@@ -527,13 +533,6 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
 
   @ReactProp(name = "keyboardType")
   public void setKeyboardType(ReactEditText view, @Nullable String keyboardType) {
-    int passwordVisibilityFlag = InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD &
-        ~InputType.TYPE_TEXT_VARIATION_PASSWORD;
-
-    int keyboardTypeFlags = INPUT_TYPE_KEYBOARD_NUMBERED |
-            InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS |
-            InputType.TYPE_CLASS_TEXT | InputType.TYPE_CLASS_PHONE |
-            passwordVisibilityFlag;
     int flagsToSet = InputType.TYPE_CLASS_TEXT;
     if (KEYBOARD_TYPE_NUMERIC.equalsIgnoreCase(keyboardType)) {
       flagsToSet = INPUT_TYPE_KEYBOARD_NUMBERED;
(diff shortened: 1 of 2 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -527,6 +527,13 @@ public void setAutoCapitalize(ReactEditText view, int autoCapitalize) {
 
   @ReactProp(name = "keyboardType")
   public void setKeyboardType(ReactEditText view, @Nullable String keyboardType) {
+    int passwordVisibilityFlag = InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD &
```
janicduplessis: So this flag is only the visibility bit if I understand correctly? It won't unset the fact that it is a password or not.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -527,6 +527,13 @@ public void setAutoCapitalize(ReactEditText view, int autoCapitalize) {
...
+        ~InputType.TYPE_TEXT_VARIATION_PASSWORD;
+
+    int keyboardTypeFlags = INPUT_TYPE_KEYBOARD_NUMBERED |
+            InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS |
+            InputType.TYPE_CLASS_TEXT | InputType.TYPE_CLASS_PHONE |
+            passwordVisibilityFlag;
     int flagsToSet = InputType.TYPE_CLASS_TEXT;
     if (KEYBOARD_TYPE_NUMERIC.equalsIgnoreCase(keyboardType)) {
       flagsToSet = INPUT_TYPE_KEYBOARD_NUMBERED;
```
janicduplessis: A bit later in this code we set the TYPE_TEXT_VARIATION_VISIBLE_PASSWORD flag, should we just set passwordVisibilityFlag instead?

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -527,6 +527,13 @@ public void setAutoCapitalize(ReactEditText view, int autoCapitalize) {
 
   @ReactProp(name = "keyboardType")
   public void setKeyboardType(ReactEditText view, @Nullable String keyboardType) {
+    int passwordVisibilityFlag = InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD &
+        ~InputType.TYPE_TEXT_VARIATION_PASSWORD;
+
+    int keyboardTypeFlags = INPUT_TYPE_KEYBOARD_NUMBERED |
```
janicduplessis: Let's move this and passwordVisibilityFlag to private static final members in the class.

janicduplessis:
Thanks for figuring this out, this flag code is pretty hard to understand -_-

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
BrandonWilliamsCS: That is correct. We wouldn't want to set/unset general password status in `keyboardType`, only in `secureTextEntry`.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
BrandonWilliamsCS: Will do, commit incoming.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
BrandonWilliamsCS: That's a good question. I think it comes down to how the field should behave when `keyboardType="visible-password"` but `secureTextEntry="false"`. Due to the way Android handles the flags, if we set the flag I called `passwordVisibilityFlag` without setting the `InputType.TYPE_TEXT_VARIATION_PASSWORD` flag, it will actually treat it like `InputType.TYPE_TEXT_VARIATION_URI`.

So three options I see:
1. Somehow forbid `keyboardType="visible-password"` with `secureTextEntry="false"`
2. Let `keyboardType="visible-password"` override the `secureTextEntry` prop (equivalent to the current PR code)
3. Allow the consumer to set the props inconsistently (probably with a warning) and risk an unexpected `InputType.TYPE_TEXT_VARIATION_URI`

I should probably leave that decision up to the core maintainers.

janicduplessis: Makes sense, let's keep the current behavior (current PR code).

BrandonWilliamsCS:
Glad I could help!
I moved the flag values to private static finals and added a comment noting the quirk, so that future contributions can have a little more context into the mess that is coordinating InputType flags between declarative and imperative contexts. Let me know if there's more I can do here.

BrandonWilliamsCS:
Pinging @janicduplessis just to make sure my update didn't get lost in the shuffle of the holidays.

janicduplessis:
Thanks @BrandonWilliamsCS, will merge after holidays!

BrandonWilliamsCS:
@janicduplessis I don't mean to be a pest, but is this still slated for merge?

janicduplessis:
Thanks for reminding me, yes!

@facebook-github-bot shipit

colinux:
This PR seems to not have been merged.

I'm facing the same issue on RN 64, which should have been fixed by this PR, and I don't see this code in master. Or did I miss something ?

hikaaam:
this issue still happen react-native v.0.63, keyboardType visible-password make my password visible

PR #17262, opened 2017-12-18
SectionList renderItem should be optional

shichongrui:
## Motivation

`renderItem` on `SectionList` is within the `OptionalProps` group of props but it is not actually marked as optional. Which means that doing things such as in the example where each section has its own `renderItem` and no `renderItem` prop is passed into `SectionList` will fail flow.

## Test Plan

Create a `SectionList` where each section has it's own `renderItem` and do not pass in a `renderItem` into `SectionList`. Run flow, it should error.

## Release Notes
[GENERAL] [MINOR] [SectionList] -Makes `renderItem` prop on `SectionList` optional for flow.


janicduplessis:
Thanks for the fix!

@facebook-github-bot shipit

Issue #12112, opened 2017-01-29
react-native-git-upgrade changes nothing


joIivier:
### Description

I am trying to upgrade from react native 35 to 40. Running 'react-native-git-upgrade' in my project tells me the upgrade has been done but nothing has changed in my repository.

Output:

```
/usr/local/Cellar/node/7.4.0/bin/react-native-git-upgrade 
git-upgrade info Check for updates
git-upgrade info Using yarn 0.19.1
git-upgrade info Read package.json files
git-upgrade info Check declared version
git-upgrade info Check matching versions
git-upgrade info Check React peer dependency
git-upgrade info Check that Git is installed
git-upgrade info Get information from NPM registry
git-upgrade info Upgrading to React Native 0.40.0, React ~15.4.0-rc.4
git-upgrade info Setup temporary working directory
git-upgrade info Configure Git environment
git-upgrade info Init Git repository
git-upgrade info Add all files to commit
git-upgrade info Commit current project sources
git-upgrade info Create a tag before updating sources
git-upgrade info Generate old version template
git-upgrade info Add updated files to commit
git-upgrade info Commit old version template
git-upgrade info Install the new version
git-upgrade info Generate new version template
git-upgrade info Add updated files to commit
warning: CRLF will be replaced by LF in android/gradlew.bat.


The file will have its original line endings in your working directory.
git-upgrade info Commit new version template
git-upgrade info Generate the patch between the 2 versions
git-upgrade info Save the patch in temp directory
git-upgrade info Reset the 2 temporary commits
git-upgrade info Apply the patch
error: patch failed: .flowconfig:1
Falling back to three-way merge...
Applied patch to '.flowconfig' with conflicts.
error: patch failed: .gitignore:22
Falling back to three-way merge...
Applied patch to '.gitignore' with conflicts.
error: android/app/src/main/java/com/appname/MainApplication.java: does not exist in index
error: cannot apply binary patch to 'android/gradle/wrapper/gradle-wrapper.jar' without full index line
Falling back to three-way merge...
error: cannot apply binary patch to 'android/gradle/wrapper/gradle-wrapper.jar' without full index line
error: android/gradle/wrapper/gradle-wrapper.jar: patch does not apply
error: patch failed: ios/appname.xcodeproj/project.pbxproj:22
Falling back to three-way merge...
Applied patch to 'ios/appname.xcodeproj/project.pbxproj' with conflicts.
[...]

PR #17266, opened 2017-12-18
Fix issue #12112: patch with --reject

enthal:
FIX https://github.com/facebook/react-native/issues/12112

Per https://github.com/facebook/react-native/issues/12112#issuecomment-292619346 :

> As others have identified, the upgrade process creates a patch file, then tries to apply the patch. When it tries to apply a --3way patch, it will fail if it encounters any errors such as those above ☝️. Unfortunately, it reverses the whole patch when a 3-way patch fails to apply.
> 
> If you change the git strategy for applying the patch from --3way to --reject, it will allow your changes to go through, then you just have to resolve and remove the .rej files where it had errored out. You can either do this after it fails using the patch file @ajwhite found, or change the cliEntry.js directly as described in the PR description.

## Motivation

This has been an open issue since 2017-01-29 with a well-defined solution (see issue).

### DIFF
--- a/react-native-git-upgrade/cliEntry.js
+++ b/react-native-git-upgrade/cliEntry.js
@@ -334,7 +334,7 @@ async function run(requestedVersion, cliArgs) {
 
     try {
       log.info('Apply the patch');
-      await exec(`git apply --3way ${patchPath}`, true);
+      await exec(`git apply --reject ${patchPath}`, true);
     } catch (err) {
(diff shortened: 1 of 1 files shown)


hramos:
Thanks for sending this pull request! We would love to review your changes however it looks like you haven't signed the CLA yet. You can do so at https://code.facebook.com/cla.

PR #17268, opened 2017-12-19
Add quotes back to script path in xcode bundler script

jonathonlui:
Building "Release" configuration using xcode fails if the project path has spaces because the xcode bundler script doesn't quote the path to the cli script.

This issue was previously fixed by https://github.com/facebook/react-native/pull/15608 but reintroduced by https://github.com/facebook/react-native/pull/15317.


lindskogen:
:+1: I encountered this issue too!

LcTwisk:
I've got the same issue!

jonathonlui:
Closing because fixed in https://github.com/facebook/react-native/commit/8aa568e867bbbe7e23ded3651f23581ff2753323

PR #17284, opened 2017-12-20
Xcode 9 supports running multiple simulators

koenpunt:
## Motivation

Since Xcode 9 you can run multiple simultaneously. And since I believe React Native advocates using the latest version of Xcode, we can safely remove this constraint.

## Test Plan

Updated the unit tests. Furthermore it can be found in the [Xcode release notes](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/WhatsNewXcode/xcode_9/xcode_9.html#//apple_ref/doc/uid/TP40004626-CH8-SW12) that multiple simulators are now supported.

This can be tested with the CLI by running `react-native run-ios` twice, but with a different `--simulator` flag, e.g.;

    react-native run-ios --simulator "iPhone SE"
    react-native run-ios --simulator "iPhone X"

## Release Notes

[IOS] [ENHANCEMENT] [local-cli/runIOS/findMatchingSimulator.js] - Allow running multiple simulators

### DIFF
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -115,7 +115,15 @@ function runOnSimulator(xcodeProject, args, scheme) {
 
     const simulatorFullName = formattedDeviceName(selectedSimulator);
     console.log(`Launching ${simulatorFullName}...`);
-    child_process.spawnSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
+    try {
+      child_process.spawnSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
+    } catch (e) {
+      throw new Error(
+        `Could not boot ${args.simulator} simulator. Is there already a simulator running? ` +
+        'Running multiple simulators is only supported from Xcode 9 and up. ' +
+        'Try closing the simulator or run the command again without specifying a simulator.'
+      );
+    }
 
     buildProject(xcodeProject, selectedSimulator.udid, scheme, args.configuration, args.packager, args.verbose)
       .then((appName) => resolve(selectedSimulator.udid, appName));
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -113,16 +113,18 @@ function runOnSimulator(xcodeProject, args, scheme) {
       throw new Error(`Could not find ${args.simulator} simulator`);
     }
 
-    const simulatorFullName = formattedDeviceName(selectedSimulator);
-    console.log(`Launching ${simulatorFullName}...`);
-    try {
-      child_process.spawnSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
-    } catch (e) {
-      throw new Error(
-        `Could not boot ${args.simulator} simulator. Is there already a simulator running? ` +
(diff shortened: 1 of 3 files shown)


koenpunt:
Ping ✨ @hramos almost all green now! (failing tests are not related to my changes I assume)

local-cli/runIOS/runIOS.js
```diff
@@ -107,22 +107,20 @@ function runOnSimulator(xcodeProject, args, scheme) {
...
-      // but we want it to only launch the simulator
-    }
+    child_process.spawnSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
     resolve(selectedSimulator.udid);
   })
-  .then((udid) => buildProject(xcodeProject, udid, scheme, args.configuration, args.packager, args.verbose))
-  .then((appName) => {
+  .then((udid) => new Promise((resolve) => {
+    buildProject(xcodeProject, udid, scheme, args.configuration, args.packager, args.verbose)
+      .then((appName) => resolve(udid, appName));
```
ide: this resolve call looks buggy

koenpunt: I kind of agree, wasn't really happy with I either, but couldn't really think of a better way.. So I'm open to suggestions :)

koenpunt: Oh now that I look at it again I think I can simply move the build project call into the first promise. Let me try

koenpunt: I've moved the invocation of `buildProject` into the first promise. But now I wonder why it wasn't there in the first place..

koenpunt: Another option would be to add the `udid` to the resolve call in `buildProject`, but that doesn't feel right either.

koenpunt:
> However, many people that I've met are still on an older version.

Well as far as I know there shouldn't be a problem if they do not give a `--simulator` flag to the `run-ios` command, since then it will just default to the `booted` one. 

Altho I'm not completely sure what happens if they do specify one when there's another booted already, and that's hard to test since I'm no longer having Xcode 8.

But again, I think we should move forward, Xcode 9 has been around for over 3 months, this might be another good reason to upgrade.

koenpunt:
Oh but this I can do:

> can we log a message in case of run-ios failing that it was expected on Xcode < 9?

ide:
@shergin Does Facebook use Xcode 8 anymore? That'd help inform whether to keep legacy support around.

shergin:
@ide Facebook does not use `run-ios` at all, afaik. And, honestly, I have no idea how it works.
So, I think, as soon OSS core community (@grabbou, @janicduplessis, someone else) can reach a consensus about this change, I can land it. (This is usually also true for many PRs here.)

koenpunt:
Ping ✨

koenpunt:
@grabbou FYI; just brought my branch up-to-date with master

hramos:
Can you take a look at the failing tests? We recently got all tests passing. If you're sure the failure is unrelated, at least rebase the PR. That will trigger tests to run again.

koenpunt:
@hramos tests are passing now. **js_checks** isn't, but I'm pretty sure that's unrelated.

yosimasu:
@koenpunt 

The default value of --simulator should be null, or the booted device never be used.

PR #17290, opened 2017-12-20
Fix ReactLegacy and delete RCTViewControllerProtocol

javache:
It's always bothered that we have this protocol and I figured it could just be merged with RCTWrapperViewController.

Test plan:

Build and Run both RNTester and RNTesterLegacy


mmmulani:
why wasn't this just a diff 😅

javache:
I was lazy and didn't want to get on VPN..

PR #17295, opened 2017-12-20
[Circle CI] Fix buck failures on master

hramos:
WIP.


hramos:
Not 100% there yet but landing to clear out some already-resolved errors on master.

PR #17305, opened 2017-12-21
Add top offset for react loading view on Kitkat

kmagiera:
## Motivation

This PR fixes the problem with dev loading view on KitKat and older Android devices after #16596

## Test Plan

Install RNTester app on Android API 19 or lower device. See green loading view show up under status bar. Do the same with full screen theme set and see it show up correctly at the top of the screen.
Verify the green loading bar displays correctly on devices with API > 19 too.

## Related PRs

This fixes an issue introduced in #16596

## Release Notes

[ANDROID][MINOR][DevSupport] - Fix green dev loading bar on Android Kitkat and below


ayc1:
nice. thanks for the quick fix!

PR #17234, opened 2017-12-15
[Dev] Update to Danger 2.0

orta:
## Motivation

Danger 2.0 is pretty stable now (Apollo-client just updated to 2.0, so I was reminded of RN) and more importantly for RN, it doesn't depends on jest's internals now. This means Danger doesn't need special casing like before. 

/cc @hramos @grabbou 

## Test Plan

I ran Danger PR to validate that any PR runs:

screen shot 2017-12-15 at 6 04 55 pm

I expect the CI will tell me whether it works or not

## Related PRs

https://github.com/facebook/react-native/pull/16350

PR #17310, opened 2017-12-21
Update to Danger 2.0

hramos:
See #17234 for discussion.


orta:
Awesome 👍 - was just setting up to do this. Thanks @hramos

PR #17370, opened 2017-12-27
SystemUiVisibility overwritten bug

rogerkerse:
Make StatusBar style respect previously set systemUiVisibility

## Motivation

I tried to set `SystemUiVisibility` flag to `SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR` to enable Android's light soft navigation bar, but when I had `<StatusBar />` component in my view, then it was always overwritten. This is how I found the bug and fixed it.

## Test Plan

1. In MainActivity you can set systemUiFlags like this in onCreate method
```
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            Window window = getWindow();
            View rootView = window.getDecorView().getRootView();
            rootView.setSystemUiVisibility(View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR);
        }
```
2. Add <StatusBar /> to a view
3. In `android/app/build.gradle` file set `targetSdkVersion` to **27** instead of 22 or something like that
4. In `android/app/src/main/res/values/styles.xml` add 2 lines to `AppTheme`
```
        <item name="android:windowDrawsSystemBarBackgrounds">true</item>
        <item name="android:navigationBarColor">#ffffff</item>
```
5. Run the app.
6. Test. Previously bottom soft navigation bar was set to white and buttons also to white (so they weren't visible anymore), because StatusBar was overwriting previously set systemUiVisibility Flags. Now the bar should be white and buttons dark as expected.

Previous buggy android bottom navbar
screen shot 2017-12-27 at 17 11 57

New fixed android bottom navbar
screen shot 2017-12-27 at 17 12 07

## Related PRs

This pull request does not change any existing feature. It only makes StatusBar coloring work more properly without affecting bottom navigation bar on android or any other system ui visibility feature.

## Release Notes

[ANDROID] [BUGFIX] [StatusBar] - Fixed StatusBar overwriting previously set SystemUiVisibility flags

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/statusbar/StatusBarModule.java
@@ -189,8 +189,13 @@ public class StatusBarModule extends ReactContextBaseJavaModule {
           @Override
           public void run() {
             View decorView = activity.getWindow().getDecorView();
-            decorView.setSystemUiVisibility(
-              "dark-content".equals(style) ? View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR : 0);
+            int systemUiVisibilityFlags = decorView.getSystemUiVisibility();
+            if ("dark-content".equals(style)) {
+              systemUiVisibilityFlags |= View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR;
+            } else {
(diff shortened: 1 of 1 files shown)


rogerkerse:
I checked the `ci/circleci: test-android` and the failure is not because of my code changes but instead some third party library missing. Can you still validate this specific pull request?

PR #17373, opened 2017-12-28
Add devDependencies support

lucianomlima:
Add devDependencies support to React Native templates.
Template can have a devDependencies.json file with devDependencies inside.

## Motivation

Allows React Native templates to have devDependencies, which can help applications to have better organization, quality and testability.

## Test Plan

Add a devDependencies.json file with at least one dependency (like eslint)

## Release Notes

[CLI] [FEATURE] [local-cli/generator/templates.js] - Add support to devDependencies for react native templates


lucianomlima:
@mkonicek

lucianomlima:
@hramos You can review this PR?

hramos:
@lucianomlima looks like some tests are failing, but they are green on master. The failures may be unrelated, and could be fixed by rebasing your PR on top of master.

hramos:
For CLI changes, @satya164 @grabbou might want to take a look.

lucianomlima:
Ok, I will do a rebase. I need to send again?

grabbou:
I am not really sure this is useful (also adds an extra file, which is in my opinion not a preferred way of doing things).

Would you mind elaborating a bit on the use case to help me understand better what this PR has to do with the CLI? Like a template you are using, the dependencies you need for that better quality etc.

Thank you

lucianomlima:
@grabbou It may be useful to start new React Native applications from a template that already comes with devDependencies included.
Example all packages needed for eslint, prettier, commitzen, reactotron and others.

Currently it is only possible to add non-dev dependencies.

grabbou:
I see, makes sense then.

Do you think it's possible to make a single function "installDependencies" and avoid having two separate functions with a similar body?

It looks to me that they just have a reference to a different file and one uses `dev` flag for adding dependencies.

lucianomlima:
Of Course! I'll do a rebase and these changes. Thanks

lucianomlima:
@grabbou What is the best approach? Add a dev flag (boolean) and call same method?

lucianomlima:
Rebase my branch causes this warnings. Maybe close this PR and send a new is better?

lucianomlima:
@grabbou I was testing the possibility of set devDependencies in a single file but this generate a error.
If someone tries to install a template with current RN version and dependencies.json has the following structure:
```json
{
    "react-navigation": "^ 1.3.0",
    "devDependencies": {
        "prettier": "^ 1.9.0"
    }
}
```
He finds an error because the current code goes through all the dependencies found in json file and will try to install a dependency that is referenced as an object.
The error will be: `Error: Command failed: yarn add devDependencies@[object Object]` resulting in installation failure.

Following my proposal with separate files to dependencies and devDependencies, it maintains compatibility with the current version.

lucianomlima:
@oguzbilgener @turnrye Rebase was asked and this messed up the whole PR. I'm closing this one and creating a new pull request #18164

Issue #12237, opened 2017-02-06
react-native-git-upgrade deletes my non-versioned files and fails


antoinerousseau:
### Description

I installed react-native-git-upgrade@0.2.5 and ran it on my project.

### Reproduction

	$ react-native-git-upgrade
	git-upgrade info Check for updates
	git-upgrade info Read package.json files
	git-upgrade info Check declared version
	git-upgrade info Check matching versions
	git-upgrade info Check React peer dependency
	git-upgrade info Check that Git is installed
	git-upgrade info Get information from NPM registry
	git-upgrade info Upgrading to React Native 0.41.2, React ~15.4.0
	git-upgrade info Setup temporary working directory
	git-upgrade info Configure Git environment
	git-upgrade info Init Git repository
	git-upgrade info Add all files to commit
	git-upgrade info Commit current project sources
	git-upgrade info Create a tag before updating sources
	git-upgrade info Generate old version template
	git-upgrade info Add updated files to commit
	git-upgrade info Commit old version template
	git-upgrade info Install the new version
	npm WARN deprecated node-uuid@1.4.7: use uuid module instead
	git-upgrade info Generate new version template
	git-upgrade info Add updated files to commit
	git-upgrade info Commit new version template
	git-upgrade info Generate the patch between the 2 versions
	git-upgrade info Save the patch in temp directory
	git-upgrade info Reset the 2 temporary commits
	git-upgrade info Apply the patch
	fatal: unrecognized input
	git-upgrade WARN The upgrade process succeeded but there might be conflicts to be resolved. See above for the list of files that have merge conflicts.
	git-upgrade info Upgrade done
	$

And in the end, no file is modified, and some files are missing, for example `android/app/google-services.json` (present in my `.gitignore`)

### Solution

No idea

### Additional Information

* React Native version: just installed 0.41.2, wanted to upgrade RN templates according to it
* Platform: n/a
* Operating System: macOS

Note: my Android/iOS package name is `net.tribeez` (my `package.json` name is `tribeez`)

ncuillery:
Ok I reproduce the deletion by running the Git commands manually, just like the tool does: https://gist.github.com/ncuillery/4b09e36d772601301cf0d0c94dcde692. I think I got it:

The first step of the process consists in restoring the template of your current RN version, just like it would have been generated by the init command. So your `.gitignore` file is replaced by this one https://github.com/facebook/react-native/blob/master/local-cli/templates/HelloWorld/_gitignore for the rest of the process.

[...]

PR #17393, opened 2017-12-29
Keep the ignored files during the upgrade process

ncuillery:
## Motivation

This PR fixes the issue #12237 relative to react-native-git-upgrade.

When the user adds new files to the .gitignore file, those files are deleted during the upgrade process because it starts by creating a fresh .gitignore file without user's modification, so the user's ignored files are no longer ignored and they are deleted during the upgrading process.

The best solution I've found to keep the user's ignored files... ignored, consists in appending the content of the .gitignore file in the `<TEMP DIR>/.git-rn/info/exclude` file.
The user's ignored files are now ignored at the repository level, they no longer interfere with the upgrade process.

Reminder: The tool uses a temporary local Git repository generated on the fly, so we can do whatever we want in it.

## Test Plan

- Publish react-native-git-upgrade to sinopia
- `npm install -g react-native-git-upgrade`
- Init a new project with an old version: `react-native init MyApp --version=0.50.0`
- Create a new file named `dummy-ignored-file.json` in the project's root
- Append `dummy-ignored-file.json` to the `.gitignore` file
- Run `react-native-git-upgrade`

👉 The `dummy-ignored-file.json` is not deleted.

## Release Notes

[CLI][BUGFIX][react-native-git-upgrade] - Keep the user's ignored files while upgrading

### DIFF
--- a/react-native-git-upgrade/cliEntry.js
+++ b/react-native-git-upgrade/cliEntry.js
@@ -303,7 +303,7 @@ async function run(requestedVersion, cliArgs) {
     await exec('git init', verbose);
 
     log.info('Save current .gitignore file');
-    await copyCurrentGitIgnoreFile(tmpDir);
+    copyCurrentGitIgnoreFile(tmpDir);
 
     log.info('Add all files to commit');
     await exec('git add .', verbose);
--- a/react-native-git-upgrade/cliEntry.js
+++ b/react-native-git-upgrade/cliEntry.js
@@ -147,6 +147,28 @@ function configureGitEnv(tmpDir) {
   process.env.GIT_WORK_TREE = '.';
 }
 
+function copyCurrentGitIgnoreFile(tmpDir) {
+  /*
+   * The user may have added new files or directories in the .gitignore file.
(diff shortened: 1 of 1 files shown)


react-native-git-upgrade/cliEntry.js
```diff
@@ -277,9 +299,12 @@ async function run(requestedVersion, cliArgs) {
     log.info('Configure Git environment');
     configureGitEnv(tmpDir);
 
-    log.info('Init Git repository');
+    log.info('Init temporary Git repository');
     await exec('git init', verbose);
 
+    log.info('Save current .gitignore file');
+    await copyCurrentGitIgnoreFile(tmpDir);
```
janicduplessis: copyCurrentGitIgnoreFile is not async so no need to use await here.

janicduplessis:
Actually I can just fix it myself :)

@facebook-github-bot shipit

janicduplessis:
cc @hramos

ncuillery:
Rebase done @hramos

Issue #15756, opened 2017-09-01
react native run android failing 


DanialV:
### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes
### Environment

1. `react-native -v`:0.47.2
2. `node -v`:7.6.0
3. `npm -v`:5.3.0
4. `yarn --version`:0.27.5

Then, specify:

- Target Platform:Android

- Development Operating System:Linux Ubuntu 16.04.3 LTS

- Build tools:Atom editor with react-native run-android

### Steps to Reproduce

(Write your steps here:)

1.react-native init AwesomeProject
2.react-native run-android

### Expected Behavior
I follow all instructions from react native android guide page, however when I create project with **react-cli** and run `react-native run-android` and android emulator was already opened

### Actual Behavior
here is the error I face with:
>     FAILURE: Build failed with an exception.
>     
>     * What went wrong:
>     A problem occurred configuring root project 'test'.
>     > Could not resolve all dependencies for configuration ':classpath'.
>        > Could not resolve org.apache.commons:commons-compress:1.8.1.
>          Required by:
>              :test:unspecified > com.android.tools.build:gradle:2.2.3 > com.android.tools.build:gradle-core:2.2.3 > com.android.tools.build:builder:2.2.3 >
> com.android.tools:sdklib:25.2.3
>              :test:unspecified > com.android.tools.build:gradle:2.2.3 > com.android.tools.build:gradle-core:2.2.3 > com.android.tools.build:builder:2.2.3 >
> com.android.tools:sdklib:25.2.3 > com.android.tools:repository:25.2.3
>           > Could not resolve org.apache.commons:commons-compress:1.8.1.
>              > Could not get resource 'https://jcenter.bintray.com/org/apache/commons/commons-compress/1.8.1/commons-compress-1.8.1.pom'.
>                 > Could not HEAD 'https://jcenter.bintray.com/org/apache/commons/commons-compress/1.8.1/commons-compress-1.8.1.pom'.
>                    > Connect to jcenter.bintray.com:443 [jcenter.bintray.com/75.126.118.188] failed: Connection timed out
> (Connection timed out)
>     
>     * Try:
>     Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
>     
>     BUILD FAILED
>     
>     Total time: 2 mins 16.65 secs
>     Could not install the app on the device, read the error above for details.
[...]

PR #17395, opened 2017-12-30
add google() to build.gradle in CLI template

riwu:
## Motivation

Fixes android build issue


chirag04:
@riwu what problem does this fix specifically? Can you provide a repro?

local-cli/templates/HelloWorld/android/build.gradle
```diff
@@ -3,6 +3,7 @@
 buildscript {
     repositories {
         jcenter()
+        google()
```
chirag04: also i think this only works for gradle 4 and up. RN is not using gradle 4 yet.

riwu:
Indeed this is only a problem on newer versions of gradle.
See https://github.com/facebook/react-native/issues/15756 and [this](https://stackoverflow.com/questions/44071080/could-not-find-com-android-tools-buildgradle3-0-0-alpha1-in-circle-ci).

Is there documentation on supported versions of gradle?

dulmandakh:
closing this PR because the same change is introduced in https://github.com/facebook/react-native/commit/6e356895e79fb92640295a14483af1a430732247

PR #17397, opened 2017-12-30
Fix #8615: NetInfo.isConnected for iOS

### LINKED ISSUE #8615: iOS: NetInfo.isConnected returns always false
I am currently running RN 0.28.0...

```
NetInfo.isConnected.fetch().then(isConnected => {
      // variable isConnected is always false
    });
```

I would do a PR but I can't code for iOS.

alburdette619:
## Motivation

This is fixing #8615. The problem was that, on initialization, the `_connectionType` variable was still set to its default value of `RCTConnectionTypeUnknown`. The exposed API method did nothing to determine this unless a subscription had be established and then the method would simply return the last reported value. Instead, the exposed oneshot API call now actually checks the connection status through the same methods as the subscription and updates RCTNetInfo’s values before returning.

In order to avoid reporting events without a subscription, a flag is set and unset on calls to start/stopObserving.

## Test Plan

- start app
- observe the (in)correct reporting of the manual status
- change network status to offline
- press refresh
- observe the manual fetch
- start subscription
- change network status to online
- press refresh to show that the manual refresh works (only now working for current RN version)
- change network status to offline
- stop subscription
- change network status to online
- press refresh to show manual refresh does(n't) work without subscription
- start subscription to show it updates to current

Current Behavior: https://drive.google.com/file/d/1Ods6HORgp_vfm1mQVjGwhtH1D7issxjo/view?usp=sharing
Fixed Behavior: https://drive.google.com/file/d/11H1UOF33LeMGvXEOoapU62ARDSb7qoYv/view?usp=sharing

## Release Notes
[IOS] [BUGFIX] [Libraries\Network\RCTNetInfo.m] - Fixed #8615, `iOS: NetInfo.isConnected returns always false`, by decoupling the fetch from the status of the subscription.


alburdette619:
@shergin I'm tagging you here because you were the last one to do a PR on NetInfo.  I don't see any need to rebase currently since no changes have been made to the only file here since last year.  This is my first time attempting to contribute to RN, please let me know if anything more is needed from me.  Thanks!

alburdette619:
Trying to find a reviewer and taking anyone off of this file's last PR I can @christopherdro @hramos.

Issue #13025, opened 2017-03-20
contentInsets always overwrites scrollIndicatorInsets of ScrollView/ListView


allenhsu:
### Description

I set `contentInsets` and `scrollIndicatorInsets` of a `ScrollView`, actually  `ListView`, to different values. But seems the value of `contentInsets` always overwrites the `scrollIndicatorInsets` one.

### Reproduction

https://sketch.expo.io/Byi2M7psg

In this demo, `scrollIndicatorInsets.top` is 20, it's expected to be 100. After deleting `contentInset={{top: 20, left: 0, right: 0, bottom: 0}}`, the `scrollIndicatorInsets` is correct.

### Solution

On line 262 of https://github.com/facebook/react-native/blob/master/React/Views/RCTView.m#L262 `scrollIndicatorInsets` is set to `baseInset` which is `contentInset` most of the time. This method is called when `setContentInset` for `RCTScrollView`.

On line 372 of https://github.com/facebook/react-native/blob/v0.42.3/React/Views/RCTComponentData.m#L372 when enumerate props, the order should not be predictable, but actually `contentInset` always comes after `scrollIndicatorInsets` during my debugging. So `contentInset` always overwrites `scrollIndicatorInsets` unless it's Zero (returned on `if (UIEdgeInsetsEqualToEdgeInsets(contentInset, _contentInset)) return`).

### Additional Information

* React Native version: 0.42.3
* Platform: iOS
* Operating System: MacOS
* Dev tools: iOS 10.2.1

hramos:
Thanks for reporting this. Would you consider sending a PR? It seems like you've already done some of the work here.

gastonmorixe:
Any update?

flowkraD:
I would like to give it a shot if this still exists

Salakar:
👋 hey all, thanks for reporting this issue. 

There was a commit some time ago (https://github.com/timotew/react-native/commit/d1b9d329314b18497a310a443ea195d88eb1e06f) that was meant to have fixed this issue, because of this I'm going to go ahead and close this issue for now. 

Should someone be able to provide a reproduction on the latest version of React Native I'd be happy to re-open this, please let me know.

Thank you

gabrielbull:
@Salakar This bug is back, the code from the PR seems to have disappeared. Can you open the issues again?

PR #17399, opened 2017-12-31
Fix: Avoid contentInsets overwrites scrollIndicatorInsets on iOS #13025

timotew:
Fix to avoid contentInsets always overwrites scrollIndicatorInsets of ScrollView/ListView.
According to issue #13025 https://github.com/facebook/react-native/blob/a8391bde7d757d01521a6d12170fb9090c17a6a0/React/Views/RCTView.m#L299
always overide `scrollIndicatorInsets` when called and only works when `contentInsets` is not set.
We only need to worry about when `contentInsets` is set, We only have to check and reset it after `autoAdjustInsetsForView` is called.

## Motivation

Fix: https://github.com/facebook/react-native/issues/13025

## Related PRs

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -394,6 +394,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
     _automaticallyAdjustContentInsets = YES;
     _DEPRECATED_sendUpdatedChildFrames = NO;
     _contentInset = UIEdgeInsetsZero;
+    _scrollIndicatorInsets = UIEdgeInsetsZero;
     _contentSize = CGSizeZero;
     _lastClippedToRect = CGRectNull;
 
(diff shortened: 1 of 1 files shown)


timotew:
@janicduplessis Hi there! How do we move forward on this PR.

timotew:
@hramos  Hi there! Are these changes allowed?

hramos:
@timotew do you have a test plan?

timotew:
@hramos  my team has been using this branch every since. 
I will try to add a screenshot.

brentvatne:
this is definitely a bug (I repro'd it on https://snack.expo.io/S1eOAIRrM -- `scrollIndicatorInsets` is completely ignored if `contentInset` is specified). i don't think this is the best fix, though. we should probably change [RCTView. autoAdjustInsetsForView](https://github.com/facebook/react-native/blob/22efd95be1f0b236eeaaa8a8e6d01e89771c9543/React/Views/RCTView.m#L299) so that it doesn't reset `scrollIndicatorInsets` to `baseInset` if the scrollview has a `scrollIndicatorInsets` prop set on it

timotew:
@brentvatne that was my initial approach but I was considering the effects on other components that make use of `autoAdjustInsectsForViews`.

brentvatne:
@timotew - I see.

I'm also concerned because `scrollIndicatorInsets` doesn't even seem to be in scope here? did you run this code? looks like it was made from Github directly

timotew:
@brentvatne yes.  I did this before the pull then try the other approach. I later revert back to this. What fix do you think will be best and won't affect other components?

PR #17407, opened 2017-12-31
Migrate from minimist to gar

ethanent:
- Updated all package.json references.
- Updated all code references.

## Motivation

minimist is about 775% the size of the gar package at 31KB! gar is only 4KB in size.

## Test Plan

The changes made were reasonably straightforward. All related tests passed when tests were executed!

image

## Release Notes

[GENERAL] [ENHANCEMENT] [react-native] - Replaced minimist with gar for args parsing.


Kureev:
Hi there! Thanks for the PR, but I'm wondering which problem does it solve? Since we talk about CLI, it does not really matter if it'll be 13kb or 2kb (in my humble opinion). However, on the other side, minimist is widely used in node ecosystem while gar is developed by one person + there is no guaranty it'll be maintained in the future.

grabbou:
What's most important when talking about dependencies is whether they are already part of the dependency tree. 

Minimist is already used by other dependencies of `react-native` and replacing it with another dependency will not result in `minimist` not being downloaded. Other dependencies will still get it. That said, `gar` will be actually an overhead rather than optimisation as no other dependency of `react-native` uses it.

PR #17415, opened 2018-01-02
fix: account for `ListHeaderComponent` length when calculating offset…

### LINKED ISSUE #16612: The calculations for `onViewableItemsChanged` are incorrect if Flatlist has ListHeaderComponent
### Is this a bug report?

Yes

Yes

### Environment

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  6.9.4
  Yarn:  0.19.1
  npm:  3.10.10
  Watchman:  4.7.0
  Android Studio:  2.3 AI-162.4069837

  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: 0.49.5 => 0.49.5

  Target Platform: iOS (10.3) (any though)

### Steps to Reproduce

1. Add a Flatlist component
2. Add a `ListHeaderComponent` of any length
3. use `onViewableItemsChanged` to track viewable rows in local state
4. That's it, the calculations will be off by the header height.

### Expected Behavior

`onViewableItemsChanged` fires at the correct time

(Write what you thought would happen.)

### Actual Behavior

calculations for `onViewableItemsChanged` are always off by the length of the header.

(Write what happened. Add screenshots!)

### Reproducible Demo

- Example repo: https://github.com/shakyShane/flatlist-bug (from react-native init, it's a single file)
- Snack https://snack.expo.io/HJz_uaS0b
- gif

switch

## Flatlist Bug with ListHeaderComponent

`Flatlist` exposes `onViewableItemsChanged` - which can be used to determine which elements are currently
visible within a list. This works correctly.

If a `ListHeaderComponent` is also given, the calculations used for determining which items are in view
will be incorrect - by exactly the height of the header.

in `Libraries/Lists/VirtualizedList.js`, `this._headerLength` is initialised as `0` and then updated on render
to store the length of the header dynamically, but then this value is *not* used anywhere else.

## Calculations need to account for the 'length' of the Header component

We found the easiest way was to subtract the `this._headerLength` from the offset calculation, such as 

```diff
_selectOffset(metrics: {x: number, y: number}): number {
-   return !this.props.horizontal ? metrics.y : metrics.x;
+   return (!this.props.horizontal ? metrics.y : metrics.x) - this._headerLength;
}
```

shakyShane:
… in VirtualizedList - fixes #16612 

Issue is detailed in #16612

## Motivation

I need `onViewableItemsChanged` to account for the length of any ListHeaderComponents

## Test Plan

I couldn't find any tests that currently cover the use-case of a VirtualizedList + ListComponent + scroll with onViewableItemsChanged, so I have not added any tests - all previous tests pass however.

## Release Notes

[GENERAL][BUGFIX][VirtualizedList] - account for `ListHeaderComponent` length


shakyShane:
@bestander Thanks for the update. This is a crucial aspect of creating super-high perf on lists, so I appreciate it now getting some focus :)

bestander:
Hey all.

This is the revert commit https://github.com/facebook/react-native/commit/604bcfa4a83396c402ba8beaa13f40d05d6e9f5c.

So what happened.
We have a fairly popular app that uses FlatList to display an infinite list of elements.
Each element can be a complicated thing and it can have logic that logs position in the FlatList, does click handling with animations etc.
So what we learned is that after this change after some initial scrolling the FlatList could get into a state where it would start scrolling down indefinitely without a stop.
This happened only on iOS and was hard to reproduce, we noticed it only after users started reporting it.

We'll be doing a follow up on this issue internally: what was actually causing the infinite scroll and also how to prevent such break in the future. We'll also raise a question of the original issue that this PR tried to fix.

@shakyShane, do you have some insight what might have caused the infinite scroll?

AndrewJack:
We experienced the same infinite scrolling issue when using this patch. Reverting this back or not using a header component fixes the issue.

shakyShane:
It's so sad that this caused such regressions elsewhere as it's an absolutely essential feature :(

if you have complicated lists, windowing is the solution to perf, and when the calculations for `onViewableItemsChanged` is off as detailed in this issue, you have no chance of getting this right

shakyShane:
I've just tried the latest React Native (0.58.1) and this is still an issue. :(

Akash-T2S:
Hey @shakyShane I'm facing a issue same has you have mentioned - [https://github.com/facebook/react-native/issues/16612](url). Is there any solution for it?

Issue #16381, opened 2017-10-15
CxxBridge, when used with Cocoapods, depends on an older version of "boost" that is also 4x larger than the one downloaded by ios-install-third-party.sh


itchingpixels:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Scanning folders for symlinks in /Users/mark/Dev/drops-react-native/node_modules (9ms)

Environment:
  OS:  macOS High Sierra 10.13
  Node:  8.6.0
  Yarn:  1.1.0
  npm:  5.3.0
  Watchman:  4.9.0
  Xcode:  Xcode 9.0 Build version 9A235
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: 0.49.3 => 0.49.3

### Steps to Reproduce

1. Install react-native via Cocoapods 
2. Add CxxBridge instead of BatchedBridge to your Podfile
2. `pod install`

### Expected Behavior

Download the same version & bundle of the `boost` library that is used by ios-install-third-party.sh.

### Actual Behavior

The `boost` dependency defined in Folly.podspec points to:
https://github.com/CocoaPods/Specs/blob/master/Specs/9/9/d/boost/1.59.0/boost.podspec.json

Which is:
- Version 1.59.0 instead of `boost`
- A way larger bundle (83 Mb instead of 16 Mb)

File specified in the `boost` pod spec:
https://sourceforge.net/projects/boost/files/boost/1.59.0/boost_1_59_0.tar.gz/download

File that ios-install-third-party.sh downloads:
https://github.com/react-native-community/boost-for-react-native/releases/download/v1.63.0-0/boost_1_63_0.tar.gz

karpov3:
Hi @itchingpixels! I have the same problem. My project now is too large cause of boost library. Did you already resolve your issue?

itchingpixels:
@karpov3 : not just yet, but I think we're close:
https://github.com/facebook/react-native/issues/16381#issuecomment-340189502
so you can download the podspec from our fork, fix the podspec file according to fishmwei 's instructions, and place it in a folder, where you can reference it I believe.

the relevant podspec file in our fork:
https://github.com/languagedrops/react-native/commit/1c3f1ab55aa8541bde2adbc4c4f25914a5731021

CFKevinRef:
[...]

PR #16756, opened 2017-11-08
Add boost.podspec speed up download time & reduce size of the package

itchingpixels:
## Motivation

Speed up download time & reduce size of the `boost` dependency. This closes https://github.com/facebook/react-native/issues/16381

## Test Plan

1. Install react-native via Cocoapods (or move a helloWorld react-native application over to use Cocoapods for dependency management)
2. Add CxxBridge instead of BatchedBridge to your Podfile
3. run `pod install`
4. run `react-native run-ios`

## Release Notes
 [IOS] [FEATURE] [third-party-podspecs] - added boost.podspec

PR #17476, opened 2018-01-08
Use boost-for-react-native CocoaPod

CFKevinRef:
## Motivation

Use the same header-only distribution of Boost that Android and the automatic ios-install-third-party.sh script uses. Greatly speeds up `pod install`, even when the distribution was already cached.

The dependency tree looks like this:

* React
  * cxxreact subspec
    * **Boost** (versioned, 1.63.0)
    * Folly
      * **Boost** (any version, will resolve to 1.63.0 from above)

Alternative implementation of PR https://github.com/facebook/react-native/pull/16756. Fixes issue #16381.

## Test Plan
(same as https://github.com/facebook/react-native/pull/16756)
* Install react-native via Cocoapods (or move a helloWorld react-native application over to use Cocoapods for dependency management)
* Add CxxBridge instead of BatchedBridge to your Podfile
* run pod install
* run react-native run-ios

## Related PRs
[Add boost.podspec speed up download time & reduce size of the package](https://github.com/facebook/react-native/pull/16756)

## Release Notes
[IOS] [BUGFIX] [third-party-podspecs] - optimize CocoaPods installation (Boost distribution)


ide:
@facebook-github-bot shipit

xiaoyongchen:
boost-for-react-native/boost/compatibility/cpp_c_headers/cstddef:10:1: error: unknown type name 'namespace'
Semantic Issue
why？？

PR #17499, opened 2018-01-09
Add support for "thread-id" in PushNotificationIOS

Ashoat:
`thread-id` is an Apple-defined key. `PushNotificationIOS.getDeliveredNotifications()` includes it, but the notification object currently ignores it. Since it's in the `aps` key, it's impossible to access without this change, as `getData()` does not include it.

Test Plan: send a notif from APNS with "thread-id" and confirm it's registered correctly.

Release Notes: add "thread-id" key to PushNotificationIOS

### DIFF
--- a/Libraries/PushNotificationIOS/PushNotificationIOS.js
+++ b/Libraries/PushNotificationIOS/PushNotificationIOS.js
@@ -453,7 +453,7 @@ class PushNotificationIOS {
    * If you do not call this method your background remote notifications could
    * be throttled, to read more about it see the above documentation link.
    */
-  finish(fetchResult: FetchResult) {
+  finish(fetchResult: string) {
     if (!this._isRemote || !this._notificationId || this._remoteNotificationCompleteCallbackCalled) {
       return;
     }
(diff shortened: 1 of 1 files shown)


Ashoat:
@shergin this is a tiny JS-only PR, any chance you could import it? Without this, it's impossible to read the "thread-id" field on an incoming iOS notification.

Ashoat:
I also corrected a Flow type error - finish takes a `string`, not a `FetchResult` (which is the object containing the valid strings).

Libraries/PushNotificationIOS/PushNotificationIOS.js
```diff
@@ -451,7 +453,7 @@ class PushNotificationIOS {
    * If you do not call this method your background remote notifications could
    * be throttled, to read more about it see the above documentation link.
    */
-  finish(fetchResult: FetchResult) {
+  finish(fetchResult: string) {
```
shergin: Is this intentional. Why?

Ashoat: Made a comment on the PR but probably should've mentioned it inline too.

> I also corrected a Flow type error - finish takes a string, not a FetchResult (which is the object containing the valid strings).

This type is incorrect, and it shows when you attempt to use the API.

PR #17501, opened 2018-01-09
Fix documentation links

Taym95:
Old link gives `404 File not found`.


ide:
@facebook-github-bot shipit

Issue #17267, opened 2017-12-19
Android: Border radius renders in a blurred / choppy way in 0.51.0


dannycochran:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  6.11.4
  Yarn:  Not Found
  npm:  3.10.10
  Watchman:  4.9.0
  Xcode:  Xcode 8.3.3 Build version 8E3004b
  Android Studio:  Not Found

Packages: (wanted => installed)
  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: 0.53.0 => 0.53.0

  Target Platform: Android (22)

### Steps to Reproduce

Create a simple view:

```javascript
export default class App extends Component {
  render() {
    return (
      <View style={styles.background}>
        <View style={styles.container}>
          <Text>Welcome to React Native!</Text>
        </View>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  background: {
    backgroundColor: 'purple',
    flex: 1
  },
  container: {
    position: 'absolute',
    left: 100,
    top: 100,
    width: 200,
    height: 100,
    backgroundColor: 'orange',
    borderWidth: 2,
    borderColor: 'orange',
    borderRadius: 50,
    justifyContent: 'center',
    alignItems: 'center'
  }
});
```

### Expected Behavior

I would expect that the border does not render in a strange and pixelated way. Previously in react-native 0.49.5, the exact same code would render like so:

<img width="615" alt="screen shot 2017-12-18 at 5 16 14 pm" src="https://user-images.githubusercontent.com/1217178/34135720-3be7184e-e417-11e7-9e2f-5b4c351bab9b.png">

### Actual Behavior

In react-native 0.53.0, that same code renders this blurry border:

<img width="707" alt="screen shot 2017-12-18 at 5 15 54 pm" src="https://user-images.githubusercontent.com/1217178/34135728-45aa6930-e417-11e7-98d2-3cd75ee404cb.png">

### Reproducible Demo

The code snippet above compiles and should be sufficient to reproduce this problem.

dannycochran:
Pinging @rsnara who is listed on release notes for 0.51.0 for border work, and @AaaChiuuu who worked on "overflow: hidden" for 0.50.0.

SirNeuman:
[...]

Issue #17311, opened 2017-12-21
Jagged rendering of round borders on Android


firefueled:
### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment
Environment:
  OS: Linux 4.4
  Node: 6.10.0
  Yarn: 0.24.5
  npm: 3.10.10
  Watchman: 4.7.0
  Xcode: N/A
  Android Studio: Not Found

Packages: (wanted => installed)
  react: ~16 => 16.2.0
  react-native: ~0.51.0 => 0.51.0

Android (any)

### Steps to Reproduce
Simply create a round bordered View.

### Expected Behavior
The bordered View should be smooth, or as smooth as a border-less rounded View

### Actual Behavior
The View with a round border (left) is jagged and ugly while a rounded border-less, but background-colored, View (right) is fine.
![round views](http://pablo.universomw8.com.br/Screenshot_1513881456.png)

### Reproducible Demo
https://github.com/firefueled/rn-bad-borders
The problem is not visible on Snack or a current phone, tested on Samsung S5, but it is so on devices and emulators with lower screen DPI.
This latest release where this starts happening in 0.51.0. It doesn't on 0.50.4
[This commit](https://github.com/facebook/react-native/commit/4994d6a389b4e41ba25e802edab5d3fdc9e8a4f1) seems to have changed something related to border drawing on Android but I haven't looked to deep into it.

firefueled:
The exact same issue has been reported here #17267 
A work-around is "faking" a border by having a smaller view on top of a larger one like:

```jsx
<View style={styles.border}>
  <View style={styles.content}>
  </View>
</View>
...
const styles = StyleSheet.create({
  border: {
    backgroundColor: 'blue',
    borderRadius: 50,
    height: 50,
    width: 50,
    justifyContent: 'center',
    padding: 5,
  },
  content: {
    borderRadius: 40,
    height: '100%',
    width: '100%',
    backgroundColor: 'white',
  },
});
```

mezod:
+1

alpamys-qanybet:
Hey guys, I just copied the ```ReactViewBackgroundDrawable.java, ReactViewBackgroundManager.java, ReactViewGroup.java``` from ```RN0.50.1``` and IT WORKS.

Albert0405:
@alpamys-qanybet  where exactly can i find that files?

[...]

Issue #17224, opened 2017-12-15
Border not animating with scale


wassgha:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS: macOS High Sierra 10.13.2
  Node: 8.5.0
  Yarn: 1.1.0
  npm: 5.3.0
  Watchman: 4.9.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: ^0.51.0 => 0.51.0

  Target Platform: Android (26+)

### Steps to Reproduce

After upgrading to react-native 0.51.0, a component that used to be animated (scales up and down based on scroll) and that has a border (with border-radius too) no longer animates. The component itself still scales (verified by adding a background) but the borders don't. Since there has been a lot of work on borders in 0.51.0 I am assuming this has to do with that.

Steps

1. Create a component using `Animated.View` that has styles including a border. In my case this is
```
    borderRadius: 88 / 2,
    width: 88,
    height: 88,
    borderWidth: 4,
    borderColor: 'rgba(0, 0, 0, 0.8)'
```
and add an animated variable as its `transform: [{scale: ...}]`.
2. Make sure you are using `react-native` 0.51.0
3. Every aspect of the component will scale as the animated value changes, except for borders

### Expected Behavior

Prior to upgrading to `0.51.0` the borders used to scale

### Actual Behavior

![ezgif com-video-to-gif 11](https://user-images.githubusercontent.com/591655/34030793-7034f4be-e13c-11e7-97db-845238af6285.gif)

### Reproducible Demo

[https://snack.expo.io/Sk6o8ZWMG](https://snack.expo.io/Sk6o8ZWMG)

(Works on Appetize but not on actual phone)

wassgha:
Pinging @rsnara and @AaaChiuuu

Druux:
same for me! any updates on this?

zachgibson:
@wassgha I ran the snack and it was a different app example. Where is the snack for the GIF you posted?

wassgha:
@zachgibson the snack is an isolated version of the bug that occurs inside the beta if our app (which you can see in the gif). I can invite you to the beta if you'd like but the effect is the same.

arma7x:
[...]

Issue #17144, opened 2017-12-11
borderWidth not working on RN 0.51 Android


maartenvandillen:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 7.6.0
  Yarn: 0.27.5
  npm: 4.1.2
  Watchman: 4.7.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: 0.51.0 => 0.51.0

  Target Platform: Android (7.0)

### Steps to Reproduce

1. render a View with style borderWidth: 2

### Expected Behavior

In previous RN releases, the border would be rendered with a width of 2.

### Actual Behavior

Since upgrading to RN 0.51 the border is rendered with a width of 1.
Using a similar style like for instance borderTopWidth: 2 does work.

### Reproducible Demo

Let me know if you cannot reproduce this

dseipp:
Having the same issue.  Seems to fail when using borderRadius.  Was able to fix it by moving to a sub view that didn't have the border radius set.  (disclaimer: I'm on 50.3)

Check out this post: https://github.com/facebook/react-native/issues/12403

maartenvandillen:
The UI design has changed since I posted this, I don't need to use borderWidth anymore. You are correct that I was not using v0.53 at the time, because that wasn't released yet. If you really want to know if it's still not working go check yourself Mr react-native-bot.

PR #17514, opened 2018-01-10
Android: Adjust border radius to maximum of half the smallest side

yedidyak:
## Motivation
Currently, if a View has a border radius set to larger than the size of the View, and there s a borderWidth set, there is a very serious performance problem on Android.

## Test Plan

This can be verified easily by making any view with a thin border, and watching the Android GPU performance or memory as it is rendered. Native memory allocation will immediately max out, and the GPU render profiler will look like this:

screenshot_1515418843

Memory:
screen shot 2018-01-08 at 16 44 25

With the fix, these problems disappear.

## Release Notes
 [ANDROID] [BUGFIX][View] - Treat larger than possible border radius as meaning 50% of the smallest side

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewBackgroundDrawable.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewBackgroundDrawable.java
@@ -277,7 +277,10 @@ public class ReactViewBackgroundDrawable extends Drawable {
   }
 
   public float getFullBorderRadius() {
-    return YogaConstants.isUndefined(mBorderRadius) ? 0 : mBorderRadius;
+    float desiredRadius = YogaConstants.isUndefined(mBorderRadius) ? 0 : mBorderRadius;
+    float minSide = Math.min(getBounds().height(), getBounds().width());
+    float radius = Math.min(minSide / 2, desiredRadius);
+    return radius;
   }
 
   public float getBorderRadius(final BorderRadiusLocation location) {


wodin:
These test failures are unrelated to this pull request.
The test-android one seems to be because ReactCommon/yoga/BUCK is missing the following:

include_defs("//ReactCommon/DEFS")

yedidyak:
@hramos @grabbou ^

This is being used in production as a patch in Wix.

grabbou:
Looks good to me, @janicduplessis and @emilsjolander would you mind looking at this?

grabbou:
For the test failures, please rebase your branch as the master is passing :)

janicduplessis:
Does it also work when setting a specific corner border radius like bottomLeftBorderRadius?

I find the bug kind of weird though, would prefer to find what is actually causing the high memory usage. Have you tried looking at what the path we are drawing looks like to make sure that is ok?

yedidyak:
I rebased, thanks.

@janicduplessis The end result looked correct, but the bounds of the ReactViewBackgroundDrawable expanded to huge numbers in all directions. I tried to analyse what was happening to cause that, I don't remember exactly what I found but the algorithm simply doesn't expect larger than possible border radii.

janicduplessis:
@rsnara Any idea about what the best fix would be here since you are the last one to have worked on this? :)

yedidyak:
@janicduplessis Whatever is the issue with the underlying algorithm, limiting the borderRadius to the maximum possible size is still worthwhile, no?

yedidyak:
@janicduplessis @grabbou bump

sercand:
After upgrading to React-Native 0.54 our app becomes so slow on because of this issue.

yedidyak:
@janicduplessis @grabbou bump?

grabbou:
I'll leave the final word to @janicduplessis as he seemed to be more active than me in this thread.

itchingpixels:
I assume this would fix:
https://github.com/facebook/react-native/issues/17267
https://github.com/facebook/react-native/issues/17311
https://github.com/facebook/react-native/issues/17144
https://github.com/facebook/react-native/issues/17224

And also help us greatly!

Is there any chance it can be approved/merged?
@janicduplessis @grabbou

@yedidyak says they have been using it in production for a while.

atlanteh:
@janicduplessis can you please have a look?

mezod:
+1

atlanteh:
I'm not sure, but I think this PR can be closed, as the issue #17267 has been fixed which might be the root cause.

Issue #16546, opened 2017-10-26
Fatal Exception: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.


iamdurui:
Is this a bug report?

Yes

Have you read the Contributing Guidelines?

No

Environment

react-native: 0.46.2
react: 16.0.0-alpha.12
I got this crash report from Fabric.io but I dont know where and when it happened.

Fatal Exception: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
       at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:6398)
       at android.view.ViewRootImpl.invalidateChildInParent(ViewRootImpl.java:905)
       at android.view.ViewGroup.invalidateChild(ViewGroup.java:4340)
       at android.view.View.invalidate(View.java:10968)
       at android.view.View.invalidate(View.java:10923)
       at android.widget.TextView.updateAfterEdit(TextView.java:7558)
       at android.widget.Editor.finishBatchEdit(Editor.java:1121)
       at android.widget.Editor.endBatchEdit(Editor.java:1103)
       at android.widget.TextView.endBatchEdit(TextView.java:5980)
       at com.android.internal.widget.EditableInputConnection.endBatchEdit(EditableInputConnection.java:77)
       at android.view.inputmethod.BaseInputConnection.finishComposingText(BaseInputConnection.java:283)
       at android.view.inputmethod.InputMethodManager.checkFocusNoStartInput(InputMethodManager.java:1330)
       at android.view.inputmethod.InputMethodManager.onWindowFocus(InputMethodManager.java:1374)
       at android.view.ViewRootImpl$ViewRootHandler.handleMessage(ViewRootImpl.java:3253)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:136)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(Unknown Source)
       at java.lang.Thread.run(Thread.java:841)

edo1493:
I have just seen this while reloading my android app.

3210jr:
Hey, same issue here.
I am having this issue being reported a lot on my crash reports... I am not sure if its relevant but after trying on a few devices i am notice it becoming very common among the lower end devices like android 4.4.2.

KarandikarMihir:
Can confirm!

[...]

Issue #16042, opened 2017-09-21
Only the original thread that created a view hierarchy can touch its views.


AkilLis:
This "Only the original thread that created a view hierarchy can touch its views" error given after react native version upgrade 44.2 to 47.2. 

  Environment:
    OS:  ubuntu 16.04
    Node:  6.10.0
    npm:  3.10.10
    Watchman:  4.7.0
    Android Studio:  2.3 AI-162.4069837

  Packages: (wanted => installed)
    react-native: 0.44.2 => 0.47.2
    react: 16.0.0-alpha.12 => 16.0.0-alpha.12

  Target Platform: Android (6.0.0)

hramos:
Hey, thanks for reporting this issue!

It looks like your description is missing some necessary information, or the list of reproduction steps is not complete. Can you please add all the details specified in the [template](https://github.com/facebook/react-native/blob/master/.github/ISSUE_TEMPLATE.md)? This is necessary for people to be able to understand and reproduce the issue being reported.

I am going to close this, but feel free to open a new issue that meets the requirements set forth in the template. Thanks!

jhalborg:
I second this. Using Expo SDK 21 along with TypeScript 2.5, and getting consistent crashes on recompilation each time on Android.

I found my issue by viewing the stacktrace I could send to the dev (through the Android report bug feature). It was due to the following call:

abi21_0_0.com.facebook.react.devsupport.DebugOverlayController.setFpsDebugViewVisible(DebugOverlayConroller.java:49)

By turning off the Inspector module (for FPS perf etc), Expo stopped crashing.

ThaJay:
It happens on every reload from the dev server for me.
```
"react-native": "^0.48.4",
"react": "^16.0.0",
```
```
E/AndroidRuntime(17714): FATAL EXCEPTION: main
E/AndroidRuntime(17714): Process: com.chipta, PID: 17714
E/AndroidRuntime(17714): android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
E/AndroidRuntime(17714): 	at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:7093)
E/AndroidRuntime(17714): 	at android.view.ViewRootImpl.requestLayout(ViewRootImpl.java:1076)
E/AndroidRuntime(17714): 	at android.view.View.requestLayout(View.java:17416)
[...]

PR #17544, opened 2018-01-11
Android: Fix crash when perf monitor does not run within UI thread.

bae-unidev:
## Motivation

Application crash when FpsDebugView(Perf monitor) does not run within UI Thread.
#16042, and [this comment](https://github.com/facebook/react-native/issues/16546#issuecomment-355214148) are relative issue.
```
E/AndroidRuntime(17714): android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
...
E/AndroidRuntime(17714): at com.facebook.react.devsupport.DebugOverlayController.setFpsDebugViewVisible(DebugOverlayController.java:49
```
I think that it will be possible to reproduce when move the task that updates the UI onto the another thread. (Android Main Thread <> JS Thread?).
So I have changed the code that create FpsDebugView to work only within the UI Thread. 

## Test Plan

1. Turn on Perf monitor.
2. Open dev menu, and click "reload"
3. Move the task that updates the UI onto the another thread.

## Release Notes

[ANDROID] [BUGFIX] [Devtools] - Fix crash when perf monitor does not run within UI thread.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
@@ -99,12 +99,12 @@ import javax.annotation.Nullable;
     UiThreadUtil.runOnUiThread(new Runnable() {
       @Override
       public void run() {
-        if (!permissionCheck(mReactContext)) {
-          FLog.d(ReactConstants.TAG, "Wait for overlay permission to be set");
+        if (mFPSDebugViewContainer != null) {
           return;
         }
 
-        if (mFPSDebugViewContainer != null) {
+        if (!permissionCheck(mReactContext)) {
+          FLog.d(ReactConstants.TAG, "Wait for overlay permission to be set");
           return;
         }
 
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
@@ -21,7 +21,6 @@ import android.widget.FrameLayout;
 
 import com.facebook.common.logging.FLog;
 import com.facebook.react.bridge.ReactContext;
-import com.facebook.react.bridge.UiThreadUtil;
 import com.facebook.react.common.ReactConstants;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


bae-unidev:
@hramos  Hi! Can you take a look at this commit?

hramos:
Happy to take another look once it's reviewed! Take a look around and see if there's someone who has contributed to a similar area who may be able to take a look at the PR.

bae-unidev:
@kmagiera Hi! I think this change is related to your commit. My changes are in the similar area as your last commit. (d19afc73f5048f81656d0b4424232ce6d69a6368)
Can you take a look at this commit?

ReactAndroid/src/main/java/com/facebook/react/devsupport/DebugOverlayController.java
```diff
@@ -96,25 +97,51 @@ public DebugOverlayController(ReactContext reactContext) {
...
+  private void showFpsDebugView() {
+    UiThreadUtil.runOnUiThread(new Runnable() {
+      @Override
+      public void run() {
+        if (!permissionCheck(mReactContext)) {
+          FLog.d(ReactConstants.TAG, "Wait for overlay permission to be set");
+          return;
+        }
+
+        if (mFPSDebugViewContainer != null) {
```
danilobuerger: @bae-unidev you might want to swap that with the `permissionCheck` as this null check should run faster.

bae-unidev: Thanks, I will fix it.

bae-unidev:
@react-native-bot I updated the release notes.

kmagiera:
@bae-unidev Thanks for looking into this.

IMO adding runOnUIThread is not the best solution as it just introduces more complexity without understanding what is happening underneath.

I'd suggest that instead of adding `runOnUIThread` you add an assertion to make sure `setFpsDebugViewVisible` always is run from the UI thread. Then try to figure out why it isn't run from the UI thread.

Just by quickly looking at the code it appears that the reason is that `DevSupportManagerImpl.resetCurrentContext` is now being called from non-ui thread (via `DevSupportManagerImpl.onReactInstanceDestroyed`. And it appears like it would've been better for `resetCurrentContext` to also run on the UI thread as it modifies `mCurrentContext` that is then only accessed from the UI thread. So to avoid concurrent modification issues it would be better to synchronize access to `mCurrentContext` or just make sure it can be modified in the same thread it is accessed (that it on the UI thread).

Therefore my suggestion would be to wrap `resetCurrentContext` method with `UiThreadUtil.runOnUiThread` to make sure `mCurrentContext` is being updated on the main thread only and this will also fix the problem with fps monitor method that would require to be called from the ui thread.

Does that make sense?

bae-unidev:
@kmagiera Thank you for your comment.

Does `getJSModule(HMRClient.class).enable` in `resetCurrentContext` need to be run in the UI thread? If not, I think it would be better to wrap only the part that recreate `DebugOverlayController` than to wrap the resetCurrentContext method with `UiThreadUtil.runOnUiThread`.
If `mDevSettings.isHotModuleReplacementEnabled` is true, the HMRClient seems to be making network requests. Then the UI thread should wait. So I think it would be better to wrap the recreate part.

Thanks for your kindness, Can you take a look at this commit?

danilobuerger:
This seems to have been fixed at https://github.com/facebook/react-native/commit/4fcd9970bd2dfb24890bc87e9c82e16dab71ec09

Issue #13671, opened 2017-04-26
CameraRoll.getPhotos breaks with assetType = photos


zackify:
### Description

I'm using this library: https://github.com/jeanpan/react-native-camera-roll-picker/blob/master/index.js under the hood it calls the cameraRoll api. When I pass in 'All' for the asset type, it works. When I use 'Photos' as the type, the app crashes inside of the RCTCameraRoll code:

<img width="1139" alt="screen shot 2017-04-26 at 10 11 19 am" src="https://cloud.githubusercontent.com/assets/449136/25438878/c1078c02-2a68-11e7-8454-a30b3fd97d72.png">

### Reproduction Steps and Sample Code

I was going to try and add an example on expo, but it won't work unless you have more than just photos in your camera roll. To reproduce, just call `CameraRoll.getPhotos({assetType: 'Photos'})`

* React Native version: 0.43.4
* Platform: iOS
* Development Operating System: MacOS
* Dev tools: Xcode 8.3.2, iOS 10.3.1

elanini:
Hi zackify, I'm having the same issue. How did you solve it?

klvs:
@zackify I'm running into this problem. I noticed you closed it, did you solve it or find a work around, if so what is it?

zackify:
It works fine when I call that method in my own component. Doesn't seem to
be a bug with React native but rather that 3rd party library. I did the
exact call to get all media types and it worked fine for me, so I closed.

pentarex:
I am having the same issue with CameraRoll from RN.

```javascript
CameraRoll.getPhotos({
            first: 20,
            groupTypes: 'All',
            assetType: 'Photos'
        }).then(r => {
            console.log(r);
            
        }).catch((err) => {
             console.log('gallery', err);
        });
```

The thing is that some photo for some reason filename is missing (maybe iCloud). And I think its crashing because of that (not completely sure), but as I see @zackify also his filename was nil.

It doesn't fail under the simulator, but on my real device, where most of my photos are on iCloud.

sujit13666:
Same Here guys.

pentarex:
The pull request was approved, it only needs to get merged.

facuescobar:
@pentarex was this update merged? 

I can not see it in the latest version of RN

pentarex:
I don’t think. Check my profile for alternative library that I’ve wrote if you want

facuescobar:
@pentarex I did the same update on my project and I think that fixed some cases.

[...]

PR #17549, opened 2018-01-11
iOS: Fix Crash when CameraRoll is getting assets from iCloud and no filename is provided. #13671

pentarex:
## Motivation

RCTCameraRollManager.m crashes when there is no filename provided for the asset (usually from iCloud). #13671

## Test Plan

Tested on real device with iCloud library using the library without and after my fix. It worked after my changes.

## Release Notes
[IOS] [BUGFIX] [CameraRoll] - Changed the filename if nil to be empty when the asset is coming from iCloud.

### DIFF
--- a/Libraries/CameraRoll/RCTCameraRollManager.m
+++ b/Libraries/CameraRoll/RCTCameraRollManager.m
@@ -199,7 +199,7 @@ RCT_EXPORT_METHOD(getPhotos:(NSDictionary *)params
               @"group_name": [group valueForProperty:ALAssetsGroupPropertyName],
               @"image": @{
                 @"uri": uri,
-                @"filename" : filename,
+                @"filename" : filename ? filename : @"",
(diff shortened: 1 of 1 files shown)


Libraries/CameraRoll/RCTCameraRollManager.m
```diff
@@ -199,7 +199,7 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
               @"group_name": [group valueForProperty:ALAssetsGroupPropertyName],
               @"image": @{
                 @"uri": uri,
-                @"filename" : filename,
+                @"filename" : filename ? filename : @"",
```
ide: `filename ?: [NSNull null]` might be better

ide:
Thanks for the update. This looks good to me. 

@facebook-github-bot shipit

PR #17389, opened 2017-12-29
Add missing BUCK import

esam091:
## Motivation

A lot of builds seem to be failing in test-android with the error message `NameError: name 'fb_xplat_cxx_library' is not defined`
https://circleci.com/gh/facebook/react-native/29412
https://circleci.com/gh/facebook/react-native/29401
https://circleci.com/gh/facebook/react-native/29393

This has problem been [solved before](https://github.com/facebook/react-native/commit/33da6047a3650e5724bb4a063ccb7ccceeaf3be9#diff-74f91ea600239f82d539d42c17cab119), but for some reason the include_defs was [removed again](https://github.com/facebook/react-native/commit/1fd7315c3b31b4378d1ee18ed98a1d941a0feecf#diff-74f91ea600239f82d539d42c17cab119)

## Test Plan

Make sure build passes

PR #17551, opened 2018-01-11
CI for Android fixed

dryganets:
Fix for android github CI.

## Motivation
Github CI doesn't work without this change.

## Test Plan
CI passes.

## Release Notes
[INTERNAL] Android CI fix.

### DIFF
--- a/ReactCommon/DEFS
+++ b/ReactCommon/DEFS
@@ -2,8 +2,7 @@
 
 # Building is not supported in OSS right now
 def rn_xplat_cxx_library(name, **kwargs):
-  cxx_library(name = name)
-
+  cxx_library(name, **kwargs)
 # Helper for referring to an Android RN target
 def react_native_target(path):
   return '//ReactAndroid/src/main/' + path
--- a/ReactCommon/yoga/BUCK
+++ b/ReactCommon/yoga/BUCK
@@ -1,4 +1,6 @@
-fb_xplat_cxx_library(
+include_defs("//ReactCommon/DEFS")
+
+rn_xplat_cxx_library(
     name = "yoga",
     srcs = glob(["yoga/*.cpp"]),
     header_namespace = "",
--- a/ReactAndroid/src/main/third-party/java/robolectric3/robolectric/BUCK
+++ b/ReactAndroid/src/main/third-party/java/robolectric3/robolectric/BUCK
@@ -115,7 +115,7 @@ prebuilt_jar(
 export_file(
     name = "robolectric-android-all-binary-jar",
     src = ":robolectric-android-all-binary-remote-jar",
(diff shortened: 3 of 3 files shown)


dryganets:
I rolled back to the commit where iOS tv worked to make all unit tests pass.
tvOS CI is broken in the master.

hramos:
Thanks for fixing this. It was reverted internally due to a misunderstanding because usually Yoga wouldn't need these DEFS. Let's land this for now as it fixes OSS CI until a better long term fix can be implemented.

ReactCommon/yoga/BUCK
```diff
@@ -1,4 +1,6 @@
-fb_xplat_cxx_library(
+include_defs("//ReactCommon/DEFS")
+
+rn_xplat_cxx_library(
```
hramos: Why `rn_` and not `fb_`?

ReactCommon/DEFS
```diff
@@ -2,8 +2,7 @@
 
 # Building is not supported in OSS right now
 def rn_xplat_cxx_library(name, **kwargs):
-  cxx_library(name = name)
-
+  cxx_library(name, **kwargs)
```
hramos: What is this change for?

dryganets: I'm still playing around. The android tests are failing randomly.
This fixed the fetch for me on the master.

dryganets: As I don't know all the steps needed to run tests I'm just pushing the branch and see the results.
Then fixing the related part of the script locally.

ReactCommon/yoga/BUCK
dryganets: fb is not defined in common.

dryganets:
There is one more android fail:

File does not exist: buck-out/gen/ReactAndroid/src/main/third-party/java/robolectric3/robolectric/android-all-4.1.2_r1-robolectric-0.jar
java.lang.IllegalArgumentException: File does not exist: buck-out/gen/ReactAndroid/src/main/third-party/java/robolectric3/robolectric/android-all-4.1.2_r1-robolectric-0.jar

dryganets:
All robolectric dependencies by some reason are spread across set of binary-jar directories.
This causes the failure.

I guess it is a result of the BUCK tool update.

dryganets:
@hramos this time it should build for Android.
Somebody needs to fix iosTv linking for x86_64.

dryganets:
yey!
Android passed.
I think it is ready for merge.

dryganets:
This is a commit which broke  tvTests:
c491b2223313676bd4900de7a8c70a10051fa9f0

dryganets:
This PR is related and could be closed now:
https://github.com/facebook/react-native/pull/17389

PR #17564, opened 2018-01-12
CI for iOS fixed

dryganets:
## Motivation
iOS TV CI is currently broken.

## Test Plan

CI for both iOS and iOS TV successfully pass.
## Release Notes
 [INTERNAL] Continous integration for iOS fixed.


dryganets:
@hramos this is a fix for regression introduced here: c491b22
Project file was edited manually and some ID's messed up, as a result, one implementation file didn't compile in iOS TV project.

PR #17575, opened 2018-01-12
All android logs are dispatched through FLog java class

dryganets:
## Motivation
Currently, c++ & javascript logs on Android are written directly to logcat instead of using the FLog java class.
As a result, it is impossible to customize C++ and javascript logging by providing a different LoggerDelegate implementation.

With this change, it is possible to enable saving logs to file by providing a custom implementation of LoggingDelegate.

Without this change, the logs written by JS and C++ code will never land to the file.

## Test Plan
Make sure that logging still works.
Make custom LoggingDelegate which logs to the file.
Make sure that all logs are saved.

## Release Notes
[ANDROID] [ENHANCEMENT] [Logging] - All logs are written using the FLog class. It is possible to customize logging behavior across the board.

### DIFF
--- a/ReactAndroid/src/main/jni/first-party/fb/log.cpp
+++ b/ReactAndroid/src/main/jni/first-party/fb/log.cpp
@@ -26,8 +26,9 @@ int fb_printLog(int prio, const char *tag,  const char *fmt, ...) {
   va_end(va_args);
   if (gLogHandler != NULL) {
       gLogHandler(prio, tag, logBuffer);
+  } else {
+    __android_log_write(prio, tag, logBuffer);
   }
-  __android_log_write(prio, tag, logBuffer);
   return result;
 }
 
--- a/ReactAndroid/src/main/jni/react/jni/JSLogging.cpp
+++ b/ReactAndroid/src/main/jni/react/jni/JSLogging.cpp
@@ -5,10 +5,49 @@
 
 #include "JSLogging.h"
 
+#include <fb/Environment.h>
 #include <fb/log.h>
+
+#include <jni/LocalReference.h>
+#include <jni/LocalString.h>
+
+using namespace facebook::jni;
 
 namespace facebook {
 namespace react {
+namespace JSLogging {
+
+static jclass gLoggerClass;
+static jmethodID gVerboseLogMethod;
+static jmethodID gDebugLogMethod;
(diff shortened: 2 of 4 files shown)


ReactAndroid/src/main/jni/first-party/fb/log.cpp
```diff
@@ -28,8 +28,9 @@ int fb_printLog(int prio, const char *tag,  const char *fmt, ...) {
   va_end(va_args);
   if (gLogHandler != NULL) {
       gLogHandler(prio, tag, logBuffer);
+  } else {
+    __android_log_write(prio, tag, logBuffer);
```
dryganets: We don't need to write the log to logcat in case we have handler as this is a handler responsibility now.
If we do we will have duplicate logs.

ReactAndroid/src/main/jni/react/jni/JSLogging.cpp
```diff
@@ -8,29 +8,89 @@
...
+          methodId = gErrorLogMethod;
+          break;
+        default:
+          methodId = gVerboseLogMethod;
+      }
+      return methodId;
+  }
+
+  void log(int prio, const char* tag, const char* message) {
+    ThreadScope threadScope;
```
dryganets: Logging everything using FLog class.

hramos:
Still need to get back to this PR. It's failing internally with:

```
error: use of undeclared identifier 'reactAndroidLoggingHook'; did you mean 'JSLogging::reactAndroidLoggingHook'?
                                   reactAndroidLoggingHook(message, logLevel);
```

dryganets:
@hramos, I could move the reactAndroidLoggingHook out of the namespace if it will make contribution easier.
I think it is better to have it inside of the namespace as this is global variable though.

cpojer:
Hello,

I spent some time to try to make this work at Facebook and it seems that we have made too many changes to React Native for this to work at the moment, so I cannot merge this without breaking and crashing things. Unfortunately I have to close your PR but I promise we'll do better in the future and review code more promptly.

dryganets:
@cpojer,
Would remove the namespace from the variable help to merge it?
Or there is more work involved?

cpojer:
I tried that but it wasn't enough to not make our apps crash. I'm not sure why that happened though and unfortunately don't know how to properly debug it.

PR #17596, opened 2018-01-15
Remove redundant config in AndroidManifest.xml

gengjiawen:
* remove redundant version code, name, miniSdk, targetSdk, they now are configed in build.gradle
* allowbackup = true is not a secure config

## Motivation

Clean up redundant config and remove security risk config.

## Test Plan

test android template still works.

## Related PRs

none

### DIFF
--- a/local-cli/templates/HelloWorld/android/app/src/main/AndroidManifest.xml
+++ b/local-cli/templates/HelloWorld/android/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
       android:name=".MainApplication"
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
+      android:allowBackup="false"
       android:theme="@style/AppTheme">
       <activity
         android:name=".MainActivity"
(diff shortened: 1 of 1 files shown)


local-cli/templates/HelloWorld/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,18 +1,11 @@
...
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
 
-    <uses-sdk
-        android:minSdkVersion="16"
-        android:targetSdkVersion="22" />
-
     <application
       android:name=".MainApplication"
-      android:allowBackup="true"
```
janicduplessis: According to https://stackoverflow.com/questions/12648373/what-is-androidallowbackup this will cause a lint warning if not set. I think we could just set it to false by default.

gengjiawen: @janicduplessis fixed.

janicduplessis:
Thanks!

@facebook-github-bot shipit

local-cli/templates/HelloWorld/android/app/src/main/AndroidManifest.xml
f3ist: Wouldn't it be beneficial to most app users to have automatic backups? Developers who aren't android savvy will not know to change this setting; and malicious user's can root their phone and get the data regardless. At the very least, there should be RN documentation about it.

https://developer.android.com/guide/topics/data/autobackup.html#EnablingAutoBackup

PR #17617, opened 2018-01-15
Make DevTools configurable

vovkasm:
## Motivation

The main motivator for this change is #10027 ("__nw_connection_get_connected_socket_block_invoke Connection has no connected handler" error on iOS 10)
This issue about spam in log output makes application very hard to debug. Error logs generated from inside Apple frameworks when websocket connection can't be established and do not do any harm themselves, but they are repeated every 2 seconds by reconnects to React DevTools and this leads to many meaningless lines in console. This makes quick analysis of logs impossible without running DevTools.
Commit 878b7e42fae990df745af2db6ac50cbac8a8ce1c tries to fix this by hacking system logging functions and filtering some messages. It works sometimes, but has drawbacks:
1. It introduces stubbing of system API functions which is hard to debug and understand.
2. It can hide really useful logs, we do not really know when and why similar errors can be logged from Apple frameworks
3. It needs maintenance with each release of iOS because such errors can change - they are not a public API
4. It is already not working on real devices (https://github.com/facebook/react-native/issues/10027#issuecomment-356484971), because errors are different,
more here https://github.com/facebook/react-native/issues/10027#issuecomment-320993090

All this can be workarounded. I can simply run React DevTools utility and these logs disappear because now websocket connection is successful, but:
1. Connection from real device to React DevTools is much harder and doesn't work "out of the box"
2. I want to test my application in offline mode, want to switch networks, etc...

So as in https://github.com/facebook/react-native/issues/10027#issuecomment-297020061
I implemented an option to disable DevTools connection from dev menu. Now developer can press `Cmd+D` and select `Disable DevTools` in menu to stop application's infinite attempts to reconnect to DevTools.

There are other variants to deal with this:
1. Do not run DevTools automatically from `InitializeCore.js`, developers who need it can insert code like below in their application themselves:
```
require('react-devtools-core').connectToDevTools({
     host: '192.168.2.115',
     port: '8097',
});
```
2. Allow to enable/disable DevTools by another mechanism which I am not aware of. I'd prefer not to pollute `globals` with magic flags but `InitializeCore.js` is called too early in application start, I simply can't (or don't see how) pass config to JS with other mechanisms.

## Test Plan

0. Make sure `react-devtools` are not running
1. Open `RNTester/RNTester.xcodeproj`
2. Run application
3. See periodic spam in console like:
```
2018-01-16 00:39:40.801162+0300 RNTester[76181:13118879] [] nw_connection_get_connected_socket 11 Connection has no connected handler
2018-01-16 00:39:40.801606+0300 RNTester[76181:13118879] TCP Conn 0x60400016ff00 Failed : error 0:61 [61]
2018-01-16 00:39:42.819030+0300 RNTester[76181:13118882] [] nw_connection_get_connected_socket 12 Connection has no connected handler
2018-01-16 00:39:42.819472+0300 RNTester[76181:13118882] TCP Conn 0x60400016fa80 Failed : error 0:61 [61]
```
4. Open dev menu (press `Cmd+D`)
5. Press `Disable DevTools`
6. After app reloading there is no more spam in console.

This setting persists like other dev menu settings and can be enabled/disabled. By default DevTools are enabled to not break current behavior.

## Release Notes

[IOS] [ENHANCEMENT] [developer menu] - Automatic connection to React DevTools can be enabled/disabled from In-App Developer Menu


vovkasm:
This PR is actually not big! It simply removes `fishhook` which is not used anymore.

vovkasm:
@emilsjolander Now I have some code to continue discussion at https://github.com/facebook/react-native/issues/10027#issuecomment-321186109
Does this looks as acceptable workaround for you?

vovkasm:
Actually we at work support our own fork for RN and can rebase and/or improve this concrete change at any time (within 1-2 days). But I want to be sure that this work not be lost... So I will do anything only if requested from human, not a bot ;-)
PS: Our fork is open. Current branches:
RN 0.53: https://github.com/vovkasm/react-native/tree/v0.53.0-woof.0
RN 0.52: https://github.com/vovkasm/react-native/tree/v0.52.1-woof.0

mmmulani:
DevTools are a pretty useful feature though and it's reasonable to think that most people will have them enabled always.

if we fix the logspew, what benefit does this provide?

cpojer:
I don't think we'll want to support this change. It's rather complex and the majority of people won't really benefit from it. I'd much rather work on reducing the log spew than adding all this complexity to React Native.

PR #17619, opened 2018-01-15
[HMR] Fix HMR syntax error messages (`message` instead of `description`)

ide:
## Motivation
The code to display HMR errors on the client was reading the `description` field from Metro payloads. Metro does not include `description` in the body of its error payloads -- only in its `body.errors[]` items. This commit changes RN's HMR code to show `body.message` (set consistently with https://github.com/facebook/metro/pull/124) instead of the non-existent `body.description`.

## Test Plan
Open a test RN app, enable HMR, and then introduce a syntax error in an app source file. See that the redbox provides information about the syntax error instead of just saying "TransformError undefined".

## Related PRs
- https://github.com/facebook/metro/pull/124

## Release Notes

[GENERAL][ENHANCEMENT][HMR] - Fix display of syntax error messages when HMR is enabled

### DIFF
--- a/Libraries/Utilities/HMRClient.js
+++ b/Libraries/Utilities/HMRClient.js
@@ -121,7 +121,7 @@ Error: ${e.message}`
         }
         case 'error': {
           HMRLoadingView.hide();
-          throw new Error(data.body.type + ' ' + data.body.description);
+          throw new Error(`${data.body.type}: ${data.body.message}`);
         }
         default: {
           throw new Error(`Unexpected message: ${data}`);


ide:
@rafeca @mjesun The code in this PR hasn't been touched in over a year but since it uses Metro's error payload, Metro team members are probably the best people to look at this.

mjesun:
When I simplified `JSTransformer/index.js` in Metro [here](https://github.com/facebook/metro/commit/d4dcd4c6fa463c372bb92a960d2ef259f4437544#diff-0fe982fedea89f4de950e3ba9fe908e4L239), I removed the duality between `message` and `description`, standardizing over the first because of

* `description` is set to `''` when it's a generic error, and
* `message` is the standard one in the ECMA specification.

I then tried identifying other places where description was used, but I obviously not looked exhaustively 😅. Thanks for pointing out that one and fixing it!

PR #11788, opened 2017-01-09
Add accessibilityElementsHidden Prop To View Comp.

omeid:
accessibilityElementsHidden is used to control wheter elemetns contained
within a view are hidden from accessiblity services that query the
screen.

On iOS, it simply reflects UIKit's accessibilityElementsHidden, on
Android it is implemented with importantForAccessibility.

This commit also improves accessiblity of Navigation Experimental
CardStack by marking inactive (off screen) scenes invsible.

Closes #9725

PR #10918, opened 2016-11-14
UIViewManager: Add accessibilityElementsHidden

omeid:
UIKit `accessibilityElementsHidden` property is similar to
Android's `importantForAccessibility` and very important for
creating a good accessible experience by hidding off the screen
and invisible elements/views from VoiceOver and other assistive
technologies.

This commit also improves accessibility of Navigation Experimental
CardStack by marking inactive (off screen) scenes invisible for iOS
VoiceOver and Android's Talkback.

Closes #9725

PR #17627, opened 2018-01-16
Add accessibilityElementsHidden prop

aputinski:
## Motivation

Allow iOS to have similar accessibility functionality to Android. This PR exposes the `accessibilityElementsHidden` property on iOS which is similar to Android's `importantForAccessibility="no-hide-descendants"`

## Test Plan

I didn't see any existing examples for testing native props being passed through, but I did add an example to the RNTester app. I've attached some screenshots using the Accessibility Inspector to verify the property was correctly passed through.

a
b

## Related PRs

I've updated the website with appropriate documentation.

https://github.com/facebook/react-native-website/pull/141

screen shot 2018-01-16 at 10 23 50
screen shot 2018-01-16 at 10 23 59

## Release Notes

[IOS] [FEATURE] [View] - Added accessibilityElementsHidden property


shergin:
Well, this is really great.
Thank you for raising this topic! However, I think we have to land https://github.com/facebook/react-native/pull/10918 instead... finally.

@aputinski What do you think?

aputinski:
Looks like that was closed and we're now discussing https://github.com/facebook/react-native/pull/11788. I don't have a strong preference, but it seems like there is quite a bit of confusion about trying to unify this behavior. Would it be possible to merge this (since it's non-breaking) and then continue discussing unifying this property down the road?

@shergin @omeid

.gitignore
```diff
@@ -48,6 +48,7 @@ node_modules
 *.log
 .nvm
 /danger/node_modules/
+scripts/.packager.env
```
shergin: Why do we need this?

shergin:
@aputinski Yeah, that's reasonable. Even if I am going to land that PR as well, it's "safer" to land yours first.

omeid:
I think landing this is a mistake without understanding the android part. The reason I abandoned my initial PR was because no one from facebook actually showed interested and were willing to consider a practical solution.

I am still interested to see this feature landed so I don't have to maintain a fork that has some decent a18y.

.gitignore
aputinski: When I was running the build locally, that file was generated and I didn't want to commit any build artifacts. It's unrelated to my change and I can remove it if necessary.

PR #17628, opened 2018-01-16
Fix Xcode Archive task failing if project path contains whitespace

jevakallio:
## Motivation
Currently, Xcodebuild Archive task "Bundle React Native code and images" step fails if the project path contains whitespace characters. This is due to `react-native-xcode.sh` not escaping the local CLI path.

Ideally, of course, folks would use sane directory names, but this affects people in the community. I have seen this at a beginner hackathon, as well as Stack Overflow, e.g. https://stackoverflow.com/questions/48209829/cannot-create-offline-bundle-for-react-native-ios

### Verify build is currently broken
```
mkdir Bad\ Name
cd Bad\ Name
react-native init TestProject
cd TestProject
open ios/TestProject.xcodeproj
```
* Run `Build > Archive` 
* `cli.js bundle` command fails with "Cannot find module '.../Bad'"

### Verify fix works

* Patch `/node_modules/react-native/scripts/react-native-xcode.sh` with changed file
* Run `Build > Archive`
* Build succeeds

## Release Notes

 [IOS] [MINOR] [scripts/react-native-xcode.sh] - Escape directory paths with spaces


hramos:
A rebase is needed due to some changes that happened earlier this week in our internal monorepo outside of the synced react-native directory (but nonetheless prevent this PR from landing). Simply hitting Update branch should resolve it, let's see.

Issue #2814, opened 2015-09-17
Support third-party 64-bit libraries on Android


corbt:
Excerpt from the Android Developers Blog article, [Get your apps ready for the 64-bit requirement](https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html):
*[added by @hramos]*

### The 64-bit requirement: what it means for developers

Starting **August 1, 2019**:

* All new apps and app updates that include native code are required to provide 64-bit versions in addition to 32-bit versions when publishing to Google Play.
* Extension: Google Play will continue to accept 32-bit only updates to existing games that use Unity 5.6 or older until August 2021.

Starting **August 1, 2021**:

* Google Play will stop serving apps without 64-bit versions on 64-bit capable devices, meaning they will no longer be available in the Play Store on those devices.
* This will include games built with Unity 5.6 or older.

The requirement does not apply to:

* APKs or app bundles explicitly targeting Wear OS or Android TV, which are form factors that do not currently support 64-bit code.
* APKs or app bundles that are not distributed to devices running Android 9 Pie or later.

## Original Issue

Hey, it appears that React Native on Android doesn't provide a 64-bit version of the `libreactnativejni.so` native library, which can cause compatibility issues on 64-bit devices. I ran into this while attempting to integrate React Native with a large existing application I'm developing.

Steps to reproduce:
1. Create a new application with `react-native init`.
2. In the "android/app/build.gradle" file, add `compile 'io.realm:realm-android:0.82.2'` to the end of the `dependencies {}` block. (Note that I don't think this is an issue specifically caused by Realm, it's just an example dependency with a 64-bit native library).
3. Attempt to run on a Nexus 9. 

Stack trace:

```
E/AndroidRuntime(32171): FATAL EXCEPTION: main
E/AndroidRuntime(32171): Process: com.androidemberall, PID: 32171
E/AndroidRuntime(32171): java.lang.UnsatisfiedLinkError: could find DSO to load: libreactnativejni.so
E/AndroidRuntime(32171):    at com.facebook.soloader.SoLoader.loadLibraryBySoName(SoLoader.java:213)
E/AndroidRuntime(32171):    at com.facebook.soloader.SoLoader.loadLibrary(SoLoader.java:178)
E/AndroidRuntime(32171):    at com.facebook.react.bridge.JSCJavaScriptExecutor.<clinit>(JSCJavaScriptExecutor.java:19)
...
```

[...]

PR #17640, opened 2018-01-17
[WIP] Android 64-bit

govi:
## Motivation

Scratching my own itch #2814  to make sure react-native works on 64-bit Android devices. Using abifilters to restrict it to bundle only 32-bit .so files drops performance heavily on 64-bit devices.

I have attempted to tackle the issue. Much of it 'just' worked, but without being able to verify them with the full suite of tests, I will leave this as Work in Progress. I pored over https://facebook.github.io/react-native/docs/testing.html but it wasn't a lot of help. I'll leave the details of the errors from the PR description for now.

## Headline changes

1. Enabled the 64bit arm eabi and x86_64
2. Changed the CXX11 flags for folly & cxxreact
3. Upgraded to use the new soloader(0.3.0) that no longer restricts the so files to be copied.
4. Upgraded double-conversion to 1.1.6, for 64-bit support.
5. Updated generator templates

## Test Plan

1. Used gradle to make the .aar file and verified that it contains the .so files for all architectures ("armeabi-v7a", "x86", "x86_64", "arm64-v8a").
2. Used that aar as a gradle dependency to build my react native app and verified it on a Nexus 6p.
3. SoLoader copies the right files, and React Native screens work as they should. 

### Automated Tests

It's not ideal. Probably because this is the first time I have attempted to run these.

Android Unit Tests pass but having massive trouble getting the integration tests going. Its the same both on master and on my branch. Would like some help with it.

Gradle tests fail on master (at least on my setup) for all yoga tests as its unable to load yoga.so. We have the same failures on this branch as well.

There also an oddity with robolectric, where the jars are not getting copied to their correct locations. Had to do it manually.

Buck tests for iOS fail on snapshots for Scroll View. Looks like the scroll view test page now has a new component that is not there in the snapshots. But that component has loading animation, and it is not going to be possible to have a 100% match (it's probable, but the probability is low). Have the same issue on master.

## Other Considerations 

Finally, we also need the 64-bit enabled JSC. There are two options,

1. Is to resurrect this https://github.com/facebook/android-jsc/pull/19. The tar location is gone now, so the script has to be modified to checkout the repo at the correct revision. Its a known version and since there are no "tests", seems like a safer option for this PR. Which is what I had to do locally and pointed it to our internal mvn that has the 64-bit enabled JSC.

2. The other option is to https://github.com/SoftwareMansion/jsc-android-buildscripts. Seems to be what expo are using/planning to use. Of course, it needs more testing. Last I have seen is expo rolling it back 

## Related PRs

Nothing at the moment.

## Release Notes

[ANDROID] [BUGFIX] [ReactAndroid] - Enabled 64-bit support


gengjiawen:
Why bundle so  instead of generate from source?

govi:
@gengjiawen that was a mistake.. i have since cleaned it up. Would really like some help to complete this PR.

gengjiawen:
I think first we need to get https://github.com/facebook/android-jsc/pull/19 merged.

AaaChiuuu:
@ayc1 @mdvacca @axe-fb ex-FB Android Messenger people at Uber are asking for this.

govi:
@gengjiawen yup.. but we prolly have to update the PR. The tar ball location is 404. I had to checkout their SVN repo and up to the revision.

gengjiawen:
@hramos @janicduplessis Can you guys to push https://github.com/facebook/android-jsc/pull/19 get merged, we have been stuck with old jsc and lack of armeabi-64 support for a long time.

hovox:
This issue is super important for us. We have started RN integration into our app, and this may hold us back. Please let us know what we need to do to increase the priority of this issue, or if we can help somehow to get issue resolved.

hey99xx:
Can we close this now? New JSC has been merged into master branch and it's 64-bit architectures are enabled in Application.mk

PR #13326, opened 2017-04-05
Added option to disable xip

jkimbo:
## Motivation (required)

When running an app on a real iPhone in debug mode, I occasionally get issues when trying to load the js bundle and I suspect the issue is with xip.io (and it doesn't look like I'm the only one: https://github.com/facebook/react-native/issues/12786 https://github.com/facebook/react-native/issues/9688#issuecomment-272591839). So I've added the ability to optionally disable the use of xip.io if an env variable `DISABLE_XIP` is set.

## Test Plan (required)

Add `export DISABLE_XIP=true` to the `Bundle React Native code and images` build phase. Run the app on a real iPhone and ensure that it can load the JS bundle from the host computers IP.

PR #8091, opened 2016-06-13
Implemented automatic IP detection for iOS

nathanajah:
Implemented automatic IP detection for iOS, based on #6345 and #6362.
As the previous pull requests did, this works by writing the IP address of the host to a file.

Test plan:
`npm pack` on the commit and create a new project by doing `react-native init` using the generated package.
With both the computer and the device on the same wireless connection, run the generated project on the device.
Make sure that the device loads the bundle successfully from the packager.
Turn off the packager, and run the app again.
Make sure that the device loads the pre-packaged bundle successfully.

PR #17642, opened 2018-01-17
Removed use of xip.io, as it's not needed.

jvranish:
By default, when a react-native app launches in development mode on a physical iOS device, it attempts to load the JS bundle from a packager at `http://_your-local-ip-address_.xip.io:8081/`.

This change removes the use of `xip.io`, which changes that url to: `http://_your-local-ip-address_:8081/`

## Motivation

Background:
The automatic IP detection feature (introduced [here](https://github.com/facebook/react-native/pull/8091)) is super handy. However, it’s use of `xip.io` has caused myself and others much grief.  Some routers do not allow or have intermittent errors when trying to resolve DNS names to local IP addresses. This prompted the introduction of a [DISABLE_XIP feature](https://github.com/facebook/react-native/pull/13326), which helps. 

However, I don’t believe the use of `xip.io` is needed at all.

Based on [this comment](https://github.com/facebook/react-native/commit/8c29a52c54392ce52148e7d3aa9f835537453aa4#commitcomment-18224788), it appears the original reason for using `xip.io` was to “circumvent the numeric IP address limitation in ATS”.

But, the reason you can’t create ATS exceptions for raw IP addresses is that ATS is not enforced for raw IP addresses _at all_. You can read the Apple documentation [here](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html), the relevant portion is: 

> App Transport Security (ATS) applies only to connections made to public host names. The system does not provide ATS protection to connections made to:
> * Internet protocol (IP) addresses
> * Unqualified host names
> * Local hosts employing the .local top-level domain (TLD)

For example, in iOS, if you attempt to make an http request (note: _not_ https) to `http://www.google.com` you will get an error due to ATS. 
However, you can make the same request to `http://172.217.6.14/`  (which for me is the same server) and the request will succeed.

And in fact, if an ATS exception _was_ needed, the DISABLE_XIP feature shouldn’t work at all, but it does.

In short, using `xip.io` with ATS exceptions is unnecessary, causes some very annoying problems for some people, and I think it should just be removed. 

## Test Plan

Run the app on a physical iOS device and verify that it can load the JS bundle from the host computer's IP.

## Related PRs

[Implemented automatic IP detection for iOS #8091](https://github.com/facebook/react-native/pull/8091)
[Added option to disable xip #13326](https://github.com/facebook/react-native/pull/13326)

## Release Notes

 [INTERNAL] [BUGFIX] [./scripts] - Removed use of xip.io


janicduplessis:
Makes sense and test plan looks good! Thanks 👍 

@facebook-github-bot shipit

hramos:
Can you rebase and import again? There was an internal directory change that basically means any PR that is imported has to be rebased past a certain commit for it to land cleanly. Clicking on Update branch above should do it.

janicduplessis:
@facebook-github-bot shipit

Issue #17504, opened 2018-01-09
[RN 0.52] RCTImageLoader requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup`


fungilation:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

```
Environment:
  OS: macOS Sierra 10.12.6
  Node: 9.3.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: Not Found

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.52.0 => 0.52.0
```

### Steps to Reproduce

`$ react-native run-ios`

### Expected Behavior

iOS simulator run without errors.

### Actual Behavior

<img width="487" alt="screenshot 2018-01-09 14 53 14" src="https://user-images.githubusercontent.com/3803466/34747390-edde8370-f54c-11e7-954c-3d7615387bab.png">
<img width="1180" alt="screenshot 2018-01-09 15 02 18" src="https://user-images.githubusercontent.com/3803466/34747677-2fa5ef68-f54e-11e7-886d-b2a592756345.png">

^
```
Module RCTImageLoader requires main queue setup since it overrides `init` but doesn't implement `requiresMainQueueSetup`. In a future release React Native will default to initializing all native modules on a background thread unless explicitly opted-out of.
```

### Reproducible Demo

Any app using the Image component, although in my app I wrap that in a higher level component I use: `react-native-fit-image`.

I've never seen `RCTImageLoader` complain about `requiresMainQueueSetup` before. Only started now after I upgraded RN from 0.51 to 0.52.

rashtay:
For me, it doesn't show any warning in the simulator but appears as a warning in the browser console.

hramos:
PR #17679 is an attempt to fix this.

animaonline:
Started getting this in 0.54.0, is there a way to actually disable the warning?

This doesn't seem to work:

YellowBox.ignoreWarnings([
  'Warning: Module RCTImageLoader requires',
]);

fungilation:
because the warning has no "warning: " prefix. Take that out in your ignore

alzalabany:
can we merge this soon ?

hramos:
Merged

fungilation:
[...]

PR #17679, opened 2018-01-20
fix(RCTImageLoader): Adds `requiresMainQueueSetup` to fix v0.52

iRoachie:
## Motivation

In react-native v0.52 this warning shows:
```
"RCTImageLoader requires main queue setup 
since it overrides `init` but doesn't implement `requiresMainQueueSetup`".
```
This removes the warning by implementing `requiresMainQueueSetup` on RCTImageLoader

## Test Plan
* Create new react-native project 
* Enable Remote JS Debugging
* Should see no warning regarding RCTImageLoader requiring main queue setup 

## Related Issues
#17504 

## Release Notes

[IOS] [BUGFIX] [Libraries/Image/RCTImageLoader.m] - Implements `requiresMainQueueSetup`

### DIFF
--- a/Libraries/Image/RCTImageLoader.m
+++ b/Libraries/Image/RCTImageLoader.m
@@ -60,6 +60,11 @@ RCT_EXPORT_MODULE()
   return [self initWithRedirectDelegate:nil];
 }
 
++ (BOOL)requiresMainQueueSetup
+{
+    return NO;
+}
+
 - (instancetype)initWithRedirectDelegate:(id<RCTImageRedirectProtocol>)redirectDelegate
 {
     if (self = [super init]) {


janicduplessis:
Does this actually require main queue setup? All the init does is settings an ivar to nil. Might be a case where we can set requiresMainQueueSetup to NO as an optimisation. cc @shergin

shergin:
I agree with Janic @janicduplessis, but someone has to do the audit of this module. After one minute looking I think we should return `NO` but one minute is not enough.
I am totally not comfortable to make it main-threaded just because we don't know for sure.

janicduplessis:
@shergin Do  you know what the `RedirectDelegate` is? This isn't used at all in the repo so we can't really tell if it's safe or not without knowing the code that uses it. With the default init that sets it to nil it is safe for sure to init off the main queue.

iRoachie:
Confirming still a warning in 0.53

ide:
The assignment of the ivar should be able to run on any queue, whether or not the `RCTImageRedirectProtocol` implementations are thread-safe. The queues on which the redirect delegate is invoked are independent from `requiresMainQueueSetup` so it looks to me like we can correctly and safely return NO here.

iRoachie:
@ide Is that a confirmation to change it to NO?

ide:
I think it should be NO but would appreciate another analysis.

hramos:
Looks like @janicduplessis @shergin and @ide are all on the same page -- this should return NO.

iRoachie:
@hramos Updated

fkgozali:
Thanks for fixing this! Sorry for the delay in merging...

jose920405:
Still happening in `0.56`

iRoachie:
This does not show in 0.56 for me

jose920405:
captura de pantalla 2018-08-17 a la s 6 10 25 pm

iRoachie:
@jose920405 None of your warnings refers to RCTImageLoader. They may come from other components within the react-native core or any native components you may be using in your project. You may want to create issues in the respective projects.

PR #17697, opened 2018-01-22
GLog fix on case sensitive APFS macOS

hovox:
## Motivation

This will fix compile error on case sensitive APFS(Apple File System) macs. 

## Test Plan

Integrate  RN on case sensitive formatted macOS via cocoa pods you will get compile errors on the include lines something like :
```
#include <glog/logging.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
** BUILD FAILED **
```

If you change `#include <glog/logging.h>` into `#include <GLog/logging.h>` (replace `glog` with `GLog`) it will fix the build error. After fixing this you will get same kind of errors on a few other places.

## Release Notes

[IOS] [BUGFIX] [Framework] - `GLog` fix on case sensitive APFS macOS


third-party-podspecs/Folly.podspec
```diff
@@ -10,17 +10,17 @@ Pod::Spec.new do |spec|
...
-  spec.source_files = 'folly/Bits.cpp',
-                      'folly/Conv.cpp',
-                      'folly/Demangle.cpp',
-                      'folly/StringBase.cpp',
-                      'folly/Unicode.cpp',
-                      'folly/dynamic.cpp',
-                      'folly/json.cpp',
-                      'folly/portability/BitsFunctexcept.cpp',
-                      'folly/detail/MallocImpl.cpp'
+  spec.source_files = 'folly/Bits.{h,cpp}',
```
shergin: Why do we add `.h` files here?

hovox: We are creating static frameworks and using that instead of directly using pods. Headers are needed to create folly framework. Will revert it. For us would be good to have headers here. Please let me know if it will be merged, I will create a separate pull request with changed headers. Thanks.

hovox: To clarify, we are doing that for build performance.

PR #17712, opened 2018-01-22
Removing broken check for height auto-sizing

MattFoley:
I implemented this feature originally. 

Not sure exactly what happened, probably some squashing issue, or something changing over the years, but without removing these elements, the font always scales down as low as possible, rather than just low enough to fit the text into the frame in the view. 

Once they're removed, it works properly, and functions just like the native iOS adjustsFontSizeToFit.

With these changes, this is what we ship to Production. @shergin would you mind chiming in here?

## Motivation

This has a bug where the text shrinks forever without these changes.

## Test Plan

We ship this to production.

### DIFF
--- a/Libraries/Text/Text/RCTTextShadowView.m
+++ b/Libraries/Text/Text/RCTTextShadowView.m
@@ -28,7 +28,6 @@ NSString *const RCTReactTagAttributeName = @"ReactTagAttributeName";
 static NSString *const kShadowViewAttributeName = @"RCTShadowViewAttributeName";
 
 static CGFloat const kAutoSizeWidthErrorMargin  = 0.05f;
-static CGFloat const kAutoSizeHeightErrorMargin = 0.025f;
 static CGFloat const kAutoSizeGranularity       = 0.001f;
 
 @implementation RCTTextShadowView
@@ -45,7 +44,6 @@ static YGSize RCTMeasure(YGNodeRef node, float width, YGMeasureMode widthMode, f
 {
   RCTTextShadowView *shadowText = (__bridge RCTTextShadowView *)YGNodeGetContext(node);
   NSTextStorage *textStorage = [shadowText buildTextStorageForWidth:width widthMode:widthMode];
-  [shadowText calculateTextFrame:textStorage];
   NSLayoutManager *layoutManager = textStorage.layoutManagers.firstObject;
   NSTextContainer *textContainer = layoutManager.textContainers.firstObject;
   CGSize computedSize = [layoutManager usedRectForTextContainer:textContainer].size;
@@ -576,8 +574,7 @@ static YGSize RCTMeasure(YGNodeRef node, float width, YGMeasureMode widthMode, f
   textContainer.maximumNumberOfLines == 0;
 
   if (fitLines && fitSize) {
-    if ((requiredSize.width + (CGRectGetWidth(frame) * kAutoSizeWidthErrorMargin)) > CGRectGetWidth(frame) &&
-        (requiredSize.height + (CGRectGetHeight(frame) * kAutoSizeHeightErrorMargin)) > CGRectGetHeight(frame))
+    if ((requiredSize.width + (CGRectGetWidth(frame) * kAutoSizeWidthErrorMargin)) > CGRectGetWidth(frame))
     {
       return RCTSizeWithinRange;
     } else {


shergin:
Well, actually, I made some changes that I plan to land very soon which... kinda conflicts with this changes (like e.g. whole this feature was rewritten from scratch because of some reasons).
So, we should discuss this a week or so later.
I am sorry about this. And I am not 100% confident that my changes cover all possible cases (related to auto-resizing) and I hope you can help me to make it better.

MattFoley:
@shergin Did you add support for Android? They just added support internally it looks like: https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview.html#setting-textview-autosize

We use a home grown layer on top of RN using a native UI module to make this work on Android.

The whole idea is just that it works exactly like adjustsFontSizeToFit on UILabel on iOS, which will scale a text component's font size until it fits within the frame given to it by iOS. This is super useful for I18n management, (or any dynamic text really) where copy can be variable length. 

The native UILabel support is still way better than mine. My implementation doesn't handle multi-line as well, and sometimes just behaves in a way you wouldn't expect, shrinking a bit too far, etc.

shergin:
@MattFoley Nope, iOS only. But we (at Facebook) really need this feature on Android as well but don't have bandwidth to implement it. 😢 So, if you want/can add this for android, I will love to merge this.

MattFoley:
For Android, it'll probably be a bit before we could implement this for you, but I do think that you'd need to have ReactTextView.java start extending the app compat TextView.

motiz88:
Chiming in to mention I've started implementing this for Android a couple of times recently. It kinda worked the last time, but I got stuck on the way it should interact with layout. I may yet crack it on another attempt. (Would appreciate any pointers)

shergin:
@MattFoley Well, here it is 2716f53220f947c690d5f627286aad51313256a0! Could you please take a look at the new implementation of auto-sizing feature? 😳

shergin:
Feel free to open new PR. ❤️

MattFoley:
@shergin I haven't gotten a chance to checkout your update and get it running. I did however take a real deep dive into getting this to work on Android.

It's actually super easy to add it to a TextView. ReactTextView just needs to inherit from AppCompatTextView, and the RN version of the Android Support Library needs to be updated.

Here's the gist of it:
```java
//ReactTextAnchorViewManager.java
  @ReactProp(name = "adjustsFontSizeToFit", defaultBoolean = false)
  public void setAdjustsFontSizeToFit(ReactTextView view, boolean enabled) {
    view.setAdjustsFontSizeToFit(enabled);
  }

  @ReactProp(name = "minimumFontScale", defaultFloat = 0.1f)
  public void setMinimumFontScale(ReactTextView view, float scale) {
    view.setMinimumFontScale(scale);
  }

//ReactTextView.java
  public void setMinimumFontScale(float scale) {
    if (this.mMinimumFontScale != scale) {
      this.mMinimumFontScale = scale;
      this.updateFontScaling();
    }
  }

  public void setAdjustsFontSizeToFit(boolean enabled) {
    if (this.mAdjustsFontSizeToFit != enabled) {
      this.mAdjustsFontSizeToFit = enabled;
      this.updateFontScaling();
    }
  }

  public float getMinimumFontScale() {
    return (this.mMinimumFontScale != 0) ? this.mMinimumFontScale : DEFAULT_SCALE;
  }

  public void updateFontScaling() {
    if (this.mAdjustsFontSizeToFit) {
      int min = (int)(this.getTextSize() * this.getMinimumFontScale());
      int max = (int)this.getTextSize();

      TextViewCompat.setAutoSizeTextTypeUniformWithConfiguration(this, min, max, 1, TypedValue.COMPLEX_UNIT_PX);
    } else {
      TextViewCompat.setAutoSizeTextTypeWithDefaults(this, TextViewCompat.AUTO_SIZE_TEXT_TYPE_NONE);
    }
  }
```

Buuuuuuutttt, I think maybe the ReactTextView isn't actually what's drawn on screen? I get how the custom text drawing works on iOS, but I still don't fully understand this on Android. 

I assumed if I could get access to the ReactBaseTextShadowNode from the ReactTextView or ReactTextAnchorViewManager, I could just set mFontSize over there and tell it to redraw. But I couldn't figure out how to do that. If you point me in the right direction, we'll probably finish adding it.

linjson:
@MattFoley "TextViewCompat.setAutoSizeTextTypeUniformWithConfiguration" don't work in rn,right?

Issue #17618, opened 2018-01-15
0.52 - CORS issue with JS Remote Debugging when using xip.io


njbmartin:
### Is this a bug report?

Yes.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes.

### Environment

Environment:
  OS: macOS High Sierra 10.13.2
  Node: 9.4.0
  Yarn: Not Found
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.2.0 => 16.2.0
  react-native: ^0.52.0 => 0.52.0

Target Platform: iOS 8

### Steps to Reproduce

> Note: I upgraded to `0.52` last night using `react-native-git-upgrade`, it had been working previously on `0.51`.

1. Run development app on real iOS device, which loads the JS bundle from `http://192.168.0.13.xip.io:8081/index.delta?platform=ios&dev=true&minify=false`
2. Enable JS Remote debug tools, which opens `http://localhost:8081/debugger-ui/` in Chrome.
3. Bundle reloads and Chrome DevTools console displays the following error:

> Failed to load http://192.168.0.13.xip.io:8081/index.delta?platform=ios&dev=true&minify=false: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8081' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

### Expected Behavior

App should load correctly and not throw an error when trying to debug JS remotely.

### Actual Behavior

> Failed to load http://192.168.0.13.xip.io:8081/index.delta?platform=ios&dev=true&minify=false: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8081' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

### Reproducible Demo

(Paste the link to an example project and exact instructions to reproduce the issue.)

> Fresh react-native app using `react-native init`.

https://github.com/njbmartin/react-native-xip

njbmartin:
[...]

PR #17720, opened 2018-01-23
Prevents JS Debugger issues with CORS

njbmartin:
## Motivation

See #17618

On certain networks, `xip.io` is used, but as the debugger will always using localhost a change upstream has resulted in a CORS issue (see screenshots in #17618). This change ensures that the debugger will always open with whatever configuration.

> This should be merged in as a patch of `0.52.x` as it affects current release, and `0.51.x` didn't have this issue.

## Test Plan

Tested locally, could do with the people having the same issue in #17618 testing it out prior to merging.

## Related PRs

## Release Notes

[CLI] [BUGFIX] [local-cli/server/middleware/getDevToolsMiddleware.js] - Prevents JS Debugger issues with CORS


local-cli/server/middleware/getDevToolsMiddleware.js
```diff
@@ -14,8 +14,8 @@ const launchChrome = require('../util/launchChrome');
 
 const {exec} = require('child_process');
 
-function launchChromeDevTools(port, args = '') {
-  var debuggerURL = 'http://localhost:' + port + '/debugger-ui' + args;
+function launchChromeDevTools(host, args = '') {
+  var debuggerURL = 'http://' + host + '/debugger-ui' + args;
```
cagataycali: Why eslint does not throw error when using "var" ?

local-cli/server/middleware/getDevToolsMiddleware.js
```diff
@@ -39,12 +39,13 @@ function launchDevTools({port, projectRoots}, isChromeConnected) {
...
   } else if (!isChromeConnected()) {
     // Dev tools are not yet open; we need to open a session
-    launchChromeDevTools(port);
+    launchChromeDevTools(host);
   }
 }
 
 module.exports = function(options, isChromeConnected) {
   return function(req, res, next) {
+    var host = req.headers.host;
```
cagataycali: Same here.  https://github.com/facebook/react-native/pull/17720/files#r163449140

local-cli/server/middleware/getDevToolsMiddleware.js
```diff
@@ -62,7 +63,7 @@ module.exports = function(options, isChromeConnected) {
       launchDevTools(options, isChromeConnected);
       res.end('OK');
     } else if (req.url === '/launch-js-devtools') {
-      launchDevTools(options, isChromeConnected);
+      launchDevTools(host, options, isChromeConnected);
```
wd: I think maybe here should be
```
      launchDevTools({...options, host}, isChromeConnected);
```

wd: This PR works here after this modify.

njbmartin: You're right. Messed up the copy pasta

local-cli/server/middleware/getDevToolsMiddleware.js
ide: Later it might be nice to honor the protocol too (ex: setting up an HTTPS reverse proxy in front of the packager server) and to use something like Express’s approach to inferring the protocol: https://github.com/expressjs/express/blob/master/lib/request.js#L306

njbmartin:
> Thanks for the PR! It’d be nice to see more passing tests (looks like this PR is unrelated to the failures but would prefer to see more green). Would you mind waiting a bit till master has more tests passing and then rebasing?

Absolutely. I thought it was me, but every PR is getting ❌

vlkosinov:
Perfectly working with iOS device

@njbmartin thank you!

njbmartin:
That reminds me, I've not tested it on Android.

cagataycali:
Also rebase needed, four commits for one PR is not eligible.

njbmartin:
I take it squash and merge isn't enabled?

cagataycali:
Instead `git rebase head~4`[rebase doc](https://git-scm.com/docs/git-rebase)

janicduplessis:
No need to squash, it will land as a single commit when the PR gets imported.

local-cli/server/middleware/getDevToolsMiddleware.js
janicduplessis: There's still a lot of files that are not converted to es6 so the rule is not enabled.

janicduplessis:
@njbmartin Just landed a commit that removes xip.io, is this still needed?

ide:
It'd be nice to have this in general for xip.io, ngrok, etc.

ide:
This fixes and issue blocking several people and looks safe - let’s try it. 

@facebook-github-bot shipit

afilp:
I see this closed, won't be merged eventually? We are having lots of problems with inability to debug.

jackwilsdon:
It looks like this was merged into 0.53-stable in df8da3b6c4038c066e12530fc66a1563bd3f58b7.

PR #17739, opened 2018-01-25
Support received intent in Linking class

koukch:
## Motivation

Allow react native apps to handle received intent that have action=ACTION_SEND

## Test Plan

It's just upgrade

## Related PRs

None

## Release Notes
[ANDROID] [FEATURE] [Linking] - add support to receive intent that have action=ACTION_SEND


mkonicek:
Hi! It would help reviewers to understand the use case (why this PR). Why do you need to send this intent to your app?
Looks like it's designed for sending data to the (React Native) app but we don't read the data from the Intent `extras`.

https://developer.android.com/training/sharing/send.html

cpojer:
Closing this as there hasn't been any comment or activity by the person who created the PR with an explanation of why this change is necessary.

rohit-ganapathy:
I cannot stress enough how important this PR Is. I  honestly don't get why the need for this feature is lost on you'll.

PR #17745, opened 2018-01-25
Enable platforms to configure CLI commands

rozele:
This change adds hooks via the `package.json` for a platform plugin to specify hooks for generating platform configurations. This change also adds hooks to allow platform plugins to participate in `link` and `unlink` commands. The goal is to move platform-specific code for platform plugins into their own repositories / node_modules.

## Motivation

We need to be able to configure the CLI commands for plugin platforms (e.g., react-native-windows) in their own repositories.

## Test Plan

- All jest tests, including new tests, pass.
- `link` and `unlink` commands are successful on iOS and Android.

## Related PRs

https://github.com/Microsoft/react-native-windows/pull/1601

## Release Notes

[CLI][FEATURE][local-cli/core/index.js] - Allow platform plugins to contribute to the RNConfig.
[CLI][FEATURE][local-cli/link/link.js] - Allow platform plugins to participate in the `link` command.
[CLI][FEATURE][local-cli/link/unlink.js] - Allow platform plugins to participate in the `unlink` command.


grabbou:
This is great! Looking into the code *right now*, but I think the ability to extend built-in commands is another big step towards decoupling CLI from being android / ios only.

local-cli/core/findPlugins.js
```diff
@@ -52,26 +60,30 @@ const findPluginInFolder = (folder) => {
...
  * @param {String} folder Path to the folder to get the package.json from
- * @type  {Array}         Array of plugins or an empty array if no package.json found
+ * @type  {Object}        Object of commands and platform plugins
  */
 module.exports = function findPlugins(folders) {
-  return uniq(flatten(folders.map(findPluginInFolder)));
+  const plugins = folders.map(findPluginInFolder);
+  return {
+    commands: uniq(flatten(plugins.map(p => flatten(p.commands)))),
+    platforms: uniq(flatten(plugins.map(p => flatten(p.platforms))))
```
grabbou: I am not sure I am getting that line - the `commands` flattening is to make a flat array of all commands (if plugin defines many, we handle that).

What's the purpose of it for `platforms`? I get that we want to get a list of *all* platforms defined throughout the plugins - but when do we use it later?

local-cli/core/index.js
```diff
@@ -70,35 +84,51 @@ const defaultRNConfig = {
         );
       });
 
-    return flatten(plugins);
+    return flatten(commands);
+  },
+
+  getPlatformConfig(): Object {
```
grabbou: Looks like this can be just a function, so that no need to call `this....` later to refer to it?

local-cli/link/link.js
```diff
@@ -76,31 +74,40 @@ const linkDependencyAndroid = (androidProject, dependency) => {
...
-      dependency.config.windows,
-      params,
-      windowsProject
-    );
-
-    log.info(`Windows module ${dependency.name} has been successfully linked`);
-  });
+const linkDependencyPlatforms = (platforms, project, dependency) => {
+  const ignorePlatforms = ['android', 'ios'];
+  Object.keys(platforms || [])
```
grabbou: Okay, so the approach you have taken here is to explicitly do platform `link`. How about making it so that when plugin defines `link` command, it will be run like this:
```js
originalLink().then(linkOfPlugin)
```
That way, we could not only allow `react-native-windows` to hook into `link` but potentially `init` and others. It would be also a bit less hardcoded then the following code.

local-cli/link/link.js
```diff
@@ -76,31 +74,40 @@ const linkDependencyAndroid = (androidProject, dependency) => {
...
-      dependency.name,
-      dependency.config.windows,
-      params,
-      windowsProject
-    );
-
-    log.info(`Windows module ${dependency.name} has been successfully linked`);
-  });
+const linkDependencyPlatforms = (platforms, project, dependency) => {
+  const ignorePlatforms = ['android', 'ios'];
```
grabbou: Is there any reason why `android` and `ios` can't have the same code path?

rozele: They can, I just wanted to keep the diff small.

local-cli/core/findPlugins.js
rozele: Not sure if the question here is related to `platforms` and why it's mangled with `commands`, or specifically why we `flatten` the `platforms` collection, so I'll address both.

`platforms` is used by `local-cli/core/index.js` and the `link` and `unlink` command to evaluate `projectConfig()`, `dependencyConfig()`, and `linkConfig()`. I would have created a separate module to extract the platform info, but I didn't want to impact CLI perf. If I created a separate module, I would have to read `package.json` for each potential plugin twice (instead of once as is done here).

In terms of why we flatten the collection of platforms - plugins could potentially define multiple platforms (e.g., uwp and win32, or ios and android), so I flatten the list so we have one array of all platforms used for configuration and in `link`.

local-cli/core/index.js
rozele: True. I'll fix.

local-cli/link/link.js
rozele: @grabbou - for most commands there wouldn't be any problems with this.  `commander` doesn't enforce that one command to one key. E.g., I could easily define `link` in the `react-native-windows` project, and `commander` would just execute both the `link` from `react-native` and from `react-native-windows`. *However* - the semantics of `link` are as follows:
1. Run prelink scripts
2. Run link for all platforms
3. Run postlink scripts

If I did `originalLink().then(linkOfPlugin)`, we would probably need to breakdown prelink and postlink per platform as well (not the worst idea), e.g.:
1. prelink for iOS
2. link iOS
3. postlink for iOS
4. prelink for Android
...

That is a much more involved change that I didn't want to undertake.

Also - `init` is a very difficult case because it's actually implemented partially by the `react-native-cli` global tool, and we currently do some heuristics to match the version of `react-native-windows` to the current installed version of `react-native`.  Not saying its impossible, but I think `init` will require some other infrastructure. For other commands, there's no reason why we can't just define another `upgrade`, `eject`, `info` and just have it work as you propose.

local-cli/core/index.js
rozele: #WontFix - @grabbou, we need the expose the `getPlatformConfig()` on the `RNConfig` so that it can be used later by other commands (like `link`).

local-cli/link/link.js
rozele: #WontFix - There's currently some custom linking logic related to `podfile` for iOS.  Let's do these changes in two phases:
1. Remove `windows` and support platform extensions.
2. Treat iOS and Android as any platform extension.

local-cli/core/findPlugins.js
```diff
@@ -37,11 +37,19 @@ const findPluginsInReactNativePackage = (pjson) => {
   return path.join(pjson.name, pjson.rnpm.plugin);
 };
 
+const findPlatformsInPackage = (pjson) => {
+  if (!pjson.rnpm || !pjson.rnpm.plugin) {
```
rozele: plugin -> platform

local-cli/link/unlink.js
```diff
@@ -94,10 +107,12 @@ const unlinkDependencyIOS = (iOSProject, dependency, packageName, iOSDependencie
...
   const packageName = args[0];
 
-  var project;
-  var dependency;
+  let platforms;
+  let project;
+  let dependency;
 
   try {
+    platforms = config.getPlatformConfig();
```
rozele: @grabbou - here is one example of where we need to call `getPlatformConfig()` on the `RNConfig`.

local-cli/core/findPlugins.js
rozele: @grabbou - I removed the inner `flatten`. The problem was that I had switched from `concat` to `push`.

rozele:
@grabbou by custom platform-specific code, you're referring to the iOS and Android stuff, right? I really think the best way to move forward is incremental change (i.e., fix iOS and Android in a separate PR). It's safer / gives us more confidence we're not breaking iOS and Android and keeps the diff small. cc @Kureev

rozele:
cc @hramos (I think you said you were interested in this PR).

grabbou:
@rozele true, I am just afraid that with no-one being interested in getting the codebase into its intended final shape, `link` will be left half-way done for a long time.

I have a plan on how to make this more "clean", but with you, being the only `react-native-x` I am aware (a platform), I am not sure if it's still a valid concern.

I guess we can merge it as is having another pair of the eyes to look at it.

rozele:
Hey @grabbou - this is no longer half-baked, can we merge without a second opinion?

grabbou:
Looks good to me - let's ship and I'll check before cherry-picking

grabbou:
@facebook-github-bot shipit

hramos:
Can you rebase? Nothing wrong with your PR per se, we just had a problematic commit land internally early last week. In order to land cleanly, the patch from GitHub must be based on a revision created later than roughly Monday of last week.

rozele:
@hramos I rebased, is there anything else I need to do to trigger the import again?

hramos:
Anyone with access using the shipit command should trigger it. I'll trigger one manually now.

hramos:
This PR did not land as it could not be applied cleanly. Looks like this still needs to be rebased. Can you try again?

Kureev:
Rebase (or let me know if I can do it for you) and ping me, I'll shit it 😉

rozele:
Thanks @Kureev and @hramos - I just rebased.

Kureev:
@facebook-github-bot shipit

ethanx94:
Any idea when this will land? Not having a functional link for UWP apps is a real downer.

grabbou:
@ethanx94 this is out as a `0.54.0-rc.0` right now - I'll be releasing patches in a backwards compatible manner tomorrow.

PR #17762, opened 2018-01-28
 Support sticky headers for inverted Lists

janicduplessis:
## Motivation

Sticky headers for inverted lists should still stick at the top of the list instead of the bottom.

## Test Plan

Tested by adding the inverted prop to the SectionList example in RNTester.

## Related PRs

It does add a prop to ScrollView but it's very specific to the inverted list implementation, not sure if it should be documented.

## Release Notes

[GENERAL][ENHANCEMENT][LISTS] -  Support sticky headers for inverted Lists

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -191,6 +191,11 @@ const ScrollView = createReactClass({
       'white',
     ]),
     /**
+     * If sticky headers should stick at the bottom instead of the top of the
+     * ScrollView. This is usually used with inverted ScrollViews.
+     */
+    invertStickyHeaders: PropTypes.bool,
+    /**
      * When true, the ScrollView will try to lock to only vertical or horizontal
      * scrolling while dragging.  The default value is false.
      * @platform ios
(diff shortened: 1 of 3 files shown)


Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -676,6 +684,15 @@ const ScrollView = createReactClass({
     this.scrollResponderHandleScroll(e);
   },
 
+  _handleLayout: function(e: Object) {
+    if (this.props.invertStickyHeaders) {
```
janicduplessis: Fast path here since we only need the height of the ScrollView for the inverted case which is pretty rare.

janicduplessis:
@sahrens Might need to re-import there was a flow error.

janicduplessis:
@sahrens Now I'm waiting for that blog post 😄

terrysahaidak:
Hi @janicduplessis, now headers stick for the top instead of the bottom, but still, they are footers but not headers. Should I do some extra setup to fix it?

brunoro:
Is this making it into upstream anytime soon? I just stumbled across this exact use case.

DinoZenka:
> Hi @janicduplessis, now headers stick for the top instead of the bottom, but still, they are footers but not headers. Should I do some extra setup to fix it?

Did you found a solution how to make sticky headers stick to the top in inverted section list?

Issue #16661, opened 2017-11-03
DX bug: ExceptionsManager does not log root exception cause


slorber:
### Is this a bug report?

Yes (somehow) this is a developer experience bug for me but it's subjective.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

I'm not using RN directly but it's plain Expo SDK 22 / CRNA

I don't think it's relevant but here are the dependencies:

```
 "dependencies": {
    "@mapbox/polyline": "^0.2.0",
    "@storybook/addon-actions": "^3.2.14",
    "@storybook/addon-links": "^3.2.14",
    "babel-plugin-module-resolver": "^2.7.1",
    "expo": "^22.0.0",
    "glamorous-native": "^1.2.0",
    "intl": "^1.2.5",
    "left-pad": "^1.1.3",
    "lodash": "^4.17.4",
    "moment": "^2.18.1",
    "murmurhash3js": "^3.0.1",
    "native-base": "^2.3.3",
    "polished": "^1.8.1",
    "prop-types": "^15.6.0",
    "react": "16.0.0-beta.5",
    "react-intl": "^2.4.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-22.0.1.tar.gz",
    "react-native-autolink": "^1.1.1",
    "react-native-calendars": "^1.9.0",
    "react-native-collapsible": "^0.9.0",
    "react-native-gallery": "^0.0.17",
    "react-navigation": "^1.0.0-beta.11",
    "react-redux": "^5.0.5",
    "recompose": "^0.23.5",
    "redux": "^3.6.0",
    "redux-actions": "^2.0.3",
    "redux-batched-actions": "^0.1.6",
    "redux-batched-subscribe": "^0.1.6",
    "redux-logger": "^3.0.6",
    "redux-saga": "^0.15.5",
    "redux-thunk": "^2.2.0",
    "reselect": "^3.0.1",
    "sentry-expo": "^1.6.0",
    "uuid": "^3.0.1"
  },
  "devDependencies": {
    "@storybook/cli": "^3.0.1",
    "@storybook/react-native": "^3.0.1",
    "babel-eslint": "^7.2.3",
    "babel-plugin-transform-remove-console": "^6.8.5",
    "babel-preset-react-optimize": "^1.0.1",
    "eslint": "^3.19.0",
    "eslint-import-resolver-reactnative": "^1.0.2",
[...]

Issue #16205, opened 2017-10-04
Feature request: useful stack traces for 'cannot read property 'x' of undefined'


jslz:
I would like the stack traces to point to my application code. Apparently it only ever points to internal JS engine type code?

react-native-cli: 2.0.1
react-native: 0.46.4

```
ExceptionsManager.js:63 Cannot read property 'explore' of undefined
handleException @ ExceptionsManager.js:63
handleError @ InitializeCore.js:125
reportFatalError @ error-guard.js:44
__guard @ MessageQueue.js:230
callFunctionReturnFlushedQueue @ MessageQueue.js:100
(anonymous) @ debuggerWorker.js:71

ExceptionsManager.js:71 Unhandled JS Exception: Cannot read property 'explore' of undefined
reactConsoleErrorHandler @ ExceptionsManager.js:71
console.error @ YellowBox.js:67
logToConsole @ RCTLog.js:43
logIfNoNativeHook @ RCTLog.js:29
__callFunction @ MessageQueue.js:260
(anonymous) @ MessageQueue.js:101
__guard @ MessageQueue.js:228
callFunctionReturnFlushedQueue @ MessageQueue.js:100
(anonymous) @ debuggerWorker.js:71
```

slorber:
@jslz I created a new issue for that: https://github.com/facebook/react-native/issues/16661

PR #17771, opened 2018-01-29
DX: log full js errors instead of just error messages (ExceptionsManager)

slorber:
## Motivation

This is really annoying DX to not have full stacktrace of errors in Chrome developer tools.

This kind of error is totally unactionnable without deeper debugging, as we don't have a single stacktrace, while it could be made actionnable easily

image

Some issues already have been opened regarding this problem, yet they did not get much attention and got closed by stale bot.

https://github.com/facebook/react-native/issues/16661
https://github.com/facebook/react-native/issues/16205

## Test Plan

Errors are logged appropriately in Chrome developer tools with full stacktrace

## Release Notes

[ENHANCEMENT][MINOR][ExceptionsManager.js] Log full errors instead of just error messages

## Other

I must admin I may not see all the implications of why the message did got logged instead of the full exception in the first place. 

If you know why and there's a good reason, please tell me and we can probably do something better than current state to help people debug RN apps.

### DIFF
--- a/Libraries/Core/ExceptionsManager.js
+++ b/Libraries/Core/ExceptionsManager.js
@@ -62,9 +62,9 @@ function handleException(e: Error, isFatal: boolean) {
     e = new Error(e);
   }
   if (console._errorOriginal) {
-    console._errorOriginal(e.message);
+    console._errorOriginal(e);
   } else {
-    console.error(e.message);
+    console.error(e);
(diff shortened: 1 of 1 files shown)


slorber:
not sure this actually solves the problem, I'm closing as I don't really know how to solve it anyway

PR #17774, opened 2018-01-29
[Android] Add support annotations dependency

hramos:
Should fix current test-android error in Circle CI.


hramos:
Landing, as this has fixed the buck issue present in test-android.

PR #17789, opened 2018-01-30
Fix: CUICatalog: Invalid asset name supplied: '(null)'

atkit:
## Motivation

if imageName is nil in (RCTImageFromLocalAssetURL) we should not call [UIImage imageNamed].
For example if it gets file://url from of some cached image.

## Test Plan

I checked that RNTester still works and shows all images. 
I also checked the case when cached image is shown from documents folder it does not show warning. 

## Related PRs

It does not change functionality, it only removes bunch of warnings on iOS when using cached images through library for example: react-native-img-cache

## Release Notes

[IOS] [BUGFIX] [File] - Fixed warning: CUICatalog: Invalid asset name supplied: '(null)'

### DIFF
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -698,10 +698,13 @@ UIImage *__nullable RCTImageFromLocalAssetURL(NSURL *imageURL)
   }
 
   UIImage *image = nil;
-  if (bundle) {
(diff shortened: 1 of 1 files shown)


atkit:
Signed CLA

atkit:
rebased

atkit:
@shergin is there anything necessary to make it happen?

PR #17792, opened 2018-01-30
[Circle] Store JUnit artifacts

hramos:
Should help visualize failures more easily.


hramos:
Will merge as part of #17794

PR #17795, opened 2018-01-31
Allow specifying location of metro cache dir.

anp:
## Motivation

When many users try Expo or CRNA for the first time, they often have to wait for a few minutes on slower hardware for the initial Metro cache to be populated. I'd really like for us to be able to include a prewarmed transform cache in the template projects to further decrease the time it takes for someone to experience their first view rendered by React Native.

## Test Plan

A few weeks ago when I wrote this patch, I:

1. ran `react-native start --cache-dir ./.cache` and left it running
2. ran `curl http://localhost:8081/index.ios.bundle`
3. ran `curl http://localhost:8081/index.android.bundle`

The cache directory I specified was correctly populated and subsequent packager runs with and without the `--cache-dir` flag had performance in line with what I expected re: using or not using a warmed cache.

Note that I've since rebased against current master and my test project is having issues resolving the haste map, but I'm fairly certain that's unrelated to these changes as it's also happening for me on master. ¯\_(ツ)_/¯

## Open Question

The way that `run-android` and `run-ios` set up the server is significantly more complicated to change, and since I expect this to be primarily useful in situations where a tarball template is provided (i.e. Expo, where run-android and run-ios aren't used) I don't see a huge advantage to adding this flag there as well. That said, I can see an argument in favor of including those commands as well for consistency's sake. Let me know if you think that should block merging and I'll try it out.

## Release Notes

[CLI] [MINOR] [local-cli] - Added CLI arg to specify transform cache directory.


hramos:
FYI, this is diff has not landed because the touched files are failing Flow, though it seems the issues were not introduced by this PR itself. For example (fb-internal filepaths truncated, and keep in mind this is a monorepo, therefore the reference to metro):

> Error: local-cli/bundle/buildBundle.js line 95 col 29-3: object literal
> This type is incompatible with the expected param type of
metro/packages/metro/src/Server/index.js line 111 col 24-30: Options

Line 95 was not touched by this PR. The CLI is not used internally, so we probably shouldn't even be blocking for Flow failures in these files anyway. I'll come back to this PR later today.

anp:
@hramos gentle ping -- I know you've been doing some work on CI stuff lately, does it seem like this might be landable now?

anp:
@hramos I fixed the merge conflict.

hramos:
Importing again, I apologize for the delay getting this merged!

local-cli/server/runServer.js
```diff
@@ -190,7 +191,8 @@ function getPackagerServer(args, config, reporter) {
     resetCache: args.resetCache,
     sourceExts: args.sourceExts.concat(defaultSourceExts),
     transformModulePath: transformModulePath,
-    transformCache: TransformCaching.useTempDir(),
+    transformCache: args.cacheDir || TransformCaching.useTempDir(),
+    useDeltaBundler: false,
```
anp: @hramos I didn't see this before, but I suspect this might break some of the Metro team's plans. Will it break the import if I edit again?

hramos: Go for it - the PR did not land, anyway.

anp:
OK, ready to go again. Sorry for the runaround.

PR #17797, opened 2018-01-31
tvOS: TV nav event emitter should check for bridge

dlowder-salesforce:
## Motivation

When running with the packager in the tvOS simulator, reloading from the packager hits an assert in `RCTEventEmitter`, causing a crash.  The solution is for `RCTTVNavigationEventEmitter` to check for the existence of the bridge before attempting to send an event.

## Test Plan

Manual testing.

## Release Notes

[IOS] [BUGFIX] [RCTTVNavigationEventEmitter.m] - Fix crash when reloading in tvOS


dlowder-salesforce:
@shergin please review

hramos:
We finally got tests back to green yesterday. If you don't mind, can you rebase? That way we can get these failing checks back to green and ensure this PR doesn't introduce any issues.

(We still need the rebase due to some changes we made to the internal repo, anyway)

dlowder-salesforce:
@hramos rebase done, tests are all green except for analyze_pr

PR #17809, opened 2018-02-01
Bundle download progress on Android

janicduplessis:
## Motivation

Android equivalent of #15066

## Test Plan

Tested that download progress shows up properly when reloading the app.

## Release Notes

[ANDROID] [FEATURE] [DevSupport] - Show bundle download progress on Android

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
@@ -28,12 +28,16 @@ public class MultipartStreamReader {
   private final String mBoundary;
   private long mLastProgressEvent;
 
-  public interface ChunkCallback {
-    void execute(Map<String, String> headers, Buffer body, boolean done) throws IOException;
-  }
+  public interface ChunkListener {
+    /**
+     * Invoked when a chunk of a multipart response is fully downloaded.
+     */
+    void onChunkComplete(Map<String, String> headers, Buffer body, boolean isLastChunk) throws IOException;
 
-  public interface ProgressCallback {
-    void execute(Map<String, String> headers, long loaded, long total) throws IOException;
+    /**
+     * Invoked as bytes of the current chunk are read.
+     */
+    void onChunkProgress(Map<String, String> headers, long loaded, long total) throws IOException;
   }
 
   public MultipartStreamReader(BufferedSource source, String boundary) {
@@ -60,23 +64,23 @@ public class MultipartStreamReader {
     return headers;
   }
 
-  private void emitChunk(Buffer chunk, boolean done, ChunkCallback callback) throws IOException {
+  private void emitChunk(Buffer chunk, boolean done, ChunkListener listener) throws IOException {
     ByteString marker = ByteString.encodeUtf8(CRLF + CRLF);
     long indexOfMarker = chunk.indexOf(marker);
     if (indexOfMarker == -1) {
-      callback.execute(null, chunk, done);
+      listener.onChunkComplete(null, chunk, done);
     } else {
       Buffer headers = new Buffer();
       Buffer body = new Buffer();
       chunk.read(headers, indexOfMarker);
       chunk.skip(marker.size());
       chunk.readAll(body);
-      callback.execute(parseHeaders(headers), body, done);
+      listener.onChunkComplete(parseHeaders(headers), body, done);
     }
   }
 
-  private void emitProgress(Map<String, String> headers, long contentLength, boolean isFinal, ProgressCallback callback) throws IOException {
-    if (headers == null || callback == null) {
+  private void emitProgress(Map<String, String> headers, long contentLength, boolean isFinal, ChunkListener listener) throws IOException {
+    if (headers == null || listener == null) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


janicduplessis:
Sadly this wont work currently since progress is disabled on android (https://github.com/facebook/react-native/commit/33057aacbf5ff6cd4d0e44eeddf9c328574883b4#diff-948206404ada96979d687d628e935ee1), but we've reverted this commit in the Expo fork so it would be great to merge it anyway :)

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
```diff
@@ -113,7 +148,10 @@ public boolean readAllParts(ChunkCallback callback) throws IOException {
         Buffer chunk = new Buffer();
         content.skip(chunkStart);
         content.read(chunk, length);
+        emitProgress(currentHeaders, chunk.size() - currentHeadersLength, true, progressCallback);
```
ide: true -> isCloseDelimiter?

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
```diff
@@ -113,7 +148,10 @@ public boolean readAllParts(ChunkCallback callback) throws IOException {
...
         emitChunk(chunk, isCloseDelimiter, callback);
+        currentHeaders = null;
```
ide: clear these only when we return (or don't clear at all) so we don't always need to reparse?

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
```diff
@@ -98,6 +119,20 @@ public boolean readAllParts(ChunkCallback callback) throws IOException {
...
+          long indexOfHeaders = content.indexOf(headersDelimiter, searchStart);
+          if (indexOfHeaders >= 0) {
+            mSource.read(content, indexOfHeaders);
+            Buffer headers = new Buffer();
+            content.copyTo(headers, searchStart, indexOfHeaders - searchStart);
+            currentHeadersLength = headers.size();
+            currentHeaders = parseHeaders(headers);
+          }
+        } else {
+          emitProgress(currentHeaders, content.size() - currentHeadersLength, false, progressCallback);
```
ide: Instead of currentHeadersLength (assuming we want to calculate the size of the response body downloaded so far), don't we want to track the starting index of the response body?

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
janicduplessis: Each chunk of the multipart response has it's own headers so this is why we clear it every time we emit a chunk so we parse the next one.

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
janicduplessis: We do emit progress for each chunk so even if it is not the last chunk (what isCloseDelimiter means) we still want to pass isFinal to true. It's only use is to make sure the last event isn't throttled so we see 100%.

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
janicduplessis: Yea there might be an error here, we don't take into consideration the length of the headers delimiter. I could keep currentHeadersLength and do `content.size() - currentHeadersLength - headersDelimiter .size()` or change currentHeadersLength for something like contentStart.

ReactAndroid/src/main/java/com/facebook/react/devsupport/MultipartStreamReader.java
ide: If I understand this correctly, we'd reach 100% only on the last chunk -> don't we want to force the progress event only in that scenario then? Otherwise I'd say `isFinal` should be named something else if it gets set to true many times and not just for the final progress event.

janicduplessis: No it is not the case, progress is reset for each chunk so let's say we have three chunks progress would look like: 0, 1, 0, 1, 0, 1. In the case of RN packager we only care about the progress of the last chunk (the JS bundle so we ignore progress of the other ones (https://github.com/facebook/react-native/pull/17809/files/bbe67864421b1de228d7af3ddd7327ed8bca060c#diff-948206404ada96979d687d628e935ee1R204).

janicduplessis: Just to clarify a chunk is a "part" from the multipart response. You can pretty much see it as multiple http responses nested inside the normal http response. It has it's own headers and we track progress of downloading each chunk separately.

ide: I think I get it. I thought that the progress was for the entire multipart response (e.g. like a multipart upload) rather than for just the chunk.

What do you think about refactoring the ChunkCallback interface so that it's clear the progress events are just for that one chunk? (Breaking change to the API)

```java
  public interface ChunkListener {
    /**
     * Invoked when a chunk of a multipart response is fully downloaded.
     */
    void onChunkComplete(Map<String, String> headers,
                         Buffer body,
                         boolean isLastChunk) throws IOException;

    /**
     * Invoked as bytes of the current chunk are read and may be invoked multiple times per chunk.
     * This method is guaranteed to be invoked one last time when the chunk is fully downloaded,
     * with `isChunkComplete` set.
     */
     void onChunkProgress(Map<String, String> headers,
                          long chunkBytesRead,
                          long chunkTotalBytes,
                          boolean isChunkComplete) throws IOException;
  }
```

janicduplessis: Makes sense, I agree it will be clearer like that, not sure if we have to mark this as breaking since I don't think it is meant to be public api.

ide:
@facebook-github-bot shipit

Issue #15497, opened 2017-08-15
BackHandler and remote debug


sm2017:
### Is this a bug report?

(Yes)

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

(Yes)

### Environment

1. `react-native -v`:
```
react-native-cli: 2.0.1
react-native: 0.47.1
```

2. `node -v`:
```
v8.1.3
```

3. `npm -v`:
```
5.0.3
```

4. `yarn --version`:
```
0.27.5
```

Then, specify:

- Target Platform: Android

- Development Operating System: Windows10

- Build tools: `react-native run-android`

### Steps to Reproduce

(Write your steps here:)

```js
class App extends Component {
  componentDidMount() {
    BackHandler.addEventListener('hardwareBackPress', this.onBackPress);
  }
  componentWillUnmount() {
    BackHandler.removeEventListener('hardwareBackPress', this.onBackPress);
  }
  onBackPress = () => {
    return true;
  };
  render() {
    return (
      <Text>Press back buton</Text>
    );
  }
}
```
If you run the code in `development` environment , it don't works and when you press back button , app will be closed , but in `development` environment if you turn on `remote debug` , the code works well and app not closed

### Expected Behavior

(Prevent closing app on back button pressed)

### Actual Behavior

(On `development` environment when `remote debug` is off it don't woks but when `remote debug` is on it works)

### Reproducible Demo

(N/A)

sm2017:
Any updates??

Sean-Snow:
+1

hramos:
Can you reproduce the issue using [Snack](http://snack.expo.io)? This step is necessary for people to be able to see and debug the issue being reported.

sm2017:
@hramos I explain in details how to reproduce the issue , I cannot reproduce the issue using Snack because there is no dev menu in Snack and you cannot turn on remote debug

hramos:
You can run the Snack within the Expo Client and shake the device to trigger a menu to show up, no? 

In any case, this is a community project. To answer your "any updates?" question, we'd take a look at any PR that proposes a fix.

sm2017:
@hramos , I tried before using Expo Client , there is no dev menu , You can try it yourself
exp://expo.io/@snack/SJn2le4cZ

dantman:
@sm2017 Is there a Symbol polyfill included by your app or any of your dependencies?

[...]

Issue #15031, opened 2017-07-15
Inconsistent dynamic array-length rendering behavior between iOS and Android.


dsimmons:
### Is this a bug report?

Yes.

### Have you read the Bugs section of the Contributing to React Native Guide?

Yes. If there is any open issue(s) related to this behavior, I must have missed it.

### Environment

1. `react-native -v`:

```
react-native-cli: 2.0.1
react-native: 0.45.0
```

2. `node -v`: `v7.10.0`
3. `npm -v`: `4.2.0`
4. `yarn --version` (if you use Yarn): `0.24.5`

- Target Platform: iOS, Android

- Development Operating System: macOS 10.12.5

- Build tools: Standard.

(**NOTE:** Although the above is reflective of my actual environment, I also just reproduced on a fresh `react-native init` using RN `v0.46.3`).

### Steps to Reproduce

1. `react-native init MyNewApp`

2. Embed the minimal code snippet:

```Javascript
  <View>
    <StatusBar hidden />
    {Array.from(Array(5).keys()).map(idx => (
      <Text key={idx}>{idx}</Text>
    ))}
  </View>

```

3. Run on iOS and Android separately, observe the below behavior (screenshots):

<img width="628" alt="screen shot 2017-07-15 at 5 17 34 pm" src="https://user-images.githubusercontent.com/694034/28242769-d6ca80fe-6982-11e7-85aa-46c28341ab9f.png">

### Expected Behavior

I'd expect the rendered output to be **identical** on **both** iOS and Android. Either:

  - ... it works on **both** iOS and Android.

  - ... it works on works on **neither** iOS **nor** Android.

Currently, it works as expected on iOS. On Android, **nothing is rendered**.

Although I do realize this is a relatively new language feature that may require transpilation, as I mentioned above, if it weren't to work, I'd expect it to not work on **either** platform rather than the inconsistent results I'm seeing presently.

### Actual Behavior

See above screenshot.

### Reproducible Demo

See above code snippet.

  - Specifically, the pertinent line: `Array.from(Array(5).keys())`.

  - I'd also originally tried: `[...Array(5).keys()]`

### Additional Context

Although I used a **static** (literal) numeric value in the minimal reproducible snippet above, in my actual app code, the ***actual*** value (quantity) is **dynamic** (depending on props).

For example, imagine the "pegs" of a paging indicator for a swipeable view, where the number of elements shown is dynamic based on some resultant API call to populate the view with data.

[...]

PR #17820, opened 2018-02-01
Remove Array.from in BackHandler.android.js to fix bug on back press

AuroreM:
## Motivation

An issue on the BackHandler of Android has been introduced by [this commit](https://github.com/facebook/react-native/commit/165b38ab7abdf9d769efddde39da35a03564284b#diff-9c3e02e844d6379291698a3c1210055a). When we want to customise the back press behaviour, no matter if the `hardwareBackPress`listener returns true or false, the app is exited. It must due to a conflict between the `Set` polyfill and an other one as it does not happen on every app.

See [the issue](https://github.com/facebook/react-native/issues/15497) here.

## Test Plan

Use a custom listener like this one for `hardwareBackPress`:
```javascript
  backAction = () => {
    return true
  };
```

Here are 2 videos.
The first one is made with the fix (RN 0.52.2) : 
ezgif-1-744ca685e9

The second one is without the fix (RN 0.52.2 as well) :
ezgif-1-9f7a6d3dcc

## Release Notes
[ANDROID] [BUGFIX] [BackHandler] - Fix the backPress: if the listener returns true, the app does not exit

### DIFF
--- a/Libraries/Utilities/BackHandler.android.js
+++ b/Libraries/Utilities/BackHandler.android.js
@@ -22,9 +22,7 @@ var _backPressSubscriptions = new Set();
 
 RCTDeviceEventEmitter.addListener(DEVICE_BACK_EVENT, function() {
   var invokeDefault = true;
-  let subscriptions = [];
-  _backPressSubscriptions.forEach(subscription => subscriptions.push(subscription));
-  subscriptions = subscriptions.reverse();
+  var subscriptions = Array.from(_backPressSubscriptions.values()).reverse();
 
   for (var i = 0; i < subscriptions.length; ++i) {
     if (subscriptions[i]()) {
--- a/Libraries/Utilities/BackHandler.android.js
+++ b/Libraries/Utilities/BackHandler.android.js
@@ -22,7 +22,9 @@ var _backPressSubscriptions = new Set();
 
 RCTDeviceEventEmitter.addListener(DEVICE_BACK_EVENT, function() {
   var invokeDefault = true;
-  var subscriptions = Array.from(_backPressSubscriptions.values()).reverse();
+  let subscriptions = [];
+  _backPressSubscriptions.forEach(subscription => subscriptions.push(subscription));
+  subscriptions = subscriptions.reverse();
 
   for (var i = 0; i < subscriptions.length; ++i) {
     if (subscriptions[i]()) {
--- a/Libraries/polyfills/Array.es6.js
+++ b/Libraries/polyfills/Array.es6.js
@@ -28,9 +28,7 @@ if (!Array.from) {
 
     var C = this;
     var items = Object(arrayLike);
-    var symbolIterator = typeof Symbol === 'function'
-      ? Symbol.iterator
-      : '@@iterator';
+    var symbolIterator = '@@iterator';
     var mapping = typeof mapFn === 'function';
     var usingIterator = typeof items[symbolIterator] === 'function';
     var key = 0;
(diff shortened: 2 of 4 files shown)


Minishlink:
Also having this issue on [this open-source](https://github.com/Minishlink/DailyScrum) repo.
Temporary workaround with [this commit](https://github.com/Minishlink/DailyScrum/commit/b6525764c81633169be5380eeba6bddc3e62c7ce).

janicduplessis:
I'm a bit hesitant to merge this since I feel like we should fix the polyfill issue instead working around it. A similar issue could just come up again in some other code in the future.

I'm not sure exactly what the problem is but from what I understand we require the Set and Map polyfills [here)(https://github.com/facebook/react-native/blob/master/Libraries/Core/InitializeCore.js#L40) and they depend on having global.Symbol available or use the '@@symbol' string key [here](https://github.com/facebook/react-native/blob/master/Libraries/vendor/core/toIterator.js#L25). I guess this causes conflict with whatever else Symbol polyfill is being used in other apps.

Could you try adding a Symbol polyfill instead or maybe just always use the '@@iterator' string key just to see if it fixes the problem? Then we can decide on which symbol polyfill we use.

AuroreM:
@janicduplessis I've tried to force the polyfill Array to use '@@iterator' as the Symbol.iterator. It fixes the issue, but only for the conflicts with the Array polyfills. The logic
```
var symbolIterator = typeof Symbol === 'function' ? Symbol.iterator : '@@iterator';
```
is used elsewhere in the app (https://github.com/facebook/react-native/blob/master/babel-preset/transforms/transform-symbol-member.js) so it's fixing the issue when we use the Array polyfill but it could break between other polyfills

AuroreM:
@janicduplessis Have you got a chance to look at it ? :)

AuroreM:
@janicduplessis I've rebased on the current version of the `master` branch. Could you have a look at it please ? :)

janicduplessis:
@AuroreM Woops I missed the notification.

I wonder if actually including a Symbol polyfill would work better here, we'd have to make sure it is included before the Set / Map ones.

@TheSavior What do you think about this (TLDR: Map/Set breaks when an external symbol polyfill is used because it is added after the Map/Set polyfills)? These polyfills are vendored from some internal fb codebase right? Do you know if there is also a symbol one we could add?

TheSavior:
@janicduplessis 
I'm not very familiar with how our polyfill systems are set up, but looking through both of our internal repositories I don't see any polyfills for symbol, and core-js (which includes a polyfill) isn't being pulled in directly. 

The only thing I have found for symbol support is [this babel transform](https://github.com/facebook/react-native/blob/f2ad77f90a5553d05448acbc4eac1eb7ad4809d5/babel-preset/transforms/transform-symbol-member.js)

hey99xx:
I think BackHandler bug might be fixed in another issue https://github.com/facebook/react-native/commit/c3e42b55312e196d7568b5fd4b3d28f50bfa0fd9 (currently in master branch, not in any release). 

But going forward I think there's still value fixing `Array.from` bugs, this one is probably also related https://github.com/facebook/react-native/issues/15031

I wonder if the polyfill from MDN docs is more compliant https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill

cpojer:
Closing as @hey99xx pointed out that it was fixed by another change recently. I hope we can get rid of the polyfills soon!

hey99xx:
@cpojer Actually lot of these issues might be fixed once RN releases with new JSC which is currently in master. At that point a lot of these polyfills will be obsolete for Android too.

PR #17841, opened 2018-02-02
Increase Tls1.2.0 support on Android (5.0.1 and below) 

tuncaulubilge:
## Motivation

On older Android devices (5.0.1 and below), TLS1.2 is not supported by default. There is already a fix in place for this on React Native, but it requires some improvements. Added a few extra details to cover wider range of devices:

- Current fix covers SDK 19 and below, but same issue is reported up to api 21 for some Samsung devices. Increased version check for the fix
- Current fix only forces Tlsv1.2 but we should keep the support for 1.1 and 1.0, so switched to default connection specs from OkHttpClient
- `sslSocketFactory` method in OkHttpClient is deprecated. Applied the recommended fix from OkHttpClient documentations. Deprecated method uses reflection to access TrustManager, which requires an extra proguard rule on projects. 

## Test Plan

- This PR is an improvement on an existing fix. All tests are running as expected.

## Related PRs

Previous commit on the issue: https://github.com/facebook/react-native/commit/55ebb89916bd084c3c352d0f9948ae0583e583ac

## Release Notes

[ANDROID] [BUGFIX][OkHttpClientProvider.java] Fixed issues with TLSv1.2 on older devices

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
@@ -23,7 +23,6 @@ import java.util.concurrent.TimeUnit;
 
 import okhttp3.ConnectionSpec;
 import okhttp3.OkHttpClient;
-import okhttp3.TlsVersion;
 
 /**
  * Helper class that provides the same OkHttpClient instance that will be used for all networking
@@ -47,7 +46,7 @@ public class OkHttpClientProvider {
     }
     return sClient;
   }
-  
+
   // okhttp3 OkHttpClient is immutable
   // This allows app to init an OkHttpClient with custom settings.
   public static void replaceOkHttpClient(OkHttpClient client) {
@@ -91,12 +90,8 @@ public class OkHttpClientProvider {
 
         client.sslSocketFactory(new TLSSocketFactory(trustManager), trustManager);
 
-        ConnectionSpec cs = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
-                .tlsVersions(TlsVersion.TLS_1_0, TlsVersion.TLS_1_1, TlsVersion.TLS_1_2)
-                .build();
-
         List<ConnectionSpec> specs = new ArrayList<>();
-        specs.add(cs);
+        specs.add(ConnectionSpec.MODERN_TLS);
         specs.add(ConnectionSpec.COMPATIBLE_TLS);
         specs.add(ConnectionSpec.CLEARTEXT);
 
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
@@ -65,16 +65,16 @@ public class OkHttpClientProvider {
       .writeTimeout(0, TimeUnit.MILLISECONDS)
       .cookieJar(new ReactCookieJarContainer());
 
-    return enableTls12OnPreLollipop(client);
+    return enableTls12OnLollipopAndBelow(client).build();
   }
 
   /*
-    On Android 4.1-4.4 (API level 16 to 19) TLS 1.1 and 1.2 are
+    On Android 4.1-5.0 (API level 16 to 21) TLS 1.1 and 1.2 are
     available but not enabled by default. The following method
     enables it.
    */
-  public static OkHttpClient.Builder enableTls12OnPreLollipop(OkHttpClient.Builder client) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN && Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
+  public static OkHttpClient.Builder enableTls12OnLollipopAndBelow(OkHttpClient.Builder client) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN && Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
       try {
         client.sslSocketFactory(new TLSSocketFactory());
 
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
@@ -16,12 +16,9 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
 import java.security.KeyStore;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
```diff
@@ -64,21 +68,30 @@ public static OkHttpClient createClient() {
...
+        if (trustManagers.length != 1 || !(trustManagers[0] instanceof X509TrustManager)) {
+          throw new IllegalStateException("Unexpected default trust managers:" + Arrays.toString(trustManagers));
+        }
+        X509TrustManager trustManager = (X509TrustManager) trustManagers[0];
+
+        client.sslSocketFactory(new TLSSocketFactory(trustManager), trustManager);
 
         ConnectionSpec cs = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
-                .tlsVersions(TlsVersion.TLS_1_2)
+                .tlsVersions(TlsVersion.TLS_1_0, TlsVersion.TLS_1_1, TlsVersion.TLS_1_2)
```
ide: Not sure we want to enable older TLS, especially 1.0. Probably better to leave it at 1.2 only.

tuncaulubilge: TLS 1.0 is still supported by Android platform and by Okhttp, so react native should not force to TLS1.2 just to fix an issue. For Lolipop+, tls 1.0 is already available, so I'd rather enable it on older devices too in order to keep version behaviour consistent. By adding all tls versions, we let the server make the decision

ide: Is there a way to see which protocols the Android device supports by default and add only more secure ones?

ide:
@hramos Would it be possible to find someone especially familiar with the networking code and security to review this carefully? In particular, this commit affects TLS and also assumes the app has just one TrustManager.

tuncaulubilge:
Here is the javadoc by Okhttp about trustmanagers. https://github.com/square/okhttp/blob/9cb3af2a03870f54a0b49c4b26042e1ec98d3513/okhttp/src/main/java/okhttp3/OkHttpClient.java#L673

ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
tuncaulubilge: TLS version is decided together by the client and the server. Server and client both share the list of tls versions supported and the highest version that is supported on both platforms is selected. 

Adding TLS 1.0 here is only useful if the server doesn't support TLS 1.1 and 1.2, which is possible for a development server or a qa environment.

newyankeecodeshop: I have a PR #17309 that relates to this. It is important to recognize that some Android environments do not support TLS v1.0. Does this code require checking the list of supported protocols explicitly?

hey99xx: At this point why have a custom `ConnectionSpec` instance. As far as I see you can just use [`ConnectionSpec.MODERN_TLS`](https://github.com/square/okhttp/blob/okhttp_36/okhttp/src/main/java/okhttp3/ConnectionSpec.java#L68-L70) as is, without creating a builder on top of it. It is already set to 

     .tlsVersions(TlsVersion.TLS_1_3, TlsVersion.TLS_1_2, TlsVersion.TLS_1_1, TlsVersion.TLS_1_0)

In addition [`ConnectionSpec.COMPATIBLE_TLS`](https://github.com/square/okhttp/blob/okhttp_36/okhttp/src/main/java/okhttp3/ConnectionSpec.java#L75-L76) is built on top of `ConnectionSpec.MODERN.TLS`.

So if I am not misunderstanding the entire `specs` list could just be

    List<ConnectionSpec> specs = new ArrayList<>();
    specs.add(ConnectionSpec.MODERN_TLS);
    specs.add(ConnectionSpec.CLEARTEXT);

Would this not be far simpler (unless I misunderstand OkHttp/TLS API)?

tuncaulubilge: We still need COMPATIBLE_TLS as a fallback on older devices but it is a good idea to use MODERN_TLS instead of a custom builder. Updated in a6d30f7

ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
```diff
@@ -90,12 +89,8 @@ public static OkHttpClient createClient() {
...
 
-        ConnectionSpec cs = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
-                .tlsVersions(TlsVersion.TLS_1_0, TlsVersion.TLS_1_1, TlsVersion.TLS_1_2)
-                .build();
-
         List<ConnectionSpec> specs = new ArrayList<>();
-        specs.add(cs);
+        specs.add(ConnectionSpec.MODERN_TLS);
         specs.add(ConnectionSpec.COMPATIBLE_TLS);
         specs.add(ConnectionSpec.CLEARTEXT);
```
hey99xx: Now you have same as [`OkHttpClient.DEFAULT_CONNECTION_SPECS`](https://github.com/square/okhttp/blob/okhttp_36/okhttp/src/main/java/okhttp3/OkHttpClient.java#L125-L126).

    static final List<ConnectionSpec> DEFAULT_CONNECTION_SPECS = Util.immutableList(
      ConnectionSpec.MODERN_TLS, ConnectionSpec.COMPATIBLE_TLS, ConnectionSpec.CLEARTEXT);

tuncaulubilge: I'll test this with default specs and we can remove this if it works for all devices. Important part is the TlsSocketFactory, which enables TLS1_1 & TLS1_2 on older devices. Before this PR, we were disabling TLS1_1 and TLS1_0 for newer ones, but it's better to stick with default specs from okhttp

perrosnk:
I still have a problem connecting to my server which uses Let's encrypt certificate, even after using this code.
I have opened this issue https://github.com/facebook/react-native/issues/18032

ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
tuncaulubilge: I removed the custom connection specs overall, and tested on KitKat, this seems to be working as expected.

tuncaulubilge:
@perrosnk Have you tried it on a native Android app with only OkHttp installed? Might be an issue with okHttp

cpojer:
What is the status of this PR?

cpojer:
It seems this PR is outdated and hasn't gotten any attention in a long time. Additionally, very few people seem to care a ton about making this change. I recognize that this is still valuable though so if anyone feels strongly about putting up a new PR with a proper implementation of this and is able to answer all the questions we may have, please make it happen. For now, I'm going to close this one.

PR #17842, opened 2018-02-02
Fix DynamicFromMap object pool synchronization

haitaoli:
## Motivation
DynamicFromMap internally uses SimplePool object to recycle dynamic prop objects. But the pool is not multi-thread safe. Currently the most used dynamic props are size props such as left, paddingVertical, marginTop and so on. These props are only accessed from the layout thread so the pool works fine. If a dynamic prop is needed in UI thread, then the two threads can access the same pool object and cause random errors. This PR make the pool object thread local to avoid synchronization. After this change there are two pool objects created in the process.

## Test Plan
Tested in official Airbnb app after updating accessibilityComponentType to be dynamic.

## Related PRs
Once this is merged, I'll send another PR to support "disabled" state in `accessibilityComponentType`.

## Release Notes
[ANDROID] [BUGFIX] [DynamicFromMap] - Fix a crash caused by dynamic props.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/DynamicFromMap.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/DynamicFromMap.java
@@ -9,13 +9,18 @@ package com.facebook.react.bridge;
 
 import javax.annotation.Nullable;
 
-import android.support.v4.util.Pools;
+import android.support.v4.util.Pools.SimplePool;
 
 /**
  * Implementation of Dynamic wrapping a ReadableMap.
  */
 public class DynamicFromMap implements Dynamic {
-  private static final Pools.SimplePool<DynamicFromMap> sPool = new Pools.SimplePool<>(10);
+  private static final ThreadLocal<SimplePool<DynamicFromMap>> sPool = new ThreadLocal<SimplePool<DynamicFromMap>>() {
+    @Override
+    protected SimplePool<DynamicFromMap> initialValue() {
+      return new SimplePool<>(10);
+    }
+  };
 
   private @Nullable ReadableMap mMap;
   private @Nullable String mName;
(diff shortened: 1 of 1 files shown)


haitaoli:
Rebased

hey99xx:
Our team also noticed this bug after inspecting a crash in production. We actually got a very similar stacktrace to https://github.com/facebook/react-native/issues/19141 starting with 

```
Caused by java.lang.ArrayIndexOutOfBoundsException: length=10; index=-1
       at android.support.v4.util.Pools$SimplePool.release(Pools.java:120)
       at com.facebook.react.bridge.DynamicFromMap.recycle(DynamicFromMap.java:42)
```

only the line number at DynamicFromMap class is now 40 instead of 42 (running more recent version of RN). So this is an obviously an issue that needs to be fixed, either with `ThreadLocal` usage as in this PR or by using `Pools.SynchronizedPool< DynamicFromMap>`.

@haitaoli Is this PR abandoned (probably since AirBnb is leaving RN) or will you continue working on it? I think the same fix should also be done for `DynamicFromArray` class for correctness, even if it's not an issue yet.

haitaoli:
@hey99xx I don't think anyone will review and merge this PR, since this probably doesn't affect FB's own apps. I have another PR to fix a bug on iOS and that's not going anywhere either. As you mentioned Airbnb is deprecating RN so I won't be pushing this fix to get merged. It might be easier for you to fork RN and merge any fixes you need.

janicduplessis:
@haitaoli Sorry this PR got unnoticed. We actually hit an issue caused by this in react-native-svg. Looks good to me so we can go ahead and ship this.

janicduplessis:
@facebook-github-bot shipit

janicduplessis:
@haitaoli Can you link me the iOS PR, I can have a look.

haitaoli:
@janicduplessis thanks! Here is the iOS PR I mentioned: https://github.com/facebook/react-native/pull/17230

PR #17860, opened 2018-02-05
Support non-integer modulus in ModulusAnimatedNode on Android

motiz88:
## Motivation

`Animated.modulo(value, modulus)` supports a non-integer modulus in the iOS and JS implementations but crashes on Android when `useNativeDriver` is set to `true`.

## Test Plan

Unfortunately, I'm not fluent enough with this codebase to add a Java test for this fix - especially as I couldn't find any analogous tests to extrapolate from. However, the fix itself seems straightforward enough.

## Related PRs

None needed

## Release Notes
[ANDROID] [BUGFIX] [Animated] - Support non-integer modulus in .modulo()

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/animated/ModulusAnimatedNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/animated/ModulusAnimatedNode.java
@@ -17,14 +17,14 @@ import com.facebook.react.bridge.ReadableMap;
 
   private final NativeAnimatedNodesManager mNativeAnimatedNodesManager;
   private final int mInputNode;
-  private final int mModulus;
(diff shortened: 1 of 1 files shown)


janicduplessis:
Thanks for fixing this!

@facebook-github-bot shipit

janicduplessis:
For future reference native animated tests are located [here](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/test/java/com/facebook/react/animated/NativeAnimatedNodeTraversalTest.java#L655) but sadly there isn't one for the module node. It should be possible to add one that's similar to the addition / multiplication ones, if you ever feel like writing android tests 🙂.

PR #17878, opened 2018-02-06
[Circle] Switch e2e to yarn

hramos:
Addresses CI build failures due to use of scoped packages by metro.


hramos:
The CI build is still breaking due to a Flow issue, that was introduced a couple of weeks ago. I'm going to land this PR as it brings the e2e workflow back to green. A fix for the Metro/Flow issue will need to be applied in a different internal diff.

PR #17882, opened 2018-02-07
[BUCK] Remove nonexistant dependency

hramos:
Should fix buck failures. Tested on OSS. Will need to confirm internal tests succeed.


hramos:
Fixing internally.

PR #17887, opened 2018-02-07
Exclude jsbundle files from VCS

aneophyte:
## Motivation

jsbundle files can be generated, and are quite large and therefore, I think should be excluded from being committed to the repo.

[ GENERAL ][ MINOR ][.gitignore] - Included a new entry to exclude jsbundle files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ node_modules
 
 /coverage
 /third-party
+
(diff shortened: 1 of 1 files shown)


aneophyte:
Whoops, just realized this might be the wrong .gitignore file for that.

PR #17896, opened 2018-02-07
[Native Animated] Support for animated tracking in native driver

kmagiera:
This PR adds support for Animated tracking to Animated Native Driver implementation on Android and iOS.

## Motivation

Animated tracking allows for animation to be started with a "dynamic" end value. Instead of passing a fixed number as end value we can pass a reference to another Animated.Value. Then when that value changes, the animation will be reconfigured to drive the animation to the new destination point. What is important is that animation will keep its state in the process of updating "toValue". That is if it is a spring animation and the end value changes while the previous animation still hasn't settled the new animation will start from the current position and will inherit current velocity. This makes end value transitions very smooth.

Animated tracking is available in JS implementation of Animated library but not in the native implementation. Therefore until now, it wasn't possible to utilize native driver when using animated tracking. Offloading animation from JS thread turns out to be crucial for gesture driven animations. This PR is a step forward towards feature parity between JS and native implementations of Animated.

Here is a link to example video that shows how tracking can be used to implement chat heads effect: https://twitter.com/kzzzf/status/958362032650244101

In addition this PR fixes an issue with frames animation driver on Android that because of rounding issues was taking one extra frame to start. Because of that change I had to update a number of Android unit tests that were relying on that behavior and running that one additional animation step prior to performing checks.

## Test Plan

As a part of this PR I'm adding three unit tests for each of the platforms that verifies most important aspects of this implementation. Please refer to the code and look at the test cases top level comments to learn what they do.

I'm also adding a section to "Native Animated Example" screen in RNTester app that provides a test case for tracking. In the example we have blue square that fallows the red line drawn on screen. Line uses Animated.Value for it's position while square is connected via tracking spring animation to that value. So it is ought to follow the line. When user taps in the area surrounding the button new position for the red line is selected at random and the value updates. Then we can watch blue screen animate to that position. 

You can also refer to this video that I use to demonstrate how tracking can be linked with native gesture events using react-native-gesture-handler lib: https://twitter.com/kzzzf/status/958362032650244101

## Release Notes

[GENERAL][FEATURE][Native Animated] - Added support for animated tracking to native driver. Now you can use `useNativeDriver` flag with animations that track other Animated.Values

### DIFF
--- a/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
+++ b/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
@@ -11,31 +11,43 @@
 #import "RCTValueAnimatedNode.h"
 #import "RCTNativeAnimatedNodesManager.h"
 
-@implementation RCTTrackingAnimatedNode
+@implementation RCTTrackingAnimatedNode {
+  NSNumber *_animationId;
+  NSNumber *_nodeTag;
+  NSNumber *_valueNodeTag;
+  NSMutableDictionary *_animationConfig;
+}
 
 - (instancetype)initWithTag:(NSNumber *)tag
-                     config:(NSDictionary<NSString *, id> *)config;
+                     config:(NSDictionary<NSString *, id> *)config
 {
   if ((self = [super initWithTag:tag config:config])) {
-
+    _animationId = config[@"animationId"];
+    _nodeTag = config[@"toValue"];
+    _valueNodeTag = config[@"value"];
+    _animationConfig = config[@"animationConfig"];
   }
   return self;
+}
+
+- (void)onDetachedFromNode:(RCTAnimatedNode *)parent
+{
+  [self.manager stopAnimation:_animationId];
+  [super onDetachedFromNode:parent];
 }
 
 - (void)performUpdate
 {
   [super performUpdate];
 
-  NSNumber *nodeTag = self.config[@"toValue"];
-  RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:nodeTag];
-
-  NSNumber *animationId = self.config[@"animationId"];
-  NSNumber *valueNodeTag = self.config[@"value"];
-
-  NSMutableDictionary *config = [NSMutableDictionary dictionaryWithDictionary:self.config[@"animationConfig"]];
+  // clone animation config and update "toValue" to reflect updated value of the parent node
+  RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:_nodeTag];
+  NSMutableDictionary *config = [NSMutableDictionary dictionaryWithDictionary:_animationConfig];
   [config setValue:@(node.value) forKey:@"toValue"];
 
-  [self schedulePostUpdate:^(RCTNativeAnimatedNodesManager *manager) {
+  NSNumber *animationId = _animationId;
+  NSNumber *valueNodeTag = _valueNodeTag;
+  [self.manager schedulePostUpdateOperation:^(RCTNativeAnimatedNodesManager * _Nonnull manager) {
     [manager startAnimatingNode:animationId
                         nodeTag:valueNodeTag
                          config:config
--- a/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
+++ b/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
@@ -25,7 +25,7 @@
     _animationId = config[@"animationId"];
     _nodeTag = config[@"toValue"];
     _valueNodeTag = config[@"value"];
-    _animationConfig = config[@"animationConfig"];
+    _animationConfig = [NSMutableDictionary dictionaryWithDictionary:config[@"animationConfig"]];
   }
   return self;
 }
@@ -40,11 +40,11 @@
 {
   [super performUpdate];
 
-  // clone animation config and update "toValue" to reflect updated value of the parent node
+  // change animation config's "toValue" to reflect updated value of the parent node
   RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:_nodeTag];
-  NSMutableDictionary *config = [NSMutableDictionary dictionaryWithDictionary:_animationConfig];
-  [config setValue:@(node.value) forKey:@"toValue"];
+  [_animationConfig setValue:@(node.value) forKey:@"toValue"];
 
+  NSDictionary *config = _animationConfig;
   NSNumber *animationId = _animationId;
   NSNumber *valueNodeTag = _valueNodeTag;
   [self.manager schedulePostUpdateOperation:^(RCTNativeAnimatedNodesManager * _Nonnull manager) {
--- a/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
+++ b/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
@@ -44,15 +44,10 @@
   RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:_nodeTag];
   [_animationConfig setValue:@(node.value) forKey:@"toValue"];
 
-  NSDictionary *config = _animationConfig;
-  NSNumber *animationId = _animationId;
-  NSNumber *valueNodeTag = _valueNodeTag;
-  [self.manager schedulePostUpdateOperation:^(RCTNativeAnimatedNodesManager * _Nonnull manager) {
-    [manager startAnimatingNode:animationId
-                        nodeTag:valueNodeTag
-                         config:config
-                    endCallback:nil];
-  }];
+  [self.manager startAnimatingNode:_animationId
+                           nodeTag:_valueNodeTag
+                            config:_animationConfig
+                       endCallback:nil];
 }
 
 @end
--- a/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
+++ b/Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "RCTTrackingAnimatedNode.h"
+#import "RCTValueAnimatedNode.h"
+#import "RCTNativeAnimatedNodesManager.h"
+
+@implementation RCTTrackingAnimatedNode
+
+- (instancetype)initWithTag:(NSNumber *)tag
+                     config:(NSDictionary<NSString *, id> *)config;
+{
+  if ((self = [super initWithTag:tag config:config])) {
+
+  }
+  return self;
+}
+
+- (void)performUpdate
+{
+  [super performUpdate];
+
+  NSNumber *nodeTag = self.config[@"toValue"];
+  RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:nodeTag];
+
+  NSNumber *animationId = self.config[@"animationId"];
+  NSNumber *valueNodeTag = self.config[@"value"];
+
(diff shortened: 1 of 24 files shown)


ReactAndroid/src/main/java/com/facebook/react/animated/AnimationDriver.java
```diff
@@ -27,4 +29,15 @@
...
    */
   public abstract void runAnimationStep(long frameTimeNanos);
+
+  /**
+   * This method will get called when some of the configuration gets updated while the animation is
+   * running. In that case animation should restart keeping its internal state to provide a smooth
+   * transision. E.g. in case of a spring animation we want to keep the current value and speed and
+   * start animating with the new properties (different destination or spring settings)
+   */
+  public void resetConfig(ReadableMap config) {
```
janicduplessis: If we decide to update the name in the iOS impl update this too.

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
```diff
@@ -49,7 +57,7 @@ public void runAnimationStep(long frameTimeNanos) {
       mFromValue = mAnimatedValue.mValue;
     }
     long timeFromStartMillis = (frameTimeNanos - mStartFrameTimeNanos) / 1000000;
-    int frameIndex = (int) (timeFromStartMillis / FRAME_TIME_MILLIS);
+    int frameIndex = (int) Math.round(timeFromStartMillis / FRAME_TIME_MILLIS);
```
janicduplessis: Why was this change needed? Does it match the iOS / JS implementation better like this?

ReactAndroid/src/main/java/com/facebook/react/animated/TrackingAnimatedNode.java
```diff
@@ -0,0 +1,37 @@
...
+  private final NativeAnimatedNodesManager mNativeAnimatedNodesManager;
+  private final int mAnimationId;
+  private final int mParentNode;
+  private final int mAnimatingNode;
+  private final JavaOnlyMap mAnimationConfig;
+
+  TrackingAnimatedNode(ReadableMap config, NativeAnimatedNodesManager nativeAnimatedNodesManager) {
+    mNativeAnimatedNodesManager = nativeAnimatedNodesManager;
+    mAnimationId = config.getInt("animationId");
+    mParentNode = config.getInt("toValue");
```
janicduplessis: Let's name these ivars the same as the iOS implementation if possible.

ReactAndroid/src/main/java/com/facebook/react/bridge/JavaOnlyArray.java
```diff
@@ -60,12 +88,12 @@ public boolean isNull(int index) {
 
   @Override
   public double getDouble(int index) {
-    return (Double) mBackingList.get(index);
```
janicduplessis: Were these broken before?

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
kmagiera: Tried to explain that in my PR description, you might have missed it as it is at the very end:

> In addition this PR fixes an issue with frames animation driver on Android that because of rounding issues was taking one extra frame to start. Because of that change I had to update a number of Android unit tests that were relying on that behavior and running that one additional animation step prior to performing checks.

So basically because of unfortunate rounding here frames driver on android would always "play" first frame twice. This could also be seen in all the unit tests where we run update loop twice at the beginning for all the framedriver tests. I am also updating these in the PR.

ReactAndroid/src/main/java/com/facebook/react/bridge/JavaOnlyArray.java
kmagiera: So there was a small inconsistency with the way getDouble and getInt work with JavaOnlyArray or Map vs NativeArray or Map. The problem was that in JS we only have "numbers" so both int and double map to the same type. Therefore NativeArray or Map is also casting the value to an appropriate type at the moment when getDouble or getInt is called. So even if we send a floating point from the JS, java code can call getInt and it will return a rounded integer. Same if we send an integer from JS. 

This has became a problem since in animated driver tracking node I need to clone NativeArray/Map to JavaOnlyArray/Map, and when there are fields of type "Number" I can't tell if these are floating point numbers or not. Because of that I always read and write them as double. Then when I pass that JavaOnlyMap/Array down to the driver it knows about semantics of the fields and thus it may call getInt on some fields. In that case it was failing because the number has been set as a double.

Libraries/NativeAnimation/Nodes/RCTTrackingAnimatedNode.m
```diff
@@ -0,0 +1,54 @@
...
+  [self.manager stopAnimation:_animationId];
+  [super onDetachedFromNode:parent];
+}
+
+- (void)performUpdate
+{
+  [super performUpdate];
+
+  // change animation config's "toValue" to reflect updated value of the parent node
+  RCTValueAnimatedNode *node = (RCTValueAnimatedNode *)[self.parentNodes objectForKey:_nodeTag];
```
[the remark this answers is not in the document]
kmagiera: parentNodes is of type `NSMapTable` so that won't work. I can use it for the next line tho

janicduplessis:
@facebook-github-bot shipit

hramos:
I looked into the failure and it looks like this is not passing some of our internal tests:

```
js/react-native-github/RNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:987:18: error: unused variable 'frame' [-Werror,-Wunused-variable]
  for (NSNumber *frame in frames) {
                 ^
js/react-native-github/RNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:1010:18: error: unused variable 'frame' [-Werror,-Wunused-variable]
  for (NSNumber *frame in frames) {
                 ^
stderr: js/react-native-github/RNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:987:18: error: unused variable 'frame' [-Werror,-Wunused-variable]
  for (NSNumber *frame in frames) {
                 ^
js/react-native-github/RNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:1010:18: error: unused variable 'frame' [-Werror,-Wunused-variable]
  for (NSNumber *frame in frames) {
                 ^
```

If you're surprised by this (Circle shows this PR as passing `test_objc_ios`), it's likely because I just found out yesterday that we haven't been running iOS tests at all since we moved them from Travis to Circle in https://github.com/facebook/react-native/pull/16354. On top of that, these tests were not running internally since December for an unrelated reason, but we fixed that internally last week.

The failure is not necessarily coming from this PR -- I haven't digged into that just yet -- but just wanted to let you know what's currently blocking it. I'll report back if it turns out this PR needs to be updated, but until then, assume that it's unrelated and that we need to fix our own internal breakage first.

janicduplessis:
@hramos Sometimes this happens on iOS because OSS and FB use different clang warning flags when building (I think because fb build uses buck internally to build iOS which is not open source).

janicduplessis:
@facebook-github-bot shipit

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
axemclion: Looks like this line is starting to cause IllegalStateException - 
```
Caused by: java.lang.IllegalStateException: Calculated frame index should never be lower than 0
	at com.facebook.react.animated.FrameBasedAnimationDriver.runAnimationStep(FrameBasedAnimationDriver.java:60)
	at com.facebook.react.animated.NativeAnimatedNodesManager.runUpdates(NativeAnimatedNodesManager.java:444)
	at com.facebook.react.animated.NativeAnimatedModule$1.doFrameGuarded(NativeAnimatedModule.java:100)
```

Is there a reason why timeFromStartMills will be negative ?

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
```diff
@@ -49,7 +57,7 @@ public void runAnimationStep(long frameTimeNanos) {
...
     if (frameIndex < 0) {
       throw new IllegalStateException("Calculated frame index should never be lower than 0");
```
axemclion: Looks like this line is starting to cause IllegalStateException  in apps- 
```
Caused by: java.lang.IllegalStateException: Calculated frame index should never be lower than 0
	at com.facebook.react.animated.FrameBasedAnimationDriver.runAnimationStep(FrameBasedAnimationDriver.java:60)
	at com.facebook.react.animated.NativeAnimatedNodesManager.runUpdates(NativeAnimatedNodesManager.java:444)
	at com.facebook.react.animated.NativeAnimatedModule$1.doFrameGuarded(NativeAnimatedModule.java:100)
```

Is there a reason why timeFromStartMills will be negative ?

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
axemclion: Looks like this line is starting to cause IllegalStateException  in apps- 
```
Caused by: java.lang.IllegalStateException: Calculated frame index should never be lower than 0
	at com.facebook.react.animated.FrameBasedAnimationDriver.runAnimationStep(FrameBasedAnimationDriver.java:60)
	at com.facebook.react.animated.NativeAnimatedNodesManager.runUpdates(NativeAnimatedNodesManager.java:444)
	at com.facebook.react.animated.NativeAnimatedModule$1.doFrameGuarded(NativeAnimatedModule.java:100)
```

Is there a reason why timeFromStartMills will be negative ? The exception seems to have started with this PR merged in

ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
```diff
@@ -60,7 +68,7 @@ public void runAnimationStep(long frameTimeNanos) {
     if (frameIndex >= mFrames.length - 1) {
       nextValue = mToValue;
       if (mIterations == -1 || mCurrentLoop < mIterations) { // looping animation, return to start
-        mStartFrameTimeNanos = frameTimeNanos;
+        mStartFrameTimeNanos = frameTimeNanos + ((long) FRAME_TIME_MILLIS) * 1000000L;
```
janicduplessis: @axemclion @kmagiera I think this line could be the culprit. I guess it could be possible that mStartFrameTimeNanos is bigger than the curent clock time.

sophiebits: Maybe this would be better as `mStartFrameTimeNanos = -1;`.

kmagiera: I remember there was some other problem when I was setting it to -1 but can't remember what exactly. Will try that and send a new PR

kmagiera: Ah, ok I remember now, I didn't want `mFromValue` to be updated with the next loop and that happens when mStartFrameTimeNanos is -1. Will add another check for this case and send a PR

kmagiera: Submitted it here: https://github.com/facebook/react-native/pull/18061

syaau:
@kmagiera Did you know that the native tracking doesn't work as it should when the tracked value is animated (rather than being set directly). The tracking starts only after the initiating animation is complete. Is this some issue with how the native animations pre-calculate and serialize the values when the animation is started? Here are the screen shots for animations without and with native driver -
without native driver
Without useNativeDriver
with native driver
With useNativeDriver

I have animated the red circle to move with timing of duration 500, the green circle tracks the red with spring animation and the blue circle tracks the green one.

Here's a code for your reference:
```
<View style={{ flex: 1 }}>
    <View style={{ flex: 1 }}>
      <Circle color="red" position={first} />
      <Circle color="green" position={second} />
      <Circle color="blue" position={third} />
    </View>
    <Button
      title="Start"
      onPress={() => {
        const useNativeDriver = true;
        Animated.spring(second, { toValue: first, duration: 0, friction: 3, useNativeDriver }).start();
        Animated.spring(third, { toValue: second, duration: 0, friction: 3, useNativeDriver }).start();


        Animated.timing(first, {
          toValue: { x: 300, y: 300 },
          duration: 500,
          useNativeDriver,
        }).start();
      }}
    />
</View>
```

Issue #17899, opened 2018-02-07
RNTester/js/http_test_server.js Uses express API with connect middleware. 


ej3:
### Is this a bug report?

Yes

### Steps to Reproduce
1. `git clone git@github.com:facebook/react-naive.git`
2. `cd react-native`
3. `npm install`
4. `./RNTester/js/http_test_server.js &`
5. `curl -D - localhost:5556`

### Expected Behavior

*Server Output*
```cli
Test server for WebSocketExample

This will set a cookie named "wstest" on the response of any incoming request.

received request
```

*Server Response*
```html
HTTP/1.1 200 OK
X-Powered-By: Express
Set-Cookie: wstest=OK; Path=/
Date: Wed, 07 Feb 2018 21:36:13 GMT
Connection: keep-alive
Content-Length: 21

Cookie has been set!
```

### Actual Behavior

*Server Output*
```cli
Test server for WebSocketExample

This will set a cookie named "wstest" on the response of any incoming request.

received request
TypeError: res.cookie is not a function
    at /mnt/c/Users/ejbru/dev/appittome/react-native/RNTester/js/http_test_server.js:37:7
    at call (/mnt/c/Users/ejbru/dev/appittome/react-native/node_modules/connect/index.js:239:7)
    at next (/mnt/c/Users/ejbru/dev/appittome/react-native/node_modules/connect/index.js:183:5)
    at Function.handle (/mnt/c/Users/ejbru/dev/appittome/react-native/node_modules/connect/index.js:186:3)
    at Server.app (/mnt/c/Users/ejbru/dev/appittome/react-native/node_modules/connect/index.js:51:37)
    at emitTwo (events.js:106:13)
    at Server.emit (events.js:194:7)
    at parserOnIncoming (_http_server.js:565:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
```
*Server Response*
```html
HTTP/1.1 500 Internal Server Error
Content-Security-Policy: default-src 'self'
X-Content-Type-Options: nosniff
Content-Type: text/html; charset=utf-8
Content-Length: 960
Date: Wed, 07 Feb 2018 21:22:09 GMT
Connection: keep-alive

<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
[...]

PR #17900, opened 2018-02-07
RNTester http_server send cookie fix

ej3:
Signed-off-by: Evan J Brunner <ej3@appitto.me>

## Motivation

Motivation can be found in #17899

## Test Plan

This `RNTester/js/http_test_server.js` is part of a internal websocket test suite / devtool.

Can be tested with `curl -D - localhost:5556` observing that the `Set-Cookie: wstest=OK; Path=\` header is present, and the service throws no exceptions.. etc

## Release Notes
[INTERNAL][MINOR][./RNTester/js/http_test_server.js] - fixed set cookie with connect framework

### DIFF
--- a/RNTester/js/http_test_server.js
+++ b/RNTester/js/http_test_server.js
@@ -30,11 +30,7 @@ const app = connect();
 
 app.use(function(req, res) {
   console.log('received request');
-  const cookieOptions = {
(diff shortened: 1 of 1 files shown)


hramos:
Can you rebase? A recent internal change is blocking this from landing cleanly.

PR #17912, opened 2018-02-08
[local-cli] Restore copyright header

hramos:
Restoring the copyright header on some files, in order to fix an internal lint failure that is raised whenever these files are edited on open source.


grabbou:
Looks like we did a small mistake when copying & pasting stuff :)

PR #17927, opened 2018-02-09
[IOS][BUGFIX] Fix: Added scroll Bounds check in scrollToOffset method in RCTScrollView on iOS

siddhantsoni:
## Motivation

The scrollToOffset method of RCTScrollView for iOS does not include bound check for the scroll offset provided to the method. This can cause the whole view to scroll out of screen if the offset provided is greater than the bounds of the View.
The same does not happen on Android, where the scroll halts once the last item of the scrollView is in the viewport.
I have added bounds check so the offset resets to the MaxOffset which makes sure the view does not scroll out of the viewport.

The issue can be observed in the following snack:
https://snack.expo.io/H1363Uo8f

ezgif com-optimize 1

## Test Plan

To test my changes I ran the code provided in the snack above with the react-native dependency pointing to my branch. As can be see in the video attached below, the scroll halts once it hits the end of the ScrollView even if the scroll offset provided is higher than the bounds of the ScrollView. It also does not scroll up for negative scroll offset.

ezgif com-optimize

## Release Notes

[IOS] [BUGFIX] [ScrollView] - Added bounds check to prevent ScrollView from scrolling to an offset which is out of bounds of the ScrollView for iOS.

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -577,6 +577,23 @@ static inline void RCTApplyTransformationAccordingLayoutDirection(UIView *view, 
 - (void)scrollToOffset:(CGPoint)offset animated:(BOOL)animated
 {
   if (!CGPointEqualToPoint(_scrollView.contentOffset, offset)) {
+  BOOL isHorizontal = [self isHorizontal:_scrollView];
+  if (isHorizontal) {
+    CGFloat offsetX = _scrollView.contentSize.width - _scrollView.bounds.size.width + _scrollView.contentInset.right;
+    if(offset.x > offsetX) {
+      offset = CGPointMake(fmax(offsetX, 0), 0);
+    } else if(offset.x < 0) {
+      offset = CGPointMake(0, 0);
+    }
+  } else {
+    CGFloat offsetY = _scrollView.contentSize.height - _scrollView.bounds.size.height + _scrollView.contentInset.bottom;
+    if(offset.y > offsetY) {
+      offset = CGPointMake(0, fmax(offsetY, 0));
+    } else if(offset.y < 0) {
+      offset = CGPointMake(0, 0);
+    }
+  }
+
     // Ensure at least one scroll event will fire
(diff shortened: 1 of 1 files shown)


React/Views/ScrollView/RCTScrollView.m
```diff
@@ -577,6 +577,23 @@ - (void)scrollToOffset:(CGPoint)offset
 - (void)scrollToOffset:(CGPoint)offset animated:(BOOL)animated
 {
   if (!CGPointEqualToPoint(_scrollView.contentOffset, offset)) {
+  BOOL isHorizontal = [self isHorizontal:_scrollView];
```
shergin: Please follow existing code style guides.

React/Views/ScrollView/RCTScrollView.m
```diff
@@ -577,6 +577,23 @@ - (void)scrollToOffset:(CGPoint)offset
...
+  if (isHorizontal) {
```
shergin: Why should we have this condition?
Seems we just have to have one single expression which ensure that `offset` inside allowed range.

siddhantsoni: Updated PR to handle offset range in single expression.

siddhantsoni:
@shergin Can you help with one of the tests that is failing. I am not sure what kind of correction the test expects. The failing test is: analyze_pr

hramos:
The anaylze_pr failure seems unrelated to this PR.

naqvitalha:
@shergin @hramos Is this getting pulled in anytime soon? Not sure about your labels.

naqvitalha:
@shergin Why was this reverted? This leads to inconsistent behaviour between iOS and Android. This might be a breaking change but one that should be made eventually.

hramos:
@naqvitalha this is covered in the revert commit at https://github.com/facebook/react-native/commit/8466db0fd355aea778ff71cfb10c927cfd18a551#diff-2f26f4ef865c6f1f1fb4485aa7fdb3a9: "The pull-request caused an issue with the keyboard covering the text field at the bottom of the scroll view."

PR #17936, opened 2018-02-10
Update DevLoadingView to Support iPhone X

mrtnrst:
## Motivation

The current implementation of DevLoadingView for iPhone currently gives a static height of `22` and does not take into account iPhoneX screen dimensions. 

## Test Plan
Devices: All iPhone devices currently available with Xcode v9.2
SDK: 8.1, 9, 10, 11

Validate resize only occurs on iPhone X devices and others remain consistent. 

### Screenshots / GIFs
Before: 
feb-10-2018 12-30-20

After:
feb-10-2018 12-28-15

## Release Notes
[GENERAL][ENHANCEMENT][{React}] - Improvements to DevLoadingView for iPhone X


brunolemos:
Thanks for this PR.

I see it was already merged, but we should not use magical numbers.
The correct implementation would be to use `safeAreaInsets`, like I did here: https://github.com/cooperka/react-native-snackbar/pull/65/files

PR #17963, opened 2018-02-13
Change irrelevant Products path in ./local-cli/runIOS/runIOS.js

blackneck:
https://github.com/facebook/react-native/issues/7308
and many of the same

Seems like xcode cli tool isn't create ./build directory any more but place Products
right inside ./Build (may check while creating single view ios application
in XCode)

* edit hardcoded path to {appName}.app

## Motivation

Command `$ react-native run-ios` is often lead to `Entry, ":CFBundleIdentifier", Does Not Exist` 
I find out how to fix it in current version of RN and Xcode cli tools

## Test Plan

Just try to do `$ react native init {appName}` and then `$ react native run-ios` if everything fine it means my approach is right :)

### DIFF
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -26,7 +26,7 @@ const getBuildPath = function (configuration = 'Debug', appName, isDevice) {
     device = 'iphonesimulator';
   }
 
-  return `build/Build/Products/${configuration}-${device}/${appName}.app`;
+  return `Build/Products/${configuration}-${device}/${appName}.app`;
 };
 const xcprettyAvailable = function() {
   try {


hramos:
Three of the tests are failing, but it's unrelated to your PR. You can rebase past https://github.com/facebook/react-native/commit/cc6d0937c37d2dc7e97a903f5fbf4fa40404e767 which should fix the failures.

blackneck:
@hramos I think I've done what you suggest, but still got weird test failure for Android, besides my fix target only js codebase and could hardly affect Android part

shivawu:
Just to be precise, it seems that xcodebuild would create the folder in `ios/build`, not `ios/Build`. So technically this is only correct because Apple's file system is case-insensitive.

angly-cat:
> ## Test Plan
> 
> Just try to do `$ react native init {appName}` and then `$ react native run-ios` if everything fine it means my approach is right :)

Ironically, if one would do `$ react-native init {appName}` and then `$ react-native run-ios` with `0.55.0` version of `react-native`, then because of this patch it would fail, causing `Entry, ":CFBundleIdentifier", Does Not Exist` error, which this patch was meant to fix.

grabbou:
With your path reverting those changes, seems to me the issue never existed?

angly-cat:
> With your path reverting those changes, seems to me the issue never existed?

There're many issues that people describe as `Entry, ":CFBundleIdentifier", Does Not Exist`. See #18710.

In this case I think OP fixed the issue, whatever is was, for himself by making this change and changing his Xcode project's local settings.

grabbou:
Interesting. Thanks for explaining this a bit more. The pull request fixing this regression is landing soon.

PR #17894, opened 2018-02-07
[Android] Fix: setSelection methods is always returned.

jainkuniya:
## Motivation

Because mMostRecentEventCount < mNativeEventCount is always true as mMostRecentEventCount is not updated.

So determine whether text is set or not from getting current text. Also handle out of bound selection.

## Test Plan
Use this component to test with & without this PR

Before this PR, on clicking Button, nothing will happen.
On this PR: text will be selected starting from 0 and ending at 3. (Just for testing I have barcoded `start` & `end` value)
Just created a component to test (below). 

```
/* @flow */
import React, { PureComponent } from 'react';
import { View, TextInput, StyleSheet, TouchableHighlight, Text } from 'react-native';

const componentStyles = StyleSheet.create({
  input: {
    flex: 1,
  },
  button: {
    //position: 'absolute',
    right: 0,
    alignItems: 'center',
    padding: 10,
  },
  buttonText: {
    color: '#000',
  },
  row: {
    flexDirection: 'row',
    justifyContent: 'space-between',
  },
});

export default class PasswordInput extends PureComponent<Props, State> {
  props: Props;

  state: State;

  state = {
    selection: { start: 0, end: 0 },
  };

  handleSelectionChange = event => {
    const { selection } = event.nativeEvent;

    this.setState({
      selection,
    });
  };

  render() {
    const { selection } = this.state;

    return (
      <View>
        <View style={componentStyles.row}>
          <TextInput
            style={[componentStyles.input]}
            placeholder={'Start typing here'}
            placeholderTextColor={'green'}
            autoCorrect={false}
            autoCapitalize="none"
            selection={selection}
            onSelectionChange={event => this.handleSelectionChange(event)}
          />
        </View>
        <TouchableHighlight
          style={componentStyles.button}
          onPress={() => this.setState({ selection: { start: 0, end: 3 } })}
        >
          <Text style={componentStyles.buttonText}>
            {'Selection (Click me to select text starting from 0 to 3 in above textInput)'}
          </Text>
        </TouchableHighlight>
      </View>
    );
  }
}
```

Current behaviour
ezgif com-video-to-gif-5

This PR
ezgif com-video-to-gif-6

## Release Notes

[ANDROID] [BUGFIX] [TextInput] Fix: setSelection methods is always returned.

PR #17990, opened 2018-02-15
[Android][TextInput] Fix: mostRecentEventCount is not updated.

jainkuniya:
## Motivation

Update it's value as it changes on the JS thread, so that updated value is used (https://github.com/facebook/react-native/blob/f7f5dc66493ad25a85927a9503728b0491c8aab9/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L243) whenever setSelection(int start, int end) is called (https://github.com/facebook/react-native/blob/f7f5dc66493ad25a85927a9503728b0491c8aab9/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L315).

## Test Plan

**On [current master](https://github.com/facebook/react-native/tree/f7f5dc66493ad25a85927a9503728b0491c8aab9)**
Log the value of `mMostRecentEventCount` in the `setSelection` method. https://github.com/facebook/react-native/blob/f7f5dc66493ad25a85927a9503728b0491c8aab9/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L243
It is always `0`.

**On this PR** 
It's values gets updated as it new value is set here (https://github.com/facebook/react-native/blob/f7f5dc66493ad25a85927a9503728b0491c8aab9/Libraries/Components/TextInput/TextInput.js#L921)

To try this PR working, test plan of #17894 can also be used :).

## Release Notes

 [ANDROID] [BUGFIX] [TextInput] - Fix: mostRecentEventCount is not updated.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -233,6 +233,10 @@ public class ReactEditText extends EditText {
     mContentSizeWatcher = contentSizeWatcher;
   }
 
+  public void setMostRecentEventCount(int mostRecentEventCount) {
+    mMostRecentEventCount = mostRecentEventCount;
+  }
+
   public void setScrollWatcher(ScrollWatcher scrollWatcher) {
     mScrollWatcher = scrollWatcher;
   }
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -367,6 +367,11 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
     } catch (IllegalAccessException ex) {}
   }
 
+  @ReactProp(name= "mostRecentEventCount", defaultInt = 0)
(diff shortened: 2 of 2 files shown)


shergin:
@jainkuniya Thank you for tackling this!
Why do you think that setting the variable in `public void maybeSetText(ReactTextUpdate reactTextUpdate) {` is not enough? (Looks like it is, but why?) Why should we keep both update paths?

jainkuniya:
hi @shergin 

`public void maybeSetText(ReactTextUpdate reactTextUpdate) {` is only called once (and sometimes it is never called) and it is definitely not called as soon as event count changes so `mMostRecentEventCount` is not updated to recent value via this path. I am thinking to remove this path https://github.com/jainkuniya/react-native/blob/bc874a55da16770034afb3a343c948d8328d9227/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L362 . Should I?

jainkuniya:
hi @shergin, @janicduplessis, @hramos 
waiting for your review :)

hramos:
I don't usually perform reviews, that's up to the community. Ping me again once this has been reviewed.

jainkuniya:
It's waiting for review :)

cc @shergin, @janicduplessis

theolavaux:
When will the change be included in a release?

Issue #17348, opened 2017-12-25
Unhandled JS Exception: TypeError: undefined is not an object (evaluating 'this._subscribableSubscriptions.forEach')


a3diti:
When changing scheme to Release in able to run the app on the device, it triggered an error
```
Unhandled JS Exception: TypeError: undefined is not an object (evaluating 'this._subscribableSubscriptions.forEach')
```
This crashed the application.

I searched for the error code and found it: `project/node_modules/react-native/Libraries/Components/Subscribable.js` on line 33 with RN 0.51.0

a3diti:
I fixed it by evaluating variable `this._subscribableSubscriptions` on line 33 on react native file `project/node_modules/react-native/Libraries/Components/Subscribable.js`

Changing this:
```
this._subscribableSubscriptions.forEach(
```

To this:
```
this._subscribableSubscriptions && this._subscribableSubscriptions.forEach(
```

avshenuk:
@facebook-github-bot reopen please. It's the same issue for me. No addition information. Just started to throw this error today without any changes.

avshenuk:
React Native version 0.50.3 is affected too

a3diti:
My bad for closing it, the issue is fixed for the moment as i mentioned it above, the problem seems to be related to the **Subscribable** component of RN. 
<img width="626" alt="screen shot 2017-12-26 at 00 14 01" src="https://user-images.githubusercontent.com/3646190/34343510-b8e9142c-e9d1-11e7-9aaf-582198334771.png">

noambonnie:
I also have this issue. RN 0.51.0, running a release build on Android 7.

leethree:
I'm seeing the exact same issue. Looks like `Subscribable.componentWillUnmount` is called before `Subscribable.componentWillMount` is called.

zhangsn:
Same issue for me. No code changes. It suddenly fails. RN 0.50.4

gveeden:
@a3diti I Could kiss you!

brownieboy:
Getting this error too, on a new test app, created with react-native-cli@2.0.1, which used react-native 0.51.0 and react 16.0.0.

@a3diti 's code change to the `Subscribable.js` file worked for me too, but that's a workaround, not a fix though.  You would have to put that code change in every time you cloned your repository and ran `npm install`.  We need a proper fix for this.

luhui:
same issue for me.  debug build works fine, but fail in release build. RN 0.51.0

[...]

PR #17992, opened 2018-02-15
fixes callstack described in #17348

arichnad:
## Motivation

fixes call-stack described in #17348

Bug shows up as:

com.facebook.react.common.JavascriptException: TypeError: undefined is not an object (evaluating 'this._subscribableSubscriptions.forEach')

This error is located at:  in ScrollView

## Test Plan

I've tested that this does indeed fix the problem described in #17348

## Release Notes
[MINOR] [BUGFIX] [Subscribable.js] fixed javascript exception on startup.


TheSavior:
Duplicate of 17463

PR #17999, opened 2018-02-16
Add tvOS Remote Tap Gestures

ultramiraculous:
## Motivation

On tvOS there is a gesture for tapping a region of the touch pad on the remote. For example, the common use for the left/right tap gesture is to skip -/+ 10s during playback. Currently this gesture is recognized but reported as a generic "left"/"right"/"up"/"down" event, which is the same event type used for swipes. This PR is to add a "tap" event that is separate from the "swipe" event.

## Test Plan

Using `RNTester`'s `TVEventHandlerExample`, the `tap<Left/Right/Top/Bottom>` event type should be displayed when the edge of the remote trackpad are tapped, and `swipe<Left/Right/Up/Down>` should be displayed after a swipe.

## Related PRs

None

## Release Notes

[IOS][BREAKING][ENHANCEMENT][TVRemoteHandler] - Change tvOS Gesture Recognizers to have separate event types for swipes and taps

### DIFF
--- a/React/Base/RCTTVRemoteHandler.h
+++ b/React/Base/RCTTVRemoteHandler.h
@@ -17,10 +17,10 @@ extern NSString *const RCTTVRemoteEventSelect;
 extern NSString *const RCTTVRemoteEventLongPlayPause;
 extern NSString *const RCTTVRemoteEventLongSelect;
 
-extern NSString *const RCTTVRemoteEventLeft;
-extern NSString *const RCTTVRemoteEventRight;
-extern NSString *const RCTTVRemoteEventUp;
-extern NSString *const RCTTVRemoteEventDown;
+extern NSString *const RCTTVRemoteEventTapLeft;
+extern NSString *const RCTTVRemoteEventTapRight;
+extern NSString *const RCTTVRemoteEventTapTop;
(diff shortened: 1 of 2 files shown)


ultramiraculous:
So there's two minor things:

* Currently this is a minor breaking change if you are using the `left`/`right/`up`/`down` events, because those events are broken into tap and swipe events. If keeping the old behavior is desired, the old events could be additionally emitted, but that seems suspect given that the old behavior didn't really make sense in the first place.

* I've gone with "top" and "bottom" for the tap location names, which differs from Apple's usage of "up" and "down". I think this makes more sense spatially, but it might be confusing if you're aware of Apple's naming.

shergin:
cc @dlowder-salesforce

dlowder-salesforce:
@ultramiraculous this looks fine, but should be marked as BREAKING in the release notes.

ultramiraculous:
Sounds good to me @dlowder-salesforce. I looks like I add an entry to the wiki after the change lands?

dlowder-salesforce:
@ultramiraculous no, you edit the "Release Notes" section of your first comment above, and add "[BREAKING]".

PR #18002, opened 2018-02-16
Bump buck to v2018.02.16.01

hramos:
This should fix our current Android test failures.


hramos:
Confirmed, BUCK issue fixed.

PR #18019, opened 2018-02-19
npmignore: ignore tests and fixtures

jeanlauliac:
## Motivation

This try to address #17672 and https://github.com/facebook/metro/issues/139. We should probably not include these folders in the released version of React Native, as they are used only for testing—am I incorrect?

cc @MoOx, @t4deu.

## Test Plan

I ran:

```
npm pack
tar -t -f react-native-1000.0.0.tgz | less
```

Then verified that `__fixture__` was not part of the package.

## Related PRs

https://github.com/facebook/react-native/pull/17672

## Release Notes

[GENERAL] [BUGFIX] [.npmignore] - Do not publish test-specific files


vovkasm:
@jeanlauliac Is there a chance that `__mocks__` (or other) folder contains files that can be useful for testing purposes in projects that use react-native?

For ex: [Libraries/Lists/ListView/__mocks__/ListViewMock.js](https://github.com/facebook/react-native/blob/master/Libraries/Lists/ListView/__mocks__/ListViewMock.js)

jeanlauliac:
> @jeanlauliac Is there a chance that __mocks__ (or other) folder contains files that can be useful for testing purposes in projects that use react-native?

Yes, I think you may be right! Do you know if anyone would have context on it? I'd think mocks are only for RN's own tests, but it's possible it's relied on by some projects.

vovkasm:
We at work currently use `jest.mock('NativeEventEmitter')` in our additional jest setup. It is not critical, we always can implement it ourself. But other users can be potentially more dependent on content in those folders...

jeanlauliac:
Sounds good, I'll remove that. As for `fixtures` and `tests` themselves, I suspect there shouldn't be any use outside of RN development?

jeanlauliac:
@rafeca, @mjesun, thoughts on this? :)

jeanlauliac:
I had to use the technique talked about there: https://github.com/npm/npm/issues/4479. Because the `files` package field overrides the `.npmignore`.

PR #18029, opened 2018-02-20
Added support for using animated values as the target of an interpolation.

DylanVann:
animated

## Motivation

More complex animations can be created by combining an interpolation with an animated `outputRange`.

An example of basic usage is shown here: https://twitter.com/atomarranger/status/965802252597387264

## Test Plan

An example is added to RNTester to demonstrate the usage of this feature with and without `useNativeDriver`.

## Release Notes

[GENERAL] [FEATURE] [Native Animated] - Added support for using animated values as the target of an interpolation.

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -39,6 +39,34 @@ import type {TimingAnimationConfig} from './animations/TimingAnimation';
 import type {DecayAnimationConfig} from './animations/DecayAnimation';
 import type {SpringAnimationConfig} from './animations/SpringAnimation';
 import type {Mapping, EventConfig} from './AnimatedEvent';
+import type {InterpolationConfigType} from './nodes/AnimatedInterpolation';
+
+const interpolateMethod = function(
+  config: InterpolationConfigType,
+): AnimatedInterpolation {
+  console.warn(
+    'The animation.interpolate(config) method will be removed from animated nodes in favour of Animated.interpolate(animation, config).',
+  );
+  return new AnimatedInterpolation(this, config);
+};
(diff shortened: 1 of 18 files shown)


Freddy03h:
It can be cool to also have it for `inputRange` !

DylanVann:
@Freddy03h Yeah that could be cool, it would introduce some weird cases though. Currently `inputRange` is required to be monotonically increasing.

My use case for animating `outputRange` is to animate the ending position of a gesture based animation. The animated value I'm interpolating is being set by `react-native-interactable`.

I think it would make sense to address animating `inputRange` in another PR.

janicduplessis:
Hey thanks for working on this. Just so I understand better do you have a concrete use case where this feature is needed to accomplish a certain behaviour?

DylanVann:
@janicduplessis Thanks for checking it out. The concrete use case I have is involves using `react-native-interactable`.

interactable

On this screen the card animation is driven by `react-native-interactable` which also updates an `AnimatedValue`. That `AnimatedValue` is interpolated to determine the size and position of the profile picture. The size and location of the profile picture also changes based on wether the user is viewing the leaderboard or the profile. So then the desired effect is to be able to animate the ending value of the interpolation.

I'm not a big fan of how I did the implementation of this. Let me know what you think about the proposed breaking change because I think that would make it a lot cleaner.

DylanVann:
I've updated https://github.com/facebook/react-native/pull/18030 so it's only a deprecation, and all tests pass.

I think that change should be made before continuing with this feature.

DylanVann:
Closing in light of `react-native-reanimated` being created, also future changes to React Native's architecture may make it no longer necessary.

PR #18030, opened 2018-02-20
Deprecate interpolate method from animated nodes and add to AnimatedImplementation.

DylanVann:
## Motivation

```js
// Current form.
myAnimation.interpolate(config)
```

```js
// Consistent with other types of animated nodes and animations.
Animated.interpolate(myAnimation, config)
Animated.diffClamp(myAnimation, min, max)
Animated.add(myAnimation, myOtherAnimation)
Animated.sping(myAnimation, config)
Animated.timing(myAnimation, config)
```

- Less code duplication.
- More consistent.
- Will allow for using `AnimatedValue` within `AnimatedInterpolation`.

## Test Plan

Have existing tests pass.

## Could eventually be a breaking change:

* **Who does this affect**: Anyone using `animation.interpolate(config)`.
* **How to migrate**: Switch to using `Animated.interpolate(animation, config)`.
* **Why make this breaking change**: It is more consistent, reduces code duplication, and reduces the risk of creating a circular dependency.
* **Severity (number of people affected x effort)**: This could affect a lot of people. It is relatively easy to fix though. It could probably be fixed using a codemod or even just some regex.

## Release Notes

[GENERAL] [MINOR] [Native Animated] - Deprecate interpolate method from animated nodes and add to AnimatedImplementation.

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -38,6 +38,34 @@ import type {TimingAnimationConfig} from './animations/TimingAnimation';
 import type {DecayAnimationConfig} from './animations/DecayAnimation';
 import type {SpringAnimationConfig} from './animations/SpringAnimation';
 import type {Mapping, EventConfig} from './AnimatedEvent';
+import type {InterpolationConfigType} from './nodes/AnimatedInterpolation';
+
+const interpolateMethod = function(
+  config: InterpolationConfigType,
+): AnimatedInterpolation {
+  console.warn(
+    'The animation.interpolate(config) method will be removed from animated nodes in favour of Animated.interpolate(animation, config).',
+  );
+  return new AnimatedInterpolation(this, config);
+};
+
+// To avoid some code duplication and a circular dependency we
+// are adding the interpolate method directly onto these prototypes.
+// This should eventually be removed.
+//$FlowFixMe
+AnimatedAddition.prototype.interpolate = interpolateMethod;
+//$FlowFixMe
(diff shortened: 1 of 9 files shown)


satya164:
Can we leave the `interpolate` method on the instance with a deprecation warning for now and remove it entirely in a later version so that libraries have time to migrate their code? Otherwise, this will break a ton of apps and libraries.

DylanVann:
The motivation for doing this is to resolve a circular dependency that would be created by something I'm trying to do in the linked PR. So a deprecation warning wouldn't solve that problem.

There's probably some other way to resolve the circular dependency though. The interpolate method could maybe be added onto the classes in `AnimatedImplementation.js`. Then there would be no circular dependency and there would be less code duplication. Although I still think a deprecation warning and adding the `Animate.interpolate` method would be an improvement to the API, but we could definitely go with a deprecation warning.

I'll try that out later today.

DylanVann:
Have to figure out how the flow types would work, but something like this:

```js
const addInterpolateMethod = function(BaseClass) {
  BaseClass.prototype.interpolate = function interpolate(config: InterpolationConfigType): AnimatedInterpolation {
    return new AnimatedInterpolation(this, config);
  };
};

const AnimatedAdditionWithInterpolate = addInterpolateMethod(AnimatedAddition);
const AnimatedValueWithInterpolate = addInterpolateMethod(AnimatedValue);
```

grabbou:
FYI this is already how you do it in Reason with `bs-react-native`, cc: @wokalski

wokalski:
indeed, this is the `bs-react-native` api, although it compiles to the current form.

PR #18033, opened 2018-02-21
Update node-notifier (has mem leak fix)

rickhanlonii:
## Motivation

Noticed that we're on a version of node-notifier that has a leak mentioned [here](https://github.com/mikaelbr/node-notifier/issues/183) and fixed in the newest version. 

## Test plan

Automated tests

## Release Notes
 [INTERNAL] [BUGFIX] [package.json] - Update node-notifier dependency


rickhanlonii:
@hramos thanks for the review! Looks like the android test timed-out - could you retry it?

PR #18037, opened 2018-02-21
Refactor flow const binding

### LINKED ISSUE #16228: Support flow experimental.const_params=true
### Is this a bug report?

Yes

Yes

### Environment

Environment:
  OS:  macOS High Sierra 10.13
  Node:  6.11.2
  Yarn:  1.1.0
  npm:  3.10.10
  Watchman:  4.7.0
  Android Studio:  Not Found

### Steps to Reproduce

(Write your steps here:)

1. Create a new react-native app with create-react-native-app
2. `yarn add --dev flow-bin@^0.49.1` (same version specified in `.flowconfig`)
3. Add `experimental.const_params=true` under `[options]` in `.flowconfig`
4. `./node_modules/.bin/flow`

### Expected Behavior

`No errors.`

(Write what you thought would happen.)

### Actual Behavior

```
$ flow
node_modules/fbjs/lib/UnicodeBidi.js.flow:90
 90:   fallback = fallback || UnicodeBidiDirection.NEUTRAL;
       ^^^^^^^^ fallback. const param cannot be reassigned (see experimental.const_params=true in .flowconfig)
 89: function resolveBlockDir(str: string, fallback: ?BidiDirection): BidiDirection {
                                           ^^^^^^^^^^^^^^^^^^^^^^^^ const param fallback

node_modules/fbjs/lib/UnicodeBidi.js.flow:92
 92:     return fallback;
                ^^^^^^^^ null. This type is incompatible with the expected return type of
 89: function resolveBlockDir(str: string, fallback: ?BidiDirection): BidiDirection {
                                                                      ^^^^^^^^^^^^^ string enum

…[removed for brevity]…

... 88 more errors (only 50 out of 138 errors displayed)
To see all errors, re-run Flow with --show-all-errors
```

### Reproducible Demo

https://github.com/ckknight/ReactNative-const_params-repro

klvs:
Refactor to treat function parameters as const binding.

Flow config can now set experimental.const_params=true without throwing any errors.

Mostly uses the solution proposed [here](https://eslint.org/docs/rules/no-param-reassign)

## Motivation

Fixes #16228 

## Test Plan

Run `yarn flow`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -51,5 +51,7 @@ suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(<VERSION>\\)? *\\(site=[a-z,_]*
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
 suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
 
+experimental.const_params=true
+
 [version]
 ^0.66.0
--- a/IntegrationTests/AsyncStorageTest.js
+++ b/IntegrationTests/AsyncStorageTest.js
@@ -56,10 +56,11 @@ function expectEqual(lhs, rhs, testname : string) {
 }
 
 function expectAsyncNoError(place, err) {
-  if (err instanceof Error) {
-    err = err.message;
+  let error = err;
+  if (error instanceof Error) {
+    error = error.message;
   }
-  expectTrue(err === null, 'Unexpected error in ' + place + ': ' + JSON.stringify(err));
+  expectTrue(error === null, 'Unexpected error in ' + place + ': ' + JSON.stringify(err));
 }
 
 function testSetAndGet() {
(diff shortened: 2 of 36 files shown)


TheSavior:
I really *really* want to do this. Unfortunately, we have a monorepo at facebook and everything is under a single flow config. If we merge this, it wouldn't change how flow treats these files internally and we'd be likely to break flow for everyone in open source. 

Fortunately, I actually wrote a pretty intensive codemod to do it at Facebook so we could convert our entire codebase and turn this flow config on. Unfortunately, there is an issue with the codemod that has kept me from converting our entire codebase. The codemod erroneously converts:

```
function foo(myVar) {
  var myVar = 2; // valid shadowing of the function argument
}
```
into
```
function foo(myVar) {
  let localMyVar = myVar;
  var localMyVar = 2; // Uncaught SyntaxError: Identifier 'localMyVar' has already been declared
}
```

If you would be interested in detecting this scope issue and getting the codemod to handle it, we'd be able (and happy) to make this change to react native. The codemod would of course also be open source then. :)

PR #18039, opened 2018-02-21
Remove unused reference in Text.js

rozele:
Previously, the Platform module was used to detect when RCTVirtualText should be used. Recently, this has changed to detecting the availability of a native virtual text component on the UIManager. The import for the Platform module can be deleted.

## Motivation

I was cleaning up source code copied in react-native-windows and noticed this reference is no longer used in Text.js.

## Test Plan

Run jest tests.

## Related PRs

N/A


chirag04:
@facebook-github-bot shipit

PR #18040, opened 2018-02-21
Fix `onLayout` prop for TextInput on Android

rozele:
When the `autogrow` prop was removed from `TextInput` on Android, the `_onLayout` helper method was removed. This helper method implemented the hook required to make `autogrow` work, then dispatched the `onLayout` event to the handler in `this.props`. This change points passes the `onLayout` handler from `this.props` directly to the inner component.

## Motivation

I was updating copied code in react-native-windows and noticed this bug.

## Test Plan

Run jest tests.

## Related PRs

N/A

## Release Notes

 [ANDROID][MINOR][BUGFIX] [TextInput] - Fix `onLayout` prop for TextInput on Android

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -880,7 +880,7 @@ const TextInput = createReactClass({
 
     return (
       <TouchableWithoutFeedback
-        onLayout={this._onLayout}
+        onLayout={props.onLayout}
         onPress={this._onPress}
         accessible={this.props.accessible}
         accessibilityLabel={this.props.accessibilityLabel}


Libraries/Components/TextInput/TextInput.js
```diff
@@ -880,7 +880,7 @@ const TextInput = createReactClass({
 
     return (
       <TouchableWithoutFeedback
-        onLayout={this._onLayout}
```
chirag04: can we do this.props.onLayout to maintain equality?

rozele: @chirag04 - `props` is copied into a local variable:
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L836

This is exactly how it's done for iOS:https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L771
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L822

janicduplessis:
Good catch!

@facebook-github-bot shipit

PR #18056, opened 2018-02-22
Remove Platform check from WebSocket module

rozele:
WebSocket uses the Platform module to check how many arguments for the `close` method should be used. In react-native-windows, we have the same number of arguments for `close` as Android, so we're prevented from using this module as-is because of the platform check (see https://github.com/Microsoft/react-native-windows/blob/master/Libraries/WebSocket/WebSocket.windows.js#L136).

By switching to an argument count check, this module becomes more useful cross-platform. If you'd like to keep the platform check, I'm also open to inverting the conditional, e.g., `if (Platform.OS !== 'ios')`.

## Motivation

I'd like to minimize the amount of code I need to copy over to react-native-windows for platform-specific overrides.

## Test Plan

Run jest tests.

## Related PRs

N/A

## Release Notes

[GENERAL][MINOR][ENHANCEMENT][Libraries/WebSocket/WebSocket.js] - Better enable cross-platform support of WebSocket.js

### DIFF
--- a/Libraries/WebSocket/WebSocket.js
+++ b/Libraries/WebSocket/WebSocket.js
@@ -202,7 +202,7 @@ class WebSocket extends EventTarget(...WEBSOCKET_EVENTS) {
   }
 
   _close(code?: number, reason?: string): void {
-    if (WebSocketModule.close.length > 1) {
+    if (WebSocketModule.close.length === 3) {
       // See https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
       const statusCode = typeof code === 'number' ? code : CLOSE_NORMAL;
       const closeReason = typeof reason === 'string' ? reason : '';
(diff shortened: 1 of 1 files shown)


Libraries/WebSocket/WebSocket.js
```diff
@@ -203,7 +202,7 @@ class WebSocket extends EventTarget(...WEBSOCKET_EVENTS) {
   }
 
   _close(code?: number, reason?: string): void {
-    if (Platform.OS === 'android') {
+    if (WebSocketModule.close.length > 1) {
```
TheSavior: How about `WebSocketModule.close.length === 3`?

rozele: Works for me. Coming right up.

PR #18058, opened 2018-02-22
[ANDROID] [BUGFIX] [WebView] Fix touchmove event inside the webview 

vomchik:
## Motivation

After adding `touchmove` event listener it triggers only once and then triggers `touchcanel` event.
This PR fix it and now `touchmove` event continue triggering until not finished the touch.

## Test Plan

Use WebView component.

```js

const html = `
  <html>

  <head>
    <title>test</title>
  </head>

  <body>
    <script>
      document.addEventListener('touchmove', console.log.bind(console.log, 'move'));
      document.addEventListener('touchcancel', console.log.bind(console.log, 'cancel'));
    </script>
  </body>

  </html>
`;

<WebView source={{ baseUrl: '/', html }}   />
```

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/webview/ReactWebViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/webview/ReactWebViewManager.java
@@ -22,6 +22,7 @@ import android.graphics.Picture;
 import android.net.Uri;
 import android.os.Build;
 import android.text.TextUtils;
+import android.view.MotionEvent;
 import android.view.ViewGroup.LayoutParams;
 import android.webkit.ConsoleMessage;
 import android.webkit.GeolocationPermissions;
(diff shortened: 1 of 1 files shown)


hramos:
Thanks for the PR! I looked at your proposal but it's not clear what issue it is addressing. Is there a Snack that demonstrates the problem?

Do you think we can add a test to ensure this does not happen again?

vomchik:
Hey, unfortunately, it can't be demonstrated via Snack because my changes require run the demo from the source.

vomchik:
Also, I don't know how I can test it. Maybe you have some suggestion?

vomchik:
BTW, if it will helpful i can record the video.

hramos:

RSNara:
@vomchik, thanks for this PR! 

To address your earlier question, a video would be helpful for the reviewers. Oftentimes, reviewers are busy people juggling several issues at once, so any help that makes issues/PRs easier to understand makes those issues/PRs more likely to be solved/landed. 

Unfortunately, we're planning on removing WebViews from this repository. They're now going to live in [react-native-community/react-native-webview](https://github.com/react-native-community/react-native-webview). Do you mind opening up this PR there? Since the scope of that repository is small, they should be able to get issue to this fairly quickly.

Issue #14706, opened 2017-06-23
Support iOS accessibilityValue and accessibilityHint


randyzhao:
Currently we only support `accessibilityLabel`. It would be great if we could also support [`accessibilityValue` ](https://developer.apple.com/documentation/objectivec/nsobject/1615117-accessibilityvalue?language=objc) and [`accessibilityHint`](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint) for iOS.

Platform: iOS

hramos:
Thanks for opening the issue, however it looks like a feature request. As noted in the [Issue template](https://github.com/facebook/react-native/blob/master/.github/ISSUE_TEMPLATE.md) we'd like to use the GitHub issues to track bugs only. Can you implement the feature as a standalone npm module? If not consider sending a pull request or a create an entry on [Canny](https://reactnative.canny.io/feature-requests). It has a voting system and if the feature gets upvoted enough it might get implemented by someone from the community. Closing this now, thanks for understanding!

cannona:
I do not believe that this is a feature, unless accessibility in general is considered a feature. This is a basic part of the accessibility API of IOS, and is necessary for making some types of interactions accessible.

PR #18093, opened 2018-02-26
Add accessibilityHint for iOS

draperunner:
This adds the accessibilityHint for View, Text and Touchable* on iOS.
The accessibilityHint provides some more information about an element
when the accessibilityLabel is not enough.

## Motivation
The accessibilityHint is a core accessibility property on iOS. 

From https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint:
> An accessibility hint helps users understand what will happen when they perform an action on the accessibility element when that result is not obvious from the accessibility label.

Related issue: https://github.com/facebook/react-native/issues/14706

## Test Plan
The npm scripts `test`, `flow`, `lint` and `prettier` are satisfied.

I added a couple of examples to the RNTester app. The Accessibility Inspector on Mac helps debugging accessibility stuff on a simulator, but it does not show the accessibilityHint. Therefore I tested the RNTester app on an iPhone 8 device using VoiceOver to verify the hint functionality. It works fine, and I've tested disabling and enabling "read hints" in the VoiceOver settings on the phone.

## Related PRs
https://github.com/facebook/react-native-website/pull/222

## Release Notes
[IOS][FEATURE][Accessibility] - Add accessibilityHint for View, Text, Touchable* on iOS


React/Views/RCTView.m
```diff
@@ -154,6 +154,15 @@ - (NSString *)accessibilityLabel
   return RCTRecursiveAccessibilityLabel(self);
 }
 
+- (NSString *)accessibilityHint
```
janicduplessis: Why is this needed? Can't we just use the prop directly?

draperunner: You're right, it is not needed. I tested after removing it, and it still works. I've updated this in the latest commit, 196ce0658b7c71ec5f30b88018d784b49272cc02

janicduplessis:
Thanks for the PR!

@facebook-github-bot shipit

draperunner:
Hi @janicduplessis, seems like facebook-github-bot met some issues. Anything I can do?

hramos:
@draperunner Janic does not have visibility into the failure but I can help in this regard. It looks like this is held back due to a failing snapshot test. Getting this PR merged will require some manual intervention on my part.

It's possible that the failing test is unrelated to this PR, so let me run the internal diff through the test suite one more time. Please **do not add any more commits** to this PR as doing so will prevent it from landing. Thanks!

janicduplessis:
@hramos This adds a prop to View so it is very possible that some snapshots need to be updated.

hramos:
Indeed that's the case. The failing snapshot tests are all due to `accessibilityHint` being added to the props. I'll take care of this.

hramos:
Snapshot tests updated and test suites are currently running. I'll check again in tomorrow morning and land this if all looks green.

hramos:
```
** BUILD FAILED **
The following commands produced analyzer issues:
Analyze Base/RCTModuleMethod.mm
Analyze Modules/RCTUIManager.m
(2 commands with analyzer issues)
The following build commands failed:
Ld /Users/facebook/Library/Developer/Xcode/DerivedData/RNTester-gmzkrbslacaqklccqmdfkejruuag/Build/Products/Debug-iphonesimulator/RNTester.app/RNTester normal x86_64
```

draperunner:
Any updates on this @hramos?

janicduplessis:
@hramos @draperunner  The PR doesn't touch any of the files from the log, maybe rebase on master and we can try shipping again.

draperunner:
@hramos @janicduplessis Hi guys. I have now rebased this PR and added Flow types.

janicduplessis:
@hramos Wanna try landing this again? It will still require some snapshot updates.

knitcodemonkey:
Cc @thesavior

hramos:
This will be landing soon.

PR #18100, opened 2018-02-26
Add platform hook for View style props

rozele:
Similar to the `ViewPropTypes` hook to allow customization of prop types for iOS and Android, it's also useful to allow platforms to override style prop types.

## Motivation

(Write your motivation here.)

## Test Plan

Run jest tests.

## Related PRs

https://github.com/facebook/react-native/pull/15175

## Release Notes

[GENERAL][ENHANCEMENT][MINOR][View] - Add platform hook for adding View style props

### DIFF
--- a/Libraries/Components/View/PlatformViewStylePropTypes.android.js
+++ b/Libraries/Components/View/PlatformViewStylePropTypes.android.js
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @providesModule PlatformViewStylePropTypes
+ * @flow
+ */
+
+module.export = {
+  /**
+   * (Android-only) Sets the elevation of a view, using Android's underlying
+   * [elevation API](https://developer.android.com/training/material/shadows-clipping.html#Elevation).
+   * This adds a drop shadow to the item and affects z-order for overlapping views.
+   * Only supported on Android 5.0+, has no effect on earlier versions.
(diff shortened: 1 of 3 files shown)


hramos:
You asked elsewhere how this might impact the docs. Right now, `elevation` is documented as part of the [View Style Props](http://facebook.github.io/react-native/docs/view-style-props.html) docs. I've opened a PR on the docs repo with a proposal on how this might be documented: https://github.com/facebook/react-native-website/pull/223

hramos:
It looks like with this change, some of our own internal apps are failing to build. I'll need to add this to my queue of diffs that need to be manually landed.

rozele:
@hramos - any luck?

hramos:
I haven't gotten around to it yet. Some of our apps are failing to build with this error:

```
Invariant Violation: "elevation" is not a valid style property.
```

rozele:
@hramos - I imagine this may be because the `elevation` prop is set on iOS even if it's not used.

rozele:
@hramos - perhaps we just leave `elevation` in the original `ViewStylePropTypes` file, any chance you could produce an isolated repro of the tests so I can try to come up with a workaround? It seem either:
1. We shouldn't produce errors when a platform-specific prop is set that isn't relevant to another platform.
2. We should provide a simple syntax for setting a prop on only the relevant platform, something like:
```
...Platform.select({
  ios: {},
  android: {
    elevation: 1
  },
  windows: {}
})
```

cpojer:
We are currently working on getting rid of PropTypes altogether and using Flow Types to describe View props. I'm gonna close this PR and I'm sorry we didn't get to it earlier.

Issue #18087, opened 2018-02-25
cannot catch error from fetch call when used a malformed URL


Trenrod:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS: Windows 10
  Node: 6.12.3
  Yarn: Not Found
  npm: 3.10.10
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Version  3.0.0.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.53.3 => 0.53.3

Target Platform: Android 6 Emulator (From the Getting Started howto)

### Steps to Reproduce

Cannot catch the fetch call when used a malformed url.

1. Call fetch() with a malformed url like 'asd' or 'https:://facebook.github.io'
2. try to catch with '.catch' when called as promise or try/catch within async/await

### Expected Behavior

catch should be executed

### Actual Behavior

red screen shows up with "unexpected url: [URL USED WITHIN THE CALL]"

### Reproducible Demo

```
  async asyncFetch(url) {
    try {
      console.log('DEBUG - START FETCHING');
      const response = await fetch(url);
      console.log('DEBUG - END FETCHING');
      const xml = await response.text();
      const json = xml2json.parse(xml);
      return { data: json };
    } catch (error) {
      console.log('DEBUG - START ERROR: ' + error.message);
      throw error;
    }
  }

  const url = 'httasd:afdasdfads';
  this.asyncFetch(url)
  .catch(err => {
    console.log('DEBUG - Error: ' + err.message);
  })
```

Actual console output is:
DEBUG - START FETCHING

Note: Same issue like #17843

ksegla:
fetch has been having this very bad problem for some time now, crashing apps in production whenever something it doesn't like happen. At this point, I'm looking for good alternatives :-(

Dorentin1997:
Check if url is good and returns something. Send an request at that link with postman check what it returns.

Trenrod:
@Dorentin1997 thank you for your comment.
Yes, a workaround can be using  'url-parse' to validate the url.
But my assumption is that if someone enters an invalid url, fetch should throw a handleble exception and not crashing the whole application.

Dorentin1997:
Use url-parse do the logic  there,  if it`s wrong throw the exception there.

Trenrod:
[...]

PR #18103, opened 2018-02-26
Catch exception and report it when making a network request with invalid URL on Android

### LINKED ISSUE #7436: fetch("aaa") has incorrect behavior
Unable to catch error thrown from `fetch` with invalid url. Instead I get a message about "unexpected url".
- [x] Provide a **minimal code snippet** / [rnplay](https://rnplay.org/) example that reproduces the bug.

``` js
try {
   fetch("aaa");
} catch (ex) { }
```
- [x] Provide **screenshots** where appropriate
  image
- [x] What's the **version** of React Native you're using?
  0.24.1
- [x] Does this occur on iOS, Android or both?
  Android
- [x] Are you using Mac, Linux or Windows?
  OS X

jcurtis:
Currently if you invoke `fetch()` with an invalid URL ("aaa" for
example) you cannot catch the error in javascript since it's not
reported. Instead the entire app crashes.

Fixes #7436 and #18087

Hopefully.

## Motivation

Fix using fetch on Android with user generated input. 

## Test Plan

Added relevant unit test

`./scripts/run-android-local-unit-tests.sh` all pass

## Release Notes

[ANDROID] [BUGFIX] [fetch] - Allow "unexpected url" exception to be caught on Android when using fetch

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -266,7 +266,13 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
       return;
     }
 
-    Request.Builder requestBuilder = new Request.Builder().url(url);
+    Request.Builder requestBuilder;
+    try {
+      requestBuilder = new Request.Builder().url(url);
+    } catch (Exception e) {
+      ResponseUtil.onRequestError(eventEmitter, requestId, e.getMessage(), null);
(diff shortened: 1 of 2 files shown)


jcurtis:
@laurent22 I'm not sure that's necessary. We're just testing that the error gets reported and not the actual URL builder itself. If it's any invalid URL it should get reported.

satya164:
We also want to add check for valid URL on the JS side to synchronously report the error, but this is a good start.

@facebook-github-bot shipit

Issue #17974, opened 2018-02-14
IndexOutOfBoundsException when using Google Indic Keyboard with RN 0.53.0


SuhairZain:
### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment
Environment:
  OS: macOS High Sierra 10.13.2
  Node: 8.4.0
  Yarn: 0.27.5
  npm: 5.3.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.53.0 => 0.53.0

Target Platform: Android (6.0.1)

### Steps to Reproduce
1. Create and run a new project with RN 0.53.0, using `react-native init`. (Most likely exists for projects created using other methods as well, didn't check.)
1. Switch the keyboard to Google Indic Keyboard from phone's input settings
1. Steps for reproducing the crash with emojis:
a. Create a TextInput with no additional props (or switch to the master branch of the provided repo).
a. Type an emoji and press the backspace twice.
1. Steps for reproducing the crash with any character:
a. Create a TextInput with autoCorrect={false} (or switch to the auto-correct branch of the provided repo).
a. Type any character and press the backspace once.

### Expected Behavior
The emojis and characters get cleared with no crashes

### Actual Behavior
![screenshot_2018-02-14-10-04-04-422_com android settings](https://user-images.githubusercontent.com/5249794/36188740-111d28e4-1174-11e8-8e1e-ff493049e922.png)

The app crashes on clearing the emojis, with no additional props set on the TextInput
![2018-02-14 10_37_49](https://user-images.githubusercontent.com/5249794/36188618-380c45b2-1173-11e8-97f3-a3d355532b3d.gif)
The app crashes on clearing the last character in case autoCorrect={false} is set
![2018-02-14 10_38_17](https://user-images.githubusercontent.com/5249794/36188619-386d66ee-1173-11e8-945a-2ecb96e13f22.gif)

### Reproducible Demo
https://github.com/SuhairZain/react-native-gindic-keyboard-crash
The `master` branch crashes when deleting an emoji character
The `auto-correct` branch crashes when deleting the last character in the input

[...]

Issue #17922, opened 2018-02-09
TextInput secureTextEntry crashing on Android custom keyboard?


apparition47:
### Is this a bug report?

Yes.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes.

### Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 9.4.0
  Yarn: 1.0.2
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.2.0 => 16.2.0
  react-native: 0.53.0 => 0.53.0

Tested on Devices: 
* Plus One Japan Limited FTJ161B (Android 6.0, API 23)
* Samsung Galaxy Note 8 (Android 7.1.1)

### Steps to Reproduce

Using [`com.justsystems.atokmobile.tv.service` custom keyboard](https://play.google.com/store/apps/details?id=com.justsystems.atokmobile.tv.service&hl=ja) then typing into a `<TextInput secureTextEntry={true}>` field yields the following crash:

>13:16:19.098 4410-4410/com.example.dev E/AndroidRuntime: FATAL EXCEPTION: main
>Process: com.example.dev, PID: 4410
>java.lang.IndexOutOfBoundsException: charAt: -1 < 0
>at android.text.SpannableStringBuilder.charAt(SpannableStringBuilder.java:122)
>at com.facebook.react.views.textinput.ReactEditTextInputConnectionWrapper.setComposingText(ReactEditTextInputConnectionWrapper.java:104)
>at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:340)
>at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:78)
>at android.os.Handler.dispatchMessage(Handler.java:111)
>at android.os.Looper.loop(Looper.java:207)
>at android.app.ActivityThread.main(ActivityThread.java:5728)
>at java.lang.reflect.Method.invoke(Native Method)
>at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:789)
>at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:679)

1. Focus the `<TextInput>`
[...]

PR #18114, opened 2018-02-27
Fix crashes onKeyPress Android

joshyhargreaves:
## Motivation
There appear to be two different types of crashes related to the recent addition of `onKeyPress` on Android introduce in `0.53`. This PR addresses the cause of both of them.

Firstly, it seems possible to get an `indexOutOfBoundsException` with some 3rd-party keyboards as observed in https://github.com/facebook/react-native/issues/17974 & https://github.com/facebook/react-native/issues/17922. I have simplified the backspace determining logic slightly, and also put in an explicit check for zero case so it is not possible to get an indexOutOfBoundsException & it should make sense in the context of the onKeyPress logic.

Secondly, it appears that `EditText#onCreateInputConnection` can return null. In this case, if we set `null` to be the target of our subclass of `ReactEditTextInputConnectionWrapper`, we will see the crashes as seen [here](https://github.com/facebook/react-native/issues/17974#issuecomment-368471737), whereby any of methods executed in the `InputConnection` interface can result in a crash. It's hard to reason about the state when `null` is returned from `onCreateInputConnection`, however I would might reason that any soft keyboard input cannot update the `EditText` with a `null` `input connection`, as there is no way of interfacing with the `EditText`. I'm am not sure, if there is a later point where we might return/set this input connection at a later point? As without the `InputConnection` onKeyPress will not work. But for now, this will fix this crash at least.

## Test Plan
I have not managed to reproduce these crashes myself yet, but users have confirmed that the `indexOutOfBounds` exception is fixed with the 'zero' case and has been confirmed on the respective issues https://github.com/facebook/react-native/issues/17974#issuecomment-368471737.

For the `null` inputConnection target case, I have verified that explicitly setting the target as null in the constructor of `onCreateInputConnection` results in the same stack trace as the one linked. Here is also a [reference](https://github.com/stripe/stripe-android/pull/392/files#diff-6cc1685c98457d07fd4e2dd83f54d5bb) to the same issue closed with the same fix for another project on github.

It is also important to verify that the behavior of `onKeyPress` still functions the same after this change, which can be verified by running the RNTesterProject and the `KeyboardEvents` section in `InputText`.
The cases to check that I think are important to check are: 
- Cursor at beginning of input & backspace
- Return key & return key at beginning of input
- Select text then press delete
- Selection then press a key
- Space key
- Different keyboard types

This should not be a breaking change. 

## Release Notes
 [ANDROID] [BUGFIX] [TextInput] - Fixes crashes with TextInput introduced in 0.53.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -172,14 +172,16 @@ public class ReactEditText extends EditText {
   @Override
   public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
     ReactContext reactContext = (ReactContext) getContext();
-    ReactEditTextInputConnectionWrapper inputConnectionWrapper =
-        new ReactEditTextInputConnectionWrapper(super.onCreateInputConnection(outAttrs), reactContext, this);
+    InputConnection inputConnection = super.onCreateInputConnection(outAttrs);
+    if (inputConnection != null) {
+      inputConnection = new ReactEditTextInputConnectionWrapper(inputConnection, reactContext, this);
+    }
 
     if (isMultiline() && getBlurOnSubmit()) {
       // Remove IME_FLAG_NO_ENTER_ACTION to keep the original IME_OPTION
       outAttrs.imeOptions &= ~EditorInfo.IME_FLAG_NO_ENTER_ACTION;
     }
-    return inputConnectionWrapper;
+    return inputConnection;
   }
 
   @Override
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditTextInputConnectionWrapper.java
@@ -92,14 +92,15 @@ class ReactEditTextInputConnectionWrapper extends InputConnectionWrapper {
     int previousSelectionEnd = mEditText.getSelectionEnd();
     String key;
     boolean consumed = super.setComposingText(text, newCursorPosition);
+    int currentSelectionStart = mEditText.getSelectionStart();
     boolean noPreviousSelection = previousSelectionStart == previousSelectionEnd;
-    boolean cursorDidNotMove = mEditText.getSelectionStart() == previousSelectionStart;
-    boolean cursorMovedBackwards = mEditText.getSelectionStart() < previousSelectionStart;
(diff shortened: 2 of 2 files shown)


kelset:
It looks like the only test that failed is because of bad credentials :/

But thanks for your PR @joshyhargreaves!

hramos:
The bad credential issue is fixed on master. It's OK to ignore the failing step, as it's only trying to run Danger, our PR linter. Everything else appears to check out.

joshyhargreaves:
@hramos do you think it's worth also landing this PR on 0.54?

hramos:
@joshyhargreaves I opened #18128 where we can discuss further.

grabbou:
Since it appears to me that some of the crashes were affecting users on `0.53.x`, do you think we should backport it there too?

kelset:
TBH I think it should depend on how 'breaking' is upgrading to 0.54. If it's a big jump then it may be worth it, if it's ex. only upgrading Flow then no need.

grabbou:
Yeah, let's actually wait for someone to request it (esp. who is blocked) than being proactive in this space.

krewllobster:
I am experiencing this bug using react-native 0.53.0. I get the following stack trace after app crashes when deleting text in a <TextInput /> element. Running on a USB connected Moto X2 via "react-native run-android". Is there a way for me to get this patch without upgrading to react-native 0.5.4?

Stack Trace

java.lang.IndexOutOfBoundsException: charAt: -1 < 0
at android.text.SpannableStringBuilder.charAt(SpannableStringBuilder.java:117)
at com.facebook.react.views.textinput.ReactEditTextInputConnectionWrapper.setComposingText(ReactEditTextInputConnectionWrapper.java:104)
at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:340)
at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:78)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:148)
at android.app.ActivityThread.main(ActivityThread.java:5443)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:728)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:618)

adhithiravi:
We are experiencing this crash too. Any plans on merging this PR soon?

adhithiravi:
Do you know when this will be merged over, this crash is affecting us on android.

hramos:
This fix has been on master for close to a year.

PR #17335, opened 2017-12-24
Add option to hide context menu for TextInput

amhinson:
## Motivation

There is currently no way to disable to context menu that automatically appears over a TextInput. This is especially troublesome if you would like to disable the user from pasting text into certain fields. This PR adds a `contextMenuHidden` property to TextInput that will hide it.

## Test Plan

I'm not sure if testing is necessary here. I would be happy to investigate further on how this would be tested, if deemed necessary!

## Related PRs

https://github.com/facebook/react-native-website/pull/95

## Release Notes

[FEATURE][TextInput] - Added `contextMenuHidden` property

PR #18125, opened 2018-02-27
Add option to hide context menu for TextInput #17335

amhinson:
## Motivation

There is currently no way to disable to context menu that automatically appears over a TextInput. This is especially troublesome if you would like to disable the user from pasting text into certain fields. This PR adds a `contextMenuHidden` property to TextInput that will hide it.

## Test Plan

I'm not sure if testing is necessary here. I would be happy to investigate further on how this would be tested, if deemed necessary!

## Related PRs

https://github.com/facebook/react-native-website/pull/95

## Release Notes

[FEATURE][TextInput] - Added `contextMenuHidden` property


amhinson:
@hramos Here is the new PR.

Balasnest:
Its still showing the context menu. Able to paste into the field.

Balasnest:
In which react-native version this works?

amhinson:
This hasn’t been included in a release yet.

hramos:
You can always see what release any commit is in by clicking on the link and looking at the tags. This commit landed only 7 days ago, so I wouldn't expect it to be on a stable release until 0.55.

robinheinze:
@amhinson Were you able to confirm that this works on Android? I'm using 0.55.4, which includes the new prop, and it works on iOS, but on Android, the context menu still appears when tapping on text in the input.

d2luu:
@amhinson 
Relate to issue @robinheinze mentioned above.
On Android, this not works by double-tap to TextInput field. I read your PR and found that only handle onLongClick case, not for double-tap case. So, the context menu still appears on Android just by double-tap to TextInput field.

amhinson:
Ah good catch! I was definitely only accounting for long press.

zmefz:
@amhinson Can you please add double click handler as the issue still present after almost a year? The solution works perfectly for the long press but not for the double press. Thanks a lot in advance!

PR #18131, opened 2018-02-28
Fix android duplicate linking

### LINKED ISSUE #18053: react-native-cli link infinite android link
### Is this a bug report?

Yes : 
I have updated my package.json to react-native 0.53 and now my android depencencies are not linking once.

### Environment

  **Environment:**
  OS: macOS High Sierra 10.13.3
  Node: 9.5.0
  Yarn: Not Found
  npm: 5.6.0
  Watchman: Not Found

**Packages: (wanted => installed)**
  react: ^16.2.0 => 16.2.0
  react-native: 0.53.3 => 0.53.3

  Target Platform: iOS (10.3)

### Steps to Reproduce

1. upgrade react-native
2.npm install
3.react-native link

### Reproducible Demo

You can view my Stack Overflow post : 
[infinite-react-native-link-with-android-dependencies](https://stackoverflow.com/questions/48895308/infinite-react-native-link-with-android-dependencies)

guperrot:
android isInstalled has a different signature than ios counterpart

## Motivation

Run `react-native link` on a project with an android library that is already linked: the check will fail and will link again.

This should fix #18053

## Test Plan

AppCenter SDK for React-Native uses linking. Running react-native link on appcenter package would constantly add the imports, the moduleConfig strings and add duplicated code to MainApplication.java without this patch.

I also tested it didn't break ios linking.

## Release Notes

[ANDROID] [BUGFIX] Fix `react-native link` adding duplicate code and configuration when running a second time.

### DIFF
--- a/local-cli/link/link.js
+++ b/local-cli/link/link.js
@@ -49,7 +49,8 @@ const linkDependency = (platforms, project, dependency) => {
         return null;
       }
 
-      const isInstalled = linkConfig.isInstalled(project[platform], dependency.config[platform]);
+      const isInstalledParameter = platform == "android" ? dependency.name : dependency.config[platform];
+      const isInstalled = linkConfig.isInstalled(project[platform], isInstalledParameter);
 
       if (isInstalled) {
         log.info(chalk.grey(`Platform '${platform}' module ${dependency.name} is already linked`));


local-cli/link/link.js
```diff
@@ -49,7 +49,8 @@ const linkDependency = (platforms, project, dependency) => {
         return null;
       }
 
-      const isInstalled = linkConfig.isInstalled(project[platform], dependency.config[platform]);
+      const isInstalledParameter = platform == "android" ? dependency.name : dependency.config[platform];
+      const isInstalled = linkConfig.isInstalled(project[platform], isInstalledParameter);
```
grabbou: That makes sense, although I would be much more in favor of having it use the same signature. 

CC: @Kureev

Kureev:
Can you please elaborate why we need an exception for Android? I'd also prefer to keep signature the same.

guperrot:
@Kureev the object ios uses, I debugged it and does not contain the `dependency.name` that Android needs. Are you suggesting having 3 parameters instead of 2?

dseipp:
`react-native link` on Android is definitely broken when moving from react-native 0.53.0 to 0.53.2.  New entries are created every time the command is executed.

Seems to be this commit:  https://github.com/facebook/react-native/commit/a40bfa730e05c68da49e6f217ae0f161dcc7ba98#diff-ca7cc59db13ac8eb27fe0a9a6c9debee

guperrot:
@Kureev @grabbou any suggestion for going forward? This impacts every single library having native android code. Without this patch every single time you run `react-native link` it will duplicate everything in `build.gradle` (compile project statements), duplicate imports in MainApplication.java, duplicate modules returned in the same file and duplicate string resources, and duplicate settings.gradle changes.

This patch fixes the regression when the code was refactored in https://github.com/facebook/react-native/commit/a40bfa730e05c68da49e6f217ae0f161dcc7ba98#diff-ca7cc59db13ac8eb27fe0a9a6c9debee

local-cli/link/link.js
guperrot: The thing is `dependency.config[platform]` does not contain `dependency.name`, should we add `dependencyName` to `dependency.config[platform]` object then?

Titozzz: Unlink suffers the same issue for android and iOS unlink is broken

alien3d:
Been  hit with this bug.. no matter what react-native  version.. new create and old  also effected.

grabbou:
I've pinged @Kureev to look into it - let's wait for a day or two and then, I'll cherry-pick this back to the releases that are affected. I am really sorry for any inconvenience.

local-cli/link/link.js
rozele: Let's just add the `name` parameter to the `isInstalled` signature.

rozele:
I am even more sorry for this. This is absolutely not @grabbou's fault - I had mixed up a two branches where I was working on support for linking platforms other than iOS and Android and ended up testing the wrong code.  Very sorry @grabbou and @Kureev that we have to patch those releases again. :-(

guperrot:
Hi @rozele do you confirm #18207 supersedes this one? Should I close mine?

rozele:
@guperrot - #18207 includes this fix in the style requested by @Kureev as well as another fix for unlinking iOS pointed out by @Titozzz

buixuanthe2212:
- react-native upgrade
and choose y change new version
- yarn install 
- react-native link

;))

PR #18136, opened 2018-02-28
Expose React Native version as a symbol rather than macro

LeoNatan:
## Motivation

Add RCTGetReactNativeVersion() to expose version in native code. Right now, version is exposed internally to RN using a MACRO constant. This exposes a symbol (function) that can be called to retrieve the React Native version in iOS.

Also exposed RCTVersion.h as a public header in the React project so it is available to developers.

The motivation behind this is for https://github.com/wix/detox —we need to know what RN version the user has, if any, so we can properly handle support and abstract differences.

## Test Plan

Ran bump-oss-version.js to ensure the template is applied properly. Also compiled the project to make sure nothing is broken.

## Release Notes

 [IOS] [ENHANCEMENT] [RCTVersion.h] - Expose version as a compile-time symbol for native queries


ide:
@facebook-github-bot shipit

LeoNatan:
Thanks!

PR #18164, opened 2018-03-02
Add devDependencies support for templates

lucianomlima:
Add devDependencies support to React Native templates.
Template can have a devDependencies.json file with devDependencies inside.
Using separate files to dependencies and devDependencies, it maintains compatibility with the current version.

## Motivation

Allows React Native templates to have devDependencies, which can help applications to have better organization, quality and testability. It's possible start a new app with some dependencies for dev support like prettier, reactotron, eslint packages and others.

## Test Plan

Add a devDependencies.json file with at least one dependency (like prettier)

## Release Notes

[CLI] [FEATURE] [local-cli/generator/templates.js] - Add support to devDependencies for react native templates

### DIFF
--- a/local-cli/generator/templates.js
+++ b/local-cli/generator/templates.js
@@ -99,6 +99,7 @@ function createFromBuiltInTemplate(templateName, destPath, newProjectName, yarnV
     newProjectName,
   );
   installTemplateDependencies(templatePath, yarnVersion);
+  installTemplateDevDependencies(templatePath, yarnVersion);
 }
 
 /**
@@ -143,6 +144,7 @@ function createFromRemoteTemplate(template, destPath, newProjectName, yarnVersio
       }
     );
     installTemplateDependencies(templatePath, yarnVersion);
+    installTemplateDevDependencies(templatePath, yarnVersion);
   } finally {
     // Clean up the temp files
     try {
@@ -194,6 +196,38 @@ function installTemplateDependencies(templatePath, yarnVersion) {
   }
   console.log('Linking native dependencies into the project\'s build files...');
   execSync('react-native link', {stdio: 'inherit'});
+}
+
+function installTemplateDevDependencies(templatePath, yarnVersion) {
+  // devDependencies.json is a special file that lists additional develop dependencies
+  // that are required by this template
+  const devDependenciesJsonPath = path.resolve(
+    templatePath, 'devDependencies.json'
+  );
+  console.log('Adding develop dependencies for the project...');
+  if (!fs.existsSync(devDependenciesJsonPath)) {
+    console.log('No additional develop dependencies.');
+    return;
+  }
+
+  let dependencies;
+  try {
+    dependencies = JSON.parse(fs.readFileSync(devDependenciesJsonPath));
+  } catch (err) {
+    throw new Error(
+      'Could not parse the template\'s devDependencies.json: ' + err.message
+    );
+  }
+  for (let depName in dependencies) {
+    const depVersion = dependencies[depName];
+    const depToInstall = depName + '@' + depVersion;
+    console.log('Adding ' + depToInstall + '...');
+    if (yarnVersion) {
+      execSync(`yarn add ${depToInstall} -D`, {stdio: 'inherit'});
+    } else {
+      execSync(`npm install ${depToInstall} --save-dev --save-exact`, {stdio: 'inherit'});
+    }
+  }
 }
 
 module.exports = {


lucianomlima:
@grabbou @hramos I closed the PR and made a new clean PR based on last commit on master.
See my comment about one file for dependencies and devDependencies [here](https://github.com/facebook/react-native/pull/17373#issuecomment-369837662)

lucianomlima:
@grabbou I was testing the possibility of set devDependencies in a single file but this generate a error.
If someone tries to install a template with current RN version and dependencies.json has the following structure:
```json
{
    "react-navigation": "^ 1.3.0",
    "devDependencies": {
        "prettier": "^ 1.9.0"
    }
}
```
He finds an error because the current code goes through all the dependencies found in json file and will try to install a dependency that is referenced as an object.
The error will be: `Error: Command failed: yarn add devDependencies@[object Object]` resulting in installation failure.

Following my proposal with separate files to dependencies and devDependencies, it maintains compatibility with the current version.

grabbou:
Sounds good - having too much of an abstraction is also a bad thing. Do you have a set of steps to do in order to reproduce this feature and see that it works?

lucianomlima:
Here is a repo that can be used to test when start a new project passing a template param
https://github.com/lucianomlima/react-native-template-exemplo

grabbou:
Looks good to me. I am not familiar with the generator code since I haven't authored any part of that before. I think @mkonicek would be the best person to land this (or at least confirm whether this is good to go). From a technical point of view, I think it's fine.

Let's wait for few days, otherwise I'll go ahead and merge it in.

tangkunyin:
Very cool PR but not work until 0.54.4.....

Hope this can work as soon as possiable

lucianomlima:
For devDependencies will only work when have been merged. For dependencies.json it's working with current version since version 0.43.0

lucianomlima:
@grabbou some new from @mkonicek ?

mkonicek:
Hey @lucianomlima, thanks so much for doing this! Looks good.

I think it would be possible to refactor the code a bit so there's some shared code with `installTemplateDependencies` but not a huge deal to copy&paste, it's not a lot of code.

For me it looks good to be merged.

As a side note, I think when designing the templates the first time, I could have made this logic more flexible. The `dependencies.json` file wasn't the best idea probably - as we can see now we need more flexibility we need another file called `devDependencies.json`. It might be simpler to have a file like `_package.json` in the template, which would allow people to specify dependencies, devDependencies, and also scripts and other things. But that's for a separate discussion. I think this PR is good to go as-is and it's an improvement on the current version!

mkonicek:
@facebook-github-bot shipit

mkonicek:
@grabbou Could you merge this please?

lucianomlima:
Thanks @mkonicek. Now, change the file name and structure is possible but with a break change for who uses the current version. Maybe a new PR that allow use same versions with warning message for old ones.

grabbou:
Thank you @mkonicek for doing this review for us. I am landing it now.

grabbou:
@facebook-github-bot shipit

savelichalex:
@mkonicek You are absolutely right! I'm right now in a middle of research how to add few things to package.json in generate phase, for i.e. we have settings for prettier, jest and etc. Can you share your thoughts how to make it better? I'm ready to help with it.

mkonicek:
Hey @savelichalex, thank you so much for offering to help!

I was thinking just providing a normal `package.json` file and replacing "HelloWorld" with the project name where needed in that file, like we do with the rest of the template.

The only issue is the file can't be called `package.json` because we need a `package.json` file for the template itself: https://www.npmjs.com/search?q=react-native-template-

It could be called e.g. `_package.json`.

The change should be backwards-compatible, so that existing templates that use `dependencies.json` and `devDependencies.json` still work.

savelichalex:
@mkonicek Right now I use in our template simple and naive [solution](https://github.com/Sprit3Dan/react-native-supercharged/blob/master/scripts/completeSetup.js). Actually it might work well and even be backward compatible.
Also we might want to not just add another one method of changing package.json, instead add ability to run some script, like `postinstall` in npm.

savelichalex:
It's just patch for `detox`, because detox is monorepo and install it from fork isn't work well :) Actually it's temp decision

savelichalex:
@mkonicek I just thought, that if we can change whole package.json, then we easily can add our own `postinstall` hook, that do the trick, so maybe there is no need to add our own logic for this

padupuy:
@lucianomlima  @grabbou  I just try to setup my react-native project with the testing template 

```sh
react-native init test --template https://github.com/lucianomlima/react-native-template-exemplo
```

The `devDependencies.json` file is not use and still exits in the source code after the template installation. Is this a bug ?

lucianomlima:
@padupuy What RN version your projects uses?

padupuy:
@lucianomlima  RN version is react-native@0.55.4
I also tested with `npx react-native init test --template https://github.com/lucianomlima/react-native-template-exemplo` and the result is the same (devDependencies KO)

PR #18181, opened 2018-03-04
Add ability for Animated views to be created with scale X or scale Y

Liamandrew:
## Motivation

Sometimes it can be useful to have an animated view be created with either scale X or scale Y in cases where scaleXY might not be as visually appealing. 

## Test Plan

Tested on both ios and android in the sample project: 

https://github.com/Liamandrew/ScaleAnimationSample

scaleanimation


package.json
```diff
@@ -173,6 +173,7 @@
     "glob": "^7.1.1",
     "graceful-fs": "^4.1.3",
     "inquirer": "^3.0.6",
+    "jest-cli": "^22.4.2",
```
hramos: Can you explain why this is necessary?

PR #18187, opened 2018-03-04
Support Interpolation of strings when using native driver in Animated, fix Expected node to be marked as "native", optimize AnimatedNode creation and connections

msand:
## Summary

Allow interpolation of strings with useNativeDriver. This allows animating much more of react-native-svg. This PR adds support for native animation of lengths with units, path data, colors etc. Plus, fixing the redundantly created nodes and (and thus, previously incorrect) connection of native animated nodes, improving performance.

## Related PRs

Docs will need to change, specifying that string interpolation works with the native driver as well.

## Changelog:

[GENERAL] [Added] Add support for native driven string interpolation in Animated
[GENERAL] Fix exception: Expected node to be marked as "native"
[GENERAL] Fix connection of AnimatedNodes and creation of redundant AnimatedNodes

## Test Plan

Make a set of tests, verifying that both number and string properties of svg content can be animated with the native driver. Showing that animation continues even if js thread is blocked.

### DIFF
--- a/Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
+++ b/Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
@@ -9,27 +9,89 @@
 
 #import "RCTAnimationUtils.h"
 
+static NSRegularExpression *regex;
+
 @implementation RCTInterpolationAnimatedNode
 {
   __weak RCTValueAnimatedNode *_parentNode;
   NSArray<NSNumber *> *_inputRange;
   NSArray<NSNumber *> *_outputRange;
+  NSArray<NSArray<NSNumber *> *> *_outputs;
+  NSArray<NSString *> *_soutputRange;
   NSString *_extrapolateLeft;
   NSString *_extrapolateRight;
+  NSUInteger _numVals;
+  bool _hasStringOutput;
+  bool _shouldRound;
+  NSArray<NSTextCheckingResult*> *_matches;
 }
 
 - (instancetype)initWithTag:(NSNumber *)tag
                      config:(NSDictionary<NSString *, id> *)config
 {
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    regex = [NSRegularExpression regularExpressionWithPattern:@"[0-9.-]+" options:NSRegularExpressionCaseInsensitive error:nil];
+  });
   if ((self = [super initWithTag:tag config:config])) {
     _inputRange = [config[@"inputRange"] copy];
     NSMutableArray *outputRange = [NSMutableArray array];
+    NSMutableArray *soutputRange = [NSMutableArray array];
+    NSMutableArray<NSMutableArray<NSNumber *> *> *_outputRanges = [NSMutableArray array];
+
+    _hasStringOutput = NO;
     for (id value in config[@"outputRange"]) {
       if ([value isKindOfClass:[NSNumber class]]) {
         [outputRange addObject:value];
+      } else if ([value isKindOfClass:[NSString class]]) {
+        /**
+         * Supports string shapes by extracting numbers so new values can be computed,
+         * and recombines those values into new strings of the same shape.  Supports
+         * things like:
+         *
+         *   rgba(123, 42, 99, 0.36) // colors
+         *   -45deg                  // values with units
+         */
+        NSMutableArray *output = [NSMutableArray array];
+        [_outputRanges addObject:output];
+        [soutputRange addObject:value];
+
+        _matches = [regex matchesInString:value options:0 range:NSMakeRange(0, [value length])];
+        for (NSTextCheckingResult *match in _matches) {
+          NSString* strNumber = [value substringWithRange:match.range];
+          [output addObject:[NSNumber numberWithDouble:strNumber.doubleValue]];
+        }
+
+        _hasStringOutput = YES;
+        [outputRange addObject:[output objectAtIndex:0]];
       }
     }
+    if (_hasStringOutput) {
+      // ['rgba(0, 100, 200, 0)', 'rgba(50, 150, 250, 0.5)']
+      // ->
+      // [
+      //   [0, 50],
+      //   [100, 150],
+      //   [200, 250],
+      //   [0, 0.5],
+      // ]
+      _numVals = [_matches count];
+      NSString *value = [soutputRange objectAtIndex:0];
+      _shouldRound = [value containsString:@"rgb"];
+      _matches = [regex matchesInString:value options:0 range:NSMakeRange(0, [value length])];
+      NSMutableArray<NSMutableArray<NSNumber *> *> *outputs = [NSMutableArray arrayWithCapacity:_numVals];
+      NSUInteger size = [soutputRange count];
+      for (NSUInteger j = 0; j < _numVals; j++) {
+        NSMutableArray *output = [NSMutableArray arrayWithCapacity:size];
+        [outputs addObject:output];
+        for (int i = 0; i < size; i++) {
+          [output addObject:[[_outputRanges objectAtIndex:i] objectAtIndex:j]];
+        }
+      }
+      _outputs = [outputs copy];
+    }
     _outputRange = [outputRange copy];
+    _soutputRange = [soutputRange copy];
     _extrapolateLeft = config[@"extrapolateLeft"];
     _extrapolateRight = config[@"extrapolateRight"];
   }
@@ -61,11 +123,47 @@
 
   CGFloat inputValue = _parentNode.value;
 
-  self.value = RCTInterpolateValueInRange(inputValue,
-                                          _inputRange,
-                                          _outputRange,
-                                          _extrapolateLeft,
-                                          _extrapolateRight);
+  CGFloat interpolated = RCTInterpolateValueInRange(inputValue,
+                                                    _inputRange,
+                                                    _outputRange,
+                                                    _extrapolateLeft,
+                                                    _extrapolateRight);
+  self.value = interpolated;
+  if (_hasStringOutput) {
+    // 'rgba(0, 100, 200, 0)'
+    // ->
+    // 'rgba(${interpolations[0](input)}, ${interpolations[1](input)}, ...'
+    if (_numVals > 1) {
+      NSString *text = _soutputRange[0];
+      NSMutableString *formattedText = [NSMutableString stringWithString:text];
+      NSUInteger i = _numVals;
+      for (NSTextCheckingResult *match in [_matches reverseObjectEnumerator]) {
+        CGFloat val = RCTInterpolateValueInRange(inputValue,
+                                                          _inputRange,
+                                                          _outputs[--i],
+                                                          _extrapolateLeft,
+                                                          _extrapolateRight);
+        NSString *str;
+        if (_shouldRound) {
+          // rgba requires that the r,g,b are integers.... so we want to round them, but we *dont* want to
+          // round the opacity (4th column).
+          bool isAlpha = i == 3;
+          CGFloat rounded = isAlpha ? round(val * 1000) / 1000 : round(val);
+          str = isAlpha ? [NSString stringWithFormat:@"%1.3f", rounded] : [NSString stringWithFormat:@"%1.0f", rounded];
+        } else {
+          str = [NSString stringWithFormat:@"%1f", val];
+        }
+
+        [formattedText replaceCharactersInRange:[match range] withString:str];
+      }
+      self.animatedObject = formattedText;
+    } else {
+      self.animatedObject = [regex stringByReplacingMatchesInString:_soutputRange[0]
+                                                 options:0
+                                                   range:NSMakeRange(0, _soutputRange[0].length)
+                                            withTemplate:[NSString stringWithFormat:@"%1f", interpolated]];
+    }
+  }
 }
 
 @end
--- a/Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
+++ b/Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
@@ -24,6 +24,7 @@
 - (void)extractOffset;
 
 @property (nonatomic, assign) CGFloat value;
+@property (nonatomic, strong) id animatedObject;
 @property (nonatomic, weak) id<RCTValueAnimatedNodeObserver> valueObserver;
 
 @end
--- a/ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
@@ -9,6 +9,12 @@ package com.facebook.react.animated;
 import com.facebook.react.bridge.JSApplicationIllegalArgumentException;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.ReadableType;
+
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import javax.annotation.Nullable;
 
 /**
@@ -21,6 +27,7 @@ import javax.annotation.Nullable;
   public static final String EXTRAPOLATE_TYPE_IDENTITY = "identity";
   public static final String EXTRAPOLATE_TYPE_CLAMP = "clamp";
   public static final String EXTRAPOLATE_TYPE_EXTEND = "extend";
+  static final Pattern regex = Pattern.compile("[0-9.-]+");
 
   private static double[] fromDoubleArray(ReadableArray ary) {
     double[] res = new double[ary.size()];
@@ -105,13 +112,68 @@ import javax.annotation.Nullable;
 
   private final double mInputRange[];
   private final double mOutputRange[];
+  private String mPattern;
+  private double mOutputs[][];
+  private final boolean mHasStringOutput;
+  private final Matcher mSOutputMatcher;
   private final String mExtrapolateLeft;
   private final String mExtrapolateRight;
   private @Nullable ValueAnimatedNode mParent;
+  private boolean mShouldRound;
+  private int mNumVals;
 
   public InterpolationAnimatedNode(ReadableMap config) {
     mInputRange = fromDoubleArray(config.getArray("inputRange"));
-    mOutputRange = fromDoubleArray(config.getArray("outputRange"));
+    ReadableArray output = config.getArray("outputRange");
+    mHasStringOutput = output.getType(0) == ReadableType.String;
+    if (mHasStringOutput) {
+      /*
+       * Supports string shapes by extracting numbers so new values can be computed,
+       * and recombines those values into new strings of the same shape.  Supports
+       * things like:
+       *
+       *   rgba(123, 42, 99, 0.36) // colors
+       *   -45deg                  // values with units
+       */
+      int size = output.size();
+      mOutputRange = new double[size];
+      mPattern = output.getString(0);
+      mShouldRound = mPattern.startsWith("rgb");
+      mSOutputMatcher = regex.matcher(mPattern);
+      ArrayList<ArrayList<Double>> mOutputRanges = new ArrayList<>();
+      for (int i = 0; i < size; i++) {
+        String val = output.getString(i);
+        Matcher m = regex.matcher(val);
+        ArrayList<Double> outputRange = new ArrayList<>();
+        mOutputRanges.add(outputRange);
+        while (m.find()) {
+          Double parsed = Double.parseDouble(m.group());
+          outputRange.add(parsed);
+        }
+        mOutputRange[i] = outputRange.get(0);
+      }
+
+      // ['rgba(0, 100, 200, 0)', 'rgba(50, 150, 250, 0.5)']
+      // ->
+      // [
+      //   [0, 50],
+      //   [100, 150],
+      //   [200, 250],
+      //   [0, 0.5],
+      // ]
+      mNumVals = mOutputRanges.get(0).size();
+      mOutputs = new double[mNumVals][];
+      for (int j = 0; j < mNumVals; j++) {
+        double[] arr = new double[size];
+        mOutputs[j] = arr;
+        for (int i = 0; i < size; i++) {
+          arr[i] = mOutputRanges.get(i).get(j);
+        }
+      }
+    } else {
+      mOutputRange = fromDoubleArray(output);
+      mSOutputMatcher = null;
+    }
     mExtrapolateLeft = config.getString("extrapolateLeft");
     mExtrapolateRight = config.getString("extrapolateRight");
   }
@@ -142,6 +204,33 @@ import javax.annotation.Nullable;
       // unattached node.
       return;
     }
-    mValue = interpolate(mParent.getValue(), mInputRange, mOutputRange, mExtrapolateLeft, mExtrapolateRight);
+    double value = mParent.getValue();
+    mValue = interpolate(value, mInputRange, mOutputRange, mExtrapolateLeft, mExtrapolateRight);
+    if (mHasStringOutput) {
+      // 'rgba(0, 100, 200, 0)'
+      // ->
+      // 'rgba(${interpolations[0](input)}, ${interpolations[1](input)}, ...'
+      if (mNumVals > 1) {
+        StringBuffer sb = new StringBuffer(mPattern.length());
+        int i = 0;
+        mSOutputMatcher.reset();
+        while (mSOutputMatcher.find()) {
+          double val = interpolate(value, mInputRange, mOutputs[i++], mExtrapolateLeft, mExtrapolateRight);
+          if (mShouldRound) {
+            // rgba requires that the r,g,b are integers.... so we want to round them, but we *dont* want to
+            // round the opacity (4th column).
+            boolean isAlpha = i == 4;
+            int rounded = (int)Math.round(isAlpha ? val * 1000 : val);
+            mSOutputMatcher.appendReplacement(sb, isAlpha ? String.valueOf((double)rounded / 1000) : String.valueOf(rounded));
+          } else {
+            mSOutputMatcher.appendReplacement(sb, String.valueOf(val));
+          }
+        }
+        mSOutputMatcher.appendTail(sb);
+        mAnimatedObject = sb.toString();
+      } else {
+        mAnimatedObject = mSOutputMatcher.replaceFirst(String.valueOf(mValue));
+      }
+    }
   }
 }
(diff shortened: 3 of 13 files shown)


msand:
@hramos @grabbou Is it likely to get this merged? You know anyone I can ask / who can review this? Or should I change strategy?

hramos:
cc @janicduplessis, @skevy

grabbou:
I believe people quoted by @hramos are the best to review this change. Also @kmagiera

janicduplessis:
Thanks for working on this! This only support interpolating a single number inside a string right? So something like "10deg" would work but not colors like "rgb(255, 255, 255)"? Also do we handle this in a similar way in the JS version of animated? It's usually a good place to get started to figure out where to go implementation wise.

I'm not a big fan of adding a stringValue prop to the AnimatedValue node but I can't really think of a better way that doesn't involve a lot more code changes. Ideally the value would be id / Object and could be accessed with asDouble or asString methods a bit like JavaOnlyMap. The problem is currently a lot of code just accesses the prop directly and assumes it's a double.

@kmagiera Any thoughts on this?

msand:
@janicduplessis Yes, that's correct. I can add support for arbitrary number of occurrences of numbers in a string, this would be a bit more generic than just rgb, and would allow animating e.g. path data as well.

At least one difference to the js version, is that the native one could at least in theory have one interpolation output both a string and a number, such that the string can be used as an attribute, and the number can be used for further computation/interpolation.

Yeah, asDouble/String definitely seems nice, but probably a bigger change indeed.

zachgibson:
@hramos Any idea of the status of this PR?

msand:
@janicduplessis @hramos @skevy @kmagiera @zachgibson @shergin I've added support for native animation of rgb/rgba now as well, what do you think of merging this?

msand:
I'm not completely satisfied with my current attempt at supporting native animation of svg. It requires adding a native view for each node, just to be able to set values on the shadow nodes. Does anyone have a better idea for how to achieve this? Or is this unavoidable with the current architecture? @sebmarkbage @sophiebits @rafeca @ss18 
https://github.com/react-native-community/react-native-svg/pull/757/files#diff-50f7e7f8735df31a6ac08ae41511a2a6R34

msand:
Here's some simple code I've used to exercise the native svg animation logic and demonstrating string animations of path data, rgba colors, and values with units:
```jsx
import { StyleSheet, View, Dimensions, Animated } from 'react-native';
import { Svg, Rect, Path } from 'react-native-svg';

const { width, height } = Dimensions.get('window');
const AnimatedRect = Animated.createAnimatedComponent(Rect);
const AnimatedPath = Animated.createAnimatedComponent(Path);

function getInitialState() {
  const anim = new Animated.Value(0);
  const fillOpacity = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 1],
  });
  const offset = fillOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 10],
  });
  const strokeOpacity = offset.interpolate({
    inputRange: [0, 10],
    outputRange: [0, 1],
    extrapolateRight: 'clamp',
  });
  const path = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['M20,20L20,80L80,80L80,20Z', 'M40,40L33,60L60,60L65,40Z'],
  });
  const fill = strokeOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: ['rgba(255, 0, 0, 0.5)', 'rgba(0, 255, 0, 0.99)'],
  });
  const oneToFivePx = offset.interpolate({
    inputRange: [0, 10],
    outputRange: ['1px', '5px'],
  });
  return { anim, fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx };
}

  state = getInitialState();

  componentDidMount() {
    const { anim } = this.state;
    Animated.timing(anim, {
      toValue: 1,
      duration: 10000,
      useNativeDriver: true,
    }).start();
  }

  render() {
    const { fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx } = this.state;
    return (
        <Svg width={width} height={height} viewBox="0 0 100 100">
          <AnimatedRect
            x="5"
            y="5"
            width="90"
            height="90"
            stroke="blue"
            fill={fill}
            strokeDasharray="1 1"
            strokeWidth={oneToFivePx}
            strokeDashoffset={offset}
            strokeOpacity={strokeOpacity}
            fillOpacity={fillOpacity}
          />
          <AnimatedPath
            d={path}
            stroke="blue"
          />
        </Svg>
      </View>
    );
  }
}

  container: {
    backgroundColor: '#ecf0f1',
  },
});
```
I'll try to rebase onto the latest master branch as well, as there's some quite useful fixes for the native animation merged recently. Should I push the rebased version or what is preferred here?

msand:
Now with an expensive js computation to introduce some lag if useNativeDriver is false:
```jsx
import { StyleSheet, View, Dimensions, Animated } from 'react-native';
import { Svg, Rect, Path } from 'react-native-svg';

const { width, height } = Dimensions.get('window');
const AnimatedRect = Animated.createAnimatedComponent(Rect);
const AnimatedPath = Animated.createAnimatedComponent(Path);

function getInitialState() {
  const anim = new Animated.Value(0);
  const fillOpacity = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 1],
  });
  const offset = fillOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 10],
  });
  const strokeOpacity = offset.interpolate({
    inputRange: [0, 10],
    outputRange: [0, 1],
    extrapolateRight: 'clamp',
  });
  const path = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['M20,20L20,80L80,80L80,20Z', 'M40,40L33,60L60,60L65,40Z'],
  });
  const fill = strokeOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: ['rgba(255, 0, 0, 0.5)', 'rgba(0, 255, 0, 0.99)'],
  });
  const oneToFivePx = offset.interpolate({
    inputRange: [0, 10],
    outputRange: ['1px', '5px'],
  });
  return { anim, fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx };
}

  state = getInitialState();

  expensiveComputation = () => {
    let result = 1;
    for (let i = 0; i < 100000000; i++) {
      result += i;
    }
    console.log(result);
    setTimeout(this.expensiveComputation, 200)
  };

  componentDidMount() {
    const { anim } = this.state;
    Animated.timing(anim, {
      toValue: 1,
      duration: 10000,
      useNativeDriver: true,
    }).start();
    setTimeout(this.expensiveComputation, 1000)
  }

  render() {
    const { fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx } = this.state;
    return (
        <Svg width={width} height={height} viewBox="0 0 100 100">
          <AnimatedRect
            x="5"
            y="5"
            width="90"
            height="90"
            stroke="blue"
            fill={fill}
            strokeDasharray="1 1"
            strokeWidth={oneToFivePx}
            strokeDashoffset={offset}
            strokeOpacity={strokeOpacity}
            fillOpacity={fillOpacity}
          />
          <AnimatedPath
            d={path}
            stroke="blue"
          />
        </Svg>
      </View>
    );
  }
}

  container: {
    backgroundColor: '#ecf0f1',
  },
});
```

msand:
Seems to work fine both when rebased onto v0.56 and onto v0.57.0-rc.0

msand:
Animation without js workload (performance when not recording is 50-60fps, about 2x compared to recording)
nativesvganimationwithoutwork

Animation with js workload
nativesvganimationwithwork

Code:
```jsx
import {
  StyleSheet,
  View,
  Dimensions,
  Animated,
  Button,
} from 'react-native';
import { Svg, Rect, Path } from 'react-native-svg';

const { width, height } = Dimensions.get('window');
const AnimatedRect = Animated.createAnimatedComponent(Rect);
const AnimatedPath = Animated.createAnimatedComponent(Path);

function getInitialState() {
  const anim = new Animated.Value(0);
  const fillOpacity = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 1],
  });
  const offset = fillOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 10],
  });
  const strokeOpacity = offset.interpolate({
    inputRange: [0, 10],
    outputRange: [0, 1],
    extrapolateRight: 'clamp',
  });
  const path = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['M20,20L20,80L80,80L80,20Z', 'M40,40L33,60L60,60L65,40Z'],
  });
  const fill = strokeOpacity.interpolate({
    inputRange: [0, 1],
    outputRange: ['rgba(255, 0, 0, 0.5)', 'rgba(0, 255, 0, 0.99)'],
  });
  const oneToFivePx = offset.interpolate({
    inputRange: [0, 10],
    outputRange: ['1px', '5px'],
  });
  return { anim, fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx };
}

  state = getInitialState();

  expensiveComputation = () => {
    let result = 1;
    for (let i = 0; i < 1000000000; i++) {
      result += i;
    }
    console.log(result);
    this.timeout = setTimeout(this.expensiveComputation, 200);
  };

  componentDidMount() {
    const { anim } = this.state;
    Animated.timing(anim, {
      toValue: 1,
      duration: 10000,
      useNativeDriver: true,
    }).start();
  }
  toZero = () => {
    console.log('0');
    Animated.timing(this.state.anim, {
      toValue: 0,
      duration: 10000,
      useNativeDriver: true,
    }).start();
  };
  toOne = () => {
    console.log('1');
    Animated.timing(this.state.anim, {
      toValue: 1,
      duration: 10000,
      useNativeDriver: true,
    }).start();
  };
  toggleWork = () => {
    const { timeout } = this;
    if (timeout) {
      clearTimeout(timeout);
      this.timeout = null;
    } else {
      this.timeout = setTimeout(this.expensiveComputation, 1000);
    }
  };

  render() {
    const {
      fillOpacity,
      offset,
      strokeOpacity,
      path,
      fill,
      oneToFivePx,
    } = this.state;
    return (
        <View style={{ flexDirection: 'row', marginTop: 30 }}>
          <Button onPress={this.toZero} title="To 0" />
          <Button onPress={this.toOne} title="To 1" />
          <Button onPress={this.toggleWork} title="Toggle work" />
        </View>
        <Svg width={width} height={height} viewBox="0 0 100 100">
          <AnimatedRect
            x="5"
            y="5"
            width="90"
            height="90"
            stroke="blue"
            fill={fill}
            strokeDasharray="1 1"
            strokeWidth={oneToFivePx}
            strokeDashoffset={offset}
            strokeOpacity={strokeOpacity}
            fillOpacity={fillOpacity}
          />
          <AnimatedPath d={path} stroke="blue" />
        </Svg>
      </View>
    );
  }
}

  container: {
    backgroundColor: '#ecf0f1',
  },
});

```

msand:
@janicduplessis Could you have another look at this? Thanks!

msand:
Managed to improve the performance in react-native-svg on android a bit. It was rendering a bitmap per animated node per animation frame, rather than once per svg root per frame. Should scale significantly better now.

msand:
@kmagiera What do you think about this? I've released the changes for native animation in react-native-svg v7.0.0 now. Would be great to get access to this more broadly.

todorone:
I would be amazing if we can get SVG animations on the native thread with the help of this PR. Will give us so much freedom in vector shapes high-performance animations w/o need to use 3rd-party tools like Lottie...

IljaDaderko:
Would be great to have this merged in order to enable whats possible in react-native-svg

Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
```diff
@@ -24,6 +24,7 @@
 - (void)extractOffset;
 
 @property (nonatomic, assign) CGFloat value;
```
janicduplessis: I think we should change the type of value to be `id` on iOS or `Object` on Android. Then we can use instanceof checks in when we need to use the actual value.

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
```diff
@@ -9,27 +9,72 @@
...
+  NSUInteger _numVals;
+  bool _hasStringOutput;
+  bool _shouldRound;
+  NSArray<NSTextCheckingResult*> *_matches;
 }
 
 - (instancetype)initWithTag:(NSNumber *)tag
                      config:(NSDictionary<NSString *, id> *)config
 {
+  if (!regex) {
```
janicduplessis: Could you explain the changes and add a comment here that gives the overall flow of the following code. I think it would help understand what is going on.

Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
msand: Is Object preferable to Dynamic here? Both require boxing of primitives, but Object is probably cheaper anyway. Just not completely sure how this would interact with other native libraries for react-native, or how many of them depend on this being a primitive data type. Also the cost of boxing can be quite significant and can even cause heap allocations. A float would take 4 bytes normally, but a Float can be 16(-ish) bytes, plus the cost of the pointer (device/cpu dependent). I'm afraid this could cause a noticeable performance regression in some cases.

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
msand: I essentially just tried to translate the logic from the js implementation straight. I'll take a look at documenting this once I have some time to spare. Thanks for the review!

msand:
@tobycox Btw, react-native-svg now supports using plain numbers with useNativeDriver, so any properties which can take unit-less numbers work fine with native animation at this point, without needing this PR. I've even tested down to v0.50.4 of react-native (latest react-native-svg related breaking change was v0.50) and it worked fine. But if you want to animate e.g. path-data or other strings containing numbers you'll need to use this fork, patch yourself or wait for this to get merged.

Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
msand: I just realised that the JavaOnlyMap will cause autoboxing of the double into a Double already anyway (and the android implementation already uses doubles rather than floats). So using Object should be fine.

osdnk:
Hi, @msand / @janicduplessis, @grabbou 

I've just case across this PR and found it really interesting. Looks fine for me, but appears not to be a full answer for strings support problem in animation.

I suggest considering adding some special kind of extra animated node, e.g. `concat` node which could be used for concatenation of numerical value and some string. Then API could look like:
```javascript
const v = Animated.concat(val, 'rad');
```

In next step it could be used in implementation of interpolation.

I find it simpler and more generic solution.

If you seek for inspiration, I've done it in `react-native-reanimated` (https://github.com/kmagiera/react-native-reanimated) so you might get inspired. If I will find some time and you're fine with this approach, I could try even do it on my own and add a PR here.

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
```diff
@@ -9,27 +9,72 @@
 
 #import "RCTAnimationUtils.h"
 
+NSRegularExpression *regex;
```
osdnk: If it actually has to be a global var (I think it could local as well -- it's just a regex), can't we initialize it immediately? 
Also, the name is not very clear.

ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
```diff
@@ -105,13 +112,51 @@ private static int findRangeIndex(double value, double[] ranges) {
...
   private final double mInputRange[];
   private final double mOutputRange[];
+  private String mPattern;
+  private double mOutputs[][];
+  private final boolean mHasStringOutput;
+  private final Matcher mSOutputMatcher;
   private final String mExtrapolateLeft;
   private final String mExtrapolateRight;
   private @Nullable ValueAnimatedNode mParent;
+  private boolean shouldRound;
```
osdnk: `mShouldRound`?

ReactAndroid/src/main/java/com/facebook/react/animated/ValueAnimatedNode.java
```diff
@@ -16,6 +16,7 @@
  * library.
  */
 /*package*/ class ValueAnimatedNode extends AnimatedNode {
+  /*package*/ String mStringValue = null;
```
osdnk: Why it can't be `protected`?

ReactAndroid/src/main/java/com/facebook/react/animated/PropsAnimatedNode.java
```diff
@@ -89,7 +89,12 @@ public final void updateView() {
       } else if (node instanceof StyleAnimatedNode) {
         ((StyleAnimatedNode) node).collectViewUpdates(mPropMap);
       } else if (node instanceof ValueAnimatedNode) {
-        mPropMap.putDouble(entry.getKey(), ((ValueAnimatedNode) node).getValue());
+        String stringValue = ((ValueAnimatedNode) node).getStringValue();
+        if (stringValue != null) {
```
osdnk: It's not very generic. I'd rather allow value to be an `Object` and check whether it's an instance of `Double` or `String`.

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
msand: True, was mostly thinking it can be reused across instances, so made it shared among them.

ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
msand: Ah, good catch

ReactAndroid/src/main/java/com/facebook/react/animated/PropsAnimatedNode.java
msand: True, but that requires wider changes in react-native, as the rest of the code depends on getValue returning a double.

ReactAndroid/src/main/java/com/facebook/react/animated/ValueAnimatedNode.java
msand: Should work as well yeah

msand:
@osdnk I'm not currently working on this, nor have it scheduled for the near future. Please feel free to work on improving string animation support either based on this or the reanimated style. Either one would be great to have 👍

cpojer:
Given that @msand is not currently working on this and nobody has volunteered to take it over, I'm going to close this PR. Please open a new PR with a rebased and complete version of this if you'd like us to ship it in RN soon.

msand:
I've rebased this against the 0.59-stable branch, in case anyone is interested: https://github.com/facebook/react-native/commit/f462f97598b38d2071e569a37fc2c24acb7a744f

ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
msand: Resolved in https://github.com/msand/react-native/commit/1e984241ef926c83cdf590a6384c72ac58afc37d

ReactAndroid/src/main/java/com/facebook/react/animated/ValueAnimatedNode.java
msand: @osdnk I think it's probably better to keep this restricted to package-private, or do you think AnimatedNode is likely to be extended by other packages?

ReactAndroid/src/main/java/com/facebook/react/animated/PropsAnimatedNode.java
msand: @osdnk  You can try changing value to Object and see that quite a lot of the Animated logic would need to change to even make it compile. Or do you think we should keep value as a double, and add objValue of type Object, which can be something else?

Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
msand: @janicduplessis How about keeping value as a double, and adding another property called objValue? This can be used for general purposes such as String interpolation. Changing all the uses of value to have dynamic type will probably come with a noticeable performance impact for the current Animated use cases. Having it isolated to a single if statement, like this pr has it, but changing it to Object and using instanceof to handle the String case, seems like the best tradeoff for me atm. What do you think?

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
msand: @osdnk I think it's not possible to initialize global vars like this in obj-c, or do you know some way?

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
msand: @janicduplessis Added comments here: https://github.com/msand/react-native/commit/a69e8817561f098df4a9f4f2593fb1c43b19894c
Would this suffice?

msand:
@cpojer Could we reopen this?

cpojer:
Oh, thanks for getting back to us. I reopened the PR, can you update it?

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
msand: Improved it to use dispatch_once

msand:
@cpojer Rebased it against master as a single squashed commit.

msand:
@cpojer I added some relatively significant fixes
Fix exception: Expected node to be marked as "native"
Fix connection of AnimatedNodes and creation of redundant AnimatedNodes

Libraries/NativeAnimation/Nodes/RCTValueAnimatedNode.h
msand: Renamed stringValue to animatedObject, and changed String to id in obj-c and Object in java https://github.com/facebook/react-native/pull/18187/commits/0521b3e085864d632cb257310c4b34195fab92f3

ReactAndroid/src/main/java/com/facebook/react/animated/PropsAnimatedNode.java
msand: Renamed stringValue to animatedObject, and changed String to id in obj-c and Object in java https://github.com/facebook/react-native/pull/18187/commits/0521b3e085864d632cb257310c4b34195fab92f3

msand:
@cpojer Fixed all the review feedback. I think this might be ready for merging.

msand:
@janicduplessis Could you check the changes I've made in response to your review? And give any feedback needed before merging?

msand:
Cleaned up a bit and added comments to the android implementation as well. Now I don't know about anything to improve in this anymore.

msand:
@hramos @grabbou could this be merged and included in 0.59?

cpojer:
I'll wait for another review by @janicduplessis.

We aren't picking features or non-essential fixes to release branches, so this will go into 0.60 the earliest.

msand:
@cpojer Alright, good to know. But, this pr essentially just aligns the native animation with what the plain Animated api already does / is able to, when useNativeDriver is falsy. In a sense this could probably be considered a bugfix. But I understand if it counts as a feature and the window has closed.

msand:
@janicduplessis Could you have another look at this? Or someone else on the team? @cpojer
The code should be relatively straightforward to follow with the comment. If there's any questions then I'm available to explain/comment further.

todorone:
@janicduplessis @shergin @hramos @cpojer
Can we somehow push reviewing/merging this commit(it's more then 1 year old now 😸)?
Handling svg shapes animations on native side will open door to great performant vector animations which will be substantial step up for React Native. Currently only option of implementing such animations on RN is Lottie but it doesn't work well on Android and requires Adobe AfterEffects.

msand:
@cpojer Would it be possible to get this into v0.61? useNativeDriver is currently quite broken, or at least very hard to make efficient, as you cannot use a single Animated.Value to drive multiple interpolations to set several properties on one View. For now, it requires finding very specific way to chain the interpolations together, for it not to crash and actually animate instead. That also makes the interpolations unnatural, as they need to use the range of values from the interpolation they're using as their input, rather than the simple Animated.Value which goes e.g. from 0 to 1.
To reproduce the exceptions, you can use the example I have used to test this pr, but set all the interpolations to use the Animated.Value, rather than have the very specific order in which they're chained now. So changing the getInitialState to something like this:

```jsx
function getInitialState() {
  const anim = new Animated.Value(0);
  const fillOpacity = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 1],
  });
  const offset = anim.interpolate({
    inputRange: [0, 1],
    outputRange: [0, 10],
  });
  const strokeOpacity = anim.interpolate({
    inputRange: [0, 0.8],
    outputRange: [0, 1],
    extrapolateRight: 'clamp',
  });
  const path = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['M20,20L20,80L80,80L80,20Z', 'M40,40L33,60L60,60L65,40Z'],
  });
  const fill = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['rgba(255, 0, 0, 0.5)', 'rgba(0, 255, 0, 0.99)'],
  });
  const oneToFivePx = anim.interpolate({
    inputRange: [0, 1],
    outputRange: ['1px', '5px'],
  });
  return { anim, fillOpacity, offset, strokeOpacity, path, fill, oneToFivePx };
}
```

cpojer:
Unfortunately I have to unland this PR because it breaks external tests. Even worse, I cannot share that code externally but from what it looks like we have a some Android stuff that uses Animated and the element our test is looking for doesn't seem to show up any more with this diff applied :( Is there anything obvious that you could point to that may be the problem that you can fix and re-send the PR?

msand:
Oh, I see. Well, we can probably fix it. Any chance of getting some minimal reproduction? My first main suspect would be the change of types crossing the bridge. It used to be Number in all cases, now it can also be a String. Try chainging this one line:
https://github.com/facebook/react-native/pull/18187/files#diff-4d4bfac95b96bfe2f42b7eddd79418e1L275
It'll probably make it work, and would imply that some other part needs to check the type or adapt to the change some way.

cpojer:
It seems like it broke more than just one thing and unfortunately it would take me a while to get a repro :( Your guess seems good. Is there any way you can repro elements not appearing with the existing PR and then making changes to fix it? In general, it's best if you can make sure we do not change the API between JS and native at the moment, and best not to change any JS only API unless there are flow types covering things that will make it easy to detect issues.

msand:
@cpojer Is the issue on iOS, Android or both? Could it be related to this: https://github.com/facebook/react-native/pull/18872

cpojer:
From what I could see, it happened on both platforms. Not sure if it's related to the other PR because that one landed in January.

msand:
@cpojer Could you give some indication what types of components/properties/transforms are used in the failing test?

cpojer:
Let's take this to Discord.

PR #18188, opened 2018-03-05
Use react.gradle from repo root instead of copy in RNTester

janicduplessis:
## Motivation

RNTester used a copy of the main react.gradle file instead of the real one. Recent changes to the real file caused builds to fail with older gradle versions but was not caught by CI because it wasn't using that file for RNTester. That copy of react.gradle is just a leftover from when projects included a copy instead of importing the one in RN directly.

Note: CI WILL fail with this PR, if we have trouble landing this I can add the revert in this commit too but wanted to keep it as 2 separate commits.

## Test Plan

Tested that building RNTester actually fails now that it uses react.gradle with recent changes, then tested that is builds properly when reverting d16ff3bd8b92fa84a9007bf5ebedd8153e4c089d.

## Release Notes

[INTERNAL] [MINOR] [RNTester] - Use react.gradle from repo root instead of copy in RNTester


janicduplessis:
CI is failing properly now :D

PR #18189, opened 2018-03-05
Revert "Better Android Gradle Plugin 3.x integration"

janicduplessis:
This reverts commit d16ff3bd8b92fa84a9007bf5ebedd8153e4c089d.

## Motivation

Currently breaks with the gradle version used by RN, I think there has been some work to update that to a more recent one but for now I think we should just revert it.

It errors with: 

```
Could not find method registerGeneratedResFolders() for arguments [file collection] on object of type com.android.build.gradle.internal.api.ApplicationVariantImpl.
```

## Test Plan

Tested that RN tester now builds when using the right react.gradle (#18188)

## Release Notes

[ ANDROID ] [ BUGFIX ] [ react.gradle ] - REVERT "Support Android Gradle Plugin 3.x and AAPT2"
[ ANDROID ] [ FEATURE ] [ react.gradle ] - REVERT "Expose the bundling task and its outputs via ext properties"

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -6,117 +6,112 @@ def cliPath = config.cliPath ?: "node_modules/react-native/local-cli/cli.js"
 def bundleAssetName = config.bundleAssetName ?: "index.android.bundle"
 def entryFile = config.entryFile ?: "index.android.js"
 def bundleCommand = config.bundleCommand ?: "bundle"
-def reactRoot = file(config.root ?: "../../")
+
+// because elvis operator
+def elvisFile(thing) {
+    return thing ? file(thing) : null;
+}
+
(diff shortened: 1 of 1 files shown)


janicduplessis:
@CFKevinRef If you can find a way that also works with gradle 2.14.1 it would be great. Otherwise I think we'll have to hold on a bit before relanding this.

CFKevinRef:
Not a problem, I do have an idea but will be a few days before I can get back to this. Thanks for the heads up!

janicduplessis:
@CFKevinRef Thanks! If this is a blocker for anyone it is possible to copy the content of the react.gradle file inside your project and change the plugin apply path to point to that instead of the one under node_modules/react-native.

PR #18206, opened 2018-03-05
Don't wrap ListEmptyComponent in an extra view

janicduplessis:
## Motivation

A common UI pattern for list empty states is some text/images centered inside the visible part of the list. This is pretty hard to do currently because we wrap ListEmptyComponent with an extra view with no way to style it so we cannot just use `flex: 1` to make it fill the available space.

## Test Plan

- Added an example of ListEmptyComponent in the FlatList example in RNTester

Before (no way to make ListEmptyComponent fill the space):
screen shot 2018-03-05 at 5 24 15 pm

After:
screen shot 2018-03-05 at 5 09 20 pm

- Tested some edge cases like returning null from the ListEmptyComponent

- Tested in an app that uses FlatList + ListEmptyComponent

## Release Notes

[GENERAL] [MINOR] [VirtualizedList] - Don't wrap ListEmptyComponent in an extra view


Libraries/Lists/VirtualizedList.js
```diff
@@ -844,19 +844,25 @@ class VirtualizedList extends React.PureComponent<Props, State> {
         );
       }
     } else if (ListEmptyComponent) {
-      const element = React.isValidElement(ListEmptyComponent) ? (
+      const element: React.Element<any> = (React.isValidElement(
```
janicduplessis: I don't think flow understands this so the cast seems fine

RNTester/js/FlatListExample.js
```diff
@@ -109,8 +111,9 @@ class FlatListExample extends React.PureComponent<{}, $FlowFixMeState> {
               {renderSmallSwitchOption(this, 'virtualized')}
               {renderSmallSwitchOption(this, 'horizontal')}
               {renderSmallSwitchOption(this, 'fixedHeight')}
-              {renderSmallSwitchOption(this, 'logViewable')}
+              {renderSmallSwitchOption(this, 'log')}
```
janicduplessis: Renamed so controls fit on 2 lines :/

sullenor:
Is there any information in what version it will be available?

hramos:
The release process is explained at http://facebook.github.io/react-native/versions.html


db061ea8c7b78d7e9df4a450c9e7a24d9b2382b4 just landed 8 days ago, so I would expect it to make it into 0.56 when the release candidate is cut next month.

sullenor:
@hramos thank you!

PR #18220, opened 2018-03-06
Add ability for Animated views to be created with scale X or scale Y

Liamandrew:
*Accidentally closed previous PR*

## Motivation

Sometimes it can be useful to have an animated view be created with either scale X or scale Y in cases where scaleXY might not be as visually appealing.

Test Plan
Tested on both ios and android in the sample project:

https://github.com/Liamandrew/ScaleAnimationSample

scaleanimation


janicduplessis:
@hramos Should we add yarn.lock / package-lock.json to .gitignore since it won't be used in OSS for now?

hramos:
@janicduplessis we should, for now. Ideally, we could allow them to be checked in, as long as they are removed prior to import by our tooling (this is because we use our own `yarn.lock` file from a parent directory in the monorepo, and having a second `yarn.lock` here would be problematic for us at FB).

hramos:
https://github.com/facebook/react-native/pull/18281

janicduplessis:
@facebook-github-bot shipit

PR #18281, opened 2018-03-08
Ignore yarn.lock and package-lock.json in oss

hramos:
Attn @janicduplessis


janicduplessis:
@facebook-github-bot shipit

PR #18299, opened 2018-03-09
Added nestedScrollEnabled prop to scroll view for android

tuncaulubilge:
## Motivation

Nested scrolling in scrollViews, listViews and flatLists are enabled on iOS by default, but needs to be enabled manually on Android. This PR introduces a `nestedScrollEnabled` property to ScrollViews to support nested scrolling on Android 21 and above.

Enabling nested scroll will resolve issues with coordinator layout in android and required to support a collapsing toolbar. 

## Test Plan

Tested on the test app. We are also using this property in our app currently to support scrolling behaviour required by coordinator layouts.

## Release Notes

[ANDROID] [ENHANCEMENT] [ScrollView] - Added a prop to enable nested scrolling

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -277,7 +277,7 @@ const ScrollView = createReactClass({
     minimumZoomScale: PropTypes.number,
      /**
      * Enables nested scrolling for Android API level 21+.
-     * Nested scrolling is supported by default in iOS platforms
+     * Nested scrolling is supported by default on iOS
      * @platform android
      */
     nestedScrollEnabled: PropTypes.bool,
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -275,6 +275,12 @@ const ScrollView = createReactClass({
      * @platform ios
      */
     minimumZoomScale: PropTypes.number,
+     /**
+     * Enables nested scrolling for Android API level 21+.
+     * Nested scrolling is supported by default in iOS platforms
+     * @platform android
+     */
+    nestedScrollEnabled: PropTypes.bool,
     /**
      * Called when the momentum scroll starts (scroll which occurs as the ScrollView glides to a stop).
      */
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
@@ -124,7 +124,7 @@ public class ReactHorizontalScrollViewManager
   }
 
   @ReactProp(name = "nestedScrollEnabled")
-  public void setNestedScrollEnabled(ReactScrollView view, boolean value) {
+  public void setNestedScrollEnabled(ReactHorizontalScrollView view, boolean value) {
     ViewCompat.setNestedScrollingEnabled(view, value);
   }
 
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
@@ -123,9 +123,6 @@ public class ReactHorizontalScrollViewManager
     view.setOverScrollMode(ReactScrollViewHelper.parseOverScrollMode(value));
   }
 
-  /**
-   * Enables/Disables nested scrolling
-   */
   @ReactProp(name = "nestedScrollEnabled")
   public void setNestedScrollEnabled(ReactScrollView view, boolean value) {
     ViewCompat.setNestedScrollingEnabled(view, value);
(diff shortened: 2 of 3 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
```diff
@@ -120,6 +123,14 @@ public void setOverScrollMode(ReactHorizontalScrollView view, String value) {
     view.setOverScrollMode(ReactScrollViewHelper.parseOverScrollMode(value));
   }
 
+  /**
+   * Enables/Disables nested scrolling
```
janicduplessis: No need for that comment since it is pretty self explanatory.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -275,6 +275,12 @@ const ScrollView = createReactClass({
      * @platform ios
      */
     minimumZoomScale: PropTypes.number,
+     /**
+     * Enables nested scrolling for Android API level 21+.
```
janicduplessis: Could you also open a PR on https://github.com/facebook/react-native-website to add this prop?

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewManager.java
```diff
@@ -121,6 +124,14 @@ public void setOverScrollMode(ReactScrollView view, String value) {
     view.setOverScrollMode(ReactScrollViewHelper.parseOverScrollMode(value));
   }
 
+  /**
+   * Enables/Disables nested scrolling
```
janicduplessis: Same, remove comment.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -275,6 +275,12 @@ const ScrollView = createReactClass({
...
+     * Nested scrolling is supported by default in iOS platforms
```
janicduplessis: `Nested scrolling is supported by default on iOS` sounds better I think.

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
tuncaulubilge: 983e69f

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollViewManager.java
tuncaulubilge: 983e69f

Libraries/Components/ScrollView/ScrollView.js
tuncaulubilge: 👍  983e69f

Libraries/Components/ScrollView/ScrollView.js
tuncaulubilge: Added to https://github.com/facebook/react-native-website/pull/249

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
```diff
@@ -120,6 +123,11 @@ public void setOverScrollMode(ReactHorizontalScrollView view, String value) {
     view.setOverScrollMode(ReactScrollViewHelper.parseOverScrollMode(value));
   }
 
+  @ReactProp(name = "nestedScrollEnabled")
+  public void setNestedScrollEnabled(ReactScrollView view, boolean value) {
```
mdvacca: Don't you need to use a ReactHorizontalScrollView here?

tuncaulubilge: Fixed in 95b1e95
Thanks

Albert0405:
Hi, Is this a fix for nested ScrollView or ScrollView inside FlatList to have scrolling event for the child ScrollView? 

Sorry im newbie and i have no idea how to use this.

tuncaulubilge:
@Albert0405 FlatList extends ScrollView and this PR adds the prop to ScrollView, so it fixes the issues for both. 

This PR has been accepted but not yet released in the latest version, so you can not use this yet. When it's available, you can just add `nestedScrollEnabled: true` to your nested view.

A temporary solution to this issue would be to manually find the FlatList in the native android code, and call `setNestedScrollEnabled` on that view. Something like this: 
```
reactRootView.setOnHierarchyChangeListener(object : OnHierarchyChangeListener {
            override fun onChildViewRemoved(parent: View?, child: View?) {

            }

            override fun onChildViewAdded(parent: View?, child: View?) {
                if (child is ScrollView) {
                    ViewCompat.setNestedScrollingEnabled(child, true)
                }
            }
        })
```

Albert0405:
@tuncaulubilge Ok thank you, i guess i have to wait untill they release it. I actually found a solution using this [react-native-nested-scroll-view](https://github.com/cesardeazevedo/react-native-nested-scroll-view) i was able to scroll a large amount of text wrapped inside a scrollview that is rendered inside a flatlist but it only works on latest android version. 4.4.4 Android Version below is not working using this plugin.

tuncaulubilge:
@Albert0405 This pr only supports 5.0 and above. I'm afraid there is no easy fix for nested scrolling on 4.4 and below

PR #18302, opened 2018-03-09
Make perspective transformation look exactly same on iOS and Android

syaau:
Summary:
There is a variation in iOS and Android output while performing perspective transformation. The variation exists even when used in Android devices with different screen density. 

Test Plan:
I carried out tests on multiple devices with different perspective value and varying rotation X angle, and back calculated the Camera Distance value from the screen shots. On iOS the values were almost equal to the `Pixel Density * Perspective` value wheres on Android devices the Distance would always come out to be approximately `2.24 * Perspective` value irrespective of the device screen density. I used the following code to use perspective transformation and verified that output matched exactly on all platforms.
```javascript
const PERSPECTIVE = Platform.OS === 'ios' ? P : (P * PixelRatio.get()) / Math.sqrt(5);
```
The code was then translated to the `BaseViewManager.java` and verified that the output matched on all platforms.

Changelog:
[ANDROID] [BUGFIX] [Perspective Transformation] - Corrected perspective transformation to look exactly same on iOS and Android devices.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -234,11 +234,11 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
       float scale = DisplayMetricsHolder.getScreenDisplayMetrics().density;
 
       // The following converts the matrix's perspective to a camera distance
-      // such that the camera perspective looks the same on Android and iOS
-      // While comparing the output between android and ios, it seemed like
-      // the native android implementation removed the screen density from the
-      // calculation, so taking square and also the normalization value at
-      // sqrt(5) produced an exact replica with ios
+      // such that the camera perspective looks the same on Android and iOS.
+      // The native Android implementation removed the screen density from the
+      // calculation, so squaring and a normalization value of
+      // sqrt(5) produced an exact replica with ios.
+      // For more information, see https://github.com/facebook/react-native/pull/18302
       float normalizedCameraDistance = scale * scale * cameraDistance * CAMERA_DISTANCE_NORMALIZATION_MULTIPLIER;
       view.setCameraDistance(normalizedCameraDistance);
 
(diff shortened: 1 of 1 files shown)


spotifysupport:
Agreed with Syaau, as he mentioned above all the prospectives are a valuable end to end, iOS and android are very far different from each other, for further information do visit our website. [Rocketmail Customer Service ](babasupport.org/email/rocketmail-customer-service-number/384)

cpojer:
@syaau thanks for this PR. Unfortunately it is a bit hard through the test plan and code to understand what is changing here. Would you be able to provide some before/after screenshots that helps us understand your change? Thank you.

syaau:
Here is an example code, that rotates a 200x200 square by 45deg on x-axis with a very near perspective distance (120 units).

```javascript

import React from 'react';
import { StyleSheet, View, Platform, PixelRatio } from 'react-native';

const PERSPECTIVE = 120;
const perspective = PERSPECTIVE;
// const perspective = Platform.OS === 'ios' ? PERSPECTIVE : (PERSPECTIVE * PixelRatio.get()) / Math.sqrt(5);

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: '#F5FCFF',
  },
  ref: {
    position: 'absolute',
    width: 200,
    height: 200,
    backgroundColor: 'yellow',
  },
  perspective: {
    position: 'absolute',
    width: 200,
    height: 200,
    backgroundColor: 'red',
    opacity: 0.5,
    transform: [
      { perspective },
      { rotateX: '45deg' },
    ],
  },
});

const App = () => (
    <View style={styles.ref} />
    <View style={{ position: 'absolute', justifyContent: 'center', alignItems: 'center' }}>
      <View style={styles.perspective} />
    </View>
  </View>
);

export default App;
```

iPhone Output (same irrespective of pixel density of the phone)
simulator screen shot - iphone x - 2019-01-30 at 00 08 09

Android Output (on a iphone5x, pixel density of 2.625)
screenshot_1548787040

Android Output (on Pixel 2 XL, pixel density of 3.5)
screenshot_1548788035
**The output went below the screen with the same perspective value**

Android Output (after applying the fix, Use the commented line on above code)
screenshot_1548787625

syaau:
@cpojer Please check the screenshots above, where you could see that the perspective values didn't work correctly in between the android devices with different pixel densities. After applying the fix, the output looks exactly the same with the iOS and irrespective of the Pixel Density. Above in the example code, I have applied the fix on the perspective value itself, whereas this pull request implements the same thing in android.

I am not sure why the magic number 5 had been used previously to calculate the camera distance.
`private static final float CAMERA_DISTANCE_NORMALIZATION_MULTIPLIER = 5;`.

I guess it has got something to do with the matrix that has been used for transformation. But sqrt(5) produced the same output as iOS.

cpojer:
Having a bit of a hiccup with the import right now but I'm gonna rebase this PR and will get it landed as soon as possible.

PR #18340, opened 2018-03-12
Adds _lineHeight = NAN; to RCTTextAttributes

aparedes:
## Motivation

On iOS, when the parent Text contains the `lineHeight` style prop, and the children are also Text components they don't inherit the lineHeight prop.

## Test Plan

This is for **iOS** only.
Create a react-native project with React Native 0.54.0 or 0.54.1 and change the app to:
```js

type Props = {};
  render() {
    return (
        <Text style={{ fontWeight: 'bold', lineHeight: 40 }}>
          <Text style={{ color: 'orange' }}>I am bold and orange, </Text>
          <Text style={{ color: 'red' }}>I am bold and red, </Text>
          <Text style={{ color: 'blue' }}>I am bold and blue, </Text>
          <Text style={{ color: 'purple' }}>I am bold and purple, </Text>
          <Text style={{ color: 'yellow' }}>I am bold and yellow, </Text>
          <Text style={{ color: 'pink' }}>I am bold and pink</Text>
        </Text>
      </View>
    );
  }
}

  container: {
    backgroundColor: 'black',
    flex: 1,
    justifyContent: 'center',
    padding: 50,
  },
});
```
It displays: 
iphone - 2018-03-12 at 21 51 53

But should look like:
iphone - 2018-03-12 at 21 48 15

## Related PRs

New &lt;Text> iOS https://github.com/facebook/react-native/commit/2716f53

## Release Notes

[IOS] [MINOR] [Text] - Inherit lineheight

### DIFF
--- a/Libraries/Text/RCTTextAttributes.m
+++ b/Libraries/Text/RCTTextAttributes.m
@@ -21,6 +21,7 @@ NSString *const RCTTextAttributesTagAttributeName = @"RCTTextAttributesTagAttrib
   if (self = [super init]) {
     _fontSize = NAN;
     _letterSpacing = NAN;
+    _lineHeight = NAN;
     _textDecorationStyle = NSUnderlineStyleSingle;
     _fontSizeMultiplier = NAN;
     _alignment = NSTextAlignmentNatural;


shergin:
@aparedes ❤️

PR #18348, opened 2018-03-13
[Android] Add HTTP cache by default (like iOS)

### LINKED ISSUE #16795: Networking on Android doesn't use Cache
The HTTP Client on Android (OkHttpClient), that is used for networking, doesn't have cache, so Cache-Control headers are not respected.

### Is this a bug report?

Yes

Yes

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.9.0
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.9.0
  Xcode: Xcode 9.0.1 Build version 9A1004

  react: 16.0.0 => 16.0.0
  react-native: ^0.50.1 => 0.50.1

### Steps to Reproduce

If you use `fetch` to perform a request, and the response has `Cache-Control` headers, then they should be respected. For example, if the response has the `Cache-Control: max-age=60` headers, then, after a successful request the response should be cached. Subsequent requests in the next minute should serve the content from cache.

1. Perform sequential requests for resource that has cache control set
2. Check the response to realize if it came from the device http cache or the server was hit

### Expected Behavior

Cache headers should be respected and subsequent requests should be served from the cache.

### Actual Behavior

On iOS cache headers are respected and responses are served from the cache when necessary.
On Android cache headers are not respected and server is always hit.

### Reproducible Demo

Server app that returns a random number between 0 - 100. It sets cache-control header:
https://github.com/andon/cache-tester-server/blob/master/index.js
https://evening-temple-40153.herokuapp.com/route/random

React-Native App that targets the server app and can be used to showcase the different behaviour on iOS and Android:
https://snack.expo.io/SJJLdpBkz

Minishlink:
## Motivation

On iOS, the cache is enabled by default but it's not the case on Android. This PR adds 10Mo of HTTP cache by default. 10Mo was chosen arbitrarily.

Fixes #16795

## Test Plan

I wrote it so there's no breaking change for those already using the `OkHttpClientProvider.createClient` and `OkHttpClientProvider.createClientBuilder` public methods.

1. react-native init TestAndroidCache
2. Followed https://facebook.github.io/react-native/docs/android-building-from-source.html to build  with this branch as react-native
3. Added the following code in the App.js (with correct URL_WITH_CACHE_CONTROL)
4. Made sure the debugger is closed
5. Throttled my mac's network speed with the `Network link conditioner` (edge)
6. Tapped "Fetch", it shows ~4000
7. Tapped "Fetch" again, it shows ~6

```js
import {
  StyleSheet,
  Alert,
  View,
  Button
} from 'react-native';

type Props = {};
  fetch = () => {
    const start = Date.now();
    fetch('{URL_WITH_CACHE_CONTROL}').then(() => {
      const diff = Date.now() - start;
      Alert.alert(diff.toString());
    });
  };

  render() {
    return (
        <Button title={"fetch"} onPress={this.fetch} />
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
});
```

## Related PRs

None

## Release Notes

[ANDROID] [FEATURE] [Networking] Add HTTP caching

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -161,7 +161,7 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
    * @param context the ReactContext of the application
    */
   public NetworkingModule(final ReactApplicationContext context) {
-    this(context, null, OkHttpClientProvider.createClient(), null);
+    this(context, null, OkHttpClientProvider.createClient(context), null);
   }
 
   /**
@@ -172,7 +172,7 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
   public NetworkingModule(
     ReactApplicationContext context,
     List<NetworkInterceptorCreator> networkInterceptorCreators) {
-    this(context, null, OkHttpClientProvider.createClient(), networkInterceptorCreators);
+    this(context, null, OkHttpClientProvider.createClient(context), networkInterceptorCreators);
   }
 
   /**
@@ -181,7 +181,7 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
    * caller does not provide one explicitly
    */
   public NetworkingModule(ReactApplicationContext context, String defaultUserAgent) {
-    this(context, defaultUserAgent, OkHttpClientProvider.createClient(), null);
+    this(context, defaultUserAgent, OkHttpClientProvider.createClient(context), null);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


Minishlink:
cc @hramos as you reviewed recent PRs about networking ;)

dulmandakh:
Could you please make cache size to be configurable from manifest or gradle scripts.

Minishlink:
@hramos [OKHTTP](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Cache.html) docs say these are `bytes`, I may be wrong but I think the unit should then be `MB` (`megabytes`) or `Mo` (`megaoctets`), as `Mb` stands for `megabits`

PR #18355, opened 2018-03-13
Wrong height when TextInput has an empty string

reyalpsirc:
## Motivation

The caret/cursor did not appear when the TextInput was empty. Found that the cause was because the frame of the TextInput had an height of 0

## Test Plan

Just fill and clear a TextInput and the caret/cursor will always appear there.

## Release Notes

[IOS] [BUGFIX] [TextInput] - This was causing the cursor/caret to not appear since the size of the frame had an height of 0.


janicduplessis:
Won't this cause width of an empty string to be measured incorrectly? 

cc @shergin

reyalpsirc:
All I know is that with the zero-width space the input gets 0 width because of that measurement and then the field disappears (which includes the caret).
To the user, it seems like no field is selected when filling an empty form.

PR #18362, opened 2018-03-14
Fix hyperlinks for Native Modules / UI Component

trentrand:
## Motivation

Multiple links on the More Resources documents page point to an invalid path for both platforms (iOS and Android).

## Release Notes
[DOCS][BUGFIX][more-resources.html] - Fix broken hyperlinks for Native UI Components and Native Modules.


hramos:
The docs were moved a while ago to https://github.com/facebook/react-native-website. Can you open the PR there instead?

PR #18370, opened 2018-03-14
Add inverted prop in SectionListExample

kyo504:
## Motivation
This PR adds an option to pass`inverted` prop to SectionListExample in RNTester. FlatListExample already has this option but it's not available in SectionListExample.

## Test Plan
Run RNTester app on device or simulator and select SectionListExample. Depending on switching `inverted` option, you can see either inverted list or not.

## Release Notes
[GENERAL][ENHANCEMENT][RNTeater] - Add inverted prop to SectionListExample


janicduplessis:
Looks good, thanks!

@facebook-github-bot shipit

kyo504:
@janicduplessis Thanks 👍

PR #18378, opened 2018-03-14
Make generic event types less platform-specific

rozele:
A previous PR (#18308) enabled lazy loading of view manager constants.

This PR enables any platform to use of generic bubbling and direct event types by exposing the constants `genericBubblingEventTypes` or `genericDirectEventTypes` on the UIManagerModule.

## Motivation

I am adding lazy view managers to react-native-windows, and would like to use the same approach of loading "generic" event types that apply to all view managers only once.

## Test Plan

Run jest tests. Run sample app on iOS and Android.

## Related PRs

#18308

## Release Notes

[GENERAL][ENHANCEMENT][MINOR][Libraries/ReactNative/requireNativeComponent.js] - Enable any platform to use generic event types.

PR #18381, opened 2018-03-14
feat: Remove platform check in requireNativeComponent to support additional platforms

jhampton:
## Motivation

Using platform checks can make it difficult to incorporate changes for other platforms (e.g. Windows). 
 Using a check for the underlying function accomplishes the same goal without relying on an Android-specific check.  This change allows this JS file to be re-used instead of copied and modified. 

## Test Plan

[X] Run Jest tests
[X] Test in RNTester on simulators
[X] Test in Playground

## Related PRs

No related PR's found :)

## Release Notes

[GENERAL] [ENHANCEMENT] [Libraries/ReactNative/requireFabricComponent.js] - Simplified check against UIManager to support additional platforms, removing Android-specific check

### DIFF
--- a/Libraries/ReactNative/requireFabricComponent.js
+++ b/Libraries/ReactNative/requireFabricComponent.js
@@ -55,7 +55,7 @@ function requireNativeComponent(
   function attachDefaultEventTypes(viewConfig: any) {
     // This is supported on UIManager platforms (ex: Android),
     // as lazy view managers are not implemented for all platforms.
-    // See [UIManager] for details on constants and implementation.
+    // See [UIManager] for details on constants and implementations.
     if (UIManager.ViewManagerNames) {
       // Lazy view managers enabled.
       viewConfig = merge(viewConfig, UIManager.getDefaultEventTypes());
--- a/Libraries/ReactNative/requireFabricComponent.js
+++ b/Libraries/ReactNative/requireFabricComponent.js
@@ -53,22 +53,21 @@ function requireNativeComponent(
   extraConfig?: ?{nativeOnly?: Object},
 ): React$ComponentType<any> | string {
   function attachDefaultEventTypes(viewConfig: any) {
-    if (Platform.OS === 'android') {
-      // This is supported on Android platform only,
-      // as lazy view managers discovery is Android-specific.
-      if (UIManager.ViewManagerNames) {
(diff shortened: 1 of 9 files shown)


rozele:
@jhampton #18378 😆

jhampton:
Quick question @rozele – I wasn't sure if the `else if` you've included was necessary.  Is there another implementation we might need to support (either bubbling or lazy and ...?)

rozele:
@jhampton the else if is to prevent iOS from calling `merge` - but I now see that merge works as expected when either source or destination is null.

rozele:
I'm good with this PR if you add the same change to `requireNativeComponent`

jhampton:
Great, thank you!  Quick question @rozele : is there a reason this couldn't/shouldn't be refactored into a separate module and `require`d in both files?

rozele:
@jhampton - I'm supportive of a separate module.

jhampton:
I ended up making the same change to `requireNativeComponent` per @rozele 's original PR.  I'm not sure about the future of these two files vis-a-vis the transition to Fabric and historical context, so this is sufficient for now.

Also, it looks like CI is failing on `yarn` ATM. I'll check back in a bit later.

hramos:
The `yarn` failure appears to be related to a Circle CI outage. The outage is over, and my colleagues at github.com/facebook/docusaurus found that jobs spawned from newer commits are no longer seeing the `yarn` failure, but I haven't had luck on this repo yet.

You may want to try pushing a dummy commit to trick Circle into running again, just in case.

jhampton:
Thank you @hramos!  I tried again, and it looks like `yarn` is still not found.

image

hramos:
If you rebase past https://github.com/facebook/react-native/commit/6f6084db69de372cbc6dcec259389e4b7dfd7fa5, it should be fixed. Sorry for the trouble!

jhampton:
Interesting...it looks like there are quite a few eslint failures in the new tests.  I should be up to date as of this morning's `master`.

grabbou:
I think I've seen CI to fail for other lint related reasons as well earlier today. I'll look into that too as I am planning on cutting few more releases today.

Meanwhile, let's try merging that.

grabbou:
Interesting... 🤔

Issue #18384, opened 2018-03-14
Build warning: m_jsThread will be initialized after m_delegate


fbartho:
When debugging react-native iOS, and it crashes (for unrelated reasons), it drops me in RCTObjcExecutor and metaphorically stabs me in the eye with a build warning:
`Field 'm_jsThread' will be initialized after field 'm_delegate'`

I believe this is simply an issue with initializer list ordering vs private member-variable ordering, and I'll be submitting a PR to fix it.

### Environment

```bash
$ react-native info
Environment:
  OS: macOS High Sierra 10.13.3
  Node: 9.7.1
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.x => 16.2.0
  react-native: 0.54.x => 0.54.2
```

### Expected Behavior

react-native should strive to minimize the build warnings that occur with normal use.

### Actual Behavior

![screen shot 2018-03-14 at 1 37 23 pm](https://user-images.githubusercontent.com/209712/37429540-e205b57e-278c-11e8-92a9-f0e9a1bc7b25.png)

Warning documented above is thrown on these lines: https://github.com/facebook/react-native/blob/0d03fddecad7af0bb7d7a1dc1846c91f12c195cd/React/CxxBridge/RCTObjcExecutor.mm#L42-L43

### Steps to Reproduce

Go to [RCTObjExecutor.mm Line 42](https://github.com/facebook/react-native/blob/0d03fddecad7af0bb7d7a1dc1846c91f12c195cd/React/CxxBridge/RCTObjcExecutor.mm#L42) & notice the build warning that Xcode shows you.

fbartho:
Yes, I've updated the Environment to show that it is reproducible on `0.54.2`

PR #18385, opened 2018-03-14
RCTObjcExecutor: fix initializer-list initialization order build warning.

fbartho:
Fixes: #18384

## Motivation

Previously line 42 had a warning of: `Field 'm_jsThread' will be initialized after field 'm_delegate'`

This appears to be because the private member fields are declared in a different order than the initializer-list was told to initialize things.

I chose to re-order the initializer list because that fixes the warning by matching the code structure to what the compiler was telling us it was going to do.

An alternate fix would be to change the order of the member variables declared in the `private:` section to match the coded order in the initializer list. This might be the right move if the initializer list had the correct intent, but this would be a breaking change from a behavior perspective, so I'd need somebody with more knowledge to suggest this is the right move.

## Test Plan

A. Examine line to see the warning is gone.
B. Get a C++ guru to figure out the intent of this file and see if this actually exposed a bug that recommends we go with the alternate fix suggested above.

## Release Notes

[IOS][MINOR][React/CxxBridge/] - Build warning fix

### DIFF
--- a/React/CxxBridge/RCTObjcExecutor.mm
+++ b/React/CxxBridge/RCTObjcExecutor.mm
@@ -39,8 +39,8 @@ public:
                   std::shared_ptr<ExecutorDelegate> delegate)
     : m_jse(jse)
     , m_errorBlock(errorBlock)
-    , m_jsThread(std::move(jsThread))
     , m_delegate(std::move(delegate))
+    , m_jsThread(std::move(jsThread))
   {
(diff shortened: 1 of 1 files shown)


hramos:
>> B. Get a C++ guru [...]

Paging @mhorowitz

mhorowitz:
@facebook-github-bot shipit

PR #18431, opened 2018-03-18
Add missing file to xcodeproj to fix iOS build

janicduplessis:
Some new files were missing in the xcodeproj

## Test Plan

Build RNTester

## Release Notes

[INTERNAL] [BUGFIX] [TextInput] - Add missing file to xcodeproj to fix iOS build

### DIFF
--- a/Libraries/Text/RCTText.xcodeproj/project.pbxproj
+++ b/Libraries/Text/RCTText.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


janicduplessis:
@hramos Do you think it would be possible to run iOS / tvOS / android gradle builds on FB internal CI? Maybe as a non-blocking check, it would probably help prevent those failures since otherwise it's hard to tell for fb employees if they break the OSS build.

janicduplessis:
Just looked at CI and the build failure got caught by test_ios_e2e and not test_ios (probably since it doesn't build the RCTText project). I think a good compromise to catch most OSS build errors would be to just build RNTester with xcodebuild / gradle, don't really need to run tests since those are probably run already with the internal buck build.

janicduplessis:
Looks like @fkgozali was on top of it already!

PR #18447, opened 2018-03-19
Make 'button' default accessibility trait / comp. type for Touchables

draperunner:
Hi! I first wrote a FR here https://react-native.canny.io/feature-requests/p/default-accessibility-traitaccessibilitycomponenttype-on-touchables, but I thought I might as well make the PR.

I find myself writing

```
accessibilityTraits="button"
accessibilityComponentType="button"
```

on all Touchable* components in my app. Why do not the Touchable* components have this set by default? I'm sure you have good reasons for having it as it is today, but I can't find them, and I'd like to argue for having these props set to "button" as default:

* All react-native apps are more accessible "out-of-the-box"

* I struggle to think of cases where you would want to have an accessible Touchable that should NOT have a trait/componentType specified

* Native Android/iOS buttons are all announcing themselves as buttons without the developer having to explicitly set trait/type.

* Less boilerplate code!

More on the first bullet point: Think about all the developers that do not care about accessibility. Their apps would become a lot more user friendly for people dependent on screen readers, without the developer having to do anything.
## Test Plan

* Ran `yarn test`.
* Used the Accessibility Inspector to inspect the RNTester app in simulator to see that buttons got the Button trait.

screen shot 2018-03-19 at 13 07 38

## Related PRs

https://github.com/facebook/react-native-website/pull/263

## Release Notes

Am I doing this correctly? :)

[IOS] [ENHANCEMENT] [Libraries/Components/Touchable] - Make "button" default accessibility trait / component type for Touchable components
[ANDROID] [ENHANCEMENT] [Libraries/Components/Touchable] - Make "button" default accessibility trait / component type for Touchable components

### DIFF
--- a/Libraries/Components/Touchable/TouchableBounce.js
+++ b/Libraries/Components/Touchable/TouchableBounce.js
@@ -76,7 +76,12 @@ const TouchableBounce = createReactClass({
   },
 
   getDefaultProps: function() {
-    return {releaseBounciness: 10, releaseVelocity: 10};
+    return {
+      accessibilityComponentType: 'button',
+      accessibilityTraits: 'button',
+      releaseBounciness: 10,
+      releaseVelocity: 10,
+    };
   },
 
   getInitialState: function(): State {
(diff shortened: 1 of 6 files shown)


getaaron:
This looks great 👍 would love to see it merged

PR #18455, opened 2018-03-20
[android] fix ReadableNativeMap.toHashMap() for nested maps and arrays

esamelson:
Commit https://github.com/facebook/react-native/commit/7891805d22e3fdc821961ff0ccc5c450c3d625c8 broke the previous behavior of `ReadableNativeMap.toHashMap()` for nested maps and arrays. Previously, all nested `ReadableNativeMap`s and `ReadableNativeArray`s were recursively converted to `HashMap`s and `ArrayList`s, but this is lost when only `getLocalMap()` is returned.

## Test Plan

Call `ReadableNativeMap.toHashMap()` on a map with values of type `ReadableNativeMap` and `ReadableNativeArray`. Verify the returned hash map has these converted to `HashMap` and `ArrayList`, respectively.

## Release Notes

[ANDROID] [BUGFIX] [ReadableNativeMap] - Fix toHashMap() for nested maps and arrays

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableNativeMap.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableNativeMap.java
@@ -13,6 +13,7 @@ import com.facebook.jni.HybridData;
 import com.facebook.proguard.annotations.DoNotStrip;
 
 import java.util.HashMap;
+import java.util.Iterator;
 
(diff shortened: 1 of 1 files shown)


mdvacca:
Thank you for working on this PR @esamelson
We are unable to import your PR because of CLA permissions in your account. Could you email cla@fb.com? they will set the right permissions for your account.

Thank you

Issue #12599, opened 2017-02-27
Possible bug with TextInput and Chinese input method on iOS


Kulbear:
### Description

I have some TextInput components in my project and there should be some problems with `onEndEditting` and `onBlur` props of TextInput component. If I make the TextInput blur before I confirm the entered content from the Chinese input method, all the changes I made during this round of focus will be lost. Check the video below, starts from 0:17, the video demonstrates the issue.

Basically, it should be an issue between the system Chinese input method and the RN TextInput component. I made a 30-second video record to demonstrate this issue.

https://www.youtube.com/watch?v=BG0WNHW2MEc

![repro](https://cloud.githubusercontent.com/assets/14886380/23396112/f80edaac-fdcc-11e6-933a-8457a13080fa.gif)

![repro2](https://cloud.githubusercontent.com/assets/14886380/23398121/4b954c34-fdd6-11e6-9ba2-d31987a1c682.gif)

### Reproduction

I don't know how to work with Chinese input methods on rnplay, so I've pushed a sample project with minimal code for reproduction.

[Current code with a workaround](https://github.com/Kulbear/ChineseTextInputIssue)

[Original bug reproduction](https://github.com/Kulbear/ChineseTextInputIssue/tree/be0bc064f1d900f7db28d5d3a47faebdef5b537b)

I created a new project. The only modified file is the `index.ios.js`

```
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Text,
  TextInput,
  View
} from 'react-native';

export default class ChineseTextInputIssue extends Component {
  constructor (props) {
    super(props);
    this.state = {
      text: ''
    };
  }

  render () {
    return (
      <View style={styles.container}>
        <TextInput
          style={styles.inputField}
          placeholder="Type here to translate!"
          value={this.state.text}
          onBlur={(evt) => console.log('onBlur event:', evt.nativeEvent.text)}
          onChange={(evt) => console.log('onChange event:', evt.nativeEvent.text)}
          onChangeText={(text) => this.setState({ text })}
          onSubmitEditing={(evt) => console.log('onSubmitEditing event:', evt.nativeEvent.text)}
          onEndEditing={(evt) => console.log('onEndEditing event:', evt.nativeEvent.text)} />
        <Text style={styles.previewText}>
          {this.state.text}
        </Text>
        <TextInput style={styles.inputField} placeholder="Another to focus, meaningless" />
      </View>
    );
  }
}

const styles = StyleSheet.create({
[...]

PR #18456, opened 2018-03-20
[iOS][TextInput] Fix controlled TextInput for Chinese (and other languages)

### LINKED ISSUE #18403: Controlled TextInput broken for Chinese (and other languages) in v0.54 on iOS
Controlled ```TextInput``` breaks the Chinese pinyin keyboard's autocomplete feature. A similar issue was raised a year and a half ago (#8265) and fixed by #7496. However, it appears that others are having the same problem (#12599, #18260, #18379), but re-filing the issue with the template filled out and an example to reproduce.

Note that this works correctly in v0.53. This may have something to do with the big ```Text```, ```TextInput``` refactor that dropped with v0.54 (Thanks, btw, @shergin and @hovox!).

I've included both a working (v0.53) and broken (v0.54) example below.

### Environment

Environment:
  OS: macOS High Sierra 10.13
  Node: 9.3.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.9.0
  Android Studio: Not Found

  react: 16.2.0 => 16.2.0

### Expected Behavior

Typing on a US keyboard would bring up Chinese characters corresponding to letters typed.

### Actual Behavior

Each new letter typed is considered individually, instead of along with the previous untranslated characters.

Shamelessly stealing an image from @ForU who did a nice job of showing the issue:

image

### Steps to Reproduce

Unfortunately, expo hasn't updated to the latest version of React Native. I've prepared a small project that demonstrates the issue [here](https://github.com/danielsuo/react-native-pinyin-bug). There is a working version on a branch [here](https://github.com/danielsuo/react-native-pinyin-bug/tree/v0.53); same project, just using RN v0.53 instead of v0.54.

magicien:
## Motivation

Fix #18403. It should support multistage text input (marked text).

In v0.54.2, when TextInput component has `defaultValue` or `value` prop, marked text is confirmed as soon as typed, and it doesn't show candidate words.

## Test Plan

- All tests of `yarn run test` are passed
- Test with [the sample app 1](https://github.com/magicien/react-native-pinyin-bug).
    - It worked. (It shows candidate words corresponding to the marked text)

### react-native 0.54.2 (with bug) (from #18403 @danielsuo)

app1_before

### with bugfix

app1_after

- Test with [the sample app 2](https://github.com/magicien/react-native-marked-text).
    - It worked with both singleline and multiline TextInput.
    - When `blur` event was emitted, the marked text was confirmed. (#12599)

### react-native 0.54.2 (with bug)

app2_before

### with bugfix

app2_after

## Release Notes

[IOS] [BUGFIX] [TextInput] - Fix controlled TextInput for Chinese (and other languages)


johnxie:
Hi

Is this resolved in the latest release?

bunnyc1986:
What's the plan merging this PR?

rplankenhorn:
@bunnyc1986 Given my above comment, I don't think this should be merged in at all.  It may fix #18403 but it causes more issues for #18874.

ssssssssssss:
@shergin

jamsch:
Confirmed. React Native treats every text entered separately which shouldn't be the case for languages like Chinese and Japanese since there's a large amount of characters that just won't fit on a keyboard.

In the instance of Japanese, you're only able to write in the single letter phonetics. For example:
* a => あ
* i => い
* u => う

When combining two (or more) letters, the following happens:

* ka => kあ - supposed to be: **か**
* shi => shい - supposed to be **し**
* ku => kう - supposed to be **く**
* tsu => tsう - supposed to be **つ**

Manually applying the fix above (from @iwasaki-d) seems to have solved the issue.

johnxie:
Thanks for confirming @jamsch.  Any chance to merge the changes from @iwasaki-d?

watanabeyu:
Hi.

I resolve this issue, the way use shouldComponentUpdate.  
Please see below snack.

https://snack.expo.io/Hkp55vn6G

timscott:
Here's a reusable component using @watanabeyu's workaround:

```jsx
import {TextInput} from 'react-native';

class MyTextInput extends Component {
  shouldComponentUpdate(nextProps){
    return this.props.value === nextProps.value;
  }
  render() {
    return <TextInput {...this.props} />;
  }
};

export default MyTextInput;
```

Please note that this will not work if you want to modify the value any time after initial render—for example, formatting phone numbers in `onChangeText`. In that case, just use TextInput or maybe introduce `alwaysUpdate` prop.

deboyblog:
Part of the test failure doesn't seem to be because of this pull request.
Is the test environment lacking watchman?
snipaste_2018-05-17_09-56-10

bruno02100:
What needs to be done for this pull request to be merged? I am also having problems with the portuguese keyboard and this solves it.

mpyw:
@timscott `shouldComponentUpdate` hacking seems to be abodon `<ref>.clear` and `<ref>.setNativeProps({ text: '' })`...

Ashoat:
Does this PR achieve the same goals as #19087? Can somebody explain the differences?

elliscwc:
hey guys, is this going to be reviewed and merged?

lhcn:
image
by the way,can anyone show me how to translate 'Cut' 'Copy' 'Paste' into Chinese

huowenxuan:
@lhcn 
设置Info.plist->Localization native development region，改为cn

lhcn:
@huowenxuan 谢谢大佬指导。我的是Expo，是在app.json里面改的，按照你的思路，搜索了一下，最终有点差别但是OK了。值是用代码获取到的，是zh_CN
```json
{
    "ios": {
      "infoPlist": {
        "CFBundleDevelopmentRegion":"zh_CN"
      }
    }
}
```

zerosrat:
I modify files in `node_modules/react-native/...` as file changes in this pull request, but the bug is still here. I wonder what I did wrong.
react-native: 0.55.4

roycclu:
@magicien thanks this works magically!
@zerosrat did you recompile the app after making the changes?

zerosrat:
@roycclu I did not compile react native code after I modify. Should I follow this [guide](https://facebook.github.io/react-native/docs/building-from-source) to make it work? Thanks for your reply!

sunnylqm:
ping  @shergin @kelset

mmmulani:
@randomor thanks for that example. Could you provide it as a Expo Snack or change to RNTester? I'm trying to fix this and want to make sure this problem is also solved

dwyanelin:
inspired by watanabeyu.

`import React, {Component} from 'react';
import {TextInput, Platform} from 'react-native';

export default class MyTextInput extends Component{
	state={
		fakeValue:this.props.value,//pure display here
		refresh:false,//right clear
	}

	componentDidUpdate(prevProps){//right clear
		if(prevProps.value!==this.props.value&&this.props.value===''&&Platform.OS==='ios'){
			this.setState({fakeValue:'', refresh:true}, ()=>{
				this.setState({refresh: false}, ()=>this.input.focus());//render null make blur perhaps, so do focus.
			});
		}
	}

	focus=()=>{
		this.input.focus();
	}

	blur=()=>{
		this.input.blur();
	}

	clear=()=>{
		this.input.clear();
	}

	setFakeValue=fakeValue=>this.setState({fakeValue});//when parent setState({input}), call this too.

	render(){
		if(this.state.refresh){//right clear
			return null;
		}

		return(
			<TextInput
				{...this.props}
				ref={ref=>this.input=ref}
				value={Platform.OS==='ios'?this.state.fakeValue:this.props.value}//only ios need this.
			/>
		);
	}
}`

mmmulani:
I instead fixed this in 892212bad2daadd373f4be241e4cd9889b0a1005. The approach I took still sends onChange events, allowing you to process the input in RN if you wish.

kelset:
Sweet!

magicien:
Thank you!! Finally I can use Japanese with React Native.

johnxie:
This is great, thank you.

watanabeyu:
Thank

ganlanshu0211:
Thanks!!!

471685249:
Thanks!!!

Issue #12994, opened 2017-03-17
Proguard app crash (Caused by: java.lang.NoSuchFieldError: no "Lcom/facebook/jni/HybridData)


jjdp:
### Description

I am getting a fatal exception when i use proguard on my release build.
`Caused by: java.lang.NoSuchFieldError: no "Lcom/facebook/jni/HybridData;" field "mHybridData" in class "Lcom/facebook/react/cxxbridge/CatalystInstanceImpl;" or its superclasses`

Here is the stacktrace:

```
03-17 07:06:58.852  1057  1432 I ActivityManager: START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.myApp.staging/com.myApp.MainActivity (has extras)} from uid 10008 on display 0
03-17 07:06:58.867  1057  1451 I ActivityManager: Start proc 2234:com.myApp.staging/u0a62 for activity com.myApp.staging/com.myApp.MainActivity
03-17 07:06:58.892  2234  2234 D FirebaseApp: com.google.firebase.auth.FirebaseAuth is not linked. Skipping initialization.
03-17 07:06:58.894  2234  2234 W InstanceID/Rpc: Failed to resolve REGISTER intent, falling back
03-17 07:06:58.895  2234  2234 W InstanceID/Rpc: Both Google Play Services and legacy GSF package are missing
03-17 07:06:58.896  2234  2234 W GooglePlayServicesUtil: Google Play Store is missing.
03-17 07:06:58.896  2234  2234 I DynamiteModule: Considering local module com.google.android.gms.flags:2 and remote module com.google.android.gms.flags:0
03-17 07:06:58.896  2234  2234 I DynamiteModule: Selected local version of com.google.android.gms.flags
03-17 07:06:58.897  2234  2234 W DynamiteModule: Local module descriptor class for com.google.android.gms.crash not found.
03-17 07:06:58.897  2234  2234 W GooglePlayServicesUtil: Google Play Store is missing.
03-17 07:06:58.897  2234  2234 I DynamiteModule: Considering local module com.google.android.gms.crash:0 and remote module com.google.android.gms.crash:0
03-17 07:06:58.897  2234  2234 E FirebaseCrash: Failed to initialize crash reporting: com.google.android.gms.d.bk$a: No acceptable module found. Local version is 0 and remote version is 0.
[...]

Issue #18460, opened 2018-03-20
Is RNTester still in maintain?


gengjiawen:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [x] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

From the commit history, it looks like that it not active for a while (like old api like mixin and no actual change to js file). And it also contains build bug in android (react root is not configure right). 

So my question is that:
* If it is not in the plan, can I create a pr to replace it with a simple Hello world app for easily do the android test. 
* If it still in the plan, I will create a pr to fix the android build problem.

hramos:
RNTester is still maintained, as it's part of our test infrastructure. It was renamed last year to RNTester to make it clear it's part of the test suite and not intended to be a showcase of example code (as it was originally intended).

I am not aware of the Android issue, and I would appreciate a PR to address it.

hramos:
Answered.

gengjiawen:
Thanks, I will create a pr to fix this, maybe integrate this into ci.

Issue #6624, opened 2016-03-24
Android release build with proguard enabled throws exception at start


dgladkov:
I get following exception on my release APK with proguard enabled after upgrading RN to 0.22.+. Disabling proguard solves the issue.

```
java.lang.NoSuchMethodError: no non-static method "Lcom/facebook/react/bridge/queue/NativeRunnable;.<init>(Lcom/facebook/jni/HybridData;)V"
   at com.facebook.react.bridge.ReactBridge.callFunction(Native Method)
   at com.facebook.react.bridge.CatalystInstanceImpl.callFunction(CatalystInstanceImpl.java:189)
   at com.facebook.react.bridge.JavaScriptModuleRegistry$JavaScriptModuleInvocationHandler.invoke(JavaScriptModuleRegistry.java:97)
   at java.lang.reflect.Proxy.invoke(Proxy.java:397)
   at $Proxy5.emit(Unknown Source)
   at com.facebook.react.modules.appstate.AppStateModule.sendAppStateChangeEvent(AppStateModule.java:73)
   at com.facebook.react.modules.appstate.AppStateModule.onHostResume(AppStateModule.java:51)
   at com.facebook.react.bridge.ReactContext.onHostResume(ReactContext.java:147)
   at com.facebook.react.ReactInstanceManagerImpl.moveToResumedLifecycleState(ReactInstanceManagerImpl.java:559)
   at com.facebook.react.ReactInstanceManagerImpl.moveReactContextToCurrentLifecycleState(ReactInstanceManagerImpl.java:904)
   at com.facebook.react.ReactInstanceManagerImpl.setupReactContext(ReactInstanceManagerImpl.java:728)
   at com.facebook.react.ReactInstanceManagerImpl.access$800(ReactInstanceManagerImpl.java:104)
   at com.facebook.react.ReactInstanceManagerImpl$ReactContextInitAsyncTask.onPostExecute(ReactInstanceManagerImpl.java:210)
   at com.facebook.react.ReactInstanceManagerImpl$ReactContextInitAsyncTask.onPostExecute(ReactInstanceManagerImpl.java:183)
   at android.os.AsyncTask.finish(AsyncTask.java:636)
   at android.os.AsyncTask.access$500(AsyncTask.java:177)
   at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:653)
   at android.os.Handler.dispatchMessage(Handler.java:102)
   at android.os.Looper.loop(Looper.java:135)
[...]

PR #18461, opened 2018-03-20
[android] Inline and fix proguard rules

gengjiawen:
## Motivation
The original proguard rules are put in the template, which is not very convenient and easy to get wrong. Because new rules get put, people also has two copy paste the rule. And there are also existing project import react native as a dependency. So the best way to  keep a android library project proguard rule is to manage the rule itself, using `consumerProguardFiles` like [dagger](https://github.com/JakeWharton/butterknife/blob/46baef6d96e93a1872efa17fdfc1d2af59b36f0c/butterknife/build.gradle#L9) and other android library project.

## Test Plan

Use RNTester to build the release flavor (now it has bugs https://github.com/facebook/react-native/issues/18460, I keep my change in local for now), after build success, run to check if crash.

In the process, I also fix https://github.com/facebook/react-native/issues/12994 and https://github.com/facebook/react-native/issues/6624 by adding the following to proguard rules
```proguard
-keep,includedescriptorclasses class com.facebook.react.bridge.** { *; }
```

## Release Notes

[ANDROID] [ENHANCEMENT and BUGFIX] [Proguard rules] - inline and fix proguard rules .


hramos:
The current test_android failure already exists on master. We'll need to fix that so you can rebase and ensure Android tests are green.

gengjiawen:
Can the ci thing fixed in today ?

hramos:
Of course.

gengjiawen:
@hramos Any chance we can fix ci asap, this pr has stuck for a while.

hramos:
@gengjiawen the major issues in the Android job were fixed a week or two ago. That fix surfaced some failures in the unit tests. Fixing those is not something that is a priority for me at the moment, and PRs to fix them would be appreciated.

gengjiawen:
I will try to fix it, though I have few knowledge about buck.

hramos:
If you're seeing buck errors, you may want to rebase. `test_android` is failing on master due to failing unit tests, not buck: https://circleci.com/gh/facebook/react-native/39551

In any case, the failing Android job shouldn't block your PR from being reviewed. These unit tests are failing on master and are unrelated to your changes.

PR #18470, opened 2018-03-20
Trigger onFocus/onBlur instead of onPressIn/onPressOut (eventually, but for now just deprecate)

jayphelps:
Currently on iOS and Android focus/blur events trigger onPressIn/onPressOut. Based on discussions with people are several companies who use react-native we're proposing instead triggering new events onFocus/onBlur. Initial discussion on Slack with some from the core team on Slack seemed positive.

Couple reasons:

* The current API behavior overloads onPressIn/onPressOut. That means on platforms like react-native-web, if focus/blur support was added (as we're hoping for), even though onPressIn/onPressOut would be useful as the name describes, you wouldn't be able to distinguish between it and browser element focus/blur events.
* The names aren't as self-documenting/intuitive as onFocus/onBlur, especially for react-dom users.

## Test Plan

There aren't any current tests around this, but I intend to add them if we solidify the API.

There's also an option question on the transition--do we deprecate the existing API with a warning? This PR just deprecates them, though it will on any TV platform when something becomes focused regardless of whether they use the API or not. This isn't ideal. It's not clear if there are alternatives or if just right away breaking the API for TV users is the correct solution, if we can get consensus between the few parties who are using it.

***

I'm interested to hear counter points or prior discussions.

Cc/ @matthargett @dlowder-salesforce @rozele

### DIFF
--- a/Libraries/Components/Touchable/Touchable.js
+++ b/Libraries/Components/Touchable/Touchable.js
@@ -325,11 +325,17 @@ const TouchableMixin = {
       evt.dispatchConfig = {};
       if (myTag === evt.tag) {
         if (evt.eventType === 'focus') {
-          cmp.touchableHandleActivePressIn &&
+          if (cmp.touchableHandleActivePressIn) {
+            console.warn('Using `touchableHandleActivePressIn` or `onPressIn` to listen for focus events is deprecated. Use `touchableHandleFocus` or `onFocus` instead.');
             cmp.touchableHandleActivePressIn(evt);
+          }
+          cmp.touchableHandleFocus(evt);
         } else if (evt.eventType === 'blur') {
-          cmp.touchableHandleActivePressOut &&
+          if (cmp.touchableHandleActivePressOut) {
+            console.warn('Using `touchableHandleActivePressOut` or `onPressOut` to listen for blur events is deprecated. Use `touchableHandleBlur` or `onBlur` instead.');
             cmp.touchableHandleActivePressOut(evt);
+          }
+          cmp.touchableHandleBlur(evt);
         } else if (evt.eventType === 'select') {
           cmp.touchableHandlePress &&
             !cmp.props.disabled &&
@@ -526,6 +532,27 @@ const TouchableMixin = {
       this._cancelLongPressDelayTimeout();
       this._receiveSignal(Signals.LEAVE_PRESS_RECT, e);
     }
+  },
+
+  /**
+   * Invoked when the item receives focus. Mixers might override this to
+   * visually distinguish the `VisualRect` so that the user knows that it
+   * currently has the focus. Most platforms only support a single element being
+   * focused at a time, in which case there may have been a previously focused
+   * element that was blurred just prior to this.
+   */
+  touchableHandleFocus: function (e: Event) {
+    this.props.onFocus && this.props.onFocus(e);
+  },
+
+  /**
+   * Invoked when the item loses focus. Mixers might override this to
+   * visually distinguish the `VisualRect` so that the user knows that it
+   * no longer has focus. Most platforms only support a single element being
+   * focused at a time, in which case the focus may have moved to another.
+   */
+  touchableHandleBlur: function (e: Event) {
+    this.props.onBlur && this.props.onBlur(e);
   },
 
   // ==== Abstract Application Callbacks ====
--- a/Libraries/Components/Touchable/TouchableBounce.js
+++ b/Libraries/Components/Touchable/TouchableBounce.js
@@ -56,7 +56,6 @@ const TouchableBounce = ((createReactClass({
 
   propTypes: {
     ...TouchableWithoutFeedback.propTypes,
-
     // The function passed takes a callback to start the animation which should
     // be run after this onPress handler is done. You can use this (for example)
     // to update UI before starting the animation.
--- a/Libraries/Components/Touchable/TouchableWithoutFeedback.js
+++ b/Libraries/Components/Touchable/TouchableWithoutFeedback.js
@@ -81,6 +81,18 @@ const TouchableWithoutFeedback = ((createReactClass({
       PropTypes.arrayOf(PropTypes.oneOf(AccessibilityTraits)),
     ]),
     /**
+     * When `accessible` is true (which is the default) this may be called when
+     * the OS-specific concept of "focus" occurs. Some platforms may not have
+     * the concept of focus.
+     */
+    onFocus: PropTypes.func,
+    /**
+     * When `accessible` is true (which is the default) this may be called when
+     * the OS-specific concept of "blur" occurs, meaning the element lost focus.
+     * Some platforms may not have the concept of blur.
+     */
+    onBlur: PropTypes.func,
+    /**
      * If true, disable all interactions for this component.
      */
     disabled: PropTypes.bool,


matthargett:
Needs release notes and an update to RNTester, but LGTM otherwise! :D

CoryG89:
Awesome. Looks good to me as well.

+1 for `onFocus` and `onBlur`.

dlowder-salesforce:
I'm not sure I understand these changes.... I had the impression that developers using RN would not use `Touchable` directly, but use `TouchableOpacity`, `TouchableHighlight`, etc., which implement `onTouchableHandleActivePressIn` (and `Out`).  So it seems to me that those files also need to change, yes?

Libraries/Components/Touchable/Touchable.js
```diff
@@ -325,9 +325,17 @@ const TouchableMixin = {
       evt.dispatchConfig = {};
       if (myTag === evt.tag) {
         if (evt.eventType === 'focus') {
-          cmp.touchableHandleActivePressIn && cmp.touchableHandleActivePressIn(evt);
+          if (cmp.touchableHandleActivePressIn) {
```
dlowder-salesforce: I had the impression that developers using RN would not use `Touchable` directly, but use `TouchableOpacity`, `TouchableHighlight`, etc., which implement `onTouchableHandleActivePressIn` (and `Out`).  So it seems to me that those files also need to change, yes?  And the way this has been done, it appears that components implementing `Touchable` methods aren't allowed to handle both press and focus events.  It seems to me that that would be a big breaking change for Apple TV, at the very least.

jayphelps: I'll double check the code soon. If I'm remembering correctly, this base mixin is included into TouchableOpacity, etc and is used as the single event handler entry point. This code then defers to `touchableHandleActivePressIn` and `touchableHandleFocus`. `touchableHandleActivePressIn` has default implementations in some of those components, e.g. TouchableBounce provides a `touchableHandleActivePressIn` which does the bounce effect and then calls `props.onPressIn` if it was provided. For focus, the idea in this PR would be that Touchable itself has the default implementation of `touchableHandleFocus` which then defers to `onFocus` if it was indeed provided. This was done mostly because some custom Touchables might want to have a style/effect change on focus, but I didn't want to make that opinion just yet, though I can definitely see some sort of default style being desirable--wanted to keep the scope narrow to make the discussion easier.

I'll circle back after I've had a chance to review the codebase again, but do let me know if I'm off base. 👍

jayphelps: Also to clarify, this PR as it stands does not remove the behavior of onPressIn being called for focus events, it just deprecates it so that it could be removed in a future release so that a true distinction could be made between onPressIn and onFocus, and in follow up PRs potentially opening up onFocus to other platforms than just TV

rozele: >touchableHandleActivePressIn [](start = 18, length = 28)

Would this preclude us from supporting keyboard navigation scenarios. E.g., get a focus event when tabbing to and pressIn event when you hit Enter or space?

Libraries/Components/Touchable/TouchableBounce.js
```diff
@@ -44,7 +44,19 @@ const TouchableBounce = createReactClass({
      * all the touchable elements are accessible.
      */
     accessible: PropTypes.bool,
-
+    /**
+     * When `accessible` is true (which is the default) this may be called when
+     * the OS-specific concept of "focus" occurs. Some platforms may not provide
+     * focus support have the concept of focus.
+     */
+      onFocus: PropTypes.func,
```
rozele: >   [](start = 4, length = 2)

looks like whitespace is a bit sloppy here.

jayphelps: good catch!

Libraries/Components/Touchable/Touchable.js
jayphelps: To try and minimize controversy this PR deprecates the behavior of onPressIn/Out being called for focus events on TVs, with the goal of eventually removing it so that on platforms where onFocus _and_ onPressIn make sense it would be supported. As far as Enter/space specifically, I believe that'd be a platform-specific behavior, e.g. on the web hitting space inside of an input does not by default do anything other than add a space character to the input value.

Does that make sense?

jayphelps: @dlowder-salesforce I checked the code again, my above description appears correct. Lmk I'm mistaken.

jayphelps:
Does anyone have further feedback on this before I continue to try and add tests?

It's so far unclear how exactly I'd test this as there aren't any existing tests AFAICT and testing it fully require interactions between native->JS. Regardless I'm gonna spend some additional time attempting to do so--if you have suggestions or prior art, lmk! ✌️

Libraries/Components/Touchable/Touchable.js
dlowder-salesforce: So eventually, TouchableHighlight and TouchableOpacity would not work on tvOS?

jayphelps: @dlowder-salesforce oops that's not the intent, can you elaborate?

dlowder-salesforce: Currently in tvOS, a focus event fires onPressIn, a blur event fires onPressOut.  This allows TouchableHighlight and TouchableOpacity to work very naturally without changes on tvOS, and therefore an iOS and tvOS app can share a lot of React Native UI code.  It seemed to me that the changes in this PR were heading in a direction where this would eventually not be possible -- did I misunderstand what you are trying to do?

rozele: @dlowder-salesforce how do you differentiate between focus and holding a button on the remote? I would think the onFocus event would trigger a different highlight than holding a button down.

dlowder-salesforce: Thanks @jayphelps .... sorry for my slowness.  I read it over again, and had a good discussion about it with @matthargett .  Looks good.

jayphelps:
btw the build is failing on android because a facebook-specific buck flag has slipped into master (unrelated to this PR) ✌️ 

```
BUILD FAILED: Buck wasn't able to parse /home/circleci/react-native/ReactAndroid/src/main/jni/react/jni/BUCK:
Incorrect arguments to cxx_library with name jni: Extra unknown kwargs: fbandroid_allow_jni_merging
```

grabbou:
@facebook-github-bot shipit

jayphelps:
Does RN have an official deprecation process? When this is merged should I create the follow up PR to remove the deprecated calling of onPressIn/Out for focus events?

grabbou:
I don't think there's anything official - we usually do the depreciation PR
and then, remove it few versions later entirely.

jayphelps:
@grabbou Thanks for your help! Anything I can do to help get this over the finish line?

grabbou:
I think just a final word from @matthargett and @dlowder-salesforce as people interested in this proposal would be good.

Was there anyone else involved in Slack discussions? To be honest, I don't recall what was the conclusion there.

Libraries/Components/Touchable/Touchable.js
dlowder-salesforce: @rozele on Apple TV, we don't do any special highlighting when the remote button is held down.  In general, TV apps highlight a view only when it comes into focus as a result of arrow key or touchpad swipe navigation.

matthargett:
final word(s): LGTM #shipit

jayphelps:
:shipit:

grabbou:
@jayphelps can I ask you for resolving conflicts for the last time and we shipt it?

jayphelps:
@grabbou master is currently failing 😢 looks like the snapshots need updating

grabbou:
@jayphelps we just need to resolve two conflicts in this PR by hitting `Resolve conflicts` button and I'll merge straight away (even if the tests fail later on since that's reasonable).

jayphelps:
@grabbou rebased!

jayphelps:
bump

jayphelps:
rebased.

grabbou:
@facebook-github-bot shipit

PR #18488, opened 2018-03-21
Improve cryptic error message when creating a component starting with a lowercase letter

### LINKED ISSUE #18473: Pointlessly cryptic error message when creating a component starting with a lowercase letter
If you try to render a component and it starts with a lowercase letter, you'll see an error like:

> Invariant Violation: View config not found for name myComponent

This error message is confusing and doesn't offer any recovery suggestions.

We should change this to say:

> Invariant Violation: View config not found for name myComponent. You probably made a component that starts with a lowercase letter. Try changing it to an uppercase letter.

Looks like this would need to be changed in a total of four places (twice in each of two files):

https://github.com/facebook/react-native/blob/22004db81912f66650e9f9776d0045b88df61b5b/Libraries/Renderer/ReactFabric-prod.js#L5652-L5663

https://github.com/facebook/react-native/blob/22004db81912f66650e9f9776d0045b88df61b5b/Libraries/Renderer/ReactNativeRenderer-prod.js#L6060-L6069

I would be happy to open a PR to resolve this if someone can explain the difference between `ReactFabric-prod.js` and `ReactNativeRenderer-prod.js` and which of the 4 strings should be modified.

getaaron:

Fixes #18473

## Test Plan

Ran automated test suite and visually observed changes.

## Related PRs

N/A

## Release Notes

[GENERAL] [ENHANCEMENT] [Libraries/Renderer/*.js] - Improve cryptic error message when creating a component starting with a lowercase letter

### DIFF
--- a/Libraries/Renderer/ReactFabric-dev.js
+++ b/Libraries/Renderer/ReactFabric-dev.js
@@ -3838,7 +3838,7 @@ function get$1(name) {
     var callback = viewConfigCallbacks.get(name);
     invariant(
       typeof callback === "function",
-      "View config not found for name %s",
+      "View config not found for name %s. You probably made a component that starts with a lowercase letter. Try changing it to an uppercase letter.",
       name
     );
     viewConfigCallbacks.set(name, null);
@@ -3847,7 +3847,7 @@ function get$1(name) {
   } else {
     viewConfig = viewConfigs.get(name);
   }
-  invariant(viewConfig, "View config not found for name %s", name);
+  invariant(viewConfig, "View config not found for name %s. You probably made a component that starts with a lowercase letter. Try changing it to an uppercase letter.", name);
   return viewConfig;
 }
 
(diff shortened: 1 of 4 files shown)


getaaron:
@hramos The `test_android` failure looks unrelated to this PR

fkgozali:
The source of truth of this error message is actually the React codebase, here: https://github.com/facebook/react/blob/94f44aeba72eacb04443974c2c6c91a050d61b1c/packages/react-native-renderer/src/ReactNativeViewConfigRegistry.js#L47

You should actually create a PR for https://github.com/facebook/react repo because we sync React to RN repo periodically, and this change would then get clobbered if React repo doesn't have it yet.

cc @sebmarkbage

hramos:
Import cancelled, please open PR against react repo.

getaaron:
@hramos @fkgozali Thanks for the quick review 👍 I moved the changes over to https://github.com/facebook/react/pull/12421 in case you want to comment/approve there.

PR #18496, opened 2018-03-21
Update fresco to v1.8.1

AndrewJack:
The current version of fresco in use is [v1.3.0](https://github.com/facebook/fresco/releases/tag/v1.3.0) from April 2017. There has been a lot of [improvements](https://github.com/facebook/fresco/releases) since then.

Fresco also depends on OkHttp 3.8.0 and soloader 0.3.0 so I have updated these too. Let me know if there's any reason either of these shouldn't be updated, however this will prevent Fresco from being updated.

The latest version of OkHttp is 3.10.0, but I have kept it as low as possible to allow developers to choose the version.

## Test Plan

Test with local build.
Check CI passes.

## Release Notes

[ANDROID] [ENHANCEMENT] [Fresco/OkHttp/SoLoader] - Updates Fresco to 1.8.1, OkHttp to 3.8.0, & SoLoader to 0.3.0

cc @foghina

PR #18499, opened 2018-03-22
Add fresco proguard rules

AndrewJack:
Adds proguard rules, which I missed off from the #18496 PR.

I've added the proguard rules as suggested in http://frescolib.org/docs/shipping.html, with the duplicates & unused libraries removed.

## Test Plan

CI passes.
Build example project with proguard turned on.

## Related PRs

#18496

## Release Notes

[ANDROID] [BREAKING] [proguard] - Update Proguard config in template. If upgrading make sure you update your project's proguard file.

cc @foghina & @hramos


foghina:
Thanks again <3

PR #18500, opened 2018-03-22
Fixed 0.54 - iOS] fetch HEAD requests fail with `Invalid response for blob`

### LINKED ISSUE #18223: [0.54 - iOS] fetch HEAD requests fail with `Invalid response for blob`
Performing a `HEAD` request with `fetch` causes an `Invalid response for blob` error on iOS

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.6.0
  Yarn: Not Found
  npm: 5.3.0
  Watchman: 4.9.0

### Expected Behavior

The request should complete without error.

### Actual Behavior

The RCTNetworking native module in iOS does not construct an empty blob for `HEAD` responses like Android does. Instead it returns an empty string as the response. However, `whatwg-fetch` always sets the `responseType` to `blob` for the XMLHttpRequest so the empty string is invalid.

Relevant lines of code:
https://github.com/github/fetch/blob/fcc4e1b48cfb5a2b1625fcd6eac06d954b00ccb6/fetch.js#L454
https://github.com/facebook/react-native/blob/6eef7de46ee5c5004854c1cdd5cd83f2a608f622/Libraries/Network/XMLHttpRequest.js#L253-L258

```
Invalid response for blob: 
handleException @ index.bundle?platform=ios&dev=true&minify=false:13132
handleError @ index.bundle?platform=ios&dev=true&minify=false:12174
reportFatalError @ index.bundle?platform=ios&dev=true&minify=false:783
__guardSafe @ index.bundle?platform=ios&dev=true&minify=false:2360
callFunctionReturnFlushedQueue @ index.bundle?platform=ios&dev=true&minify=false:2165
t @ RNDebuggerWorker.js:1
```

### Steps to Reproduce

1 Clone and run https://github.com/lingokids/FetchError

(I tested on an iPhone 7)

animaonline:
Fixes #18223

## Test Plan
 
```
fetch('https://apipre.monkimun.com/whoami', {
  body: null,
  method: 'HEAD',
  headers: {
    Accept: 'application/json',
    'Content-Type': 'application/json',
  },
})
  .then(response => response.json())
  .then(json => console.log(json));
```

Additional information can be found in the issue #18223

## Release Notes
CATEGORY

[ANDROID] [BUGFIX] [XMLHttpRequest] - Fixed an issue where fetch HEAD requests would fail with `Invalid response for blob`
[IOS] [BUGFIX] [XMLHttpRequest] - Fixed an issue where fetch HEAD requests would fail with `Invalid response for blob`

### DIFF
--- a/Libraries/Network/XMLHttpRequest.js
+++ b/Libraries/Network/XMLHttpRequest.js
@@ -254,7 +254,11 @@ class XMLHttpRequest extends EventTarget(...XHR_EVENTS) {
         if (typeof this._response === 'object' && this._response) {
           this._cachedResponse = BlobManager.createFromOptions(this._response);
         } else {
-          throw new Error(`Invalid response for blob: ${this._response}`);
+          if (this._response === '')  {
+            this._cachedResponse = null;
+          } else {
+            throw new Error(`Invalid response for blob: ${this._response}`);
+          }
         }
         break;
 


janicduplessis:
Do you think it would be better to not parse the body as a blob when the method is HEAD instead? I wonder what the spec says about that.

Otherwise we might end up returning null when it should actually throw.

animaonline:
As far as I know there’s no body in a HEAD request.

janicduplessis:
@animaonline Yea, HEAD responses don't have a body so my question is why are we trying to parse it as a blob. I think fixing that would be better.

janicduplessis:
Alright so it looks like whatwg-fetch uses blob whenever it is available (https://github.com/github/fetch/blob/27473e937c6bdbb61212cc078ad60d22ae77ffb1/fetch.js#L454).

Not sure if this is ideal, but XMLHttpRequest should be able to handle null body + responseType blob. I read through the spec and it doesn't really explicitly mention how to handle it but I tested it in chrome and it does return a blob instance of size 0 and type text/plain. It seems like this is also what the Android implementation does.

I suggest we fix the iOS implementation so it also returns an empty blob. Furthermore we should investigate if having the responseType always set to blob when using fetch could cause issues.

janicduplessis:
Also found this PR that makes sure GET and HEAD returns a null body as per the spec, but it is only implemented for Android.

https://github.com/facebook/react-native/commit/d52569c4a1b6bd19792e4bda23e3a8c3ac4ad8df

https://xhr.spec.whatwg.org/#dom-xmlhttprequest-send

animaonline:
It should definitely be fixed in the future, but for now I believe this will do, this issue is critical, it also seems to happen when executing a GET fetch request that is expected to return JSON, but returns html (for example a 404 page)., or when executing a fetch request while offline. 

My PR takes care of that problem, but I agree that it should be handled in a better way.

xiamx:
This patch is identical to the one I proposed on https://github.com/facebook/react-native/issues/18223#issuecomment-372844446 10 days ago... Yet, no attribution.

The solution proposed here is only a workaround that masks the source of the issue. The creation of `_cachedResponse` should be fixed for the case of empty body instead.

animaonline:
Well, the logic in this "patch" is similar to how XMLHttpRequest handles other response types, as I've written before this bug also applies to GET requests, and in our case it's really critical, you don't really expect me to attribute you for 2 lines of code that borrows from the existing logic, do you?

janicduplessis:
@animaonline Thanks for giving a shot at fixing this, I submitted a new patch that fixes the underlying issue in the iOS implementation, so this workaround won't be needed. See #18547 for a more detailed explanation of the issue.

peacechen:
This error also manifests with RN Android.  This PR should be applied as a safety net in case the underlying platform generates an empty blob.  While an empty blob may not be in spec, this is causing fatal errors in multiple platforms (iOS *and* Android, has anyone test Windows?).  Expecting all platforms to be maintained equally isn't a good assumption and makes RN more brittle.

PR #18526, opened 2018-03-23
[iOS][TextInput] Add iOS 10 textContentType for TextInput

cbrevik:
## Motivation

Setting `textContentType` will provide the keyboard and system with semantic meaning for inputs. Should enable password/username autofill in apps running on iOS 11+ as demonstrated here: https://developer.apple.com/videos/play/wwdc2017/206/

Also gives you the ability to disable autofill by setting `textContentType="none"`: https://stackoverflow.com/questions/48489479/react-native-disable-password-autofill-option-on-ios-keyboard

## Test Plan
Adding `textContentType` equal to `username` or `password`  should give you an autofill-bar over the keyboard which will let you fill in values from the device Keychain:

image

Setting the appropriate `textContentType` will fill in the correct value in the `TextInput`.
I have only been able to get this to work on device, and not simulator. 

Usage:
```jsx
<TextInput
    value={this.state.username}
    onChangeText={this.setUserName}
    textContentType="username"
/>
```

```jsx
<TextInput
    value={this.state.password}
    onChangeText={this.setPassword}
    secureTextEntry={true}
    textContentType="password"
/>
```

To disable:
```jsx
<TextInput
    value={this.state.password}
    onChangeText={this.setPassword}
    secureTextEntry={true}
    textContentType="none"
/>
```

This will set `textContentType` to an empty string: https://stackoverflow.com/a/46474180/5703116

## Related PRs

Docs PR: https://github.com/facebook/react-native-website/pull/277

## Release Notes

[IOS] [MINOR] [TextInput] - Added `textContentType` prop for iOS 10+. Will enable password autofill for iOS 11+.

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -167,10 +167,12 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
   }
 }
 
-- (void)setTextContentType:(NSString*)type
+- (void)setTextContentType:(NSString *)type
 {
   #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0
     if (@available(iOS 10.0, *)) {
+        // Setting textContentType to an empty string will disable any
+        // default behaviour, like the autofill bar for password inputs
         self.backedTextInputView.textContentType = [type isEqualToString:@"none"] ? @"" : type;
     }
   #endif
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -167,6 +167,15 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
   }
 }
 
+- (void)setTextContentType:(NSString*)type
+{
+  #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0
+    if (@available(iOS 10.0, *)) {
+        self.backedTextInputView.textContentType = [type isEqualToString:@"none"] ? @"" : type;
+    }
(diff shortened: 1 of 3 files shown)


Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -167,6 +167,15 @@ - (void)setSelection:(RCTTextSelection *)selection
   }
 }
 
+- (void)setTextContentType:(NSString*)type
+{
+  #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0
+    if (@available(iOS 10.0, *)) {
+        self.backedTextInputView.textContentType = [type isEqualToString:@"none"] ? @"" : type;
```
janicduplessis: According to apple's doc the default value is nil, so I guess we should use that instead of an empty string.

Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -167,6 +167,15 @@ - (void)setSelection:(RCTTextSelection *)selection
   }
 }
 
+- (void)setTextContentType:(NSString*)type
```
janicduplessis: Nit: Space between NSString and *

Libraries/Text/TextInput/RCTBaseTextInputView.m
janicduplessis: Nevermind, just read the part about it in your test plan, maybe just add a comment to explain why we use the empty string.

cbrevik: It seems like for password inputs, it still defaults to `textContentType` equal to `password`, even if you haven't explicitly set that value. So setting an empty string will let you disable autofill for passwords. I can add a comment in the obj-c code :+1:

cbrevik:
Updated according to comments. 

Have not noticed any side effects from setting an invalid string. I was thinking that since the valid ones were defined in `propTypes`, the warning you receive was enough.

janicduplessis:
@facebook-github-bot shipit

Gregoirevda:
What's the correct way to get a callback when the form was autofilled to submit the form directly?

PR #18527, opened 2018-03-23
Add support for resource-id

### LINKED ISSUE #9777: Add support for resource-id
### Issue Description

There currently isn't a way to find elements by resource-id using appium/uiautomatorviewer.

`react-native` currently supports `contentDescription` via `accessibilityLabel`.  This is great for clickable elements, but not so great for testing other things like splash screen visibility.

I'd be more than willing to add support for this, but am currently not aware of how `react-native` builds the android views.  If someone could give me some guidance on how `android:id` could be added to a view then I'd most certainly be willing to give it a shot.  My understanding of `android:id` (per the [View docs on IDs](https://developer.android.com/reference/android/view/View.html)), is that it must be set in the XML prior to view creation.

I don't believe this is a duplicate of #7135, because I'm not proposing that we break the way testID currently works, more that I feel it would be nice if `testID` could add it's value to the corresponding view component in the XML.
### Steps to Reproduce / Code Snippets
- Add `testID="something"` to any component (e.g. `View`)
- Run the app on an android device
- Run `uiautomatorviewer`

You'll notice that `resource-id` is empty

without-id
#### Expected Results

`resource-id` has a value like `com.myApp:id/something`
### Additional Information
- React Native version: `0.32`
- Platform(s) (iOS, Android, or both?): Android.
- Operating System (macOS, Linux, or Windows?): All

lightboys22:
## Summary

Track react tag using android.view.View.tag.  Allow developers to set resource id by setting component.testID

Fixes #9777 

## Test Plan

First, you need to declare the resource id by creating /_your_android_studio_folder_/res/values/ids.xml

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <item type="id" name="test_id"/>
</resources>
```

Second.  Optionally, you can expose the resource id names so we can access them in our JSX code.  Or you can just hard code a list of resource id names.

```
public class ConstantsModule extends BaseJavaModule {
  @Override
  public String getName() {
    return REACT_NAME;
  }

  @Override
  public Map<String, Object> getConstants() {
    return // return the map with resource ids here.
  }
}
```

Third, set the testID of your component to the name of a resource id.

```
<Button
    onPress={onButtonPress}
    title="Press Me"
    testID='test_id'
    />
```

Forth, run the app and observe the resource id using UIAutomator viewer (/Android/sdk/tools/bin/uiautomatorviewer)

resource_id_snapshot

## Related PRs

This PR doesn't require a documentation change.

## Changelog

[ANDROID] [Fixes] - BaseViewManager: Allow developers to set resource id by setting component.testID


nes123:
any plans to continue with that?

erez-guesty:
hi 
whats the status of this feature?

ronilitman:
@lightboys22 any news on that?

icew1nd:
Is there any progress on this? Would help a lot with bot testing 🎉

lightboys22:
Sorry, I won’t be working on this. You can trying using the Espresso JavaScript framework to find view based on tags, (which is the testID）

LuisBonsembiante:
Waht happen with this PR?... why everithing is complicated in RN?

lightboys22:
Are you trying to write an unit test? If so, you can try using Expresso

LuisBonsembiante:
@lightboys22  ...It is for robo test... so I need a resource-id....

PR #18540, opened 2018-03-24
Prevents android crash due to unsupported ellipsize mode

### LINKED ISSUE #18474: Android: Text ellipsizeMode="clip" crashes on Android

## Environment
Environment:
  OS:  macOS High Sierra 10.13.3
  Node:  9.0.0
  Yarn:  1.3.2
  npm:  5.5.1
  Watchman:  4.7.0
  Xcode:  Xcode 9.2 Build version 9C40b
  Android Studio:  3.0 AI-171.4408382

  react: 16.2.0 => 16.2.0
  react-native: ^0.54.2 => 0.54.2

## Steps to Reproduce
* Use a [`<Text />`](https://facebook.github.io/react-native/docs/text.html) component
* Set ellipsizeMode to "clip"

## Expected Behavior
While the documentation clearly states `clip is working only for iOS` I assumed this will have no effect on Android.

## Actual Behavior
Causes an exception manifesting as a crash inside the app:
`RuntimeException
com.facebook.react.bridge.JSApplicationIllegalArgumentException: Error while updating property 'ellipsizeMode' of a view managed by: RCTText`
```
com.facebook.react.bridge.ReactContext in handleException at line 313
com.facebook.react.uimanager.GuardedFrameCallback in doFrame at line 33
com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher in doFrame at line 136
com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1 in doFrame at line 107
android.view.Choreographer$CallbackRecord in run at line 909
android.view.Choreographer in doCallbacks at line 723
android.view.Choreographer in doFrame at line 655
android.view.Choreographer$FrameDisplayEventReceiver in run at line 897
android.os.Handler in handleCallback at line 790
android.os.Handler in dispatchMessage at line 99
android.os.Looper in loop at line 164
android.app.ActivityThread in main at line 6494
java.lang.reflect.Method in invoke
com.android.internal.os.RuntimeInit$MethodAndArgsCaller in run at line 438
com.android.internal.os.ZygoteInit in main at line 807
```

t4deu:
Fixes #18474

This allows use clip as ellipsize mode for truncated text on android

## Test Plan

Added a test to RNTester, so it can be tested from there:
1. Run RNTester project
2. Navigate to `<Text>` tests
3. Scroll down to "Ellipsize mode" examples
4. Check the default behavior being applied when the value is set to "clip"

## Release Notes

[ANDROID] [BUGFIX] [Text] - Prevents android crash due to unsupported "clip" ellipsize mode

### DIFF
--- a/RNTester/js/TextExample.android.js
+++ b/RNTester/js/TextExample.android.js
@@ -453,6 +453,9 @@ class TextExample extends React.Component<{}> {
           <Text ellipsizeMode="head" numberOfLines={1}>
             This very long text should be truncated with dots in the beginning.
           </Text>
+          <Text ellipsizeMode="clip" numberOfLines={1}>
+            This very long text should be clipped and this will not be visible.
+          </Text>
         </RNTesterBlock>
         <RNTesterBlock title="Include Font Padding">
           <View style={{flexDirection: 'row', justifyContent: 'space-around', marginBottom: 10}}>
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextAnchorViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextAnchorViewManager.java
@@ -52,6 +52,8 @@ public abstract class ReactTextAnchorViewManager<T extends View, C extends React
       view.setEllipsizeLocation(TextUtils.TruncateAt.START);
     } else if (ellipsizeMode.equals("middle")) {
       view.setEllipsizeLocation(TextUtils.TruncateAt.MIDDLE);
+    } else if (ellipsizeMode.equals("clip")) {
+      view.setEllipsizeLocation(null);
     } else {
       throw new JSApplicationIllegalArgumentException("Invalid ellipsizeMode: " + ellipsizeMode);
     }


janicduplessis:
I'm not sure about this, maybe it would be better to redscreen with a good error message instead?

t4deu:
Hi @janicduplessis, thanks for the review!

The error message is the current behavior as described in the issue, the current message seems clear to me `JSApplicationIllegalArgumentException`, if it's the desired result we could just close the issue, what you think?

t4deu:
@borisyankov any thoughts?

janicduplessis:
@t4deu I had a quick look and it might actually be possible to support "clip" mode on Android. I think if we just pass null to setEllipsizeLocation it will work.

t4deu:
Yes @janicduplessis, it really works :v:!!

I will update the PR with the change and also the docs.

Thanks again!

Here a print with the result:
print

t4deu:
Done!

mdvacca:
Thanks @t4deu and @janicduplessis

Issue #14586, opened 2017-06-18
React Native WebView onMessage doesn't work on Android


fernandofranca:
### Description

WebView's onMessage function never gets called on some Android versions. 

This behaviour is not consistent on the Expo simulator. Test on physical devices.

### Reproduction Steps
Se line 40 on the following snack. 

If you test on a physical device and call `window.postMessage` without the second parameter it will throw "Failed to execute 'postMessage' on 'Window': 2 arguments required, but only 1 present".

The second parameter should be `'*'`.

See postMessage api on Mozilla: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

`postMessage` seems to work only after the DOMContentLoaded event but not immediately. Expo simulator catches the two last events, but on a physical device I was able to trigger and receive messages before DOMContentLoaded. I think this should've been clarified on the docs.

#### Sample Code

https://snack.expo.io/HybU5stmW

### Solution

Update the docs. Where is stated:
"window.postMessage accepts one argument, data, which will be available on the event object, event.nativeEvent.data. data must be a string."

Should state:
window.postMessage requires two arguments, data, ...., and `targetOrigin` which should be `'*'`.

### Additional Information

* React Native version: react-native: 0.44.2
* Platform: Android 6.0 and 5.0 (didn't tested on previous versions)
* Development Operating System: macOS 10.12.4
* Build tools: Xcode 8.3.3, Android Studio 2.3

hjh5696:
@fernandofranca I am in the exact same situation.
You mean `window.postMessage (SOME_DATA, '*')` should be used instead of `window.postMessage (SOME_DATA)`. right?
But it does not work.. if I am doing something wrong, please let me know

fernandofranca:
@hjh5696 Yup. 

In my environment `window.postMessage(someString, '*');` works.  YMMV as it's not documented and the behaviour is not consistent.

I'd suggest you to wrap your call in a try/catch and check what kind of error is being thrown.

hjh5696:
@fernandofranca Thank you for reply. However, even if I use try/catch, how do I get an error message within the html code of webview? is it from Android studio?

fernandofranca:
@hjh5696 I didn't knew any other way and had to improvise. 

Eg.: Inside your webview, use the catch block to add a new DOM element containing the stringified error.
```
function printStuff(stuff){
  var el = document.createElement("pre");
  var str = JSON.stringify(stuff);
  el.innerHTML = str;
[...]

PR #18546, opened 2018-03-24
[WebView] [BUGFIX] onMessage should not break postMessage

chrmod:
`onMessage` handler, replaces original `window.postMessage` function, so messages being sent by the applications never arrive to their target destinations. Some application like `twitter.com` are simply broken in such environments.

This PR, changes the implementation of replaced `window.postMessage` to call the original function in order to prevent website breakage. 

Those issues are probably caused by described problem:
* https://github.com/facebook/react-native/issues/14586
* https://github.com/facebook/react-native/issues/10865
There are probably more reports like that.

I've tested changes on Android and I'm assuming the iOS version should work as well.

## Test Plan

Create a WebView component with `onMessage` handler and `url` set to `https://twitter.com/`. Log in to twitter.

With changes of this PR, twitter feed will load correctly. Without changes, no tweets will be loaded.

## Release Notes

[WebView] [BUGFIX] - onMessage should not break postMessage

### DIFF
--- a/React/Views/RCTWebView.m
+++ b/React/Views/RCTWebView.m
@@ -322,6 +322,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
         "}"
 
         "window.postMessage = function(data) {"
+          "window.originalPostMessage.apply(window, arguments);"
           "messageQueue.push(String(data));"
           "processQueue();"
         "};"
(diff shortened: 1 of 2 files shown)


chrmod:
Would it be possible to backport this bugfix to some older releases? I'm interested in 0.53. Not sure what's react-native policy for such changes.

PS. `ci/circleci: test_android` fails, but I don't this it is related to changes in of this PR.

andrew--r:
@shergin Could you please review and merge this PR? It has only two lines of code changed.

hramos:

RSNara:
@chrmod, I doubt that we'd be able to back-port this to v0.53.0. 😔

This change is reasonable, and I'd love to review and merge this, but we can't because we're trying to pull out WebViews into [react-native-community/react-native-webview](https://github.com/react-native-community/react-native-webview). Would you mind re-launching this PR there instead? 

Before we exported WebViews, we also launched the new `WKWebView` implementation of `iOS` WebViews, which suffer from this exact same problem. So you'll have to make a similar fix in `RNCWKWebView.m`.

fungilation:
I've written this issue: https://github.com/react-native-community/react-native-webview/issues/66 for this as a roundup of issues related to 2 way communitaction between RN<->WebView. Please discuss there.

Issue #9612, opened 2016-08-27
react.gradle file not bundling Js code


waqas19921:
i uncommented app.gradle code to following

```
/**
 * The react.gradle file registers a task for each build variant (e.g. bundleDebugJsAndAssets
 * and bundleReleaseJsAndAssets).
 * These basically call `react-native bundle` with the correct arguments during the Android build
 * cycle. By default, bundleDebugJsAndAssets is skipped, as in debug/dev mode we prefer to load the
 * bundle directly from the development server. Below you can see all the possible configurations
 * and their defaults. If you decide to add a configuration block, make sure to add it before the
 * `apply from: "../../node_modules/react-native/react.gradle"` line.
 */
  project.ext.react = [
    // the name of the generated asset file containing your JS bundle
    bundleAssetName: "index.android.bundle",

    // the entry file for bundle generation
    entryFile: "index.android.js",

    // whether to bundle JS and assets in debug mode
    bundleInDebug: true,

    // whether to bundle JS and assets in release mode
    bundleInRelease: true,

    // whether to bundle JS and assets in another build variant (if configured).
    // See http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Build-Variants
    // The configuration property can be in the following formats
    //         'bundleIn${productFlavor}${buildType}'
    //         'bundleIn${buildType}'
    // bundleInFreeDebug: true,
    // bundleInPaidRelease: true,
    // bundleInBeta: true,

    // the root of your project, i.e. where "package.json" lives
    root: "../../",

    // where to put the JS bundle asset in debug mode
    jsBundleDirDebug: "$buildDir/intermediates/assets/debug",

    // where to put the JS bundle asset in release mode
    jsBundleDirRelease: "$buildDir/intermediates/assets/release",

    // where to put drawable resources / React Native assets, e.g. the ones you use via
    // require('./image.png')), in debug mode
    resourcesDirDebug: "$buildDir/intermediates/res/merged/debug",

    // where to put drawable resources / React Native assets, e.g. the ones you use via
    // require('./image.png')), in release mode
    resourcesDirRelease: "$buildDir/intermediates/res/merged/release",

    // by default the gradle tasks are skipped if none of the JS files or assets change; this means
    // that we don't look at files in android/ or ios/ to determine whether the tasks are up to
    // date; if you have any other folders that you want to ignore for performance reasons (gradle
    // indexes the entire tree), add them here. Alternatively, if you have JS files in android/
[...]

PR #18557, opened 2018-03-26
Use afterEvaluate to make gradle bundle task work with configureondemand

janicduplessis:
The js bundle task does not run when `org.gradle.configureondemand` is set to true. This uses `afterEvaluate` instead of `gradle.projectsEvaluated` which is executed properly.

## Test Plan

Add `org.gradle.configureondemand=true`, run RNTester in release mode and make sure the bundle task is run.

## Release Notes

[ANDROID] [BUGFIX] [LOCATION] - Fix release bundle task when org.gradle.configureondemand=true

### DIFF
--- a/RNTester/android/app/gradle.properties
+++ b/RNTester/android/app/gradle.properties
@@ -1,4 +1,6 @@
 android.useDeprecatedNdk=true
+org.gradle.parallel=true
+org.gradle.configureondemand=true
 MYAPP_RELEASE_STORE_FILE=my-release-key.keystore
 MYAPP_RELEASE_KEY_ALIAS=my-key-alias
 MYAPP_RELEASE_STORE_PASSWORD=*****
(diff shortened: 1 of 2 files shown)


mdvacca:
@janicduplessis parallel and configureondemand are still incubating features in Gradle, is there a significant change in building time when using these features?

janicduplessis:
@mdvacca I haven't profiled it but I wanted to make this change mostly to avoid issues like https://github.com/facebook/react-native/issues/9612. I only added the config to RNTester (not the new project template) just to make sure we don't break it in the future.

According to https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:configuration_on_demand (paragraph 2 line 3) it might become the default mode eventually so I think it makes sense to support it.

For parallel I don't think it is an incubating feature, it won't really have an impact since RNTester doesn't have many subprojects but I think it can be useful for real apps that might depend on a lot of 3rd party dependencies. So I think adding it is good just to make sure we support it too.

PR #18587, opened 2018-03-27
Handle update of `secureTextEntry` of `TextInput` in iOS

### LINKED ISSUE #5859: [TextInput] Dynamic changing of secureTextEntry breaks font and doesn't change cursor position
I need to have show/hide password option for TextInput.
Dynamic changing of secureTextEntry breaks font and doesn't change cursor position. It probably happens because font for "*" is fixed and it is not changed back when securyTextEntry is false.

lucasbento:
This PR fixes #5859, which causes the `TextInput` to have a space appended to the end when you change the value of `secureTextEntry` prop in iOS:

Before | After
:--- | :---
before|after

I found a few workarounds online and the simplest one in my opinion is [this SO answer](https://stackoverflow.com/a/22537788/4252781).

I'm definitely open to hear more opinions on the implementation.

## Test Plan

My test plan is very basic, having a `<TextInput />` and dynamically switching `secureTextEntry` prop between `true`/`false` will show the bug. 

## Release Notes

[IOS] [BUGFIX] [Libraries/Text/TextInput/RCTBaseTextInputView.m] - Fix bug with space being appended to the end of input value when changing `secureTextEntry`.

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -412,10 +412,23 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
 - (void)didSetProps:(NSArray<NSString *> *)changedProps
 {
+  [self shouldHandleSecureTextEntryChange:changedProps];
+
   if ([changedProps containsObject:@"inputAccessoryViewID"] && self.inputAccessoryViewID) {
     [self setCustomInputAccessoryViewWithNativeID:self.inputAccessoryViewID];
   } else if (!self.inputAccessoryViewID) {
     [self setDefaultInputAccessoryView];
+  }
+}
+
+- (void)shouldHandleSecureTextEntryChange:(NSArray<NSString *> *)changedProps
+{
+  BOOL didSecureTextEntryUpdate = [changedProps containsObject: @"secureTextEntry"];
+  if (didSecureTextEntryUpdate) {
+    NSAttributedString *originalText = self.backedTextInputView.attributedText;
+    
+    self.backedTextInputView.attributedText = nil;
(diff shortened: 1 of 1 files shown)


lucasbento:
Hey @janicduplessis, thank you for reviewing.

Where exactly should I put it? Do I have to add it and listen to the prop changes somewhere?

lucasbento:
@janicduplessis: can you take a look? :)

lucasbento:
@janicduplessis: would you mind taking a look? :)

RWOverdijk:
@lucasbento I read somewhere that the people responsible are not in atm. Might be related? https://github.com/react-native-community/react-native-releases/issues/11#issuecomment-385354468

> Unfortunately, the Facebook team members in charge of this are currently on
a leave and I can't give you any estimate when this gets unblocked. We are
in touch with the team and will do our best to make it as fast as possible.

But yes, 👍 on this PR.

RWOverdijk:
A month later I'd like to bump this again. We're using a patch on react native, which is a bandaid we'd love to stop applying.

lucasbento:
@shergin: perhaps you can take a look at this?

RWOverdijk:
We're in July now. Anyone?

lucasbento:
@hramos: I'm waiting for a reply on this: https://github.com/facebook/react-native/pull/18587#issuecomment-378170642.

I just need to know where to put the code specified by @janicduplessis.

justinsoong:
👍 keen for this

hramos:
Looks like the review feedback hasn't been addressed yet. I know there's a question out to Janic on how the feedback should be implemented, but that's not for me to answer.

I'll add that this PR would greatly benefit from adding a test case.

ericlewis:
@hramos I opened https://github.com/facebook/react-native/pull/23524 which addresses Janic's feedback.

cpojer:
Closing in favor of the new PR.

PR #18603, opened 2018-03-28
ModalHostView's child view size error

yk3372:
Signed-off-by: yk3372 <yk3372@gmail.com>

## Motivation
In some Android device such as Samsung S8, the ModalHostView's child view can't fill the screen.
before:
before
The JS ModalHostView can't fill the bottom, the area is device's navigation bar.

In class ModalHostShadowNode, follow code calculate is error:
`Point modalSize = ModalHostHelper.getModalHostSize(getThemedContext());`
This way not care the device's navigation bar, so the height is smaller than real.
For Samsung S8, real height is 2220 and modalSize.y is 2076.

### Cause:
1. The inner class ModalHostView.DialogRootViewGroup size is right, but when onSizeChanged is called, the getChildCount is 0. So the updateNodeSize not called.

2. After called onSizeChanged, addView is called, then ModalHostShadowNode.addChildAt is called.
So the child view used a smaller size. And will not correct after added. (onSizeChanged has called)

3. The main reason is onSizeChanged and addView call order is not sure.

---

### Fix Way
1. I add the boolean variable `isSizeChangedNotEffectChildView`, used to record whether the onSizeChanged effect the child view or not.
(The variable `isSizeChangedNotEffectChildView` is true mean the method onSizeChanged is called but not affect the child view's size.)

2. if not and `isSizeChangedNotEffectChildView` is true, when add child view, I called the method updateFirstChildView to affect the child view. 

And this fix way, has no side effect.
The result is:
after

## Test Plan
Use ModalHostView component in JS file. And show it in Samsung S8. 
The result is similar to above two pic. We has test it in many devices has no problem.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/modal/ReactModalHostView.java
@@ -362,10 +362,7 @@ public class ReactModalHostView extends ViewGroup implements LifecycleEventListe
 
     @Override
     public void handleException(Throwable t) {
-      
-      
-      
-      .handleException(new RuntimeException(t));
+      getReactContext().handleException(new RuntimeException(t));
     }
 
     private ReactContext getReactContext() {


yk3372:
How to add Release Notes?

yk3372:
see: #18615

sophister:
works +1

Issue #18416, opened 2018-03-16
New in RN 0.54, Debugging on Android no longer resolve source maps 


aarondail:
Prior to Android 0.54 (at least as of 0.52.2) when I run my app on iOS or Android and turn on JS Debugging, then in Chrome I can find, see, and debug my source files (rather than the bundled file).

As of 0.54, this still seems to work on iOS, but no longer works on Android.

### Environment

```
Environment:
  OS: macOS Sierra 10.12.6
  Node: 9.8.0
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.5.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 2.3 AI-162.3934792

Packages: (wanted => installed)
  react: ^16.3.0-alpha.1 => 16.3.0-alpha.2
  react-native: 0.54.2 => 0.54.2
```

### Expected Behavior

When I run a RN app on Android, and launch the Chrome debugger I can see my source files, and debug them. 

### Actual Behavior

I can only see and debug the bundled file.

### Steps to Reproduce

Create a sample app using `react-native init` and run it on Android, and start JS Debugging.  Note that you can't find `App.js` via `CMD+P` and on the console lists all statements as coming from some `blob:http....` path.  E.g.
![android_source_maps_broken](https://user-images.githubusercontent.com/3768353/37542044-05819804-291a-11e8-906d-cf5758feb574.png)

*NOTE* you can workaround this by going to the dev settings (via CMD+M) and turning OFF `Use JS Deltas`

suhshinhwang:
I also had to reset the packager and reload in Android.

brookhong:
For those who need to work around this issue, https://stackoverflow.com/questions/49340074/chrome-debug-tools-for-react-native-mutiple-issues/49676564

Issue #17970, opened 2018-02-13
Debugging: Dev Tools launch to 10.0.2.2 instead of localhost


foggy1:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

yes

### Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 8.7.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.7.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.53.0 => 0.53.0

### Steps to Reproduce

1. Run app on Android.
2. Turn on Debug mode.
3. Bundler says "Launching Dev Tools..." and it opens chrome to http://10.0.2.2:8081/debugger-ui, which hangs because the debugger is on localhost.

### Expected Behavior

I expect the debugger to open Chrome to localhost:8081/debugger-ui as it always has before I upgraded to version 0.53.0.

### Actual Behavior

As mentioned, it opens to http://10.0.2.2:8081/debugger-ui, which is inoperable. Nothing ever loads. The Devtools middleware has the wrong host for some reason.

### Reproducible Demo

You can literally `react-native init` a new app right now, and it will do this. Here's a package json of an app I just generated that has this issue

```
{
  "name": "testingtesting",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "jest"
  },
  "dependencies": {
    "react": "16.2.0",
    "react-native": "0.53.0"
  },
  "devDependencies": {
    "babel-jest": "22.2.2",
    "babel-preset-react-native": "4.0.0",
    "jest": "22.3.0",
    "react-test-renderer": "16.2.0"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

It appears the app bundle is also being loaded from 10.0.2.2, but the debugger is not functioning at this host, which I'm guessing means that even if I navigate to a "working" debugger @ localhost, it's still not going to work. Credit to #17910 for first assuring me I was not crazy and this was a real issue.

Edit: Complete shot in the dark here, but could it have something to do with fa574c6 ? (happy to pr this with some direction, although I know at least part of the issue is in a separate file where it is also incorrect about the host when launching chrome).

[...]

PR #18607, opened 2018-03-28
Turn off JS deltas by default

dannycochran:
Use JS deltas prevents resolving source maps when using remote debugging:

https://github.com/facebook/react-native/issues/18416

This doesn't fix the underlying issue but I don't think a feature that describes itself as experimental should be enabled by default.

## Test Plan

Attempt to do remote debugging on an Android device in 0.54.2

## Related PRs

None.

## Release Notes

[ANDROID] [BUGFIX] [Preferences] - Disable using JS deltas by default.

### DIFF
--- a/ReactAndroid/src/main/res/devsupport/xml/preferences.xml
+++ b/ReactAndroid/src/main/res/devsupport/xml/preferences.xml
@@ -23,7 +23,7 @@
       android:key="js_bundle_deltas"
       android:title="Use JS Deltas"
       android:summary="Request delta bundles from metro to get faster reloads (Experimental)"
-      android:defaultValue="true"
+      android:defaultValue="false"
       />
     <CheckBoxPreference
         android:key="animations_debug"


dannycochran:
@rafeca who authored the original PR, and @davidaurelio  who approved.

angly-cat:
@dannycochran btw, if this would be disabled by default, how would one enable it?

dannycochran:
@angly-cat you can turn it on by bringing up the developer options menu.

It probably should be configurable from the command line as well (and maybe already is?).

davidaurelio:
We just never removed the “experimental” label. It is actually stable. 

Instead of disabling it completely, we should either fix it for debugging, or disable it for debugging.

cc @rafeca

rafeca:
@dannycochran the actual issue with remote debugging comes from https://github.com/facebook/react-native/issues/17970

Disabling JS deltas by default won't help much in #18416: in this specific case what is allowing you workaround the issue is the fact that you're switching from one mode to the other (restarting the simulator or restarting Metro Bundler will have the same effect)

cpojer:
Closing this PR based on @davidaurelio's and @rafeca's comments.

PR #18613, opened 2018-03-28
Always write the manifest in multiRemove

### LINKED ISSUE #9196: AsyncStorage.getAllKeys() can sometimes include previously removed keys
Found on iOS
1. Add an item to async storage with a value larger than 1024:
   `AsyncStorage.setItem('@myapp:mykey',someValueLongerThan1024);`
2. Later on, remove that item:
   `AsyncStorage.removeItem('@myapp:mykey');`
3. Restart your iOS app and get your list of keys:
   `AsyncStorage.getAllKeys();`

The key that was removed will still be returned by getAllKeys.
This is because of logic in multiRemove:callback of RCTAsyncLocalStorage.m at line 397.
When a key with a large (stored in a separate file) value is removed, manifest.json is not rewritten.
The next time the module is loaded from scratch and _ensureSetup is called, it loads the incorrect key from manifest.json.

I would expect `AsyncStorage.getAllKeys();` to always return an accurate set of valid keys

esprehn:
RCTAsyncLocalStorage did not write the manifest after removing a value that was larger than RCTInlineValueThreshold. This meant that the values would still be on disk as null in the manifest.json, and if the app didn't do anything to make the manifest get written out again the null values would persist in the AsyncStorage and be returned by getAllKeys. We need to always write out the manifest.json even if the value is in the overflow storage files.

Fixes #9196.

## Test Plan

Not sure where the tests are for this?

## Related PRs

none.

## Release Notes

[IOS] [BUGFIX] [AsyncStorage] - Correctly remove keys of large values from AsyncStorage.

@tadeuzagallo @nicklockwood @dannycochran

### DIFF
--- a/React/Modules/RCTAsyncLocalStorage.m
+++ b/React/Modules/RCTAsyncLocalStorage.m
@@ -423,10 +423,8 @@ RCT_EXPORT_METHOD(multiRemove:(NSArray<NSString *> *)keys
         NSString *filePath = [self _filePathForKey:key];
         [[NSFileManager defaultManager] removeItemAtPath:filePath error:nil];
         [RCTGetCache() removeObjectForKey:key];
-        // remove the key from manifest, but no need to mark as changed just for
-        // this, as the cost of checking again next time is negligible.
-        [_manifest removeObjectForKey:key];
-      } else if (_manifest[key]) {
+      }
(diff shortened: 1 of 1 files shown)


esprehn:
Ping?

PR #18619, opened 2018-03-29
VirtualizedSectionList: Prevent out of bounds array access

rbrosboel:
## Motivation

Using mobx arrays with SectionList will trigger out-of-bounds warnings.

This commit checks bounds before access.

## Test Plan

Verified all items are rendered, without triggering mobx warnings.

## Release Notes
[GENERAL] [BUGFIX] [VirtulizedSectionList] - Prevent out of bounds array access


cpojer:
It seems like this should be solved on a higher level, like the component that passes data into this component, and not by this component itself. If you are using array-like objects as arrays, all sorts of things may be weird.

PR #18648, opened 2018-04-02
added in snapAlignment for horizontal android scrollView

philvasseur:
`snapToAlignment` is available on iOS but not android yet. This PR is to add support for `snapToAlignment` on android as `snapToInterval` was recently added to android and they are very useful together.

## Test Plan

Make a `Flatlist` in android with `pagingEnabled`, `horizontal`, `snapToInterval` and `snapToAlignment` set and see how adjusting between the three values of `snapToAlignment` aligns just like it does in iOS.
## Release Notes

[ANDROID] [MINOR] [ScrollView] - On Android, **ScrollView** now takes snapToAlignment like iOS

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
@@ -97,6 +97,9 @@ public class ReactHorizontalScrollView extends HorizontalScrollView implements
 
   public void setSnapInterval(int snapInterval) {
     mSnapInterval = snapInterval;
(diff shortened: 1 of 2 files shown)


mdvacca:
Thank you for working on this @philvasseur,

Could you also update the documentation https://facebook.github.io/react-native/docs/scrollview.html#snaptoalignment for Android? Thanks

philvasseur:
@mdvacca made a small change so pagingEnabled doesn't need to be turned on for snapToInterval to override it (as it functions in iOS), and have updated the docs and submitted a [PR here](https://github.com/facebook/react-native-website/pull/279). 

I also don't *think* that failed test has to do with my change(s), seems semi obscure from what I was touching.

PR #18651, opened 2018-04-02
Handle layout updates during LayoutAnimation animations on Android

lnikkila:
On Android, LayoutAnimation directly updates the layout since a generic
scaling animation is more difficult to implement. This causes a problem
if the layout is updated during an animation, as the previous layout is
stored with the animation and is not updated. As a result the view gets
the old layout instead once the animation completes.

This commit fixes this issue by storing the layout handling animations
while those animations are active, and updating the animations on the
fly if one of the views receives a new layout. The resulting behaviour
mirrors what iOS currently does.

This bug has real world consequences, for example if a LayoutAnimation
happens right after a VirtualizedList has mounted, it’s possible that
some list rows are mounted while the animation is active, making the
list content view bigger. If the content view is being animated, the
new size will not take effect and it becomes impossible to scroll to
the end of the list.

## Test Plan

I wrote a minimal test case to verify the bug, which I’ve also added to
RNTester. You can find the standalone app here:

<https://gist.github.com/lnikkila/18096c15b2fb99b232795ef59f8fb0cd>

The app creates a 100x300 view that gets animated to 200x300 using
LayoutAnimation. In the middle of that animation, the view’s dimensions
are updated to 300x300.

The expected result (which is currently exhibited by iOS) is that the
view’s dimensions after the animation would be 300x300. On Android the
view keeps the 200x300 dimensions since the animation overrides the
layout update.

The test app could probably be turned into an integration test by
measuring the view through UIManager after the animation, however I
don’t have time to do that right now...

Here are some GIFs to compare, click to expand:

<details>
  <summary><b>Current master (iOS vs Android)</b></summary>
  <p></p>
  
</details><p></p>

<details>
  <summary><b>With this patch (iOS vs Android, fixed)</b></summary>
  <p></p>
  
</details><p></p>

## Related PRs

No documentation changes needed.

## Release Notes

[ANDROID] [BUGFIX] [LayoutAnimation] - View layout is updated correctly during an ongoing LayoutAnimation, mirroring iOS behaviour.


lnikkila:
Android test failures on CircleCI seem to be coming from SoLoader, I don’t think this affects it.

lnikkila:
@janicduplessis 

Hi! I noticed you requested a review some time ago, would you still be interested in reviewing this?

janicduplessis:
@facebook-github-bot shipit

PR #18703, opened 2018-04-04
TextInput - Add hideKeyboardOnSubmit prop

renatoagds:
Today, when we press `Enter` after write in a input, the keyboard is being hide by default, this is a normal approach, but here at @loggi we come into a case that we want to keep the keyboard open when the user press enter in a input.

## Test Plan
To test it, you pass the prop as `false` for `TextInput`, type something and press `enter`, the keyboard should stay open.
* I added a unit test to check if prop is being propagate for the component, but looks like unit tests are disable for **textinput**.

## Release Notes

[GENERAL] [ENHANCEMENT] [TextInput] - Adding a prop called `hideKeyboardOnSubmit`, it controls if the keyboard will be hidden or not when press `enter/return`.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -82,6 +82,7 @@ public class ReactEditText extends EditText {
   private final InternalKeyListener mKeyListener;
   private boolean mDetectScrollMovement = false;
   private float mLetterSpacingPt = 0;
+  private boolean mHideKeyboardOnSubmit = true;
 
   private ReactViewBackgroundManager mReactBackgroundManager;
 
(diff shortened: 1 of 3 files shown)


janicduplessis:
Hello @renatoagds, have you tried using http://facebook.github.io/react-native/docs/textinput.html#bluronsubmit? I think it is supposed to do what you want to do with this PR, maybe it is broken on android?

renatoagds:
Hey @janicduplessis. I'm using `blurOnSubmit` as false, it keeps the input focused, but it didn't avoid the keyboard to be hide on **submit/enter**, I don't know if it's the expected, but looking at the source code, the keyboard is being forced hide when the `keyevent` is `ENTER`:

https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L158-L161

janicduplessis:
From what I understand the logic should be that blurOnSubmit is true by default for single line and false for multiline, but setting the blurOnSubmit prop should override that. We only hide the keyboard if blurOnSubmit is true.

From the iOS implementation:
image

https://github.com/facebook/react-native/blob/38197c8230657d567170cdaf8ff4bbb4aee732b8/Libraries/Text/TextInput/RCTBaseTextInputView.m#L204

renatoagds:
@janicduplessis Makes sense, but It's not the behaviour in Android, did you test this behaviour in Android devices?  In my tests, even when I use `blurOnSubmit` as `false`, the keyboard is hidden when I press `enter`.

PR #18710, opened 2018-04-05
"Fix" `Print: Entry, ":CFBundleIdentifier", Does Not Exist` error

angly-cat:
Hi. If a build triggered by `run-ios` command fails, the execution continues and `Print: Entry, ":CFBundleIdentifier", Does Not Exist` error is shown. Often that's the last thing people see in console output and they decide that's an issue they're having, when in reality they have build errors.

Take a look at this sample output (don't mind failed build command, that's just me messing with some random files to make builds to fail):

```
** BUILD FAILED **


The following build commands failed:
	CompileC /Users/angly/Work/test/ios/build/Build/Intermediates.noindex/React.build/Debug-iphonesimulator/yoga.build/Objects-normal/x86_64/YGNodePrint.o /Users/angly/Work/test/node_modules/react-native/ReactCommon/yoga/yoga/YGNodePrint.cpp normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)

Installing build/Build/Products/Debug-iphonesimulator/test.app
An error was encountered processing the command (domain=NSPOSIXErrorDomain, code=2):
Failed to install the requested application
An application bundle was not found at the provided path.
Provide a valid path to the desired application bundle.
Print: Entry, ":CFBundleIdentifier", Does Not Exist

Command failed: /usr/libexec/PlistBuddy -c Print:CFBundleIdentifier build/Build/Products/Debug-iphonesimulator/test.app/Info.plist
Print: Entry, ":CFBundleIdentifier", Does Not Exist
```

The part of script that produces output that starts with `Installing` shouldn't be executed if a build is failed.

I've checked `local-cli/runIOS/runIOS.js` file and discovered that `buildProcess.error` is never truthy and there isn't really any checks if a build is failed. This PR changes that. And make process to exit with non-zero exit code on build fails.

## Test Plan

Create a blank project. Run `react-native run-ios` in it's folder. See that all is ok (that this PR doesn't break successful builds). Mess with some random files to get a build to fail, run `react-native run-ios` again. See that on build fails the script doesn't try to install non-existent app and therefore doesn't produce `Print: Entry, ":CFBundleIdentifier", Does Not Exist` error.

## Release Notes

[CLI][BUGFIX][local-cli/runIOS/runIOS.js] - Stop script execution on build fails

### DIFF
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -227,10 +227,8 @@ function buildProject(xcodeProject, udid, scheme, configuration = 'Debug', launc
       }
       //FULL_PRODUCT_NAME is the actual file name of the app, which actually comes from the Product Name in the build config, which does not necessary match a scheme name,  example output line: export FULL_PRODUCT_NAME="Super App Dev.app"
       let productNameMatch = /export FULL_PRODUCT_NAME="?(.+).app"?$/m.exec(buildOutput);
-      if (productNameMatch && productNameMatch.length && productNameMatch.length > 1) {
-        return resolve(productNameMatch[1]);//0 is the full match, 1 is the app name
-      }
-      return buildProcess.error ? reject(buildProcess.error) : resolve();
+      //[0] of productNameMatch is the full match, [1] is the app name
+      return resolve(productNameMatch && productNameMatch.length && productNameMatch.length > 1 ? productNameMatch[1] : undefined);
     });
   });
 }
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -99,7 +99,8 @@ function runOnDeviceByUdid(args, scheme, xcodeProject, devices) {
 }
 
 function runOnSimulator(xcodeProject, args, scheme) {
-  return new Promise((resolve) => {
+  return Promise.resolve()
+  .then(() => {
     try {
       var simulators = JSON.parse(
       child_process.execFileSync('xcrun', ['simctl', 'list', '--json', 'devices'], {encoding: 'utf8'})
@@ -142,8 +143,8 @@ function runOnSimulator(xcodeProject, args, scheme) {
       }
     }
 
-    buildProject(xcodeProject, selectedSimulator.udid, scheme, args.configuration, args.packager, args.verbose, args.port)
-      .then((appName) => resolve({ udid: selectedSimulator.udid, appName }));
+    return buildProject(xcodeProject, selectedSimulator.udid, scheme, args.configuration, args.packager, args.verbose, args.port)
+      .then((appName) => ({ udid: selectedSimulator.udid, appName }));
   })
   .then(({udid, appName}) => {
     if (!appName) {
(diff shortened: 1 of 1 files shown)


angly-cat:
@grabbou, since you've been dealing with all kind of `runIOS` issues recently, may I bring your attention to this?

local-cli/runIOS/runIOS.js
```diff
@@ -198,19 +198,22 @@ function buildProject(xcodeProject, udid, scheme, configuration = 'Debug', launc
...
+        process.exit(1);
+      }
       //FULL_PRODUCT_NAME is the actual file name of the app, which actually comes from the Product Name in the build config, which does not necessary match a scheme name,  example output line: export FULL_PRODUCT_NAME="Super App Dev.app"
       let productNameMatch = /export FULL_PRODUCT_NAME="?(.+).app"?$/m.exec(buildOutput);
-      if (productNameMatch && productNameMatch.length && productNameMatch.length > 1) {
-        return resolve(productNameMatch[1]);//0 is the full match, 1 is the app name
-      }
-      return buildProcess.error ? reject(buildProcess.error) : resolve();
+      //[0] of productNameMatch is the full match, [1] is the app name
+      resolve(productNameMatch && productNameMatch.length && productNameMatch.length > 1 && productNameMatch[1] || undefined);
```
grabbou: I think I'd prefer a ternary expression here to make it easier to understand.

grabbou:
Thank you! That looks really good to me. Left one styling request/comment, rather than that, ready to go!

local-cli/runIOS/runIOS.js
angly-cat: Changed.

angly-cat:
@grabbou Fixed and rebased.

local-cli/runIOS/runIOS.js
```diff
@@ -213,19 +213,22 @@ function buildProject(xcodeProject, udid, scheme, configuration = 'Debug', launc
...
     buildProcess.stderr.on('data', function(data) {
-      console.error(data.toString());
+      error = data.toString();
+      console.error(error);
     });
     buildProcess.on('close', function(code) {
       if (xcpretty) {
         xcpretty.stdin.end();
       }
+      if (error) {
```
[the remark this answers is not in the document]
angly-cat: > Out of curiosity, have you tried playing around with `code`? I missed it for the first time.

I missed it too, actually. But now when you mentioned it, I'll go and check if it contains zero exit code for successful builds and non-zero code for failed builds. If it does, it would be more clean to use it.

> Also, since we don't really do anything than just `if(error)`, maybe instead of doing `error = data.toString()` we simply store a variable `let hasError` or something.

I agree. If `code` argument contains what I suspect it to contain, then this additional variable may not even be needed. Otherwise, I'll change this code according to your suggestion.

grabbou:
Thanks! Sorry for not providing this feedback for the first time (I also
hate that back and forth stuff). Anyways, looking forward for your updates
:)

angly-cat:
@grabbou Your hunch was right. The `code` arguments indeed is an exit code for building process, `0` on success builds and non-zero on failed builds, so I changed this PR accordingly. It's more simple now.

BTW the reason I call `process.exit` instead of `reject`ing a promise is that the script has no rejection handlers, however `process.exit` works just fine.

grabbou:
> the script has no rejection handlers

Interesting, that sounds to me like something that we should fix eventually? Here is the code that executes `runIOS` and all other commands. As you can see, it has a rejection handler: https://github.com/facebook/react-native/blob/master/local-cli/cliEntry.js#L110-L115

angly-cat:
> Here is the code that executes `runIOS` and all other commands. As you can see, it has a rejection handler

Indeed it has.

The reason I wrote that there aren't rejection handlers, because when I tried to reject that promise in `runIOS`, I got "unhandled promise rejection" `node` error and I hadn't second thought about it.

I'll take another look into this.

cpojer:

grabbou:
Really sorry that the review time on this PR took so long. I must have gotten busy with other things. Either way, we have ended up fixing this issue while working on the extraction of the CLI and in fact, refactored few other files too.

Feel free to look at it in your free time and implement any of the outstanding features or bugs. I promise help and immediate feedback time :)

PR #18711, opened 2018-04-05
Fix launching iOS simulator regression

angly-cat:
PR #17284 (accepted in 2ad34075f1d048bebb08ef30799ac0d081073150) introduced a couple of regressions.

~1. There's the code:~
```
.then((appName) => resolve(selectedSimulator.udid, appName));
/* ... */
.then((udid, appName) => {
```

~~This makes `appName` to be always `undefined` as per `resolve` accepts only 1 argument. This regression causes issues if an app name differs from a scheme name.~

~This PR fixes this by wrapping both values in an array.~

This was fixed in 589eae1432cc4bbc16221f841e27b038099fd128.

2. The code
```
child_process.execFileSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
```
makes a simulator *boot*, but the simulator *doesn't launch*. That's a regression, which forces developers to launch simulators by other means (by running a number of elaborate console commands, by running Xcode, or by running a simulator manually).

This PR reverts that part of changes.

## Test Plan

Create a blank project with a name that differs from scheme name. Try to `react-native run-ios` in it. See that a simulator is launched and installing succeeds. Without this changes simulator wouldn't launch, and installing step would fail because of app name mismatch.

## Related PRs

#17284

## Release Notes

[CLI][BUGFIX][local-cli/runIOS/runIOS.js] - Fix running on multiple simulators feature regressions

### DIFF
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -115,15 +115,12 @@ function runOnSimulator(xcodeProject, args, scheme) {
 
     if (!selectedSimulator.booted) {
       const simulatorFullName = formattedDeviceName(selectedSimulator);
-      console.log(`Booting ${simulatorFullName}...`);
+      console.log(`Launching ${simulatorFullName}...`);
       try {
-        child_process.execFileSync('xcrun', ['simctl', 'boot', selectedSimulator.udid]);
+        child_process.spawnSync('xcrun', ['instruments', '-w', selectedSimulator.udid]);
       } catch (e) {
-        throw new Error(
-`Could not boot ${args.simulator} simulator. Is there already a simulator running?
-Running multiple simulators is only supported from Xcode 9 and up.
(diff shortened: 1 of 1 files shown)


grabbou:
I wasn't able to get that `instruments` code to run with my XCode. Anyway, I made an alternative implementation (which, should be a bit better as it uses a tool designed for spawning instead of instruments).

Let me know what you think: #18721

grabbou:
I am on 9.2. Have you tested your PR with multiple simulators? The commit
you are referring to in the original description added the way to run
multiple simulators at once (--simulator iPhone X, --simulator iPhone 6). I
can't check right now if `instruments -w` will support that case.

grabbou:
@facebook-github-bot shipit

grabbou:
BTW, thanks for the test plan. I really appreciate that :)

Issue #18681, opened 2018-04-03
[0.55.0] Metro Bundler can't listen on port undefined


ASteinheiser:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)
- [x] I have searched [existing issues](https://github.com/facebook/react-native/issues)

I am following up on closed issue [#18673](https://github.com/facebook/react-native/issues/18673). This still seems to be an issue. Here's a screenshot of what I'm experiencing. Is there something I am missing?

![screen shot 2018-04-03 at 2 44 27 pm](https://user-images.githubusercontent.com/9949512/38277525-88b64012-374d-11e8-9f2e-f925c4fa2d8b.png)

## Environment

  OS: macOS High Sierra 10.13.3
  Node: 7.9.0
  Yarn: Not Found
  npm: 4.2.0
  Watchman: 4.9.0
  Xcode: Xcode 9.3 Build version 9E145
  Android Studio: 3.1 AI-173.4670197

Packages: (wanted => installed)
  react: 16.3.0 => 16.3.0
  react-native: 0.55.0 => 0.55.0

## Steps to Reproduce

```
react-native init AwesomeProject
cd AwesomeProject
react-native run-ios
```

## Expected Behavior

Build Successful!

## Actual Behavior

Metro Bundler can't listen on port undefined

ASteinheiser:
Can confirm that using react-native: 0.52.2 fixed this issue for me.

leofontes:
This is happening to me too, 0.52.2 with React 16.2.0 fixed it.

shivawu:
Apparently this is related to #18525, that issue fixes device but simulator is still broken, because `port` is not passed into `buildProject`. 

https://github.com/facebook/react-native/blob/master/local-cli/runIOS/runIOS.js#L130

Adding `, port` to this function call should temporarily fix the issue

kaushal9678:
after updating npm i also get  error when running command **react-native init-ios**. 
**React Native 0.55.0 CFBundleIdentifier error after fixing metro port**
But the projects that are created with lower version is works fine without any error.

I have following versions of NPM, react-native, node js
**react-native-cli: 2.0.1
react-native: 0.55.0,
 I downgrade NPM version to this npm 3.10.10
node -verison v6.14.1**

hatlestad:
I am experiencing the same issue with 0.55.0.

[...]

PR #18721, opened 2018-04-06
Fix `run-ios` not turning on Simulator app

grabbou:
Right now, `run-ios` will "boot" Simulator in the headless mode in the background, as long as the Simulator.app is not running. It is exactly what "detox" does - it executes your test suite in the background.

It seems to me that the author of this change was testing it with `Simulator.app` open.

In order to fix this behavior, we have to make sure it runs before we attempt booting.

This passed through the release process since we don't use `run-ios` there (it recommends turning on XCode and testing manually which is what I have done recently too).

CC: @kelset @koenpunt @hramos 

Related: #18681

### DIFF
--- a/local-cli/runIOS/runIOS.js
+++ b/local-cli/runIOS/runIOS.js
@@ -112,6 +112,24 @@ function runOnSimulator(xcodeProject, args, scheme) {
     if (!selectedSimulator) {
       throw new Error(`Could not find ${args.simulator} simulator`);
     }
+    
+    /**
+		 * Booting simulator through `xcrun simctl boot` will boot it in the `headless` mode
+		 * (running in the background).
+		 *
+		 * In order for user to see the app and the simulator itself, we have to make sure
+		 * that the Simulator.app is running.
+		 *
+		 * We also pass it `-CurrentDeviceUDID` so that when we launch it for the first time,
+		 * it will not boot the "default" device, but the one we set. If the app is already running,
+		 * this flag has no effect.
+		 */
+		child_process.execFileSync('open', [
+			'/Applications/Xcode.app/Contents/Developer/Applications/Simulator.app',
+			'--args',
+			'-CurrentDeviceUDID',
+			selectedSimulator.udid
+		]);
 
     if (!selectedSimulator.booted) {
       const simulatorFullName = formattedDeviceName(selectedSimulator);


angly-cat:
Why not just revert the changes that made simulators launch in headless mode?

Check #18711.

koenpunt:
I believe I tested it with simulator not running as well, but apparently there was an oversight. Anyway, nice there's a solution already.

local-cli/runIOS/runIOS.js
```diff
@@ -112,6 +112,24 @@ function runOnSimulator(xcodeProject, args, scheme) {
...
+		 * (running in the background).
+		 *
+		 * In order for user to see the app and the simulator itself, we have to make sure
+		 * that the Simulator.app is running.
+		 *
+		 * We also pass it `-CurrentDeviceUDID` so that when we launch it for the first time,
+		 * it will not boot the "default" device, but the one we set. If the app is already running,
+		 * this flag has no effect.
+		 */
+		child_process.execFileSync('open', [
```
koenpunt: I believe `open -a Simulator` should launch the simulator as well, without specifying a full path. Although not sure if you're then able to pass it additional flags

koenpunt: After further investigation, I think you should be able to use;

```
open -a Simulator --args -CurrentDeviceUDID ...
```

grabbou: Yeah, that is nice!

grabbou:
@angly-cat @koenpunt I am wondering whether we should stick to what we already have in master or keep using this approach.

Any ideas? This one uses sort of dedicated and less hacky API for launching the simulator (unfortunately, in headless mode so hack for Simulator app is still needed).

koenpunt:
I think if we use the `open -a Simulator --args -CurrentDeviceUDID ...` approach, the `xcrun simctl boot ...` is no longer necessary, since the simulator will be booted already, right?

grabbou:
Not really - the `open -a Simulator --args -CurrentDeviceUDID` will only
launch the given simulator first time. That means if Simulator app is
already running, `open` will be a noop action and it will just make the
Simulator app focus. That's why we need `xcrun simctl` to do the real work.

Since both commands seem to handle all states w/o issues, there are no if
statements around.

koenpunt:
Ah, I see, yeah that's true.

hramos:
Should I try to land this again?

hramos:
CLI has been moved.

PR #18722, opened 2018-04-06
Fix: "Metro listenting on port undefined"

grabbou:
Regression introduced in https://github.com/facebook/react-native/commit/2ad34075f1d048bebb08ef30799ac0d081073150.

Fixes https://github.com/facebook/react-native/issues/18681


grabbou:
Note: This is already part of `0.55.1` (the change was applied there as a "hot-fix"). No need to cherry-pick back.

Issue #18509, opened 2018-03-22
No way to access React instance of a pure C++ native module in order to communicate back to JS


tomduncalf:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [x] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

I am trying to create a native module to communicate both from JS->Native and Native->JS in pure C++, in order to improve portability and reduce boiler plate of native code in our application (which has a large C++ component with lots of React Native integration going on).

There's not much documentation out there about this so I have been figuring it out as I go, and have successfully managed to create a `CxxModule` instance, based on `SampleCxxModule`, which is registered via `registerExtraModules` in my custom bridge delegate and can receive method calls from Javascript.

However, I'd also like to be able to call Javascript methods from C++, for example to emit events. I was hoping I would be able to do so by getting the module from the bridge a bit like:

```
TestCxxModule* module = [bridge moduleForName:@"TestCxxModule"];
module->emitSomething();
```        

But there are a few issues I have found in doing so:

1. I can't see a way to get the actual C++ module from the RCTCxxModule wrapper, unless I keep my own reference to it (which is fine, just seems a bit odd)

2. Unless I have accessed the module via JS, or call e.g. `[module methodsToExport]`, the module has not been lazily initialised when I try to use it

and the main one:

3. The C++ module inside the RCTCxxModule that I get back from `[bridge moduleForName...]` has no React `_instance` set (it is `nullptr`), so I have no way to call back to JS (as it just crashes when trying to `callJSFunction`). If I put a breakpoint inside one of the methods exposed to JS, I can see when it is called from the Javascript context, `_instance` is populated as I'd expect, but not when I get the module from the bridge. 

For example, I'd like to be able to call this from anywhere in my app code (as long as I have a pointer to the module):

```
void TestCxxModule::emitEvent (dynamic eventData)
{
    auto reactInstance = this->getInstance().lock().get();
    reactInstance->callJSFunction("RCTDeviceEventEmitter", "emit", std::move(eventData));
}
```

[...]

PR #18728, opened 2018-04-06
Add requiresImmediateCxxNativeModuleSetup property to RCTCxxModule and immediately initialize CxxNativeModule if true.

tomduncalf:
This allows for CxxModules with methods which require a bridge instance (for example, emitting an event) to be called from native code, without requiring them to be initialised by some other means (e.g. being called from Javascript, or calling `getMethods`) first.

## Rationale

I have been experimenting with using pure C++ modules with React Native, as our application is made up of a C++ core with a React Native UI, and makes use of a lot of communication both ways across the bridge. Creating Objective C wrappers for every call has proven to be quite verbose, so native C++ modules would be quite a productivity and code complexity win.

Unfortunately there isn't a great deal of documentation around this topic, so I've been figuring it out as I go along.

I was able to create a CxxModule accessible from Javascript, but as described in https://github.com/facebook/react-native/issues/18509, I've been having issues getting a reference to the CxxModule from native code in order to emit events from C++ to Javascript.

The crux of the issue seems to be that the module instance returned from e.g. `moduleForName` is the `CxxModule` rather than `CxxNativeModule` instance, and so has no `_instance` set, and therefore cannot communicate back to React Native. In addition to this, the module is not initialised until you either call into it from Javascript or call e.g. `getMethods`, and accessing modules this way requires Objective C code to access the bridge and get the module instance so is not pure C++.

I've spent quite a lot of time looking at the React Native source code, attempting to determine why things work the way they do, and have considered several approaches to solving this problem (e.g. providing a way to get the `CxxNativeModule` from the module registry via the bridge), but after a lot of experimentation, I decided upon the approach in this PR. It is the smallest change I could think of to achieve the goals, and is low risk in the sense that it opt-in and does not modify the default behaviour of React Native.

## Example code

I've created a simple example of how this can be used at https://github.com/tomduncalf/react-native-cxx-module/tree/working - the relevant part is in https://github.com/tomduncalf/react-native-cxx-module/tree/working/ios/rnfresh/ReactNative, and the changes from the "not working" version can be seen at https://github.com/tomduncalf/react-native-cxx-module/commit/40235ce2e541508a09a3463133c5b7a9bae60f06

By specifying that a module requires immediate CxxNativeModule setup, we know that the constructor of the CxxModule will be called at app startup time, then we can store a reference to it somewhere shared (in this case, the `ModuleRegistry`, which could equally be a singleton), and we can then access a fully working instance of our module anywhere from our native codebase.

Note that this code is currently iOS only - any advice on the right approach for Android would be welcome, but I am currently not targetting Android so have no device to test on.

If this is not the right approach I'd be very happy to discuss further - I thought this was a good way to get the ball rolling, as there are no docs about this and it would be a massive help for our project!

## Test Plan

Tested in existing project and in the simple example above. There appear to be no automated tests around this area, but the change is low risk in that it does not change defaults.

## Related PRs

There is no related documentation, but I'd be happy to create some if we are happy with this approach!

## Release Notes

[INTERNAL] [ENHANCEMENT] [CxxModule] - Allow a RCTCxxModule to specify that it `requiresImmediateCxxNativeModuleSetup`, which will cause the module to be immediately initialised rather than lazily.

### DIFF
--- a/React/Base/RCTModuleData.h
+++ b/React/Base/RCTModuleData.h
@@ -60,6 +60,11 @@ typedef id<RCTBridgeModule>(^RCTBridgeModuleProvider)(void);
 @property (nonatomic, assign) BOOL requiresMainQueueSetup;
 
 /**
+ * Returns YES if Cxx native module instance must be created at startup.
+ */
+@property (nonatomic, assign) BOOL requiresImmediateCxxNativeModuleSetup;
+
+/**
  * Returns YES if module has constants to export.
  */
 @property (nonatomic, assign, readonly) BOOL hasConstantsToExport;
--- a/React/Base/RCTModuleData.mm
+++ b/React/Base/RCTModuleData.mm
@@ -67,6 +67,13 @@
         "on a background thread unless explicitly opted-out of.", _moduleClass, methodName);
     }
   }
+    
+  const BOOL implementsRequiresImmediateCxxNativeModuleSetup = [_moduleClass respondsToSelector:@selector(requiresImmediateCxxNativeModuleSetup)];
+  if (implementsRequiresImmediateCxxNativeModuleSetup) {
+    _requiresImmediateCxxNativeModuleSetup = [_moduleClass requiresImmediateCxxNativeModuleSetup];
+  } else {
+    _requiresImmediateCxxNativeModuleSetup = NO;
+  }
 }
 
 - (instancetype)initWithModuleClass:(Class)moduleClass
--- a/React/CxxModule/RCTCxxModule.mm
+++ b/React/CxxModule/RCTCxxModule.mm
@@ -31,6 +31,11 @@ using namespace facebook::react;
   return NO;
 }
 
++ (BOOL)requiresImmediateCxxNativeModuleSetup
+{
+  return NO;
+}
+
 - (void)lazyInit
 {
   if (!_module) {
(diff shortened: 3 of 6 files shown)


tomduncalf:
Sorry, will look into the test failure Monday

tomduncalf:
Hi @mhorowitz, you seem like the person who has worked most on this code – is there any chance you could take a look at this? Thanks!

oNaiPs:
@tomduncalf you might be able to get away without the changes if you reimplement `RCTCxxModule` outside of the react native sources

tomduncalf:
True. For now we are working around it by calling in from JavaScript at startup time, which works okay but feels messy!

matthargett:
why not rename lazyInit() and make it public instead?

tomduncalf:
Yeah that would probably be a valid approach too. Unfortunately it’s not been possible to get any RN contributors attention on this so as I mentioned above, our solution for now is to call into the module from JS at app start up time to make it initialise - would rather avoid forking/hacks if possible. 

I believe they are making large changes to some of the native interop stuff in React Native next year, so it might be that a different solution emerges then!

cpojer:
Thank you for this PR. I think this somewhat conflicts with our plans for TurboModules, which should support your use case in some form as well once it is ready to use. See https://github.com/react-native-community/discussions-and-proposals/issues/40 (It seems like you are already giving input there, but just making sure there is a link for other people who may see this PR).

tomduncalf:
Thanks Christoph - yeah, I’m already investigating JSI and turbo modules and I think it will suit our use case perfectly so I’m happy for this to be closed.

PR #18732, opened 2018-04-06
Fix TouchHistoryMath import

sjmueller:
The reference needed to be updated after the file was moved in this commit:
https://github.com/facebook/react-native/commit/06085d38366373f3135074dc14e2c9871ca4fe29

Otherwise, results in the packager failing with the following error:
> Unable to resolve module TouchHistoryMath from /node_modules/react-native/Libraries/Interaction/PanResponder.js: Module TouchHistoryMath does not exist in the Haste module map

### DIFF
--- a/Libraries/Interaction/PanResponder.js
+++ b/Libraries/Interaction/PanResponder.js
@@ -10,7 +10,7 @@
 'use strict';
 
 const InteractionManager = require('./InteractionManager');
-const TouchHistoryMath = require('TouchHistoryMath');
+const TouchHistoryMath = require('./TouchHistoryMath');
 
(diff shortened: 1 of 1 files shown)


janicduplessis:
Not sure if we should just add `@providesModule` to TouchHistoryMath instead, RN still uses haste module for pretty much everything.

cc @sebmarkbage

sebmarkbage:
Yea, providesModule would be better.

I think the plan is to eventually not have to specify it (which is probably why it works internally at FB).

gengjiawen:
Any plan to fix this, current RNTester won't compile because of this.

PR #18746, opened 2018-04-08
enhance RNTester android config

gengjiawen:
## Motivation

* cliPath is not config right
* root config can be better config instead of a relative path which can easily go wrong if location changed
* DeveloperSupport should only be in debug mode.

## Test Plan

* make https://github.com/facebook/react-native/pull/18732 change on local. 
* config signingConfig 
* execute `./gradlew :RNTester:android:app:assembleRelease`, and run app on device to check everything is fine.

## Related PRs

none

## Release Notes

[GENERAL][ENHANCEMENT][RNTester]

### DIFF
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -58,9 +58,10 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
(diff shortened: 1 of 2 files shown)


janicduplessis:
Looks good, thanks!

@facebook-github-bot shipit

PR #18782, opened 2018-04-10
Implement Image.queryCache on iOS

ladipro:
The API was available only on Android (with no mention to that effect in the docs, AFAICT).
This commit adds a simple iOS implementation based on NSURLCache. It should be possible to
query the decoded image cache as well to provide higher fidelity (i.e. "disk", "memory",
"decoded") if the caller passes size, scale, etc. in addition to the image URL, but it's
probably not worth the complexity. The assumption is that callers are interested in the
durability rather than performance aspect of the returned information.

## Test Plan

Tested with RNTester on iPhone emulator.

## Release Notes

[IOS] [ENHANCEMENT] [Image] - Implemented queryCache


ladipro:
(Working on getting added to my company's CLA list of designated employees.)

Libraries/Image/RCTImageLoader.m
```diff
@@ -779,6 +779,25 @@ - (RCTImageLoaderCancellationBlock)getImageSizeForURLRequest:(NSURLRequest *)ima
...
+  NSMutableDictionary *results = [NSMutableDictionary dictionary];
+  for (id request in requests) {
+    NSURLRequest *urlRequest = [RCTConvert NSURLRequest:request];
+    if (urlRequest) {
+      NSCachedURLResponse *cachedResponse = [NSURLCache.sharedURLCache cachedResponseForRequest:urlRequest];
+      if (cachedResponse) {
+        if (cachedResponse.storagePolicy == NSURLCacheStorageAllowedInMemoryOnly) {
+          [results setObject:@"memory" forKey:urlRequest.URL.absoluteString];
+        } else {
+          [results setObject:@"disk" forKey:urlRequest.URL.absoluteString];
```
rigdern: >[results setObject:@"disk" forKey:urlRequest.URL.absoluteString]; [](start = 10, length = 65)

Do we want this branch to run when `storagePolicy` is `NSURLCacheStorageNotAllowed`?

Libraries/Image/RCTImageLoader.h
```diff
@@ -129,6 +129,14 @@ typedef dispatch_block_t RCTImageLoaderCancellationBlock;
  */
 - (RCTImageLoaderCancellationBlock)getImageSizeForURLRequest:(NSURLRequest *)imageURLRequest
                                                        block:(void(^)(NSError *error, CGSize size))completionBlock;
+/**
+ * Determines whether given image URLs are cached locally. The `requests` array is expected
+ * to contain objects convertible to NSURLRequest. The return value maps URLs to strings:
```
rigdern: Does this match the kind of things that can appear in `requests` on Android?

Libraries/Image/RCTImageLoader.m
ladipro: My understanding is that responses obtained from the cache will never have such ```storagePolicy``` because they would not be inserted in the cache in the first place. I recognize that it would be safer to explicitly check for ```NSURLCacheStorageAllowed``` in the second if, though.

Libraries/Image/RCTImageLoader.h
rigdern: Okay, it's good that it's consistent with the existing cross-platform image APIs

ladipro:
Yes, this pull request is still relevant. Tagging @mhenaire as a super wild guess :)

ladipro:
@gaguirre rebased and fixed the fallout. The failing CI tests seem to be infra issues.

gaguirre:
@ladipro Great! Hope @shergin approves this PR 😃

PR #18812, opened 2018-04-11
Remove extra space in source_files

hsousa:
This was causing tvOS builds to fail because .m files weren’t being included.

## Test Plan

 - Add `tvOS` subspec to your project with a tvOS target.
 - Build
 - Profit.

## Release Notes

[TVOS] [BREAKING] [React.podspec] - Fix building for tvOS with CocoaPods


janicduplessis:
Thanks for the fix!

@facebook-github-bot shipit

PR #18814, opened 2018-04-11
Android ndk-build fix for ndk-r14+

dryganets:
According to [ndk release notes](https://github.com/android-ndk/ndk/wiki/Changelog-r14-beta1#ndk-build) flag APP_ALLOW_MISSING_DEPS should be set to true in order to fix ndk-build tool.

## Test Plan
If it builds it will work. On older ndk flag is ignored.

## Release Notes
[ANDROID] [MINOR] [Application.mk]  - build with android-ndk-14+ supported

### DIFF
--- a/ReactAndroid/src/main/jni/react/jni/Android.mk
+++ b/ReactAndroid/src/main/jni/react/jni/Android.mk
@@ -37,7 +37,7 @@ LOCAL_CFLAGS += $(CXX11_FLAGS)
 LOCAL_EXPORT_CPPFLAGS := $(CXX11_FLAGS)
 
 LOCAL_LDLIBS += -landroid
-LOCAL_SHARED_LIBRARIES := libfolly_json libfbjni libjsc libglog_init libyoga libprivatedata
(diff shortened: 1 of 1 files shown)


dryganets:
error An unexpected error occurred: "Couldn't find any versions for \"metro\" that matches \"^0.32.0\"".

No changes in package.json here obviously.

dryganets:
@hramos, should be trivial enough.
Unlike an intitial fix with adding the flag `ALLOW_MISSING_DEPENDENCIES=true` to Application.mk it fixes the rootcause.
Looks like there was a fbjni library before and somebody forgot to update it to fb during refactoring.
android-ndk-14+ don't allow missing dependencies by default. See: https://github.com/android-ndk/ndk/wiki/Changelog-r14-beta1#ndk-build

hramos:
Alright, I'm running tests on Circle CI for this PR now over at https://circleci.com/workflow-run/ecd42b0e-3940-4167-aef0-4aa9c85efbbd. If they pass, I'll land this.

Issue #11828, opened 2017-01-11
Removing items with LayoutAnimation causes adjacent views/texts to disappear on Android


vinceyuan:
### Description

When remove items with LayoutAnimation, the adjacent views/texts should not be impacted. But they disappear on Android.

In the following screenshot of the sample app, we can see some texts/buttons disappeared when we press `Remove with opacity animation` button and `Remove with scaleXY animation` button. I found this issue when I worked on #11769.

![Screenshot](https://github.com/vinceyuan/ReactNativeOpacityBugDemo/raw/master/ReactNativeOpacityBug3.gif)

### Reproduction

Reproduced on rnplay.org https://rnplay.org/apps/JbdOpQ
The sample project I created: https://github.com/vinceyuan/ReactNativeOpacityBugDemo

### Solution

N/A

Update on Jan 18, 2017:
I figured out a solution. See below.

### Additional Information

* React Native version: 0.33+
* Platform: Android
* Operating System: MacOS

vinceyuan:
### Cause

When LayoutAnimation is enabled, indices of to-be-removed items are incorrect (to reproduce it, we need to delete at least two items). So the adjacent items are removed. In https://github.com/facebook/react-native/blob/0.41-stable/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java#L364-L373 because `indexToRemove` is wrong, we get wrong `viewToRemove`. `arrayContains(tagsToDelete, viewToRemove.getId())` returns `false`. That view is removed immediately in`viewManager.removeViewAt(viewToManage, indexToRemove);`.

### Essential Cause
The essential cause is in [https://github.com/facebook/react-native/blob/0.41-stable/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java#L278-L285](https://github.com/facebook/react-native/blob/0.41-stable/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java#L278-L285). It gets the index of a `ReactShadowNode` and adds a managing children operation into a queue.

Says we want to delete 2 items `i0` and `i2`  from `[i0, i1, i2, i3]`. It creates two operations `op0` and `op1`. `op0` is created with index `0` for deleting `i0`, and `op1` is created with index `1` for `i2`.  `op1`'s index is `1`, instead of `2`, because `ReactShadowNode` of `i0` has been deleted.

[...]

PR #11962, opened 2017-01-18
Fix issue in removing items with LayoutAnimation on Android. When LayoutAnimation is enabled, indicesToRemove are incorrect, which makes adjacent views disappear.

vinceyuan:
Fix issue #11828 Removing items with LayoutAnimation causes adjacent views/texts to disappear on Android

Screenshot

I explained my solution in https://github.com/facebook/react-native/issues/11828#issuecomment-273434006

PR #18830, opened 2018-04-12
Fix view indices with Android LayoutAnimation

lnikkila:
Fixes issue #11828 that causes layout animations for removed views to
remove some adjacent views as well. This happens because the animated
views are still present in the ViewGroup, which throws off subsequent
operations that rely on view indices having updated.

This issue was addressed in #11962, which was closed in favour of a more
reliable solution that addresses the issue globally since it’s difficult
to account for animated views everywhere. @janicduplessis [recommended][0]
handling the issue through ViewManager.

Since API 11, Android provides `ViewGroup#startViewTransition(View)`
that can be used to keep child views visible even if they have been
removed from the group. ViewGroup keeps an array of these views, which
is only used for drawing. Methods such as `ViewGroup#getChildCount()`
and `ViewGroup#getChildAt(int)` will ignore them.

I believe relying on these framework methods within ViewManager is the
most reliable way to solve this issue because it also works if callers
ignore ViewManager and reach into the native view indices and counts
directly.

[0]: https://github.com/facebook/react-native/pull/11962#pullrequestreview-21862640

## Test Plan

I wrote a minimal test app that you can find here:

<https://gist.github.com/lnikkila/87f3825442a5773f17ead433a810d53f>

The expected result is that the red and green squares disappear, a blue
one appears, and the black one stays in place. iOS has this behaviour,
but Android removes the black square as well.

We can see the bug with some breakpoint logging.

Without LayoutAnimation:

```
    NativeViewHierarchyOptimizer: Removing node from parent with tag 2 at index 0
    NativeViewHierarchyOptimizer: Removing node from parent with tag 4 at index 1
    NativeViewHierarchyManager: Removing indices [0] with tags [2]
    RootViewManager: Removing child view at index 0 with tag 2
    NativeViewHierarchyManager: Removing indices [1] with tags [4]
    RootViewManager: Removing child view at index 1 with tag 4
```

With LayoutAnimation tag 3 gets removed when it shouldn’t be:

```
    NativeViewHierarchyOptimizer: Removing node from parent with tag 2 at index 0
    NativeViewHierarchyOptimizer: Removing node from parent with tag 4 at index 1
    NativeViewHierarchyManager: Removing indices [0] with tags [2]
    NativeViewHierarchyManager: Removing indices [1] with tags [4]
->  RootViewManager: Removing child view at index 1 with tag 3
    RootViewManager: Removing child view at index 2 with tag 4

    (Animation listener kicks in here)

    RootViewManager: Removing child view at index 1 with tag 2
```

Here are some GIFs to compare, click to expand:

<details>
  <summary><b>Current master (iOS vs Android)</b></summary>
  <p></p>
  
</details><p></p>

<details>
  <summary><b>With this patch (iOS vs Android, fixed)</b></summary>
  <p></p>
  
</details><p></p>

## Related PRs

Previously addressed in #11962, which wasn’t merged.

Tangentially related to my other LayoutAnimation PR #18651.

No documentation changes needed.

## Release Notes

[ANDROID] [BUGFIX] [LayoutAnimation] - Removal LayoutAnimations no longer remove adjacent views as well in certain cases.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -373,11 +373,12 @@ public class NativeViewHierarchyManager {
         if (mLayoutAnimationEnabled &&
             mLayoutAnimator.shouldAnimateLayout(viewToRemove) &&
             arrayContains(tagsToDelete, viewToRemove.getId())) {
-          // The view will be removed and dropped by the 'delete' layout animation
-          // instead, so do nothing
-        } else {
-          viewManager.removeViewAt(viewToManage, indexToRemove);
+          // Display the view in the parent after removal for the duration of the layout animation,
+          // but pretend that it doesn't exist when calling other ViewGroup methods.
+          viewManager.startViewTransition(viewToManage, viewToRemove);
         }
+
+        viewManager.removeViewAt(viewToManage, indexToRemove);
 
         lastIndexToRemove = indexToRemove;
       }
@@ -423,7 +424,9 @@ public class NativeViewHierarchyManager {
           mLayoutAnimator.deleteView(viewToDestroy, new LayoutAnimationListener() {
             @Override
             public void onAnimationEnd() {
-              viewManager.removeView(viewToManage, viewToDestroy);
+              // Already removed from the ViewGroup, we can just end the transition here to
+              // release the child.
+              viewManager.endViewTransition(viewToManage, viewToDestroy);
               dropView(viewToDestroy);
             }
           });
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
@@ -93,6 +93,14 @@ public abstract class ViewGroupManager <T extends ViewGroup>
     }
   }
 
+  public void startViewTransition(T parent, View view) {
+    parent.startViewTransition(view);
+  }
+
+  public void endViewTransition(T parent, View view) {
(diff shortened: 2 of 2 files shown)


janicduplessis:
It seems similar to the technique I wanted to use in #16091 and that iOS uses but using a platform implementation which is way better (since my implementation had a bug left that I didn't have time to investigate 😆).

lnikkila:
Thanks for taking a look! I’ll try it out with that gist first thing tomorrow.

lnikkila:
Tried it out, works fine! I managed to get current master to crash due to a wrong index by removing lots of squares rapidly and then adding some, but this branch was stable.

There’s still a bug with the view opacities not updating when interrupting an ongoing animation (compared to iOS where they get the final value) but it looks like a separate issue to me.

mdvacca:
@lnikkila I'm reverting this PR, as we found a java.lang.IndexOutOfBoundsException when removing views:

Exception in native call java.lang.IndexOutOfBoundsException
	at android.view.ViewGroup.removeViewsInternal(ViewGroup.java:4795)
	at android.view.ViewGroup.removeViewsInLayout(ViewGroup.java:4652)
	at com.facebook.react.views.view.ReactViewGroup.removeViewWithSubviewClippingEnabled(ReactViewGroup.java:530)
	at com.facebook.react.views.view.ReactViewManager.removeViewAt(ReactViewManager.java:285)
	at com.facebook.react.views.view.ReactViewManager.removeViewAt(ReactViewManager.java:36)
	at com.facebook.react.uimanager.NativeViewHierarchyManager.manageChildren(NativeViewHierarchyManager.java:381)

lnikkila:
Hmm could be due to `removeClippedSubviews`, I’ll take a closer look with that enabled and see if I can figure out why that’s happening.

PR #18850, opened 2018-04-13
Add Image.getSizeWithHeaders

se-bastiaan:
## Summary

This adds new functionality to the `Image` component by allowing you to retrieve the width and height of an image just like you'd do with [`Image.getSize`](https://facebook.github.io/react-native/docs/image.html#getsize) but _with_ the ability to provide headers to your request.

Why would you need this you ask? Well, imagine that you have an image that you're loading into your `Image` component that is protected and you get access by using a token in a header (or something similar). That would work. However, getting the dimensions isn't possible since you can't provide those same headers.
This is something that is bothering me when using a third-party library (https://github.com/archriss/react-native-image-gallery) and instead of implementing this just for that single library I imagined that it would be useful for anyone else that needs to get the image dimensions before displaying it.

## Changelog

[Android] [Added] - Added Image.getSizeWithHeaders
[iOS] [Added] - Added Image.getSizeWithHeaders

## Test Plan

I initialised a new RN project and changed it so that I could use React Native build from source by following the [guide](https://facebook.github.io/react-native/docs/building-from-source.html).
Then I used `Image.getSize` in that sample project to check if it loaded a sample image.
I applied my changes and tested if `Image.getSize` still worked. And last I checked if getting the dimensions using a protected image using `Image.getSizeWithHeaders` worked.

## Related PRs

https://github.com/facebook/react-native-website/pull/304

### DIFF
--- a/Libraries/Image/Image.ios.js
+++ b/Libraries/Image/Image.ios.js
@@ -42,6 +42,24 @@ function getSize(
   );
 }
 
+function getSizeWithHeaders(
+  uri: string,
+  headers: {[string]: string},
+  success: (width: number, height: number) => void,
+  failure?: (error: any) => void,
+) {
+  return ImageViewManager.getSizeWithHeaders({uri, headers})
+    .then(function(sizes) {
+      success(sizes.width, sizes.height);
+    })
+    .catch(
+      failure ||
+        function() {
+          console.warn('Failed to get size for image: ' + uri);
+        },
+    );
+}
+
 function prefetch(url: string) {
   return ImageViewManager.prefetchImage(url);
 }
@@ -56,6 +74,7 @@ declare class ImageComponentType extends ReactNative.NativeComponent<
   ImagePropsType,
 > {
   static getSize: typeof getSize;
+  static getSizeWithHeaders: typeof getSizeWithHeaders;
   static prefetch: typeof prefetch;
   static queryCache: typeof queryCache;
   static resolveAssetSource: typeof resolveAssetSource;
@@ -135,6 +154,17 @@ Image.displayName = 'Image';
  * error found when Flow v0.89 was deployed. To see the error, delete this
  * comment and run Flow. */
 Image.getSize = getSize;
+
+/**
+ * Retrieve the width and height (in pixels) of an image prior to displaying it
+ * with the ability to provide the headers for the request.
+ *
+ * See https://facebook.github.io/react-native/docs/image.html#getsizewithheaders
+ */
+/* $FlowFixMe(>=0.89.0 site=react_native_ios_fb) This comment suppresses an
+ * error found when Flow v0.89 was deployed. To see the error, delete this
+ * comment and run Flow. */
+Image.getSizeWithHeaders = getSizeWithHeaders;
 
 /**
  * Prefetches a remote image for later use by downloading it to the disk
--- a/Libraries/Image/RCTImageViewManager.m
+++ b/Libraries/Image/RCTImageViewManager.m
@@ -10,6 +10,7 @@
 #import <UIKit/UIKit.h>
 
 #import <React/RCTConvert.h>
+#import <React/RCTImageSource.h>
 
 #import "RCTImageLoader.h"
 #import "RCTImageShadowView.h"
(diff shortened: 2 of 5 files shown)


se-bastiaan:
I've rebased this, looks like the same tests are failing that fail on the `master` branch too. Fixing those tests seems a bit out-of-scope.

se-bastiaan:
@hramos Any idea if this can be merged with the failing tests? I've noticed that the tests on the master branch have been failing since mid-July, I don't think I should fix those issues here.

Libraries/Image/RCTImageViewManager.m
```diff
@@ -49,11 +50,11 @@ - (UIView *)view
   view.renderingMode = json ? UIImageRenderingModeAlwaysTemplate : defaultView.renderingMode;
 }
 
-RCT_EXPORT_METHOD(getSize:(NSURLRequest *)request
+RCT_EXPORT_METHOD(getSize:(RCTImageSource *)source
```
hramos: Should this be considered a breaking change?

hramos:
Tests look fine, it shouldn't block getting this imported. I do have a question on whether this could break existing uses of getSize() that expect it to accept a NSURLRequest object.

Libraries/Image/RCTImageViewManager.m
se-bastiaan: If someone is using this directly then it is, but I don't think you should be using it like that. The real exposed _documented_ API is defined in [Image.ios.js](https://github.com/facebook/react-native/pull/18850/files#diff-741217105b8bfdafc4248cbd8efa8eda) and is not changed in behaviour at all.

dulmandakh:
I propose to close this PR because even if you get size of an image by specifying HTTP headers, but you cannot render it or use as a source for Image component. Supporting that would require a lot of work, maybe breaking change.

dulmandakh:
@se-bastiaan I couldn't find a way to specify headers in source of Image component. So thought that it's not supported in RN.

se-bastiaan:
It is: https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageSource.js#L42

dulmandakh:
@se-bastiaan sorry, we should've updated docs.

se-bastiaan:
np. saw that this needed a rebase, so I've fixed that again.

Libraries/Image/Image.ios.js
```diff
@@ -135,6 +145,14 @@ Image = React.forwardRef(Image);
  * comment and run Flow. */
 Image.getSize = getSize;
 
+/**
+ * Retrieve the width and height (in pixels) of an image prior to displaying it
+ * with the ability to provide the headers for the request.
+ *
+ * See https://facebook.github.io/react-native/docs/image.html#getsizewithheaders
```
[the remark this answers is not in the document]
se-bastiaan: https://github.com/facebook/react-native/pull/18850#discussion_r253265098

Libraries/Image/Image.ios.js
```diff
@@ -135,6 +144,14 @@ Image = React.forwardRef(Image);
...
+ */
```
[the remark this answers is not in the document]
se-bastiaan: https://github.com/facebook/react-native/pull/18850#discussion_r253265098

Libraries/Image/Image.ios.js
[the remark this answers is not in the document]
se-bastiaan: Missed comment

Libraries/Image/RCTImageViewManager.m
se-bastiaan: @hramos

se-bastiaan:
'new native methods' being some a new interface between native and JS?

cpojer:
Exactly. Basically we cannot change a JS API calling a native function with different args so we have to create completely new APIs like `ImageLoader.getSizeWithHeaders`.

se-bastiaan:
Okay I get it I think: keep the old interface intact, create a new one for the headers call.

cpojer:
@se-bastiaan did you have any chance to get back to this PR?

se-bastiaan:
Sorry, forgot. Have made some quick changes now. Can fix any comments later today/this week.

Libraries/Image/Image.ios.js
```diff
@@ -42,6 +42,22 @@ function getSize(
   );
 }
 
+function getSizeWithHeaders(
+  uri: string,
+  headers: {[string]: string},
+  success: (width: number, height: number) => void,
+  failure?: (error: any) => void,
+) {
+  ImageViewManager.getSizeWithHeaders(
```
cpojer: Can you make this API a promise based API? See the PR here: https://github.com/facebook/react-native/pull/18875/files

We should try to adopt this pattern for new APIs since we cannot change the old one.

cpojer:
Nice, this is much better, thank you :) My only request is to turn this into a Promise based API so it aligns with what we do on Android better, as discussed in https://github.com/facebook/react-native/pull/18875/files.

se-bastiaan:
Wouldn't it be possible to align the iOS JavaScript `getSize` with Android by changing it to a Promise on the JS side? I know it's not completely related to this PR, was just wondering.

```
new Promise((resolve, reject) => {
        Image.getSize(uri, (width, height) => resolve({ width, height }), reject);
})
```

cpojer:
Yeah actually that’s a good idea, we should do this (in a separate PR after this one).

I’m on vacation for a week but I’ll get back to this PR soon.

se-bastiaan:
Whoohooo! :)

RockyF:
Image.getSizeWithHeaders is undefined in version 0.59.9

bumbummen99:
It is working fine for me on 0.62.2

PR #18872, opened 2018-04-16
Fix Native Rotation Android

### LINKED ISSUE #14161: Error when using rotation transform and "useNativeDriver: true" in Android
### Description

Animation do not work on Android when `useNativeDriver` is `true` and using a rotation transform setting the its angle as a `String` value. But it works as expected when setting it in radians using a `double` value. I usually set the rotation using a String as it is specified in the documentation, but in this case it crashes as it is expecting a double value.

The error I am getting is the following:

error screenshot

The following code sample would work either changing `useNativeDriver` to `false` or setting the rotation in radians.

```
  animateOpacity() {
    console.log('animation')
    Animated.timing(
      this.state.opacity_anim,
      {toValue: 1, duration: 600, useNativeDriver: true}
    ).start()
  }
  
  render() {
    return (
        {/* If the rotate transform is set in radians it works, but it doesn't when setting it in degrees with a string value. */}
        <Animated.View style={[
            styles.block,
            {
              transform: [{rotate: '90deg'}],
              opacity: this.state.opacity_anim
            }
        ]}>
          <Text style={styles.message}>Some text</Text>
        </Animated.View>
      </View>
    );
  }
```
I wrote an example in Expo as well:  https://snack.expo.io/SJGbutsl-

### Additional Information

* React Native version: v0.44.0, v0.43.0
* Platform: Android
* Development Operating System: macOS 10.12.5
* Dev tools: Android SDK 25.2.5, Android 7.0

scisci:
Fixes #14161
Android crashes in some cases if an animated transform config contains a string value, like a rotation.
This PR fixes that by ensuring all values sent to the native side are doubles. It adds `__transformDataType` to AnimatedTransform.js.

## Test Plan

Added integration test `ReactAndroid/src/androidText/js/AnimatedTransformTestModule.js` This test fails with the following error `INSTRUMENTATION_RESULT: longMsg=java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Double`, if the changes to AnimatedTransform.js are reverted.

## changelog

[Android] [Fixed] - Fixes Android crash on animated style with string rotation

### DIFF
--- a/Libraries/Animated/src/nodes/AnimatedTransform.js
+++ b/Libraries/Animated/src/nodes/AnimatedTransform.js
@@ -87,6 +87,22 @@ class AnimatedTransform extends AnimatedWithChildren {
     super.__detach();
   }
 
+  __transformDataType(value: any): number {
+    // Change the string array type to number array
+    // So we can reuse the same logic in iOS and Android platform
+    if (typeof value !== 'string') {
+      return value;
+    }
+    if (/deg$/.test(value)) {
+      const degrees = parseFloat(value) || 0;
+      const radians = degrees * Math.PI / 180.0;
+      return radians;
+    } else {
+      // Assume radians
+      return parseFloat(value) || 0;
+    }
+  }
+
   __getNativeConfig(): any {
     const transConfigs = [];
 
@@ -103,7 +119,7 @@ class AnimatedTransform extends AnimatedWithChildren {
           transConfigs.push({
             type: 'static',
             property: key,
-            value,
+            value: this.__transformDataType(value),
           });
         }
       }
--- a/Libraries/Animated/src/nodes/AnimatedTransform.js
+++ b/Libraries/Animated/src/nodes/AnimatedTransform.js
@@ -87,22 +87,6 @@ class AnimatedTransform extends AnimatedWithChildren {
     super.__detach();
   }
 
-  __transformDataType(value: any): number {
-    // Change the string array type to number array
-    // So we can reuse the same logic in iOS and Android platform
-    if (typeof value !== 'string') {
-      return value;
-    }
-    if (/deg$/.test(value)) {
-      const degrees = parseFloat(value) || 0;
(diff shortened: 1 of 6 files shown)


scisci:
@ janicduplessis

henrikra:
I am waiting for this PR :)

Libraries/Animated/src/nodes/AnimatedTransform.js
```diff
@@ -87,6 +87,22 @@ class AnimatedTransform extends AnimatedWithChildren {
     super.__detach();
   }
 
+  __transformDataType(value: any): number {
```
janicduplessis: Could you extract transformDataType to NativeAnimatedHelpers and use it here https://github.com/facebook/react-native/blob/4148976a83ab96029de1b26b515ff95d3cb58c10/Libraries/Animated/src/nodes/AnimatedInterpolation.js#L356 also.

scisci: OK I extracted to the common file. Interpolation works slightly differently (on an array) so it still needed the mapping function. This is from so long ago that I wasn't able to get the tests running again so maybe if you can check that it works.

cpojer:
@scisci it seems there are some problems with your PR. Would you mind rebasing it on master and making sure the JS tests pass?

Libraries/Animated/src/nodes/AnimatedTransform.js
janicduplessis: Yea sounds good, the main thing that is nice to extract out to be able to re-use is the deg to number conversion part. 👍

scisci:
@cpojer, I rebased and tests are now passing

janicduplessis:
@scisci Could you just address the lint issues that the bot pointed out? Especially the one about prettier since it might block landing this.

scisci:
@janicduplessis I've fixed the eslint errors.

PR #18879, opened 2018-04-16
Fix native postMessage test

### LINKED ISSUE #10865: [WebView] onMessage failing when there are JS warnings and errors on the page

ivank:
The current way we're testing if postMessage has been changed is no longer valid. Lodash has already changed the way they test against this, as it no longer works in blink (thow continues to work in webkit)

In chrome browser (Version 65.0.3325.181 (Official Build) (64-bit)) console type:

```js
String(window.postMessage) === String(Object.hasOwnProperty).replace('hasOwnProperty', 'postMessage')
```

It still works in safari (Version 11.0.3 (13604.5.6))

A simplified version of how lodash are testing against it - https://github.com/lodash/lodash/blob/4.0.1-npm-packages/lodash.isnative/index.js

With a simpler description about this method can be found here: https://davidwalsh.name/detect-native-function

Running:

```js
RegExp('^' +
  String(Object.prototype.toString)
  .replace(/[.*+?^${}()|[\]\/\\]/g, '\\$&')
  .replace(/toString|(function).*?(?=\\\()| for .+?(?=\\\])/g, '$1.*?') + '$'
).test(String(window.postMessage))
```

Works correctly in both chrome and safari.

Fixes #10865 - some hacks were suggested here to sidestep this issue. This PR should fix it outright

## Test Plan

In chrome browser (Version 65.0.3325.181 (Official Build) (64-bit)) console type:

```js
String(window.postMessage) === String(Object.hasOwnProperty).replace('hasOwnProperty', 'postMessage')
```
Should return "true" indicating that its "native" and not been changed. Currently returns "false"

## Release Notes

[IOS] [BUGFIX] [React/Views/RCTWebView.m] - Fix postMessage overwriting warning.
[ANDROID] [BUGFIX] [ReactAndroid/src/main/java/com/facebook/react/views/webview/ReactWebViewManager.java] - Fix postMessage overwriting warning.


ivank:
Hey @shergin - the tests for this seem to be failing, but that's not part of the changes in this PR. They seem to have been failing for quite some time now. If you can tell me which commit / tag to base this PR on that has passing tests, I could change the base and possible add some tests for this myself. As it stand thats rather hard to do unfortunately.

ivank:
@shergin after your own fixes in master (and my own fix as well) the tests are now passing. Care to review this?

PR #18892, opened 2018-04-17
Allow bundle task to be disabled for release

deecewan:
## Summary

The current system would always end up falling back to `true` for release builds

## Test Plan

- I ran my build after setting `bundleInProductionRelease` to false and noticed it still bundled.
- I updated the code in `react.gradle` and ran my build again - this resulted in the correct behaviour
- I removed the `false` setting and it correctly bundled.

A lot of logging revealed that when the `bundleIn${targetName}` flag was set to `false`, nothing actually changed.

## Changelog

[ANDROID] [Fixed] Allow disabling of bundling for release builds


deecewan:
I'm not really sure why my builds are breaking. They are breaking on code that has nothing to do with my code.

Anyone have ideas?

deecewan:
@janicduplessis sorry for the tag, but it looks like you've worked on this file before and I've seen you around the project.

I'm getting test failures that have seemingly nothing to do with the code I changed. In fact, I'm not sure how the code I'm changing could have possibly had an effect on `test_android`.

It seems like a pretty simple change - anything I can do to get it through?

deecewan:
@shergin any ideas on how I can get this through? as above, no idea why these tests would fail on this build.

PR #18948, opened 2018-04-19
Prevent console logging on iOS 11.3+ within WebSocket

maicki:

Fixes Xcode console output for web socket connections. Apple uses OSLog for logging within libnetworking starting 11.3+. The old way we hook into logging to prevent it will not work anymore.  

Let's hook into `__nwlog_pack` that is exclusively used by libnetworking and prevent the logging in there.

### DIFF
--- a/Libraries/WebSocket/RCTReconnectingWebSocket.m
+++ b/Libraries/WebSocket/RCTReconnectingWebSocket.m
@@ -19,6 +19,27 @@
 
 #if RCT_DEV // Only supported in dev mode
 
+// From https://github.com/apple/swift/blob/ad40c770bfe372f879b530443a3d94761fe258a6/stdlib/public/SDK/os/os_log.m
+typedef struct os_log_pack_s {
+  uint64_t olp_continuous_time;
+  struct timespec olp_wall_time;
+  const void *olp_mh;
+  const void *olp_pc;
+  const char *olp_format;
+  uint8_t olp_data[0];
+} os_log_pack_s, *os_log_pack_t;
+
+static void (*orig__nwlog_pack)(os_log_pack_t pack, os_log_type_t logType);
+
+static void my__nwlog_pack(os_log_pack_t pack, os_log_type_t logType)
+{
+  if (logType == OS_LOG_TYPE_ERROR) {
+    if (strstr(pack->olp_format, "%{public}s %s Connection has no connected handler") == NULL) {
+      orig__nwlog_pack(pack, logType);
+    }
+  }
+}
+
 static void (*orig_nwlog_legacy_v)(int, char*, va_list);
 
 static void my_nwlog_legacy_v(int level, char *format, va_list args) {
@@ -46,7 +67,7 @@ static void my_os_log_error_impl(void *dso, os_log_t log, os_log_type_t type, co
   }
 }
 
-#endif /* __IPHONE_11_0 */
+#endif /* RCT_DEV */
 
 @interface RCTReconnectingWebSocket () <RCTSRWebSocketDelegate>
 @end
@@ -63,6 +84,9 @@ static void my_os_log_error_impl(void *dso, os_log_t log, os_log_type_t type, co
     rebind_symbols((struct rebinding[1]){
       {"nwlog_legacy_v", my_nwlog_legacy_v, (void *)&orig_nwlog_legacy_v}
     }, 1);
+    rebind_symbols((struct rebinding[1]){
+      {"__nwlog_pack", my__nwlog_pack, (void *)&orig__nwlog_pack}},
+                   1);
 #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 /* __IPHONE_11_0 */
     rebind_symbols((struct rebinding[1]){
       {"_os_log_error_impl", my_os_log_error_impl, (void *)&orig_os_log_error_impl}


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

lhecker:
@maicki @hramos I tested this PR and can confirm that this solves #10027 in my case. 🙂
With this `__nwlog_pack` hook in place the `nwlog_legacy_v` hook could be removed though, right?

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -63,6 +84,9 @@ + (void)load
     rebind_symbols((struct rebinding[1]){
       {"nwlog_legacy_v", my_nwlog_legacy_v, (void *)&orig_nwlog_legacy_v}
     }, 1);
+    rebind_symbols((struct rebinding[1]){
+      {"__nwlog_pack", my__nwlog_pack, (void *)&orig__nwlog_pack}},
+                   1);
```
lhecker: I believe you need to "fix" the indentation here.

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -19,6 +19,27 @@
 
 #if RCT_DEV // Only supported in dev mode
 
+// From https://github.com/apple/swift/blob/ad40c770bfe372f879b530443a3d94761fe258a6/stdlib/public/SDK/os/os_log.m
```
lhecker: It might be necessary to add a `__IPHONE_OS_VERSION_MAX_ALLOWED >= 100300` check here since the entire API around this (e.g. `_os_log_pack_size` etc.) has only been added in iOS 10.3, but React supports up to iOS 8.

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -19,6 +19,27 @@
...
+  const char *olp_format;
+  uint8_t olp_data[0];
+} os_log_pack_s, *os_log_pack_t;
+
+static void (*orig__nwlog_pack)(os_log_pack_t pack, os_log_type_t logType);
+
+static void my__nwlog_pack(os_log_pack_t pack, os_log_type_t logType)
+{
+  if (logType == OS_LOG_TYPE_ERROR) {
+    if (strstr(pack->olp_format, "%{public}s %s Connection has no connected handler") == NULL) {
```
lhecker: Since you're comparing the entire string from start to end anyways I suggest changing this check to:
```c
strcmp(pack->olp_format, "%{public}s %s Connection has no connected handler") != 0
```

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -46,7 +67,7 @@ static void my_os_log_error_impl(void *dso, os_log_t log, os_log_type_t type, co
   }
 }
 
-#endif /* __IPHONE_11_0 */
+#endif /* RCT_DEV */
```
lhecker: This change is incorrect. This endif belongs to its previous `__IPHONE_11_0` check

Libraries/WebSocket/RCTReconnectingWebSocket.m
maicki: I change it to just compare to `Connection has no connected handler`, the same as in `my_nwlog_legacy_v`

Libraries/WebSocket/RCTReconnectingWebSocket.m
maicki: Correct, changed back.

Libraries/WebSocket/RCTReconnectingWebSocket.m
maicki: Added

Libraries/WebSocket/RCTReconnectingWebSocket.m
maicki: Fixed

maicki:
@lhecker Just addressed the comments, thanks for reviewing.

I would suggest we keep the `nwlog_legacy_v` hook to have at least for the near term some backwards compatibility.

Libraries/WebSocket/RCTReconnectingWebSocket.m
```diff
@@ -84,9 +88,12 @@ + (void)load
     rebind_symbols((struct rebinding[1]){
       {"nwlog_legacy_v", my_nwlog_legacy_v, (void *)&orig_nwlog_legacy_v}
     }, 1);
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 100300 /* __IPHONE_10_3 */
     rebind_symbols((struct rebinding[1]){
-      {"__nwlog_pack", my__nwlog_pack, (void *)&orig__nwlog_pack}},
-                   1);
+      {"__nwlog_pack", my__nwlog_pack, (void *)&orig__nwlog_pack}
+
```
lhecker: Sadly there's still a leftover newline here. 😕

lhecker:
@maicki Yeah you're right about the `nwlog_legacy_v` hook.
P.S.: Sorry that I didn't bundle those 4 reviews into one before. I forgot about this "Start Review" functionality Github provides.

Libraries/WebSocket/RCTReconnectingWebSocket.m
maicki: Done

vovkasm:
Oh no... please stop insert hard to debug code here... instead, please remove fishhook at all. See wip patch https://github.com/facebook/react-native/pull/17617

mmmulani:
thanks for this!

overall this change looks good to me, the only thing I might fix up is combining the if's to a single one.

I'm also a fan of this approach as it's only in dev mode.

vovkasm:
- It introduces forward incompatibilty, what if `os_log_pack_s` change in new version of ios?
- Moreover it will be security issue... field `olp_format` compared with `strstr`, but if it change, what address of memory will be supplied to `strstr`?
- It increases complexity and future support burden...
- It hides users errors, not only RN's

One can say "it's safe because it only in dev mode", but I will say:
- Each new dev mode deviation from production increases the need to run/test/debug production mode builds, so developer will create/enable additional level of debug itself... then he will forgot to disable some of them...
- Dev mode can be too easy be enabled in production software... I need to forgot about one single switch...

mmmulani:
we want dev mode to be a good experience for people though. Of course we're deviating from prod because we want to make it easy to develop. But I don't think this will cause much unexpected changes for people when comparing debug to prod.

> Dev mode can be too easy be enabled in production software... I need to forgot about one single switch...

Again, I don't think this is a good reason for us to not make the development experience good.

Regarding security, the attack vector would be another native module that you import. At which point, there's not much hope from defending against it.

mmmulani:
separately this got reverted for a build error, I'll try to reland

PR #18988, opened 2018-04-23
BREAKING - default underlineColorAndroid to transparent

### LINKED ISSUE #18938: Make the default for underlineColorAndroid transparent
https://twitter.com/notbrent/status/986843528004292608

This will help reduce differences between iOS and Android, the default style is actually kind of broken right now because we no longer add extra padding on Android.

Before:
image

After:
image

The easiest implement this might be just to add `underlineColorAndroid: 'transparent'` to TextInput default props.

mikaoelitiana:
Set default `underlineColorAndroid` to `transparent`.

Fixes #18938

## Test Plan

Use a TextInput in a component without defining `underlineColorAndroid`, the underline color should be transparent. 

## Release Notes

[ANDROID] [BREAKING] [TextInput] - set default underlineColorAndroid to transparent

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -638,6 +638,7 @@ const TextInput = createReactClass({
   getDefaultProps(): Object {
     return {
       allowFontScaling: true,
+      underlineColorAndroid: 'transparent',
     };
   },
   /**


janicduplessis:
@mikaoelitiana Thanks for the PR! I just updated the title and release not to mention this is a breaking change.

@mjesun Could you have a quick look to see if this could break anything @fb before we land. I guess looking for TextInputs that don't define underlineColorAndroid are the ones that will be affected.

mjesun:
@janicduplessis I can import the PR internally, but I'm probably not the best one to check this 🙂

cc'ing @yungsters who knows better than I do 😀

janicduplessis:
@mjesun Thanks for redirecting my request to the right person, I wasn’t sure who to ping on this one 😁

yungsters:
@mdvacca dug into this a bit just now and we were both surprised that the default was _not_ transparent after looking here at [ReactTextInputManager.java](https://github.com/facebook/react-native/blob/c595509048cc5f6cab360cd2ccbe7c86405baf92/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L417).

It turns out that at least in API 19, the default was a black underline. In API 25, the default is now no underline. So making `transparent` the default makes a lot of sense.

This change makes a lot of sense (instead of setting it in the view manager) because in the event that `underlineColorAndroid` is provided, it's likely better to send this value over the bridge and update the underlying view one time rather than initializing it to transparent and then re-setting it.

I'll get this merged internally. Thanks for the pull request!

janicduplessis:
@yungsters @mdvacca Thanks for looking at this, yea from what I understand how it works is that it applies a color filter to the underline drawable, which is by default based on the theme color. In this case clearing the color filter will just bring back the color to the default and not remove the underline. Applying a transparent color filter using default props seemed like the best way to make it work.

martnst:
@mikaoelitiana While I can see the point of changing the default `underlineColorAndroid` to transparent -  I think this change was a little bit poorly executed. 

- The docs do not tell the new default value https://facebook.github.io/react-native/docs/textinput#underlinecolorandroid

- Neither do they explain that it is possible to set `underlineColorAndroid={null}` to get back the old behaviour, which I picked up from the Android Native code [ReactTextInputManager.java#L420-L434](https://github.com/facebook/react-native/blob/4936d284df36071047ce776d9e2486c0371f7b97/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L420-L434).

- Also the PropType is not matching to the native code actually still taking `null` as a valid input. 

image


**Generally speaking I think changing a default value should effect the ability to set some value to achieve the behaviour before.* *

PR #18995, opened 2018-04-23
Remove @providesModule from all modules

rubennorte:
This PR removes the need for having the `@providesModule` tags in all the modules in the repository.

It configures Flow, Jest and Metro to get the module names from the filenames (`Libraries/Animated/src/nodes/AnimatedInterpolation.js` => `AnimatedInterpolation`)

## Test Plan

* Checked the Flow configuration by running flow on the project root (no errors):

```
yarn flow
```

* Checked the Jest configuration by running the tests with a clean cache:

```
yarn jest --clearCache && yarn test
```

* Checked the Metro configuration by starting the server with a clean cache and requesting some bundles:

```
yarn run start --reset-cache
curl 'localhost:8081/IntegrationTests/AccessibilityManagerTest.bundle?platform=android'
curl 'localhost:8081/Libraries/Alert/Alert.bundle?platform=ios'
```

## Release Notes

[INTERNAL] [FEATURE] [All] - Removed @providesModule from all modules and configured tools.


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  :exclamation: Big PR - <i>This PR is extremely unlikely to get reviewed because it touches 642 lines.</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

jest/hasteImpl.js
```diff
@@ -0,0 +1,75 @@
...
+    return hasteName;
+  },
+};
+
+function isHastePath(filePath/*: string*/)/*: bool*/ {
+  if (!filePath.endsWith('.js') && !filePath.endsWith('.js.flow')) {
+    return false;
+  }
+
+  if (!filePath.startsWith(ROOT)) {
```
chirag04: hey @rubennorte e2e tests are failing right now and my guess is this might be one cause. Refer: https://circleci.com/gh/facebook/react-native/42132?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link.

I'm seeing `ReferenceError: SHA-1 for file /private/tmp/react-native-9ke7hCEE/EndToEndTest/node_modules/fbjs/lib/invariant.js is not computed`. I might be completely wrong but looks like we don't handle haste modules in the node_modules directory?

rozele:
@rubennorte - can you explain how metro is configured by this PR? Does metro use the `.flowconfig`?

jest/hasteImpl.js
rubennorte: Hi @chirag04, I don't think that specific error is caused by the @providesModule change, but a bad integration between Metro and Watchman. @mjesun or @rafeca can you please confirm that?

mjesun: Hi, yes, that might be. `jest-haste-map` is not computing SHA-1s for `node_modules`; I'm going to put up a fix for this.

rozele:
@rubernorte - as far as I can tell, this will break `react-native-windows` because that haste config does not strip `.windows.js` platform extensions. We need to ensure there's a way to make bundling `react-native-windows` modules without customizing `rn-cli.config.js` because we need a fully automated way of installing `react-native-windows`.

rozele:
What's the point of removing `@providesModule` anyway? It's actually a really nice hook for plugin platforms to override core behavior. It has a solid, and very clear use case.

vincentriemer:
Just chiming in to say this will likely break `react-native-dom` as well

mjesun:
@rozele @vincentriemer We haven't removed support for `@providesModule`; we've just removed duplicated information: for a very long time we've been enforcing filenames to match with the `@providesModule` value, up to the point where the annotation became useless.

In fact, the change speeds up crawling process by removing the need - in certain cases - of opening the file to check the annotation, since the filename contains the same information. Haste maps should work the same way they've been working so far.

rozele:
@mjesun - Thanks for clarifying.

jest/hasteImpl.js
hramos: Hi @mjesun! When might that fix be ready? It's keeping us from running Android tests.

rozele:
@mjesun - hoping to get some clarity on your last response. I assumed when you said "we haven't removed support for `@providesModule`, you implied that I would still be able to override haste modules across NPM packages. This is how `react-native-windows` works, as there are many core `react-native` modules that need to be modified slightly to get things working on UWP.  Is this an accurate assumption? Currently, this behavior is not working on react-native 0.56 RC, so I'm trying to figure out if this is "by design" (i.e., no more cross-package haste overrides) or if it's some other bug.

mjesun:
Sorry, that's a FB "deviation" of the language, where we indistinctly use `@providesModule` to refer both to Haste and `@providesModule` itself. What was removed was the `@providesModule` annotations, but not the Haste support.

Haste is the ability of referring to a file from a `require` call just by its name, instead of its relative or absolute path. As long as your `View` file is called `View.*`, and Haste is instructed to read that, everything should still work.

rozele:
Okay - that's good to know @mjesun. Sounds like the experience I'm currently having with metro is definitely a "bug" as opposed to a new limitation of the package then. I've followed up with bug reports here and on metro:
https://github.com/facebook/react-native/issues/19953
https://github.com/facebook/metro/issues/188

PR #19025, opened 2018-04-25
reset _prevRenderedRowsCount when all ListView rows are loaded

mrousal:
Intention of this PR is to fix bug in RN.ListView as it doesn't re-render rows if the initial render happened in multiple steps. 
My guess is that _prevRenderedRowsCount member is used there to prevent re-render of already rendered rows when filling ListView in multiple steps. However it should be reset when all rows are loaded otherwise it prevents further row updates during ListView lifetime.

## Test Plan

Scenario:
Create ListView with more rows than set in initialListSize prop. Without the fix this ends with this._prevRenderedRowsCount == total number of rows. Change ListView state/props so some of the rows should be re-rendered.

Existing behavior:
No rows are re-rendered as shouldUpdateRow is always false. 

Expected behavior:
Rows are re-rendered if dataSource.rowShouldUpdate() returns true.

## Release Notes
[GENERAL] [BUGFIX] [ListView] - Fixing bug that prevents rows to be updated after being rendered in multiple steps

### DIFF
--- a/Libraries/Lists/ListView/ListView.js
+++ b/Libraries/Lists/ListView/ListView.js
@@ -645,7 +645,10 @@ var ListView = createReactClass({
       },
       () => {
         this._measureAndUpdateScrollProps();
-        this._prevRenderedRowsCount = this.state.curRenderedRowsCount;
+        this._prevRenderedRowsCount =
+          (this.props.dataSource.getRowCount() === this.state.curRenderedRowsCount
+            ? 0
+            : this.state.curRenderedRowsCount);
       },
     );
   },


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

cpojer:
ListView is deprecated and will be removed from React Native soon.

PR #19065, opened 2018-04-28
Fix root view cleanup

viotti:
The tearDownReactContext method of ReactInstanceManager was not removing
root views from its private mAttachedRootViews member.

The perceived problem was that when an Android application with multiple
root views was reloaded, root JS components that existed prior to that
reload were being remounted. They did not appear on screen, but React
life cycle methods such as componentWillMount were being called.

## Test Plan

I've created a MCVE project to illustrate the problem.

https://github.com/viotti/react-native-arvc-mcve

## Related PRs

None.

## Release Notes

[ANDROID] [BUGFIX] [ReactInstanceManager.java] - Fix root view cleanup


viotti:
Rebased. Disregard the previous close & reopen, happened by accident.

Issue #18954, opened 2018-04-19
Android hardware back exit or back page


forster-thomas:
Hi guys!

I need some help. I'm newest on react-native. And I don't know how fix that problem. I try any solutions, but neither one fix my occurrence.
That's the hell.

I have on "Home.js" and a "Detail.js".
When I'm on Home.js, and press android hardware button, I need to app close. That work's perfectly. But, when I go to Detail, and press android hardware button, the app close too. BUT I want to go back, and not close it.

See my files and show me the WAY. Thnks 4all.

Home.js

import React, { Component } from 'react';
import { connect } from 'react-redux';
import AXIOS from 'axios';
import { AsyncStorage, BackHandler, FlatList } from 'react-native';
import { Body, Button, Container, Content, Footer, FooterTab, Header, Icon, Left, Right, Text, Title } from 'native-base';
import { List, ListItem } from 'react-native-elements';
import { SetUserArticles, UserSession, modifySESSIONID } from '../actions/auth';
import Loading from '../components/Loading/Loading';
import { Login } from '../routes/Screens';

class Home extends Component {
  constructor(props) {
    super(props);

    this.state = {
      keyname: '',
      visible: true,
      loading: true,
      state: false,
      sessionid: '',
      url: '',
      email: '',
      password: '',
      ticketids: '',
      chamados: []
    };
  }

  componentWillMount() {
    this.setState({ keyname: this.props.navigation.state.key });
  }

  componentDidMount() {
    console.disableYellowBox = true;
    console.log(this.props);
    
    BackHandler.addEventListener('hardwareBackPress', () => {
      if (this.state.keyname === 'Home') {
        this._handleBackButton();
      }
    });
    this._loadAsyncArticles();
  }

  _handleBackButton() {
    BackHandler.exitApp();
  }

  _loadAsyncArticles() {
    AsyncStorage.getItem('UserArticles').then((ARTICLES) => {
      if (ARTICLES === null) {
        this._loadSession();
      } else {
        const DataArticles = JSON.parse(ARTICLES);
        this.setState({
          visible: false,
          chamados: DataArticles.Articles
        });
      }
    });
  }

  _reloadLogin = async () => {
    const url = this.state.url;
    const email = this.state.email;
    const password = this.state.password;

    try {
      return await AXIOS({
        url: `${url}/nph-genericinterface.pl/Webservice/AppTicket/Session`,
        method: 'POST',
        params: {
          UserLogin: email,
          Password: password,
        }
      });
[...]

PR #19077, opened 2018-04-30
Fix that BackHandler was not called on Android

jerolimov:
This fixes issues #18954 and #15497

The transformation from Set to (an reversed) Array does not work on Android devices when the remote debugging was started with `react-native run-android`.

## Test Plan

Add the following code to your App.js and run the app with `react-native run-android` in an Android device:

```
  componentDidMount() {
    BackHandler.addEventListener('hardwareBackPress', () => {
      console.warn('Callback called. Do not close app. Do nothing at all.');
      return true; // Do not close the app.
    });
  }
```

After adding this, the app should NOT close when you press the back button. Without this fix, the app closes (which is the default behaviour).

See also the comments from @Victoriayangx in #18954 and #15497.

## Related PRs

#17820 by @AuroreM tries to fix this also by improving the iterator polyfill. (See this after creating this PR...)

## Release Notes

[ANDROID] [BUGFIX] [BackHandler] - Fix that BackHandler was not called on Android

### DIFF
--- a/Libraries/Utilities/BackHandler.android.js
+++ b/Libraries/Utilities/BackHandler.android.js
@@ -18,22 +18,16 @@ type BackPressEventName = $Enum<{
   backPress: string,
 }>;
 
-const _backPressSubscriptions = new Set();
+const _backPressSubscriptions = [];
 
 RCTDeviceEventEmitter.addListener(DEVICE_BACK_EVENT, function() {
-  let invokeDefault = true;
-  const subscriptions = Array.from(_backPressSubscriptions.values()).reverse();
-
-  for (let i = 0; i < subscriptions.length; ++i) {
-    if (subscriptions[i]()) {
-      invokeDefault = false;
-      break;
+  for (let i = _backPressSubscriptions.length - 1; i >= 0; i--) {
+    if (_backPressSubscriptions[i]()) {
+      return;
     }
   }
 
-  if (invokeDefault) {
-    BackHandler.exitApp();
-  }
+  BackHandler.exitApp();
 });
 
 /**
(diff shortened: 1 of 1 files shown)


jerolimov:
@facebook-github-bot ping

jerolimov:
Over 50 days and zero feedback 😞

jerolimov:
Ping @TheSavior @sophiebits @rubennorte because you already worked on the changed file.

jerolimov:
@eslint-bot ping

TheSavior:
Fwiw, it is hard to feel confident merging this because I’m not sure anyone on our team is familiar with this code and the potential edge cases that could occur which might not be in the happy path you mention in your test plan.

For this to get some more traction we probably need to hear from more people that it works for them and didn’t cause any problems. If there are other edge cases you can think of which you can test and include in your test plan that would be great.

Essentially, can you help us feel more confident in this change? :-/

jerolimov:
@TheSavior In some environments the Set operators .add .delete .reverse does not work as expected. I did not understand why exactly this happen, because I have also projects which has no problem here. But you can see some other people running into this issue: #18954 and #15497

I'm also unhappy that nobody verifies this.. But in the one project we had, changing this implementation from Set to an Array works fine.

The API works still the same way.

https://github.com/facebook/react-native/blob/e09622c68aec0da301cd0c622b7baf50eaaf5683/Libraries/Utilities/BackHandler.android.js

1. The app could add and remove event listeners to BackHandler (line 74 and 89). Sure, a Set provides some easier api, but based on the problem that it doesn't work.. I switched this to Array.push and Array.splice.
2. The listeners was called in reverse order, which means last listener first (line 24)
3. If one if these added listeners return true (or truthy), it means "I handled the back button", stop calling the other listeners and do not close the app. (line 26)
4. If no listener return a true (or truthy) value, "nobody" handles the back button and we want to close the app (line 30)

TheSavior:
Alright, this seems fine. Let’s give it a shot. I can’t import from my phone so pinging @hramos to land, otherwise I’ll land later.

Issue #17091, opened 2017-12-06
[VirtualList] Render blockages for multiple seconds


mikelambert:
### Is this a bug report?

Yes.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes.

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 6.10.2
  Yarn: 0.15.1
  npm: 4.6.1
  Watchman: 4.1.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: ^16.0.0 => 16.0.0
  react-native: file:../../react-native => 0.50.2

### Steps to Reproduce

When you use a VirtualList with refresh component, and then hide the refresh component to show a list of variable sized elements (ie, a section list with headers that are differently-sized than the elements), you can have extraordinarily-long delays during the second rendering frame.

### Expected Behavior

It renders quickly.

### Reproducible Demo

You can see a repro case here: https://snack.expo.io/HkRgOcEZG

After the simulated 500ms load, it proceeds to render item 0. Then it gets "stuck" for a 5+ seconds before rendering items 1-to-N. You can make it even worse by uncommenting maxToRenderPerBatch={1} (ie, rendering in smaller batches makes it even slower, and makes it take even longer to render an initial batch of items!)

===============

My long-breakdown of what's going on from poking around with debug statements:

When the list loads and the refresh indicator hides, in `_scheduleCellsToRenderUpdate` the `distTop` is negative, and the velocity is negative (due to the refresh indicator being hidden). Because it's negative, it triggers a "hi priority" updates whenever `_scheduleCellsToRenderUpdate` is called. Even when there are more than enough elements on screen. This is Bug 1 (ie, logically, the user is not scrolling, so I'm not sure this should be a high priority update).

After the first cell is laid out in `_onCellLayout` (ie, the first `initialNumToRender` cells) and it has an average cell size , it calls `_onCellLayout->_scheduleCellsToRenderUpdate->_updateCellsToRender`, and uses `computeWindowedRenderLimits` to compute the new state of the `{first, last}` render window. And `computeWindowedRenderLimits`, uses the average cell size to figure out how many cells are visible (and how many should be rendered for the overscan), and computes the new `{first, last}`.

[...]

PR #19081, opened 2018-05-01
Prevent unintended VirtualizedList high priority rendering

jabbawookiees:
VirtualizedList sets high priority rendering when scrolled to the top of a list even if the bottom of the rendered list is still far away and there's nothing further to render above. This causes severe non-responsiveness while dozens of items are waiting to be rendered and user inputs can't get through.

The fix is simply to not consider it high priority to render more items when we reach the top and the first item has already been rendered, or when we reach the bottom and the last item has been rendered.

The code change just splits the two cases (hitting the top and bottom) instead of combining them into one.

## Test Plan

I wrote a small application that switches between two FlatLists. 

Demo of the original VirtualizedList

I'm clicking pretty often on the button, but you can see that it only switches lists a split second after the rendering is all done because the javascript thread was being blocked.

Demo of the fixed VirtualizedList

You can see this one is significantly more responsive than the previous one.

The source code for the demo application is available [here](https://gist.github.com/jabbawookiees/ba93a4e7b4f9b8f3acbc157e4fd04877)

## Release Notes

[ GENERAL ] [ ENHANCEMENT ] [ Libraries/Lists/VirtualizedList.js  ] - Prevent high priority rendering at the top or bottom of a list.

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1344,19 +1344,23 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     const {offset, visibleLength, velocity} = this._scrollMetrics;
     const itemCount = this.props.getItemCount(this.props.data);
     let hiPri = false;
-    if (first > 0 || last < itemCount - 1) {
+    const scrollingThreshold =
+      /* $FlowFixMe(>=0.63.0 site=react_native_fb) This comment suppresses an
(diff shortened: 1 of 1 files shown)


jochem725:
@jabbawookiees This PR seems to fix https://github.com/facebook/react-native/issues/17091. 
The list no longer shows blocking behavior 😄

ssssssssssss:
ping @sahrens

Issue #10822, opened 2016-11-09
Touchable* onPress not working on ScrollView composition


7ynk3r:
### Description

I have one `ScrollView`(a) containing another `ScrollView`(b) and a `TouchableOpacity`(c) floating on top for (b). While scrolling on (b), `onPress` on (c) is not triggered. 
- This only happens when `scrollEventThrottle` is 50 or less
- Scroll on (a) is disabled.
- It doesn't matter the direction of (a) or (b). 

### Reproduction

Use the following code, scroll and press the yellow square button. The button will only work if you're not scrolling.

```
import React, { Component } from 'react';
import {
  AppRegistry,
  View,
  ScrollView,
  TouchableOpacity,
  Dimensions,
} from 'react-native';

const {width,height} = Dimensions.get('window');

export default class scrollplusbutton extends Component {
  render() {
    const color = ['green','red','blue'];
    return (
      <ScrollView scrollEnabled={false}>

        <View style={{width, height}}>

          <ScrollView scrollEventThrottle={50}>
            {[...Array(20)].map((_, i) =>
              <View key={i} style={{height:i*50, backgroundColor:color[i%color.length]}}/>
            )}
          </ScrollView>

          <TouchableOpacity
            style={{
              position:'absolute',
              left:0,
              top:0,
              width:100,
              height:100,
              backgroundColor:'yellow',
            }}
            onPress={()=>alert('hello!')}
          />

        </View>

      </ScrollView>
    );
  }
}

AppRegistry.registerComponent('scrollplusbutton', () => scrollplusbutton);
```

### Additional Information

* React Native version: "react": "15.3.2" and "react-native": "^0.37.0"
* Platform: iOS (works well on Android)
* Operating System: MacOS

mdvacca:
same problem here

michaelcpuckett:
+1

sachinrekhi:
Running into the same problem with React Native 0.39.2.

I've tried both TouchableWithoutFeedback and TouchableOpacity and in neither case the onPress handler is triggered when rendered within ScrollViews.

This appears to be a regression as it was working fine for me in React Native 0.34.1 and only saw the issue occur immediately after upgrading to RN 0.39.2.

Strange thing is the onPress handler is properly triggered for me in the iOS Simulator, but is not triggered when I actually load the app on my iPhone.

sachinrekhi:
[...]

PR #19110, opened 2018-05-03
Prop to disable the default pan responder on ScrollView that blocks touches when snapToInterval is enabled

PimDeWitte:
## Summary

The ScrollView component seems to have been written for a default touch up / touch down scrolling behavior. However, when using snapToInterval, this default behavior of adding PanResponders while the ScrollView is animating or has an active touch causes bugs with the PanResponder hijacking touches it is not supposed to hijack. 

### Video
Video that shows the changes (First FlatList shown is WITH the native props disabled, second one is WITHOUT it disabled (you can see it hijacking touches):

https://imgur.com/a/2F4XbkJ

You may recreate this behavior as follows:

1) Create a FlatList or ScrollView 
2) Fill it with objects of the size of 400 px that has any onPress function
3) Set snapToAlignment to 'start'
4) set snapToInterval to the height of the object you created

Snap down once, you will notice that your first touch after scrolling does not respond, and the onPress function you defined does not fire. This is due to the default ScrollView PanResponder not having received an event that the touch was released, due to the scrolling happening manually when snapping.

This has fixed the issue for our production application for ~10k people, and is a simple prop that is optional to use. Therefore should not cause any issues other than give people like myself trying to figure out why a touch wasn't responding the option to fix it in 5 minutes rather than spend a week looking that React Native apparently attaches a pan responder in JS to a scrollview.

## Related PRs

Docs change:
https://github.com/facebook/react-native-website/pull/337

Other related issues (though not exact since our use case with snapToInterval is so specific)
https://github.com/facebook/react-native/issues/10822

## Changelog

[GENERAL] [Added] - ScrollView: New disableScrollViewPanResponder that disables the default ScrollView pan responder, and allow for complete control over the touch behavior within a scrollview. 

This is particularly useful when using vertical snapToInterval and snapToAlignment. Without it, it can hijack certain certain onPress/PanResponders. Be careful to not use this on a default ScrollView without snapToInterval since it may cause unexpected behavior and cause accidental touches while scrolling.


Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -95,6 +95,13 @@ type IOSProps = $ReadOnly<{|
    * @platform ios
    */
   bounces?: ?boolean,
+  /**
+   * By default, ScrollView has an active pan responder that hijacks panresponders 
+   * deeper in the render tree in order to prevent accidental touches while scrolling.
+   * However, in certain occasions (such as when using snapToInterval) in a vertical scrollview
+   * You may want to disable this behavior in order to prevent the ScrollView from blocking touches 
+   */
+  disableScrollViewPanResponder: PropTypes.bool,
```
hramos: Flow found some issues here:
```
Unexpected reserved type

Libraries/Components/ScrollView/ScrollView.js:104:44
  104|   disableScrollViewPanResponder: PropTypes.bool,                           
                                                  ^^^^
```

To reproduce locally try running
```
flow . --flowconfig-name .flowconfig.android
```

hramos: Changed to ?boolean internally.

PR #19192, opened 2018-05-08
[Docker] Update Android docker images

hramos:
Quick update to switch to a new React Native CI org on Docker Hub. Note that the images are not yet automatically generated on CI. We could do this on Circle CI in certain scenarios:

- Base image needs to be updated whenever the Android development environment has changed (e.g. switch to a new SDK version, build tools, etc)
- Tests image should ideally be updated on each commit

This PR should be safe to land as Circle CI is not yet using these images.


hramos:
Landing as this should not affect the rest of the codebase.

PR #19220, opened 2018-05-11
fixes(Android): while alert method buttons param type not array

jzhangdev:
## Summary

On Android platform, buttons param type is array, should be judged:

```js
static alert(title, message?, buttons?, options?, type?)
```

## Test Plan

```js
Alert.alert(
  'title',
  'some message',
  {cancelable: true}
)
```

So let's write that up, iOS platform is ok, but Android platform was a mistake, buttons.slice(0, 3), there is no correct judgment on the type of buttons.

## Changelog

[ANDROID] [Fixes] - Alert: while alert method buttons param type not array


chirag04:
@jzhangdev Can you provide a test plan? I'm not sure this is the right fix.

jzhangdev:
@chirag04 Hi, I just updated Test Plan, you can try this.

hramos:
Rebased and resolved conflicts. @chirag04, how do you feel about this PR?

chirag04:
@jzhangdev I don't mean to block but I don't understand what this is fixing. The code in question is all inside of a `if (android)` check. so ios should be good. for Android, button type is marked as `Array`. Don't think we gain anything from runtime type check. cc @hramos

Issue #19226, opened 2018-05-11
[Node 8] Prepare for Node 8


hramos:
- [X] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [X] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

The time has come to bump the minimum required Node version to 8. [Node 4.x LTS reached end of life, and Node 6.x entered maintenance mode, on April 30, 2018](https://medium.com/the-node-js-collection/april-2018-release-updates-from-the-node-js-project-71687e1f7742). Furthermore, this will allow us to be consistent with the rest of the JavaScript code at Facebook.

The codebase is ready for this, as our tests have been [running on Node 8 for a while now](https://github.com/facebook/react-native/commit/8fa9984b11f6dfd00531f424b05e68394bd30c03).

Things that need to be done:

- [X] Update the documentation to tell people to install Node.js 8 or newer ([`react-native-website/#355`](https://github.com/facebook/react-native-website/pull/355))
- [X] Test with Node 8
- [X] Upgrade the Circle CI tests to run against Node 8 (8fa9984b11f6dfd00531f424b05e68394bd30c03)
- [X] Convert `Libraries/` to `let`/`consts` (8f5ebe5952d0675b463137103a82f3fb0c26ae0d)
- [X] Allow trailing commas in the codebase (1e2de712907e5fe0d17648f0ff5c81d4384ca85b) 
- [x] Run Prettier (86b6f5d39ecc571260dee891aa422f8d3cc807e8, dca21c8f234803a32a74a2080b7bc9c8493e7c45, d01ab66b47a173a62eef6261e2415f0619fefcbb, additional commits incoming)
- [x]  Update the version check message in the packager (`local-cli‎/server‎/‎
checkNodeVersion.js`)
- [x] Remove the Node 6 compatibility tests on Circle CI (`.circleci‎/‎
config.yml`)

The last three action items are in flight and should be landing soon.

cc @ide (h/t #2545), @TheSavior, @grabbou 

## Environment

N/A

## Steps to Reproduce

N/A

## Expected Behavior

N/A

## Actual Behavior

N/A

PR #19230, opened 2018-05-11
[Node 8] Bump to Node 8

hramos:
Moving to Node 8 in React Native will allow us to be consistent with Facebook, e.g. trailing commas.

As of April 30, Node 4.x is end of life, and Node 6.x entered maintenance mode (see https://medium.com/the-node-js-collection/april-2018-release-updates-from-the-node-js-project-71687e1f7742).

This will require our docs to be updated to require Node 8 as a minimum. This should be done as part of the following PR: https://github.com/facebook/react-native-website/pull/355

Circle CI already runs all JS tests using Node 8, and it has an extra compatibility check that uses Node 6. This workflow is dropped as we no longer need to check for failures when Node 6 is used.

See https://github.com/facebook/react-native/issues/19226

Test Plan
=========

flow
CircleCI

Release Notes
=============

[GENERAL] [BREAKING] [Node] - Bump minimum req. Node version to 8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,24 +365,6 @@ jobs:
       - store_test_results:
           path: ~/react-native/reports/junit
 
-  # Runs JavaScript tests on Node 6
-  test_javascript_node6_compatibility:
-    <<: *defaults
-    docker:
-      - image: circleci/node:6
-    steps:
-      - checkout
-      - run: *setup-artifacts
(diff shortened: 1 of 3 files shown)


hramos:
This is a mirror of D7961929 which was approved by @TheSavior. Opening a PR to ensure Circle CI works as expected.

Issue #17799, opened 2018-01-31
'RCTPackagerClient.h' file not found


zaiz91:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
Yes

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.6.0
  Yarn: 0.27.5
  npm: 5.3.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.52.2 => 0.52.2

### Steps to Reproduce

(Write your steps here:)

1. Upgrade react from 16.0.0-alpha.12 to 16.2.0
2. Upgrade react-native from 0.47.2 to 0.52.2
3. Reinstall all node_modules and Pods

### Expected Behavior

Xcode Build with no Errors

### Actual Behavior

Got following error:

../node_modules/react-native/React/Modules/RCTDevSettings.mm:23:9: 
'RCTPackagerClient.h' file not found

![image](https://user-images.githubusercontent.com/9715008/35616278-16c47040-0675-11e8-8ae5-b647e2f0adb6.png)

I did not find any Issues with the same bug. But I also have no clue why I get this bug. Maybe someone can help me?

zaiz91:
Adding 'DevSupport' to podfile fixed the issue.

pod 'React', :path => '../node_modules/react-native', :subspecs => [
    'DevSupport',

mattop1709:
@zaiz91 i have inserted 'DevSupport', but it still won't work..

Sadanandteggi:
I am still getting this error despite having     'DevSupport', in pod file. Can I expect any help here ?
Thanks

Baskerville42:
+1
the same problem!

PR #19265, opened 2018-05-15
Fix CocoaPods integration without DevSupport subspec

ikesyo:
This is the thing which is handled in https://github.com/orta/cocoapods-fix-react-native today.

https://github.com/orta/cocoapods-fix-react-native/blob/89a78ad34950b5d6b3de657914c8ced8bf37ee8e/lib/cocoapods-fix-react-native/versions/0_55_3-post.rb#L140-L164

This should be the correct fix for #17799.

## Test Plan

Integrating React Native into an existing iOS project using CocoaPods without `DevSupport` subspec and seeing if the project successfully compiles.

## Release Notes

[IOS] [BUGFIX] [DevSupport] - Fix CocoaPods integration without DevSupport subspec

### DIFF
--- a/React/Modules/RCTDevSettings.mm
+++ b/React/Modules/RCTDevSettings.mm
@@ -18,7 +18,6 @@
 #import "RCTEventDispatcher.h"
 #import "RCTJSCSamplingProfiler.h"
 #import "RCTLog.h"
-#import "RCTPackagerClient.h"
 #import "RCTProfile.h"
 #import "RCTUtils.h"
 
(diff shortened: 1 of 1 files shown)


ikesyo:
I think this should also be backported to `0.55-stable` branch.

hramos:
0.55 has been out for a while, this can probably wait for 0.56, but best to discuss over in https://github.com/react-native-community/react-native-releases

ikesyo:
What should I do for the test failures? Those are apparently unrelated  to the changes in this PR.

ikesyo:
Could you please review this, maintainers?

hramos:
@orta, can you take a look at this PR?

orta:
Yep, this PR does what I'd expect it to - moving some of the headers and imports to be hidden behind compiler flags that are only set when the dev web socket is included 👍

orta:
Also, want to emphasize that you can always ping me on any of these kind of issues 👍

ikesyo:
Thanks @orta!

PR #19281, opened 2018-05-16
[GENERAL] [BUGFIX] [Image] Fix props width, height and tintColor

brunolemos:
## Test Plan

Currently this already works with the live version:
- `<View width={100} height={100} margin={100} />`
- `<Image width={100} height={100} source={[{ uri: '' }]} />` `// width and height when source is an array`
- `<Image margin={100} />` `// any style as prop, except tintColor`

This only works after this pr:
- `<Image width={100} height={100} source={{ uri: '' } />` `// width and height when source is not an array`
- `<Image tintColor="#0000FF" />` `// tintColor as prop`

```jsx
<Image
  width={200}
  height={200}
  borderRadius={100}
  margin={20}
  source={{ uri: 'https://arcweb.co/wp-content/uploads/2016/10/react-logo-1000-transparent.png' }}
  // style={{ width: 300, height: 300, tintColor: 'purple' }}
  tintColor="green"
/>
```

## Release Notes

[GENERAL] [BUGFIX] [Image] Fix props width, height and tintColor

### DIFF
--- a/Libraries/Image/Image.android.js
+++ b/Libraries/Image/Image.android.js
@@ -239,7 +239,7 @@ const Image = createReactClass({
     let sources;
     if (source.uri) {
       const {width, height} = source;
-      style = flattenStyle([{width, height}, styles.base, this.props.style]);
+      style = flattenStyle(['width' in source && { width }, 'height' in source && { height }, styles.base, this.props.style].filter(Boolean)) || {};
       sources = [{uri: source.uri}];
     } else {
       style = flattenStyle([styles.base, this.props.style]);
--- a/Libraries/Image/Image.ios.js
+++ b/Libraries/Image/Image.ios.js
@@ -106,8 +106,8 @@ const Image = createReactClass({
     }
 
     const resizeMode =
-      this.props.resizeMode || (style || {}).resizeMode || 'cover'; // Workaround for flow bug t7737108
-    const tintColor = this.props.tintColor || (style || {}).tintColor; // Workaround for flow bug t7737108
+      (style || {}).resizeMode || this.props.resizeMode || 'cover'; // Workaround for flow bug t7737108
+    const tintColor = (style || {}).tintColor || this.props.tintColor; // Workaround for flow bug t7737108
 
     if (this.props.src) {
       console.warn(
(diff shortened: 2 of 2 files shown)


brunolemos:
Ping @shergin @TheSavior this is a small one

TheSavior:
Hmm. 

We recommend putting those props in the `style` prop. React Native originally shipped supporting setting styles as props instead of in the styles object but we are moving away from that. We actually expect to be removing that functionality in a future release, forcing those styles to be defined in `style`.

If you set them in `style` instead does the functionality you expect work as-is?

brunolemos:
@TheSavior Thanks. Yes, style works as expected.

I understand this is not the recommended way of doing it, but while RN still haven't moved away from it, I see no harm on including this fix, right? Like, all other props works, except for these three.

TheSavior:
We are actively removing this functionality, so I don't think we want to reverse direction on this and open the door to it being used more.

brunolemos:
> We are actively removing this functionality

@TheSavior can you point me to PRs or discussions about this?

TheSavior:
I don't actually think we've posted about this anywhere yet (you're hearing it first!) :'( 

We have been working on rewriting the core of React Native (project is called Fabric) and want to be stricter with what props get passed in to native components. There are some perf wins to be gained here as well. One of the other problems we've had with styles being set as props are that if you define a prop on a component with the name of a style then you can run into conflicts.

We have been flow typing the core components and disallowing unsupported props being passed in that way as a first step. To support open source better we need to get a dev mode runtime warning or something out so people can find and have time to update.

brunolemos:
@TheSavior I'm removing the "styles as props" support from TypeScript definition (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28217), please let me know if this is still the plan or if someone decided that it will keep working as today on Fabric.

TheSavior:
We still intend to remove styles as top level props and the React Native flow definitions enforce it as such. It would be good to have typescript be consistent with that. Thanks for making the change to the typescript definition!

brunolemos:
Ok! Thanks for the answer.

brunolemos:
TypeScript change merged and deployed.

Issue #6228, opened 2016-03-01
[Android] [DialogModule] IllegalStateException


grabbou:
Hey,

Again, not sure how to reproduce - ~20 crashes so it's still a super small percentage of overall app install. 

```
IllegalStateException: Can not perform this action after onSaveInstanceState
  Module "android.app.FragmentManagerImpl", line 1343, in checkStateLoss
  Module "android.app.FragmentManagerImpl", line 1361, in enqueueAction
  Module "android.app.BackStackRecord", line 729, in commitInternal
  Module "android.app.BackStackRecord", line 705, in commit
  Module "android.app.DialogFragment", line 292, in dismissInternal
  Module "android.app.DialogFragment", line 258, in dismiss
  Module "com.facebook.react.modules.dialog.DialogModule$FragmentManagerHelper", line 115, in dismissExisting
  Module "com.facebook.react.modules.dialog.DialogModule$FragmentManagerHelper", line 121, in showNewAlert
  Module "com.facebook.react.modules.dialog.DialogModule", line 234, in showAlert
  Module "java.lang.reflect.Method", in invoke
  Module "java.lang.reflect.Method", line 372, in invoke
  Module "com.facebook.react.bridge.BaseJavaModule$JavaMethod", line 249, in invoke
  Module "com.facebook.react.bridge.NativeModuleRegistry$ModuleDefinition", line 158, in call
  Module "com.facebook.react.bridge.NativeModuleRegistry", line 58, in call
  Module "com.facebook.react.bridge.CatalystInstanceImpl$NativeModulesReactCallback", line 428, in call
  Module "com.facebook.react.bridge.queue.NativeRunnableDeprecated", in run
  Module "android.os.Handler", line 739, in handleCallback
  Module "android.os.Handler", line 95, in dispatchMessage
  Module "com.facebook.react.bridge.queue.MessageQueueThreadHandler", line 31, in dispatchMessage
  Module "android.os.Looper", line 145, in loop
  Module "com.facebook.react.bridge.queue.MessageQueueThreadImpl$3", line 184, in run
  Module "java.lang.Thread", line 818, in run
```

mkonicek:
Looks like a bug in the Android Dialog module, in the `dismissExisting` method.

mkonicek:
I'll try to find some time to take a look / ask @foghina if he has time :)

grabbou:
Yeah, or I could look into that as well just to get some Java :fire:

DoingLiaw:
[...]

PR #19332, opened 2018-05-17
DialogModule crash fixes.

dryganets:
I'm trying to fix DialogModule crashes we have in production: #6228

In this PR I'm fixing the following problem:

The fragment manager methods should be called only from the foreground. Now dismissExisting is protected by the mIsForeground variable as well as showNewAlert method.

## Test Plan

Make sure that dialogs work after this change.

## Release Notes
[ANDROID] [BUGFIX] [DialogModule] - Race condition in dialog module fixed.


webdevbyjoss:
@dryganets hey, I'm facing these crashes in a DialogModule on React Native v57. We didn't have this addressed yet in the latest RN version, aren't we? Shall we update this PR and merge it?

dryganets:
@webdevbyjoss, I updated the PR. It fixes the problem for us.

@hramos, could you take a look?

PR #19333, opened 2018-05-17
Vendor fetch polyfill, remove default blob response type

janicduplessis:
While investigating an issue about blobs (https://github.com/facebook/react-native/issues/18223), I noticed that the fetch polyfill (https://github.com/github/fetch) uses blobs as the response type by default if the module is available (https://github.com/github/fetch/blob/master/fetch.js#L454). This surfaced some issue with the blob implementation on iOS that has since been fixed.

However after further review of the fetch polyfill and the way Blobs work in RN, I noticed a major issue that causes blobs created by fetch to leak memory. This is because RN blobs are not deallocated automatically like in the browser (see comment https://github.com/facebook/react-native/blob/master/Libraries/Blob/Blob.js#L108) and the fetch polyfill does not deallocate them explicitly using the close method.

Ideally we should implement automatic blob cleanup when the instance is garbage collected but implementing that is probably somewhat complex as it requires integrating with JSC. For now I suggest disabling the default handling of requests as blobs in the fetch polyfill. This will mitigate the issue for people not using Blobs directly. I'm not sure how well documented the Blob module is but we should make it clear that they currently require explicit deallocation with the close method for people using them directly.

## Test Plan

Run a simple http request using fetch and make sure it does not use the Blob module anymore.

## Release Notes

[GENERAL] [BUGFIX] [fetch] - Do not use blobs to handle responses in the fetch polyfill, fixes potential memory leak.

### DIFF
--- a/Libraries/vendor/core/whatwg-fetch.js
+++ b/Libraries/vendor/core/whatwg-fetch.js
@@ -1,0 +1,529 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+// Fork of https://github.com/github/fetch/blob/master/fetch.js that does not
+// use reponseType: 'blob' by default. RN already has specific native implementations
+// for different response types so there is no need to add the extra blob overhead.
+
+// Copyright (c) 2014-2016 GitHub, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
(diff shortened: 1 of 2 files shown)


hramos:
I'm checking with our open source team on how to proceed with licensing here.

react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

rozele:
@janicduplessis - maybe it will be easier to just fix the issue to begin with? I was thinking about how we would implement this in `react-native-windows`, and I came across the `JsObjectBeforeCollectCallback` in ChakraCore, which would allow us to attach finalizers to JS object:
https://github.com/Microsoft/ChakraCore/wiki/JsObjectBeforeCollectCallback

My thought is, in the BlobModule for react-native, we add a new native callback for react-native, e.g., `nativeFinalizer` that takes an object ref and a function ref. Internally, this would maintain some native finalizer mapping, wire up the call to `JsObjectBeforeCollectCallback` (or in the case of JSC, [`JsObjectFinalizeCallback`](https://developer.apple.com/documentation/javascriptcore/jsobjectfinalizecallback)), and invoke the JavaScript function just before garbage collection. This would allow us to wire up a finalizer on Blob objects that would make sure native blobs are cleaned up. Consequently, this could also help support an interesting `ErrorModule` for promises that would allow us to hold on to native exception objects while a rejected Promise is still being handled, so the user could query the `ErrorModule` for additional native exception details.

janicduplessis:
@rozele Makes sense, I don't have time to work on implementing this at the moment though, I suggest we still land this to fix the issue ASAP and we can work on implementing finalizers after. I think I'd rather not use blobs when using fetch anyway since RN already has optimized native methods for loading the different content types.

janicduplessis:
Ping @hramos

shyaniv7:
This PR never made it into 0.55 or 0.56 and this still an issue.

Can anyone merge it?
Thanks in advance

hramos:
@shyaniv7 this fix is clearly in 0.56, as you can see in the list of tags in the commit that closed the PR: https://github.com/facebook/react-native/commit/122b3791ede095345f44666691aa9ce5aa7f725a.

jamesmoss:
This change has broken `fetch` with blobs in 0.57 on iOS.

If we read a jpeg from disk via `fetch` the `Blob` we're getting back is now roughly 2x the file size and the contents can't be read by any image editor:

```
const response = await fetch('/path/to/some/file/on/disk');
const blob = await response.blob();

console.log(blob.size); // incorrect value returned
```

After some investigation, this block being removed from the `fetch` polyfill when it was copied across has caused this:

```
if ('responseType' in xhr && support.blob) {
  xhr.responseType = 'blob'
}
```

I'm putting together a proper bug report with example code but leaving this here for now in case anybody else is hitting this issue.

PR #19361, opened 2018-05-20
Update and rename glog.podspec to GLog.podspec

onyoon7:
An error occurs when running `pod install` as described in [[the official documentation](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#configuring-cocoapods-dependencies)]

Changed glog file similar to official document and other third-party-podspecs forms

## Release Notes


orta:
This should stay as `glog` IMO, as it's the official name in the CocoaPods trunk repo, the docs should be updated instead 👍

PR #19416, opened 2018-05-23
iOS: Resolve a build warning for 32bit devices

chuganzy:
This PR resolves a warning `Implicit conversion loses floating-point precision: 'double' to 'CGFloat' (aka 'float')` when building not only this library **but also apps** for 32bit devices since it's declared in public header file.

## Test Plan

This does not affect anything I believe.

## Related PRs

N/A

## Release Notes

[IOS] [MINOR] [RCTAnimationDriver] - Resolve a build warning for 32bit devices


chuganzy:
I think @javache can review this.

PR #19451, opened 2018-05-25
Fixes UI glitch when drawing selective corners with borderRadius

### LINKED ISSUE #11897: Artifacts on View with borderRadius on only 2 corners on iOS
### Description

I need a couple of Views with border radius only on the top or bottom corners, to use as container for TextInputs (its a email/password form).

### Reproduction

Using this code:
```javascript
    // style
    inputExample: {
        borderColor: '#9b9b9b',
        backgroundColor: 'white',
        borderWidth: 1,
        borderStyle: 'solid',
        height: 50,
    },
    inputTop: {
        borderTopLeftRadius: 6,
        borderTopRightRadius: 6,
        borderBottomLeftRadius: 0,
        borderBottomRightRadius: 0,
    },
    inputBottom: {
        borderTopLeftRadius: 0,
        borderTopRightRadius: 0,
        borderBottomLeftRadius: 6,
        borderBottomRightRadius: 6,
        marginBottom: 60
    },
```

```javascript
// render
    <View style={[style.inputExample, style.inputTop]}>
    </View>
    <View style={[style.inputExample, style.inputBottom]}>
    </View>
```

The result is as expected, but with artifacts on the corners with radius:0

screen shot 2017-01-13 at 7 30 12 pm

### Solution

Don't know yet

### Additional Information

* React Native version: 0.36.1
* Platform: iOS
* Operating System: MacOS

mattijsf:
Fixes #11897

On devices (not simulator) running iOS 11 there seems to be a bug when drawing selective borders with a corner radius. A "bleeding" line appears on corners that don't have a radius while an adjacent corner has.

It’s most likely a bug with iOS 11 itself as the original code is according to spec and it works fine in an iOS 11 simulator. However it turns out that this patch is a workaround for this problem.

This patch increases the stretchable area to 2pt instead of 1pt after which the glitch disappears. I assume shouldn’t have a negative performance impact as it also enforces resizingMode UIImageResizingModeStretch.

In all fairness, though this workaround seems to have a positive impact on the bug, the exact reason why some cap inset areas are partially stretched across the resizable area remains unknown to me.

Screenshots:

The screenshots below are created using an iPhone 5s Device (e.g. in simulator the bug doesn't manifest):

From left to right:
- Raw resizable image that React Native uses to render the borders. Where I added:
  - Blue lines for debugging to indicate the corners (cap insets) of the image that are supposed to be [unchanged when resized](https://developer.apple.com/documentation/uikit/uiimage/1624102-resizableimagewithcapinsets)
  - Green pixels that indicate stretchable content area
- How the image is resized in the actual rendered view (with the blue lines for debugging)
- How the image is resized in the actual rendered view (without the blue lines for debugging)

Before fix:
before_fix

After fix:
after_fix

As you can see, there is still some bleeding happening but not in the final result.

## Test Plan

See https://github.com/facebook/react-native/issues/11897#issuecomment-385778877

## Release Notes

[IOS] [BUGFIX] [View] - Fixes bleeding corner borders on iOS 11 devices

### DIFF
--- a/React/Views/RCTBorderDrawing.m
+++ b/React/Views/RCTBorderDrawing.m
@@ -225,9 +225,9 @@ static UIImage *RCTGetSolidBorderImage(RCTCornerRadii cornerRadii,
   };
 
   const CGSize size = makeStretchable ? (CGSize){
-    // 1pt for the middle stretchable area along each axis
-    edgeInsets.left + 1 + edgeInsets.right,
-    edgeInsets.top + 1 + edgeInsets.bottom
+    // 2pt for the middle stretchable area along each axis
+    edgeInsets.left + 2 + edgeInsets.right,
+    edgeInsets.top + 2 + edgeInsets.bottom
   } : viewSize;
 
   CGContextRef ctx = RCTUIGraphicsBeginImageContext(size, backgroundColor, hasCornerRadii, drawToEdge);
@@ -385,7 +385,7 @@ static UIImage *RCTGetSolidBorderImage(RCTCornerRadii cornerRadii,
   UIGraphicsEndImageContext();
 
   if (makeStretchable) {
-    image = [image resizableImageWithCapInsets:edgeInsets];
+    image = [image resizableImageWithCapInsets:edgeInsets resizingMode: UIImageResizingModeStretch];
   }
 
   return image;


mattijsf:
@sophiebits Can help reviewing this? Or do you know someone who can?

hramos:
Which devices have you tested this on? Might want to ensure it works as expected on @1x, @2x, and @3x devices.

mattijsf:
I already tested it on a iPhone 5 and 5s, I just also tested on a iPhone 7 plus and iPhone X and it works. However, on these 3x devices the glitch never manifested but at least I can confirm that everything still works as expected on these devices.

I'm not sure how to test 1x though... I might have to get my hands on an iPad mini which runs iOS 9 and has a screen scale of 1x.

<details>
  <summary>Screenshots</summary>
 iPhone 7+:

iphone 7plus_after

iPhone X:

iphone x_after

</details

mattijsf:
Done. Tests are passing again.

jamesreggio:
Hey @mattijsf — @hramos would like us to coordinate to determine which of our fixes to accept. (Mine is over in #21208.)

Are there are test cases you're aware of where your approach does not properly address the issue?

mattijsf:
Sorry @jamesreggio I didn't have the test readily available for re-creating the "intermediate"-state images that I used for my PR. But I think your PR has a proper solution compared to the workaround I used in my PR. 

Nice work! 👍 Good to see that it landed in master now!

PR #19501, opened 2018-05-30
better place for dismiss all button

skv-headless:
This pr makes a little bit simpler to dismiss all warnings (1 click instead of 2)
Sometimes you don't want to use `YellowBox.ignoreWarnings` to remember.

## Test Plan

RNTester sceenshot:

screen shot 2018-05-30 at 09 51 25
Or RNTester -> Native Animation Example -> Force JS Stalls (in the end of the list) also cause warning to test. 

## Release Notes

[GENERAL][ENHANCEMENT][YellowBox] - Move `Dismiss All` yellow box button to another place which makes a little bit simpler to dismiss warnings.

### DIFF
--- a/Libraries/ReactNative/YellowBox.js
+++ b/Libraries/ReactNative/YellowBox.js
@@ -247,7 +247,6 @@ const WarningInspector = ({
   warning,
   stacktraceVisible,
   onDismiss,
-  onDismissAll,
   onMinimize,
   toggleStacktrace,
 }) => {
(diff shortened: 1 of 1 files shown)


hramos:
The `Dismiss All` button is now showing up in cases where there are no visible warnings. It's likely this is failing to account for ignored warnings. The fix might be as simple as checking if `rows` is non-zero before displaying the button.

Can you send a fix?

hramos:
Might be as simple as 

```
{!inspector && rows.length > 0 && (
```

skv-headless:
@hramos tested this case. Nice catch. Fix here https://github.com/facebook/react-native/pull/19564

PR #19510, opened 2018-05-30
[local-cli] Remove explicit cwd config, fix tests

hramos:
This reverts a change from 23741470ca706776815f77c757e1dbddb706889e as it was causing local-cli tests to fail

cc @jeanlauliac

# Test Plan

Run on Circle CI and confirm tests pass.

# Release Notes

[INTERNAL][MINOR][local-cli] - Fix tests.

### DIFF
--- a/local-cli/__mocks__/fs.js
+++ b/local-cli/__mocks__/fs.js
@@ -9,7 +9,7 @@
 
 'use strict';
 
-const fs = new (require('metro-memory-fs'))({cwd: process.cwd});
+const fs = new (require('metro-memory-fs'))();
 
 function setMockFilesystem(object) {
   fs.reset();


hramos:
screen shot 2018-05-30 at 10 50 25 am

jeanlauliac:
> This reverts a change from 2374147 as it was causing local-cli tests to fail

Thank you, can you point out to the error that is happening? Using `process.cwd` is essentially what was the default before, so I'm surprised it caused a new error.

jeanlauliac:
Maybe just something like `{cwd: () => '/'}` would be fine? The point of having to specify the cwd is to enforce people understand what exactly is happening when some code somewhere uses absolute paths. Alternatively, something like `{cwd: () => '/some/folder'}` but then we need to `mkdir` these folders alongside creating a new `memory-fs` instance.

hramos:
@jeanlauliac you're correct, this is unnecessary and not the actual source of the problem.

The failures I'm investigating are outlined in [this comment](https://github.com/facebook/react-native/issues/19308#issuecomment-393242414), and are generally in the form of:

```
EINVAL: `/empty/android/app`: path is invalid because it cannot start with a separator

      21 |   const nested = path.join('android', 'app');
      22 |
    > 23 |   if (fs.existsSync(path.join(folder, nested))) {
         |          ^
      24 |     return nested;
      25 |   }
      26 |
```

The `test_javascript` job has been broken for a while so it's not quite straightforward to determine when the `local-cli` tests stopped passing, but it seems like D7652914 might be related as that's where the old mock fs implementation was removed.

jeanlauliac:
This error is coming from the `metro-memory-fs` module and happens when the Memory/mock FS is set in "win32" platform mode. For Windows paths, it's invalid to have slash as first character: it needs a drive (`C:/`, etc). Granted, the error message should actually say that. I'll open a task to fix that.

Can you try this:

    const fs = new (require('metro-memory-fs'))({cwd: process.cwd, platform: 'posix'});

If that works, then it's a bug in `metro-memory-fs` as it should already default to the `posix` platform.

hramos:
No change:

```
android::findAndroidAppFolder › returns `null` if there is no android app folder

    EINVAL: `/empty/android/app`: path is invalid because it cannot start with a separator

      21 |   const nested = path.join('android', 'app');
      22 |
    > 23 |   if (fs.existsSync(path.join(folder, nested))) {
         |          ^
      24 |     return nested;
      25 |   }
      26 |

      at makeError (node_modules/metro-memory-fs/src/index.js:951:15)
      at MemoryFs._parsePath (node_modules/metro-memory-fs/src/index.js:588:15)
      at MemoryFs._resolve (node_modules/metro-memory-fs/src/index.js:610:27)
      at MemoryFs.existsSync.filePath [as existsSync] (node_modules/metro-memory-fs/src/index.js:311:30)
      at findAndroidAppFolder (local-cli/core/android/findAndroidAppFolder.js:23:10)
      at Object.it (local-cli/core/__tests__/android/findAndroidAppFolder.spec.js:40:12)
```

PR #19531, opened 2018-06-01
android compileSdk 26, build-tools 26.0.2

dulmandakh:
This PR uses android build tools 26.0.2, support library 26.1.0 (with android lifecycle) and setting compileSdkVersion to 26, but leaving minSdkVersion and targetSdkVersion intact.

CI builds just fine, but having error with bundling JS or same error as master.

Sorry, that I messed up with PR git history 😄

@hramos

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -7,6 +7,7 @@
 
 [maven_repositories]
   central = https://repo1.maven.org/maven2
+  google = https://maven.google.com
 
 [alias]
   rntester = //RNTester/android/app:app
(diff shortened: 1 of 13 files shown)


dulmandakh:
once merged, we can add more features to Android that makes developers life easier like custom fonts 😄 and utilize Android Lifecycle for better and robust application, activity and fragment lifecycle handling.

gengjiawen:
I think put them in one pr is more clear, any thought ?

PR #19541, opened 2018-06-02
fix for haste module lookup

fatfatson:
the new haste module lookup method instead of  `providesModule` has some path-related problems in windows, such as 
1. the type of slash
2. the insensitive case of path, especially the drive letter

the best solution is converting all the slash to `/` and the drive letter to lowercase before all follow-up checks.

### DIFF
--- a/jest/hasteImpl.js
+++ b/jest/hasteImpl.js
@@ -12,12 +12,12 @@
 
 const path = require('path');
 
-const ROOT = path.join(__dirname, '..');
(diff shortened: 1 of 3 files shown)


hramos:
Can you share an example of how this fails prior to the fix, and how the fix addresses this? A test plan would be great, as well; ideally, you'd add a test case to cover this.

hramos:
Closing as the author has not signed the CLA and there has been no updates in a while.

PR #19548, opened 2018-06-03
Remove unused include.

sayrer:
`LayoutableShadowNode.cpp` includes `"LayoutableShadowNode.h"` as well as `<fabric/core/LayoutContext.h>`. In turn, `LayoutContext.h` then includes `<fabric/core/LayoutableShadowNode.h>`. `LayoutContext.h` doesn't actually require `LayoutableShadowNode.h`, but this unnecessary inclusion can cause duplicate definition errors if the two include paths don't map to exactly the same file. This patch removes the unnecessary include.

## Test Plan

The CI's build system should cover the testing needed.

## Release Notes

[INTERNAL] [MINOR] [fabric] - Remove an unnecessary include in fabric/core/layout.

### DIFF
--- a/ReactCommon/fabric/core/layout/LayoutContext.h
+++ b/ReactCommon/fabric/core/layout/LayoutContext.h
@@ -9,7 +9,6 @@
 
 #include <unordered_set>
 
-#include <fabric/core/LayoutableShadowNode.h>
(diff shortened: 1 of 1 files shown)


shergin:
@sayrer Congrats! This is the first known OSS contribution for React Native Fabric effort. :)

PR #19571, opened 2018-06-05
Change error message on interpolation

osdnk:
Change message in Animated.Interpolation to "inputRange must be monotonically non-decreasing" as it's allowed to give the same x's like in the test [example](https://github.com/facebook/react-native/blob/4435f087713e1d0ac3639e3b3196d71c6402898e/Libraries/Animated/src/__tests__/Interpolation-test.js#L71)

## Test Plan

Simply giving improper value of interpolation input

[GENERAL] [MINOR] [AnimatedInterpolation.js] - Change error message on interpolation improper range error


TheSavior:
Thanks for contributing! Improving our documentation and error messages is incredibly valuable work. Hopefully this will help make it more clear for other people who run into it.

PR #19612, opened 2018-06-07
[FlatList] [VirtualizedList] Fix issue with nested horizontal VirtualizedLists

### LINKED ISSUE #19452: Nesting two horizontal Flatlist's causes the second Flatlist to display content vertically

When you nest two horizontal FlatLists the second FlatList displays content vertically and not horizontally.

See here: https://gist.github.com/hgale/672f5a3b1782ffecb14e3f5a246de9ca

screen shot 2018-05-25 at 1 59 10 pm

**Other things to note**

- If you use the flag legacyImplementation={true}, the issue does not occur.
- If you make the second list a ListView the implementation does not occur.
 
It should be noted that the legacyImplementation flag doesn't support the scrollToIndex method, which I want to use with the component I am building.

I also really don't want to have to use ListView and then swap it out for a FlatList once this issue is fixed.

- If you open node_modules/react-native/Libraries/Lists/VirtualizedList.js and edit the following code:

 https://gist.github.com/hgale/e92e8fff418c92f486bb68ef93e7c6c3

The issue goes away. 

## Environment

Environment:
  OS: macOS High Sierra 10.13.4
  Node: 9.10.1
  Yarn: 1.6.0
  npm: 5.6.0
  Watchman: 4.9.0

  react: ^16.3.0-alpha.1 => 16.3.1

Also, this has been tested on:

"react-native": "0.54.2"

## Steps to Reproduce

See above description and code sample. This bug occurs on iOS, I have not tested it on Android.

Here is an example implementation:

 https://gist.github.com/hgale/410c1a0f596f594a2c34f6790896e64a

## Actual Behavior

The second Flatlist should display content horizontally not vertically.

screen shot 2018-05-25 at 2 05 28 pm
screen shot 2018-05-25 at 2 05 34 pm

hgale:
This fixes #19452.  When you nest two horizontal FlatLists the second FlatList displays content vertically and not horizontally.

I created a new gist that repo's this issue just using VirtualizedLists see [here](https://gist.github.com/hgale/a7db66dd56670047ffcde8ed573ec2d8).

Before my fix the above example looks like this: 

before

After it looks like this:

after
 
Other things to note: this diff reverts part of a change made by @sahrens see [here](https://github.com/facebook/react-native/commit/63f7efcd32dcfc5ff0bb742aae8d4fc35665cbbd). If he has a better suggestion as to how to fix this issue I would love to hear about it.

@bvaughn reviewed original diff by @sahrens and might have some more context on this.

## Test Plan

1. Create App using the following: `react-native init TestNestedLists`
2. Replace App.js with [this](https://gist.github.com/hgale/a7db66dd56670047ffcde8ed573ec2d8).
3. Run.

App should look like after screenshot.

## Release Notes

[GENERAL] [BUGFIX] [VirtualizedList] - Fix nested horizontal VirtualizedList/Flatlists.

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1013,10 +1013,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
 
   _defaultRenderScrollComponent = props => {
     const onRefresh = props.onRefresh;
-    if (this._isNestedWithSameOrientation()) {
-      // $FlowFixMe - Typing ReactNativeComponent revealed errors
-      return <View {...props} />;
-    } else if (onRefresh) {
+    if (onRefresh) {
       invariant(
         typeof props.refreshing === 'boolean',
         '`refreshing` prop must be set as a boolean in order to use `onRefresh`, but got `' +
--- a/Libraries/Lists/__tests__/VirtualizedList-test.js
+++ b/Libraries/Lists/__tests__/VirtualizedList-test.js
@@ -160,6 +160,31 @@ describe('VirtualizedList', () => {
     expect(component).toMatchSnapshot();
   });
 
+  it('handles nested horizontal lists', () => {
+    const component = ReactTestRenderer.create(
+      <VirtualizedList
+        data={[1, 2, 3]}
+        getItem={(data, index) => data[index]}
+        getItemCount={data => data.length}
+        keyExtractor={dataNumer => dataNumer.toString()}
+        horizontal={true}
+        renderItem={outerInfo => (
+          <VirtualizedList
+            data={['A', 'B', 'C']}
+            getItem={(data, index) => data[index]}
+            getItemCount={data => data.length}        
+            keyExtractor={imageId => imageId}
+            horizontal={true}
+            renderItem={({ item: imageId}) => {
+              return <item title={imageId} />;
+            }}
+          />
+        )}
+      />,
+    );
+    expect(component).toMatchSnapshot();
+  });
+
   it('returns the viewableItems correctly in the onViewableItemsChanged callback after changing the data', () => {
     const ITEM_HEIGHT = 800;
     let data = [{key: 'i1'}, {key: 'i2'}, {key: 'i3'}];
--- a/Libraries/Lists/__tests__/VirtualizedList-test.js
+++ b/Libraries/Lists/__tests__/VirtualizedList-test.js
@@ -172,10 +172,10 @@ describe('VirtualizedList', () => {
           <VirtualizedList
             data={['A', 'B', 'C']}
             getItem={(data, index) => data[index]}
-            getItemCount={data => data.length}        
+            getItemCount={data => data.length}
             keyExtractor={imageId => imageId}
             horizontal={true}
-            renderItem={({ item: imageId}) => {
+            renderItem={({item: imageId}) => {
               return <item title={imageId} />;
             }}
           />
--- a/Libraries/Lists/__tests__/__snapshots__/VirtualizedList-test.js.snap
+++ b/Libraries/Lists/__tests__/__snapshots__/VirtualizedList-test.js.snap
@@ -1,5 +1,303 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`VirtualizedList handles nested horizontal lists 1`] = `
+<RCTScrollView
+  data={
+    Array [
+      1,
+      2,
+      3,
+    ]
+  }
+  disableVirtualization={false}
+  getItem={[Function]}
+  getItemCount={[Function]}
+  horizontal={true}
+  initialNumToRender={10}
+  keyExtractor={[Function]}
+  maxToRenderPerBatch={10}
+  onContentSizeChange={[Function]}
+  onEndReachedThreshold={2}
+  onLayout={[Function]}
+  onMomentumScrollEnd={[Function]}
+  onScroll={[Function]}
+  onScrollBeginDrag={[Function]}
+  onScrollEndDrag={[Function]}
+  renderItem={[Function]}
+  scrollEventThrottle={50}
+  stickyHeaderIndices={Array []}
+  updateCellsBatchingPeriod={50}
+  windowSize={21}
+>
+  <View>
+    <View
+      onLayout={[Function]}
+      style={
+        Array [
+          Object {
+            "flexDirection": "row",
+          },
+          null,
+        ]
+      }
+    >
+      <RCTScrollView
+        data={
+          Array [
+            "A",
+            "B",
+            "C",
+          ]
+        }
+        disableVirtualization={false}
+        getItem={[Function]}
+        getItemCount={[Function]}
+        horizontal={true}
+        initialNumToRender={10}
+        keyExtractor={[Function]}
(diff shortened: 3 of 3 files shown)


Libraries/Lists/VirtualizedList.js
```diff
@@ -1013,10 +1013,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
 
   _defaultRenderScrollComponent = props => {
     const onRefresh = props.onRefresh;
-    if (this._isNestedWithSameOrientation()) {
-      // $FlowFixMe - Typing ReactNativeComponent revealed errors
-      return <View {...props} />;
```
sahrens: I think you just need to add a style with `flexDirection: 'row'` when horizontal is true.

hgale: Thanks for the feedback @sahrens! I'll give that a go and report bak if it works.

hgale:
When I add the style with `flexDirection: 'row'` like so:

```
    if (this._isNestedWithSameOrientation()) {
      return <View style={{flexDirection: 'row'}} {...props} />;
    } 

```

The following occurs:

screen shot 2018-06-08 at 8 13 28 pm

It looks like all of the views in my nested example VirtualizedLists render on top of each other. 

The code for the VirtualizedLists example looks like this:


```
          <VirtualizedList
              data={[1, 2, 3]}
              getItemCount={data => data.length}
              getItem={(data, index) => data[index]}
              horizontal={true}
              debug={true}
              getItemLayout={(data, index) => (
                  {length: 200, offset: 200 * index, index}
                )}                
              keyExtractor={pageNumber => pageNumber.toString()}
              renderItem={({ item: pageNumber }) => (
              <View style={styles.page}>
                  <Text>{`page ${pageNumber}`}</Text>
                  <VirtualizedList
                        data={imageIds}
                        getItemCount={data => data.length}
                        getItem={(data, index) => data[index]}                        
                        getItemLayout={(data, index) => (
                            {length: 200, offset: 200 * index, index}
                        )}
                        listKey={Math.random().toString()}
                        debug={true}                        
                        initialNumToRender={3}
                        horizontal={true}
                        extraData={'asdasd'}
                        keyExtractor={imageId => imageId}
                        renderItem={({ item: imageId}) => ( 
                            <View style={styles.pageStyleOne}>
                        <Text>{`image ${imageId}`}</Text> 
                            </View>
                        )}
                    />                  
              </View>
              )}
          />
```

Was that what you had in mind when you suggested I add a style with `flexDirection: 'row'` to the View? Note, I checked props to see if it contained a style property, it did not, so even if I did `[props.style, {flexDirection: 'row'}]`, it wouldn't have made a difference in this case.

> I don't think this is the right fix - we don't want to be rendering nested ScrollViews in the same direction. It sounds like all you need to do is change the flex direction to row when horizontal is true.

@sahrens could you explain why you don't want to render nested ScrollViews in the same direction? I come from a native iOS background and this is supported out of the box and is explicitly mentioned in the UIScrollView [docs](https://developer.apple.com/library/archive/documentation/WindowsViews/Conceptual/UIScrollView_pg/NestedScrollViews/NestedScrollViews.html)

_Same direction scrolling occurs when a UIScrollView that is a subview of a UIScrollView both scroll in the same direction. This is shown in the left image in Figure 6-1._

I also cooked up a little example app in Swift that nests two horizontal UIScrollViews. that behaves the way I would expect this to work:

https://www.dropbox.com/s/2ttfkxmdqw9kxr1/TestNestedScrollViews.zip?dl=0

If this use case isn't supported right now, should I make this an invariant and update the docs so that others don't get into this situation?  Thanks for all your help.

sahrens:
Hmm, seems like that should work. Can you please include the styles you're using in your example? It's nearly impossible to debug layout issues without them.

And yes it's possible for nested scrolling to work nicely in iOS, but it can cause unexpected behavior that is inconsistent on other platforms (e.g. Android), and it is unnecessary overhead to render unneeded `UIScrollView`s (we can completely optimize away many `View` components in native code).

hgale:
@sahrens Sorry, I included them in the original gist linked above. You can see them [here](https://gist.github.com/hgale/a7db66dd56670047ffcde8ed573ec2d8).  If you create a new app using:` react-native init TestNestedLists` and replace app.js with the contents of that gist you should be able to see the example. 

I tried getting this working with expo so I could link a snack but I couldn't repo it in that for some reason. Thanks for taking the time to look at this.

sahrens:
Ah sorry, didn't realize it was all still the same.

I think at least part of the problem is your styles.page which wraps your inner VirtualizedList and has width set to the width of the window. Since the items inside it also have a width the size of the window via styles.pageOne, that might be messing things up.

hgale:
"I think at least part of the problem is your styles.page which wraps your inner VirtualizedList and has width set to the width of the window. Since the items inside it also have a width the size of the window via styles.pageOne, that might be messing things up."

When I remove the width field from styles.page in the gist I posted [here](https://gist.github.com/hgale/a7db66dd56670047ffcde8ed573ec2d8) to be like this:

```
    page: {
        // width: Dimensions.get('window').width,
        borderWidth: 2,
        borderColor: 'blue',
        marginTop: 20,
        alignItems: 'center',
    },
```
And when I  alter the VirtualizedList.js to use the suggested flexDirection fix:

```
    if (this._isNestedWithSameOrientation()) {
      return <View style={{flexDirection: 'row'}} {...props} />;
    } else  if (props.onRefresh) {
```

I get this result:

ezgif-1-3591756490

Is this what you would expect to happen or do you think their are other issues with the layout code I am using?  I.e given the fact I am using nested VirtualizedLists should all of the pages in the nested horizontal VirtualizedList render at once as one big View? This is not what I would have intuitively expected to happen. 

If this is expected behavior do you think it would be worthwhile for me to update the docs to document this?

> And yes it's possible for nested scrolling to work nicely in iOS, but it can cause unexpected behavior that is inconsistent on other platforms (e.g. Android), and it is unnecessary overhead to render unneeded UIScrollViews (we can completely optimize away many View components in native code).

Could you elaborate a bit more on this inconsistent behavior? Are you referring to how the current VirtualizedList implementation works or are you talking about how nested lists work on native iOS compared to native Android?

It's been a while since I have done native Android development but aren't nested ScrollView's a supported feature?  While googling around I came across this [article](http://www.devexchanges.info/2016/07/nested-scroll-views-in-android.html) describing how to use Android's [NestedScrollView](https://developer.android.com/reference/android/support/v4/widget/NestedScrollView) component.

I've included the gif from the article below:

16drupn

The demo shows nested vertical views where one of the cells with some text in it is scrollable.
This looks very similar behavior to the scenario outlined in the picture from Apple's UIScrollView docs with regard to nested UIScrollViews:

scrollviews

arivanandan:
This is still an issue on Android. If someone could point me in the right direction, maybe I could take a look.

cpojer:
I'm gonna go ahead and close this PR as we haven't identified a proper solution yet and it probably makes sense to discuss it here or in an issue first before proceeding back to the PR stage.

FrickHazard:
Ran into this issue Recently. A fix would be nice.

beingArkReal:
> Ran into this issue Recently. A fix would be nice.

is this actually affecting any performance issue?

Kinark:
Any news on this?

Issue #19260, opened 2018-05-15
Text label backgroundColor has alpha display error


helele90:
- I have reviewed the [documentation](https://facebook.github.io/react-native)
- I have searched [existing issues](https://github.com/facebook/react-native/issues)
- I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

### code
```
import React, { Component } from 'react';
import { Text, View, StyleSheet } from 'react-native';

export default class App extends Component {
  render() {
    return (
      <View style={styles.container}>
      <Text style={[styles.text, {backgroundColor: '#f30301'}]}>backgroundColor: #f30301</Text>
      <Text style={[styles.text, {backgroundColor: '#f3030150'}]}>backgroundColor: #f3030150</Text>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    justifyContent: 'center',
    alignItems: 'center',
    flex: 1,
  },
  text: {
    marginTop: 10,
    height: 40,
    width: 200,
    fontSize: 13,
    textAlign: 'center',
    color: 'white',
  },
});

```
### Image
![9d8b85aea74007e56e808264fca008f7](https://user-images.githubusercontent.com/4125469/40091469-879e8fae-58ea-11e8-918c-a158d3c9d1e4.png)

### Template
template: https://snack.expo.io/H1DvLbYRz

if Text style backgroundColor has alpha, the background display error.
## Environment
Environment:
  OS: macOS High Sierra 10.13.4
  Node: 8.11.1
  Yarn: 1.6.0
  npm: 6.0.1
  Watchman: 4.9.0
  Xcode: Xcode 9.3 Build version 9E145
  Android Studio: Not Found

Packages: (wanted => installed)
  react: ^16.3.2 => 16.3.2
  react-native: ^0.55.4 => 0.55.4

hramos:
Can you please fill out the template?

helele90:
@hramos I made a expo demo.

PR #19614, opened 2018-06-08
Text BackgroundColor bug

helele90:
The background of the Text does not need to use NSBackgroundColorAttributeName. When the background color contains alpha, it will conflict with the UIView background color, cause a bug.
#19260 

## Test Plan

simulator screen shot - iphone 8 - 2018-06-08 at 10 51 25

## Release Notes

[GENERAL] [BUGFIX] [Libraries/Text/RCTTextAttributes.m] - Fix Text background has alpha


cpojer:
Closing because of inactivity.

PR #16792, opened 2017-11-11
Migrate RCTWebView to WKWebView.

bowerman0:
## Motivation

https://github.com/facebook/react-native/issues/321 - Use WKWebView instead of UIWebView, because WKWebView is faster, consumes less memory, and is more stable.

This is meant to be a drop-in replacement for the existing UIWebView-based RCTWebView.

## Test Plan

Used a web view loosely based on [RNTester's WebViewExample.js](https://github.com/facebook/react-native/blob/master/RNTester/js/WebViewExample.js), and tested in an ejected AwesomeProject. Browsed around the internet.

## Release Notes

[IOS][ENHANCEMENT][RCTWebView] - Migrate RCTWebView to use WKWebView instead of UIWebView.

PR #19617, opened 2018-06-08
iOS: WebView stays in loading state

### LINKED ISSUE #5974: WebView not loading source / Staying in loading state
Hello,

seems like there Is an issue with the WebView not loading its source.
I have followed the basic example on the DOCS and tried to load google.com and plain html. Either one didnt work.
RN V0.19. 
Staying in loading state.

Sample code (nested component):

```

'use strict';
import React, {
  AppRegistry,
  Component,
  StyleSheet,
  Text,
  View,
  Image,
  TextInput,
  Button,
  TouchableHighlight,
  TouchableOpacity,
  StatusBarIOS,
  WebView
} from 'react-native';

StatusBarIOS.setStyle(1);

import Dimensions from 'Dimensions';
var width = Dimensions.get('window').width;
var height = Dimensions.get('window').height;

var CStyles = require('./CStyle');

class TestComp extends Component {
  constructor(){
    super();

    this.state = {
      url: 'https://www.google.at/'
    }
  }
  render() {
    return (
      <View style={{flex: 1}}>
          <WebView
            ref="webview"
            style={{width: width}}
            automaticallyAdjustContentInsets={false}
            source={{uri: this.state.url}}
            javaScriptEnabled={true}
            domStorageEnabled={true}
            decelerationRate="normal"
            startInLoadingState={true}
          />
      </View>
    );
  }
}
module.exports = TestComp;
```

johncblandii:
## MOTIVATION

Fixes #5974

`WebView` does not load when `startInLoadingState` is `true` on iOS; Android works perfectly fine. The issue is a hidden `UIWebView` never exits the loading state.

## Test Plan

React Native: 0.52.2
iOS: 11.3

I added a simple logging statement and it logs:
```
2018-06-07 22:32:34.481172-0500 HarvestPoint[52403:2799087] LOADING:: TRUE, TRUE
```

That's the only log it gets when `startInLoadingState={true}` on JS. If I turn that off, you'll see a list of `TRUE, TRUE` then a `FALSE, FALSE` which then renders the UI.

The fix sets the default height to `1` which actually loads the UI. Once there is a height, the iOS code properly triggers the `onLoadingFinish` callback.

## Related PRs

#16792: Migrate RCTWebView to WKWebView

## Release Notes

[IOS] [BUGFIX] [WebView] - Set a default height on a hidden `WebView` to avoid `UIWebView` problem

### DIFF
--- a/Libraries/Components/WebView/WebView.ios.js
+++ b/Libraries/Components/WebView/WebView.ios.js
@@ -678,7 +678,7 @@ const styles = StyleSheet.create({
     marginBottom: 10,
   },
   hidden: {
-    height: 0,
+    height: 1, // UIWebView does not finish loading when 100% hidden; see https://github.com/facebook/react-native/issues/5974
     flex: 0, // disable 'flex:1' when hiding a View
   },
   loadingView: {


johncblandii:
This is the error. Any clue as to whether this is intermittent or really broken with this small change?

```
error: bundling failed: TypeError: Cannot read property 'bindings' of null
    at Scope.moveBindingTo (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/scope/index.js:978:13)
    at BlockScoping.updateScopeInfo (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/babel-plugin-transform-es2015-block-scoping/lib/index.js:364:17)
    at BlockScoping.run (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/babel-plugin-transform-es2015-block-scoping/lib/index.js:330:12)
    at PluginPass.BlockStatementSwitchStatementProgram (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/babel-plugin-transform-es2015-block-scoping/lib/index.js:70:24)
    at newFn (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/visitors.js:237:21)
    at NodePath._call (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/path/context.js:65:20)
    at NodePath.call (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/path/context.js:100:12)
    at TraversalContext.visitQueue (/private/tmp/react-native-NhQrzVYk/EndToEndTest/node_modules/@babel/traverse/lib/context.js:142:16)
```

mailyokesh:
I have this issue as well where as when i try to load a webpage in ios all i see is a white screen. Android the page loads fine. I have links for 4 pages so irrespective of the order the first 3 will work and the 4 th one will show just the white screen. Any way this change will fix this? I dont see any errors being returned by the webview but all IOS webview shows is a white screen.

hramos:

RSNara:
This looks interesting, although I'm not exactly sure if this hack is the right solution.

We're moving WebViews to [react-native-community/react-native-webview](https://github.com/react-native-community/react-native-webview) as a part of an effort to reduce the surface area of React Native. Would you mind launching this PR/issue over at the community repository instead?

PR #19619, opened 2018-06-08
Line height

helele90:
If Text style has lineHeight, The English, chinese, japanese, number is not on a same baseline.

## Related Notes
[GENERAL] [BUGFIX] [Libraries/Text/RCTTextShadowView.m] - Fix Text background has alpha


ikesyo:
This should be superseded by #19653.

cpojer:
Thanks for your PR. This should already be fixed on master, yay!

PR #19627, opened 2018-06-08
Add onInsetsChange property to SafeAreaView

clmntcrl:
The aim of this PR is to avoid the use of magic numbers to handle safe area inset size.

## Test Plan

Test is implemented in RNTester (see below the results for iPhone X and iPhone 8, running iOS 11 of course 🤗).

iPhone X                             |  iPhone 8
:-------------------------:|:-------------------------:
 | 

## Related PRs

[react-native-website#401](https://github.com/facebook/react-native-website/pull/401)

## Release Notes

[IOS] [ENHANCEMENT] [SafeAreaView] - Add onInsetsChange to SafeAreaView


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  :exclamation: Big PR - <i>This PR is extremely unlikely to get reviewed because it touches 31671 lines.</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

hramos:
Some Flow issues are preventing this from landing...

clmntcrl:
Sorry, I'm not used to work with Flow so I'm not sure this fix anything. However reading Flow doc it seems better using `children?: ?React.Node` in this case. Hope this fix Flow issues.

clmntcrl:
@hramos does this fix the Flow issues ?

Libraries/Components/SafeAreaView/SafeAreaView.ios.js
```diff
@@ -13,13 +13,16 @@ const React = require('React');
...
 
 const RCTSafeAreaView = requireNativeComponent('RCTSafeAreaView');
 
-type Props = ViewProps & {
-  children: any,
-};
+type Props = $ReadOnly<{|
+  ...ViewProps,
+  children?: ?React.Node,
+  onInsetsChange?: ?Function,
```
janicduplessis: We could use a better flow type here, see 
https://github.com/clmntcrl/react-native/blob/d9568123d5f5dc491fd321e2ed463bffb13868ec/Libraries/Components/Switch/Switch.js#L66 for example.

clmntcrl:
@shergin Indeed, the same result can be achieved by using a `View` inside a `SafeAreaView` and then compute safe area insets by comparing `SafeAreView` size to the `View` size and position (obtained by using `onLayout`).

The only difference is that `onInsetsChange' purpose is only to provide safe area insets size, so it doesn't need calculation and is only call when insets changes.
Is this sufficient? I don't know but I think it will bring more visibility to how to get the size of these insets and maybe avoid a lot of implementation which rely on magic numbers.

shergin:
All that I said here https://github.com/facebook/react-native/pull/20999#issuecomment-476742284 is also applicable to this PR.

And because it's relatively simple, I am okay to land this.

shergin:
(But we need to resolve conflicts first.)

Libraries/Components/SafeAreaView/SafeAreaView.js
```diff
@@ -33,7 +35,7 @@ let exported;
 if (Platform.OS === 'android') {
   exported = class SafeAreaView extends React.Component<Props> {
     render(): React.Node {
-      const {emulateUnlessSupported, ...props} = this.props;
+      const {emulateUnlessSupported, onInsetsChange, ...props} = this.props;
```
cpojer: You are destructuring the props here and then not doing anything with `onInsetsChange`. Do you mind patching up the code to make it work properly and making sure the test plan and example show that this is working properly?

cpojer:
This looks mostly good to me but I left a comment inline where I think you missed hooking up the event properly unless I'm missing something.

Please also rebase it and resolve the last merge conflict before I can land this change :) Thank you!

zhigang1992:
Do we need to add a no-ops method on android? Or does it apply to android as well?

cpojer:
Unfortunately it looks like this PR is now stale. I'm happy to merge this if somebody makes a version that works on master. I'll close this one but feel free to send a new PR with the rebased and fully tested version.

PR #19632, opened 2018-06-09
Add ability to customise WebSocket OkHttp client on Android

### LINKED ISSUE #18920: WebSocket https connection with self signed certificate on Android
- [] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

I tried to connect using websocket to a server using https with a self signed certificate. I customized the OkHttpClient using the factory reintroduced in the 0.54 version and it works as expected for https requests but it doesn't work for websocket connection. By looking at the WebSocketModule.java I found that in the connect method the OkHttpClient is created as a default OkHttpClient instead of using the OkHttpClientProvider.createClient() method that, from 0.54 version, allow us to customize the OkHttpClient.

## Environment

Environment:
 OS: macOS Sierra 10.12.5
 Node: 9.9.0
 Yarn: 1.5.1
 npm: 5.6.0
 Watchman: 4.9.0
 Xcode: Xcode 8.3.2 Build version 8E2002

 react: ^16.2.0 => 16.2.0
 react-native: ^0.54.0 => 0.54.4

## Steps to Reproduce

Try to establish a secure web socket connection to a web server with an https certificate by using a custom OkHttpClientFactory

## Expected Behavior

Connection established

## Actual Behavior

This error occur 

> java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.

tolbkni:
## Motivation

Summary:
Prior to 22efd95, users could customise the OkHttp client used by React Native on Android by calling setOkHttpClientFactory in OkHttpClientProvider.

This functionality has a variety of legitimate applications from changing connection timeouts or pool size to Stetho integration. In this commit also provide the ability for WebSocket.

Closes #18920

## Test Plan

Create React Native application and set a custom factory in the constructor, e.g. OkHttpClientProvider.setOkHttpClientFactory(new CustomNetworkModule());

Where a custom factory would look like:

```
class CustomNetworkModule implements OkHttpClientFactory {
    public OkHttpClient createNewNetworkModuleClient() {
        return new OkHttpClient.Builder().build();
    }
}
```

## Related PRs

None

## Release Notes

[ANDROID] [MINOR] [WebSocket] - Also use the custom factory set in OkHttpClientProvider for customising the WebSocket OkHttp client used by React Native

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
@@ -22,6 +22,8 @@ import com.facebook.react.common.ReactConstants;
 import com.facebook.react.module.annotations.ReactModule;
 import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.facebook.react.modules.network.ForwardingCookieHandler;
+import com.facebook.react.modules.network.OkHttpClientProvider;
+
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -84,7 +86,7 @@ public final class WebSocketModule extends ReactContextBaseJavaModule {
     @Nullable final ReadableArray protocols,
     @Nullable final ReadableMap options,
     final int id) {
-    OkHttpClient client = new OkHttpClient.Builder()
+    OkHttpClient client = OkHttpClientProvider.getOkHttpClient().newBuilder()
       .connectTimeout(10, TimeUnit.SECONDS)
       .writeTimeout(10, TimeUnit.SECONDS)
       .readTimeout(0, TimeUnit.MINUTES) // Disable timeouts for read


hey99xx:
Do you think there's value moving creation of the OkHttp client into the constructor of `WebSocketModule`? Right now it looks to me like it creates a new client for every new websocket connection, which looks expensive.

tolbkni:
Sounds good idea, here I just change the way to create OkHttp client, to provider ability for WebSocket. And we can create another PR to improve.

tolbkni:
Please review and merge @hramos @mdvacca.

boliveira:
Hi, I need this fix to be available so we can use certificate pinning on websocket connections!!! Please, fix this asap facebook!!!

ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
```diff
@@ -84,11 +86,7 @@ public void connect(
     @Nullable final ReadableArray protocols,
     @Nullable final ReadableMap options,
     final int id) {
-    OkHttpClient client = new OkHttpClient.Builder()
```
hramos: Would the default client provided by `OkHttpClientProvider.createClient()` have similar defaults to the ones defined below?

tolbkni: The one defined before
```java
OkHttpClient client = new OkHttpClient.Builder()
        .connectTimeout(10, TimeUnit.SECONDS)
        .writeTimeout(10, TimeUnit.SECONDS)
        .readTimeout(0, TimeUnit.MINUTES) // Disable timeouts for read
        .build();
```

and the default client provided by `OkHttpClientProvider.createClient()`
```java
// No timeouts by default
OkHttpClient client = new OkHttpClient.Builder()
        .connectTimeout(0, TimeUnit.MILLISECONDS)
        .readTimeout(0, TimeUnit.MILLISECONDS)
        .writeTimeout(0, TimeUnit.MILLISECONDS)
        .cookieJar(new ReactCookieJarContainer())
        .build();
```

they are a little different.

cpojer:
It doesn't seem like this is the right change as it changes the defaults. Can you make it so the default behavior stays?

tolbkni:
@cpojer @hramos I have updated the code to stay the default behavior.

boliveira:
Hello,

I expected that by release 0.59 this would land. Why this hasn't landed yet? This makes upgrading to newer versions of react native harder for us as we have implemented it in a fork.

hramos:
There's a lot of work going on to get 0.59 and future releases ready. This one missed the 0.59 release, but the good news is that 0.60 and 0.61 are around the corner. Please avoid adding unnecessary pressure to maintainers! We'd appreciate it.

boliveira:
Hi Hector, I am sorry for the language used. I love react native and the work you do.
It's just that without this change I am not able to upgrade react native to newer versions without doing a lot of work and as time goes on, more and more versions keep getting released and I fear I'll not be able to keep the pace.

cpojer:
It seems like this diff got stuck when I tried to land it :(

cpojer:
I've been trying to land this for a while at FB and it fails in the apps that rely on this. Does this change come with any actual behavior changes that are not obvious?

boliveira:
Maybe because the defaults are different for Websockets and fetchs.
Maybe a different implementation should be chosen, that would differentiate if the okhttpclient to be created is for use on a websocket or a normal fetch request... just a suggestion...

tolbkni:
So what can I do? The default way of `OkHttpClientProvider.java` to create `OkHttpClient.Builder` is
```java
// No timeouts by default
OkHttpClient.Builder client = new OkHttpClient.Builder()
  .connectTimeout(0, TimeUnit.MILLISECONDS)
  .readTimeout(0, TimeUnit.MILLISECONDS)
  .writeTimeout(0, TimeUnit.MILLISECONDS)
  .cookieJar(new ReactCookieJarContainer());

try {
  Class ConscryptProvider = Class.forName("org.conscrypt.OpenSSLProvider");
  Security.insertProviderAt((Provider) ConscryptProvider.newInstance(), 1);
  return client;
} catch (Exception e) {
  if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.KITKAT) {
    try {
      client.sslSocketFactory(new TLSSocketFactory());

      ConnectionSpec cs = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS)
              .tlsVersions(TlsVersion.TLS_1_2)
              .build();

      List<ConnectionSpec> specs = new ArrayList<>();
      specs.add(cs);
      specs.add(ConnectionSpec.COMPATIBLE_TLS);
      specs.add(ConnectionSpec.CLEARTEXT);

      client.connectionSpecs(specs);
    } catch (Exception exc) {
      FLog.e("OkHttpClientProvider", "Error while enabling TLS 1.2", exc);
    }
  }
  return client;
}
```

cpojer:
I'm going to close this as we cannot land it without breaking changes as is. If you can find a way to change this PR not to cause breaking changes by keeping the existing behavior and allowing it to be modified I'm happy to reopen it (or feel free to send a new PR).

ianbrind:
> I've been trying to land this for a while at FB and it fails in the apps that rely on this. Does this change come with any actual behavior changes that are not obvious?

Do you have any more information on what's causing it to crash?

sxlvalue:
Oops, I also encountered the same problem and couldn't find a solution. There is a problem with the basic network connection components of React Native. Is there any way to fix and solve this problem?

PR #13014, opened 2017-03-19
Add custom eslint plugin to check fbjs imports

janicduplessis:
Adds a custom ESLint plugin in the repo to be able to easily create rules and adds a rule to check for haste import of fbjs modules which often breaks OSS since it works on fb infra. Fun little weekend project to learn how to make eslint rules :)

Will need someone from fb to make sure it works properly internally.

**Test plan**
Checked that it works in atom and when running `npm run lint -- --fix`

giphy

PR #19633, opened 2018-06-09
Disallow requiring from invariant/warning

TheSavior:
In this repo we have to require from fbjs/lib/invariant and issues occur when we require directly from invariant/warning. Flow doesn't complain about those requires right now because we have a transitive dependency on invariant and warning causing them to happen to be at the root of node_modules. 

This change blacklists requiring directly from invariant and warning using Flow. @janicduplessis opened a pull request to do this with ESLint back in 2017: https://github.com/facebook/react-native/pull/13014

Test Plan:
I changed a require to point to invariant instead of fbjs/lib/invariant, got the warning:
screen shot 2018-06-08 at 9 39 19 pm

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -25,6 +25,11 @@
 ; Ignore metro
 .*/node_modules/metro/.*
 
+; These should not be required directly
+; require from fbjs/lib instead: require('fbjs/lib/invariant')
+.*/node_modules/invariant/.*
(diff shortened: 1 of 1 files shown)


TheSavior:
Whoops. Didn't mean to push the branch to this repo. Closing in favor of https://github.com/facebook/react-native/pull/19634

Issue #18916, opened 2018-04-18
TextInput has performance issue in ^0.55


armenbadalyan:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [x] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

## Environment
Environment:
  OS: macOS High Sierra 10.13.3
  Node: 8.8.0
  Yarn: 1.3.2
  npm: 5.4.2
  Watchman: 4.7.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: Not Found

Packages: (wanted => installed)
  react: ^16.0.0 => 16.3.2
  react-native: ^0.55.0 => 0.55.3

## Steps to Reproduce

1. Create a new project with react-native init
2. Replace App.js content with the following code:

```
import React, { Component } from 'react';
import {    
    TextInput,    
    View
} from 'react-native';

type Props = {};
export default class App extends Component<Props> {
    state = {
        text: ''
    }

    onChangeText = (event) => {
        const { text } = event.nativeEvent
        this.setState({ text })
    }

    render() {
        return (<View>
            <TextInput value={this.state.text} onChange={this.onChangeText} />
        </View>);
    }
}
```
3. Run the app on Android device with react-native run-android
4. Type and clear some text into the TextInput several times

## Expected Behavior

Input should perform the same after several type/clear cycles.

## Actual Behavior

Input becomes laggy after a couple of type/clear cycles.

ilonashub:
Having the same issue exactly... no need to clear the text even - if I type 3-4 lines it starts to become very laggy.

wuyunqiang:
may be you can try like that:
```
class Input extends Component{
    constructor(){
        super();
        this.amount=0;
    }
    componentWillReceiveProps(nextProps) {
        if(this.props.amount!=nextProps.amount){
            this._textInput&&this._textInput.setNativeProps({text: nextProps.amount});//重点
        }
    }
    setAmout = (text)=>{
        this.amount = text;
    };

    getAmout = ()=>{
        this.props.getAmout&&this.props.getAmout(this.amount);
    };

    render(){
        return (<View style={styles.inputStyle}>
          <TextInput
                    ref={component => this._textInput = component}
                    numberOfLines={1}
                    underlineColorAndroid={'transparent'}
                    style={{padding:0,backgroundColor:'white',width:SCALE(400)}}
[...]

PR #17398, opened 2017-12-31
Implement letterSpacing on Android >= 5.0

motiz88:
## Motivation

`letterSpacing` is completely missing from RN Android at the moment.

I've reviewed the `letterSpacing` implementations in #13199, #13877 and #16801 (that all seem to have stalled) and managed to put together an improved one based on #13199, updated to merge cleanly post https://github.com/facebook/react-native/commit/6114f863c3aed826355530bcf404ee5aed2f927b, that resolves the [issues](https://github.com/facebook/react-native/pull/13199#issuecomment-354568863) I've identified with that code.

I believe this is the closest PR yet to a correct implementation of this feature, with a few caveats:

- As with the other PRs, this only works on Android >= 5.0 (silently falling back to no letter spacing on older versions). Is this acceptable for a RN feature, in general? Would a dev mode warning be desirable?
- The other PRs seem to have explored the space of potential solutions to the layout issue ([Android renders space _around_ glyphs](https://issuetracker.google.com/issues/37079859), iOS to the _right_ of each one) and come up empty, so I've opted to merely document the difference.
- I have neither updated nor tested the "Flat" UI implementation - everything compiles but I've taken [this comment](https://github.com/facebook/react-native/issues/12770#issuecomment-294052694) to mean there's no point in trying to wade through it on my own right now; I'm happy to tackle it if given some pointers.
- The implementation in `ReactEditText` is only there to handle the placeholder text, as `ReactBaseTextShadowNode` already affects the input control's contents correctly.
  - I'm not sure whether `<TextInput>` is meant to respect `allowFontScaling`; I've taken my cue here from `ReactTextInputManager.setFontSize()`, and used the same units (SP) to interpret the value in `ReactEditText.setLetterSpacingPt()`.
  - I'm not sure whether `<TextInput>` is even meant to support `letterSpacing` - it doesn't actually work on iOS. I'm not going to be able to handle the Objective-C side of this, not as part of this PR at least.
- I have not added unit tests to `ReactTextTest` - is this desirable? I see that some other props such as `lineHeight` aren't covered there (unless I'm not looking in the right place).
- Overall, I'm new to this codebase, so it's likely I've missed something not mentioned here.

## Test Plan

Note comment re: unit tests above; RNTester screenshots follow.

### RNTester - `<Text>`
| iOS (existing functionality, amended test) | Android (new functionality & test) |
| - | - |
|  |  |

### RNTester - `<TextInput placeholder>`
[...]

PR #19645, opened 2018-06-11
Fix extreme TextInput slowness on Android

### LINKED ISSUE #19126: TextInput becomes slow after writing and clearing
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [X] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

## Environment
  OS: Linux 4.16
  Node: 8.11.1
  Yarn: 1.6.0
  npm: 6.0.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: 3.1.2 AI-173.4720617

  react: 16.3.2 => 16.3.2
  react-native: 0.55.3 => 0.55.3

## Steps to Reproduce

```javascript
import { View, TextInput } from 'react-native';

class App extends Component {
  render() {
         <View>
             <TextInput />
         </View>
   }
}
```

1. Create a basic app with a single TextInput
2. Run the app on Android, with or without debug active, also in production build (signed APK)
3. Write in the TextInput anything, delete it and repeat the process until it becomes laggy.

## Expected Behavior
TextInput should not start lagging after something has been written inside and deleted.

## Actual Behavior
TextInput lags after 40-50 times you delete the input, programmatically or via keyboard, doesn't really matter also the amount of text inside of it because happens also with a single character.

gnprice:
This reverts 5898817fc "Implement letterSpacing on Android >= 5.0".
Testing shows that that commit is the cause of #19126, where in a
controlled TextInput after some text is first added, then deleted,
further interaction with the TextInput becomes extremely slow.

Fixes #19126.

## Test Plan

Tried the repro case from #19126 without this change, then with it.
The issue reproduces, then doesn't.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
@@ -119,7 +119,7 @@ public abstract class ReactBaseTextShadowNode extends LayoutShadowNode {
                 start, end, new BackgroundColorSpan(textShadowNode.mBackgroundColor)));
       }
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-        if (textShadowNode.mLetterSpacing != Float.NaN) {
+        if (!Float.isNaN(textShadowNode.mLetterSpacing)) {
           ops.add(new SetSpanOperation(
             start,
             end,


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

gnprice:
/cc @motiz88 as the author of #17398, which became the commit that caused #19126 and which this PR would revert. If the code can be fixed to keep that feature while fixing this issue, that would of course be better than merging this revert :-)

I'd like to CC the reviewer of #17398 too, but AFAICT from the PR thread, the review happened internally within Facebook and I don't see any other details.

motiz88:
Huh, weird! I wasn't aware of #19126 until now, thanks @gnprice.

/cc @janicduplessis @hramos 

One thing I'd personally love to see is if we can preserve the `<Text>` behaviour and only regress `<TextInput>`.

Secondarily, it would be nice to figure out _why_ this is slow at all, and maybe extract an Android platform bug report out of this (though now that I've said this, I've shifted the probability towards a glaring memory leak in my own code 😅)

I'll have some time this weekend to investigate. I don't know what the release plans are exactly, but if the maintainers feel this patch should go out as-is, I don't have a problem with that.

hramos:
#17398 was merged by https://github.com/facebook/react-native/commit/5898817fc1a66bd317d65ce96520159df2f96045 which was reviewed internally by @mdvacca

janicduplessis:
I agree it would be nice to only revert the text input part, I think it should work if we just revert the code in ReactEditText.java and ReactTextInputManager.java.

shafy:
Any updates on in which release this PR will be included? #18916 is also the same issue.

kelset:
@shafy depending on the author responsiveness, we'll surely try to cherry pick this for 0.56.0 since its release is still a couple weeks (more or less) away

stueynet:
@gnprice Any chance you can respond to @hramos change request? This would allow it to be merged to master and cherry picked for 56 - https://github.com/react-native-community/react-native-releases/issues/14

hramos:
Feel free to send a PR that reverts just the Text Input changes, if the feedback here is not addressed.

gnprice:
Hi all, and thanks @janicduplessis and @hramos for the feedback!

Today I spent some more time experimenting with this. I tried this suggestion:

> I think it should work if we just revert the code in ReactEditText.java and ReactTextInputManager.java.

Unfortunately it doesn't work :-/ -- the issue still repros.

If on top of that I revert this one additional hunk:
```
       }
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if (textShadowNode.mLetterSpacing != Float.NaN) {
+          ops.add(new SetSpanOperation(
+            start,
+            end,
+            new CustomLetterSpacingSpan(textShadowNode.mLetterSpacing)));
+        }
+      }
       if (textShadowNode.mFontSize != UNSET) {
```
then that fixes the issue! That leaves part of the original changes from #17398 in place -- there's a `setLetterSpacing` method, and it's controlled by a `letterSpacing` prop and also called from `setAllowFontScaling`; and there's a `CustomLetterSpacingSpan` class. But as far as I understand how the original commit works, this hunk is at the crux of it, and without it the rest of the changes don't have any effect.

So you can see exactly what I tried, here's my branch:
https://github.com/gnprice/react-native/compare/v0.55.4...text-experiments-0.55

So it looks like this is going to take some debugging of the actual logic, if we want to fix the issue without just reverting the feature completely.

gnprice:
... Oho. I bet this is the problem:

```
        if (textShadowNode.mLetterSpacing != Float.NaN) {
```

Because NaN is special, that will always be true -- even if `mLetterSpacing` is also `Float.NaN`, which is its default value.

That means every text shadow node will create a `CustomLetterSpacingSpan` around its contents, even when the `letterSpacing` prop was never set. It must be that we're somehow ending up with large numbers of these shadow nodes -- that sounds like a bug in itself, but I guess it's normally low-impact -- and this condition is failing to prune them from causing a bunch of work here.

The correct way to write this condition is with `Float.isNaN`. I'll go see if that fixes the issue.

gnprice:
New version pushed! This one is a simple one-line fix, just fixing the specific bug in #17398 without reverting anything:

```
       if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-        if (textShadowNode.mLetterSpacing != Float.NaN) {
+        if (!Float.isNaN(textShadowNode.mLetterSpacing)) {
           ops.add(new SetSpanOperation(
```

@hramos , @janicduplessis , please take a look!

hramos:
I tried to approve and merge this, but it looks like you haven't signed the CLA yet, @gnprice  :)

gnprice:
CLA signed! :) Hopefully the bot will chime in confirming that shortly.

motiz88:
@gnprice Argh, I should have known better than to introduce a NaN comparison bug! Well spotted & thanks for being on top of this.

shafy:
@motiz88 happens to the best =P

gnprice:
Ouch, the commit message in the final commit is wrong because it's taken from my PR message (which describes the original version, a revert of that previous commit) rather than from my actual commit (which describes what it actually does).

I guess as a former (happy!) Phabricator user, I should have known that was how it was going to work. :-P Next time I'll be sure to edit the PR message, just like when using Phabricator.

Thanks @hramos and @janicduplessis !

Noitidart:
@gnprice I don't get why `if (!Float.isNaN(textShadowNode.mLetterSpacing)) {` is a fix. Because if we do set `letterSpacing` then this will wrap everything in a span no? So this only fixes if `letterSpacing` is not set no?

Maybe we should remove the `letterSpacing` support in this case?

gnprice:
@Noitidart 

> if we do set letterSpacing then this will wrap everything in a span no? So this only fixes if letterSpacing is not set no?

That's right. If you look at the surrounding code, this follows the pattern of about 8 other features:
```
      if (textShadowNode.mFontSize != UNSET) {
        ops.add(new SetSpanOperation(start, end, new AbsoluteSizeSpan(textShadowNode.mFontSize)));
      }
// ...
      if (textShadowNode.mIsUnderlineTextDecorationSet) {
        ops.add(new SetSpanOperation(start, end, new UnderlineSpan()));
      }
      if (textShadowNode.mIsLineThroughTextDecorationSet) {
        ops.add(new SetSpanOperation(start, end, new StrikethroughSpan()));
      }
```
In each case: if some prop is set, then it's implemented by creating a span. This makes sense to me architecturally, if the implementation works out the way I imagine the authors of this class intended: one formatting prop on one React component, or one spot where the formatting needs to change inside a React component, produces one Android text span (or possibly an O(1) handful for some features.) That corresponds well with my understanding of how the Android text span API is designed to be used.

It's still a bit of a puzzle to me why we're going through this code so many times that getting this condition wrong was such a problem. As I wrote above, it seems like that means we're ending up with tons and tons of instances of this `ReactBaseTextShadowNode` class (well, of its subclasses), just from typing a bunch of text into a single boring `TextInput`. That isn't how this class reads to me like it's designed to be used.

Whatever's causing that, though, I think it's unrelated to the `letterSpacing` feature, which looks like it's doing the same thing here (after this bugfix) as a bunch of other formatting features.

Take a look at #20119, which I just filed today; I think that issue is related to this puzzle.

Noitidart:
Thanks @gnprice for that detail!

chris-divvito:
Is there a way to get a version of this patch onto 0.55 short of us just forking RN and applying it?

hramos:
You can check out the `0.55-stable` branch and cherry-pick this fix onto your local checkout, and build a release from that.

Issue #17123, opened 2017-12-08
Android - App crashes with ReactRootView inside viewpager


tdekoning:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Steps to Reproduce

(Write your steps here:)

1. Create a native `ViewPager` with a `ReactRootView` as content, preferably with some images inside the `ReactRootView`
2. Swipe fast
3. See this stack trace

### Expected Behavior

I would expect no stack trace or errors to occur.

### Actual Behavior

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void com.facebook.react.uimanager.ReactShadowNode.addChildAt(com.facebook.react.uimanager.ReactShadowNode, int)' on a null object reference
       at com.facebook.react.uimanager.UIImplementation.setChildren(UIImplementation.java:472)
       at com.facebook.react.uimanager.UIManagerModule.setChildren(UIManagerModule.java:436)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:374)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:162)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
       at android.os.Looper.loop(Looper.java:158)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:194)
       at java.lang.Thread.run(Thread.java:818)
```

This occurs after swiping through the ViewPager quickly.

### Reproducible Demo

Proposed fix:
Below [this line of code](https://github.com/facebook/react-native/blob/d19afc73f5048f81656d0b4424232ce6d69a6368/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIImplementation.java#L474): Add the same fix as [this one](https://github.com/facebook/react-native/blob/d19afc73f5048f81656d0b4424232ce6d69a6368/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIImplementation.java#L152)

[...]

Issue #17178, opened 2017-12-13
com.facebook.react.uimanager.IllegalViewOperationException


henrikra:
### Is this a bug report?

Yeah.

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yeah.

### Environment

Environment:
  OS:  macOS High Sierra 10.13.1
  Node:  8.9.2
  Yarn:  1.3.2
  npm:  5.5.1
  Watchman:  4.7.0
  Xcode:  Xcode 9.2 Build version 9C40b
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: 0.50.3 => 0.50.3

### Steps to Reproduce

1. Have Android phone
2. Add `LayoutAnimation` to some of your components
3. App crashes randomly

### Expected Behavior

The `UIImplementation` is trying to add some `View` which is not there yet.

### Actual Behavior

1. Happening only on Android
2. OS versions having this crash so far: 
![image](https://user-images.githubusercontent.com/12229968/33937832-ac66b330-e00d-11e7-93a6-42eb733efd06.png)

Logs:
```
com.facebook.react.uimanager.IllegalViewOperationException Trying to add unknown view tag: 430 
    UIImplementation.java:396 com.facebook.react.uimanager.UIImplementation.setChildren
    UIManagerModule.java:310 com.facebook.react.uimanager.UIManagerModule.setChildren
    Method.java:-2 java.lang.reflect.Method.invoke
    JavaMethodWrapper.java:363 com.facebook.react.bridge.JavaMethodWrapper.invoke
    JavaModuleWrapper.java:162 com.facebook.react.bridge.JavaModuleWrapper.invoke
    NativeRunnable.java:-2 com.facebook.react.bridge.queue.NativeRunnable.run
    Handler.java:789 android.os.Handler.handleCallback
    Handler.java:98 android.os.Handler.dispatchMessage
    MessageQueueThreadHandler.java:31 com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage
    Looper.java:164 android.os.Looper.loop
    MessageQueueThreadImpl.java:194 com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run
    Thread.java:764 java.lang.Thread.run
```

### Backstory
[...]

PR #19651, opened 2018-06-11
Add null check in UIIMplementation#setChildren

fourlastor:
This PR implements the suggested fix for #17123, basically adding a null check after getting the view via tag in `UIImplementation#setChildren`.

## Test Plan

As per issue:

1. Create a native `ViewPager` with a `ReactRootView` as content, preferably with some images inside the `ReactRootView`
2. Swipe fast (back and forth)
3. The app will crash with an NPE before the fix, no NPE and a warning is logged after the fix.

## Release Notes

[ANDROID] [BUGFIX] [UIImplementation] - Added null check in UIImplementation#setChildren

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIImplementation.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIImplementation.java
@@ -479,6 +479,11 @@ public class UIImplementation {
 
     ReactShadowNode cssNodeToManage = mShadowNodeRegistry.getNode(viewTag);
 
+    if (cssNodeToManage == null) {
+      FLog.w(ReactConstants.TAG, "Tried to set children on non-existent root tag: " + viewTag);
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


fourlastor:
@mdvacca as far as I understood, the viewpager removes the view (or a view containing the view), then the method gets called. At that point the view isn't there anymore and getting it via tag returns null. Can investigate more but it's kinda tricky, because it's very hard to reproduce reliably (you have to keep swiping back and forth).

patrickkempff:
I agree with @mdvacca. I think this is an threading issue. I don't think that silencing the exception is the right approach. PR #20025 should fix this issue. 

Util this has landed you can try the steps described by @SudoPlz here: https://github.com/facebook/react-native/issues/17178#issuecomment-394662148

fourlastor:
That's fair enough, I'll close this then! Thanks for getting back to me (I tried to investigate but got pretty much stuck at the same point every time)

PR #19653, opened 2018-06-11
iOS: fix the baseline issue when displaying a mixture of different-language characters

### LINKED ISSUE #19193: Setting text contains combination of Chinese and English with `lineHeight` style cause those characters not on one baseline

## Environment

Environment:
  OS: macOS High Sierra 10.13.4
  Node: 10.0.0
  Yarn: 1.6.0
  npm: 6.0.0
  Watchman: 4.9.0

  react: 16.3.1 => 16.3.1
  react-native: ~0.55.2 => 0.55.3

## Steps to Reproduce

```
import React from 'react';

  render() {
    return (
        <Text
          style={styles.text}
          numberOfLines={2}>
          这里是中文 This is English 1234567890 这里是数字 This is numbers 123456789
        </Text>
      </View>
    );
  }
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
  text: {
    maxWidth: '80%',
    fontSize: 17,
    color: '#4e4e4e',
    lineHeight: 24,
  },
});
```
## Expected Behavior

I migrated my app from 0.52.3 and previous version works fine. All characters are on one baseline. 

There's the screenshot. 

**Please note that I took this screenshot without `lineHeight` style applied, so the expected behavior should render larger line spacing.**

simulator screen shot - iphone 6 - 2018-05-09 at 10 41 42

## Actual Behavior

Chinese characters on one baseline, while English and numbers on another. 

There's the screenshot. 

simulator screen shot - iphone 6 - 2018-05-09 at 10 41 32

**And seems the problem only affect with iOS. **

lyb0307:
This fixes #19193 

Characters of different languages may have different height and different baseline, even with a same font and same font-size. When they mixed together, iOS will adjust their baselines automatically, to display a suitable view of text. 

The problem is the behavior of dealing with the text-style property 'lineHeight'.

It once to be a right way at version 0.52.3, to setting a base-line-offset attribute for the whole range of the string. However, in current version, it enumerated the string, and set a different  base-line-offset for different font-height characters. 

And this behavior broke the baseline adjustment made by the iOS. It just make every character's baseline aligned to a same line. But it is not supposed to displaying characters of different languages like that. Chinese characters' baseline is the bottom of each, however, it is not for English characters.

So maybe it is the better way to give a same value of base-line-offset attribute for the whole string. And this PR just did that: found the biggest value of font-height in the text, and calculate the offset with that biggest value, then set to the whole string. This will keep the origin baseline adjustment for different languages' chars made by iOS.

## Test Plan

Since I always got an error when running the snapshot test locally, I can't even pass the them with the unmodified code in master branch.

The error is "Nesting of <View> within <Text> is not currently supported."

After I comment all of the check of that error from the source code, I got a different snapshot from the reference ones. It seems that all components which will cause that error are not rendered in the reference images. 

Since this PR changed the behavior of 'lineHeight' property, it's better to add a new snapshot case for the situation of different-language-characters' mixture. So maybe somebody could help me with that or maybe it should be a issue?

## Release Notes

[IOS] [BUGFIX] [Text] - fix the baseline issue when displaying a mixture of different-language characters

### DIFF
--- a/Libraries/Text/Text/RCTTextShadowView.m
+++ b/Libraries/Text/Text/RCTTextShadowView.m
@@ -133,30 +133,37 @@
     return;
   }
 
-  [attributedText beginEditing];
-
+  __block CGFloat maximumFontLineHeight = 0;
+  
   [attributedText enumerateAttribute:NSFontAttributeName
                              inRange:NSMakeRange(0, attributedText.length)
                              options:NSAttributedStringEnumerationLongestEffectiveRangeNotRequired
                           usingBlock:
-    ^(UIFont *font, NSRange range, __unused BOOL *stop) {
-      if (!font) {
-        return;
-      }
-
-      if (maximumLineHeight <= font.lineHeight) {
-        return;
-      }
-
-      CGFloat baseLineOffset = maximumLineHeight / 2.0 - font.lineHeight / 2.0;
-
-      [attributedText addAttribute:NSBaselineOffsetAttributeName
-                             value:@(baseLineOffset)
-                             range:range];
+   ^(UIFont *font, NSRange range, __unused BOOL *stop) {
+     if (!font) {
+       return;
      }
-   ];
-
-   [attributedText endEditing];
+     
+     if (maximumFontLineHeight <= font.lineHeight) {
+       maximumFontLineHeight = font.lineHeight;
+     }
+     
+   }
+  ];
+  
+  if (maximumLineHeight < maximumFontLineHeight) {
+    return;
+  }
+  
+  [attributedText beginEditing];
+  
+  CGFloat baseLineOffset = maximumLineHeight / 2.0 - maximumFontLineHeight / 2.0;
+  
+  [attributedText addAttribute:NSBaselineOffsetAttributeName
+                         value:@(baseLineOffset)
+                         range:NSMakeRange(0, attributedText.length)];
+  
+  [attributedText endEditing];
 }
 
 - (NSAttributedString *)attributedTextWithMeasuredAttachmentsThatFitSize:(CGSize)size


kgdev:
Can we do a review for this PR. really need this fix.

lyb0307:
@facebook-github-bot Thank you for reply. Yes, this PR is still relevant. As I can see, the files effected by this PR still exist in latest master. And the last modification of the related lines was made by @shergin at commit > https://github.com/facebook/react-native/commit/2716f53220f947c690d5f627286aad51313256a0 So, @shergin could you please take a review for this？

MrPluto:
any progress about this PR ?

ikesyo:
We want this as well.

@shergin Could you please review and merge this?

watanabeyu:
@shergin 
we want this pr merge.
please check!!!

ikesyo:
Is there any chane to ge this (or #19619) reviewd?

Ping to @shergin (or any other appropriate reviewer? @hramos @kelset)

hramos:
Thanks for the PR. Can you include screenshots of the issue before and after the patch is applied?

How did you test this change?

Libraries/Text/Text/RCTTextShadowView.m
```diff
@@ -133,30 +133,37 @@ - (void)postprocessAttributedText:(NSMutableAttributedString *)attributedText
...
+     }
+     
+   }
+  ];
+  
+  if (maximumLineHeight < maximumFontLineHeight) {
+    return;
+  }
+  
+  [attributedText beginEditing];
```
shergin: We don't need this (and closing part) here because we change the string just one.

PR #19663, opened 2018-06-12
macOS : request interactive input when killing the packager

zedtux:
Do not request interactive input when running on a CI

When I'm running the build on my CI server, many Terminal windows are remaining on the server because of the `read` command.

This PR prevent this behaviour when the `CI` environment variable is defined so that it doesn't alter the behaviour on local machines.

## Test Plan

 * Run a build when the packager killing step at the end
 * Ensure no window remain opened on the server (macOS)

## Release Notes

[IOS] [MINOR] [launchPackager.command] - Do not request interactive input when killing packager on a CI server


zedtux:
Build doesn't seem to be broken by my change

zedtux:
Come on ... Am I trapped in an infinite loop here? 🤦‍♂️

zedtux:
@cpojer do you need anything else from me?

cpojer:
No thanks, I'm landing the change now :)

zedtux:
Awesome, thank you @cpojer

PR #19664, opened 2018-06-12
Fix memory usage display in Perf Monitor

JunyiXie:
According to: https://github.com/apple/darwin-xnu/blob/master/osfmk/kern/bsd_kern.c#L420
The result is the same as Xcode app memory uesd.

## Test Plan

Before Fix:

before
before1

After Fix:
after
after1

## Release Notes

[IOS] [BUGFIX] [Monitor] - fix memory perf monitor. Make sure the memory footprint is the same as Jetsam.

### DIFF
--- a/React/Profiler/RCTPerfMonitor.m
+++ b/React/Profiler/RCTPerfMonitor.m
@@ -68,8 +68,7 @@ static vm_size_t RCTGetResidentMemorySize(void)
     kern_return_t kernelReturn = task_info(mach_task_self(), TASK_VM_INFO, (task_info_t) &vmInfo, &count);
     if(kernelReturn == KERN_SUCCESS) {
         memoryUsageInByte = (vm_size_t) vmInfo.phys_footprint;
-    } else {
-    }
+    } 
     return memoryUsageInByte;
 }
 


hramos:
Can you elaborate? What issue does this fix, how was this tested?

JunyiXie:
Make sure the memory footprint is the same as Xcode monitor.
iOS Jetsam(out of memory) use this method to measure memory usage.
Before change:
before
before1

After change: 
after
after1

zhongwuzw:
@JunyiXie `Xcode` 's memory gauge is good for noticing memory usage trends, and is only a barometer refer to the developer from `Apple`, see [Difference in iOS memory allocation measurement between XCode and instruments](https://forums.developer.apple.com/thread/52186).

You can have `Allocation template` of `Instruments` to see the memory use.

JunyiXie:
iOS Jetsam(out of memory) use this method to measure process memory usage.

JunyiXie:
Is the pull request format right now?

React/Profiler/RCTPerfMonitor.m
```diff
@@ -62,17 +62,15 @@ static BOOL RCTJSCSetOption(const char *option)
...
-
-  return info.resident_size;
+    vm_size_t memoryUsageInByte = 0;
+    task_vm_info_data_t vmInfo;
+    mach_msg_type_number_t count = TASK_VM_INFO_COUNT;
+    kern_return_t kernelReturn = task_info(mach_task_self(), TASK_VM_INFO, (task_info_t) &vmInfo, &count);
+    if(kernelReturn == KERN_SUCCESS) {
+        memoryUsageInByte = (vm_size_t) vmInfo.phys_footprint;
+    } else {
+    }
```
henrikra: Can you remove this empty `else` block safely?

JunyiXie:
Delete useless logic.

PR #19673, opened 2018-06-12
add RNTester to ci

gengjiawen:
## Motivation
* Current ci is missing an important part to test the whole part. With this we can make sure the js and android part compiles.
* Ensure the current android proguard rules is okay.

The `my-release-key.keystore` is just a copy of debug.keystore in `react-native/keystores`.
## Test Plan
Pass all ci.
## Related PRs
none
## Release Notes
 [GENERAL] [ENHANCEMENT] [CI] - Add RNTester to ci


.circleci/config.yml
```diff
@@ -563,6 +563,11 @@ jobs:
       - store_test_results:
           path: ~/react-native/reports/junit
 
+      - run:
```
hramos: Can you move this step to just before # Collect Results?

gengjiawen: @hramos done.

gengjiawen:
@hramos The request change is done.

PR #19681, opened 2018-06-13
[CI] Use ShellCheck

hramos:
As recommended in https://circleci.com/docs/2.0/using-shell-scripts/#use-shellcheck

It will only run on PRs for now.


scripts/circleci/analyze_scripts.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
```
hramos: SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.

Issue #18890, opened 2018-04-16
iOS Dictation TextInput ends abruptly between words


sedevccb:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [x] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

When using dictation on a TextInput in iOS, the dictation will end abruptly between words. This was not an issue previously on React Native 53. Moving to version 54+ causes this behavior.

## Environment
Environment:
  OS: macOS High Sierra 10.13.4
  Node: 9.7.1
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: Not Found
  Xcode: Xcode 9.3 Build version 9E145
  Android Studio: Not Found

Packages: (wanted => installed)
  react: 16.3.0-alpha.1 => 16.3.0-alpha.1
  react-native: https://github.com/expo/react-native/archive/sdk-26.0.0.tar.gz => 0.54.2

## Steps to Reproduce
Try using iOS dictation in the TextInput. Be sure to use a long sentence...

```
import React from 'react';
import { TextInput } from 'react-native';

export default class App extends React.Component {
  state = { value: '' }
  onChangeText = value => console.log(value) || this.setState({ value })
  render() {
    return (
        <TextInput
          onChangeText={this.onChangeText}
          value={this.state.value}
          style={{ borderWidth: 2, borderColor: 'black', width: 200, height: 48 }}
        />
    );
  }
}
```

## Expected Behavior
I expect the TextInput to capture the entire dictation.

## Actual Behavior
The dictation will end abruptly between words.

marcosrdz:
If I place a breakpoint on setAttributedText  (line 101) on this file:

 react-native/Libraries/Text/TextInput/RCTBaseTextInputView.m

and wait a second or two, then resume the app, the dictated sentence does not get cut off.

gancan110:
I have this problem too. Is that a problem? Or are you ready to use the onBlur method to get the data that is being entered by Textinput? Version 0.49.2 is fine, but 0.55 will also appear.

lbelavoir:
@marcosrdz moving `setState` in `onBlur` will make useless all the other props such as `onEndEditing` or `onSubmitEditing` because they will fire before `setState` take the previous `state`

arvidkahl:
I have solved this issue using `lodash.debounce`.

```jsx
import _ from 'lodash'

[...]

PR #19687, opened 2018-06-13
Possible fix for 18890

eoinnorris:
Potential fixes for #18890. The issue was that when the setAttributedText method in RCTBaseTextInputView.m is called it does two main things:

Check that the attributed text that it receives is equal to the existing attributed text in the underlying backed up view ( backedTextInputView)
If not set the attributed text of the backed up view to the value passed into the method. This kills the dictation as it is effectively the equivalent of typing in the backed up text view.
    self.backedTextInputView.attributedText = attributedText;
is the problem. It kills the dictation. It may have other effects as well.

In all cases I have seen the underlying text of the attributed string that is passed in the same as the text in the backedTextInputView, what was said to the dictation; however the attributes are different, which causes the isEqualToAttributedString: check to fail and thus the update happens, and the keyboard is killed.

Fix is to test for the underlying string equality not the attributed string equality when the input mode is dictation. 

## Test Plan

By necessity this had to be an integration test on an existing application. To test I enabled the keyboard and started the dictation. It worked with this fixes and not without. Will upload videos later. 

## Risks

This might break attributes on dictation, as it is happening. However anything set on the existing underlying text view should hold. 

## Release Notes
[IOS] [BUG] [Textinput] fixed an issue where the keyboard dictation ended abruptly.

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -98,40 +98,54 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
   return self.backedTextInputView.attributedText;
 }
 
+- (BOOL)textOf:(NSAttributedString*)newText equals:(NSAttributedString*)oldText{
+  UITextInputMode *currentInputMode =  self.backedTextInputView.textInputMode;
+  if ([currentInputMode.primaryLanguage isEqualToString:@"dictation"]) {
+    // when the dication is running we cant update the attibuted text on the backed up text view
+    // because setting the attributed string will kill the dictation. This means that we cant impose the
+    // settings on a dication.
+    return ([newText.string isEqualToString:oldText.string]);
+  } else {
+    return ([newText isEqualToAttributedString:oldText]);
+  }
+}
+
 - (void)setAttributedText:(NSAttributedString *)attributedText
 {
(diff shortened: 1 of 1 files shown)


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

th317erd:
This appears to be the best fix I can think of (I was about to create and submit a PR for the same issue before I found this one. Thanks @eoinnorris !), however, as a future improvement should the text field emit "onDictationStarted" and "onDictationEnded" events to let the Javascript application know that 1) dictation is in progress, and 2) that any `value` updates happening while dictation is active will be ignored? Where / how would a person go about updating the code to emit these events? I am no iOS expert...

hramos:
This fix is in 0.57.

th317erd:
@eoinnorris Thank you for the information on this. I will look into this when I get a spare moment.

@hramos Are you saying PR 19687 made it into 0.57? Or that what I am proposing has already been implemented by someone?

hramos:
This PR has already been merged. You can look up the commit and see if it’s
in 0.57. There will be a v0.57.0 tag if it is.

chrisbobbe:
> You can look up the commit and see if it’s in 0.57. There will be a v0.57.0 tag if it is.

None of the three commits have any tags shown in the GitHub UI; was this released in v0.57.0?

PR #19705, opened 2018-06-14
Show packager error stack

iyegoroff:
Motivation: get more info from packager errors to easier debug them

## Test Plan
add `throw new Error('error')` here https://github.com/iyegoroff/react-native/blob/635f2740c28da594f236d4227b6450fddef4b599/local-cli/cliEntry.js#L116 and run
`react-native start` to see error stack

[CLI] [ENHANCEMENT] [local-cli/cliEntry.js] - Show packager error stack

### DIFF
--- a/local-cli/cliEntry.js
+++ b/local-cli/cliEntry.js
@@ -34,10 +34,14 @@ commander.version(pkg.version);
 
 const defaultOptParser = val => val;
 
-const handleError = err => {
+const handleError = (err) => {
   console.error();
   console.error(err.message || err);
   console.error();
+  if (err.stack) {
+    console.error(err.stack);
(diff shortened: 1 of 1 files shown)


hramos:
Can you share an example of how the output is changed after this fix is applied?

iyegoroff:
Output for throw `new Error('some error inside addCommand')` without error stack:

screen shot 2018-06-15 at 20 11 34

Output for `throw new Error('some error inside addCommand')` with error stack:

screen shot 2018-06-15 at 20 11 55

Output for `throw new Error('some error inside metro/Server/index.js')` with error stack:

screen shot 2018-06-15 at 20 15 02

PR #18350, opened 2018-03-13
[ANDROID] Add number-pad keyboardType

khaeransori:
## Motivation

On Android there's no number-pad but numeric instead, for my use-case I need number only (without decimal and sign, for PIN input)

## Test Plan

I write it so there's no breaking change for those already using the `TextInput` Component

## Related PRs

None

## Release Notes

[ANDROID][ENHANCEMENT][Component] Add number-pad keyboardType

Issue #17474, opened 2018-01-07
Can't see decimal key when keyboardType="numeric" on my HTC device


RituKhetan:
Is this a bug report?
YES

Environment
Environment:
OS: Windows 8
Node: v9.3.0

npm:  4
Using EXPO app for testing
Packages: (wanted => installed)
"expo": "^23.0.4",
"react": "16.0.0",
"react-native": "0.50.3"

Target Platform: Android (4.4.2)

Steps to Reproduce
(Write your steps here:)

Install create-react-native-app (Run command: npm install -g create-react-native-app
Create a react-native-app (Go to a directory and run command: create-react-native-app tipcalc
Make changes in the App.js file: https://gist.github.com/RituKhetan/2306e5bed13f78cb5602dc9c1c21c90d
Expected Behavior
Should show numeric keypad with decimal key on my HTC device

Actual Behavior
Cannot see the decimal key on my HTC device when I try to type into the TextInput component.

RituKhetan:
Can someone please help?

aemdy:
Same here as well on at least Samsung Galaxy A3.

xerotolerant:
This problem still persits on Samsung devices.

JaEdmuva:
I have the same issue with a HTC One and RN 0.54.2, anybody has a clean and safe solution?

elly0t:
any update on this?

duyphambhtech:
It has the same issue for my s8+ but not for my s5

aenciso:
This is related to this

https://github.com/facebook/react-native/issues/12988

Potential solution is to add `autoCapitalize="none"` to TextInput. I know it's not ideal but at least it shows decimal point on the keyboard.

scriptgb:
When user uses Samsung Keyboard the issue, still persists. Anyone find a solution to that?

scriptgb:
@JPanneel I'm using react-native 0.48.4, I tried to change the files manually but the behaviour stills the same. So, I need to upgrade the react-native version or there is something else that I can do to make this works.

yanhaijing:
Samsung s8 has the same issue

amcsi:
I'm getting the same issue on Samsung Galaxy S7. My girlfriend as well on her Xiaomi Redmi Plus 3.

```
    "expo": "^26.0.0",
    "react": "^16.3.2",
"react-native": "https://github.com/expo/react-native/archive/sdk-26.0.0.tar.gz",
```

[...]

Issue #17473, opened 2018-01-07
Can't see decimal key when keyboardType="numeric" on my HTC device


RituKhetan:

RituKhetan:
### Is this a bug report?

YES

### Environment

  Environment:
    OS:  Windows 8
    Node:  v9.3.0
 
    npm:  4
    Using EXPO app for testing

  Packages: (wanted => installed)
    "expo": "^23.0.4",
    "react": "16.0.0",
    "react-native": "0.50.3"

  Target Platform: Android (4.4.2)

 
### Steps to Reproduce

(Write your steps here:)

1. Install create-react-native-app (Run command: npm install -g create-react-native-app
2. Create a react-native-app (Go to a directory and run command: create-react-native-app tipcalc
3. Make changes in the App.js file: https://gist.github.com/RituKhetan/2306e5bed13f78cb5602dc9c1c21c90d

### Expected Behavior

Should show numeric keypad with decimal key on my HTC device

### Actual Behavior

Cannot see the decimal key on my HTC device when I try to type into the TextInput component.

PR #19714, opened 2018-06-14
Feature/add decimal pad to android

JPanneel:
For a current use-case we need the a keyboard with characters 0-9 and a decimal point (or comma depending on language settings)

This exists on iOS as UIKeyboardType "decimalPad" and this is what react-native maps to for both "numeric" and "decimal-pad". This also exists on Android as inputType "numberDecimal", but is currently not accessible through react-native.

This PR maps the value "decimal-pad" of the keyboardType property of TextInput to the Android inputType "numberDecimal", effectively making "decimal-pad" cross platform without breaking anything.

## Relevant links to documentation
* https://facebook.github.io/react-native/docs/textinput.html#keyboardtype
* https://developer.apple.com/documentation/uikit/uikeyboardtype
* https://developer.android.com/reference/android/widget/TextView#attr_android:inputType

## Additional benefit 
There is this bug in some Samsung keyboards where both the - sign and decimal sign disappear when the keyboardType is set to "number" and both the "signed" and "decimal" flags are set. (Like is the case when using the react-native keyboardType prop "numeric".) https://androidforums.com/threads/numeric-soft-keyboard-missing-minus-sign-in-android-8-0-samsung-a5.1272628/ 

For developers that need decimal numbers but not negative ones, using "decimal-pad" will provide a workaround. I reproduced this on a Samsung A5 only, but maybe other phones have this exact issue. https://github.com/facebook/react-native/issues/12988 https://github.com/facebook/react-native/issues/12977 https://github.com/facebook/react-native/issues/17473 https://github.com/facebook/react-native/issues/17474

## Test Plan

* Added testcase consistent with existing keyboardType tests
* Also added testcase for the related, but missing number-pad

## Related PRs

This PR follows the same approach as the recently merged PR introducing "number-pad" https://github.com/facebook/react-native/pull/18350/commits/b638847a46491bd75e7ce7928c73f7cb78399195

Documentation PR: https://github.com/facebook/react-native-website/pull/405

## Release Notes

 [ANDROID] [ENHANCEMENT] [TextInput] - Added "decimal-pad" keyboard type

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -86,12 +86,12 @@ export type KeyboardType =
   | 'numeric'
   | 'phone-pad'
   | 'number-pad'
+  | 'decimal-pad'
   // iOS-only
   | 'ascii-capable'
   | 'numbers-and-punctuation'
   | 'url'
   | 'name-phone-pad'
-  | 'decimal-pad'
   | 'twitter'
   | 'web-search'
   // Android-only
@@ -388,6 +388,7 @@ const TextInput = createReactClass({
      * - `default`
      * - `numeric`
      * - `number-pad`
+     * - `decimal-pad`
      * - `email-address`
      * - `phone-pad`
      *
@@ -399,7 +400,6 @@ const TextInput = createReactClass({
      * - `numbers-and-punctuation`
      * - `url`
      * - `name-phone-pad`
-     * - `decimal-pad`
      * - `twitter`
      * - `web-search`
      *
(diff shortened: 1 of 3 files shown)


mdvacca:
Thanks for working on this @JPanneel.
Can you also update the documentation here: https://facebook.github.io/react-native/docs/textinput.html#keyboardtype

JPanneel:
@mdvacca Is that something other than what is included in https://github.com/facebook/react-native-website/pull/405 ? If so, I am not sure where to update the documentation you mention. 
https://deploy-preview-405--react-native.netlify.com/docs/next/textinput.html#keyboardtype

JPanneel:
No problem. Thanks for the fast feedback!

DickLemm:
See my comment https://github.com/facebook/react-native/issues/22005?_pjax=%23js-repo-pjax-container.

PR #19722, opened 2018-06-14
Web Audio API support

ramirezd42:
Hey all! 

Sorry if this isn't the correct forum for this, but I am interested in investigating a means bringing support for the Web Audio API to react-native. This is something I think could have some cool potential but probably wouldn’t be a priority roadmap item any time soon.

There are some points of the api that I think would be problematic to implement as a standard NativeModule. A good example is the [AnalyserNode - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode). The AnalyserNode allows for access to the underlying data of the audio stream flowing through it in real-time. A good use case for this might be implementing some form of visualization like a volume meter. However if we want the meter to animate smoothly we would need to get data from native objects fast enough to render frames for an animation, which could create performance issues if done over the bridge.

Because of this I’ve considered exposing wrapped native objects directly to the JSC context, bypassing the bridge entirely. [I’ve played around with a similar idea before](https://github.com/ramirezd42/node-audio) to introduce a native-backed web audio implementation to Node with v8, but given that this is a pretty large departure from existing react-native architecture I wanted to get some feedback or at least some stern warnings on what I might be bumping into!


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

kelset:
Hey @ramirezd42 I think you may want to open an issue instead ;)

ramirezd42:
OK will do!

PR #19724, opened 2018-06-15
Bugfix/ci lint proguard

gengjiawen:
## Motivation
* avoid lint to interrupt ci
* add proguard rules
## Test Plan
pass all ci
## Related PRs
none
## Release Notes
 [GENERAL] [ENHANCEMENT] [CI] - Fix lint and proguard

### DIFF
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -120,6 +120,11 @@ android {
             signingConfig signingConfigs.release
         }
     }
+
+    lintOptions {
+        abortOnError false
+    }
+
     // applicationVariants are e.g. debug, release
     applicationVariants.all { variant ->
         variant.outputs.each { output ->
--- a/ReactAndroid/proguard-rules.pro
+++ b/ReactAndroid/proguard-rules.pro
@@ -55,10 +55,6 @@
 # See libs/proxy/src/main/java/com/facebook/fbui/textlayoutbuilder/proxy for details.
 -dontwarn android.text.StaticLayout
 
-# So loader
--keep class com.facebook.soloader.SoLoader {*;}
--dontwarn com.facebook.soloader.SoLoader
-
 # okhttp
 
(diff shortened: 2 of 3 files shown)


gengjiawen:
@hramos Technically this should be split two pr, but it fix one problem. Any opinion ?

hramos:
Let's try and fix test_android in a single PR if possible. What broke it?

gengjiawen:
Two things
* The lint rule broke it after bump to andorid 26.
* Something wrong with proguard with current toolchain. The build passes, but it crashed on phone, still finding the reason.

And also sometimes circle ci broke with `Received 'killed' signal`, but it works on local machine.

gengjiawen:
About the soloader, decompiled code showed it's there, weired.
image

gengjiawen:
I find the reason, looks like something wrong with soloader 0.5.0, revert it back works fine.

RNTester/android/app/build.gradle
```diff
@@ -120,6 +120,11 @@ android {
             signingConfig signingConfigs.release
         }
     }
+
+    lintOptions {
+        abortOnError false
```
chirag04: can we rather fix the lint errors? might be masking errors, no?

hramos: @chirag04 we just added RNTester to this job, so I think it's fine to ignore lint errors *right now* to keep tests green. We should, however, have someone send PR(s) to fix any lint warnings that currently exist in master so that we can re-enable lint errors.

I have a similar issue where ShellCheck has surfaced a lot of issues in our shell scripts, but enabling it would turn tests red right away. In that particular case I plan to chip away at the shell issues until ShellCheck can be enabled fully.

chirag04: I gave it a quick run: https://github.com/facebook/react-native/pull/19743 CI is passing

PR #19741, opened 2018-06-15
deprecate ReactFragmentActivity

dulmandakh:
Android Support Library page says https://developer.android.com/topic/libraries/support-library/
> Caution: Starting with Support Library release 26.0.0 (July 2017), the minimum supported API level across most support libraries has increased to Android 4.0 (API level 14) for most library packages. For more information, see Version Support and Package Names in this document.

_android.support.v4.app.FragmentActivity_ is used to support **Fragment**s on Android API versions below 11. Now, we support only API version 14 and above, it's ok to remove _ReactFragmentActivity_ that extends _FragmentActivity_.

Once ReactFragmentActivity removed, we can remove some codes that use _android.support.v4_ to work support _ReactFragmentActivity_ on **unsupported** Android versions. 

CI is failing for unknown reasons: https://circleci.com/gh/dulmandakh/react-native/278
> Received 'killed' signal
## Test Plan

Everything will build and work just fine, except showing _ ReactFragmentActivity_ as deprecated

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactFragmentActivity.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactFragmentActivity.java
@@ -19,9 +19,11 @@ import com.facebook.react.modules.core.PermissionAwareActivity;
 import com.facebook.react.modules.core.PermissionListener;
 
 /**
- * Base Activity for React Native applications. Like {@link ReactActivity} but extends
- * {@link FragmentActivity} instead of {@link android.app.Activity}.
+ * @deprecated
+ * ReactFragmentActivity will be removed in 0.57 release.
+ * Use {@link ReactActivity} instead.
  */
+@Deprecated
 public abstract class ReactFragmentActivity extends FragmentActivity implements
   DefaultHardwareBackBtnHandler, PermissionAwareActivity {
 


dulmandakh:
@hramos what is your opinion about this PR? I think this will give developers to change their code, then cleanup RN codebase to remove outdated Android versions.

hramos:
SGTM

kmagiera:
I don't think this is the right approach. It is still recommended by android team to use classes from support library instead of relying on the corresponding classes from SDK. It makes it more future proof and avoid compatibility related problems when new methods gets added.


For source please see this paragraph on the link you cited: https://developer.android.com/topic/libraries/support-library/
screen shot 2018-08-03 at 12 37 09


Also note that android is deprecating a number of methods that's been a part of SDK and recommend to use corresponding methods from the support library. For example `getFragmentManager` is marked as deprecated as of API 28 (which is the most recent API version) and it is recommended that `FragmentManager.getSupportFragmentManager` is used instead that is a part of `support-v4` lib. Source: https://developer.android.com/reference/android/app/Activity.html#getFragmentManager()
screen shot 2018-08-03 at 12 41 34


If we remove `ReactFragmentActivity` that would prevent browfield RN apps from using fragments in android 28 and above.

My suggestion is to actually follow the android team recommendation I posted in first paragraph and change `ReactActivity` to extend `android.support.v4.app.ActivityCompat` instead of plain `android.app.Activity`. Then we can leave `ReactFragmentActivity` as deprecated but instead of removing it in the near future we should just change it to extend `ReactActivity` directly and became an empty class. Note that such a change will have no effect on that class consumers as `ActivityCompat` extends `FragmentActivity`.

hramos:
Sounds good. @dulmandakh do you think you can follow @kmagiera's suggestion above and send a PR?

dulmandakh:
@kmagiera got it, @hramos will create a PR shortly.

dulmandakh:
@hramos @kmagiera made some quick changes, and it seems that it requires more work than expected. I'll try to make it work tomorrow.

dulmandakh:
@kmagiera here is the PR to that incorporated your suggestions https://github.com/facebook/react-native/pull/19950

PR #19753, opened 2018-06-16
remove redundant babel dependency

gengjiawen:
## Motivation
Remove babel for it already included by metro
## Test Plan
pass all current ci.
## Related PRs
none
## Release Notes
 [GENERAL] [INTERNAL] [MISC] - remove redundant dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,6 @@
     "yargs": "^9.0.0"
   },
   "devDependencies": {
-    "@babel/core": "7.0.0-beta.47",
     "async": "^2.4.0",
     "babel-eslint": "8.2.1",
     "babel-generator": "^6.26.0",


vovkasm:
But it is required by `jest/preprocessor.js`, does it safe to make such dependency implicit?

gengjiawen:
It's not here in the first place. It's added in solving ci when upgrade to babel 7. I think since metro has all the babel packages, there is no need to declared it multi times.

vovkasm:
But it's here at least today (2018-06-17): https://github.com/facebook/react-native/blob/15f84420f0c53f53f11a93b609385ca741379898/jest/preprocessor.js#L15

File configured to be used by jest: https://github.com/facebook/react-native/blob/15f84420f0c53f53f11a93b609385ca741379898/package.json#L21-L24

And jest can run against react-native working copy:
https://github.com/facebook/react-native/blob/15f84420f0c53f53f11a93b609385ca741379898/package.json#L124-L126

So it is clearly dev dependency of RN. The fact that `@babel/core` required by `metro` do not remove this dependency, imho. But may be RN devs can manage dependency between RN, jest and metro consistently (but they can't), then ok, it's redunant.

gengjiawen:
When use babel 6, it's there too. Check the git history.

PR #19760, opened 2018-06-16
use same signing config for debug build

gengjiawen:
## Motivation
use same signing config for debug build 
## Test Plan
pass all current ci.
## Related PRs
none
## Release Notes
 [GENERAL] [INTERNAL] [RNTester] - use same signing config for debug build .


gengjiawen:
@janicduplessis @hramos

Issue #19793, opened 2018-06-18
NoSuchKeyException Thrown From ReadableNativeMap


Bhavik-P:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
Run `react-native info` in your terminal and paste its contents here.

Environment:
  OS: macOS High Sierra 10.13.4
  Node: 10.3.0
  Yarn: Not Found
  npm: 6.1.0
  Watchman: 4.9.0
  Xcode: Xcode 9.4 Build version 9F1027a
  Android Studio: 3.1 AI-173.4819257

Packages: (wanted => installed)
  react: 16.3.2 => 16.3.2
  react-native: 0.55.4 => 0.55.4

## Description
Describe your issue in detail. Include screenshots if needed. If this is a regression, let us know.

After we recently upgraded our React Native package to 0.55.4, I’ve started to notice an Android-specific bug that is hit from within the library. Hundreds of our users hit this error daily. The bug is thrown in ReadableNativeMap:122 because the app is trying to look for a key with name “toValue” however it’s not in the config - FrameBasedAnimationDriver.java:44. We generally see this happening on slower devices, but it’s not limited to slower devices. We’ve seen majority of the events happen on these devices:
LG K20 Plus LGMP260
Samsung Galaxy S7 - SM-G930V
ZTE Blade™ Z MAX Z982

The only way I am able to recreate it on my own is by adding a debugger in the native code as the app is launching and then pressing continue to step out of debugging mode. For example, I make a call to load my in-app products by calling querySkuDetailsAsync at the start on the app. Within the onSkuDetailsResponse function, I add a debugger. After the debugger is hit and I press continue, the error is recreated. This leads me to believe it is a timing issue (race condition).

Here is the stacktrace:

```
06-18 19:26:55.111 7871-7871/app.hurdlr.com E/unknown:ReactNative: Exception in native call
    com.facebook.react.bridge.NoSuchKeyException: toValue
    	at com.facebook.react.bridge.ReadableNativeMap.getValue(ReadableNativeMap.java:122)
    	at com.facebook.react.bridge.ReadableNativeMap.getDouble(ReadableNativeMap.java:147)
    	at com.facebook.react.animated.FrameBasedAnimationDriver.resetConfig(FrameBasedAnimationDriver.java:44)
[...]

PR #19808, opened 2018-06-19
Check if config contains a key before fetching. Default to 0

Bhavik-P:
Potential Fix to #19793 

## Test Plan

modified the code and tried to recreate the bug and was unable to. The red screen never popped up and nothing else seemed to be affected in a negative way.

## Release Notes

[ANDROID] [BUGFIX] [FrameBasedAnimationDriver.java] - Safely unwrapping ReadableMap by defaulting to 0 if key not present.


Bhavik-P:
Signed the CLA

PR #19834, opened 2018-06-21
Fix `currentlyFocusedField` by Removing `this` usage in TextInputState

janicduplessis:
I broke `currentlyFocusedField` when adding it back in ce3b7b8204dad0fd62a76a0ce66472eca4b25bc8 because `this` no longer refers to the proper object because it is assigned here https://github.com/facebook/react-native/commit/ce3b7b8204dad0fd62a76a0ce66472eca4b25bc8#diff-b48972356bc8dca4a00747d002fc3dd5R330. This code was pretty prone to breaking so I simply removed the `this` usage and rely on a top level variable instead. Also moved everything to named functions.

Test Plan:
----------
Tested that calling currentlyFocusedField actually returns a value if a text input is focused.

Release Notes:
--------------

[GENERAL] [BUGFIX] [TextInputState] - Fix `currentlyFocusedField` by Removing `this` usage in TextInputState

### DIFF
--- a/Libraries/Components/TextInput/TextInputState.js
+++ b/Libraries/Components/TextInput/TextInputState.js
@@ -18,73 +18,74 @@
 const Platform = require('Platform');
 const UIManager = require('UIManager');
 
+let currentlyFocusedID: ?number = null;
 const inputs = new Set();
 
-const TextInputState = {
-  /**
-   * Internal state
-   */
-  _currentlyFocusedID: (null: ?number),
+/**
+ * Returns the ID of the currently focused text field, if one exists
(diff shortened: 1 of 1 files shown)


blairvanderhoof:
Just found this issue, realized that currentlyFocusedField was returning undefined and thus not properly avoiding the keyboard from this library: https://github.com/APSL/react-native-keyboard-aware-scroll-view/blob/master/lib/KeyboardAwareHOC.js#L301. Patching this PR does fix the issue.

janicduplessis:
cc @hramos

PR #19856, opened 2018-06-22
[CI] Remove pull request check on npm deploys

hramos:
The publish script will fail on forked PRs anyway as the $CIRCLE_NPM_TOKEN envvar will be missing or incorrect.

We also move buck fetches to their own own shell script. These are shared by the Android and Deploy jobs, and using -ex will allow us to see which specific command failed without the need to list all steps in the config file.

Finally, cache keys are updated as architecture is only relevant in caches that may be reused across macOS and linux, which is not the case for Android.

Test Plan: Circle CI 

Release Notes:
[INTERNAL] [MINOR] [Circle] - Remove deploy check and update cache keys


.circleci/config.yml
```diff
@@ -646,6 +597,47 @@ jobs:
             fi
           when: always
 
+  # Publishes new version onto npm
```
hramos: Moving this block down makes PR review less convenient, I am afraid.

Here's what has changed. The following if/else block was removed (leaving the contents of the if block in place):

```
if [ -z "$CIRCLE_PULL_REQUEST" ]; then
 ...
else	
  echo "Skipping deploy."	
fi
```

The "save-*-cache" steps have been removed. No need to slow down a deploy while waiting for a cache to upload. This is already a noop in most cases because the cache is already up to date from the preceding jobs.

Consolidating `buck fetch` steps into a single script, matching `test_android`.

hramos:
`test_end_to_end` and `test_objc` were failing on master when the PR was opened.

PR #19881, opened 2018-06-25
fix android ci

dulmandakh:
This PR fixes buck and CI
CI: https://circleci.com/gh/dulmandakh/react-native/533

CC: @hramos 

Test Plan:
----------
CI is green again


hramos:
Thanks for the fix @dulmandakh. I just checked with the Buck team and they will push out a new Buck release that should include support for this feature. We probably will not need to apply this patch.

hramos:
Looks like the new Buck release does fix CI. Closing this PR in favor of #19882. Again, thanks for jumping on this quickly!

PR #19890, opened 2018-06-26
bump glog to 0.3.5

dulmandakh:
bump glog to 0.3.5.

Version 0.3.4 added support for libc++ or clang. Starting with revision 11, Android NDK recommends and defaults to clang, so it'll add support for it.

notable changes in 0.3.4 and 0.3.5:
* add libc++ support
* reduce dynamic allocation from 3 to 1 per log message
* style fix for C++11
* Add CMake support

Test Plan:
----------
Builds and runs as usual


dulmandakh:
@hramos please merge 😄

PR #19536, opened 2018-06-01
Build 64bit binaries for Android

marafat:
## Motivation
Add 64bit support to Android. Fixes #2814

**How is this different from the previous effort done in #17640 and #18754**
1. This PR uses the JSC from the SoftwareMansion's [fork](https://github.com/SoftwareMansion/jsc-android-buildscripts) [mentioned](https://github.com/facebook/react-native/issues/2814#issuecomment-391825614) under #2814 
2. It builds with NDK 16/clang
3. it updates the `APP_STL` to `c++_shared`

## Test Plan
1. Make the RNTester app build a separate APK per abi
2. assemble debug and release APK
```shell
$ cd react-native
react-native$ rm -rf ./node_modules && yarn install
react-native$ ./gradlew :ReactAndroid:cleanReactNdkLib :RNTester:android:app:clean :RNTester:android:app:assembleRelease
```
I found these to be beneficial against some build errors:
```shell
$ cd react-native
react-native$ rm -rf ./.buckd/
react-native$ rm -rf ./buck-out/
```
3. use `adb install` to install the `arm64-v8a` APK on a PIXEL 2XL
```shell 
$ cd react-native
react-native$ adb install ./RNTester/android/app/build/outputs/apk/app-arm64-v8a-release.apk
```
4. the app runs fine in both cases.

## Related PRs
[X] ~~Todo: change [instructions](https://github.com/facebook/react-native-website) on building from source.~~ facebook/react-native-website#394

## Release Notes
[ANDROID] [ENHANCEMENT] [ReactAndroid] - add x86_64 arm64-v8a support.

PR #19893, opened 2018-06-26
bump Android NDK to r14b

dulmandakh:
Bump Android NDK to r14b.

NDK tool chain is set to GCC 4.9 (in *Application.mk* file), because clang fails to build RN.

Below are notable changes to r14b
> GCC is no longer supported. It will not be removed from the NDK just yet, but is no longer receiving backports.

and

> NDK_TOOLCHAIN_VERSION now defaults to Clang.

CI: https://circleci.com/gh/dulmandakh/react-native/573

Test Plan:
----------
Builds and runs as before

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [NDK] - bump version to r14b


dulmandakh:
RN community, let's make it compile with clang 😄

gengjiawen:
This pr https://github.com/facebook/react-native/pull/19536 bump ndk to 16, r13b maybe to old.

dulmandakh:
IMHO, NDK 13 requires almost no change in source code and still builds the source code. I see it as a small, but non-disruptive step in supporting newer NDK versions and 64bit CPUs.

dulmandakh:
Bumped NDK to r14b, which became possible with https://github.com/facebook/react-native/commit/8ce758f48adc263924db0349d3b7059e52fc0be7.

I couldn't build with versions above 14, and I'm not proficient enough in C++ to fix build errors.

CC: @hramos

gengjiawen:
I still prefer to the latest ndk, since it's for all cpu arch, not just arm64.

dulmandakh:
@gengjiawen me too, if you read through revision history NDK is not adding support for CPU architectures but removing them.

NDK is switched from GCC to Clang, GNU C++ standard library to Clang's libc++ and unified headers therefore it might require some code changes. For example, folly github page states that it requires GCC to build. If merged, we'll land clang, libc++ and unified headers to play with.

I think it's better to have small incremental changes rather than nothing or disruptive big change.

dulmandakh:

gengjiawen:
@ulmandakh I think the removed abi won't has effect for this project, since they removed the abi is no one used or deprecated as you can see https://github.com/android-ndk/ndk/wiki/Changelog-r17. I agree with we should small incremental changes, but I think the latest ndk is okay.

dulmandakh:
Please review @fkgozali @hramos @mdvacca . Once merged I would like help with landing Unified Headers introduced in NDK r14 and default in r15.

* https://android.googlesource.com/platform/ndk.git/+/ndk-r14-release/docs/UnifiedHeaders.md
* https://android.googlesource.com/platform/ndk/+/ndk-r15-release/docs/UnifiedHeadersMigration.md

PR #19899, opened 2018-06-26
bump gradle to 4.4

dulmandakh:
Bump gradle to 4.4, version used by android studio 3.x and gradle plugin 3.x. This will help make migration easier and smoother.

CI: https://circleci.com/gh/dulmandakh/react-native/591

FYI: I found that gradle 4.8 works fine with gradle plugin 2.3, but gradle plugin 3.x displays warnings and asks to disable on demand configuration.

Test Plan:
----------
Builds and runs as usual

Release Notes:
--------------

[ANDROID] [ENHANCEMENT] [GRADLE] - bump to 4.4

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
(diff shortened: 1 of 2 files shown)


dulmandakh:

hramos:
@dulmandakh this won't be merged just yet, as I need to update our internal Gradle tests to use 4.8 as well. I've imported your changes and I'll land them alongside my internal test changes. It's on my queue, however, so it should not be long.

dulmandakh:
@hramos it's 4.4. Version 4.8 will cause problems if used with Gradle
plugin 3.1, maybe 3.0.

slorber:
@hramos in the meantime, is there any documentation on how to dev RN apps as of today? Because latest Android Studio version requires Gradle 4.4 as minimum version.

Just wondering what is the best option as of today to solve my problems. Maybe downgrading to 2.x AS is the solution: https://developer.android.com/studio/archive

hramos:
@slorber you can update the `gradle-wrapper.properties` file in your own project even if this PR is not yet merged. It should not be long, though this won't make it into a release until 0.57 most likely.

PR #19910, opened 2018-06-26
Fix: D8450498 breaks test_end_to_end job

hramos:
Summary: Because of the way react-native local-cli communicates with metro, getWatchFolders function was always returning an empty array that triggered an error.

Differential Revision: D8650733


hramos:
Verified. Closing as this will land to the internal repo soon.

PR #19919, opened 2018-06-27
Update window size when the in-call status bar is toggled

janicduplessis:
When the in-call status bar is toggled, the Dimensions module window doesn't resize properly. The current code assumes that window and screen is always the same on iOS but it is not the case for when the in-call status bar is on. The viewport size is reduced by 20px in that case.

This results in the following behaviour changes:
- The `Dimensions` module `change` event is now triggered when the in-call status bar is toggled
- The `window` and `screen` values of the `Dimensions` module are not always the same on iOS, currently only different when the in-call status bar is opened.
- The `Dimensions` module `change` event is now triggered when the status bar is shown / hidden. Right now this is not ideal but I'm planning to add `safeAreaInsets` to the `Dimensions` module and in that case the change event will be meaningful. 

`RCTKeyWindow().rootViewController.view.bounds` seemed the best way I found to get the proper value for when the in-call status bar is on. Might want to verify this also work properly in non-pure RN apps.

Test Plan:
----------
- Tested that a view that has exactly the size of the window (using `Dimensions.get('window')` still fits properly when the in-call status bar is on.
- Tested that the `Dimensions` module values are still updated properly on screen rotation (and only once). Also works if the status bar is hidden.

Release Notes:
--------------
[IOS] [BREAKING] [Dimensions] - Update window size when the in-call status bar is toggled

### DIFF
--- a/React/Modules/RCTDeviceInfo.m
+++ b/React/Modules/RCTDeviceInfo.m
@@ -13,11 +13,7 @@
 #import "RCTUIUtils.h"
 #import "RCTUtils.h"
 
-@implementation RCTDeviceInfo {
-#if !TARGET_OS_TV
-  UIInterfaceOrientation _currentInterfaceOrientation;
-#endif
-}
+@implementation RCTDeviceInfo
 
 @synthesize bridge = _bridge;
 
@@ -42,11 +38,9 @@ RCT_EXPORT_MODULE()
                                                name:RCTAccessibilityManagerDidUpdateMultiplierNotification
                                              object:_bridge.accessibilityManager];
 #if !TARGET_OS_TV
-  _currentInterfaceOrientation = [RCTSharedApplication() statusBarOrientation];
-
   [[NSNotificationCenter defaultCenter] addObserver:self
-                                           selector:@selector(interfaceOrientationDidChange)
-                                               name:UIApplicationDidChangeStatusBarOrientationNotification
+                                           selector:@selector(applicationDidChangeStatusBarFrame)
+                                               name:UIApplicationDidChangeStatusBarFrameNotification
                                              object:nil];
 #endif
 }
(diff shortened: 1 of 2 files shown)


vovkasm:
But in-call status bar does not reduces window height. Same as normal status bar, it draws over application window, but not changes it's size. References:
- https://developer.apple.com/design/human-interface-guidelines/ios/bars/status-bars/
- I can't currently found exact description in apple docs, but hope this post also shed some light on this topic: https://forums.developer.apple.com/thread/18351

janicduplessis:
@vovkasm I haven’t tested on all iOS versions but on iOS 11 it does resize the window by 20pt when the in call status bar is toggled. I assume apple did this to make it easier to manage since you can assume the status bar always covers 20pt at the top of the screen, instead of having to consider that it sometimes cover 40pt when the in call status bar is on.

vovkasm:
@janicduplessis I just did some tests:
1. Normal status bar (we can see normal UIWindow size, and normal root view, that covers whole UIWindow):
Normal status bar

2. In-call status bar (we still see normal UIWindow size, but UIViewController automatically adjust constraints of its view)
In-call status bar

I'm as a developer can set any constraints in UIViewController and make RCTRootView do not relayout on in-call status bar. Right?

vovkasm:
I think what you want to use `Dimensions` for, can be easily implemented by `onLayout` props of root view, or its child with `felx: 1` style. Window, screen and root view can all be different in size and we need all these metrics.

janicduplessis:
With the default RN setup, window size should be the same as the rootview. You can always get the real screen size using “screen”. In this case I am just measuring the root VC so if you decide to remove thr default 20 top constraint it would still return the right value.

vovkasm:
> With the default RN setup, window size should be the same as the rootview.

1. Why it should? ))) It is not the case currently (as we can see above) and if this should be fixed, then it should be fixed by instantiate custom VC (say `RCTRootViewController`) instead of pristine `UIViewController` that will implement proper layout logic for its root view.

2. The default RN setup - not the only one used.

> You can always get the real screen size using “screen”.

Screen - is totally different thing... Do not mix it with window please.
- Screen size - should be constant for main screen of application. And can change only if application needs to manage distinct view on other screen (as I know RN currently can't deal with this without native coding)
- Window size - is the area where an application can draw. In split view mode on iPad it will differ from screen size

And again, why you can't get size of root view with onLayout event? It will work always - for rotations, for split view, for custom layout logic in custom view controller, etc...

janicduplessis:
Alright, let's leave this as is. I'm working on a new api to get the root view layout metrics instead.

PR #19921, opened 2018-06-27
[WIP] Add safeAreaInsets to the Dimensions module

janicduplessis:
There are some use cases where `SafeAreaView` doesn't work because we need to do some computations or apply the values manually in JS (ex.: margin instead of padding). This exposes the safe area insets of the screen as a new value on the `Dimensions` module.

Kind of depends on #19919

TODO:
- RNTester example
- Docs
- Test iOS < 11 polyfill
- Android

Test Plan:
----------
TODO

Release Notes:
--------------
[GENERAL] [FEATURE] [Dimensions] - Add safeAreaInsets to the Dimensions module

### DIFF
--- a/React/UIUtils/RCTUIUtils.m
+++ b/React/UIUtils/RCTUIUtils.m
@@ -21,6 +21,24 @@ RCTDimensions RCTGetDimensions(CGFloat fontScale)
   result.window = dims;
   result.screen = dims;
 
+  if (@available(iOS 11.0, *)) {
+    UIEdgeInsets windowInsets = RCTKeyWindow().rootViewController.view.safeAreaInsets;
+    typeof (result.safeAreaInsets) safeAreaInsets = {
+      .left = windowInsets.left,
+      .top = windowInsets.top,
+      .right = windowInsets.right,
+      .bottom = windowInsets.bottom
+    };
+    result.safeAreaInsets = safeAreaInsets;
+  } else {
+    // Polyfill `safeAreaInsets` for iOS < 11. We only care about the status bar.
+    typeof (result.safeAreaInsets) safeAreaInsets = {
+      // TODO: Test this with in-call status bar.
+      .top = RCTSharedApplication().statusBarFrame.size.height
+    };
+    result.safeAreaInsets = safeAreaInsets;
+  }
+
   return result;
 }
 
--- a/Libraries/Utilities/Dimensions.js
+++ b/Libraries/Utilities/Dimensions.js
@@ -66,6 +66,7 @@ class Dimensions {
       eventEmitter.emit('change', {
         window: dimensions.window,
         screen: dimensions.screen,
+        safeAreaInsets: dimensions.safeAreaInsets,
       });
     } else {
       dimensionsInitialized = true;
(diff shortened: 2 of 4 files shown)


React/UIUtils/RCTUIUtils.m
```diff
@@ -21,6 +21,24 @@ RCTDimensions RCTGetDimensions(CGFloat fontScale)
   result.window = dims;
   result.screen = dims;
 
+  if (@available(iOS 11.0, *)) {
+    UIEdgeInsets windowInsets = RCTKeyWindow().rootViewController.view.safeAreaInsets;
```
patrickkempff: @janicduplessis why aren't you using `RCTKeyWindow().safeAreaInsets`? It does not force loading the view.

janicduplessis: I tries using that but the insets were always 0.

patrickkempff: Oh that is strange. Just to verify you are not running the code in a app extension right? Because RCTSharedApplication is returning nil when RCTRunningInAppExtension returns true. 

As an example we are using this is a large scale production app:

```objc
- (NSDictionary *) constantsToExport {
    return @{
             @"SAFE_AREA_INSETS_TOP": @([self safeAreaInsetsTop]),
             @"SAFE_AREA_INSETS_BOTTOM": @([self safeAreaInsetsBottom])
             };
}

- (CGFloat) safeAreaInsetsTop {
    if (@available(iOS 11.0, *)) {
        return UIApplication.sharedApplication.keyWindow.safeAreaInsets.top;
    }
    return 0;
}

- (CGFloat) safeAreaInsetsBottom {
    if (@available(iOS 11.0, *)) {
        return UIApplication.sharedApplication.keyWindow.safeAreaInsets.bottom;
    }
    return 0;
}
```

I wanted to open a PR myself but awesome that you are doing this already. thanks!

patrickkempff:
Android P has support for the "notch" (or other type of insets) as well:

```js
getSafeInsetTop()
getSafeInsetBottom()
getSafeInsetLeft()
getSafeInsetRight()
```

https://developer.android.com/reference/android/view/WindowInsets#getDisplayCutout()

If you want some help I can join to implement the android side. Let me know!

React/UIUtils/RCTUIUtils.m
janicduplessis: Weird, I can try to investigate this further. I agree it would be cleaner to user keyWindow.safeAreaInsets.

janicduplessis:
@patrickkempff Sure, I planned adding support for Android too but help would be appreciated! If you want you could open a PR on top of this one that adds support for Android.

janicduplessis:
Ideally we would also polyfill the notch API to work on older Android versions. It would return the proper top offset when using translucent status bar.

patrickkempff:
@janicduplessis will invest time this weekend into this

clmntcrl:
Hi, as you could see here #19627, I used another approach to get safe area insets values on JS side (for iOS only).

janicduplessis:
@clmntcrl I think the issue with this approach is that you cannot have access to the insets in a synchronous way, before the first render. This could cause some flickers on initial mount. I still think your PR can be useful, I'll have a look soon.

janicduplessis:
Closing in favor of #20999 which is a lot less hacky

PR #19951, opened 2018-06-28
bump metro to 0.40.1 in babel-preset

dulmandakh:
Bump metro to 0.40.1 in babel-preset, that is left out from https://github.com/facebook/react-native/commit/770dd38bbbec20e40e4895d512da19af20e0d0d0.

Test Plan:
----------


hramos:
cc @mjesun - not sure why this one was missed in 770dd38, but I'm hesitant to import this as we have a process in place to upgrade metro across all our internal surfaces which resulted in 770dd38.

hramos:
I just chatted with @rafeca about this. `babel-preset-react-native` will be moved to the metro repository shortly, and as a result it will be upgraded alongside the other metro packages in the future.

PR #19965, opened 2018-06-29
[iOS] Expose enabled prop on RCTPicker

Moreno97:
**Test Plan:**
----------
All tests passed (flow, jest) and work as expected.

Changelog:

With this change on iOS API, the user can't change the value of the UIPickerView is enabled prop is false (works like Android, the goal is unify both platforms to work as expected)

[iOS] [Added] PickerIOS - Expose enabled prop on RCTPicker


Moreno97:
Hello, someone to review that? Maybe @shergin ??

hramos:
Attempted to land this, but it's effectively a breaking change as `flow check` now fails on any app that fails to set this prop on Picker. It affects many of our own internal apps.

This should not be a blocker necessarily, but I do not have the bandwidth to perform this work right now.

Moreno97:
@hramos ok, ping me if I've to do something else or if there's news about the review of the PR.

hramos:
@Moreno97 I think this would be on us, as it would require updating every app that uses PickerIOS at Facebook.

Moreno97:
Hi @hramos , some news about this PR??

Moreno97:
@shergin @hramos any updates? Hope to see this merged on the repo!

radeno:
@jainkuniya @cpojer PickerIOS and PickerAndroid is merged together with this PR https://github.com/facebook/react-native/pull/23359 as a Lean Core project. 

@Moreno97 @jainkuniya will be still an issue after merging?

cpojer:
Yes, our plan is to merge the two components and then extract them out of this repo, so this PR will have to be reapplied to the extracted repo soon. I'm gonna close this PR here since we aren't likely to merge it at this time.

PR #19972, opened 2018-06-29
Add back ImageEditingExample example

gengjiawen:
## Motivation
add back missing file.
## Test Plan
pass all current ci.
## Related PRs
none
## Release Notes
 [GENERAL] [INTERNAL] [RNTester] - Add back ImageEditingExample example.

### DIFF
--- a/RNTester/js/ImageEditingExample.js
+++ b/RNTester/js/ImageEditingExample.js
@@ -1,9 +1,0 @@
-/**
- * Copyright (c) 2015-present, Facebook, Inc.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- * @format
- * @flow strict
- */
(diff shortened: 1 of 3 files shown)


TheSavior:
Woah, interesting find. 

It looks like this file got accidentally messed up in this commit: https://github.com/facebook/react-native/commit/4a80dceac7e35b669301c7aa8180dcd0348f4e93#diff-9bff2388f38e11550ff316a7e78092bd

Instead of deleting this file could you add back the original version that was deleted in that commit?

gengjiawen:
@TheSavior I add it back. I check the file git history and see it it empty in first place, not realizing that it was deleted accidentally :)

TheSavior:
Thanks! Can you confirm that it works in RNTester now?

Also, it would be great if you migrated this file away from var to let/const. We have done that for most everything else in the codebase already but this file got missed since it was accidentally deleted.

gengjiawen:
Currently I only have an windows computer, but RNTester doesn't work on windows see https://github.com/facebook/react-native/pull/19974. 
As for the var issue, most of RNTester still has the issue and use commonjs, I have a plan to migrate them all. But that's another story.

gengjiawen:
I test it on a mac via teamviewer after I grant permission , looks like ui rendered correctly

image

PR #20007, opened 2018-07-02
Hotfix to include react-native-windows in hasteImpl accepted paths

hramos:
Summary:
We removed support for providesModule annotations and maintained support for Haste names in installed modules via `providesModuleNodeModules`, but our default `hasteImpl` doesn't take them into account. We need to find a better way to override core components from plugins but meanwhile this adds an exception for react-native-windows in the default `hasteImpl` to unblock their upgrade to the latest RC.

Fixes https://github.com/facebook/metro/issues/188

Reviewed By: mjesun

Differential Revision: D8695207

### DIFF
--- a/jest/hasteImpl.js
+++ b/jest/hasteImpl.js
@@ -12,7 +12,10 @@
 
 const path = require('path');
 
-const ROOT = path.join(__dirname, '..');
+const ROOTS = [
+  path.resolve(__dirname, '..') + path.sep,
+  path.resolve(__dirname, '../../react-native-windows') + path.sep,
+];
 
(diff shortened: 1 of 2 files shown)


hramos:
Test passed. Closing.

rozele:
Thanks @hramos - what's the status update on a generalization of this? I still need a solution for the `react-native-windows` dev environment, where the "haste overrides" aren't necessarily in the `react-native-windows` folder in `node_modules` (they actually in the `Libraries` folder at the root of the project).

hramos:
cc @rubennorte, original author of the diff that synced out as 5494274.

PR #14792, opened 2017-07-01
Add gradle support for multi-dimensional product flavors

wsxyeah:
#14791 

Replace `productFlavors.each ` with `applicationVariants.each` to support multi-dimensional product flavors.

PR #20012, opened 2018-07-03
React native gradle support for flavor dimensions

ishaanbahal:
Related info -> https://developer.android.com/studio/build/build-variants

Currently react native gradle only supports `{flavor}{BuildType}` for example, `devDebug`, `qaDebug`, `qaRelease` but android now also supports `flavorDimensions` which can be utilised by people to have different environments like prod, staging etc and make builds according to those, `{flavor}{flavorDimension}{buildType}` is the final name for it, example could be qaStagingDebug, qaProdDebug, devProdRelease, etc.

To support these one can simply collect `applicationVariants` instead of `productFlavors` and `buildTypes`. Application variants give full flavorName with dimension and buildType is also available to it.

Similar PR opened a while ago https://github.com/facebook/react-native/pull/14792 for the same feature.

Integrating react-native into an existing app which uses flavors is a challenge and we had to take a fork of it in my organisation. Please inform of any other testing steps needed to be done.

The changes may seem quite a lot, but actually only 3 line have been modified, 2 added and one changed, the change is large due to the reduced indent caused by the removal of a loop on `buildTypes` 

Test Plan:
----------
- Tested this on `create-react-native-app` project after eject from expo (don’t know how to build with expo tbh)
screen shot 2018-07-03 at 2 10 49 pm

- Tested in a project with product flavors like `qaDevStagingDebug`, `qaDevProdDebug` etc, where `prod` and `staging` are flavor dimensions.

> Testing done on MacOS only.

Release Notes:
--------------

[ANDROID] [ENHANCEMENT] [Gradle] - Added flavor dimension support using application variants

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -27,110 +27,111 @@ afterEvaluate {
     def isAndroidLibrary = plugins.hasPlugin("com.android.library")
     // Grab all build types and product flavors
     def buildTypes = android.buildTypes.collect { type -> type.name }
-    def productFlavors = android.productFlavors.collect { flavor -> flavor.name }
+    def productFlavors = android.applicationVariants.collect { flavor -> flavor }
 
     // When no product flavors defined, use empty
     if (!productFlavors) productFlavors.add('')
 
-    productFlavors.each { productFlavorName ->
-        buildTypes.each { buildTypeName ->
-            // Create variant and target names
-            def flavorNameCapitalized = "${productFlavorName.capitalize()}"
-            def buildNameCapitalized = "${buildTypeName.capitalize()}"
-            def targetName = "${flavorNameCapitalized}${buildNameCapitalized}"
-            def targetPath = productFlavorName ?
-                    "${productFlavorName}/${buildTypeName}" :
-                    "${buildTypeName}"
+    productFlavors.each { productFlavor ->
+        def productFlavorName = productFlavor.flavorName
+        def buildTypeName = productFlavor.buildType.name
 
-            // React js bundle directories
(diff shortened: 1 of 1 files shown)


dulmandakh:
This is not necessary once Android App Bundling becomes Stable, which will be very soon. App Bundling will generate separate APKs for each architecture, screen sizes and users will automatically download APK that suits their device.

ishaanbahal:
@dulmandakh I don't think app bundle will play a role for flavor dimensions. You can for instance create a separate branch for every environment your APIs support and give those to QA and do other things. App bundle only comes to picture when you're pushing to playstore. The purpose for app bundle is mostly to split apk into parts and do dynamic features. 

The change requested here supports `flavorDimensions`. Read https://developer.android.com/studio/build/build-variants. This feature isn't for users, but for devs that have flavorDimensions in their app to be able to bundle react native with their app using the react.gradle config and not manually bundle the file everytime they there is a release or debug app created.

wangluu:
this is a useful function ,and I need it recently.  so when the RN release version can include it?

ishaanbahal:
@wangluu my company maintains a separate fork of RN for the ever so slow bug fixes and other such reasons. I suggest you try that too, because it may take sometime for this to get into stable or it even might not.

wangluu:
@ishaanbahal thanks for your reply, marked and  I will try it then.

cpojer:
Thanks for this PR! Unfortunately we have made large changes to the gradle file since this PR was last updated (I'm sorry for taking so long to get back to you). I'm going to close this PR but feel free to send a version of this that's fully rebased on master and I'll look at it promptly.

Issue #16906, opened 2017-11-21
[Android] Requiring images with Gradle 4 fails release builds 


Almouro:
Hello guys,

I ran into this issue when updating Gradle and Android Studio. 

Thanks for your help!

### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.5.0
  Yarn: 1.3.2
  npm: 5.3.0
  Watchman: 4.7.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 3.0 AI-171.4408382

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: 0.50.3 => 0.50.3

  Target Platform: Android (22)

### Steps to Reproduce

Requiring images with latest `gradle` fails release builds

I created a small [test project](https://github.com/Almouro/react-native-with-gradle-4) but here are the steps to reproduce it:

1. Init a new project: `react-native init Test`
2. Update to latest Gradle: 
  - Open Android Studio and install recommended version
  - Build from Android Studio, it should ask you to update build tools, do so
3. Require an image in `index.js`
4. Run `cd android && ./gradlew assembleRelease`

### Expected Behavior

The command `./gradlew assembleRelease` should not fail

### Actual Behavior

Running `./gradlew assembleRelease` when the image is required from the JS fails with:
```
/Users/almouro/bam/uefa/TestGradle3/android/app/build/intermediates/res/merged/release/drawable-mdpi/image.png: error: uncompiled PNG file passed as argument. Must be compiled first into .flat file..
error: failed parsing overlays.
```

If the image is not required, the command works fine.

### Reproducible Demo

```shell
git clone https://github.com/Almouro/react-native-with-gradle-4
yarn
cd android
./gradlew assembleRelease
```

### Workaround

Adding `android.enableAapt2=false` to `android/gradle.properties` fixes the issue but it is only a workaround.

macs03:
I have the same issue

strawferry:
+1 😂

vladp:
+1 have same issue with react-native 0.50
setting android.enableAapt2=false to android/gradle.properties  
helps with this issue. However, I am running into a next problem (and not sure if this due to the above setting).

> 
> > Task :app:bundleReleaseJsAndAssets
[...]

PR #20026, opened 2018-07-04
Bump Android SDK to 28

dulmandakh:
This PR will bump Android SDK to 28, build tools to 28.0.2 and support library to 27.1.1. But issue related to aapt2 is breaking RNTester app. 

## Solution

After some research I found that we need to compile (or copy) resources using **aapt2 compile** command. **aapt2** is located in  **${sdkDir}/${buildToolsVersion}** directory.

And resource bundling is handled in **copy** function of the **local-cli/bundle/saveAssets.js** file. In **react.gradle** file we can access **sdkDir** via **project.android.sdkDirectory**, **buildToolsVersion** via **project.android.sdkDirectory**.

The solution might be pass aapt2 path to bundler and use it to compile resources.

I'm not proficient in gradle, and would like to discuss to find better solution.

Please help @gengjiawen @hramos @kelset 

Resources
* https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt2/readme.md
* https://stackoverflow.com/questions/42082352/how-to-use-aapt2-where-is-the-documentation

Test Plan:
----------

Everything should build and run as usual.

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [SDK] - Bump to 27


dulmandakh:
There is no major or breaking change in 27, as it carries Oreo codename and 8.1 version number.

Once merged I would like to focus more on NDK and JSC part.

gengjiawen:
aapt2 through path is not the right way, I am looking into gradle dsl to find the related api. But I am not expert too. Any chance you know something we can hook @janicduplessis ?

AndrewJack:
Someone updated the gradle file to work with plugin v3 here https://github.com/facebook/react-native/blob/d16ff3bd8b92fa84a9007bf5ebedd8153e4c089d/react.gradle

but it was reverted.

try adding that back in?

gengjiawen:
Maybe not adding it back, since it's not a clean fix. We need aapt2 to support instant run.

MarkOSullivan94:
@gengjiawen do we need to support instant run? If not, could we not disable AAPT2 for now?

> If you are experiencing issues while using AAPT2, you can disable it by setting android.enableAapt2=false in your gradle.properties file and restarting the Gradle daemon by running ./gradlew --stop from the command line.

Source: https://developer.android.com/studio/build/gradle-plugin-3-0-0

gengjiawen:
This is some snippet in `AaptV2CommandBuilder`, maybe related. But I have find the method we can get the aapt2 path.
```java
    public static ImmutableList<String> makeCompile(@NonNull CompileResourceRequest request) {
        ImmutableList.Builder<String> parameters = new ImmutableList.Builder<>();

        if (request.isPseudoLocalize()) {
            parameters.add("--pseudo-localize");
        }

        if (!request.isPngCrunching()) {
            // Only pass --no-crunch for png files and not for 9-patch files as that breaks them.
            String lowerName = request.getInputFile().getPath().toLowerCase(Locale.US);
            if (lowerName.endsWith(SdkConstants.DOT_PNG)
                    && !lowerName.endsWith(SdkConstants.DOT_9PNG)) {
                parameters.add("--no-crunch");
            }
        }

        parameters.add("--legacy");
        parameters.add("-o", request.getOutputDirectory().getAbsolutePath());
        parameters.add(request.getInputFile().getAbsolutePath());

        return parameters.build();
    }
```

gengjiawen:
The commit doesn't handle aapt2. If you disable aapt2, current code works fine.

kelset:
Side-related to appt2 (and reason why I think we should care/have aapt2 support): https://github.com/facebook/react-native/issues/16906 (and by extension https://github.com/react-navigation/react-navigation/issues/3097)

AndrewJack:
Could keep both scripts in react native, if there's a concern that it'll break a lot of builds. 
`react.gradle` &`react-v3.gradle`

We are currently using the linked gradle script with appt2 enabled.

dulmandakh:
Also android.enableAapt2=false is breaking App Bundling, and we need to fix the issue.

janicduplessis:
If you try https://github.com/janicduplessis/react-native/blob/5d0d70b7388c1eea39f2fa568a3b52a985107ce2/react.gradle it should work without android.enableAapt2=false and assets should bundle properly. I'm looking at landing this soon.

jwaldrip:
Any update here?

dulmandakh:
CI is green again 😄

ReactAndroid/build.gradle
```diff
@@ -293,30 +293,30 @@ android {
...
+    implementation 'com.android.support:appcompat-v7:27.1.1'
+    implementation 'com.facebook.fbui.textlayoutbuilder:textlayoutbuilder:1.0.0'
+    implementation 'com.facebook.fresco:fresco:1.10.0'
+    implementation 'com.facebook.fresco:imagepipeline-okhttp3:1.10.0'
+    implementation 'com.facebook.soloader:soloader:0.5.1'
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
+    implementation 'com.squareup.okhttp3:okhttp:3.10.0'
+    implementation 'com.squareup.okhttp3:okhttp-urlconnection:3.10.0'
+    implementation 'com.squareup.okio:okio:1.14.0'
     compile 'org.webkit:android-jsc:r174650'
```
gengjiawen: Looks like forget to change here.

.circleci/config.yml
```diff
@@ -28,11 +28,11 @@ aliases:
```
gengjiawen: Appveyor need to change to.

dulmandakh: @gengjiawen change to what?

gengjiawen: I mean windows ci need related change too.

ReactAndroid/build.gradle
gengjiawen: I think maybe `api` is better.

.circleci/config.yml
dulmandakh: @gengjiawen sorry, I have no experience of Appveyor. Please provide me exact change to make.

dulmandakh:
Bumped Android SDK to 28, which was released just yesterday. We still use Support Library 27.x because 28.x is not stable yet.

.circleci/config.yml
gengjiawen: https://github.com/facebook/react-native/blob/master/appveyor.yml, change ANDROID_BUILD_VERSION and ANDROID_TOOLS_VERSION is okay

ReactAndroid/build.gradle
gengjiawen: I mean most of them, like okhttp, fresco, appcompat. And also the compile is still exist.

build.gradle
```diff
@@ -7,13 +7,10 @@ buildscript {
...
         mavenLocal()
-        maven {
-            url 'https://maven.google.com/'
-            name 'Google'
-        }
+        google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
+        classpath 'com.android.tools.build:gradle:3.1.3'
```
gengjiawen: Maybe 3.1.4.

dulmandakh: Changing compile to something else is breaking ReactAndroid build, and think that it needs some refactoring. IIRC, 3.1.3 is the latest Android Plugin for Gradle.

dulmandakh: IIRC, 3.1.3 is the latest version.

ReactAndroid/build.gradle
dulmandakh: Changing compile to something else if breaking ReactAndroid build. So I think that it requires some refactoring to fix the issue.

.circleci/config.yml
dulmandakh: @gengjiawen done

dulmandakh:
Android CI was green before Appveyor change requested by @gengjiawen, and it's broken due to issue on Circle CI.

ReactAndroid/build.gradle
gengjiawen: If you change it to api or implementation, what's the error ?

gengjiawen: Using `implementation` instead of `api` may break many things, since docs has show the usage too such as okhttp and fresco.

dulmandakh: @gengjiawen yes, you're right. I'll make changes to use api instead of compile

gengjiawen:
maybe change the title ?

dulmandakh:
Closing this PR, because I decided to break it into smaller and easy to review PRs.

PR #20097, opened 2018-07-08
Allow third party components to render strings

iyegoroff:
After introducing new invariant checks in 0.56.0 release, third party native components that can render strings became broken - e.g. https://github.com/iyegoroff/react-native-text-gradient/issues/11

This PR allows third party components to render strings if they have static property `canRenderString` set to `true`.

Usage:

```js
const ICanRenderString = createReactClass({
  statics: {
    canRenderString: true
  },
  ...
};

class ICanRenderStringToo extends React.Component {
  static canRenderString = true;
}
```

Test Plan:
----------
This can be tested on react-native-text-gradient example:
1) ```bash
     $ git clone git@github.com:iyegoroff/react-native-text-gradient.git
     $ cd react-native-text-gradient
     $ git checkout fb-test
     $ cd TextGradientExample
     $ npm i
     $ react-native run-ios
     ```
2) example app should compile and run without errors, on the screen you should see text "Welcome to React Native!"
3) optionally, you can check that react-native core components behavior is not changed:
  ```js
  <Text>text</Text> // OK
  <TextInput>text-input</TextInput> // OK
  <View>view</View> // Invariant Violation: Text strings must be rendered within a <Text> component
  ```
4) Open "Dev Settings", uncheck "JS Dev Mode", reload the bundle and repeat items 2 and 3

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [Renderer] - Allow third party components to render strings

### DIFF
--- a/Libraries/Renderer/oss/ReactFabric-dev.js
+++ b/Libraries/Renderer/oss/ReactFabric-dev.js
@@ -3733,14 +3733,15 @@ function getRootHostContext(rootContainerInstance) {
   return { isInAParentText: false };
 }
 
-function getChildHostContext(parentHostContext, type, rootContainerInstance) {
+function getChildHostContext(parentHostContext, fiber, rootContainerInstance) {
   var prevIsInAParentText = parentHostContext.isInAParentText;
   var isInAParentText =
-    type === "AndroidTextInput" || // Android
-    type === "RCTMultilineTextInputView" || // iOS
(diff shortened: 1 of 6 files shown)


janicduplessis:
Hi @iyegoroff, thanks for the PR!

The source for these files are actually located in the react repo (https://github.com/facebook/react/tree/master/packages/react-native-renderer/src) so you will have to open a PR there.

PR #20109, opened 2018-07-09
[GENERAL] Add .nvmrc + fix node version for eslint 5 compat

slorber:
A .nvmrc file is practical when working on many node projects with different version requirements.

Also, the current required node engine is incompatible with eslint5 requirements:

```js
  "engines": {
    "node": ">=8"
  },
```

It produces the following warning on install:

> error eslint@5.0.1: The engine "node" is incompatible with this module. Expected version "^6.14.0 || ^8.10.0 || >=9.10.0".
> error Found incompatible module

I made the requirement compatible with eslint5 requirements

Test Plan:
----------

`yarn test` is still passing

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] Add .nvmrc and ensure node version required is compatible with ESLint 5


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

slorber:
not sure exactly why ios e2e tests are failing on CI, any idea?

slorber:
Ok thanks, I'll update my PR with doc + checkNodeVersion to 8.10 soon

slorber:
@janicduplessis actually not sure if I should make the changes you wanted

ReactNative can actually work for node >= 8.3, it's just ESLint which requires 8.10. So it may be a good idea to have a .nvmrc 8.10 for contributors, but not necessary to force end users to use 8.10 as they won't run ESLint anyway. I don't think the engines attribute allow that granularity between distinct version requirement for devtools and consumer code.

Maybe it's better to let engine >= 8.3 and only add nvmrc for a recommended contributor version to use?

cc @hramos who changed the version recently: https://github.com/facebook/react-native/pull/20236/files

hramos:
@slorber that approach sounds good - let's use engine for end-user minimum required Node version, .nvmrc for contributors.

slorber:
ok,

I've reverted the PR's code to only add a .nvmrc and let the engine >= 8.3

PR #20111, opened 2018-07-09
[TEST] Add missing tests for Animated.forkEvent

slorber:
`forkEvent` is generally used to intercept an existing listener and add a new js listener to it, considering the original listener can be null/js/Animated.event(). I added tests to ensure the 3 cases of the implementation are all covered.

Test Plan:
----------
Tests are passing

Release Notes:
--------------

[INTERNAL] [ENHANCEMENT] [Animated] Add Animated.forkEvent tests for better coverage


janicduplessis:
Looks good, thanks!

janicduplessis:
@facebook-github-bot shipit

PR #20162, opened 2018-07-12
Fix local-cli assetRegistryPath and middlewares

### LINKED ISSUE #20008: [CI] Test failure - D8711717 test_objc, test_end_to_end
https://github.com/facebook/react-native/commit/c4a66a89a28152cd4b81e2b0f80ab3aac34d6872 breaks open source tests on `master` as argument mapping is not preserved. This should be fixed in a future commit.

janicduplessis:
This fixes some regressions with local-cli introduced in c4a66a89a28152cd4b81e2b0f80ab3aac34d6872. Fixes #20008.

- We didn't pass `assetRegistryPath` which caused the following error when loading the bundle:
```
error: bundling failed: Error: Unable to resolve module `missing-asset-registry-path` from `/Users/janic/Developer/react-native/RNTester/js/uie_thumb_normal@2x.png`: Module `missing-asset-registry-path` does not exist in the Haste module map
```
- The middlewares were not added to the metro server. This causes some packager server features to fail. The one I noticed is that the /status endpoint didn't exist anymore which causes CI to fail and also Android to not load the bundle from the packager initially. The remote debugging feature was also broken.

Test Plan:
----------
- Run RNTester and make sure the app loads properly.
- Test the /status endpoint to make sure middlewares are added.
- Test remote debugging and make sure it works as expected.

Release Notes:
--------------
Marking this as internal since the regression hasn't made it to a release.

[INTERNAL] [BUGFIX] [runServer.js] - Fix local-cli assetRegistryPath and middlewares

### DIFF
--- a/local-cli/server/runServer.js
+++ b/local-cli/server/runServer.js
@@ -16,9 +16,13 @@ const Metro = require('metro');
 
 const {Terminal} = require('metro-core');
 
+const messageSocket = require('./util/messageSocket');
 const morgan = require('morgan');
 const path = require('path');
+const webSocketProxy = require('./util/webSocketProxy');
 const MiddlewareManager = require('./middleware/MiddlewareManager');
+
+const {ASSET_REGISTRY_PATH} = require('../core/Constants');
 
 import type {ConfigT} from 'metro';
 
@@ -55,6 +59,9 @@ async function runServer(args: Args, config: ConfigT) {
   const serverInstance = await Metro.runServer({
     config: {
       ...config,
+      assetRegistryPath: ASSET_REGISTRY_PATH,
+      enhanceMiddleware: middleware =>
+        middlewareManager.getConnectInstance().use(middleware),
       hmrEnabled: true,
       maxWorkers: args.maxWorkers,
       reporter,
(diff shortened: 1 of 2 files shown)


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

janicduplessis:
CI is back to green with this :)

janicduplessis:
I found more issues with remote debugging because we don't attach the websocket proxy, looking into a fix.

janicduplessis:
Looks like hmr is not working also 🙃

janicduplessis:
A lot of config keys were on the wrong object, hmr now works.

local-cli/server/runServer.js
```diff
@@ -55,21 +59,32 @@ async function runServer(args: Args, config: ConfigT) {
   const serverInstance = await Metro.runServer({
```
hramos:
```
Error ---------- local-cli/server/runServer.js:59:48
Cannot call `Metro.runServer` with object literal bound to the first parameter because property `watch` is missing in `RunServerOptions` [1] but exists in object literal [2].

                                                     v
  59|   const serverInstance = await Metro.runServer({
  60|     config: {
  61|       ...config,
    :
  77|     watch: !args.nonPersistent,
  78|   });
      --^

References:
  metro/packages/metro/src/index.js:201:4
  201| }: RunServerOptions) => {
          ^^^^^^^^^^^^^^^^ [1]
                                                     v
  59|   const serverInstance = await Metro.runServer({
  60|     config: {
  61|       ...config,
    :
  77|     watch: !args.nonPersistent,
  78|   });
      --^ [2]
```

janicduplessis:
@hramos Oh right, we ignore metro flow types in OSS so I it must be why I missed it. Working on a fix.

janicduplessis:
The `watch` key isn't used by metro (it overrides it to true always), not sure what the `nonPersistent` is for but breaking this is better than the current state of things :)

janicduplessis:
@hramos Should be good to land now

Issue #19859, opened 2018-06-23
Can't run jest tests with 0.56.0


adamivancza:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.4
      CPU: x64 Intel(R) Core(TM) i7-3740QM CPU @ 2.70GHz
      Memory: 27.74 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.9.0 - /usr/local/bin/node
      npm: 6.1.0 - /usr/local/bin/npm
      Watchman: 4.5.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
      Android SDK:
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 24.0.2, 25.0.0, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.0, 27.0.3
        API Levels: 19, 22, 23, 24, 25, 26, 27
    IDEs:
      Android Studio: 2.2 AI-145.3360264
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.3.2 => 16.3.2
      react-native: 0.56.0-rc.2 => 0.56.0-rc.2
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7

## Description
My jest tests are broken - it displays this error for each test file:
Plugin 0 specified in "project/node_modules/babel-preset-react-native/index.js" provided an invalid property of "default" (While processing preset: "project/node_modules/babel-preset-react-native/index.js")

at Plugin.init (node_modules/babel-core/lib/transformation/plugin.js:131:13)
      at Function.normalisePlugin (node_modules/babel-core/lib/transformation/file/options/option-manager.js:152:12)
      at node_modules/babel-core/lib/transformation/file/options/option-manager.js:184:30
          at Array.map (<anonymous>)
      at Function.normalisePlugins (node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
[...]

PR #20178, opened 2018-07-12
[ GENERAL ][ BUGFIX ][ jest/preprocessor.js ] - Invalid use of destructuring

umairda:
Test Plan:
----------
I had the same issues running tests as mentioned in: https://github.com/facebook/react-native/issues/19859

I followed the suggestions but ran into another error:
```
  ● Test suite failed to run

    /Users/umair/projects/my-project/node_modules/react-native/jest/preprocessor.js:49
              {sourceType: 'script', ...nodeOptions, ast: false},
                                     ^^^

    SyntaxError: Unexpected token ...
```

The `preprocessor.js` code in question is:
```
      // node specific transforms only
      return babelTransformSync(
        src,
        Object.assign(
          {filename: file},
          {sourceType: 'script', ...nodeOptions, ast: false},
        ),
      ).code;
```

Specifically line 46: `{sourceType: 'script', ...nodeOptions, ast: false},` is an invalid use of object destructuring and since this object is part of an `Object.assign` statement I simply broke the constituents up into arguments as follows:
```
        Object.assign(
          {filename: file},
          {sourceType: 'script'},
          nodeOptions, 
          {ast: false}
        ),
```

Release Notes:
--------------
[GENERAL][BUGFIX][jest/preprocessor.js] - fixes invalid use of object destructuring in preprocessor.js

### DIFF
--- a/jest/preprocessor.js
+++ b/jest/preprocessor.js
@@ -43,7 +43,9 @@ module.exports = {
         src,
         Object.assign(
           {filename: file},
-          {sourceType: 'script', ...nodeOptions, ast: false},
+          {sourceType: 'script'},
+          nodeOptions,
+          {ast: false}
         ),
       ).code;
     }
--- a/jest/preprocessor.js
+++ b/jest/preprocessor.js
@@ -41,12 +41,12 @@ module.exports = {
       // node specific transforms only
       return babelTransformSync(
         src,
-        Object.assign(
-          {filename: file},
-          {sourceType: 'script'},
-          nodeOptions,
-          {ast: false}
-        ),
+        {
+          filename: file,
+          sourceType: 'script',
+          ...nodeOptions,
+          ast: false
+        },
       ).code;
     }
 


TheSavior:
What version of node are you running? My understanding is that the ... there is reasonable.

umairda:
v8.0.0

It looks like I have to pass the `--harmony` flag when running node to enable object destructuring

But the code in general is strange. If destructuring is being used then why is Object.assign also used? It seems like you could write the second argument simply as:
```
{
   filename:file,
   sourceType:'script',
   ...nodeOptions,
   ast:false
}
```

TheSavior:
It looks like node rest/spread syntax started working without the harmony flag in 8.3.0. This should work if you update to that. 

@hramos, we should make sure our minimum supported version is set to 8.3.0.

————
I agree, this function call is strange and is unnecessarily using Object.assign and spread. If you change this PR to just be the spread version, we would merge this in.

umairda:
You are right. I updated to node v8.3.0 and it works.

I went ahead and updated the pr so that portion of the code uses destructuring only and got rid of the Object.assign. 

I re-ran all my tests and they work fine whereas before they were complaining about the use of the spread operator.

janicduplessis:
Maybe we should update https://github.com/facebook/react-native/blob/master/local-cli/server/checkNodeVersion.js#L17 to 8.3 then.

hramos:
https://github.com/facebook/react-native/pull/20236

TheSavior:
It looks like the commit didn't auto close this PR: https://github.com/facebook/react-native/commit/9d5bd507372c7b63e59a94383c3e3091d96409de

PR #20194, opened 2018-07-13
add flow typing for transform prop

kekami:
Fixes:
----------
Fixes a flow error that occurs when using the transform prop on the View component.

`[flow] Cannot create `View` element because property `transform` is missing in object type [1] but exists in props [2]. (References: [1] [2])`

Test Plan:
----------
Flow no longer yield any errors when using transform on the View component.

Release Notes:
--------------
[ GENERAL ] [ BUGFIX ] [ {./Libraries} ] - Fixed flow error


TheSavior:
Transform props should be set via style, and not as top level props. Flow understands the transform style props as defined in StyleSheetTypes.js

PR #20208, opened 2018-07-14
VirtualizedList function optimization

### LINKED ISSUE #20174: FlatList component inefficient function passing for CellRenderer
- [x] Review the documentation
  - [x] Search for existing issues
  - [ ] Use the latest React Native release - *using version 0.55.2 did not see any changes to the FlatListComponent since that version.*

## Environment
```
Environment:
  OS: Windows 10
  Node: 8.11.1
  Yarn: 0.21.3
  npm: 5.8.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Version  3.0.0.0 AI-171.4408382

  react: 16.3.1 => 16.3.1
  react-native: ^0.55.2 => 0.55.4

```
## Description
I used the npm package why-did-you-update to help optimize my components and prevent re-renders. However there is a re-render that I cannot control. This happens for every item in the list.  My gues

```
CellRenderer.props: Changes are in functions only. Possibly avoidable re-render?
Functions before: {onLayout: ƒ}
Functions after: {onLayout: ƒ}
```

It seems this is caused by an anonymous function at https://github.com/facebook/react-native/blob/d756d94b3a7e2812f17f549c57767ac63734b49c/Libraries/Lists/VirtualizedList.js#L691

I propose that this function be defined on the class to prevent unnecessary re-renders.

jasekiw:
Fixes
----------
Fixes #20174 

Changes Made
----------
* Optimized function passing to the CellRenderer component.
* Removed all components being passed in through the parentProps prop since they were not being used but causing a re-render since component props always get a new reference.
* Add shouldComponentUpdate to CellRenderer and compared the old props to the new props using shallowEquals. I had to pull in shallowEquals and modify it to allow exceptions. I would like some feedback here.

I noticed the onLayout method has this comment `// This is extracted by ScrollViewStickyHeader` . I'm not sure if my modification effected anything in regards to that.

Test Plan:
----------
Tested on a real device with `why-did-you-update` running to view when the CellRendere component re-rendered. Applied updates to the list items to make sure they still get updated.

Added a test that makes sure the initial onLayout is still called properly.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT][VirtualList]  - Prevent unnecessary re-renders of the CellRenderer Component.

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -688,7 +688,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           key={key}
           prevCellKey={prevCellKey}
           onUpdateSeparators={this._onUpdateSeparators}
-          onLayout={e => this._onCellLayout(e, key, ii)}
+          onLayout={this._onCellLayout}
           onUnmount={this._onCellUnmount}
           parentProps={this.props}
           ref={ref => {
@@ -1046,7 +1046,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     }
   };
 
-  _onCellLayout(e, cellKey, index) {
+  _onCellLayout = (e, cellKey, index): void => {
     const layout = e.nativeEvent.layout;
     const next = {
       offset: this._selectOffset(layout),
@@ -1086,7 +1086,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     }
 
     this._computeBlankness();
-  }
+  };
 
   _onCellUnmount = (cellKey: string) => {
     const curr = this._frames[cellKey];
@@ -1604,7 +1604,7 @@ class CellRenderer extends React.Component<
     index: number,
     inversionStyle: ?DangerouslyImpreciseStyleProp,
     item: Item,
-    onLayout: (event: Object) => void, // This is extracted by ScrollViewStickyHeader
+    onLayout: (event: Object, key: string, index: number) => void, // This is extracted by ScrollViewStickyHeader
     onUnmount: (cellKey: string) => void,
     onUpdateSeparators: (cellKeys: Array<?string>, props: Object) => void,
     parentProps: {
@@ -1670,6 +1670,9 @@ class CellRenderer extends React.Component<
     this.props.onUnmount(this.props.cellKey);
   }
 
+  _onLayout = (e): void =>
+    this.props.onLayout(e, this.props.cellKey, this.props.index);
+
   render() {
     const {
       CellRendererComponent,
@@ -1694,7 +1697,7 @@ class CellRenderer extends React.Component<
        * comment and run Flow. */
       getItemLayout && !parentProps.debug && !fillRateHelper.enabled()
         ? undefined
-        : this.props.onLayout;
+        : this._onLayout;
     // NOTE: that when this is a sticky header, `onLayout` will get automatically extracted and
     // called explicitly by `ScrollViewStickyHeader`.
     const itemSeparator = ItemSeparatorComponent && (
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1594,7 +1594,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   }
 }
 
-class CellRenderer extends React.Component<
+class CellRenderer extends React.PureComponent<
   {
     CellRendererComponent?: ?React.ComponentType<any>,
     ItemSeparatorComponent: ?React.ComponentType<*>,
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1736,6 +1736,7 @@ class CellRenderer extends React.Component<
   }
 
   _onLayout = (e): void =>
+    this.props.onLayout &&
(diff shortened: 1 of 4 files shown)


Libraries/Lists/VirtualizedList.js
```diff
@@ -1670,6 +1735,19 @@ class CellRenderer extends React.Component<
     this.props.onUnmount(this.props.cellKey);
   }
 
+  _onLayout = (e): void =>
+    this.props.onLayout(e, this.props.cellKey, this.props.index);
```
ssssssssssss: You should guard onLayout before calling it.
```
this.props.onLayout && this.props.onLayout(...args)
```

Libraries/Lists/VirtualizedList.js
```diff
@@ -1046,7 +1111,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     }
   };
 
-  _onCellLayout(e, cellKey, index) {
+  _onCellLayout = (e, cellKey, index): void => {
     const layout = e.nativeEvent.layout;
```
ssssssssssss:
```
if (isNaN(index)) {
  index = 0
}
```
Maybe you should check index before actually use it. 
```
this._highestMeasuredFrameIndex = Math.max(
         this._highestMeasuredFrameIndex,
         index,	 
       );	
  ```
If index is NaN, then _highestMeasuredFrameIndex will be also NaN, which may leads to crash.

**Comment by ssssssssssss** [comment]
IMO, Ping @sahrens @shergin

File: Libraries/Lists/VirtualizedList.js [hunk 3]
```diff
@@ -675,6 +734,12 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       if (stickyIndicesFromProps.has(ii + stickyOffset)) {
         stickyHeaderIndices.push(cells.length);
       }
+      // remove all components from parentProps so that these components don't cause the CellRenderer to re-render.
+      const parentProps: Props = {...this.props};
+      delete parentProps.ItemSeparatorComponent;
+      delete parentProps.ListEmptyComponent;
+      delete parentProps.ListFooterComponent;
+      delete parentProps.ListHeaderComponent;
```
sahrens: Instead of copying the whole thing and then removing things you don't want to keep, can you explicitly copy just the props that we do want to keep?

Libraries/Lists/VirtualizedList.js
```diff
@@ -234,6 +234,65 @@ type ChildListState = {
...
+    objB === null
+  ) {
+    return false;
+  }
+
+  var keysA = Object.keys(objA);
+  var keysB = Object.keys(objB);
+
+  if (keysA.length !== keysB.length) {
+    return false;
```
sahrens: should comparing key length consider exceptions?

Libraries/Lists/VirtualizedList.js
```diff
@@ -1670,6 +1735,20 @@ class CellRenderer extends React.Component<
     this.props.onUnmount(this.props.cellKey);
   }
 
+  _onLayout = (e): void =>
+    this.props.onLayout &&
+    this.props.onLayout(e, this.props.cellKey, this.props.index);
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (!shallowEqual(this.props, nextProps, ['parentProps'])) {
```
sahrens: Can you use `deepEquals` with a depth limit of 2 instead? That will do a shallow compare of all the props, including `parentProps`. Hopefully that works and you can get rid of your `shallowEqual` fork. If that doesn't work for some reason, could you at least rename to `shallowEqualWithExceptions` and put it in a util folder?

christianbach:
@jasekiw I will be happy to help move this forward if you don’ have the time?

Libraries/Lists/VirtualizedList.js
ssssssssssss: In terms of perf, seems like copy whole thing and set not used property to null will be much better. https://jsperf.com/delete-vs-undefined-vs-null/16

christianbach:
@jasekiw It would be so awesome if this could make it into master. If you feel that you don't have the time to respond to the change request. I would be happy to help.

Libraries/Lists/VirtualizedList.js
christianbach: @sahrens just to be clear, you are referring to deepDiffer `Libraries/Utilities/differ/deepDiffer` ?

Libraries/Lists/VirtualizedList.js
christianbach: Looking at the original code the only parent props used by CellRenderer are `const {renderItem, getItemLayout} = parentProps;` and then later `!parentProps.debug`.
Couldn't we just pass them as regular props and make CellRenderer a PureComponent? The ShallowCompare used by PureComponent would just work and we don't have the need shallowEqual or deepDiffer. Or perhaps I'm missing something?

christianbach:
@jasekiw I asked @sahrens a few questions and depending on his answer we can choose the best approach?

christianbach:
@jasekiw & @sahrens I made a new pull request and simplified the solutions. CellRenderer is now a PureComponent and the props are now flat utilizing PureComponents shouldComponentUpdate

#21163

cpojer:
Closing in favor of the new PR.

PR #20229, opened 2018-07-16
Make AccessibilityInfo.setAccessibilityFocus cross platform

draperunner:
Currently, `AccessibilityInfo.setAccessibilityFocus` is only available on iOS. The same behaviour can be achieved on Android by dispatching the proper accessibility event. I implemented the same function for Android, to make life slightly more convenient for the developer.

Today, developers must write something like this:
```
if (Platform.OS === 'ios') {
     AccessibilityInfo.setAccessibilityFocus(reactTag)
} else {
     UIManager.sendAccessibilityEvent(reactTag, 8)
}
```

With this change, the following is enough for both Android and iOS:
```
AccessibilityInfo.setAccessibilityFocus(reactTag)
```

Test Plan:
----------
I made the same change in the `react-native` package in my project's node_modules, and checked that it worked on an Android device with TalkBack activated. 

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [AccessibilityInfo] - Implement AccessibilityInfo.setAccessibilityFocus for Android

### DIFF
--- a/Libraries/Components/AccessibilityInfo/AccessibilityInfo.android.js
+++ b/Libraries/Components/AccessibilityInfo/AccessibilityInfo.android.js
@@ -12,6 +12,7 @@
 
 const NativeModules = require('NativeModules');
 const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
+const UIManager = require('UIManager');
 
 const RCTAccessibilityInfo = NativeModules.AccessibilityInfo;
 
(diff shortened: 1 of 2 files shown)


Libraries/Components/AccessibilityInfo/AccessibilityInfo.android.js
```diff
@@ -66,6 +67,15 @@ const AccessibilityInfo = {
...
     _subscriptions.delete(handler);
   },
+
+  /**
+   * Set accessibility focus to a react component.
+   *
+   * See http://facebook.github.io/react-native/docs/accessibilityinfo.html#setaccessibilityfocus
+   */
+  setAccessibilityFocus: function(reactTag: number): void {
+    UIManager.sendAccessibilityEvent(reactTag, 8);
```
mdvacca: Instead of set the value 8 here, you should export a new constant in UIManagerModuleConstants.AccessibilityEventTypes:

see: https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIManagerModuleConstants.java#L138

and then use that new constant from here

mdvacca:
Thank you for this PR @draperunner. 
I added a small comment.
Can you also update the documentation in: https://facebook.github.io/react-native/docs/accessibilityinfo#setaccessibilityfocus

Thank you!

PR #20233, opened 2018-07-16
Fix React native portal unmount

### LINKED ISSUE #19084: (undocumented) RN.createPortal works but leaves fiber tree in weird state
I know portals aren't officially support on React Native, but using them seems to work just great and allows me to cheaply render eg a title outside the root view (in my case the titleView of a native UINavigationController):

```js
class MyScreen extends React.Component {

    render() {
        const portal = require('ReactNative').createPortal(<Text>My Title</Text>, this.props.titleTag);
        return <View>{portal}<Text>My screen content</Text></View>;
    }
}
AppRegistry.registerComponent('MyScreen', () => MyScreen);
```

on the native (iOS) side, I add a separate `RCTRootContentView` where I want the portal to render to, and then pass that view's tag to my main render method:

```objc
- (void)loadView {
  NSNumber* titleTag = RCTAllocateRootViewTag();
  self.titleView = [[RCTRootContentView alloc] initWithFrame:CGRectMake(0,0,100,100)
                                                                     bridge:self.bridge
                                                                   reactTag:titleTag
                                                             sizeFlexiblity:RCTRootViewSizeFlexibilityWidth];
 
  self.view = [[RCTRootView alloc] initWithBridge:self.bridge
                                       moduleName:@"MyScreen"
                                initialProperties:@{@"titleTag": titleTag}];
}
```

Initial rendering works as expected (and my react components are portaled to outside the original `RCTRootView`, really cool). It is when `MyScreen` is redrawn and the portal needs to be removed an internal error is triggered in `unmountHostComponents`:

`undefined is not an object (evaluating children.indexOf)`

Trying to debug this proves rather challenging. I suspect the fiber tree is already in an incorrect state by then. What happens is that on https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-dev.js#L10776, `currentParent` is a number instead of an object. 

Any clues on how to debug this issue further? Eg determine at which point the tree ends up in this weird state.

Thanks!

Titozzz:
*I need help to know how to submit this fix correctly dev renderer + prod renderer*

Fixes #19084 

[INTERNAL] [BUGFIX] [ReactNativeRenderer] - Missing line in unmountHostComponents()

### DIFF
--- a/Libraries/Renderer/oss/ReactNativeRenderer-dev.js
+++ b/Libraries/Renderer/oss/ReactNativeRenderer-dev.js
@@ -11911,6 +11911,7 @@ function unmountHostComponents(current) {
       // When we go into a portal, it becomes the parent to remove from.
       // We will reassign it back when we pop the portal on the way up.
       currentParent = node.stateNode.containerInfo;
+      currentParentIsContainer = true;
       // Visit children because portals might contain host components.
       if (node.child !== null) {
         node.child.return = node;


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

satya164:
cc @hramos

gaearon:
I don't know if this fix is right, but this file should not be edited directly. You'll want to send a PR to the `facebook/react` repository (which is where this line is coming from). And it'll need some tests.

gaearon:
Relevant line: https://github.com/facebook/react/blob/43ffae2d1750ee32409e140e41ea562ff46d2516/packages/react-reconciler/src/ReactFiberCommitWork.js#L724

Since this code is generic I expect that the issue also exists for other renderers. It's possible that it doesn't reproduce with React DOM because `removeChild()` exists in both cases (and so it doesn't matter whether it's a container or a normal instance). But it matters for other renderers.

One way to reproduce it might be to add assertions to `createNoop` renderer that enforce `removeChild` can only be called with an instance, and `removeChildFromContainer` can only be called with a container. Then write a portal test against it (it also exposes `createPortal`), like the one in `ReactIncrementalPerf-test`.

gaearon:
Closing as this should be sent against React repo.

Titozzz:
Thank you for the guidelines, I'm going to submit this ASAP. 🎉

gaearon:
The fix landed into React in https://github.com/facebook/react/pull/13220 and will be carried over with the next RN sync.

PR #20257, opened 2018-07-17
Update Readable/WritableNativeArray

### LINKED ISSUE #20256: WriteableNativeArray regression
## Environment

    System:
      OS: macOS High Sierra 10.13.4
      CPU: x64 Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz
      Memory: 1.26 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.1.0 - /usr/local/bin/node
      npm: 5.6.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 11.3, macOS 10.13, tvOS 11.3, watchOS 4.3
      Android SDK:
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 26.0.3, 27.0.3
    IDEs:
      Xcode: 9.3.1/9E501 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.4.1 => 16.4.1 
      react-native: 0.56.0 => 1000.0.0 
    npmGlobalPackages:
      react-native-cli: 2.0.1

## Description

5649aed6d3 batched native array access for ReadableNativeArray into a single native call. However, this change failed to take into account WriteableNativeArray which extends ReadableNativeArray. When items are pushed to WriteableNativeArray, they are pushed to the native array. However, the methods provided by ReadableNativeArray are not aware of the fact that the cached native array is no longer valid. This leads to writes to WriteableNativeArray failing.

## Reproducible Demo

https://github.com/skaplan/thirddebug/
This repo contains an example app with a native module https://github.com/skaplan/thirddebug/blob/master/android/app/src/main/java/com/thirddebug/TestModule.java. The module has a method `testMethod` which is invoked in App.js. The method demonstrates that pushing to the array will not update the reads after the first read has occured.

skaplan:
Fixes #20256 

`mUseNativeAccessor` is a static property of the ReadableNativeArray class. This fix works by introducing an instance variable `iUseNativeAccessor` which determines whether the instance should always make native accesses. The array will make native accesses if either  `mUseNativeAccessor` or `iUseNativeAccessor` is true. WriteableNativeArray then sets `iUseNativeAccessor` true in the constructor. 

The main problem I faced here is how to make the read methods read natively in the WriteableNativeArray subclass. You can't override variables/static variables. One option I explored was to create a method shouldUseNative which is overridable. However, I thought the performance hit would be too large from requiring a method invocation on every read.

Happy to put more work into fixing this. Let me know what the best way to do things is.

Test Plan:
----------
Re-ran https://github.com/skaplan/thirddebug/ which was used to reproduce the issue. Output is correct.

Release Notes:
--------------

[ANDROID] [BUGFIX] [WriteableNativeArray] - Fixed bug that cached reads don't respect new writes.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableNativeArray.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableNativeArray.java
@@ -32,6 +32,7 @@ public class ReadableNativeArray extends NativeArray implements ReadableArray {
   //WriteOnce but not in the constructor fields
   private @Nullable Object[] mLocalArray;
   private @Nullable ReadableType[] mLocalTypeArray;
+  protected boolean iUseNativeAccessor = false;
 
   private static int jniPassCounter = 0;
   private static boolean mUseNativeAccessor = false;
@@ -77,7 +78,7 @@ public class ReadableNativeArray extends NativeArray implements ReadableArray {
 
   @Override
   public int size() {
-    if (mUseNativeAccessor) {
+    if (mUseNativeAccessor || iUseNativeAccessor) {
       jniPassCounter++;
       return sizeNative();
     }
@@ -87,7 +88,7 @@ public class ReadableNativeArray extends NativeArray implements ReadableArray {
 
   @Override
   public boolean isNull(int index) {
-    if (mUseNativeAccessor) {
+    if (mUseNativeAccessor || iUseNativeAccessor) {
       jniPassCounter++;
       return isNullNative(index);
     }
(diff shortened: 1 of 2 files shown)


skaplan:
@dcestari @emilsjolander @cjhopman

cpojer:
Thanks for the PR. I'm sorry we didn't get back to you earlier. We are actually planning on getting rid of this code "soon" so we aren't planning on including this change right now. I hope you'll understand and that the new native module architecture will delight you! See https://github.com/react-native-community/discussions-and-proposals/issues/40

PR #20264, opened 2018-07-18
add Image border*Radius tests

zholobov:
Test Plan:
----------
Run RNTester and check visuals for Image borders are what is expected.

Release Notes:
--------------
[GENERAL] [MINOR] [RNTester] - Added tests for Image's border*Radius props.

### DIFF
--- a/RNTester/js/ImageExample.js
+++ b/RNTester/js/ImageExample.js
@@ -443,6 +443,30 @@ exports.examples = [
             style={[styles.base, styles.leftMargin, {borderRadius: 19}]}
             source={fullImage}
           />
+          <Image
+            style={[styles.base, styles.leftMargin, {borderTopLeftRadius: 20}]}
+            source={fullImage}
+          />
+          <Image
+            style={[styles.base, styles.leftMargin, {
+              borderWidth: 10,
+              borderTopLeftRadius: 10,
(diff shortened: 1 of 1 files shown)


zholobov:
Not sure why CI is failing on iOS

TheSavior:
Awesome! Thanks for the additional tests. Since these are visual tests, can you add screenshot tests for these? I believe there are instructions in the docs somewhere for how to add these tests. Maybe in the contributing docs.

zholobov:
@TheSavior I see snapshot tests described in [RN docs here](https://facebook.github.io/react-native/docs/testing.html#screenshot-snapshot-tests). But it's only for iOS. For Android, even though I see [screenshot-tests-for-android](https://facebook.github.io/screenshot-tests-for-android/) in the wild, I do not see React Native using them (e.g. there is no mention of the screenshot-tests-for-android plugin in  a build.gradle).

Unfortunately, I do not have access to an iOS dev environment (I'm on Windows). So I cannot add screenshot tests to existing iOS ones. And since Android does not have snapshot testing, this would require adding this feature to the Android test suite. Do you require me to do this?

zholobov:
@TheSavior gentle ping.

TheSavior:
Thanks for the ping!

Issue #20262, opened 2018-07-18
Bridge promises don't handle multiple resolve or reject calls properly


esprehn:
- [ x] Review the documentation: https://facebook.github.io/react-native
  - [ x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [ x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment

Environment:
  OS: macOS High Sierra 10.13.5
  Node: 8.9.4
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.7.0
  Xcode: Xcode 9.4.1 Build version 9F2000
  Android Studio: 3.1 AI-173.4720617

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.54.4 => 0.54.4

## Description
PromiseImpl and RCTModuleMethod don't handle promise resolution properly. Resolving a promise a second time is supposed to have no effect per ES semantics, similarly rejecting it a second time, or if it's rejected already, should have no effect.

PromiseImpl.resolve a second time will throw an error:

```
Exception java.lang.RuntimeException: Illegal callback invocation from native module. This callback type only permits a single invocation from native code.
com.facebook.react.bridge.CallbackImpl.invoke (CallbackImpl.java:30)
com.facebook.react.bridge.PromiseImpl.resolve (PromiseImpl.java:32)
```

For iOS this doesn't crash in ObjC, instead it sends another message over the bridge, but the callbacks are gone, so it incorrectly warns in a dev build (`Callback with id ${cbID}: ${module}.${method}() not found`) and is correctly a no-op in a release build (though it is inefficient to send the bridge response).

## Reproducible Demo

Create a `@RectMethod` and do resolve() in the Java twice, or do do resolve twice in an ObjC iOS method.

esprehn:
This happens in the most recent version too. The code is the same.

hramos:
@esprehn can you update your post accordingly? 0.54.4 is not the latest as of this writing, and we'd like to know the exact version you tested in.

esprehn:
This does actually crash on iOS in a debug build, I misread the code:
https://github.com/facebook/react-native/blob/30b9d81087cb86f5fb272d00bfee63a0632009f5/React/Base/RCTModuleMethod.mm#L119

esprehn:
This bug still exists. @hramos The bot probably shouldn't close issues with pending PRs?

hramos:
[...]

PR #20303, opened 2018-07-19
Don't crash on second resolve or reject in PromiseImpl.

esprehn:
Promise semantics per JS should allow calling resolve or reject repeatedly without crashing. Only the first one should do anything, but others should be allowed. This change fixes the Java bridge for Android. A separate change is needed for iOS.

Issue #20262 

Test Plan:
----------
Where are the bridge tests?

Make a @ReactMethod that takes a Promise. Call promise.resolve() twice.

Release Notes:

--------------
[ANDROID] [BUGFIX] [Bridge] - Ignore resolve or reject calls on bridge Promises in Java.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/PromiseImpl.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/PromiseImpl.java
@@ -28,6 +28,8 @@ public class PromiseImpl implements Promise {
   public void resolve(Object value) {
     if (mResolve != null) {
       mResolve.invoke(value);
+      mResolve = null;
+      mReject = null;
     }
   }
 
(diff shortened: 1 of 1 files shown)


yungsters:
Looks good to me. Are you up for making the same change on iOS?

fkgozali:
Rather than just silently ignoring the 2nd calls onwards, what about throwing a redbox (at least on debug build)?

yungsters:
> Rather than just silently ignoring the 2nd calls onwards, what about throwing a redbox (at least on debug build)?

As proposed by @esprehn, I think we should preserve the semantics of Promise if we're going to expose an API that looks like Promise's resolve and reject functions.

Moving forward with Fabric, we can re-evaluate whether we think these are reasonable semantics (and whether we should not emulate Promise's API).

esprehn:
> Looks good to me. Are you up for making the same change on iOS?

Yeah I'll look into how to do this for iOS as well. I think it just means adding a shared __block boolean across the resolve and reject blocks in [RCTModuleMethod processMethodSignature].

cpojer:
@esprehn are you still able to take a look at the iOS version of this issue and could you send a PR for it? :)

PR #20314, opened 2018-07-20
fix openURL frozen app in iPhone7 iOS10.3.1 

yurnery:
### Test Plan:
as this issue is a native issue, I wrote a demo to test it , 

- `git clone https://github.com/yurnery/OpenURLDemo-.git`
- use iPhone7 iOS10.3.1 run the demo(maybe other devices and version will also work, but I don't have old version device to check out )
- the demo have two button, one call `openURL`, the other call `open(url:option:completion)`, click then you will see the difference between them, the` openURL` will frozen the app 4~5s, `open(url:option:completion)` will not, in this PR I judge the version to choose different method(this OpenURLDemo not judge ), this is the way i fix this issue correctly.

 after merge this PR, an easy way to test this out is to generate a new empty project with `react-native --init` and rewrite the render like this, then  run app and touch screen, the call alert will present quickly, **must be  physical Device, simulator will not work**,
```
render() {
    return (
      <TouchableOpacity
        style={{ flex: 1 }}
        onPress={() => {
          Linking.openURL("tel://10086");
        }}
      />
    );
  }
```

### Release Notes:
[IOS] [BUGFIX] [LINKING] - fix iOS ` Linking.openURL()` frozen the app in  10.3.1 system version

Desc: 
1. the` openURL` method is deprecated in iOS10.
2. in my app, there is a case to use Linking, like `Linking.openURL('tel://911')`, it works well in iOS11 and later, but in iPhone 7 iOS10.3.1, the app is frozen about 4~5s, i think it's too long
3. so i judge the system version to call different Native API, it works well
4. the `openURL` method only frozen "tel:" in iPhone7 iOS10.3.1, other URL type  works well

### DIFF
--- a/Libraries/LinkingIOS/RCTLinkingManager.m
+++ b/Libraries/LinkingIOS/RCTLinkingManager.m
@@ -95,7 +95,8 @@ RCT_EXPORT_METHOD(openURL:(NSURL *)URL
     }
   };
   
-  if (@available(iOS 10.0, *)) {
+  BOOL canRespond = [[UIApplication sharedApplication] respondsToSelector:@selector(openURL:options:completionHandler:)];
+  if (canRespond) {
     [RCTSharedApplication() openURL: URL options:@{} completionHandler: ^(BOOL opened) {
       callBack(opened);
     }];


yurnery:
@hramos hi, i had add the testPlan and the release note

hramos:
Thanks. It looks like your test plan describes how to reproduce the issue, but what we're looking for is a list of steps performed to test and ensure your change fixes the actual underlying issue.

yurnery:
@facebook-github-bot no, I don't have any updates for this pull request right now

lichen19853:
@hramos can we please merge this? The code change and test plan all look good. I also tested in my app, it fixed major issues. And thanks @yurnery for the PR.

yurnery:
I had delete my repository with mistake

PR #20320, opened 2018-07-20
Replace our local types by flow-typed

charpeni:
These comments were confusing since we don't use flow-typed.

cc @bvaughn.

I have replaced our local types by flow-typed, removed metro custom stub and deleted `flow-github` directory.

As result, createReactClass, PropTypes and Metro types won't be distributed to RN users. If you need it, you should relay on flow-typed.

Test Plan:
----------

```
yarn flow

No errors!
```

Release Notes:
--------------

[INTERNAL] [BREAKING] [Flow] - Replace our local types by flow-typed and don't distribute them


bvaughn:
> since we don't use flow-typed

These files [get published to NPM](https://unpkg.com/react-native@0.56.0/flow/). I assume they were for external RN users? I actually have no idea.

Maybe we can just remove the file entirely now since the linked PR has been merged?

charpeni:
> I assume they were for external RN users?

I would say it is for this codebase, but they're exposed to flow:
https://github.com/facebook/react-native/blob/ebf5aeab280f2ebc439ec39d25c48fdf1980cf73/local-cli/templates/HelloWorld/_flowconfig#L24-L27

I think that external RN users should rely on flow-typed and not directly on our `flow` directory?

It looks like [`create-react-class.js`](https://github.com/facebook/react-native/blob/ebf5aeab280f2ebc439ec39d25c48fdf1980cf73/flow/create-react-class.js) can be safely removed from our codebase without generating any flow errors on our side or from RN users. But `createReactClass` will be `any` instead of `React$CreateClass`. Did I miss something?

[`prop-types.js`](https://github.com/facebook/react-native/commit/00330987c8e5e609fa76f30d0a90bd5895a8aeb9#diff-ff51053ae0dcfde739883ae6c234f1ea) was introduced by you when you added the `prop-types` package (00330987c8e5e609fa76f30d0a90bd5895a8aeb9) because `prop-types` is not typed and if we remove it, we'll have some `untyped-import` errors with the `prop-types` package. The official way would be to use flow-typed, but I guess it will generate many other errors? Also, from a RN project, it will be a "breaking change" if we remove `prop-types.js` from the flow directory as consumers will lose typing of PropTypes if they don't use flow-typed.

I'm not sure why we don't use flow-typed, I'll look at this.

charpeni:
We could possibly remove `flow/prop-types.js` from this repo.

@TheSavior Can we safely remove `@flow strict` from these files like I've done in my previous commit?

TheSavior:
I’m not sure that we publish flow defs for React Native to flow typed. I haven’t heard about that before.

What causes you need to remove flow strict?

charpeni:
I don't know about RN on flow typed, but createReactClass and prop-types are on flow-typed (https://github.com/flow-typed/flow-typed/pull/773).

The initial goal of this PR was about these two TODOs since we do not use flow typed and https://github.com/flow-typed/flow-typed/pull/773 is merged.
https://github.com/facebook/react-native/blob/ebf5aeab280f2ebc439ec39d25c48fdf1980cf73/flow/create-react-class.js#L12-L12
https://github.com/facebook/react-native/blob/ebf5aeab280f2ebc439ec39d25c48fdf1980cf73/flow/prop-types.js#L12

Brian made the point that we probably don't need these files anymore, so I check how RN would react (no pun intended) without these type definitions. It flags two untyped-import errors because the lib `prop-types` is not typed (needs flow-typed) and stylesheet proptypes edited in 5b8e5de were on `@flow strict`.

In brief, we can safely remove `flow/create-react-class.js` and `flow/prop-types.js` if we remove `@flow strict` from some stylesheet proptypes. Some users who depend on our published types of prop-types will lose it, but they can use the prop-types definition on `flow-typed`.

TheSavior:
We definitely need to make sure that createReactClass and prop-types are not typed as `any`. This might mean figuring out how to include them from flow typed.

react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  :exclamation: Big PR - <i>This PR is extremely unlikely to get reviewed because it touches 1026 lines.</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

charpeni:
I have replaced our local types by flow-typed, I have also deleted `flow-github` which was containing a stub for metro because flow-typed can stub it.

I have also removed all references to `flow-github` (.flowconfig, _flowconfig, and package.json).

`flow-typed` directory won't be published on npm, so we'll only have these types for RN repo.

createReacteClass, PropTypes, and Metro won't be distributed to RN users. IMO, if they need it, they should relay on flow-typed directly.

Sounds good?

flow-typed/npm/prop-types_v15.x.x.js
```diff
@@ -1,15 +1,15 @@
 /**
- * Copyright (c) 2015-present, Facebook, Inc.
+ * Copyright (c) 2018-present, Facebook, Inc.
```
gengjiawen: Why change this line ?

hramos: This line will be updated to have no reference to the year in an upcoming commit. I'd leave it untouched, to avoid merge conflicts...

hramos:
What should we do with this PR?

charpeni:
Well, I've redone the integration of flow-typed on top of the recent changes with Flow.

I've also checked that it has:
- No news Flow errors in this repo;
- No Flow errors on a freshly init react native project;

cpojer:
Do you mind rebasing this?

charpeni:
Thanks, @rickhanlonii!

PR #20386, opened 2018-07-25
execute calls to AsyncStorageModule serially

### LINKED ISSUE #18372: AsyncStorage.getItem() doesn't seem to work
When I try to set a value via `AsyncStorage.getItem()`, I cannot request it back.

### Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 9.8.0
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.9.0
  Android Studio: Not Found

  react: ^16.3.0-alpha.1 => 16.3.0-alpha.1

### Expected Behavior

    await AsyncStorage.setItem('foo', 'bar');
    await AsyncStorage.getItem('foo').then(console.log); // 'bar'
    await AsyncStorage.getAllKeys().then(console.log); // ['foo']

### Actual Behavior

    await AsyncStorage.setItem('foo', 'bar');
    await AsyncStorage.getItem('foo').then(console.log); // null
    await AsyncStorage.getAllKeys().then(console.log); // []

### Steps to Reproduce

    import { AsyncStorage } from 'react-native';

    it('should be able to request a value after it was set', async () => {
      await AsyncStorage.setItem('foo', 'bar');
      const output = await AsyncStorage.getItem('foo');

      expect(output).toBe('bar');
    });

    it('should be able to see a new key after a value was set', async () => {
      await AsyncStorage.setItem('foo', 'bar');
      const output = await AsyncStorage.getAllKeys();

      expect(output).toContain('foo');
    });

dulmandakh:
This PR creates single threaded executor in AsyncStorageModule and use it to execute calls serially, so that below code works as expected

``` js
await AsyncStorage.setItem('foo', 'bar');
await AsyncStorage.getItem('foo').then(console.log); // 'bar'
await AsyncStorage.getAllKeys().then(console.log); // ['foo']
```

Previously, it was using AsyncTask that executes tasks in its own thread pool, which may execute calls in parallel.

Since it uses its own executor, it don't need to wrap the executions in the AsyncTask.

Fixes #18372 #14101

Test Plan:
----------
Before (it's inconsistent)
```js
await AsyncStorage.setItem('foo', 'bar');
await AsyncStorage.getItem('foo').then(console.log); // null
await AsyncStorage.getAllKeys().then(console.log); // []
```

After
```js
await AsyncStorage.setItem('foo', 'bar');
await AsyncStorage.getItem('foo').then(console.log); // 'bar'
await AsyncStorage.getAllKeys().then(console.log); // ['foo']
```

Release Notes:
--------------

[ANDROID] [BUGFIX] [AsyncStorage] - Execute calls serially


dulmandakh:
@hramos @gengjiawen I have no experience of writing android tests, could you please help update tests for AsyncStorageModule?

ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
gengjiawen: why touch this line.

gengjiawen:
I have little test experience on android test. Do you test your code in some way , like tested it on real device.

dulmandakh:
I was thinking that example for AsyncStorage in RNTester app would be
useful, but I develop in TypeScript and have no experience of Flow. The
example must be trivial, enter a key and value to store and enter key to
retrieve its value.

gengjiawen:
I find flow difficulty to write too. I want it in plain js. And also current project has to many #FlowFixme.

ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java
```diff
@@ -83,9 +85,9 @@ public void multiGet(final ReadableArray keys, final Callback callback) {
       return;
     }
 
-    new GuardedAsyncTask<Void, Void>(getReactApplicationContext()) {
+    execute(new Runnable() {
```
mdvacca: What happens if an exception is thrown when executing the Runnable?

GuardedAsyncTask provided error management, see: https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/GuardedAsyncTask.java#L34

ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java
```diff
@@ -42,6 +43,7 @@
 
   private ReactDatabaseSupplier mReactDatabaseSupplier;
   private boolean mShuttingDown = false;
+  private Executor executor = Executors.newSingleThreadExecutor();
```
mdvacca: I'm not sure if adding a new executor is the best approach here, but If you are adding this you should override the  onCatalystInstanceDestroy method and then shutdown the executor.
See:
https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/ReactAndroid/src/main/java/com/facebook/react/bridge/NativeModule.java#L51

dulmandakh:
Thank you @mdvecca for review and suggestions.

Could you please suggest other options to make serially execute calls than
executor.

I intentionally wrapped in private execute method so we can easily change
execution method in the future. I'll change it to handle runtime exception,
and shutdown executor.

dulmandakh:
@mdvacca made changes as you suggested 😄

ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java
dulmandakh: Android Studio was showing warning, so fixed it 😄

dulmandakh:
ExecutorService in AsyncStorageModule is customizable, and test is passing. It'll fix many issues on Android, including me.

@mdvacca please review and merge

gengjiawen:
Maybe add RNTester showcase too. Can we just add plain js in RNTester ? @hramos @janicduplessis

ReactAndroid/src/main/java/com/facebook/react/modules/storage/AsyncStorageModule.java
janicduplessis: Could we use `executeOnExecutor` (https://developer.android.com/reference/android/os/AsyncTask.html#executeOnExecutor(java.util.concurrent.Executor,%20Params...)) instead of `execute` and keep using GuardedAsyncTask instead?

dulmandakh: @janicduplessis got it and ready to push the change, but waiting for CI to become green again.

hramos:
CI looks good, still waiting on review from @mdvacca or other. I'm not the best person to review this.

dulmandakh:
@mdvacca could you please review ASAP. I would like it to land in 0.57 if possible, because I'm planning to use it for my projects. Thank you

ahilles107:
Would be great to have it in 0.57!

cpojer:
Hey @dulmandakh. We moved AsyncStorage to its own repo, it would be great if you could re-send your PR to that one: https://github.com/react-native-community/react-native-async-storage

PR #20393, opened 2018-07-25
Whitelist react-native-dom in haste/cli config defaults

vincentriemer:
This adds `react-native-dom` to `hasteImpl.js` because it's not currently possible to configure it from an out of tree platform. Also adds the relevant `providesModuleNodeModules` and `platforms` values to the default RN CLI config. This should hopefully be able to be removed once better support for out of tree platforms is implemented.

Test Plan:
----------

Confirmed that the updated config correctly builds an RNDom project by pulling my fork as a dependency (though it required some other changes for some unrelated `local-cli` issues).

Release Notes:
--------------

[CLI] [ENHANCEMENT] [jest/hasteImpl.js] - Whitelist react-native-dom in haste/cli config defaults

### DIFF
--- a/jest/hasteImpl.js
+++ b/jest/hasteImpl.js
@@ -15,6 +15,7 @@ const path = require('path');
 const ROOTS = [
   path.resolve(__dirname, '..') + path.sep,
   path.resolve(__dirname, '../../react-native-windows') + path.sep,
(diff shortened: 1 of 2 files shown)


kelset:
Context: it's the same approach that was used for `react-native-windows` here: https://github.com/facebook/react-native/commit/54942746d4037e1153e14fcfc95e4edc772d296a

PR #20416, opened 2018-07-27
Adding support to cancel a headless task

psivaram:
On UWP, when the app is executing a Background task, the OS can request it to cancel the task (for various reasons). If the app does not cancel its background task quickly, the OS might terminate the app.
This change adds support to cancel the headless task which can be used by native code to forward the cancellation request from the OS to the JS code.

Test Plan:
----------
There don't seem to be any tests for startHeadlessTask, so tested both startHeadlessTask and cancelHeadlessTask manually.
Our app based on private forks of react-native and react-native-windows already has been using these two functions for a few months now. So pushing the private changes to the public repo.

Release Notes:
--------------
[GENERAL][ENHANCEMENT][AppRegistry] - Added support to cancel a headless-task


psivaram:
There is a companion PR for updating the documentation at https://github.com/facebook/react-native-website/pull/473

react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Libraries/ReactNative/AppRegistry.js
```diff
@@ -237,6 +251,19 @@ const AppRegistry = {
...
+   * Only called from native code. Cancels a headless task.
+   *
+   * See http://facebook.github.io/react-native/docs/appregistry.html#cancelheadlesstask
+   */
+  cancelHeadlessTask(taskId: number, taskKey: string): void {
+    const taskCancelProvider = taskCancelProviders.get(taskKey);
+    if (!taskCancelProvider) {
+      throw new Error(`No task canceller registered for key '${taskKey}'`);
+    }
+    taskCancelProvider()();
```
rigdern: >taskCancelProvider()(); [](start = 4, length = 23)

Why did you decide to have two layers of indirection instead of just one. In other words, with the current design people have to do this:

```
function taskCancelProvider() {
  return () => {
    // cancel the task
  };
}
registerCancellableHeadlessTask(someKey, taskProvider, taskCancelProvider);
```

Why not instead design the API like this?:

```
function taskCancelProvider() {
  // cancel the task
}
registerCancellableHeadlessTask(someKey, taskProvider, taskCancelProvider);
```

psivaram: I followed the pattern used for TaskProvider.
The motivation for the original choice seems to be to allow delay loading of the corresponding JS code by having it in a separate JS file and set taskProvider as () => require('SomeTaskName')

See https://facebook.github.io/react-native/docs/headless-js-android.html

PR #20474, opened 2018-07-31
Force version of build tools and compile SDK from root project on all android modules

futuun:
It resolves all those:
```
Error:The SDK Build Tools revision (??.?.?) is too low for project ‘:(…) Minimum required is 25.0.0
```
by forcing the same version of build tools and compile SDK on all android modules.

Almost all native android modules will not compile on rn 0.56 because buildToolsVersion and
compileSdkVersion were updated and what most packages have is too low.

Test Plan:
----------
You can add any native module that haven't been updated after rn 0.56 ie. `react-native-config@0.11.5` and try to run on android. It'll error with `The SDK Build Tools revision (...)` message. Adding changes proposed in this PR to your `build.gradle` will fix that.

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [build.gradle] - force version of build tools and compile SDK from root project on android modules

### DIFF
--- a/local-cli/templates/HelloWorld/android/build.gradle
+++ b/local-cli/templates/HelloWorld/android/build.gradle
@@ -38,3 +38,14 @@ ext {
     targetSdkVersion = 26
     supportLibVersion = "26.1.0"
 }
+
+subprojects { subproject ->
+    afterEvaluate {
+        if ((subproject.plugins.hasPlugin('android') || subproject.plugins.hasPlugin('android-library'))) {
+            android {
+                compileSdkVersion rootProject.ext.compileSdkVersion
+                buildToolsVersion rootProject.ext.buildToolsVersion
+            }
+        }
+    }
+}


kelset:
I don't think this should land because it's responsibility of the people creating all those "native modules" to ensure that they work with more recents versions of the SDK/Build Tools.

And if you want to "take the risk" of having your project use the newer version, even if that third-party-lib doesn't explicitly support it, you should do it in your own implementation.

futuun:
That’s fair but all those “native modules” will work with more recents versions. `compileSdkVersion` doesn't change runtime behaviour. Android studio even offers you autofix (not very useful given that it changes content of node_modules).
- `compileSdkVersion` is the version of the API the app is compiled against. As long as you are not changing `targetSdkVersion` you cannot get compatibility issues
-  `buildToolsVersion` is the version of compiler you want to use. It’s tied to your `compileSdkVersion` (if your compileSdkVersion is `23` then you use `23.x.x` if you are using `26` then you have `26.x.x` etc). No issues with backward compatibility either.

kelset:
Uhm, understood.

Yeah if that's the case (of no backward compatibility) I think I can relax my concern.

@dulmandakh do you think this should be enforced?

dulmandakh:
@futuun @kelset I don't like the idea of forcing Android tools from React Native. Personally, I use latest versions of SDK in my projects, rather than waiting RN to bump Android SDK versions. To make development of native modules easy, RN landed using project-wide settings with 0.56 https://github.com/facebook/react-native/commit/0a3055d98a36e49746144e883edc7e20afec4fcb. I would like propose to close the PR.

PR #20539, opened 2018-08-05
Update docs on splitting APKs

alvinthen:
Gradle v3.1 complains about conflicting configuration as below

<pre>
Conflicting configuration : 'armeabi-v7a,x86' in ndk abiFilters cannot be present when splits abi filters are set : x86,armeabi-v7a
</pre>

Release Notes:
--------------
[ DOCS][MINOR][signed-apk-android.md] - Improve docs on splitting APKs for newer version of gradle


hramos:
The docs have been moved to https://github.com/facebook/react-native-website

PR #20541, opened 2018-08-06
Remove newLayout check when begin layout calculation && fix comment of RCTShadowView

zhongwuzw:
1.  Please see [layout calculate](https://github.com/facebook/react-native/blob/73d57461229ebb398af769883bf994801943ceca/React/Views/RCTShadowView.m#L275), it would forcedly set `hasNewLayout` of `yogaNode` to `true`. So we only need to reset it, don't need to check again.
2. We have removed `collectUpdatedProperties:parentProperties:` from this [commit](https://github.com/facebook/react-native/commit/0f9fc4b2953d52fa1754e786dc5c74bfecbeaaca) and give these states all to `Yoga`.

Test Plan:
----------
Please see [layout calculate](https://github.com/facebook/react-native/blob/73d57461229ebb398af769883bf994801943ceca/React/Views/RCTShadowView.m#L275), it would forcedly set `hasNewLayout` of `yogaNode` to `true`. So we only need to reset it, don't need to check again.

Release Notes:
--------------

[IOS][ENHANCEMENT][RCTShadowView] - Remove unnecessary newLayout check

### DIFF
--- a/React/Views/RCTShadowView.m
+++ b/React/Views/RCTShadowView.m
@@ -281,10 +281,6 @@ static void RCTProcessMetaPropsBorder(const YGValue metaProps[META_PROP_COUNT], 
 
   RCTAssert(!YGNodeIsDirty(yogaNode), @"Attempt to get layout metrics from dirtied Yoga node.");
 
-  if (!YGNodeGetHasNewLayout(yogaNode)) {
-    return;
-  }
-
   YGNodeSetHasNewLayout(yogaNode, false);
 
   RCTLayoutMetrics layoutMetrics = RCTLayoutMetricsFromYogaNode(yogaNode);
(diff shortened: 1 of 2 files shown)


React/Views/RCTShadowView.m
```diff
@@ -281,10 +281,6 @@ - (void)layoutWithMinimumSize:(CGSize)minimumSize
 
   RCTAssert(!YGNodeIsDirty(yogaNode), @"Attempt to get layout metrics from dirtied Yoga node.");
 
-  if (!YGNodeGetHasNewLayout(yogaNode)) {
```
shergin: I am a bit concerned about this change because it relies on current implementation details of Yoga. Yoga is an independent library and its implementation can change any time. And, if we remove change, why should not we remove the next line that resets the flag?

zhongwuzw: @shergin Thanks for review. It's true that we should not rely details on third library. The reason I don't remove the line of resets the flag is maybe other code may query `YGNodeGetHasNewLayout` of this `node`, but actually, we don't have this check currently, IMO, maybe keep this reset is ok, we can keep the same way, when we reset it, it means we already layout the views.
So what's your opinion? 🤔 We can remove `hasNewLayout` check? Or just to abandon this `PR`?

cpojer:
I talked to @davidaurelio who works on Yoga. He doesn't think this is the right change and that it won't have much impact so I'm gonna close this PR.

zhongwuzw:
@cpojer Thanks for review, I revisited this again and I also think it's unnecessary. 😂 I'll make a new `PR` to fix the shadow view process comment.

PR #20584, opened 2018-08-08
[Linux/OSX] Add missing "--terminal" argument to run-android

### LINKED ISSUE #18122: JS Server Never Terminates
### Is this a bug report?
Yes

Yes

### Environment

Environment:
  OS: Linux 4.13
  Node: 8.9.4
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A

  react: 16.2.0 => 16.2.0
  react-native: ^0.53.3 => 0.53.3

Target Platform: android:minSdkVersion="16"
                           android:targetSdkVersion="23"

### Steps to Reproduce
(On a linux machine)
1. run command "sudo lsof -i :8081" and kill any processes on that port with the command "kill -9 <PID>"
2. react-native init test
3. cd test
4. react-native run-android
5. close/stop terminal (either by closing out of the terminal or pressing ctrl+c; doesn't matter)
6. run command "sudo lsof -i :8081".

### Expected Behavior
There are no processes active on port 8081 and you can run "react-native run-android" any time in the future and it works.

### Actual Behavior
The JS server is still running on port 8081. I do not understand why, but this creates a problem when running the app again. After starting the intent (usually the last thing in the terminal after running react-native run-andriod), it just stops automatically as seen in my terminal below:
react-native-run-android-abrupt-termination

Versus how I usually stop it as seen in my terminal below:
react-native-correct-termination

### Reproducible Demo
https://github.com/BWillis98/test
I apologize that I know nothing about "snack", but I've uploaded the test repository I created when going through "Steps to reproduce". I don't think it's really necessary since all I did was run the command "react-native init test" and then push it to a github repository.

### Workaround To Problem
Run the command "sudo lsof -i :8081", make note of the PID of the process where the column under "COMMAND" is titled "node", and then run the command "kill -9 [PID]" as demonstrated here:

workaround

However, I don't want to do this every time I want to load the app onto my phone and it took a lot of time to narrow down the issue to this and I'd like to save future developers the time of fixing this issue.

jamsch:
Commit 79e498b7fbf19ab50bb73697602910fb827ff928 claims that "--terminal" has been added as an argument to the "run-android" script which it hasn't. Instead it adds a new environment variable "process.env.REACT_TERMINAL" which seems to be undocumented and nowhere else in the repository. 

This commit now allows Linux and OSX users to specify their custom terminal :

    react-native run-android --terminal=x-terminal-emulator

Additionally the metro bundler child process "detached" option has been set to false if no terminal has been specified on Linux which fixes an issue where the packager would be running in the background and would have to manually be closed by finding the process. 

Fixes #18122

Test Plan:
----------
None

Release Notes:
--------------
[CLI] [BUGFIX] [local-cli/runAndroid/runAndroid.js] - Add missing "--terminal" argument

### DIFF
--- a/local-cli/runAndroid/runAndroid.js
+++ b/local-cli/runAndroid/runAndroid.js
@@ -348,7 +348,7 @@ function runOnAllDevices(
   }
 }
 
-function startServerInNewWindow(port, terminalArg) {
+function startServerInNewWindow(port, terminal = process.env.REACT_TERMINAL) {
   // set up OS-specific filenames and commands
   const isWindows = /^win/.test(process.platform);
   const scriptFile = isWindows
@@ -363,7 +363,6 @@ function startServerInNewWindow(port, terminalArg) {
   const scriptsDir = path.resolve(__dirname, '..', '..', 'scripts');
   const launchPackagerScript = path.resolve(scriptsDir, scriptFile);
   const procConfig = {cwd: scriptsDir};
-  const terminal = terminalArg || process.env.REACT_TERMINAL;
 
   // set up the .packager.(env|bat) file to ensure the packager starts on the right port
   const packagerEnvFile = path.join(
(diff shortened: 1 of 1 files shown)


hramos:
Re-importing. This failed earlier, because the PR was updated after it was approved but before it could be imported.

PR #20600, opened 2018-08-09
Remove TextLayoutBuilder from ReactAndroid dependency

dulmandakh:
This PR removes TextLayoutBuilder, which is not used anywhere.

Test Plan:
----------
Everything builds and runs as usual


dulmandakh:
@hramos please view and merge

hramos:
Importing without approval, as we need to check if this is OK to remove internally.

PR #20603, opened 2018-08-09
[android] fix bug overflow:hidden does not work

tsangint:
### Test Code
```
<View
    style={{width:60,height:40,overflow:'hidden'}}
>
    <View style={{width:40,height:80,backgroundColor:"#F04F43"}}></View>
</View>
```
### Before
When execute above code, overflow:hidden is not render correctly.(iOS is ok)  

### After
overflow:hidden render rightly,it should be a 40*40 red foursquare.  

All above situation can render overflow rightly.

## Release Notes

[ Android ] [BUGFIX] [ Yoga ]

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java
@@ -12,6 +12,7 @@ import com.facebook.infer.annotation.Assertions;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMapKeySetIterator;
 import javax.annotation.Nullable;
+import com.facebook.yoga.YogaOverflow;
 
 /**
  * Class responsible for optimizing the native view hierarchy while still respecting the final UI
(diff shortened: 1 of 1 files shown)


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

yungsters:
Thanks for submitting this pull request.

But this is unexpected. Would you be open to figuring out why the logic in [`ViewProps.isLayoutOnly`](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewProps.java#L262) is insufficient? It would be great to not have this logic scattered between the two methods.

tsangint:
It's my fault. my code is base on 0.54.0 and this version had not deal with OVERFLOW in `isLayoutOnly(ReadableMap map, String prop)`.
I checkout to master and the latest code in master will judge overflow props(`map.isNull(OVERFLOW) || VISIBLE.equals(map.getString(OVERFLOW))`).This problem has gone,`overflow:hidden` can be rendered rightly.
The logic in ` ViewProps.isLayoutOnly` is more  reasonable .I will close the pr.

PR #20605, opened 2018-08-09
Enable metro hmr server in local-cli

janicduplessis:
The `hmrEnabled` flag was not passed to the metro runServer config which caused HMR to be broken.

Test Plan:
----------
Test that hmr works by applying the fix locally

Release Notes:
--------------
[CLI] [MINOR] [runServer.js] - Enable metro hmr server in local-cli


janicduplessis:
cc @CompuIves

PR #20664, opened 2018-08-14
No force layout of superview and contentSize recalculation if intrinsicContentSize not changed

zhongwuzw:
Test Plan:
----------
No need test.

Release Notes:
--------------
[IOS][ENHANCEMENT][RCTRootView] - No force layout of superview and contentSize recalculation if intrinsicContentSize not changed


cpojer:
I verified the original change that added this and I think it makes sense to move it up.

PR #20719, opened 2018-08-17
Expose scrollEnabled as iOS prop for TextInput

hramos:
Summary: Expose scrollEnabled as iOS prop for TextInput

Differential Revision: D9383477


hramos:
This was an export of an upcoming internal diff.

PR #20731, opened 2018-08-19
refine android config

gengjiawen:
## Motivation
By this change we can unify version early, like we can make kotlin version share in buildscript and app dependency the same time ([butterknife](https://github.com/JakeWharton/butterknife/blob/master/build.gradle) use the same).
## Test Plan
pass all current ci and one of my project https://github.com/gengjiawen/ReactNative64bitDemo.
## Related PRs
none
## Release Notes
 [GENERAL] [ANDROID] [TEMPLATE] - refine android config

### DIFF
--- a/local-cli/templates/HelloWorld/android/build.gradle
+++ b/local-cli/templates/HelloWorld/android/build.gradle
@@ -1,6 +1,13 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext {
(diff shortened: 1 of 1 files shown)


dulmandakh:
You can still access these properties from an app and any library using rootProject variable.

gengjiawen:
not in buildscript.

dulmandakh:
how about gradle.properties file?

gengjiawen:
Nope, I think this is better.

PR #20742, opened 2018-08-20
Consolidate native dependencies versions

dulmandakh:
This PR tries to consolidate Android native dependencies versions to make it less error prone to version bumps.

Test Plan:
----------
CI is green


ReactAndroid/build.gradle
```diff
@@ -3,10 +3,11 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-apply plugin: 'com.android.library'
-apply plugin: 'maven'
-
-apply plugin: 'de.undercouch.download'
+plugins {
```
gengjiawen: I think old-fashioned is good and more readable.

dulmandakh:
But this is recommended way and Kotlin DSL compatible. IMHO, Gradle is
recommending to have configuration blocks to make more readable and
performant.

gengjiawen:
Yes, but most project like butter-knife and rxjava use the old-fashioned way. I think kotlin dsl still has a long way to go. Even sqldelight and okio which written in kotlin use the old-fasioned way too https://github.com/square/sqldelight/blob/master/sqldelight-core/build.gradle.

dulmandakh:
Plugins block or configuration blocks in general are recommended way, and
it's more readable. I see no reason to stick with old syntax.

FYI, Kotlin DSL is now 1.0RC3 and will be stable with Gradle 4.10 very
soon. It has superior IDE support and more performant compared to Groovy
DSL.

gengjiawen:
I think the old way is more readable. And adopt to new kotlin dsl we can wait for google officially support it. No need to rush to new syntax since few people using it.

FibreFoX:
Instead of following "other peoples recommendations", does using `plugin`-block fix something? If not, then why change?

dulmandakh:
@hramos please review and merge. plugins block is recommended way to configure Gradle plugins and I found it's used in fresco.

PR #20773, opened 2018-08-21
Make Flow and PropType types checks consistent for view style props.

palexs:
Description:
------------
While creating a custom component I added a prop `style: ?ViewStyleProp`. When I tried to pass the following object 
```
style={{
  opacity: new Animated.Value(0),
}}
```
to this `style` prop I received a yellow box with the warning:
```
Warning: Failed prop type: Invalid prop `opacity` of type `object` supplied to `MyComponent`, expected `number`.
```

Then I browsed through the source code and found an inconsistency in types checks. Namely,
https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js#L158 has the following declaration `opacity?: number | AnimatedNode`, whereas https://github.com/facebook/react-native/blob/master/Libraries/Components/View/ViewStylePropTypes.js#L50 is willing to accept **only** a `number`: `opacity: ReactPropTypes.number`.

Test Plan:
----------
1. Create the following component:
```
import { View, Animated, ViewPropTypes } from 'react-native';
import PropTypes from 'prop-types';

class MyComponent extends React.Component {
  static propTypes = {
    style: ViewPropTypes.style,
  };
  render() {
    return null;
  }
}

  render() {
    return (
      <View>
        <MyComponent
           style={{ opacity: new Animated.Value(0) }}
        />
      </View>
    );
  }
}
```
2. Observe a yellow box warning saying that `opacity` is expected to be a `number`, not an `object`.
3. Apply the fix in this PR.
4. Observe the issue is gone and `Animated.Value` is accepted as a legit value for `opacity` without warnings.

Release Notes:
--------------
[ INTERNAL ] [ MINOR ] [ ViewStylePropTypes ] - Make view style props Flow and PropType checks consistent.

### DIFF
--- a/Libraries/Components/View/ViewStylePropTypes.js
+++ b/Libraries/Components/View/ViewStylePropTypes.js
@@ -10,6 +10,7 @@
 
 'use strict';
 
+const AnimatedNode = require('AnimatedNode');
 const ColorPropType = require('ColorPropType');
 const LayoutPropTypes = require('LayoutPropTypes');
 const ReactPropTypes = require('prop-types');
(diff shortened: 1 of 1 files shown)


TheSavior:
This looks good to me. I’ll land it at work tomorrow if nobody else has landed it by then.

Note that we are starting to deprecate the prop-types in favor of flow and will eventually be removing them. I’d recommend starting to think about not relying on these and instead leveraging flow or typescript. Many of the core components already are fully moved over to Flow.

palexs:
@TheSavior Got it! Thanks!

PR #20787, opened 2018-08-22
[VirtualizedSectionList] - Externalise and handle any sort of data blob

### LINKED ISSUE #20770: VirtualizedSectionList not exposed

## Environment

    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
      Memory: 50.32 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Yarn: 1.7.0 - /usr/local/bin/yarn
      npm: 6.2.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react: 16.4.1 => 16.4.1

## Description
Hello,

I just want to understand why VirtualizedSectionList is not exposed ?
SectionList inherit from it and is exposed, so I'm a little bit confused...

I need this class because my datas are immutable, so I need to specify the getItem props, to handle this kind of data, something that isn't possible with SectionList if I'm right.

EDIT: After a bit of reading of the code, I saw that getItem is override in VirtualizedSectionList, to follow the same logic as with VirtualizedList, getItem should be specify in SectionList, to allow the same behaviour as excepted when we use VirtualizedList with immutable datas.

magrinj:
## Summary

Fixes #20770

## Test Plan:

I test this class in my project with immutable datas, and it's perfectly working:

```
                    <VirtualizedSectionList
                        sections={ sections }
                        renderItem={ this._renderItem }
                        renderSectionHeader={ this._renderSectionHeader }
                        getItem={ this.getItem }
                        getItemCount={ this.getItemCount }
                        keyExtractor={ this.keyExtractor }
                        ItemSeparatorComponent={ this._renderSeparator }
                    />

    getItem(data, key) {
        return data.get(key);
    }

    getItemCount(items) {
        return items.size || 0;
    }

    keyExtractor(item, index) {
        return String(index);
    }
```

## Changelog:

[General] [Changed] - Externalise and handle any sort of data blob

In this PR I externalise VirtualizedSectionList and make it working with any sort of data blob, so sections can be an Immutable Map by exemple.
I do this to follow the logic of VirtualizedList.
Hoping to have made the changes correctly!


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Libraries/Lists/VirtualizedSectionList.js
```diff
@@ -20,34 +20,24 @@ import type {Props as VirtualizedListProps} from 'VirtualizedList';
...
-  }) => ?React.Element<any>,
-  ItemSeparatorComponent?: ?React.ComponentType<any>,
-  keyExtractor?: (item: SectionItem, index: ?number) => string,
-
-  // TODO: support more optional/override props
-  // FooterComponent?: ?ReactClass<any>,
-  // HeaderComponent?: ?ReactClass<any>,
-  // onViewableItemsChanged?: ({viewableItems: Array<ViewToken>, changed: Array<ViewToken>}) => void,
-};
+type SectionBase = any;
```
sahrens: These `any`s are unfortunate - how tricky is it to use more specific types?

magrinj: Unfortunately, I don't think we can specify the type of SectionBase, because sections can be plain JS, Immutable or any data blob, so, by exemple the object tree of Immutable is really different of plain JS. If we keep this definition, and the user specify something else than plain JS, a lot of warning would appear. I understand, that keeping this kind of type definition is important, but in this case I don't find any other solutions, maybe I'm wrong, but if you know a way to handle it, let me know, and I will try to implement it.

magrinj:
Thanks for the review @sahrens, I take a look on the test failure, and try to solve it!

magrinj:
@sahrens From what I've read the flow errors due to `getItem`, `getItemCount` and `getItemParam` are due because the props are inherited from parent to children, and some are no longer required in child. I saw a comment added to fix this error in `VirtualizedList` by Gabe Levi in version 0.44.
So I did the same thing, adding this comment:
```
/* $FlowFixMe - this Props seems to be missing a bunch of stuff. Remove this
 * comment to see the errors */
```
on the top of Props declaration.
I think for the flow inheritance of these classes, it's better to only inherit of OptionalProps by exporting them, what do you think about that ?

magrinj:
@sahrens I also try to implement basic test for `VirtualizedSectionList`, let me know if you don't want it

sahrens:
I think exporting the flow types is a good way to go - gives much more control. Do you want to do that in this PR or separately?

Libraries/Lists/VirtualizedSectionList.js
```diff
@@ -202,6 +168,30 @@ class VirtualizedSectionList<SectionT: SectionBase> extends React.PureComponent<
     );
   }
 
+  scrollToLocation(params: {
```
sahrens: Please avoid unnecessary changes - makes it harder to review. If you think this should be moved here, please submit a separate PR.

Libraries/Lists/VirtualizedSectionList.js
sahrens: There isn't anything more specific we can assume about it? Isn't it at least always an Object?

Libraries/Lists/SectionList.js
```diff
@@ -322,14 +324,40 @@ class SectionList<SectionT: SectionBase<any>> extends React.PureComponent<
...
+    super(props);
+    this._checkProps(this.props);
+  }
+
+  componentDidUpdate(prevProps: Props<SectionT>) {
+    this._checkProps(this.props);
+  }
+
+  _checkProps(props: Props<SectionT>) {
+    const {getItem, getItemCount, getItemParam} = props;
```
sahrens: Can you add a check to only do this with `__DEV__` and use `warning` instead of `invariant`?

sahrens:
And yes, thank you for adding the test :)

Libraries/Lists/VirtualizedSectionList.js
magrinj: You're right, it can be an object! :) I had misread the declaration of `sections` type and confused with the flow type of an element in the array.
I will change it! Sorry

magrinj:
@sahrens I made the changes, but not for the type of the props, I think it's better to be done in another PR, I try to implement it, but a lot of other errors comes out, like `onScroll` that is not declared anywhere in Props, but used in the the classes.
I will try to submit another PR later to achieve this :)
By the way, I hope this one is good for you, and that the changes you requested are being properly implemented.

magrinj:
@sahrens My last commit as you request is ok for you ? :)

magrinj:
@sahrens Any news on this ?
@javache @nicklockwood

magrinj:
@hramos I resolved the conflicts, but I'm stuck with the flow error between `SectionBase` and `SectionT`, any idea ? :)

sahrens:
Let me take a look - can you also run prettier and update?

magrinj:
> Let me take a look - can you also run prettier and update?

@sahrens Thanks! I ran prettier, no change needed on the files I modified :)
What do you mean by update, update the branch again to master change ?

cpojer:
@magrinj can you run `yarn && yarn lint --fix` and rebase this branch?

@sahrens Otherwise, is this ready to be merged?

sahrens:
I'm happy for this to land as soon as it passes lint / flow.

magrinj:
@cpojer I normally already passed the l’inter some commit ago, maybe I need to take a look again but I think it’s ok 
Otherwise I’m little bit confused with the flow error, if you or @sahrens can help me understand and fix it 🙂
I’m in holiday until 25 February, I’ll try again at this time when I have a computer with me

cpojer:
Can we make the new prop optional? Otherwise this will require a bunch of work at FB so I'm not sure if we can land this anytime soon.

magrinj:
@cpojer You mean `getItemParam` prop of `VirtualizedSectionList` ? If it's this one, it can be optional, but I don't think it's a good choice, it should be mandatory to force decomposing the datas correctly without hidden comportement

cpojer:
Basically this is a big breaking change and it's unlikely we'll be able to land it anytime soon if we have to go and update every single place we use this at FB. If you can make this change non-breaking, then it will be much easier to land it.

magrinj:
@cpojer So I remove `getItemParam` prop, since `getItem` should be enough to handle any sort of data blob, I update the tests in consequence.
I think I've also fix the flow errors, I'm not sure if I did it in a good way, but the errors disappear :)

Libraries/Lists/VirtualizedSectionList.js
```diff
@@ -19,35 +19,11 @@ import type {ViewToken} from 'ViewabilityHelper';
...
-  }) => ?React.Element<any>,
-  ItemSeparatorComponent?: ?React.ComponentType<any>,
-  keyExtractor?: (item: SectionItem, index: ?number) => string,
-
-  // TODO: support more optional/override props
-  // FooterComponent?: ?ReactClass<any>,
-  // HeaderComponent?: ?ReactClass<any>,
-  // onViewableItemsChanged?: ({viewableItems: Array<ViewToken>, changed: Array<ViewToken>}) => void,
-};
+type SectionBase = Object;
```
cpojer: This is a regression. Could you work on putting back the proper Flow Type and running flow on the react-native repo to verify it?

Libraries/Lists/VirtualizedSectionList.js
```diff
@@ -115,7 +91,8 @@ type OptionalProps<SectionT: SectionBase> = {
    */
   refreshing?: ?boolean,
 };
-
+/* $FlowFixMe - this Props seems to be missing a bunch of stuff. Remove this
```
cpojer: Can we fix this instead?

cpojer:
I think this looks good but I don't feel comfortable with relaxing the flow types. If you can fix that, I'm happy to land this PR.

Overall, I think it would be great if we could change this API to use an iterator protocol, but I assume that is something to discuss later.

Libraries/Lists/VirtualizedSectionList.js
magrinj: The SectionBase type can't be specified, because it can handle different type of datas, like classic js object or Immutable object, by exemple the data type of `VirtualizedList` is `any`.
For `SectionList` component `SectionBase` can be specified because we know the data must be javascript object.

Libraries/Lists/VirtualizedSectionList.js
magrinj: I'm not really an expert with complex flow typing, I can try to remove this, but I'm not sure if I'm going to fix something of fix it in a good way :/

magrinj:
@cpojer I answer in the comment above for the flow type :)

For the iterator it's a good idea, that's should work very well for `VirtualizedList`, but in `VirtualizedSectionList` some data need to be extract from item, and an iterator protocol wouldn't be enough, I think.

cpojer:
If you can rebase this PR, I can take a look at the flow types.

magrinj:
@cpojer Thanks ! :) I update the branch to the current master and fix the conflicts

Libraries/Lists/VirtualizedSectionList.js
```diff
@@ -523,22 +515,28 @@ class ItemWithSeparator extends React.Component<
...
+  sections: ?$ReadOnlyArray<Item>,
+  index: number,
+): ?Item {
   if (!sections) {
     return null;
   }
   let itemIdx = index - 1;
-  for (let ii = 0; ii < sections.length; ii++) {
-    if (itemIdx === -1 || itemIdx === sections[ii].data.length) {
+  for (let ii = 0; ii < props.getItemCount(sections); ii++) {
```
cpojer: Working on this at FB. I think this is supposed to stay `sections.length` here because `sections` is always an array. I'm changing this internally, no need to do anything, just wanted to put this comment here for your information.

magrinj: @cpojer Thanks ! In fact it can be something else, by exemple if you use an immutable list, to get his length the method is `size` and not `length` so it's better to keep it outside to handle immutable data

cpojer:
Hey @magrinj,

I'm about to land this change from FB's side after fixing up a few things. It seems that there were a bunch of bugs in the code that I fixed, among them the tests for `handles separators correctly` were not actually working properly but they are now.

I made a larger change to remove `getItem` and `getItemCount` calls on `sections` itself altogether. I think we want to be able to use custom data structure for `section.data` but I prefer keeping `sections` as arrays themselves. This should give you most of the value that you were looking for without adding too much dynamism (one `getItem` function to be called on different data structures, which is hard to type and prone to errors). I hope that works for you :)

magrinj:
@cpojer Thanks a lot ! Sorry for the bugs, I didn't have time to test everything again between the different merges. I understand that you want to keep `sections` as a basic javascript array, it will be more easy to maintain in the future, and customisation on `section.data` should be enough for me :)
Again thank you !

Issue #18439, opened 2018-03-18
Alert.alert without message kills apps even after refreshing


fullhdpixel:
Unrecognized selector sent to instance was thrown while invoking alertWithArgs on target AlertManager

![image](https://user-images.githubusercontent.com/7371397/37571169-568b47f0-2af9-11e8-8eb9-931736883f9f.png)
  
## Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 9.2.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.52.0 => 0.52.0

### Steps to Reproduce

https://snack.expo.io/SJG9v82Kz

See: https://stackoverflow.com/questions/41942681/alert-alert-alertmanager-with-params-exception

## Expected Behavior

No crash

## Actual Behavior

Crash

fullhdpixel:
It has all necessary information

siamakEsh:
Same issue.

trev91:
Same issue. Anyone have a work around?

fullhdpixel:
To fix it at least supply an empty string ''

Here is a fix for the reproduction 

 ```
Alert.alert('Uitloggen', 'Weet je zeker dat je wilt uitloggen?', [
      {
        text: 'Annuleer',
        style: 'cancel'
      }, {
        text: 'Ok',
        onPress: () => { //Go to walkthrough
          Meteor.logout()

          this
            .props
            .navigation
            .dispatch(NavigationActions.reset({
              index: 0,
              actions: [NavigationActions.navigate({routeName: 'Walkthrough'})]
            }))
        }
      }
    ], {cancelable: false})
```

shamanking:
I have same issue with react-native version 0.55.2

Tnarita0000:
If there is anything other than a string in the second argument, how do you issue an appropriate warning message in `console.warn`?

https://github.com/facebook/react-native/blob/d01ab66b47a173a62eef6261e2415f0619fefcbb/Libraries/Alert/AlertIOS.js#L89-L94  
If the type is not matched with the above optional argument of alert, it is better to issue an appropriate error.

I can add error message for not matched type errror.

Samiksha19:
Alert.alert(
[...]

PR #20810, opened 2018-08-23
Fix issue #18439: Alert.alert without message kills apps even after refreshing

palexs:
Fixes issue #18439 

Test Plan:
----------
1. Create a simple component:
```
import React from 'react';
import { Text, View, TouchableOpacity, Alert } from 'react-native';

  showAlert = () => {
     Alert.alert('Title', [
      {
        text: 'Cancel',
        style: 'cancel'
      },
      {
        text: 'Ok',
        onPress: () => console.log('Pressed Ok.')
      }
    ], { cancelable: false })
  }
  
  render() {
    return (
      <View>
        <TouchableOpacity onPress={() => this.showAlert()}>
          <Text>Show alert</Text>
        </TouchableOpacity>
      </View>
    );
  }
}
```
2. Run the app and observe an issue described in #18439.
3. Apply the fix.
4. Verify the issue is gone.

Release Notes:
--------------
[ IOS ] [ BUGFIX ] [ AlertIOS ] - Fix AlertIOS so that it doesn't crash when no message is provided.

### DIFF
--- a/Libraries/Alert/AlertIOS.js
+++ b/Libraries/Alert/AlertIOS.js
@@ -96,10 +96,10 @@ class AlertIOS {
       console.warn(
         'AlertIOS.alert() with a 4th "type" parameter is deprecated and will be removed. Use AlertIOS.prompt() instead.',
       );
-      this.prompt(title, message, callbackOrButtons, type);
+      this.prompt(title, message = '', callbackOrButtons, type);
       return;
     }
-    this.prompt(title, message, callbackOrButtons, 'default');
+    this.prompt(title, message = '', callbackOrButtons, 'default');
   }
 
   /**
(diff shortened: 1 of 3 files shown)


palexs:
@shergin Is there a way to re-trigger validations?

palexs:
Not sure why some of the CI validations failed.

Also  the other option is to make the fix more specific to alerts and do not touch `RCTConvert` class. For example, `NSString *message = [RCTConvert NSString:args[@"message"]];` (https://github.com/facebook/react-native/blob/master/React/Modules/RCTAlertManager.m#L67) might be changed by
```
NSString *message = nil;
if (![args[@"message"] isKindOfClass:[NSNull class]]) {
  message = [RCTConvert NSString:args[@"message"]];
}
```

Any thoughts, @shergin?

palexs:
@joshjhargreaves What do you think of this fix?

cpojer:
Thank you for your PR. Unfortunately this doesn't look like the right fix to me. There is some debug code you left in there (unsetting variables) and I think just silencing such an error doesn't seem like the right idea. I think we should probably change the API to not allow undefined values through Flow, and enforce it through typing.

PR #20854, opened 2018-08-25
Fix all build/test issues on master for android, detox & objc CI stages + update to Babel 7

Salakar:
This 'fixes' the following failing CI stages:

  - test_android
  - test_objc
  - test_detox_end_to_end

#### Missing Buck Dependency

This fixes Android tests failing to build on master when running:

`./scripts/run-android-local-unit-tests.sh`

Without this master fails to build with the following error:

image

This commit appears to be the culprit: https://github.com/facebook/react-native/commit/bce77c8b3bb6fc50acee3cf05460a8a4d718bbdb - Android support annotations imported but buck file was not updated.

-----

#### JS Bundles not being transformed with the React Native Babel preset

There's an issue with building JS bundles - no Babel preset is being applied to `RNTester` when using `local-cli` which causes a syntax error on Flow types and on JSX syntax:

image

**KEEP:** To fix this I've added a new format `babel.config.js` file at the root which works for both `jest` (`babelrc: false` is set on options so does not interfere) and `RNTester/local-cli` bundled output (`babel-preset-react-native`). Additionally `react-native-xcode.sh` was updated to run Node from the project root - otherwise it never picks up the babel config and will fail to run with the same issues as before.

**EDIT**: Below is no longer relevant - all updated to now use the release version of Babel v7.0.0
**REMOVE:** Once `babel-preset-react-native` has been updated to use `beta.56` the `babel.config.js` can be updated to just be the following and all the deps taken out the package.json again (except for RN preset):

> This one is for reference and not merging as `babel-preset-react-native` needs it's Babel deps updating to `v7.0.0-beta.56` to match metro's Babel dep version to fix this. To confirm it's building on these versions i've added them to the package.json for now.

```js
// KEEP
module.exports = {
  presets: ['metro-react-native-babel-preset'],
  plugins: [
    '@babel/plugin-transform-runtime',
  ],
};

```

I would PR to `babel-preset-react-native` but I can no longer find it, doesn't seem to exist on GitHub anymore - it was in the root of the project before?
**EDIT**: this is now `metro-react-native-babel-preset` - and latest version is now added.

-----

Test Plan:
----------

  1) Ran the Android local unit test shell script (as above), tests have all passed, output:
image

  2) Ran the jest tests (`npm run test`), tests have all passed, output:
image

  3) Full CI passing 👀 

Release Notes:
--------------

N/A - Internals


react-linter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

package.json
```diff
@@ -203,9 +203,37 @@
...
+    "@babel/plugin-transform-spread": "7.0.0-beta.56",
+    "@babel/plugin-transform-sticky-regex": "7.0.0-beta.56",
+    "@babel/plugin-transform-template-literals": "7.0.0-beta.56",
+    "@babel/plugin-transform-unicode-regex": "7.0.0-beta.56",
+    "@babel/plugin-transform-runtime": "7.0.0-beta.56",
+    "@babel/runtime": "7.0.0-beta.56",
     "async": "^2.4.0",
     "babel-eslint": "9.0.0-beta.2",
-    "babel-generator": "^6.26.0",
+    "babel-preset-react-native": "^5.0.2",
```
kelset: Why is this not `"metro-react-native-babel-preset": "0.43.5"`?

package.json
```diff
@@ -203,9 +203,37 @@
   },
   "devDependencies": {
     "@babel/core": "7.0.0-beta.56",
+    "@babel/generator": "7.0.0-beta.56",
```
kelset: for all the `beta`, I think we want to have babel 7.0 support ASAP so I think we should not merge this until we have everything on `7.0`.

jest/preprocessor.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
         inlineRequires: true,
         minify: false,
         platform: '',
+        babelrc: false,
```
kelset: uhm maybe this fixes https://github.com/facebook/react-native/issues/19859? Not sure

package.json
Salakar: Ah, I could not find where it had moved to as per my comment in the PR. Will update 👍

package.json
Salakar: Agree, should be switching over to the release version of v7 - I had to add these in at a fixed version to align with metro's version of `babel/core` - once metro has been updated to use Babel 7 then all of these except for `@babel/core` can be removed.

`metro-react-native-babel-preset`'s Babel packages would also need updating to the same version of Babel as used here and in `metro` - to avoid random bugs/issues.

jest/preprocessor.js
Salakar: `babelrc: false` was added so that the `babel.config.js` file is ignored by the jest preprocessors transform logic in this file - this babel config file is only needed for `local-cli` whilst running inside CI  & local tests. 

The jest preprocessor requires it's own babel plugins here: https://github.com/facebook/react-native/blob/3af30dfcc73a8fb879c55e573b4072b2f591b33c/jest/preprocessor.js#L65 - so we don't want the `babel.config.js` file to interfere with this otherwise it loads RN specific Babel plugins and causes all sorts of havoc.

package.json
kelset: yup, there is already a PR for that :)

Salakar:
@kelset no worries! 

I'd personally avoid merge this yet as is as we don't really want all these extra Babel dev deps being added to the project - these will be gone shortly; once I add in `"metro-react-native-babel-preset": "0.44.x"` all the others can be removed - will sort and add commits here to cleanup, give me a couple hours 👍 

(I was unaware that the preset had moved to `metro-react-native-babel-preset` - which is why I added these extra dev deps in as a interim solution to waiting for `babel-preset-react-native` to be updated - which as you've pointed out is now part of metro packages - so I will sort)

package.json
Salakar: Ooo noice, missed that one - ok is it worth merging this on `beta.56` (after I sort below changes though) to get build/CI working as is - then once that metro PR is landed and published to NPM I can drop another PR here to update it to the latest versions.

Or we can just wait for metro to be updated - is another option 🙃

kelset: TBH since the PR is passing all the tests  I'd prefer to first have the Metro team release a new version and synch up RN master https://github.com/facebook/metro/pull/233

package.json
kelset: ah yeah sorry I'm dumb 🙈

Salakar:
@kelset changes done and passing again. 

I've switched it over to the `metro-react-native-babel-preset` preset at version `^0.44.0` and also updated the other `metro*` dependencies to match this version (internally these all use beta.56 of Babel 7). 

All the other Babel dependencies have been removed as no longer required.

### Notes

----

 - `@babel/plugin-transform-runtime` (and in the babel.config) is required otherwise the following error occurs on bundling:
image

Potentially related PR: https://github.com/facebook/metro/pull/198/files

----

 - `@babel/runtime` is required to pair with `@babel/plugin-transform-runtime` otherwise the following error occurs:

image

----

Once `metro` has been updated to use Babel 7 the following diff can be applied:

```diff
--- a/package.json
+++ b/package.json
@@ -170,10 +170,10 @@
     "graceful-fs": "^4.1.3",
     "inquirer": "^3.0.6",
     "lodash": "^4.17.5",
-    "metro": "^0.43.6",
-    "metro-babel-register": "^0.44.0",
-    "metro-core": "^0.44.0",
-    "metro-memory-fs": "^0.44.0",
+    "metro": "^0.45.0",
+    "metro-babel-register": "^0.45.0",
+    "metro-core": "^0.45.0",
+    "metro-memory-fs": "^0.45.0",
     "mime": "^1.3.4",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
@@ -202,12 +202,12 @@
     "yargs": "^9.0.0"
   },
   "devDependencies": {
-    "@babel/core": "7.0.0-beta.56",
-    "@babel/runtime": "7.0.0-beta.56",
-    "@babel/plugin-transform-runtime": "7.0.0-beta.56",
-    "metro-react-native-babel-preset": "^0.44.0",
+    "@babel/core": "^7.0.0",
+    "@babel/runtime": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "metro-react-native-babel-preset": "^0.45.0",
     "async": "^2.4.0",
-    "babel-eslint": "9.0.0-beta.2",
+    "babel-eslint": "^9.0.0",
     "detox": "^8.0.0",
     "eslint": "5.1.0",
     "eslint-config-fb-strict": "22.1.0",
```

(`babel-eslint` included as this runs on a beta Babel 7 currently - updating to v9.0.0 will on `^7.0.0`)

Salakar:
Master now already on metro `^0.44.0` versions - synced up master to remove merge conflict. Was passing before merging (in case something on master breaks 😫): 
image

Salakar:
Latest master changes break test_objc and analyze (specifically flow) stages  ¯\_(ツ)_/¯

TheSavior:
The circle jobs might be passing because Hector made them skip running and return success for known broken jobs. You should remove that code from the circle config and verify that the jobs are actually running and passing properly on circle. You should check the detox PR I put up for an example of how to change the circle config.

Salakar:
@TheSavior I'm guessing you mean the following tests on the circle config: https://github.com/facebook/react-native/blob/ee543ecf86824d1c0aec22cfaede4ed3c9548311/.circleci/config.yml#L299-L315

Are these in the scope of this PR? They were skipped prior to this PR, this PR was to address the build issues on tests/CI stages that were currently active but failing, namely the 2 / 3 stages I mentioned above.

I'd be happy to take a stab at these skipped tests in another PR though 👍

TheSavior:
Ah, got it. I thought you were trying to fix the slipped ones.

Cc @hramos for getting this merged then.

Salakar:
https://github.com/facebook/react-native/commit/bf8e1b4ffab4958587efdf3ce97e4ebdd887a20c just been added to master - causes the Babel versions to be out of sync again - it's using Babel 7 beta 52, beta 56 and non beta again on master. 

Once they're all on a release version of Babel 7 it shouldn't matter too much if the versions differ - just there was issues on beta that made some versions not cross-compatible.

kelset:
👋 @Salakar yeah basically you are technically correct - the TLDR of that is that sometimes the releases on the OSS side are a bit off-beat across the different projects, but yeah they'll release a new version of Metro today since that commit landed in that repo too (https://github.com/facebook/metro/commit/add3a4dd0c77dfe80b34ca2f00293ca41e2bf28e) so that we can all proceed :)

kelset:
👋 @Salakar - looks like there is a new commit which seems to me it may be fixing the issue -> https://github.com/facebook/react-native/commit/a549a5377e917f7a4f59c485448b503b43e04d10

could you let me know if it does?

Salakar:
@kelset on it!

Salakar:
@kelset should be good to go now 🤞 

I had to fix a few more things after updating from master; specifically, buck issues - there were redundant & duplicate dependencies defined on master that was breaking build/buck fetch.

Sorry for the delay - on a different Mac today so had to re-setup my environment 😫

kelset:
No problem! Thanks for doing this 💪 I'll check with Mike, maybe we can even get this in for rc4 :)

Salakar:
@kelset no problem :) will also give `RC4` a whirl this weekend and upgrade our tests suite/app for [invertase/react-native-firebase](https://github.com/invertase/react-native-firebase) to use it and see what happens 🔥

hramos:
I just saw this PR - @Salakar thanks so much for working on this!

.flowconfig
```diff
@@ -23,7 +23,7 @@
 .*/Libraries/polyfills/.*
 
 ; Ignore metro
-.*/node_modules/metro/.*
+; .*/node_modules/metro/.*
```
TheSavior: This needs to be kept in sync with the .flowconfig.android file and will fail to land if they are not in sync.

jest/preprocessor.js
TheSavior: Can you add more context here? Maybe include a comment at the top of the Babel.config.js?

What needs it and what needs to not use it?

From your comment I couldn’t tell which thing you are saying jest needs.

Salakar: @TheSavior, `babelrc: false` was added to prevent the Jest tests in the React Native project from using the `babel.config.js` file which was added at the root of the project - this new config file is only for `local-cli` usage whilst testing locally and in CI - without this config file Metro will bundle RN app bundles for the test apps without using the `metro-react-native-babel-preset` preset - which was causing the CI failures..

Without this option (`babelrc: false`) the custom Jest preprocessor for this project ([here](https://github.com/facebook/react-native/blob/master/jest/preprocessor.js)) will read the additional config in `babel.config.js` and subsequently merge that config into the configuration defined in the `transformer.transform` options - https://github.com/facebook/react-native/blob/master/jest/preprocessor.js#L50.

We don't want this to happen as all the babel plugins defined in `metro-react-native-babel-preset` are added to the custom Jest preprocessor - on top of the one already defined in the `transformer` configuration.

.flowconfig
Salakar: Good spot - pushed a commit to update android. Thanks

jest/preprocessor.js
Salakar: @TheSavior have also just added a descriptive comment header on the config file: https://github.com/facebook/react-native/pull/20854/commits/6ad1be5d12fe25db84c6f20c12827d19963bf8f0

Salakar:
Ignore last 2 commits - accidentally merged my local master into my PR branch - reverted immediately though. 🙈

gengjiawen:
@hramos The import failed, need to import again.

Salakar:
Will close this PR now as several commits have now been merged that are a subset of this PR.

Thanks all.

PR #20883, opened 2018-08-28
[fix][android] allow bundling to be disabled for release

deecewan:
It was not previously possible to disable the bundling of assets for android when bundling in release mode. That meant that it would *always* fall back to `true`, because setting `bundleInRelease` to `false` will fall through to `targetName.toLowerCase().contains("release")` (which would be true).

Closes #18892 where I tried this earlier but had problems with the builds that I couldn't understand and seemed unrelated.

Test Plan:
----------
I've been running this patch using `patch-package` for a while, without any dramas. 

It can be tested pretty easily. On current `master`, set `bundleIn<variant>Release` to be false. The bundle will still happen. Switch to this branch and make the same change and the bundle will be disabled.

Release Notes:
--------------

[ANDROID] [BUGFIX] [react.gradle] - Allow disabling of bundling for release builds

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -69,9 +69,11 @@ afterEvaluate {
                     "--reset-cache", "--entry-file", entryFile, "--bundle-output", jsBundleFile, "--assets-dest", resourcesDir, *extraArgs)
             }
 
-            enabled config."bundleIn${targetName}" ||
-                config."bundleIn${variant.buildType.name.capitalize()}" ?:
-                targetName.toLowerCase().contains("release")
+            enabled config."bundleIn${targetName}" != null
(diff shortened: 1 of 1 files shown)


deecewan:
How good `react-native` master builds are failing too.

cpojer:
Ah, seems like we actually made the same change on master already. Thanks for your PR and sorry for not getting back to you earlier, but at least it is fixed :)

deecewan:
@cpojer I think it was my previous PR (#18892) where this got merged in to master. All good 😄

kpelelis:
Hey! Is there an equivalent of this issue for iOS?

PR #20889, opened 2018-08-28
[Android/EditText]: Fix case where TextInput loses text selection ability

### LINKED ISSUE #20887: [Android] Using TextInput inside ViewPagerAndroid causes context menu (copy/paste) in some cases to not display

## Environment

```bash
    System:
      OS: Linux 4.15 Ubuntu 18.04.1 LTS (Bionic Beaver)
      CPU: x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
      Memory: 284.57 MB / 15.54 GB
      Shell: 4.4.19 - /bin/bash
    Binaries:
      Node: 8.11.2 - ~/.nvm/versions/node/v8.11.2/bin/node
      Yarn: 1.9.4 - /usr/bin/yarn
      npm: 6.3.0 - ~/.nvm/versions/node/v8.11.2/bin/npm
    SDKs:
      Android SDK:
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.0, 27.0.1, 27.0.3
        API Levels: 23, 24, 25, 26, 27
    npmPackages:
      @storybook/react-native: ^4.0.0-alpha.16 => 4.0.0-alpha.16
      react: ^16.4.2 => 16.4.2
      react-native: ^0.56.0 => 0.56.0
```

## Description

1. Place a `TextInput` in `ViewPagerAndroid`
2. Attempt to select text from TextInput
3. Android logcat will display: "TextView does not support text selection. Selection cancelled."

The following steps will not *always* reproduce the bug but in some configurations it can cause the effect. One of the apps I'm working on always has this bug however I am yet to find the exact reason why (perhaps due to a heavier component mount tree or subsequent component updates).

Another observation is that this is most prevalent if the device orientation is locked. When rotating to landscape and back to portrait, it seems to regain function. This may be due to [`View.onMeasure(int, int)`](https://developer.android.com/reference/android/view/View.html#onMeasure(int,%20int)) and following `View.onAttachedToWindow()` being called on Native Android.

Another observation is that resuming the app after being in the background for a while will cause this issue.

This has not been working for at least the last couple of RN versions. It does not work on 0.57.0-rc.3 either. Here's someone experiencing the same issue in 2016: [link](https://github.com/facebook/react-native/issues/9958#issuecomment-267965224).

## Reproducible Demo

https://snack.expo.io/SJurHkrvm

Note: Perform a small change in the snack and let the app hot reload. Text selection ability should be lost.

jamsch:
In some instances setting style properties such as height to a TextInput that's inside a ViewPager will not allow the user to select and edit text. 

Android logcat will instead display: "TextView does not support text selection. Selection cancelled."

This seems to be an Android bug -  https://code.google.com/p/android/issues/detail?id=208169

This workaround manually toggles the "enabled" state from false to true when the EditText is attached to the window to circumvent this bug.

Fixes #20887 

Test Plan:
----------
No test plan.

Release Notes:
--------------

[ANDROID] [BUGFIX] [TextInput] - Fix case where TextInput loses text selection ability

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -588,6 +588,15 @@ public class ReactEditText extends EditText {
   @Override
   public void onAttachedToWindow() {
     super.onAttachedToWindow();
+    
+    // Bug workaround for losing text selection ability, see:
+    // https://code.google.com/p/android/issues/detail?id=208169
+    // This should allow users to copy and paste text from TextInput elements inside ViewPagerAndroid
+    if (isEnabled()) {
+      setEnabled(false);
+      setEnabled(true);
+    }
+
     if (mContainsImages) {
       Spanned text = getText();
       TextInlineImageSpan[] spans = text.getSpans(0, text.length(), TextInlineImageSpan.class);


jamsch:
0.58.0-rc.1 seems to have solved this issue

PR #20940, opened 2018-09-01
[android] Allow 'userInfo' for native promise.reject + add native error stack support

Salakar:
### Changes

As mentioned [here](https://github.com/react-native-community/react-native-releases/issues/34#issuecomment-417718601), Android is missing native Promise reject with a `userInfo` `WritableMap` support and also `nativeStack` support (which addresses `TODO(8850038)`). This PR adds Android support for both of these.

React Native on iOS ([here](https://github.com/facebook/react-native/blob/master/React/Base/RCTUtils.m#L433)) and Windows ([here](https://github.com/Microsoft/react-native-windows/pull/732)) already support this so this is a relatively minor addition to bring Android in line with the other platforms. (JS support is also [here](https://github.com/facebook/react-native/blob/master/Libraries/BatchedBridge/NativeModules.js#L145-L148))

Existing methods remain unchanged other than general cleanup of variable names (`e -> throwable`) and adding code comments/docs. 

Additionally, the `ShareTestModule` implementation of Promise (SimplePromise) was updated to reflect these changes - other line changes in this file are from formatting in Android Studio - if this is an issue let me know.

#### Reason for changes

 - Currently inconsistent with other platforms.
 - Blocking a couple of issues over at [invertase/react-native-firebase](https://github.com/invertase/react-native-firebase) - save for re-writing everything to Promise resolve only - which is no small change and isn't a great solution either.

Test Plan:
----------

1. Fixed all build/test issues on master for android, detox & obj-c CI stages and sent up a PR which is now in the process of being imported - #20854. CI for this PR will only pass after that has been merged into master - will rebase when this has happened.
2. With the fixes from '1' locally applied I ran the following suites/scripts with no issues:
	2.1. `npm run test` - Jest tests
	2.2. `./scripts/run-android-local-integration-tests.sh`
	2.3. `./scripts/run-android-local-unit-tests.sh`
	2.4. `./scripts/test-manual-e2e.sh` - built Android and ran the app, clicked through several samples/examples including the `Share` one with no issues.
3. [✈️🔥] Switched over [`invertase/react-native-firebase`](https://github.com/invertase/react-native-firebase/tree/react-native-promise-reject/tests) tests app (powered by [Detox](https://github.com/wix/detox) & [Jet](https://github.com/invertase/jet)) to build from RN Android source with the diff from this PR applied. All our e2e test suites ran with no issues - several hundred methods using Androids native Promise Implementation tested with no issues. 
4. [✈️] For bonus points and sanity checking: using Jet in our tests project (3) I built an extensive e2e test suite in our tests app (3) to test every native promise method (resolve & reject) and the new additions. Code can be see in the following places:
	4.1. [/com/testing/RNPromiseTest.java](https://github.com/invertase/react-native-firebase/blob/react-native-promise-reject/tests/android/app/src/main/java/com/testing/RNPromiseTest.java) - native module exposing all variations of `PromiseImpl`'s methods
	4.2. [/tests/e2e/promise.e2e.js](https://github.com/invertase/react-native-firebase/blob/react-native-promise-reject/tests/e2e/promise.e2e.js) JS test suite directly calling these methods from 4.1 and validating codes, messages, default codes & messages, types, instanceof checks and `nativeStack` frames. 

Imgur

## Feedback

Would be good to get some feedback on the following:

  - `void reject(String message);` - can this deprecated method be removed? It's several versions old and I couldn't find any usages of it anymore.
  - `nativeStack*` platform format - the support I added for `nativeStackAndroid` provides an array of Stack Frames in the same format as already utilised in RN JavaScript code [here](https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/parseErrorStack.js#L13) - this makes it easier to convert to a JS like stack trace if needed or for Crash/Error reporting frameworks to utilise. However, the pre-existing iOS support at [`nativeStackIOS`](https://github.com/facebook/react-native/blob/master/React/Base/RCTUtils.m#L428) provides an unparsed array of call stack symbols which personally I'm not too keen on. What may be good is if both platforms return the same Stack Frame format as implemented on Android in this PR. Once both platforms are in the same format they can then just use `nativeStack` rather than the current post-fixed key names. Feedback would be appreciated :) I could unify these in a subsequent PR based on the feedback - right now not a major issue as the keys are post-fixed by platform.
  - Stack Frames are hard set at 10 frames - easily changable but not from userland code. 10 too much too little?  
	

Release Notes:
--------------

 - [ANDROID] [ENHANCEMENT] [NativeModules] - added support for rejecting a promise with an additional `WritableMap` arg for extra properties (`userInfo`). See the inteface defined in `Promise.java` for available methods. Accessible in JavaScript as `Error.userInfo`. This is to match iOS's existing `Error.userInfo` support.
 - [ANDROID] [ENHANCEMENT] [NativeModules] - added a `nativeStackAndroid` property to promises rejected with an `Exception/Throwable` - making native error stacks available inside Javascript: `Error.nativeStackAndroid`. This is to match iOS's existing `Error.nativeStackIOS` support.


ReactAndroid/src/main/java/com/facebook/react/bridge/PromiseImpl.java
```diff
@@ -5,67 +5,234 @@
...
+    // this matches iOS behavior - iOS adds a `nativeStackIOS` property
+    // iOS: /React/Base/RCTUtils.m -> RCTJSErrorFromCodeMessageAndNSError
+    if (throwable != null) {
+      StackTraceElement[] stackTrace = throwable.getStackTrace();
+      WritableNativeArray nativeStackAndroid = new WritableNativeArray();
+
+      // Build an an Array of StackFrames to match JavaScript:
+      // iOS: /Libraries/Core/Devtools/parseErrorStack.js -> StackFrame
+      for (int i = 0; i < ERROR_STACK_FRAME_LIMIT; i++) {
+        StackTraceElement frame = stackTrace[i];
```
igalmizrahi: Chance of IndexOutOfBoundsException if stacktrace is shorter than 10 elements?

hey99xx: Chance of IndexOutOfBoundsException if stacktrace is shorter than 10 elements?

Salakar: Hmm maybe, I'll update this to iterate to the lowest of the two: ERROR_STACK_FRAME_LIMIT or stackTrace.length

Thanks for the review 👌

PR #20952, opened 2018-09-03
Throw an error if an invalid platform is passed to the bundler

empyrical:
This pull request will have the React Native CLI throw an error if an invalid platform is passed to it, rather than the previous behavior where it would fail silently and give errors about missing files (like `AccessibilityInfo`)

```bash
> react-native bundle --entry-file index.js --bundle-output test.bundle --platform test
Invalid platform (test) selected.
Available platforms are:
  ios, android, windows, web, native
If you are trying to bundle for an out-of-tree platform, it may not be installed.
```

Test Plan:
----------
I've tested passing a registered `--platform` to `react-native bundle`, and it builds without issue. I've also passed an invalid string and this error is thrown as expected.

Release Notes:
--------------
[CLI] [ENHANCEMENT] [local-cli/bundle/buildBundle.js] - Added a check to verify that the `--platform` passed to the bundler is valid
[CLI] [ENHANCEMENT] [local-cli/core/index.js] - Using lodash's `uniq` to get rid of duplicate entries in `platforms` and `providesModuleNodeModules`. Before, "ios" and "android" were in the list twice.

### DIFF
--- a/local-cli/bundle/buildBundle.js
+++ b/local-cli/bundle/buildBundle.js
@@ -17,6 +17,7 @@ const Server = require('metro/src/Server');
 /* $FlowFixMe(site=react_native_oss) */
 const outputBundle = require('metro/src/shared/output/bundle');
 const path = require('path');
+const chalk = require('chalk');
 const saveAssets = require('./saveAssets');
 
 import type {RequestOptions, OutputOptions} from './types.flow';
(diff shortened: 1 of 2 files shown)


cpojer:

grabbou:
Hey @empyrical - that was a great PR. I don't recon you ended up submitting it again to CLI - I would love to see it because it helps with the developer experience which is our priority now!

empyrical:
Hey! I can remake this over in the RNCLI repo 😄

grabbou:
Let's do it!

PR #20992, opened 2018-09-05
Fix #20991 - use name from app.json if available

### LINKED ISSUE #20991: react-native-git-upgrade should use name from app.json instead of package.json

## Environment
```
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-3720QM CPU @ 2.60GHz
      Memory: 82.54 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.9.0 - /usr/local/n/versions/node/10.9.0/bin/node
      npm: 6.4.0 - /usr/local/n/versions/node/10.9.0/bin/npm
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react: ^16.4.1 => 16.4.2
      react-native: ^0.56.0 => 0.56.0
    npmGlobalPackages:
```

## Description
Right now, `react-native-git-upgrade` [gets the app name from `package.json`](https://github.com/facebook/react-native/blob/d985b2e6615bcc49740e8c65485a6e0251eae0cc/react-native-git-upgrade/cliEntry.js#L310) instead of `app.json`. This can be problematic, as NPM allows package names which don't work as iOS/Android package names, or which break the directory structure assumed by react native (for example, the package name `@my-namespace/my-app`).

If an `app.json` file exists (such as when the folder structure is generated by `react-native-cli`), its `name` field should be used instead.

mike-marcacci:
Fix #20991 - use name from `app.json` if available, instead of using `package.json`.

Test Plan:
----------
I haven't tested my plan 😬but didn't want to file a trivial, low-priority issue here without a PR, since those just take up space!

Release Notes:
--------------
[CLI] [ENHANCEMENT] [react-native-git-upgrade] - Use name from app.json if available

### DIFF
--- a/react-native-git-upgrade/cliEntry.js
+++ b/react-native-git-upgrade/cliEntry.js
@@ -114,10 +114,18 @@ function readPackageFiles(useYarn) {
     'package.json',
   );
   const pakPath = path.resolve(process.cwd(), 'package.json');
+  const appPath = path.resolve(process.cwd(), 'app.json');
+  let app = null;
+  try {
+    app = parseJsonFile(appPath);
+  } catch (err) {
+    log.warn('Unable to parse app.json', err.message);
+  }
   return {
     reactNativeNodeModulesPak: parseJsonFile(reactNativeNodeModulesPakPath),
     reactNodeModulesPak: parseJsonFile(reactNodeModulesPakPath),
     pak: parseJsonFile(pakPath),
+    app: app,
   };
 }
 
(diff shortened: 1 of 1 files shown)


hramos:
Looks fine. I'd still like to see a test plan, to make sure this has at least been tested and verified before merging.

This only touches the upgrade CLI, which is not used at Facebook, so any of our contributors should feel free to land this if they approve of the PR.

mike-marcacci:
OK, FWIW I used this change when upgrading `@boltline/boltline-native`. Upgrading ran into [another unrelated issue](https://github.com/facebook/react-native/issues/21156) but the name was no longer a problem. Let me know if anyone has a programatic test strategy you'd like implemented here.

mikemorris:
This looks like a good fix to me, only caution is that IIRC iOS preserves `AppName` app names throughout, while Android downcases to `appname` when creating directory paths, would want to double-check that behavior stays consistent.

PR #20999, opened 2018-09-06
New LayoutContext API to access the frame and safe area of the root view

janicduplessis:
Motivation:
----------
There is currently no way to access layout metrics from the root view that contains the component tree. This can be useful when you need to access these values from JS synchronously without having to rely on the `onLayout` event of a top level view which is async and can cause flickers.

It is roughly the equivalent of the `Dimensions` module for screen / window size but instead based on the `RootView` that the component is rendered in with a more modern `Context` based api. This also be useful for hybrid apps where the root view might not cover the entire screen.

Another goal of this is to be able to expose `safeAreaInset` values which cannot be accessed currently from JS. `SafeAreaView` is sometimes not flexible enough if we want to apply the insets manually in JS (for example use margins instead of padding).

Test Plan:
----------
### iOS RNTester

The red square is a view that has width and height set to the LayoutContext frame with padding equal to the LayoutContext safeAreaInsets. The blue square is a view that fills the content of its parent to show the safe areas.

Sample code from the RNTester example to illustrate this

```js
return (
  <LayoutContext>
    {({layout, safeAreaInsets}) => (
      <View
        style={{
          width: layout.width,
          height: layout.height,
          paddingTop: safeAreaInsets.top,
          paddingRight: safeAreaInsets.right,
          paddingBottom: safeAreaInsets.bottom,
          paddingLeft: safeAreaInsets.left,
          backgroundColor: 'red',
        }}>
          <View style={{flex: 1, backgroundColor: 'blue'}} />
        </View>
    )}
  </LayoutContext>
);
```

### iOS

- Test that safe area insets and frame match the actual app frame for iOS 11 and iOS < 11 (polyfilled version)

- Test insets on iPhone X with screen rotation

- Test insets on iPhone X with root view that doesn't cover the screen

- Test that showing and hiding the status bar updates the safe area. This doesn't work on iOS < 11 but also doesn't work for the SafeAreaView component. I couldn't find a way to make it work since the status bar frame changed event is not triggered when hiding / showing. Also kvo for the statusBarHidden prop did not work either for some reason... This isn't a big deal anyway.

- Test toggling the on-call status bar

### Android

- Test changing the status bar hidden prop
- Test changing the status bar translucent prop
- Test screen rotation
- Test making the navigation bar transparent (https://stackoverflow.com/a/31596735)

Release Notes:
--------------
[General] [Added] - New LayoutContext API to access the frame and safe area of the root view

### DIFF
--- a/Libraries/ReactNative/RootViewLayout.js
+++ b/Libraries/ReactNative/RootViewLayout.js
@@ -1,0 +1,82 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow local-strict
+ * @format
+ */
+
+const React = require('React');
+const EmitterSubscription = require('EmitterSubscription');
+const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
+
+import type {Layout} from 'CoreEventTypes';
+
+export type LayoutContext = $ReadOnly<{|
+  layout: Layout,
+  safeAreaInsets: $ReadOnly<{|
+    top: number,
+    right: number,
+    bottom: number,
+    left: number,
+  |}>,
+|}>;
+
+/**
+ * Context used to provide layout metrics from the root view the component
+ * tree is being rendered in. This is useful when sync measurements are
+ * required.
+ */
+const Context: React.Context<LayoutContext> = React.createContext({
+  layout: {x: 0, y: 0, width: 0, height: 0},
+  safeAreaInsets: {top: 0, right: 0, bottom: 0, left: 0},
+});
+
+type Props = {|
+  children: React.Node,
+  initialLayoutContext: LayoutContext,
+  rootTag: number,
+|};
+
+type State = {
+  layoutContext: LayoutContext,
+};
+
+class RootViewLayoutManager extends React.Component<Props, State> {
+  state = {
+    layoutContext: this.props.initialLayoutContext,
+  };
+  _subscription: EmitterSubscription | null = null;
+
+  componentDidMount() {
+    this._subscription = RCTDeviceEventEmitter.addListener(
+      'didUpdateLayoutContext',
+      ({rootTag, layoutContext}) => {
+        if (rootTag === this.props.rootTag) {
+          this.setState({layoutContext});
+        }
+      },
+    );
+  }
+
+  componentWillUnmount() {
+    if (this._subscription !== null) {
+      this._subscription.remove();
+    }
+  }
+
+  render() {
+    return (
+      <Context.Provider value={this.state.layoutContext}>
+        {this.props.children}
+      </Context.Provider>
+    );
+  }
+}
+
+module.exports = {
+  Context,
+  Manager: RootViewLayoutManager,
+};
--- a/Libraries/ReactNative/RootViewLayout.js
+++ b/Libraries/ReactNative/RootViewLayout.js
@@ -9,7 +9,6 @@
  */
 
 const React = require('React');
-const EmitterSubscription = require('EmitterSubscription');
 const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
 
 import type {Layout} from 'CoreEventTypes';
@@ -40,40 +39,29 @@ type Props = {|
   rootTag: number,
 |};
 
-type State = {
-  layoutContext: LayoutContext,
-};
-
-class RootViewLayoutManager extends React.Component<Props, State> {
-  state = {
-    layoutContext: this.props.initialLayoutContext,
-  };
-  _subscription: EmitterSubscription | null = null;
-
-  componentDidMount() {
-    this._subscription = RCTDeviceEventEmitter.addListener(
+function RootViewLayoutManager({
+  children,
+  initialLayoutContext,
+  rootTag,
+}: Props) {
+  const [layoutContext, setLayoutContext] = React.useState<LayoutContext>(
+    initialLayoutContext,
+  );
+  React.useEffect(() => {
+    const subscription = RCTDeviceEventEmitter.addListener(
       'didUpdateLayoutContext',
-      ({rootTag, layoutContext}) => {
-        if (rootTag === this.props.rootTag) {
-          this.setState({layoutContext});
+      event => {
+        if (rootTag === event.rootTag) {
+          setLayoutContext(event.layoutContext);
         }
       },
     );
-  }
+    return () => {
+      subscription.remove();
+    };
+  }, [rootTag]);
 
-  componentWillUnmount() {
-    if (this._subscription !== null) {
-      this._subscription.remove();
-    }
-  }
-
-  render() {
-    return (
-      <Context.Provider value={this.state.layoutContext}>
-        {this.props.children}
-      </Context.Provider>
-    );
-  }
+  return <Context.Provider value={layoutContext}>{children}</Context.Provider>;
 }
 
 module.exports = {
--- a/Libraries/ReactNative/RootViewLayout.js
+++ b/Libraries/ReactNative/RootViewLayout.js
@@ -47,7 +47,7 @@ function RootViewLayoutManager({
   const [layoutContext, setLayoutContext] = React.useState<LayoutContext>(
     initialLayoutContext,
   );
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     const subscription = RCTDeviceEventEmitter.addListener(
       'didUpdateLayoutContext',
       event => {
--- a/Libraries/ReactNative/renderApplication.js
+++ b/Libraries/ReactNative/renderApplication.js
@@ -16,8 +16,11 @@ import type {IPerformanceLogger} from 'createPerformanceLogger';
 import PerformanceLoggerContext from 'PerformanceLoggerContext';
 const React = require('React');
 const ReactFabricIndicator = require('ReactFabricIndicator');
+const RootViewLayout = require('RootViewLayout');
 
 const invariant = require('invariant');
+
+import type {LayoutContext} from 'RootViewLayout';
 
 // require BackHandler so it sets the default handler that exits the app if no listeners respond
 require('BackHandler');
@@ -26,6 +29,7 @@ function renderApplication<Props: Object>(
   RootComponent: React.ComponentType<Props>,
   initialProps: Props,
   rootTag: any,
+  initialLayoutContext: LayoutContext,
   WrapperComponent?: ?React.ComponentType<*>,
   fabric?: boolean,
   showFabricIndicator?: boolean,
@@ -36,12 +40,16 @@ function renderApplication<Props: Object>(
   let renderable = (
     <PerformanceLoggerContext.Provider
       value={scopedPerformanceLogger ?? GlobalPerformanceLogger}>
-      <AppContainer rootTag={rootTag} WrapperComponent={WrapperComponent}>
-        <RootComponent {...initialProps} rootTag={rootTag} />
-        {fabric === true && showFabricIndicator === true ? (
-          <ReactFabricIndicator />
-        ) : null}
-      </AppContainer>
+      <RootViewLayout.Manager
+        initialLayoutContext={initialLayoutContext}
+        rootTag={rootTag}>
+        <AppContainer rootTag={rootTag} WrapperComponent={WrapperComponent}>
+          <RootComponent {...initialProps} rootTag={rootTag} />
+          {fabric === true && showFabricIndicator === true ? (
+            <ReactFabricIndicator />
+          ) : null}
+        </AppContainer>
+      </RootViewLayout.Manager>
     </PerformanceLoggerContext.Provider>
   );
 
--- a/RNTester/js/LayoutContextExample.js
+++ b/RNTester/js/LayoutContextExample.js
@@ -1,0 +1,112 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const React = require('react');
+const {
+  LayoutContext,
+  StyleSheet,
+  View,
+  StatusBar,
+  Button,
+  Platform,
+} = require('react-native');
+
+const BORDER_WIDTH = 2;
+
+type Props = {
+  onExampleExit: () => void,
+};
+
+type State = {
+  hidden: boolean,
+  translucent: boolean,
+};
+
+class LayoutContextExample extends React.Component<Props, State> {
+  static title = '<LayoutContext>';
+  static description =
+    'LayoutContext allows getting layout metrics for the current root view.';
+  static external = true;
+
+  state = {
+    hidden: false,
+    translucent: false,
+  };
+
+  render() {
+    return (
+      <LayoutContext>
+        {({layout, safeAreaInsets}) => {
+          return (
+            <>
+              <StatusBar
+                hidden={this.state.hidden}
+                translucent={this.state.translucent}
+                backgroundColor="rgba(0, 0, 0, 0.3)"
+              />
+              <View
+                style={[
+                  styles.container,
+                  {
+                    width: layout.width,
+                    height: layout.height,
+                    paddingTop: safeAreaInsets.top - BORDER_WIDTH,
+                    paddingRight: safeAreaInsets.right - BORDER_WIDTH,
+                    paddingBottom: safeAreaInsets.bottom - BORDER_WIDTH,
+                    paddingLeft: safeAreaInsets.left - BORDER_WIDTH,
+                  },
+                ]}>
+                <View style={styles.content}>
+                  <Button
+                    title="Toggle status bar hidden"
+                    onPress={() =>
+                      this.setState(state => ({hidden: !state.hidden}))
+                    }
+                  />
+                  {Platform.OS === 'android' && (
+                    <Button
+                      title="Toggle status bar translucent"
+                      onPress={() =>
+                        this.setState(state => ({
+                          translucent: !state.translucent,
+                        }))
+                      }
+                    />
+                  )}
+                  <Button title="Close" onPress={this.props.onExampleExit} />
+                </View>
+              </View>
+            </>
+          );
+        }}
+      </LayoutContext>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    borderWidth: BORDER_WIDTH,
+    borderColor: 'red',
+    backgroundColor: '#589c90',
+  },
+  content: {
+    flex: 1,
+    backgroundColor: 'white',
+    alignItems: 'center',
+    justifyContent: 'center',
+    borderWidth: BORDER_WIDTH,
+    borderColor: 'blue',
+  },
+});
+
+module.exports = LayoutContextExample;
--- a/RNTester/js/LayoutContextExample.js
+++ b/RNTester/js/LayoutContextExample.js
@@ -15,6 +15,7 @@ const {
   LayoutContext,
   StyleSheet,
   View,
+  Text,
   StatusBar,
   Button,
   Platform,
@@ -45,7 +46,7 @@ class LayoutContextExample extends React.Component<Props, State> {
   render() {
     return (
       <LayoutContext>
-        {({layout, safeAreaInsets}) => {
+        {ctx => {
           return (
             <>
               <StatusBar
@@ -57,15 +58,23 @@ class LayoutContextExample extends React.Component<Props, State> {
                 style={[
                   styles.container,
                   {
-                    width: layout.width,
-                    height: layout.height,
-                    paddingTop: safeAreaInsets.top - BORDER_WIDTH,
-                    paddingRight: safeAreaInsets.right - BORDER_WIDTH,
-                    paddingBottom: safeAreaInsets.bottom - BORDER_WIDTH,
-                    paddingLeft: safeAreaInsets.left - BORDER_WIDTH,
+                    width: ctx.layout.width,
+                    height: ctx.layout.height,
+                    paddingTop: ctx.safeAreaInsets.top - BORDER_WIDTH,
+                    paddingRight: ctx.safeAreaInsets.right - BORDER_WIDTH,
+                    paddingBottom: ctx.safeAreaInsets.bottom - BORDER_WIDTH,
+                    paddingLeft: ctx.safeAreaInsets.left - BORDER_WIDTH,
                   },
                 ]}>
                 <View style={styles.content}>
+                  <Text
+                    style={{
+                      marginBottom: 32,
+                      backgroundColor: '#eee',
+                      padding: 16,
+                    }}>
+                    {JSON.stringify(ctx, null, 2)}
+                  </Text>
                   <Button
                     title="Toggle status bar hidden"
                     onPress={() =>
@@ -102,10 +111,9 @@ const styles = StyleSheet.create({
   content: {
     flex: 1,
     backgroundColor: 'white',
-    alignItems: 'center',
-    justifyContent: 'center',
     borderWidth: BORDER_WIDTH,
     borderColor: 'blue',
+    padding: 32,
   },
 });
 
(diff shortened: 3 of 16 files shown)


TheSavior:
Cc @yungsters @shergin, what do you think of this approach?

yungsters:
This is pretty cool. My main concern would be that when the values change, we must trigger a re-render. With this approach, my understanding is that when something causes the safe area insets to change (e.g. orientation change), we would have to force a re-render with the new values.

Ideally, we would not have to necessitate a re-render in JavaScript. But in reality, I think this solves the problem very elegantly and 99% of the time, people will not experience consequences of my concern.

@shergin, I know you had more thoughts than anyone else about the future of `SafeAreaView`. What do you think about this approach?

janicduplessis:
@yungsters I think in a lot of cases the re-render will get blocked by some pure component up in the tree, and even if it doesn't I guess re-rendering on orientation change is not a bad thing.

janicduplessis:
The main difference between this API and `SafeAreaView` is that `SafeAreaView` does take into consideration it's position inside the `RootView` and adjusts its insets accordingly. The drawback is that you have no control over how the insets are applied to the view and cannot know their value.

This API has the advantage of exposing the inset values to JS with a synchronous API so they can be used during the first render, this allow more flexibility to choose how the insets are being applied to views. The main drawback is that it only considers the position of the `RootView` in the screen so you have to make sure the view covers the whole `RootView` if you want to apply the insets directly. In the case the view doesn't cover the screen or you don't need the flexibility of knowing values in JS then SafeAreaView might be a better choice.

I think the APIs are kind of complementary for now since they both have different tradeoffs.

Libraries/ReactNative/RootViewLayoutContext.js
```diff
@@ -0,0 +1,33 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+const React = require('React');
+
+export type LayoutContext = $ReadOnly<{|
+  frame: $ReadOnly<{|
```
AlicanC: Cloud this be "layout" instead of "frame" and use the type "Layout" from "Libraries/Types/CoreEventTypes.js" instead of defining its own?

janicduplessis: Good point, I'll update it.

janicduplessis:
@yungsters This should address your concern about re-rendering.

janicduplessis:
Actually we don't target api 28 yet so I cannot use the notch api. I finished the android implementation and just need to do some more testing on different device / api version combinations.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -471,6 +472,87 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+        rotation == Surface.ROTATION_90 ? navbarHeight : 0,
+        rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_180 ? navbarHeight : 0,
+        rotation == Surface.ROTATION_270 ? navbarHeight : 0);
+    }
+
+    // Calculate the part of the root view that overlaps with window insets.
+    int[] windowLocation = new int[2];
+    getLocationInWindow(windowLocation);
+    DisplayMetrics screenMetrics = DisplayMetricsHolder.getScreenDisplayMetrics();
+    windowInsets.top = Math.max(windowInsets.top - windowLocation[1], 0);
```
kmagiera: This seem inconsistent with iOS. When view occupies bottom half of the screen on iOS the top inset will be 0 whereas on Android it will be `windowHeight/2`

(same applies to left, bottom and right insets)

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -471,6 +472,87 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+    windowInsets.left = Math.max(windowInsets.left - windowLocation[0], 0);
+    windowInsets.bottom = Math.max(windowLocation[1] + getHeight() + windowInsets.bottom - screenMetrics.heightPixels, 0);
```
kmagiera: note that `screenMetrics.heightPixels` includes the height of navigation bar (https://developer.android.com/training/system-ui/navigation) while it is not included in the window height.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -471,6 +472,87 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+        insets.getSystemWindowInsetBottom(),
+        insets.getSystemWindowInsetLeft());
+    } else {
+      int rotation =
+        ((WindowManager) getContext().getSystemService(Context.WINDOW_SERVICE))
+          .getDefaultDisplay().getRotation();
+      int statusBarHeight = 0;
+      int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
+      if (resourceId > 0) {
+        statusBarHeight = getResources().getDimensionPixelSize(resourceId);
```
kmagiera: this method seem not to care whether status bar is shown or not it returns a system constant. I'm not 100% sure about it but I think this value does not reflect the actual status of the UI

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -471,6 +472,87 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+      int navbarHeight = 0;
+      resourceId = getResources().getIdentifier("navigation_bar_height", "dimen", "android");
+      if (resourceId > 0) {
+        navbarHeight = getResources().getDimensionPixelSize(resourceId);
+      }
+
+      windowInsets = new RootViewInsets(
+        statusBarHeight,
+        rotation == Surface.ROTATION_90 ? navbarHeight : 0,
+        rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_180 ? navbarHeight : 0,
```
kmagiera: Apparently some tablet devices can render navigation bar on the bottom while reporting landscape rotation 😢  Maybe we can use this method as a polyfil -> https://developer.android.com/reference/android/view/View.html#getWindowVisibleDisplayFrame(android.graphics.Rect)

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
janicduplessis: Are you sure about this?

If the root view covers the bottom half of the screen then it would be:
Lets say that the status bar is visible and `windowInsets.top` = 25 initially

`Math.max(25 - ( windowHeight / 2), 0) -> Math.max(-somebignumber, 0) -> 0`

Let me know if there is something that could make the code clearer.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
janicduplessis: I *think* this is fine, do you have a case in mind where this would not work? I did test showing / hiding the navigation bar and it did work properly.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
janicduplessis: Yea I did not handle the case in the polyfilled version, I guess it can be done easily by checking the window flags. On a sidenote the iOS polyfill also doesn't handle this properly so it is partly why I didn't bother with it.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
janicduplessis: Ew, I'll look into this method to see if I can use this for the polyfill instead of using status bar / navigation bar dimensions.

pribeh:
Tested but getting "use of undeclared identifier '_emulateUnlessSupported' on line 25 of RCTSafeAreaView.m" from xcode. I will attempt to patch again and retest. Amazing work so far on this @kmagiera

cpojer:
Hey @janicduplessis. I know you are super busy right now with other stuff but I'm wondering what the status of this PR is?

Libraries/ReactNative/RootViewLayout.js
```diff
@@ -0,0 +1,82 @@
...
+ * tree is being rendered in. This is useful when sync measurements are
+ * required.
+ */
+const Context: React.Context<LayoutContext> = React.createContext({
+  layout: {x: 0, y: 0, width: 0, height: 0},
+  safeAreaInsets: {top: 0, right: 0, bottom: 0, left: 0},
+});
+
+type Props = {|
+  children: React.Node,
```
sahrens: I think React.Children is preferred:
```suggestion
  children: React.Children,
```

Libraries/ReactNative/RootViewLayout.js
```diff
@@ -0,0 +1,82 @@
...
+
+type State = {
+  layoutContext: LayoutContext,
+};
+
+class RootViewLayoutManager extends React.Component<Props, State> {
+  state = {
+    layoutContext: this.props.initialLayoutContext,
+  };
+  _subscription: EmitterSubscription | null = null;
```
sahrens:
```suggestion
  _subscription: ?EmitterSubscription = null;
```

Libraries/ReactNative/RootViewLayout.js
```diff
@@ -0,0 +1,82 @@
...
+
+  componentDidMount() {
```
sahrens: Want to use a functional component with the `useEvent` hook?

Libraries/ReactNative/renderApplication.js
```diff
@@ -13,29 +13,37 @@
...
   invariant(rootTag, 'Expect to have a valid rootTag, instead got ', rootTag);
 
   let renderable = (
-    <AppContainer rootTag={rootTag} WrapperComponent={WrapperComponent}>
-      <RootComponent {...initialProps} rootTag={rootTag} />
-      {fabric === true && showFabricIndicator === true ? (
-        <ReactFabricIndicator />
-      ) : null}
-    </AppContainer>
+    <RootViewLayout.Manager
```
sahrens: I wonder if we should just bake this new root layout functionality into AppContainer since it already has subscriptions and context and state - that would reduce the "wrapper hell" we have going on. Ideally we could factor out the functionality into separate hook modules `useRootLayoutProvider` / `useRootLayout`?

RNTester/js/LayoutContextExample.js
```diff
@@ -0,0 +1,112 @@
...
+type Props = {
+  onExampleExit: () => void,
+};
+
+type State = {
+  hidden: boolean,
+  translucent: boolean,
+};
+
+class LayoutContextExample extends React.Component<Props, State> {
```
sahrens: What do you think about also dumping the content of the context to the screen with JSON.stringify? That can be really handy for seeing what's available and debugging in examples like this.

Libraries/ReactNative/renderApplication.js
janicduplessis: The reason for not having it inside `AppContainer` is that it is also used in modal here https://github.com/facebook/react-native/blob/007e00fa70d0e1d69f7148e901790d2092e15b8c/Libraries/Modal/Modal.js#L244

RNTester/js/LayoutContextExample.js
janicduplessis: 👍

janicduplessis:
@cpojer It's mostly done, the main thing left is address the review by @kmagiera and have the ios part reviewed.

Libraries/ReactNative/renderApplication.js
janicduplessis: Adding a `useRootLayout` hook sounds good, it think it will just be a simple wrapper around `useContext`.

janicduplessis:
@sahrens Also the main reason for having the `ContextProvider` and subscription logic in a separate component is to avoid re-rendering the whole app when the layout changes. I explained the reasoning in more details here https://github.com/facebook/react-native/pull/20999#issuecomment-450526535

cpojer:
@kmagiera could you do us the honors? :)

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
kmagiera: ah, sorry I think you are right. Please ignore this comment

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
kmagiera: Ah, I think you are right about this too. I got confused because of that difference between `getHeight` and `screenMetrcs.heightPixels` but I think `windowInsets.bottom` exactly accounts for that (that is it is equal to the height of the navbar)

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -471,6 +472,87 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+      int rotation =
+        ((WindowManager) getContext().getSystemService(Context.WINDOW_SERVICE))
+          .getDefaultDisplay().getRotation();
+      int statusBarHeight = 0;
+      int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
+      if (resourceId > 0) {
+        statusBarHeight = getResources().getDimensionPixelSize(resourceId);
+      }
+      int navbarHeight = 0;
+      resourceId = getResources().getIdentifier("navigation_bar_height", "dimen", "android");
```
kmagiera: Sadly, navbar height is different on portrait vs landscape (in landscape it is a bit more narrow). This is a frequent mistake people make when they think their app W/H ratio stays the same when the device rotates. We should use `navigation_bar_height_landscape` in landscape.

And if that wasn't enough sometimes the navbar is vertical in which case you need to use `navigation_bar_width` instead 🤦‍♂️ -> you may want to check this SO answer https://stackoverflow.com/questions/21057035/detect-android-navigation-bar-orientation

Libraries/ReactNative/RootViewLayout.js
```diff
@@ -0,0 +1,82 @@
...
+        if (rootTag === this.props.rootTag) {
+          this.setState({layoutContext});
+        }
+      },
+    );
+  }
+
+  componentWillUnmount() {
+    if (this._subscription !== null) {
+      this._subscription.remove();
```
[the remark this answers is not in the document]
sahrens: Change the check to `!=` instead of `!==` to also check for `undefined`

React/Base/RCTRootView.m
```diff
@@ -366,6 +391,35 @@ - (void)contentViewInvalidated
   [self showLoadingView];
 }
 
+- (void)maybeUpdateLayoutContext
+{
+  CGRect frame = self.frame;
+  UIEdgeInsets safeAreaInsets = RCTSafeAreaInsetsForView(self, YES);
+  if (
+    RCTUIEdgeInsetsEqualToEdgeInsetsWithThreshold(safeAreaInsets, _lastSafeAreaInsets, RCTScreenScale()) &&
+    CGRectEqualToRect(frame, _lastFrame)
```
sahrens: why not use threashold for both?

kmagiera:
I think this is mostly ok. I left one Android related comment last time I reviewed and I think it would be good to address it prior to merging.

However, one more thing I've been thinking of recently is about what kind of impact this change would make for the case when someone renders RN root views as list view cells (or in any scrollable container). In that case I think what is going to happen is that those events will get triggered for every scroll event which may be a bit too often (e.g. in RN we normally throttle scroll events). So maybe we should think of adding a way to disable this behavior for apps that don't just have one full-screen root view?

grabbou:
Can we also test the navigation bar being toggled? There's few issues with that in the current Dimensions API. 

Also, shall we deprecate Dimensions API as a result of that?

janicduplessis:
@grabbou I did test that and it works :)

Theorically this api is a bit different since it is dimensions for the root view and not the actual screen so it could be different from Dimensions.

sahrens:
I'd love to get this in - are you going to address @kmagiera's comments?

janicduplessis:
@sahrens We discussed this and the plan is to cleanup the JS side and land it as is. We can iterate on fixing the android edge cases after. Since it is a new api it should not cause issues.

React/Base/RCTRootView.m
janicduplessis: The existing safe area view code used this when comparing safe area values so I kept it. For frames it does not seem to be needed, I did a quick search for `threshold` and we don't seem to do anything special when comparing frames. Not sure why we need `RCTUIEdgeInsetsEqualToEdgeInsetsWithThreshold` in the first place.

janicduplessis:
@sahrens @cpojer @kmagiera @grabbou Should be good to ship. The only things left is to address feedback around the android safe area polyfill code but we can do this in a follow up when we hit some edge cases on some android devices. It would also be nice to use the android notch api on API 28+.

janicduplessis:
Discovered an issue when using a release build on iOS. When we get the initial `layoutContext` the view has not yet been added to the view hierarchy so its measurements are 0. Since `useEffect` is async it caused the event to get completely lost since the layout context change event would be triggered between the first render and the time effects are called. Using `useLayoutEffect` works around the issue and makes sure no events are lost since it is called synchronously (same timing as cDM). This still cause a bad initial render and a re-render that should not be needed.

Looking into a way to defer app initialization after the view has been added to the hierarchy instead of when the RootView instance is created. This usually happen pretty much at the same time so it should be fine but I'll have to make sure to measure the perf impact of this.

shergin:
Objective-C part of the code looks good to me.

Even if I don't think that this is a right way/API to expose the desired functionality, I think it's fair to land this because seems nobody still has a better idea.

I would like to highlight some concerns that I have:
 * I believe the API should be declarative and *just work*. That will help us resolve most of the problem that I mention below. (That's still unclear how exactly that can be implemented and designed though.)
 * The proposed API describes the properties of RootView, not a particular hosting view of `Context`; that's inflexible and confusing.
 * Layout is multi-pass operation by design. Asynchronous handling of layout events will make the usage quirky and unreliable.
 * Exposing `{x: 0, y: 0}` is not reliable. The coordinates might change without notifying RCTRootView.

janicduplessis:
@shergin Can you elaborate on

> Exposing {x: 0, y: 0} is not reliable. The coordinates might change without notifying RCTRootView.

I agree with the other points, we can probably re-visit post fabric for a better solution as a true replacement for SafeAreaView. In the meantime this is a good replacement for the `Dimensions` module for most use cases.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -669,6 +763,22 @@ private void checkForDeviceDimensionsChanges() {
...
+        mFrameHeight != getHeight() ||
+        mFrameX != getX() ||
+        mFrameY != getY() ||
+        !areInsetsEqual(mSafeAreaInsets, safeAreaInsets)) {
+        mSafeAreaInsets = safeAreaInsets;
+        mFrameWidth = getWidth();
+        mFrameHeight = getHeight();
+        mFrameX = getX();
+        mFrameY = getY();
+        emitLayoutContextEvent(safeAreaInsets);
```
satya164: Will this trigger if the insets change when the app is drawn below the statusbar/navigation bar? Maybe we need a `setOnApplyWindowInsetsListener`?

janicduplessis: I think it did work, I remember testing different combinations of toggling the status bar hidden and translucent props. Seems like the global layout listener does get called when system insets change.

brunolemos:
I've been meaning to have a `useSafeAreaInsets` hook and it seems this pr will allow this 🙌🙌

:shipit:

sahrens:
Quick update - there are a lot of flow errors internally trying to pull this in. I'll work through them soon.

janicduplessis:
Also little update: I’ve shipped this to production for a few weeks and so far iOS is solid but I got a crash that happens kind of infrequently on Android that I need to investigate before we can ship.

janicduplessis:
@sahrens 

- You are talking about https://github.com/facebook/react-native/pull/20999/files#diff-5fb3b29563ba1e3091b5b0ce41f006d2R32 right? It is basically there just to avoid having to type the context as nullable since it does require to pass an initial value but it will always be initialized when the RootView is created.
- Agree
- Agree, I also ran into this when creating a hook version in my app and made the change already in my fork.
- Good point, it probably doesn't work at the moment. The modal could create a new context to provide its size to components that are inside of it. Not sure if this is blocking to land this, especially if we make it clear that the size is the one of the current RootView.
- AFAIK Fabric integration is just a matter of re-implementing the logic to pass the layout context in the Fabric surface `initialProps` and calling `DeviceEventEmitter` when the layout context changes.
- Making the LayoutContext tied to the RootView instead of some global value like the old Dimensions module solves a lot of technical issues because some elements like safeAreaInsets are only available at the View level. I think it also plays nicely with multiple RootViews and eventually multiple screens.

sahrens:
(1) I'm more concerned with people relying on x/y and having them change without an event firing as opposed to just the initial value in javascript.

(3) can you update this PR? I still see this:
```
  get LayoutContext() {
    return require('RootViewLayout').Context.Consumer;
  },
```
(4) Fabric decouples starting the app with the initial props from the root view so it will be tricky to link the two for a clean initial render.
(5) I think if the naming is clear then lack of modal support isn't blocking, but would be nice. I think a lot or most people conceptually think of modals as a new root, especially from a layout perspective.

janicduplessis:
@sahrens  I'm still not sure I get (1), x/y should get updated whenever it changes (in `layoutSubview` on iOS and the global layout listener on Android). Or do you mean these values should always be zero? Like `bounds` vs `frame` on `UIView`.

sahrens:
I got this working internally but unfortunately it really hurt our performance benchmarks on iOS.  I’m guessing the delay of bundleFinishedLoading is the most likely issue - any ideas?

shergin:
@janicduplessis @sahrens 

By this:
> Exposing {x: 0, y: 0} is not reliable. The coordinates might change without notifying RCTRootView.

I mean that the action position of a RootView depends on all views around it and seems there is no easy/cheap way to listen for those changes. :(

nyanev:
> I got this working internally but unfortunately it really hurt our performance benchmarks on iOS. I’m guessing the delay of bundleFinishedLoading is the most likely issue - any ideas?

Have you fixed the performance?

janicduplessis:
@sahrens Can you give me more info about what the benchmark is and what it measures?

janicduplessis:
@shergin In this case maybe we should just make x,y always 0,0 like UIKit `bounds`. I don't think there values are very useful anyway.

sahrens:
@janicduplessis - sorry for the delay. The performance benchmark simply measures initial render time, that is, the time from the user tapping a button to the new content being visible on screen. The test in question is for the Marketplace tab in the Facebook app, so it's a brownfield cold start, and the "done" time is marked in native code via a special native component that listens for something like `onAttachedToWindow`. It would actually be really nice if we had a simple equivalent in Open Source...any idea how other folks measure performance?

janicduplessis:
@sahrens Thanks for the info, would it be possible to re-run the tests without the change to delay of bundleFinishedLoading. This would help isolate the issue.

I know @axe-fb did some work on perf measurement in OSS. Could be interesting to see if we can include that in RNTester and maybe integration tests.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -468,6 +475,85 @@ public void setAppProperties(@Nullable Bundle appProperties) {
...
+    }
+  }
+
+  private RootViewInsets getSafeAreaInsets() {
+    // Window insets are parts of the window that are covered by system views (status bar,
+    // navigation bar, notches). There are no apis the get these values for android < M so we
+    // do a best effort polyfill.
+    // TODO: Use the DisplayCutout api when we target sdk 28+
+    RootViewInsets windowInsets;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
greenfrvr: Why do you have this version check?`WindowInsets` is actually available from SDK level 20.

LaszloDev:
Any updates on this?

janicduplessis:
Building something similar outside of core for now, gonna loop back if it makes sense to include at some point.

https://github.com/th3rdwave/react-native-safe-area-context

PR #21010, opened 2018-09-07
If source.method is "post", use HTTP POST method

dignite:
The React Native Web View takes a method option to determine if HTTP GET or HTTP POST should be used.
For iPhone this is not case sensitive, meaning that `method: "post"` is allowed.
For Android "post" is not understood and therefore the request will be using HTTP GET.

I suggest we ignore case for the method, or throw an Exception.

Test Plan:
----------
Given that I add a `<WebView />` with `source.method` as lowercase `"post"` and source.uri as `http://httpbin.org/post`
When I view the WebView
Then the web view should show post data
It should not say "Method not allowed"

Release Notes:
--------------
 [ANDROID] [BUGFIX] [WebView] - Use postUrl if method is "post"


dulmandakh:
@facebook-github-bot shipit

PR #21015, opened 2018-09-07
Change the number of days to kill stales from 3 to 7

kelset:
Basically, saw this commit this morning and had a chance to discuss it with @matthargett and @AndrewJack and overall I agree with their criticism about 3 days being too little time for people to react to it.


hramos:
Thanks for the PR! 30 days was too long of a time period, and I agree that 3 days turned out to be too aggressive.

PR #21027, opened 2018-09-09
Add desktop platform-related stylesheet props

empyrical:
This PR adds the ReactPropTypes and Flow definitions for two style properties that are not useful for mobile platforms, but will be useful for desktop implementations:

 * `cursor` - for setting what the cursor's style will be when hovering over a `<View />`
 * `appRegion` - to make `View`s into draggable areas of a window, and also for marking children of said views non-draggable (like buttons). This is named after the `--webkit-app-region` property seen in [Chrome apps](https://developer.chrome.com/apps/app_window) and [Electron apps](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#draggable-region)

These could also be utilized by React Native Dom and React Native Web. (mapping `appRegion` back to `--webkit-app-region`)

Test Plan:
----------
In a test project I was able to use `appRegion` and `cursor` in a `StyleSheet` and have `flow check` pass.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/StyleSheet/DesktopPropTypes.js] - Added Desktop Prop Types
[GENERAL] [ENHANCEMENT] [Libraries/Components/View/ViewStylePropTypes.js] - Added Desktop Prop Types
[GENERAL] [ENHANCEMENT] [Libraries/StyleSheet/StyleSheetTypes.js] - Added Desktop style flow types
[GENERAL] [ENHANCEMENT] [Libraries/StyleSheet/StyleSheet.js] - Added Desktop style flow types

### DIFF
--- a/Libraries/Components/View/ViewStylePropTypes.js
+++ b/Libraries/Components/View/ViewStylePropTypes.js
@@ -15,6 +15,7 @@ const LayoutPropTypes = require('LayoutPropTypes');
 const ReactPropTypes = require('prop-types');
 const ShadowPropTypesIOS = require('ShadowPropTypesIOS');
 const TransformPropTypes = require('TransformPropTypes');
+const DesktopPropTypes = require('DesktopPropTypes');
 
 /**
  * Warning: Some of these properties may not be supported in all releases.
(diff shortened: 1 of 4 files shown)


empyrical:
Looks like `test_node10` failed due to a network hiccup:

```
yarn install v1.9.4
[1/5] Validating package.json...
[2/5] Resolving packages...
error Received malformed response from registry for undefined. The registry may be down.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Exited with code 1
```

But other than that, seemingly no regressions.

PR #21031, opened 2018-09-10
Add the new RCTWKWebView* files to build target on iOS

lwinkler-mathspace:
These files support backing a `WebView` component with the WebKit-based
`WKWebView` on iOS.  These files need to be included in the Xcode
build target so that they are compiled, linked and ultimately available
to the UIManager at runtime, so that the correct views can be instantiated
when a `WebView` is passed the prop `useWebKit={true}`.

Test Plan:
----------

Verify code not working with minimal test: 

- initialise new project
- upgrade to `0.57.0-rc.3`
- display traditional webview with `<WebView source={{}} />`; verified working
- try to use `WKWebView` with `<WebView useWebKit={true} source={{}} />`

fails with RedBox:

```Invariant Violation: requiredNativeComponent: "RCTWKWebView" was not found in the UIManager```

Repeat this test after adding the necessary files to the build target as in this PR; able to display `WKWebView`-backed `WebView` component.

Release Notes:
--------------

[IOS] [BUGFIX] [WebView] - Make sure RCTWKWebView* files are included in the Xcode build target


lwinkler-mathspace:
I feel like I must be in error, since this seems like a strange thing to be missing from RC4.  ie. it would have been noticed by now.  But I wonder if maybe the files are being included internally at Facebook by the Buck build system?  And therefore it's gone unnoticed that they're missing from the Xcode build?

hramos:
Facebook does not use the WebKit-backed version of WebView that is on master right now, so I'm not surprised this was not caught.

PR #20853, opened 2018-08-25
Update link command for Android project

matt-block:
Motivation:
--------------
PR #20767 bumped the version of the Android Gradle Plugin to v3 which uses the newer Gradle dependency configurations `implementation` and `api` which make `compile` obsolete.

While the PR updated the template Gradle configuration, it did not cover the `link` command which will still link native modules using `compile` resulting in a warning message beeing displayed during an app build.

Since `compile` will be eventually removed by Gradle, this commit updates the `link` command to attach native modules using `implementation`.

Test Plan:
----------
Ran the tests on my branch:

screenshot from 2018-08-26 12-37-28

**Note:** *test_android*, *test_detox_end_to_end* and *test_objc* seem to fail for other reasons discussed in https://github.com/facebook/react-native/pull/20854.

then,

1. Initialized a new RN project using the most recent candidate which is necessary for the project to understand the new `implementation` dependency configuration keyword
```
react-native init demoProject --version="0.57.0-rc.3"
```
2. Added a RN native module that targets Android as a dependency
```
yarn add react-native-custom-tabs
```
3. Copied the `local-cli/link` folder from my branch into `demoProject/node_modules/react-native/local-cli`
4. Followed the [additional step](https://github.com/droibit/react-native-custom-tabs#installation) in the module guide (added a repository).
5. Linked the added module using the local modified react-native
```
yarn run react-native link react-native-custom-tabs
```
screenshot from 2018-08-26 12-47-58

6. Checked that `android/app/build.gradle` has been patched correctly with `implementation`.

screenshot from 2018-08-26 12-53-41

7. Checked that `android/settings.gradle` and `MainApplication.java` have been patched as usual.
8. Built and installed successfully on a physical device without the deprecation warning (ignore the ones in the screenshot as they are related to the custom-tabs module as itself is not using the new configurations).
```
yarn run react-native run-android
```
screenshot from 2018-08-26 13-13-18

9. Unlinked and checked that is was successful.

Related PRs:
--------------
- https://github.com/facebook/react-native/pull/20767.

Release Notes:
--------------
[CLI] [ENHANCEMENT] [local-cli/link] - Updated `link` to link Android modules as `implementation` dependencies.

PR #21043, opened 2018-09-10
Fix regression in Android unlink command

matt-block:
Motivation:
--------------
Commit 4dfdec9b28074b311ffd75084d9d2a3cbd41a800 introduced a regression in the `unlink` command for Android projects: native modules previously linked with the now-deprecated `compile` configuration will not be detected anymore and by extension will not be unlinked.

`unlink` should be aware of both new and deprecated options to ensure functionality also in existing projects that previously linked modules with the old convention.

This PR fixes the regression by:
1. `makeDeprecatedNativeModule` is added so to have the deprecated patch line that could be present in the Gradle build file.
2. updating the Android `isInstalled` to check, for a given package name, for both `compile` (using the function introduced in step 1) and `implementation` dependency records in `build.gradle`.
3. Android `unregisterNativeModule` will proceed to revoke both possible patch lines from `build.gradle`.

This PR together with commit 4dfdec9b28074b311ffd75084d9d2a3cbd41a800 provide the following behaviour:

- `link` will continue to work as intended using `implementation`. This works without issues even with older packages that still use `compile` internally for their own dependencies.
- `link` will detect previously linked packages with `compile` and will not proceed with a new linkage. Instead, it will log that `Platform 'android' module X is already linked`.
- `unlink` will detect dependencies with both `compile` and `implementation` and proceed to unlink as expected.

Test Plan:
----------
- Ran tests locally on my branch, all related tests passed.
- Manually tried the changes on my existing projects that still use `compile` and the described behaviour listed above was matched.

Related PRs:
--------------
- #20853

Release Notes:
--------------
[CLI] [BUGFIX] [local-cli/link] - Fixed regression in Android `unlink`.

### DIFF
--- a/local-cli/link/__fixtures__/android/patchedBuild.gradle
+++ b/local-cli/link/__fixtures__/android/patchedBuild.gradle
@@ -3,6 +3,7 @@ dependencies {
     implementation(project(':test2')) {
       exclude(group: 'org.unwanted', module: 'test10')
     }
+    compile project(':testDeprecated')
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:27.1.1"
     implementation "com.facebook.react:react-native:+"
--- a/local-cli/link/__tests__/android/isInstalled.spec.js
+++ b/local-cli/link/__tests__/android/isInstalled.spec.js
@@ -26,6 +26,10 @@ describe('android::isInstalled', () => {
     expect(isInstalled(projectConfig, 'test2')).toBeTruthy();
   });
 
+  it('should return true when project is already in build.gradle using compile', () => {
+    expect(isInstalled(projectConfig, 'testDeprecated')).toBeTruthy();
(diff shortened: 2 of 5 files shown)


matt-block:
@Kureev @grabbou Let me know your thoughts.

matt-block:
@grabbou since [your proposal](https://github.com/react-native-community/discussions-and-proposals/pull/13) passed (congrats 🎉), shall we postpone this until we separate the CLI or proceed as always?

cpojer:

grabbou:
Hey @matt-block! Thanks for submitting this PR! Apologies for not getting back, but the whole CLI extraction process got me super busy with other notifications.

We ended up fixing this bug in the CLI with this PR: https://github.com/react-native-community/react-native-cli/pull/81 which takes a slightly different approach compared to your take. Would love to hear your opinion about that take - please feel free to comment on the PR or message me!

PR #21069, opened 2018-09-12
Add oneTimeCode to textContentType typings

BlakeSimpson:
iOS 12 supports a new `textContentType` attribute with the value `oneTimeCode` for extracting security codes from single-factor SMS login flows.

https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_a_text_input_view

This commit adds the Flow typings and Prop Types.

As far as I can see no changes need to be made to the iOS code for this change. 

If this change is accepted I would also make a follow-up PR for the documentation.

Release Notes:
--------------

 [IOS] [ENHANCEMENT] [TextInput] - Added `oneTimeCode` typings and prop types to the `textContentType` prop of `TextInput`.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -155,6 +155,7 @@ type IOSProps = $ReadOnly<{|
     | 'telephoneNumber'
     | 'username'
     | 'password'
+    | 'oneTimeCode'
   ),
   scrollEnabled?: ?boolean,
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for contributing this PR but I'm landing #21079 in favor of this since it adds another one as well.

BlakeSimpson:
That makes sense, thanks for the response.

PR #21101, opened 2018-09-13
Add support for resource-id

### LINKED ISSUE #9777: Add support for resource-id

lightboys22:
Track react tag using android.view.View.tag. Allow developers to set resource id by setting component.testID

Fixes #9777

## Test Plan

First, you need to declare the resource id by creating /_your_android_studio_folder_/res/values/ids.xml

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
<item type="id" name="my_test_id"/>
</resources>
```

Second, set the testID of your component to a resource name.

```
<Button
onPress={onButtonPress}
title="Press Me"
testID='test_id'
/>
```

Third, run the app and observe the resource id using UIAutomator viewer (/Android/sdk/tools/bin/uiautomatorviewer)

## Changelog

[ANDROID] [Fixes] - BaseViewManager: Allow developers to set resource id by setting component.testID

### DIFF
--- a/ReactAndroid/src/androidTest/java/com/facebook/react/testing/BUCK
+++ b/ReactAndroid/src/androidTest/java/com/facebook/react/testing/BUCK
@@ -35,6 +35,7 @@ rn_android_library(
         react_native_target("java/com/facebook/react/modules/debug:interfaces"),
         react_native_target("java/com/facebook/react/shell:shell"),
         react_native_target("java/com/facebook/react/uimanager:uimanager"),
+        react_native_target("java/com/facebook/react/uimanager/util:util"),
         react_native_target("res:uimanager"),
     ],
 )
--- a/ReactAndroid/src/androidTest/java/com/facebook/react/testing/ReactAppTestActivity.java
+++ b/ReactAndroid/src/androidTest/java/com/facebook/react/testing/ReactAppTestActivity.java
@@ -41,7 +41,9 @@ import com.facebook.react.uimanager.UIManagerModule;
 import com.facebook.react.uimanager.ViewManager;
 import com.facebook.react.uimanager.ViewManagerRegistry;
 import com.facebook.react.uimanager.events.EventDispatcher;
+import com.facebook.react.uimanager.util.ReactFindViewUtil;
 import java.util.Arrays;
+import com.facebook.react.uimanager.util.ReactFindViewUtil;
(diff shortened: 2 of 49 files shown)


craigrbruce:
@lightboys22 This is great work! I would suggest renaming `testId` as we have other applications for this facility beyond testing, namely gaining control of navigation (using a d-pad for example).

Perhaps a better name would be `resource-id`?

hramos:
Is this a duplicate of https://github.com/facebook/react-native/pull/18527, which you also opened?

cpojer:
@lightboys22 could you rebase this commit on master? Let's figure out how to get this merged!

lightboys22:
@cpojer Hi Christopher,
You honored me with your presence. My team bypassed this issue by using Espresso. 

Although Espresso only allows us to find UI elements by it content-description, Espresso is good enough for us.

harking:
@cpojer We've recently run into this again for some of our projects. Are you still open to including this fix?

lightboys22:
@harking I am too lazy to continue to work on this pull request. My previous team was using Espresso, which allow you to find elements based on the “test-is” view tag.

PR #21147, opened 2018-09-17
README: correct missing reference for contributing section

vcapretz:

Test Plan:
----------

Release Notes:
--------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See the official [React Native website](https://facebook.github.io/react-native/
 - [Getting Started](#getting-started)
 - [Documentation](#documentation)
 - [Upgrading](#upgrading)
-- [Contributing](#contributing)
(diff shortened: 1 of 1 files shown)


hramos:
There's already a reference to the CONTRIBUTING file right before the line you added.

vcapretz:
right, my bad!

but then on line 11 we would be missing a Contributing section for the anchor tag right? we can also point line 11 directly to the contributing file instead, what are your thoughts?

hramos:
We can link from line 11. About the anchor tag, is there something specific you had in mind? Is there a convention to have a Contributing anchor on every readme?

vcapretz:
@hramos not actually! it's just that currently if you click the Contributing link on line 11 it doesn't lead you anywhere, but I was not sure on what would be the best fix for that!

PR #21148, opened 2018-09-17
ios-simulator: change default iphone version

vcapretz:
this is just a suggestion, but as the iPhone 6 is discontinued by Apple
and the new iPhone line will match the X version, it could be good to update the default simulator.

it could also be the iPhone 8, feel free to suggest other changes!

Test Plan:
----------

Release Notes:
--------------


hramos:
Thanks for the PR. Let's go with iPhone X for now, as it's supported by both Xcode 9.4 and Xcode 10.

Once we drop support for Xcode 9, we can use iPhone XS or iPhone XR as the default.

PR #21157, opened 2018-09-17
[NativeAnimated] Support the `Slow Animations` option of the iOS simulator

janicduplessis:
RN animations currently ignore the `Slow Animations` option on the iOS simulator because we don't use UIKit animations directly. This uses a private api to get the slow coefficient and use it in the native animated driver. We only compile the private api code on simulator so this won't cause issues for app store approval. One possible issue is that the api changes in new iOS versions but I think it's reasonable to do this.

Note that this won't work with JS driven animations, we could expose the slow coefficient as a constant and use that in JS but I decided not to implement it.

Test Plan:
----------
Tested that animations are properly slowed like native UIKit ones in RNTester native animated example. 

Release Notes:
--------------
[IOS] [ENHANCEMENT] [NativeAnimated] - Support the `Slow Animations` option of the iOS simulator

### DIFF
--- a/Libraries/NativeAnimation/RCTAnimationUtils.m
+++ b/Libraries/NativeAnimation/RCTAnimationUtils.m
@@ -93,3 +93,17 @@ CGFloat RCTDegreesToRadians(CGFloat degrees)
 {
   return degrees / 180.0 * M_PI;
 }
+
+#if TARGET_IPHONE_SIMULATOR
+// Based on https://stackoverflow.com/a/13307674
+float UIAnimationDragCoefficient(void);
+#endif
+
+CGFloat RCTAnimationDragCoefficient()
+{
+#if TARGET_IPHONE_SIMULATOR
+  return (CGFloat)UIAnimationDragCoefficient();
+#else
+  return 1.0;
+#endif
+}
--- a/Libraries/NativeAnimation/Drivers/RCTDecayAnimation.m
+++ b/Libraries/NativeAnimation/Drivers/RCTDecayAnimation.m
@@ -10,6 +10,7 @@
 #import <UIKit/UIKit.h>
 #import <React/RCTConvert.h>
 
+#import "RCTAnimationUtils.h"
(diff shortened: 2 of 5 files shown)


janicduplessis:
cc @kmagiera 

Could be nice to integrate this in reanimated too.

TheSavior:
Including a video in your test plan would be cool

sahrens:
@janicduplessis: we're seeing animated test failures after merging this - any idea why that could be? e.g.


RCTNativeAnimatedNodesManagerTests/testSpringAnimationLoop:
Failure: ((didComeToRest) is true) failedRNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:484

RCTNativeAnimatedNodesManagerTests/testAnimationCallbackFinish:
Failure: ((endCallbackCalls) equal to (1)) failed: ("0") is not equal to ("1")
RNTester/RNTesterUnitTests/RCTNativeAnimatedNodesManagerTests.m:514

Libraries/NativeAnimation/RCTAnimationUtils.m
```diff
@@ -93,3 +93,17 @@ CGFloat RCTDegreesToRadians(CGFloat degrees)
...
+
+#if TARGET_IPHONE_SIMULATOR
+// Based on https://stackoverflow.com/a/13307674
+float UIAnimationDragCoefficient(void);
+#endif
+
+CGFloat RCTAnimationDragCoefficient()
+{
+#if TARGET_IPHONE_SIMULATOR
+  return (CGFloat)UIAnimationDragCoefficient();
```
sahrens: Based on this POP code, maybe we need to guard against 0 for some reason?

https://github.com/facebook/pop/blob/master/pop/POPAnimationExtras.mm#L37

janicduplessis: Oh that's strange. Thanks for investigating this!

PR #21204, opened 2018-09-19
Add `onScrollToTop` to ScrollView for iOS

jamesreggio:
This PR exposes the `onScrollToTop` event on iOS using the same event-forwarding infrastructure as other ScrollView events. (As such, its `nativeEvent` object reflects the same fields as other ScrollView events.)

Motivation:
----------

If your app is only interested in knowing the position of a ScrollView after a scroll has completed, it can use `onScrollEndDrag` and `onMomentumScrollEnd` to inspect the `contentOffset` after a drag-initiated scroll has finished. (This is much less expensive than observing the `onScroll` event if you only want to know the end position.) However, neither of these `End` events fire if the ScrollView is scrolled to the top by tapping the status bar.

By exposing `onScrollToTop`, it is now possible for an app to cheaply know when such a scroll has completed.

Test Plan:
----------

We've been using this patch in our production app without issue since December 2017.

I've recorded a video that shows the various ScrollView events firing, too. Notice that `onScroll(Begin|End)Drag` and `onMomentumScrollEnd` do not fire when scrolling to top (thus illustrating the value added by exposing `onScrollToTop`).

2018-09-19-react-native-ios-onscrolltotop

Release Notes:
--------------

[IOS] [ENHANCEMENT] [ScrollView] - Add `onScrollToTop` event handler to ScrollView


jamesreggio:
As is typical, the `prettier` bot is complaining about a `.` character that doesn't exist in my diff. If there's something actionable, please advise.

Also, it's not clear why the release notes bot didn't like my release notes... maybe it was the missing hyphen? Either way, I'm 99% certain that the release notes are good.

janicduplessis:
@facebook-github-bot shipit

PR #21207, opened 2018-09-19
Mark old OkHttp targets as deprecated

hramos:
Differential Revision: D9875254


dulmandakh:
why deprecate okhttp3?

PR #21208, opened 2018-09-19
Fix artifacting on RN-drawn borders with asymmetric radii

jamesreggio:
This PR fixes an obscure rendering bug on iOS for borders with asymmetric radii. It appears to be a problem with the custom drawing that React Native performs when it cannot use native UIKit/CoreAnimation border drawing.

Test Plan:
----------

We've been using this patch in our production app without issue since July 2018. I've seen this issue on the latest version of iOS 10, as well as iOS 11.2 (though it seems to not be an issue on iOS 11.4 any longer).

I've included before/after screenshots below, which also show the CSS needed to cause the artifacting.

react-native-border-before

react-native-border-after

Release Notes:
--------------

[IOS] [BUGFIX] [BorderDrawing] - Fixed visual artifacts on borders with asymmetric radii

### DIFF
--- a/React/Views/RCTBorderDrawing.m
+++ b/React/Views/RCTBorderDrawing.m
@@ -217,11 +217,19 @@ static UIImage *RCTGetSolidBorderImage(RCTCornerRadii cornerRadii,
   (borderInsets.top + cornerInsets.topRight.height +
    borderInsets.bottom + cornerInsets.bottomLeft.height <= viewSize.height);
 
-  const UIEdgeInsets edgeInsets = (UIEdgeInsets){
+  UIEdgeInsets edgeInsets = (UIEdgeInsets){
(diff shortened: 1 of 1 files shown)


jamesreggio:
Ah, it looks like this PR addresses the broadly-felt issue in #11897.

There's another PR from @mattijsf (#19451) that also attempts to fix this, though test cases remain where it doesn't seem resolved.

@mattijsf — can you either share code on the still-broken middle example from your PR, or try my solution here and see if it resolves all of your test cases?

jamesreggio:
Hi @hramos — I haven't heard anything from @mattijsf on this or the other PR. Given the age of his original PR, I think he may have moved on.

My suggestion would be to evaluate and merge my fix on its merits, and if the chorus of users in Issue #11897 continue to experience problems after merging my fix, we can revive the other PR.

PR #21230, opened 2018-09-20
Merged PR 76925: garbage collection is called in case of memory warning

dryganets:
JS garbage collection is called in case of memory warning for both iOS and Android.

Test Plan:
----------
Make sure that app calls garbage collector in case memory needs to be reclaimed.

[GENERAL] [ENHANCEMENT] [JSEngine] - OSS version calls the garbage collector in case memory warning occurs.

### DIFF
--- a/React/CxxBridge/RCTCxxBridge.mm
+++ b/React/CxxBridge/RCTCxxBridge.mm
@@ -200,6 +200,15 @@ struct RCTInstanceCallback : public InstanceCallback {
   return _reactInstance ? _reactInstance->isInspectable() : NO;
 }
 
+- (void)memWarning
+{
+    auto instance = _reactInstance.get();
+    if (instance) {
+        // there is no deferentiation in iOS between memory warning levels
+        instance->handleMemoryPressure(CRITICAL);
+    }
+}
+
 - (instancetype)initWithParentBridge:(RCTBridge *)bridge
 {
   RCTAssertParam(bridge);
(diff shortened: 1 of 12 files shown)


hramos:
Importing to get it onto our review queue

hramos:
This can't be merged, as it is, just yet. It looks like we'll need some manual work to land this from someone more familiar with Facebook's custom JSC.

dryganets:
@hramos, thank you for trying to land it!

I think the problem in the re-defenition of the:

```
enum JSMemoryPressure {
  UI_HIDDEN,
  MODERATE,
  CRITICAL,
};
```

I defined it in the Platform.h to make code more clean. Now we are using int only in the place where we have to (jni integration).
It is defined inside of facebook::react namespace so depending on your defenition error will be "JSMemoryPressure redefenition" if you are using the same namespace or "ambiguous variable ..." otherwise.

 **You could delete your own defenition and include Platform.h to your memory pressure implementation instead. It should fix the problem.**

davedets:
I think before landing this that we ought to make sure this doesn't have the potential for making bad situations worse. I can't find any reliable detailed information on memory warnings. For example, is there any limit on the rate at which they can occur? Let's say that the process threshold that triggers the warning is T. With just the live data in the JS heap, total memory is below T; with another 10% of the heap occupied, total memory is above T. We're at 10% above, we do a full GC, and we're below; good. Then a bit of JS is executed, promotions happen, we get back to 10% above the live data steady state, and above T. We get another warning, another full GC. Rinse, repeat: the effective heap size is 1.1x live data here, and we might be doing a lot of full GCs, and go very slowly. I present this scenario to argue that this sort of slow, lingering death might be a *worse* outcome than just having the whole process killed by the low memory killer.

So I guess I'd argue that if we're doing this, we should have some limit on the rate at which full GCs invoked this way happen, to avoid these "lingering death" scenarios.

dryganets:
Thanks for the heads up about ongoing work. In case of Skype calling GC and cleaning the in-memory caches helped us to keep application longer in the background.

I agree that it might be not the best possible solution, but it proved to be helpful as we reached a glass ceiling with startup time optimizations and in most of the scenarios it was a better user experience to keep the app alive in the background in dorm state rather than restarting it again. It is beneficial in terms of both battery usage as well as we don't have to start JS engine and reload the JS on every new push notification if we app stays alive in the background.

That's why I thought it might be beneficial for the community to have some solution in the public react-native version.

Would be great if your new implementation will be available in the public version @davedets!

PR #21231, opened 2018-09-20
OkHttp is more strict than other http libraries.

dryganets:
It crashes with IllegalStateException in case you pass a wrong URL.
It crashes if it meets unexpected symbols in the header name and value,
while standard says it is not recommended to use those symbols not that
they are prohibited.

The headers handing is a special use case as a client might have an auth token in the header. In this case, we want to get 401 error response
from the server to find out that token is wrong. In case of the onerror
client will continue to retry with an existing token.

Test Plan:
----------
Verify that android app receives an error instead of crashing in case of following network request:
"contacts/v2/users/"
Verify the case with passing the headers with

Release Notes:
--------------
[ANDROID] [BUGFIX] [Networking] - Passing invalid URL not crashes the app instead onerror callback of HttpClient is called. Invalid symbols are stripped from the headers to allow HTTP query to fail with 401 error code in case of the broken token.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/HeaderUtil.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/HeaderUtil.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * <p>This source code is licensed under the MIT license found in the LICENSE file in the root
+ * directory of this source tree.
+ */
+package com.facebook.react.modules.network;
+
+/**
+ *
+ * The class purpose is to weaken too strict OkHttp restriction on http headers.
+ * See: https://github.com/square/okhttp/issues/2016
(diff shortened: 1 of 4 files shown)


dryganets:
This is a workaround for the problem.

You can refer to this okhttp issue:
https://github.com/square/okhttp/issues/2802
Square is not going to fix it.

The ugly part of this PR comes from the validation of the headers in okhttp.
Other http libraries are able to send HTTP request to the server. As token is corrupt server returns 401 and everything is handled correctly on js side as a result.
In case of rejecting the promise js code won't be able to handle the situation correctly.

dryganets:
@hramos, did it merged to the master?
looks like you've been trying to import it but PR is still open.

hramos:
I looked at this and it's unclear why it wasn't landed. It was approved back in October. I'm going to rebase and run it through our tests again, then land.

PR #21250, opened 2018-09-21
[CI] [npm] Deploy to npm on tagged commits only

hramos:
This PR applies some fixes I made to the 0.57-stable branch to ensure we only run on commits tagged as "v0.57.1", as an example. This also ensures we only deploy after all tests pass.

Test Plan:
----------
Tested on Circle CI

Release Notes:
--------------

[INTERNAL] [BUGFIX] [CircleCI] - Ensure releases are attempted only on tagged commits

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -677,12 +677,11 @@ workflows:
       # Only runs on vX.X.X tags if all tests are green
       - publish_npm_package:
           filters:
-            branches:
-              only:
-                - /.*-stable/
             tags:
               only: /v[0-9]+(\.[0-9]+)*(\-rc(\.[0-9]+)?)?/
           requires:
+            - analyze
+            - test_detox_end_to_end
             - test_javascript
             - test_objc
             - test_android
--- a/scripts/publish-npm.js
+++ b/scripts/publish-npm.js
@@ -13,26 +13,28 @@
  * This script publishes a new version of react-native to NPM.
  * It is supposed to run in CI environment, not on a developer's machine.
  *
- * To make it easier for developers it uses some logic to identify with which version to publish the package.
+ * To make it easier for developers it uses some logic to identify with which
+ * version to publish the package.
  *
  * To cut a branch (and release RC):
  * - Developer: `git checkout -b 0.XY-stable`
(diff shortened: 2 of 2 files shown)


scripts/publish-npm.js
```diff
@@ -43,23 +45,28 @@
...
- * If tag v0.XY.0-rcZ is present on the commit then publish to npm with version 0.XY.0-rcZ and tag next
+ * If tag v0.XY.0-rc.Z is present on the commit then publish to npm with version 0.XY.0-rc.Z and tag next
  * If tag v0.XY.Z is present on the commit then publish to npm with version 0.XY.Z and no tag (npm will consider it latest)
  */
 
 /*eslint-disable no-undef */
 require('shelljs/global');
 
-const buildBranch = process.env.CIRCLE_BRANCH;
+const buildTag = process.env.CIRCLE_TAG;
```
hramos: We are no longer running the `publish_npm_package` job on stable branches, but rather we look for the actual git tag (e.g. v0.57.1, see Circle CI config).

$CIRCLE_TAG is available when a Circle CI job is initiated due to a commit that matches the pattern established in `.circleci/config.yml`.

This envvar is empty when a job is initiated due to a commit to a branch, so we need to switch to using $CIRCLE_TAG instead.

.circleci/config.yml
```diff
@@ -677,12 +677,11 @@ workflows:
       # Only runs on vX.X.X tags if all tests are green
       - publish_npm_package:
           filters:
-            branches:
```
hramos: Prior to removing this block, we'd end up with two separate `publish_npm_package` jobs.

Removing it ensures we only run that job on commits that are tagged with a string that matches /v[0-9]+(\.[0-9]+)*(\-rc(\.[0-9]+)?)?/ (from the `tags:` block).

.circleci/config.yml
```diff
@@ -677,12 +677,11 @@ workflows:
...
-              only:
-                - /.*-stable/
             tags:
               only: /v[0-9]+(\.[0-9]+)*(\-rc(\.[0-9]+)?)?/
           requires:
+            - analyze
```
hramos: Circle CI is more stable these days, let's avoid pushing a release when tests are red.

PR #21278, opened 2018-09-23
Remove remaining PropTypes from and modernize Swipable* Classes

empyrical:
Part of: https://github.com/react-native-community/discussions-and-proposals/issues/29

This PR removes all remaining PropTypes from the classes in `Libraries/Experimental/SwipeableRow`, and converts `SwipeableRow` from an old-style `createReactClass` to a proper subclass of `React.Component`.

This needed an alternative to `TimerMixin`, so I created a special `<Timer />` component based off of it that can be used through a ref. So it can still use timers that will be automatically terminated when the component is unmounted.

Test Plan:
----------
`flow check` passes.

A new test for the `Timer` component based off of the `TimerMixin` test has been added to `Libraries/Components/Timer/__tests__/Timer-test.js`

Release Notes:
--------------

[GENERAL] [FEATURE] [Libraries/Components/Timer/Timer.js] - Added a new `Timer` component based off of `react-timer-mixin`
[GENERAL] [FEATURE] [Libraries/Components/Timer/__tests__/Timer-test.js] - Added a test for the `Timer` component
[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableFlatList.js] - Cleaned up and exported props
[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableListView.js] - Removed proptypes and moved prop documentation to the flow types
[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableQuickActionButton.js] - Moved props to a separate declaration and exported them
[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableQuickActions.js] - Converted proptypes to flow types
[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableRow.js] - Converted from `createReactClass` to a class component, and removed all proptypes
[GENERAL] [MINOR] [Libraries/Lists/ListView/ListView.js] - Exported props type so `SwipeableFlatList` can use them.
[GENERAL] [ENHANCEMENT] [RNTester/js/SwipeableListViewExample.js] - Finished ES6 modernization

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
@@ -12,28 +12,31 @@
 import type {Props as FlatListProps} from 'FlatList';
 import type {renderItemType} from 'VirtualizedList';
 
-const PropTypes = require('prop-types');
 const React = require('React');
 const SwipeableRow = require('SwipeableRow');
 const FlatList = require('FlatList');
 
-type SwipableListProps = {
+type Props = {
   /**
    * To alert the user that swiping is possible, the first row can bounce
    * on component mount.
    */
   bounceFirstRowOnMount: boolean,
-  // Maximum distance to open to after a swipe
+  /**
+   * Maximum distance to open to after a swipe
+   */
   maxSwipeDistance: number | (Object => number),
-  // Callback method to render the view that will be unveiled on swipe
+  /**
+   * Callback method to render the view that will be unveiled on swipe
+   */
   renderQuickActions: renderItemType,
 };
 
-type Props<ItemT> = SwipableListProps & FlatListProps<ItemT>;
+export type SwipableListProps<ItemT> = Props & FlatListProps<ItemT>;
 
-type State = {
+type State = {|
   openRowKey: ?string,
-};
+|};
 
 /**
  * A container component that renders multiple SwipeableRow's in a FlatList
@@ -52,29 +55,12 @@ type State = {
  * - More to come
  */
 
-class SwipeableFlatList<ItemT> extends React.Component<Props<ItemT>, State> {
-  props: Props<ItemT>;
-  state: State;
-
+class SwipeableFlatList<ItemT> extends React.Component<
+  SwipableListProps<ItemT>,
+  State,
+> {
   _flatListRef: ?FlatList<ItemT> = null;
   _shouldBounceFirstRowOnMount: boolean = false;
-
-  static propTypes = {
-    ...FlatList.propTypes,
-
-    /**
-     * To alert the user that swiping is possible, the first row can bounce
-     * on component mount.
-     */
-    bounceFirstRowOnMount: PropTypes.bool.isRequired,
-
-    // Maximum distance to open to after a swipe
-    maxSwipeDistance: PropTypes.oneOfType([PropTypes.number, PropTypes.func])
-      .isRequired,
-
-    // Callback method to render the view that will be unveiled on swipe
-    renderQuickActions: PropTypes.func.isRequired,
-  };
 
   static defaultProps = {
     ...FlatList.defaultProps,
@@ -82,7 +68,7 @@ class SwipeableFlatList<ItemT> extends React.Component<Props<ItemT>, State> {
     renderQuickActions: () => null,
   };
 
-  constructor(props: Props<ItemT>, context: any): void {
+  constructor(props: SwipableListProps<ItemT>, context: any): void {
     super(props, context);
     this.state = {
       openRowKey: null,
--- a/Libraries/Experimental/SwipeableRow/SwipeableListView.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableListView.js
@@ -11,26 +11,49 @@
 'use strict';
 
 const ListView = require('ListView');
-const PropTypes = require('prop-types');
 const React = require('React');
 const SwipeableListViewDataSource = require('SwipeableListViewDataSource');
 const SwipeableRow = require('SwipeableRow');
 
-type DefaultProps = {
-  bounceFirstRowOnMount: boolean,
-  renderQuickActions: Function,
-};
+import type {Props as ListViewProps} from 'ListView';
 
-type Props = {
+export type SwipeableListViewProps = $ReadOnly<{|
+  ...ListViewProps,
+
+  /**
+   * To alert the user that swiping is possible, the first row can bounce
+   * on component mount.
+   */
   bounceFirstRowOnMount: boolean,
+  /**
+   * Use `SwipeableListView.getNewDataSource()` to get a data source to use,
+   * then use it just like you would a normal ListView data source
+   */
   dataSource: SwipeableListViewDataSource,
+  /**
+   * Maximum distance to open to after a swipe
+   */
   maxSwipeDistance:
     | number
     | ((rowData: any, sectionID: string, rowID: string) => number),
   onScroll?: ?Function,
-  renderRow: Function,
-  renderQuickActions: Function,
-};
+  /**
+   * Callback method to render the swipeable view
+   */
+  renderRow: (
+    rowData: any,
+    sectionID: string,
+    rowID: string,
+  ) => React.Element<any>,
+  /**
+   * Callback method to render the view that will be unveiled on swipe
+   */
+  renderQuickActions: (
+    rowData: any,
+    sectionID: string,
+    rowID: string,
+  ) => React.Element<any>,
+|}>;
 
 type State = {
   dataSource: Object,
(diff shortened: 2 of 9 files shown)


TheSavior:
Awesome!

I believe @yungsters already had a solution for TimerMixin and many things that used to use TimerMixin have already been moved away. Can you decouple those changes from this PR?

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
```diff
@@ -12,28 +12,31 @@
 import type {Props as FlatListProps} from 'FlatList';
 import type {renderItemType} from 'VirtualizedList';
 
-const PropTypes = require('prop-types');
 const React = require('React');
 const SwipeableRow = require('SwipeableRow');
 const FlatList = require('FlatList');
 
-type SwipableListProps = {
+type Props = {
```
TheSavior: Can you make this $ReadOnly and exact as well?

```
type Props = $ReadOnly<{||}>;
```

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
```diff
@@ -12,28 +12,31 @@
...
   maxSwipeDistance: number | (Object => number),
-  // Callback method to render the view that will be unveiled on swipe
+  /**
+   * Callback method to render the view that will be unveiled on swipe
+   */
   renderQuickActions: renderItemType,
 };
 
-type Props<ItemT> = SwipableListProps & FlatListProps<ItemT>;
+export type SwipableListProps<ItemT> = Props & FlatListProps<ItemT>;
```
TheSavior: Does this need to be exported?

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
```diff
@@ -12,28 +12,31 @@
...
 
-type State = {
+type State = {|
```
TheSavior: 👍

Libraries/Experimental/SwipeableRow/SwipeableListView.js
```diff
@@ -11,26 +11,49 @@
...
-  renderQuickActions: Function,
-};
+  /**
+   * Callback method to render the swipeable view
+   */
+  renderRow: (
+    rowData: any,
+    sectionID: string,
+    rowID: string,
+  ) => React.Element<any>,
```
TheSavior: Should this be `React.Element` or `React.Node`? I'm not sure what this function can return. It probably is a React.Element. 👍

Libraries/Experimental/SwipeableRow/SwipeableQuickActions.js
```diff
@@ -10,11 +10,16 @@
...
 'use strict';
 
-const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
 const React = require('React');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
 
+import type {ViewStyleProp} from 'StyleSheet';
+
+export type SwipeableQuickActionsProps = $ReadOnly<{|
```
TheSavior: Does this need to be exported? We try not to export a bunch of things, only the things that are actually necessary or expected to be used by other files.

Libraries/Experimental/SwipeableRow/SwipeableQuickActions.js
```diff
@@ -25,11 +30,9 @@ const View = require('View');
...
-class SwipeableQuickActions extends React.Component<{style?: $FlowFixMe}> {
-  static propTypes = {
-    style: DeprecatedViewPropTypes.style,
-  };
-
+class SwipeableQuickActions extends React.Component<
+  SwipeableQuickActionsProps,
+> {
   render(): React.Node {
     // $FlowFixMe found when converting React.createClass to ES6
```
TheSavior: While typing this component, can you clean up these FlowFixMes as well, as best as possible. This one for example is because children isn't defined in the props: `children: React.Node`

Libraries/Experimental/SwipeableRow/SwipeableRow.js
```diff
@@ -57,83 +53,72 @@ const RIGHT_SWIPE_BOUNCE_BACK_DURATION = 300;
  */
 const RIGHT_SWIPE_THRESHOLD = 30 * SLOW_SPEED_SWIPE_FACTOR;
 
-type Props = $ReadOnly<{|
+export type SwipableRowProps = $ReadOnly<{|
   children?: ?React.Node,
   isOpen?: ?boolean,
-  maxSwipeDistance?: ?number,
+  maxSwipeDistance: number,
```
TheSavior: Why this change?

RNTester/js/SwipeableListViewExample.js
```diff
@@ -10,42 +10,35 @@
...
-  },
-
-  getInitialState: function() {
-    var ds = SwipeableListView.getNewDataSource();
-    return {
-      dataSource: ds.cloneWithRowsAndSections(...this._genDataSource({})),
-    };
-  },
+class SwipeableListViewSimpleExample extends React.Component<
+  $FlowFixMeProps,
```
TheSavior: Can we do better than this?

RNTester/js/SwipeableListViewExample.js
```diff
@@ -165,19 +153,26 @@ var THUMB_URLS = [
   require('./Thumbnails/superlike.png'),
   require('./Thumbnails/victory.png'),
 ];
-var LOREM_IPSUM =
-  'Lorem ipsum dolor sit amet, ius ad pertinax oportere accommodare, an vix civibus corrumpit referrentur. Te nam case ludus inciderint, te mea facilisi adipiscing. Sea id integre luptatum. In tota sale consequuntur nec. Erat ocurreret mei ei. Eu paulo sapientem vulputate est, vel an accusam intellegam interesset. Nam eu stet pericula reprimique, ea vim illud modus, putant invidunt reprehendunt ne qui.';
+
+const LOREM_IPSUM = [
```
TheSavior: Could also use template literal

TheSavior:
This is great. Thanks for cleaning up / updating these files! For the sections that might require more back and forth or are less obvious on the approach, splitting those changes out can let us merge the rest of it and mark them off the list as completed!

Libraries/Experimental/SwipeableRow/SwipeableQuickActions.js
empyrical: That makes sense! Reducing the public api surface area is a good thing.

Libraries/Experimental/SwipeableRow/SwipeableListView.js
empyrical: I had a look around the codebase and for callbacks like this it seems to use `Element`

Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
empyrical: Can reverse this.

empyrical:
For the `TimerMixin` thing, I will have to revert it back to a `createReactClass` for now. I tried looking around and seeing what was being used as an alternative to `TimerMixin` before I did that, but I couldn't see it.

empyrical:
I have made a copy of the current version of this branch so my Timer component is more easily referenced to once I update this PR to the feedback:

https://github.com/empyrical/react-native/tree/swipable-updating-bak/Libraries/Components/Timer

Libraries/Experimental/SwipeableRow/SwipeableRow.js
empyrical: After I converted it to a React class, it being nullable had a lot of type errors like this:

```
undefined [1] is not a number.

 [1]  59│   maxSwipeDistance?: ?number,
        :
     306│     const maxSwipeDistance = IS_RTL
     307│       ? -this.props.maxSwipeDistance
     308│       : this.props.maxSwipeDistance;
     309│     this._animateTo(-maxSwipeDistance, duration);
     310│   }
     311│
     312│   _animateToClosedPosition(duration: number = SWIPE_DURATION): void {
```

But it has a default prop for `maxSwipeDistance`, so I removed the nullability because [the flow docs](https://flow.org/en/docs/react/components/) mention this:

> Note: You don’t need to make foo nullable in your Props type. Flow will make sure that foo is optional if you have a default prop for foo.

When it's a `creatReactClass` again, I should revert this because that notice will no longer be relevant.

TheSavior:
Apparently @yungsters's approach was to just inline the timer:

```
this._timerID = setTimeout(…);
```

And then in componentWillUnmount:
```
if (this._timerID != null) {
  clearTimeout(this._timerID);
}
```

PR #21283, opened 2018-09-24
Remove global babelHelpers and regenerator

janicduplessis:
Follow up to https://github.com/facebook/metro/commit/8932a9caa8f9ffe82160f2da25342a0bde3bbb26

Babel helpers and regenerator runtime will be imported automatically from `@babel/runtime`. We still need to add the global regeneratorRuntime for jest tests since we disable babel-runtime currently.

Test Plan:
----------
- Run RN tester and make sure it works fine for screens using async functions.
- Inspect the generated bundle to make sure helpers and regenerator are imported from @babel/runtime
- Made sure the global babelHelpers and the old regenerator-runtime are not in the generated bundle.
- Run jest tests

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [polyfills] - Remove global babelHelpers and regenerator


rafeca:
Thanks @janicduplessis for the PR! I'm planning to release a new version of Metro later today (it's going to be v0.46.0 since the babel runtime change is breaking.

With this you're going to be able to test it locally and afterwards I can import it 😄

Libraries/Core/InitializeCore.js
```diff
@@ -86,18 +86,6 @@ ReactNativeVersionCheck.checkVersions();
...
 
-// Set up regenerator.
-polyfillGlobal('regeneratorRuntime', () => {
-  // The require just sets up the global, so make sure when we first
-  // invoke it the global does not exist
-  delete global.regeneratorRuntime;
-  /* $FlowFixMe(>=0.54.0 site=react_native_oss) This comment suppresses an
-   * error found when Flow v0.54 was deployed. To see the error delete this
-   * comment and run Flow. */
-  require('regenerator-runtime/runtime');
```
newyankeecodeshop: Can you also remove `regenerator-runtime` from the dependencies in `package.json`?

janicduplessis: We still need it for jest tests, I guess I could move it to dev deps

janicduplessis:
@rafeca I was able to tests it, the only change missing was to add @babel/runtime to RN deps.

janicduplessis:
@rafeca I also didn't update the yarn.lock file, for some reason I couldn't generate a clean set of changes to it. It might be out of date already or a different version of yarn was used to generate it.

janicduplessis:
Also CI seems to be failling because of metro package that isn't published properly

```
error Couldn't find any versions for "metro-cache" that matches "0.46.0"
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: Couldn't find any versions for "metro-cache" that matches "0.46.0"
    at MessageError.ExtendableBuiltin (/opt/yarn-v1.9.4/lib/cli.js:243:66)
    at new MessageError (/opt/yarn-v1.9.4/lib/cli.js:272:123)
    at Function.<anonymous> (/opt/yarn-v1.9.4/lib/cli.js:51264:13)
    at Generator.next (<anonymous>)
    at step (/opt/yarn-v1.9.4/lib/cli.js:92:30)
    at /opt/yarn-v1.9.4/lib/cli.js:103:13
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
Exited with code 1
```

PR #21289, opened 2018-09-24
Handling the presence of multiple devices for android logging

MrScorp:
When testing on multiple android devices, its very common for the user to want to see the device logs from a specific device. I had a hard time with this, eventually deciding to run the adb command directly. Wanted to provide some solution to the problem, so I haveupdated  logAndroid.js to present an option to the user to specify/select the device when multiple devices are available. A device name will be picked up from the list of connected devices, and this device name will be passed on to the adb argument array with a "-s" flag preceding it, and logcat will be run against that device.

Test Plan:
-----------
1. Run "react-native log-android" without any device/emulators connected. 
Expected Result: The logger should wait till a device is connected

2. Run "react-native log-android" with only one device/emulators connected. 
Expected Result: The logger should output logs from the connected device/emulator

3. Run "react-native log-android" with more than one device/emulators connected. 
Expected Result : The logger should present the user with a list of connected devices, and an option to select the device from which logs has to be fetched.

4. Run "react-native log-android" with more than one device/emulators connected and provide a wrong input.
Expected Result : The logger should inform the user about the invalid entry, and exit the program.

Release Notes:
--------------
[CLI] [FEATURE] [local-cli/logAndroid/logAndroid.js] - Added option to run the logger for a specific android device.

### DIFF
--- a/local-cli/logAndroid/logAndroid.js
+++ b/local-cli/logAndroid/logAndroid.js
@@ -4,21 +4,49 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @format
  */
 
 'use strict';
 
 const chalk = require('chalk');
 const child_process = require('child_process');
+const adb = require('../runAndroid/adb');
+const prompt = require('../generator/promptSync')();
 
 /**
  * Starts adb logcat
  */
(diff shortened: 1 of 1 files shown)


cpojer:

PR #21293, opened 2018-09-24
StatusBar: Remove PropTypes

empyrical:
Part of: https://github.com/react-native-community/discussions-and-proposals/issues/29

This PR removes the remaining PropTypes from `StatusBar` and moves its flowtypes to its own definition.

Test Plan:
----------
`flow check` passes for android and ios.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Components/StatusBar/StatusBar.js] - Removed PropTypes

### DIFF
--- a/Libraries/Components/StatusBar/StatusBar.js
+++ b/Libraries/Components/StatusBar/StatusBar.js
@@ -11,8 +11,6 @@
 'use strict';
 
 const React = require('React');
-const PropTypes = require('prop-types');
-const ColorPropType = require('ColorPropType');
 const Platform = require('Platform');
 
 const processColor = require('processColor');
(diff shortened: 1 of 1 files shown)


Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -55,9 +53,47 @@ export type StatusBarAnimation = $Enum<{
   slide: string,
 }>;
 
-type DefaultProps = {
-  animated: boolean,
-};
+type StatusBarProps = $ReadOnly<{|
```
TheSavior: I think I'd prefer if you just named these types `Props`. I believe that is consistent with all the other files

Libraries/Components/StatusBar/StatusBar.js
```diff
@@ -55,9 +53,47 @@ export type StatusBarAnimation = $Enum<{
...
+  /**
+   * If the transition between status bar property changes should be animated.
+   * Supported for backgroundColor, barStyle and hidden.
+   */
+  animated?: ?boolean,
+  /**
+   * The background color of the status bar.
+   * @platform android
+   */
+  backgroundColor?: ?string,
```
TheSavior: When there are a bunch of Props that are supported on different platforms, I like doing this pattern:

```
type AndroidProps = //
type IOSProps = //

type Props = $ReadOnly<{|
  ...IOSProps,
  ...AndroidProps,
  // shared props
|}>
```

Libraries/Components/StatusBar/StatusBar.js
empyrical: Renamed it

Libraries/Components/StatusBar/StatusBar.js
empyrical: And also re-organized these

PR #21298, opened 2018-09-24
SwipeableListView: Remove PropTypes

empyrical:
This PR removes the remaining proptypes from `SwipeableListView`, and cleans up its flow types a bit. Its RNTester example has also been cleaned up, and turned into an ES6 class.

`ListView`'s props have been exported so this can use it.

Test Plan:
----------
`flow check` passes.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableListView.js] - PropTypes have been removed
[GENERAL] [MINOR] [Libraries/Lists/ListView/ListView.js] - Flow proptypes have been exported
[GENERAL] [ENHANCEMENT] [RNTester/js/Shared/RNTesterTypes.js] - Created basic flow types for RNTester components 
[GENERAL] [ENHANCEMENT] [RNTester/js/SwipeableListViewExample.js] - Converted from `createReactClass` to ES6 class, and other general cleanup.

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableListView.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableListView.js
@@ -11,30 +11,53 @@
 'use strict';
 
 const ListView = require('ListView');
-const PropTypes = require('prop-types');
 const React = require('React');
 const SwipeableListViewDataSource = require('SwipeableListViewDataSource');
 const SwipeableRow = require('SwipeableRow');
 
-type DefaultProps = {
-  bounceFirstRowOnMount: boolean,
-  renderQuickActions: Function,
-};
+import type {Props as ListViewProps} from 'ListView';
 
-type Props = {
+type Props = $ReadOnly<{|
+  ...ListViewProps,
+
+  /**
+   * To alert the user that swiping is possible, the first row can bounce
+   * on component mount.
+   */
   bounceFirstRowOnMount: boolean,
+  /**
+   * Use `SwipeableListView.getNewDataSource()` to get a data source to use,
+   * then use it just like you would a normal ListView data source
+   */
   dataSource: SwipeableListViewDataSource,
+  /**
+   * Maximum distance to open to after a swipe
+   */
   maxSwipeDistance:
     | number
-    | ((rowData: any, sectionID: string, rowID: string) => number),
+    | ((rowData: Object, sectionID: string, rowID: string) => number),
   onScroll?: ?Function,
-  renderRow: Function,
-  renderQuickActions: Function,
-};
+  /**
+   * Callback method to render the swipeable view
+   */
+  renderRow: (
+    rowData: Object,
+    sectionID: string,
+    rowID: string,
+  ) => React.Element<any>,
+  /**
+   * Callback method to render the view that will be unveiled on swipe
+   */
+  renderQuickActions: (
+    rowData: Object,
+    sectionID: string,
+    rowID: string,
+  ) => React.Element<any>,
+|}>;
 
-type State = {
+type State = {|
   dataSource: Object,
-};
+|};
 
 /**
  * A container component that renders multiple SwipeableRow's in a ListView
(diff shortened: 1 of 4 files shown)


RNTester/js/SwipeableListViewExample.js
```diff
@@ -165,19 +159,26 @@ var THUMB_URLS = [
...
+const LOREM_IPSUM = [
+  'Lorem ipsum dolor sit amet, ius ad pertinax oportere accommodare, an vix ',
+  'civibus corrumpit referrentur. Te nam case ludus inciderint, te mea facilisi ',
+  'adipiscing. Sea id integre luptatum. In tota sale consequuntur nec. Erat ',
+  'ocurreret mei ei. Eu paulo sapientem vulputate est, vel an accusam ',
+  'intellegam interesset. Nam eu stet pericula reprimique, ea vim illud modus, ',
+  'putant invidunt reprehendunt ne qui.',
+].join('');
 
 /* eslint no-bitwise: 0 */
```
TheSavior: Can you change this too to:

```
/* eslint-disable no-bitwise */
// code
/* eslint-enable no-bitwise */
```

TheSavior:
Did you test that SwipeableListViewExample still works in RNTester? It is a pretty major refactor that testing it manually and not just relying on Flow would be great.

empyrical:
Will give it a try and get back when I do! Currently on XCode 10, so I will have to set up an Android simulator for myself.

RNTester/js/SwipeableListViewExample.js
```diff
@@ -100,45 +94,45 @@ var SwipeableListViewSimpleExample = createReactClass({
...
       </TouchableHighlight>
     );
-  },
+  };
 
-  _genDataSource: function(pressData: {[key: number]: boolean}): Array<any> {
-    var dataBlob = {};
-    var sectionIDs = ['Section 0'];
-    var rowIDs = [[]];
+  _genDataSource(pressData: {[key: number]: boolean}): Array<any> {
```
rsnara: The return type here might be overly generic, given that we know the types of each of the elements in the array. 

```typescript
const dataBlob: {[key: string]: { id: string, text: string }} = {}
const sectionIDs: Array<string> = ['Section 0'];
const rowIds: Array<Array<string>> = [[]]
```

I wonder if it's better to remove it. Flow is smart enough to infer the return type.

RNTester/js/SwipeableListViewExample.js
```diff
@@ -100,45 +94,45 @@ var SwipeableListViewSimpleExample = createReactClass({
...
 
-  _renderSeperator: function(
-    sectionID: number,
-    rowID: number,
+  _renderSeperator = (
+    sectionID: string,
+    rowID: string,
     adjacentRowHighlighted: boolean,
-  ) {
+  ) => {
```
rsnara: Since we're not using `this` inside the `_renderSeparator` body, it's probably fine for us to use a regular function here. 

**Also:** This is not necessary, but since we're not using any instance state of `SwipeableListViewSimpleExample` objects, we don't even have to keep this an instance method.

RNTester/js/SwipeableListViewExample.js
```diff
@@ -10,42 +10,41 @@
...
+import type {RNTesterProps} from 'RNTesterTypes';
 
-  getInitialState: function() {
-    var ds = SwipeableListView.getNewDataSource();
-    return {
-      dataSource: ds.cloneWithRowsAndSections(...this._genDataSource({})),
-    };
-  },
+type State = {|
+  dataSource: Object,
```
rsnara: I think we can be more specific here:

```typescript
type State = {|
  dataSource: SwipeableListViewDataSource
|};
```

Libraries/Experimental/SwipeableRow/SwipeableListView.js
```diff
@@ -11,30 +11,53 @@
...
-const PropTypes = require('prop-types');
 const React = require('React');
 const SwipeableListViewDataSource = require('SwipeableListViewDataSource');
 const SwipeableRow = require('SwipeableRow');
 
-type DefaultProps = {
-  bounceFirstRowOnMount: boolean,
-  renderQuickActions: Function,
-};
+import type {Props as ListViewProps} from 'ListView';
```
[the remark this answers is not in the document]
empyrical: Thanks, didn't know about that one. Much cleaner!

RNTester/js/SwipeableListViewExample.js
empyrical: Flow seemed fine with removing it

RNTester/js/SwipeableListViewExample.js
empyrical: Thinking a top level function in the module, or a static method?

rsnara:
Thanks for the corrections! I'll land this tomorrow when I get to work :)

empyrical:
@rsnara I was able to confirm that `SwipableListViewExample.js` works on Android RNTester, but I have XCode 10 on my machine and am not able to test it on iOS right now. Eli suggested I ask you to try it out on iOS and double check if it works right there.

RSNara:
@empyrical it works! Landing now.

PR #21314, opened 2018-09-25
Re-sourcing profile Xcode to update node version

rush1506:
Re-sourcing profile to update Xcode node version since Xcode use old node version and for some people (our team). This can cause the bundle to fail building main.jsbundle

Test Plan:
----------

**Problem**
XCode 10 using wrong node version (older) so it cannot build the main.jsbundle when release the app on IOS

**Test failed case before editing the code:**

Run command react native for release:
```
react-native run-ios --configuration Release
```
-> The bundle will not be generated, XCode throw in error

**Test success case after editing the code:**

Run command react native for release:
```
react-native run-ios --configuration Release
```
-> IOS app is upload to test flight or ipa file is generated

Release Notes:
--------------
[IOS][BUGFIX] [./scripts]  - Sourcing XCode node version again when release to help re-bundling mainjs file when X code generate react native app version `^0.57.x`

### DIFF
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -8,6 +8,9 @@
 # This script is supposed to be invoked as part of Xcode build process
 # and relies on environment variables (including PWD) set by Xcode
 
+#Re-sourcing profile to update Xcode node version since Xcode use old node version
+source /.bash_profile
+
 # Print commands before executing them (useful for troubleshooting)
 set -x
 DEST=$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH


rush1506:
> Hey, thanks for the PR! Can you fill in your test plan?

Sorry for the late reply. I am away for couple of weeks after the projects so I leave everything at home, here is the test plan text:

##Environment 
XCode: 10.0.0
React native: 0.57.0
React: 16.5
Node version: 8.9.4

##Problem
XCode 10 using wrong node version (older) so it cannot build the main.jsbundle when release the app on IOS

##Test Case

####Test failed case before editing the code:
Run command react native for release:
```
react-native run-ios --configuration Release
```
-> The bundle will not be generated, XCode throw in error

####Test success case after editing the code:
Run command react native for release:
```
react-native run-ios --configuration Release
```
-> IOS app is upload to test flight or ipa file is generated

hramos:
Thanks for adding it in!

I could use some help understanding why this fix is needed. Why is Xcode using a different version of Node? The fix appears to assume the user has set up their `.bash_profile` with a path to Node. Is this a reasonable assumption to make?

vovkasm:
Oh, no! Please don't source .bash_profile by default... I use zsh, and other shells exists in the world!

vovkasm:
Moreover, `/.bash_profile` simple does not exists in my system. Why in root of fs? May be it should be in `/etc/profile`, but I sure problem has other solutions...

rush1506:
> Moreover, `/.bash_profile` simple does not exists in my system. Why in root of fs? May be it should be in `/etc/profile`, but I sure problem has other solutions...

Sorry, it seems like I was not being considerate about this. Please reject my pull request

vovkasm:
@rush1506 Actually I'd recommend as a workaround for your issue, use [nvm](https://github.com/creationix/nvm) or [nodenv](https://github.com/nodenv/nodenv). Both are supported by `react-native-xcode.sh` script. (moreover, actually it can be source of you initial issue, because `react-native-xcode.sh` script calls these managers not only if they are installed locally, but even if they simple exists in homebrew paths, see: https://github.com/facebook/react-native/blob/0.57-stable/scripts/react-native-xcode.sh#L59)

PR #21326, opened 2018-09-26
package.json: Add flow check commands for iOS and Android

empyrical:
This PR adds these two scripts to `package.json`:

 * `flow-check-ios` - for running `flow check` on `.ios.js` files
 * `flow-check-android` - for running `flow check` on `.android.js` files

The Android command makes use of the new `--flowconfig-name` option added to Flow in `0.80.0`

Test Plan:
----------
Both commands have been ran and work as expected.

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [package.json] - Added flow check commands for iOS and Android

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,9 +185,12 @@ aliases:
     name: Lint code
     command: scripts/circleci/exec_swallow_error.sh yarn lint --format junit -o ~/react-native/reports/junit/eslint/results.xml
 
-  - &run-flow-checks
-    name: Check for errors in code using Flow
+  - &run-flow-checks-ios
+    name: Check for errors in code using Flow (iOS)
     command: yarn flow-check-ios
+
+  - &run-flow-checks-android
+    name: Check for errors in code using Flow (Android)
     command: yarn flow-check-android
 
   - &run-sanity-checks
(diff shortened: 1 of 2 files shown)


empyrical:
Added!

.circleci/config.yml
```diff
@@ -187,7 +187,8 @@ aliases:
 
   - &run-flow-checks
     name: Check for errors in code using Flow
-    command: yarn flow check
+    command: yarn flow-check-ios
+    command: yarn flow-check-android
```
TheSavior: Woah! We weren't checking this!? Scary!

TheSavior: I think the second command overwrites the first. Looking at the CI logs for this step in this PR show that it only ran flow on android: https://circleci.com/gh/facebook/react-native/53561?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

You'll probably need to change this from `run-flow-checks` to `run-flow-ios` and have another one for android.

empyrical: Good catch, fixed that up

TheSavior: That only runs checking with the default config which checks .ios.js files but not .android.js files.

empyrical:
I don't know a whole lot about the AppVeyor syntax yet - would just adding more commands after `npm test` work?

hramos:
Landing. The checks can be added to Appveyor's config on a separate PR.

PR #21338, opened 2018-09-26
Fix DatePickerAndroid error when passing "minDate: null"

### LINKED ISSUE #21253: [0.57] [Android] DatePickerAndroid doesn't open because of null minDate
## Environment

```bash
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
      Memory: 55.30 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.10.0 - /usr/local/bin/node
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react: ^16.5.1 => 16.5.1 
      react-native: ^0.57.0 => 0.57.0 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description
 
**Problem**: when no minDate is provided, the DatePickerAndroid won't open because of this error:
> Cannot read property 'getMonth' of null

For info, I use [react-native-modal-datetime-picker](https://github.com/mmazzarolo/react-native-modal-datetime-picker), itself using [DatePickerAndroid](https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js).

## Cause 

See the [_toMillis](https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js#L18-L24) function below.

```js
function _toMillis(options: Object, key: string) {
  const dateVal = options[key];
  // Is it a Date object?
  if (typeof dateVal === 'object' && typeof dateVal.getMonth === 'function') {
    options[key] = dateVal.getTime();
  }
}
```

` if (typeof dateVal === 'object' && typeof dateVal.getMonth === 'function') ` doesn't block when `dateVal` is `null`. Cause `typeof null` is `object` (wtf!?)

## Reproducible Demo

I can provide a reproducible demo, but I want first to have your opinion about the issue I described above. 

Thank you

MaxInMoon:
It avoids an error when passing an options object with minDate: null. (see #21253)
Inspired by isObjectLike from Loadash.

Fixes #21253 

@rsnara reviewer?

## Test Plan:

This [jsbin](https://jsbin.com/hiferipene/1/edit?js,console) allow to test the old and the new `_toMillis` function.

You can observe that `_toMillis` doesn't throw error (fixed) and that if you decomment the `_toMillis_before_pr` call, it will generate an error as before.

## Release Notes:

[ANDROID] [BUGFIX] [DatePickerAndroid] fix error when passing `minDate: null` in `options`

### DIFF
--- a/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js
+++ b/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js
@@ -18,7 +18,7 @@ const DatePickerModule = require('NativeModules').DatePickerAndroid;
 function _toMillis(options: Object, key: string) {
   const dateVal = options[key];
   // Is it a Date object?
-  if (typeof dateVal === 'object' && typeof dateVal.getMonth === 'function') {
+  if (dateVal != null && typeof dateVal == 'object' && typeof dateVal.getMonth === 'function') {
     options[key] = dateVal.getTime();
   }
 }


TheSavior:
Thanks for contributing! This file really looks like it needs to be flow typed. This change just feels like a small patch that might not completely handle all the cases of things not being typesafe.

Would you be willing to take a stab at flow typing this file to ensure we are properly handling all of the different edge cases of `options`?

MaxInMoon:
@TheSavior I'll do some flow-typing within the two next days as soon as I have a quiet moment.

Just two questions about `optionsMs` and `options` in the [function](https://github.com/facebook/react-native/blob/master/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js#L68) below:

```js

static async open(options: Object): Promise<Object> {
  const optionsMs = options;
  if (optionsMs) {
    _toMillis(options, 'date');
    _toMillis(options, 'minDate');
    _toMillis(options, 'maxDate');
  }
  return DatePickerModule.open(options);
}
```

1 - Why doing this: `const optionsMs = options;`?

2 - Why mutating the `options` object within `_toMillis()`? Would this not be a bad practice?

TheSavior:
The answer to both of your questions are "this is legacy code, I wouldn't want it written this way today". Good news! You can make it better! :D

Fortunately, if you flow type `.open()` then we shouldn't even need the `_toMillis` function.

RSNara:
Since this is a pretty simple change, I think we can merge this PR as is. 

But as @TheSavior said, it'd be better if this file were flow-typed. The error you caught at run-time could have easily been caught at compile-time. Would you like to launch another PR? Feel free to refactor as you see fit.

cpojer:
I agree with @TheSavior and @RSNara that this file should be flow typed instead of adding more runtime checks. Since there haven't been any updates in more than three months I'm gonna close this PR. I'm happy to reopen or land a new PR that adds Flow types to this module. Thank you for bringing this issue to our attention!

PR #21340, opened 2018-09-26
Rounded corner rendering fixed on Android N.

dryganets:
On Android N (API 24) rounded corners rendering has issues in case scale factor is set for the view.

Test Plan:
----------
Setup transform scale not equal 1.0 for view with rounded corners. On Android 7.0 view will be incorrectly positioned in case hardware optimization is turned on.

Release Notes:
--------------
[Android] [UI] [View] - Rounded corner rendering fixed on android 7.0.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
@@ -241,8 +241,7 @@ public class ReactViewGroup extends ViewGroup implements
     ReactViewBackgroundDrawable backgroundDrawable = getOrCreateReactViewBackground();
     backgroundDrawable.setRadius(borderRadius);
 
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
```diff
@@ -241,8 +241,7 @@ public void setBorderRadius(float borderRadius) {
     ReactViewBackgroundDrawable backgroundDrawable = getOrCreateReactViewBackground();
     backgroundDrawable.setRadius(borderRadius);
 
-    if (Build.VERSION_CODES.HONEYCOMB < Build.VERSION.SDK_INT
```
dryganets: You don't support targets below HONEYCOMB anyway. The only API level affected by the original check was 16.

Issue #21342, opened 2018-09-26
Removing propTypes from core components


TheSavior:
We have been using Flow for our components for a long while and have enjoyed much stricter type safety, enabling us to validate our code at compile time instead of with propTypes at runtime. We want to replace the remaining callsites of propTypes with Flow and deprecate the propType shapes exposed by React Native core. To do this, we need your help!

*This is step 1 of our multi step goal. Step 2 can be [found here](https://github.com/facebook/react-native/issues/21485).*

There are two classes of things that need to happen to complete this work. The first step is removing references from propTypes in our core components. The following are the individual files that need to be converted / updated. Each file should be converted in its own PR and each file is about the size of a good first issue to learn and get familiar with a part of the React Native codebase. If you’d like to convert one, pick one from the list, comment on this issue that you are interested in converting it, and follow the tips in [@empyrical's comment for help](https://github.com/facebook/react-native/issues/21342#issuecomment-424907040) successfully updating that file.

## Remove callsites from components:
- [x] Libraries/Components/Button.js -> Delete PropTypes
 - [x] MaskedViewIOS.ios.js -> Delete PropTypes
 - [x] Picker/Picker.js -> Delete PropTypes
 - [x] PickerAndroid.android.js -> Delete PropTypes
 - [x] StaticRenderer.js -> Delete PropTypes
 - [x] StatusBar/StatusBar.js -> Delete PropTypes
 - [x] TabBarIOS/TabBarIOS.ios.js -> Delete PropTypes
 - [x] TabBarIOS/TabBarItemIOS.ios.js -> Convert to Flow Types
 - [x] ViewPagerAndroid.android.js -> Delete PropTypes
 - [x] IncrementalPresenter.js -> Delete PropTypes, leave contextTypes
 - [x] SwipeableFlatList.js -> Delete PropTypes
 - [x] SwipeableListView.js -> Delete PropTypes
 - [x] SwipeableQuickActions.js -> DeletePropTypes (style changed to ViewStyleProp)
 - [x] ElementProperties.js -> Delete PropTypes
 - [x] InspectorPanel -> Delete PropTypes
 - [x] InspectorOverlay.js -> Delete PropTypes
 - [x] SnapshotViewIOS.ios.js -> Delete PropTypes
 - [x] LinkingExample.js -> Convert to Flow Types
 - [x] RNTesterBlock.js -> Delete PropTypes, strengthen Flow Types
 - [x] RNTesterButton.js -> Delete PropTypes
[...]

PR #21343, opened 2018-09-27
Remove PropTypes from IncrementalPresenter

julioxavierr:
Related to https://github.com/facebook/react-native/issues/21342


julioxavierr:
Just did the request changes, @empyrical. Thanks for the advise.

empyrical:
Looks like there's a flow failure here:

```
Cannot create `View` element because mixed [1] is incompatible with undefined [2] in the return value of property
`onLayout`.

   Libraries/Experimental/IncrementalPresenter.js:85:9
               v----
   85|         <View
   86|           children={this.props.children}
   87|           ref="view"
   88|           style={style}
   89|           onLayout={this.props.onLayout}
   90|         />
               -^

References:
   Libraries/Experimental/IncrementalPresenter.js:38:38
   38|   onLayout?: (event: LayoutEvent) => mixed,
                                            ^^^^^ [1]
   Libraries/Components/View/ViewPropTypes.js:31:39
   31|   onLayout?: ?(event: LayoutEvent) => void,
                                             ^^^^ [2]
```

This is because the `onLayout` type in `ViewPropTypes.js` needs to be fixed so it returns `mixed`, rather than `void`. Then your PR should pass tests! :+1:

TheSavior:
Thanks for contributing!

PR #21345, opened 2018-09-27
Remove PropTypes from InspectorOverlay

kdastan:
Related to #21342

### DIFF
--- a/Libraries/Inspector/InspectorOverlay.js
+++ b/Libraries/Inspector/InspectorOverlay.js
@@ -23,11 +23,13 @@ type EventLike = {
   nativeEvent: Object,
 };
 
+type Inspected = $ReadOnly<{|
+  frame?: Object,
+  style?: ViewStyleProp,
+|}>;
+
 type Props = $ReadOnly<{|
-  inspected?: ?{
-    frame?: ?Object,
-    style?: ?ViewStyleProp,
-  },
+  inspected?: Inspected,
   inspectedViewTag?: ?number,
   onTouchViewTag: (tag: number, frame: Object, pointerY: number) => mixed,
 |}>;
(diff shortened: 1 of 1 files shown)


kdastan:
@empyrical is it ok, if I leave frame as Object and style as any?

empyrical:
You should type `style` as `ViewStyleProp`, which you can import like this:

```js
import type {ViewStyleProp} from 'StyleSheet';
```

You should be able to leave the type of `frame` as `Object` for now. Filling this out this will require filling out the types of `ElementBox` too, which is out of the scope of this pull request.

kdastan:
@empyrical got it, thanks!

empyrical:
LGTM :+1:

Libraries/Inspector/InspectorOverlay.js
```diff
@@ -12,33 +12,27 @@
...
   nativeEvent: Object,
 };
 
-class InspectorOverlay extends React.Component<{
-  inspected?: {
-    frame?: Object,
-    style?: any,
+type Props = $ReadOnly<{|
+  inspected?: ?{
+    frame?: ?Object,
```
TheSavior: Let's make these nullable either. The goal of nullable props is so that a component can take optional props and pass them to another component which takes the optional prop.

kdastan: Should I make Props nullable?

```
type Props = ?$ReadOnly<{| ...
```

do not quite understand

TheSavior: I'm looking for:

```
inspected?: {
  frame?: Object 
  style?: ViewStyleProp
}
```

Although inspected should also use `$ReadOnly<{||}>`. So:

```
inspected?: $ReadOnly<{|
  frame?: Object 
  style?: ViewStyleProp
|}>
```

kdastan:
I created new Type, then used it as type of inspected 
```
type Inspected = $ReadOnly<{|
  frame?: Object,
  style?: ViewStyleProp,
|}>;
```

TheSavior:
Thanks for contributing!

PR #21346, opened 2018-09-27
Remove PropTypes from MaskedViewIOS.ios.js

nd-02110114:
related #21342 

This is a first PR for this repo. 
So, if there are any problem, please tell me 🙇 

TODO
* delete props types 
* apply read only interface

CheckList
 - [x] `yarn prettier`
 - [x] `yarn flow-check-ios`

### DIFF
--- a/Libraries/Components/MaskedView/MaskedViewIOS.ios.js
+++ b/Libraries/Components/MaskedView/MaskedViewIOS.ios.js
@@ -8,8 +8,6 @@
  * @flow
  */
 
-const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
-const PropTypes = require('prop-types');
 const React = require('React');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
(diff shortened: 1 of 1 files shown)


Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -20,7 +18,7 @@ import type {ViewProps} from 'ViewPropTypes';
 
 const RCTMaskedView = requireNativeComponent('RCTMaskedView');
 
-type Props = {
+type Props = $ReadOnly<{
```
empyrical: Should make it exact using `|`, like:

```js
type Props = $ReadOnly<{|
 //...
|}>;
```

Libraries/Components/MaskedView/MaskedViewIOS.ios.js
```diff
@@ -20,7 +18,7 @@ import type {ViewProps} from 'ViewPropTypes';
...
   ...ViewProps,
 
   children: any,
```
empyrical: `children` should be typed like this:

```js
children?: React.Node,
```

nd-02110114:
Thank you for comments!  I fixed by following your comments.

empyrical:
Thanks, LGTM! 👍

TheSavior:
Thanks for contributing!

PR #21348, opened 2018-09-27
StaticRenderer: Removed prop types

rmcp1g15:
related #21342 

 The `render` function, I was not able to specifically type since the props passed to it may vary. At the moment only `renderRow` function from ListView component is using StaticRenderer, and the type of the renderRow function is `Function`.
Let me know what your thoughts are on this. Thank you

Test Plan:
Ran against flow check for android and ios

Release Notes:
 [GENERAL] [ENHANCEMENT] [StaticRenderer.js] - Converted PropTypes to Flow

### DIFF
--- a/Libraries/Components/StaticRenderer.js
+++ b/Libraries/Components/StaticRenderer.js
@@ -12,20 +12,20 @@
 
 const React = require('React');
 
-type StaticRendererProps = $ReadOnly<{|
+type Props = $ReadOnly<{|
   /**
    * Indicates whether the render function needs to be called again
    */
   shouldUpdate: boolean,
   /**
-   * (props) => renderable
-   * A function that takes in certain props that are used to render a component
+   * () => renderable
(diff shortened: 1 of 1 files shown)


Libraries/Components/StaticRenderer.js
```diff
@@ -12,18 +12,20 @@
...
+type StaticRendererProps = $ReadOnly<{|
+  /**
+   * Indicates whether the render function needs to be called again
+   */
   shouldUpdate: boolean,
+  /**
+   * (props) => renderable
+   * A function that takes in certain props that are used to render a component
+   */
   render: Function,
```
empyrical: It doesn't look like it passes any props to this callback. So you should be able to type this like:

```js
render: () => React.Element<any>,
```

rmcp1g15: Changed it to 
```js 
render: () => React.Node,
 ```

PR #21350, opened 2018-09-27
Remove PropTypes from ElementProperties

kdastan:
Related to #21342

### DIFF
--- a/Libraries/Inspector/ElementProperties.js
+++ b/Libraries/Inspector/ElementProperties.js
@@ -11,7 +11,6 @@
 'use strict';
 
 const BoxInspector = require('BoxInspector');
-const PropTypes = require('prop-types');
 const React = require('React');
 const StyleInspector = require('StyleInspector');
 const StyleSheet = require('StyleSheet');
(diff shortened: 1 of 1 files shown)


Libraries/Inspector/ElementProperties.js
```diff
@@ -26,27 +25,16 @@ const openFileInEditor = require('openFileInEditor');
...
 import type {DangerouslyImpreciseStyleProp} from 'StyleSheet';
 
-class ElementProperties extends React.Component<{
+type Props = $ReadOnly<{|
   hierarchy: Array<$FlowFixMe>,
-  style?: DangerouslyImpreciseStyleProp,
-  source?: {
-    fileName?: string,
-    lineNumber?: number,
+  style?: ?DangerouslyImpreciseStyleProp,
```
empyrical: You should change this up, so instead of importing and using `DangerouslyImpreciseStyleProp`, it uses `ViewStyleProp`

empyrical:
Had a look at the flow failures and looks like a fix will need to be done;

`frame`, `selection` and `setSelection` needs to be added to props, they'll have types like this:

```js
frame?: ?Object,
selection?: ?number,
setSelection?: ?number => mixed,
```

Frame is not well typed yet, so it can be stubbed out using `?Object`

empyrical:
LGTM! 👍

Libraries/Inspector/ElementProperties.js
```diff
@@ -24,29 +23,21 @@ const flattenStyle = require('flattenStyle');
...
-class ElementProperties extends React.Component<{
+type Props = $ReadOnly<{|
   hierarchy: Array<$FlowFixMe>,
-  style?: DangerouslyImpreciseStyleProp,
-  source?: {
-    fileName?: string,
-    lineNumber?: number,
+  style?: ?ViewStyleProp,
+  source?: ?{
+    fileName?: ?string,
```
TheSavior: I'm not sure these should be nullable. Lets leave fileName and lineNumber the way they were:

```
fileName?: string,
lineNumber?: string
```

Libraries/Inspector/ElementProperties.js
```diff
@@ -24,29 +23,21 @@ const flattenStyle = require('flattenStyle');
...
-      PropTypes.number,
-    ]),
-    source: PropTypes.shape({
-      fileName: PropTypes.string,
-      lineNumber: PropTypes.number,
-    }),
-  };
+  frame?: ?Object,
+  selection?: ?number,
+  setSelection?: (?number) => mixed,
```
TheSavior: setSelection is only ever called with a number, never with null, so the argument shouldn't be nullable. `setSelection?: (number) => mixed`

TheSavior:
Thanks for contributing!

PR #21351, opened 2018-09-27
Split EdgeInsets in DeprecatedPropTypes and flow types

danibonilha:
This PR splits EdgeInsetsPropTypes into EdgeInsetsPropTypes with only flow types and DeprecatedEdgeInsetsPropTypes inside DeprecatedProptypes with only PropTypes.

Related to #21342 

Test Plan:
----------
All flow checks pass.

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [DeprecatedEdgeInsetsPropTypes.js] - Created.
[GENERAL] [ENHANCEMENT] [EdgeInsetsPropTypes.js] - Flow types only.

### DIFF
--- a/Libraries/Components/Touchable/TouchableBounce.js
+++ b/Libraries/Components/Touchable/TouchableBounce.js
@@ -11,7 +11,7 @@
 
 const Animated = require('Animated');
 const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
-const EdgeInsetsPropType = require('DeprecatedEdgeInsetsPropType');
+const DeprecatedEdgeInsetsPropType = require('DeprecatedEdgeInsetsPropType');
 const NativeMethodsMixin = require('NativeMethodsMixin');
 const PropTypes = require('prop-types');
 const React = require('React');
@@ -70,7 +70,7 @@ const TouchableBounce = ((createReactClass({
      * reactivated! Move it back and forth several times while the scroll view
      * is disabled. Ensure you pass in a constant to reduce memory allocations.
      */
-    pressRetentionOffset: EdgeInsetsPropType,
+    pressRetentionOffset: DeprecatedEdgeInsetsPropType,
     releaseVelocity: PropTypes.number.isRequired,
     releaseBounciness: PropTypes.number.isRequired,
     /**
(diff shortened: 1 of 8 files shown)


danibonilha:
@empyrical I'm not sure why flow-check is failing in CI, should I find all references of require('EdgeInsetsPropType'),  replace with require('DeprecatedEdgeInsetsPropType') and export DeprecatedEdgeInsetsPropType in modules.exports ?

empyrical:
I missed that flow failure, sorry!

`TouchableWithoutFeedback.js` still uses `prop-types`. It still uses an old style of creating components called `createReactClass`, so for now just swap out the import by changing this:

```js
const EdgeInsetsPropType = require('EdgeInsetsPropType');
```

to this:

```js
const EdgeInsetsPropType = require('DeprecatedEdgeInsetsPropTypes');
```

Libraries/Components/Touchable/TouchableBounce.js
```diff
@@ -11,7 +11,7 @@
 
 const Animated = require('Animated');
 const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
-const EdgeInsetsPropType = require('EdgeInsetsPropType');
+const EdgeInsetsPropType = require('DeprecatedEdgeInsetsPropType');
```
TheSavior: Can you rename the variable to match the require?

```
const DeprecatedEdgeInsetsPropType = require('DeprecatedEdgeInsetsPropType');
```

Libraries/DeprecatedPropTypes/DeprecatedEdgeInsetsPropType.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @format
+ * @flow strict
+ */
+
+'use strict';
+
+const PropTypes = require('prop-types');
+
+const EdgeInsetsPropType = PropTypes.shape({
```
TheSavior: Should probably call this DeprecatedEdgeInsetsPropType here as well

Libraries/Components/Touchable/TouchableBounce.js
danibonilha: Sure, also should I changed in react-native-implementation 
```
  // Prop Types
  get ColorPropType() {
    return require('ColorPropType');
  },
  get EdgeInsetsPropType() {
    return require('EdgeInsetsPropType');
  },
```
to
```
  // Prop Types
  get ColorPropType() {
    return require('ColorPropType');
  },
  get EdgeInsetsPropType() {
    return require('DeprecatedEdgeInsetsPropType');
  },
```
?

TheSavior:
Thanks for contributing!

PR #21372, opened 2018-09-27
TVViewPropTypes, PlatformViewPropTypes, DeprecatedTVViewPropTypes

exced:
Related to #21342

- Split TVViewPropTypes
- PlatformViewPropTypes (dependencies) flow types and old prop-type definitions
- ViewStylePropTypes (dependencies) rm prop-type

### Test Plan:

Flow tests succeed

### Release Notes:

[GENERAL] [ENHANCEMENT] [TVViewPropTypes.js] - rm prop-types
[GENERAL] [ENHANCEMENT] [PlatformViewPropTypes.android.js] - replace prop-types by Flow
[GENERAL] [ENHANCEMENT] [PlatformViewPropTypes.ios.js] - replace prop-types by Flow
[GENERAL] [ENHANCEMENT] [DeprecatedTVViewPropTypes.js] - old prop-types


Libraries/DeprecatedPropTypes/DeprecatedTVViewPropTypes.js
```diff
@@ -0,0 +1,76 @@
...
+ * @format
+ */
+'use strict';
+var PropTypes = require('prop-types');
+
+/**
+ * Additional View properties for Apple TV
+ */
+var DeprecatedTVViewPropTypes = {
+  /**
```
TheSavior: You can remove the comments from here so that the source of truth is on the Flow types

Libraries/DeprecatedPropTypes/DeprecatedTVViewPropTypes.js
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ * @format
+ */
+'use strict';
+var PropTypes = require('prop-types');
```
TheSavior: Let's use `const` instead of `var` while we are here

Libraries/Components/AppleTV/TVViewPropTypes.js
```diff
@@ -34,51 +39,36 @@ const TVViewPropTypes = {
    * shiftDistanceY: Defaults to 2.0.
    * tiltAngle: Defaults to 0.05.
    * magnification: Defaults to 1.0.
-   * pressMagnification: Defaults to 1.0.
```
TheSavior: Do we still want these lines?

Libraries/Components/AppleTV/TVViewPropTypes.js
```diff
@@ -34,51 +39,36 @@ const TVViewPropTypes = {
...
-   * pressDuration: Defaults to 0.3.
-   * pressDelay: Defaults to 0.0.
    *
    * @platform ios
    */
-  tvParallaxProperties: PropTypes.object,
+  tvParallaxProperties?: Object,
```
TheSavior: Can you flow type this based on the docblock comment above?

Libraries/Components/AppleTV/TVViewPropTypes.js
```diff
@@ -1,30 +1,35 @@
 /**
- * Copyright (c) Facebook, Inc. and its affiliates.
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
  *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
+ * This source code is licensed under the BSD-style license found in the
```
TheSavior: I believe this is the old header as it is now MIT licensed.

empyrical:
Looks like `checkout_code` is failing because the test that detects if files with the `PATENTS` license snuck in. You'll want to adjust the license headers in these files to match the license headers in other JS files

TheSavior:
You'll need to fix the headers for the other files as well, I just only commented on one of them.

Libraries/Components/AppleTV/TVViewPropTypes.js
exced: I've never used this, do they bring useful information ?

TheSavior: I'm not sure either, but it seems like we should be keeping them unless the options are no longer valid.

Libraries/Components/AppleTV/TVViewPropTypes.js
```diff
@@ -10,6 +10,33 @@
 
 'use strict';
 
+type TVParallaxPropertiesType = {
```
empyrical: Looks very good! Should be made `$ReadOnly` and exact (`{| |}`) as well.

TheSavior:
Thanks for contributing!

PR #21386, opened 2018-09-28
SwipeableRow: Remove PropTypes, convert to ES6 class

empyrical:
Part of: https://github.com/facebook/react-native/issues/21342

This PR removes prop types from `SwipeableRow`, and converts it from a `createReactClass` to an ES6 class.

Test Plan:
----------
`flow check` passes. Nothing seems off in Android RNTester.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Experimental/SwipeableRow/SwipeableRow.js] - Removed PropTypes, converted to ES6 class

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableRow.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableRow.js
@@ -16,13 +16,8 @@ const PanResponder = require('PanResponder');
 const React = require('React');
 const PropTypes = require('prop-types');
 const StyleSheet = require('StyleSheet');
-/* $FlowFixMe(>=0.54.0 site=react_native_oss) This comment suppresses an error
- * found when Flow v0.54 was deployed. To see the error delete this comment and
- * run Flow. */
-const TimerMixin = require('react-timer-mixin');
 const View = require('View');
 
-const createReactClass = require('create-react-class');
 const emptyFunction = require('fbjs/lib/emptyFunction');
 
 import type {LayoutEvent, PressEvent} from 'CoreEventTypes';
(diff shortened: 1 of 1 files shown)


Libraries/Experimental/SwipeableRow/SwipeableRow.js
```diff
@@ -200,61 +188,61 @@ const SwipeableRow = createReactClass({
         {swipeableView}
       </View>
     );
-  },
+  }
 
   close(): void {
-    this.props.onClose();
+    this.props.onClose && this.props.onClose();
```
TheSavior: onClose is required based on the propTypes, you shouldn't need this check. If you do, then the flow types for onClose aren't properly showing that it is required.

Libraries/Experimental/SwipeableRow/SwipeableRow.js
```diff
@@ -200,61 +188,61 @@ const SwipeableRow = createReactClass({
...
-  ): void {},
+  ): void {}
 
   _handlePanResponderMove(event: PressEvent, gestureState: GestureState): void {
     if (this._isSwipingExcessivelyRightFromClosedPosition(gestureState)) {
       return;
     }
 
-    this.props.onSwipeStart();
+    this.props.onSwipeStart && this.props.onSwipeStart();
```
TheSavior: same with `onSwipeStart`

TheSavior: If you could check the other ones in this file too that would be great.

empyrical:
Cleaned up those nullchecks, and removed the nullability of the other props that have default props

PR #21398, opened 2018-09-28
Expose a getter for overflow setting in ReactViewGroup

kmagiera:
This change adds getter for overflow attribute in ReactViewGroup class. Overflow setting can affect how view children are drawn but also how hit testing behaves when receiving touch. Exposing this setting makes it possible for gesture-handler library to implement proper hit testing that takes into account overflow property of a view.

Test Plan:
----------
This change only adds a getter function that will be used by external libraries. 

Release Notes:
--------------
[ANDROID] [MINOR] [ReactViewGroup] - Expose a getter for overflow setting in ReactViewGroup


sahrens:
Are you planning on fixing overflow hit testing for regular touch handling as well?

Issue #21311, opened 2018-09-25
Cannot resolve module when running Flow on Windows


zenz:
## Environment
Run `react-native info` in your terminal and paste its contents here.
react-native info

  React Native Environment Info:
    System:
      OS: Windows 10
      CPU: x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
      Memory: 2.09 GB / 7.91 GB
    Binaries:
      Yarn: 1.9.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.1.0 - C:\Program Files\nodejs\npm.CMD

## Description
using flow not working and got error with "Cannot resolve module XXXXXX". but same action under macOS works well.

## Reproducible Demo
react-native init testapp
cd testapp
yarn add flow-bin@0.78.0 --dev
yarn flow

under windows will get errors.

zenz:
with
```
"@babel/runtime": "^7.1.2",
 "metro-react-native-babel-preset": "0.48.0",
```
The issue seems had been fixed. So I close this now.

zenz:
Sorry, While create a new project with init, the issue shows up again.

zenz:
No other one using flow under windows or don't have any problem on this?

jlkalberer:
I had the issue but I just added more ignores to the flowconfig

zenz:
@gengjiawen Thank you!

gengjiawen:
@TheSavior Looks like a flow issue, can you find some people work on this ?
```cmd
$ flow check --flowconfig-name .flowconfig.android
Error --------------------------------------------------------------- IntegrationTests/AccessibilityManagerTest.js:16:39

Cannot resolve module `RCTDeviceEventEmitter`.

   16| const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
                                             ^^^^^^^^^^^^^^^^^^^^^^^


Error -------------------------------------------------------------------------- IntegrationTests/AppEventsTest.js:18:28

Cannot resolve module `deepDiffer`.

   18| const deepDiffer = require('deepDiffer');
```

gengjiawen:
I have reported to flow team too, https://github.com/facebook/flow/issues/7138.

gengjiawen:
Update: fb look into this, hopefully we can fix this soon.

jamesisaac:
I believe this should fix it: https://github.com/facebook/flow/pull/7419

zenz:
[...]

PR #21401, opened 2018-09-28
add flow check to windows ci

gengjiawen:
## Motivation
add flow check to windows ci
## Test Plan
none
## Related PRs
none
## Release Notes
 [GENERAL] [INTERNAL] [CI] - add flow check to windows ci.

### DIFF
--- a/.appveyor/config.yml
+++ b/.appveyor/config.yml
@@ -30,16 +30,15 @@ install:
   - appveyor DownloadFile "%NDK_TOOLS_URL%" -FileName "%TMP%/ndk.zip"
   - 7z x "%TMP%/ndk.zip" -o"%ANDROID_HOME%" > nul
 
-  - ps: Install-Product node $env:nodejs_version
-  - node --version
-  - yarn --version
+  - ps: Install-Product node $env:nodejs_version x64
+  - npx envinfo@latest
   - appveyor-retry yarn install
 
 build_script:
+  - yarn run flow-check-android
+  - yarn run flow-check-ios
+  - yarn run test
   - gradlew.bat RNTester:android:app:assembleRelease
-
-test_script:
-  - npm test
 
 cache:
   - node_modules


hramos:
Should we run Flow for iOS files as well? Since this is only JavaScript, I expect it to work on Windows.

TheSavior:
Is it even worth running this? Since it runs on Circle, running it on Appveyor is essentially just testing Flow itself to make sure it doesn't have weird windows specific bugs. It doesn't seem like it really tells us anything extra about our codebase.

gengjiawen:
@TheSavior Yes, because now flow is broken on windows, if we have this, we can find this earlier see https://github.com/facebook/react-native/issues/21311.

@hramos Yes, I think we can add it. My first thought is that iOS check maybe redundant on windows platform.

TheSavior:
Flow is broken on Windows!? :O

gengjiawen:
If there is no future question, I think this can be merged.

gengjiawen:
@hramos Is anyone working on the flow issue in windows ?

hramos:
This is not something I am tracking closely at the time. We should definitely merge this once #21311 is fixed, though, to prevent regressions as you suggested.

hramos:
Appveyor build failed with:

```
$ flow check --flowconfig-name .flowconfig.android
C:\projects\react-native\node_modules\flow-bin\cli.js:17
  throw new Error('Platform not supported.');
  ^
```

gengjiawen:
After swith to x64 platform, here is the new error:
```console
$ flow check --flowconfig-name .flowconfig.android
Error -------------------------------------------------------------------- node_modules/invariant/invariant.js.flow:4:14
Unclear type. Using `any`, `Object`, or `Function` types is not safe! (`unclear-type`)
   4|   condition: any,
                   ^^^
Error -------------------------------------------------------------------- node_modules/invariant/invariant.js.flow:6:18
Unclear type. Using `any`, `Object`, or `Function` types is not safe! (`unclear-type`)
   6|   ...args: Array<any>
                       ^^^
```

jamesisaac:
It's complaining about this file: https://github.com/zertosh/invariant/blob/master/invariant.js.flow

Imo that file should be erroring... it's annotated Flow strict mode, but it's using `any` types, which correctly triggers the `unclear-type` rule.  So the bug seems to me more like why does that file not error on Linux?

I can't see anything in RN's flowconfig which indicates that file is meant to be ignored.  So don't think this is to do with normalizing file paths like the previous errors were.

jamesisaac:
@gengjiawen Maybe this explains it: https://github.com/facebook/flow/commit/ba425e403cf40e9232aab68aff96b8e3fca192b0

gengjiawen:
> @gengjiawen Maybe this explains it: [facebook/flow@ba425e4](https://github.com/facebook/flow/commit/ba425e403cf40e9232aab68aff96b8e3fca192b0)

Let's wait for next release :)

cpojer:
@gengjiawen the latest Flow version should now be deployed on master. Can you rebase and make this PR pass on Windows? Then we can finally land and enable it!

gengjiawen:
> @gengjiawen the latest Flow version should now be deployed on master. Can you rebase and make this PR pass on Windows? Then we can finally land and enable it!

Done :)

hramos:
🎉 It worked! Landing.

PR #21429, opened 2018-09-30
pass through the displayName from the mocked component

probablyup:
Without this, most tooling will just say "Component" instead of "Text", "View" etc.

Test Plan:
----------

Copied the change over and verified that it worked locally in the styled-components tests.

Release Notes:
--------------

[INTERNAL] [BUGFIX] [jest/mockComponent.js] - resume passing the mocked component's displayName through so it will appear in downstream tooling

### DIFF
--- a/jest/__mocks__/FakeComponent.js
+++ b/jest/__mocks__/FakeComponent.js
@@ -1,0 +1,1 @@
+export default () => null
--- a/jest/__mocks__/FakeComponent.js
+++ b/jest/__mocks__/FakeComponent.js
@@ -1,1 +1,1 @@
-export default () => null
+export default () => null;
--- a/jest/__mocks__/FakeComponent.js
+++ b/jest/__mocks__/FakeComponent.js
@@ -1,1 +1,1 @@
-export default () => null;
+module.exports = () => null;
--- a/jest/__mocks__/FakeForwardRefComponent.js
+++ b/jest/__mocks__/FakeForwardRefComponent.js
@@ -1,0 +1,9 @@
+const React = require('react');
+
+const ForwardedComponent = () => null;
+
+const ForwardRefComponent = React.forwardRef((props, ref) => <ForwardedComponent {...props} forwardedRef={ref} />);
+
+ForwardRefComponent.displayName = 'ForwardedComponent';
+
+module.exports = ForwardRefComponent;
--- a/jest/__tests__/mockComponent-test.js
+++ b/jest/__tests__/mockComponent-test.js
@@ -1,0 +1,13 @@
+const mockComponent = require('../mockComponent');
+const FakeComponent = require('FakeComponent')
+const TextInput = require('TextInput')
+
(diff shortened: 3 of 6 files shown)


TheSavior:
Thanks for contributing! This looks reasonable to me at first glance, but it is hard to tell from your test plan how I can reproduce the issue you are fixing. Can you provide a minimal jest test that reproduces this error if your fix wasn't included? Better yet, can you add a jest test for `mockComponent` to the codebase to ensure this keeps working in the future?

probablyup:
Sure thing, I'll add a test

probablyup:
Side question, why was this done? https://github.com/facebook/react-native/commit/76eebce3c20152fdd5386ec405524d0810770eea#diff-a562e4a7ac9e900c6d2bc644453e9152

The displayName was removed from a bunch of components which is kind of a mistake because it makes them hard to identify in devtools and react warnings.

TheSavior:
Our understanding is that React.forwardRef was updated to utilize the displayName of the passed in component so those lines were no longer necessary.

probablyup:
It doesn’t as far as I know. It just respects if you set a displayName on
the forwardRef

probablyup:
@TheSavior The behavior is if you set a displayName on the ForwardRef wrapper it'll take it into account in devtools I believe. `forwardRef` doesn't know what component it's rendering until `render` is called so there's no way to acquire the underlying component's displayName statically.

TheSavior:
`View` and `Text` both use `forwardRef` and don't explicitly set a `displayName`. Using React DevTools with our current production app, I see the proper name in `forwardRef`:

screen shot 2018-10-01 at 12 57 39 pm

Is this inconsistent with what you are seeing?

probablyup:
@TheSavior I think devtools is kind of a special case since it literally executes the render function of the forwardRef to obtain the underlying component and then shows the displayName as a derivation of that. Enzyme and other tooling likely will not go to that length since the fact that forwardRef has a function called "render" is an implementation detail and not part of the public API.

[I actually wrote the change](https://github.com/facebook/react/pull/13615) that was recently merged into react core to start directly honoring `displayName` set on `forwardRef` and this is an extension of that effort.

TheSavior:
Okay, so to make sure I understand, is your request that we add back the annotations to all the components that use `forwardRef` so that tooling like enzyme is able to read those names? 

It seems like that will be hard to enforce everywhere forwardRef is used and is a workaround that adds runtime code to help make build and test time tooling easier. Am I misunderstanding something?

probablyup:
I’d say that’s phase 2 of this PR/changeset, yes. I’m not sure what other
recourse there is without having React core do displayName as a getter for
forwardRef and emulating the same behavior as devtools because most testing
output becomes less useful without the names.

probablyup:
And actually that wouldn’t even work consistently because it’s entirely
possible to dynamically render different things inside the forwardRef
render based on props.

cpojer:
I'm gonna close this as there hasn't been progress in many months and the conclusion seems to be that this isn't what we want to do.

PR #21456, opened 2018-10-02
iOS: Support inline view truncation

rigdern:
If text is truncated and an inline view appears after the truncation point, the user should not see the inline view. Instead, we have a bug such that the inline view is always visible at the end of the visible text.

This commit fixes this by marking the inline view as hidden if it appears after the truncation point.

This appears to be a regression. React Native used to have logic similar to what this commit is adding: https://github.com/facebook/react-native/blob/1e2a924ba60001c6f0587c7561536f00b2922cbf/Libraries/Text/RCTShadowText.m#L186-L192

**Before fix**

Inline view (blue square) is visible even though it appears after the truncation point:

image

The full text being rendered was:

```
<Text numberOfLines={1}>
  Lorem ipsum dolor sit amet, consectetur adipiscing elit,
  sed do eiusmod tempor incididunt ut labore et dolore magna
  <View style={{width: 50, height: 50, backgroundColor: 'steelblue'}} />
</Text>
```

**After fix**

Inline view is properly truncated:

image

**Test Plan**

Tested that the inline view is hidden if it appears after the truncation point when `numberOfLines` is 1 and 2. Similarly, verified that the inline view is visible if it appears before the truncation point.

**Release Notes**

[IOS] [BUGFIX] [Text] - Fix case where inline view is visible even though it should have been truncated

Adam Comella
Microsoft Corp.

### DIFF
--- a/Libraries/Text/Text/RCTTextShadowView.m
+++ b/Libraries/Text/Text/RCTTextShadowView.m
@@ -290,6 +290,9 @@
         RCTRoundPixelValue(attachmentSize.width),
         RCTRoundPixelValue(attachmentSize.height)
       }};
+      
+      NSRange truncatedGlyphRange = [layoutManager truncatedGlyphRangeInLineFragmentForGlyphAtIndex:range.location];
+      BOOL viewIsTruncated = NSIntersectionRange(range, truncatedGlyphRange).length != 0;
 
       RCTLayoutContext localLayoutContext = layoutContext;
       localLayoutContext.absolutePosition.x += frame.origin.x;
@@ -300,9 +303,9 @@
                         layoutDirection:self.layoutMetrics.layoutDirection
                           layoutContext:localLayoutContext];
 
-      // Reinforcing a proper frame origin for the Shadow View.
       RCTLayoutMetrics localLayoutMetrics = shadowView.layoutMetrics;
-      localLayoutMetrics.frame.origin = frame.origin;
+      localLayoutMetrics.frame.origin = frame.origin; // Reinforcing a proper frame origin for the Shadow View.
+      localLayoutMetrics.isHiddenDueToClipping = viewIsTruncated;
       [shadowView layoutWithMetrics:localLayoutMetrics layoutContext:localLayoutContext];
     }
   ];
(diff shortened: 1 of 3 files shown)


Libraries/Text/Text/RCTTextShadowView.m
```diff
@@ -298,9 +301,9 @@ - (void)layoutSubviewsWithContext:(RCTLayoutContext)layoutContext
                         layoutDirection:self.layoutMetrics.layoutDirection
                           layoutContext:localLayoutContext];
 
-      // Reinforcing a proper frame origin for the Shadow View.
       RCTLayoutMetrics localLayoutMetrics = shadowView.layoutMetrics;
-      localLayoutMetrics.frame.origin = frame.origin;
+      localLayoutMetrics.frame.origin = frame.origin; // Reinforcing a proper frame origin for the Shadow View.
```
shergin: Why not just set `display: none` here instead of introducing a new prop?
`localLayoutMetrics.display = RCTDisplayTypeNone;`

rigdern: @shergin good question. It's possible that my lack of familiarity with the code incorrectly turned me off of this approach.

I was afraid that setting `localLayoutMetrics.display = RCTDisplayTypeNone;` might prevent layout from running on subsequent layout passes and consequently preventing `localLayoutMetrics.display` from ever being updated back to the visible value (also, how do I make sure to change it back to the value the user set -- `RCTDisplayTypeFlex` vs `RCTDisplayTypeInline`?). I also had a concern about multiple pieces of code fighting to set the same field (what if between the time I set `display` and rendering somebody overwrites my value?).

Are my concerns unfounded? Is `shadowView.layoutMetrics` recreated from scratch on each layout pass? Please enlighten me 😄.

shergin: I think this would be totally legit and work.
An `RCTLayoutMetrics` thing represents final computed layout metrics and can/should not be used as some input information for future layout passes; it's a product meant to be consumed by mounting layer.

If it's a View (and it it a View), it always should be `RCTDisplayTypeFlex`. Inline is reserved for VirtualText nodes.

PR #21457, opened 2018-10-03
Added groupCollapsed polyfill

### LINKED ISSUE #21446: groupCollapsed is not available in the console polyfill when debugging in JSC (iOS)
## Environment
```
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 132.79 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.12.0 - ~/.nvm/versions/node/v8.12.0/bin/node
      Yarn: 1.6.0 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v8.12.0/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 26.0.3, 27.0.3, 28.0.2
        API Levels: 21, 22, 23, 24, 25, 26, 27
    IDEs:
    npmPackages:
      react: 16.5.1 => 16.5.1 
      react-native: 0.57.1 => 0.57.1 
```

## Description
The console polyfill doesn't have the `groupCollapsed` function when debugging using Safari. 
This makes the console log very spammy as we can't collapse any console log calls.

## Reproducible Demo
- Run the hello world app for ios `react-native run-ios`
- Use Safari to open the JSContext for the app
image
- Type `console.groupCollapsed` in the console and you will get `undefined`
image

### Notes
- I did some digging to find the polyfill and looks like its this https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/polyfills/console.js#L539
- Some might ask, why would you want to debug using Safari instead of Chrome?
because not having to "Debug JS remotely" is crash free and much faster.

Eyesonly88:
`groupCollapsed` is used to group logs but collapsed which is very useful when the console has many logs (e.g. when using `redux-logger`).

For developers who prefer to debug iOS apps using Safari JSC, the `groupCollapsed` was not polyfilled.

Fixes #21446 

Test Plan:
----------
I tested it in the Safari Developer console.

**Input**:

```
console.groupCollapsed('Hello RN')
console.log('some data', {hello: "world"})
console.log('another line')
console.log('and another one')
console.groupEnd()
```

**Output**:
A single collapsed log containing the other logs as expected. See demo below.

**Demo**:

groupcollapsed_demo

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [JSC] - Added `groupCollapsed` to the console polyfill to enable log collapsing in consoles such as Safari

### DIFF
--- a/Libraries/polyfills/console.js
+++ b/Libraries/polyfills/console.js
@@ -520,7 +520,6 @@ function consoleGroupPolyfill(label) {
   groupStack.push(GROUP_PAD);
 }
 
-
 function consoleGroupCollapsedPolyfill(label) {
   global.nativeLoggingHook(groupFormat(GROUP_CLOSE, label), LOG_LEVELS.info);
   groupStack.push(GROUP_PAD);
--- a/Libraries/polyfills/console.js
+++ b/Libraries/polyfills/console.js
@@ -520,6 +520,12 @@ function consoleGroupPolyfill(label) {
   groupStack.push(GROUP_PAD);
 }
 
+
+function consoleGroupCollapsedPolyfill(label) {
+  global.nativeLoggingHook(groupFormat(GROUP_CLOSE, label), LOG_LEVELS.info);
+  groupStack.push(GROUP_PAD);
+}
+
 function consoleGroupEndPolyfill() {
   groupStack.pop();
   global.nativeLoggingHook(groupFormat(GROUP_CLOSE), LOG_LEVELS.info);
@@ -537,6 +543,7 @@ if (global.nativeLoggingHook) {
     table: consoleTablePolyfill,
     group: consoleGroupPolyfill,
     groupEnd: consoleGroupEndPolyfill,
+    groupCollapsed: consoleGroupCollapsedPolyfill,
   };
 
   // If available, also call the original `console` method since that is


Libraries/polyfills/console.js
```diff
@@ -537,6 +543,7 @@ if (global.nativeLoggingHook) {
     table: consoleTablePolyfill,
     group: consoleGroupPolyfill,
     groupEnd: consoleGroupEndPolyfill,
+    groupCollapsed: consoleGroupCollapsedPolyfill,
```
newyankeecodeshop: I think this should be the `consoleGroupPolyfill`. I believe `console.groupCollapsed` should behave the same as `console.group` in a non-debugger environment logging to the OS.

Eyesonly88: Hey @newyankeecodeshop if I use `consoleGroupPolyfill` then the log won't be collapsed because of `GROUP_OPEN` inside `consoleGroupPolyfill`.

newyankeecodeshop: @Eyesonly88 I don't understand how the `GROUP_OPEN` character vs. `GROUP_CLOSE` makes a difference about whether the group is shown collapsed. When you attach the Safari debugger to the JSContext, is it still using the console polyfill?

Eyesonly88: @newyankeecodeshop Yes it is using the console polyfill. Without my code, there is no `groupCollapsed`. As you can see in the demo, using `GROUP_CLOSE` makes it collapsed.

Eyesonly88: Exactly.

cpojer:
Thanks for the PR. Seems like it previously fell through the radar. Sorry about that!

Issue #21485, opened 2018-10-04
Removing TimerMixin, SubscribableMixin from components.


TheSavior:
As part of our goal to modernize and type our core components, we need to migrate off of mixins. This task is to track the work to remove `TimerMixin` from our codebase.

*This is step 2. Step 1 can be [found here](https://github.com/facebook/react-native/issues/21342).*

[`TimerMixin`](https://github.com/reactjs/react-timer-mixin) enables components to call functions like `this.setTimeout` and have that timeout automatically cleared when the component is unmounted. In order to remove `TimerMixin` from these components, we will need to do this work manually.

The following are the individual files that currently use `TimerMixin` which will need to be migrated. Each file should be migrated in its own PR and each file is about the size of a good first issue to learn and get familiar with a part of the React Native codebase. If you’d like to convert one, pick one from the list, comment on this issue that you are interested in converting it, and follow the tips below for help successfully updating that file.

- [x] IntegrationTests/ReactContentSizeUpdateTest.js
- [x] IntegrationTests/TimersTest.js
- [x] Libraries/Components/TextInput/TextInput.js
- [x] Libraries/Components/Touchable/TouchableOpacity.js
- [x] Libraries/Components/Touchable/TouchableWithoutFeedback.js
- [x] Libraries/Experimental/SwipeableRow/SwipeableRow.js
- [x] Libraries/Lists/ListView/ListView.js (in progress)
- [x] RNTester/js/ProgressBarAndroidExample.android.js
- [x] RNTester/js/ProgressViewIOSExample.js
- [x] RNTester/js/TimerExample.js

Here is an example component that uses `setTimeout` from `TimerMixin`.
```
createReactClass({
  mixins: [TimerMixin],

  render() {
    return (
      <View
        onPress={() => {
          this.setTimeout(() => {
            alert('hi');
          });
        }}>
        Whee
      </View>
    );
  },
});
```

This would need to be changed to:
```
createReactClass({
  _timeoutID: (null: ?TimeoutID),

  componentWillUnmount() {
    if (this._timeoutID != null) {
      clearTimeout(this._timeoutID);
    }
  },

  render() {
    return (
      <View
        onPress={() => {
          this._timeoutID = setTimeout(() => {
            alert('hi');
          });
        }}>
        Whee
      </View>
    );
  },
});
```

[...]

PR #21488, opened 2018-10-04
[TimerMixin] Remove TimerMixin from ListView

exced:
Related to #21485 

- Remove TimerMixin from ListView

### Test Plan:

- All flow tests succeed.
- RNTester: <ListView> iOS (this change should only affect iOS because calculateChildFrames is iOS only)
Show perf monitor, show ListView* screen, start scrolling. UI frame Rate is used at the beginning. When scrolling there is no drop in FPS rate.

TODO: I think a load test would be more relevant: 
- Update props multiple times and scroll

### Release Notes:

[GENERAL] [ENHANCEMENT] [ListView.js] - rm TimerMixin

### DIFF
--- a/Libraries/Lists/ListView/ListView.js
+++ b/Libraries/Lists/ListView/ListView.js
@@ -215,7 +215,7 @@ type Props = $ReadOnly<{|
 
 const ListView = createReactClass({
   displayName: 'ListView',
-  _rafId: (null: ?AnimationFrameID),
+  _rafIds: ([]: Array<AnimationFrameID>),
   _childFrames: ([]: Array<Object>),
   _sentEndForContentLength: (null: ?number),
   _scrollComponent: (null: ?React.ElementRef<typeof ScrollView>),
@@ -348,6 +348,7 @@ const ListView = createReactClass({
       offset: 0,
     };
 
+    this._rafIds = [];
     this._childFrames = [];
     this._visibleRows = {};
     this._prevRenderedRowsCount = 0;
@@ -355,15 +356,14 @@ const ListView = createReactClass({
   },
 
   componentWillUnmount: function() {
-    if (this._rafId != null) {
-      cancelAnimationFrame(this._rafId);
-    }
+    this._rafIds.forEach(cancelAnimationFrame);
+    this._rafIds = [];
   },
 
   componentDidMount: function() {
     // do this in animation frame until componentDidMount actually runs after
     // the component is laid out
-    this._rafId = requestAnimationFrame(() => {
+    this._requestAnimationFrame(() => {
       this._measureAndUpdateScrollProps();
     });
   },
@@ -391,7 +391,7 @@ const ListView = createReactClass({
   },
 
   componentDidUpdate: function() {
-    this._rafId = requestAnimationFrame(() => {
+    this._requestAnimationFrame(() => {
       this._measureAndUpdateScrollProps();
     });
   },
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for the PR! Please note the comment about the importance of test plans from the main issue. I don't think relying on Flow for this is sufficient. We've seen some surprising errors sneak in after having made these changes when they weren't tested by running the code.

>When submitting a PR removing an instance of TimerMixin, it is important to include a good test plan. This test plan helps reviewers understand the work you did to verify that your changes are correct and complete. Since this changes the runtime logic of these components, it is important that these changes are tested in an app. For most of these components, you should be able to validate your change via RNTester. If there is nothing that exists as-is in RNTester which gives you confidence in your change, adding something to RNTester would be greatly appreciated. :)

Libraries/Lists/ListView/ListView.js
```diff
@@ -216,14 +215,15 @@ type Props = $ReadOnly<{|
 
 const ListView = createReactClass({
   displayName: 'ListView',
+  _rafId: (null: ?AnimationFrameID),
```
RSNara:
```JavaScript
_rafIds: Array<AnimationFrameID>
```

Libraries/Lists/ListView/ListView.js
```diff
@@ -347,16 +347,23 @@ const ListView = createReactClass({
...
+    if (this._rafId != null) {
+      cancelAnimationFrame(this._rafId);
+    }
+  },
+
   componentDidMount: function() {
     // do this in animation frame until componentDidMount actually runs after
     // the component is laid out
-    this.requestAnimationFrame(() => {
+    this._rafId = requestAnimationFrame(() => {
```
RSNara: Both `componentDidMount` and `componentDidUpdate` will write to `this._rafId`, which means that the `rafId` from `componentDidUpdate` will overwrite the `rafId` from `componentDidMount`. Suppose that both hooks are executed, and then the component is unmounted, all before the first `requestAnimationFrame` hook is executed. Then, we'll end up executing `this._measureAndUpdateScrollProps();`. 

To get around this, we could add the following function to ListView:

```JavaScript
requestAnimationFrame(fn: () => void): void {
  const rafId = requestAnimationFrame(() => {
    clearAnimationFrame(rafId);
    this._rafIds.splice(this._rafIds.indexOf(rafId));
    fn();
  })
  this._rafIds.push(rafId);
},
```

Libraries/Lists/ListView/ListView.js
```diff
@@ -347,16 +347,23 @@ const ListView = createReactClass({
...
     this._visibleRows = {};
     this._prevRenderedRowsCount = 0;
     this._sentEndForContentLength = null;
   },
 
+  componentWillUnmount: function() {
+    if (this._rafId != null) {
+      cancelAnimationFrame(this._rafId);
+    }
+  },
```
RSNara:
```JavaScript
componentWillUnmount() {
  this._rafIds.forEach(clearAnimationFrame);
  this._rafIds = [];
}
```

Libraries/Lists/ListView/ListView.js
```diff
@@ -347,16 +347,23 @@ const ListView = createReactClass({
       contentLength: null,
       offset: 0,
     };
+
```
RSNara:
```JavaScript
this._rafIds = [];
```

exced:
You're right, I'll update it to have multiple references and come up with a test plan

PR #21492, opened 2018-10-04
Critical improvements for Map and Set polyfills.

benjamn:
Test Plan:
---
See included tests.

Release Notes:
---
* [POLYFILL] [SPEC] The `Map` and `Set` polyfills no longer reject non-extensible object keys, through a combination of tagging objects _before_ they become non-extensible, and falling back to a linear `Array#indexOf` lookup when there is no other option.
* [POLYFILL] [BUGFIX] Fixed a critical bug that caused `Map` and `Set` object hashes to collide with the hashes of other objects in their prototype chains.

Details
---
When I first discovered these bugs in the course of investigating https://github.com/apollographql/react-apollo/issues/2442, I was disappointed and disheartened, and I was tempted to recommend that Apollo developers find a replacement for the React Native `Map` and `Set` polyfills.

However, as I dug into the code and investigated alternative polyfills, I realized that this `Map` polyfill has the potential to be one of the fastest and best polyfills available, since almost every other polyfill avoids tagging object keys, and instead resorts to linear lookup. I hope this PR demonstrates that we can provide constant-time lookup in (almost) all cases, even for frozen object keys.

While I can certainly provide more creative solutions to developers who use my libraries, from monkey-patching `Map.prototype` to using the `react-native` override field in `package.json`, I think the best solution is simply to improve the official polyfills, which is what I've tried to do in this PR.

Please read the commit messages and the comments that I've added to the code for more explanation, and thanks for your consideration!

### DIFF
--- a/Libraries/vendor/core/Map.js
+++ b/Libraries/vendor/core/Map.js
@@ -26,6 +26,9 @@ module.exports = (function(global, undefined) {
     return global.Map;
   }
 
+  // In case this module has not already been evaluated, import it now.
+  require('./_wrapObjectFreezeAndFriends');
+
   /**
    * == ES6 Map Collection ==
    *
@@ -38,15 +41,17 @@ module.exports = (function(global, undefined) {
    *
    * https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map-objects
    *
-   * There only two -- rather small -- diviations from the spec:
+   * There only two -- rather small -- deviations from the spec:
    *
-   * 1. The use of frozen objects as keys.
-   *    We decided not to allow and simply throw an error. The reason being is
-   *    we store a "hash" on the object for fast access to it's place in the
-   *    internal map entries.
-   *    If this turns out to be a popular use case it's possible to implement by
-   *    overiding `Object.freeze` to store a "hash" property on the object
-   *    for later use with the map.
+   * 1. The use of untagged frozen objects as keys.
+   *    We decided not to allow and simply throw an error, because this
+   *    implementation of Map works by tagging objects used as Map keys
+   *    with a secret hash property for fast access to the object's place
+   *    in the internal _mapData array. However, to limit the impact of
+   *    this spec deviation, Libraries/Core/InitializeCore.js also wraps
+   *    Object.freeze, Object.seal, and Object.preventExtensions so that
+   *    they tag objects before making them non-extensible, by inserting
+   *    each object into a Map and then immediately removing it.
    *
    * 2. The `size` property on a map object is a regular property and not a
    *    computed property on the prototype as described by the spec.
--- a/Libraries/vendor/core/Map.js
+++ b/Libraries/vendor/core/Map.js
@@ -553,15 +553,17 @@ module.exports = (function(global, undefined) {
       }
 
       if (isExtensible(o)) {
-        hashCounter += 1;
         if (isES5) {
           Object.defineProperty(o, hashProperty, {
             enumerable: false,
             writable: false,
             configurable: false,
-            value: hashCounter,
+            value: ++hashCounter,
           });
-        } else if (o.propertyIsEnumerable) {
+          return hashCounter;
+        }
+
+        if (o.propertyIsEnumerable) {
           // Since we can't define a non-enumerable property on the object
           // we'll hijack one of the less-used non-enumerable properties to
           // save our hash on it. Additionally, since this is a function it
@@ -569,22 +571,19 @@ module.exports = (function(global, undefined) {
           o.propertyIsEnumerable = function() {
             return propIsEnumerable.apply(this, arguments);
           };
-          o.propertyIsEnumerable[hashProperty] = hashCounter;
-        } else {
-          throw new Error('Unable to set a non-enumerable property on object.');
+          return o.propertyIsEnumerable[hashProperty] = ++hashCounter;
         }
-        return hashCounter;
-      } else {
-        // If the object is not extensible, fall back to storing it in an
-        // array and using Array.prototype.indexOf to find it.
-        let index = nonExtensibleObjects.indexOf(o);
-        if (index < 0) {
-          index = nonExtensibleObjects.length;
-          nonExtensibleObjects[index] = o;
-          nonExtensibleHashes[index] = ++hashCounter;
-        }
-        return nonExtensibleHashes[index];
       }
+
+      // If the object is not extensible, fall back to storing it in an
+      // array and using Array.prototype.indexOf to find it.
+      let index = nonExtensibleObjects.indexOf(o);
+      if (index < 0) {
+        index = nonExtensibleObjects.length;
+        nonExtensibleObjects[index] = o;
+        nonExtensibleHashes[index] = ++hashCounter;
+      }
+      return nonExtensibleHashes[index];
     };
   })();
 
--- a/Libraries/vendor/core/Map.js
+++ b/Libraries/vendor/core/Map.js
@@ -29,6 +29,8 @@ module.exports = (function(global, undefined) {
   // In case this module has not already been evaluated, import it now.
   require('./_wrapObjectFreezeAndFriends');
 
+  const hasOwn = Object.prototype.hasOwnProperty;
+
   /**
    * == ES6 Map Collection ==
    *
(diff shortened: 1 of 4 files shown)


benjamn:
Were the release notes supposed to be in some other file, or should I simply have formatted them differently in the PR description?

Libraries/Core/InitializeCore.js
```diff
@@ -44,6 +44,28 @@ if (_shouldPolyfillCollection('Set')) {
   polyfillGlobal('Set', () => require('Set'));
 }
 
+// Wrap Object.{freeze,seal,preventExtensions} so each function adds its
```
yungsters: Can you avoid doing this if the polyfill is not needed (i.e. for versions of JSC where `Map` and `Set` are present)?

Libraries/vendor/core/Map.js
```diff
@@ -39,15 +41,17 @@ module.exports = (function(global, undefined) {
    *
    * https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map-objects
    *
-   * There only two -- rather small -- diviations from the spec:
```
yungsters: lol

Libraries/Core/InitializeCore.js
yungsters: Also, should this be in a new file/module that lives closer to `Map.js`?

hramos:
re: release notes, you did nothing wrong. The bot as currently implemented will only recognize categories from the list provided by the [PULL_REQUEST_TEMPLATE](https://github.com/facebook/react-native/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (e.g. CLI, GENERAL, IOS, ...), and POLYFILL is not one of those. We should just relax this check, thanks for pointing it out.

I also noticed it's been too verbose with its `eslint` nits, that can be fixed.

Libraries/Core/InitializeCore.js
benjamn: Definitely!

benjamn: Improved: 114008536e40c394e990682026d2a8b55922fec6

Libraries/Core/InitializeCore.js
```diff
@@ -38,9 +38,11 @@ if (global.window === undefined) {
 // Set up collections
 const _shouldPolyfillCollection = require('_shouldPolyfillES6Collection');
 if (_shouldPolyfillCollection('Map')) {
+  require('_wrapObjectFreezeAndFriends');
```
benjamn: I'm a little surprised that this bare module identifier just works, without `@providesModule _wrapObjectFreezeAndFriends` or something like that. Should I be worried?

benjamn:
@yungsters Ok, I think I've addressed your feedback. Moving the `Object.freeze` wrapping logic into its own module was a great idea, since it means we can safely import that module in multiple places (first in `InitializeCore.js` and then in `Map.js`, just to be sure it gets imported if the polyfill is used).

Still interested in your thoughts on https://github.com/facebook/react-native/pull/21492#discussion_r222995540, though I'm fine with it (since it works) if you are!

Libraries/Core/InitializeCore.js
yungsters: This was a change to Haste (or `jest-haste[-map]`) about a year ago. It is configured to use file base names as haste identifiers instead.

Libraries/vendor/core/_wrapObjectFreezeAndFriends.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @author Ben Newman (@benjamn) <ben@benjamn.com>
```
yungsters: Can you enable Prettier for this file with `@format`?

Libraries/vendor/core/Map.js
```diff
@@ -607,7 +620,15 @@ module.exports = (function(global, undefined) {
         }
         return hashCounter;
       } else {
-        throw new Error('Non-extensible objects are not allowed as keys.');
+        // If the object is not extensible, fall back to storing it in an
+        // array and using Array.prototype.indexOf to find it.
+        let index = nonExtensibleObjects.indexOf(o);
+        if (index < 0) {
+          index = nonExtensibleObjects.length;
+          nonExtensibleObjects[index] = o;
```
pvdz: I'm slightly worried about a memory leak here. I think it's fine since the temporary arrays should not be retained once the map is released, but I'm not certain.

Libraries/vendor/core/_wrapObjectFreezeAndFriends.js
```diff
@@ -0,0 +1,35 @@
...
+  const method = Object[name];
+  if (typeof method === "function") {
+    Object[name] = function (obj) {
+      try {
+        // If .set succeeds, also call .delete to avoid leaking memory.
+        getTestMap().set(obj, obj).delete(obj);
+      } finally {
+        // If .set fails, the exception will be silently swallowed
+        // by this return-from-finally statement, and the method will
+        // behave exactly as it did before it was wrapped.
```
pvdz: ... with the potential of leaking memory, unreported, if the error happens in the `.delete()` portion of things.

Hypothetically; if something wants to call freeze or seal on an object, how likely is it that it actually wouldn't want the `__MAP_POLYFILL_INTERNAL_HASH__` expando applied to the object itself, in the first place? Wouldn't we be better and safer off letting it use the (new-in-this-PR) fallback method instead?

Libraries/vendor/core/Map.js
```diff
@@ -607,7 +620,15 @@ module.exports = (function(global, undefined) {
         }
```
pvdz: (_The error above this line, to which github is not allowing me to comment;_)

This could also fallback to the solution below.

Libraries/vendor/core/_wrapObjectFreezeAndFriends.js
benjamn: > Wouldn't we be better and safer off letting it use the (new-in-this-PR) fallback method instead?

No, because the fallback method has linear lookup time, compared to the tagging approach, which has constant lookup time. That's a nice advantage of this polyfill over other (better tested, more widely used) polyfills that settle for linear time in order to avoid tagging objects. If we don't care about the performance benefit of tagging, then we should really replace this whole polyfill with one that wasn't invented here.

benjamn: I'm confident that `.delete(obj)` will always succeed immediately after `.set(obj, obj)` succeeds in putting `obj` into the map. If that's not the case, there's a bug, and we will fix it.

Libraries/vendor/core/Map.js
benjamn: I don't think enough objects will ever be added to this array to matter from a memory perspective, and I'm worried that removing and re-adding objects will give them different hashes over time. Since these hashes are just made-up (not based on the memory address of the object), and these objects are non-extensible, there's no way to give them the same hash every time without keeping them in memory somewhere.

benjamn: I could be convinced that changing hashes doesn't matter in this case, but it worries me.

Libraries/vendor/core/_wrapObjectFreezeAndFriends.js
benjamn: Sure.

Libraries/vendor/core/Map.js
benjamn: Good idea!

yungsters:
What's the status of this pull request?

cpojer:
Also asking, @benjamn mind responding to the latest comments (adding @format)?

benjamn:
Rebased and addressed feedback from @pvdz and @yungsters. It's still possible that the `nonExtensibleObjects` array will retain objects that were frozen before `_wrapObjectFreezeAndFriends.js` was imported, but those objects should be exceedingly rare because the wrapping logic begins very early in the setup of the environment, and the memory will be freed whenever the `Map` is garbage collected.

Apologies to @format for the unintentional @-mention-spam.

benjamn:
Looks like **test_detox_end_to_end** is failing on `master` currently, in case that wasn't clear.

Issue #21539, opened 2018-10-06
Installation fails when path contains spaces due to using an obsolete version of detox


dantman:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
```
  React Native Environment Info:
    System:
      OS: macOS Sierra 10.12.6
      CPU: x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
      Memory: 1.05 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.11.0 - /usr/local/bin/node
      Yarn: 1.10.1 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.2, macOS 10.13, tvOS 11.2, watchOS 4.2
      Android SDK:
        Build Tools: 23.0.0, 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.2
        API Levels: 19, 23, 24, 25, 26, 27
    IDEs:
      Android Studio: 3.0 AI-171.4443003
      Xcode: 9.2/9C40b - /usr/bin/xcodebuild
    npmPackages:
      react: 16.5.0 => 16.5.0
      react-native: 0.57.2 => 0.57.2
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
```

## Description
Installation of react-native with yarn fails when the path to your project contains a space in it anywhere in the directory tree, with this error:
```
[3/3] ⠄ detox
error /Users/daniel/Workspace/work/Mind Monkey RN/node_modules/react-native/node_modules/detox: Command failed.
Exit code: 1
Command: node scripts/postinstall.js
Arguments:
Directory: /Users/daniel/Workspace/work/Mind Monkey RN/node_modules/react-native/node_modules/detox
Output:
+ xcodebuild -version
+++ dirname /Users/daniel/Workspace/work/Mind Monkey RN/node_modules/react-native/node_modules/detox/scripts/build_framework.ios.sh
usage: dirname path
[...]

PR #21570, opened 2018-10-08
Update to Detox 9

kelset:
Updating to Detox 9 - this should help with better Xcode 10 support, and should fix https://github.com/facebook/react-native/issues/21539.

I've updated to fixed `9.0.4` since it seems that each version is fairly different from the next, and I've followed the [migration guide](https://github.com/wix/detox/blob/master/docs/Guide.Migration.md#migrating-from-detox-8xx-to-9xx) for the android native dependencies.

Test Plan:
----------
Run against CI. Should not affect other builds.

Release Notes:
--------------
[INTERNAL] [ENHANCEMENT] [DETOX] - Update to version 9.0.4


hramos:
This should be landing soon.

PR #21575, opened 2018-10-08
[Android + Crossplatform][TextInput] Add autoComplete prop

cassiozen:
# Motivation
TL;DR: Setting `autoComplete` will allow the system to suggest autofill options for the `<TextInput>` component.

Android Oreo introduced the AutoFill Framework, for secure communication between an app and autofill services (e.g. Password managers). When using `<TextInput>` on Android Oreo+, the system already tries to autofill (based on heuristics), but there is no way to set configuring options or disable.

The quick solution would be to just add the same Android attributes (`autofillHints` & `importantForAutofill`) in React Native TextInput, but that doesn't bond well with the cross-platform nature of the library.

## This PR

Introduces an `autoComplete` prop based on HTML's `autocomplete` attribute, mapping to Android `autofillHints` & `importantForAutofill` and serving as a proper placeholder for autofill/autocomplete in other platforms:

Also gives you the ability to disable autofill by setting autocomplete="off".

# Test Plan
Adding autoComplete equal to username, password or any other available options should give you an autofill-bar bellow the TextInput which will let you fill in values from an AutoFill Service.

(The image show an example using Google's [sample AutoFill Service](https://github.com/googlesamples/android-AutofillFramework)

screen shot 2018-10-08 at 10 16 44 am

Setting the appropriate autoComplete will fill in the correct value in the TextInput.

Usage:

```js
<TextInput
    value={this.state.username}
    onChangeText={this.setUserName}
    autoComplete="username"
/>
<TextInput
    value={this.state.password}
    onChangeText={this.setPassword}
    secureTextEntry={true}
    autoComplete="password"
/>
```

To disable:

```js
<TextInput
    value={this.state.password}
    onChangeText={this.setPassword}
    secureTextEntry={true}
    autoComplete="off"
/>
```

# References

- Feature Request on React Native's Canny: https://react-native.canny.io/feature-requests/p/autocomplete-prop-for-textinput

- Android AutoFill Framework documentation: https://developer.android.com/guide/topics/text/autofill

- HTML autocomplete attribute spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls%3A-the-autocomplete-attribute

# Related PR:

Docs PR: facebook/react-native-website#606

Release Notes:
--------------

[ANDROID] [FEATURE] [AutoComplete] - Allow control over autofill suggestions for the `<TextInput>` component.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -162,6 +162,21 @@ type IOSProps = $ReadOnly<{|
 |}>;
 
 type AndroidProps = $ReadOnly<{|
+  autoCompleteType?: ?(
+    | 'username'
+    | 'password'
+    | 'email'
+    | 'name'
+    | 'tel'
+    | 'street-address'
+    | 'postal-code'
+    | 'cc-number'
+    | 'cc-csc'
+    | 'cc-exp'
+    | 'cc-exp-month'
+    | 'cc-exp-year'
+    | 'off'
+  ),
   returnKeyLabel?: ?string,
   numberOfLines?: ?number,
   disableFullscreenUI?: ?boolean,
@@ -349,6 +364,45 @@ const TextInput = createReactClass({
       'sentences',
       'words',
       'characters',
+    ]),
+    /**
+     * Determines which content to suggest on auto complete, e.g.`username`.
+     * To disable auto complete, use `off`.
+     *
+     * *Android Only*
+     *
+     * The following values work on Android only:
+     *
+     * - `username`
+     * - `password`
+     * - `email`
+     * - `name`
+     * - `tel`
+     * - `street-address`
+     * - `postal-code`
+     * - `cc-number`
+     * - `cc-csc`
+     * - `cc-exp`
+     * - `cc-exp-month`
+     * - `cc-exp-year`
+     * - `off`
+     *
+     * @platform android
+     */
+    autoCompleteType: PropTypes.oneOf([
+      'username',
+      'password',
+      'email',
+      'name',
+      'tel',
+      'street-address',
+      'postal-code',
+      'cc-number',
+      'cc-csc',
+      'cc-exp',
+      'cc-exp-month',
+      'cc-exp-year',
+      'off'
     ]),
     /**
      * If `false`, disables auto-correct. The default value is `true`.
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -402,7 +402,7 @@ const TextInput = createReactClass({
       'cc-exp',
       'cc-exp-month',
       'cc-exp-year',
-      'off'
+      'off',
     ]),
     /**
      * If `false`, disables auto-correct. The default value is `true`.
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -163,18 +163,18 @@ type IOSProps = $ReadOnly<{|
 
 type AndroidProps = $ReadOnly<{|
   autoCompleteType?: ?(
-    | 'username'
-    | 'password'
-    | 'email'
-    | 'name'
-    | 'tel'
-    | 'street-address'
-    | 'postal-code'
-    | 'cc-number'
     | 'cc-csc'
     | 'cc-exp'
     | 'cc-exp-month'
     | 'cc-exp-year'
+    | 'cc-number'
+    | 'email'
+    | 'name'
+    | 'password'
+    | 'postal-code'
+    | 'street-address'
+    | 'tel'
+    | 'username'
     | 'off'
   ),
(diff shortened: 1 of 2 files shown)


gaodeng:
+1

janicduplessis:
@cassiozen Thanks for the PR! We already have https://facebook.github.io/react-native/docs/textinput#textcontenttype for iOS, I think we could reuse this prop and make it cross platform. We probably just have to specify which values are platform specific.

cassiozen:
Another argument in favor of a cross-platform solution: iOS is introducing password recipes (and Android might do something similar in the future) - we could embrace all those solutions with a single, unified API.

nhayfield:
Hopefully this gets merged soon. I need it to fix  a whole host of bugs involving the autofill manager: 

```
android.os.RemoteException: Remote stack trace: at com.android.server.autofill.AutofillManagerServiceImpl.assertCallerLocked(AutofillManagerServiceImpl.java:552) at com.android.server.autofill.AutofillManagerServiceImpl.createSessionByTokenLocked(AutofillManagerServiceImpl.java:510) at com.android.server.autofill.AutofillManagerServiceImpl.startSessionLocked(AutofillManagerServiceImpl.java:375) at com.android.server.autofill.AutofillManagerService$AutoFillManagerServiceStub.startSession(AutofillManagerService.java:897) at android.view.autofill.IAutoFillManager$Stub.onTransact(IAutoFillManager.java:117) 
```

the-felipeal:
Hi there,

I'm the technical lead of the [Android Autofill Framework](https://developer.android.com/guide/topics/text/autofill). See my replies below..

> @cassiozen We already have https://facebook.github.io/react-native/docs/textinput#textcontenttype for iOS

When we designed the Autofill Framework, we consider using Android's InputType to convey the semantic of the field type, but we decide to add a new property (`autofillHints`) for 2 reasons:

1.The `inputType` is only associated with text fields, and the framework supports more types, so that information would not be available on those types (for example, a `Spinner`).
2.This "automatic" mapping is error prone. For example, an app could use a `inputType="textPassword"` in a CVC field, which would make its "hint"  to be _"password"_, not _"creditCardVerificationNumber"_

> I think we could reuse this prop and make it cross platform. 

I think an "autocomplete" property would be cross-platform as well, as this is a [W3C standard](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls%3A-the-autocomplete-attribute). In fact, for Android apps using `WebView`, the WebView implementation automatically maps the `autocomplete` tag to `autofillHints`.

-- Felipe

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -523,6 +523,39 @@ public void setMaxLength(ReactEditText view, @Nullable Integer maxLength) {
     view.setFilters(newFilters);
   }
 
+  @ReactProp(name = "autoComplete")
+  public void setTextContentType(ReactEditText view, @Nullable String autocomplete) {
+    if ("username".equals(autocomplete)) {
```
lostatseajoshua: You should check for null first. It is a valid type on the JS side to pass undefined or null. Since you don't check for it and return out it will cause an exception to be thrown. We should allow null and set auto fill to its default.

Libraries/Components/TextInput/TextInput.js
```diff
@@ -162,6 +162,21 @@ type IOSProps = $ReadOnly<{|
...
+    | 'name'
+    | 'tel'
+    | 'street-address'
+    | 'postal-code'
+    | 'cc-number'
+    | 'cc-csc'
+    | 'cc-exp'
+    | 'cc-exp-month'
+    | 'cc-exp-year'
+    | 'off'
```
lostatseajoshua: Nit: You should alpha order this list. Unless you have a specific reason for it being in this order. I would alpha order all values and then add off at the end for clear readability.

Libraries/Components/TextInput/TextInput.js
```diff
@@ -350,6 +365,45 @@ const TextInput = createReactClass({
...
+     * - `cc-csc`
+     * - `cc-exp`
+     * - `cc-exp-month`
+     * - `cc-exp-year`
+     * - `off`
+     *
+     * @platform android
+     */
+    autoCompleteType: PropTypes.oneOf([
+      'username',
```
lostatseajoshua: Nit: You could alpha order this last same as above.

hramos:
We noticed some crashes on pre-26 devices. We're rolling out a fix to prevent `setTextContentType` from calling the Oreo+ APIs on pre-26.

grabbou:
Going to revert manually in 0.60-stable now.

hramos:
Fix is in https://github.com/facebook/react-native/commit/d4aa1e7a52b51fa5d7fc9ded132b7b50170f2190.

PR #21585, opened 2018-10-09
[NativeMethodsMixin] CheckBox: Convert NativeMethodsMixin to forwardedRef, convert to class

empyrical:
This PR converts the use of `NativeMethodsMixin` in `CheckBox.android.js`, and converts it to an ES6-style class.

Test Plan:
----------
I have tested it in RNTester, and I ran in to no issues.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Components/CheckBox/CheckBox.android.js] - Convert NativeMethodsMixin to forwardedRef, converted to ES6 class

### DIFF
--- a/Libraries/Components/CheckBox/CheckBox.android.js
+++ b/Libraries/Components/CheckBox/CheckBox.android.js
@@ -10,7 +10,6 @@
 'use strict';
 
 const React = require('React');
-const ReactNative = require('ReactNative');
 const StyleSheet = require('StyleSheet');
 
 const requireNativeComponent = require('requireNativeComponent');
--- a/Libraries/Components/CheckBox/CheckBox.android.js
+++ b/Libraries/Components/CheckBox/CheckBox.android.js
@@ -192,8 +192,8 @@ const styles = StyleSheet.create({
 type CheckBoxType = Class<NativeComponent<Props>>;
 
 // $FlowFixMe - TODO T29156721 `React.forwardRef` is not defined in Flow, yet.
-const CheckBoxWithRef = React.forwardRef((props, ref) => (
-  <CheckBox {...props} forwardedRef={ref} />
-));
+const CheckBoxWithRef = React.forwardRef(function CheckBoxWithRef(props, ref) {
+  return <CheckBox {...props} forwardedRef={ref} />;
+});
 
 module.exports = (CheckBoxWithRef: CheckBoxType);
--- a/Libraries/Components/CheckBox/CheckBox.android.js
+++ b/Libraries/Components/CheckBox/CheckBox.android.js
@@ -15,10 +15,9 @@ const StyleSheet = require('StyleSheet');
 const requireNativeComponent = require('requireNativeComponent');
 const nullthrows = require('nullthrows');
 
-const RCTCheckBox = requireNativeComponent('AndroidCheckBox');
-
 import type {ViewProps} from 'ViewPropTypes';
 import type {SyntheticEvent} from 'CoreEventTypes';
+import type {NativeComponent} from 'ReactNative';
 
 type CheckBoxEvent = SyntheticEvent<
   $ReadOnly<{|
@@ -62,6 +61,18 @@ type Props = $ReadOnly<{|
    */
   ref?: ?Object,
 |}>;
+
+type NativeProps = $ReadOnly<{|
+  ...Props,
+  on?: ?boolean,
+  enabled?: boolean,
+|}>;
+
+type CheckBoxNativeType = Class<NativeComponent<NativeProps>>;
+
+const RCTCheckBox = ((requireNativeComponent(
+  'AndroidCheckBox',
+): any): CheckBoxNativeType);
 
 /**
  * Renders a boolean input (Android only).
@@ -119,8 +130,7 @@ type Props = $ReadOnly<{|
  * @keyword toggle
  */
 class CheckBox extends React.Component<Props> {
-  // $FlowFixMe How to type a native component to be able to call setNativeProps
-  _nativeRef: ?React.ElementRef<typeof RCTCheckBox> = null;
+  _nativeRef: ?React.ElementRef<CheckBoxNativeType> = null;
 
   static defaultProps = {
     value: false,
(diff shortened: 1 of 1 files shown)


Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -9,21 +9,70 @@
...
+  testID?: ?string,
+
+  /**
+   * Used to get the ref for the native checkbox
+   */
+  ref?: ?Object,
+|}>;
+
+type NativeProps = $ReadOnly<{|
+  ...Props,
```
TheSavior: I don't think this is true. The native component doesn't take `disabled` or `value` for example.

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -9,21 +9,70 @@
...
+
+  /**
+   * Used to locate this view in end-to-end tests.
+   */
+  testID?: ?string,
+
+  /**
+   * Used to get the ref for the native checkbox
+   */
+  ref?: ?Object,
```
TheSavior: I don't think we have a type for this yet in Flow but I'd rather it be `$FlowIssue` here than `Object.`

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -80,81 +135,59 @@ type DefaultProps = {
...
-    props.style = [styles.rctCheckBox, this.props.style];
+  render() {
+    const {disabled, value, ...props} = this.props;
+    const nativeProps = {
+      ...props,
+      onStartShouldSetResponder: () => true,
+      onResponderTerminationRequest: () => false,
+      enabled: !disabled,
+      on: value,
+      style: [styles.rctCheckBox, props.style],
```
RSNara: nit: We should probably also pull out the style, like how you pulled out `disabled` and `value`. 

```JavaScript
const {disabled, value, style, ...props} = this.props;
const nativeProps = {
  ...props
  // ...
  style: [styles.rctCheckBox, style],
}
```

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -80,81 +135,59 @@ type DefaultProps = {
...
+    // String-based refs cannot be shared.
+    if (typeof ref === 'function') {
+      ref(nativeRef);
+    } else if (typeof ref === 'object' && ref != null) {
+      ref.current = nativeRef;
+    }
+  };
+
+  _onChange = (event: CheckBoxEvent) => {
+    nullthrows(this._nativeRef).setNativeProps({value: this.props.value});
```
RSNara: Since the original code doesn't have `nullthrows`, I'm not sure if we should introduce it here.

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -9,21 +9,76 @@
...
+  /**
+   * Used to get the ref for the native checkbox
+   */
+  // $FlowFixMe: Properly type the ref prop type
+  ref?: $FlowIssue,
+|}>;
+
+type Props = $ReadOnly<{|
+  ...ViewProps,
+  ...CommonProps,
```
RSNara: nit: `ViewProps` are also common props. So, shouldn't we spread `ViewProps` into `CommonProps`?

```JavaScript
type Props = $ReadOnly<{|
  ...CommonProps
  // ...
```

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -80,81 +135,59 @@ type DefaultProps = {
...
-      disabled: false,
-    };
-  },
+  static defaultProps = {
+    value: false,
+    disabled: false,
+  };
+
+  _setAndForwardRef = nativeRef => {
+    const {ref} = this.props;
```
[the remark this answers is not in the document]
TheSavior: Yeah, it has to be passed in as a prop of a different name than ref (like forwardedRef)

TheSavior:
@empyrical, you did a bunch of work to validate that this ref set and forwarding work. Do you think you could add some unit tests for this component that check that behavior? It would be super valuable to set a good example for other components we will want to be updating as well.

empyrical:
Do you think a new section to the RNTester page that uses its ref would cover it? (like a `<Button />` that toggles the state of the checkbox, perhaps)

I have a PR coming up that includes a better way to handle these forwarded refs in classes, and would like to rebase this and address the code review points once it makes its way in. (Will also rename the ref prop to `forwardedRef`)

The `nullthrows` thing I did because the type of the ref is nullable, and Flow understandably does not like trying to access a method on it without a null check. This is also how `RefreshControl` does it:

https://github.com/facebook/react-native/blob/95174d4ea8d050305885f3cc2a4c99e2b83fc961/Libraries/Components/RefreshControl/RefreshControl.js#L159-L161

And I didn't do `this._nativeRef && this._nativeRef.method()` because if the ref handling broke at some point, then it would be a silent failure.

TheSavior:
I think you can test these things using a Jest test which would be run fast and in an automated way. I think this would be better here than an RNTester example.

Libraries/Components/CheckBox/CheckBox.android.js
empyrical: 👍

Libraries/Components/CheckBox/CheckBox.android.js
empyrical: 👍

empyrical:
Rebased this and changed it to properly forward refs using `setAndForwardRef`.

Libraries/Components/CheckBox/CheckBox.android.js
```diff
@@ -80,85 +135,65 @@ type DefaultProps = {
...
-
-  getDefaultProps: function(): DefaultProps {
-    return {
-      value: false,
-      disabled: false,
-    };
-  },
+class CheckBox extends React.Component<Props> {
+  _nativeRef: ?React.ElementRef<CheckBoxNativeType> = null;
+  _setNativeRef = setAndForwardRef({
```
TheSavior: We should flow type `setAndForwardRef` to ensure that `getForwardedRef` and `setLocalRef` operate on the same ref type.

TheSavior: Separate is best. I just noticed as part of some weirdness in how it is being used here.

PR #21586, opened 2018-10-09
Allow CxxModules to implement methods with two callbacks

acoates-ms:
When writing a native module in C++ using CxxModule, its not currently possible to write async methods which take two callbacks, that shouldn't be projected to JS as a promise.  I hit this when trying to implement the AppState module in c++.  AppState has a single method on it, getCurrentAppState, which takes two callbacks (a success and a failure callback).

This change adds a couple of new CxxModule::Method ctors, which allow devs to specify that they want two callbacks, but not a promise.  This is done using an extra tag in the ctor, similar to how you specify that you want to make a synchronous method.  I used the existing AsyncTag to indicate that the 2 callback version should be async, and not a promise.

Test Plan:
----------
Implemented the AppState module using a CxxModule for an out of tree platform.  Added sample usage to the Sample CxxModule.

Release Notes:
--------------

[MINOR] [ENHANCEMENT] [CxxModule.h] - Allow CxxModules to implement functions which take two callbacks

### DIFF
--- a/ReactCommon/cxxreact/SampleCxxModule.cpp
+++ b/ReactCommon/cxxreact/SampleCxxModule.cpp
@@ -114,7 +114,7 @@ auto SampleCxxModule::getMethods() -> std::vector<Method> {
         sample_->hello();
         return nullptr;
       }, SyncTag),
-    Method("addIfPositiveAsPromise", [this](dynamic args, Callback cb, Callback cbError) {
+    Method("addIfPositiveAsPromise", [](dynamic args, Callback cb, Callback cbError) {
         auto a = jsArgAsDouble(args, 0);
         auto b = jsArgAsDouble(args, 1);
         if (a < 0 || b < 0) {
@@ -123,7 +123,7 @@ auto SampleCxxModule::getMethods() -> std::vector<Method> {
           cb({a + b});
         }
       }),
-    Method("addIfPositiveAsAsync", [this](dynamic args, Callback cb, Callback cbError) {
+    Method("addIfPositiveAsAsync", [](dynamic args, Callback cb, Callback cbError) {
         auto a = jsArgAsDouble(args, 0);
         auto b = jsArgAsDouble(args, 1);
         if (a < 0 || b < 0) {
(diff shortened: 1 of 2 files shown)


matthargett:
Looks pretty good to me, some questions:
* is the AppState module you reference using platform agnostic C++? If so, why not contribute that in this PR as well?
* getType is computed on each call, but is a Native Module even able to change this dynamically? You didn’t change this, but I wonder if this can be a constant calculated in the ctor and assigned to a field, in preparation for it being constexpr later.

acoates-ms:
AppState itself is pretty platform specific.  And the one I'm working on isn't one of the base RN platforms.
getType is only called in getMethods which should only be called once when the native module is preparing the config to send to JS.  So its probably not worth the optimization?

ReactCommon/cxxreact/SampleCxxModule.cpp
```diff
@@ -114,6 +114,25 @@ auto SampleCxxModule::getMethods() -> std::vector<Method> {
         sample_->hello();
         return nullptr;
       }, SyncTag),
+    Method("addIfPositiveAsPromise", [this](dynamic args, Callback cb, Callback cbError) {
```
fkgozali: Looks like the capture of `this` is actually not used, failing the build internally:

```
react-native-github/ReactCommon/cxxreact/SampleCxxModule.cpp:117:39: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
    Method("addIfPositiveAsPromise", [this](dynamic args, Callback cb, Callback cbError) {
                                      ^
react-native-github/ReactCommon/cxxreact/SampleCxxModule.cpp:126:37: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
    Method("addIfPositiveAsAsync", [this](dynamic args, Callback cb, Callback cbError) {
                                    ^
```

acoates-ms: Sure I can remove that.

PR #21588, opened 2018-10-09
[NativeMethodsMixin] ProgressViewIOS: Remove PropTypes and NativeMethodsMixin, convert to functional component

empyrical:
This PR converts `ProgressViewIOS` from a `createReactClass` component to a functional component, and removes the remaining proptypes. Its use of `NativeMethodsMixin` has been ported to a `forwardRef` to the native component.

Test Plan:
----------
I have used it in RNTester, and the example progress bars work fine.

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [Libraries/Components/ProgressViewIOS/ProgressViewIOS.ios.js] - Remove PropTypes and NativeEventsMixin, convert to functional component

### DIFF
--- a/Libraries/Components/ProgressViewIOS/ProgressViewIOS.ios.js
+++ b/Libraries/Components/ProgressViewIOS/ProgressViewIOS.ios.js
@@ -21,8 +21,6 @@ import type {ImageSource} from 'ImageSource';
 import type {ColorValue} from 'StyleSheetTypes';
 import type {ViewProps} from 'ViewPropTypes';
 
-const RCTProgressView = requireNativeComponent('RCTProgressView');
-
 type Props = $ReadOnly<{|
   ...ViewProps,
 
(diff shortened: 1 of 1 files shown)


Libraries/Components/ProgressViewIOS/ProgressViewIOS.ios.js
```diff
@@ -29,70 +25,59 @@ const RCTProgressView = requireNativeComponent('RCTProgressView');
...
-      <RCTProgressView
-        {...this.props}
-        style={[styles.progressView, this.props.style]}
-      />
-    );
-  },
-});
+const ProgressViewIOS = (
+  props: Props,
+  forwardedRef?: ?React.Ref<'RCTProgressView'>,
```
TheSavior: I don't know if this is right.

empyrical: This is the way the other components (like `Text`) seem to do it - but I notice that `View` does it a better way, so I updated it to do it the way `View` does it. The other ones should be cleaned up too, perhaps!

Issue #21581, opened 2018-10-09
Convert components off of createReactClass


TheSavior:
As part of our goal to modernize and type our core components, we need to migrate off of `createReactClass`. This task is to track the work to remove `createReactClass` from our codebase.

We can only convert off of `createReactClass` if the components don't use any mixins. [Step 1](https://github.com/facebook/react-native/issues/21342) and [Step 2](https://github.com/facebook/react-native/issues/21485) were removing a few of the mixins. There are still a few mixins left and we will need to completely finish that before we can complete this issue. 

However, we can at least get started by removing `createReactClass` from the components that currently don't use any mixins. The following is that list:

- [x] ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
- [x] ReactAndroid/src/androidTest/js/UIManagerTestModule.js
- [x] RNTester/js/ListViewExample.js
- [x] RNTester/js/ImageExample.js
- [x] RNTester/js/TimerExample.js
- [x] RNTester/js/TextExample.ios.js
- [x] RNTester/js/ProgressBarAndroidExample.android.js
- [x] RNTester/js/PanResponderExample.js
- [x] RNTester/js/ListViewGridLayoutExample.js
- [x] RNTester/js/SwipeableFlatListExample.js
- [x] RNTester/js/ProgressViewIOSExample.js
- [x] RNTester/js/CameraRollView.js
- [x] IntegrationTests/LayoutEventsTest.js
- [x] IntegrationTests/ReactContentSizeUpdateTest.js
- [x] Libraries/Experimental/SwipeableRow/SwipeableRow.js

I got this list by running `grep -L "mixins:" $(grep -l -r createReactClass .)`. None of these files seem to be convertible by the [codemod provided by the react team](https://github.com/reactjs/react-codemod#explanation-of-the-new-es2015-class-transform-with-property-initializers).

I'm providing less instructions on this one because I think there are a lot of posts online about how to convert off of `createReactClass`.

peaonunes:
I will take `RNTester/js/TimerExample.js` :)

nd-02110114:
I will take `RNTester/js/ImageExample.js` and `RNTester/js/TextExample.ios.js`

sopranolinist:
I would like to reserve ScrollViewTestModule.js and UIManagerTestModule.js, will be able to start later this evening sometime.

himanshusoni:
I will take `SwipeableFlatListExample.js` and `ProgressViewIOSExample`

[...]

PR #21600, opened 2018-10-09
[createReactClass] Converting ProgressbarAndroid

gugakatsi:
[Android] [RNTester] converted ProgressbarAndroid.js to class 
but you own build is failing

relates to #21581

### DIFF
--- a/RNTester/js/ProgressBarAndroidExample.android.js
+++ b/RNTester/js/ProgressBarAndroidExample.android.js
@@ -16,33 +16,31 @@ var createReactClass = require('create-react-class');
 var RNTesterBlock = require('RNTesterBlock');
 var RNTesterPage = require('RNTesterPage');
 
-var MovingBar = createReactClass({
-  displayName: 'MovingBar',
-  _intervalID: (null: ?IntervalID),
+class MovingBar extends React.Component<{}> {
+  displayName = 'MovingBar';
+  _intervalID = IntervalID ? IntervalID : null;
 
-  getInitialState: function() {
-    return {
-      progress: 0,
-    };
-  },
+  state = {
+    progress: 0,
+  }
 
-  componentDidMount: function() {
+  componentDidMount() {
     this._intervalID = setInterval(() => {
       var progress = (this.state.progress + 0.02) % 1;
       this.setState({progress: progress});
     }, 50);
-  },
+  }
 
-  componentWillUnmount: function() {
+  componentWillUnMount() {
     if (this._intervalID != null) {
       clearInterval(this._intervalID);
     }
-  },
+  }
 
-  render: function() {
+  render() {
     return <ProgressBar progress={this.state.progress} {...this.props} />;
-  },
-});
+  }
+}
 
 class ProgressBarAndroidExample extends React.Component<{}> {
   static title = '<ProgressBarAndroid>';


RNTester/js/ProgressBarAndroidExample.android.js
```diff
@@ -16,33 +16,31 @@ var createReactClass = require('create-react-class');
 var RNTesterBlock = require('RNTesterBlock');
 var RNTesterPage = require('RNTesterPage');
 
-var MovingBar = createReactClass({
-  displayName: 'MovingBar',
-  _intervalID: (null: ?IntervalID),
+class MovingBar extends React.Component<{}> {
+  displayName = 'MovingBar';
+  _intervalID = IntervalID ? IntervalID : null;
```
TheSavior: I don't think this is valid. I think it should be

```
_intervalID = (null: ?IntervalID)
```

RNTester/js/ProgressBarAndroidExample.android.js
```diff
@@ -16,33 +16,31 @@ var createReactClass = require('create-react-class');
 var RNTesterBlock = require('RNTesterBlock');
 var RNTesterPage = require('RNTesterPage');
 
-var MovingBar = createReactClass({
-  displayName: 'MovingBar',
-  _intervalID: (null: ?IntervalID),
+class MovingBar extends React.Component<{}> {
```
TheSavior: Can you make this an exact type so that people don't pass additional props? `{||}`

RSNara: @gugakatsi It looks like `MovingBar` just forwards the props to `ProgressBar`. So, `MovingBarProps` should probably be defined like this:

```typescript
// Get the props of ProgressBar
type ProgressBarProps = React.ElementConfig<typeof ProgressBar>

// Remove the `progress` prop from ProgressBarProp, and make the resultant props ReadOnly
type MovingBarProps = $ReadOnly<$Diff<ProgressBarProps, {
  progress: $ElementType<ProgressBarProps, 'progress'> 
}>>
```

RNTester/js/ProgressBarAndroidExample.android.js
RSNara: @gugakatsi `IntervalID` is a type (and not a JavaScript object). Since JavaScript doesn't understand Flow types, we can't assign Flow types to JavaScript properties, which is why this is invalid. Also, what we really want is what @TheSavior suggested.

RNTester/js/ProgressBarAndroidExample.android.js
```diff
@@ -16,33 +16,31 @@ var createReactClass = require('create-react-class');
...
+  displayName = 'MovingBar';
```
RSNara: I think this should be:
```typescript
static displayName = 'MovingBar';
```

But since we have a babel plugin to automatically generate the `displayName` from components, we don't really need this. So, you should get rid of this line entirely.

gugakatsi: @RSNara thanks for helping me out , will fix pull request soon , learned a lot.

gugakatsi:
@RSNara @TheSavior  thanks for helping and understanding , your replies helped me lot , will fix it in few hrs.

RSNara:
@gugakatsi, ping. It's been some time since there was activity on this PR. Do you still have time to move this forward?

RSNara:
Due to inactivity, I'm closing this PR.

PR #21619, opened 2018-10-10
[createReactClass] Remove createReactClass from CameraRollView

exced:
Related to #21581

Remove createReactClass from CameraRollView.

## Test Plan:

- All flow tests succeed.
- Run RNTester app, go to CameraViewExample component, scroll, update image size, groupTypes.

## Release Notes:

[GENERAL] [ENHANCEMENT] [CameraRollView.js] - rm createReactClass


RNTester/js/CameraRollView.js
```diff
@@ -25,105 +23,111 @@ var {
...
 
-var groupByEveryN = require('groupByEveryN');
-var logError = require('logError');
+const groupByEveryN = require('groupByEveryN');
+const logError = require('logError');
+
+import type {RNTesterProps} from 'RNTesterTypes';
+
+type Props = $ReadOnly<{|
+  ...RNTesterProps,
```
RSNara: How come you're spreading the `RNTesterProps` into `Props`?

RNTester/js/CameraRollView.js
```diff
@@ -225,11 +225,11 @@ var CameraRollView = createReactClass({
     });
 
     return <View style={styles.row}>{images}</View>;
-  },
+  };
 
-  _appendAssets: function(data: Object) {
-    var assets = data.edges;
-    var newState: Object = {loadingMore: false};
+  _appendAssets = (data: Object) => {
```
RSNara: This shouldn't be bound.

RNTester/js/CameraRollView.js
```diff
@@ -187,9 +191,9 @@ var CameraRollView = createReactClass({
         enableEmptySections
       />
     );
-  },
+  }
 
-  _rowHasChanged: function(r1: Array<Image>, r2: Array<Image>): boolean {
+  _rowHasChanged(r1: Array<Image>, r2: Array<Image>): boolean {
```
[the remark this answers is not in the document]
RSNara: Also, this doesn't have to be a method on the component class.

RNTester/js/CameraRollView.js
```diff
@@ -25,105 +23,123 @@ var {
...
   /**
    * This should be called when the image renderer is changed to tell the
    * component to re-render its assets.
    */
-  rendererChanged: function() {
-    var ds = new ListView.DataSource({rowHasChanged: this._rowHasChanged});
+  rendererChanged() {
+    const ds = new ListView.DataSource({rowHasChanged: rowHasChanged});
     this.state.dataSource = ds.cloneWithRows(
       // $FlowFixMe(>=0.41.0)
```
RSNara: I think the general rule of thumb should be: If you see a `$FlowFixMe`, try to see if you can remove it.

Pretty sure this is suppressing some valid flow errors. For example, `rowHasChanged` takes an array of `Image` components, but `this.state.assets` is of type `Array<GetPhotosEdge>`.

RNTester/js/CameraRollView.js
```diff
@@ -25,105 +23,123 @@ var {
...
+  componentDidMount() {
     this.fetch();
-  },
+  }
 
   /* $FlowFixMe(>=0.68.0 site=react_native_fb) This comment suppresses an error
    * found when Flow v0.68 was deployed. To see the error delete this comment
    * and run Flow. */
-  UNSAFE_componentWillReceiveProps: function(nextProps: {groupTypes?: string}) {
+  UNSAFE_componentWillReceiveProps(nextProps: {groupTypes?: string}) {
```
RSNara: Can we remove this `$FlowFixMe`?

RNTester/js/CameraRollView.js
```diff
@@ -25,105 +23,123 @@ var {
...
 
-var groupByEveryN = require('groupByEveryN');
-var logError = require('logError');
+const groupByEveryN = require('groupByEveryN');
+const logError = require('logError');
 
-var propTypes = {
+import type {GetPhotosEdge, GetPhotosReturn} from 'CameraRoll';
+
+const rowHasChanged = function(r1: Array<Image>, r2: Array<Image>): boolean {
```
RSNara: I think the flow type of this function is wrong. Shouldn't it be this👇🏼?

```JavaScript
function rowHasChanged(r1: Array<GetPhotosEdge>, r2: Array<GetPhotosEdge>) {}
```

RNTester/js/CameraRollView.js
```diff
@@ -225,11 +227,11 @@ var CameraRollView = createReactClass({
     });
```
RSNara: Can we get rid of this `$FlowFixMe`?

RNTester/js/CameraRollView.js
```diff
@@ -245,16 +247,16 @@ var CameraRollView = createReactClass({
     }
```
RSNara: Can we get rid of this `$FlowFixMe`?

Libraries/CameraRoll/CameraRoll.js
```diff
@@ -79,34 +79,36 @@ const getPhotosParamChecker = createStrictShapeTypeChecker({
...
-      },
-      timestamp: number,
-      location?: {
-        latitude?: number,
-        longitude?: number,
-        altitude?: number,
-        heading?: number,
-        speed?: number,
-      },
+export type GetPhotosEdge = {
```
RSNara: What do you think about renaming this to:
```JavaScript
export type PhotoIdentifierObject = {
```

Or:
```JavaScript
export type PhotoIdentifier = {
```

Libraries/CameraRoll/CameraRoll.js
```diff
@@ -79,34 +79,36 @@ const getPhotosParamChecker = createStrictShapeTypeChecker({
...
+      latitude?: number,
+      longitude?: number,
+      altitude?: number,
+      heading?: number,
+      speed?: number,
+    },
+  },
+};
+
+export type GetPhotosReturn = {
```
RSNara: What do you think about renaming this to:

```JavaScript
export type PhotoIdentifierObjectsPage = {
```

Or: 

```JavaScript
export type PhotoIdentifiersPage = {
```

RNTester/js/CameraRollView.js
exced: Not sure how we can remove it.

This is the Flow error I get :
`Cannot extend property Component [1] with CameraRollView because property groupTypes is read-only in object type [2] but
writable in object type [3] in the first argument of property UNSAFE_componentWillReceiveProps.`

exced:
The Props "sub" type representation, the actual prop types object (wrapped by $ReadOnly<>)

RSNara:
Ah, that was no biggie. I fixed that on my end. 😁

RSNara:
Closing manually since this has been merged.

PR #21621, opened 2018-10-10
[createReactClass] Remove createReactClass from LayoutEventsTest

exced:
Related to #21581

Remove createReactClass from LayoutEventsTest and refactor a little bit.

## Test Plan:

- All flow tests succeed.
- `yarn run test-android-instrumentation` && CI integration tests ? Did not run the command yet

## Release Notes:

[GENERAL] [ENHANCEMENT] [LayoutEventsTest.js] - rm createReactClass


IntegrationTests/LayoutEventsTest.js
```diff
@@ -36,56 +35,80 @@ type State = {
...
-    return {
-      didAnimation: false,
-    };
-  },
-  animateViewLayout: function() {
+class LayoutEventsTest extends React.Component<$FlowFixMeProps, State> {
+  _view: ?View;
+  _img: ?Image;
+  _txt: ?Text;
+  _isMounted: boolean = true;
```
RSNara: Why are you setting `_isMounted` to true initially, instead of doing this:

```JavaScript
  _isMounted: boolean = false;
  componentDidMount() {
    this._isMounted = true;
  }
```

IntegrationTests/LayoutEventsTest.js
```diff
@@ -36,56 +35,80 @@ type State = {
...
     this.setState({containerStyle: {width: 280}}, () =>
       this.checkLayout(TestModule.markTestCompleted),
     );
-  },
-  checkLayout: function(next?: ?Function) {
-    if (!this.isMounted()) {
+  };
+
+  checkLayout = (next?: ?Function) => {
+    if (!this._isMounted) {
```
RSNara: Hmm... It seems like we were using the `isMounted()` check to ensure that `this.refs.{view,txt,img}`were all non-null. I wonder if we should replace this check with: `this._view == null || this._txt == null || this._img == null`.

```JavaScript
  if (this._view == null || this._txt == null || this._img == null) {
    return;
  }
  
  this._view.measure((x, y, width, height) => {
    this.compare('view', {x, y, width, height}, this.state.viewLayout);
    if (typeof next === 'function') {
      next();
    } else if (!this.state.didAnimation) {
      // Trigger first state change after onLayout fires
      this.animateViewLayout();
      this.state.didAnimation = true;
    }
  });
  this._txt.measure((x, y, width, height) => {
    this.compare('txt', {x, y, width, height}, this.state.textLayout);
  });
  this._img.measure((x, y, width, height) => {
    this.compare('img', {x, y, width, height}, this.state.imageLayout);
  });  

```

If we do this, we could get rid of `isMounted` completely.

IntegrationTests/LayoutEventsTest.js
exced: Because... there's no reason, it's a mistake !

IntegrationTests/LayoutEventsTest.js
exced: Since it's only used here I think we can get rid of isMounted, I'll update it

exced:
@RSNara I removed _isMounted variable

IntegrationTests/LayoutEventsTest.js
```diff
@@ -36,56 +35,75 @@ type State = {
...
-    this.refs.txt.measure((x, y, width, height) => {
-      this.compare('txt', {x, y, width, height}, this.state.textLayout);
-    });
-    this.refs.img.measure((x, y, width, height) => {
-      this.compare('img', {x, y, width, height}, this.state.imageLayout);
-    });
-  },
-  compare: function(node: string, measured: any, onLayout: any): void {
+
+    if (this._view) {
```
RSNara: I think we can (and should) get rid of these guards.

IntegrationTests/LayoutEventsTest.js
```diff
@@ -36,56 +35,75 @@ type State = {
...
 
-const LayoutEventsTest = createReactClass({
-  displayName: 'LayoutEventsTest',
-  getInitialState(): State {
-    return {
-      didAnimation: false,
-    };
-  },
-  animateViewLayout: function() {
+class LayoutEventsTest extends React.Component<$FlowFixMeProps, State> {
```
RSNara: If we look at `react-native/IntegrationTests/IntegrationTestsApp.js`, it looks like this component is rendered with no props. So, we can remove this `$FlowFixMe`! Should have caught this earlier. 😅

```JavaScript
type Props = $ReadOnly<{||}>

class LayoutEventsTest extends React.Component<Props, State> {
```

IntegrationTests/LayoutEventsTest.js
```diff
@@ -36,56 +35,75 @@ type State = {
...
+  _view: ?View;
+  _img: ?Image;
+  _txt: ?Text;
+
+  state: State = {
+    didAnimation: false,
+  };
+
+  animateViewLayout = () => {
```
RSNara: This doesn't need to be auto-bound.

exced:
@RSNara I updated bound function, types and guards. For the guards, I used the && syntax instead, not sure if you are big fans of it though

RSNara:
Strange. This was merged merged, so I'm closing this PR.

PR #21627, opened 2018-10-10
[createReactClass] remove createReactClass from ScrollViewTestModule.js

sopranolinist:
Related to issue #21581 
Removed createReactClass from ReactAndroid/src/androidTest/js/ScrollViewTestModule.js

Test Plan:
----------
- [x]   npm run prettier
- [x]   npm run flow-check-ios
- [x]   npm run flow-check-android

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [ScrollViewTestModule.js] - rm createReactClass

### DIFF
--- a/ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
+++ b/ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
@@ -36,46 +36,44 @@ class Item extends React.Component {
   }
 }
 
-class ScrollViewTestApp extends React.Component {
-  constructor() {
-    super();
-    this.scrollView = React.createRef();
+const getInitialState = function() {
+  var data = [];
+  for (var i = 0; i < NUM_ITEMS; i++) {
+    data[i] = {text: 'Item ' + i + '!'};
   }
-
-  getInitialState = () => {
-    var data = [];
-    for (var i = 0; i < NUM_ITEMS; i++) {
-      data[i] = {text: 'Item ' + i + '!'};
-    }
-    return {
-      data: data,
-    };
+  return {
+    data: data,
   };
+};
 
-  onScroll = e => {
-    ScrollListener.onScroll(
-      e.nativeEvent.contentOffset.x,
-      e.nativeEvent.contentOffset.y,
-    );
-  };
+const onScroll = function(e) {
+  ScrollListener.onScroll(
+    e.nativeEvent.contentOffset.x,
+    e.nativeEvent.contentOffset.y,
+  );
+};
 
-  onItemPress = itemNumber => {
-    ScrollListener.onItemPress(itemNumber);
-  };
+const onScrollBeginDrag = function(e) {
+  ScrollListener.onScrollBeginDrag(
+    e.nativeEvent.contentOffset.x,
+    e.nativeEvent.contentOffset.y,
+  );
+};
 
-  onScrollBeginDrag = e => {
-    ScrollListener.onScrollBeginDrag(
-      e.nativeEvent.contentOffset.x,
-      e.nativeEvent.contentOffset.y,
-    );
-  };
+const onScrollEndDrag = function(e) {
+  ScrollListener.onScrollEndDrag(
+    e.nativeEvent.contentOffset.x,
+    e.nativeEvent.contentOffset.y,
+  );
+};
 
-  onScrollEndDrag = e => {
-    ScrollListener.onScrollEndDrag(
-      e.nativeEvent.contentOffset.x,
-      e.nativeEvent.contentOffset.y,
-    );
-  };
+const onItemPress = function(itemNumber) {
+  ScrollListener.onItemPress(itemNumber);
+};
+
+class ScrollViewTestApp extends React.Component {
+  scrollView = React.createRef();
+  state = getInitialState();
 
   scrollTo = (destX, destY) => {
     this.scrollView.scrollTo(destY, destX);
@@ -87,14 +85,14 @@ class ScrollViewTestApp extends React.Component {
       <Item
         key={index}
         text={item.text}
-        onPress={this.onItemPress.bind(this, index)}
+        onPress={onItemPress.bind(this, index)}
       />
     ));
     return (
       <ScrollView
-        onScroll={this.onScroll}
-        onScrollBeginDrag={this.onScrollBeginDrag}
-        onScrollEndDrag={this.onScrollEndDrag}
+        onScroll={onScroll}
+        onScrollBeginDrag={onScrollBeginDrag}
+        onScrollEndDrag={onScrollEndDrag}
         ref={this.scrollView}>
         {children}
       </ScrollView>
@@ -103,31 +101,8 @@ class ScrollViewTestApp extends React.Component {
 }
 
 class HorizontalScrollViewTestApp extends React.Component {
-  constructor() {
-    super();
-    this.scrollView = React.createRef();
-  }
-
-  getInitialState = () => {
-    var data = [];
-    for (var i = 0; i < NUM_ITEMS; i++) {
-      data[i] = {text: 'Item ' + i + '!'};
-    }
-    return {
-      data: data,
-    };
-  };
-
-  onScroll = e => {
-    ScrollListener.onScroll(
-      e.nativeEvent.contentOffset.x,
-      e.nativeEvent.contentOffset.y,
-    );
-  };
-
-  onItemPress = itemNumber => {
-    ScrollListener.onItemPress(itemNumber);
-  };
+  scrollView = React.createRef();
+  state = getInitialState();
 
   scrollTo = (destX, destY) => {
     this.scrollView.scrollTo(destY, destX);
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
 }
 
-var getInitialState = function() {
-  var data = [];
-  for (var i = 0; i < NUM_ITEMS; i++) {
-    data[i] = {text: 'Item ' + i + '!'};
+class ScrollViewTestApp extends React.Component {
+  constructor() {
+    super();
+    this.scrollView = React.createRef();
```
RSNara: nit: We should just use class properties for this.

```JavaScript
class ScrollViewTestApp extends React.Component {
  scrollView = React.createRef();
  //...
```

This way, we won't have to implement our own constructor.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
   }
 }
 
-var getInitialState = function() {
-  var data = [];
-  for (var i = 0; i < NUM_ITEMS; i++) {
-    data[i] = {text: 'Item ' + i + '!'};
+class ScrollViewTestApp extends React.Component {
+  constructor() {
+    super();
```
RSNara: React components are instantiated with arguments (the first of which is just `props`). To avoid potential problems, we should just forward them all to `React.Component` constructor.

```JavaScript
class ScrollViewTestApp extends React.Component {
  constructor(...args) {
    super(...args);
    // ...
  }
  // ...
}
```

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
-};
+  getInitialState = () => {
+    var data = [];
+    for (var i = 0; i < NUM_ITEMS; i++) {
+      data[i] = {text: 'Item ' + i + '!'};
+    }
+    return {
+      data: data,
+    };
+  };
```
RSNara: Classes that extend `React.Component` do not call `getInitialState` to get their initial state. With the current configuration, `this.state` will be `null` inside the `render()` method, which will raise an exception. Instead of implementing `getinitialState`, you should create the state and assign it to `this.state` inside the constructor. Since we support class properties, we should just do this:

```JavaScript
class ScrollViewTestApp extends React.Component {
  scrollView = React.createRef();
  state = getInitialState();
```

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
-    e.nativeEvent.contentOffset.x,
-    e.nativeEvent.contentOffset.y,
-  );
-};
+  onScroll = e => {
+    ScrollListener.onScroll(
+      e.nativeEvent.contentOffset.x,
+      e.nativeEvent.contentOffset.y,
+    );
+  };
```
RSNara: All functions need to meet two conditions before you should consider binding them:

1. You're using `this` inside the function.
2. You're passing the function by reference to component's prop or as a callback to some function. 

Since we're not using `this` inside `onScroll`, we don't need to bind this function. In fact, since we're not using `this`, we could leave `onScroll` as is (i.e: Implemented outside the `ScrollViewTestApp` class). 

Okay, lets just move this back to where it was before.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
 
-var onScrollEndDrag = function(e) {
-  ScrollListener.onScrollEndDrag(
-    e.nativeEvent.contentOffset.x,
-    e.nativeEvent.contentOffset.y,
-  );
-};
+  onItemPress = itemNumber => {
+    ScrollListener.onItemPress(itemNumber);
+  };
```
RSNara: Let's move this back to where it was before.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
 
-var onItemPress = function(itemNumber) {
-  ScrollListener.onItemPress(itemNumber);
-};
+  onScrollBeginDrag = e => {
+    ScrollListener.onScrollBeginDrag(
+      e.nativeEvent.contentOffset.x,
+      e.nativeEvent.contentOffset.y,
+    );
+  };
```
RSNara: Let's move this back to being outside the component.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -37,54 +36,52 @@ class Item extends React.Component {
...
-  onScroll: onScroll,
-  onItemPress: onItemPress,
-  onScrollBeginDrag: onScrollBeginDrag,
-  onScrollEndDrag: onScrollEndDrag,
+  onScrollEndDrag = e => {
+    ScrollListener.onScrollEndDrag(
+      e.nativeEvent.contentOffset.x,
+      e.nativeEvent.contentOffset.y,
+    );
+  };
```
RSNara: Let's move this out.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -98,24 +95,45 @@ var ScrollViewTestApp = createReactClass({
         onScroll={this.onScroll}
         onScrollBeginDrag={this.onScrollBeginDrag}
         onScrollEndDrag={this.onScrollEndDrag}
```
RSNara: I don't think there's a reason for these functions to be attached to the component instance. So, after we move the functions out of the `ScrollViewTestApp` class, lets change this to:

```javascript
  <Item
    // ...
    onScroll={onScroll}
    onScrollBeginDrag={onScrollBeginDrag}
    onScrollEndDrag={onScrollEndDrag}
   // ...
```

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -98,24 +95,45 @@ var ScrollViewTestApp = createReactClass({
...
-var HorizontalScrollViewTestApp = createReactClass({
-  displayName: 'HorizontalScrollViewTestApp',
-  getInitialState: getInitialState,
-  onScroll: onScroll,
-  onItemPress: onItemPress,
+class HorizontalScrollViewTestApp extends React.Component {
+  constructor() {
+    super();
+    this.scrollView = React.createRef();
+  }
```
RSNara:
```javascript
class HorizontalScrollViewTestApp extends React.Component {
  scrollView = React.createRef();
```

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -98,24 +95,45 @@ var ScrollViewTestApp = createReactClass({
...
+      data: data,
+    };
+  };
+
+  onScroll = e => {
+    ScrollListener.onScroll(
+      e.nativeEvent.contentOffset.x,
+      e.nativeEvent.contentOffset.y,
+    );
+  };
```
RSNara: There's no need to duplicate this function.

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -98,24 +95,45 @@ var ScrollViewTestApp = createReactClass({
...
-  },
+  getInitialState = () => {
+    var data = [];
+    for (var i = 0; i < NUM_ITEMS; i++) {
+      data[i] = {text: 'Item ' + i + '!'};
+    }
+    return {
+      data: data,
+    };
+  };
```
RSNara:
```javascript
class HorizontalScrollViewTestApp extends React.Component {
  scrollView = React.createRef();
  state = getInitialState()
```

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -125,12 +143,15 @@ var HorizontalScrollViewTestApp = createReactClass({
       />
     ));
     return (
-      <ScrollView horizontal={true} onScroll={this.onScroll} ref="scrollView">
+      <ScrollView
+        horizontal={true}
+        onScroll={this.onScroll}
```
RSNara:
```javascript
  <ScrollView
     horizontal={true}
     onScroll={onScroll}
```

RSNara:
This file should also be flow-typed. But it's up to you if you want to tackle that here, or tackle that at all. 😛

ReactAndroid/src/androidTest/js/ScrollViewTestModule.js
```diff
@@ -47,90 +46,84 @@ var getInitialState = function() {
...
-  onScrollEndDrag: onScrollEndDrag,
+class ScrollViewTestApp extends React.Component {
+  scrollView = React.createRef();
+  state = getInitialState();
 
-  scrollTo: function(destX, destY) {
-    this.refs.scrollView.scrollTo(destY, destX);
-  },
+  scrollTo = (destX, destY) => {
+    this.scrollView.scrollTo(destY, destX);
```
RSNara: I should have caught this earlier, but when you create a ref using `React.createRef`, you have to access what it's referring to using `ref.current`. So, this should be something like:

```JavaScript
const { scrollView: { current: scrollView } } = this;
if (scrollView == null) {
  return;
}
scrollView.scrollTo(destY, destX);
```

But don't worry about it. 😄 I'll make the fix internally.

PR #21628, opened 2018-10-10
bump Android NDK to r17c

dulmandakh:
bump Android NDK to r17c, which is only available revision to download for r17. And it includes bug fixes.

Test Plan:
----------
CI is green


dulmandakh:
@gengjiawen thank you 👍

PR #21634, opened 2018-10-10
remove ListView examples from RNTester

dulmandakh:
This PR removes ListView examples from RNTester. ListView is deprecated (https://github.com/facebook/react-native/commit/e90f5fa2630f8a89e15fa57c70ada83e75a20642), and I assume that it's safe to remove ListView examples from RNTester.

Test Plan:
----------
CI is green, and RNTester app works as expected.


dulmandakh:
I thought that someone would look into the application and code, maybe copy
and paste. So it might be better to remove examples before actually
removing the code.

>

dulmandakh:
@mdvacca please feel free to close the PR, if you feel it's not the time to remove the examples

TheSavior:
I’d like to leave the examples and tests until we actually remove it. That will help make sure it isn’t broken while deprecated.

PR #21702, opened 2018-10-10
Feature/replace create react class

ronal2do:
Related to #21581 .
Removed createReactClass from ReactAndroid/src/androidTest/js/UIManagerTestModule.js

Test Plan:
----------
yarn prettier
yarn flow-check-android

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [ReactAndroid/src/androidTest/js/UIManagerTestModule.js] - remove createReactClass dependency


ReactAndroid/src/androidTest/js/UIManagerTestModule.js
```diff
@@ -9,214 +9,240 @@
...
+const StyleSheet = require('StyleSheet');
+const View = require('View');
+const Text = require('Text');
+
+const renderApplication = require('renderApplication');
+
+type FlexTestAppProps = $ReadOnly<{||}>;
+class FlexTestApp extends React.Component<FlexTestAppProps> {
+  render() {
     return (
```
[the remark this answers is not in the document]
ronal2do: @RSNara Can you help me understand this problem with prettier and eslint?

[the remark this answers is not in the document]
RSNara: Super weird... I think the bot's just bugging out, so I wouldn't worry about it.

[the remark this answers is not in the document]
ronal2do: Thanks

ReactAndroid/src/androidTest/js/UIManagerTestModule.js
```diff
@@ -9,214 +9,248 @@
...
+class UpdatePositionInListTestApp extends React.Component<
+  UpdatePositionInListTestAppProps,
+  UpdatePositionInListTestAppState,
+> {
+  state = {
+    active: false,
+  };
+
+  flushUpdatePositionInList = () => this.setState({active: true});
+
```
RSNara: This seems like a mistake. 

Right now, this is creating a property on `UpdatePositionInListTestApp` instances called `flushUpdatePositionInList`, which isn't what the original code does. What we actually need is something like this:

```JavaScript
class UpdatePositionInListTestApp extends React.Component<
  UpdatePositionInListTestAppProps,
  UpdatePositionInListTestAppState,
> {
  state = {
    active: false,
  }

  constructor(...args) {
    super(...args);
    flushUpdatePositionInList = () => this.setState({active: true});
  }
```

ReactAndroid/src/androidTest/js/UIManagerTestModule.js
```diff
@@ -9,214 +9,248 @@
...
-  }),
-  getInitialState: function() {
-    flushUpdatePositionInList = () => this.setState({active: true});
-    return {active: false};
-  },
-  render: function() {
+let flushUpdatePositionInList = null;
+type UpdatePositionInListTestAppProps = $ReadOnly<{||}>;
+type UpdatePositionInListTestAppState = $ReadOnly<{|
+  active?: ?boolean,
```
TheSavior: Are you sure this is nullable? It looks like it defaults to false and is only ever set to true.

RSNara: It looks like we don't need these to be optional?

```JavaScript
type UpdatePositionInListTestAppState = $ReadOnly<{|
  active: boolean,
|}>
```

TheSavior: Heh, same comment as @RSNara

ReactAndroid/src/androidTest/js/UIManagerTestModule.js
ronal2do: Thanks, for your feedback, fixed.

ronal2do: @RSNara It seemed a bit confusing to me that the instance of this class is being created [here,](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/androidTest/java/com/facebook/react/tests/CatalystUIManagerTestCase.java#L176), I believe that we do not need method inside the constructor.


```
  public void testUpdatePositionInList() {
    FrameLayout rootView = createRootView();
    jsModule.renderUpdatePositionInListTestApplication(rootView.getId());
    waitForBridgeAndUIIdle();
...

    // Let's make the second elements 50px height instead of 10px
    jsModule.flushUpdatePositionInList();
    waitForBridgeAndUIIdle();
....
  }

```
And now I'm not sure if we need this 
```
  flushUpdatePositionInList() {
    flushUpdatePositionInList();
  },
```
Let me know what do you think, to help me.

RSNara: I'm afraid I don't understand where you're confused. 

That test looks like it renders the component, makes a few assertions, calls `flushUpdatePositionInList`, which updates the state and re-renders the component, and then makes a few more assertions. 

We need to expose `flushUpdatePositionInList` because we need a way to re-render the component with new state after we make the first batch of assertions. 

We can't call `this.flushUpdatePositionInList();` inside the constructor, because (1) `setState` doesn't work inside the constructor and (2), even if it did work, we don't want to immediately re-render the component with the new state. Doing so would make the first batch of assertions fail in the test. 

I think there may have been some sort of miscommunication earlier. Let's change the code to this:

```JavaScript
let flushUpdatePositionInList = null;

// ...
class UpdatePositionInListTestApp extends React.Component<
  UpdatePositionInListTestAppProps,
  UpdatePositionInListTestAppState,
> {
  state = {
    active: false,
  }

  constructor(...args) {
    super(...args);
    flushUpdatePositionInList = () => this.setState({active: true});
  }
  // ...
}

const UITestManagerModule = {
  // ...
  flushUpdatePositionInList,
}
```

ronal2do: Thanks a lot, awesome

ronal2do:
@RSNara @TheSavior Thanks for your patience.

PR #21717, opened 2018-10-11
[Animated] check for undefined values in the config object for vector animations

adamjuhasz:
If you passed in a config object with an `undefined` value for XY animation, the destructuring of the config keys result in an exception since you can't destrcuture a null or undefined value.

Test Plan:
----------

- Test case added

Release Notes
--------------

[ GENERAL ] [ BUGFIX ] [ {Animated.spring} ] - Fix bug with config object for vector animations when a key has an undefined value

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -110,6 +110,7 @@ const maybeVectorAnim = function(
     const configX = {...config};
     const configY = {...config};
     for (const key in config) {
+      if (config[key] === undefined || config[key] === null) continue;
       const {x, y} = config[key];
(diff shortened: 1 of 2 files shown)


TheSavior:
I think I'd rather push people to pass in valid config than to silently handle invalid config. This could also be solved by using a stronger type for config than `Object` and to make the values not voidable.

adamjuhasz:
This actually came from writing in Reason where the optional parameters are turned into `undefined` values automatically before being passed to react native. I can attempt to fix this from the Reason project side by not creating the keys instead of setting them `undefined`, but I'm worried that some other projects would consider `undefined` values valid and pass them inside the config object.

PR #21730, opened 2018-10-11
[ANDROID][MINOR]: ReactAndroid - "compile" to "implementation"

ferrannp:
With this syncing with gradle gives 0 beautiful warnings.

# Test Plan:
- RNTester to sync without warnings.

# Release Notes
[ANDROID] [MINOR] - ReactAndroid - "compile" to "implementation" for `android-jsc`


ferrannp:
So without `compile` (deprecated) somehow tests are failing. I guess that's the reason it was not changed so probably not worth for now.

PR #21748, opened 2018-10-12
[createReactClass, TimerMixin] remove createReactClass and TImerMixin from TimerTest.js

ronal2do:
Related to https://github.com/facebook/react-native/issues/21485#issuecomment-428742440
Remove createReactClass and TimerMixin from IntegrationTests/TimersTest.js

Test Plan:
----------
Flow tests succeed

- [x] yarn prettier
- [x] yarn flow-check-android
- [x] yarn flow-check-ios

Release Notes:
--------------
[GENERAL] [ENHANCEMENT] [IntegrationTests/TimersTest.js] Remove createReactClass and TimerMixin


IntegrationTests/TimersTest.js
```diff
@@ -111,62 +119,77 @@ const TimersTest = createReactClass({
...
 
   render() {
     return (
       <View style={styles.container}>
         <Text>
-          {this.constructor.displayName + ': \n'}
+          {/* $FlowFixMe(>=0.54.0 site=react_native_fb,react_native_oss) This
+             * comment suppresses an error found when Flow v0.54 was deployed.
+             * To see the error delete this comment and run Flow. */
+          this.constructor.displayName + ': \n'}
```
RSNara: IIRC, the displayName is added on to our components in a babel transform. Flow doesn't know anything about that, so that's probably why it's raising this error. 

I think there're three fixes:
1. Change `this.constructor.displayName` to `this.constructor.name`. (Both should be equivalent).
2. Redundantly define the `displayName` static property in this class. 
3. Hardcode 'TimerTest' instead of `this.constructor.displayName`. 

Your call. 😄

IntegrationTests/TimersTest.js
```diff
@@ -111,62 +119,77 @@ const TimersTest = createReactClass({
...
-  },
+  }
 
   _incrementInterval() {
     if (this.state.count > 3) {
       throw new Error('interval incremented past end.');
     }
     if (this.state.count === 3) {
-      this.clearInterval(this._interval);
+      clearInterval(this._intervalId);
```
RSNara: The signature of `clearInterval` is this:
```JavaScript
function clearInterval(interval?: IntervalID) {
```
So `clearInterval` expects an argument that is either an `IntervalID`, `undefined`, or not present at all. You're passing in a value that's `?IntervalID`, which is equivalent to `IntervalID`, `undefined`, or `null`. (The bottom line  is that `clearInterval` does not accept `null` as an argument). Therefore, If you guard this function call with a null check, your flow error should go away. 

```JavaScript
if (this.state.count === 3) {
  if (this._intervalId != null) {
    clearInterval(this._intervalId);
  }
}
```

IntegrationTests/TimersTest.js
```diff
@@ -11,98 +11,107 @@
...
+type State = {|
+  count: number,
+  done: boolean,
+|};
+class TimersTest extends React.Component<Props, State> {
+  _nextTest = () => {};
+  _interval = -1;
+  _timeoutID: ?TimeoutID = null;
+  // $FlowFixMe
+  _intervalId: ?IntervalID = -1;
```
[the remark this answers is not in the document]
RSNara: I commented below how we can fix this.

IntegrationTests/TimersTest.js
```diff
@@ -11,98 +11,106 @@
 'use strict';
 
 const React = require('react');
-const createReactClass = require('create-react-class');
 const ReactNative = require('react-native');
 /* $FlowFixMe(>=0.54.0 site=react_native_oss) This comment suppresses an error
```
RSNara: Let's get rid of all `$FlowFixMe`s in this file.

IntegrationTests/TimersTest.js
```diff
@@ -111,62 +119,77 @@ const TimersTest = createReactClass({
...
 
   _fail(caller: string): void {
     throw new Error('_fail called by ' + caller);
-  },
-});
+  }
+}
 
 const styles = StyleSheet.create({
   container: {
```
RSNara: Let's get rid of: 
```JavaScript
TimersTest.displayName = 'TimersTest';
```

If we want to define static properties on `TimerTest`, we should do it using the `static` keyword inside the `TimerTest` class. That way, flow will be able to analyze the static property.

PR #21758, opened 2018-10-12
Fix npm commands in DockerTests.md

prog1dev:
[DOCS] [BUGFIX] [DockerTests.md] - Fix npm commands after https://github.com/facebook/react-native/pull/18597


RSNara:
Nice catch! 👌🏽

PR #21763, opened 2018-10-12
Add fix for refresh control state's race condition.

### LINKED ISSUE #21762: RCTRefreshControl.m contains race condition
## Environment
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz
      Memory: 56.93 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.12.0 - /usr/local/bin/node
      Yarn: 1.10.1 - /usr/local/bin/yarn
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 23.0.3, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3
        API Levels: 23, 25, 26
    IDEs:
    npmPackages:
      react-native: 0.57.3 => 0.57.3
    npmGlobalPackages:
      react-native-cli: 2.0.1

## Description
Current implementation of RCTRefreshControl contains race condition which results in incorrect state of refresh control. The worst case scenario is that refresh control displays ongoing state and it's not possible to cancel or re-dispatch onRefresh event. 

## Logical scenario
1. assign refreshing to true, which initiates beginRefreshingProgrammatically
2. before beginRefreshingProgrammatically animation ends, change refreshing state by assigning refreshing to false.
3 animation ends and overrides real state (refreshing: false) into previous state (refreshing: true)

rostislav-simonik:
Fixes #21762 

Test Plan:
----------
1. assign refreshing to true, which initiates beginRefreshingProgrammatically
2. before beginRefreshingProgrammatically animation ends, change refreshing state by assigning refreshing to false.
3. animation ends and overrides real state (refreshing: false) into previous state (refreshing: true)

Release Notes:
----------
[BUGFIX][iOS][RefreshControl] Fix race condition that would overwrite state changes

### DIFF
--- a/React/Views/RCTRefreshControl.m
+++ b/React/Views/RCTRefreshControl.m
@@ -12,6 +12,7 @@
 @implementation RCTRefreshControl {
   BOOL _isInitialRender;
   BOOL _currentRefreshingState;
+  UInt64 _currentRefreshingStateTimestamp;
   BOOL _refreshingProgrammatically;
   NSString *_title;
   UIColor *_titleColor;
@@ -21,6 +22,7 @@
 {
   if ((self = [super init])) {
     [self addTarget:self action:@selector(refreshControlValueChanged) forControlEvents:UIControlEventValueChanged];
+    _currentRefreshingStateTimestamp = 0;
     _isInitialRender = true;
     _currentRefreshingState = false;
   }
@@ -49,6 +51,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
 
 - (void)beginRefreshingProgrammatically
 {
+  UInt64 beginRefreshingTimestamp = _currentRefreshingStateTimestamp;
   _refreshingProgrammatically = YES;
   // When using begin refreshing we need to adjust the ScrollView content offset manually.
   UIScrollView *scrollView = (UIScrollView *)self.superview;
@@ -62,7 +65,10 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
                      animations:^(void) {
                        [scrollView setContentOffset:offset];
                      } completion:^(__unused BOOL finished) {
-                       [super beginRefreshing];
+                       if(beginRefreshingTimestamp == _currentRefreshingStateTimestamp) {
+                         [super beginRefreshing];
+                         [self setCurrentRefreshingState:super.refreshing];
+                       }
                      }];
 }
 
@@ -72,6 +78,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
   // endRefreshing otherwise the next pull to refresh will not work properly.
   UIScrollView *scrollView = (UIScrollView *)self.superview;
   if (_refreshingProgrammatically && scrollView.contentOffset.y < 0) {
+    UInt64 endRefreshingTimestamp = _currentRefreshingStateTimestamp;
     CGPoint offset = {scrollView.contentOffset.x, 0};
     [UIView animateWithDuration:0.25
                           delay:0
@@ -79,7 +86,10 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
                      animations:^(void) {
                        [scrollView setContentOffset:offset];
                      } completion:^(__unused BOOL finished) {
-                       [super endRefreshing];
+                       if(endRefreshingTimestamp == _currentRefreshingStateTimestamp) {
+                         [super endRefreshing];
+                         [self setCurrentRefreshingState:super.refreshing];
+                       }
                      }];
   } else {
     [super endRefreshing];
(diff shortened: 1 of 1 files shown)


matthargett:
can you add an e2e test?  also release notes like ```[BUGFIX][iOS][RefreshControl] Fix race condition that would overwrite state changes``` should suffice.

rostislav-simonik:
> can you add an e2e test? also release notes like `[BUGFIX][iOS][RefreshControl] Fix race condition that would overwrite state changes` should suffice.

- Release notes changed. 
- e2e test - Could you please give me an example, maybe another pull request. I've never done e2e test for this platform, but it's not problem to do it. Just need some directions.

cpojer:
Fails internally with:

```
error: unable to find developer root
react-native/React/Views/RCTRefreshControl.m:68:55: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
                       if(beginRefreshingTimestamp == _currentRefreshingStateTimestamp) {
                                                      ^
react-native/React/Views/RCTRefreshControl.m:89:53: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
                       if(endRefreshingTimestamp == _currentRefreshingStateTimestamp) {
                                                    ^
stderr: react-native/React/Views/RCTRefreshControl.m:68:55: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
                       if(beginRefreshingTimestamp == _currentRefreshingStateTimestamp) {
                                                      ^
react-native/React/Views/RCTRefreshControl.m:89:53: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
                       if(endRefreshingTimestamp == _currentRefreshingStateTimestamp) {
                                                    ^
```

Could you fix this up?

@matthargett do you know if there is a way to enable the same error on React Native GitHub? This seems to come up frequently.

rostislav-simonik:
> Alright, I'll land it. One thing to keep in mind here is because it is using local client time, changing the device time could break this behavior, no?

Yes, I've got idea later that could be done by logical clock instead.

rostislav-simonik:
> Fails internally with:
> 
> ```
> error: unable to find developer root
> react-native/React/Views/RCTRefreshControl.m:68:55: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
>                        if(beginRefreshingTimestamp == _currentRefreshingStateTimestamp) {
>                                                       ^
> react-native/React/Views/RCTRefreshControl.m:89:53: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
>                        if(endRefreshingTimestamp == _currentRefreshingStateTimestamp) {
>                                                     ^
> stderr: react-native/React/Views/RCTRefreshControl.m:68:55: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
>                        if(beginRefreshingTimestamp == _currentRefreshingStateTimestamp) {
>                                                       ^
> react-native/React/Views/RCTRefreshControl.m:89:53: error: block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior [-Werror,-Wimplicit-retain-self]
>                        if(endRefreshingTimestamp == _currentRefreshingStateTimestamp) {
>                                                     ^
> ```
> Could you fix this up?
Done.

cpojer:
Thank you so much for the changes!

@matthargett could you do a second review of this given that it is now using a slightly different approach? Thank you.

React/Views/RCTRefreshControl.m
```diff
@@ -132,9 +144,15 @@ - (void)setRefreshing:(BOOL)refreshing
   }
 }
 
+- (void)setCurrentRefreshingState:(BOOL)refreshing
+{
+  _currentRefreshingState = refreshing;
+  _currentRefreshingStateTimestamp = _currentRefreshingStateClock++;
```
matthargett: currentRefreshingStateClock is never decremented. Was this meant to track/prevent other potential races where it could go above 1 or 2?

rostislav-simonik: @matthargett no, it shouldn't because I used logical clock. Every time `refreshSetCurrentRefreshingState` is called i need unique timestamp. This time stamp is used in animation window which starts on 
https://github.com/facebook/react-native/pull/21763/files#diff-cfafa8d7e42ab87ab863f917e187a4aeR56
and when animation window ends 
https://github.com/facebook/react-native/pull/21763/files#diff-cfafa8d7e42ab87ab863f917e187a4aeR70
and timestamp is still the same, then state can be safely set, otherwise it should be thrown away due there is another animation in progress with higher clock timestamp.

rostislav-simonik: I've just changed that due i used local date time before, and as @cpojer has noted changing os date time can have impact on that logic. But in real life it wouldn't harm logic at all because chance that somebody changes os date time, and in that moment will trigger race condition described above, is astronomically low. But to prevent confusion, I've rid of that logic and replaced local date time with logical clock.

PR #21772, opened 2018-10-13
[TimerMixin] rm TimerMixin from TimersTest

exced:
Related to #21485

Remove TimerMixin from TimersTest

## Test Plan:
- All flow tests succeed
- `yarn run test-android-instrumentation` && CI integration tests ? (Did not run the command yet and not sure how to properly test it)

## Release Notes:

[GENERAL] [ENHANCEMENT] [TimersTest.js] - rm TimerMixin

### DIFF
--- a/IntegrationTests/TimersTest.js
+++ b/IntegrationTests/TimersTest.js
@@ -25,7 +25,7 @@ const TimersTest = createReactClass({
   _rafIDs: ([]: Array<AnimationFrameID>),
 
   _nextTest: () => {},
-  _interval: -1,
+  _interval: (null: ?IntervalID),
 
   getInitialState() {
     return {
@@ -45,7 +45,7 @@ const TimersTest = createReactClass({
     this._intervalIDs = [];
     this._immediateIDs.forEach(clearImmediate);
     this._immediateIDs = [];
-    this._rafIs.forEach(cancelAnimationFrame);
+    this._rafIDs.forEach(cancelAnimationFrame);
     this._rafIDs = [];
   },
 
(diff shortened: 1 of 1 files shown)


IntegrationTests/TimersTest.js
```diff
@@ -151,12 +164,75 @@ const TimersTest = createReactClass({
...
+      this._timeoutIDs.splice(this._timeoutIDs.indexOf(timeoutID));
+      fn();
+    }, ms);
+    this._timeoutIDs.push(timeoutID);
+    return timeoutID;
+  },
+
+  _setInterval: function(fn: () => void, ms?: number): IntervalID {
+    const intervalID = setInterval(() => {
+      this._intervalIDs.splice(this._intervalIDs.indexOf(intervalID));
```
RSNara: I think this is a bug? When the callback is first called, we'll clear the interval id from `this_intervalIDs`, which means that we'll never clear the interval in `componentWillUnmount`.

exced:
@RSNara You're write I shouldn't clear in _setInterval ! I fixed it

RSNara:
> @RSNara You're write I shouldn't clear in _setInterval ! I fixed it
> I also found out an issue with `splice` method !
> I still don't know if you prefer using `splice(index,1)` or `filter` (what's the most efficient method ?)
> (#21802 has the same issue)

Nice catch on the `Array.prototype.splice`! 

I think your call about using `Array.prototype.filter` is correct. It's less confusing than `Array.prototype.splice`.

RSNara:
@exced, I just noticed that #21748 does the exact same thing. I modified it to take the approach you took in this PR. Since that's already landing internally, I'll close this PR. Thanks for this PR! (And sorry that I didn't notice the conflict earlier 😢).

Issue #21639, opened 2018-10-10
TextInput duplicate value when inserting at the beginning (ios) maxLength


MadeinFrance:
* [x]  Review the documentation: https://facebook.github.io/react-native
* [x]  Search for existing issues: https://github.com/facebook/react-native/issues
* [x]  Use the latest React Native release: https://github.com/facebook/react-native/releases
 
## Environment

React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.10.0 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
   IDEs:
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.5.0 => 16.5.0 
      react-native: 0.57.2 => 0.57.2 
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
      react-native: 0.57.2

## Description
 [iOS only]

 When entering 1 character at the beginning of the TextInput, we expect to have only 1 character inserted. However we experience that the character is inserted and the current value is duplicated.  

 This behaviour is reproducible when the prop `maxLength` is used and when length of the current value is above the half of the `maxLength` prop.

 I have created an app from scratch using `create-react-native-app`, and I have modified `App.js` to include the following:

```js
type Props = {};
export default class App extends Component<Props> {
  constructor(props) {
    super(props);
    this.state = { text: '10000' };
  }

  render() {
    return (
      <View style={styles.container}>
        <TextInput
          style={{height: 40, width:120, borderColor: 'gray', borderWidth: 1}}
          maxLength={10}
          onChangeText={(text) => this.setState({text})}
          value={this.state.text}/>
      </View>
    );
  }
}
```
 
## Reproducible Demo
![demo](https://user-images.githubusercontent.com/1864359/46726492-c4881b80-cc7e-11e8-95d2-df954f515a9e.gif)

<hr>

## Core investigation

[...]

Issue #18374, opened 2018-03-14
[0.54] [iOS] Wrong 'Backspace' event fired onKeyPress after clearing TextInput


michalpetrov:
In IOS, after clearing TextInput value (setting it to empty string) and then pressing any key (e.g. 'a') two onKeyPress events are fired: A, Backspace. First is for the pressed key and second is a Backspace.

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.9.1
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.3.0-alpha.1 => 16.3.0-alpha.1
  react-native: 0.54.2 => 0.54.2

### Expected Behavior

Only one event for the pressed key should be fired (e. g. 'A').

### Actual Behavior

An additional 'Backspace' event is fired.

### Steps to Reproduce

1. write anythig in the textfield
2. press clear
3. press 'a' >>> events 'A' and 'Backspace' will be fired

```javascript
type Props = {};
type State = {
  text: string,
  keys: string
};
export default class App extends Component<Props, State> {
  state = {text: '', keys: ''}
  render() {
    return (
      <View style={styles.container}>
        <TextInput style={styles.textInput} value={this.state.text} onChangeText={this.onChangeText} onKeyPress={this.onKeyPress}/>
        <Button title="Clear" onPress={this.onClear}/>
        <Text>Text: {this.state.text}</Text>
        <Text>Keys: {this.state.keys}</Text>
      </View>
    );
  }

  onChangeText = (text: string) => {
    this.setState({text})
  }

  onKeyPress = ({ nativeEvent }: Object) => {
    this.setState({keys: this.state.keys + nativeEvent.key + ', '})
  }

  onClear = () => {
    this.setState({text: '', keys: ''})
  }
}
```

johnckendall:
+1, validated this is happening

edo1493:
+1

choiRyan:
I also encountered this. It causes react-native-tag-input@0.0.18 to not work properly. (Used the example code in the repo for replication) https://github.com/jwohlfert23/react-native-tag-input

hamaron:
+1; I've got the same issue on 0.54.

barsumek:
+1, encountered when using https://github.com/jwohlfert23/react-native-tag-input

[...]

PR #21798, opened 2018-10-15
TextInput component fix maxLength property when entering characters at the beginning.

MadeinFrance:
Fix TextInput component and the maxLength property when entering characters at the beginning. (#21639).

Improvements: In `TextInputTest.js` We could place the native cursor at the beginning of the input in the function `componentDidMount`. Not sure how to do that from the JavaScript.

Summary:
----------
- In RCTBaseTextInputView.m the ivar _predictedText is not updated correctly when inserting at the beginning.
- The actual fix is to update the value before the checking the text mismatch from `backedTextInputView`.

Test Plan:
----------
- When entering 1 character at the beginning of the TextInput, we expect to have only 1 character inserted. However we experience that the character is inserted and the current value is duplicated.
- Insert `0` character before the value `123` with: `<TextInput maxLength={5} />`.
broken-input

Release Notes:
--------------
[IOS] [BUGFIX] [TextInput] - Fix maxLength property when entering characters at the beginning.

Result:
--------------

ios-input-fixed

### DIFF
--- a/IntegrationTests/TextInputTest.js
+++ b/IntegrationTests/TextInputTest.js
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const React = require('react');
+const ReactNative = require('react-native');
+const {View, TextInput} = ReactNative;
+const {TestModule} = ReactNative.NativeModules;
+
+function expectTrue(condition: boolean, message: string): boolean {
+  if (!condition) {
+    throw new Error(message);
+  }
+
+  return condition;
+}
+
+class TextInputTest extends React.Component<{}, $FlowFixMeState> {
+  textInputRef: any;
+  state = {
+    text: '123',
+  };
+
+  componentDidMount() {
+    if (!TestModule.verifySnapshot) {
+      throw new Error('TestModule.verifySnapshot not defined.');
+    }
+
+    this.updateText();
+  }
+
+  updateText() {
+    // set native cursor to 0
+    this.textInputRef.setNativeProps({selection: {start: 0, end: 0}});
+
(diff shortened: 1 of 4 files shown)


IntegrationTests/TextInputTest.js
```diff
@@ -0,0 +1,67 @@
...
+  componentDidMount() {
+    if (!TestModule.verifySnapshot) {
+      throw new Error('TestModule.verifySnapshot not defined.');
+    }
+
+    this.updateText();
+  }
+
+  updateText() {
+    this.setState({text: '0'.concat(this.state.text)}, () => {
```
MadeinFrance: Improvement: Set the natively the cursor position at the beginning of the input field.

shergin:
@MadeinFrance Thank you!
Could you please describe *why* exactly your fix fixes things (doing them right)?

MadeinFrance:
Thank you for the feedback, I appreciate that.

Lets use the TestCase I described in the description (`<TextInput maxLength={5} defaultValue="123" />` and try to insert `0` at the beginning in order to have `0123` as value.) 

What happen when inserting `0`:

1. `RCTBackedTextInputDelegateAdapter.shouldChangeCharactersInRange` triggers and unfortunately override the ivar `_predictedText` to `0`.
2. At the same time, from `BackedInputDelegateAdapter`, `textFieldDidChange` delegate is called. 

The bug is that the `BaseTextInputView` [(line)](https://github.com/facebook/react-native/blob/master/Libraries/Text/TextInput/RCTBaseTextInputView.m#L369) will find a mismatch between `0123` and `0` and try to add a replacementText of `123` at `NSRange: location=1, length=0`.


3. The error result will be `01123`.

<hr>

- My current implementation keeps the ivar `_predictedText` to the correct value. 
- Another possible implementation including the clean up of (#18374) could be:

```objective-c
if (_predictedText) {
    _predictedText = [_predictedText stringByReplacingCharactersInRange:range withString:text];
}
```

MadeinFrance:
Fixed in a later [commit](https://github.com/facebook/react-native/commit/36507e4a3c2fa58dcfdc9bd389b37536c66006d0). Closing.

PR #21802, opened 2018-10-15
[FIX] ListView requestAnimationFrame leak

exced:
Related to https://github.com/facebook/react-native/pull/21488
Disclaimer: I made this PR.

I think there's some requestAnimationFrame events that are not cleared on unmount because of bad use of `splice` method.

## Test Plan:

- All flow tests succeed.
- RNTester: iOS (this change should only affect iOS because calculateChildFrames is iOS only)
Show perf monitor, show ListView* screen, start scrolling. UI frame Rate is used at the beginning. When scrolling there is no drop in FPS rate.
- TODO: I'll write a load test for ListView

## Release Notes:
[GENERAL] [ENHANCEMENT] [ListView.js] - rm TimerMixin

### DIFF
--- a/Libraries/Lists/ListView/ListView.js
+++ b/Libraries/Lists/ListView/ListView.js
@@ -544,10 +544,7 @@ const ListView = createReactClass({
 
   _requestAnimationFrame: function(fn: () => void): void {
     const rafId = requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


exced:
@RSNara Yep for sure, I find it clearer !

PR #21819, opened 2018-10-17
TextInput: Convert to ES6 Class-based component

empyrical:
This pull request converts `TextInput` from a `createReactClass`-based component to an ES6 class-based component. It also removes all remaining proptypes, and moves their documentation to the Flow declarations.

To cover the lack of `NativeMethodsMixin` here, I've introduced a new helper function called `setAndForwardRef` which allows a class based component, when used in conjunction with `React.forwardRef`, to forward a ref to user code while still being able to use the ref inside of the class. As a consequence of this, only function-based and `createRef`-based refs will work with this - string refs will throw.

Minor, but I also added documentation for `NativeComponent` in `ReactNativeTypes` while I was documenting the types.

Test Plan:
----------
I've tried this out in both Android and iOS RNTester, and nothing seems amiss in the behavior of this component.

A Jest-based test for `setAndForwardRef` has been added.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Components/TextInput/TextInput.js] - Converted to an ES6 class-based component. Moved documentation to the types.
[GENERAL] [MINOR] [Libraries/Renderer/shims/ReactNativeTypes.js] - Added documentation for `NativeComponent`
[GENERAL] [ENHANCEMENT] [Libraries/Utilities/__tests__/setAndForwardRef-test.js] - Added a test for `setAndForwardRef`
[GENERAL] [ENHANCEMENT] [Libraries/Utilities/setAndForwardRef.js] - Added `setAndForwardRef` - a helper function for class based components for forwarding a ref, while still keeping a local copy of the ref.
[GENERAL] [ENHANCEMENT] [RNTester/js/TextInputExample.android.js] - Converted string-based refs to functions and assorted cleanup
[GENERAL] [ENHANCEMENT] [RNTester/js/TextInputExample.ios.js] - Converted string-based refs to functions and assorted cleanup

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -9,13 +9,8 @@
  */
 'use strict';
 
-const DeprecatedColorPropType = require('DeprecatedColorPropType');
-const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
 const DocumentSelectionState = require('DocumentSelectionState');
-const EventEmitter = require('EventEmitter');
-const NativeMethodsMixin = require('NativeMethodsMixin');
 const Platform = require('Platform');
-const PropTypes = require('prop-types');
 const React = require('React');
 const ReactNative = require('ReactNative');
 const StyleSheet = require('StyleSheet');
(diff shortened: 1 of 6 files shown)


empyrical:
Yep, can do that!

PR #21837, opened 2018-10-17
Convert network overlay to use flatlist

AlanFoster:
Test Plan:
----------

I ran through the following scenarios:

- No Requests
- New Requests coming in, an ensuring that scroll to bottom works as expected
- Testing both XHR + WebSocket events
- Focusing on a particular request/response element
- Closing the detail view

network-update

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Network Inspector] - The network inspector has been updated to use FlatList, as ListView has been marked as deprecated.

### DIFF
--- a/Libraries/Inspector/NetworkOverlay.js
+++ b/Libraries/Inspector/NetworkOverlay.js
@@ -10,7 +10,7 @@
 
 'use strict';
 
-const ListView = require('ListView');
+const FlatList = require('FlatList');
 const React = require('React');
 const ScrollView = require('ScrollView');
 const StyleSheet = require('StyleSheet');
@@ -21,12 +21,12 @@ const WebSocketInterceptor = require('WebSocketInterceptor');
 const XHRInterceptor = require('XHRInterceptor');
 
 const LISTVIEW_CELL_HEIGHT = 15;
-const SEPARATOR_THICKNESS = 2;
 
 // Global id for the intercepted XMLHttpRequest objects.
 let nextXHRId = 0;
 
 type NetworkRequestInfo = {
+  id: number,
   type?: string,
   url?: string,
   method?: string,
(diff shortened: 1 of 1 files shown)


Libraries/Inspector/NetworkOverlay.js
```diff
@@ -216,10 +214,13 @@ class NetworkOverlay extends React.Component<
           return;
         }
         if (statusCode !== null && closeReason !== null) {
-          this._requests[socketIndex].status = statusCode;
-          this._requests[socketIndex].closeReason = closeReason;
+          this.setState(({requests}) => {
+            const networkRequestInfo = requests[socketIndex];
+            networkRequestInfo.status = statusCode;
+            networkRequestInfo.closeReason = closeReason;
+            return {requests};
```
AlanFoster: I'm not a massive fan of the localised state mutation, but it feels like an improvement to the existing code.
I assumed the previous mutation was for performance reasons, and I've aligned with the most simplistic approach for now.

AlanFoster:
@TheSavior I agree - it's quite the PR. The code in the file was previously quite complected, but thankfully it's a single component still!

> Can you break this into a couple different changes?
>
> - ListView -> FlatList
> - Changing styles
> - Changing the usage of setState

The main responsibilities are:

- The network handling logic is between lines 89-300, and is mostly just updating mutation of `this._requests` to instead be calls to `setState`
- The rest of the file is ListView -> FlatList, with the changes to styles being minimal with simple changes to align with using FlatList

I can try to extract this into two separate commits, but the commits wouldn't be atomic as they wouldn't work in isolation without lots of extra glue code to help with an intermediate step, which might make the review more confusing!

Let me know if the additional insights make this PR easier to review, or if you'd still like two separate commits :]

TheSavior:
Thanks for that explanation. I'm not sure I understand why the changes to network handling logic / setState are deeply coupled to ListView/FlatList. Can you help me understand this better? It seems like it should be possible to migrate to FlatList without also migrating the business logic very much.

AlanFoster:
> It seems like it should be possible to migrate to FlatList without also migrating the business logic very much.

@TheSavior That was my initial plan too, but once I learnt what the existing code was doing - it ended up easier to modify the code to be closer to idiomatic React. Let me know your thoughts after you've had a chance to review it :+1:

TheSavior:
I tried to import the PR to test it manually but noticed that your target branch is `0.57-stable`. Our tooling can only handle PRs onto master. Can you rebase?

AlanFoster:
@TheSavior Done! Apologies :)

PR #21849, opened 2018-10-18
use requireActual and requireMock from jest instead of require

thymikee:
A while back Jest introduced `jest.requireActual` and `jest.requireMock` which are aliases to `require.requireActual` and `require.requireMock`. We believe that users should use official Jest API and are planning to deprecate the latter.

Test Plan:
--------------

Only Jest-related changes, CI should be green.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [tests] - Use requireActual and requireMock from jest object instead of require.

### DIFF
--- a/jest/mockComponent.js
+++ b/jest/mockComponent.js
@@ -10,7 +10,7 @@
 'use strict';
 
 module.exports = (moduleName, instanceMethods) => {
-  const RealComponent = require.requireActual(moduleName);
+  const RealComponent = jest.requireActual(moduleName);
   const React = require('react');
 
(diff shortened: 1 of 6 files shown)


TheSavior:
Thanks for this. If you are curious, it looks like the majority of Facebook's callsites to `require.requireActual` is in the source to `react-relay`. A PR there would further help Facebook migrate off.

thymikee:
Sure, gonna update it there too, thanks!

thymikee:
Relay: https://github.com/facebook/relay/pull/2550

PR #21876, opened 2018-10-21
[createReactClass] Remove createReactClass from SwipeableRow

exced:
Related to #21581

Remove createReactClass from SwipeableRow.

## Test Plan:
- All flow tests succeed.
-RNTester > SwipeableFlatList, scroll, swipe left, swipe right
-RNTester > SwipeableListView, scroll, swipe left, swipe right

## Release Notes:

[GENERAL] [ENHANCEMENT] [SwipeableRow.js] - rm createReactClass

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableRow.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableRow.js
@@ -57,15 +57,15 @@ const RIGHT_SWIPE_THRESHOLD = 30 * SLOW_SPEED_SWIPE_FACTOR;
 type Props = $ReadOnly<{|
   children?: ?React.Node,
   isOpen?: ?boolean,
-  maxSwipeDistance: number,
-  onClose?: ?Function,
-  onOpen?: ?Function,
-  onSwipeEnd?: ?Function,
-  onSwipeStart?: ?Function,
+  maxSwipeDistance?: ?number,
+  onClose?: ?() => void,
+  onOpen?: ?() => void,
+  onSwipeEnd?: ?() => void,
+  onSwipeStart?: ?() => void,
   preventSwipeRight?: ?boolean,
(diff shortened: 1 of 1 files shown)


Libraries/Experimental/SwipeableRow/SwipeableRow.js
```diff
@@ -59,79 +57,57 @@ const RIGHT_SWIPE_THRESHOLD = 30 * SLOW_SPEED_SWIPE_FACTOR;
...
+  static defaultProps = {
+    isOpen: false,
+    preventSwipeRight: false,
+    maxSwipeDistance: 0,
+    onOpen: emptyFunction,
+    onClose: emptyFunction,
+    onSwipeEnd: emptyFunction,
+    onSwipeStart: emptyFunction,
+    swipeThreshold: 30,
+  };
```
RSNara: The problem with `defaultProps` is that it only defaults `undefined` props (and ignores when props are `null`). To make a component's API flexible, we should also default `null` props. This is why all of the optional props in this component are also nullable. However, the typings in this case are actually incorrect because we don't default nullable props. For example, `maxSwipeDistance` is nullable, so `defaultProps` doesn't default it to 0, but everywhere in the code, we assume that it's been defaulted to a number. 

Instead of relying on `defaultProps`, we should probably use the `??` operator:

```JavaScript
_animateToOpenPosition(): void {
  const maxSwipeDistance = this.props.maxSwipeDistance ?? 0;
  const directionAwareMaxSwipeDistance = IS_RTL ? -maxSwipeDistance : maxSwipeDistance;
  this._animateTo(-directionAwareMaxSwipeDistance);
},
```

What do you think about getting rid of `defaultProps` and using the `??` operator, which has this definition:
```JavaScript
  x ?? y === x != null ? x : y
```

PR #21880, opened 2018-10-21
Bump fbjs-scripts to ^1.0.0

jmheik:
This PR bumps also fbjs-scripts to latest version. Benefit is smaller node_modules and less deps to download as newer version doesn't depend on babel6 anymore.

Test Plan:
----------
All the test should be passing. As far as I see, only `fbjs-scripts/jest/createCacheKeyFunction` is used, and its implementation has not changed in a year.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] - Bump fbjs-scripts dependency to ^1.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,7 +165,7 @@
     "escape-string-regexp": "^1.0.5",
     "event-target-shim": "^1.0.5",
     "fbjs": "^1.0.0",
-    "fbjs-scripts": "^0.8.1",
+    "fbjs-scripts": "^1.0.0",
     "fs-extra": "^1.0.0",
     "glob": "^7.1.1",
     "graceful-fs": "^4.1.3",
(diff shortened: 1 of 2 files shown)


kelset:
Closing since this commit already accomplish this https://github.com/facebook/react-native/commit/ee034596fecfb47ff9e6e1fc30cefb0e970e7d80

jmheik:
@kelset Commit you linked updates fbjs, not fbjs-scripts. My motivation for this PR was to update fbjs-scripts as the old version transitively pulls lots of babel6 deps that are not needed.

hramos:
Thanks for the PR. Lorenzo is correct, this will require some work on our end to land since it touches `package.json`, and I can take care of it.

PR #21884, opened 2018-10-21
[ANDROID][BUGFIX][TextInput] - Fixed line not breaking when multiline set to true and keyboard type set to numeric (or other numeric type)

Muritavo:
Native Android sets the EditText widget to multiline only if the InputType flags are set to Text (or its variants like textEmailAddress) and Multiline, which causes the React Native TextInput to not break the line when set to multiline={true} and keyboardType={'numeric'} as it only have the flags Multiline and Number set.

This fix forces the widget to enable multiline, by calling setSingleLine(false) everytime a state change needs to be commited and the multiline prop is set to true.

Test Plan:
----------

This bug only occurs on Android devices

1. Create a TextInput component with multiline set to true and keyboardType set to 'numeric' or 'phone-pad' or any numeric type.
```javascript
<TextInput
   keyboardType={"numeric"}
   numberOfLines={6}
   multiline={true}/>
```

2. Input some text until it reaches the end of the component

3. See as you keep input more data, the line doesn't break

The test plan can be reproduced [HERE](https://snack.expo.io/@muritavo/textinput-numeric-type-bug)

Release Notes:
----------

[ANDROID] [BUGFIX] [TextInput] - Fixed line not breaking when multiline set to true and keyboard type set to numeric (or other numeric type)


RSNara:
@Muritavo, thanks for the PR!

One question: Are you sure that this is sufficient to fix the bug? If you replace `numberOfLines` with a really long placeholder, we still can't get multiline numeric `<TextInput/>`s. 🤔

```JavaScript
<TextInput
  multiline={true}
  keyboardType="numeric"
  style={styles.input}
  placeholder="some really really really really really really long text"
/>
```

Do you think that's a different problem? I'm asking because maybe there's a better solution that fixes both cases.

mmamaniwolox:
It would be awesome to have this working, great job @Muritavo . I hope this is merged soon :P Need it so bad haha

PR #21893, opened 2018-10-22
[createReactClass] remove createReactClass from ToolbarAndroid/ToolbarAndroid.android.js

nd-02110114:
Related to #21581 .
Removed createReactClass from the Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js

## Test Plan
- [x] npm run prettier
- [x] npm run flow-check-ios
## Release Notes
[GENERAL] [ENHANCEMENT] [Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js] - remove createReactClass dependency

### DIFF
--- a/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
+++ b/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
@@ -101,7 +101,7 @@ type ToolbarAndroidProps = $ReadOnly<{|
    * Callback that is called when an action is selected. The only argument that is passed to the
    * callback is the position of the action in the actions array.
    */
-  onActionSelected?: ?(postition: number) => void,
+  onActionSelected?: ?(position: number) => void,
   /**
    * Callback called when the icon is selected.
    */
--- a/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
+++ b/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
@@ -97,11 +97,11 @@ type Props = $ReadOnly<{|
    * Callback that is called when an action is selected. The only argument that is passed to the
    * callback is the position of the action in the actions array.
    */
-  onActionSelected?: ?Function,
+  onActionSelected?: ?(ToolbarAndroidChangeEvent) => void,
   /**
    * Callback called when the icon is selected.
    */
-  onIconClicked?: ?Function,
+  onIconClicked?: ?() => void,
   /**
    * Sets the overflow icon.
    */
@@ -168,9 +168,11 @@ class ToolbarAndroid extends React.Component<Props> {
   };
 
   render() {
-    const nativeProps = {
-      ...this.props,
-    };
+    const {
+      onIconClicked,
+      onActionSelected,
+      ...nativeProps,
+    } = this.props;
     if (this.props.logo) {
       nativeProps.logo = resolveAssetSource(this.props.logo);
     }
--- a/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
+++ b/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
@@ -9,24 +9,16 @@
 
 'use strict';
 
-const DeprecatedColorPropType = require('DeprecatedColorPropType');
-const DeprecatedViewPropTypes = require('DeprecatedViewPropTypes');
-const Image = require('Image');
-const NativeMethodsMixin = require('NativeMethodsMixin');
-const PropTypes = require('prop-types');
 const React = require('React');
 const UIManager = require('UIManager');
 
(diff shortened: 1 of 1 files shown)


Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
```diff
@@ -63,106 +55,119 @@ const optionalImageSource = PropTypes.oneOfType([
...
+  navIcon?: ?ImageSource,
+  /**
+   * Callback that is called when an action is selected. The only argument that is passed to the
+   * callback is the position of the action in the actions array.
+   */
+  onActionSelected?: ?Function,
+  /**
+   * Callback called when the icon is selected.
+   */
+  onIconClicked?: ?Function,
```
RSNara: This should be:

```JavaScript
onIconClicked?: ?() => void,
```

Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
```diff
@@ -63,106 +55,119 @@ const optionalImageSource = PropTypes.oneOfType([
...
+  logo?: ?ImageSource,
+  /**
+   * Sets the navigation icon.
+   */
+  navIcon?: ?ImageSource,
+  /**
+   * Callback that is called when an action is selected. The only argument that is passed to the
+   * callback is the position of the action in the actions array.
+   */
+  onActionSelected?: ?Function,
```
RSNara: This should be:

```JavaScript
onActionSelected?: ?(ToolbarAndroidChangeEvent) => void
```

Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
```diff
@@ -63,106 +55,119 @@ const optionalImageSource = PropTypes.oneOfType([
...
+    if (position === -1) {
+      this.props.onIconClicked && this.props.onIconClicked();
+    } else {
+      this.props.onActionSelected && this.props.onActionSelected(position);
+    }
+  };
+
+  render() {
     const nativeProps = {
       ...this.props,
```
RSNara: nit: It looks like `onIconClicked` and `onActionSelected` aren't native props.

```JavaScript
const {
  onIconClicked,
  onActionSelected,
  ...nativeProps,
} = this.props;
```

empyrical:
This used to use NativeMethodsMixin, so you should forward a ref to the native component with this:

https://github.com/facebook/react-native/blob/master/Libraries/Utilities/setAndForwardRef.js

empyrical:
Actually, on closer inspection this class doesn't need setAndForwardRef. Should just need to forward the ref

Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
```diff
@@ -195,17 +202,8 @@ const ToolbarAndroid = createReactClass({
...
-    const position = event.nativeEvent.position;
-    if (position === -1) {
-      this.props.onIconClicked && this.props.onIconClicked();
-    } else {
-      this.props.onActionSelected && this.props.onActionSelected(position);
-    }
-  },
-});
+  }
+}
```
RSNara: To successfully get rid of `NativeMethodMixin`, you need to do this:

```JavaScript
const ToolbarAndroidWithRef = React.forwardRef((props: Props, ref: React.Ref<typeof NativeToolbar>) => {
  return <ToolbarAndroid {...props} forwardedRef={ref} />
});
```

And then inside `ToolbarAndroid`, you need to do this:
```JavaScript
type Props = $ReadOnly<{|
  // ...
  forwardedRef: React.Ref<typeof NativeToolbar>,
|}>

class ToolbarAndroid extends React.Component<Props> {
  // ...
  render() {
    const {
      onIconClicked,
      onActionSelected,
      forwardedRef: ref,
      ...nativeProps,
    } = this.props;    
    // ...

    return <NativeToolbar onSelect={this._onSelect} {...nativeProps} ref={ref}/>
  }
}
```

I think React.forwardRef isn't flow typed yet, so you'll have to find a workaround for that. There are other PRs removing NativeMethodsMixin that accomplish this (e.g: #21885)

empyrical: The bit that needs to be copied for now specifically is the FlowFixMe comment above the forwardref call, to suppress the type errors

nd-02110114: 👍

nd-02110114:
CI doesnot work ...

nd-02110114:
@empyrical 
CI passed 🎉

nd-02110114:
@empyrical @RSNara 
Thank you for your patience 🙇

Issue #21907, opened 2018-10-23
android :  Failed to resolve


kmend:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
Run `react-native info` in your terminal and paste its contents here.

 React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
      Memory: 740.27 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.2 - /usr/local/bin/node
      Yarn: 1.10.1 - /usr/local/bin/yarn
      npm: 5.6.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
      Android SDK:
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.3, 28.0.3
        API Levels: 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.0/10A255 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.0-alpha.8af6728 => 16.6.0-alpha.8af6728 
      react-native: 0.57.3 => 0.57.3 
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7

## Description
Describe your issue in detail. Include screenshots if needed. If this is a regression, let us know.

android

react-native init AwesomeProject

open  android studio

err: 

The option 'android.enableAapt2' is deprecated and should not be used anymore.
Use 'android.enableAapt2=true' to remove this warning.
It will be removed at the end of 2018..

Failed to resolve: support-vector-drawable
Open File

Failed to resolve: livedata-core
Open File

Failed to resolve: viewmodel
Open File

Failed to resolve: common
Open File

Failed to resolve: runtime
[...]

PR #21910, opened 2018-10-23
Make google repo priority higher than jcenter

sunnylqm:
Fixes https://github.com/facebook/react-native/issues/21907#issuecomment-432319128 and prevent similar issues happening again

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,9 +20,9 @@ buildscript {
 
 allprojects {
     repositories {
+        google()
         jcenter()
         mavenLocal()
-        google()
 
         def androidSdk = System.getenv("ANDROID_SDK")
         maven {
--- a/local-cli/templates/HelloWorld/android/build.gradle
+++ b/local-cli/templates/HelloWorld/android/build.gradle
@@ -22,13 +22,13 @@ buildscript {
 
 allprojects {
     repositories {
+        google()
         mavenLocal()
         jcenter()
         maven {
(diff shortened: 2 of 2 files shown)


sunnylqm:
ping @kelset @grabbou 
This is a hotfix. Now all new 0.57.x project can not build on android.

kelset:
Thanks for the PR, we'll try to have this merged ASAP as it should fix the issue we are having on the CI for android test ✅

kelset:
btw @sunnylqm shouldn't this change be done in the root level `build.gradle`? 🤔

sunnylqm:
@kelset For end user only `build.gradle` in template works right? I dont know what is the root `build.gradle` for? If that is for CI, I can add that change too

kelset:
I think that yeah, the root one defines the "overall" react native android config. Anyway, in the meantime, I've tried to replicate your change on the `0.57-stable` branch to see if it fixes the CI there (and now the test_android is green) -> https://github.com/facebook/react-native/commit/7a766bf2935208d0e01376f8e60ca009d31fbe27

mmccartney:
I think it is safer to move `jcenter()` below `google()` rather than moving `google()` to the top (above `mavenLocal()`).  I agree we desperately need the templates fixed because the following doesn't work:

```
react-native init AwesomeProject
cd AwesomeProject
react-native run-android # <-- build fails because jcenter is borken
```

hramos:
Will land this one as is, but @mmccartney feel free to send another PR with your suggested changes.

PR #21913, opened 2018-10-24
fix android ci

dulmandakh:
This PR adds support-annotations to devsupport buck dependency, which fixes Android CI.


dulmandakh:
I prepared changes to bump fresco to 1.11, will create a PR when this gets merged.

PR #21935, opened 2018-10-25
bump Fresco to 1.11

dulmandakh:
This PR bumps Fresco to 1.11

Test Plan:
----------
CI is green 👍 

Changelog:
----------
[Android][Security] - bump Fresco to 1.11, includes many security fixes


dulmandakh:
Fresco fixed many security bugs in this version

dulmandakh:
@hramos Did FB switched to use Fresco 1.11, which added nativeimagefilters and it might be the cause of import failure.

ReactAndroid/src/main/libraries/fresco/fresco-react-native/BUCK
```diff
@@ -32,6 +32,7 @@ rn_android_library(
         ":bolts",
         ":imagepipeline-base",
         ":imagepipeline-core",
+        ":nativeimagefilters",
```
hramos: After importing your PR, I'm seeing the following error in one of our internal checks:

```
Failed parsing: The rule //libraries/fresco/fresco-react-native:nativeimagefilters could not be found.

Please check the spelling and whether it exists in fbandroid/libraries/fresco/fresco-react-native/BUCK.
```

That BUCK file at `fbandroid/libraries/fresco/fresco-react-native/BUCK` is quite different from `react-native-github/ReactAndroid/src/main/libraries/fresco/fresco-react-native/BUCK` though. There's no nativeimagefilters reference in it, though I found a reference to `native-filters` in it:

```
// ...
fb_android_library(
    name = "imagepipeline",
    srcs = glob(["**/*.java"]),
    visibility = DEFAULT_VISIBILITY,
    exported_deps = [
        "//libraries/fresco/imagepipeline-base/src/main/java/com/facebook/imagepipeline:imagepipeline",
        "//libraries/fresco/imagepipeline/src/main/java/com/facebook/imagepipeline:imagepipeline",
        "//libraries/fresco/native-filters:native-filters",
    ],
)
// ...
```

I'm not very familiar with BUCK, but should be use `native-filters` below instead of `nativeimagefilters` as the name for this dep?

dulmandakh: nativeimagefilters is maven artifact or gradle package name, so if we change that into **native-filters** Circle CI will fail to find it. Do you have any solution for this? Otherwise, we'll stick with with 1.10 forever.

dulmandakh:
@hramos changed BUCK target name, please re-try to import this

PR #21952, opened 2018-10-25
Update network inspector to have smarter scroll stickiness

AlanFoster:
When making use of the network inspector on a react-native app, it can be quite annoying that as new requests come in the network inspector instantly sticks to the bottom.

This PR makes this logic smarter by allowing the user to be scrolled away from the bottom by two rows to override this automatic scrolling to the bottom logic.

Test Plan:
----------

### Scenario 1: Being close to the bottom

Given the user has not scrolled away from the bottom
When a new network requests come in
Then the new request should be scrolled to

### Scenario 2: Intentionally scrolling away

Given the user has scrolled away from the bottom
When a new network request comes in
Then the new request should not be scrolled to
but the scroll bars should highlight to notify additional requests have been made

network-update

GIF depicts testing the first scenario of scrolling to the bottom when new requests arrive, followed by testing the scenario scenario of being scrolled away from the bottom

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Network Inspector] - The network inspector has been updated to no longer automatically scroll to the bottom if you have intentionally moved away

### DIFF
--- a/Libraries/Inspector/NetworkOverlay.js
+++ b/Libraries/Inspector/NetworkOverlay.js
@@ -88,6 +88,16 @@ class NetworkOverlay extends React.Component<Props, State> {
   _requestsListView: ?React.ElementRef<typeof FlatList>;
   _detailScrollView: ?React.ElementRef<typeof ScrollView>;
 
+  // Metrics are used to decide when if the request list should be sticky, and
+  // scroll to the bottom as new network requests come in, or if the user has
+  // intentionally scrolled away from the bottom - to instead flash the scroll bar
+  // and keep the current position
+  _requestsListViewScrollMetrics = {
+    offset: 0,
+    visibleLength: 0,
+    contentLength: 0,
+  };
+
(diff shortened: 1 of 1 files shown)


Libraries/Inspector/NetworkOverlay.js
```diff
@@ -383,11 +393,35 @@ class NetworkOverlay extends React.Component<Props, State> {
...
+      }
     }
-  }
+  };
+
+  _captureRequestsListView = (listRef: ?FlatList<NetworkRequestInfo>): void => {
+    this._requestsListView = listRef;
+  };
+
+  _requestsListViewOnScroll = (e: Object): void => {
```
TheSavior: What about?

```suggestion
_requestsListViewOnScroll = (e: Object): void => {
    this._requestsListViewScrollMetrics = {
      offset: e.nativeEvent.contentOffset.y,
      visibleLength: e.nativeEvent.layoutMeasurement.height,
      contentLength: e.nativeEvent.contentSize.height,
  };
```

PR #21970, opened 2018-10-27
Feature: dev menu option to change packager location during runtime

chubakueno:
Add an option to dev menu to change packager location on the fly on iOS (similar to Dev Settings in Android)

Test Plan:
----------
Tested the code change in iOS 12.1 beta 5, Xcode 10.1 beta. Set up two different PCs in the same LAN, then pointed to each one of them during runtime. Then closed the app and reopened to confirm that the last setting persists in user defaults.

Demo

Release Notes:

 [IOS] [FEATURE] [DevMenu] - Added an option to change packager location during runtime


TheSavior:
It looks like this PR doesn't work with changing bundler ports. Created an issue here: https://github.com/facebook/react-native/issues/25772

PR #21976, opened 2018-10-27
Upgrade folly to v2018.10.22.00 for iOS

Kudo:
Fixes #20302 (For iOS)

Note:
------

1. Checked the changes did not break CocoaPods integration.
2. The change for glog copying header into exported/ is to prevent build break for folly.
    `folly/detail/Demangle.h` will try to use libstdc++'s demangle.h. Unfortunately, glog also has a demangle.h in source code. So I copy exported headers and only search headers in exported/ folder during build.

Test Plan:
----------
Test with RNTester and check if the application could run without problems.

Release Notes:
--------------
[IOS] [ENHANCEMENT] [folly] - Upgrade folly to v2018.10.22.00

PR #21977, opened 2018-10-27
Upgrade folly to v2018.10.22.00 for Android

### LINKED ISSUE #20302: [META] Upgrading Folly
# For Discussion

The Folly version used in open source is very old. We're looking for volunteers interested in making the necessary changes to upgrade Folly in React Native.

## Known Incompatibilities

This is in no way an exhaustive list, but be aware you'll need to tackle the following when upgrading Folly:

- folly::Optional's has_value() doesn't exist
- `No template named 'Optional' in namespace 'folly'` stuff

Kudo:
Fixes #20302 (For Android)

Note:
------
1. New folly will have build break for a gcc-4.9 and gcc-4.9 seems to be deprecated for latest folly.
    As we only use partial folly implementations, I just fixed the build break part.
    To support building RN on Windows, the patches are written by gradle ReplaceTokens.

2. The change for glog copying header into exported/ is to prevent build break for folly.
    `folly/detail/Demangle.h` will try to use libstdc++'s demangle.h. Unfortunately, glog also has a demangle.h in source code. So I copy exported headers and only search headers in exported/ folder during build.

Test Plan:
----------
Test with RNTester and check if the application could run without problems.

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [folly] - Upgrade folly to v2018.10.22.00


dulmandakh:
Thank you @Kudo for your efforts.

I compiled ReactAndroid and RNTester, and RNTester starts up and works as usual. Also I requested @fkgozali and @hramos to review the PR.

fkgozali:
Thank you for this - @hramos, let’s verify this internally in case there’s issue.

We will also need to upgrade the version for iOS. Would you be willing to help after this PR is done?

gengjiawen:
@fkgozali https://github.com/facebook/react-native/pull/21976.

ReactAndroid/src/main/jni/third-party/folly/Android.mk
```diff
@@ -2,20 +2,35 @@ LOCAL_PATH:= $(call my-dir)
...
+  folly/memory/detail/MallocImpl.cpp \
+  folly/String.cpp \
+  folly/dynamic.cpp \
+  folly/Format.cpp \
+  folly/json_pointer.cpp \
+  folly/lang/ColdClass.cpp \
+  folly/detail/Demangle.cpp \
+  folly/hash/SpookyHashV2.cpp \
+  folly/container/detail/F14Table.cpp \
+  folly/ScopeGuard.cpp \
```
fkgozali: Longer term it would be great to figure out a way to not list these out manually, so that upgrades can be easier. I don't know what the best solution is, but perhaps the folly repo has a cmake configuration we can use.

fkgozali:
landing this

PR #21993, opened 2018-10-29
fix(image): return dest target rect if source zero

mattyclarkson:
If the provided source size if `{0. 0}` we cannot calculate the aspect ratio due to a divide by zero error. If the provided source size is zeroed, and the destination size is non-zero we can assume we want the destination size.

### DIFF
--- a/Libraries/Image/RCTImageUtils.m
+++ b/Libraries/Image/RCTImageUtils.m
@@ -55,6 +55,9 @@ CGRect RCTTargetRect(CGSize sourceSize, CGSize destSize,
   if (CGSizeEqualToSize(destSize, CGSizeZero)) {
     // Assume we require the largest size available
     return (CGRect){CGPointZero, sourceSize};
+  } else if (CGSizeEqualToSize(sourceSize, CGSizeZero)) {
+    // Assume we require the destination size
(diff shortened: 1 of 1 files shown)


matthargett:
1. Does this apply to Android as well?
2. Can you add a JS test that fails without the fix and passes with the fix?

mattyclarkson:
> Does this apply to Android as well?

Haven't seen the issue on Android. It only occurs on iOS when you change the dimensions during image load.

> Can you add a JS test that fails without the fix and passes with the fix?

Maybe but I'm not sure how. No error occurs you just end up with a blurry rendered image. The JavaScript side thinks everything is a'OK. Is there a way I can do a image diff on the rendered output?

hramos:
Same as your other PR, we cannot import this without a CLA on file.

hramos:
Looks like this is stale.

mattyclarkson:
It's not stale, I just don't want to sign the CLA. Someone is completely welcome to implement this separately.

ericlewis:
I will implement this sep. ty @mattyclarkson 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #21995, opened 2018-10-29
Add specific exception for rendering raw text outside of <Text>

yedidyak:
This matched the iOS implementation and is a more explicit exception for the most common case.

Release Notes:
--------------

[ANDROID] [ENHANCEMENT] [ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactShadowNodeImpl.java] - Add specific exception for rendering raw text outside of `<Text>`

Currently when trying to render raw text outside of a <Text> tag, the iOS message is explicit about what the error is whilst the Android error is too abstract to be useful. This PR makes the Android error message identical to the iOS one.

Image of current exceptions side by side:
screen shot 2018-10-29 at 12 54 42

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactShadowNodeImpl.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactShadowNodeImpl.java
@@ -13,6 +13,7 @@ import com.facebook.debug.tags.ReactDebugOverlayTags;
 import com.facebook.infer.annotation.Assertions;
 import com.facebook.react.common.build.ReactBuildConfig;
 import com.facebook.react.uimanager.annotations.ReactPropertyHolder;
+import com.facebook.react.views.text.ReactRawTextShadowNode;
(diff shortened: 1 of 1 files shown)


yedidyak:
Note: this is a common error when trying to shortcut existing, non boolean but falsy conditions inline.

For example: 
```js
<View>
    {cond && <MyConditionalComponent/>}
</View>
```

will work if `cond` is `null`, `undefined`, or any truthy value, but fail with this native exception if `cond` is `0` or `''`.

mdvacca:
CC @yungsters this was fixed in the React side few months ago, right?

yungsters:
Interesting... I'm not sure why this is not being caught by React in JS: https://github.com/facebook/react/blob/master/packages/react-native-renderer/src/ReactNativeHostConfig.js#L138

@yedidyak, do you have a test case that shows how you reproduce the error?

yedidyak:
My bad, I was testing in an older version. I checked the native code still handled the error in the same way, and didn't think to check that this case was handled now on the JS side.

Closing.

PR #21999, opened 2018-10-29
Fixes animated gifs incorrectly looping/not stopping on last frame

staufman:
Currently, if you load an animated gif using the standard `Image` component, it will not correctly respect the loop count property found in the Netscape App Extension block of the file. The issues are as follows:

1) If the App Extension isn't present, the animated gif loops indefinitely when it should not loop at all.
2) If the App Extension is present, the animated gif loops one less time than it should.

The other issue is that once the looping completes, the image doesn't pause at the last frame but instead, loops back to the beginning of the animation e.g. frame 1.

The fix does a few things:

1) If there is _no_ App Extension present, the image doesn't loop at all
2) If there _is_ an App Extension present, it loops the correct amount of times. For instance, if the loop count is 1, it means the gif should loop _once_ after it finishes playing, for a total of _two_ total loops.
3) Once the number of loops completes (assuming loop count isn't set to 0 which means infinite), the animation pauses on the last frame.

Test Plan:
----------

The easiest way to verify this works is to take the attached images which are labeled with how many times an image should loop. The number of times the lips in the image come to the foreground of the animation should match how many loops the animation is set to play.

You can verify the behavior for animated gifs now matches most other animated gif implementations like the one found in Chrome.

giphy-no-loop-no-app-extension
No loop

giphy-loop-once
Loop once

giphy-loop-twice
Loop twice

giphy-loop-thrice
Loop thrice

giphy-forever
Loop forever

Release Notes:
--------------

 [IOS] [BUGFIX] [Image] - Fixes an issue where animated gifs weren't looping correctly
 [IOS] [BUGFIX] [Image] - Fixes an issue where animated gifs weren't stopping at the last frame

### DIFF
--- a/Libraries/Image/RCTGIFImageDecoder.m
+++ b/Libraries/Image/RCTGIFImageDecoder.m
@@ -32,8 +32,18 @@ RCT_EXPORT_MODULE()
 {
   CGImageSourceRef imageSource = CGImageSourceCreateWithData((CFDataRef)imageData, NULL);
   NSDictionary<NSString *, id> *properties = (__bridge_transfer NSDictionary *)CGImageSourceCopyProperties(imageSource, NULL);
-  NSUInteger loopCount = [properties[(id)kCGImagePropertyGIFDictionary][(id)kCGImagePropertyGIFLoopCount] unsignedIntegerValue];
-
+  NSUInteger loopCount = 0;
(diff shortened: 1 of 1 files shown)


matthargett:
Did you verify this is only a problem in iOS and there's no fix needed on the Android side?

staufman:
Good question! I thought I read in another thread that this had already been fixed in Android but I will try to verify this over the weekend. Stay tuned!

staufman:
@matthargett Confirmed this has already been fixed on Android. So this should fix up iOS.

PR #22009, opened 2018-10-30
fix(image): compare incoming image size

mattyclarkson:
If a frame calculation occurs when a image is pending then it can fail to reload the image when the image size and pending size are out of whack. This results in a situation where the image is scaled for no reason and ends up all blurry.


hramos:
This looks like a small fix that would be OK to pull in. We do need the CLA to be signed though.

hramos:
This can be closed, now that https://github.com/facebook/react-native/commit/b758d63bc9fb4180ebab4b81f08c54a9741d0e84 has landed.

mattyclarkson:
It's not stale, I just don't want to sign the CLA. Someone is completely welcome to implement this separately.

PR #22044, opened 2018-10-31
Remove AndroidManifest.xml from UIManager

radeno:
AndroidManifest.xml file in UIManager only defines minimum SDK version for whole namespace "com.facebook". But it is always redefined by build.gradle file because Gradle file has precedence.

https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L273

Anyway Gradle is prefered way to define it. So i think it should be removed if it is not used for another purpose.

Test Plan:
--------------
CI green

Release Notes:
--------------
[INTERNAL] [BUGFIX] [com.facebook.react.uimanager] - Removed AndroidManifest.xml from UIManager in advantage of build.grafle.


dulmandakh:
@facebook-github-bot shipit

PR #22045, opened 2018-10-31
Remove undefined value on init cli command

ignacioola:
Fixes `undefined` description message when running `react-native init --help` on an existing React Native project.

Test Plan:
----------
Run the cli from `react-native` repo:
```
> cd react-native
> yarn
> node local-cli/cli.js init --help
```
And see that the description no longer says `undefined`

Release Notes:
--------------
[CLI][BUGFIX][local-cli/cliEntry.js] - Fix undefined description message when running `init` cli command help.

### DIFF
--- a/local-cli/cliEntry.js
+++ b/local-cli/cliEntry.js
@@ -60,7 +60,7 @@ function printHelpInformation() {
   let output = [
     '',
     chalk.bold(chalk.cyan(`  react-native ${cmdName} ${this.usage()}`)),
-    `  ${this._description}`,
+    this._description ? `  ${this._description}` : '',
     '',
     ...sourceInformation,
     `  ${chalk.bold('Options:')}`,


TheSavior:
Maybe we should be requiring every command to have a description instead of patching this here.

For reference, this is where commanderjs defines it's help behavior. https://github.com/tj/commander.js/blob/master/index.js#L1083-L1101

cc @grabbou

I'll land this in the mean time though, seems reasonable enough short term patch.

grabbou:
Interesting @ignacioola - sorry to comment after that amount of time, but it's weird that you have run into this issue. 

Description itself is optional and we set `noHelp` to avoid printing information when it's not provided. See here: https://github.com/react-native-community/react-native-cli/blob/master/packages/cli/src/cliEntry.js#L97

Let's take a closer look here. Maybe `noHelp` is not working as expected?

grabbou:
I am moving this discussion to the CLI repository.

PR #22048, opened 2018-11-01
Fix Promise polyfill flow type

empyrical:
This pull request adds a `declare module 'Promise'` entry to `flow/Promise.js`. Previously, it seemed that the type of the `Promise` module was `any`.

This fixes a flow error that shows up for me and some other people:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Libraries/Core/polyfillPromise.js:21:33

Importing from an untyped module makes it any and is not safe! Did you mean to add // @flow to the top of Promise?
(untyped-import)

     18│  * If you don't need these polyfills, don't use InitializeCore; just directly
     19│  * require the modules you need from InitializeCore for setup.
     20│  */
     21│ polyfillGlobal('Promise', () => require('Promise'));
     22│

```

An improper use of `Promise<T>` in `AccessibilityInfo.ios.js` that was revealed after I did this is now also fixed.

Test Plan:
----------
`yarn flow-check-ios` and `yarn flow-check-android` have been ran and both pass.

Release Notes:
--------------

[GENERAL] [MINOR] [Libraries/Components/AccessibilityInfo/AccessibilityInfo.ios.js] - Fixed a type annotation
[GENERAL] [MINOR] [flow/Promise.js] - Added a `declare module` entry

### DIFF
--- a/flow/Promise.js
+++ b/flow/Promise.js
@@ -45,3 +45,7 @@ declare class Promise<+R> {
 
   static cast<T>(object?: T): Promise<T>;
 }
+
+declare module 'Promise' {
+  declare module.exports: typeof Promise;
+}
--- a/Libraries/Components/AccessibilityInfo/AccessibilityInfo.ios.js
+++ b/Libraries/Components/AccessibilityInfo/AccessibilityInfo.ios.js
@@ -44,7 +44,7 @@ const AccessibilityInfo = {
    *
    * See http://facebook.github.io/react-native/docs/accessibilityinfo.html#fetch
    */
-  fetch: function(): Promise {
+  fetch: function(): Promise<boolean> {
     return new Promise((resolve, reject) => {
       AccessibilityManager.getCurrentVoiceOverState(resolve, reject);
     });


flow/Promise.js
```diff
@@ -45,3 +45,7 @@ declare class Promise<+R> {
 
   static cast<T>(object?: T): Promise<T>;
 }
+
+declare module 'Promise' {
```
TheSavior: This is only needed because AccessibilityInfo is requiring the Promise module instead of using Promise from global. That seems weird. Does everything else use it from global?

empyrical: It also fixes this Flow error that is showing up for me locally and some other people:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Libraries/Core/polyfillPromise.js:21:33

Importing from an untyped module makes it any and is not safe! Did you mean to add // @flow to the top of Promise?
(untyped-import)

     18│  * If you don't need these polyfills, don't use InitializeCore; just directly
     19│  * require the modules you need from InitializeCore for setup.
     20│  */
     21│ polyfillGlobal('Promise', () => require('Promise'));
     22│

```

I don't know why this error isn't showing up on CircleCI (I also did a clean re-install of all `node_modules` and this still appears)

TheSavior: This fix seems detrimental to that particular issue. `require('Promise')` should be resolving to `Libraries/Promise.js` via Haste (which does have @flow at the top so there is something else weird happening). By specifying `declare module 'Promise' {` in `flow/Promise.js` it is bypassing the haste module which is probably not what we want.

Can you try running flow get-def on `require('Promise')` in the statement you have above and see where Flow is trying to resolve that module from and debug from there?

empyrical: It's resolving to `node_modules/promise/index.js`.

The particular version of `promise` RN is using doesn't have a Flow type stub which is probably why this is showing up as `any`. The latest version (`8.0.2`) has it, however:

https://github.com/then/promise/commit/c98808978c0eda314c36268e284da219064348cb

A less ideal but simpler alternative however could be to just remove `strict-local` on `polyfillPromise`, since it's (supposed to be) the only place where promise is required directly.

(I am still not sure why it's working on CircleCI but not on my machine however 🤔 )

TheSavior:
After more digging it looks like this is a case sensitivity issue. This is a well known problem with Flow and we are talking with the Flow team about this. All of our CIs run on an OS that is case insensitive but Mac is case sensitive (or it is the other way around). Either way, the two operating systems are resolving to a different module and causing the types to be incorrect.

The simplest fix is pretty easy, just changing `@flow strict-local` to `@flow`. I'm landing this directly internally but trying to give you credit for it via this PR.

PR #22052, opened 2018-11-01
Comment out invalid Buck ruletypes

scottrice:
Summary:
My previous diff broke the open source build, because it uses a Buck version from July when the fb_native_wrapper list was generated in October. Therefore, any rules added to Buck after that point won't exist and break the build.

This list was generated by doing the following:

```
$ git clone https://github.com/facebook/buck.git ~/buck --branch v2018.07.23.01 --depth=1
$ cd buck && ant
$ bin/buck audit buildruletypes
```

Then comparing the list it gives back with the one currently in fb_native_wrapper.bzl

Reviewed By: TheSavior

Differential Revision: D12878931

### DIFF
--- a/tools/build_defs/fb_native_wrapper.bzl
+++ b/tools/build_defs/fb_native_wrapper.bzl
@@ -3,6 +3,13 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+# NOTE: Certain values are commented out due to the open source build using a
+# buck version from July 2018, while this list was generated with a version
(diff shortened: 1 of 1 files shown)


TheSavior:
Looks like this still fails with 
```
AttributeError: 'native' object has no attribute 'prebuilt_go_library'
```

I'm going to land the buck bump instead: https://github.com/facebook/react-native/pull/22049

scottrice:
Closing because  #22049 fixes the same issue this does and that one _doesnt_ fail on CI.

Issue #16333, opened 2017-10-12
'babelHelpers.asyncToGenerator is not a function' error only on iOS


NathHorrigan:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

No

### Environment
  Environment:
  OS:  macOS Sierra 10.12.1
  Node:  7.9.0
  Yarn:  0.20.3
  npm:  4.2.0
  Watchman:  4.9.0
  Xcode:  Xcode 8.3.3 Build version 8E3004b
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0-alpha.12 => 16.0.0-alpha.12
  react-native: 0.48.3 => 0.48.3

  Target Platform: iOS (10.3) Android

### Expected Behavior

App to load as it does on android .

### Actual Behavior
Error: `babelHelpers.asyncToGenerator is not a function`

![iOS Simulator (10.3 @ iPhone 6)](https://user-images.githubusercontent.com/13197111/31492331-7373a8ea-af42-11e7-9362-617f0e05b874.png)

KarthikMAM:
Duplicate of #4844

Leko:
I got same error in react-native `0.50.4`.
I think that babelHelpers.asyncToGenerator is missing in [`Libraries/polyfills/babelHelpers.js`](https://github.com/facebook/react-native/blob/7a4eda2f70aea5bcbd46d295b873788a100c18bb/Libraries/polyfills/babelHelpers.js), right ?

It works fine after add `babelHelpers.asyncToGenerator` to Libraries/polyfills/babelHelpers.js.
(It is partial of `require('babel-core').buildExternalHelpers('asyncToGenerator'.split(' '))`.)

```js
babelHelpers.asyncToGenerator = function (fn) {
  return function () {
    var gen = fn.apply(this, arguments);
    return new Promise(function (resolve, reject) {
      function step(key, arg) {
        try {
          var info = gen[key](arg);
          var value = info.value;
        } catch (error) {
          reject(error);
          return;
        }

        if (info.done) {
          resolve(value);
        } else {
          return Promise.resolve(value).then(function (value) {
            step("next", value);
          }, function (err) {
            step("throw", err);
          });
        }
      }

      return step("next");
    });
  };
};
```

It is bug ? I made a mistake ?

joeyfigaro:
[...]

PR #22058, opened 2018-11-01
Add asyncToGenerator to the babelHelper

nicholaslee119:
This is an existing problem which is discussed in this [issue](https://github.com/facebook/react-native/issues/16333#issuecomment-350425444)

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Polyfills] - Add asyncToGenerator to the babelHelper


newyankeecodeshop:
I don't think this is necessary now that RN 0.57 uses the babel runtime instead of the helpers in this file. See #21283.

cpojer:
Closing as per the last comment, agree this shouldn't be necessary.

PR #20877, opened 2018-08-27
Improve lint config and fix some errors

janicduplessis:
We recently re-enabled the lint bot for PRs and I was getting some unrelated or too strict warnings for certain files so I relaxed the eslint config for certain directories (integration tests, RNTester) and fixed some easy warnings. Also added react-native/no-unused-styles which I think is very useful to avoid shipping unused code.

Test Plan:
----------
Trivial / automated changes
Lint is down to ✖ 131 problems (0 errors, 131 warnings) from 30+ errors, 200+ warnings
Make sure flow / tests pass

Release Notes:
--------------
[INTERNAL] [ENHANCEMENT] [eslint] - Improve lint config and fix some errors

PR #22062, opened 2018-11-01
Fix linting issues

ignacioola:
Fixes lots of ESLint warnings. Many of them where in PR #20877 by @janicduplessis which requested to split the linting fixes from configuration and package changes. 

I solved only the issues that I was most certain about but I would love to get hands on all of them with a little bit of input.

Test Plan:
----------
Run linting from the React Native project:
```
> cd react-native
> yarn
> yarn lint
```
And see less linting issues than before. 

Release Notes:
--------------
[INTERNAL] [ENHANCEMENT] [eslint] - Fix eslint issues

### DIFF
--- a/RNTester/e2e/sanity.test.js
+++ b/RNTester/e2e/sanity.test.js
@@ -30,7 +30,7 @@ describe('Sanity', () => {
     await element(by.text('OK')).tap();
   });
 
-  it('Two buttons with JustifyContent:\'space-between\' should be tappable', async () => {
+  it("Two buttons with JustifyContent:'space-between' should be tappable", async () => {
     await element(by.label('This looks great!')).tap();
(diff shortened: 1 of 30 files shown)


RNTester/e2e/sanity.test.js
```diff
@@ -30,7 +30,7 @@ describe('Sanity', () => {
     await element(by.text('OK')).tap();
   });
 
-  it(`Two buttons with JustifyContent:'space-between' should be tappable`, async () => {
+  it('Two buttons with JustifyContent:\'space-between\' should be tappable', async () => {
```
TheSavior: The linter would be happy with double quotes here and then you don't need to escape the inner ones.
```suggestion
  it("Two buttons with JustifyContent:'space-between' should be tappable", async () => {
```

TheSavior:
Thanks for these fixes!

Issue #21851, opened 2018-10-18
App crashes when uploading blob with fetch API


adam-hari:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues

I found two instances where an issue was logged describing the same behaviour.
https://github.com/facebook/react-native/issues/20895
https://github.com/facebook/react-native/issues/21076
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

This issue is present in the latest React Native release (0.57.3)

## Environment
React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz
      Memory: 1.00 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 9.5.0 - /usr/local/bin/node
      Yarn: 1.7.0 - /usr/local/bin/yarn
      npm: 6.1.0 - /usr/local/bin/npm
      Watchman: 4.9.4 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
      Android SDK:
        Build Tools: 26.0.3, 27.0.3, 28.0.1
        API Levels: 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.1 AI-173.4819257
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.0-alpha.8af6728 => 16.6.0-alpha.8af6728 
      react-native: 0.57.3 => 0.57.3 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
      react-native-init: 0.6.7

## Description
I am using the fetch API to upload a locally stored media file to a remote URI. With react-native@0.55.2 the upload will succeed or execute the catch block on fail. With react-native@0.57.3 Android will propagate the 'TypeError: Network Request Failed' to the catch block whereas iOS will intermittently not reach the catch block and will instead crash with the following exception:

```
ExceptionsManager.js:84 Exception '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]' was thrown while invoking sendRequest on target Networking with params (
        {
        data =         {
            blob =             {
[...]

Issue #19717, opened 2018-06-14
Fetching blobs from filesystem results in blob-size: 0


ma-pe:
## Environment
```
React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.4
      CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 769.02 MB / 16.00 GB
      Shell: 2.7.1 - /usr/local/bin/fish
    Binaries:
      Node: 8.9.3 - ~/.nvm/versions/node/v8.9.3/bin/node
      Yarn: 1.7.0 - /usr/local/bin/yarn
      npm: 6.1.0 - ~/.nvm/versions/node/v8.9.3/bin/npm
      Watchman: 4.9.1 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
    IDEs:
      Android Studio: 3.0 AI-171.4443003
      Xcode: 9.4/9F1027a - /usr/bin/xcodebuild
    npmPackages:
      react: 16.4.1 => 16.4.1
      react-native: 0.56.0-rc => 0.56.0-rc
```

## Description
After upgrading from `react native 0.55.4` (which introduced blob fetching) to `react native 0.56.0-rc` I experience issues with blob fetching on iOS (android not tested). When fetching an image/video from filesystem and building a blob from it, the blob-`size` is `0`.

```
const mediaResponse = await fetch('file:///var/mobile/Containers/Data/Application/.../Library/CachesCamera/....mov');

// convert response to blob
const mediaBlob = await mediaResponse.blob();

console.log(mediaBlob.size) 
// 0
```

I am aware of this [react-native-blob-test](https://github.com/expo/react-native-blob-test) repo but these tests doesn't seem to fail on `size === 0` since an `&&` operator is used instead of an `||` operator : https://github.com/expo/react-native-blob-test/blob/c0dbd66f4e2a193a5031c52c9b49ea2a6b4419ed/index.common.js#L291

## Steps to Reproduce
[https://github.com/expo/react-native-blob-test/blob/master/index.common.js#L134](https://github.com/expo/react-native-blob-test/blob/master/index.common.js#L134)

## Expected Behavior
Since the image/video exists (i can save it via `CameraRoll.saveToCameraRoll`) the resulting blob should have a valid size (positive number).

## Actual Behavior
The blob has a size of `0`.

[...]

PR #22063, opened 2018-11-01
set blob as default XMLHttpRequest header response type if supported

reime005:
This is a problem that is discussed in issue #21092 

Related issues: #21851 #19717

Found the code to eventually fix this issue [here](https://github.com/github/fetch/blob/899b155746630c32d83ee29a38642da16b314ecb/fetch.js#L486)

### Test Plan

- [x] Fetching an image locally and check if the blob is there, as well as its size > 0.

### Release Notes:
___

[GENERAL] [ENHANCEMENT] [whatwg-fetch] - set blob as default XMLHttpRequest header response type if supported

### DIFF
--- a/Libraries/vendor/core/whatwg-fetch.js
+++ b/Libraries/vendor/core/whatwg-fetch.js
@@ -516,6 +516,10 @@
         xhr.withCredentials = false;
       }
 
+      if ('responseType' in xhr && support.blob) {
+        xhr.responseType = 'blob'
+      }
+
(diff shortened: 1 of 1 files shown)


farzd:
Please merge :'(

farzd:
This fixes the quotes issues, i've just patched manually and tested - Please merge

cpojer:
Thank you for sending this PR. It makes me wonder though: what is the reason RN ships with its own copy of fetch?

PR #22070, opened 2018-11-01
Modified declarations from vars to const

fernando-sendMail:
Replaces the keywords `var` with `const` in `Libraries/Lists/ListView/ListViewDataSource.js`

Test Plan:
----------
- [x] Check `npm run flow`
- [x] Check `npm run flow-check-ios`
- [x] Check `npm run flow-check-android`

Release Notes:
--------------

[GENERAL] [INTERNAL] [Libraries/Lists/ListView/ListViewDataSource.js] - Changed declarations from var to const in ListViewDataSource.js

### DIFF
--- a/Libraries/Lists/ListView/ListViewDataSource.js
+++ b/Libraries/Lists/ListView/ListViewDataSource.js
@@ -62,7 +62,7 @@ type ParamType = {
  *
  * ```
  * getInitialState: function() {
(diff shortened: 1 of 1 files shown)


matthargett:
I would change 'ENHANCEMENT' to 'INTERNAL' since it doesn't change functionality and probably shouldn't be a line item in the generated release notes.

hramos:
Sorry about that!

PR #22072, opened 2018-11-01
Improving Modal `visible` prop check to handle undefined and null

MateusAndrade:
Changing the the this.props.visible if to be ` if (!!this.props.visible === false)` . So passing undefined, or other values wont set the modal to be visible. Granting that anything that is not true, will set the modal to null on the render.

I make this this PR, because on the company that im working, we used a lot of RN. At the moment, we arent using anything like flow or TS. But to grant that the modals will only show if they are set visible to true i have changed the if made.

Test Plan:
----------
Just initiate any project, and import the Modal from RN. The try pass undefined, or any value different from a boolean. The expected behaviour was to only show the modal if you pass true.

Before
` if (this.props.visible === false)`

After
` if (this.props.visible !== true)`

Release Notes:
--------------

[GENERAL] [MINOR] [Libraries/Modal/Modal.js] - Adjusting the verification to set the modal to visible on the render. Ensuring that the modal will only show if true is passed to it.

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -217,7 +217,7 @@ class Modal extends React.Component<Props> {
   }
 
   render(): React.Node {
-    if (!!this.props.visible === false) {
+    if (this.props.visible !== true) {
       return null;
     }
 
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -217,7 +217,7 @@ class Modal extends React.Component<Props> {
   }
 
   render(): React.Node {
-    if (this.props.visible === false) {
+    if (!!this.props.visible === false) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


TheSavior:
Even though you don't use Flow or TypeScript, since the flow type is:
```
visible?: ?boolean
```
it needs to support `undefined` and `null` as well. So it seems like this fix is necessary not just for people not using type systems, but for those using them as well.

Instead of casting visible to a boolean, what about changing the conditional to:

```
if (this.props.visible !== true) {
```

MateusAndrade:
@TheSavior  i changed the visible check to `if (this.props.visible !== true)` to become more concise. I have one doubt: how i need to add these tags [CATEGORY] [TYPE] [COMPONENT]? On my PR comment?

PR #22080, opened 2018-11-02
[ANDROID TV][RN 0.57] Add setNextFocus support

### LINKED ISSUE #20100: [Android TV] Adjacent Scrollviews strange D-Pad behaviour at start/end of view
  - [ ] Use the latest React Native release: https://github.com/facebook/react-native/releases

Unable to test this on `0.56` as it's currently broken on windows. 

## Environment
```
Environment:
  OS: Windows 10
  Node: 8.5.0
  Yarn: 1.1.0
  npm: 5.3.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Not Found

  react: 16.3.1 => 16.3.1
```

## Description
When we have 2 or more horizontal ScrollViews, the D-Pad navigation behaviour is strange when navigating past the beginning or end of the ScrollView. 

When at the beginning or end of the scrollview, if there are any off screen elements in an adjacent scroll view,  navigating with the D-Pad in their direction will navigate to an item in that horizontal scrollview.

For example:
- There are 2 horizontal scroll views, with 10 items each, of which only 5 are displayed on screen
- Navigate to the last item in the first scroll view
- Navigate right once again
- Notice the focus moves to the right-most visible item in the *second* scroll view. 

Alternatively:
- There are 2 horizontal scroll views, with 10 items each, of which only 5 are displayed on screen
- On the top list, navigate all the way to the last item
- Switch to the second list, and make sure the first item is highlighted
- Navigate *left* (the end of the list)
- Notice focus shifts up to the item in the horizontal scroll view above. 

In these situations I would not expect any navigation to occur at all. 

Here is a gif illustrating the issue, where only D-Pad left or right are used: 

scollbug2

This combined with #20099, #19917 make it impossible to make a well behaving, standard style android TV app with nested scrollviews (a la youtube etc). : 

## Reproducible Demo
```
import {View, Text, TouchableOpacity, ScrollView} from 'react-native';

  render() {
    const data = [];
    for (let i = 0; i < 10; i++)
      data.push(i);

    return (
      <View>
        {[1, 2].map(() => (
          <ScrollView horizontal style={{height: 210}}>
            {data.map(i => (
              <TouchableOpacity key={i}>
                <View
                  style={{
                    backgroundColor: 'grey',
                    width: 200,
                    height: 200,
                  }}
                >
                  <Text style={{fontSize: 60}}>{i}</Text>
                </View>
              </TouchableOpacity>
            ))}
          </ScrollView>
        ))}
      </View>
    );
  }
```

### LINKED ISSUE #20593: [Android TV] Multiple Scrollviews of Touchable strange D-Pad navigation pattern

## Environment
    System:
      OS: Linux 4.15 Ubuntu 16.04.5 LTS (Xenial Xerus)
      CPU: x64 Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz
      Memory: 609.16 MB / 15.32 GB
      Shell: 4.3.48 - /bin/bash
    Binaries:
      Node: 8.11.3 - /usr/bin/node
      npm: 5.6.0 - /usr/bin/npm
    SDKs:
      Android SDK:
        Build Tools: 23.0.1, 26.0.1, 26.0.3, 27.0.3
        API Levels: 23, 26, 27
    npmPackages:
      react: 16.4.1 => 16.4.1 
    npmGlobalPackages:
      react-native-cli: 2.0.1

## Description
See original closed ticket from @Weetbix here:  #20100
Reproduced the issue on latest react version.

When having more than one scrollList of touchables, D-Pad navigation pattern is wrong using left-right keys to scroll horizontally. When reaching the end of the first list, the next touchable receiving focus is one from the second list. Same goes when reaching the start of the second list, next focused touchable is one from the above list.

**Expected behavior** : only reach touchables into the same horizontal scroll list when using left-right and goes to next list touchables only when using up/down keys.

## Reproducible Demo

```
import {View, Text, TouchableOpacity, ScrollView} from 'react-native';`

type Props = {};
  render() {
  const data = [];
  for (let i = 0; i < 10; i++)
    data.push(i);

  return (
    <View>
      {[1, 2].map((index) => (
        <ScrollView key={index} horizontal style={{height: 210}}>
          {data.map(i => (
            <TouchableOpacity key={i}>
              <View
                style={{
                  backgroundColor: 'grey',
                  width: 200,
                  height: 200,
                }}
              >
                <Text style={{fontSize: 60}}>{i}</Text>
              </View>
            </TouchableOpacity>
          ))}
        </ScrollView>
      ))}
    </View>
  );
  }
}

```

gregaou:
Add properties to be able to set the nextFocus. It can be very useful with Android TV.

New android View properties :

- nextFocusDown binded to [setNextFocusDownId](https://developer.android.com/reference/android/view/View.html#setNextFocusDownId(int))
- nextFocusForward binded to [setNextFocusForwardId](https://developer.android.com/reference/android/view/View.html#setNextFocusForwardId(int))
- nextFocusLeft binded to [setNextFocusLeftId](https://developer.android.com/reference/android/view/View.html#setNextFocusLeftId(int))
- nextFocusRight binded to [setNextFocusRightId](https://developer.android.com/reference/android/view/View.html#setNextFocusRightId(int))
- nextFocusUp binded to [setNextFocusUpId](https://developer.android.com/reference/android/view/View.html#setNextFocusUpId(int))

Can be used to fix :
- Fixes #20593
- Fixes #20100

I did this PR for RN 0.57 but i will adapt my changes to open a PR for master soon

Test Plan:
----------
I fixed this issue #20100 by doing this

```jsx
import {View, Text, TouchableOpacity, ScrollView, findNodeHandle} from 'react-native';

class Item extends Component {
  myRef = null;

  componentDidMount() {
    const { id } = this.props;
    if (id === 0) {
      this.myRef.setNativeProps({ nextFocusLeft:  findNodeHandle(this.myRef)})
    } else if (id === 9) {
      this.myRef.setNativeProps({ nextFocusRight:  findNodeHandle(this.myRef)})
    }
  }

  render() {
    const { id } = this.props;
    return (
      <TouchableOpacity
        key={id}
        ref={(c) => this.myRef = c}
      >
        <View
          style={{
            backgroundColor: 'grey',
            width: 200,
            height: 200,
          }}
        >
          <Text style={{fontSize: 60}}>{id}</Text>
        </View>
      </TouchableOpacity>
    )
  }

}

  render() {
    const data = [];
    for (let i = 0; i < 10; i++)
      data.push(i);

    return (
      <View>
        {[1, 2].map(() => (
          <ScrollView horizontal style={{height: 210}}>
            {data.map(i => <Item id={i} />)}
          </ScrollView>
        ))}
      </View>
    );
  }
}
```

Release Notes:
--------------

[ANDROID] [FEATURE] [View] - Add properties to set next Focus


gregaou:
PR for master opened #22082

cpojer:
Thanks for this PR but we are unlikely to merge this against a previous stable release.

PR #22081, opened 2018-11-02
Fix linting for examples

ignacioola:
Fixes linting for all examples and tests with React components by disabling the ESLint rule `react-native/no-inline-styles`.

Test Plan:
----------
Run linting from the React Native project:
```
> cd react-native
> yarn
> yarn lint
```
And see less linting issues than before.

Release Notes:
--------------
[INTERNAL] [ENHANCEMENT] [eslint] - Fix eslint issues for examples

### DIFF
--- a/IntegrationTests/AsyncStorageTest.js
+++ b/IntegrationTests/AsyncStorageTest.js
@@ -10,6 +10,8 @@
 
 'use strict';
 
+/* eslint-disable react-native/no-inline-styles */
+
 const React = require('react');
 const ReactNative = require('react-native');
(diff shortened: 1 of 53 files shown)


TheSavior:
I think I'd rather we follow the rule and not define inline styles. Was there a particular reason you went this direction instead?

ignacioola:
I thought it was mostly for an easier comprehension of the examples, but I like the idea of not disabling the linting rules ;) 

I'll change it for `StyleSheet` then.

ignacioola:
This PR was not good, I'll close it and open other PRs to incrementally solve the inline styles warnings

PR #22083, opened 2018-11-02
Remove unused styles

vcalvello:
Removes unused styles.

NOTE: Lint rule `react-native/no-unused-styles` not added because of custom lint rule internally at Facebook that does this.

Test Plan:
----------
Add `react-native/no-unused-styles` rule in `.eslintrc` file.
```
> yarn lint
```
No warnings related to `react-native/no-unused-styles` rule.
Remove `react-native/no-unused-styles` rule from `.eslintrc` file. 

Release Notes:
--------------
[INTERNAL] [ENHANCEMENT] - Remove unused styles.

### DIFF
--- a/ReactAndroid/src/androidTest/js/SubviewsClippingTestModule.js
+++ b/ReactAndroid/src/androidTest/js/SubviewsClippingTestModule.js
@@ -21,32 +21,31 @@ const ClippableView = requireNativeComponent('ClippableView');
 
 class ClippingSample1 extends React.Component {
   render() {
-    const styles = sample1Styles;
     return (
       <View>
         <ClippableView
           clippableViewID="outer"
-          style={styles.outer}
+          style={sample1Styles.outer}
           removeClippedSubviews={true}>
           <ClippableView
             clippableViewID="inner1"
-            style={[styles.inner, styles.inner1]}
+            style={[sample1Styles.inner, sample1Styles.inner1]}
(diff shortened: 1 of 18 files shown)


ReactAndroid/src/androidTest/js/SubviewsClippingTestModule.js
```diff
@@ -54,6 +54,7 @@ class ClippingSample1 extends React.Component {
   }
 }
 
+/* eslint-disable react-native/no-unused-styles */
```
TheSavior: It seems like we need these because the examples use a local alias like `var styles = sample2Styles;`. I haven't looked too closely but does it make sense to you why they do that and why we should ignore the rule for these blocks? Or should we remove the aliases and have these blocks referenced directly without the need for the lint disable?

vcalvello: Yes, I think removing aliases is a better option. I'll work on that.

.eslintrc
```diff
@@ -248,6 +248,7 @@
   // The following rules are made available via `eslint-plugin-react-native`
 
     "react-native/no-inline-styles": 1,
+    "react-native/no-unused-styles": 1,
```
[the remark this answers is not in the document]
vcalvello: Sure, no problem.

PR #20621, opened 2018-08-10
Bump Robolectric to 3.6.1

dulmandakh:
This PR bums Robolectric version to 3.6.1, which is the latest version we can have due to that later versions require SDK 27 or above.

**Notable changes**:
* Robolectric 3.3 improved Android support, please see http://robolectric.org/blog/2017/03/02/robolectric-3-3-and-roadmap/.
* Starting with 3.6 version they are adding support for resources http://robolectric.org/blog/2017/11/13/resources-for-real/.

Notes, I'll try to bump testing tools in near future. Then try to cleanup or refactor some RN tests.

Test Plan:
----------
Tests run and succeed as normal.

Release Notes:
--------------
[ANDROID] [ENHANCEMENT] [Robolectric] - Bump version to 3.6.1

PR #20602, opened 2018-08-09
ReactActivity extends AppCompatActivity

dulmandakh:
With this PR, ReactActivity extends AppCompatActivity per Google recommendation. 

This is subset of https://github.com/facebook/react-native/pull/19950, based on lessons learned from conversation. If merged, I'll create smaller, less distruptive PRs to land changes from the original PR.

Test Plan:
----------
Everything builds and works normally.

PR #19768, opened 2018-06-17
use AppCompat widgets

dulmandakh:
Android documentation recommends to use widgets from AppCompat, also Android Studio warns about it. This PR changes Picker, Slider, Text, TextInput components to use AppCompat widgets.

CI is GREEN https://circleci.com/gh/dulmandakh/react-native/453

## Test Plan

Android builds and runs as usual, almost no visual changes

## Release Notes
[ANDROID] [ENHANCEMENT] [WIDGET] - extend widgets from AppCompat

PR #22096, opened 2018-11-03
refactor Android Support Library in Buck

dulmandakh:
Refactor Android Support Library in Buck

Rationale:
----------
Many PRs involving Android Support Library were failed to import, even though Android CI is green (https://github.com/facebook/react-native/pull/20602, https://github.com/facebook/react-native/pull/19768, https://github.com/facebook/react-native/pull/20621). Also exports from FB to GitHub broke Android CI and fixed with https://github.com/facebook/react-native/commit/99632e16d98c8252d9a61c470558a5acdb1d34fc, https://github.com/facebook/react-native/commit/1edeaef6cab001e59b763a24e5669269e4d87146.

After some investigation and help from FB developers, I found that BUCK target names for Android Support Library are different in GitHub and FB repos. Also there might be a software tool that translates/maps FB's Android Support Library dependencies into something like below in GitHub. But it's not working properly and results in **cannot find symbol** errors.
```
provided_deps = [
    react_native_dep("third-party/android/support/v4:lib-support-v4"),
],
```

And this issue is blocking many improvements involving Android Support Library, and should be resolved as soon as possible.

Proposed solution:
-------------------
Instead of translating/mapping Android Support Library target names in BUCK dependencies, which might break anytime, why not have same target names in both FB and GitHub, therefore have Android Support Library BUCK files to have different contents in FB and GitHub, due to FB internal requirements. This requires less mapping and less error prone.

Here are proposed steps to make it work.
1. Merge this PR, and ASL Buck files separate. But must have same target paths.
2. Remove Android Support Library target mappings as dependency and export BUCK files to GitHub.
3. Remove old Android Support Library target names from GitHub once we confirm step 2 worked, which was kept to make transition smoother.

Test Plan:
----------
CI is green


hramos:
Failed with `The rule //third-party/android/support:all-support-v4 could not be found.`

dulmandakh:
@mdvacca @hramos I added some comments about the PR to make its intentions clear.

dulmandakh:
closing in favor of https://github.com/facebook/react-native/pull/22962

Issue #22101, opened 2018-11-04
Flow error Promise


exced:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
[skip envinfo]

## Description
It seems there's a Flow type Error in Libraries/Core/polyfillPromise.js

Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Libraries/Core/polyfillPromise.js:21:33

Importing from an untyped module makes it any and is not safe! Did you mean to add // @flow to the top of Promise?
(untyped-import)

     18│  * If you don't need these polyfills, don't use InitializeCore; just directly
     19│  * require the modules you need from InitializeCore for setup.
     20│  */
     21│ polyfillGlobal('Promise', () => require('Promise'));
     22│

This error could be solved by changing // @flow strict-local annotation to // @flow in Libraries/Core/polyfillPromise.js file but I guess we would like to fix the error in Promise module.

## Reproducible Demo
- `yarn && yarn run flow`
You'll see a Flow type error

TheSavior:
Trying to fix this here: https://github.com/facebook/react-native/pull/22048

exced:
Sorry for the duplicate

TheSavior:
No worries. This issue caused me to take that report more seriously. Thanks for posting. :-)

Issue #22100, opened 2018-11-04
Strengthening Flow Types for Core Components


TheSavior:
We want to tighten the Flow types of the props for our core components. We need the help of the community to comb through these files and improve these types.

Right now a lot of props are typed using `any`, `Function`, or `Object`. 

> These types can cause flow to silently skip over portions of your code, which would have otherwise caused type errors.
> \- From [eslint-plugin-flowtype](https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-no-weak-types)

Many of these types are used to validate what we send to native. If JS defines a function that expects a string but Native calls it with a number, applications can crash. 

We'd like to fix that by removing all references to `any`, `Function`, or `Object`. While we ideally want to remove all references of these files from the codebase, there are a lot, so we should prioritize the ones that are used in the props type definitions for components. 😄 

If your are able to remove all of these weak types from the file, try to change `@flow` at the top of the file to `@flow strict-local`. That should ensure that weak types can't come back to these files in the future.

[Step 1](https://github.com/facebook/react-native/issues/21342), [Step 2](https://github.com/facebook/react-native/issues/21485), and [Step 3](https://github.com/facebook/react-native/issues/21581) helped prepare our components for this step.

# How to submit quality PRs
Since many of these weak types are used at the boundary between JS and native, you will likely need to read through the native code for these components to see how these props are being used. Unfortunately, Flow will likely pass with invalid types. Paying close attention to what iOS and Android expects will be helpful to ensure the types are accurate.

I also urge those that submit PRs for this to help out with reviewing the PRs for this issue from other contributors. Code review is a great opportunity to learn and improve your own code as well as make sure everyone is on the same page and consistent. If you find tips that would have helped you investigate and improve the types, commenting on this issue with those tips would be appreciated. Help each other. ❤️ 

Also note that since you are improving these types you will likely help catch a bunch of bugs at Facebook (and elsewhere) where code isn't handling the types correctly. This means that PRs will likely take longer to land then the other issues like this we have asked for help on. This is a good thing, it is direct impact on the stability of React Native projects and catching bugs.

# The files
The following is a list of files that I'd like to address first. If you want to take one of these files please comment on this issue with the file name so that others don't work on it as well and waste work. There are plenty of files to go around. 😄

[...]

PR #22105, opened 2018-11-04
Turn Flow strict mode on for DatePickerIOS

nd-02110114:
Related to #22100

Turn Flow strict mode on for DatePickerIOS.

## Test Plan:
- [x] npm run prettier
- [x] npm run flow-check-ios
 
## Release Notes:
[GENERAL] [ENHANCEMENT] [Components/DatePicker/DatePickerIOS.ios.js] - Flow strict mode

### DIFF
--- a/Libraries/Components/DatePicker/DatePickerIOS.ios.js
+++ b/Libraries/Components/DatePicker/DatePickerIOS.ios.js
@@ -159,7 +159,11 @@ class DatePickerIOS extends React.Component<Props> {
                 ? props.initialDate.getTime()
                 : undefined
           }
-          locale={props.locale}
+          locale={
+            props.locale != null && props.locale !== ''
+              ? props.locale
+              : undefined
+          }
           maximumDate={
             props.maximumDate ? props.maximumDate.getTime() : undefined
           }
--- a/Libraries/Components/DatePicker/DatePickerIOS.ios.js
+++ b/Libraries/Components/DatePicker/DatePickerIOS.ios.js
@@ -27,7 +27,7 @@ const RCTDatePickerIOS = requireNativeComponent('RCTDatePicker');
 
 type Event = SyntheticEvent<
   $ReadOnly<{|
-    timestamp: string | number,
+    timestamp: number,
   |}>,
 >;
 
--- a/Libraries/Components/DatePicker/DatePickerIOS.ios.js
+++ b/Libraries/Components/DatePicker/DatePickerIOS.ios.js
@@ -8,7 +8,7 @@
  * This is a controlled component version of RCTDatePickerIOS
  *
  * @format
- * @flow
+ * @flow strict-local
  */
 
 'use strict';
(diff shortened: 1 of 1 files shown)


Libraries/Components/DatePicker/DatePickerIOS.ios.js
```diff
@@ -154,7 +159,7 @@ class DatePickerIOS extends React.Component<Props> {
                 ? props.initialDate.getTime()
                 : undefined
           }
-          locale={props.locale ? props.locale : undefined}
+          locale={props.locale}
```
TheSavior: I don't think this is the same behavior. React handles null (allowed by the flow type) differently than undefined. Was there a flow error that required this change?

Libraries/Components/DatePicker/DatePickerIOS.ios.js
```diff
@@ -21,10 +21,15 @@ const View = require('View');
...
 
 import type {ViewProps} from 'ViewPropTypes';
+import type {SyntheticEvent} from 'CoreEventTypes';
 
 const RCTDatePickerIOS = requireNativeComponent('RCTDatePicker');
 
-type Event = Object;
+type Event = SyntheticEvent<
+  $ReadOnly<{|
+    timestamp: string | number,
```
TheSavior: Could you link to the file(s) that shows that this is the type for `timestamp` from native?

empyrical: It looks like `timestamp` can only be a number:

https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/React/Views/RCTDatePicker.m#L32-L37

Libraries/Components/DatePicker/DatePickerIOS.ios.js
nd-02110114: Yes... I got this error...
```
Sketchy null check on string [1] which is potentially an empty string. Perhaps
you meant to check for null or undefined [2]? (sketchy-null-string)
```

I think this is a problem of flow. The situation is simlilar to this issue https://github.com/facebook/flow/issues/4608

Libraries/Components/DatePicker/DatePickerIOS.ios.js
nd-02110114: OK! I fixed.

Libraries/Components/DatePicker/DatePickerIOS.ios.js
TheSavior: Ah, this should be changed to the following then to be semantically the same. 

```
locale={(props.locale != null && props.locale != "")  ? props.locale : undefined}
```

TheSavior:
Just needs the other change to the local prop that I commented then we can land this. :+1:

nd-02110114:
@TheSavior 
I fixed!!

PR #22106, opened 2018-11-04
[flow-strict] Flow strict in DatePickerAndroid.android.js, DatePickerAndroid.ios.js

nd-02110114:
Related to #22100

Turn Flow strict mode on for DatePickerAndroid

## Test Plan:
- [x] npm run prettier
- [ ] npm run flow-check-ios
- [ ] npm run flow-check-android
 
## Release Notes:
[GENERAL] [ENHANCEMENT] [Components/DatePickerAndroid/DatePickerAndroid.android.js] - Flow strict mode
[GENERAL] [ENHANCEMENT] [Components/DatePickerAndroid/DatePickerAndroid.ios.js] - Flow strict mode

### DIFF
--- a/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js
+++ b/Libraries/Components/DatePickerAndroid/DatePickerAndroid.android.js
@@ -5,17 +5,38 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @format
- * @flow
+ * @flow strict-local
  */
 
 'use strict';
 
 const DatePickerModule = require('NativeModules').DatePickerAndroid;
 
+type Options = $ReadOnly<{|
+  date?: ?(Date | number),
+  minDate?: ?(Date | number),
+  maxDate?: ?(Date | number),
+  mode?: ?('calender' | 'spinner' | 'default'),
+|}>;
+
+type DatePickerModuleOpen =
+  | {|
+      action: 'dateSetAction',
+      year: number,
+      month: number,
+      day: number,
+    |}
+  | {|
+      action: 'dismissedAction',
+      year: typeof undefined,
+      month: typeof undefined,
+      day: typeof undefined,
+    |};
+
 /**
  * Convert a Date to a timestamp.
  */
-function _toMillis(options: Object, key: string) {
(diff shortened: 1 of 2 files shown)


nd-02110114:
I got this error...
```
// Getters and setters can have side effects and are unsafe. (unsafe-getters-setters)
static get dismissedAction() {
    return 'dismissedAction';
}
```

I should add `unsafe.enable_getters_and_setters` in .flowconfig?

empyrical:
I am not quite sure why those are getters. (Perhaps in an earlier revision, they were getting things from `NativeModules`?)

But you should be able to change them to normal static properties like:

```js
static dismissedAction = 'dismissedAction';
```

nd-02110114:
@empyrical 
Thank you for advice! I fixed.

TheSavior:
We have two internal flow errors from this change:

```
Cannot call await with `DatePickerAndroid.open(...)` bound to `p` because property `action` is missing in `Promise` [1].
                                                   v-----------------------
  87|     const {action, year, month, day} = await DatePickerAndroid.open({
  88|       date: this._getDefaultStartDate(),
  89|       minDate: minDate,
  90|       maxDate: maximumDate,
  91|     });
      -----^
References:
  react-native/Libraries/Components/DatePickerAndroid/DatePickerAndroid.ios.js:35:33
  35|   async open(options: Options): Promise<DatePickerModuleOpen> {
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
```

It looks like it expects action as part of the resolved promise. Is that something native actually provides and the type should be including or does it seem like a bug in our callsites?

nd-02110114:
I refer to these lines 
https://github.com/facebook/react-native/blob/0a39cda39a83313fa1628759b2735d58ffcc02ec/ReactAndroid/src/main/java/com/facebook/react/modules/datepicker/DatePickerDialogModule.java#L116-L120

TheSavior:
@nd-02110114, was there a reason that you closed this PR? We hadn't gotten a chance to finish landing it but it was on our radar. I'd still like to land it. :-/

nd-02110114:
@TheSavior 
I'm so sorry 🙇 Reopened!!

PR #22109, opened 2018-11-04
Warn that 'WebView' has moved to 'react-native-webview'

empyrical:
This pull request adds a warning to the getter for `WebView` that warns users that it has moved to `react-native-webview`, and that it will be removed in a future release.

A new generic `warnMoved` function has also been added that can be used to create these warnings for other APIs/components that get moved out in the slimmening.

The warning looks like this:

```
'WebView' has moved to another module and will be removed from 'react-native'
in a future release. You can instead import it from the module 'react-native-webview
  import {WebView} from 'react-native-webview';
```

Test Plan:
----------
I have tested this in RNTester, and the warning appears as expected. Flow and ESLint have also been ran.

Release Notes:
--------------

[GENERAL] [ENHANCEMENT] [Libraries/Utilities/warnMoved.js] - Added new `warnMoved` function to warn users that an import has moved to another module
[GENERAL] [MINOR] [Libraries/react-native/react-native-implementation.js] - Added a warning to the `WebView` export


TheSavior:
I really want to land this but Facebook hasn’t yet swapped over to the community one and we need to build a tad bit of infrastructure to support pulling in those 3rd party modules. 

Can we delay adding this warning for a few weeks until we can make that switch?

Libraries/Utilities/warnMoved.js
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * A simple function that warns the user once if an API has been moved to
+ * another module.
+ *
+ * @param {string} api - The name of the API or component that moved
+ * @param {string} destModule - The name of the module that the API moved to
+ * @param {string} [renamed] - Optional; name of the API in the dest module,
+ *                             if it was renamed
+ */
+function warnMoved(api: string, destModule: string, renamed?: string) {
```
TheSavior: If you make another function called warnOnce then it can be used for the existing warnings in the api as well.

empyrical:
That sounds good!

I wonder, maybe for cases like this we could have an `__FB__` or `__OSS__` variable, so stuff could be stripped out of bundles like they can with `if(__DEV__)` blocks

Libraries/Utilities/warnMoved.js
empyrical: sounds like a good idea! with the delay, maybe I should instead split this helper function into its own pull request?

TheSavior:
That seems reasonable but I would worry that it would make it too easy for FB to have a diverging version of react-native which would be detrimental for everyone. For now, I'd rather keep the pressure on us to fix our stuff and get out of the way of progress. :) We currently have the work to migrate to the community WebView on our short term roadmap

empyrical:
For the JS side of things, perhaps we should try and make a codemod to help rewrite imports of `{WebView}`. I imagine that's only a small part of the work that needs to be done, however!

TheSavior:
Writing a codemod would be great! The community will need that when they upgrade to the version without WebView. If you don't want to work on it I'm sure there is someone around who would appreciate that opportunity!

cpojer:
What's the status of this PR?

empyrical:
Looks like I have to resolve a merge conflict, but I think the issue @TheSavior brought up above might still be relevant (I also need to still split this up into a separate helper function too)

Libraries/Utilities/warnMoved.js
gengjiawen: Split maybe a good idea. We may have more usage for this if we move more component out.

Libraries/Utilities/warnMoved.js
```diff
@@ -0,0 +1,43 @@
...
+  if (warnedApis[api]) {
+    return;
+  }
+
+  const destName = renamed != null ? renamed : api;
+
+  warning(
+    false,
+    `'${api}' has moved to another module and will be removed from 'react-native' ` +
```
cpojer: Could we add something like:

```
yarn add ${destModule}
```

to the instructions?

hramos:
I missed this PR way back then. https://github.com/facebook/react-native/pull/22980 landed, which adds the deprecation warning. We should still consider #22109 as a way to handle exported components in a consistent manner.

empyrical:
I've been meaning to address this PR's issues, I will try to get to it this week!

cpojer:
@empyrical do you wanna pick this back up or close it?

empyrical:
It looks like I took so long to pick this back up that there is another warning in place on `master` now. Sorry about that! 😅 

I can pick this back up though, it would still be useful for the generic "warnOnce()" function so all the warnings in `react-native-implementation.js` don't need to add `x_warning_showed` variables, hopefully cleaning things up a bit.

cpojer:
Yes exactly! A generic warning function is still valuable. Do you wanna modify your existing PR to do that or close this one and open a new one?

empyrical:
I've modified this PR instead of making a new one 👍

PR #22114, opened 2018-11-04
Turn Flow strict mode on for KeyBoard

nd-02110114:
Related to #22100

Turn Flow strict mode on for KeyBoard

## Test Plan:
- [x] npm run prettier
- [ ] npm run flow-check-ios
- [ ] npm run flow-check-android
 
## Release Notes:
[GENERAL] [ENHANCEMENT] [Components/Keyboard/Keyboard.js] - Flow strict mode

### DIFF
--- a/Libraries/Components/Keyboard/Keyboard.js
+++ b/Libraries/Components/Keyboard/Keyboard.js
@@ -158,7 +158,7 @@ Keyboard = KeyboardEventEmitter;
 Keyboard.dismiss = dismissKeyboard;
 Keyboard.scheduleLayoutAnimation = function(event: KeyboardEvent) {
   const {duration, easing} = event;
-  if (duration != null) {
+  if (duration != null && duration !== 0) {
     LayoutAnimation.configureNext({
       duration: duration,
       update: {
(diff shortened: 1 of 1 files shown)


Libraries/Components/Keyboard/Keyboard.js
```diff
@@ -155,12 +158,12 @@ Keyboard = KeyboardEventEmitter;
 Keyboard.dismiss = dismissKeyboard;
 Keyboard.scheduleLayoutAnimation = function(event: KeyboardEvent) {
   const {duration, easing} = event;
-  if (duration) {
+  if (duration != null) {
```
TheSavior: This changes the behavior in the case that duration is 0. If you add the explicit check for that then flow should be okay.

```
if (duration != null && duration != 0) {
```

Libraries/Components/Keyboard/Keyboard.js
```diff
@@ -155,12 +158,12 @@ Keyboard = KeyboardEventEmitter;
...
     LayoutAnimation.configureNext({
       duration: duration,
       update: {
         duration: duration,
-        type: (easing && LayoutAnimation.Types[easing]) || 'keyboard',
+        type: (easing != null && LayoutAnimation.Types[easing]) || 'keyboard',
```
TheSavior: :+1:

TheSavior: In fact, the `easing` type at the top of this file could probably be even stronger by using `Type` from `LayoutAnimation.js` instead of any string.

PR #22119, opened 2018-11-04
Flow type RefreshControl

exced:
Related to #22100

Enhance Flow types for RefreshControl specifying `onRefresh` props type.
There are still 2 `any` left using `requireNativeComponent` that need to be addressed to turn Flow to strict mode.

I went through `RCTRefreshControl` and `AndroidSwipeRefreshLayout` classes to understand where this method came from.

### Test Plan:
- All flow tests succeed.

### Release Notes:

[GENERAL] [ENHANCEMENT] [RefreshControl.js] - Flow onRefresh type

### DIFF
--- a/Libraries/Components/RefreshControl/RefreshControl.js
+++ b/Libraries/Components/RefreshControl/RefreshControl.js
@@ -87,7 +87,7 @@ export type RefreshControlProps = $ReadOnly<{|
   /**
    * Called when the view starts refreshing.
    */
-  onRefresh?: ?Function,
+  onRefresh?: ?() => void,
 
   /**
    * Whether the view should be indicating an active refresh.


turnrye:
I'm planning to combine this with the other related #22100 changes to a single changelog entry. Thanks for the contribution!

TheSavior:
There are some callsites internally that are returning from `onRefresh` an object `{handled: boolean}`. 

Do you see anything in the native code that implies that this exists? I'm trying to track down the authors and see where they found that.

TheSavior:
I was able to confirm this is a bug with an internal callsite.

exced:
Sorry, did not have time to look at it. Alright I'll patch it, do you know what kind of parameters this function accepts ?

TheSavior:
No need to change this. It was a bug with internal code. I am landing this PR

PR #22122, opened 2018-11-04
[flow-strict] Flow strict TextProps

exced:
Related to #22100

Turn on Flow strict mode for TextProps.
I used ResponseHandlers type definition defined in Text.js.

I wanted to move ResponseHandlers type to TextProps and reuse it inside the file.
I know I could use $Shape<> to maybe keys but how do I elegantly maybe every values ? 
Unless having a straightforward solution, I found it clearer to copy paste these types.

### Test Plan:
- All flow tests succeed.

### Release Notes:
[GENERAL] [ENHANCEMENT] [TextProps.js] - Flow strict mode


turnrye:
Hi @exced! Thanks for the help! I plan to combine those entries in the changelog to a single entry, so the changelog section in this PR is n/a. I'll make sure that once this PR lands, you're attributed for your contribution.

TheSavior:
cc @empyrical, you two have similar types here for the responders between ViewProps and TextProps. Can you work together to figure out how these types should be consistent / share the same types?

empyrical:
Maybe `PanResponder` can export a `ResponderProps` type that both `View` and `Text` can use the `...spread` operator to make use of them?

exced:
@empyrical Sounds good, do you want me to add it into this PR ?

empyrical:
Actually I'm not sure if this can be done - it looks like `Text` and `View` don't all use the same responder props.

But for the type of `event` on your callbacks, you should import and use `GestureState` from `PanResponder`. I realize I improperly typed the event type in my PR and I will be updating them too. Then I think we are good on the responder types!

empyrical:
Sorry ignore that suggestion, I had that wrong because I was looking at something else 🤦‍♂️ Import `PressEvent` from `CoreEventTypes` and use that instead of `SyntheticEvent<>`

exced:
No worries 😊

exced:
@empyrical I replaced `SyntheticEvent<>` by `PressEvent`

PR #22132, opened 2018-11-05
Removed unnecessary code in Libraries/Text/Text.js

ifsnow:
I don't think it's necessary to use spread properties (prevState) in `getDerivedStateFromProps` 

Test Plan:
----------
- [x] Check npm run flow
- [x] Check npm run flow-check-ios
- [x] Check npm run flow-check-android

Changelog:
----------
[GENERAL] [ENHANCEMENT] [Libraries/Text] - remove unnecessary code

### DIFF
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -108,10 +108,9 @@ class TouchableText extends React.Component<Props, State> {
     responseHandlers: null,
   };
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State): ?State {
+  static getDerivedStateFromProps(nextProps: Props, prevState: State): $Shape<State> | null {
     return prevState.responseHandlers == null && isTouchable(nextProps)
       ? {
-          ...prevState,
           responseHandlers: prevState.createResponderHandlers(),
         }
       : null;


TheSavior:
I don't feel confident in the behavior of `getDerivedStateFromProps` to know what the return value is for this. The [docs]() just say:

>It should return an object to update the state, or null to update nothing.

It isn't clear to me whether this is a partial object similar to what you can use in `setState` or if it needs to be the complete object as would be used to set state in the constructor.

Do you have any resources that make it clear which it is?

ifsnow:
@TheSavior 

- `Text` already has a process of initializing state. (https://github.com/facebook/react-native/blob/v0.57.4/Libraries/Text/Text.js#L103)

- In `getDerivedStateFromProps`, only `responseHandlers` are changed when the value of props changes or it is touchable from the beginning. (https://github.com/facebook/react-native/blob/v0.57.4/Libraries/Text/Text.js#L110)

I'm sure `getDerivedStateFromProps` only needs to return the value of the change, not all values of the state. This is because previous and new values of state are merged internally. You can see it in the code below.

```js
function applyDerivedStateFromProps(workInProgress, ctor, getDerivedStateFromProps, nextProps) {
....
var partialState = getDerivedStateFromProps(nextProps, prevState);
var memoizedState = partialState === null || partialState === undefined ? prevState : _extends({}, prevState, partialState);
....
}
```

turnrye:
Hi @ifsnow, is this part of the larger effort under #22100? I plan to combine those entries in the changelog to a single entry, so the changelog section in this PR is n/a.

ifsnow:
@turnrye I agree with your plan to combine those entries. Thank you.

PR #22156, opened 2018-11-05
Remove useless additionnal blur call

gazoudoudou:
I noticed that the _onBlur method was not exactly similar to the _onFocus one in the TextInput component. 

After digging, I found that the blurTextInput method in the TextInputState.js file was call twice in a raw instead of once when the textinput component should blur.

By removing this line, I fix this unecessary multiple call.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -1198,7 +1198,6 @@ const TextInput = createReactClass({
   },
 
   _onBlur: function(event: Event) {
-    this.blur();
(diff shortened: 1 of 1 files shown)


RSNara:
cc @sahrens. It seems like you introduced this line back in 2015. This looks like it's fairly harmless, but I wonder if you intentionally added this in to fix some weird bug. Should we just land this as is?

sahrens:
Apparently this "extra" blur call was necessary on Android. Reverting this PR fixes https://github.com/facebook/react-native/issues/22425

PR #22174, opened 2018-11-06
Extract CLI out

grabbou:
Removes CLI and uses `react-native-local-cli` instead. Opening this PR to make sure tests are green and that all requires to `local-cli` are either removed or replaced with a new version.

Right now, the `react-native-local-cli` is synced with `local-cli` folder present in 0.57.0 of React Native. I will work on upgrading that shortly as we tried to make minimal changes to the CLI before it actually gets extracted.

Next steps:
- [x] Make sure `react-native init` works and new projects work w/o issues
- [x] CI is green
- [x] All call-sites to `local-cli` are replaced
- [x] Sync with latest master 
- [ ] Ship it!


grabbou:
Note: Android tests seem to be failing anyway - I am ignoring them locally on my machine too. Will report back once I find out. It's likely because this is super fresh PR based on latest master.

grabbou:
`react-native init` works without issues. Latest updates required to make it work are present in CLI repository master branch.

grabbou:
CLI is now updated to the latest commit. I am going to work on it tomorrow. I need to set up proper CI tests for the CLI itself and make sure all the Flow/Unit tests are passing w/o issues.

We are moving in the right direction.

package.json
```diff
@@ -191,6 +191,7 @@
     "prop-types": "^15.5.8",
     "react-clone-referenced-element": "^1.0.1",
     "react-devtools-core": "^3.4.0",
+    "react-native-local-cli": "^0.57.0",
```
TheSavior: Do you expect that the CLI and react-native will continue to be versioned together?

Libraries/Image/assetPathUtils.js
```diff
@@ -0,0 +1,85 @@
...
+
+'use strict';
+
+import type {PackagerAsset} from './AssetRegistry';
+
+/**
+ * FIXME: using number to represent discrete scale numbers is fragile in essence because of
+ * floating point numbers imprecision.
+ */
+function getAndroidAssetSuffix(scale: number): string {
```
grabbou: This file is now duplicated between React Native and React Native CLI. Shall we consider making it a separate package, e.g. `react-native-cli-path-utils` and publish to `npm`?

CC: @TheSavior

package.json
kelset: I agree with having separate versioning, but this means that the new repo must have a table of compatibility of some sort

Libraries/Image/assetPathUtils.js
kelset: I feel that to keep RN and the CLI separated but having them both using these utils the separate package option is a valid solution

package.json
grabbou: Don't think it's *needed* at the beginning since React Native contains a dependency on CLI for now. It is similar to the way we handle Metro - it also doesn't have compatibility table on React Native.

However, I do think keeping a `peerDependency` is going to be the right choice to make it easier to sync CLI and React Native versions together.

grabbou: FYI I already published `1.0.0-alpha.0` version and updated this PR.

Libraries/Image/assetPathUtils.js
grabbou: Yeah. For now, I think we can ship *the copy* and figure out this as a next PR. Problem with this one is that I need to update `local-cli` folder every time something changes which is very hard. Hopefully, we can get this reviewed ASAP.

grabbou:
Update: There's one last issue I need to fix related to Metro Bundler. Once that happens, this will be ready for review. I am hoping to do it by today.

grabbou:
This pull request is ready to be reviewed.

I have tested it by:
a) running most of the commands manually confirming they work
b) making sure all tests pass (Flow & Jest)
c) running `./scripts/test-manual-e2e.sh` and initing an `RNDemoApp` project to use the new CLI tools. It run w/o issues by invoking `yarn react-native run-ios` and `yarn react-native start`.

Migration strategy:
- If you have a require call to `local-cli` folder, replace it with `react-native-local-cli` or consider removing entirely.

For every new release, we are going to verify that CLI works (by running step "c" from the above, which is exactly the same workflow we used to have) and if there are no issues, the `peerDependency` will be updated in the CLI or new version released to indicate the support. I expect small changes to happen as we put this into practice.

To Metro team:
Stick to your current approach where you bump Metro dependencies in React Native master and we will be following you inside `react-native-cli`. Ideally, you'd send a PR to this repository with bumped versions too and required changes to the API if any.

This is non-breaking and transparent change to users unless they were requiring files from `local-cli` (private API, not guaranteed to work). In this case, they just need to update imports.

The sooner we merge it, the easier it will be for us to keep up with changes that happen to CLI itself and React Native.

grabbou:
Note: We don't have to ship it in next RC. This commit can be easily excluded if it causes any concerns.  It can also be easily cherry-picked to the next version.

grabbou:
Note: Once the PR is merged, I will work on:
- all open PRs and ask authors to submit the same diff against `react-native-cli` instead
- all open issues and move them to the `react-native-cli` where it makes sense

grabbou:
Tests are green. Please ignore Appveyor error - I don't know why it happens.

local-cli/cli.js
```diff
@@ -7,21 +7,10 @@
...
-require('graceful-fs').gracefulify(require('fs'));
-
-// This file must be able to run in node 0.12 without babel so we can show that
-// it is not supported. This is why the rest of the cli code is in `cliEntry.js`.
-require('./server/checkNodeVersion')();
-
-require('../setupBabel')();
-
-var cliEntry = require('./cliEntry');
+var cli = require('react-native-local-cli');
```
axe-fb: (nit): why change the name of the variable ?

Libraries/Image/AssetSourceResolver.js
```diff
@@ -22,7 +22,7 @@ import type {PackagerAsset} from 'AssetRegistry';
 const PixelRatio = require('PixelRatio');
 const Platform = require('Platform');
 
-const assetPathUtils = require('../../local-cli/bundle/assetPathUtils');
```
axe-fb: assetPathUtils are not limited to images. They also impact all other types of assets.

Libraries/Image/assetPathUtils.js
```diff
@@ -10,7 +10,7 @@
 
 'use strict';
 
-import type {PackagerAsset} from '../../Libraries/Image/AssetRegistry';
```
axe-fb: Not sure if moving this utility and making it specific to Image makes sense. This is for all assets. Maybe this should be a separate util.

Libraries/Image/assetPathUtils.js
axe-fb: Note that this file is for more than just images, so it does need to be in RN, to be used by other asset types.

local-cli/cli.js
grabbou: I ended up overwriting this file with a local copy I created upfront. Not intentional, I can revert if needed.

Libraries/Image/AssetSourceResolver.js
grabbou: Right, but my understanding is that the "impact" is more present in the `local-cli` context where the utils are used to generate a path to any type of asset.

On the React Native side of things, this file is only used in `Image` module. Anyway, this shouldn't be an issue now that the `local-cli` folder will not be removed. I am going to revert this change and w can leave this to be refactored at some point.

Libraries/Image/assetPathUtils.js
grabbou: Again, not an issue that we leave `local-cli`. However, this is only used in `Image` module on the React Native side.

grabbou:
Closing in favor of #22337. We have decided to not remove `local-cli` folder, for now, to not break the code that relies on its private API and might be using it in some places.

JAStanton:
@grabbou excuse my ignorance, but I see that local-cli has been removed in favor of [react-native-local-cli](https://www.npmjs.com/package/react-native-local-cli) which is published on npm but I cannot find the source anywhere? Is it in source control somewhere?

kelset:
@JAStanton you can find the repo here -> https://github.com/react-native-community/react-native-cli

And the original proposal -> https://github.com/react-native-community/discussions-and-proposals/pull/13

JAStanton:
Thank you!

PR #22176, opened 2018-11-06
[flow-strict] Flow TouchableNativeFeedback.android.js

mottox2:
Related to #22100

Turn Flow strict mode on for Libraries/Components/Touchable/TouchableNativeFeedback.android.js.

Test Plan:
----------
- [x] npm run prettier
- [x] npm run flow
- [x] npm run flow-check-ios
- [x] npm run lint
- [x] npm run test
 
Changelog:
----------
[GENERAL][FIXED] - apply flow


mottox2:
Sorry. I will submit PR to apply flow strict to `Touchable.js` before this PR.

Libraries/Components/Touchable/TouchableNativeFeedback.android.js
```diff
@@ -13,6 +14,7 @@ const Platform = require('Platform');
 const React = require('React');
 const PropTypes = require('prop-types');
 const ReactNative = require('ReactNative');
+// $FlowFixMe
```
TheSavior: This is to fix the flow warning about not importing from untyped files, right? If that is the case then lets just not turn on strict-local in this file.

mottox2: Thank you for reviewing.
I applied flow to `Touchable.js` then I removed `FlowFixMe` from `TouchableNativeFeedback.android.js`.

Libraries/Components/Touchable/Touchable.js
```diff
@@ -652,12 +698,12 @@ const TouchableMixin = {
     );
   },
 
-  _handleDelay: function(e) {
+  _handleDelay: function(e: LayoutEvent) {
```
RSNara: This is a bit curious. How come you decided to type this with `LayoutEvent`?

Libraries/Components/Touchable/Touchable.js
```diff
@@ -652,12 +698,12 @@ const TouchableMixin = {
...
     this.touchableDelayTimeout = null;
     this._receiveSignal(Signals.DELAY, e);
   },
 
-  _handleLongDelay: function(e) {
+  _handleLongDelay: function(e: LayoutEvent) {
```
RSNara: Same deal here. Why is this a `LayoutEvent` and not a `PressEvent`? This function calls `_receiveSignal`, which I think should accept a `PressEvent`.

Libraries/Components/Touchable/Touchable.js
```diff
@@ -725,14 +771,14 @@ const TouchableMixin = {
...
-  _isHighlight: function(state) {
+  _isHighlight: function(state: State) {
     return (
       state === States.RESPONDER_ACTIVE_PRESS_IN ||
       state === States.RESPONDER_ACTIVE_LONG_PRESS_IN
     );
   },
 
-  _savePressInLocation: function(e) {
+  _savePressInLocation: function(e: LayoutEvent) {
```
RSNara: Later on in the function, we call `TouchEventUtils.extractSingleTouch`, which expects to be called with the native event of a `PressEvent`. So, I think this type should be `PressEvent` and not `LayoutEvent`.

Libraries/Components/Touchable/Touchable.js
```diff
@@ -813,12 +869,12 @@ const TouchableMixin = {
...
   },
 
-  _startHighlight: function(e) {
+  _startHighlight: function(e: LayoutEvent) {
     this._savePressInLocation(e);
     this.touchableHandleActivePressIn && this.touchableHandleActivePressIn(e);
   },
 
-  _endHighlight: function(e) {
+  _endHighlight: function(e: LayoutEvent) {
```
RSNara: This method calls `this.touchableHandleActivePressOut`, which you typed as accepting a `PressEvent` inside `TouchableNativeFeedback.android.js`. So, shouldn't the type of `e` also be `PressEvent`?

Libraries/Components/Touchable/Touchable.js
```diff
@@ -813,12 +869,12 @@ const TouchableMixin = {
     UIManager.playTouchSound();
   },
 
-  _startHighlight: function(e) {
+  _startHighlight: function(e: LayoutEvent) {
```
RSNara: Same deal as `_endHighlight` here.

Libraries/Components/Touchable/Touchable.js
```diff
@@ -758,7 +809,12 @@ const TouchableMixin = {
    * @param {Event} e Native event.
    * @sideeffects
    */
-  _performSideEffectsForTransition: function(curState, nextState, signal, e) {
+  _performSideEffectsForTransition: function(
+    curState: State,
+    nextState: State,
+    signal: Signal,
+    e: LayoutEvent,
+  ) {
```
RSNara: This event is forwarded to `_startHighlight` and `_endHighlight`. Shouldn't it therefore also be typed as a `PressEvent`?

Libraries/Components/Touchable/Touchable.js
```diff
@@ -685,7 +731,7 @@ const TouchableMixin = {
    * @throws Error if invalid state transition or unrecognized signal.
    * @sideeffects
    */
-  _receiveSignal: function(signal, e) {
+  _receiveSignal: function(signal: Signal, e: LayoutEvent) {
```
RSNara: This gets forwarded to `_performSideEffectsForTransition`. So, shouldn't `e` be typed as `PressEvent`?

Libraries/Components/Touchable/Touchable.js
```diff
@@ -391,7 +430,7 @@ const TouchableMixin = {
    * @param {SyntheticEvent} e Synthetic event from event system.
    *
    */
-  touchableHandleResponderGrant: function(e) {
+  touchableHandleResponderGrant: function(e: SyntheticEvent<any>) {
```
RSNara: I think this function accepts a PressEvent.

Libraries/Components/Touchable/Touchable.js
```diff
@@ -432,21 +471,21 @@ const TouchableMixin = {
...
-  touchableHandleResponderRelease: function(e) {
+  touchableHandleResponderRelease: function(e: PressEvent) {
     this._receiveSignal(Signals.RESPONDER_RELEASE, e);
   },
 
   /**
    * Place as callback for a DOM element's `onResponderTerminate` event.
    */
-  touchableHandleResponderTerminate: function(e) {
+  touchableHandleResponderTerminate: function(e: PressEvent) {
```
RSNara: You've typed this function as accepting a `PressEvent`, but later on in the function, we forward this event to `_receiveSignal`, which accepts a LayoutEvent. There's a contradiction here.

Issue #19376, opened 2018-05-22
Cookies set in a redirect response are not persisted


michaeleisel:
When I go to a site that returns a 303 along with a Set-Cookie header, it does not use the new cookie from that redirect when it goes to the new URL. This would be somewhat manageable if I could use `redirect: 'manual'` in `fetch`'s request options and inspect the redirect myself, but that is also broken.

- [X] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [X] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

## Environment
Environment:
  OS: macOS High Sierra 10.13.2
  Node: 10.0.0
  Yarn: Not Found
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.1 AI-173.4697961

Packages: (wanted => installed)
  react: 16.3.1 => 16.3.1
  react-native: ^0.55.2 => 0.55.2

## Steps to Reproduce
Go to a site that redirects with a 303 response and sets a cookie via Set-Cookie.

## Expected Behavior
Cookie that it sets is sent up when the redirect continues with whatever URL the 303 sent you to.

## Actual Behavior
Cookie is not sent up

michaeleisel:
When I try this using an `NSURLSession` in Objective-C, the cookie is set as expected, so react native is behaves differently from native.

michaeleisel:
From my testing, android doesn't have the same problem, so I guess it's an iOS thing

km16:
It just stopped cookies for me from 0.56. Was working fine till 0.55. I am facing the issue in a normal fetch request.

michaeleisel:
@kelset why add a “no repro steps” label? There are repro steps in the ticket

km16:
Even I am facing this on android. It works perfectly till 0.55.4. I just feel https://github.com/facebook/react-native/pull/19770 might fix this, when it get's merged.

Jacse:
Please see #16127

gtebbutt:
@Jacse Would that explain a difference between 0.55 and 0.56? I thought that PR was referencing an issue from much further back than that. 

I'm seeing what appears to be a related issue on iOS: since updating from 0.55 to 0.57, the `react-native-video` package is no longer passing cookies to the CDN. 

[...]

PR #16127, opened 2017-09-28
Restore platform-specific cookie behavior

jamesreggio:
This PR reverts commit 047961fbf77cb012b53978184102e8ca3d00c7ec, which was introduced by #10575.

The original patch was intended to make iOS behave more like a standards-compliant web browser in its handling of cookies. Specifically, it was built to seamlessly share the cookie jar between standard HTTP and WebSocket requests.

While the goals of the original patch are understandable, they contradict the outcome of a spirited debate within the community in https://github.com/facebook/react-native/issues/14063#issuecomment-303490425. The debate was whether React Native should default to platform-specific networking behavior, or whether it should try to normalize across platforms (with an eye towards behaving as close to a web browser as possible). **The community resoundingly voted for preserving platform-specific behavior, 31 to 3.**

This custom behavior arose as an issue because the patch in question breaks the semantics around `withCredentials` on `fetch`. Instead of mapping that value to `setHTTPCookieAcceptPolicy`, it pins `setHTTPCookieAcceptPolicy` to `.Always` and defines custom behavior. Naturally, the custom behavior fails to emulate the many edge-cases that exist in the system level cookie handling, which has led to a number of people complaining on #14869 and #14931.

I laid out an argument for both sides in https://github.com/facebook/react-native/pull/14931#issuecomment-318149698 and was met with mostly silence. However, stating my intention to revert the original patch garnered some support (as you'll see on that thread), and so here we are.

Hopefully this PR can provoke the conversation that needs to be had around cookie behavior, but in the absence of strong objection, I think the correct course of action is to merge this reversion and get RN a little further out of the business of abstracting the networking stack on iOS.

PR #22178, opened 2018-11-06
[iOS] Fix cookies not being sent after a redirected response

corradio:
Fixes https://github.com/facebook/react-native/issues/19376

On iOS, the `HTTPShouldHandleCookies` boolean has no effect when custom cookies are set (see https://developer.apple.com/documentation/foundation/nsmutableurlrequest/1415485-httpshouldhandlecookies?preferredLanguage=occ). Setting custom cookies prevents the cookies of a redirect response from being re-used in the subsequent request.
This is why issue https://github.com/facebook/react-native/issues/19376 is opened.
The fix is to intercept the redirect request, and to manually set the cookies. This effectively makes the Android and iOS behaviours converge.

Changelog:
----------
  [iOS] [Fixed] - Fix cookies not being sent on iOS after redirect

### DIFF
--- a/Libraries/Network/RCTHTTPRequestHandler.mm
+++ b/Libraries/Network/RCTHTTPRequestHandler.mm
@@ -109,6 +109,21 @@ totalBytesExpectedToSend:(int64_t)totalBytesExpectedToSend
 }
 
 - (void)URLSession:(NSURLSession *)session
+              task:(NSURLSessionTask *)task
+willPerformHTTPRedirection:(NSHTTPURLResponse *)response
+        newRequest:(NSURLRequest *)request
+ completionHandler:(void (^)(NSURLRequest *))completionHandler
+{
+  // Reset the cookies on redirect.
+  // This is necessary because we're not letting iOS handle cookies by itself
+  NSMutableURLRequest *nextRequest = [request mutableCopy];
+
+  NSArray<NSHTTPCookie *> *cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookiesForURL:request.URL];
+  nextRequest.allHTTPHeaderFields = [NSHTTPCookie requestHeaderFieldsWithCookies:cookies];
+  completionHandler(nextRequest);
+}
+
+- (void)URLSession:(NSURLSession *)session
           dataTask:(NSURLSessionDataTask *)task
 didReceiveResponse:(NSURLResponse *)response
  completionHandler:(void (^)(NSURLSessionResponseDisposition))completionHandler


jamesreggio:
I'm a little sad to see that this PR contributes additional proprietary cookie behavior that deviates from the platform.

I still think the correct course of action is what I proposed in #16127, which is to remove React Native's half-baked efforts to replicate browser cookie handling, and instead let the native networking on iOS and Android dictate the behavior.

Either way, I applaud your effort, and hope you hear from the maintainers.

morenoh149:
If the choice is between a uniform api that is  inconsistent with iOS/android or an inconsistent api for each platform I'd choose the first. To do the second you'll have to expose the underlying apis and developers would have to read two sets of docs. I've seen this first hand with a native module I maintain, no easy answer.

jamesreggio:
@morenoh149 — I appreciate your viewpoint, but how do you ensure that the APIs are consistent between iOS and Android? As it stands, cookie handling on RN is already inconsistent between iOS and Android, and is inconsistent with the platform defaults for each. And because of the former inconsistencies, people just open PRs as needed to add/fix behavior they think is reasonable.

I still think the easiest way to avoid this thicket of complexity is to get RN out of the business of handling cookie-related edge cases.

shergin:
@jamesreggio

> I'm a little sad to see that this PR contributes additional proprietary cookie behavior that deviates from the platform.

Despite the famous "learn one, write anywhere" moto, React Native *is* a platform. There is a HUGE value in unifying APIs and behaviors across underlying platforms.
I am not aware of any value that relying on underlying-platform-specifics provide. I would love to learn abot that.

Using web standards as a common determinator is usually a good thing, it provides consistency and unified security model.

> How do you ensure that the APIs are consistent between iOS and Android?
Even if we cannot ensure or proor that it's consistent, there is a value to have them unified even *at some* level. 

> Cookie handling on RN is already inconsistent between iOS and Android, and is inconsistent with the platform defaults for each.

That's terrible. I don't care about being consisten with iOS or Android specifics, IMHO they should be consistent across RN on all platfroms and with web best practices.

shergin:
@corradio
Could you please provide a link to some standard that describe the implemented behaviour?

corradio:
@shergin I haven't had time to investigate but the native iOS implementation actually works like that. The fact that RN uses a custom cookie store forces us to add this edge case in order to restore default behaviour.

shergin:
Yeah, I think it satisfies "3.3.6  Sending Cookies in Unverifiable Transactions" from  https://www.ietf.org/rfc/rfc2965.txt .

PR #22181, opened 2018-11-06
[flow-strict] Flow strict ScrollResponder

saitoxu:
Related to #22100 

Enhance Flow types for Libraries/Components/ScrollResponder.js

Test Plan:
----------

- [x] npm run prettier
- [x] npm run lint
- [x] npm run flow
- [x] npm run flow-check-ios
- [x] npm run test

Changelog:
----------

[GENERAL] [ENHANCEMENT] [ScrollResponder.js] - Flow types

### DIFF
--- a/Libraries/Components/ScrollResponder.js
+++ b/Libraries/Components/ScrollResponder.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @format
- * @flow strict-local
+ * @flow
  */
 
 'use strict';
--- a/Libraries/Components/ScrollResponder.js
+++ b/Libraries/Components/ScrollResponder.js
@@ -24,7 +24,8 @@ const warning = require('fbjs/lib/warning');
 
 const {ScrollViewManager} = require('NativeModules');
 
-import type {PressEvent} from 'CoreEventTypes';
+import type {PressEvent, ScrollEvent} from 'CoreEventTypes';
+import type {KeyboardEvent} from 'Keyboard';
 import type EmitterSubscription from 'EmitterSubscription';
 
 /**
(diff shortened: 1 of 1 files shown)


Libraries/Components/ScrollResponder.js
```diff
@@ -113,7 +114,16 @@ type State = {
   observedScrollSinceBecomingResponder: boolean,
   becameResponderWhileAnimating: boolean,
 };
-type Event = Object;
+type Event = $ReadOnly<{|
```
TheSavior: Is `SyntheticEvent` from `CoreEventTypes` helpful here?

Also, are we sure this shouldn't be `ScrollEvent` from `CoreEventTypes`? Where do you see that this event contains touches?

saitoxu: In response to your review, I changed PR a lot.

I found that the event defined by the existing flow type is getting passed in the arguments of all the functions defined in `ScrollResponder`, so we don't need to use `SysntheticEvent` here.

Libraries/Components/ScrollResponder.js
```diff
@@ -584,8 +588,8 @@ const ScrollResponderMixin = {
     this.preventNegativeScrollOffset = false;
   },
 
-  scrollResponderTextInputFocusError: function(e: Event) {
-    console.error('Error measuring text field: ', e);
+  scrollResponderTextInputFocusError: function(msg: string) {
```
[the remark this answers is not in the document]
saitoxu: Yes, the function is finally passed to `UIManager.measureLayout` as error callback.
In Java, the error callback argument is string, and in Objective-C, the callback is never called.

ref:
* Objective-C https://github.com/facebook/react-native/blob/master/React/Modules/RCTUIManager.m#L1321
* Java https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIImplementation.java#L645

PR #22182, opened 2018-11-06
Various minor changes to allow compliation with msvc

acoates-ms:
This change is aiming to reduce some of the forking changes we have internally in order to use CxxReact for some additional out of tree platforms.

Some of the fixes allow more of the code to compile when using Microsoft Visual Studio Compiler. In particular the change around the default value of RN_EXPORT and some changes around how to enable the packing attribute.

Another change moves more of the code for JSBigFileString into the cpp file, so that people can share the header but replace the implementation as appropriate for other platforms.

And finally the removal of an unused header include.

This is unlikely to be the extent of the changes required for MSVC, but at least gets one of our complication blocks to work against an unforked RN.

Test Plan:
----------
No behavior changes.  Code is just being moved around. This just contains changes to help other projects share code in CxxRreact.


acoates-ms:
@shergin , I think I just pushed the required change.  Do the circleci tests not use BUCK?  Or is it only something that would show up if inserted into a bigger buck environment?

ReactCommon/cxxreact/Portability.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+// Packing uses a push/pop mechanic in MSVC.
```
Orvid: Folly already defines all of these in `folly/Portability.h`, so it's probably better to just use those rather than have another copy of them :)

ReactCommon/cxxreact/JSBigString.cpp
```diff
@@ -14,6 +14,59 @@
...
+    munmap((void *)m_data, m_size);
+  }
+  close(m_fd);
+}
+
+
+const char *JSBigFileString::c_str() const {
+  if (!m_data) {
+    m_data =
+      (const char *) mmap(0, m_size, PROT_READ, MAP_PRIVATE, m_fd, m_mapOff);
```
Orvid: `#include <folly/portability/SysMMan.h` and change `MAP_PRIVATE` to `MAP_SHARED` (this is a read-only `mmap`, so copy-on-write is irrelevant) and this should work just fine on Windows :)

shergin:
Unfortunately, it does not compile for Android:

```
stderr: js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:37: error: undefined reference to 'facebook::react::JSBigFileString::JSBigFileString(int, unsigned int, long)'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:40: error: undefined reference to 'facebook::react::JSBigFileString::c_str() const'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:41: error: undefined reference to 'facebook::react::JSBigFileString::~JSBigFileString()'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:41: error: undefined reference to 'facebook::react::JSBigFileString::~JSBigFileString()'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:52: error: undefined reference to 'facebook::react::JSBigFileString::JSBigFileString(int, unsigned int, long)'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:55: error: undefined reference to 'facebook::react::JSBigFileString::size() const'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:57: error: undefined reference to 'facebook::react::JSBigFileString::c_str() const'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:59: error: undefined reference to 'facebook::react::JSBigFileString::~JSBigFileString()'
js/react-native-github/ReactCommon/cxxreact/tests/jsbigstring.cpp:59: error: undefined reference to 'facebook::react::JSBigFileString::~JSBigFileString()'
```

cpojer:
@acoates-ms ping! What's the status of this PR? Did you see @shergin's last comment?

acoates-ms:
@cpojer 
Unfortunately its a little hard for me to verify the internal fb build.  I don't believe I'm able to build that code locally outside of fb.  I can try to fumble my way though a fix, but I'm not that familiar with BUCK, and given that the test flow for me is to publish a new GitHub PR, have it signed off, merged into phabricator, then have some fb engineer ping me with updated failure messages... it seems less than ideal.  The basic change is just to move some code from the header to the cpp, so that I can reuse the same header on other platforms, but use a different cpp implementation.  

The cpp is currently listed under TEST_CPPs, whatever that does.  I'm guessing it doesn't link it into the test, otherwise I dont think you'd get the undefined reference.  Maybe the cpp could just be added into the test link.  But again... I can't really verify any fix.

Would it be possible for a fb engineer to help fixup the internal tests?

cpojer:
@acoates-ms thanks for your reply. You are right, it's not ideal that these files require updating by us and that it's not a good workflow to share back what happens internally and ask you to fix it. I'm also sorry it took so long to get back to you to land this diff. That being said, I made the required change in the BUCK file and landed your change. Thank you so much!

Going forward I want to make sure we do everything we can to help you unfork RN and move faster on your pull requests. Do you mind assigning me to the PRs that you create so I get a notification? Thank you :)

PR #22184, opened 2018-11-06
Replace global.alert use to fix eslint warnings

vcalvello:
Replaces `alert` with `Alert.alert` to fix eslint warnings.

Test Plan:
----------
```
yarn lint
```
Check `no-alert` lint warning displayed.

Changelog:
----------
[INTERNAL] [ENHANCEMENT] - Replace `alert` with `Alert.alert` to fix eslint warning.

### DIFF
--- a/RNTester/js/AccessibilityIOSExample.js
+++ b/RNTester/js/AccessibilityIOSExample.js
@@ -12,18 +12,22 @@
 
 const React = require('react');
 const ReactNative = require('react-native');
-const {AccessibilityInfo, Text, View, TouchableOpacity} = ReactNative;
+const {AccessibilityInfo, Text, View, TouchableOpacity, Alert} = ReactNative;
 
 class AccessibilityIOSExample extends React.Component<{}> {
   render() {
     return (
       <View>
         <View
-          onAccessibilityTap={() => alert('onAccessibilityTap success')}
+          onAccessibilityTap={() =>
+            Alert.alert('Alert', 'onAccessibilityTap success')
+          }
           accessible={true}>
           <Text>Accessibility normal tap example</Text>
         </View>
-        <View onMagicTap={() => alert('onMagicTap success')} accessible={true}>
+        <View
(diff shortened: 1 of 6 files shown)


RNTester/js/AccessibilityIOSExample.js
```diff
@@ -12,18 +12,20 @@
...
-const {AccessibilityInfo, Text, View, TouchableOpacity} = ReactNative;
+const {AccessibilityInfo, Text, View, TouchableOpacity, Alert} = ReactNative;
 
 class AccessibilityIOSExample extends React.Component<{}> {
   render() {
     return (
       <View>
         <View
-          onAccessibilityTap={() => alert('onAccessibilityTap success')}
+          onAccessibilityTap={() => Alert.alert('onAccessibilityTap success')}
```
TheSavior: This is almost the same. global's alert sets the title of the alert to "Alert". https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpAlert.js#L20

RSNara:
What's the rationale behind removing reliance on `global.alert`? It seems like a fairly innocent/standard function that provides the convenience of calling `Alert.alert` in a fairly standardized manner. Considering how all browsers implement this function, I'm not really sure why we should diverge and try to get rid of it. Is that what the intent of this PR is?

TheSavior:
I think the intent is to fix the eslint warning which is pretty standard to disallow alert and console calls. I think the question is do we disable the linter for this file, or throughout all of RNTester, or use the Alert module which doesn’t trigger the lint. *shrug*

vcalvello:
The idea is to fix the eslint warning. I'm not pretty sure how, though. I usually use alert and console calls for debugging purposes and I prefer not to disable linting rules, that's why I chose using `Alert.alert`. Anyway we can disable linting in that line with `/*eslint-disable-line no-alert*/`. Definitely, I wouldn't disable the linter for the whole RNTester project. What do you prefer?

PR #22208, opened 2018-11-08
Fixed for supporting mediaPlaybackRequiresUserAction under iOS 10.

ifsnow:
There is a problem that the `mediaPlaybackRequiresUserAction` property does not work in WKWebView(`useWebKit`) under iOS 10.

I fully know you are currently working to migrate the core's WebView to the standalone `react-native-webview` project. This has already been submitted to PR in `react-native-webview` and will be merged soon. I hope this fix applies to `react-native` before your migration is done.

Test Plan:
----------
I checked that `mediaPlaybackRequiresUserAction` property works in my service as intended. I think this is a small change that is less likely to cause problems.

Release Notes:
--------------
[IOS] [BUGFIX] [React/Views/RCTWKWebView.m] - Fixed for `mediaPlaybackRequiresUserAction` in WKWebView under iOS 10.


ifsnow:
This fix was merged into [react-native-webview](https://github.com/react-native-community/react-native-webview/pull/129) and released.  I hope it will be merged into RN if possible.

PR #22212, opened 2018-11-08
Fixed for supporting flow of AsyncStorage.

ifsnow:
As you know, `AsyncStorage` does not support flow. This is a little inconvenient when developing.

I hope this helps.

Test Plan:
----------
- [x] Check npm run flow
- [x] Check npm run flow-check-ios
- [x] Check npm run flow-check-android

Release Notes:
----------
[GENERAL] [ENHANCEMENT] [Libraries/Storage/AsyncStorage.js] - support flow

### DIFF
--- a/IntegrationTests/AsyncStorageTest.js
+++ b/IntegrationTests/AsyncStorageTest.js
@@ -71,7 +71,6 @@ function testSetAndGet() {
     AsyncStorage.getItem(KEY_1, (err2, result) => {
       expectAsyncNoError('testSetAndGet/getItem', err2);
       expectEqual(result, VAL_1, 'testSetAndGet setItem');
-      expectTrue(result !== null, 'should not be null');
       updateMessage('get(key_1) correctly returned ' + String(result));
       runTestCase('should get null for missing key', testMissingGet);
     });
@@ -106,7 +105,6 @@ function testRemoveItem() {
     AsyncStorage.setItem(KEY_2, VAL_2, () => {
       AsyncStorage.getAllKeys((err, result) => {
         expectAsyncNoError('testRemoveItem/getAllKeys', err);
-        expectTrue(result !== null, 'should not be null');
         expectTrue(
           Array.isArray(result) &&
             result.indexOf(KEY_1) >= 0 &&
--- a/IntegrationTests/AsyncStorageTest.js
+++ b/IntegrationTests/AsyncStorageTest.js
@@ -71,7 +71,8 @@ function testSetAndGet() {
     AsyncStorage.getItem(KEY_1, (err2, result) => {
       expectAsyncNoError('testSetAndGet/getItem', err2);
       expectEqual(result, VAL_1, 'testSetAndGet setItem');
-      updateMessage('get(key_1) correctly returned ' + result);
+      expectTrue(result !== null, 'should not be null');
+      updateMessage('get(key_1) correctly returned ' + String(result));
       runTestCase('should get null for missing key', testMissingGet);
     });
   });
@@ -81,7 +82,8 @@ function testMissingGet() {
   AsyncStorage.getItem(KEY_2, (err, result) => {
     expectAsyncNoError('testMissingGet/setItem', err);
     expectEqual(result, null, 'testMissingGet');
-    updateMessage('missing get(key_2) correctly returned ' + result);
+    expectTrue(result === null, 'should be null');
+    updateMessage('missing get(key_2) correctly returned ' + String(result));
     runTestCase('check set twice results in a single key', testSetTwice);
   });
 }
@@ -104,9 +106,10 @@ function testRemoveItem() {
     AsyncStorage.setItem(KEY_2, VAL_2, () => {
       AsyncStorage.getAllKeys((err, result) => {
         expectAsyncNoError('testRemoveItem/getAllKeys', err);
+        expectTrue(result !== null, 'should not be null');
         expectTrue(
-          result.indexOf(KEY_1) >= 0 && result.indexOf(KEY_2) >= 0,
-          'Missing KEY_1 or KEY_2 in ' + '(' + result + ')',
+          Array.isArray(result) && result.indexOf(KEY_1) >= 0 && result.indexOf(KEY_2) >= 0,
+          'Missing KEY_1 or KEY_2 in ' + '(' + String(result) + ')',
         );
         updateMessage('testRemoveItem - add two items');
         AsyncStorage.removeItem(KEY_1, err2 => {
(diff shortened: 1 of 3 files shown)


Libraries/Storage/AsyncStorage.js
```diff
@@ -333,19 +332,19 @@ if (!RCTAsyncStorage.multiMerge) {
...
 
-function convertError(error) {
+function convertError(error): ?Error {
   if (!error) {
     return null;
   }
-  const out = new Error(error.message);
-  out.key = error.key; // flow doesn't like this :(
+  const out: Object = new Error(error.message);
+  out.key = error.key;
```
TheSavior: Can you put // $FlowIssue above this line instead of casting the error to Object? That is more consistent with our intent.

Libraries/Storage/AsyncStorage.js
```diff
@@ -40,7 +39,7 @@ const AsyncStorage = {
   getItem: function(
     key: string,
     callback?: ?(error: ?Error, result: ?string) => void,
-  ): Promise {
+  ): Promise<any> {
```
TheSavior: Could you type these Promises so that they aren't returning `any`?

ifsnow:
@TheSavior Thanks for your review feedbacks. I've tried to apply your feedback. Please confirm.

IntegrationTests/AsyncStorageTest.js
```diff
@@ -71,7 +71,8 @@ function testSetAndGet() {
     AsyncStorage.getItem(KEY_1, (err2, result) => {
       expectAsyncNoError('testSetAndGet/getItem', err2);
       expectEqual(result, VAL_1, 'testSetAndGet setItem');
-      updateMessage('get(key_1) correctly returned ' + result);
+      expectTrue(result !== null, 'should not be null');
```
TheSavior: Aren't these expectations covered by the previous statement? `VAL_1` isn't null, so we know that result isn't null.

IntegrationTests/AsyncStorageTest.js
```diff
@@ -104,9 +106,12 @@ function testRemoveItem() {
     AsyncStorage.setItem(KEY_2, VAL_2, () => {
       AsyncStorage.getAllKeys((err, result) => {
         expectAsyncNoError('testRemoveItem/getAllKeys', err);
+        expectTrue(result !== null, 'should not be null');
```
TheSavior: The following expectation would fail if result was null

Libraries/Storage/AsyncStorage.js
```diff
@@ -39,8 +38,8 @@ const AsyncStorage = {
    */
   getItem: function(
     key: string,
-    callback?: ?(error: ?Error, result: ?string) => void,
-  ): Promise {
+    callback?: ?(error: ?Error, result: string|null) => void,
+  ): Promise<string|null> {
```
TheSavior: Promise<?string>

Libraries/Storage/AsyncStorage.js
```diff
@@ -154,12 +154,13 @@ const AsyncStorage = {
    */
   getAllKeys: function(
     callback?: ?(error: ?Error, keys: ?Array<string>) => void,
-  ): Promise {
+  ): Promise<?Array<string>> {
```
TheSavior: keys is nullable?

Can this use `$ReadOnlyArray` instead of `Array`? `$ReadOnlyArray` is stricter so we prefer it where we can.

Libraries/Storage/AsyncStorage.js
```diff
@@ -223,15 +224,15 @@ const AsyncStorage = {
...
+  ): Promise<?Array<Array<string>>> {
     if (!this._immediate) {
       this._immediate = setImmediate(() => {
         this._immediate = null;
         this.flushGetRequests();
       });
     }
 
-    const getRequest = {
+    const getRequest: Object = {
```
TheSavior: Is this annotation necessary?

Libraries/Storage/AsyncStorage.js
```diff
@@ -333,19 +334,20 @@ if (!RCTAsyncStorage.multiMerge) {
   delete AsyncStorage.multiMerge;
 }
 
-function convertErrors(errs) {
+function convertErrors(errs): ?Array<?Error> {
```
TheSavior: $ReadOnlyArray

Libraries/Storage/AsyncStorage.js
ifsnow: The annotation is necessary to pass flow. I'll define a more specific type.

ifsnow:
@TheSavior Thanks for your review. I fixed all things you requested. Please review it positively.

Libraries/Storage/AsyncStorage.js
```diff
@@ -20,6 +19,14 @@ const RCTAsyncStorage =
   NativeModules.AsyncSQLiteDBStorage ||
   NativeModules.AsyncLocalStorage;
 
+type MultiRequestType = {|
+  keys: Array<string>,
```
TheSavior: We prefer $ReadOnlyArray over Array

Libraries/Storage/AsyncStorage.js
```diff
@@ -20,6 +19,14 @@ const RCTAsyncStorage =
...
+  callback: ?(errors: ?Array<Error>, result: ?Array<Array<string>>) => void,
+  keyIndex: number,
+  resolve: ?(result?: Promise<any>) => void,
```
TheSavior: Why does this Promise resolve the `any` type? Are we able to be more specific with this type?

RNTester/js/AsyncStorageExample.js
```diff
@@ -31,7 +31,7 @@ class BasicStorageExample extends React.Component<{}, $FlowFixMeState> {
   _loadInitialState = async () => {
     try {
       const value = await AsyncStorage.getItem(STORAGE_KEY);
-      if (value !== null) {
+      if (value) {
```
TheSavior: We prefer the explicit null checking rather than truthy checking. For example, we don't want to be checking for the empty string since `"" == false`.

ifsnow: This change was because of your review(`?String` intead of `String | null`).
As you know, `?String`(Maybe types) can also be null or undefined. so, `if (value !== null)` has error in Flow.

We should not use `Myabe types` to check for null. What would be better?

ifsnow:
@TheSavior Took care of your last review. :)

RNTester/js/AsyncStorageExample.js
cpojer: @ifsnow sorry for taking a while to get back to you. Yeah, using `string | null` is fine. Do you mind changing it back to that? Then I can try to land this internally and see how many Flow errors it is causing there :D

ifsnow: @cpojer I changed to use `string | null` instead of `?string` as you said. Please check. :)

cpojer:
Just getting back to this now. As expected, this caused quite a lot of issues at Facebook and a bunch of work to manually fix up Flow types. In the meantime, we also decided to move AsyncStorage into a separate package that is maintained by the community. We'd love to have you as a maintainer for that project! The repository is here: https://github.com/react-native-community/react-native-async-storage

Would you mind submitting the same PR to the repository there? I think it has a much higher chance of landing quickly without being blocked by us at Facebook :)

PR #22216, opened 2018-11-08
resizeMode applies to Image.defaultSource

dulmandakh:
resizeMode applies to Image.defaultSource, because depending on screen resolution sometimes we see padding around defaultSource image while source loading. Now we can control scale of both source and defaultSource.

Changelog:
----------
[Android] [Fixed] - Image.resizeMode will apply to defaultSource, enhanced user experience. Previously, resizeMode didn't applied to defaultSource, therefore depending on screen resolution sometimes we see padding around defaultSource image while source loading.


dulmandakh:
just dug a little bit, and found proper solution 👍

radko93:
You should ask release notes (look into new PR template).

kelset:
Hey @dulmandakh can you add the release notes to the post - then I'll try to trigger the import.

dulmandakh:
@kelset done

PR #22220, opened 2018-11-09
TextInput: Remove use of legacy context API

empyrical:
This removes the use of the legacy context API in `TextInput`.

Nothing in OSS appears to make use of the `focusEmitter`.

Test Plan:
----------
Flow and ESLint has been ran. No regressions have been noticed in RNTester

Changelog:
----------

[General] [Removed] - TextInput: Removed use of legacy context API


empyrical:
It appears to be! Neither me or @TheSavior were able to find any uses for this. But he suggested that the removal of this be done as its own separate diff, in case there was dependant code that got looked over.

PR #22231, opened 2018-11-09
Update gradle and mk files

hramos:
Summary:
- Use clang instead of the deprecated ndkbuild
- Use libc++ instead of the deprecated gnustl
- Updated gradle and android plugin version
- Fixed missing arch in local-cli template
- `clean` task should now always succeed
- `clean` task deletes build artifacts
- No need to specify buildToolsVersion. It's derived.
- Elvis operator for more readable code

Differential Revision: D13004499


akshetpandey:
This still needs to address that android-jsc is build with gnustl.
Also, in the summary `Use clang instead of the deprecated ndkbuild` should instead be `Use clang instead of the deprecated gcc` as the build still uses the deprecated ndkbuild instead of cmake in case the summary ends up in the changelog

hey99xx:
Oh maybe I'm missing this part https://github.com/react-community/jsc-android-buildscripts/pull/66#issuecomment-438416242

DanielZlotin:
> Won't this break in runtime as android-jsc still depends on libgnustl?

Right, I'm trying to allow users to choose which JSC to use, the current old one or [the new one](https://github.com/react-community/jsc-android-buildscripts.git). But seeing that the older jsc uses gnustl might prove to be a problem. We probably need to remove [this line](https://github.com/facebook/android-jsc/blob/master/install.sh#L15-L18), good catch. Also the new JSC forces min api 21 but there might be a [workaround](https://github.com/react-community/jsc-android-buildscripts/pull/66) in the works.

hey99xx:
Wouldn't it be better to wait for minSdk downgrade that you've already approved and then upgrade to new JSC in here, then android-jsc would be obsolete.
Anyways I'll leave that up to you, maybe you prefer other way as safer; but FYI according to the other PR you can also clean up Folly patch in [ReactAndroid/build.gradle#L89-L99](https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L89-L99).

Kudo:
@DanielZlotin FYI, I checked the ticket at https://code.google.com/p/android/issues/detail?id=158630.
The packagingOptions.pickFirst seems to work from android library now.
If this works, we don't need to remove libc++_shared.so from JSC handy.
I will try and update if this works later.

Kudo:
Updates:
packagingOptions.pickFirst works, but it should be added in user's application build.gradle.
I.e. at `local-cli/templates/HelloWorld/android/app/build.gradle`, but not at  `ReactAndroid/build.gradle`
Not pretty perfect.

The usage somehow like this.
```gradle
android {
    ...
    ...
    packagingOptions {
        pickFirst 'lib/armeabi-v7a/libc++_shared.so'
        pickFirst 'lib/arm64-v8a/libc++_shared.so'
        pickFirst 'lib/x86/libc++_shared.so'
        pickFirst 'lib/x86_64/libc++_shared.so'
    }
}
```

DanielZlotin:
Yes please continue discussion on #22263

PR #22240, opened 2018-11-12
CxxReact: Silence 'unused lambda capture' warnings in open-source

empyrical:
This pull request silences build warnings like this in open-source:

```
{snip}/ReactCommon/cxxreact/CxxNativeModule.cpp:134:85: warning: lambda capture 'callId' is not used [-Wunused-lambda-capture]
  messageQueueThread_->runOnQueue([method, params=std::move(params), first, second, callId] () {
```

These are variables used by "fbsystrace", which is not open-sourced.

An  unused statement has been added to the affected files in the `#else` for the `#ifdef WITH_FBSYSTRACE` conditionals

Test Plan:
----------
These warning no longer appear when building these files.

Changelog:
----------

[General] [Fixed] - Silence "unused lambda capture" warnings in CxxReact


shergin:
I am a bit concerned about defining the macro twice. Maybe we can just inline this?

empyrical:
Yep, can do that!

PR #22242, opened 2018-11-12
Updated RedBox screen

Monte9:
[Re: RedBox screen is a bit scary - Discussions and Proposals](https://github.com/react-native-community/discussions-and-proposals/issues/42)

Per @hramos:
> The RedScreen was inspired by Ruby on Rails's error screen

> I do see the RedBox screen could be made less jarring while still successfully displaying all the information we need.

Hence @jamonholmgren came up with the idea that only the header & footer of the RedBox screen could be red. This makes the content a bit more readable as well as makes the screen a little less intimidating.

Also @frantic made the suggestion that since the bottom buttons are not as important, they don't need to stand out. Hence only the header of the RedBox screen which displays the error is made red.

Screenshots:
----------

### iOS

<div style="flex-direction: row">
orginal
redbox_v2_ios

</div>

### Android:

<div style="flex-direction: row">
original_android
redbox_v2_android

</div>

Test Plan:
----------
Use `RNTester` iOS & Android app to generate an error and test the updated error screen.

Changelog:
----------
[General] [Changed] - RedBox Screen has new dark mode design to improve readability.

### DIFF
--- a/React/Modules/RCTRedBox.m
+++ b/React/Modules/RCTRedBox.m
@@ -134,7 +134,7 @@
         [extraButton addTarget:self action:@selector(showExtraDataViewController) forControlEvents:UIControlEventTouchUpInside];
 
         CGFloat buttonWidth = self.bounds.size.width / 4;
-        CGFloat bottomButtonHeight = self.bounds.size.height - buttonHeight - ([self isPhoneX] ? bottomSafeViewHeight : 0);
+        CGFloat bottomButtonHeight = self.bounds.size.height - buttonHeight - ([self isIPhoneX] ? bottomSafeViewHeight : 0);
 
         dismissButton.frame = CGRectMake(0, bottomButtonHeight, buttonWidth, buttonHeight);
         reloadButton.frame = CGRectMake(buttonWidth, bottomButtonHeight, buttonWidth, buttonHeight);
@@ -166,14 +166,14 @@
         [rootView addSubview:copyButton];
         [rootView addSubview:extraButton];
 
-        if ([self isPhoneX]) {
+        if ([self isIPhoneX]) {
             [rootView addSubview:bottomSafeView];
         }
     }
     return self;
 }
 
-- (BOOL)isPhoneX
+- (BOOL)isIPhoneX
 {
     CGSize screenSize = [UIScreen mainScreen].nativeBounds.size;
     CGSize iPhoneXScreenSize = CGSizeMake(1125, 2436);
--- a/React/Modules/RCTRedBox.m
+++ b/React/Modules/RCTRedBox.m
@@ -47,7 +47,7 @@
 #else
         self.windowLevel = UIWindowLevelStatusBar - 1;
 #endif
-        self.backgroundColor = [UIColor colorWithRed:0.8 green:0 blue:0 alpha:1];
+        self.backgroundColor = [UIColor colorWithRed:0.1 green:0.1 blue:0.1 alpha:1];
         self.hidden = YES;
 
         UIViewController *rootController = [UIViewController new];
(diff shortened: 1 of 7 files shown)


axe-fb:
This looks nice !!

kelset:
This looks awesome, thanks for the PR 👏

Titozzz:
Shouldn't the bottom text on iOS be moved up a bit, It was not designed with IphoneX in mind and it could be annoying to use the buttons?

Look way better tho, great work 👍

TheSavior:
This is great. 

Can we put the hot key under the title in the bottom bar on iOS? More like Androids. That should help keep the buttons from running into each other and the edges.

frantic:
The author of the original Red Box here :)

This is awesome! A couple of minor suggestions:

- [ ] Make the buttons at the bottom consistent between iOS and Android (specifically shortcuts placement as suggested by @TheSavior)
- [ ] Can we make the bottom part black too? With two reds in top and bottom the screen looks almost symmetrical but not quite. Also the buttons aren't as important as the error message.
- [ ] Are the subtitles intentionally red-tinted? Can we make them more neutral (grey-ish)?
- [ ] Add extra padding on iPhone X
- [ ] Changelog: use a more positive phrasing, e.g. "RedBox has new dark mode design to improve readability"

I understand that addressing this is a little outside of the initial scope of just changing the colors, but I truly believe all this little thing will add up to a much better error message experience.

Monte9:
@Titozzz, @TheSavior that's a good idea. Will update iOS so the layout of the buttons are more consistent with Android.

@frantic appreciate your feedback on this PR! I really like your suggestions and agree that these would be good to work on as part of this PR to improve the overall error screen dx.

In regards to updating the bottom bar to be black.. I agree with you in principle about the bottom buttons not being as important as the error message at the top. I'll give it a shot and will post a screenshot here so we can compare the two and see which one we like better.

Also, the reason the subtitle is red-tinted is so it's easier to read on a black background. But I'll try a shade of gray that might contrast well and see if that works too.

And I really like the way you phrased the changelog. Will end up using that instead. Thanks!

I'll work on what we talked about above and follow up in a couple days. 🏃

henrikra:
Hmm buttons are too close to bottom of the iPhone X 🤔

Monte9:
- [x] Make the buttons at the bottom consistent between iOS and Android
- [x] Make the bottom part black too
- [x] Make the subtitles more neutral (grey-ish)?
  **New color: #B3B3B3**
- [x] Add extra padding on iPhone X
  **Added conditional `SafeAreaView`** 
- [x] Changelog: **RedBox has new dark mode design to improve readability.**

/cc @frantic please take a look and let me know if you have any other suggestions.

## Updated Screenshots

### iOS:

<div style="flex-direction: row">
redbox_v2_iphone_8
redbox_v2_ios
</div>

### Android:

redbox_v2_android

React/Modules/RCTRedBox.m
```diff
@@ -88,53 +89,102 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+    CGSize screenSize = [UIScreen mainScreen].nativeBounds.size;
+    CGSize iPhoneXScreenSize = CGSizeMake(1125, 2436);
+    CGSize iPhoneXMaxScreenSize = CGSizeMake(1242, 2688);
+    CGSize iPhoneXRScreenSize = CGSizeMake(828, 1792);
+
+    return CGSizeEqualToSize(screenSize, iPhoneXScreenSize) ||
+        CGSizeEqualToSize(screenSize, iPhoneXMaxScreenSize) ||
+        CGSizeEqualToSize(screenSize, iPhoneXRScreenSize);
+}
+
```
Monte9: @kelset this is not ideal and the code is redundant with `RCTDeviceInfo`.

`RCTDeviceInfo` has a func called `RCTIsIPhoneX` but since I am not very proficient in Obj-C, I couldn't figure out how to reuse it. Any suggestions on how to reuse/improve this code would be appreciated!

React/Modules/RCTRedBox.m
```diff
@@ -88,53 +89,102 @@ - (instancetype)initWithFrame:(CGRect)frame
...
         copyButton.titleLabel.font = [UIFont systemFontOfSize:13];
+        copyButton.titleLabel.lineBreakMode = NSLineBreakByWordWrapping;
+        copyButton.titleLabel.textAlignment = NSTextAlignmentCenter;
+        copyButton.backgroundColor = [UIColor colorWithRed:0.1 green:0.1 blue:0.1 alpha:1];
         [copyButton setTitle:copyText forState:UIControlStateNormal];
-        [copyButton setTitleColor:[UIColor colorWithWhite:1 alpha:0.5] forState:UIControlStateNormal];
-        [copyButton setTitleColor:[UIColor whiteColor] forState:UIControlStateHighlighted];
+        [copyButton setTitleColor:[UIColor whiteColor] forState:UIControlStateNormal];
+        [copyButton setTitleColor:[UIColor colorWithWhite:1 alpha:0.5] forState:UIControlStateHighlighted];
         [copyButton addTarget:self action:@selector(copyStack) forControlEvents:UIControlEventTouchUpInside];
```
Monte9: @kelset Also ideally all those buttons should be in a func that just takes a param and returns a `UIButton`.

This would clean up the code a lot, but I haven’t done Obj-C in a while. Let me know if that makes sense, and if something like would work.

React/Modules/RCTRedBox.m
frantic: Yeah there's no easy way to reuse that function unless we extract it out to a util module.

I think this could work: [`safeAreaInsets`](https://developer.apple.com/documentation/uikit/uiview/2891103-safeareainsets?language=objc): according to the docs, if the `rootView` is part of the view hierarchy, this value should return the extra padding we need to add to our UI.

@shergin what's the best way to do this?

React/Modules/RCTRedBox.m
```diff
@@ -88,53 +89,102 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+        topBorderReloadButton.backgroundColor = [UIColor colorWithRed:0.70 green:0.70 blue:0.70 alpha:1.0];
+        [reloadButton addSubview:topBorderReloadButton];
+
+        UIView *topBorderCopyButton = [[UIView alloc] initWithFrame:CGRectMake(0, 0, copyButton.frame.size.width, 1)];
+        topBorderCopyButton.backgroundColor = [UIColor colorWithRed:0.70 green:0.70 blue:0.70 alpha:1.0];
+        [copyButton addSubview:topBorderCopyButton];
+
+        UIView *topBorderExtraButton = [[UIView alloc] initWithFrame:CGRectMake(0, 0, extraButton.frame.size.width, 1)];
+        topBorderExtraButton.backgroundColor = [UIColor colorWithRed:0.70 green:0.70 blue:0.70 alpha:1.0];
+        [extraButton addSubview:topBorderExtraButton];
```
frantic: Instead of creating top border for each button, can we use one full-width view and add it to the `rootView` directly?

henrikra:
Red top bar should expand to the top on iphonex

radeno:
@Monte9 awesome work! No more flash bangs. Eyes saver. Hope one day YellowBox will be same usable and readable.

React/Modules/RCTRedBox.m
cpojer: @Monte9 do you mind following up on @frantic's last comment? I think that's the only remaining blocker to getting this shipped :)

React/Modules/RCTRedBox.m
```diff
@@ -88,53 +89,102 @@ - (instancetype)initWithFrame:(CGRect)frame
...
+        bottomSafeView.backgroundColor = [UIColor colorWithRed:0.1 green:0.1 blue:0.1 alpha:1];
+        bottomSafeView.frame = CGRectMake(0, self.bounds.size.height - bottomSafeViewHeight, self.bounds.size.width, bottomSafeViewHeight);
 
         [rootView addSubview:dismissButton];
         [rootView addSubview:reloadButton];
         [rootView addSubview:copyButton];
         [rootView addSubview:extraButton];
+
+        if ([self isIPhoneX]) {
+            [rootView addSubview:bottomSafeView];
```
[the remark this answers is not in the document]
cpojer: 👍 please do

React/Modules/RCTRedBox.m
Monte9: Yes, will update it tonight. Was waiting for you to respond. 👍

React/Modules/RCTRedBox.m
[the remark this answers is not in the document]
Monte9: @frantic good point. Will update.

React/Modules/RCTRedBox.m
Monte9: Yeah, that does sound better. Will give it a shot.

hramos:
@radeno YellowBox was updated this past summer, and it should be easier to contribute to. It's out of the scope of this PR, but feel free to open an issue describing how the design could be improved. That way we won't lose track of the issue once this PR is merged.

React/Modules/RCTRedBox.m
Monte9: @frantic quick follow up: when I try to use the `safeAreaInsets` on the rootview I get 0 on an iPhoneX. 

```
NSLog(@"safeArea: %f", rootView.safeAreaInsets.bottom);
> 2018-12-13 14:42:24.955719-0600 RNTester[37397:1022268] safeArea: 0.000000
```

I looked up a few issues on stack overflow and I figure out that I need to do this instead:
```
[UIApplication sharedApplication].delegate.window.safeAreaInsets.bottom;
````

Hence I came up with this which is pretty abstracted and much better than what I had before. Thanks for the suggestion. 👍 

```
- (NSInteger)bottomSafeViewHeight
{
    if (@available(iOS 11.0, *)) {
        return [UIApplication sharedApplication].delegate.window.safeAreaInsets.bottom;
    } else {
        return 0;
    }
}
```

Monte9:
@frantic thanks for all the feedback. Def cleaned up & improved the code. I have addressed the feedback. Can you please take a look at it again?

@hramos @cpojer this PR should be ready to go once @frantic approves. I'd love to get it out in one of the v0.58 RC's so that we can get some feedback from on the community on this improved DX.

Monte9:
Ping @frantic

Monte9:
@hramos @cpojer what else needs to be done in order to get this PR merged in? I have addressed @frantic's comments although he seems to be afk.

I see that @cpojer imported this PR.. but not sure what that means? @kelset Can you give me a tl;dr of the lifecycle of a RN PR. (afaik it gets imported into FB's version of RN and is used internally first then it's merged into open source?)

kelset:
Basically, the tl;dr (or at least my understanding of it) is that once it's imported, someone internally need to 'manually merge it' into the FB monorepo. Once that's done, the bot will close the PR and a commit will appear on master.

So it's basically in the internal phase right now, which means that it's probably getting reviewed & tested. It may take a bit longer since probably the team is "fully operative" starting next week.

hramos:
What @kelset said. I just got back from being away for a few weeks on holiday break - I'll take a look at the imported diff ASAP :)

PR #22248, opened 2018-11-12
Android: Close websocket properly when remote server initiates close

syaau:
When the remote server closes the websocket, the app doesn't get a 'close' event in Android, There is no issue on iOS.

Test Plan:
----------
After making the changes, the app behaved similar to iOS. The 'close' event was sent on WebSocket on android as well.

Changelog:
----------
[Android] [Fixed] - Close websocket properly on android when remote server initiates close


dulmandakh:
@facebook-github-bot shipit

PR #22254, opened 2018-11-13
WebView:IOS Fix scalesPagesToFit while using loader

### LINKED ISSUE #2087: [WebView] startInLoadingState Causes page Scaling to Fail
If a page requires scaling and `startInLoadState={true}` then the page will load larger than the screen size. Setting this to `{false}` fixes the error. I have created an example over at RNPlay.org

https://rnplay.org/apps/1PCy0Q

prasadpilla:
Summary:
This fixes #2087

scalesPageToFit is enabled by default for Android
which is somehow not the same for IOS. So while using
webView with Loader and startInLoadingState, it renders
the page un-scaled, which is inconsistent.

It renders scaled pages if we pass scalesPageToFit explicitly
to the webView, but we should have same defaults across Android
and IOS unless there's a very strong reason not to be consistent

### DIFF
--- a/Libraries/Components/WebView/WebView.ios.js
+++ b/Libraries/Components/WebView/WebView.ios.js
@@ -404,6 +404,7 @@ class WebView extends React.Component {
 
   static defaultProps = {
     originWhitelist: WebViewShared.defaultOriginWhitelist,
+    scalesPageToFit: true,
   };
 
   state = {


prasadpilla:
We also have other default params which are in-consistent between Android and IOS. Please let me know if we want to change them as well, I'll make the necessary changes. 

Android:
` static defaultProps = {
    javaScriptEnabled: true,
    thirdPartyCookiesEnabled: true,
    scalesPageToFit: true,
    saveFormDataDisabled: false,
    originWhitelist: WebViewShared.defaultOriginWhitelist,
  };`

IOS:
`static defaultProps = {
    originWhitelist: WebViewShared.defaultOriginWhitelist,
    scalesPageToFit: true,
  };
`

PR #22263, opened 2018-11-14
Use new JavaScriptCore from npm

DanielZlotin:
Summary:
Pull Request resolved: https://github.com/facebook/react-native/pull/22231

- Use clang instead of the deprecated gcc
- Use libc++ instead of the deprecated gnustl
- Updated gradle and android plugin version
- Fixed missing arch in local-cli template
- `clean` task should now always succeed
- `clean` task deletes build artifacts
- No need to specify buildToolsVersion. It's derived.
- Elvis operator for more readable code

Differential Revision: D13004499


DanielZlotin:
- I published the old jsc-android (the current one used by RN) to npm of the new jsc (jsc-android), under tag `stock-react-native` (version is `174650.0.0`).
- new jsc is on `latest` tag
- The goal right now is that the user will be able to choose which version of jsc to use
- Easiest way I see right now to solve the gnustl <--> libc++ issue is to bundle `gnustl_shared.so` inside the package of the old jsc (under the `stock-react-native` tag). The issue right now is that I need to find a way to compile it, gradle screams at me that it is no longer supported -_-"
```
Android NDK: APP_STL gnustl_shared is no longer supported. Please switch to either c++_static or c++_shared.
```

ReactAndroid/build.gradle
```diff
@@ -242,16 +242,22 @@ task buildReactNdkLib(dependsOn: [prepareJSC, prepareBoost, prepareDoubleConvers
             "REACT_COMMON_DIR=$projectDir/../ReactCommon",
             "REACT_SRC_DIR=$projectDir/src/main/java/com/facebook/react",
             '-C', file('src/main/jni/react/jni').absolutePath,
-            '--jobs', project.hasProperty("jobs") ? project.property("jobs") : Runtime.runtime.availableProcessors()
+            '--jobs', project.findProperty("jobs") ?: Runtime.runtime.availableProcessors()
 }
 
 task cleanReactNdkLib(type: Exec) {
+    ignoreExitValue true
```
gengjiawen: why ignore exitvalue ?

DanielZlotin: Because running `gradle clean` before building executes ndkbuild which in turn throws errors for not being able to find the build folders. `clean` should always succeed.

gengjiawen:
@dulmandakh

kelset:
Can you add the Release notes please? 😇

DanielZlotin:
Using npm is just not the right way to manage jsc as it's a native dependency. After fighting with it for a while I abandoned this approach. The correct approach (which will require some work in https://github.com/react-community/jsc-android-buildscripts) is to manage publications of jsc in maven. This way the user will be able to override / use other versions of jsc. 

This PR just updates react-native to use the latest jsc from npm, with full x64 support.

gengjiawen:
I think it can, maven maybe too slow.

package.json
```diff
@@ -223,6 +223,7 @@
     "flow-bin": "^0.86.0",
     "jest": "24.0.0-alpha.6",
     "jest-junit": "5.2.0",
+    "jsc-android": "latest",
```
gengjiawen: maybe keep it simple, not introduce jsc in this pr.

gengjiawen:
Maven can make developer use different jsc more easily, it's more flexible.

build.gradle
```diff
@@ -30,8 +30,3 @@ allprojects {
         }
     }
 }
-
```
gengjiawen: Don't delete this, it's used to upgrade gradle.

local-cli/templates/HelloWorld/android/build.gradle
```diff
@@ -31,9 +30,3 @@ allprojects {
         }
     }
 }
-
-
-task wrapper(type: Wrapper) {
```
gengjiawen: Same as above.

build.gradle
DanielZlotin: Overriding wrapper scripts is deprecated and will be removed in future gradle versions

local-cli/templates/HelloWorld/android/build.gradle
DanielZlotin: Overriding wrapper scripts is deprecated and will be removed in future gradle versions

ReactAndroid/build.gradle
```diff
@@ -84,17 +85,17 @@ task prepareFolly(dependsOn: dependenciesPath ? [] : [downloadFolly], type: Copy
     from dependenciesPath ?: tarTree(downloadFolly.dest)
     from 'src/main/jni/third-party/folly/Android.mk'
     include "folly-${FOLLY_VERSION}/folly/**/*", 'Android.mk'
-    eachFile {fname -> fname.path = (fname.path - "folly-${FOLLY_VERSION}/")}
+    eachFile { fname -> fname.path = (fname.path - "folly-${FOLLY_VERSION}/") }
     includeEmptyDirs = false
 
     // Patch for folly build break on gcc 4.9 and could be removed after build by clang
```
hey99xx: Should clean the block below this.

package.json
hey99xx: By clearing old JSC is there anything else except 64-bit libicu_common.so and libjsc.so files checked into this package?

akshetpandey:
Any updates on this?

build.gradle
radeno: @gengjiawen looks like there it is (hope) https://docs.gradle.org/4.8/release-notes.html#overwriting-gradle's-built-in-tasks

hramos:
It's still being worked on. There's some changes we've had to make internally. We're close to shipping.

ben-manes:
This is great, but unfortunately there is a memory leak in an unresolved JSC (#23259). The GC does not support Android, so the default behavior only collects the young generation and OOME when the old is full. You can either compile with flags set to disable generational GC or apply the provided patch. Unfortunately without fixing JSC itself, RN will crash in long running application.

PR #22269, opened 2018-11-14
Fixing easing support on React Native Animated

santiph:
Fixing easing support on React Native Animated

Test Plan:
----------
By interpolating an AnimatedValue, documentation says easing could be configured, but instead I'm getting "Error: Interpolation property 'easing' is not supported by native animated module".

Changelog:
----------
[General][Fixed] - Support 'easing' as configuration option for Animated.Value

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -179,6 +179,7 @@ const TRANSFORM_WHITELIST = {
 const SUPPORTED_INTERPOLATION_PARAMS = {
   inputRange: true,
   outputRange: true,
+  easing: true,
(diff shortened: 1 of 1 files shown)


cpojer:
@janicduplessis do you know about this? Is this the right fix?

iamiota:
have any progress?

janicduplessis:
I'm pretty sure that customizing easing is not supported by NativeAnimated. Easing functions are only implemented in JS currently. @santiph Did you test that changing the easing function actually has an impact when using the native driver?

cpojer:
Yeah that was my thought as well. I'm gonna close this for now but happy to reopen and merge if this is confirmed to work.

PR #22281, opened 2018-11-15
default hitSlop values to 0

Taylor123:
## Motiviation

Experienced a `TouchableOpacity` releasing the `PanResponder` without invoking `onPress` due to a missing a direction key in the `hitSlop` prop; The missing key caused the corresponding pressExpand to become NaN which causes `isTouchWithinActive` to become falsey, when it should be truthy.

If defaulting to 0 is undesired behavior, I'm happy to take a different approach.

Test Plan:
----------
Used local version of react-native within the application as well as running existing test suite.

New tests don't seem necessary based on the added code.

Changelog:
----------
[General] [Fixed] - Fix unexpected responder release by defaulting hitSlop object values to 0 within touchableHandleResponderMove

### DIFF
--- a/Libraries/Components/Touchable/Touchable.js
+++ b/Libraries/Components/Touchable/Touchable.js
@@ -481,10 +481,10 @@ const TouchableMixin = {
       : null;
 
     if (hitSlop) {
-      pressExpandLeft += hitSlop.left;
-      pressExpandTop += hitSlop.top;
-      pressExpandRight += hitSlop.right;
-      pressExpandBottom += hitSlop.bottom;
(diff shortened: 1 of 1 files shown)


tychota:
That super nice.

I encountred the same issue just 2 weeks ago (as it was preventing apple validation on some device) and I was trying to find the cause in native which did not works (RCTConvert correctly convert incomplete hitslop). Now it makes sense since the part of the touch handling is handle in JS (so happy to learn that). 

That make it it crystal clear why the native component did seems to receive the touch (touchable opacity got the touched opacity) but nothing changed on JS. 

I think 0 is the good default.

PR #22285, opened 2018-11-15
WIP add type for functions

tsukushi0104:
Related to #22100
This PR is work in progress.

@MengjueW 
Hello! I'm trying to enhance flowtypes for TouchableWithoutFeedback.
And I think `_linearTap` has problem now... onPress would pass only 1 argument but this function requires `refName` as the first argument and `e` as the second. I tried to fix this but I don't know the purpose of this file and don't know the specification... Could you please advice for me??

https://github.com/facebook/react-native/blob/06c13b3e066636b414f5dc19c919dcb138763c71/RNTester/js/RTLExample.js#L529-L544

### DIFF
--- a/Libraries/Components/Touchable/TouchableWithoutFeedback.js
+++ b/Libraries/Components/Touchable/TouchableWithoutFeedback.js
@@ -26,7 +26,7 @@ const {
   DeprecatedAccessibilityTraits,
 } = require('DeprecatedViewAccessibility');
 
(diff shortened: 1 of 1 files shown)


hramos:
I'm going through open PRs and it looks like this one has been flagged as a work in progress for a while. I'm going to close it, but feel free to open a new one once you're ready for review.

PR #22300, opened 2018-11-15
Added `persistentScrollbar` prop in ScrollView component

kpsroka:
This PR is a follow-up from the closed [discussions-and-proposals PR](https://github.com/react-native-community/discussions-and-proposals/pull/39), that explains in more detail the rationale for adding this feature.

Test Plan:
----------
Tested through modification of `FlatList` example in the RNTester app on Android. 

See the following screenrecordings:

| Base | `persistentScrollbar` |
| ------------- | ------------- |
| rn-scrollview-base | rn-scrollview-persistentscrollbar|

Changelog:
----------

[Android] [Added] - Added `persistentScrollbar` prop in ScrollView component

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
@@ -257,7 +257,7 @@ public class ReactHorizontalScrollViewManager
   }
 
   @ReactProp(name = "persistentScrollbar")
-  public void setPersistentScrollbar(ReactScrollView view, boolean value) {
+  public void setPersistentScrollbar(ReactHorizontalScrollView view, boolean value) {
     view.setScrollbarFadingEnabled(!value);
   }
 }


Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -301,6 +301,13 @@ type AndroidProps = $ReadOnly<{|
    * @platform android
    */
   overScrollMode?: ?('auto' | 'always' | 'never'),
+  /**
+   * Causes the scrollbars not to turn transparent when they are not in use.
+   * The default value is false.
+   *
+   * @platform android
+   */
+  persistentScrollbar?: ?boolean,
```
thymikee: O guess we can be more strict about this type and just go with `boolean` if it's set

grabbou: Yeah, I think this is a good suggestion.

grabbou:
Looks good to me, thanks for sending it over! One comment from @thymikee to be fixed and I think we can ship it.

mdvacca:
Hey @kpsroka, I imported the PR but it failed because of the following error

```
error: incompatible types: com.facebook.react.views.scroll.ReactHorizontalScrollView cannot be converted to com.facebook.react.views.scroll.ReactScrollView
        manager.setPersistentScrollbar(view, props.getBoolean(name, false));
                                       ^

```
You can find the full error here: https://circleci.com/gh/facebook/react-native/60942?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

kpsroka:
@mdvacca Yeah, I've noticed that -- I'll fix it tomorrow.

Libraries/Components/ScrollView/ScrollView.js
kpsroka: All the prop types are defined like this. I'd rather keep it consistent than strict, and I assume there might be a reason for this. @grabbou Do you know who might shed some light on why these types are defined as optional on both sides?

kpsroka:
@mdvacca I've fixed the PR now.

Libraries/Components/ScrollView/ScrollView.js
grabbou: Then I'd leave it as is. I think it might be for historical reasons. @TheSavior do you happen to know the answer to this?

PR #22301, opened 2018-11-15
[flow-strict] Flow strict ScrollView; get rid of InternalScrollViewType

thymikee:
## Summary

Relates to #22100.

I left 2 `$FlowFixMe`s as I was not sure how to handle generic `React.Element<>` and which native props can I pass to ScrollView (would be cool to document it once we got proper types there).

I also got rid of `InternalScrollViewType` because we have better typings in original `ScrollView` now.

## Test Plan

Flow passes. 

## Changelog

[General] [Fixed] - Stricter Flow typings for ScrollView


RNTester/js/RNTesterPage.js
```diff
@@ -41,7 +41,11 @@ class RNTesterPage extends React.Component<Props> {
     return (
       <View style={styles.container}>
         {title}
-        <ContentWrapper style={styles.wrapper} {...wrapperProps}>
+        <ContentWrapper
+          style={styles.wrapper}
+          {...wrapperProps}
+          zoomScale={2}
+          onScrollBeginDrag={event => console.log(event.nativeEvent)}>
```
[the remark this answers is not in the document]
thymikee: Whoopsie! Will fix that soon 😅

grabbou:
@facebook-github-bot shipit

grabbou:
Thank you for sending this PR and to all the reviewers!

hramos:
This may require some internal work to update some internal callsites to ensure Flow check does not fail.

thymikee:
Understood. The typings around ScrollView increased significantly with this PR (by just removing the `InternalScrollViewType`) so I would expect some failures here and there :D. But technically that's a good thing. Hopefully it will be smooth upgrade.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -1006,10 +1013,6 @@ const ScrollView = createReactClass({
   },
 });
 
-const TypedScrollView = ((ScrollView: any): Class<
```
TheSavior: The reason we were exporting this is because ScrollView still uses createReactClass and is thus untyped. I believe with the change in this PR the Props type is now unused, I think this was the only callsite.

thymikee: Oh man, how could I miss that :o. I could migrate this to class component by duplicating ScrollResponder.Mixin (it's only used in deprecated ListView, so the duplication is going to be removed soon anyway)

TheSavior: Migrating it to an ES6 class and not using mixins would be great and something we need to do anyways. It probably makes sense to do that in a separate PR as we've seen those be pretty tricky. cc @empyrical as he has converted a few other ones and ran into some edge cases. If you two work together to work out any problems you come across I'd be super excited!

Do you think that PR needs to happen first?

thymikee: Yea, I agree that it would be better to land mixins removal for this component first and rebase this one on top of it.
I'll look up @empyrical's work prior to that, to see what issues I can expect :D

TheSavior: The easiest thing to do is probably to do something like:

```
this._scrollResponder = new ScrollResponder();
```

And then in `ScrollView`'s `componentWillMount` call `this._scrollResponder.componentWillMount()`, same for `componentWillUnmount`, and then replace these lines: https://github.com/facebook/react-native/blob/master/Libraries/Components/ScrollView/ScrollView.js#L915-L936

with: 

```
const props = {
  // ... other props
  ...this._scrollResponder.getHandlers(),
  // ... other props
}
```

This should let as much as possible stay exactly the same. In fact changing the ScrollResponder in that way could probably be done without even switching off of createReactClass in the same PR. 👍

thymikee: If it's gonna be this little change, I could squeeze it into this PR, what do you think?

TheSavior: We've had other things break when making changes to this kind of stuff. I'd really rather doing small PRs for this so that things can be bisected and reverted easily without necessarily having to revert the whole thing.

thymikee: So, this makes sense?:
1) get rid of mixin in ScrollView but keep the createReactClass 
2) rebase this and change to class component

TheSavior: It potentially makes sense to do one between that just converts it to a class with the existing typings. We’ve run into problems with how we’ve handled migrating the native refs at that point which has caused lots of headaches. Having it be separate would probably make it easier to carefully test and review.

TheSavior: Sorry for what seems like a lot of overhead here. We just got burned recently and I want to help make sure we don’t repeat the same issues and that you get to learn from our problems. :-)

thymikee: Oh yea, I'm totally ok with that. It caused me headaches when I was grabbing components by name/display name in tests (bad practice to do so tho) or snapshotting them :D

thymikee:
Updated the diff, but didn't test it on RNTester yet. Feel free to comment on Flow typings though :)

cpojer:
Let's see if this passes Flow internally.

cpojer:
Imported to see latest internal Flow errors and discussed with @thymikee. Will leave this PR for @RSNara to land.

thymikee:
@RSNara @TheSavior I've tested this on RNTester for unexpected regressions and it seems just fine. Typings also work, e.g. when you add `.flowconfig` to RNTester directory.

There are still open questions (e.g. whether `getScrollResponder()` from non-ScrollView components to return `?ScrollView` as it is currently or the real `ScrollView & ScrollResponder.Mixin`?) so I'd like your input.

PR #22302, opened 2018-11-15
Android: Adding sendIntent on Linking module

ferrannp:
This PR implements "Add a standardized way to send intents on Android" discussed in https://github.com/react-native-community/discussions-and-proposals/issues/34.

Test Plan:
----------
Linking section of RNTester on Android (on iOS that part should not be there). See:

| Battery        | Facebook notifications           | 
| ------------- |:-------------:| 
| battery | facebook

Changelog:
----------
[Android] [Added] - Add `sendIntent` to Linking module

### DIFF
--- a/Libraries/Linking/Linking.js
+++ b/Libraries/Linking/Linking.js
@@ -81,6 +81,20 @@ class Linking extends NativeEventEmitter {
     return LinkingManager.getInitialURL();
   }
 
+  /*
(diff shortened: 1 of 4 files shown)


cpojer:
Thank you for this PR. Based on the discussion it seems that we have consensus that we want this feature in RN, so I'm gonna ship it and let you iterate on it with future pull requests.

Issue #22310, opened 2018-11-16
Android projects with 'apply plugin: "com.android.library"' now fail to build


jtreanor:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
Run `react-native info` in your terminal and paste its contents here.

```
React Native Environment Info:
    System:
      OS: macOS 10.14.1
      CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
      Memory: 279.43 MB / 32.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.12.0 - ~/.nvm/versions/node/v10.12.0/bin/node
      Yarn: 1.10.1 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v10.12.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 24, 26, 27, 28
        Build Tools: 27.0.0, 27.0.3, 28.0.2, 28.0.3
        System Images: android-27 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.1 => 16.6.1
      react-native: 0.57.5 => 0.57.5
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-create-library: 3.1.2
```

## Description

Any Android React Native project which uses `apply plugin: "com.android.library"` currently fails to build. The problem is that a recent change (https://github.com/facebook/react-native/pull/20526) to `react.gradle` now uses `android.applicationVariants` when configuring the project. This is not present for library projects, so it fails.

Prior to this changes, Android library projects could be build successfully, so this is a regression.

## Reproducible Demo

I have put a simple project to demonstrate the issue here: https://github.com/jtreanor/sample-library-react-native.

This project is the generated content of `react-native init SampleLibrary` with the only change being modifying `build.gradle` to add `apply plugin: "com.android.library"`

[...]

PR #22312, opened 2018-11-16
Fix build error for Android projects that use `apply plugin: "com.android.library"`

jtreanor:
This PR allows Android projects that use `apply plugin: "com.android.library"` to build successfully. A recent regression caused by https://github.com/facebook/react-native/pull/20526 means that building one of these projects always fails with this error:

```

* Where:
Script '/Users/james/src/SampleLibrary/node_modules/react-native/react.gradle' line: 15

* What went wrong:
A problem occurred configuring project ':app'.
> Could not get unknown property 'applicationVariants' for object of type com.android.build.gradle.LibraryExtension.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
```

This change updates `react.gradle` to cater to both application and library projects by selectively using `android.applicationVariants` or `android.libraryVariants`.

Fixes https://github.com/facebook/react-native/issues/22310.

Test Plan:
----------

Build any Android project with `apply plugin: "com.android.library"` using `./gradlew build` (I have a sample project here: https://github.com/jtreanor/sample-library-react-native). It will now succeed with these changes.

Changelog:
----------

[Android] [Fixed] - Fixed build error with Android library projects

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -17,7 +17,9 @@ def bundleConfig = config.bundleConfig ? "${reactRoot}/${config.bundleConfig}" :
 
 
 afterEvaluate {
-    android.applicationVariants.all { def variant ->
+    def isAndroidLibrary = plugins.hasPlugin("com.android.library")
+    def variants = isAndroidLibrary ? android.libraryVariants : android.applicationVariants
+    variants.all { def variant ->
         // Create variant and target names
         def targetName = variant.name.capitalize()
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks for the PR. This looks good to me.

PR #22317, opened 2018-11-16
Fix of the memory leaks in catalyst. 

dryganets:
ReactApplicationContext holds a reference to the CatalystInstanceImpl which holds references to the NativeModules. Many native modules store the references to objects holding the references to the Activity holders. The examples of such modules are UIManagerModule, NativeAnimatedModule and many others.

As result without this change caching the ReactApplicationContext causes Activity leaks. It is counter-intuitive.

Test Plan:
----------
Test with LeakCannery for memory leaks on JSEngine re-creation

Changelog:
----------
[Android] [Fixed] - ReactAppliclationContext leaks activity.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/NativeModuleRegistry.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/NativeModuleRegistry.java
@@ -83,6 +83,7 @@ public class NativeModuleRegistry {
       for (ModuleHolder module : mModules.values()) {
         module.destroy();
       }
+      mModules.clear();
     } finally {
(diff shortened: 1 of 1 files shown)


dryganets:
We found problems related to this commit in production. Not really high rate, but still.

There is no protection for getModule calls and in some cases, it is used in the async context. For example network requests sending events by  calling
```
reactContext.getNativeModule(UIManagerModule.class)
      .getEventDispatcher()
```
The module might not exist if JSEngine is already destroyed and we have a cleanup in place.

We got to fix a root cause - sanitize the ReactApplicationContext usage as it leaks the JSEngine if cached.

PR #22319, opened 2018-11-16
Add back TextInput.State

janicduplessis:
9ea129517922d8a23a5c81128e4dcf97f04a5129 forgot to add back the static `TextInput.State`.

Test Plan:
----------
Tested in an app that crashed because of this

Changelog:
----------
[General] [Fixed] - Add back TextInput.State


janicduplessis:
Looks like @sahrens backed out the diff, you could also include this fix when relanding it.

PR #22330, opened 2018-11-18
JSBigString: Explicitly include unistd.h

empyrical:
`JSBigString` is using functions from `unistd.h`, like `getpagesize`, `dup`, `open`, etc. but was not directly including it.

It was being included from inside the glog `logging.h` header, which in turn was getting included by the Folly headers `JSBigString` was using.

This was discovered while building CxxReact with a custom shimmed Glog.

Test Plan:
----------
No build regressions have been noticed.

Changelog:
----------

[General] [Fixed] - `JSBigString`: Explicitly included dependant header `unistd.h` (minor)

### DIFF
--- a/ReactCommon/cxxreact/JSBigString.h
+++ b/ReactCommon/cxxreact/JSBigString.h
@@ -7,6 +7,7 @@
 
 #include <fcntl.h>
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


empyrical:
Do you know why this failed to land internally?

PR #22334, opened 2018-11-18
[Do not land] test new jsc

gengjiawen:
don't land it now, it contains binary for test.

Showcase: https://github.com/gengjiawen/ReactNativeNewJsc.

Details not handled:
* remove Promise polyfill , not sure iOS need it or not cc @empyrical 
* `npm postinstall` hook for unbundle the jsc so
* maybe not bundle `libc++_shared.so` see https://github.com/facebook/react-native/pull/22263.


gengjiawen:
nope, it can close.

PR #22337, opened 2018-11-18
Extract out CLI 

grabbou:
Continuation of https://github.com/facebook/react-native/pull/22174 with an exception that `local-cli` folder is left in React Native repository to keep Facebook internal and React Native calls still working.

Separate strategy should be developed to remove all uses of `local-cli` in favor of dedicated utilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,9 +142,6 @@
     "build-ios-e2e": "detox build -c ios.sim.release",
     "test-ios-e2e": "detox test -c ios.sim.release --cleanup"
   },
-  "bin": {
-    "react-native": "local-cli/wrong-react-native.js"
-  },
   "peerDependencies": {
     "react": "16.6.0-alpha.8af6728"
   },
@@ -191,6 +188,7 @@
     "prop-types": "^15.5.8",
     "react-clone-referenced-element": "^1.0.1",
     "react-devtools-core": "^3.4.0",
+    "react-native-local-cli": "1.0.0-alpha.4",
     "regenerator-runtime": "^0.11.0",
     "rimraf": "^2.5.4",
     "semver": "^5.0.3",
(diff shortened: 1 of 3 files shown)


grabbou:
Few questions:

1) Are we going to communicate internally and externally that `local-cli` folder is "read-only" and no edits should be performed? 

2) How are we going to keep `local-cli` and `react-native-local-cli` in sync? Should maintainers of the latter keep an eye on the `local-cli` and make sure all changes (if any) are reapplied before a new version is cut? This is especially important in case of Metro team that might work on some files.

Action items:

Right now, the `react-native-local-cli` is set to `alpha.4` version which is behind `local-cli` folder by a few commits (not really relevant to this PR). The last one is f8040ed - all newer than Nov, 5 are not present. See it [here](https://github.com/facebook/react-native/commits/master/local-cli). 

The changes are not really relevant to this PR (from the stability perspective). My idea is to merge this PR and once it happens, I will send another one the same day to bump `react-native-local-cli` with all the updates applied to `local-cli` right before this PR has been merged.

package.json
```diff
@@ -142,9 +142,6 @@
     "build-ios-e2e": "detox build -c ios.sim.release",
     "test-ios-e2e": "detox test -c ios.sim.release --cleanup"
   },
-  "bin": {
```
TheSavior: Why don't we want this anymore? It seems like it is there to catch people from making mistakes. Is that mistake not possible anymore? Or we just think it isn't a big deal anymore?

grabbou: We have kept this behaviour, but made `react-native-local-cli` to have "bin" property instead. So just declaring in a different `npm` module (since React Native itself doesn't have an executable anymore)

PR #22339, opened 2018-11-19
Integrate ReactModuleSpecProcessor annotation processor into gradle

### LINKED ISSUE #11689: [Android] Why use annotation-processing by default and only with buck (not with gradle)?
It just let me confuse when I'm diving into react-native android source code, you use java annotation-processor in bellowing position:

- [ReactPropertyProcessor](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/processing/ReactPropertyProcessor.java) for generate PropertySetter.
- [ReactModuleSpecProcessor](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/module/processing/ReactModuleSpecProcessor.java) for generate ModuleInfoProvider.

But it's not added for gradle java compile procedure, but added to buck. And they are not enabled by default, we can see:

- ViewManagerSetter has a FallbackViewManagerSetter with java reflect-based annotation parse when it not found processing generated class (which has a '$$PropsSetter` suffix). See code in [ViewManagerPropertyUpdater.java#L65](https://github.com/facebook/react-native/blob/57f6cbb3dc12d2dcc3eedd9712c36bb3d39149a5/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManagerPropertyUpdater.java#L65).
- ModuleInfoProvider would only enabled in `mLazyNativeModulesEnabled` (see [XReactInstanceManagerImpl.java#L945](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/XReactInstanceManagerImpl.java#L945). But this value is never set true in source, so it would use reflect-based annotation too.

So why don't enable annotation-processing by default? Since react-native android import annotation-processing for almost a year and not a very new feature (commit [57f6cb](https://github.com/facebook/react-native/commit/57f6cbb3dc12d2dcc3eedd9712c36bb3d39149a5)), it's existence make your code more confusing.

Kudo:
Fixes #11689
Original ReactModuleSpecProcessor only being used with BUCK but not gradle.
This PR integrate the processors into gradle as well.
This will benefit LazyReactPackage and further TurboReactPackage.

Test Plan:
----------
`./gradlew :ReactAndroid:assembleDebug`
And see if there are generated java files.
`find . -name '*$$ReactModuleInfoProvider*'`
For example,
```
./ReactAndroid/build/generated/source/apt/debug/com/facebook/react/shell/MainReactPackage$$ReactModuleInfoProvider.java
./ReactAndroid/build/generated/source/apt/debug/com/facebook/react/CoreModulesPackage$$ReactModuleInfoProvider.java
./ReactAndroid/build/generated/source/apt/debug/com/facebook/react/DebugCorePackage$$ReactModuleInfoProvider.java
```

Furthermore, could attach debugger into LazyReactPackage.getReactModuleInfoProviderViaReflection() and see the code expecting the generated java file runs without problems.

Changelog:
----------

[Android] [Added] - Integrate ReactModuleSpecProcessor annotation processor into gradle


Kudo:
@axe-fb Please let me quote your questions at https://github.com/Kudo/react-native/commit/92acf11179d9036f4add9be5d43dc684796caf82#commitcomment-31331935 here.

> Is there no other way to not have a top level ReactAndroid-processor ? I was also looking at a solution like this, but did not really like having another top level folder.

Done it this PR, the main change is at setting.gradle to specify project root
`project(':ReactAndroid-processor').projectDir = new File(rootProject.projectDir, 'ReactAndroid/processor')`
BTW this will smell a little weird for gradle multiproject layout which used to expect a project a folder.
Anyway, to reduce top level folder, this is also acceptable.

> Do we really need to copy over all the files, though the annotation processor is running only on certain files ?

I just didn't find a way for gradle java plugin to only compile specific files and at last copy alternatively.
If there are someone who have clever solutions, please feel free to let me know.
 
> Note that with TurboModules, we would not need to run all this. Hence, would it make sense to generate this manually just for the CoreModule ?

If we generate all $$ReactModuleInfoProvider.java, does it benefit for bootstrap speed?
What is the advantage if we generate only for CoreModule?
I also found that current OSS BUCK seems to [lookup annotations from all java files](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/module/processing/BUCK#L17).

ReactAndroid/processor/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
gengjiawen: A java module won't need this.

Kudo: @gengjiawen Thank you. Removed the unused AndroidManifest.xml.

cpojer:
I'm sorry we didn't get to this PR earlier. It seems that now it is a bit outdated. Could you rebase it on top of master?

Kudo:
@cpojer It's original to respond @axe-fb for speed up TurboModule launch time.
However I am not quite sure if this is still feasible right now.
I am going to close this PR first.
Let me know and I will create a new PR if annotation processing to gradle is still a need.
Thank you.

PR #22372, opened 2018-11-21
export Props types for end users

chuckhacker:
Related to #22100 

When using Flow, it is handy to be able to inherit from `Props` type for components in user code because the typing can be stricter.

For example:

```
type SubmitButtonProps = $ReadOnly<{|
  ...TouchableHighlightProps,

  onPress: ?() => void,
  style?: ?ViewStyleProp,
  title: string,
|}>;
```

## Test Plan:
* [x]  npm run prettier
* [ ]  npm run flow-check-ios
* [ ]  npm run flow-check-android
 
## Release Notes:
[GENERAL] [ENHANCEMENT] [Components/Touchable/TouchableBounce.js] - export Props Flow types
[GENERAL] [ENHANCEMENT] [Components/Touchable/TouchableHighlight.js] - export Props Flow types
[GENERAL] [ENHANCEMENT] [Components/Touchable/TouchableOpacity.js] - export Props Flow types

### DIFF
--- a/Libraries/Components/Touchable/TouchableBounce.js
+++ b/Libraries/Components/Touchable/TouchableBounce.js
@@ -33,7 +33,7 @@ type State = {
 
 const PRESS_RETENTION_OFFSET = {top: 20, left: 20, right: 20, bottom: 30};
 
-type Props = $ReadOnly<{|
+export type Props = $ReadOnly<{|
   ...TouchableWithoutFeedbackProps,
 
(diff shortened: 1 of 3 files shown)


TheSavior:
You can use Flow's [React.ElementProps](https://flow.org/en/docs/react/types/#toc-react-elementprops) utility type:

```
type TouchableBounceProps = React.ElementProps<typeof TouchableBounce>
```

PR #22374, opened 2018-11-21
[ScrollView] Remove mixins from ScrollView

thymikee:
As a part of https://github.com/facebook/react-native/pull/22301 it turned out that we need to first convert `ScrollView` to class component. As a first step to do so, here's removal of using `mixins` API, in favor of populating `_scrollResponder` field with `ScrollResponder.Mixin` (still used) methods.

Test Plan:
----------
Played around with RNTester app and looks ok, but I'd like more eyes on it.

Changelog:
----------

[ScrollView][Removed] – Remove mixins from ScrollView

cc @TheSavior

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -608,6 +608,16 @@ const ScrollView = createReactClass({
     this._scrollViewRef && this._scrollViewRef.setNativeProps(props);
   },
 
+  /**
+   * Returns a reference to the underlying scroll responder, which supports
+   * operations like `scrollTo`. All ScrollView-like components should
+   * implement this method so that they can be composed while providing access
+   * to the underlying scroll responder's methods.
+   */
+  getScrollResponder: function(): ScrollView {
+    return this;
+  },
+
   getScrollableNode: function(): any {
     return ReactNative.findNodeHandle(this._scrollViewRef);
   },
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -598,7 +598,6 @@ const ScrollView = createReactClass({
 
   componentWillUnmount: function() {
     this._scrollResponder.componentWillUnmount();
-
     if (this._scrollAnimatedValueAttachment) {
       this._scrollAnimatedValueAttachment.detach();
     }
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -20,7 +20,6 @@ const StyleSheet = require('StyleSheet');
 const View = require('View');
 const InternalScrollViewType = require('InternalScrollViewType');
 
-const createReactClass = require('create-react-class');
 const dismissKeyboard = require('dismissKeyboard');
 const flattenStyle = require('flattenStyle');
 const invariant = require('fbjs/lib/invariant');
@@ -514,6 +513,10 @@ export type Props = $ReadOnly<{|
   children?: React.Node,
 |}>;
 
+type State = {|
+  layoutHeight: ?number,
+|};
+
 /**
  * Component that wraps platform ScrollView while providing
  * integration with touch locking "responder" system.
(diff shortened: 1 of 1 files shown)


Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -551,22 +551,33 @@ export type Props = $ReadOnly<{|
...
   _headerLayoutYs: (new Map(): Map<string, number>),
+  _createScrollResponder: function() {
+    for (const key in ScrollResponder.Mixin) {
+      if (typeof ScrollResponder.Mixin[key] === 'function') {
+        this._scrollResponder[key] = ScrollResponder.Mixin[key].bind(
+          this._scrollResponder,
+        );
+      }
+    }
+    this._scrollResponder.props = this.props;
```
TheSavior: This seems like it would be problematic if the props passed to ScrollView change

RSNara: Yeah, this is pretty brittle.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -551,22 +551,33 @@ export type Props = $ReadOnly<{|
  */
 const ScrollView = createReactClass({
   displayName: 'ScrollView',
-  mixins: [ScrollResponder.Mixin],
-
+  _scrollResponder: {},
```
RSNara: It might be a good idea to just turn this into a React.Component subclass right now: 

```JavaScript
const ScrollResponder = require('ScrollResponder');

function createScrollResponder(node: React.ElementRef<typeof ScrollView>): typeof ScrollResponder.Mixin {
  const scrollResponder = {... ScrollResponder.Mixin};
  for (const key of Object.keys(ScrollResponder.Mixin)) {
    if (typeof scrollResponder[key] === 'function') {
      scrollResponder[key] = scrollResponder[key].bind(node);
    }
  }
  return scrollResponder;
}

class ScrollView extends React.Component<Props, State> {
  _scrollResponder: typeof ScrollResponder.Mixin = createScrollResponder(this);
}
```

Then, we change all references to scrollResponder methods within `ScrollView` to use `this._scrollResponder`

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -551,22 +551,33 @@ export type Props = $ReadOnly<{|
...
     0,
   ): AnimatedImplementation.Value),
   _scrollAnimatedValueAttachment: (null: ?{detach: () => void}),
   _stickyHeaderRefs: (new Map(): Map<number, ScrollViewStickyHeader>),
   _headerLayoutYs: (new Map(): Map<string, number>),
+  _createScrollResponder: function() {
+    for (const key in ScrollResponder.Mixin) {
+      if (typeof ScrollResponder.Mixin[key] === 'function') {
+        this._scrollResponder[key] = ScrollResponder.Mixin[key].bind(
+          this._scrollResponder,
```
RSNara: To avoid having to proxy `props` and `state`, you should just bind this to the component instance.

Libraries/Components/ScrollView/ScrollView.js
RSNara: One way you could get around this is to define a getter. 

```JavaScript
Object.defineProperty(this._scrollResponder, 'props', {
  get: () => {
    return this.props;
  }
})
```

But I'm not sure if that's a good idea because it'll be difficult for flow to type check this.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -551,22 +551,33 @@ export type Props = $ReadOnly<{|
...
+        );
+      }
+    }
+    this._scrollResponder.props = this.props;
+    this._scrollResponder.state = ScrollResponder.Mixin.scrollResponderMixinGetInitialState();
+  },
   getInitialState: function() {
+    this._createScrollResponder();
```
RSNara: I'm not too sure it's a good idea to hack `getInitialState` to initialize the scrollResponder. What we really need is a constructor, and we can get that by converting this component to a class, which shouldn't take too much time. I think we should just do that instead. Besides, we're going to have to remove this code eventually when we convert to a class down the line anyways.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -676,7 +688,7 @@ const ScrollView = createReactClass({
    * @platform ios
    */
   flashScrollIndicators: function() {
-    this.getScrollResponder().scrollResponderFlashScrollIndicators();
+    this._scrollResponder.scrollResponderFlashScrollIndicators();
```
RSNara: I think we should just change `getScrollResponder()` to return `this._scrollResponder`.

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -551,22 +551,33 @@ export type Props = $ReadOnly<{|
...
   getInitialState: function() {
+    this._createScrollResponder();
     return {
-      ...this.scrollResponderMixinGetInitialState(),
       layoutHeight: null,
     };
   },
 
   UNSAFE_componentWillMount: function() {
+    this._scrollResponder.UNSAFE_componentWillMount();
```
RSNara:
```JavaScript
this.getScrollResponder().UNSAFE_componentWillMount();
```

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -586,6 +597,7 @@ const ScrollView = createReactClass({
   },
 
   componentWillUnmount: function() {
+    this._scrollResponder.componentWillUnmount();
```
RSNara:
```JavaScript
this.getScrollResponder(). componentWillUnmount();
```

Libraries/Components/ScrollView/ScrollView.js
thymikee: Not sure about that. According to the comments `getScrollResponder()` is used to e.g. fire `scrollTo()` which is not available in `ScrollResponder.Mixin`. Unless they're wrong (there's no actual usage of `getScrollResponder()` in this repo, so it may relate to internal FB callsites)

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -549,24 +552,31 @@ export type Props = $ReadOnly<{|
...
-  mixins: [ScrollResponder.Mixin],
-
-  _scrollAnimatedValue: (new AnimatedImplementation.Value(
+class ScrollView extends React.Component<Props, State> {
+  _createScrollResponder = () => {
+    const scrollResponder = {...ScrollResponder.Mixin};
+    for (const key in ScrollResponder.Mixin) {
+      if (typeof scrollResponder[key] === 'function') {
+        scrollResponder[key] = scrollResponder[key].bind(this);
+        (this: any)[key] = scrollResponder[key].bind(this);
```
thymikee: Turning this to class component turned out pretty straightforward. However, problem with binding `scrollResponder` to `ScrollView` is that methods from mixin will refer to `ScrollView` context, while they were not assigned to the ScrollView instance, but to `this._scrollResponder`. To mitigate that, I added a hack that copies methods from mixin back to the `ScrollView`. Wonder if it makes sense to actually do only that and only put lifecycle methods (componentWillMount/componentWillUnmount) inside `this._scrollResponder`

thymikee: Yes, I wanted to address that, but I left it in a, let's say, "request for comments" state and lost track on this issue – this was the reason I left the `this._scrollResponder` so we can differentiate on state and lifecycle of mixin vs component. This is what I meant while writing:

> Wonder if it makes sense to actually do only that and only put lifecycle methods (componentWillMount/componentWillUnmount) inside this._scrollResponder

So I think it's gonna be safer to revert this change you landed locally and get this sorted here (unless you'd like to to fix it on your own while landing it, then be my guest :D)

PR #22376, opened 2018-11-21
Make type of RefreshControl.onRefresh less strict

janicduplessis:
I think a common pattern is to use async functions with the onRefresh prop to reload data. I made the type `mixed` since we don't care about what the function returns. Could also be `void | Promise<void>` I haven't really seen a pattern for those in the codebase.

Test Plan:
----------
Run flow

Changelog:
----------
[General] [Fixed] - Make type of RefreshControl.onRefresh less strict

### DIFF
--- a/Libraries/Components/RefreshControl/RefreshControl.js
+++ b/Libraries/Components/RefreshControl/RefreshControl.js
@@ -87,7 +87,7 @@ export type RefreshControlProps = $ReadOnly<{|
   /**
    * Called when the view starts refreshing.
    */
-  onRefresh?: ?() => void,
+  onRefresh?: ?() => mixed,
 
   /**
    * Whether the view should be indicating an active refresh.


yungsters:
@TheSavior, do you an opinion for `mixed` vs. `void | Promise<void>`? We should consolidate on one or the other. I've been using the latter internally, but I can see why it might seem burdensome.

TheSavior:
Honestly, I go back and forth myself. 

I like the explicitness of void, and that it allows us to change the return type in the future without a bunch of type refinements. 

However, having a return type of `mixed` feels more "correct" and would be what the JavaScript community expects. 

We should probably go with `mixed` if we had to pick one to be consistent with. cc @RSNara

janicduplessis:
@TheSavior I think mixed is slightly more flexible since you could use a function that returns something directly instead of having to wrap it inside another function to ignore the return value. 

Should we just land this for now? I don't really have time to go through all call sites to cleanup the callback formats we use but that was the only one that surfaced an error in my app.

TheSavior:
Yeah, sorry. Forgot to land this before. Thanks for the ping.

PR #22377, opened 2018-11-22
Fix ReactAndroid build break.

Kudo:
Summary:
  During C++ build, we need the libjsc.so.
  But arm64-v8a and x86_64 libjsc.so are in different path and this error raised:

    Android NDK: ERROR:/home/circleci/react-native/ReactAndroid/build/third-party-ndk/jsc/Android.mk:jsc: LOCAL_SRC_FILES points to a missing file
    /opt/ndk/android-ndk-r17c/build/core/prebuilt-library.mk:45: *** Android NDK: Aborting    .  Stop.
    Android NDK: Check that /home/circleci/react-native/ReactAndroid/build/third-party-ndk/jsc/jni/arm64-v8a/libjsc.so exists  or that its path is correct

  The commit moves prebuilt libjsc.so into
  ReactAndroid/src/main/jni/third-party/jsc/jni and let ndkbuild script find the prebuilt libjsc.so.
  For AAR packaging, modify the jniLibs so that gradle android library plugin could find the prebuilt libjsc.so as well.

Test Plan:
----------

1. Check CI test_android is green.
2. Check AAR contains two prebuilt libjsc.so
```
./jni/arm64-v8a/libjsc.so
./jni/x86_64/libjsc.so
```

Changelog:
----------

[Android] [Fixed] - Fix ReactAndroid build break.

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -162,7 +162,7 @@ task prepareJSC(dependsOn: dependenciesPath ? [] : [downloadJSCHeaders]) << {
     copy {
         from zipTree(configurations.compile.fileCollection { dep -> dep.name == 'android-jsc' }.singleFile)
         from dependenciesPath ? "$dependenciesPath/jsc-headers" : {downloadJSCHeaders.dest}
-        from 'src/main/jni/third-party/jsc/Android.mk'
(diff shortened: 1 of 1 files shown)


dulmandakh:
@facebook-github-bot shipit

PR #22379, opened 2018-11-22
Fix allocating Buffer in early commit

radeno:
Commit https://github.com/facebook/react-native/commit/d9c2cdae41f41d879351516eeadd1fb608210242 brings compatibility with Node 10, but replaced allocating buffer incorrectly.
@dulmandakh made two comments and shows how to do it based on NodeJS documentation https://nodejs.org/api/buffer.html#buffer_new_buffer_size

This PR just fixes that commit.

Test Plan:
----------
CI Green.

Changelog:
----------

No changelog, it is internal fix.


dulmandakh:
@facebook-github-bot shipit

PR #22394, opened 2018-11-23
Fix Xcode 10 builds (broken by folly upgrade)

radko93:
[Folly upgrade](https://github.com/facebook/react-native/commit/a70625abd7bf4fba3dafb8a969a73854b7ddcd42) introduced changes that have to be applied to `Install Third Party` script in order to use `New build system` from Xcode 10. Unfortunately, this might happen again if someone changes folly. Also removes non-existent files from folly podspec.

Test Plan:
----------
Run RNTester before applying this change, apply this commit and try the build again.

Release notes:
----------
[IOS] [FIXED] - Fixed Xcode 10 build broken by folly upgrade


kelset:
@facebook-github-bot shipit

PR #22412, opened 2018-11-26
Fix crash in nativeInjectHMRUpdate

### LINKED ISSUE #22116: Hot Reloading causes crash on iOS
## Environment
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
      Memory: 5.93 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.12.0 - /usr/local/bin/node
      Yarn: 1.12.1 - /usr/local/bin/yarn
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react-native: 0.57.4 => 0.57.4 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-rename: 2.3.1

## Description
under RN 0.57.2-4, when I turn on Hot Reloading, the iPhone emulator crashes whenever I save the file.The same problem occurs when I create a new RN0.57.4 project. I initially suspected that it was an XCode or emulator problem, but it didn't matter with the XCode version and the emulator version; however, Hot Reloading works fine under RN0.57.1.

vovkasm:
Fix crash in nativeInjectHMRUpdate if it is called with only one argument from JS

Fixes #22116

For more deep explanation, see comments thread from here: https://github.com/facebook/react-native/issues/22116#issuecomment-440420713

Environment:
-------------

This fix dedicated for `0.57-stable` branch, because crashed code was part of old JSC executor, that was replaced in master with new JSI code.

Test Plan:
----------
Ensure that application no more crash when enable HMR from dev menu.

Changelog:
----------

[iOS] [Fixed] - Fix application crash when HMR enabled

### DIFF
--- a/ReactCommon/cxxreact/JSCExecutor.cpp
+++ b/ReactCommon/cxxreact/JSCExecutor.cpp
@@ -113,7 +113,7 @@ static JSValueRef nativeInjectHMRUpdate(
     const JSValueRef arguments[],
     JSValueRef* exception) {
   String execJSString = Value(ctx, arguments[0]).toString();
-  String jsURL = Value(ctx, arguments[1]).toString();
+  String jsURL = argumentCount > 1 ? Value(ctx, arguments[1]).toString() : String();
   evaluateScript(ctx, execJSString, jsURL);
   return Value::makeUndefined(ctx);
 }


ReactCommon/cxxreact/JSCExecutor.cpp
```diff
@@ -113,7 +113,7 @@ static JSValueRef nativeInjectHMRUpdate(
     const JSValueRef arguments[],
     JSValueRef* exception) {
   String execJSString = Value(ctx, arguments[0]).toString();
-  String jsURL = Value(ctx, arguments[1]).toString();
+  String jsURL = argumentCount > 1 ? Value(ctx, arguments[1]).toString() : String();
```
kelset: would you mind just adding a comment for our future selves, something like

`temporary workaround for 0.57 and Metro 48`

so that if I encounter any issue in future releases/cherry-picks I can quickly remember what this was about?

kelset:
Thanks for being so quick! Super appreciated :) If you can add the comment it would be perfect, in the meantime I'll have it double-checked & tested in my release :)

(btw, if you are on twitter, can you shoot me a [DM over there](https://twitter.com/Kelset)?)

ReactCommon/cxxreact/JSCExecutor.cpp
vovkasm: Done :-)

PR #22430, opened 2018-11-27
Fix Modal size on Xiaomi Mi 8

JChord:
Test Plan:
----------
Before:
img_5007

After:
img_5006

Changelog:
----------
[Android] [Fixed] - Modal background size error on Xiaomi Mi 8


JChord:
> Thanks for the PR! I'd like to understand a bit more what is going on here. Particularly, why is this only an issue on Mi phones?
> 
> Your test plan does not mention any testing done on non-Mi devices.

Hide the navigation bar in XiaoMi 8 will reproduce the bug, but not reproduced in my Huawei P9.

PR #22450, opened 2018-11-28
Respect GIF metadata for loop counts

EdwardDrapkin:
Right now RCTGIFImageDecoder is hard coded to loop "forever" when it encounters a gif with a loop count set to be zero.  Android already works correctly... or iOS works correctly and Android is broken.


hramos:
Thanks for the PR! It looks like https://github.com/facebook/react-native/pull/21999 fixed this issue, and it's now on master as of de759b949e4aa4904fd60a2fcbb874a3c857b50c.

PR #22468, opened 2018-12-01
Make RNTester build and pass again

TheSavior:
Detox was failing because of build errors due to the new xcode10 build system. These errors were fixed by @RSNara in https://github.com/facebook/react-native/commit/b7349f9857df331e4c618256158662ea1f819ed5 but this callsite was missed.

Test Plan:
----------
I ran `yarn build-ios-e2e && yarn test-ios-e2e`. 

Before this change I got build errors like:

```
1) Target 'React-tvOS' (project 'React') has copy command from 'react-native/React/Views/RCTRefreshControlManager.h' to 'react-native/RNTester/build/Build/Products/Release-iphonesimulator/include/React/RCTRefreshControlManager.h'
2) Target 'React' (project 'React') has copy command from 'react-native/React/Views/RCTRefreshControlManager.h' to 'react-native/RNTester/build/Build/Products/Release-iphonesimulator/include/React/RCTRefreshControlManager.h'
```

screen shot 2018-11-30 at 4 50 13 pm

Changelog:
----------

[General] [Fixed] - RNTester runs and passes again in CI

### DIFF
--- a/RNTester/e2e/test-init.js
+++ b/RNTester/e2e/test-init.js
@@ -8,7 +8,7 @@
 /* eslint-env jasmine */
 
 const detox = require('detox');
-const config = require('../../../package.json').detox;
+const config = require('../../package.json').detox;
 const adapter = require('detox/runners/jest/adapter');
 
 jest.setTimeout(480000);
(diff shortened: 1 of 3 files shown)


RNTester/e2e/test-init.js
```diff
@@ -8,7 +8,7 @@
 /* eslint-env jasmine */
```
TheSavior: I moved this file because jest requires all files in the `__tests__` directory to have tests. I believe this is getting fixed soon but this makes things pass until then.

PR #22473, opened 2018-12-01
RCTSafeAreaShadowView does not correctly adjust for RTL when setting padding

chbota:
Issue
----
`RCTSafeAreaShadowView` currently sets `paddingLeft` and `paddingRight`:

```
UIEdgeInsets insets = localData.insets;

  super.paddingLeft = (YGValue){insets.left, YGUnitPoint};
  super.paddingRight = (YGValue){insets.right, YGUnitPoint};
  super.paddingTop = (YGValue){insets.top, YGUnitPoint};
super.paddingBottom = (YGValue){insets.bottom, YGUnitPoint};
```

But, when `RCTShadowView` propagates the padding settings to Yoga, it adjusts for RTL when `doLeftAndRightSwapInRTL` is set to true:

```
static void RCTProcessMetaPropsPadding(const YGValue metaProps[META_PROP_COUNT], YGNodeRef node) {
  if (![[RCTI18nUtil sharedInstance] doLeftAndRightSwapInRTL]) {
    RCT_SET_YGVALUE(metaProps[META_PROP_START], YGNodeStyleSetPadding, node, YGEdgeStart);
    RCT_SET_YGVALUE(metaProps[META_PROP_END], YGNodeStyleSetPadding, node, YGEdgeEnd);
    RCT_SET_YGVALUE(metaProps[META_PROP_LEFT], YGNodeStyleSetPadding, node, YGEdgeLeft);
    RCT_SET_YGVALUE(metaProps[META_PROP_RIGHT], YGNodeStyleSetPadding, node, YGEdgeRight);
  } else {
    YGValue start = metaProps[META_PROP_START].unit == YGUnitUndefined ? metaProps[META_PROP_LEFT] : metaProps[META_PROP_START];
    YGValue end = metaProps[META_PROP_END].unit == YGUnitUndefined ? metaProps[META_PROP_RIGHT] : metaProps[META_PROP_END];
    RCT_SET_YGVALUE(start, YGNodeStyleSetPadding, node, YGEdgeStart);
    RCT_SET_YGVALUE(end, YGNodeStyleSetPadding, node, YGEdgeEnd);
}
...

```

Because `RCTSafeAreaShadowView` only sets `paddingLeft` and `paddingRight`, and not `paddingStart` and `paddingEnd`, `paddingLeft` is always treated as `YGEdgeStart` and `paddingRight` is always treated as `YGEdgeEnd`. This is incorrect in RTL scenarios, where left == end and right == start.

Fix
----
When `doLeftAndRightSwapInRTL` is set to true, we set `paddingStart` and `paddingEnd`, rather than `paddingLeft/Right`, in `RCTSafeAreaShadowView`. When `doLeftAndRightSwapInRTL` is set to false, we maintain the previous behavior.

Test Plan
----
Validated SafeAreaView works as it did before when in LTR and when `doLeftAndRightSwapInRTL` == false.

Validated that SafeAreaView now works correctly in RTL, with padding no longer being swapped.

Changelog
----
[iOS][Fixed] - SafeAreaView padding is incorrect in RTL

### DIFF
--- a/React/Views/SafeAreaView/RCTSafeAreaShadowView.m
+++ b/React/Views/SafeAreaView/RCTSafeAreaShadowView.m
@@ -11,6 +11,7 @@
 #import <yoga/Yoga.h>
 
 #import "RCTSafeAreaViewLocalData.h"
+#import "RCTI18nUtil.h"
 
 @implementation RCTSafeAreaShadowView
 
@@ -21,12 +22,23 @@
 
   UIEdgeInsets insets = localData.insets;
 
-  super.paddingLeft = (YGValue){insets.left, YGUnitPoint};
-  super.paddingRight = (YGValue){insets.right, YGUnitPoint};
   super.paddingTop = (YGValue){insets.top, YGUnitPoint};
   super.paddingBottom = (YGValue){insets.bottom, YGUnitPoint};
 
-  [self didSetProps:@[@"paddingLeft", @"paddingRight", @"paddingTop", @"paddingBottom"]];
+  if ([[RCTI18nUtil sharedInstance] doLeftAndRightSwapInRTL]) {
+    if ([[RCTI18nUtil sharedInstance] isRTL]) {
+      super.paddingStart = (YGValue){insets.right, YGUnitPoint};
+      super.paddingEnd = (YGValue){insets.left, YGUnitPoint};
+    } else {
+      super.paddingStart = (YGValue){insets.left, YGUnitPoint};
+      super.paddingEnd = (YGValue){insets.right, YGUnitPoint};
+    }
+    [self didSetProps:@[@"paddingStart", @"paddingEnd", @"paddingTop", @"paddingBottom"]];
+  } else {
+    super.paddingLeft = (YGValue){insets.left, YGUnitPoint};
+    super.paddingRight = (YGValue){insets.right, YGUnitPoint};
+    [self didSetProps:@[@"paddingLeft", @"paddingRight", @"paddingTop", @"paddingBottom"]];
+  }
 }
 
 /**


React/Views/SafeAreaView/RCTSafeAreaShadowView.m
```diff
@@ -21,12 +22,23 @@ - (void)setLocalData:(RCTSafeAreaViewLocalData *)localData
...
+    if ([[RCTI18nUtil sharedInstance] isRTL]) {
+      super.paddingStart = (YGValue){insets.right, YGUnitPoint};
+      super.paddingEnd = (YGValue){insets.left, YGUnitPoint};
+    } else {
+      super.paddingStart = (YGValue){insets.left, YGUnitPoint};
+      super.paddingEnd = (YGValue){insets.right, YGUnitPoint};
+    }
+    [self didSetProps:@[@"paddingStart", @"paddingEnd", @"paddingTop", @"paddingBottom"]];
+  } else {
+    super.paddingLeft = (YGValue){insets.left, YGUnitPoint};
```
shergin: paddingLeft and paddingRight always work regardless of `doLeftAndRightSwapInRTL`, so why not always use that?

React/Views/SafeAreaView/RCTSafeAreaShadowView.m
```diff
@@ -21,12 +22,23 @@ - (void)setLocalData:(RCTSafeAreaViewLocalData *)localData
...
   UIEdgeInsets insets = localData.insets;
 
-  super.paddingLeft = (YGValue){insets.left, YGUnitPoint};
-  super.paddingRight = (YGValue){insets.right, YGUnitPoint};
   super.paddingTop = (YGValue){insets.top, YGUnitPoint};
   super.paddingBottom = (YGValue){insets.bottom, YGUnitPoint};
 
-  [self didSetProps:@[@"paddingLeft", @"paddingRight", @"paddingTop", @"paddingBottom"]];
+  if ([[RCTI18nUtil sharedInstance] doLeftAndRightSwapInRTL]) {
+    if ([[RCTI18nUtil sharedInstance] isRTL]) {
```
shergin: ... and btw this is not proper way to check for RTL. That prop return a global setting which is not so useful because we have `direction` style.

And, actually, there is not way to check it properly here because the computed value of `display` prop depends on outer nodes and unabailable here.

So, see my next comment.

cpojer:
Hey @chbota,

do you mind following up on the comments by @shergin? It seems like there is a tiny amount of more work necessary :)

chbota:
Yeah totally. Just need to carve out some time to refresh my memory on these changes and take another look here. 😄

cpojer:
@chbota just wanted to ping you again about another iteration of this PR? Do you think you'll be able to get to it? Otherwise please close the PR.

chbota:
Thanks for the ping. I've blocked off some time this afternoon to get another iteration out

React/Views/SafeAreaView/RCTSafeAreaShadowView.m
chbota: I might be misunderstanding your point here, but I don't think `paddingLeft` and `paddingRight` will always work as we expect.

In the case where `doLeftAndRightSwapInRTL` == true, `RCTProcessMetaPropsPadding` will not set `paddingLeft` and `paddingRight` on the yoga node. Instead it will try to set `paddingStart` and `paddingEnd`, and fallback to `paddingLeft` and `paddingRight` if `paddingStart` and `paddingEnd` are not specified:

```
    YGValue start = metaProps[META_PROP_START].unit == YGUnitUndefined ? metaProps[META_PROP_LEFT] : metaProps[META_PROP_START];
    YGValue end = metaProps[META_PROP_END].unit == YGUnitUndefined ? metaProps[META_PROP_RIGHT] : metaProps[META_PROP_END];
    RCT_SET_YGVALUE(start, YGNodeStyleSetPadding, node, YGEdgeStart);
    RCT_SET_YGVALUE(end, YGNodeStyleSetPadding, node, YGEdgeEnd);
```

This behavior is fine in LTR, but not in RTL (in RTL, `paddingLeft` should be used for `paddingEnd`, `paddingRight` should be used for `paddingStart`).

I totally might be missing something though, so definitely sanity check my logic 😄

chbota: In other words, yes `paddingLeft` and `paddingRight` would work if `RCTProcessMetaPropsPadding` didn't flip them in RTL. It may be that the right fix is to change `RCTProcessMetaPropsPadding` or not use it in this scenario, but that was a less targeted fix so I opted not to take that path. But your comment about determining isRTL value suggests that such a fix would be a more robust option

chbota:
@shergin , I had a couple of follow ups based on your questions/feedback above. Could you take a look when you have a chance?

cpojer:
@chbota it's unclear to me what your questions were. Do you mind referring me to your questions so we can help unblock you and get this shipped?

React/Views/SafeAreaView/RCTSafeAreaShadowView.m
chbota: @cpojer

PR #22532, opened 2018-12-05
Extend reason message for `RCTFatalException`

### LINKED ISSUE #22530: iOS RCTFatal truncation preventing debugging
  - [ ] Use the latest React Native release: https://github.com/facebook/react-native/releases
(Not using the latest release yet, but have confirmed this issue exists in latest as well)

## Environment
```
    System:
      OS: macOS 10.14.1
      CPU: x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 23.93 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.2.0 - /usr/local/bin/node
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 27.0.3, 28.0.2
    IDEs:
    npmPackages:
      @types/react: ^16.4.7 => 16.4.9 
      @types/react-native: ^0.56.4 => 0.56.7 
      react: 16.4.1 => 16.4.1 
    npmGlobalPackages:
      create-react-native-app: 2.0.2
```

## Summary
*TLDR;* The very low "max message length" of 75 at [this line](https://github.com/facebook/react-native/blob/9d00d4d5bb0f312b7d907e2fc4005a73301dee9f/React/Base/RCTAssert.m#L131) in `RCTAssert.m` renders `RCTFatal` errors very difficult to debug.

## Description
What we're seeing is that errors captured by the iOS built-in crash analytics and by Sentry.io are using the `NSException.message` to describe the message and discarding the `name` value. As such, we're only seeing the 75 character truncated value. Given how much of that string is dedicated to boilerplate information about the exception, we wind up with exception logs like `RCTFatalException: Exception '*** -[__NSArrayM setObject:atIndexedSubscript:]: object cannot be nil' was thrown while invokin...` that truncate _just_ before the actually useful information for debugging would be printed.

There isn't an especially compelling reason that I know of for keeping this value truncated as short as 75 characters. (Indeed, IMO there isn't really a reason to truncate it at all, but I expect there are cases that demand that that I'm not aware of.) I'm going to make a branch and pull-request that will both extend this truncation length and add information to the `userinfo` dictionary on the exception.

## Reproducible Demo
Cause any exception and view the value in the RCTFatalException that is thrown.

zackzachariah:
## Background
Fixes #22530 

As described in the issue, the previous behavior for the `RCTFatal` macro was to truncate the `reason` on the resulting `NSException` to 75 characters. This would ensure the reason would fit on a single line, but resulted in issues debugging errors that occurred in the wild, as many crash logging tools (like Sentry) discard the `name` value of the exception and use the `reason` as their primary identifier. At 75 characters, useful information like the location of the error would usually be truncated.

## Detailed Changes
- [x] This extends the truncation threshold to 175 characters, which should be short enough to prevent full-screen-takeover length errors, but long enough to provide useful context to the error.
- [x] This adds a `userInfo` value to the resulting `NSException`. It copies over the `userInfo` from the `NSError` passed to the macro, and adds an "untruncated message" value that contains the untruncated version of the `NSException`'s reason.

## Changelog:
[iOS] [Changed] - RCTFatalExceptions now include more information in their reason and a userInfo.

### DIFF
--- a/React/Base/RCTAssert.m
+++ b/React/Base/RCTAssert.m
@@ -12,6 +12,7 @@ NSString *const RCTErrorDomain = @"RCTErrorDomain";
 NSString *const RCTJSStackTraceKey = @"RCTJSStackTraceKey";
 NSString *const RCTJSRawStackTraceKey = @"RCTJSRawStackTraceKey";
 NSString *const RCTFatalExceptionName = @"RCTFatalException";
+NSString *const RCTUntruncatedMessageKey = @"RCTUntruncatedMessageKey";
 
 static NSString *const RCTAssertFunctionStack = @"RCTAssertFunctionStack";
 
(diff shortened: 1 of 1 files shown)


zackzachariah:
CLA has been signed.

PR #22537, opened 2018-12-06
Add e2e tests, bug fixes for testIDs

rickhanlonii:
## Overview
This PR adds e2e tests for the Picker and DatePicker components.

While writing these tests, I also found and fixed two bugs where we wern't passing the `testID` down to the native components, so detox couldn't look them up. This confirms what was mentioned by @rotemmiz [here](https://github.com/wix/Detox/issues/798#issuecomment-401412276)

Test Plan:
----------

gif

Changelog:
----------
[General] [Fixed] - Fix testIDs not being passed in Picker and DatePicker

### DIFF
--- a/RNTester/e2e/config.json
+++ b/RNTester/e2e/config.json
@@ -1,5 +1,5 @@
 {
-  "setupTestFrameworkScriptFile" : "./test-init.js",
+  "setupFilesAfterEnv" : ["./test-init.js"],
   "testEnvironment": "node",
   "bail": true,
   "verbose": true
--- a/RNTester/e2e/config.json
+++ b/RNTester/e2e/config.json
@@ -1,5 +1,5 @@
 {
-  "setupFilesAfterEnv" : ["./test-init.js"],
+  "setupTestFrameworkScriptFile" : "./test-init.js",
   "testEnvironment": "node",
   "bail": true,
   "verbose": true
--- a/RNTester/js/DatePickerIOSExample.js
+++ b/RNTester/js/DatePickerIOSExample.js
@@ -46,10 +46,13 @@ class DatePickerExample extends React.Component<
     return (
       <View>
         <WithLabel label="Value:">
-          <Text>
+          <Text testID="date-and-time-indicator">
             {this.state.date.toLocaleDateString() +
               ' ' +
-              this.state.date.toLocaleTimeString()}
+              this.state.date.toLocaleTimeString([], {
+                hour: '2-digit',
+                minute: '2-digit',
+              })}
           </Text>
         </WithLabel>
         <WithLabel label="Timezone:">
@@ -62,6 +65,7 @@ class DatePickerExample extends React.Component<
         </WithLabel>
         <Heading label="Date + time picker" />
         <DatePickerIOS
+          testID="date-and-time"
           date={this.state.date}
           mode="datetime"
           timeZoneOffsetInMinutes={this.state.timeZoneOffsetInHours * 60}
@@ -69,6 +73,7 @@ class DatePickerExample extends React.Component<
         />
         <Heading label="Date picker" />
         <DatePickerIOS
+          testID="date-only"
           date={this.state.date}
           mode="date"
           timeZoneOffsetInMinutes={this.state.timeZoneOffsetInHours * 60}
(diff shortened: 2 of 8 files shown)


RNTester/js/DatePickerIOSExample.js
```diff
@@ -46,10 +46,13 @@ class DatePickerExample extends React.Component<
...
       <View>
         <WithLabel label="Value:">
-          <Text>
+          <Text testID="date-and-time-indicator">
             {this.state.date.toLocaleDateString() +
               ' ' +
-              this.state.date.toLocaleTimeString()}
+              this.state.date.toLocaleTimeString([], {
+                hour: '2-digit',
+                minute: '2-digit',
```
rickhanlonii: I switched the formatting here since having seconds is non-deterministic

RNTester/e2e/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "setupTestFrameworkScriptFile" : "./test-init.js",
+  "setupFilesAfterEnv" : ["./test-init.js"],
```
rickhanlonii: This is a Jest change that was printing a warning

TheSavior:
it is surprising to me that each example in the DatePickerIOS suite is so slow, more than twice as long as the other tests: 

Is `setColumnToValue` just really slow? Can you experiment a little and see if you can identify which command is taking so long?

```
PASS  RNTester/e2e/__tests__/Switch-test.js (40.372s)
  Switch
    ✓ Switch that starts on should switch (801ms)
    ✓ Switch that starts off should switch (704ms)
    ✓ disabled switch should not toggle (398ms)

detox[25343] INFO:  [DetoxServer.js] server listening on localhost:49370...
 PASS  RNTester/e2e/__tests__/DatePickerIOS-test.js (44.554s)
  DatePickerIOS
    ✓ Should change indicator with datetime picker (6351ms)
    ✓ Should change indicator with date-only picker (5990ms)
    ✓ Should change indicator with time-only picker (6282ms)

detox[25343] INFO:  [DetoxServer.js] server listening on localhost:49422...
 PASS  RNTester/e2e/__tests__/Button-test.js (34.574s)
  Button
    ✓ Simple button should be tappable (1546ms)
    ✓ Adjusted color button should be tappable (1909ms)
    ✓ Two buttons with JustifyContent:'space-between' should be tappable (3824ms)
    ✓ Disabled button should not interact (1271ms)
```

RNTester/js/RNTesterPage.js
```diff
@@ -41,7 +41,10 @@ class RNTesterPage extends React.Component<Props> {
     return (
       <View style={styles.container}>
         {title}
-        <ContentWrapper style={styles.wrapper} {...wrapperProps}>
+        <ContentWrapper
+          testID="scroll-view"
```
TheSavior: Should this have a more specific `testID`?

TheSavior:
Based on the video it seems like scrolling to the bottom could glitch out if the place it puts the cursor down is in the middle of the picker, scrolling the picker instead of the scrollview. Not sure if that is how detox works though. Can you see if you can cause that issue, if it can then it would be a footgun we'll for sure hit later when running this on a different device size or changing the contents of the view.

rickhanlonii:
I think it's slow in aggregate because they have 6 or 7 actions/assertion, each taking a good amount of time. Here's an analysis of work time per statement in the first test block (note I ran the analysis multiple times with similar results).

Legend for the statements tested:
```js
// 0, no actions/assertions

await expect(indicator).toBeVisible(); // 1
await expect(testElement).toBeVisible(); // 2

await testElement.setColumnToValue(0, 'Dec 4'); // 3
await testElement.setColumnToValue(1, '4'); // 4
await testElement.setColumnToValue(2, '10'); // 5
await testElement.setColumnToValue(3, 'AM'); // 6

await expect(indicator).toHaveText('12/4/2005 4:10 AM'); // 7
``` 

Results: 

Line | Method | Agg ms | Δ ms | Δ % 
----- | ----- | ----- | ----- | -----
  0 | - | 3 | - | -
  1 | toBeVisible | 768 | 765 | %%
  2 | toBeVisible | 1,224 | 456 | 59%
  3 | setColumnToValue | 2,046 | 822 | 67%
  4 | setColumnToValue | 2,349 | 303 | 15%
  5 | setColumnToValue | 3,366 |1,017 | 43%
  6 | setColumnToValue | 4,216 | 850 | 25%
  7 | toHaveText | 5,253 | 1,037 | 25%

rickhanlonii:
@TheSavior good find on the scroll issue, here's what happens

bug

I removed the test for now, but will revisit a generic filtering strategy to use on every test page instead of scrolling 👍

RNTester/e2e/config.json
cpojer: Can you do that in a separate commit?

PR #19809, opened 2018-06-19
[iOS][TextInput] Avoid using `-[UITextView setAttributedString:]` while user is typing

mandrigin:
## Motivation

iOS-specific.
For languages with complex input (such as Japanese or Chinese), a user has to type multiple characters that are then merged into a single one.
If `-[UITextView setAttributedString:]` is used while the user still types, it resets the input and characters are not being treated as typed together.

This PR avoids calling this method if possible, replacing it by just copying the attributes if the string has not been changed. That preserves the state and user can continue to type Korean or Chinese characters.

Fixes #19339

## Test Plan

Essentially, the steps to reproduce are described in [the issue](https://github.com/facebook/react-native/issues/19339):

1. Type some Korean characters in TextInput, such as "하늘". 
2. Then move the cursor to the beginning of the text, type "파란" this time.

Behaviour before this fix:
The text is rendered as "ㅍㅏㄹㅏㄴ하늘" (characters aren't combined)

Behaviour after this fix:
The text is rendered as "파란하늘" (characters are combined, the same behaviour is in vanilla iOS `UITextView`).

## Release Notes

[IOS] [BUGFIX] [TextView] - Fix Korean/Chinese/Japanese input for multiline TextView on iOS.

PR #22546, opened 2018-12-06
[iOS][TextInput] Apply the fix for CJK languages on single-line text fields.

### LINKED ISSUE #19339: ReactNative's state update via this.setState breaks text input mode for Korean, Chinese, Japanese characters in 0.54 and 0.55
## Environment
Environment:
  OS: macOS High Sierra 10.13.4
  Node: 6.11.0
  Yarn: Not Found
  npm: 3.10.10
  Watchman: Not Found

  react: ^16.3.1 => 16.3.2
  react-native: ^0.55.4 => 0.55.4

## Steps to Reproduce
This issue has been reported number of times already. #19313 #18403 #19104 #19337 #12599 #18379 #18767 #18260

However, most of them reported the issue as being the TextInput bug whereas the real bug seems to lie with ReactNative's view re-rendering logic. This report is to unify them under a more proper title so that someone who is more likely to recognize the root cause of this issue can pay attention.

If there is a <TextInput> tag in jsx file and its onChange or onChangeText callback has this.setState() call, Korean, Chinese, Japanese characters (which have multi-stage character input mode) are not input correctly.

In caes of Hangul(Korean), if the cursor is at the end of the text, the input is handled properly. However, if the cursor is at the beginning or is in the middle of the text, the input character parts are not combined as they should be. The parts of a character that should form a single character are input separately by themselves.

To reproduce, just type some Korean characters in TextInput, such as "하늘". Then move the cursor to the beginning of the text, and then type "파란" once more.

## Expected Behavior
The text should be input like "파란하늘"

## Actual Behavior
However, as of RN version 0.54 and 0.55, the text is input like "ㅍㅏㄹㅏㄴ하늘"

Chinese characters and Japanese kanji characters are input by first entering romanzi to specify the pronounciation and then the user select a character from a list of Chinese characters popup that appears. However, this popup does not appear as demonstrated in #18403.

Korean, Chineses, Japanese characters input is handled by multi-state IM mode. However, setState => view re-rendering cycle of ReactNative in v0.54 and v0.55 seems to reset the "character input is in progress" IM mode, thus causing incomplete and/or no characters to be entered into TextInput field.

This is a rather serious issue for Korean, Chinese, Japanese developers who have no options but to stick with v0.53 or below. (0.53 is particularly unstable version so I am stuck using v0.52.3) If anyone can guess where this issue might be stemming from, please share your opinions. It would be very appreciated. Thank you.

mandrigin:
Follow-up to https://github.com/facebook/react-native/pull/19809

This fix generalizes the `setAttributedString:` fix to single-line text fields.

Fixes #19339

Test Plan:
----------

⚠️ Possible areas of regressions: formatting of single-line text fields.

#### The test app

```
import {StyleSheet, Text, TextInput, View} from 'react-native';

type Props = {};
    constructor(props) {
        super(props);
        this.state = { text: '', textM: '' };
    }
    render() {
        return (
            <Text style={styles.instructions}>SINGLE_LINE</Text>
            <TextInput
                style={{height: 40, width: 200, borderWidth: 1}}
                onChangeText={(text) => this.setState({text: text})}
                value={this.state.text}
            />
            <Text style={styles.instructions}>MULTILINE</Text>
            <TextInput
                style={{height: 40, width: 200, borderWidth: 1}}
                multiline={true}
                onChangeText={(text) => this.setState({textM: text})}
                value={this.state.textM}
            />
            </View>
        );
    }
}

  container: {
    flex: 1,
    paddingTop: 100,
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  instructions: {
    textAlign: 'center',
    color: '#333333',
    marginBottom: 5,
    marginTop: 20,
  },
});
```

Essentially, the steps to reproduce are described in the issue:

* For both single-line and multi-line text fields:
    * Type some Korean characters in TextInput, such as "`하늘`" (buttons `ㅎ`,`ㅏ`,`ㄴ`,`ㅡ`,`ㄹ`).
    * Then move the cursor to the beginning of the text, type "`파란`" (buttons `ㅍ`,`ㅏ`,`ㄹ`,`ㅏ`,`ㄴ`) this time.

**Behaviour before this fix**
Actual text (Single-line): `ㅍㅏㄹㅏㄴ하늘`.
Actual text (Multi-line): `파란하늘`.
Expected text (both fields): `파란하늘`.
Characters aren't combined properly in single-line.

**Behaviour after this fix (correct)**
Actual text (Single-line): `파란하늘`.
Actual text (Multi-line): `파란하늘`.
Expected text (both fields): `파란하늘`.
Characters are combined correctly.

Changelog:
----------
[IOS] [BUGFIX] [TextView] - Fix Korean/Chinese/Japanese input for single-line TextView on iOS.


shergin:
@mandrigin I deeply appreciate your effort and I am existed about those fixes. We struggled with them for so long time! But we really need to do it right to make it sustainable.

Libraries/Text/TextInput/Singleline/RCTUITextField.m
```diff
@@ -107,6 +109,29 @@ - (CGRect)caretRectForPosition:(UITextPosition *)position
...
+- (void)setAttributedText:(NSAttributedString *)attributedText {
+  [_attributesHolder setAttributedString:attributedText];
+  [super setAttributedText:attributedText];
+}
+
+- (void)copyAttributesFrom:(NSAttributedString *)source {
+  [_attributesHolder setAttributedString:super.attributedText];
+}
+
+- (NSAttributedString *)attributedText {
```
mandrigin: @shergin it is easier to me to comment here, to show why I did it this way.
`UITextField` is a very basic UI element. It doesn't expose any API to just change the string attributes without re-setting the attributed string completely (and breaking CJK input). 
Luckily, it relies on it's own public (and overridable) property `attributedText` when rendering.

What I'm doing here is hijacking this property and replacing its value with the value I store locally, in NSMutableAttributedString, which I can replace without noticeable side-effects. So this override has to be here, if we want to continue to use UITextField. It's a hack and it can potentially break with new iOS releases, etc.

To be able to consolidate logic, we can just use UITextView everywhere, it provides all the necessary functionality as public APIs. What is the history between having 2 different implementations for single and multiline text views?

Libraries/Text/TextInput/Singleline/RCTUITextField.m
```diff
@@ -107,6 +109,29 @@ - (CGRect)caretRectForPosition:(UITextPosition *)position
   return [super caretRectForPosition:position];
 }
 
+#pragma mark - Fix for CJK Languages
+// Search for `GH-19339` to see all code affected by this fix.
+
+- (void)setAttributedText:(NSAttributedString *)attributedText {
+  [_attributesHolder setAttributedString:attributedText];
```
shergin: I mean why not to fallback to `copyAttributesFrom` here, if we see that `super.attributedText.string === attributedText`?

Libraries/Text/TextInput/Singleline/RCTUITextField.m
shergin: I mean why not consolidate all attribute specific-logic in this class making `attributedText` and `setAttributedText` behave as we want?

We have to have two different underlying components because they simply behave differently. There is no way to configure UITextView behave exactly as UITextField.

shergin: I mean we have an original problem: When we call `setAttributedText`, market fragment gets reset. So, let's add all checks that prevents that into both  `RCTUITextField` and `RCTUITextView` (iirc, now we have only one of them).

mandrigin: ah, okay, I mistread your initial comment.
So you mean, just duplicate the logic and encapsulate it inside `RCTUITextField` and `RCTUITextView` classes separately, so we can call `setAttributedString` safely on both `RCTUITextField` and `RCTUITextView`?

mandrigin:
@shergin I just simplified my fix to affect only one source file.

Libraries/Text/TextInput/Singleline/RCTUITextField.m
mandrigin: should be fixed now

Libraries/Text/TextInput/Singleline/RCTUITextField.m
mandrigin: should be fixed now

shergin: (Let's continue discussion here.)
Now we are abstracted from anything else, so we can work on it. 😄 

I think, it's great. ❤️ 

Nits:
* Instead of referring issues and commenting every single line, we have to leave one comment that describe a problem and (most importantly) the change in the behavior of the methods. That's more sustainable and easy to understand for people from the future which don't have a full context.
* Codestyle. (New line before `{` and spaces.)

Btw, I don't think it's only fix for CJK problem. It's more wide problem which happens when we are replacing some attributed string with completely new one without paying attention to what attributes were added to the string by active text input. So, if we can preserve those attributes, I bet full replacing will work with CJK. (The current solution is good enough though.)

mandrigin: pushed the fix to address this feedback

mandrigin: @shergin is there anything else I need to fix in this PR?

PeteTheHeat:
I'm going to have to revert this, it broke single line text inputs, sorry 😥

Repro:
In this playground https://gist.github.com/PeteTheHeat/52aef6290850b9840342e884325bdd0a type in the inputs, tap away, and watch them revert back to default text.

mandrigin:
Sure, I will take a look next week on what can be improved

alsun2001:
> I'm going to have to revert this, it broke single line text inputs, sorry 😥
> 
> Repro:
> In this playground https://gist.github.com/PeteTheHeat/52aef6290850b9840342e884325bdd0a type in the inputs, tap away, and watch them revert back to default text.

By the way, this issue repros on iOS 9 and iOS 10 only.

@mandrigin were you able to make any progress on this? I'm interested in a fix for a related issue where the keyboard doesn't automatically switch back from numbers to letters on typing space.

mandrigin:
@alsun2001 sorry, was a bit busy at first and then I was on vacation. I'm back, I'll take a look at it this week. Knowing that that is iOS 9 and iOS 10 issue only helps a lot, thanks!

mandrigin:
@PeteTheHeat @alsun2001 I can't reproduce the original (CJK input) issue in 0.59.3.
I tried iOS Simulator with iOS 10.0 and 12.2.
So, I guess it is fixed?

PeteTheHeat:
There have been quite a few small changes to text inputs recently, so I wouldn't be surprised if it was fixed. Thanks for following up with this!!

PR #22561, opened 2018-12-07
Making JSI_EXPORT macro definition conditional

mganandraj:
This change enables defining the macro at a more global and central location. And thereby allowing us to building this file using MSVC.

### DIFF
--- a/ReactCommon/jsi/jsi.h
+++ b/ReactCommon/jsi/jsi.h
@@ -13,7 +13,9 @@
 #include <string>
 #include <vector>
 
+#ifndef JSI_EXPORT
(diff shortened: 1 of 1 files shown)


TheSavior:
Can you add some more context to your summary for how this will be used? In some ways you are asking for this to become part of our public API so that other projects can define it differently. It would be good to know the use cases for that in a bit more detail.

mganandraj:
Yep. Andrew has perfectly explained the motivation behind the change. Essentially we want the macro to be redefinable.

RSNara:
cc @mhorowitz. Can you review this change?

PR #22570, opened 2018-12-08
Add Touchable E2E tests

CodingItWrong:
Adds some initial tests for Touchable*. It only tests the first screen worth of examples; in a separate PR I'll work on an alternate way to "scroll" to individual examples in tests, before I add tests for the rest of the Touchable examples.

On the live stream where I began writing these tests, I reorganized the "Touchable feedback examples" to the top of the list so it would be on-screen for testing. I didn't include this reorganization or test in this PR; that can be added in once the "alternative to scrolling" is added in, to avoid having to reorganize.

Changelog:
----------
[General] [Added] - Add E2E tests for Touchable

### DIFF
--- a/RNTester/e2e/__tests__/Touchable-test.js
+++ b/RNTester/e2e/__tests__/Touchable-test.js
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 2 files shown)


CodingItWrong:
@rickhanlonii I like it; that provides more coverage and the tests should still be just about as fast and easy to read. Will do.

PR #22576, opened 2018-12-09
Recycle cells during render phase on VirtualizedList component

yun77op:
This prevents view creation which is expensive, thus improves performance.

Test Plan:
----------
1. Begin with index 0
2. Increment index by 2 every 3s, then call method scrollToIndex(index) to scroll list, and measure the time taken to update the list component
3. When index reaches 60, end the interval
4. Calculate the mean time

before  507ms
after 484ms

Changelog:
----------

[General] [Added] - Recycle cells during render phase on VirtualizedList component

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -663,6 +663,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     first: number,
     last: number,
     inversionStyle: ViewStyleProp,
+    recycleEnabled = false,
   ) {
     const {
       CellRendererComponent,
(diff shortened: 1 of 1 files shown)


sahrens:
Hmm, I'm not sure a 5% performance improvement is worth all this complexity (at the very least, it would be nice if we could factor out the recycling code into a helper module if possible). I also think there might be potential for unexpected bugs from recycling because if cells have any stateful ancestors, that state could get swapped around to other items - is that something you considered?

cpojer:
Hello @yun77op. Do you mind providing the answers to the questions by @sahrens? I'm afraid without them I won't be able to merge this PR :(

cpojer:
Closing because of inactivity.

PR #22596, opened 2018-12-11
Remove persistence from RNTester app

CodingItWrong:
Previously the RNTester app saved what screen you were on and what filter text was entered into the initial screen. This made e2e testing complex, as each test needed to manually restore the state to the home screen. If the state ever got out of sync with the test's expectations, it could lead to multiple failed tests.

There is still one specific component that uses persistence: `RNTesterSettingSwitchRow`. Persistence can be removed from this component next time tests for it are updated. As a result, `RNTesterStatePersister` is not yet entirely removed from the app.

Test Plan:
----------
CI running yarn test-ios-e2e will confirm the tests still pass.

You can also run RNTester manually to confirm that after quitting and restarting the app, you're returned to the home screen of the app, and the filter state is not saved.

Changelog:
----------
[General] [Removed] - Remove persistence from RNTester app

### DIFF
--- a/RNTester/e2e/__tests__/Button-test.js
+++ b/RNTester/e2e/__tests__/Button-test.js
@@ -8,10 +8,11 @@
  * @format
  */
 
-/* global element, by, expect */
+/* global device, element, by, expect */
 
 describe('Button', () => {
   beforeAll(async () => {
+    await device.reloadReactNative();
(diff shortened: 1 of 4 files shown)


CodingItWrong:
@TheSavior bump; passing now

cpojer:
Looks good to me! Thanks for improving the RN test infra <3

Issue #13588, opened 2017-04-19
RCTFont SIGABRT crash


IlyaRepo:
### Description

Experiencing following crash:
Crash log snippet
Exception Type: SIGABRT
Exception Codes: #0 at 0x186ac7014
Crashed Thread: 4
Thread 4 Crashed:
0 libsystem_kernel.dylib 0x0000000186ac7014 __pthread_kill + 8
1 libsystem_c.dylib 0x0000000186a3b400 abort + 140
2 libc++abi.dylib 0x00000001865052d4 __cxa_bad_cast + 0
3 libc++abi.dylib 0x0000000186522ca8 default_terminate_handler() + 280
4 libobjc.A.dylib 0x0000000186530860 _objc_terminate() + 152
5 libc++abi.dylib 0x000000018651f66c std::__terminate(void ()) + 16
6 libc++abi.dylib 0x000000018651ef84 __cxxabiv1::exception_cleanup_func(_Unwind_Reason_Code, _Unwind_Exception*) + 0
7 libc++.1.dylib 0x00000001864f578c std::_1::_throw_system_error(int, char const*) + 88
8 MyApp 0x00000001001adb24 +[RCTFont updateFont:withFamily:size:weight:style:variant:scaleMultiplier:] (RCTFont.mm:92)
9 MyApp 0x0000000100223d94 -[RCTShadowText _attributedStringWithFontFamily:fontSize:fontWeight:fontStyle:letterSpacing:useBackgroundColor:foregroundColor:backgroundColor:opacity:] (RCTShadowText.m:278)
10 MyApp 0x0000000100223b08 -[RCTShadowText attributedString] (RCTShadowText.m:235)
11 MyApp 0x0000000100223a14 -[RCTShadowText recomputeText] (RCTShadowText.m:228)
12 MyApp 0x000000010021a670 -[RCTTextManager uiBlockToAmendWithShadowViewRegistry:] (RCTTextManager.m:109)
13 MyApp 0x00000001001f9c9c -[RCTComponentData uiBlockToAmendWithShadowViewRegistry:] (RCTComponentData.m:459)
14 MyApp 0x00000001001f1c18 -[RCTUIManager _layoutAndMount] (RCTUIManager.m:1120)
15 MyApp 0x00000001001c511c __36-[RCTBatchedBridge batchDidComplete]_block_invoke (RCTBatchedBridge.m:1008)
16 libdispatch.dylib 0x00000001869821fc _dispatch_call_block_and_release + 24
[...]

PR #22607, opened 2018-12-11
Leak mutex to avoid a crashing issue

chuganzy:
Try to solve https://github.com/facebook/react-native/issues/13588
ref: https://github.com/facebook/componentkit/pull/906

Test Plan:
----------
Regression / smoke tests should be enough.

Changelog:
----------
[iOS] [Fixed] - Fix a crashing issue on RCTFont / mutex

### DIFF
--- a/React/Views/RCTFont.mm
+++ b/React/Views/RCTFont.mm
@@ -126,12 +126,12 @@ static NSString *FontWeightDescriptionFromUIFontWeight(UIFontWeight fontWeight)
 static UIFont *cachedSystemFont(CGFloat size, RCTFontWeight weight)
 {
   static NSCache *fontCache;
-  static std::mutex fontCacheMutex;
+  static std::mutex *fontCacheMutex = new std::mutex;
 
   NSString *cacheKey = [NSString stringWithFormat:@"%.1f/%.2f", size, weight];
   UIFont *font;
   {
-    std::lock_guard<std::mutex> lock(fontCacheMutex);
+    std::lock_guard<std::mutex> lock(*fontCacheMutex);
     if (!fontCache) {
       fontCache = [NSCache new];
     }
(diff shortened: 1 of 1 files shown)


DimitryDushkin:
Ping. Eliminating crash is really important.

DimitryDushkin:
We've released out app with this patch and this crash gone away.

DimitryDushkin:
Where is my manners. Thank you @chuganzy!

shergin:
@chuganzy 
Wait, wait, what makes you think that the linked crash happens when application is closing? Should we leak all static mutexes?

shergin:
Here's a quote from Oliver:

> Historically, TextKit has had lots of hidden thread-safety issues despite being stated as thread safe. CK locks around all access to TextKit, but there are still issues that have cropped up in the recent past. We built the same text locking system in both CK and ASDK, and they both work this way.

As I can see, that's *not* how RN works here (and everywhere else) with fonts and co. So, maybe *this* is the problem.

chuganzy:
Here's a quote from @ocrickard

> OK, looks like std::mutex is not trivially destructible in Clang, so we should probably just leak that mutex. Anyway, I believe this crash is happening on exit() as @gkassabli said above, so it probably has no impact on user experience.

shergin:
Yes, exactly. My point is that the fix in CK fixes different type of the crash, not the one that promised in this diff's description.
I believe this change will not fix https://github.com/facebook/react-native/issues/13588 .

shergin:
Just chatted with Oliver and... I was a bit wrong. :)
So, this fix seems to fix the one of the crashes mentioned in the assigned issue (but probably not that one which in the description).

As Oliver pointed out, all TextKit-related crashes wich has something like  '0x185dfc238 std::__1::mutex::try_lock() + 34' will be fixed with this fix.

chuganzy:
@shergin Thank you, but seems like the issue in its description also indicates that it's crashing when it tries to lock from its line so I am hoping that this solves the issue🤞

PR #22579, opened 2018-12-10
Map textContentType strings to Objective-C constants

levibuzolic:
Fixes #22578

Currently the only `textContentType` values that work are: `username`, `password`, `location`, `name` and `nickname`. This is due to the strings provided by React Native not matching up with the underlying string constants used in iOS (with the exception of the aforementioned types). Issue #22578 has more detail examples/explanation.

Test Plan:
----------
Added a **Text Content Type** example to the RNTester app which shows two examples which weren't previously working.

> **Note:** In my testing, most (if not all) `textContentType` values don't work on the simulator, instead they need to be tested on a real device.

Before | After
---|---
img_1584 | img_1583

Changelog:
----------

[iOS] [Fixed] - Fixes textContentTypes not working by mapping textContentType strings to Objective-C constants - Fixes #22578

PR #22611, opened 2018-12-12
Map TextInput textContentType strings to Objective-C constants

### LINKED ISSUE #22578: Most TextInput textContentTypes don't work due to mismatched string constants
## Environment
```
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
      Memory: 36.17 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 8.11.3 - /var/folders/5f/6g410wwx743d7_wx8lphvzfc0000gn/T/yarn--1544413987073-0.5166565719343945/node
      Yarn: 1.12.1 - /var/folders/5f/6g410wwx743d7_wx8lphvzfc0000gn/T/yarn--1544413987073-0.5166565719343945/yarn
      npm: 5.6.0 - ~/.n/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 23.0.3, 25.0.0, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.3
        API Levels: 23, 25, 26
    IDEs:
      Android Studio: 3.1 AI-173.4819257
    npmPackages:
      @storybook/react-native: ^4.0.2 => 4.0.6
      react: 16.6.1 => 16.6.1
      react-native: 0.57.7 => 0.57.7
```

## Description

The strings used for `textContentType` are passed directly though to iOS, however they don't match up with the underlying string constants used in iOS.

I ran the following code in a playground to source the underlying values of the TextContentType constants:

```swift
import UIKit
print(UITextContentType.URL.rawValue)
print(UITextContentType.addressCity.rawValue)
// etc...
```

Here are the values that I got.

React Native | iOS
---|---
`URL` | `url`
`addressCity` | `address-level2`
`addressCityAndState` | `address-level1+2`
`addressState` | `address-level1`
`countryName` | `country-name`
`creditCardNumber` | `cc-number`
`emailAddress` | `email`
`familyName` | `family-name`
`fullStreetAddress` | `street-address`
`givenName` | `given-name`
`jobTitle` | `organization-title`
`location` | `location`
`middleName` | `additional-name`
`name` | `name`
`namePrefix` | `honorifix-prefix`
`nameSuffix` | `honorifix-suffix`
`nickname` | `nickname`
`organizationName` | `organization`
`postalCode` | `postal-code`
`streetAddressLine1` | `address-line1`
`streetAddressLine2` | `address-line2`
`sublocality` | `address-level3`
`telephoneNumber` | `tel`
`username` | `username`
`password` | `password`
`newPassword` | `new-password`
`oneTimeCode` | `one-time-code`

According to this table the only `textContentType`s that will work currently in React Native are: `username`, `password`, `location`, `name` and `nickname`.

## Possible Solutions

The simplest solution would be to update `textContentType` to use the iOS strings instead of current keys.

Alternatively a mapping in the native code could allow us to keep the existing strings and just silently translate them.

My preferred solution is to map each string to the named Objective-C constants which I've created a PR for #22579.

## Reproducible Demo

I've created a demo snack which allows you to try each `textContentType` with both the React Native and iOS strings: https://snack.expo.io/@levibuzolic/textcontenttype

> **Note:** In my testing, most (if not all) `textContentType` values don't work on the simulator, instead they need to be tested on a real device. 

## Example Screenshots

`addressState` | `address-level1`
---|---
img_1579 | img_1578

`newPassword` | `new-password`
---|---
img_1582 | img_1581

levibuzolic:
This is an updated version of #22579 which uses compile conditionals to prevent `use of undeclared identifier` errors when compiling on older versions of Xcode.

--------

## Summary
Currently the only `textContentType` values that work are: `username`, `password`, `location`, `name` and `nickname`. This is due to the strings provided by React Native not matching up with the underlying string constants used in iOS (with the exception of the aforementioned types). Issue #22578 has more detail examples/explanation.

## Test Plan
Added a **Text Content Type** example to the RNTester app which shows two examples which weren't previously working.

> **Note:** In my testing, most (if not all) `textContentType` values don't work on the simulator, instead they need to be tested on a real device.

Before | After
---|---
 | 

## Changelog
[iOS] [Fixed] - Fixes textContentTypes not working by mapping textContentType strings to Objective-C constants - Fixes #22578

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -141,9 +141,9 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
   [attributedTextCopy removeAttribute:RCTTextAttributesTagAttributeName
                                 range:NSMakeRange(0, attributedTextCopy.length)];
-  
+
   textNeedsUpdate = ([self textOf:attributedTextCopy equals:backedTextInputViewTextCopy] == NO);
-  
+
   if (eventLag == 0 && textNeedsUpdate) {
     UITextRange *selection = self.backedTextInputView.selectedTextRange;
     NSInteger oldTextLength = self.backedTextInputView.attributedText.string.length;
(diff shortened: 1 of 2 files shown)


levibuzolic:
Thanks for the build output logs @TheSavior -- I've updated the code and tested on an older versions of Xcode so this should work now.

@shergin would I be able to get a quick re-review?

Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -193,9 +193,65 @@ - (void)setTextContentType:(NSString *)type
...
+
+              contentTypeMap = [iOS12baseMap copy];
+            }
+          #endif
+        });
+
         // Setting textContentType to an empty string will disable any
         // default behaviour, like the autofill bar for password inputs
-        self.backedTextInputView.textContentType = [type isEqualToString:@"none"] ? @"" : type;
+        self.backedTextInputView.textContentType = type ? contentTypeMap[type] : type;
```
chuganzy: Probably `type ? contentTypeMap[type] : nil;` might be clearer?

BTW, some projects might be already using iOS internal definitions (like `given-name`, `family-name` etc) directly.
https://github.com/facebook/react-native/pull/22611/files#diff-a5239f085f0beab82ba2c1643be157acL198

So it might be better to assign directly if it was not defined in the map.

`self.backedTextInputView.textContentType = contentTypeMap[type] ?: type`

mcgaryes:
@levibuzolic - Any word on what version this PR might make it into?

levibuzolic:
@mcgaryes unfortunately not. It’s with the RN team now. It’s ready to merge but has been sitting here for a while without any attention. 🤷‍♂️

levibuzolic:
@shergin anything I can do to get this across the line for a future release? Should be good to go as far as I can tell.

levibuzolic:
@kelset sorry to ping you here, but I'm really keen to try and get this in to the next release, any chance I could get a review/merge from the team?

kelset:
Hey thanks for the ping - I'll see if we can merge it but no promises 😛

PR #22621, opened 2018-12-12
Fix E2E warnings

CodingItWrong:
Fixes two types of warnings that occur when running E2E tests:

1. A deprecation warning from Jest: 'Option "setupTestFrameworkScriptFile" was replaced by configuration "setupFilesAfterEnv", which supports multiple paths.'
2. YellowBox warnings when running the app in debug mode, about components that are deprecated and that require main queue setup.

By fixing these warnings, we increase contributors' confidence that things are working correctly, and draw attention to any warnings that they _should_ pay attention to, if and when they arise.

I feel confident that we should hide the deprecated-component warnings; we _want_ to use these components because we want them to be tested, until they're removed entirely.

For the warning "Module RCTImagePickerManager requires main queue setup", if that's something that can be fixed with reasonable effort in the RNTester code then I think it would be better to do so. Otherwise, I think it is good to hide the warning, because this is a condition we expect: it's not something a contributor should pay attention to.

Test Plan:
----------

Run:

```
detox build -c ios.sim.debug
detox test -c ios.sim.debug
```

You should not see any warnings in the console, or any YellowBox warnings in the simulator.

Changelog:
----------
[General] [Fixed] - Fix E2E warnings

### DIFF
--- a/RNTester/e2e/config.json
+++ b/RNTester/e2e/config.json
@@ -1,5 +1,5 @@
 {
-  "setupTestFrameworkScriptFile" : "./test-init.js",
+  "setupFilesAfterEnv" : ["./test-init.js"],
   "testEnvironment": "node",
   "bail": true,
   "verbose": true
--- a/RNTester/js/RNTesterApp.ios.js
+++ b/RNTester/js/RNTesterApp.ios.js
@@ -30,6 +30,7 @@ const {
   Text,
   View,
   SafeAreaView,
+  YellowBox,
 } = ReactNative;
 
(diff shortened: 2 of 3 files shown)


CodingItWrong:
Objc: `ld: symbol(s) not found for architecture x86_64`—seems like flake, quite unlikely to be related to this PR

CodingItWrong:
I reconfirmed the Detox tests work for me locally. Fixing whatever broke the Objc tests might fix Detox as well; I think we should not worry about the Detox ones yet.

PR #22659, opened 2018-12-15
iOS: Clear `Linking.getInitialURL` during bridge reload

venik:
On iOS platform, RN retains launchOptions dictionary after bridge reload which can lead to unexpected consequences to a developer. The app will receive the same value for `Linking.getInitialURL` during initial launch and during bridge reload. Here's an example from our application. We use deeplinks via custom URL scheme so a user can open the app via link. Also, we reload the bridge when a user signs out. So if a user opens the app via URL, logs out, and a second user logs into the app, the app will behave as though the second user launched the app via the same deeplink. Because reload destroys the JS engine, there's nowhere for our app to remember that it already handled the deeplink activation.

On iOS Linking.getInitialURL() gets URL from the _launchOptions dictionary, so by setting it to nil we prevent retention of initialURL after reload.

This change makes iOS's behavior consistent with Android's. On Android, the launch URL is stored on the `Intent` and reloading the app involves creating a new `Intent`. Consequently, the launch URL is dropped as desired during the reload process.
 
Test plan:
----------
I wrote a toy application to show the problem
https://github.com/venik/launchUrlExample/blob/fa8542b57b4331ee6be226db24048efe0d957b3e/App.js

Repro Steps
1. Ensure the app isn't running (swipe it out if it is running)
2. Open Safari, type in the URL mypoc://some_text, and hit enter.
3. OS asks if you want to open the app. Choose to open the app.
4. Tap on the button. Alert dialog shows "url: mypoc://some_text"
5. Now reload the app (cmd+R). This simulates an app in production reloading the bridge.
6. Tap the button again

Unexpected result (i.e. before this fix): Alert dialog shows "url: mypoc://some_text"

Expected result (i.e. after this fix): Alert dialog shows "url: null"

Here's the code from my toy application which repros the bug:
```
  render() {
    return (
        <Button
          onPress={
            () => {
              Linking.getInitialURL().then((url) => {
                AlertIOS.alert('url: ' + url);
              }).catch(err => console.error('An error occurred', err));
            }
          }
          style={styles.welcome}
          title="Learn More"
          accessibilityLabel="Learn more about this purple button"
        />
      </View>
    );
  }
}
```

Changelog:
----------
[iOS] [Fixed] [RCTBridge] - Set _launchOptions to nil when app reloads and treat reload as a fresh start. Do not respect previous launchOptions, so Linking.getInitialURL() will return null after reload.


cpojer:
This makes sense to me. Thank you for submitting a pull request with a fix and providing an extensive test plan :)

venik:
I will fix and resubmit soon. Thanks

venik:
@cpojer I fixed the PR and tested locally against POC application. It doesn't show the warning now.

Thanks in advance.

PR #22663, opened 2018-12-15
Remove injectJavascript from WebView propTypes definition

levibuzolic:
Related to https://github.com/facebook/react-native-website/pull/646

`injectJavascript` isn't a consumed prop of `WebView` -- it's a class method. This PR removes the prop type for it to prevent confusion in the future.

Test Plan:
----------
No test required, only removes an unused PropType.

Changelog:
----------
[General][Fixed] - Removed unused `injectJavascript` from `WebView` prop types

### DIFF
--- a/Libraries/Components/WebView/WebView.android.js
+++ b/Libraries/Components/WebView/WebView.android.js
@@ -191,12 +191,6 @@ class WebView extends React.Component {
     originWhitelist: PropTypes.arrayOf(PropTypes.string),
 
     /**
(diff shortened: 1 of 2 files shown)


hramos:
Can you open the PR against https://github.com/react-native-community/react-native-webview instead? The WebView code that remains in the core RN repository will be removed at some point, and we're asking people to make use of the https://github.com/react-native-community/react-native-webview fork instead.

PR #22664, opened 2018-12-15
remove appcompat-v7, support-v4 from Android Support Library BUCK

dulmandakh:
For FB import testing purpose only

Test Plan:
----------

Changelog:
----------

[CATEGORY] [TYPE] - Message


dulmandakh:
@hramos please import this, and let's see if it breaks something 😉

hramos:
Tried to import it, but we're having trouble with our OSS tooling this morning. Will try again later.

hramos:
This one didn't run into any issues, either. @dulmandakh is this for testing only?

dulmandakh:
@hramos yes, so closing it.

PR #22666, opened 2018-12-15
Fix modal and alert dismissal bugs

### LINKED ISSUE #22237: Alert disappears when modal gets hidden (re-rendered with visible=false)
## Environment

```
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (4) x64 Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
      Memory: 318.71 MB / 16.00 GB
      Shell: 5.6.2 - /usr/local/bin/zsh
    Binaries:
      Node: 11.0.0 - /usr/local/bin/node
      Yarn: 1.12.1 - /usr/local/bin/yarn
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react-native: 0.57.4 => 0.57.4
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description

On iOS when an alert is shown on top of a modal and you hide the modal (by changing its visible prop via state) the alert also gets dismissed at the same time.

## Reproducible Demo

Demo project:

[demo.zip](https://github.com/facebook/react-native/files/2569490/demo.zip)

Video of the bug:

https://drive.google.com/file/d/1eV7p9F_GLZDMWeoVZJrkcGqOXhL8ViYR/view?usp=sharing

sryze:
Fixes #22237

Summary:

1. Fixed a bug with dismissing modals that caused the modal to stay on screen after being dismissed if some other view controller is presented on top of it.

2. Fixed an issue where dismissing a modal (by changing its `visible` property) would also hide alerts presented on top of it (i.e. shown after the modal became visible). This is achieved by presenting each alert in a separate `UIWindow` instance.

3. Fixed a bug where showing two modals rapidly and hiding them would leave an empty (invisible) `RCTModalHostViewController` hanging on screen, which made it impossible to interact with the app as the view controller is display on top of everything.

These changes affect iOS only (no JS/Android changes).

Test Plan:
----------

Need some help with filling this section...

Changelog:
----------

  [iOS] [Fixed] - Fixed various issues with dismissing alerts and modals

### DIFF
--- a/React/Modules/RCTAlertManager.m
+++ b/React/Modules/RCTAlertManager.m
@@ -30,6 +30,7 @@ RCT_ENUM_CONVERTER(RCTAlertViewStyle, (@{
 @implementation RCTAlertManager
 {
   NSHashTable *_alertControllers;
+  UIWindow *_window;
 }
 
 RCT_EXPORT_MODULE()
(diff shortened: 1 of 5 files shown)


sryze:
@shergin, can you please review this pull request?

BTW I don't think the failure of "test_objc" is related to my changes in this PR.

sryze:
After more testing I discovered that [3] is not fully fixed as I ran into the same bug again today. I'll try to make a repro later. While debugging this I found that after the call to`dismissViewController:animated:` the modal view controller is not deallocated because the main (root) view controller  still holds a strong reference to it... I'm not sure if it's a React Native bug or iOS bug.

React/Modules/RCTAlertManager.m
```diff
@@ -30,6 +30,7 @@ @interface RCTAlertManager()
 @implementation RCTAlertManager
 {
   NSHashTable *_alertControllers;
+  UIWindow *_window;
```
shergin: Retaining anything in ViewManager is very bad idea in general (and especially for UIKit things).

cpojer:
As @shergin pointed out this PR should be split into three separate PRs. Would you mind opening those? I'm going to close this one because it's too big to be merged as one. Ideally every PR only has one main objective (one bugfix, one new feature, etc.).

PR #22667, opened 2018-12-15
enable proguard in release mode

gengjiawen:
## Motivation
This is default behavior  in android project. With progurad the code runs fast, the apk is smaller.
## Test Plan
pass all current ci.
## Release Notes
[Android] [Feature] - enable proguard in release mode

### DIFF
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -91,7 +91,7 @@ def enableSeparateBuildPerCPUArchitecture = false
 /**
  * Run Proguard to shrink the Java bytecode in release builds.
  */
-def enableProguardInReleaseBuilds = false
+def enableProguardInReleaseBuilds = true
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion


radeno:
@gengjiawen i like the idea about enabling it but is it safe for most of community modules ?

radeno:
I see. So every module needs to set it by own in `build.gradle`.

Maybe there could be wider propagation that is new default option and most as possible community modules should support it.

gengjiawen:
Yes, I agree we need add it to release note. But not every module need to config proguard rules.

dulmandakh:
I think that it might not be necessary and leave it to developer. Also I assume that proguard will be deprecated in favor of R8.

gengjiawen:
R8 still need time come close to production. And even with R8 the dsl may not be change.

hramos:
Can we instead recommend people enable this themselves as part of the docs?

gengjiawen:
I still think this should be default since minify in production is serious issue.

gengjiawen:
I think we should keep up with android default behavior. Let's keep it open for a while.

cpojer:
I agree with @hramos and @dulmandakh that we should add this to the documentation for now and ask users to do this manually. We should ensure reliability and correctness over size and speed, at least for now.

Issue #21966, opened 2018-10-26
[0.57.4-0.57.8/0.58.6] using textTransform causes text to disappear on Android, but works on iOS


kristerkari:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment

```
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
      Memory: 1.51 GB / 16.00 GB
      Shell: 2.7.1 - /usr/local/bin/fish
    Binaries:
      Node: 8.11.4 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 6.4.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
      Android SDK:
        Build Tools: 23.0.1, 23.0.3, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.0
        API Levels: 17, 18, 19, 21, 22, 23, 25, 26, 27
    IDEs:
      Android Studio: 3.1 AI-173.4670197
      Xcode: 10.0/10A255 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.0-alpha.8af6728 => 16.6.0-alpha.8af6728 
      react-native: 0.57.4 => 0.57.4 
```

## Description

React Native 0.57.4 brings support for using `textTransform` style property on Android. When setting `textTransform` value to `uppercase`, `lowercase` or `capitalize` the text is not shown at all on Android, but is shown correctly on iOS.

If the `textTransform` property is removed, then the text renders normally.

This can be easily reproduced by creating a new React Native project and setting the `textTransform` property to one of the texts inside `App.js`.

**Screenshots:**

iOS simulator (iPhone 8 / iOS 12):

<img width="471" alt="ios" src="https://user-images.githubusercontent.com/993108/47582471-1908e780-d95d-11e8-8405-a7a39858deea.png">

Android emulator (Nexus 5X / Android 5.0):

[...]

PR #22670, opened 2018-12-16
Fix textTransform when used with other text styles on Android

janicduplessis:
On Android `textTransform` breaks other styles applied to the text. It seems related to the usage of `ReplacementSpan` which allows drawing the text manually but seems to throw away some changes made by other span applied to the text.

To fix it I removed the usage of `ReplacementSpan` and simply transform the text before appending it to the `Spannable` string. To make sure textTransform is inherited correctly I added it to TextAttributes which handles this.

Test Plan:
----------
I added a test case to RNTester. It should render turquoise text on a blue background. It also adds line height and letter spacing to make sure the text is properly measured and all styles are applied.

Before the fix:
screen shot 2018-12-15 at 8 16 30 pm

After the fix:
screen shot 2018-12-15 at 8 15 05 pm

Also made sure all existing text transform examples produce the expected result.

Changelog:
----------
[Android] [Fixed] - Fix textTransform when used with other text styles on Android


kristerkari:
This will most likely fix https://github.com/facebook/react-native/issues/21966, right?

janicduplessis:
@kristerkari I think so, I can test with an example from the issue

janicduplessis:
@rigdern Quite possible, I just tested on snack but `textTransform` on android doesn't work on latest expo yet.

This should print `aB`
```js
<Text style={{textTransform: 'uppercase'}}>
  <Text style={{textTransform: 'none'}}>a</Text>
  <Text style={{textTransform: null}}>b</Text>
</Text>
```

I can test it when I get on a computer.

janicduplessis:
@rigdern you were right https://github.com/facebook/react-native/pull/22694

hramos:
@rigdern's changes landed in f6f8b092f9dac03e81cb01a73e748f3df85ea519. @janicduplessis do you still want your changes to be incorporated? If so, let me know if you can resolve the conflicts.

Thanks to both of you for working on getting these issues addressed!

janicduplessis:
@hramos Looks like https://github.com/facebook/react-native/pull/22917 will also conflict with this. I think we should try to land #22917 first and then I can rebase this on top of the improvements to text inheritance that @rigdern implemented.

rigdern:
@janicduplessis #22917 has been merged.

janicduplessis:
@hramos @rigdern Rebased on top of #22917 and re-tested to make sure everything still works. I also included the little fix in #22694 since I ended up changing these lines.

rigdern:
@janicduplessis The PR description says:

> To do that I can pass the parent's textTransform when recursing through text shadow nodes.

This sentence seems to need to be updated now that this information is passed around via `TextAttributes.java`

PR #22673, opened 2018-12-16
Make TextInput event prop types less strict

janicduplessis:
Similar to what was done here https://github.com/facebook/react-native/pull/22376

This allows using things like async functions with text input event props.

Test Plan:
----------
Run flow

Changelog:
----------

[General] [Fixed] - Make TextInput event prop types less strict


janicduplessis:
cc @TheSavior

TheSavior:
I can merge this from my work computer on Monday. Remind me if this doesn't get merged then.

PR #22688, opened 2018-12-18
Change location of iOS build cache directory to ~/Library/Caches/

### LINKED ISSUE #21780: Temporary files should be stored in ~/LIbrary/Caches or elsewhere, not ~/.rncache
Why does React Native put downloaded source code archives (e.g. Boost, glog) in `~/.rncache`? I don't want to track down the temporary garbage created by `react-native` next time I need to make more free disk space. 

Please put these files in `~/Library/Caches` or `/tmp` (or some other appropriate directory) on macOS.

sryze:
Instead of using ~/.rncache use the special Caches directory designed for caching files. This fixes #21780.

Changelog:
----------

[iOS] [Changed] - Moved iOS build cache directory to ~/Library/Caches/

Test Plan:
----------

1. Execute ./scripts/ios-install-third-party.sh
2. Check that the directory ~/Library/Caches/com.facebook.ReactNativeBuild is created and that the .tar.gz files of dependencies are present

or better:

1. Create a new project: `react-native init ReactNativeCacheTest`
2. 
   ```
   cd ReactNativeCacheTest
   cp /path/to/react-native-code/scripts/ios-install-third-party.sh node_modules/react-native/scripts/
   ```
3. `react-native run-ios`
4. Check that the directory ~/Library/Caches/com.facebook.ReactNativeBuild is created and that the .tar.gz files of dependencies are present

### DIFF
--- a/scripts/ios-install-third-party.sh
+++ b/scripts/ios-install-third-party.sh
@@ -4,14 +4,14 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-cachedir="$HOME/.rncache"
+cachedir="$HOME/Library/Caches/com.facebook.ReactNativeBuild"
 mkdir -p "$cachedir"
 
 function file_fail () {
     cachefile=$1
     msg=$2
 
-    echo "$msg.  Debug info:" 2>&1
+    echo "$msg. Debug info:" 2>&1
     ls -l "$cachefile" 2>&1
     shasum "$cachefile" 2>&1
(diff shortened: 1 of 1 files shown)


mcchrish:
@sryze I think it should check if `.rncache` is still existing and still use that instead. Only use the new location if not found. This prevents redownloading the cache for existing installs and new installs will use the new location.

sryze:
I agree, it's a good idea. I update the script to make it reuse the old directory if it exists.

Issue #22697, opened 2018-12-18
RNTesterIntegrationTests: Fix TheTester_ExpectError failure


hramos:
# Help Wanted: Fix `TheTester_ExpectError`

`test_objc` on Circle CI is red due to a failing integration test, TheTester_ExpectError:

`Assertion Failure: RCTTestRunner.m:231: failed: caught "NSInternalInconsistencyException", "Expected an error but the first one was missing or did not match."`

The test is defined in `RNTesterIntegrationTests.m` and the harness that should throw the expected error is at `IntegrationTests/IntegrationTestHarnessTest.js`.

Steps to reproduce:
---

1. Run `yarn` at the repository root.
2. Open `RNTester/RNTester.xcodeproj` on Xcode 10.1.0.
3. Build RNTester.
4. Run tests (CMD+U) on an iPhone 6 simulator.

Issue #22695, opened 2018-12-18
RNTesterIntegrationTests: Fix testTimersTest failure


hramos:
# Help Wanted: Fix `TimersTests`

`test_objc` on Circle CI is red due to a failing integration test, TimersTest:

```
UI Test Activity: 
Assertion Failure: RCTTestRunner.m:233: failed: caught "NSInternalInconsistencyException", "RedBox errors: (
    "undefined is not an object (evaluating 'this.setTimeout')",
    "Unhandled JS Exception: undefined is not an object (evaluating 'this.setTimeout')"
)"
```

The test is defined in JavaScript at `IntegrationTests/TimersTest.js`, and is invoked by RNTesterIntegrationTests.

Steps to reproduce:
---

1. Run `yarn` at the repository root.
2. Open `RNTester/RNTester.xcodeproj` on Xcode 10.1.0.
3. Build RNTester.
4. Run tests (CMD+U) on an iPhone 6 simulator.

PR #22720, opened 2018-12-20
[CI] test_ios → ✅: Update iOS Snapshots, remove tvOS tests, disable failing tests

hramos:
The following tests are disabled in this PR:

- testTimersTest is failing due to undefined this.setTimeout, probably introduced back in 61346d3. Tracking a fix in https://github.com/facebook/react-native/issues/22695
- testTheTester_ExpectError is failing as RCTTestRunner is not properly passing through the error. Tracking a fix in https://github.com/facebook/react-native/issues/22697

I've added a comment regarding testWebSocketTest and how to ensure it passes locally.

This PR also fixes all remaining snapshot tests, which were failing due to the use of iPhone XS as a iOS Simulator on Circle CI. We are using iPhone 6s for SST internally, and this allows us to be consistent.

Test Plan:
----------

Circle CI


hramos:
cc @ronal2do I *think* testTimersTest stopped working as of 61346d3 but it went undetected due to the issue I fixed in 5194495e9. I'll disable the test for now.

hramos:
`testScrollViewExample` is still failing on Circle CI, but it passes locally. Being a snapshot test, that's likely due to a small discrepancy in the simulator used locally vs on Circle CI.

PR #22731, opened 2018-12-20
[iOS] Add image bytes property of UIImage

zhongwuzw:
* Added a `reactDecodedImageBytes` property of `UIImage`, to give the chances if user wants to do custom memory calculation.
* Fixes wrong calculation of GIF decoded image bytes.

Changelog:
----------

[iOS] [Fixed] - Fixes image decoded bytes calculation

Test Plan:
----------

For single image, works the same as before. For GIFs, we get the decoded bytes using `total frames * single image bytes`.

### DIFF
--- a/Libraries/Image/RCTImageCache.m
+++ b/Libraries/Image/RCTImageCache.m
@@ -71,7 +71,7 @@ static NSString *RCTCacheKeyForImage(NSString *imageTag, CGSize size, CGFloat sc
   if (!image) {
     return;
   }
-  CGFloat bytes = image.size.width * image.size.height * image.scale * image.scale * 4;
+  NSInteger bytes = image.reactDecodedImageBytes;
   if (bytes <= RCTMaxCachableDecodedImageSizeInBytes) {
     [self->_decodedImageCache setObject:image
                                  forKey:cacheKey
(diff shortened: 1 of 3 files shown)


cpojer:
Can you expand on why "Added a reactDecodedImageBytes property of UIImage, to give the chances if user wants to do custom memory calculation." matters?

zhongwuzw:
Hi @cpojer , two reasons let me think expose one property of `UIImage` may better:
1. Currently, we implement `GIF` by `reactKeyframeAnimation`, it's a property of `UIImage`, so the original memory cost calculation is invalid in this situation, at least we need to find a way to fix.
2. We have `RCTImageLoader` manage loaders, which means, we support interface to register image loader for user, please see [RCTImageURLLoader](https://github.com/facebook/react-native/blob/5218932b13ad0649ff2a57aaf1ec682fe278c47d/Libraries/Image/RCTImageLoader.h#L164), in theory, user can return an `UIImage` which also has its custom property like `reactKeyframeAnimation` occupy much, a simple example is user may use [FLAnimatedImage](https://github.com/Flipboard/FLAnimatedImage) to handle `GIF`, `FLAnimatedImage` class occupy much memory, and it's not a subclass of a `UIImage`, so we need make `FLAnimatedImage` as a property of `UIImage`, the same as `reactKeyframeAnimation`. By exposing property `reactDecodedImageBytes`, user can provide its correct memory cost.

PR #22733, opened 2018-12-20
Adding JSON support to FormData polyfill

Mimetis:
Hi team !

In some circumstances, we need to send a **JSON** payload as a **part** of the `FormData`.
For instance, here is what we need to send, as a final payload:

```
--boundary_1234-abcd 
Content-Type: application/json 
Content-Disposition: form-data; name="mywine" 
{
    'key' : 1,
    'title' : 'red wine'
} 
--boundary_1234-abcd-- 
```

Today, the `FormData` class does not allow us to specify `Content-Type` if the payload *is not* a blob uri.

This PR fixes this issue, allowing the user to pass a JSON payload with a type property `Content-Type: application/json`.   
To do it, the `FormData` value provided by the user, must have a `value` property, containing the JSON object:
``` javascript

{
 value: {}, // JSON Payload
 type : 'application/json',
 name: 'arbitrary_name'
}

``` 
Basically, here is a straightforward sample:

``` JavaScript

    let payload = {
      key: 1,
      title : 'red wine'
    };
   
   formData.append('json', {
      value: payload,
      type: 'application/json',
      name :'mywine'
   });
``` 

Changelog:
----------
[General][added] - Add JSON support to FormData polyfill

Test Plan:
----------
The FormData is updated.    
See `Librairies/Network/__tests__/FormData-test.js`

### DIFF
--- a/Libraries/Network/FormData.js
+++ b/Libraries/Network/FormData.js
@@ -13,18 +13,18 @@
 type FormDataValue = any;
 type FormDataNameValuePair = [string, FormDataValue];
 
-type Headers = {[name: string]: string};
+type Headers = { [name: string]: string };
 type FormDataPart =
   | {
-      string: string,
-      headers: Headers,
-    }
+    string: string,
+    headers: Headers,
+  }
   | {
-      uri: string,
-      headers: Headers,
-      name?: string,
-      type?: string,
-    };
+    uri: string,
+    headers: Headers,
+    name?: string,
+    type?: string,
+  };
 
 /**
  * Polyfill for XMLHttpRequest2 FormData API, allowing multipart POST requests
@@ -38,9 +38,20 @@ type FormDataPart =
  *     name: 'photo.jpg',
  *   };
  *
+ *   var json = {
+ *     type: 'application/json',
+ *     value: {
+ *        a: 'a',
+ *        b: 'b',
+ *        k: 12
+ *     },
+ *     name:'ab'
+ *   };
+ * 
  *   var body = new FormData();
  *   body.append('authToken', 'secret');
  *   body.append('photo', photo);
+ *   body.append('obj', json);
  *   body.append('title', 'A beautiful photo!');
  *
  *   xhr.open('POST', serverURL);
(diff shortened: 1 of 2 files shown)


newyankeecodeshop:
This change will diverge the `FormData.append` API from the Web standard.
https://developer.mozilla.org/en-US/docs/Web/API/FormData/append
Another way to achieve this would be to have a way to create a JSON `Blob` and then pass that to append.
```
const value = new Blob([JSON.stringify(stuff)], { type: 'application/json' });
formData.append('name', value);
```

Mimetis:
Yes @newyankeecodeshop you're totally right on this point !
I'm not really proud of this implementation, at least It's working on my case...

I'm trying to evolve the code to use the `Blob` standard, as you said, using `new Blob([JSON.stringify(stuff)], { type: 'application/json' });`

The problem is that I need to read the blob, and the `getParts` method is not a Promise: 

Here is what I have so far:

``` javascript
    readBlob(blob) {
        return new Promise((rs, rj) => {
            var reader = new FileReader();
            reader.onload = () => rs(reader.result);
            reader.onerror = (error) => rj(error);
            reader.readAsText(blob);
        });
    }

    async getParts() {
        return await Promise.all(this._parts.map(async ([name, value]) => {
            var contentDisposition = 'form-data; name="' + name + '"';
            var headers = { 'content-disposition': contentDisposition };
            if (typeof value === 'string') {
                var itemstring = { string: value, headers, fieldName: name };
                return itemstring;
            }

            if (value instanceof Blob && typeof value.type === 'string' && value.type == 'application/json') {

                headers['content-type'] = value.type;
                var str = await this.readBlob(value);
                return { string: str, headers, fieldName: name };
            }

            // The body part is a "blob", which in React Native just means
            // an object with a `uri` attribute. Optionally, it can also
            // have a `name` and `type` attribute to specify filename and
            // content type (cf. web Blob interface.)
            if (typeof value.name === 'string') {
                headers['content-disposition'] += '; filename="' + value.name + '"';
            }
            if (typeof value.type === 'string') {
                headers['content-type'] = value.type;
            }

            return { ...value, headers, fieldName: name };
        }));
    }
```
I'm continuing to investigate. 
If I'm not able to resolve this *standardization* issue, I will close the PR.

Sebastien

PR #22738, opened 2018-12-20
Fixing ActionSheetIOS position after rotation on tablet

stigi:
There's been a bug on iOS and iPad that the position of an action sheet using UIActionController isn't updated if the position of its anchor view changes due to rotating the device. A common scenario would be, presenting an action sheet from a right bar button item. Rotating the device will most likely change the bar button's X coordinate. The action sheets arrow would still point to the old position due to how it has been implemented so far.

I used also reduced some code duplication between `-showActionSheetWithOptions` and `-showShareActionSheetWithOptions:` while at it.

Changelog:
----------

[iOS] [Fixed] - Action Sheet position after rotation on tablet

Test Plan:
----------

I tested on simulator and on device: on iPad and iPhone.

I added two no cases to RNTester to test the changes.

Check the iPad videos for the before:  
[ipad-broken.mp4.zip](https://github.com/facebook/react-native/files/2699202/ipad-broken.mp4.zip)

and after:  
[ipad-fixed.mp4.zip](https://github.com/facebook/react-native/files/2699203/ipad-fixed.mp4.zip)

When no anchor view is provided the popover is centered on the screen. Rotating the screen still causes a bug where the action sheet is mispositioned. This behavior is the same same as without this patch.

On iPhone, the action sheet is still presented in the regular way, pushing in from the bottom. This behavior is the same with and without providing an anchor view.

We're also testing the changes in the beta of one of our internal apps via [patch-package](https://www.npmjs.com/package/patch-package).


stigi:
Ran inlined styles, ran prettier and updated changeling entry.

PR #22759, opened 2018-12-23
Makes 0.58.0-rc.2 build on tvOS

clarsen:
Help us understand your motivation by explaining why you decided to make this change:

Changelog:
----------

[iOS] [FIXED] - Removed RCTWK* files from compilation on tvOS platform. 

These files include 'WebKit/WebKit.h' which isn't available when building for tvOS.

Test Plan:
----------

1. Built on tvOS with this version and verified my own app came up on simulator.
2. Ran RNTester-tvOS in `RNTester/RNTester.xcodeproj` in this version and verified it came up on Apple TV 4K simulator.
  Had to use xcode legacy build as per https://stackoverflow.com/a/51089264/124042


hramos:
Please open your PR against the master branch instead. We can then cherry pick your change onto 0.58.

Unfortunately I wasn't able to get a clean PR after editing your PR to use master as the base.

PR #22763, opened 2018-12-23
[Android] [TextInput]: Add prop to configure `importantForAutofill`.

jainkuniya:
In API 26, autofill framework was introduced in Android.
Read more about Autofill at https://developer.android.com/guide/topics/text/autofill.

Now, if in case for some text input if developer wants to disable
autofill then he can take help from this `importantForAutoFill` prop
and pass `no` to it.

Also important of auto fill can be configured with this prop, like:
* `auto`: Let the Android System use its heuristics to determine if the view is important for autofill.
* `no`: This view isn't important for autofill.
* `noExcludeDescendants`: This view and its children aren't important for autofill.
* `yes`: This view is important for autofill.
* `yesExcludeDescendants`: This view is important for autofill, but its children aren't important for autofill.

Default value if `auto`.

Read more at: https://developer.android.com/guide/topics/text/autofill-optimize

Changelog:
----------
[Android] [Added] - Add prop to configure `importantForAutofill` in `TextInput`.

Test Plan:
----------
Just add ` importantForAutofill="yesExcludeDescendants"` to `TextInput`.

like this:
```
 <TextInput
          style={{ height: 40, borderColor: 'gray', borderWidth: 1 }}
          onChangeText={text => this.setState({ text })}
          value={this.state.text}
          importantForAutofill="yesExcludeDescendants"
        />
```

Motivation:
----------
Android Oreo introduced the AutoFill Framework, for secure communication between an app and autofill services (e.g. Password managers). When using `<TextInput>` on Android Oreo+, the system already tries to autofill (based on heuristics), but there is no way to set configuring options or **disable**.

So I thought to add ability to **disable**/**configure** this auto fill importance.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -230,6 +230,13 @@ type AndroidProps = $ReadOnly<{|
   underlineColorAndroid?: ?ColorValue,
   inlineImageLeft?: ?string,
   inlineImagePadding?: ?number,
+  importantForAutofill?: ?(
+    | 'auto'
+    | 'no'
+    | 'noExcludeDescendants'
+    | 'yes'
+    | 'yesExcludeDescendants'
+  ),
 |}>;
 
 type Props = $ReadOnly<{|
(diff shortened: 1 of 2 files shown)


janicduplessis:
@cpojer -bot shipit :)

PR #22764, opened 2018-12-23
Fix universal links not working in iOS 12 / Xcode 10

### LINKED ISSUE #22716: [ios] Linking doesn't work with ios12 due to new continue userActivity implementation

## Environment
```
    System:
      OS: macOS 10.14.1
      CPU: x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
      Memory: 293.77 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.14.2 - /usr/local/bin/node
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      @types/react: 16.7.17 => 16.7.17 
      @types/react-native: 0.57.18 => 0.57.18 
      react: 16.7.0-alpha.2 => 16.7.0-alpha.2 
      react-native: 0.57.8 => 0.57.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description
[Per linking documentation](https://facebook.github.io/react-native/docs/linking) iOS implementation for handling universal links right now is

```objc
- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity
 restorationHandler:(void (^)(NSArray * _Nullable))restorationHandler
{

 // Debug log something here
 return [RCTLinkingManager application:application
                  continueUserActivity:userActivity
                    restorationHandler:restorationHandler];
}
```

With iOS 12 and XCode 10 this no longer seems to work (I think ios 12 forces new implementation?), there is new way to handle universal links that should look like this

```objc
- (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull NSUserActivity *)userActivity 
restorationHandler: (nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler {

  // Debug log something here
  return [RCTLinkingManager application:application
                  continueUserActivity:userActivity
                    restorationHandler:restorationHandler];
}
```

As you can see they added `UIUserActivityRestoring` as a new type for `restorationHandler`.  When running on iOS 12 and using old approach I don't see my `// Debug log something here` message at all. When I use new approach I see the message, but app crashes at `RCTLinkingManager` with following error

> Thread 1: EXC_BREAKPOINT (code=1, subcode=0x100a9cde0)

There is also an xcode issue saying

> Cannot convert value of type '([UIUserActivityRestoring]?) -> Void' to expected argument type '(([Any]?) -> Void)?'

Related apple docs entry: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623072-application?language=objc

IljaDaderko:
fix #22716

Changelog: 
----------
[iOS][Fixed] - Fix universal links in iOS 12 / Xcode 10

Test Plan:
----------

Tested this on a real device running iOS 12, in theory it should work on older versions as well as previous implementation assumed `*` type for `restorationHandler` params, however if someone could test this on iOS 11 and lower, it would be much appreciated.

### DIFF
--- a/Libraries/LinkingIOS/RCTLinkingManager.h
+++ b/Libraries/LinkingIOS/RCTLinkingManager.h
@@ -22,6 +22,6 @@
 
 + (BOOL)application:(nonnull UIApplication *)application
     continueUserActivity:(nonnull NSUserActivity *)userActivity
-      restorationHandler:(nonnull void (^)(NSArray *__nullable))restorationHandler;
+      restorationHandler:(nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler;
 
 @end
--- a/Libraries/LinkingIOS/RCTLinkingManager.h
+++ b/Libraries/LinkingIOS/RCTLinkingManager.h
@@ -22,6 +22,11 @@
 
 + (BOOL)application:(nonnull UIApplication *)application
     continueUserActivity:(nonnull NSUserActivity *)userActivity
-      restorationHandler:(nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler;
+      restorationHandler:
+        #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 12000)
+            (nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler;
+        #else
+            (nonnull void (^)(NSArray *_Nullable))restorationHandler;
+        #endif
 
 @end
--- a/Libraries/LinkingIOS/RCTLinkingManager.h
+++ b/Libraries/LinkingIOS/RCTLinkingManager.h
@@ -23,7 +23,7 @@
 + (BOOL)application:(nonnull UIApplication *)application
     continueUserActivity:(nonnull NSUserActivity *)userActivity
       restorationHandler:
-        #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 12000)
+        #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 12000) /* __IPHONE_12_0 */
             (nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler;
         #else
             (nonnull void (^)(NSArray *_Nullable))restorationHandler;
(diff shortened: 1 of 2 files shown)


radex:
@hramos Can we merge this? I was about to submit a nearly identical patch, because the app will Crash if you try to call this from Swift (as the type annotations don't match and a dumb cast will fail)

hramos:
@radex any chance you can test this on iOS 11 or earlier? At a glance it seems like the incomplete test plan might have been the reason this was not merged back when it was submitted.

IljaDaderko:
@hramos @radex 

Following snipped is from [firebase dynamic links implementation](https://firebase.google.com/docs/dynamic-links/ios/receive)

```objc
- (BOOL)application:(UIApplication *)application
continueUserActivity:(nonnull NSUserActivity *)userActivity
 restorationHandler:
#if defined(__IPHONE_12_0) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_12_0)
(nonnull void (^)(NSArray<id<UIUserActivityRestoring>> *_Nullable))restorationHandler {
#else
    (nonnull void (^)(NSArray *_Nullable))restorationHandler {
#endif  // __IPHONE_12_0

// Other Code

}
```

They target ios 12+ specifically to apply this change, I can update my PR to this tomorrow, it should fall back to old implementation on earlier versions of iOS. Sounds ok?

radex:
@IljaDaderko Right! I'm not sure if checking for __IPHONE_12_0 is the right way to go — I'd check for other RN sources to check what's the convention for checking for version (so things like tvOS are also taken into account)... there might already be a macro for that or something

IljaDaderko:
@radex I just looked through few react-native-community repos, but can't seem to find any macro for iOS version check / tvOS. Do you have any repos in mind that could use them for me to reference? Perhaps I should go ahead with similar solution firebase use for time being?

radex:
@IljaDaderko I did a quick search and this seems to be the RN convention:

https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/WebSocket/RCTReconnectingWebSocket.m#L19

IljaDaderko:
@hramos @radex sorry for the delay, but I have now added those conditional checks to only apply this change on ios 12+

radex:
If it compiles, looks good to me! @IljaDaderko Have you had the chance to check on iOS 11 (Xcode 9) if it still compiles correctly there, just to be sure?

IljaDaderko:
@radex Only tested this with latest XCode and iOS I'm afraid. Tests seem to be failing, but for things unrelated to this PR?

I can try to compile on earlier iOS versions devices latter this week if no one else can do this atm.

cpojer:
@IljaDaderko please do test on a device with an older version of iOS and report back so we have confidence before merging this PR :)

IljaDaderko:
@cpojer @radex Hey guys, I don't have access to ios 11 device, is anyone able to test this. Back when I replied I had simulator with ios11 in the back of my head, but obviously can't test linking in simulator :/

cpojer:
Hey @IljaDaderko I finally managed to land this PR. I had to add a check `__has_include(<UIKitCore/UIUserActivity.h>)` to make it pass some Facebook internal stuff. Would you mind double checking if universal links work with React Native from master?

PR #22769, opened 2018-12-23
Add allowScrollOutOfBounds to ScrollView

### LINKED ISSUE #22768: [iOS] scrollTo() can scroll past the child contents

## Environment
```
    System:
      OS: macOS 10.14.2
      CPU: x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
      Memory: 183.48 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.14.0 - /opt/local/bin/node
      Yarn: 1.12.3 - /opt/local/bin/yarn
      npm: 6.5.0 - /opt/local/bin/npm
      Watchman: 4.9.0 - /opt/local/bin/watchman
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.0, 28.0.1, 28.0.2, 28.0.3
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5014246
    npmPackages:
      react: 16.6.3 => 16.6.3
      react-native: 0.57.8 => 0.57.8
```
## Description
On iOS if one uses the scrollTo()/scrollToOffset() family of functions and the offset provided is bigger than the actual content, the content will disappears from the screen and will only be visible again if the user interacts with the UI again. It's important to note that this behavior does not occur on Android, therefore both platforms should behave equally.

## Reproducible Demo

1 - Go to https://snack.expo.io/@cabelitos/scroll-problem
2 - Run the project on an iOS simulator.
3 - After 2 seconds the scrollToOffset() function will be called with an offset bigger than the content.
4 - The screen will be white after the function executes.

One can execute the same app on an Android emulator and will notice that the app will stop at the last item and will not advance any further.

### iOS behavior
iOS scroll problem
### Android behavior
Android

cabelitos:
When using the scrollTo() function the user can specify a x/y position
that tells the ScrollView to scroll to. Prior to this patch the x/y were
not validated which could cause the SrollView to scroll to a position
outside the content view bounds, leaving the screen empty (the content
disapears, since it goes outside the screen). In order to fix this problem,
this commit imposes that the x/y must be inside the view bounds if allowScrollOutOfBounds
is set to false. This new prop is also set to true automatically to not break
existing apps which may depends on this feature.

Note that this behavior does not occur on Android, because the
Android's framework itself implements this features via [1] and [2], which are used
by React Native.

[1]: https://developer.android.com/reference/android/widget/ScrollView#smoothScrollTo(int,%20int)
[2]: https://developer.android.com/reference/android/widget/ScrollView.html#scrollTo(int,%20int)

Changelog:
----------

[iOS] [Added] - Add allowScrollOutOfBounds to ScrollView

Test Plan:
----------
Run the following code on a iOS simulator/device and play around with the new `allowScrollOutOfBounds` property.

* When set to `false` the `scrollToOffset` will stop at the last item (the item 12)
* When set to `true` the `scrollToOffset` will scroll outside the content bounds and leave the screen completely white.

```javascript
import { StyleSheet, Text, View, FlatList, SafeAreaView } from 'react-native';

const data = [
  { key: "1" },
  { key: "2" },
  { key: "3" },
  { key: "4" },
  { key: "5" },
  { key: "6" },
  { key: "7" },
  { key: "8" },
  { key: "9" },
  { key: "10" },
  { key: "11" },
  { key: "12" }
];

  componentDidMount() {
    this.timeoutID = setTimeout(() => {
      if (this.flatList) {
        console.log('sdsadsadasdads');
        this.flatList.scrollToOffset({
          offset: 10000,
          animated: true
        });
      }
      this.timeoutID = null;
    }, 2000);
  }

  componentWillUnmount() {
    clearTimeout(this.timeoutID);
  }

  renderItem({ item: { key }}) {
    return (
      <View style={styles.view}>
        <Text style={styles.text}>{key}</Text>
      </View>
    )
  }

  onRef = (flatList) => {
    this.flatList = flatList;
  };

  timeoutID = null;
  flatList = null;

  render() {
    return (
      <SafeAreaView>
        <FlatList
          allowScrollOutOfBounds={false} // NOTE THIS NEW FLAG
          ref={this.onRef}
          style={styles.container}
          data={data}
          renderItem={this.renderItem}
        />
      </SafeAreaView>
    );
  }
}

  container: {
    paddingTop: 20,
    backgroundColor: '#F5FCFF',
  },
  view: {
    marginBottom: 10,
    justifyContent: 'center',
    alignItems: 'center',
    width: '100%',
    backgroundColor: 'green',
    height: 100,
  },
  text: {
    color: 'red',
    fontSize: 20,
  },
});

```

# GIF demo

## Without the patch
iOS scroll problem

## With the patch
iOS scroll problem

fixes #22768

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -585,12 +585,40 @@ static inline void RCTApplyTransformationAccordingLayoutDirection(UIView *view, 
   [self scrollToOffset:offset animated:YES];
 }
 
+- (CGPoint)getMaxXOffset
+{
+    CGFloat offsetX = _scrollView.contentSize.width - _scrollView.bounds.size.width + _scrollView.contentInset.right;
+    return CGPointMake(fmax(offsetX, 0), 0);
+}
+
+- (CGPoint)getMaxYOffset
+{
+  CGFloat offsetY = _scrollView.contentSize.height - _scrollView.bounds.size.height + _scrollView.contentInset.bottom;
+  return CGPointMake(0, fmax(offsetY, 0));
+}
+
+- (CGPoint)getMaxOffset
+{
+  return [self isHorizontal:_scrollView] ? [self getMaxXOffset] : [self getMaxYOffset];
+}
+
 - (void)scrollToOffset:(CGPoint)offset animated:(BOOL)animated
 {
-  if (!CGPointEqualToPoint(_scrollView.contentOffset, offset)) {
+  CGPoint maxX = [self getMaxXOffset];
+  CGPoint maxY = [self getMaxYOffset];
+  CGPoint toOffset;
+
+  if (self.allowScrollOutOfBounds) {
+    toOffset = offset;
+  } else {
+    // Restrict inside the scroll view container bounds
+    toOffset = CGPointMake(fmax(0, fmin(offset.x, maxX.x)), fmax(0, fmin(offset.y, maxY.y)));
+  }
+
+  if (!CGPointEqualToPoint(_scrollView.contentOffset, toOffset)) {
     // Ensure at least one scroll event will fire
     _allowNextScrollNoMatterWhat = YES;
-    [_scrollView setContentOffset:offset animated:animated];
+    [_scrollView setContentOffset:toOffset animated:animated];
   }
 }
 
@@ -600,15 +628,7 @@ static inline void RCTApplyTransformationAccordingLayoutDirection(UIView *view, 
  */
 - (void)scrollToEnd:(BOOL)animated
 {
-  BOOL isHorizontal = [self isHorizontal:_scrollView];
-  CGPoint offset;
-  if (isHorizontal) {
-    CGFloat offsetX = _scrollView.contentSize.width - _scrollView.bounds.size.width + _scrollView.contentInset.right;
-    offset = CGPointMake(fmax(offsetX, 0), 0);
-  } else {
-    CGFloat offsetY = _scrollView.contentSize.height - _scrollView.bounds.size.height + _scrollView.contentInset.bottom;
-    offset = CGPointMake(0, fmax(offsetY, 0));
-  }
+  CGPoint offset = [self getMaxOffset];
   if (!CGPointEqualToPoint(_scrollView.contentOffset, offset)) {
     // Ensure at least one scroll event will fire
     _allowNextScrollNoMatterWhat = YES;
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -76,6 +76,12 @@ type IOSProps = $ReadOnly<{|
    */
   automaticallyAdjustContentInsets?: ?boolean,
   /**
+   * Set to false if the scroll should be clamped to the bounds of the
+   * ScrollView children when using the scrollTo() function. The default value is true.
+   * @platform ios
+   */
+  allowScrollOutOfBounds?: ?boolean,
+  /**
    * The amount by which the scroll view content is inset from the edges
    * of the scroll view. Defaults to `{top: 0, left: 0, bottom: 0, right: 0}`.
    * @platform ios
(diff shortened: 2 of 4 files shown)


TheSavior:
If I understand correctly, you are changing `scrollTo` so that if you try to scrollTo a location outside of the bounds of the scrollview it will be bounded to the size of the scrollview. And you are adding this prop so as to not change the current behavior?

I'm not very familiar with ScrollView behavior but I'm curious if the current behavior would be seen as a bug that we should just change without adding the additional prop. Maybe @sahrens or @wcandillon have some thoughts.

cabelitos:
Hello @TheSavior, yes you're correct. I'm only adding this prop to do not break the current behavior in iOS in case someone depends on this. But if you guys think it's not necessary I'll be glad to remove it.

cabelitos:
Hey guys, any news on this?

@TheSavior @sahrens

berickson1:
I was actually going to open an issue/PR to prevent this same programatic overscroll behaviour.

I don't think RN should expose a prop to keep the "existing" behaviour. This behaviour sounds entirely like a bug, especially since the first user interaction causes a rubber-band effect & the behaviour differs from Android (consistent behaviour should be used wherever possible between platforms)

I can't speak for Facebook on this matter, but I'd highly suggest taking this PR without the `allowScrollOutOfBounds` flag

cabelitos:
@berickson1 I'm happy to remove the flag. I just to know what Facebook has to say

cpojer:
@cabelitos can you rebase this PR and remove the `allowScrollOutOfBounds`?

cabelitos:
@cpojer as a matter of fact a fix for this was already merged [1]. This is soo sad, because I have this PR sitting here for months and that other PR was merged even though it was newer than mime.


[1]: https://github.com/facebook/react-native/pull/23427

cpojer:
Oh you are right! I'm so sorry about that. Up until recently we didn't have the bandwidth to take care of all contributions. We recently worked through 600 Pull Requests (See https://facebook.github.io/react-native/blog/2019/03/01/react-native-open-source-update ) and are now much more active on the React Native repo. As you can see, the number of open Pull Requests is less than 50, which is much more manageable than it was in the past :) Again, I'm so sorry this happened but I'm hoping to see your contributions again in the future, and I'll make sure to provide feedback quickly and get your changes merged.

cabelitos:
@cpojer np at all, these things happens specially when you have 500+ PRs. I have another PR open which fixes problems for ART on Android. I hope you guys are able to take a look into it!

zhongwuzw:
@cabelitos Ahha, I'm sorry, I didn't see this PR when I linked your bug report. And I see your fix now, we try to fix in different way, you add an option to enable/disable overflow, maybe you can try to create a new PR to let react native team for review.

React/Views/ScrollView/RCTScrollView.m
```diff
@@ -585,12 +585,40 @@ - (void)scrollToOffset:(CGPoint)offset
...
+- (CGPoint)getMaxXOffset
+{
+    CGFloat offsetX = _scrollView.contentSize.width - _scrollView.bounds.size.width + _scrollView.contentInset.right;
+    return CGPointMake(fmax(offsetX, 0), 0);
+}
+
+- (CGPoint)getMaxYOffset
+{
+  CGFloat offsetY = _scrollView.contentSize.height - _scrollView.bounds.size.height + _scrollView.contentInset.bottom;
+  return CGPointMake(0, fmax(offsetY, 0));
```
zhongwuzw: The minimum is not 0, we also need to consider `_scrollView.contentInset.top`.

cpojer:
@zhongwuzw we agreed that your solution is fine, and that we don't want to add a new flag for this.

zhongwuzw:
@cpojer Get it ! :+1:
@cabelitos Apologize again! :heart:

mysport12:
> @zhongwuzw we agreed that your solution is fine, and that we don't want to add a new flag for this.

So there is no chance of adding a new flag for this to re-enable the old behavior? We were using the behavior quite extensively as an alternative way to position TextInputs/Dropdowns/etc on the screen when selected/focused and oftentimes that caused some over scroll of the content.

sheepmiee:
> > @zhongwuzw we agreed that your solution is fine, and that we don't want to add a new flag for this.
> 
> @cpojer So there is no chance of adding a new flag for this to re-enable the old behavior? We were using the behavior quite extensively as an alternative way to position TextInputs/Dropdowns/etc on the screen when selected/focused and oftentimes that caused some over scroll of the content.

I agree with you, I used this old behavior  to position TextInputs in previous projects, but now,I have to edit the native code to use the old behavior

zhongwuzw:
@sheepmiee Hey, `flag` already be added in https://github.com/facebook/react-native/pull/24296, please see that.

PR #22774, opened 2018-12-24
Optimize js batch callback when find valid modules

zhongwuzw:
`batchDidComplete` would be called every time JavaScript communicates with native, it's be called in high frequency, we can manage an array to store module which needs batched callback (currently only `UIManager` needs this, but we have 81 default modules).

Changelog:
----------

[iOS] [Optimized] - Optimize js batch callback when find valid modules

Test Plan:
----------

For any bridged module, if it needs to handle batched callback, the module needs to be stored in `_batchDidCompleteCallbackModules` or `_partialBatchDidFlushCallbackModules`.


cpojer:
I talked to @fkgozali and he says we may not want to do this because of Turbomodules, so I'm gonna go ahead and close this PR.

PR #22777, opened 2018-12-24
Added filtering to RNTester example screens

CodingItWrong:
This PR adds filtering functionality to individual example screens of RNTester. This is useful for Detox testing of RNTester, since Detox requires elements to be visible on the screen before they can be interacted with. Instead of needing to scroll an arbitrary amount, the test can enter the name of the example to be tested, just as is done on the main screen. This will lead to simpler and more reliable E2E tests for long example screens. This PR doesn't add any automated tests using the filter; those will be added in a separate PR.

This is implemented by extracting the existing filtering functionality out of `RNTesterExampleList` into a shared `RNTesterExampleFilter` component that can be used both within `RNTesterExampleList` (the main screen) and `RNTesterExampleContainer` (the example screen).

simulator screen shot - iphone 8 - 2018-12-24 at 08 22 46

simulator screen shot - iphone 8 - 2018-12-24 at 08 22 51

Changelog:
----------

[General] [Added] - Added filtering to RNTester example screens

Test Plan:
----------

Run RNTester and tap into example screens. Confirm examples still display, and that typing into the filter field filters the list of examples.

### DIFF
--- a/RNTester/js/RNTesterExampleContainer.js
+++ b/RNTester/js/RNTesterExampleContainer.js
@@ -42,7 +42,7 @@ class RNTesterExampleContainer extends React.Component {
       const Example = this.props.module.examples[0].render;
       return <Example />;
     }
-    
+
     const filter = ({example, filterRegex}) => filterRegex.test(example.title);
 
     const sections = [
--- a/RNTester/js/RNTesterExampleContainer.js
+++ b/RNTester/js/RNTesterExampleContainer.js
@@ -12,6 +12,7 @@
 const React = require('react');
 const {Platform} = require('react-native');
 const RNTesterBlock = require('./RNTesterBlock');
+const RNTesterExampleFilter = require('./RNTesterExampleFilter');
 const RNTesterPage = require('./RNTesterPage');
 
 class RNTesterExampleContainer extends React.Component {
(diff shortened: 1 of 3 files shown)


RNTester/js/RNTesterExampleContainer.js
```diff
@@ -37,9 +38,25 @@ class RNTesterExampleContainer extends React.Component {
       return <this.props.module />;
     }
 
+    const filter = ({example, filterRegex}) => filterRegex.test(example.title);
```
rickhanlonii: What do you think about keeping the old behavior for examples that only export one example?

```suggestion
    if (this.props.module.examples.length === 1) {
      const Example = this.props.module.examples[0].render;
      return <Example />;
    }
    
    const filter = ({example, filterRegex}) => filterRegex.test(example.title);
```

grabbou:
Note: This PR has been merged with failing tests that are getting fixed in here: https://github.com/facebook/react-native/pull/22861

I understand that our CI may be often returning false-positive results. I think we should make sure that - when we merge a PR that has failing tests - this is expected.

rickhanlonii:
Ah, my bad here, sorry about that!

PR #22783, opened 2018-12-24
RCTAsyncLocalStorage: don't reset the data when have no access to the manifest

aleh:
Starting with an empty dictionary is a reasonable fallback in case the manifest cannot be parsed, however it would lead to data loss when the app temporarily has no access to its documents due to data protection being enabled and the device being locked. 

Changelog:
----------

[iOS] [Fixed] - Fix a data loss problem in RCTAsyncLocalStorage happening when data protection is enabled for the app which starts when the device is locked due to background push notification, geofence event, etc.

Test Plan:
----------

Only checked with an app we are making. Before the fix, when data protection was turned on and the app would start due to a geofence event while the phone was locked, then it would begin with an empty state, which would be later successfully saved when the user would unlock the phone and start the app manually. After the fix, the app would silently crash in this case, which is far from ideal behavior, but at least no data loss would occur.

### DIFF
--- a/React/Modules/RCTAsyncLocalStorage.m
+++ b/React/Modules/RCTAsyncLocalStorage.m
@@ -237,16 +237,33 @@ RCT_EXPORT_MODULE()
     }
     RCTHasCreatedStorageDirectory = YES;
   }
+
   if (!_haveSetup) {
-    NSDictionary *errorOut;
+    NSDictionary *errorOut = nil;
     NSString *serialized = RCTReadFile(RCTGetManifestFilePath(), RCTManifestFileName, &errorOut);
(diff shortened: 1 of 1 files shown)


aleh:
Hey guys! This test fails even without my fix, just tried this on master. CC @shergin

radeno:
@aleh @cpojer @Krizzu is this critical? i think losing data is important issue.
Or should be moved to https://github.com/react-native-community/react-native-async-storage

cpojer:
Yeah I think this should go into the new async-storage repo. Let's close this PR here.

@aleh we moved this code out of React Native as part of the Lean Core (#23313) effort. Please reopen your PR there. I'm sure it will be possible to merge it much faster :)

aleh:
Reopened in the async-storage repo: https://github.com/react-native-community/async-storage/pull/270

PR #22785, opened 2018-12-25
[iOS] Fix isBatchActive of RCTCxxBridge

zhongwuzw:
Seems we lost handler of `isBatchActive` from [RCTBatchedBridge a86171a](https://github.com/facebook/react-native/blob/a86171a48284c440226a79a26c6d6a4704d401e9/React/Base/RCTBatchedBridge.m) to [RCTCxxBridge 5bc7e39](https://github.com/facebook/react-native/commit/b774820dc29eae5a09e570384636d353afe213c2#diff-a2a67635fffd7b690d14dc17ae563a71).

Changelog:
----------

[iOS] [fixed] - Fix isBatchActive of RCTCxxBridge

Test Plan:
----------

`isBatchActive` always equal to `NO` before this `PR`, after it, if batched message is being invoked, `isBatchActive` equal to `YES`.

### DIFF
--- a/React/CxxBridge/RCTCxxBridge.mm
+++ b/React/CxxBridge/RCTCxxBridge.mm
@@ -170,7 +170,6 @@ struct RCTInstanceCallback : public InstanceCallback {
 
 @implementation RCTCxxBridge
 {
-  BOOL _wasBatchActive;
   BOOL _didInvalidate;
   BOOL _moduleRegistryCreated;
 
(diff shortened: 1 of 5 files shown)


cpojer:
Hello @zhongwuzw,

it seems like this was removed intentionally because it was both unused and can lead to wrong information due to races. Can you explain your use case and reasoning for why you brought it back and why we should include this in React Native again? Thank you.

zhongwuzw:
@cpojer Hi, `isBatchActive` is used to check wether batch calls are running in `setNeedsLayout` of `RCTUIManager` currently, is this really has any race issue? 🤔 the setter do on `JavaScript Thread`, seems we don't need to worry about race condition? 

Include this check can decrease layout process times because if there has batch calls, it would do layout end of it.

cpojer:
Got it, I'm gonna land it since there are still references to it in the code. If somebody deems it not useful, I'll let them remove it in the future across the whole codebase.

PR #22821, opened 2018-12-28
Fix Picker.onValueChange on Android sometimes not fired due to race condition

### LINKED ISSUE #15556: Android Picker Not Consistently Firing onValueChange
### Is this a bug report?

Yes. I have also seen this issue on [stackoverflow](https://stackoverflow.com/questions/45292719/picker-onvaluechange-not-being-called-on-every-selection).

Yes

### Environment

1. `react-native -v`: 2.0.1
2. `node -v`: v8.0.0
3. `npm -v`: 5.0.0
4. `yarn --version`: 0.24.5

- Target Platform: Android
- Development Operating System: macOS
- Build tools: N/A

### Steps to Reproduce

Using the Picker component in Android, the first `onValueChange` fires, but the next one does not. The one after does...this continues and results in about 33% of the time `onValueChange` does not fire.

I created a [snack](https://snack.expo.io/BJI5J2Edb) with a basic example.

### Expected Behavior

I expected `onValueChange` to fire if the Picker item selected is different from the current selection.

### Actual Behavior

About 33% of the time `onValueChange` will not fire, but the text on the Picker does change, which is confusing to the user. The other times, everything works as expected.

### Reproducible Demo

I created a [snack](https://snack.expo.io/BJI5J2Edb) with a basic example. Select one option, then another and, in my experience, it will not consistently log to console.

Kudo:
Changelog:
----------

[Android] [Fixed] - Fix Picker.onValueChange sometimes not fired due to race condition. Fixes #15556

Root Cause:
------------

Please check my code snippet at https://snack.expo.io/@kudochien/android-picker-issue
and try to select different items on Picker to see if console.log() hit.
If calling setState() with some latency, e.g. setTimeout() or changes from redux,
the second time changing picker item on UI, the onValueChange() will be not fired. 

The root cause comes from the `forceUpdate` in PickerAndroid.android.js.
If user's setState() update comes after forceUpdate(), the flow will be:
1. First time select picker item
2. onValueChange + forceUpdate
3. user's setState() + componentDidUpdate + setNativeProps({ selected: ... })
4. mSuppressNextEvent = true
5. Second time select picker item
6. Since mSuppressNextEvent is true, the onValueChange will not be fired.

Solution:
---------

Like other controlled components, disable change listener during setup values from JS.
Android Spinner `setSelection(int position)` is asynchronous call, i.e. will fire onItemSelected in next run loop and is not suitable for us.
`setSelection(int position, boolean animate)`, however, is synchronous call which I used.

Some more references about setSelection:
https://stackoverflow.com/a/43512925/2590265
http://androidxref.com/8.1.0_r33/xref/frameworks/base/core/java/android/widget/AbsSpinner.java#276
The two arguments version will use `setSelectionInt()` which set mBlockLayoutRequests as true to prevent onItemSelected call from next layout().

I also moved the setOnItemSelectedListener() call after onLayout to prevent onValueChange() during intialization.

Test Plan:
----------

1. Test PickerExample in RNTester
2. Test my test cases at https://snack.expo.io/@kudochien/android-picker-issue with and without setTimeout()
3. I also removed some FlowFixMe, tested with this command to ensure no flow errors:
```
yarn flow start --flowconfig-name .flowconfig.android
yarn flow status --flowconfig-name .flowconfig.android
```

### DIFF
--- a/Libraries/Components/Picker/PickerAndroid.android.js
+++ b/Libraries/Components/Picker/PickerAndroid.android.js
@@ -49,7 +49,6 @@ type Item = $ReadOnly<{|
 |}>;
 
 type PickerAndroidState = {|
-  initialSelectedIndex: number,
   selectedIndex: number,
   items: $ReadOnlyArray<Item>,
 |};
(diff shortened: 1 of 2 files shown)


cpojer:
Thanks for this PR and the extensive test plan. Fingers crossed that it is gonna cover all cases and won't cause weird issues but your code looks solid :)

PR #22827, opened 2018-12-29
Add filtering to RNTester examples

rickhanlonii:
This PR adds filtering to the examples in RNTester. This will allow us to filter the examples directly instead of scrolling (which is unreliable). Note, this replaces https://github.com/facebook/react-native/pull/22777

Additionally, it also:
- Adds and consolidates flow types for examples
- Updates all examples to export `title` `description` and `examples` consistently
- Refactors existing e2e tests to use the new filtering

Todo in a followup:
- Refactor remaining modules that include multiple examples but do not export them individually (and thus cannot be filtered)

Test Plan:
----------
Here is an example of one of the tests that now use filtering:

e2e


RNTester/js/AnimatedGratuitousApp/AnExApp.js
```diff
@@ -195,11 +195,6 @@ class Circle extends React.Component<any, any> {
 }
 
 class AnExApp extends React.Component<any, any> {
-  static title = 'Animated - Gratuitous App';
```
rickhanlonii: Many of the example files used static properties to declare the title and description. This didn't work great with flow types and is inconsistent with the other examples so I moved them all to the exports

RNTester/js/DatePickerIOSExample.js
```diff
@@ -14,20 +14,15 @@ const React = require('react');
 const ReactNative = require('react-native');
 const {DatePickerIOS, StyleSheet, Text, TextInput, View} = ReactNative;
 
-class DatePickerExample extends React.Component<
+class WithDatePickerData extends React.Component<
```
rickhanlonii: This needed refactored to export multiple examples

RNTester/js/ExampleTypes.js
```diff
@@ -1,29 +0,0 @@
-/**
```
rickhanlonii: I moved these types to shared/RNTesterTypes

RNTester/js/PickerExample.js
```diff
@@ -13,115 +13,115 @@
...
 
 const {Picker, Text} = ReactNative;
 
 const Item = Picker.Item;
 
-class PickerExample extends React.Component<{}, $FlowFixMeState> {
-  static title = '<Picker>';
-  static description =
-    'Provides multiple options to choose from, using either a dropdown menu or a dialog.';
+class BasicPickerExample extends React.Component<{}, $FlowFixMeState> {
```
rickhanlonii: Also needed to refactor this to export multiple examples

RNTester/js/RNTesterExampleContainer.js
```diff
@@ -32,17 +44,81 @@ class RNTesterExampleContainer extends React.Component {
...
+      </View>
+    );
+  }
+
   render(): React.Element<any> {
     if (!this.props.module.examples) {
       return <this.props.module />;
     }
 
+    if (this.props.module.examples.length === 1) {
```
rickhanlonii: Here are the main changes in this pull

If there is only one example exported, we just show the single example. This should maintain the existing behavior for modules until they're refactored to export multiple examples.

If there's more than one example exported, we add a filter which will filter by the example title

rickhanlonii:
cc @CodingItWrong

rickhanlonii:
Spoke to @TheSavior and @CodingItWrong offline, closing this in favor of https://github.com/facebook/react-native/pull/22777 and breaking these changes out into multiple PRs

PR #22828, opened 2018-12-30
Add filtering to e2e tests

rickhanlonii:
This PR adds filtering for e2e test examples using the new examples filter introduced in https://github.com/facebook/react-native/pull/22777

To do that we:
- Add a `testID` to `RNTesterExampleFilter` to select an example
- Refactor a few examples to export multiple examples for filtering
- Update all tests to filter by example title

Test Plan:
----------

Here is an example of one of the tests that now use filtering:

e2e

### DIFF
--- a/RNTester/e2e/__tests__/Button-test.js
+++ b/RNTester/e2e/__tests__/Button-test.js
@@ -9,37 +9,36 @@
  */
 
 /* global device, element, by, expect */
+const {
+  openComponentWithLabel,
+  openExampleWithTitle,
+} = require('../e2e-helpers');
 
 describe('Button', () => {
   beforeAll(async () => {
     await device.reloadReactNative();
-    await element(by.id('explorer_search')).replaceText('<Button>');
-    await element(
-      by.label('<Button> Simple React Native button component.'),
-    ).tap();
-  });
-
-  afterAll(async () => {
-    //TODO - remove app state persistency, till then, we must go back to main screen,
-    await element(by.label('Back')).tap();
+    await openComponentWithLabel(
+      '<Button>',
+      '<Button> Simple React Native button component.',
+    );
   });
 
   it('Simple button should be tappable', async () => {
-    await element(by.id('example_search')).replaceText('Simple Button');
+    await openExampleWithTitle('Simple Button');
     await element(by.id('simple_button')).tap();
     await expect(element(by.text('Simple has been pressed!'))).toBeVisible();
     await element(by.text('OK')).tap();
   });
 
   it('Adjusted color button should be tappable', async () => {
-    await element(by.id('example_search')).replaceText('Adjusted color');
+    await openExampleWithTitle('Adjusted color');
     await element(by.id('purple_button')).tap();
     await expect(element(by.text('Purple has been pressed!'))).toBeVisible();
     await element(by.text('OK')).tap();
   });
 
   it("Two buttons with JustifyContent:'space-between' should be tappable", async () => {
-    await element(by.id('example_search')).replaceText('Fit to text layout');
+    await openExampleWithTitle('Fit to text layout');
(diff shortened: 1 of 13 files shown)


RNTester/e2e/__tests__/Switch-test.js
```diff
@@ -13,69 +13,85 @@
 const jestExpect = require('expect');
 
 describe('Switch', () => {
-  beforeEach(async () => {
+  beforeAll(async () => {
```
rickhanlonii: Switched this to speed up the test so we reload once for the file instead of after every test

RNTester/js/DatePickerIOSExample.js
```diff
@@ -14,18 +14,15 @@ const React = require('react');
...
-  static defaultProps = {
-    date: new Date(),
-    timeZoneOffsetInHours: (-1 * new Date().getTimezoneOffset()) / 60,
-  };
+type State = {
+  date: Date,
+  timeZoneOffsetInHours: number,
+};
 
+class WithDatePickerData extends React.Component<{}, State> {
```
rickhanlonii: Needed to refactor this to export multiple examples to filter on

RNTester/js/PickerExample.js
```diff
@@ -13,115 +13,119 @@
...
 
-class PickerExample extends React.Component<{}, $FlowFixMeState> {
-  static title = '<Picker>';
-  static description =
-    'Provides multiple options to choose from, using either a dropdown menu or a dialog.';
+type State = {
+  value: string,
+};
 
+class BasicPickerExample extends React.Component<{}, State> {
```
rickhanlonii: Also needed to refactor this to export multiple examples

RNTester/js/TouchableExample.js
```diff
@@ -32,108 +32,6 @@ const forceTouchAvailable =
     NativeModules.PlatformConstants.forceTouchAvailable) ||
   false;
 
-exports.displayName = (undefined: ?string);
```
rickhanlonii: Just moved the exports to the bottom for consistency

RNTester/e2e/__tests__/Switch-test.js
CodingItWrong: This makes sense; I switched to `beforeEach` to try to decrease flake of the Switch tests, but it wasn't 100% successful, and it makes sense that the filter might serve the same purpose.

RNTester/e2e/__tests__/Switch-test.js
```diff
@@ -13,69 +13,85 @@
...
     await device.reloadReactNative();
     await element(by.id('explorer_search')).replaceText('<Switch>');
     await element(by.label('<Switch> Native boolean input')).tap();
   });
 
-  it('Switch that starts off should switch', async () => {
-    const testID = 'on-off-initial-off';
-    const indicatorID = 'on-off-initial-off-indicator';
+  describe('Switches can be set to true or false', () => {
+    beforeEach(async () => {
```
CodingItWrong: Nice, I like how the filtering encourages grouping examples.

RNTester/js/PickerExample.js
CodingItWrong: Cool, having all the example modules be more consistent will also be helpful for new contributors.

RNTester/e2e/__tests__/Button-test.js
```diff
@@ -25,18 +25,21 @@ describe('Button', () => {
   });
 
   it('Simple button should be tappable', async () => {
+    await element(by.id('example_search')).replaceText('Simple Button');
```
TheSavior: This seems like a good helper

rickhanlonii: Thanks for the suggestion, added helpers

Issue #17475, opened 2018-01-07
[Proposal] Keep RCTPropsAnimatedNode and connection to UIView around until _delete_ animation completes.


gavrix:
This is an issue of 2 independent animation modules in RN not playing nicely together.

1. `Animated` — animations are being removed as soon as the associated React component is removed for the graph.
2. `LayoutAnimation` — since it wraps UIView- related manipulation inside some animation blocks — those animations are still running when React view is gone but corresponding UIView is still being removed with animation. (UIView will be completely removed when _delete_ animation completes).

Now, what _is_ the issue: if both of those animation mechanisms are used it is possible to have a situation when UIView is still on screen, but `Animated`'s animation run on that view would be prematurely terminated.

### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment
Environment:
  OS: macOS High Sierra 10.13.2
  Node: 8.9.2
  Yarn: Not Found
  npm: 5.5.1
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.49.5 => 0.49.5

### Steps to Reproduce

Consider app provided in demo section (on Expo)

App represents a list where each item is a swipe-to-reveal. There is context "Remove" action under each item.
Revealing is done via `Animated` module and achieved using `translateX`transform.  
These are the steps to reproduce visual glitch:

1. Swipe left to reveal some items "Remove" context action
2. Press Remove
3. Observe transitions and animations within next second.

### Expected Behavior
According LayoutAnimation, item being removed is supposed to fade away, item right after it is supposed slide in on removed item's position. Visual state of item being removed is not changed except opacity animation.

### Actual Behavior
Visual state of the item being removed _immediately resets_ to default, wiping out `translateX` transform. This happens because Animation Node used to control that transform value is disconnected from React component that was just removed from the graph. On the native side, `RCTPropsAnimatedNode` upon disconnection calls `restoreDefaultValues` that removes any transform from the native UIView.

### Reproducible Demo
Here's the link to Expo fragment demonstrating described behaviour: https://snack.expo.io/HyiDbGeEG

### Thoughts
[...]

PR #22839, opened 2019-01-02
[iOS] [Animated] Fix Animated-driven transform state when deleting views from view hierarchy 

gavrix:
This is an implementation of proposal https://github.com/facebook/react-native/issues/17475.
I haven't been using ReactNative since original evaluation about a year ago, upon revisiting I see this issue is still around.

Issue:
------

When deleting a view which UI state (like transform) has been connected to AnimatedProps, said state is reset to defaults before running deleting animation (fade to transparent using opacity, by default). This makes it impossible to implement nice swipe-to-delete experience (when swipe is driven by native driver)

Proposal
---------

To iterate what I described in https://github.com/facebook/react-native/issues/17475: instead of disconnected animated component from native view when component is unmounted, keep that connection at native side around until native view is fully purged from the hierarchy.

Concerns:
----------

I'm not 100% sure this is a correct approach. May there be legitimate use cases when AnimatedProps has to be disconnected from native view while latter has to remain in the hierarchy? If anyone has broader context around this mechanism I'd like to discuss.

Changelog:
----------

[iOS] [Changed] - AnimatedProps isn't disconnected from native view (by calling `__disconnectAnimatedView`) when component is unmounted on js side. Instead, runtime on native side is keeping track of such connections and disconnects when native view is fully removed from hierarchy.

Test Plan:
----------

As a proof-of-concept please see gif provided in https://github.com/facebook/react-native/issues/17475.
I wanted to provide a proper integration test but couldn't wrap my head around how to best do it. Any help or hint on direction would much appreciated.

### DIFF
--- a/Libraries/Animated/src/nodes/AnimatedProps.js
+++ b/Libraries/Animated/src/nodes/AnimatedProps.js
@@ -75,9 +75,6 @@ class AnimatedProps extends AnimatedNode {
   }
 
   __detach(): void {
-    if (this.__isNative && this._animatedView) {
-      this.__disconnectAnimatedView();
-    }
     for (const key in this._props) {
       const value = this._props[key];
       if (value instanceof AnimatedNode) {
--- a/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h
+++ b/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h
@@ -11,7 +11,7 @@
 
 @protocol RCTValueAnimatedNodeObserver;
 
-@interface RCTNativeAnimatedNodesManager : NSObject
+@interface RCTNativeAnimatedNodesManager : NSObject<RCTUIManagerDelegate>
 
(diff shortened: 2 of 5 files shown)


matthargett:
@hramos it looks like this is dependent on the iOS snapshot test fix PR by @grabbou #22861, if we want CI to be green before merge.

@gavrix does Android not have this problem as well?

gavrix:
I came to my attention that [`Reanimated`](https://github.com/kmagiera/react-native-reanimated) implements the same functionality as Animated and somehow doesn't have this issue. I am going to close this issue and consider Reanimated preferred animation framework.

matthargett:
That solution works for you, but should this PR still be merged to fix the underlying issue in RN? It seems so to me :)

gavrix:
It was an experiment and only supported on iOS and I was mainly asking for feedback and I found that issue is not reproducible with Reanimated (which is basically an Animated replacement)

PR #22844, opened 2019-01-02
Fix expose originalConsole issue

shenjunru:
The "originalConsole" expose is incorrect in the "polyfills/console.js"

Changelog:
----------

[GENERAL] [Fixed] - Fixed originalConsole expose issue

Test Plan:
----------

Visit "originalConsole" should get original console object.
Ex: Debug iOS in the Safari Debugger


cpojer:
Thank you for this pull request, you are right it seems like this code was broken before.

PR #22861, opened 2019-01-03
Fix snapshot tests on iOS

grabbou:
This PR fixes failing snapshot tests on iOS that were caused by https://github.com/facebook/react-native/commit/bd79303d5bd043431d249f58caf86e6ec42b9a47

What it does is it hides newly introduced "filter" input when running snapshot tests. 

It also fixes a glitch that happened when there was only one example to show (e.g. ART Example - missing RNTesterPage wrapper).

CC: @rickhanlonii @CodingItWrong

### DIFF
--- a/RNTester/js/RNTesterApp.ios.js
+++ b/RNTester/js/RNTesterApp.ios.js
@@ -177,7 +177,10 @@ RNTesterList.ComponentExamples.concat(RNTesterList.APIExamples).forEach(
         render() {
           return (
             <SnapshotViewIOS>
-              <RNTesterExampleContainer module={ExampleModule} />
+              <RNTesterExampleContainer
+                module={ExampleModule}
+                displayFilter={false}
+              />
             </SnapshotViewIOS>
           );
         }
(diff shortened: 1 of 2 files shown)


grabbou:
Note: "test_ios" is expected to pass.

grabbou:
Closing to test it on the "0.58-stable" branch first.

hramos:
Thanks for updating these! How did your tests on 0.58 fare?

cpojer:
@grabbou what's your plan with this?

CodingItWrong:
I'm in the process of updating the snapshots to include the filter control, but I have a blocker I've raised in Discord on tvOS not building.

cpojer:
I think #23152 is enough to get these working again, so I'm gonna close this. Happy to reopen if you think it's still necessary.

ericlewis:
@grabbou is this active still?

grabbou:
I think this PR should be closed. If I remember, the general conclusion was to regenerate  the snapshots rather than adjusting the code.

PR #22877, opened 2019-01-05
bump okhttp to 3.12.1

dulmandakh:
Bump OkHTTP to 3.12.1.

Changelog:
----------
[Android] [Changed] - bump OkHTTP 3.12.1

Test Plan:
----------
CI is green


dulmandakh:
@gengjiawen please help with proguard issue.

gengjiawen:
I will take a look.

gengjiawen:
@dulmandakh Try add those in https://github.com/facebook/react-native/blob/master/ReactAndroid/proguard-rules.pro
```proguard
-dontwarn javax.naming.**
-keep class javax.naming.** {*;}
-dontwarn org.conscrypt.**
-keep class org.conscrypt.** {*;}
-dontwarn sun.security.**
-keep class sun.security.** {*;}
```

ReactAndroid/build.gradle
```diff
@@ -330,7 +330,9 @@ dependencies {
     api 'com.google.code.findbugs:jsr305:3.0.2'
     api "com.squareup.okhttp3:okhttp:${OKHTTP_VERSION}"
     api "com.squareup.okhttp3:okhttp-urlconnection:${OKHTTP_VERSION}"
-    api 'com.squareup.okio:okio:1.14.0'
+    api 'org.bouncycastle:bcprov-jdk15on:1.60'
```
gengjiawen: Why not just use `okhttp-tls`.

dulmandakh: I looked into https://github.com/square/okhttp/blob/master/CHANGELOG.md and added those lines

hramos:
Just curious - what do we gain from bumping to 3.12.1?

dulmandakh:
@hramos previous versions didn't have connection timeout, but timeouts for connect, read and write. Once landed I plan to create a PR to do that. But need to bump okhttp in FB repo too. Also this version added support for TLS 1.3, but requires more dependencies to work. I'll work on that too.

Thank you

PR #22883, opened 2019-01-06
Add listener for non-value animated node

osdnk:
Changelog:
----------
[Changed][General] Move callback-related logic to `AnimatedNode` class in order to make it possible to add the listener for other animated nodes than `AnimatedValue`. 

I observed that native code appears to be fully prepared for listening not only to animated value but animated nodes generally. Therefore I managed to modify js code for exposing `addListener` method from `AnimatedNode` class instead of `AnimatedValue`. It called for some minor changes, which are not breaking.

If you're fine with these changes, I could add proper docs if needed.

Test Plan:
----------
The most simple example to verify whether it works is to use code like this.
```javascript
  constructor(props) {
    super(props)
    this.A = new Animated.Value(0);
    this.B = Animated.add(this.A, 10)
    this.B.addListener(({ value }) => console.log(value))

    Animated.timing(
      // Animate value over time
      this.A, // The value to drive
      {
        useNativeDriver: false,
        duration: 5000,
        toValue: 100, // Animate to final value of 1
      },
    ).start();
  }
  render() {
    return (
        <Animated.View style={{ width: 100, height: 100, backgroundColor: 'red', transform: [{ translateX: this.B }] }}/>
      </View>
    );
  }
}
```

You might switch `useNativeDriver` in order to verify my solution with and without native driver support.


Libraries/Animated/src/nodes/AnimatedValue.js
```diff
@@ -321,9 +242,7 @@ class AnimatedValue extends AnimatedWithChildren {
     if (flush) {
       _flush(this);
```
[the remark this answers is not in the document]
osdnk: But it's in superclass

osdnk:
iOS tests appear to fail in not-related places

matthargett:
can you rebase against master to see if the latest CI fixes there make this PR go green?

cpojer:
CI is failing because of a Flow error, so this definitely doesn't seem to be ready yet. At FB, it's failing because we access `_listeners` somewhere where we shouldn't, but I can fix that once this PR is actually ready. It also seems to break about 20 tests at this time but I'm unsure whether that is because of mocking or if the Flow error on this PR is pointing at the same issue. Either way, I can try again once this is actually working.

Libraries/Animated/src/nodes/AnimatedNode.js
```diff
@@ -36,11 +43,99 @@ class AnimatedNode {
...
+  }
+
+  /**
+   * Adds an asynchronous listener to the value so you can observe updates from
+   * animations.  This is useful because there is no way to
+   * synchronously read the value because it might be driven natively.
+   *
+   * See http://facebook.github.io/react-native/docs/animatedvalue.html#addlistener
+   */
+  addListener(callback: any): string {
```
osdnk: I know it's not perfect to use `any`, but we need to override this type in `AnimatedValueXY` with another type which is related to another logic. 😭

satya164: @osdnk maybe we can use `callback: (value: any) => mixed` instead?

osdnk: `callback: (value: any) => void`

satya164: @osdnk using `mixed` or `any` is better in this case. `void` will prevent async functions to be used for the listener: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVUCuA7AxgFwEs5swBDAEwoDU5CKAZQgZ3wFNs2AnAClzJgwARmVwBrAFxgeANwGY2UstgCeASjABeAHxgZdChoDeAX3RY8REuSoBZQgA82jFu069+gkeKmz5i8lUNHTAAW0dnYzNUShoDJlYObh4yZhU8aWDdI1QwMGBgMAo4MGY4ULZ8AAtCbABzVBM1AG50WPsnF0T3FLSMniywHLyCopKyiurahqbmoA

osdnk: Okay, well, I didn't want to do it previously, cause it changes existing code

cpojer:
cc @mattharget let me know when this is ready for shipping

osdnk:
@cpojer @matthargett forgot second "t" 😃

cpojer:
Still waiting for a review on this PR.

PR #22884, opened 2019-01-06
Add ability to control scroll animation duration for Android

osdnk:
Motivation:
----------
This is one of the more sought after feature requests for RN:
react-native.canny.io/feature-requests/p/add-speed-attribute-to-scrollto

This PR adds the support to add a "duration" whenever using "scrollTo" or "scrollToEnd" with
a scrollView. Currently this only exists for Android as the iOS implementation will be somewhat more involved.

This PR is also backwards compatible and does not yet deprecate the "animated" boolean. It may not make sense to ever deprecate "animated", as it could be the flag that is used when devs want the system default duration (which is 250ms for Android). I'm not sure what it is for iOS. It would simplify things to remove "animated", though.

Test Plan:
----------

Create a ScollView component that overflows the bounds of the container, e.g.:
```javascript
  render() {
    return (
        <Button onPress={this.scrollToPosition} title="Scroll to Position" color="#841584" />
        <ScrollView style={styles.scrollView} ref={scrollView => this.scrollView = scrollView}>
          {items.map(item => {
            return <View key={item.key} style={styles.scrollViewItem}><Text>{item.text}</Text></View>
          })}
        </ScrollView>
      </View>
    );
  }
```
Using the "scrollView" ref and the Button callback, scroll to a position with a desired duration speed, e.g:
```javascript
scrollToPosition = () => {
  this.scrollView && this.scrollView.scrollTo({
    y: 500
    duration: 1000
  });
};
```
Ensure the legacy "animated" boolean works as expected:
```javascript
scrollToPosition = () => {
  this.scrollView && this.scrollView.scrollTo({
    y: 500
    animated: true
  });
};
```
Test that you can animate the x position as well (change the ScrollView to be horizontal).
Test that the scroll animation is canceled when touching on the scroll view in the middle of an animation.
Test that repeated calls to animate the scrollView don't queue up (e.g. smashing the in the example above, and then tapping into the scrollView should cancel any animation movement).
Vertical scroll:
drive.google.com/file/d/1VhaolHx63GOv0YWTg7ay0GtzW1nIk635/view?usp=sharing

Horizontal scroll:
drive.google.com/file/d/1yq5P1k4-KPPu9BhS6B9liL-NRGX1Z4gA/view?usp=sharing

Related PRs:
----------

Abandoned PR to bring "duration" to iOS:
#1334

Updated docs on react-native-website:
facebook/react-native-website#115

Changelog:
----------

[Android] [Enhancement] - Added support for controlling animation duration speed for methods "scrollTo" and "scrollToEnd"

This is a copy of https://github.com/facebook/react-native/pull/17422 with resolved conflicts and style fixes.


osdnk:
Ok, @hramos 
How about it?

osdnk:
I thought it will be a bit easier. I need to dig into this PR deeped

osdnk:
/@hramos @janicduplessis

Thanks! I simplified it a little bit following my intuition and your advices!
Take a look

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
```diff
@@ -171,8 +171,12 @@ public void flashScrollIndicators(ReactHorizontalScrollView scrollView) {
   @Override
   public void scrollTo(
       ReactHorizontalScrollView scrollView, ReactScrollViewCommandHelper.ScrollToCommandData data) {
-    if (data.mAnimated) {
-      scrollView.smoothScrollTo(data.mDestX, data.mDestY);
+    if (data.mAnimated && data.mDuration != 0) {
+      if (data.mDuration > 0) {
```
janicduplessis: Can you add a comment here about -1 being the default platform duration since it's not super obvious from the code. Same thing for the other places this code is copied.

osdnk:
Here you are @janicduplessis

fkgozali:
Sorry for the churn, but seems like this broke some internal callsites, so we'll have to revert it for now. We'll figure out how to address it and update this thread again -- @cpojer

satya164:
Damn, I was so excited for this :(

Seems the docs still mention it. I was trying to use it and scratching my head why it doesn't work.

PR #22886, opened 2019-01-07
Fix system font scaling issues on nested <Text /> component on Android

mu29:
### Motivation

`allowFontScaling` prop doesn't works as expected on nested `<Text />` component especially when specify `fontSize`. So I added allowFontScaling in validAttributes when creating `RCTVirtualText` component.

### Test Plan

I created a test app with codes below:

```javascript
type Props = {};
  render() {
    return (
        <Text
          style={styles.welcome}
          allowFontScaling={false}
        >
          Welcome to React Native!
        </Text>
        <Text
          style={styles.instructions}
          allowFontScaling={false}
        >
          To get started, edit{' '}
          <Text
            style={styles.instructions}
            allowFontScaling={false}
          >
            App.js
          </Text>
        </Text>
        <Text
          style={styles.instructions}
          allowFontScaling={false}
        >
          {instructions}
        </Text>
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  welcome: {
    fontSize: 20,
    textAlign: 'center',
    margin: 10,
  },
  instructions: {
    fontSize: 14,
    textAlign: 'center',
    color: '#333333',
    marginBottom: 5,
  },
});
```

After setting system font size as big as possible in Android settings menu, run above code.
The result was looks like this:

screenshot_20190107-122542_rnandroidfontscalingissue

And if you apply my commit, the result will looks like this:

screenshot_20190107-122534_rnandroidfontscalingissue

### Changelog

[Android] [Fixed] - Fix allowFontScaling prop to work as expected on nested <Text /> component

### DIFF
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -267,6 +267,7 @@ const RCTVirtualText =
           ...ReactNativeViewAttributes.UIView,
           isHighlighted: true,
           maxFontSizeMultiplier: true,
+          allowFontScaling: true,
         },
         uiViewClassName: 'RCTVirtualText',
       }));


janicduplessis:
cc @rigdern 

Do you know if this is still needed after your recent text inheritance fixes?

janicduplessis:
@mu29 Thanks for working on this! We landed a large refactor to the way inheritance is done for Text on Android recently (see #22917) so this fix should no longer be needed.

PR #22917, opened 2019-01-09
Android Text: More robust logic for handling inherited text props

rigdern:
Purpose
----------

This commit fixes a bug and prepares us for adding support for the `maxContentSizeMultiplier` prop (it's currently only supported on iOS).

Details
----------

Today we don't explicitly track inheritance of text props. Instead we rely on `SpannableStringBuilder` to handle this for us. Consider this example:

```
<Text style={{fontSize: 10}}>
  <Text style={{letterSpacing: 5}}>
    ...
  </Text>
</Text>
```

In today's implementation, the inner text doesn't know about `fontSize` (i.e. its `mFontSize` instance variable is `Float.NaN`). But everything works properly because the outer `Text` told `SpannableStringBuilder` to apply the font size across the entire string of text.

However, today's approach breaks down when computing the value to apply to the `SpannableStringBuilder` depends on multiple props. Suppose that RN Android supported the `maxContentSizeMultiplier` prop. Then calculating the font size to apply to the `SpannableStringBuilder` would involve both the `fontSize` prop and the `maxContentSizeMultiplier` prop. If `fontSize` was set on an outer `Text` and `maxContentSizeMultiplier` was set on an inner `Text` then the inner `Text` wouldn't be able to calculate the font size to apply to the `SpannableStringBuilder` because the outer `Text's` `fontSize` prop isn't available to it.

The `TextAttributes` class solves this problem. Every `Text` has a `TextAttributes` instance which stores its text props. During rendering, a child's `TextAttributes` is combined with its parent's and handed down the tree. In this way, during rendering a `Text` has access to the relevant text props from itself and its ancestors.

This design is inspired by the [`RCTTextAttributes`](https://github.com/facebook/react-native/blob/7197aa026b6d262faa8f4dc6bb3e591f860cdc95/Libraries/Text/RCTTextAttributes.m) class from RN iOS.

Bug Fix
----------

This refactoring happens to fix a bug. Today, when setting `fontSize` on nested Text, `allowFontScaling` is always treated as though it is true regardless of the value on the root `Text`. For example, the following snippets should render "hello" identically, Instead, the bottom snippet renders "hello" as though `allowFontScaling` is true.

```
<Text allowFontScaling={false} style={{fontSize: 50}}>hello</Text>
<Text allowFontScaling={false}><Text style={{fontSize: 50}}>hello</Text></Text>
```

(The repro assumes you've increased your device's font setting so that the font size multiplier is not 1.0.)

Introducing the `TextAttributes` class fixed this. It forced us to think about how inheritance should work for `allowFontScaling`. In the new implementation, `Text` components use the value of `allowFontScaling` from the outermost `Text` component. This matches the behavior on iOS (the `allowFontScaling` prop gets ignored on virtual text nodes because it doesn't appear [in this list](https://github.com/facebook/react-native/blob/3749da13127cb7455d533cb2bc5f2cf37470c0c7/Libraries/Text/Text.js#L266-L269).)

Test Plan:
----------

Verified that `fontSize`, `lineHeight`, and `letterSpacing` (the props that moved to `TextAttributes`) work properly when inherited, overridden, and not specified (so the default value is used). Tested that these are properly affected by `allowFontScaling`. Verified that these props also work with `TextInput`. Verified that `lineHeight` and inline images behave the same way that they did before my change.

Changelog:
----------

[Android] [Fixed] - Obey `allowFontScaling` when setting `fontSize` on nested `Text`

Adam Comella
Microsoft Corp.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -23,6 +23,7 @@ import android.text.method.QwertyKeyListener;
 import android.text.style.AbsoluteSizeSpan;
 import android.text.style.BackgroundColorSpan;
 import android.text.style.ForegroundColorSpan;
+import android.util.TypedValue;
 import android.view.Gravity;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
@@ -654,7 +655,7 @@ public class ReactEditText extends EditText {
     
     // `getEffectiveFontSize` always returns a value so don't need to check for anything like
     // `Float.NaN`.
-    setTextSize(mTextAttributes.getEffectiveFontSize());
+    setTextSize(TypedValue.COMPLEX_UNIT_PX, mTextAttributes.getEffectiveFontSize());
 
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
       float effectiveLetterSpacing = mTextAttributes.getEffectiveLetterSpacing();


ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -86,15 +87,23 @@ private static void buildSpannedFromShadowNode(
       ReactBaseTextShadowNode textShadowNode,
       SpannableStringBuilder sb,
       List<SetSpanOperation> ops,
+      TextAttributes parentTextAttributes,
       int start) {
 
+    TextAttributes textAttributes;
+    if (parentTextAttributes != null) {
```
shergin: Probably we can simplify logic if `parentTextAttributes` always be non-null. So we can implement `TextAttributes::defaultTextAttributes()` (with default values, obviously) and pass into the function initially.

I am not sure about this though. I am also curious how I did this in Fabric.

rigdern: If I had a `defaultTextAttributes()`, I could get rid of the `parentTextAttributes == null` checks.

Some tricky things to watch out for:
  - What if the data type doesn't have a good value to represent "not set" (everything is `float` now and `float` has `Float.NaN`)?
  - We have to ensure that the root text shadow node always adds `fontSize` to the span even if it matches the default value (`ViewDefaults.FONT_SIZE_SP`).

ReactAndroid/src/main/java/com/facebook/react/views/text/TextAttributes.java
```diff
@@ -0,0 +1,143 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/*
```
shergin: Can we move this to the top of the java class?

PR #22217, opened 2018-11-08
Make jest react-native preset to use typescript files

dalcib:
Metro and react-native now have first-class support to Typescript with Babel 7, but react-native Jest preset needs specific settings to work with .ts and .tsx files.
This PR updates react-native Jest preset to use Typescript out of the box.

Test Plan:
----------
* [x]  Check yarn run lint
* [x]  Check yarn run prettier
* [x]  Check yarn run test
There is a test to the hasteImpl file.

Changelog:
----------

[GENERAL] [CHANGE] - Add Typescript to Jest react-native preset.

PR #22973, opened 2019-01-13
chore: bump jest to latest beta

SimenB:
This is a partial revert of #22217 - Jest by default handles ts(x) now.

/cc @rubennorte not sure how you normally handle these updates

Changelog:
----------

[General] [Changed] - Upgrade jest to v24.0.0-alpha.12

Test Plan:
----------

I haven't tested 😅 I was here for another preset PR, so I just did this at the same time.


rubennorte:
@SimenB thanks for looking into this but we're working on an internal PR to upgrade it to this same version. I should be available soon.

SimenB:
Cool! You should include the preset cleanups 🙂

PR #22980, opened 2019-01-14
Added WebView deprecation warning

Titozzz:
Changelog:
----------

Help reviewers and the release process by writing your own changelog entry. See http://facebook.github.io/react-native/docs/contributing#changelog for an example.

[General] [Deprecated] - Deprecation warning for WebView as it has been extracted from core. 

Test Plan:
----------

importing WebView from react-native will now trigger a warning.

### DIFF
--- a/Libraries/react-native/react-native-implementation.js
+++ b/Libraries/react-native/react-native-implementation.js
@@ -14,6 +14,7 @@ const invariant = require('invariant');
 
 let showedListViewDeprecation = false;
 let showedSwipeableListViewDeprecation = false;
+let showedWebWiewDeprecation = false;
 
 // Export React, plus some native additions.
 module.exports = {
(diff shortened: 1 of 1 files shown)


Libraries/react-native/react-native-implementation.js
```diff
@@ -167,6 +168,14 @@ module.exports = {
     return require('VirtualizedList');
   },
   get WebView() {
+    if (!showedWebWiewDeprecation) {
+      console.warn(
+        'WebView has been extracted from react-native core and will be removed in a future release. ' +
+          'See https://github.com/react-native-community/react-native-webview for more information',
```
grabbou: Looks good!

I would consider adding a note about the new `npm` package, similar to what Navigator is doing: https://github.com/facebook/react-native/blob/ff329801edbd7bbfd52716f8e236d13cd46b04a1/Libraries/react-native/react-native-implementation.js#L357

Titozzz: Fixed 😄

Issue #22966, opened 2019-01-11
NetInfo.isConnected.addEventListener behaves incorrectly


benjarwar:
- [X] Review the documentation: https://facebook.github.io/react-native
  - [X] Search for existing issues: https://github.com/facebook/react-native/issues
  - [X] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
```
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 172.94 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.9.4 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
      Android SDK:
        API Levels: 23, 25, 26, 27
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 27.0.3, 28.0.3
        System Images: android-23 | Intel x86 Atom_64
    IDEs:
      Android Studio: 3.1 AI-173.4819257
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.57.8 => 0.57.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
```

## Description
The callback passed to `NetInfo.isConnected.addEventListener` does not fire consistently nor accurately in iOS, `0.57.8`. We cannot reliably determine whether or not the user is connected when following the [documented method](https://facebook.github.io/react-native/docs/netinfo#addeventlistener).

This is an enormous problem on our project, where we're trying to restrict requests and functionality when the user is offline.

## Reproducible Demo
I created a test app with `react-native init`:

https://github.com/benjarwar/NetInfoTest

Simply added a `NetInfo.isConnected` event listener that console logs the connection status on `connectionChange`:

https://github.com/benjarwar/NetInfoTest/blob/master/App.js#L21-L25

[...]

PR #22981, opened 2019-01-14
Add support for new Android NetInfo connectivity monitoring

matt-oakes:
This PR changes NetInfo to use the newer and better supported [NetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback) on API levels 24 and above.

This is in place of the now deprecated BroadcastReceiver method. Both are kept so we can continue supporting down to API level 16. This is done by splitting out the connectivity code into two different classes, organised around a base abstract class to avoid duplication. Which class to use is chosen at runtime with a simple `if` statement in the NetInfo module.

The structure of the `connectionChange` event has not been changed, so this will be a non-breaking change for most users. The slight exception to that is that I have now removed the [previously deprecated](https://github.com/facebook/react-native/commit/fc38fe1736080215e3c462e25081d530a5399dc3) code from the module to avoid needing to support it in the new NetworkCllback code, however, this should be a non-issue as it was deprecated 2 years ago and doesn't even feature in the documentation anymore.

There is a slight difference in behaviour, which is down to a limitation in the old broadcast receiver method of updating the connectivity status. If the device changes from one cellular network type to another (effective network type), for example from 2g to 4g, this previously didn't trigger a new event. Using the new Network Callback method, it does.

Changelog:
----------

[Android] [Changed] - NetInfo on API level 24+ (N) we now use [NetworkCallback]() to receive updates about the network status
[General] [Removed] - Removed the previously deprecated `change` event from the NetInfo module

Test Plan:
----------

* Open the RNTester app to the `NetInfoExample` screen.
* Toggle the Wifi on and off.
* Change the network type in the emulator (changing from LTE to GPRS for example).
* See that the events are fired and have the correct status in the RNTester screen.
* Do this for devices above and below the API level 24 threshold to ensure it works the same on both (aside from the limitation on changes to the effective network type for devices below API level 24, see above).

Gif of it working on Android: 
ezgif com-video-to-gif

### DIFF
--- a/Libraries/Network/NetInfo.js
+++ b/Libraries/Network/NetInfo.js
@@ -21,55 +21,31 @@ const DEVICE_CONNECTIVITY_EVENT = 'networkStatusDidChange';
 
 type ChangeEventName = $Enum<{
   connectionChange: string,
-  change: string,
 }>;
 
-type ReachabilityStateIOS = $Enum<{
-  cell: string,
-  none: string,
-  unknown: string,
-  wifi: string,
+type ConnectionType = $Enum<{
+  // iOS & Android
+  cell: 'cellular',
+  none: 'none',
+  unknown: 'unknown',
+  wifi: 'wifi',
+  // Android only
+  bluetooth: 'bluetooth',
+  ethernet: 'ethernet',
+  wimax: 'wimax',
 }>;
-
-type ConnectivityStateAndroid = $Enum<{
-  NONE: string,
-  MOBILE: string,
-  WIFI: string,
-  MOBILE_MMS: string,
-  MOBILE_SUPL: string,
-  MOBILE_DUN: string,
-  MOBILE_HIPRI: string,
-  WIMAX: string,
-  BLUETOOTH: string,
-  DUMMY: string,
-  ETHERNET: string,
-  MOBILE_FOTA: string,
-  MOBILE_IMS: string,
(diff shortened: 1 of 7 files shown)


matt-oakes:
Rebased this on `master` to resolve the conflicts from 5a87093e1aba11941f23c2716aa04aeb1ffbaca4.

All good to review again 👍

ngandhy:
Hey @matt-oakes did you run across the issues in #22966 (and parts of #8615)? I'm hoping maybe you came across that and fixed it :)
 
Basically that the connectionChange callback doesn't reliably fire. (e.g - if I start connected, then disconnect (get callback) and then connect again (no callback).

Thanks!

matt-oakes:
@ngandhy This PR mainly touches the Android side of that module and leaves the iOS side alone aside from removing the long deprecated API.

With the Lean Core initiative (https://github.com/react-native-community/discussions-and-proposals/issues/6) I would be happy to help move out and maintain this part of the code, however, the exact process for doing that is currently not 100% clear. It should, however, allow fixes and changes like this to move forward.

cpojer:
@matt-oakes given that we are splitting NetInfo out into its own repo, I'm gonna close this PR here and we can discuss the same change on that other repo :)

matt-oakes:
Opened the PR in the new repository 👍

PR #22984, opened 2019-01-14
feat: expose AnimatedEvent

jaulz:
Some third party components rely on the `AnimatedEvent` as a type (see https://github.com/kmagiera/react-native-gesture-handler/blob/master/Swipeable.js#L9). In order to not rely on a path which could change at any time it would be great to expose it.

Changelog:
----------

[General][added] - Expose AnimatedEvent

Test Plan:
----------

n/a


cpojer:
Importing again as something went wrong.

Issue #22985, opened 2019-01-14
Animated gif loops stops after playing once on ios hardware only


chrisnojima:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
  React Native Environment Info:
    System:
      OS: macOS 10.14.1
      CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
      Memory: 10.25 GB / 32.00 GB
      Shell: 3.0.0 - /usr/local/bin/fish
    Binaries:
      Node: 11.6.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.5.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 23.0.2, 25.0.0, 25.0.2, 26.0.0, 26.0.2, 26.0.3, 27.0.3, 28.0.0, 28.0.0, 28.0.2, 28.0.3
        System Images: android-16 | Google APIs Intel x86 Atom, android-17 | Google APIs Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3
      react-native: 0.57.8 => 0.57.8

## Description
This is a regression. There are several auto closed issues related to this (https://github.com/facebook/react-native/issues/22959, https://github.com/facebook/react-native/issues/22788)
On a real device animated gifs will play once and stop. Previously they used to loop forever correctly. Strangely they loop forever in the simulator.

## Reproducible Demo
Let us know how to reproduce the issue. Include a code sample, share a project, or     share an app that reproduces the issue using https://snack.expo.io/. Please follow the guidelines for providing a MCVE: https://stackoverflow.com/help/mcve

Just load an image and source an animated gif.

chrisnojima:
[...]

PR #22987, opened 2019-01-14
fix incorrect type which makes animated gifs not loop forever on device

chrisnojima:
https://github.com/facebook/react-native/issues/22985

This 1 liner fixes the animation looping being broken on ios devices. The original source of the bug is here: https://github.com/facebook/react-native/commit/95ef882#diff-e57b12f931820d7e0949e5cbb2701dcfR35

We set the value to a special large float, and assign it to repeatCount which is also a float, so this should be a float.

Changelog:

[iOS] [Fixed] - Fix animated GIFs not looping forever


pakyo-pan:
Would appreciate if it would be possible to merge this change, thank you!!

PR #22992, opened 2019-01-15
Android TextInput: Support `allowFontScaling` on placeholder

### LINKED ISSUE #18827: [Android] allowFontScaling prop on TextInput does not disable placeholder font scaling
- [✔️] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [✔️] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [✔️] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

I'm using the `allowFontScaling` prop on `Text` elements and `TextInput` elements for some more fine-grained control of the font scaling behavior in my team's app. Specifically, I'm setting `allowFontScaling` to false and doing the font scaling manually, so I can cap the font scaling at a reasonable amount. This approach worked for everything except the placeholder text on `TextInput` elements on Android. It appears that `allowFontScaling={false}` does not turn off the font scaling for placeholder text.

## Environment

Environment:
  OS: macOS High Sierra 10.13.2
  Node: 9.2.0
  Yarn: Not Found
  npm: 5.8.0
  Watchman: 4.9.0
  Android Studio: 3.1 AI-173.4670197

  react-native: 0.55.0 => 0.55.0

## Steps to Reproduce

I have put together a sample project at https://github.com/Elethier/react-native-placeholder-scaling-bug. I reproduced the issue on a Samsung Galaxy J7 running Android 6.0.1 and a Nexus 5X running Android 8.1.0.

I should also note that I was unable to reproduce this issue on a fresh native Android project, which suggests to me that it's a bug specific to React Native.

## Expected Behavior

I would expect that if the prop `allowFontScaling` is set to false, both the placeholder text and the user-inserted text would not scale up based on the phone's system font size.

## Actual Behavior

The placeholder text font size scales up with the system font size, even though the user-inserted text does not. Here is a GIF showing the behavior: 

placeholder scaling bug

For reference, here is a screenshot of the phone's font size settings. This is on a Samsung Galaxy J7.

font size screenshot

rigdern:
Currently, the `TextInput's` placeholder is always sized as though `allowFontScaling` is `true`.

Note that `allowFontScaling` works fine for the content of the `TextInput`. The reason is that we set the font size in two places: in the shadow node and in the Android view. The shadow node logic determines the size of the content and the Android view logic determines the size of the placeholder. The handler for the `allowFontScaling` prop is only present in the shadow node logic. Consequently, the Android view logic always uses the default value of `true` for the `allowFontScaling` prop.

The fix is to add logic for handling the `allowFontScaling` prop to the Android view.

It would be great if we could handle all text props in one spot instead of duplicating code between the shadow node and the Android view. That would eliminate this whole class of bugs. However, I don't have enough familiarity with the history of this code to know how hard that would be or if it's even possible.

Fixes #18827.

Test Plan:
----------

Verified that `Text`, the content of a `TextInput`, and the placeholder of a `TextInput` are all rendered the same way for all values of `allowFontScaling` (`undefined`, `true`, `false`).

Changelog:
----------

[Android] [Fixed] - TextInput: Support `allowFontScaling` on placeholder

Adam Comella
Microsoft Corp.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -644,6 +644,13 @@ public class ReactEditText extends EditText {
     applyTextAttributes();
   }
 
+  public void setAllowFontScaling(boolean allowFontScaling) {
+    if (mTextAttributes.getAllowFontScaling() != allowFontScaling) {
+      mTextAttributes.setAllowFontScaling(allowFontScaling);
+      applyTextAttributes();
+    }
+  }
+
   public void setFontSize(float fontSize) {
     mTextAttributes.setFontSize(fontSize);
     applyTextAttributes();
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -319,6 +319,11 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
     view.setLetterSpacingPt(letterSpacing);
   }
 
+  @ReactProp(name = ViewProps.ALLOW_FONT_SCALING, defaultBoolean = true)
+  public void setAllowFontScaling(ReactEditText view, boolean allowFontScaling) {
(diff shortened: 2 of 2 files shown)


janicduplessis:
@facebook-github-bot shipit

cpojer:
Will re-import, something went wrong yesterday.

Issue #22990, opened 2019-01-15
Creating Standalone JS File For Each Native Component


TheSavior:
We are going to be generating native code from flow types in JS for our native components. In order to do this we need to have an individual JS file for each native component with complete flow types.

[Step 1](https://github.com/facebook/react-native/issues/21342), [Step 2](https://github.com/facebook/react-native/issues/21485), [Step 3](https://github.com/facebook/react-native/issues/21581), and [Step 4](https://github.com/facebook/react-native/issues/22100) have all been steps towards this goal.

In a bunch of our JS files we choose a component based on platform, typically looking something like this:
```
const SwitchNativeComponent: SwitchNativeComponentType =
  Platform.OS === 'android'
    ? (requireNativeComponent('AndroidSwitch'): any)
    : (requireNativeComponent('RCTSwitch'): any);
```

we want to change this code so that each `requireNativeComponent` call lives in its own file, thus changing the above code to look like this:

```
const SwitchNativeComponent: SwitchNativeComponentType =
  Platform.OS === 'android'
    ? require('AndroidSwitchNativeComponent')
    : require('RCTSwitchNativeComponent');
```

We need your help to make this change! 

# How to help
Pick one of the native components from the list below. Comment that you are working on that file and try to choose one that nobody else has commented about already.

Search the codebase for .js files that include that name. We are specifically looking for `requireNativeComponent` calls. So if you pick the component `AndroidCheckBox` we'd be looking for [this callsite](https://github.com/facebook/react-native/blob/24f8d4d3db483dab913486b3eb0b68572274eac7/Libraries/Components/CheckBox/CheckBox.android.js#L79-L81).

Create a new js file suffixed with 'NativeComponent' right next to the file that you found. In this case we'd create a `AndroidCheckBoxNativeComponent.js` file in `Libraries/Components/CheckBox`. 

We should move the call to requireNativeComponent from `Checkbox.android.js` to this file, including the flow type if one exists, or defining one if it doesn't. 

A pastebin of what I think the changes look like for this component are here: https://pastebin.com/RFpdT76V

One thing to note is that the majority of types necessary for these files shouldn't be imported from other files. Our codegen can't yet handle importing types from other files so in most cases I'd prefer to duplicate the type in this new file if it is needed. There are some exceptions to this that the codegen *can* handle that I'd expect most of these files will need. Specifically:
```
import type {ViewProps} from 'ViewPropTypes';
import type {SyntheticEvent} from 'CoreEventTypes';
[...]

PR #22995, opened 2019-01-15
[Standalone for NativeComponents] AndroidViewPagers.js

VisibleMarkov:
[Android][Changed] - All the imports connected to `requireNativeComponent` in `ViewPager` was moved to  a separate file.
Issue in focus: #22990 

Test Plan:
* [x] yarn test
* [x] yarn flow
* [x] yarn lint

### DIFF
--- a/Libraries/Components/ViewPager/AndroidViewPagerNativeComponent.js
+++ b/Libraries/Components/ViewPager/AndroidViewPagerNativeComponent.js
@@ -1,0 +1,15 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow strict-local
+ */
+
+'use strict';
+
+const requireNativeComponent = require('requireNativeComponent');
+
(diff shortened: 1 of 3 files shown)


VisibleMarkov:
@TheSavior, sorry for the ping but I'm not quite able to comprehend the `NSInternalInconsistencyException` being thrown during the IOS build. Did this arise as a result of the changes I have made cause I couldn't link them to `ViewPager`

VisibleMarkov:
Thanks for the clarification. On it! (^_^)

VisibleMarkov:
Nevermind, I found the error  :sweat_smile:. The `PageScrollStateChanged` had a flaw right from the beginning where type of actual parameter wasn't matching with what was declared in the formal parameter. Changes made in the latest commit made the two consistent. I'm sorry for the ping, should have looked at the error message harder. Thank you

cpojer:
Thanks for the PR! While your change is technically breaking, I think you are changing the callback to be consistent so it makes sense to me to change.

VisibleMarkov:
I observed the call to be inconsistent with the rest and found a hard time writing  the flow type. When I cross referenced with the other calls, the parameters always seemed to be an event and hence decided to change it. I was waiting for a review to verify whether that decision was right or wrong. Thanks for the approval (^_^)

PR #23009, opened 2019-01-15
[Standalone for NativeComponents] ToolbarAndroid

sasurau4:
### Changelog
[Android] [Changed] - create standalone JS file for ToolbarAndroid, as part of #22990 

### Question
I have a question.

#### `ImageSource` is sufficiently complex not to be duplicate code?
`ToolbarAndroid` component use 2 types from other files.
They are not included files mentioned by https://github.com/facebook/react-native/issues/22990#issue-399165354.
One type `ColorValue` is very simple and I'll copy it.
But `ImageSource` type is a bit complicated I think it wolud be better to add to the codegen.
How do you think about it? @TheSavior 
Update: -> solved 👍 

### Test Plan
----------
- [x] yarn lint
- [x] yarn flow-check-android
- [x] yarn flow
- [x] yarn test

### DIFF
--- a/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
+++ b/Libraries/Components/ToolbarAndroid/ToolbarAndroid.android.js
@@ -13,7 +13,7 @@
 const React = require('React');
 const UIManager = require('UIManager');
 
-const requireNativeComponent = require('requireNativeComponent');
+const ToolbarAndroidNativeComponent = require('ToolbarAndroidNativeComponent');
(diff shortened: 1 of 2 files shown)


TheSavior:
`ImageSource` seems reasonable to support in the codegen. Thanks for checking!

sasurau4:
@TheSavior Thanks for response!
So, it's all done 👍

TheSavior:
Thanks for contributing!

PR #23011, opened 2019-01-15
Add prop "disableSwipe" to SwipeableFlatList and SwipeableRow

ivnnv:
Motivation:
----------
Sometimes you are rendering multiple SwipeableFlatList at the same time (Ex: one per tab) and some of them wont require swiping because there are not associated quickActions for any (there could be more use cases).
So instead of programatically need to render either FlatList or SwipeableFlatList, using this prop saves you from overcomplicating your render with a simple boolean prop.

Test Plan:
----------
  npm run prettier
  npm run flow-check-ios
  npm run flow-check-android
  Run RNTester app, go to SwipeableFlatList, added prop disableSwipe={true} to component, everything works as expected.

Easy to reproduce PR, just add the prop to `react-native/RNTester/js/SwipeableFlatListExample.js` and run it. Rows wont swipe anymore :)
```
<SwipeableFlatList
  data={data}
  disableSwipe={true}
  bounceFirstRowOnMount={true}
  maxSwipeDistance={160}
  renderItem={this._renderItem.bind(this)}
  renderQuickActions={this._renderQuickActions.bind(this)}
 />
```

Changelog:
----------
[General][Added] - Add prop *disableSwipe* to `SwipeableFlatList.js` and `SwipeableRow.js`

### DIFF
--- a/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
+++ b/Libraries/Experimental/SwipeableRow/SwipeableFlatList.js
@@ -34,6 +34,11 @@ type SwipableListProps = {
    * Callback method to render the view that will be unveiled on swipe
    */
   renderQuickActions: renderItemType,
+
+  /**
+   * Disable swipping on rows
+   */
+  disableSwipe: boolean,
 };
 
 type Props<ItemT> = SwipableListProps & FlatListProps<ItemT>;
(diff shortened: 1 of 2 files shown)


cpojer:
Thank you for this PR. I think the intention you have to simplify this makes sense, but I don't think React Native is the right place for this abstraction. I recommend to you instead to make a higher order component that has a `disableSwipe` prop which will render to either a `SwipeableFlatList` or a `FlatList`. That way you can solve this neatly in your own code, while we don't have to introduce complexity to React Native.

PR #23038, opened 2019-01-17
use okhttp callTimeout to set request timeout

dulmandakh:
OkHTTP 3.12 introduced callTimeout or full-operation timeouts. And here is the snippet from changelog (https://github.com/square/okhttp/blob/master/CHANGELOG.md)

> OkHttp now offers full-operation timeouts. This sets a limit on how long the entire call may take and covers resolving DNS, connecting, writing the request body, server processing, and reading the full response body. If a call requires redirects or retries all must complete within one timeout period.

Previously we set timeout for connect, but not for full-operation. Therefore, network client connect to a server within timeout, but it may take forever to read responses or write requests. This PR fixes it.

Changelog:
----------

[Android] [Changed] - Network request timeout works as expected using callTimeout

Test Plan:
----------
CI is green, and everything works as before. But network timeout will for full-operation or as expected by many developers.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -324,8 +324,8 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
     // client and set the timeout explicitly on the clone.  This is cheap as everything else is
     // shared under the hood.
     // See https://github.com/square/okhttp/wiki/Recipes#per-call-configuration for more information
-    if (timeout != mClient.connectTimeoutMillis()) {
-      clientBuilder.connectTimeout(timeout, TimeUnit.MILLISECONDS);
+    if (timeout != mClient.callTimeoutMillis()) {
+      clientBuilder.callTimeout(timeout, TimeUnit.MILLISECONDS);
     }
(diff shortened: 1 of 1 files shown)


cpojer:
Looks good!

cpojer:
Uh oh, I think we may not be on that version of okhttp at Facebook which makes this pretty hard to land.

dulmandakh:
@hramos would you please update okhttp at FB and merge this PR. IIRC, okhttp have no breaking changes.

cpojer:
@dulmandakh Unfortunately it's not that simple. It seems like we won't be able to upgrade everything depending on RN to okhttp 3.12 anytime soon (possibly months). Is there any way we could work around this?

dulmandakh:
Nope, it depends on okhttp 3.12. Let's wait until okhttp 3.12 lands in FB.

cpojer:
I asked around and there currently aren't any plans to upgrade okhttp internally. It seems that it would be a bit of work to update it for all apps. Previously it wasn't a big deal if FB apps use a different version of okhttp than RN in open source but it does matter when RN makes use of newer APIs.

I'm going to close this PR as it isn't actionable. We should revisit this whenever we update okhttp at FB.

dulmandakh:
how I can find out if okhttp at FB updated?

cpojer:
Ask me again in a few months, I'm happy to let you know :)

PR #23047, opened 2019-01-17
Update EventEmitter.js

xianglinhe:
added protection for emit() because subscription.listener may have been removed during event loop

Changelog:
----------

[GENERAL][FIX] Avoid crash when subscription listener may have been removed during event loop

### DIFF
--- a/Libraries/vendor/emitter/EventEmitter.js
+++ b/Libraries/vendor/emitter/EventEmitter.js
@@ -176,7 +176,7 @@ class EventEmitter {
         const subscription = subscriptions[i];
 
         // The subscription may have been removed during this event loop.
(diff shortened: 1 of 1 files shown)


matthargett:
[GENERAL][FIX] Avoid crash when subscription listener may have been removed during event loop

hramos:
@xianglinhe can you sign the CLA? Once you do that, we can move forward with the PR.

xianglinhe:
> @xianglinhe can you sign the CLA? Once you do that, we can move forward with the PR.

It's signed now, what else should I do?

cpojer:
Thanks for the PR! :)

PR #23052, opened 2019-01-17
Update references to the CLI

grabbou:
This updates React Native to use latest CLI. We also create Metro configuration, because CLI looks for React Native in "node_modules" by default. Since we are running React Native from source, it will fail to find required files.

To avoid hacky logic to detect if we are running from source backed into the CLI, I decided to leverage the Metro configuration instead.


grabbou:
@hramos - heads up, I fixed a Flow error and a single change to `findPlugins` that (I believe) was causing the tests to fail.

grabbou:
@facebook-github-bot ship

grabbou:
@facebook-github-bot shipit

grabbou:
@hramos I am triggering the import again - the issues seem to be fixed and failures are similar to what we have on master. Hopefully it works for you locally.

PR #23068, opened 2019-01-18
[Standalone for NativeComponents] ProgressBarAndroid

rafaellincoln:
his PR is related to #22990

Changelog:
----------

[Android][Changed] - move the call to requireNativeComponent from ProgressBarAndroid.android.js to ProgressBarAndroidNativeComponent.js

Test Plan:
----------

- [x] npm run lint
- [x] npm run flow
- [x] npm run test

### DIFF
--- a/Libraries/Components/ProgressBarAndroid/ProgressBarAndroid.android.js
+++ b/Libraries/Components/ProgressBarAndroid/ProgressBarAndroid.android.js
@@ -12,12 +12,10 @@
 
 const React = require('React');
 
-const requireNativeComponent = require('requireNativeComponent');
+const ProgressBarAndroidNativeComponent = require('ProgressBarAndroidNativeComponent');
 
-import type {NativeComponent} from 'ReactNative';
 import type {ViewProps} from 'ViewPropTypes';
-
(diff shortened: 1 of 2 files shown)


Libraries/Components/ProgressBarAndroid/ProgressBarAndroid.android.js
```diff
@@ -84,9 +82,9 @@ export type ProgressBarAndroidProps = $ReadOnly<{|
  */
 const ProgressBarAndroid = (
   props: ProgressBarAndroidProps,
-  forwardedRef: ?React.Ref<'AndroidProgressBar'>,
+  forwardedRef: ?React.Ref<ProgressBarAndroidType>,
```
TheSavior: I don't think you need this type exported from this file. You should just be able to do `React.Ref<ProgressBarAndroidNativeComponent>`, or if that doesn't work then `React.Ref<typeof ProgressBarAndroidNativeComponent>`

rafaellincoln: Resolved

PR #23099, opened 2019-01-22
Remove AlertExample from RNTester in iOS

zhongwuzw:
Changelog:
----------

[iOS] [Removed] - Removed AlertExample from RNTester in iOS.

Test Plan:
----------

Before, when we click `Alert` row from `APIS` section, RNTester came to error, after it, `Alert` row be removed, and actually `AlertIOS` row already contains `Alert` row does.

0d321aca-49dc-4742-9c57-1c16f8ffa546

The error like below, the reason is AlertExample not exports examples array.

e5907e50-67be-4636-a2c6-d3353f86868e


cpojer:
Thank you!

PR #23105, opened 2019-01-22
[Android][DevSupport] Add Network Security Config file (fixes #22375)

### LINKED ISSUE #22375: [Android] Android 9 can not access bundler via cleartext request without additional config
## Environment
```
    System:
      OS: Linux 4.18 Arch Linux undefined
      CPU: x64 AMD Ryzen 7 1700X Eight-Core Processor
      Memory: 682.77 MB / 7.79 GB
      Shell: 5.6.2 - /bin/zsh
    Binaries:
      Watchman: 4.7.0 - /usr/local/bin/watchman
    SDKs:
      Android SDK:
        Build Tools: 23.0.1, 23.0.3, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.2, 27.0.3, 28.0.1, 28.0.3
        API Levels: 23, 24, 26, 27, 28
    npmPackages:
      react-native: 0.57.x => 0.57.4 
    npmGlobalPackages:
```

## Description
As of Android  9 (API Level 28), the default behavior is to now [block all clear text requests](https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted), unless a domain is specifically configured to allow this behavior. As a result the bundler URL is not reachable in development mode, with a generic error that offers no context to what the problem could be, or what solutions are available.

The issue is fairly trivial to resolve once the root cause is apparent, but getting to that root cause can take a bit of searching.

## Reproducible Demo
Start a `react-native` app on an Android 9 devices (I only tested physical device), in development mode. The app will not start, and will complain that the bundle can not be fetched.

Salakar:
This fixes #22375. Android API level 28 and above now blocks all [clear text requests](https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted) unless a network configuration rule is added to exclude it specifically.

The packager falls into this category and therefore gets blocked; resulting in the app being unable to connect to it.

Domains/IPs for the config have been taken from [here](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java#L20-L22).

This change only adds to DevSupport android resources - so won't affect builds without DevSupport, e.g. release builds.

Changelog:
----------

[ANDROID] [DevSupport] add Network Security Config file to allow access to packager via cleartext requests in Android API 28 and above. (fixes #22375)

Test Plan:
----------

Built an app API 28 Android App with RN Android building from source, prior to this change the packager would not connect and show the following Error in Android logs;

image

After applying this change the connection is now successful.

### DIFF
--- a/ReactAndroid/src/main/res/devsupport/xml/network_security_config.xml
+++ b/ReactAndroid/src/main/res/devsupport/xml/network_security_config.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+  <domain-config cleartextTrafficPermitted="true">
+    <domain includeSubdomains="false">localhost</domain>
+    <domain includeSubdomains="false">10.0.2.2</domain>
+    <domain includeSubdomains="false">10.0.3.2</domain>
+  </domain-config>
+</network-security-config>


cpojer:
Thanks for this PR!

PR #23122, opened 2019-01-23
[iOS] Add CF_RETURNS_NOT_RETAINED annotate for Objective-C method which returns Core Foundation object

zhongwuzw:
Changelog:
----------

[iOS][Fixed] - Add CF_RETURNS_NOT_RETAINED annotate for Objective-C method which returns Core Foundation object

Test Plan:
----------

Annotate the Objective-C returned objects memory management.

### DIFF
--- a/Libraries/ART/RCTConvert+ART.h
+++ b/Libraries/ART/RCTConvert+ART.h
@@ -15,7 +15,7 @@
 
 @interface RCTConvert (ART)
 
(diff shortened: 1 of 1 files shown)


cpojer:
What does this actually fix? Can you expand on the summary in your pull request to explain why this change is necessary?

cpojer:
Thank you for the explanation!

PR #23128, opened 2019-01-23
Fix tests in template project (0.58-stable)

vovkasm:
This fixes template application sample tests.
1. Fix adding __tests__ folder to npm package
2. Fix tests itself (add babel-core to deps and rename .babelrc to babel.config.js)

Unfortunately some changes can't be applied to master branch because of cli extraction. Moreover some changes (probably addition of babel-core to deps) will not be needed when jest 24 will be released. So this is path specially for 0.58-stable branch.

I think after apply this, we also should do some things:
1. Recheck npmignore patterns in master branch, remove unused and ensure that templates packed correctly.
2. Recheck our deps and config in HelloWorld template after jest 24 will be released.

Changelog:
----------

[General] [Fixed] - Fix tests in template HelloWorld application

Test Plan:
----------

```sh
# Prepare RN tarball for testing
cd /Volumes/work/react-native
npm pack
# clear yarn cache
yarn cache clean

# Test with yarn
cd /Volumes/work/tmp
react-native init --version /Volumes/work/react-native/react-native-0.58.0-rc.3.tgz Sample1
cd Sample1
yarn test
# Tests passed

# Test without yarn (yarn not available on system)
cd /Volumes/work/tmp
react-native init --version /Volumes/work/react-native/react-native-0.58.0-rc.3.tgz Sample1
cd Sample1
npm test
# Tests passed

# Recheck dependencies with fresh reinstall
cd /Volumes/work/tmp/Sample1
rm -rf node_modules yarn.lock package-lock.json
npm install
npm test
# Tests passed
```

### DIFF
--- a/local-cli/.npmignore
+++ b/local-cli/.npmignore
@@ -1,2 +1,3 @@
 __fixtures__
 __tests__
+!/templates/HelloWorld/__tests__
--- a/local-cli/generator/copyProjectTemplateAndReplace.js
+++ b/local-cli/generator/copyProjectTemplateAndReplace.js
@@ -127,6 +127,7 @@ function translateFilePath(path) {
     .replace('_BUCK', 'BUCK')
     .replace('_gitignore', '.gitignore')
     .replace('_gitattributes', '.gitattributes')
+    .replace('_babel.config.js', 'babel.config.js')
     .replace('_babelrc', '.babelrc')
     .replace('_flowconfig', '.flowconfig')
     .replace('_buckconfig', '.buckconfig')
--- a/local-cli/init/init.js
+++ b/local-cli/init/init.js
@@ -90,7 +90,7 @@ function generateProject(destinationRoot, newProjectName, options) {
     });
   }
   if (!options['skip-jest']) {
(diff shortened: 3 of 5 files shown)


hramos:
Thanks for working on this! Since the PR is against 0.58-stable, I can't land it, but @grabbou or anyone else with write access who is working on the 0.58 release can push these changes to the 0.58-stable branch directly. Let me know if I can help with that.

grabbou:
Awesome! Thank you so much for working on it! Shall we also make sure this works on master? I think the only difference is that there's no templates/HelloWorld but template.

vovkasm:
@grabbou Yes, I'll check and make PRs needed for land this changes to master. I think that even changes to react-native-cli will not be needed, because:
1. setup babel-core@7.0.0-bridge.0 as dependency should not be needed because master currently at jest@24
2. modification of generator (https://github.com/facebook/react-native/pull/23128/files#diff-e1bddafdbb592817072925c6b7aabcd2R130) should not be needed because babel.config.js probably can live in template with its original name. Because by definition babel.config.js only processed at project root.

PR #23160, opened 2019-01-25
fix packager start in 0.58

dulmandakh:
This PR fixes packager during **react-native run-ios** and **react-native run-android**.

On reddit and discord many developers say that **react-native run-ios** failing on 0.58. But I found that **react-native start** works just fine. So did some investigation and found that if check is false when **react-native run-ios** but true when **react-native start**.
https://github.com/facebook/react-native/blob/57ad19758db60992a4d8ab55219d86eb892ac2b0/local-cli/cli.js#L23

It checks if script is run from shell, then starts the packager. Also found that **react-native start** calls this **react-natvie/local-cli/cli.js** directly and check is true. And dug little further to find that **packager.sh** is calling **react-native/cli.js** which imports **react-native/local-cli/cli.js**, therefore check is false.

### DIFF
--- a/scripts/packager.sh
+++ b/scripts/packager.sh
@@ -9,4 +9,4 @@ THIS_DIR=$(cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOUR
 # shellcheck source=/dev/null
 source "${THIS_DIR}/.packager.env"
 cd "$THIS_DIR/.." || exit
-node "./cli.js" start "$@"
+node "./local-cli/cli.js" start "$@"


grabbou:
I think this was affected by recent change by @cpojer from this commit https://github.com/facebook/react-native/commit/63a6bb7637b165f21bdddd2ebe818756dc8f4888#diff-2cce40143051e25f811b56c79d619bf5

screenshot 2019-01-25 at 16 19 01

Shall we simply revert it then?

grabbou:
It's interesting this is not present on 0.57, even though there are no changes to the files as far as I am aware.

cpojer:
I don't really get why this was failing? The top level cli.js is the same as this one :O

dulmandakh:
local-cli has a check that if it's run from cli or imported as a module, and it'll run it's called from cli. top level cli.js requires it, so the check is failed and not run.

```js
if (require.main === module) {
```

cpojer:
Ah, got it. Sorry I didn't realize it was used that way.

cpojer:
I'd like to find out why we didn't discover this during testing. Since this is basic functionality, we should ensure it never breaks. cc @grabbou

Issue #23108, opened 2019-01-23
[CI] Test failure - Various tests failing on master


hramos:
Tests help us prevent regressions, and without a clear signal, maintainers are unable to properly review pull requests. The following jobs are failing on `master`, and will need to be fixed before any new contributions can be considered:

- [x] `test_detox_end_to_end`
- [x] `test_android` (potential fix in https://github.com/facebook/react-native/pull/23106)
- [ ] `test_ios` (several failures fixed in https://github.com/facebook/react-native/commit/27617be9bb2083a58a56c1d9d20a45e9af069f42, one remaining failure)

## test_detox_end_to_end

[The job is failing with:](https://circleci.com/gh/facebook/react-native/68100)

```
Cannot find module 'react-native/jest/hasteImpl'

Error: Cannot find module 'react-native/jest/hasteImpl'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.resolve (internal/module.js:18:19)
    at load (/Users/distiller/react-native/node_modules/@react-native-community/cli/build/util/loadMetroConfig.js:89:51)
    at <anonymous>
```

## test_ios

Failing tests:

- [X] [RNTesterSnapshotTests testLayoutExample](https://circleci.com/gh/facebook/react-native/68366)
- [X] [RNTesterSnapshotTests testScrollViewExample](https://circleci.com/gh/facebook/react-native/68366)
- [X] [RNTesterSnapshotTests testSliderExample](https://circleci.com/gh/facebook/react-native/68366)
- [X] [RNTesterSnapshotTests testSwitchExample](https://circleci.com/gh/facebook/react-native/68366)
- [X] [RNTesterSnapshotTests testTextExample](https://circleci.com/gh/facebook/react-native/68366)
- [X] [RNTesterSnapshotTests testViewExample](https://circleci.com/gh/facebook/react-native/68366)
- [ ] [RNTesterSnapshotTests testTabBarExample](https://circleci.com/gh/facebook/react-native/68426)

dulmandakh:
Android CI is green https://github.com/facebook/react-native/pull/23109

guhungry:
For test_ios is it related to ios simulator 12 since `RNTesterSnapshotTests`'s `_runner` have only ios 10 and 11 condition

guhungry:
Somehow layout of snapshot is changed i'm uploading image for comparison

guhungry:
iOS CI is also green #23152

hramos:
[...]

PR #23161, opened 2019-01-25
Update testTabBarExample snapshot

hramos:
Related to https://github.com/facebook/react-native/issues/23108, addressing the `test_ios` failure. I ran the SSTs in record mode on an iPhone 6s running 12.1.

Test Plan:

Verify Circle CI succeeds when running the iOS test suite.

Changelog:

[iOS][Changed] - Update TabBar snapshot and fix failing SST.


guhungry:
I think may be it related to animation in testTabBarExample 

image

hramos:
The test was removed.

PR #23162, opened 2019-01-25
Label issues that use the Bug Report template

hramos:
This will allow the bot to more easily target issues that circumvent the bug report template.


hramos:
Landing as this is a trivial GitHub config with no code changes.

Issue #18886, opened 2018-04-16
Babel error when running react-native-git-upgrade


Nickersoft:
- [X] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [x] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

## Environment

Environment:
  OS:  macOS High Sierra 10.13.4
  Node:  9.11.1
  Yarn:  1.5.1
  npm:  5.6.0
  Watchman:  4.9.0
  Xcode:  Xcode 9.3 Build version 9E145
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: ^16.3.1 => 16.3.1
  react-native: 0.54.2 => 0.54.2

## Steps to Reproduce

I imagine this error occurs with all projects that utilize a 7.x beta version of Babel
 
## Expected Behavior

I'm trying to upgrade from React Native 0.53.x to 0.55 via `react-native-git-upgrade`

## Actual Behavior

I'm receiving the following error directly after the "Generate old version template" step:

```
Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.0". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. 
Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel. (While processing preset: "/Users/nickersoft/rnapp/node_modules/@babel/preset-flow/lib/index.js")
```

I'm currently using `@babel/core` (beta.44) for my project. My hunch is that `react-native-git-upgrade` is loading Babel via its own 6.x dependencies and calling `babel-register`, but once it hits the plugins and code I'm using it errors, seeing those expect 7.x. I've updated my code to use `@babel/core` and would prefer to not downgrade, but I'm curious if that's the only option. I tried the beta release of `react-native-git-upgrade`, thinking it might use Babel beta, but it doesn't. Any help would be appreciated. Thanks.

Nickersoft:
@react-native-bot I would, but this issue revolves around upgrading to v0.55... also, why did I respond to a bot?

hramos:
The next release should use Babel 7 - the work has already begun on master - hopefully that should help with the upgrade.

Strate:
Still actual while trying upgrade from 0.56 to 0.57. react-native-git-upgrade still uses babel@6 @hramos

[...]

PR #23167, opened 2019-01-26
Make upgrade cli compatible to babel7.

hyochan:
Old babel package makes recent `react-native` project with babel7 fails running `react-native-git-upgrade`. Refer to #18886.

Changelog:
----------

[GENERAL] [FIXED] - Fix `react-native-git-upgrade` to be compatible with `babel7` project.

Test Plan:
----------

1. Create a new `react-native` project above `0.56` which uses `babel7`.
2. Run `react-native-git-upgrade` will result failed saying `babel6` version is incompatible.

### DIFF
--- a/react-native-git-upgrade/package.json
+++ b/react-native-git-upgrade/package.json
@@ -15,7 +15,8 @@
     "url": "https://github.com/facebook/react-native.git"
   },
   "dependencies": {
-    "babel-core": "^6.18.0",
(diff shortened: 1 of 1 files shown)


cpojer:
Thank you for your pull request. I'm actually hoping on killing this module altogether soon in an effort to make upgrading less confusing, so I'm not gonna ship this change.

hyochan:
Yeah sounds better.

Issue #23176, opened 2019-01-27
bug with react-native link in android and windows


misaku:
## Environment

 React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (8) x64 Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
      Memory: 2.66 GB / 15.89 GB
    Binaries:
      Yarn: 1.13.0 - ~\AppData\Roaming\npm\yarn.CMD
      npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
      Watchman: 4.9.4 - D:\servers\watchman\watchman.EXE
    IDEs:
      Android Studio: Version  3.3.0.0 AI-182.5107.16.33.5199772
## Description

When run react-native link in windows and android
the url use \ and not /
for example:
```project(':react-native-gesture-handler').projectDir = new File(rootProject.projectDir, '..\node_modules\react-native-gesture-handler\android')```
and to fix to:
```project(':react-native-gesture-handler').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-gesture-handler/android')```
## Reproducible Demo

react-native link in windows

kelset:
That's a good idea, but maybe it's worth working directly with https://github.com/react-native-community/react-native-cli which is the new place where the CLI will be hosted in the future 👍

alainib:
same error for me

PR #23177, opened 2019-01-27
Fix for Windows - Backslashes is the escape

misaku:
Fix for Windows Backslashes is the escape character and will result in an invalid path in settings.gradle
  [issue:23176]( https://github.com/facebook/react-native/issues/23176)
  i found this fix in past commit (0.57) and i submit again.

Changelog:
----------

[Android] [Fixed] - Fix for Windows Backslashes is the escape character and will result in an invalid path in settings.gradle

Test Plan:
----------

in windows run react-native link and start error,  Backslashes is the escape character and will result in an invalid path in settings.gradle

### DIFF
--- a/local-cli/link/android/patches/makeSettingsPatch.js
+++ b/local-cli/link/android/patches/makeSettingsPatch.js
@@ -10,6 +10,8 @@
 const path = require('path');
 const normalizeProjectName = require('./normalizeProjectName');
 
+const isWin = process.platform === 'win32';
+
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks for sending this Pull Request. We actually removed the `local-cli` folder from this repository. It is now part of `react-native-cli` here: https://github.com/react-native-community/react-native-cli

Please submit your Pull Request there and it will make it into the latest version of React Native.

PR #23181, opened 2019-01-28
[fix] Fixing overscrolling issue on Virtualized List

dustinsavery:
Changelog:
----------
- [iOS][fixed] Fixed overscroll behavior on iOS virtualized lists

Test Plan:
----------
- Create a section list (for iOS)
- Attempt to scroll to an item towards the bottom

**Expected:** List scrolls to the item to the correct position without overscrolling the list
**Actual:** List is over scrolling to put the item in the proper position. This causes future scrolling to jump.

**Before:**
before

**After:**
after


cpojer:
It was pointed out to me that this was not the correct fix for this issue. You should be able to use `overScrollMode="never"` to fix this problem instead.

PR #23187, opened 2019-01-28
Update Pull Request and Issue templates.

cpojer:
As a maintainer, I noticed that the description for a PR is often missing, which makes it hard to understand why a change is made. Because some of the text isn't using HTML comments, the same text explaining how to fill out the changelog and test plan pops up and isn't removed by people. This PR fixes both of these issues.

I also made another small update to the main issue template to urge people to use one of the other templates, with an explanation that we will close poorly formatted issues.

Changelog:
----------

[General] [Changed] - Updated GitHub templates

Test Plan:
----------
yolo

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,3 +1,4 @@
-Please use one of the provided templates:
+👉 Please follow one of these issue templates:
 - https://github.com/facebook/react-native/issues/new/choose
 
+Note: to keep the backlog clean and actionable, issues may be immediately closed if they do not follow one of the above issue templates.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,16 +1,15 @@
-Thank you for sending the PR! We appreciate you spending the time to work on these changes.
-Help us understand your motivation by explaining why you decided to make this change:
+<!-- Thanks for submitting a pull request! We appreciate you spending the time to work on these changes. Please provide enough information so that others can review your pull request. The three fields below are mandatory. -->
 
+## Summary
 
-Changelog:
-----------
+<!-- Explain the **motivation** for making this change. What existing problem does the pull request solve? -->
 
-Help reviewers and the release process by writing your own changelog entry. See http://facebook.github.io/react-native/docs/contributing#changelog for an example.
+## Changelog
+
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,3 @@
-Please use one of the provided templates:
-- https://github.com/facebook/react-native/issues/new/choose
+## 👉 [Please follow one of these issue templates](https://github.com/facebook/react-native/issues/new/choose) 👈
```
hramos: I've been avoiding using markdown in these templates, as they are not rendered by default and actually make it harder to copy the URL. It's fairly safe for URLs, as GitHub will still render them as links should the user hit "Preview".

For reference, this is what a user sees when opening a blank template-less issue on master right now:
screen shot 2019-01-28 at 12 25 21 pm

This is how the proposed change would look like:
screen shot 2019-01-28 at 12 25 33 pm

This is what the current template that does not use markdown for the link looks like if the user hits "Preview":
screen shot 2019-01-28 at 12 25 42 pm

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,15 @@
...
+## Changelog
 
-[CATEGORY] [TYPE] - Message
+<!-- Help reviewers and the release process by writing your own changelog entry. See http://facebook.github.io/react-native/docs/contributing#changelog for an example. -->
 
+[CATEGORY] [TYPE] - Message
 
-Test Plan:
-----------
+## Test Plan
```
hramos: Don't drop the colon after "plan", to ensure Phabricator pulls this into the Test Plan field.

cpojer:
Created D13859925 and D13859922 to deal with both the colons for test plans and stripping comments. Let's see if they land internally first, and then ship this.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,15 @@
-Thank you for sending the PR! We appreciate you spending the time to work on these changes.
-Help us understand your motivation by explaining why you decided to make this change:
+<!-- Thanks for submitting a pull request! We appreciate you spending the time to work on these changes. Please provide enough information so that others can review your pull request. The three fields below are mandatory. -->
 
+## Summary
```
rickhanlonii: LOVE the reordering and using markdown headers here

PR #7304, opened 2016-04-29
iOS: Enable views to be nested within <Text>

rigdern:
Previously, only Text and Image could be nested within Text. Now, any
view can be nested within Text. One restriction of this feature is
that developers must give inline views a width and a height via
the style prop.

Previously, inline Images were supported by using iOS's built-in support
for rendering images with an NSAttributedString via NSTextAttachment.
However, NSAttributedString doesn't support rendering arbitrary views.

This change adds support for nesting views within Text by creating one
NSTextAttachment per inline view. The NSTextAttachments act as placeholders.
They are set to be the size of the corresponding view. After the text is
laid out, we query the text system to find out where it has positioned each
NSTextAttachment. We then position the views to be at those locations.

This commit also contains a change in `RCTShadowText.m`
`_setParagraphStyleOnAttributedString:heightOfTallestSubview:`. It now only sets
`lineHeight`, `textAlign`, and `writingDirection` when they've actually changed. This
avoids unnecessarily dirtying the shadow tree which is important in some inline
view scenarios.

Here's a screen shot demoing an inline view and an inline image:

image

Adam Comella
Microsoft Corp.

PR #23195, opened 2019-01-28
Android: Enable views to be nested within <Text>

rigdern:
Potential breaking change: The signature of ReactShadowNode's onBeforeLayout method was changed
  - Before: public void onBeforeLayout()
  - After:  public void onBeforeLayout(NativeViewHierarchyOptimizer nativeViewHierarchyOptimizer)

Implements same feature as this iOS PR: https://github.com/facebook/react-native/pull/7304

Previously, only Text and Image could be nested within Text. Now, any view can be nested within Text. One restriction of this feature is that developers must give inline views a width and a height via the style prop.

Previously, inline Images were supported via FrescoBasedReactTextInlineImageSpan. To get support for nesting views within Text, we create one special kind of span per inline view. This span is called TextInlineViewPlaceholderSpan. It is the same size as the inline view. Its job is just to occupy space -- it doesn't render any visual. After the text is rendered, we query the Android Layout object associated with the TextView to find out where it has positioned each TextInlineViewPlaceholderSpan. We then position the views to be at those locations.

One tricky aspect of the implementation is that the Text component needs to be able to render native children (the inline views) but the Android TextView cannot have children. This is solved by having the native parent of the ReactTextView also host the inline views. Implementation-wise, this was accomplished by extending the NativeViewHierarchyOptimizer to handle this case. The optimizer now handles these cases:
  - Node is not in the native tree. An ancestor must host its children.
  - Node is in the native tree and it can host its own children.
  - (new) Node is in the native tree but it cannot host its own children. An ancestor must host both this node and its children.

I added the `onInlineViewLayout` event which is useful for writing tests for verifying that the inline views are positioned properly.

Limitation: Clipping
----------

If Text's height/width is small such that an inline view doesn't completely fit, the inline view may still be fully visible due to hoisting (the inline view isn't actually parented to the Text which has the limited size. It is parented to an ancestor which may have a different clipping rectangle.). Prior to this change, layout-only views had a similar limitation.

Test Plan:
----------

Created a test app which tested a variety of cases including:
  - An inline view that is smaller than the line height
  - An inline view that is bigger than the line height
  - An inline view that is in text that wraps to multiple lines
  - An inline view that is in text that gets truncated and the inline view occurs past the truncation point.
  - An inline view that is nested under 1 and 2 layers of Text

I considered each of the above test cases under these variations:
  - Type of inline element: image or view
  - Type of hosting text component: Text or TextInput

All of the inline image cases behave identically before and after my change. The inline view behaviors are similar to the inline image behaviors. Note that TextInput doesn't support inline views (only inline images).

Lastly, we've been using a change like this in Skype for the past couple of years.

Changelog:
----------

[Android] [Added] - Enable views to be nested within Text

Adam Comella
Microsoft Corp.


matthargett:
Do snapshot and screenshot tests needs to be updated as well?

hramos:
Can you fix the copyright header on the new files? The tests are failing due to the old license being used. It should be something like:

```
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
```

RNTester/js/TextExample.android.js
```diff
@@ -494,9 +494,13 @@ class TextExample extends React.Component<{}> {
             This text will have a orange highlight on selection.
           </Text>
         </RNTesterBlock>
-        <RNTesterBlock title="Inline images">
+        <RNTesterBlock title="Inline views">
```
TheSavior: If inline views are now supported on both platforms we should consolidate the example out of this .android.js file so that the examples are used for both ios and android. 

Also, your PR summary lists some potential edge cases around clipping and different sized views. Can you add examples for not just the happy path but for all of those edge cases too? That way we can track how they change over time and ensure that they have consistent behavior across iOS and Android.

shergin:
Looks great to me! 🎉
Let's import that and see what internal tests think about that.

ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactShadowNodeImpl.java
```diff
@@ -164,6 +177,10 @@ public final boolean hasUnseenUpdates() {
   public void dirty() {
     if (!isVirtual()) {
       mYogaNode.dirty();
+    } else if (getParent() != null) {
```
shergin: (This is great and should be enough for now, but... ) I believe this will not work in all cases.
Let's imagine that the deeply nested view inside interleaving Text and View hierarchy was changed. This view is not virtual, so it will not call `getParent().dirty()`. As a result of calling `dirty`, Yoga will propagate the dirty flag to the closes root node, and that's it; the propagation will stop on the nearest inline View.

To fix this we have to additionally implement `onDirty` handler in Yoga that will dirty the semi-virtual (virtual outside, non-virtual inside) anchor View node (the inline view node) that will propagate this dirty call to ancestors until the first non-virtual one calls to Yoga again.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextView.java
```diff
@@ -74,6 +261,9 @@ public void setText(ReactTextUpdate update) {
         setBreakStrategy(update.getTextBreakStrategy());
       }
     }
+
+    // Ensure onLayout is called so the inline views can be repositioned.
```
shergin: As I said previously, I don't think we need this line, but it certainly doesn't hurt.

shergin:
@rigdern Did you fix the copyright header issue?

rigdern:
@shergin thanks for the ping.

I pushed a commit to fix the copyright headers. It looks like there are [some files in master with the wrong header](https://github.com/facebook/react-native/search?utf8=%E2%9C%93&q=%22Copyright+%28c%29+2015-present%2C+Facebook%2C+Inc.%22&type=). I didn't fix those because I want to avoid touching more files than necessary in this commit.

I want to think more about [Eli's comment](https://github.com/facebook/react-native/pull/23195#discussion_r254417878) and [your comment](https://github.com/facebook/react-native/pull/23195#discussion_r255268607) but it'll take more time than I have this morning. For your comment, I want to test your theory and if it's correct, put a comment in the code about it.

hramos:
Thanks for the heads up, looks like some of those files sneaked in after the header change. I have a [diff](https://our.internmc.facebook.com/intern/diff/D14091706/) out to fix these.

RNTester/js/TextExample.android.js
rigdern: Thanks for the suggestions.

I removed the example I added to RNTester for now because it won't work until we make a change to the React repo. Specifically, we have to remove [this assert from `ReactNativeRenderer-dev.js`](https://github.com/facebook/react-native/blob/536045276cd381aa37f72d9a5afa8ff68b40d575/Libraries/Renderer/oss/ReactNativeRenderer-dev.js#L3965-L3968).

I created a [branch with several inline view examples](https://github.com/rigdern/react-native/commit/ae11179a492283a7ab7bb7ef2b49e399eb8717c8) which I'll send as a PR after the prerequisite PRs have been merged. I shared code between iOS and Android where possible.

Unrelated to inline views, we should consider factoring out some of the common examples between `TextExample.ios.js` and `TextExample.android.js` into a shared file.

ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactShadowNodeImpl.java
rigdern: You were right. I was able to craft cases where inline images and views did not get rerendered. You can see them in the `ChangeImageSize`, `ChangeViewSize`, and `ChangeInnerViewSize` functions in `TextInlineView.js` in [my branch](https://github.com/rigdern/react-native/commit/ae11179a492283a7ab7bb7ef2b49e399eb8717c8#diff-f8c2073073da018fc6d2ed9a4e71f441). I'll be sending my branch as a separate PR so we can have these inline view examples in RNTester.

I added a TODO comment in `ReactShadowNodeImpl.java` to note this bug.

rigdern:
@shergin I pushed some commits to address the remaining feedback. Is there anything else blocking this PR?

JosephDunivan:
Not sure where this is at but there is another reason to be able to nest views within` <Text></Text>`.  

If you have a text component and a touchable opacity component nested within a view with flex-wrap set to `flexWrap: 'wrap'` and `flexDirection: 'row'`,  when the text takes up the width of the screen on the first line, but doesn't on the second line, the TouchableOpacity component will be pushed to the next block, as the row will take up the rest of the available space on the second line because the first line spans the entire width of the component. 

If you were able to nest views or other components for that matter, such as touchable opacity, then touchable content could sit on the same line as as the text component.

rigdern:
@JosephDunivan Thanks for the feedback. Are you experiencing this on iOS or Android? Can you provide a minimal code snippet that reproduces this issue?

JosephDunivan:
@rigdern 
Yes, it's actually a feature of the `<Text>` component at the moment so it's reproducible 100% of the time, both on iOS and Android. In fact, any component that is used within a view after text that takes up two lines will be pushed to the next row. I've used TouchableOpacity to demonstrate the error here because it has practical applications. 

The issue is 100% reliant on `textSize` and the `width` of the Component. For this reason please use iOS simulator running iOS 11 or 12, on a iPhone X. 

### Code Snippet where "feature"(error) occurs 
``` react-native
/**
 *
 * @format
 * @flow
 * @lint-ignore-every XPLATJSCOPYRIGHT1
 */

import {
  Platform, 
  StyleSheet, 
  Text, 
  View, 
  StatusBar, 
  Image, 
  TouchableOpacity,
  ScrollView,
  Dimensions
} from 'react-native';

var screen_width = Dimensions.get('window').width - 40

type Props = {};
  render() {
    return (
      <View style={styles.background}>
      <StatusBar barStyle="light-content"/>
          <View style={styles.wrap_text}>
            <Text style={styles.text}>By creating a new account or signing in with Facebook, you are agreeing to our </Text>
            <TouchableOpacity>
              <Text style={styles.text_touchable}>Terms of Service </Text>
            </TouchableOpacity>
            <Text style={styles.text}>and </Text>
            <TouchableOpacity>
              <Text style={styles.text_touchable}>Privacy Policy.</Text>
            </TouchableOpacity>
          </View>
        </View>
      </View>
    );
  }
}

  background: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#182B42',
  },
  container: {
    marginTop: 80,
    marginBottom: 30,
    marginLeft: 20,
    marginRight: 20,
    width: screen_width,
  },
  wrap_text: {
      flexWrap: 'wrap',
      flexDirection: 'row',
      justifyContent: 'center',
      width: 350,
  },
  text: {
    lineHeight: 25,
    fontSize: 11,
    color: '#738395',
  },
  text_touchable: {
    color: '#FFFFFF',
    lineHeight: 25,
    fontSize: 11,
    textDecorationLine: 'underline',
  },
});
```

### Code Snippet workaround for styling issues. 
#### The goal is to get all elements on the same line. 
```react-native
/**
 *
 * @format
 * @flow
 * @lint-ignore-every XPLATJSCOPYRIGHT1
 */

import {
  Platform, 
  StyleSheet, 
  Text, 
  View, 
  StatusBar, 
  Image, 
  TouchableOpacity,
  ScrollView,
  Dimensions
} from 'react-native';

var screen_width = Dimensions.get('window').width - 40

type Props = {};
  render() {
    return (
      <View style={styles.background}>
      <StatusBar barStyle="light-content"/>
          <View style={styles.wrap_text}>
            <Text style={styles.text}>By creating a new account or signing in with Facebook, </Text>
            <Text style={styles.text}>you are agreeing to our </Text>
            <TouchableOpacity>
              <Text style={styles.text_touchable}>Terms of Service </Text>
            </TouchableOpacity>
            <Text style={styles.text}>and </Text>
            <TouchableOpacity>
              <Text style={styles.text_touchable}>Privacy Policy.</Text>
            </TouchableOpacity>
          </View>
        </View>
      </View>
    );
  }
}

  background: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#182B42',
  },
  container: {
    marginTop: 80,
    marginBottom: 30,
    marginLeft: 20,
    marginRight: 20,
    width: screen_width,
  },
  wrap_text: {
      flexWrap: 'wrap',
      flexDirection: 'row',
      justifyContent: 'center',
      width: 350,
  },
  text: {
    lineHeight: 25,
    fontSize: 11,
    color: '#738395',
  },
  text_touchable: {
    color: '#FFFFFF',
    lineHeight: 25,
    fontSize: 11,
    textDecorationLine: 'underline',
  },
});
```
### Expected functionality
```react-native
/**
 *
 * @format
 * @flow
 * @lint-ignore-every XPLATJSCOPYRIGHT1
 */

import {
  Platform, 
  StyleSheet, 
  Text, 
  View, 
  StatusBar, 
  Image, 
  TouchableOpacity,
  ScrollView,
  Dimensions
} from 'react-native';

var screen_width = Dimensions.get('window').width - 40

type Props = {};
  render() {
    return (
      <View style={styles.background}>
      <StatusBar barStyle="light-content"/>
          <View style={styles.wrap_text}>
            <Text style={styles.text}>By creating a new account or signing in with Facebook, 
            you are agreeing to our 
              <TouchableOpacity>
                <Text style={styles.text_touchable}>Terms of Service </Text>
              </TouchableOpacity> 
            and
              <TouchableOpacity>
                <Text style={styles.text_touchable}>Privacy Policy.</Text>
              </TouchableOpacity>
            </Text>
          </View>
        </View>
      </View>
    );
  }
}

  background: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#182B42',
  },
  container: {
    marginTop: 80,
    marginBottom: 30,
    marginLeft: 20,
    marginRight: 20,
    width: screen_width,
  },
  wrap_text: {
      flexWrap: 'wrap',
      flexDirection: 'row',
      justifyContent: 'center',
      width: 350,
  },
  text: {
    lineHeight: 25,
    fontSize: 11,
    color: '#738395',
  },
  text_touchable: {
    color: '#FFFFFF',
    lineHeight: 25,
    fontSize: 11,
    textDecorationLine: 'underline',
  },
});
```

rigdern:
@josephdunivan Thanks for the details. Have you considered using the Text component's onPress prop for your scenario? https://facebook.github.io/react-native/docs/text#onpress

Since this also affects iOS, this is outside of the scope of this PR. If you still want to discuss this, can you create a new issue with the details from your posts here and at-mention me? Also, screenshots of the expected and actual results tend to be useful for helping people understand an issue more quickly.

JosephDunivan:
@rigdern 
Yes I did consider that, however it is an axiom that clicked text provide feedback to an iOS user. onPress does not currently provide feedback, I suppose you could set up some very complicated state change for the opacity, but TouchableOpacity already provides this, so it makes sense developers should be able to use it. 

As far as screenshots I can provide links on the new issue.

I'll set up a new issue for this and tag you.

jpstrikesback:
@shergin @rigdern apologies if this is off base but I can't seem to test this because of a check in the renderers added in https://github.com/facebook/react/commit/c802d29bd16753aef15863db8fb3f338c0067c75 via:

https://github.com/facebook/react/pull/12766/files

Do the relevant Nesting errors being thrown from that ^^ need to be reverted as well?

rigdern:
@jpstrikesback I plan to do a followup PR to remove the `invariant` for nesting views inside of text. I didn't do it yet because we wanted to give this PR some time to bake before building on it in case we needed to revert it.

It has been a couple of weeks since the merge so I'll send a followup PR soon.

jpstrikesback:
Awesome, cheers @rigdern !

PR #23207, opened 2019-01-29
keep desiredAccuracy if observer is active

mo22:
If watchPosition was called with high accuracy and getCurrentPosition is called the CLLocationManager desiredAccuracy is reset to default low value.

This patch resets the accuracy to the accuracy of the active observer if no getCurrentPosition task is queued, and to default only if there is also no observer active.

Changelog:
----------

[iOS] [Fixed] - Geolocation: Keep the desired location accuracy when an observer is active

Test Plan:
----------

Verify that geolocation getCurrentPosition and watchPosition is still working, and that the returned accuracy of the GPS signal matches the highAccuracy setting.

### DIFF
--- a/Libraries/Geolocation/RCTLocationObserver.m
+++ b/Libraries/Geolocation/RCTLocationObserver.m
@@ -355,7 +355,10 @@ RCT_EXPORT_METHOD(getCurrentPosition:(RCTLocationOptions)options
 
   // Reset location accuracy if desiredAccuracy is changed.
   // Otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes.
-  if (ABS(_locationManager.desiredAccuracy - RCT_DEFAULT_LOCATION_ACCURACY) > 0.000001) {
+  if (_observingLocation && ABS(_locationManager.desiredAccuracy - _observerOptions.accuracy) > 0.000001) {
+    _locationManager.desiredAccuracy = _observerOptions.accuracy;
+  }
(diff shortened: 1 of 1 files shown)


omnikron:
@mo22 by total coincidence I've submitted a PR (#23208) at the exact same time touching the exact same code! Maybe you can explain to me why `_locationManager.desiredAccuracy = RCT_DEFAULT_LOCATION_ACCURACY;` is necessary at all? We can't understand why it would be necessary and it seems to completely break the `enableHighAccuracy` option.

cpojer:
@mo22 as @omnikron pointed out, we just landed a change touching the same files so you'll need to rebase this PR and also please add the clarifications that @omnikron asked for.

mo22:
> @mo22 by total coincidence I've submitted a PR (#23209) at the exact same time touching the exact same code! Maybe you can explain to me why
> 
> ```
> _locationManager.desiredAccuracy = RCT_DEFAULT_LOCATION_ACCURACY;
> ```
> is necessary at all? We can't understand why it would be, and it seems to completely break the `enableHighAccuracy` option.

And that even though that bug is there since years ;)

To be honest I kept the fallback to RCT_DEFAULT_LOCATION_ACCURACY because it was in the original code. As the LocationManager is active only if there is a queued location request or an active observer that case should never be called and could be safely removed.

cpojer:
@mo22 do you wanna rebase your diff on master and make the change from your last comment?

hramos:
We'll also need the CLA to be signed before we can merge this.

mo22:
Hi,
we ended up writing a custom geolocation module as we needed some other features (location beacons) as well.
The patch should still apply correctly, and I claim no kind of ownership for those couple of lines, so if anyone wants to, feel free.

Issue #7680, opened 2016-05-22
Geolocation API automatically reverts to 100m accuracy on iOS


rburgst:
The current implementation of the geolocation API always reverts back to an accuracy of 100m after receiving the first location update.

- [x] Code snippet

https://facebook.github.io/react-native/docs/geolocation.html
    
    componentDidMount() {
        this.startGpsWatch();
    }

    startGpsWatch() {
        let geoOptions :GeoOptions = {
            enableHighAccuracy: true,
            distanceFilter: 3,
            timeout: 20000,
            maximumAge: 4000,
            accuracy: 1,
        };
        this.watchID = navigator.geolocation.watchPosition(
            this.onLastPositionReceived.bind(this),
            this.onLastPositionError.bind(this),
            geoOptions
        );

    onLastPositionReceived(position) {
        console.info("accuracy", position.accuracy);
    }

- [x] What's the **version** of React Native you're using?

latest 0.26.1

- [x] Does this occur on iOS, Android or both?

only on iOS

- [x] Are you using Mac, Linux or Windows?

Mac

The issue seems to stem from this code https://github.com/facebook/react-native/blob/2310494f328de903a4bc00636b9a89ffd2b24f62/Libraries/Geolocation/RCTLocationObserver.m#L337

    - (void)locationManager:(CLLocationManager *)manager
       didUpdateLocations:(NSArray<CLLocation *> *)locations
    {
       ...
       
    // Reset location accuracy if desiredAccuracy is changed.
    // Otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes.
    if (ABS(_locationManager.desiredAccuracy - RCT_DEFAULT_LOCATION_ACCURACY) > 0.000001) {
        _locationManager.desiredAccuracy = RCT_DEFAULT_LOCATION_ACCURACY;
    }

Where `RCT_DEFAULT_LOCATION_ACCURACY ` is 100m.

The rationale for this code is somewhat unclear to me. The comment does not seem to fit to the implementation.

lacker:
The code you are linking to is inside `didFailWithError` so it looks like the accuracy is only resetting on an error. My suspicion is that in your project it is actually erroring a lot for some reason which is causing those resets. I'm going to close this issue since it's been a while but if this theory is incorrect feel free to reopen.

rburgst:
[...]

PR #23209, opened 2019-01-29
Don't reset desiredAccuracy on every update/error

omnikron:
See https://github.com/facebook/react-native/issues/7680.

On iOS the RCTLocationObserver delegate is overriding `desiredAccuracy` every time CLLocationManager calls `didUpdateLocations` or `didFailWitError`.  `desiredAccuracy` is reset to
`RCT_DEFAULT_LOCATION_ACCURACY` (100 meters) This effectively makes it impossible for a react-native app to use any location accuracy other than the default.

This commit simply removes the code which resets the desired accuracy, as there seems to be no rationale for doing so. The reset code was added as part of [a large general
change](https://github.com/facebook/react-native/commit/705a8e0144775ff695f26a1f6b41d0baa6bed683) so the original intention is unclear from the history. If somebody can explain it to me, I'm happy to rework this PR accordingly.

Changelog:
----------
[iOS] [Fixed] - Location Services accuracy constantly reset to default of 100 meters.

Test Plan:
----------

We've been using this in production at FATMAP for nearly a year. It works. I'm unclear on how to actually test this in code, so would appreciate any tips.

### DIFF
--- a/Libraries/Geolocation/RCTLocationObserver.m
+++ b/Libraries/Geolocation/RCTLocationObserver.m
@@ -353,11 +353,6 @@ RCT_EXPORT_METHOD(getCurrentPosition:(RCTLocationOptions)options
       [_locationManager stopUpdatingLocation];
   }
 
-  // Reset location accuracy if desiredAccuracy is changed.
-  // Otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes.
-  if (ABS(_locationManager.desiredAccuracy - RCT_DEFAULT_LOCATION_ACCURACY) > 0.000001) {
-    _locationManager.desiredAccuracy = RCT_DEFAULT_LOCATION_ACCURACY;
-  }
 }
 
 - (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *)error
@@ -389,11 +384,6 @@ RCT_EXPORT_METHOD(getCurrentPosition:(RCTLocationOptions)options
   }
   [_pendingRequests removeAllObjects];
 
-  // Reset location accuracy if desiredAccuracy is changed.
-  // Otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes.
-  if (ABS(_locationManager.desiredAccuracy - RCT_DEFAULT_LOCATION_ACCURACY) > 0.000001) {
-    _locationManager.desiredAccuracy = RCT_DEFAULT_LOCATION_ACCURACY;
-  }
 }
 
 static void checkLocationConfig()


omnikron:
Well, this is an unlikely coincidence – since I started working on this PR a few hours ago a very similar one has been opened (#23207). Happy to unify if it makes sense to do so, although I'm not sure if there is 100% overlap.

hramos:
Did you observe the issue mentioned in the comment, "otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes." in your fork?

omnikron:
@hramos there's quite a lot to unpick here, so bear with me! This issue is absolutely critical for us (our users are often in mountainous back country, so precise GPS user location is critical for safety as well as usability) which is why we've been maintaining our fork for 11 months.

> Did you observe the issue mentioned in the comment, "otherwise update accuracy will force triggering didUpdateLocations, watchPosition would keeping receiving location updates, even there's no location changes."

I dispute that this is a bug that needed fixing here, so much as an application issue. When setting `enableHighAccuracy` and starting `watchPosition` the device will indeed send location updates even if you don't move. This is because the GPS signal will vary/reflect and send minute changes to very high (centimeter) accuracy, even when the device is not moving. However, my take is that this should be handled in application code, not in location services themselves. If you don't need this kind of very high accuracy, then you should simply disable `enableHighAccuracy` and go back to the default of 100m accuracy, which is probably fine for most 'what's near here' services.

The comment you mentioned was added in [this commit](https://github.com/facebook/react-native/commit/aacd2d90a4a511f1e39e7473b888b6f9fb1b723e#diff-db9ecf8dd94d42455275a6070ea72f71) which just adds this check:
```
 if (ABS(_locationManager.desiredAccuracy - RCT_DEFAULT_LOCATION_ACCURACY) > 0.000001)
```

to me this check alone makes absolutely zero sense. "If `desiredAccuracy` is not default accuracy then set `desiredAccuracy` to default accuracy" – what? I feel like I must be missing something, but I can't figure it out.

However, the actual code that I dispute is added previously in [this commit](https://github.com/facebook/react-native/commit/705a8e0144775ff695f26a1f6b41d0baa6bed683#diff-ee9ef50d237033748491b0adefa0e0c1R283) where the comment is simply 'Reset location accuracy'. I don't understand this either – when requesting a high accuracy position from the device, the device sends the first position with high accuracy (which will often be cached from earlier), and then this code resets the accuracy to 100m again, so all future positional updates will be wildly inaccurate!

My interpretation of this is that it has been added in error. If I am missing something then I'd love to know what it is, because as it stands `enableHighAccuracy` simply doesn't work for iOS devices in React Native.

hramos:
Thanks for the explanation! I agree this should be handled in product code.

cpojer:
Retrying because we had a import tooling issue.

PR #23226, opened 2019-01-31
[iOS] Enhance image freshness check before stored into cache

zhongwuzw:
## Summary

Currently, before we store the image to cache, we only respect `Cache-Control`, actually, we also may need to check `Expires`、`Last-Modified`, refer to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#Freshness), and  [okhttp](https://github.com/square/okhttp/blob/568a91c44a118b2c2ba62d310a331582c567b24a/okhttp/src/main/java/okhttp3/internal/cache/CacheStrategy.java#L268) respect the `MDN`, so in iOS, we can also respect this.

## Changelog

[iOS] [Fixed] - Respect `MDN` cache strategy before cache the image.

## Test Plan

After `PR`, if image's response do not contains `Cache-Control`, it would also checks `Expires`、`Last-Modified` refers to  [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#Freshness).

### DIFF
--- a/Libraries/Image/RCTImageLoader.h
+++ b/Libraries/Image/RCTImageLoader.h
@@ -31,8 +31,7 @@ typedef dispatch_block_t RCTImageLoaderCancellationBlock;
                    size:(CGSize)size
                   scale:(CGFloat)scale
              resizeMode:(RCTResizeMode)resizeMode
-           responseDate:(NSString *)responseDate
-           cacheControl:(NSString *)cacheControl;
+               response:(NSURLResponse *)response;
 
 @end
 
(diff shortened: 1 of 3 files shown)


Libraries/Image/RCTImageLoader.h
```diff
@@ -31,8 +31,7 @@ typedef dispatch_block_t RCTImageLoaderCancellationBlock;
                    size:(CGSize)size
                   scale:(CGFloat)scale
              resizeMode:(RCTResizeMode)resizeMode
-           responseDate:(NSString *)responseDate
-           cacheControl:(NSString *)cacheControl;
```
zhongwuzw: Only pass `responseDate` or `cacheControl` may not sufficient, we can just pass `response` that we can get any response header we want.

cpojer:
Just landed this one again in https://github.com/facebook/react-native/commit/fb8ba3fe959fd2a0c4ef0025fd84f6deffd9d03b

Issue #22609, opened 2018-12-11
☂️ Resolve Warnings in Xcode


hramos:
Xcode shows multiple warnings when either `RNTester/RNTester/xcodeproj` or `template/ios/HelloWorld.xcodeproj/` (scaffold for new apps) is built. These warnings have accumulated over the years (https://github.com/facebook/react-native/issues/19628), and it is [one of the concerns with the most reactions](https://github.com/react-native-community/discussions-and-proposals/issues/64#issuecomment-444835761) in a recent survey of React Native users.

This issue will track all efforts to reduce the number of warnings in our Xcode projects.

If you send a pull request that removes a warning listed in this issue, please let us know so we can take a look at your proposed change.

# RNTester

## Targets

Xcode breaks down the warnings by target. 

- [x] double-conversion
- [x] jsi
- [x] third-party (folly, glog)
- [x] cxxreact
- [x] jsiexecutor
- [x] React
- [x] RCTNetwork
- [x] RCTImage
- [x] RNTester

Note that some of these warnings may originate in third party libraries. Further, these warnings may be resolved in newer versions of these libraries. As such, you may need to upgrade some of these libraries to resolve a warning. Careful testing will be needed in order to make sure no breaking changes are introduced as part of these upgrades.

## Detailed list of warnings

As of this writing, Xcode 10.1 presents warnings on the following files when building `RNTester`:

<details>
<summary>
Target: double-conversion
</summary>

```
third-party/double-conversion-1.1.6/src/double-conversion.cc:825:10: Declaration shadows a local variable
```
</details>

<details>
<summary>
Target: jsi
</summary>

```
Semantic Issue Group
ReactCommon/jsi/JSCRuntime.cpp:575:22: Unused parameter 'ctx'
ReactCommon/jsi/JSCRuntime.cpp:612:22: Unused parameter 'ctx'
third-party/folly-2018.10.22.00/folly/Memory.h:51:30: Possible misuse of comma operator here
ReactCommon/jsi/JSIDynamic.cpp:6:10: In file included from ReactCommon/jsi/JSIDynamic.cpp:6:
ReactCommon/jsi/JSIDynamic.h:8:10: In file included from ReactCommon/jsi/JSIDynamic.h:8:
third-party/folly-2018.10.22.00/folly/dynamic.h:67:10: In file included from third-party/folly-2018.10.22.00/folly/dynamic.h:67:
[...]

PR #23227, opened 2019-01-31
[iOS] Remove application:openURL:sourceApplication:annotation: of LinkingManager

zhongwuzw:
## Summary

Related to #22609 . `application:openURL:sourceApplication:annotation: ` already deprecated in `iOS9`, `react-native` only supports iOS which target greater or equal to `iOS9`, so we can just remove it totally.

## Changelog

[iOS] [Removed] - Remove application:openURL:sourceApplication:annotation: of LinkingManager

## Test Plan

On iOS system which greater or equal to `iOS9`, `application:openURL:sourceApplication:annotation: ` should not be called by system or user, use `application:openURL:options:` instead.

### DIFF
--- a/Libraries/LinkingIOS/RCTLinkingManager.h
+++ b/Libraries/LinkingIOS/RCTLinkingManager.h
@@ -16,11 +16,6 @@
             options:(nonnull NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options;
 
 + (BOOL)application:(nonnull UIApplication *)application
-              openURL:(nonnull NSURL *)URL
-    sourceApplication:(nullable NSString *)sourceApplication
-           annotation:(nonnull id)annotation;
-
-+ (BOOL)application:(nonnull UIApplication *)application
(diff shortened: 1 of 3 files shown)


cpojer:
Seems like we haven't upgraded internal callsites yet.

zhongwuzw:
@cpojer Hi, so seems we need to wait?

By the way, I see #23226 has been reverted, please see [1a8e3f](https://github.com/facebook/react-native/commit/c435ff382ea2e5ba4db069eb7ed166e2301a8e3f), any thought about that? 🤔

cpojer:
Yeah we'll need to wait for a bit it seems. For the other one, I'm gonna try to get that one back in.

cpojer:
As said before, it seems we cannot remove this deprecated method yet as it causes issues at FB. I'll close this PR for now and we can revisit it in the future.

zhongwuzw:
@cpojer 👌 .

PR #23230, opened 2019-01-31
[Android] Sometimes setupReactContextRunnable is run faster than maybeRecreateR…

RoJoHub:
## Summary

> sorry,my english not good

Related to  #13439  

Sometimes setupReactContextRunnable runs faster than maybeRecreateReactContextRunnable.

The code will proceed in the following order:

---

`setupReactContext(reactApplicationContext);`→ `moveReactContextToCurrentLifecycleState();`→
`moveToResumedLifecycleState(true);`

---

```java
private synchronized void moveToResumedLifecycleState(boolean force) {
        ReactContext currentContext = getCurrentReactContext();
        if (currentContext != null) {
            // we currently don't have an onCreate callback so we call onResume for both transitions
            if (force ||
                    mLifecycleState == LifecycleState.BEFORE_RESUME ||
                    mLifecycleState == LifecycleState.BEFORE_CREATE) {
                Log.d(lhtag, "--moveToResumedLifecycleState: mCurrentActivity" + mCurrentActivity);
                currentContext.onHostResume(mCurrentActivity);
                Log.d(lhtag, "--moveToResumedLifecycleState: mCurrentActivity" + mCurrentActivity);
            }
        }
        mLifecycleState = LifecycleState.RESUMED;
    }
```
At this time, the value of mCurrentActivity is equal to null.

So,It will run `currentContext.onHostResume(null);`

Finally,other module run `getCurrentActivity()`,this value will reture null.

## Changelog
[Android] [Fixed] - Changed the order in which the code is executed

## Test Plan
Everything should run as usual.And Fix this Bug #13439 .

### DIFF
--- a/Libraries/Image/RCTImageCache.m
+++ b/Libraries/Image/RCTImageCache.m
@@ -106,33 +106,49 @@ static NSString *RCTCacheKeyForImage(NSString *imageTag, CGSize size, CGFloat sc
                    size:(CGSize)size
                   scale:(CGFloat)scale
              resizeMode:(RCTResizeMode)resizeMode
-           responseDate:(NSString *)responseDate
-           cacheControl:(NSString *)cacheControl
+               response:(NSURLResponse *)response
 {
-  NSString *cacheKey = RCTCacheKeyForImage(url, size, scale, resizeMode);
-  BOOL shouldCache = YES;
-  NSDate *staleTime;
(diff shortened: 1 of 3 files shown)


matthargett:
From the internal FB review, copying here: ```This will regress TTI for initialization of RN, can you find another way to solve this issue without changing the order of execution?```

RoJoHub:
@matthargett  it's done.

PR #23231, opened 2019-01-31
Allow headless JS tasks to retry

JoshuaJamesOng:
## Summary
`setTimeout` inside a headless JS task does not always works; the function does not get invoked until the user starts an `Activity`. 

This was attempted to be used in the context of widgets. When the widget update or user interaction causes the process and React context to be created, the headless JS task may run before other app-specific JS initialisation logic has completed. If it's not possible to change the behaviour of the pre-requisites to be synchronous, then the headless JS task blocks such asynchronous JS work that it may depend on. A primitive solution is the use of `setTimeout` in order to wait for the pre-conditions to be met before continuing with the rest of the headless JS task. But as the function passed to `setTimeout` is not always called, the task will not run to completion.

This PR solves this scenario by allowing the task to be retried again with a delay. If the task returns a promise that resolves to a `{'timeout': number}` object, `AppRegistry.js` will not notify that the task has finished as per master, instead it will tell `HeadlessJsContext` to `startTask` again (cleaning up any posted `Runnable`s beforehand) via a `Handler` within the `HeadlessJsContext`.  

Documentation also updated here: https://github.com/facebook/react-native-website/pull/771

### AppRegistry.js
If the task provider does not return any data, or if the data it returns does not contain `timeout` as a number, then it behaves as `master`; notifies that the task has finished. If the response does contain `{timeout: number}`, then it will attempt to queue a retry. If that fails, then it will behaves as if the task provider returned no response i.e. behaves as `master` again. If the retry was successfully queued, then there is nothing to do as we do not want the `Service` to stop itself.

### HeadlessJsTaskSupportModule.java
Similar to notify start/finished, we simply check if the context is running, and if so, pass the request onto `HeadlessJsTaskContext`. The only difference here is that we return a `Promise`, so that `AppRegistry`, as above, knows whether the enqueuing failed and thus needs to perform the usual task clean-up.

### HeadlessJsTaskContext.java
Before retrying, we need to clean-up any timeout `Runnable`'s posted for the first attempt. Then we need to copy the task config so that if this retry (second attempt) also fails, then on the third attempt (second retry) we do not run into a consumed exception. This is also why in `startTask` we copy the config before putting it in the `Map`, so that the initial attempt does leave the config's in the map as consumed. Then we post a `Runnable` to call `startTask` on the main thread's `Handler`. We use the same `taskId` because the `Service` is keeping track of active task IDs in order to calculate whether it needs to `stopSelf`. This negates the need to inform the `Service` of a new task id and us having to remove the old one. 

## Changelog
[Android][added] - Allow headless JS tasks to return a promise that will cause the task to be retried again with the specified delay

## Test Plan
A fork containing this fix has been integrated and tested in a project like below: 
```
export const headlessJsTask = () => () =>
    async function updateWidget() {
        const isReady = ...;
        if (!isReady) {
            return new Promise(resolve => resolve({'timeout': 1000})); // 1000ms
        }

        // Do work that relies on pre-requisite
    };
```

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java
@@ -5,12 +5,6 @@
 
 package com.facebook.react.jstasks;
 
-import java.lang.ref.WeakReference;
-import java.util.Set;
-import java.util.WeakHashMap;
-import java.util.concurrent.CopyOnWriteArraySet;
-import java.util.concurrent.atomic.AtomicInteger;
-
 import android.os.Handler;
 import android.util.SparseArray;
 
@@ -19,6 +13,14 @@ import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.UiThreadUtil;
 import com.facebook.react.common.LifecycleState;
 import com.facebook.react.modules.appregistry.AppRegistry;
+
+import java.lang.ref.WeakReference;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Helper class for dealing with JS tasks. Handles per-ReactContext active task tracking, starting /
@@ -51,6 +53,7 @@ public class HeadlessJsTaskContext {
   private final AtomicInteger mLastTaskId = new AtomicInteger(0);
   private final Handler mHandler = new Handler();
   private final Set<Integer> mActiveTasks = new CopyOnWriteArraySet<>();
+  private final Map<Integer, HeadlessJsTaskConfig> mActiveTaskConfigs = new ConcurrentHashMap<>();
   private final SparseArray<Runnable> mTaskTimeouts = new SparseArray<>();
 
   private HeadlessJsTaskContext(ReactContext reactContext) {
@@ -85,6 +88,16 @@ public class HeadlessJsTaskContext {
    * @return a unique id representing this task instance.
    */
   public synchronized int startTask(final HeadlessJsTaskConfig taskConfig) {
+    final int taskId = mLastTaskId.incrementAndGet();
+    startTask(taskConfig, taskId);
+    return taskId;
+  }
+
+  /**
+   * Start a JS task the provided task id. Handles invoking {@link AppRegistry#startHeadlessTask}
+   * and notifying listeners.
+   */
+  private synchronized void startTask(final HeadlessJsTaskConfig taskConfig, int taskId) {
     UiThreadUtil.assertOnUiThread();
     ReactContext reactContext = Assertions.assertNotNull(
       mReactContext.get(),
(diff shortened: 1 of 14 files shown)


cpojer:
@JoshuaJamesOng did you see @matthargett's comment and could you address it?

cpojer:
@JoshuaJamesOng ping?

ejanzer:
@JoshuaJamesOng I think this makes sense, the only part that feels a little weird to me is returning the timeout from the JS task. What if we allow you to configure whether or not retries are enabled (along with the timeout in ms, # of retries, etc.) from the native side (in HeadlessJsTaskConfig) and then we just retry on any JS exception if so? Maybe something like:

```
export const headlessJsTask = () => () =>
    async function updateWidget() {
        const isReady = ...;
        if (!isReady) {
          throw new Error('Not ready');
        }
       // other stuff if it's ready
    };
```

and then something like this in getTaskConfig (not sure if this is an up-to-date example, pulled it from the docs):

```
  new HeadlessJsTaskConfig(
    "SomeTaskName",
    Arguments.fromBundle(extras),
    5000, // timeout
    false, // allowed in foreground
    5, // number of retries
    1000 // retry after ms
  );
```

What do you think? Would that do what you need?.

cpojer:
@JoshuaJamesOng hey! Did you see @ejanzer's reply to your PR? Would you be able to make the changes she proposed so we can merge this?

JoshuaJamesOng:
Have made the changes suggested. Note however that this does change the behaviour. Previously the author of the could decide when to retry. The new implementation retries on all errors. I wonder if we should make it so that the user still has control over which errors are retried? I'm worried we get stuck in a loop. Thoughts?

cpojer:
@JoshuaJamesOng I think that's a good idea. Could you add it?

Libraries/ReactNative/HeadlessJsTask.js
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+'use strict';
+
+export class HeadlessJsTaskError extends Error {}
```
ejanzer: Can you change the file name to HeadlessJsTaskError to match the class? Also, instead of putting it at the top level, can you put it /Utilities, maybe?

ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java
```diff
@@ -105,7 +118,44 @@ public synchronized int startTask(final HeadlessJsTaskConfig taskConfig) {
...
+            sourceTaskConfig.getData(),
+            sourceTaskConfig.getTimeout(),
+            sourceTaskConfig.isAllowedInForeground(),
+            retryPolicy.update()
+    );
+
+    final Runnable retryAttempt = new Runnable() {
+      @Override
+      public void run() {
+        startTask(taskConfig, taskId);
```
ejanzer: Do we reuse the taskId just so JS can continue to use the same id? Or to avoid adding entries to the active tasks map? (I think it makes sense to use the same taskId, just curious if there's another reason why it's needed)

ReactAndroid/src/main/java/com/facebook/react/modules/core/HeadlessJsTaskSupportModule.java
```diff
@@ -32,6 +33,22 @@ public String getName() {
...
+  public void notifyTaskRetry(int taskId, Promise promise) {
+    HeadlessJsTaskContext headlessJsTaskContext =
+      HeadlessJsTaskContext.getInstance(getReactApplicationContext());
+    if (headlessJsTaskContext.isTaskRunning(taskId)) {
+      final boolean retryPosted = headlessJsTaskContext.retryTask(taskId);
+      promise.resolve(retryPosted);
+    } else {
+      FLog.w(
+        HeadlessJsTaskSupportModule.class,
+        "Tried to try non-active task with id %d. Did it time out?",
```
ejanzer: nit: "tried to retry"

cpojer:
Hey @JoshuaJamesOng! Do you mind addressing @ejanzer's last small comments? Then I think we can land it finally :)

PR #23245, opened 2019-02-01
Expose isLocalUserInfoKey to keyboard event notifications

nossbigg:
## Summary

Given two apps loaded side-by-side and when a `Keyboard` event is triggered, there is no way to ascertain which app triggered the keyboard event. This ambiguity can arise in slide over/split view scenarios.

This pull request exposes the `isLocalUserInfoKey` property of the native `UIKeyboard` iOS events to the `Keyboard` event listener; this property will return `true` for the app that triggered the keyboard event.

(Also, I threw in a couple of Keyboard.js tests just for fun 😅)

## Changelog

[iOS][Added] - Expose isLocalUserInfoKey to keyboard event notifications

## Test Plan

1. Load two apps side-by-side, with the app on the left side subscribing to the keyboard events (and logging out the events as they happen)
1. Trigger a keyboard to appear with the left app. The logged keyboard event will contain the `isEventFromThisApp` property which will be true.
1. Dismiss the keyboard
1. Trigger a keyboard to appear with the right app. The left app will still log the keyboard event, but the event's `isEventFromThisApp` property will be false (because the left app didn't trigger the keyboard event)

### DIFF
--- a/Libraries/Components/Keyboard/Keyboard.js
+++ b/Libraries/Components/Keyboard/Keyboard.js
@@ -37,6 +37,7 @@ export type KeyboardEvent = $ReadOnly<{|
   easing?: string,
   endCoordinates: ScreenRect,
   startCoordinates?: ScreenRect,
+  isEventFromThisApp?: boolean,
 |}>;
 
 type KeyboardEventListener = (e: KeyboardEvent) => void;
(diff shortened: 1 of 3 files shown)


hramos:
Thanks for increasing our test coverage!

PR #23266, opened 2019-02-02
ReactAppTestActivity extends ReactActivity, instead of FragmentActivity

dulmandakh:
## Summary

Since https://github.com/facebook/react-native/commit/dda2b82a0a49da52b43b50db5a2bda50a216c09b, we have only ReactActivity. But we are not using it for tests. This PR changes ReactAppTestActivity to extend ReactActivity and used for tests.

## Changelog

[Android] [Changed] - ReactAppTestActivity extends ReactActivity

## Test Plan
CI is green.


ReactAndroid/src/androidTest/java/com/facebook/react/testing/ReactAppTestActivity.java
```diff
@@ -34,18 +32,15 @@
...
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 
-public class ReactAppTestActivity extends FragmentActivity
+public class ReactAppTestActivity extends ReactActivity
     implements DefaultHardwareBackBtnHandler, PermissionAwareActivity {
```
hey99xx:
```
implements DefaultHardwareBackBtnHandler, PermissionAwareActivity
```

shouldnt be needed since `ReactActivity` implements them already:

```
public abstract class ReactActivity extends FragmentActivity
    implements DefaultHardwareBackBtnHandler, PermissionAwareActivity {
```

`invokeDefaultOnBackPressed` method can also be removed as the one here will be the same as in the parent class.

dulmandakh: @hey99xx thanks, removed

cpojer:
Thank you!

cpojer:
I'm getting:

```
java.lang.RuntimeException: Unable to resume activity {com.facebook.react.screenshots.shared.Facebook/com.facebook.react.testing.ReactAppTestActivity}: java.lang.ClassCastException: com.facebook.testing.instrumentation.TestApplication cannot be cast to com.facebook.react.ReactApplication
at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3400)
at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3440)
```

do you know how I can fix this?

dulmandakh:
please make sure that TestApplication implements ReactApplication, it might be the cause.

dulmandakh:
@cpojer please see https://github.com/facebook/react-native/blob/a82f6e164b51cf24127b6fcaee39a039789bd91a/RNTester/android/app/src/main/java/com/facebook/react/uiapp/RNTesterApplication.java

cpojer:
@dulmandakh I looked into this today. We have internal infrastructure depending on this code but I do not understand why it is failing when I import this PR given that it still matches the FragmentActivity API in the end. I cannot change the class that it fails on because that one is shared with all of FB and not just RN. I'll have to investigate some more.

ReactAndroid/src/androidTest/java/com/facebook/react/testing/ReactAppTestActivity.java
```diff
@@ -273,10 +264,6 @@ public void run() {
     }
   }
 
-  private ReactInstanceManager getReactInstanceManager() {
-    return mReactInstanceManager;
```
cpojer: @dulmandakh I'm not sure this is correct. We store a `mReactInstanceManager` in this class but with your change it will instead go to `ReactActivityDelegate` and call `((ReactApplication) getPlainActivity().getApplication()).getReactNativeHost()` (which is where it fails).

Is there a way we can retain the previous behavior here? It's also a bit odd to me to rely on different ways of getting the instance manager both within this class and through external usage when the getter is being called.

I'm not an expert here so I'd appreciate any help so I can better understand what is going on so I can make this work and land it.

dulmandakh:
@cpojer I'll look into it thoroughly, maybe tomorrow.

dulmandakh:
let's drop the idea for a while, it seems that it requires some behavior changes in ReactActivity.

PR #23277, opened 2019-02-03
remove unnecessary Android version checks

dulmandakh:
## Summary

React Native's minSdkVersion is 16, or we support Android versions 16 (Jelly Bean) and above. But in the code we have many checks if Android is Jelly Bean or newer, which are unnecessary. This PR removes unnecessary Android version checks, also uses Android version names instead of numbers.

## Changelog

[Android] [Changes] - remove unnecessary Android version checks

## Test Plan

CI is green and works as usual.


dulmandakh:
Also found that if we switch to AppCompat widgets, we can remove even more version check but provide compatibility.

PR #23278, opened 2019-02-03
ReactActivity extends AppCompatActivity

dulmandakh:
## Summary

Google recommends to extend AppCompatActivity. and This PR changes ReactActivity to extend AppCompatActivity

## Changelog

[Android] [Changed] - ReactActivity extends AppCompatActivity

## Test Plan

CI is green and everything should work as usual.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/BUCK
@@ -4,8 +4,8 @@ rn_android_library(
     name = "react",
     srcs = glob(["*.java"]),
     provided_deps = [
+        react_native_dep("third-party/android/support/v7/appcompat-orig:appcompat"),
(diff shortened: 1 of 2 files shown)


dulmandakh:
@mdvacca I just realized that I had learned something while working on landing Android Support Library 28.0.0. Please import it and see if it works. Thank you very much.

mdvacca:
There is an internal build error because of lack of dependency but buck works fine, I will update that dependency and then land this

mdvacca:
This PR produced runtime crashes with errors:
`
You need to use a Theme.AppCompat theme (or descendant) with this activity.
`

We should flag this PR as a backward in-compatible change for next release
CC @cpojer, @hramos

dulmandakh:
@mdvacca I couldn't find non AppCompat theme reference in RN repo. But I agree that we should add notes about it.

PR #23289, opened 2019-02-05
bump flow-bin to 0.92.1

gengjiawen:
## Summary

Bump flow version so we can fix the bug on windows.

## Changelog

[GENERAL] [Fixed] - bump flow version
## Test Plan

pass current ci

### DIFF
--- a/package.json
+++ b/package.json
@@ -224,7 +224,7 @@
     "eslint-plugin-prettier": "2.6.0",
     "eslint-plugin-react": "7.8.2",
     "eslint-plugin-react-native": "3.5.0",
-    "flow-bin": "^0.92.0",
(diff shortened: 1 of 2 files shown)


gengjiawen:
cc @cpojer @TheSavior @empyrical @gabelevi

TheSavior:
I don't think we do this ourselves as Flow has to be on the same version across Facebook. Typically the Flow team handles these rollouts. I imagine @gabelevi will do so.

cpojer:
As @TheSavior said, we cannot upgrade Flow through a pull request as we need to update our entire codebase in one go. The Flow team takes care of this, and I'd expect them to land this upgrade soon. Watch out for the commit landing in open source :)

PR #23305, opened 2019-02-06
Backwards compatibility in findMatchingSimulator version input

osunnarvik:
## Summary

handling dot-version input in findMatchingSimulator also when using newest xcrun to list devices. Format of xcrun output has changed

When parsing simulator names in order to run on iOS simulator, the version provided by xcrun now is separated by a dash instead of a dot. Therefore when input to run-ios --simulator is specified with an iOS version (in dot-format), the simulator will not be found. 

This started happening for me after I installed the latest XCode (beta). I am currently using XCode "Version 10.2 beta (10P82s)", but I have tried using xcode-select to point towards the latest stable XCode version as well, but the problem still occurs. My version of xcrun is "xcrun version 43.1."

## Changelog

[iOS] [Changed] - Backwards compatibility in input format for run-ios --simulator "<device-type (device-version)>"

## Test Plan

Run the command "react-native run-ios --simulator "iPhone X (12.2)". Without the fix I get the error
Could not find iPhone X simulator
With the fix, it starts fine (Assuming you have an iPhone X 12.2 simulator installed on your machine.)

### DIFF
--- a/local-cli/runIOS/findMatchingSimulator.js
+++ b/local-cli/runIOS/findMatchingSimulator.js
@@ -44,7 +44,7 @@ function findMatchingSimulator(simulators, simulatorString) {
     if (!version.includes('iOS') && !version.includes('tvOS')) {
       continue;
     }
-    let versionWithDashReplacedByDot = version.replace(/-/g,'\.')
+    let versionWithDashReplacedByDot = version.replace(/-/g,'\.');
     if (simulatorVersion && !versionWithDashReplacedByDot.endsWith(simulatorVersion)) {
       continue;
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks for making us aware of this issue. I noticed that we already had this fix in the newer version of the CLI that was extracted, so I decided to just backport that one and push it to the 0.58 branch.

* https://github.com/react-native-community/react-native-cli/pull/138
* https://github.com/facebook/react-native/commit/8bddcb6cb0914373a0aeb927f12a6d48ffc5bb84

PR #23322, opened 2019-02-07
ReactEditText extends AppCompatEditText

dulmandakh:
## Summary

Google recommends to extend AppCompat widgets, and this PR changes ReactEditText to extend AppCompatEditText.

## Changelog

[Android] [Changed] - ReactEditText extends AppCompatEditText

## Test Plan

CI is green, and everything works as before.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/BUCK
@@ -3,15 +3,14 @@ load("//tools/build_defs/oss:rn_defs.bzl", "YOGA_TARGET", "react_native_dep", "r
 rn_android_library(
     name = "textinput",
     srcs = glob(["*.java"]),
-    provided_deps = [
-        react_native_dep("third-party/android/support/v4:lib-support-v4"),
-    ],
     required_for_source_only_abi = True,
     visibility = [
         "PUBLIC",
     ],
     deps = [
         YOGA_TARGET,
+        react_native_dep("third-party/android/support/v4:lib-support-v4"),
+        react_native_dep("third-party/android/support/v7/appcompat-orig:appcompat"),
         react_native_dep("third-party/java/infer-annotations:infer-annotations"),
         react_native_dep("third-party/java/jsr-305:jsr-305"),
         react_native_target("java/com/facebook/react/bridge:bridge"),
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -12,6 +12,7 @@ import android.graphics.Rect;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
+import android.support.v7.widget.AppCompatEditText;
 import android.text.Editable;
 import android.text.InputType;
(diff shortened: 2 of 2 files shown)


dulmandakh:
@cpojer is there anything I can help with?

hey99xx:
You may want to test this with old Android OS like API 19.

Earlier I found using AppCompat widgets wrap the original context, making any `(ReactContext) getContext()` throw `java.lang.ClassCastException: android.support.v7.widget.TintContextWrapper cannot be cast to com.facebook.react.bridge.ReactContext`.

I see couple dangerous `getContext` usages that way in `ReactEditText` class.

See https://github.com/facebook/react-native/issues/22885#issuecomment-458435675

dulmandakh:
@hey99xx We cannot stick to platform widgets forever. What would be a generic solution for this?

dulmandakh:
@hey99xx quick google search reveals that it can be solved if we extend AppCompatActivity. So I think it'll gone once 0.59 released with ReactActivity extending AppCompatActivity.

hey99xx:
I dont know why activity class matters for casting context. In support library code that wraps the context with TintContextWrapper only api version is checked, not activity type. I have two suggestion for you:

---

(1) We can either have a generic method that does the context unwrapping done in https://github.com/facebook/react-native/commit/58437cd10a667bbcbc16781df855fd7c3d73bf49 like

```
public static ReactContext getReactContext(View view) { ... }
```

then when we need to get the context

```diff
@Override
public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
-  ReactContext reactContext = (ReactContext) getContext();
+  ReactContext reactContext = ReactContextHelper.getReactContext(this);
```

---

(2) or we save the ReactContext as variable when we know it'll be used again:

```diff
+final ReactContext mReactContext;

-public ReactEditText(Context context) {
-  super(context);
+public ReactEditText(ReactContext context) {
+  super(context);
+  mReactContext = context;
```

```diff
@Override
public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
- ReactContext reactContext = (ReactContext) getContext();
  InputConnection inputConnection = super.onCreateInputConnection(outAttrs);
  if (inputConnection != null && mOnKeyPress) {
-   inputConnection = new ReactEditTextInputConnectionWrapper(inputConnection, reactContext, this);
+   inputConnection = new ReactEditTextInputConnectionWrapper(inputConnection, mReactContext, this);
  }
```

---

I prefer 2nd as it's cleaner, and you don't need to worry about which context class wraps which in the future.

hey99xx:
Btw platform widgets like EditText is not getting deprecated. AppCompatEditText is a subclass of EditText that's documented as 

> A EditText which supports compatible features on older versions of the platform, ...

As I understand, AppCompat widgets only add value for old Android versions to make app look consistent with modern Android versions, as also mentioned in original checkbox PR https://github.com/facebook/react-native/pull/18318. So I support your PRs like this, just make sure you test it :)

dulmandakh:
@hey99xx I target M for my current project, because we've been experiencing many issues/inconsistencies on Android versions below. Recently, I found that RN use many APIs that are not available in Android 4.x, and probable solution to it is to use more from Support Library. Thank you.

dulmandakh:
@cpojer @mdvacca please review and merge 👍

hey99xx:
Aren't setIntrinsicContentSize and onCreateInputConnection crash on lower API versions after this?

PR #23324, opened 2019-02-07
bump gradle to 5.0

dulmandakh:
## Summary

This PR bumps gradle to 5.0, which includes Kotlin DSL 1.0. Also Gradle 5.x requires Java 8.

This is preparation for Kotlin DSL migration.

## Changelog

[Android] [Changed] - Bump Gradle to 5.0

## Test Plan

CI is green, and RNTester builds, runs as normal.


dulmandakh:
It was easy because we were tracking Gradle, plugins and addressed deprecation warnings on timely manner 👍

hramos:
Thank you, looks like tests are passing. We'll need to bump the Gradle version used to populate our offline cache internally as well when we import this. Are there any concerns with how this may affect end users?

dulmandakh:
It may affects users/developers who have custom gradle scripts (brownfield apps) that use removed APIs. We've been fixing deprecation warnings, thus have no issue with this upgrade.

dulmandakh:
@hramos tested it locally before submitting the PR, and confirmed that RNTester works just fine.

dulmandakh:
Let's bump it to 5.0 😉

>

dulmandakh:
@hramos remove Android Gradle Plugin bump to create another PR and cherry pick to 0.59 branch.

PR #23331, opened 2019-02-07
add husky and lint-staged to current workflow

gengjiawen:
## Summary

make format more easier with hooks, just like `create-react-app`.

## Changelog

[General][Workflow] - improve develop efficiency with husky and lint-staged.

## Test Plan

pass current ci

### DIFF
--- a/Libraries/vendor/core/Map.js
+++ b/Libraries/vendor/core/Map.js
@@ -546,8 +546,10 @@ module.exports = (function(global, undefined) {
       }
 
       if (!isES5) {
-        if (hasOwn.call(o, "propertyIsEnumerable") &&
-            hasOwn.call(o.propertyIsEnumerable, hashProperty)) {
+        if (
+          hasOwn.call(o, 'propertyIsEnumerable') &&
+          hasOwn.call(o.propertyIsEnumerable, hashProperty)
+        ) {
           return o.propertyIsEnumerable[hashProperty];
         }
       }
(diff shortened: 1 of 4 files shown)


cpojer:
@hramos do you think this is ok to do and will work with the facebook-github-bot? Basically I'm worried that we make a change internally that won't pass lint rules, which would mess up applying the changes from Mercurial to git and will hold up commits to GitHub until we fix it internally. That would be bad.

(also cc @zpao)

hramos:
I've had mixed experiences using a similar approach in the facebook/react-native-website repo. I probably won't have time to look at this today, let me think about it.

gengjiawen:
@hramos Which part do you dislike. I has been using `husky` + `lint-staged` for a long time. Pretty work well for me. Maybe some cases I am missing ?

zpao:
We shouldn't end up with any actual hooks in fresh clones (eg shipit syncs) since that'll require the post install hook of husky to set up (IIRC) and we don't run that. I think the problem is that since we won't be running this internally, we'll sync code and then the next PR will be making formatting changes to unrelated files.

gengjiawen:
I see https://github.com/facebook/create-react-app/blob/master/package.json using it.

cpojer:
@gengjiawen create-react-app doesn't sync with Facebook, so that's why it can be used there. I'm sorry to have to close this but it seems like FB's tooling isn't going to play nicely with this, so we'll have to keep relying on CI to give us this signal for now.

PR #23333, opened 2019-02-07
fix lint error/warnings

dulmandakh:
## Summary

Fix lint errors and warning, which might be a cause of various crashes on older Android OS, using Android Support Library.

```bash
./gradlew :ReactAndroid:lint
```

## Changelog

[Android] [Changed] - fix lint error/warnings

## Test Plan

CI is green

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java
@@ -7,7 +7,9 @@
 
 package com.facebook.react.views.textinput;
 
+import android.annotation.TargetApi;
 import android.os.Build;
+import android.support.v4.view.ViewCompat;
 import android.text.Layout;
 import android.util.TypedValue;
 import android.view.ViewGroup;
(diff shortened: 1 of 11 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java
```diff
@@ -49,7 +52,7 @@
 
   public ReactTextInputShadowNode() {
     mTextBreakStrategy = (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) ?
-        0 : Layout.BREAK_STRATEGY_SIMPLE;
+      Layout.BREAK_STRATEGY_SIMPLE : Layout.BREAK_STRATEGY_HIGH_QUALITY;
```
mdvacca: is this change of behavior correct?

dulmandakh: BREAK_STRATEGY_HIGH_QUALITY was introduced in 23 or M, and is a default text breaking strategy. Before that BREAK_STRATEGY_SIMPLE was default, which has value of 0.

PR #23359, opened 2019-02-10
Merge `PickerIOS` and `PickerAndroid` into `Picker`.

jainkuniya:
## Summary

As a part of #23313, merge `PickerIOS` to `Picker`. And also combine `PickerAndroid`.

Picker already uses `PikckerIOS` for rendering `Picker` on iOS. Now next
we will remove exporting `PickerIOS`. So merge it to the `Picker`
itself.

We are going to deprecate the use of `PickerIOS` directly, so add a
warning. It can be used via `Picker` on iOS.

Picker already uses `PickerAndroid` behind the scene. So use it directly
and remove one more re-direction. It also shares some code with iOS
picker.

## Changelog
[component] [refactor] - lib, component, picker: Merge `PickerIOS` to `Picker`.
[component] [refactor] - pickerIOS: Add `warnOnce` to the component..
[component] [refactor] - lib, component, picker: Merge `PickerAndroid` to `Picker`..

## Test Plan

```
<Picker
          mode="dropdown"
          selectedValue={this.state.language}
          onValueChange={(itemValue, itemIndex) =>
            this.setState({ language: itemValue })
          }
        >
          <Picker.Item label="Java" value="java" />
          <Picker.Item label="JavaScript" value="js" />
</Picker>
```

Use above code in the RN app and make sure that `Picker` is rendered properly.


bartolkaruza:
@jainkuniya I think you'll be done much faster if you make PickerIOS.js a simple proxy for Picker, leaving only the warning and an export of Picker.js there. It will solve your "Tried to register two views with the same name RCTPicker" issue and remove a lot of complexity.

cpojer:
I agree with @bartolkaruza, it doesn't look like this is the right approach and we should probably do it more conservatively and a less risky change. I'm gonna close this PR because I don't think it gets us closer to a solution.

PR #23367, opened 2019-02-11
JSI: Minor tweaks for building on MSVC

empyrical:
## Summary

This pull request makes two minor changes to `jsi.h`:

 * Tweak the `JSI_EXPORT` macro to automatically set itself to an empty value if `_MSC_VER` is defined - like how was done by @acoates-ms [here](https://github.com/facebook/react-native/blob/8beb4bb58ab93af8c95af6844230d62c85ccab78/ReactCommon/cxxreact/JSBigString.h#L15-L21).
 * Tweak the call to constructor `Pointer(Runtime::PointerValue* ptr)` in the constructor for `PropNameID`. I am not sure why MSVC wasn't working with the original version, but it compiles after I tweak that.

## Changelog

[General] [Fixed] - Tweaked `jsi.h` to build on MSVC

## Test Plan

JSI successfully compiled under Visual Studio 2017

### DIFF
--- a/ReactCommon/jsi/jsi.h
+++ b/ReactCommon/jsi/jsi.h
@@ -1,7 +1,7 @@
 //  Copyright (c) Facebook, Inc. and its affiliates.
 //
 // This source code is licensed under the MIT license found in the
- // LICENSE file in the root directory of this source tree.
+// LICENSE file in the root directory of this source tree.
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


empyrical:
That was a Clang format change - It appears if since this file was last formatted, the option `PointerAlignment` was set to `Right`. I've got "Format on save" in my editor which flipped the alignment on all pointers and references in this file, but I opted to discard most of the clang format changes to reduce the size of the diff. I can manually reset the spacing on the spot I missed if that's preferable as well.

PR #23371, opened 2019-02-11
DatePickerDialogModule supports only FragmentActivity

dulmandakh:
## Summary

Now RN has only ReactActivity which extends AppCompatActivity, subclass of FragmentActivity, therefore no need to check if activity is FragmentActivity or not. This PR changes DatePickerDialogModule to work only with FragmentActivity.

Also DialogFragment from Android is deprecated in API 28, and recommends to use DialogFragment from Support Library. Excerpt from DialogFragment documentation.

> **This class was deprecated in API level 28.**
> Use the Support Library DialogFragment for consistent behavior across all devices and access to Lifecycle.

**BREAKING CHANGE**: Brown field apps must extend FragmentActivity or its subclasses

## Changelog

[Android] [Changed] - DatePickerDialogModule supports only FragmentActivity

## Test Plan

CI is green, and everything works as usual if an activity extends FragmentActivity or its subclasses.


dulmandakh:
@hey99xx another PR to remove plain Activity.

PR #23387, opened 2019-02-11
[Web] Add Platform module for web target

EvanBacon:
## Summary

Added a Platform file for module for instances where internal modules need Platform.
ex: https://github.com/facebook/react-native/blob/78be6efda97096e8388af5bf2dadb9c9de4fa0ba/Libraries/Utilities/Dimensions.js#L14

## Changelog

* Added `Libraries/Utilities/Platform.web.js`
[CATEGORY] [TYPE] - Message

## Test Plan

everything should pass.


TheSavior:
Why does this need to live in core? Do we need to have Platform files for all the platforms people make? `Platform.windows.js`, `Platform.vr.js`, `Platform.firetv.js`? How do those projects currently support these additional platform checks without this file currently existing in the repo?

EvanBacon:
@TheSavior the problem with this particular module is that Dimensions requires it. In RNWeb they have a web module but Dimensions doesn't have access to it. If any module internally accesses `Dimensions` then it won't know how to properly handle this import.

Attempting to use something very unrelated like `AnimatedEvent` will create a tricky error. Currently I have a webpack monkey-patch but it adds significant time to the bundler.

```sh
Module not found: Error: Can't resolve '../Utilities/Platform' in '/node_modules/react-native/Libraries/WebSocket'

/node_modules/react-native/Libraries/WebSocket/WebSocket.js 1:949-981
/node_modules/react-native/Libraries/Core/Devtools/setupDevtools.js
/node_modules/react-native/Libraries/Core/InitializeCore.js
/node_modules/react-native/Libraries/Renderer/oss/ReactNativeRenderer-dev.js
/node_modules/react-native/Libraries/Renderer/shims/ReactNative.js
/node_modules/react-native/Libraries/Animated/src/AnimatedEvent.js
/node_modules/react-native-gesture-handler/DrawerLayout.js
/node_modules/react-native-gesture-handler/index.js
/node_modules/@react-navigation/native/dist/Scrollables.js
/node_modules/@react-navigation/native/dist/index.js
/node_modules/react-navigation/src/react-navigation.js
./App.js
```

TheSavior:
While we should probably merge this just to unblock, I'd really like to have a solution that doesn't require all the platforms to add themselves to the core. Could you think about this and see if you can come up with a solution? :)

TheSavior:
In the case of Dimensions, it looks like Android is special cased because the object we get from the device is different on iOS than Android:
https://github.com/facebook/react-native/blob/master/Libraries/Utilities/Dimensions.js#L44-L56

Dimensions comes from native here:
https://github.com/facebook/react-native/blob/master/Libraries/Utilities/Dimensions.js#L125-L140

It seems like the platforms should be unified on the data they provide to JS so that it can be calculated consistently. At which point other platforms have a standard API to report dimensions through and it doesn't need platform checks.

EvanBacon:
@TheSavior AFAIK this is (should be) the only place we define the platform `OS` value it would seem that adding a new file to Platform is the best solution. Perhaps we could read something from the bundler like `process.env` on web, but that seems hacky and not scalable. 
Depending on how native modules work, maybe a platform could have a native platform module with the same name. That native module could then export the name of it's own OS as a constant. That way it's modular and dependent on the platform. On web we need to think of a way to "install" a native module so it could be accessed the same way as iOS and Android, but this should cover windows.

matthargett:
For an example of how another out-of-tree platform does this:
https://github.com/Microsoft/react-native-windows/blob/master/Libraries/Utilities/Platform.windows.js

Fully agree with @TheSavior , figuring out a way to do this that doesn't require RN itself having a master list of all platforms is key. At my previous job, we were able to bundle RN Windows app using haul which is webpack underneath. 

Of course, I'm open to learning something if there's some counter-indication you have in your specific situation :D

PR #23389, opened 2019-02-12
[Web] Add HMRLoadingView module for web target

EvanBacon:
## Summary

* Added a default value for HMRLoadingView this will prevent errors from being thrown when `HMRClient` attempts to include it in a `web` context.

## Changelog

* [Web] [Added] - HMRLoadingView

## Test Plan

* No new features.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -26,6 +26,9 @@
 ; require from fbjs/lib instead: require('fbjs/lib/warning')
 .*/node_modules/warning/.*
 
+; Ignore duplicate for internal Facebook repo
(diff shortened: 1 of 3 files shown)


EvanBacon:
Hold pending #23387

EvanBacon:
@cpojer does this file exist in the internal repo? Outside of that I'm not sure what I would change to make this more compatible (besides the extension ofc).

cpojer:
Haven't gotten around to do this again. We actually need to ignore this file in .flowconfig for now. Could you update the two in the root of this repo? That helps me land this faster :)

PR #23409, opened 2019-02-12
Add scrollEnabled example to RNTester

berickson1:
## Summary

RNTester is missing a test for the scrollEnabled prop

## Changelog

[General] [Added] - Added RNTester ScrollView scrollEnabled prop test

## Test Plan

Tested in RNTester app


cpojer:
Can you run prettier using `yarn lint --fix` on your project and push again? Thank you!

berickson1:
Snapshot tests appear to be failing on iOS since I added the new example - can anyone explain how to update the snapshot images?

hramos:
1. Open `RNTester/RNTester.xcodeproj` in Xcode, then open `RNTesterSnapshotTests.m`. Change L29 to `_runner.recordMode = YES;`. This will make it so that snapshots are re-recorded on the next run.
2. Select the iPhone 6S simulator from the list of devices in the Xcode toolbar (the simulator used here should match the IOS_DEVICE envvar in `scripts/.tests.env`).
3. You can then hit CMD+U (Menubar > Product > Test) to run tests.
4. Afterwards, switch recordMode back to NO.

You should end up with a new set of snapshots for the ScrollViewExample test.

berickson1:
Perfect, thanks update snapshots as well

hramos:
We should look into that. We're only using the iOS12 snapshots on Circle CI, but I'm not sure if our internal test suite relies on the iOS 10 and iOS 11 snapshots for additional tests.

All looks good on this PR, merging!

PR #23413, opened 2019-02-12
[analysis-bot] Suggest running yarn lint --fix

hramos:
## Summary

This PR expands the code analysis script to allow for customizing the GitHub Review left by @analysis-bot whenever it finds code-level issues.

Some, but not all, `eslint` errors are fixable by running `yarn lint --fix`. When the `eslint` converter finds warnings or errors, the bot will suggest running the command as part of the review.

The script could be adjusted to only suggest this when `eslint` returns a non-zero `fixableErrorCount` or `fixableWarningCount`, but I think this is a good first step to get people to unblock themselves when they see this type of review from the bot.

I've also excluded `bots/` and `scripts/` from eslint as these have several `eslint` errors that need to be addressed. These directories do not contain core RN code, so they can be excluded for now.

## Changelog

[GENERAL] [Changed] - @analysis-bot will suggest fixing eslint warnings on code reviews

## Test Plan

Intentionally added a prettier issue in `RNTester/js/ScrollViewExample.js` in 7617afa, and confirmed bot left a review in https://github.com/facebook/react-native/pull/23413#pullrequestreview-202961036:
screen shot 2019-02-12 at 3 24 52 pm

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,3 +8,5 @@ pr-inactivity-bookmarklet.js
 question-bookmarklet.js
 flow/
 danger/
+bots/
+scripts/
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,6 +7,4 @@ Libraries/Renderer/*
 pr-inactivity-bookmarklet.js
 question-bookmarklet.js
 flow/
-danger/
-bots/
-scripts/
+bots/node_modules
--- a/scripts/circleci/code-analysis-bot.js
+++ b/scripts/circleci/code-analysis-bot.js
@@ -30,6 +30,12 @@ function push(arr, key, value) {
   arr[key].push(value);
 }
 
+const converterSummary = {
+  eslint: '`eslint` found some issues. Run `yarn lint --fix` to automatically fix problems.',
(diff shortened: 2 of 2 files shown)


.eslintignore
```diff
@@ -8,3 +8,5 @@ pr-inactivity-bookmarklet.js
 question-bookmarklet.js
 flow/
 danger/
+bots/
+scripts/
```
cpojer: I'm worried about excluding them because most likely we'll never actually fix them and re-enable them. Could we instead either fix the errors or modify the eslint config to disables the linters we don't care about for those folders?

hramos: Done. `eslint` only ignores the top `node_modules` directory by default, so using `bots/node_modules` here is sufficient to suppress the errors originating from third party code.

.eslintignore
```diff
@@ -7,4 +7,4 @@ Libraries/Renderer/*
 pr-inactivity-bookmarklet.js
 question-bookmarklet.js
 flow/
-danger/
```
hramos: `danger/` was removed some time ago, the dangerfile is now in `bots/`

PR #23414, opened 2019-02-13
Fabric: Fix building on MSVC

empyrical:
## Summary

This pull request modifies the source code in the `ReactCommon/fabric` directory so it will build on MSVC 2017

The following changes have been done:

 * Removed the usages of designated initializers and replaced them with list initializers. This is the vast majority of the changes done to Fabric
 * Changed the rule `AlignTrailingComments` in the `.clang-format` file to `true` so the comments on the initializers are nicely organized like this:
```cpp
    newChild = oldChild->clone({
        0,                                            // tag
        0,                                            // rootTag
        ShadowNodeFragment::nullSharedProps(),        // props
        ShadowNodeFragment::nullSharedEventEmitter(), // eventEmitter
        sharedChildren,                               // children
    });
```
* Changed the null coalescing operator usage (`?:`) to a ternary in `ReactCommon/fabric/core/shadownode/ShadowNode.cpp`
* Removed default constructor in `ReactCommon/fabric/debug/DebugStringConvertibleItem.h` that was conflicting with the constructor that has all default values - MSVC had issues with this

## Changelog

[General] [Fixed] - Fixed Fabric compilation on MSVC

## Test Plan

The Fabric test suite has been ran on builds made with Windows + MSVC 2017 and Ubuntu 18.10 + Clang 7.0.0, and all tests pass on both platforms

```
[----------] Global test environment tear-down
[==========] 31 tests from 17 test suites ran. (573 ms total)
[  PASSED  ] 31 tests.
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -5,7 +5,7 @@ AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlinesLeft: true
 AlignOperands:   false
-AlignTrailingComments: false
+AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
--- a/ReactCommon/fabric/components/image/ImageShadowNode.cpp
+++ b/ReactCommon/fabric/components/image/ImageShadowNode.cpp
@@ -48,7 +48,9 @@ ImageSource ImageShadowNode::getImageSource() const {
   auto sources = getProps()->sources;
 
   if (sources.size() == 0) {
-    return {.type = ImageSource::Type::Invalid};
+    return {
+        ImageSource::Type::Invalid // type
+    };
   }
 
   if (sources.size() == 1) {
--- a/ReactCommon/fabric/components/root/RootShadowNode.cpp
+++ b/ReactCommon/fabric/components/root/RootShadowNode.cpp
@@ -29,7 +29,12 @@ UnsharedRootShadowNode RootShadowNode::clone(
   auto props = std::make_shared<const RootProps>(
       *getProps(), layoutConstraints, layoutContext);
   auto newRootShadowNode = std::make_shared<RootShadowNode>(
-      *this, ShadowNodeFragment{.props = props});
+      *this,
+      ShadowNodeFragment{
+          0,     // tag,
+          0,     // rootTag,
+          props, // props
+      });
   return newRootShadowNode;
 }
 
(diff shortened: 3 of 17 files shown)


ReactCommon/fabric/components/root/RootShadowNode.cpp
```diff
@@ -55,11 +60,24 @@ UnsharedRootShadowNode RootShadowNode::clone(
     sharedChildren = std::make_shared<SharedShadowNodeList>(children);
 
     oldChild = ancestor.get().shared_from_this();
-    newChild = oldChild->clone(ShadowNodeFragment{.children = sharedChildren});
+    newChild = oldChild->clone({
```
shergin: Here and everywhere with `ShadowNodeFragment`:

I think if we should:
 * Rename `ShadowNodeFragment::nullSharedProps()` (and family) to `ShadowNodeFragment::propsPlaceholder()`;
 * Introduce similar functions for the rest of members (tag and rootTag);
 * Feel free to rename `rootTag` to `surfaceId`.

I believe if we go this way, we don't need comments in the code.

Seems this is only data structure worth to improve this way, the rest is fine.

.clang-format
```diff
@@ -5,7 +5,7 @@ AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlinesLeft: true
 AlignOperands:   false
-AlignTrailingComments: false
+AlignTrailingComments: true
```
shergin: I don't think we need really this. Having this option introduces bigger diffs which complicates merges... and the real value of having things aligned is kinda debatable.
This is quite unpopular setting across FB monorepo, so... I would prefer not to enable this.

empyrical:
Sounds good! Would being split up over 2 or 3 pull requests sound easier to deal with?

cpojer:
Closing this PR as you are planning on splitting it up into smaller ones.

gengjiawen:
@empyrical Do you think we can add ci to keep cpp code compiles on MSVC ?

empyrical:
I'm not sure how hard it would be to do that, but I think that fabric could be built with `-pedantic` on CI to help stop non-standard CPP from making it in

ReactCommon/fabric/uimanager/ComponentDescriptorRegistry.cpp
```diff
@@ -113,12 +113,13 @@ SharedShadowNode ComponentDescriptorRegistry::createNode(
   ComponentName componentName = componentNameByReactViewName(viewName);
   const SharedComponentDescriptor &componentDescriptor = (*this)[componentName];
 
-  SharedShadowNode shadowNode = componentDescriptor->createShadowNode(
-      {.tag = tag,
-       .rootTag = rootTag,
-       .eventEmitter =
-           componentDescriptor->createEventEmitter(std::move(eventTarget), tag),
-       .props = componentDescriptor->cloneProps(nullptr, RawProps(props))});
+  SharedShadowNode shadowNode = componentDescriptor->createShadowNode({
```
matthargett: use ```auto``` keyword here?

ReactCommon/fabric/uimanager/Scheduler.cpp
```diff
@@ -105,9 +105,15 @@ void Scheduler::renderTemplateToSurface(
           [&](const SharedRootShadowNode &oldRootShadowNode) {
             return std::make_shared<RootShadowNode>(
                 *oldRootShadowNode,
-                ShadowNodeFragment{.children =
-                                       std::make_shared<SharedShadowNodeList>(
-                                           SharedShadowNodeList{tree})});
+                ShadowNodeFragment{
+                    0,                                     // tag
```
matthargett: I think you can eliminate the type name here as well, it should be deduced --right?

ReactCommon/fabric/components/image/ImageShadowNode.cpp
```diff
@@ -48,7 +48,9 @@ ImageSource ImageShadowNode::getImageSource() const {
   auto sources = getProps()->sources;
 
   if (sources.size() == 0) {
-    return {.type = ImageSource::Type::Invalid};
+    return {
+        ImageSource::Type::Invalid // type
```
Orvid: As a note, most of the other Windows-compatible code in the fb monorepo that has migrated from designated initializers has done it with leading comments:

```
return {
    /* .type = */ ImageSource::Type::Invalid
};
```

That also makes it easier to codemod out the comment and return them to actual designated initializers when MSVC gets around to actually supporting them. (IIRC they have made it into C++20, so it shouldn't be too long a wait)

PR #23424, opened 2019-02-13
lock docker image version in ci

gengjiawen:
## Summary

Lock the version will help us to test new dependency version like ndk or buck.

## Test Plan

pass current ci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,7 +283,7 @@ js_defaults: &js_defaults
 android_defaults: &android_defaults
   <<: *defaults
   docker:
-  - image: reactnativecommunity/react-native-android
+  - image: reactnativecommunity/react-native-android:2019-1-19
   resource_class: "large"
   environment:
     - TERM: "dumb"


cpojer:
How often do you expect this image to change? Can we lock it to something that doesn't have a date in it? I'm not really sure if this is better than what we have now.

dulmandakh:
IIRC, CI will try to pull latest image from Docker registry and use it. And using this versioning scheme we can make sure that only working versions are used for CI, or prevent to use latest version automatically.

gengjiawen:
React native is very sophisticated project in my opinion,  we may update it if `jdk`, `node`, `buck` or any other tools release new version.

You can check the image repo here: https://github.com/react-native-community/docker-android.

dulmandakh:
@cpojer I expect that versions will be changed when we change tools versions, like SDK, NDK or BUCK.

gengjiawen:
@cpojer If you want to involve the docker project, just let me know. I have talked to @hramos , but he has not reply my message.

gengjiawen:
@cpojer should we import this again ?

cpojer:
Yap! Seems like the import failed on the first try for some unknown reason.

PR #23428, opened 2019-02-13
chore: remove the unused packages from package.json

nimish-gupta:
## Summary

There are couple of packages which are installed as dependency but are not directly used in the project. In this PR I have remove those packages from dependency from the package.json.

## Changelog

[General] [Removed] - Remove the unused packages from package.json


cpojer:
This is awesome! @nimish-gupta can you walk me through the process you used to find unused dependencies? I wanna just make sure we aren't removing anything that we still need.

gengjiawen:
Do they originally belong to `react-native-cli` ?

cpojer:
@gengjiawen yes, or Metro. See https://github.com/facebook/react-native/issues/23313 for more information.

nimish-gupta:
@cpojer for removing the packages
-  I just search the dependencies through out the repo, and if that is not used from anywhere I just removed it.
- I also did `yarn remove` but it doesn't cause any change because these dependencies are indirectly used by different packages.

nimish-gupta:
@cpojer yeah sure I will work on this one.
 > As a next step, do you think you could identify the largest dependencies and make an issue so we can analyze which ones we may want to consider stripping? Thank you!

dalcib:
It seems that `lodash` is no longer used either.
https://github.com/facebook/react-native/search?q=lodash&type=Code

cpojer:
Importing again because there was an issue yesterday…

PR #23429, opened 2019-02-13
improve Android Network Security config

dulmandakh:
## Summary

This PR is trying to improve Android Network Security configuration introduced in https://github.com/facebook/react-native/commit/84572c4051f11f68ddf0928d2c3df5850ae15491. I found that Android merges all manifest files into a single manifest file when building an app, so this PR provides AndroidManifest.xml with network security config in debug folder, that will be used only for debug builds. Also the network security configuration will be applied to only app that target API 28. Moved security config file to xml-v28, so that it'll only visible to API 28 and above.

See https://developer.android.com/studio/build/manifest-merge

## Changelog

[Android] [Changed] -  Android Network Security configuration.

## Test Plan

CI is green, and RN apps will work that targets API 28.


dulmandakh:
@gengjiawen @Salakar @hey99xx please review

Salakar:
@dulmandakh thanks for this, good idea changing this to limit it specifically to API 28; LGTM

PR #23431, opened 2019-02-13
bump targetSdkVersion to 28

dulmandakh:
## Summary

Bump targetSdkVersion to 28

## Changelog

[Android] [Changed] - Bump targetSdkVersion to 28

## Test Plan

CI is green, and apps will work as usual.


dulmandakh:
@cpojer maybe we can cherry pick it to 0.59, because Android Studio is highlighting **targetSdkVersion** and suggesting to use the latest one. But I don't have opinion about it.

PR #23434, opened 2019-02-13
add clang-format to current toolchain

gengjiawen:
## Summary

add `clang-format` toolchain. we have 2558 files need to process in current task.
cc @cpojer @empyrical @shergin @matthargett 

## Test Plan

pass current ci.


package.json
```diff
@@ -139,6 +139,8 @@
     "flow-check-ios": "flow check",
     "flow-check-android": "flow check --flowconfig-name .flowconfig.android",
     "lint": "eslint .",
+    "clang-format": "clang-format -i --glob=*/**/*.{h,cpp,m,mm}",
```
shergin: Are we sure that in this configuration clang-formant will pick up the style guide from the file in the root of the repo (not some default settings)?

I suspect we should pass something like `--file`.

gengjiawen: I tested it, it uses the root config. Do you want me make it explicit ?

gengjiawen: > clang-format supports two ways to provide custom style options: directly specify style configuration in the -style= command line option or use -style=file and put style configuration in the .clang-format or _clang-format file in the project directory.

From https://clang.llvm.org/docs/ClangFormatStyleOptions.html.

But maybe adding it good for fb internal development ?

package.json
```diff
@@ -162,6 +164,8 @@
     "@react-native-community/cli": "^1.1.0",
     "art": "^0.10.0",
     "base64-js": "^1.1.2",
+    "clang-format": "^1.2.4",
+    "connect": "^3.6.5",
```
cpojer: This was me during the merge. `connect` was removed in another diff accidentally, and this diff still had it as part of the conflict, and I need to patch up `yarn.lock` anyway so I figured I'll just do it here. I realize this is not the proper way of dealing with it.

PR #23438, opened 2019-02-13
Fabric: Remove null coalescing operators in ShadowNode

empyrical:
## Summary

This pull request removes the use of the GCC extension null coalescing operators (`?:`) and replaces them with ternary operators. This improves the portability of `ShadowNode.cpp` (and enables MSVC to build it)

## Changelog

[General] [Fixed] - Fabric: Removed null coalescing operators in `ShadowNode`

## Test Plan

Successfully builds and ran the Fabric test suite with Ubuntu 18.10 + Clang 7


empyrical:
Yup, clang-format was ran on this file before committing :+1:

PR #23444, opened 2019-02-13
Fix buildType issue introduced in rebase

hramos:
## Summary

While rebasing https://github.com/facebook/react-native/pull/18892, I missed that buildTypeName no longer existed. 

## Changelog

[ANDROID] [Fixed] - Fix issue in react.gradle

## Test Plan

CI


hramos:
Confirmed this fixes CI, landing.

PR #23446, opened 2019-02-14
[iOS] Fix image wrong scale factor when load image from file system

### LINKED ISSUE #22383: Regression: iOS images loaded from filesystem don't respect scaling factor in filename
## Environment
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
      Memory: 206.29 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.9.0 - /src/homebrew/bin/node
      Yarn: 1.9.4 - /src/homebrew/bin/yarn
      npm: 6.2.0 - /src/homebrew/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 28
        Build Tools: 28.0.2
    IDEs:
    npmPackages:
      react: 16.6.1 => 16.6.1 
      react-native: 0.57.5 => 0.57.5 

## Description
Image assets loaded from the filesystem don't respect iOS's image scaling naming convention.

This appears to be a regression of functionality introduced by 998197f444aca06cde0d5258469b3d314f8ea8b9 where `[UIImage imageWithContentsOfFile:filePath];` was replaced by `[UIImage imageWithData:fileData];`, but that leaves UIKit unable to inspect the `filePath` for the presence of `@3x` filename suffixes.

Reverting that change locally fixes the issue.

Where this really comes into play is in projects using CodePush where the image assets are loaded from the filesystem (e.g., `file:///.../Library/Application%20Support/CodePush/path/to/Image@3x.png`) vs. the bundle included with the application at build time (e.g., `file:///var/containers/Bundle/Application/.../path/to/Image@3x.png`) where the filename-based constructor `[UIImage imageNamed:imageName];` is used.

## Reproducible Demo
CodePush bugs contain references to sample repros:
https://github.com/Microsoft/react-native-code-push/issues/1222
https://github.com/Microsoft/react-native-code-push/issues/1276

Let me know if you need any help setting up a repro... although at a minimum if the desired functionality is to always respect iOS's image scaling filename convention then I feel it should be fairly safe to revert (a portion of) the original change.

zhongwuzw:
## Summary

Regression, fix image load from `~/Library` not respect scale factor. 
Fixes #22383 , the bug comes from [Clean up some URL path handling](https://github.com/facebook/react-native/commit/998197f444aca06cde0d5258469b3d314f8ea8b9).

## Changelog

[iOS] [Fixed] - Fix image wrong scale factor when load image from file system

## Test Plan

If we have picture `panda`, and has `@1x/@2x/@3x` photo, we can load the correct image based on screen scale factor.
`<Image source={{uri: '~/Library/panda.jpg', width: 300, height: 300}} />`

### DIFF
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -706,9 +706,9 @@ UIImage *__nullable RCTImageFromLocalAssetURL(NSURL *imageURL)
     image = [UIImage imageNamed:imageName];
   }
 
-  if (!image && imageURL.isFileURL) {
+  if (!image) {
     // Attempt to load from the file system
-    NSString *filePath = imageURL.path;
+    NSString *filePath = [NSString stringWithUTF8String:[imageURL fileSystemRepresentation]];
     if (filePath.pathExtension.length == 0) {
       filePath = [filePath stringByAppendingPathExtension:@"png"];
     }
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -708,13 +708,11 @@ UIImage *__nullable RCTImageFromLocalAssetURL(NSURL *imageURL)
 
   if (!image) {
     // Attempt to load from the file system
-    NSData *fileData;
-    if (imageURL.pathExtension.length == 0) {
-      fileData = [NSData dataWithContentsOfURL:[imageURL URLByAppendingPathExtension:@"png"]];
-    } else {
-      fileData = [NSData dataWithContentsOfURL:imageURL];
+    NSString *filePath = imageURL.path;
+    if (filePath.pathExtension.length == 0) {
+      filePath = [filePath stringByAppendingPathExtension:@"png"];
     }
-    image = [UIImage imageWithData:fileData];
+    image = [UIImage imageWithContentsOfFile:filePath];
   }
 
   if (!image && !bundle) {
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -706,7 +706,7 @@ UIImage *__nullable RCTImageFromLocalAssetURL(NSURL *imageURL)
     image = [UIImage imageNamed:imageName];
   }
 
-  if (!image) {
+  if (!image && imageURL.isFileURL) {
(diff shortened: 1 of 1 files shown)


cpojer:
This change was originally made because accessing `imageURL.path` for filesystem operations is not safe. Is there a different way to fix this that doesn't fall back to NSURL?

zhongwuzw:
@cpojer Hi, can you please give more details about unsafe? 🤔 you mean the `imageURL` may not be a valid file URL?

cpojer:
@zhongwuzw I actually don't know myself, but this is from the original diff that introduced this change:

> It's not safe to use -[NSURL path] for filesystem paths.
> This is why NSURL has methods for dealing
> with and producing filesystem paths (fileSystemRepresentation, etc).
> This change switches to some filesystem-safe NSURL methods.

zhongwuzw:
@cpojer 😂 TBO, I don't get the whole meaning about safe things, I add a file url check that ensure path is a valid file system path, can we ping anyone to check this?

React/Base/RCTUtils.m
```diff
@@ -706,15 +706,13 @@ BOOL RCTIsLocalAssetURL(NSURL *__nullable imageURL)
...
 
-  if (!image) {
+  if (!image && imageURL.isFileURL) {
     // Attempt to load from the file system
-    NSData *fileData;
-    if (imageURL.pathExtension.length == 0) {
-      fileData = [NSData dataWithContentsOfURL:[imageURL URLByAppendingPathExtension:@"png"]];
-    } else {
-      fileData = [NSData dataWithContentsOfURL:imageURL];
+    NSString *filePath = imageURL.path;
```
sahrens: Can you use this instead:

`NSString *filePath = [NSString stringWithUTF8String:[imageURL fileSystemRepresentation]];`

zhongwuzw:
@sahrens  @cpojer Done! 🍻

PR #23449, opened 2019-02-14
fix: add local-cli to react-native package

Esemesek:
Make `local-cli` part of `react-native` package to prevent breaking changes.


grabbou:
@facebook-github-bot ship

PR #23464, opened 2019-02-14
Add jsx extension to transform and testMatch in jest-preset

KarthikMAM:
## Summary

A lot of time people will write react components and tests in files with extension **.jsx**. So, tests ending with this extension don't run and the component files written with this extension don't get transpiled.

## Changelog

[General] [Fixed] - Add jsx extension to transform and testMatch in jest-preset

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -24,7 +24,7 @@ module.exports = {
   },
   modulePathIgnorePatterns: [`${dir}/Libraries/react-native/`],
   transform: {
-    '^.+\\.(js|ts|tsx)$': 'babel-jest',
+    '^.+\\.(js|ts|tsx|jsx)$': 'babel-jest',
     '^.+\\.(bmp|gif|jpg|jpeg|mp4|png|psd|svg|webp)$': require.resolve(
       './jest/assetFileTransformer.js',
     ),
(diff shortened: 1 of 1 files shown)


cpojer:
Please change your own Jest configuration in your project template to overwrite this stuff in the template. I think we don't want to promote using jsx specifically in RN.

KarthikMAM:
Ya sure 👍. But a bit curious, any specific reasons for not promoting the use of jsx extension with RN.

navilan:
@cpojer - Could you please clarify your comment when you get a chance? Is there a reason why you don't want to promote `jsx` in RN?  Since you have `tsx` allowed in the babel preset, are you gently nudging people to use `tsx` instead?

cpojer:
@navilan Absolutely! Originally, `jsx` was invented because people weren't used to compiling their plain `js` files and it wasn't common to add syntax extensions to JS. It was a way to get the community to accept custom flavors of JavaScript. The reason `tsx` exists is because originally TypeScript's syntax conflicted with JSX. That's why they invented `tsx` which doesn't have this incompatibility.

Nowadays, people compile their `js` files. I do not think we need to promote usage of `jsx` over `js`.  As you can see, `tsx` is still necessary.

Note that not including this change into React Native only means this won't be a default but you can still change your project to include `jsx` files across your project.

navilan:
Thanks for the clear reply @cpojer. Much appreciated.

PR #18627, opened 2018-03-30
iOS: Fixed the bug where a Backspace event was emitted when entering characters after clearing a text in TextInput by an empty string

hamaron:
The bug #18374 was caused by the loose condition to execute `stringByReplacingCharactersInRange` in the method `textInputShouldChangeTextInRange` . As a result, `findMismatch` wrongly returning `true` which ends up the Backspace event being fired in another `textInputShouldChangeTextInRange` call in `textInputDidChange`.

## Test Plan
1. Pass all the tests by `yarn run test`
2. Run the following code and type any text. (This code is brought from #18374. Thank you @michalpetrov!!) And then verify that 'Backspace' events are not emitted after clearing text
and entering any letters.

```javascript
type Props = {};
type State = {
  text: string,
  keys: string
};
export default class App extends Component<Props, State> {
  state = {text: '', keys: ''}
  render() {
    return (
      <View style={styles.container}>
        <TextInput style={styles.textInput} value={this.state.text} onChangeText={this.onChangeText} onKeyPress={this.onKeyPress}/>
        <Button title="Clear" onPress={this.onClear}/>
        <Text>Text: {this.state.text}</Text>
        <Text>Keys: {this.state.keys}</Text>
      </View>
    );
  }

  onChangeText = (text: string) => {
    this.setState({text})
  }

  onKeyPress = ({ nativeEvent }: Object) => {
    this.setState({keys: this.state.keys + nativeEvent.key + ', '})
  }

  onClear = () => {
    this.setState({text: '', keys: ''})
  }
}
```

## Release Notes

[IOS] [BUGFIX] [TextInput] - Fixed the bug where Backspace event was emitted when entering a character after clearing a text in TextInput by an empty string

PR #23472, opened 2019-02-15
[iOS] Fix TextInput maxLength when insert characters at begin

### LINKED ISSUE #21639: TextInput duplicate value when inserting at the beginning (ios) maxLength

zhongwuzw:
## Summary

Fixes #21639 , seems we tried to fix this before, please see related `PR` like [D10392176](https://github.com/facebook/react-native/commit/36507e4a3c2fa58dcfdc9bd389b37536c66006d0), #18627, but they don't solve it totally. 

## Changelog

[iOS] [Fixed] - Fix TextInput maxLength when insert characters at begin

## Test Plan

From the git log, we faced some issues, like [Backspace event](https://github.com/facebook/react-native/pull/18627), we need to ensure no regression happened, issues like #21639 #18627 should be fixed.

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -338,11 +338,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
   NSString *previousText = [_predictedText substringWithRange:range] ?: @"";
 
-  if (!_predictedText || backedTextInputView.attributedText.string.length == 0) {
-    _predictedText = text;
-  } else {
-    _predictedText = [_predictedText stringByReplacingCharactersInRange:range withString:text];
-  }
+  _predictedText = [backedTextInputView.attributedText.string stringByReplacingCharactersInRange:range withString:text];
 
   if (_onTextInput) {
     _onTextInput(@{
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -330,13 +330,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
     }
   }
 
-  if (range.location + range.length > _predictedText.length) {
-    // _predictedText got out of sync in a bad way, so let's just force sync it.  Haven't been able to repro this, but
-    // it's causing a real crash here: #6523822
-    _predictedText = backedTextInputView.attributedText.string;
-  }
-
-  NSString *previousText = [_predictedText substringWithRange:range] ?: @"";
+  NSString *previousText = backedTextInputView.attributedText.string ?: @"";
 
   _predictedText = [backedTextInputView.attributedText.string stringByReplacingCharactersInRange:range withString:text];
 
@@ -372,7 +366,6 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
     [self textInputShouldChangeTextInRange:predictionRange replacementText:replacement];
     // JS will assume the selection changed based on the location of our shouldChangeTextInRange, so reset it.
     [self textInputDidChangeSelection];
-    _predictedText = backedTextInputView.attributedText.string;
   }
 
   _nativeEventCount++;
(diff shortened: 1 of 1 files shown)


shergin:
@rigdern Adam, what do you think?

rigdern:
@shergin I'm not familiar with this code. I read over the PR but it's unclear to me what flaw in the code is causing the symptom and why this delta fixes it.

zhongwuzw:
@shergin @rigdern Thanks for review, my opinion is `_predictedText` should always equal to the combination of `backedTextInputView`'s original text and replaced `text`. The original `if..else..` check may leads to inconformity between `_predictedText` and `backedTextInputView`'s text.

zhongwuzw:
@shergin @cpojer Hi, any opinion?

ejanzer:
@zhongwuzw Can you share that the issue is you're seeing and how to repro it? Like @rigdern said, it's not clear to me what the issue is and how this fixes it - if you're seeing an inconsistency here, then it seems like predictedText already got out of sync with the backing textinput view. I think this change may make sense regardless, but I just want to take a look and see where things are going wrong.

zhongwuzw:
@ejanzer Hi, I borrowed the repro from #21639 :

> ## Description
> [iOS only]
> 
> When entering 1 character at the beginning of the TextInput, we expect to have only 1 character inserted. However we experience that the character is inserted and the current value is duplicated.
> 
> This behaviour is reproducible when the prop `maxLength` is used and when length of the current value is above the half of the `maxLength` prop.
> 
> I have created an app from scratch using `create-react-native-app`, and I have modified `App.js` to include the following:
> 
> ```js
> type Props = {};
> export default class App extends Component<Props> {
>   constructor(props) {
>     super(props);
>     this.state = { text: '10000' };
>   }
> 
>   render() {
>     return (
>       <View style={styles.container}>
>         <TextInput
>           style={{height: 40, width:120, borderColor: 'gray', borderWidth: 1}}
>           maxLength={10}
>           onChangeText={(text) => this.setState({text})}
>           value={this.state.text}/>
>       </View>
>     );
>   }
> }
> ```
> 
> ## Reproducible Demo
 demo

zhongwuzw:
Based on repro example, when we insert `0` at the begin, the `_predictedText` should equal to `0 10000`, but the old code, it equal to `0`, leads to the issue.

ejanzer:
Oh I see, the problem is if you set an initial value from JS then `_predictedText` never gets properly initialized, so it's nil when you get here. We could also fix this by explicitly checking if it's nil, but I think this change makes sense - from the code it looks like _predictedText is supposed to be used to detect when the backing textinput updated without `textInputShouldChangeTextInRange` getting called. 

@zhongwuzw would you mind also cleaning up the other places in this file where we set `_predictedText`, since they're no longer needed?

Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -338,11 +338,7 @@ - (BOOL)textInputShouldChangeTextInRange:(NSRange)range replacementText:(NSStrin
 
   NSString *previousText = [_predictedText substringWithRange:range] ?: @"";
```
ejanzer: We should probably move this above here so that previousText is set correctly.

zhongwuzw:
@ejanzer Hey, I clean up the code, please continue to review it.

PR #23476, opened 2019-02-15
chore: upgrade `jest-junit` to avoid older jest dependencies

SimenB:
## Summary

Upgrade `jest-junit` to the latest version (breaking change between 5 and 6 was reverted). In particular, 6.2.0 removes the dependency on `jest-config` which pulls in `babel-core@6` etc.

I noticed this since `react-native` had `jest-junit` in `dependencies` until #23276. So not as big of an issue. Still a nice cleanup, I'd say

/cc @palmerj3

Btw, any reason Jest is still at alpha?

## Changelog

[General] [Fixed] - upgrade `jest-junit` to avoid older jest dependencies

## Test Plan

I guess check that the junit report is still picked up by CI? 😀

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "eslint-plugin-react-native": "3.5.0",
     "flow-bin": "^0.92.0",
     "jest": "24.0.0-alpha.6",
-    "jest-junit": "5.2.0",
+    "jest-junit": "6.3.0",
     "jscodeshift": "^0.6.2",
     "prettier": "1.13.6",
     "react": "16.8.1",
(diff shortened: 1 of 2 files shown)


SimenB:
`babel-core` is still in the lockfile since jest is at version `alpha.6`, btw. Upgrading to `alpha.9` (or just stable) will include the internal upgrade to `babel@7`

cpojer:
Yeah, @mjesun is trying to upgrade to the latest release of Jest but he is having trouble with a `jest-worker` issue that we haven't figured out yet.

PR #23485, opened 2019-02-16
[iOS] Apply latest shadow view workflow description

zhongwuzw:
## Summary

Apply latest shadow view workflow description

## Changelog

[iOS] [Fixed] - Apply latest shadow view workflow description

## Test Plan

N/A.


cpojer:
Thanks!

Issue #22925, opened 2019-01-09
No fling/momentum on scroll (Android Pie 9.0)


alexandrius:
## Environment

>   React Native Environment Info:
>     System:
>       OS: macOS 10.14.2
>       CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
>       Memory: 990.90 MB / 16.00 GB
>       Shell: 5.3 - /bin/zsh
>     Binaries:
>       Node: 11.3.0 - /usr/local/bin/node
>       Yarn: 1.12.3 - /usr/local/bin/yarn
>       npm: 6.5.0 - /usr/local/bin/npm
>       Watchman: 4.9.0 - /usr/local/bin/watchman
>     SDKs:
>       iOS SDK:
>         Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
>     IDEs:
>       Android Studio: 3.2 AI-181.5540.7.32.5056338
>       Xcode: 10.1/10B61 - /usr/bin/xcodebuild
>     npmPackages:
>       react: 16.6.3 => 16.6.3 
>       react-native: 0.57.8 => 0.57.8 
>     npmGlobalPackages:
>       react-native-cli: 2.0.1
>       react-native-git-upgrade: 0.2.7

## Description
I installed custom rom PixelExperience on my Xiaomi Redmi 5 Plus.
Since then I'm experiencing scroll issue. If you make the page lose focus - scroll lacks the fling or momentum behaviour.
This only happens in React Native apps.
Live demo.
https://www.youtube.com/watch?v=vG2fTEb_ICg
Emulator works just fine.

I don't think is the custom rom issue since another user (@vforvasile) has reported the same.
Android has some changes for scrolling element https://developer.android.com/about/versions/pie/android-9.0-changes-28#scrolling-element
I tried using SDK 26 and 27 but it didn't solve the issue

## Reproducible Demo
The issue described above is present in expo too. 
https://snack.expo.io/@alexandrius/c3VwcG

esmatptr:
+1
I also have the same issue on Android 9 devices,

using: react-native 0.54.0

alexandrius:
I tried different ROM Resurrection Remix today. It has the same issue with fling.

alexandrius:
I tried react-native 0.58.4 today still same issue. I will try to dig into react native code today to get the idea

patrickkempff:
@alexandrius thank you for digging into this

[...]

PR #23490, opened 2019-02-16
Switch to NestedScrollView

alexandrius:
## Summary

Fixes bug described at #22925 

## Changelog

`ReactScrollView` extends `NestedScrollView` instead of normal `ScrollView`

[Android] [Changed] - Switch to NestedScrollView to fix Scrolling issue on AOSP ROMS

## Test Plan
AOSP based rom needs to be installed on the phone.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java
@@ -18,7 +18,7 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.OverScroller;
-import android.widget.ScrollView;
+import android.support.v4.widget.NestedScrollView;
 
 import com.facebook.infer.annotation.Assertions;
 import com.facebook.react.bridge.ReactContext;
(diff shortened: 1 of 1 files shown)


dulmandakh:
There is the excerpts from the documentation

> NestedScrollView is just like ScrollView, but it supports acting as both a nested scrolling parent and child on both new and old versions of Android. Nested scrolling is enabled by default.

cpojer:
Are there other behavior changes that come with this? Right now this is failing internally by an end-to-end test that tests whether the view can be scrolled down properly.

alexandrius:
There shouldn't be. Some method modifiers are different. What is cause of e2e test fail?

cpojer:
We have an internal test like this:

```
it(
      'Tests that we scroll down and find expected component',
      async (expect, scenario) => {
        await scenario
          .waitTillLaunched()
          .then(screen => screen.selectUniqueMainView())
          .then(mainView => mainView.goToScrollingScreen())
          .then(scrollingScreen => scrollingScreen.selectUniqueScrollingView())
          .then(scrollingView => scrollingView.scrollDownToCell(50));
      },
    )
```

it seems that it doesn't scroll down with this change applied. Note that this is a custom framework and the APIs probably don't translate to open source solutions but maybe it gives you an idea about what could be wrong.

alexandrius:
I highly doubt that it doesn't scroll to the requested coordinate.
test might fail if the test expects ScrollView component instead of ReactScrollView.

cpojer:
Ah yeah, I bet that's the case. I can investigate soon and see if I can fix it.

cpojer:
I gave this another look and it doesn't seem like it's an issue with the test framework or locating the scroll view component. I then took another look at the source for ScrollView and NestedScrollView in Android and despite the document saying they are "just like" each other, they do look significantly different enough that swapping out such a core piece of infrastructure for RN is bound to cause issues so I'm not gonna go ahead with this change.

I recommend you to make your own native component and use your own implementation of ScrollView for now. I'm sorry that this isn't ideal but I hope you'll understand a change like this can be very disruptive.

Issue #18177, opened 2018-03-03
Modal component does not support SafeAreaView bottom margin


LutherBaker:
`SafeAreaView` bottom margin is not honored inside of a `Modal` on an iPhone X.

### Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 9.7.1
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.7.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.53.3 => 0.53.3

### Expected Behavior

On an iPhone X, when a full screen component wrapped in a `SafeAreaView` renders correctly on the root view of the application, it should also render correctly when used inside of a `Modal`.

### Actual Behavior

On an iPhone X, a full screen component wrapped in a `SafeAreaView` renders correctly on the root view of the application **but** it does not render correctly when used inside of a `Modal`.

Specifically, the HEADER margin is present but the FOOTER margin is missing. The full screen component goes to the bottom edge of the iPhone X (simulator).

### Steps to Reproduce

1. Run the Snack on an iPhone X simulator
2. Find `render()` and note the two options (one with modal, one not)
3. Comment the options appropriately and see the issue
4. Remember, you need to run this on an iPhone X to see the problem

https://snack.expo.io/@lutherbaker/modal---safeareaview

LutherBaker:
_`SafeAreaView` bottom margin is not being honored in a `Modal`_

When components wrapped in `SafeAreaView` are rendered to the root view:

<img width="320" alt="iphonex-root" src="https://user-images.githubusercontent.com/144541/36939387-abcecc74-1ef5-11e8-9e3f-bf3da0bf4d1e.png">

When (the same) components wrapped in `SafeAreaView` are provided to a `Modal`:

<img width="334" alt="iphonex-modal" src="https://user-images.githubusercontent.com/144541/36939396-d0fd21da-1ef5-11e8-877c-fd8d24b67511.png">

LutherBaker:
I upgraded to React Native 54.0 but the provided snack demonstration still holds. The bottom safe area margin is not honored within a `Modal`.

Environment:
  OS: macOS Sierra 10.12.6
  Node: 9.7.1
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.7.0
[...]

PR #23501, opened 2019-02-17
support SafeAreaView bottom margin

elelad:
Fix for issue: https://github.com/facebook/react-native/issues/18177

Modal component does not support SafeAreaView bottom margin

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -19,6 +19,7 @@ const React = require('React');
 const PropTypes = require('prop-types');
 const StyleSheet = require('StyleSheet');
 const View = require('View');
+const SafeAreaView = require('SafeAreaView');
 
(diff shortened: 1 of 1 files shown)


fsfhiyg:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

fsfhiyg:
https://github.com/facebook/react-native/pull/23501#issuecomment-464498339

cpojer:
It doesn't seem like that is the right fix and it is a breaking change as well. Your PR also doesn't conform to the template, and there is no explanation about why this fixes the issue so I'm gonna close this PR for now. Feel free to make a similar modification in your project.

elelad:
You're right, this is not the correct approach to solving the issue, I will try another way.

PR #23510, opened 2019-02-18
feat(permissions-android): export Rationale type and add missing properties according to docs

n1ru4l:
## Summary

I checked the documentation over at https://facebook.github.io/react-native/docs/permissionsandroid and the documented properties `buttonPositive`, `buttonNegative` and `buttonNeutral` are not available in the flow-type definitions. 

___

Also the Rationale type is not exported which makes it hard to reuse it in a library or in your own application code.

However I do not know if it is actually intended to import "internal" flow-types from `react-native` since I could not find any other type or interface being exported. So I am not 100% sure if I should have done this.

## Changelog

[General] [Added] - Export Rationale flow-type and add missing properties `buttonPositive`, `buttonNegative` and `buttonNeutral` to the documentation.

## Test Plan

run flow

### DIFF
--- a/Libraries/PermissionsAndroid/PermissionsAndroid.js
+++ b/Libraries/PermissionsAndroid/PermissionsAndroid.js
@@ -12,9 +12,12 @@
 
 const NativeModules = require('NativeModules');
 
-type Rationale = {
+export type Rationale = {
   title: string,
   message: string,
+  buttonPositive: string,
+  buttonNegative?: string,
+  buttonNeutral?: string,
 };
 
 type PermissionStatus = 'granted' | 'denied' | 'never_ask_again';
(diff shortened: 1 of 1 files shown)


n1ru4l:
@cpojer Could you please add a comment on:

> However I do not know if it is actually intended to import "internal" flow-types from `react-native` since I could not find any other type or interface being exported. So I am not 100% sure if I should have done this.

Also in the documentation `buttonPositive` is marked as required. However when adding it as required to the flow types the flow tests failed (because none had it defined). I think it should be be marked as optional in the documentation?

cpojer:
Yeah, it probably requires a docs fix.

Feel free to import flow types from modules in RN. The story here isn't entirely figured out, so be prepared to change your flow imports from time to time as we improve the situation.

PR #23514, opened 2019-02-18
enable symbolicate for sectioned sourcemaps

mbardauskas:
## Summary

`react-native-symbolicate` usage output claim to work with 4 arguments, though the code supports 3. Furthermore, `Symbolication.parseFileName` is getting a lineNumber argument instead of fileName. Described issues disable usage of sectioned source map symbolication, which was discussed in https://github.com/getsentry/sentry-cli/issues/408.

## Changelog

[General] [Fixed] - enable symbolicate for sectioned/indexed/segmented source maps

## Test Plan

I've tried adding a test, but I couldn't make the suite run. Given that this package was just released, I would assume that there is still some implicit knowledge to running tests for it, which I'm not aware of. If so, could you please guide me how I could do it?
Also, I'm not entirely sure how to construct a sectioned source map for the test, thus I've used a react-native ram-bundle and stripped what I could. Any tips how I could improve it?


cpojer:
@mbardauskas I turned the existing tests into JavaScript tests here: https://github.com/facebook/react-native/blob/master/packages/react-native-symbolicate/__tests__/symbolicate-test.js

You should be able to run them just like any other Jest test in the repo. Please let me know if that unblocks you and whether that will help you iterate on this PR.

mbardauskas:
Sorry for the whole mess.
I'll create a new PR with the changes

PR #23484, opened 2019-02-15
[Lean Core] Removes SwipeableListView

ericlewis:
## Summary

As part of the effort in #23313.

## Changelog

[General] [Removed] - Remove SwipeableFlatList & SwipeableListView

## Test Plan

The relevant tests were removed.

PR #23517, opened 2019-02-18
Refactor CameraRollView to not rely on ListView

ericlewis:
## Summary

CameraRollView was the last remaining code in RN to use ListView, this refactor changes it to FlatList.

## Changelog

[General] [Fixed] - CameraRollView in RNTester uses FlatList instead of ListView

## Test Plan

Run and test that it still works.


ericlewis:
@TheSavior as requested! This keeps the existing functionality the same.

ericlewis:
@cpojer I will open a PR to remove ListView after this and #23484 land!

ericlewis:
@cpojer anything I need to do?

cpojer:
Nope, sorry it's taking a while, I'm just catching up on work stuff now and I need to actually go through and test this change ;) You could help me out by posting a before and after video of CameraRoll in RNTester, though :D

ericlewis:
@cpojer here ya go, the original & the refactor.

OG: https://youtu.be/ftP5mMBxCTw
Updated: https://youtu.be/oGnq8HNkM_Q

cpojer:
Hey @ericlewis do you mind giving this another look? I tested your checkout locally and I'm getting these issues with your PR:
* My local images are duplicated, causing a duplicated React key warning.
* After loading images from the All section I cannot switch to other sections using the slider.

cpojer:
Note I had no images in categories other than "All" when I did my testing.

ericlewis:
If testing on simulator you won't have images in other categories, unless you added new images, so that is expected. Will address other feedback. ty.

cpojer:
With the current version of master, moving the slider will remove the content in the list though. With your PR, it will not remove the images that are already in the list.

RNTester/js/CameraRollView.js
```diff
@@ -238,10 +201,15 @@ class CameraRollView extends React.Component<Props, State> {
...
-        groupByEveryN(newState.assets, this.props.imagesPerRow),
-      );
+
+      // Dedup assets we want to append
+      newState.assets = this.state.assets
+        .concat(assets)
+        .filter(
+          (a1, index, self) =>
+            index ===
+            self.findIndex(a2 => a1.node.image.uri === a2.node.image.uri),
```
cpojer: This algorithm is O(n*n). Can you instead do `Array.from(new Set(this.state.assets.concat(assets))`?

RNTester/js/CameraRollView.js
```diff
@@ -74,10 +58,9 @@ type Props = $ReadOnly<{|
   renderImage: PhotoIdentifier => React.Node,
 
   /**
-   * imagesPerRow: Number of images to be shown in each row.
+   * A boolean that indicates if we should render large or small images.
    */
-
-  imagesPerRow: number,
```
cpojer: We actually depend on this at Facebook. Does FlatList not support this? I'm not sure if we can merge this change without this feature.

ericlewis: No, not easily.. let me refactor this one more time to keep imagesPerRow & resolve the dupe issue properly.

ericlewis:
It looks to be working as it should now, the complexity is brought down to O(n). I’m not sure what the complexity is for grouping, I reused the existing functionality that is under utilities.

PR #23520, opened 2019-02-18
[Security] Update ws dependency

ericlewis:
## Summary

Our `ws` dependency is super outdated, and is [insecure](https://www.npmjs.com/advisories/550). It is used for the websocket example code in RNTester. This PR updates the dependency, and removes undefined console.logs.

## Changelog

[General] [Security] - Updates ws dependency to 6.4.1

## Test Plan

run RNTester & ensure works, check integration test too.

### DIFF
--- a/IntegrationTests/websocket_integration_test_server.js
+++ b/IntegrationTests/websocket_integration_test_server.js
@@ -34,7 +34,6 @@ server.on('connection', ws => {
       console.log('WebSocket integration test server exit');
       process.exit(0);
     }
-    console.log('Cookie:', ws.upgradeReq.headers.cookie);
     ws.send(message + '_response');
   });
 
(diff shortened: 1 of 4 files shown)


ericlewis:
And more specifically addresses this comment: https://github.com/react-native-community/discussions-and-proposals/issues/64#issuecomment-444822438

radeno:
CLI, Metro and Detox depends on ^1.1.x.

They should be also updated. Doesn't they?
Multiple different versions makes bundle bloated, because every package have own dependencies versions bundled within.

When will be WS updated for RN it should be updated for CLI and Metro also. But Detox is little bit complicated because there was initiation for update, but it breaks theirs Jenkins CI.

ericlewis:
I honestly think that we should not even have `ws` as a dependency in this repo for the only purpose of serving up a demo in RNTester, seems a little silly. Could we perhaps just remove instead?

ericlewis:
I would also be happy to find another way to deal with the websockets, it’s a very small surface area.

radeno:
@ericlewis i agree with you. I also think if it is not possible it should be moved into `devDependencies` because it is not necessary in production. Or is it required for CI with `--production` flag?

PR #23524, opened 2019-02-18
Toggle secureTextEntry cursor spacing

ericlewis:
## Summary

This is a fix for #5859, based on the feedback in #18587. Instead of using `didSetProps` it uses a setter. I will also note that setting to `nil` no longer works (crashes) so setting it to a blank string then back to the original works fine.

## Changelog

[iOS] [Fixed] - Toggling secureTextEntry correctly places cursor.

## Test Plan

Toggle the isSecureEntry, running on iOS 11.4 or lower.


cpojer:
Added a test to RNTester and verified this doesn't change behavior on iOS 12 and it fixes the issue on iOS 11.

Issue #17118, opened 2017-12-08
App crash with error "index=4 count=3" using LayoutAnimation.easeInEaseOut() on Android


jay1337:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

This bug has been there for several months and is still there with react-native@0.51.0.

Environment:
  OS: Windows 10
  Node: 8.9.1
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Version  3.0.0.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.2.0 => 16.2.0
  react-native: 0.51.0 => 0.51.0

### Steps to Reproduce

On Android, updating a specific (but very simple) component tree after calling LayoutAnimation.easeInEaseOut() leads to an error.
Here is a repro : https://snack.expo.io/H1H0AbDZM
Just click 2 times on the top text component to see the app crash.

### Expected Behavior

No app crash.

### Actual Behavior

Component are not correctly rendered after first click (some are missing).
App crash after second click.:

```
index=4 count=3
addInArray
    ViewGroup.java:3971
addViewInner
    ViewGroup.java:3902
addView
    ViewGroup.java:3733
addView
    ViewGroup.java:3678
addView
    ViewGroupManager.java:47
manageChildren
    NativeViewHierarchyManager.java:401
execute
    UIViewOperationQueue.java:180
run
    UIViewOperationQueue.java:819
flushPendingBatches
    UIViewOperationQueue.java:926
access$2100
    UIViewOperationQueue.java:47
doFrameGuarded
    UIViewOperationQueue.java:986
doFrame
    GuardedFrameCallback.java:31
doFrame
    ReactChoreographer.java:136
doFrame
    ChoreographerCompat.java:107
run
    Choreographer.java:765
doCallbacks
    Choreographer.java:580
doFrame
    Choreographer.java:549
run
    Choreographer.java:753
handleCallback
    Handler.java:739
dispatchMessage
    Handler.java:95
loop
    Looper.java:135
main
    ActivityThread.java:5221
invoke
    Method.java
invoke
    Method.java:372
run
    ZygoteInit.java:899
main
    ZygoteInit.java:694
```

### Reproducible Demo

Cf Steps to Reproduce

jay1337:
Hey !

DanGDroid:
[...]

PR #22776, opened 2018-12-24
Fix the IndexOutOfBoundsException when setChildren/manageChildren an illegal child ReactRawTextShadowNode to a parent non-ReactBaseTextShadowNode.

raphael-liu:
## Summary 

Hi, @cpojer , would you please help me reviewing the pull request? Thank you.

* Bad Case:
```
    <View>
       123
       '456'
       <Text>{"setChildren/manageChildren"}</Text>
    </View>
```

* Root Cause:
image

* Solution:
Drop the illegal children node(s) and correct the indices at which other children views to add when setChildren/manageChildren.

* Issues: #18773 #17118 #17113 #16411 #1211 #924

## Changelog:

[Android] [Fixed] -  Fix the IndexOutOfBoundsException when setChildren/manageChildren an illegal child ReactRawTextShadowNode to a parent non-ReactBaseTextShadowNode.

## Test Plan:
----------
See UIImplementationTest.java ( Tests for UIImplementation ).

PR #23530, opened 2019-02-19
[Android][Fixed] - Fix the IndexOutOfBoundsException when wrong tag is used in manageChildren because of animation.

zhangmincons:
Fix the IndexOutOfBoundsException when wrong tag is used in manageChildren because of animation.

## Summary

When using animation, some views may not be removed immediately. Instead, they will be removed when the animation is finished. This can cause incorrect index or even crashes. 

- Bad Case:
``` 
import { Text, View, UIManager, LayoutAnimation } from 'react-native';

class Screen extends Component {
  showB = true;
  toggle = () => {
    this.showB = !this.showB;
    this.forceUpdate();
  };

  componentWillUpdate(/* A garder : nextProps , nextState */) {
    LayoutAnimation.easeInEaseOut();
  }

  render() {
    return (
    <View style={{flex: 1, backgroundColor: '#F5FCFF'}}>
      <View style={{ margin: 40 }}>
        <Text onPress={this.toggle}>Click 2 times here to make app crash</Text>
        {
          [
            <View key={0}>
              <Text>A</Text>
              {
                this.showB ?
                  <Text>B</Text>
                  :
                  null
              }
            </View>,
            <View key={1}>
              <Text>A</Text>
              {
                this.showB ?
                  <Text>B</Text>
                  :
                  null
              }
            </View>,
          ]
        }
      </View>
    </View>
    );
  }
}

export default Screen;
```

- Root Cause:
image

1. Index 2(the first B) is supposed to be removed, but it is not removed immediately because of animation. 
2. Index 3(the second B) is supposed to be removed then, but the first B has  not be removed yet. Therefore the second A is removed by mistake.

- Solution:
A mark is used to adjust the index. Using an delete-mark to represent the view going to be deleted can help find out the correct index when the animation finish and therefore fix the problem. In mDeleteMark, Integer(0) represents a normal view while Integer(1) represents the view that should be removed but because of animation it should be removed later. The true index can be gotten by calculating the number of Integer(1).

- Issue:#17118

## Changelog

[Android][Fixed] - Fix the IndexOutOfBoundsException when wrong tag is used in manageChildren because of animation.

## Test Plan

- Open the Android emulator or an Android device.
- Create a react-native-app with the bad case code above.
- Tap the button and see if it will crash.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
@@ -68,6 +68,16 @@ public abstract class ViewGroupManager <T extends ViewGroup>
     return parent.getChildAt(index);
   }
 
+  /**
+   * Use the method when a view should be removed by layout animation
+   * Note that this is only used in ReactViewGroup at the moment
+   *
+   * @param parent the parent ViewGroup
+   * @param index the index that should attatch a delete mark
+   */
+  public void addDeleteMark(T parent, int index) {
+  }
+
   public void removeViewAt(T parent, int index) {
     parent.removeViewAt(index);
   }
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
@@ -44,6 +44,7 @@ import com.facebook.react.uimanager.RootViewUtil;
 import com.facebook.react.uimanager.ViewGroupDrawingOrderHelper;
 import com.facebook.react.uimanager.ViewProps;
 import com.facebook.yoga.YogaConstants;
+import java.util.ArrayList;
 import javax.annotation.Nullable;
 
 import static com.facebook.react.common.ReactConstants.TAG;
@@ -61,6 +62,11 @@ public class ReactViewGroup extends ViewGroup implements
   private static final LayoutParams sDefaultLayoutParam = new ViewGroup.LayoutParams(0, 0);
   /* should only be used in {@link #updateClippingToRect} */
   private static final Rect sHelperRect = new Rect();
+  /* should only be used when using layout animation. */
+  public @Nullable ArrayList<Integer> mDeleteMark = null;
+  /* -1 represents the delete mark is not enabled. */
+  public int mMarkedChildCount = -1;
+
 
   /**
    * This listener will be set for child views when removeClippedSubview property is enabled. When
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
@@ -26,6 +26,7 @@ import com.facebook.react.uimanager.ViewProps;
 import com.facebook.react.uimanager.annotations.ReactProp;
 import com.facebook.react.uimanager.annotations.ReactPropGroup;
 import com.facebook.yoga.YogaConstants;
+import java.util.ArrayList;
 import java.util.Locale;
 import java.util.Map;
 import javax.annotation.Nullable;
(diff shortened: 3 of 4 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -260,18 +260,52 @@ public void receiveCommand(ReactViewGroup root, int commandId, @Nullable Readabl
...
     }
   }
 
   @Override
   public int getChildCount(ReactViewGroup parent) {
+    boolean removeClippedSubviews = parent.getRemoveClippedSubviews();
+    if (removeClippedSubviews) {
+      return parent.getAllChildrenCount();
+    } else {
+      return parent.getChildCount() - getMarkedViewCount(parent);
```
mdvacca: the usage of getMarkedViewCount() here makes getChildCount() inefficient, now we iterate for every "marked child", can we avoid this?

ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java
```diff
@@ -61,6 +62,7 @@
   private static final LayoutParams sDefaultLayoutParam = new ViewGroup.LayoutParams(0, 0);
   /* should only be used in {@link #updateClippingToRect} */
   private static final Rect sHelperRect = new Rect();
+  public @Nullable ArrayList<Integer> mDeleteMark = new ArrayList<Integer>();
```
mdvacca: creating an empty ArrayList for every ReactViewGroup is quite expensive, specially because this is not going to be used in most of the cases
Also this variable is not @Nullable

ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
```diff
@@ -64,25 +64,36 @@ public int getChildCount(T parent) {
     return parent.getChildCount();
   }
 
+  public int getUnmarkedChildCount(T parent) {
```
mdvacca: can you add documentation about what is the purpose of these new methods?

Also let's revisit the names, you are mixing View with Child, e.g. markView vs getMakedChildAt

ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewGroupManager.java
```diff
@@ -64,25 +64,36 @@ public int getChildCount(T parent) {
...
+    return 0;
+  }
+
   public View getChildAt(T parent, int index) {
     return parent.getChildAt(index);
   }
 
+  public View getUnmarkedChildAt(T parent, int index) {
```
mdvacca: is this method @Nullable?

ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -260,18 +260,52 @@ public void receiveCommand(ReactViewGroup root, int commandId, @Nullable Readabl
...
+  }
+
   @Override
   public void addView(ReactViewGroup parent, View child, int index) {
     boolean removeClippedSubviews = parent.getRemoveClippedSubviews();
     if (removeClippedSubviews) {
       parent.addViewWithSubviewClippingEnabled(child, index);
     } else {
-      parent.addView(child, index);
+      parent.mDeleteMark.add(index + getPrefixMarkedViewCount(parent, index),
```
mdvacca: the usage of getPrefixMarkedViewCount() here makes addView() inefficient, now we iterate for every "marked child", and we do this twice. can we avoid this?

ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -290,6 +334,13 @@ public View getChildAt(ReactViewGroup parent, int index) {
     }
   }
 
+  @Override
+  public void markView(ReactViewGroup parent, int index) {
+    parent.mDeleteMark.remove(index + getPrefixMarkedViewCount(parent, index));
+    parent.mDeleteMark.add(index + getPrefixMarkedViewCount(parent, index),
```
mdvacca: The usage of mDeleteMark is quite inefficient here.

Every time you remove an element from a specific index of the ArrayList, you will shift to the left all the elements on the right of the element. Then you will add the element and shift all the same elements but this time to the right. 
See  https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#remove-int- 
and https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#add-int-E-

can you choose a different data-structure?

todorone:
@zhangmincons @mdvacca Isn't this PR already trying to fix the same problem?
https://github.com/facebook/react-native/pull/22776

zhangmincons:
> @zhangmincons @mdvacca Isn't this PR already trying to fix the same problem?
> #22776

Nope, I've read that PR, and it fixed a different problem.

zhangmincons:
> Hey @zhangmincons, thanks for working on this. I added few comments
> Moving the PR back to you to review my comments.

Thank you, I will read the comments and fix the problems.

zhangmincons:
hi, @mdvacca, the problems are solved. Can you review the code again? Thanks!

RWOverdijk:
Is there any progress on this?

lunaleaps:
Hi @zhangmincons this was fixed internally here:
https://github.com/facebook/react-native/commit/20b4879dfd34716997e92bb5d89b0e4f873fef98

Apologies about the lack of synchronization on this!

cpojer:
Based on @lunaleaps's comment, I'll close this PR here since the underlying problem should already have been fixed.

PR #23541, opened 2019-02-19
Fix testScrollViewExample flakiness

ericlewis:
## Summary

Sometimes images are slow to load, and they are quite heavy. This converts the ScrollViewExample to a list of simple text items, similar to the append example inside of it. 

What it looks like: https://i.imgur.com/jt083Iv.png

## Changelog

[iOS] [Fixed] - Fix testScrollViewExample flakiness

## Test Plan

Run the tests.


ericlewis:
@hramos it appears to be fixed.

PR #23548, opened 2019-02-20
fix getPhotos params on RNTester XHR example.

ericlewis:
## Summary

Fixes redscreen on RNTester's XHR page. CameraRoll getPhotos was missing params.

ss: https://i.imgur.com/JLThyrF.png

## Changelog

[iOS] [Fixed] - RNTester XHR Example crash

## Test Plan

Run RN tester

PR #23558, opened 2019-02-20
Update iOS snapshots

hramos:
## Summary

Re-generated iOS snapshots. The main difference appears to be in the font size for the "Search..." text in the search box at the top.

screen shot 2019-02-20 at 7 47 54 am

## Changelog

[iOS] [Changed] - SnapshotTests: Updated iOS snapshots

## Test Plan

Ran Xcode tests in record mode.


hramos:
Duplicate of #23548 23548

PR #23569, opened 2019-02-21
[iOS] Remove dispatch_queue nil operation when invalidate module data

### LINKED ISSUE #23564: crash in bridge code with some native module.
## 🐛 Bug Report

Crash in bridge code in Release build only. Works in Debug

```
void RCTNativeModule::invoke(unsigned int methodId, folly::dynamic &&params, int callId) {
  // capture by weak pointer so that we can safely use these variables in a callback
  __weak RCTBridge *weakBridge = m_bridge;
  __weak RCTModuleData *weakModuleData = m_moduleData;
  // The BatchedBridge version of this buckets all the callbacks by thread, and
  // queues one block on each.  This is much simpler; we'll see how it goes and
  // iterate.
  dispatch_block_t block = [weakBridge, weakModuleData, methodId, params=std::move(params), callId] {
    #ifdef WITH_FBSYSTRACE
    if (callId != -1) {
      fbsystrace_end_async_flow(TRACE_TAG_REACT_APPS, "native", callId);
    }
    #endif
    invokeInner(weakBridge, weakModuleData, methodId, std::move(params));
  };

  if (m_bridge.valid) {
    dispatch_queue_t queue = m_moduleData.methodQueue;
    if (queue == RCTJSThread) {
      block();
    } else if (queue) {
      dispatch_async(queue, block); <----- com.facebook.react.JavaScript (8): EXC_BAD_ACCESS (code=2, address=0x1c20a79b8)
    }
  } else {
    RCTLogWarn(@"Attempted to invoke `%u` (method ID) on `%@` (NativeModule name) with an invalid bridge.",
               methodId, m_moduleData.name);
  }
}
```
com.facebook.react.JavaScript (8) /!\ This thread has caused the crash

## To Reproduce
I'm using several native modules so, deciding which one is faulty is complicated for me.
I'm using Cocoapods, following is my pod file:

```
platform :ios, '9.0'
	
use_frameworks!
	
target 'MyTarget' do
    pod 'yoga', path: '../node_modules/react-native/ReactCommon/yoga/'
    pod 'React', path: '../node_modules/react-native/', :subspecs => [
    'Core',
    'ART',
    'RCTActionSheet',
    'RCTAnimation',
    'RCTLinkingIOS',
    'RCTGeolocation',
    'RCTImage',
    'RCTNetwork',
    'RCTText',
    'RCTVibration',
    'RCTWebSocket',
    'DevSupport',
    'CxxBridge',
    ]
    
    pod 'DoubleConversion', :podspec => '../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec'
    pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
    pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
    
    pod 'RNCharts', :path => '../node_modules/react-native-charts-wrapper'
    pod 'RNSVG', :path => '../node_modules/react-native-svg'

end
	
swift4 = ['Charts']

post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      if swift4.include?(target.name)
        config.build_settings['SWIFT_VERSION'] = '4.1'
        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= []
        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] << 'DEBUG=1'
      end
    end
  end
end
```

An the result of the install:

```
Oliviers-iMac:ios omatrot$ pod install
Analyzing dependencies
Fetching podspec for `DoubleConversion` from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`
Fetching podspec for `Folly` from `../node_modules/react-native/third-party-podspecs/Folly.podspec`
Fetching podspec for `RNCharts` from `../node_modules/react-native-charts-wrapper`
Fetching podspec for `RNSVG` from `../node_modules/react-native-svg`
Fetching podspec for `React` from `../node_modules/react-native/`
Fetching podspec for `glog` from `../node_modules/react-native/third-party-podspecs/glog.podspec`
Fetching podspec for `yoga` from `../node_modules/react-native/ReactCommon/yoga/`
Downloading dependencies
Installing Charts (3.1.1)
Installing DoubleConversion (1.1.6)
Installing Folly (2016.10.31.00)
Using RNCharts (0.5.0)
Using RNSVG (9.2.4)
Using React (0.57.4)
Installing SwiftyJSON (4.1.0)
Installing boost-for-react-native (1.63.0)
Installing glog (0.3.5)
Using yoga (0.57.4.React)
Generating Pods project
Integrating client project

[!] Please close any current Xcode sessions and use `MyHeartSigns.xcworkspace` for this project from now on.
Sending stats
Pod installation complete! There are 19 dependencies from the Podfile and 10 total pods installed.
```

## Expected Behavior
No crash at all.

## Environment

System:
OS: macOS 10.14.3
CPU: (4) x64 Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
Memory: 42.34 MB / 8.00 GB
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 8.11.2 - /usr/local/bin/node
SDKs:
iOS SDK:
Android SDK:
API Levels: 14, 23, 24, 25, 26, 27, 28
Build Tools: 23.0.1, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.2, 27.0.3, 28.0.0, 28.0.2, 28.0.3
IDEs:
npmPackages:
react: ^16.8.3 => 16.8.4
react-native: ^0.59.0 => 0.59.0
npmGlobalPackages:
react-native-cli: 2.0.1
react-native-rename: 2.2.2
react-native: 0.59.0

zhongwuzw:
## Summary

Try to fix #23564 , previously, we would nil the dispatch_queue when invalidate `Bridge`, seems it has race condition when get and set concurrency, leads `queue` becomes dangling pointer and throw `EXC_BAD_ACCESS` exception. Now, we let `RCTModuleData` nil it in dealloc.

## Changelog

[iOS] [Fixed] - Remove dispatch_queue nil operation when invalidate module data

## Test Plan

Race condition seems hard to reproduce, we just need to ensure this would not break currently behavior.

### DIFF
--- a/React/Base/RCTModuleData.mm
+++ b/React/Base/RCTModuleData.mm
@@ -338,7 +338,6 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init);
 
 - (void)invalidate
 {
-  _methodQueue = nil;
 }
 
 - (NSString *)description


cpojer:
Why does this fix the issue?

zhongwuzw:
@cpojer Thanks for review, from the crash log, I see it crashed at `dispatch_async(queue, block); `, which means at lease `queue` is not `nil` (PS. we have nil check before it), so `queue` is a dangling pointer, and we only call setter of `queue` in `invalidate` method after module setup, based on these infos, I think the reason is the concurrency of `invalidate`(setter) and `m_moduleData.methodQueue`(getter), so I remove the `nil` statement, actually, `RCTModuleData` can nil it automatically in `dealloc` method, we can see in `invalidate` of `RCTCxxBridge      https://github.com/facebook/react-native/blob/master/React/CxxBridge/RCTCxxBridge.mm#L1083 , we would dealloc everything, we don't need to nil `queue` explicitly.

```
  if (m_bridge.valid) {
    dispatch_queue_t queue = m_moduleData.methodQueue;
    if (queue == RCTJSThread) {
      block();
    } else if (queue) {   
      dispatch_async(queue, block); <----- com.facebook.react.JavaScript (8): EXC_BAD_ACCESS (code=2, address=0x1c20a79b8)
    }
  } else {
    RCTLogWarn(@"Attempted to invoke `%u` (method ID) on `%@` (NativeModule name) with an invalid bridge.",
               methodId, m_moduleData.name);
  }
```

aleclarson:
@zhongwuzw The `if (queue)` condition fails after `_methodQueue = nil`?
Try changing it to `if (queue != nullptr)` instead. Note: The `methodQueue` getter throws in development when `_methodQueue` does not exist.

zhongwuzw:
@aleclarson What's the difference between `if (queue != nullptr)` and `if (queue)`? I think they are the same 🤔 .

```
#ifndef nil
# if __has_feature(cxx_nullptr)
#   define nil nullptr
# else
#   define nil __DARWIN_NULL
# endif
#endif
```

aleclarson:
I only mentioned `nullptr` because that's what the `methodQueue` getter uses:
https://github.com/facebook/react-native/blob/272622a0c8b84780cc450045e5ea49af8c9092cb/React/Base/RCTModuleData.mm#L334-L335
Are you sure it's possible for a nil `dispatch_queue_t` to pass an `if (queue)` condition?

zhongwuzw:
@aleclarson From the crash log #23564 , it's not a `nullptr`, it's a zombie ptr, it's also the `EXC_BAD_ACCESS` happened.

aleclarson:
How can the `methodQueue` getter return a zombie pointer? When the local `queue` variable is non-nil, we can assume that `m_moduleData` has a strong reference to it, which means the queue hasn't been released yet. Even if `invalidate` is called on `m_moduleData` between setting the local `queue` variable and calling `dispatch_async`, doesn't the local variable ensure the queue is not released?

zhongwuzw:
> _edit:_ Could this be related? https://stackoverflow.com/a/43715625/2228559

@aleclarson Emm, maybe not 🤔 , because this crash is in `Release` mode, may not related to `libBacktraceRecording.dylib` as `stackoverflow` said.

zhongwuzw:
> How can the `methodQueue` getter return a zombie pointer? When the local `queue` variable is non-nil, we can assume that `m_moduleData` has a strong reference to it, which means the queue hasn't been released yet. Even if `invalidate` is called on `m_moduleData` between setting the local `queue` variable and calling `dispatch_async`, doesn't the local variable ensure the queue is not released?

I think the issue happened at this line `dispatch_queue_t queue = m_moduleData.methodQueue;`  and [`_methodQueue = nil;`](https://github.com/facebook/react-native/pull/23569/files#diff-f0ca0436192a5821de3a18ea4534b066L341), thus read and write concurrently. (PS. I'm not entirely sure, but I think it's the only place can happened 🤔 )

Put aside this crash issue, remove `nil` operation is also ok IMO, because `dealloc` would `nil` it directly after called `invalidate`.

[`(void)[self instance];`](https://github.com/facebook/react-native/pull/23569/files#diff-f0ca0436192a5821de3a18ea4534b066L333) has write operation but it's fine because it has a lock to guard.

aleclarson:
Both the `RCTModuleData::invalidate` and `RCTNativeModule::invoke` methods are only ever called from the JS thread, so I'm not seeing how a race condition is possible there.

https://github.com/facebook/react-native/blob/58c956768af75047b2acdca429a28945a6a8b8c0/React/CxxBridge/RCTCxxBridge.mm#L1076

https://github.com/facebook/react-native/blob/58c956768af75047b2acdca429a28945a6a8b8c0/React/CxxModule/RCTNativeModule.mm#L76-L77

But I agree that the `RCTModuleData::invalidate` method doesn't seem necessary. It should be removed altogether if you're going to remove its only statement.

zhongwuzw:
@aleclarson Yes, seems the `RN` internal call these on `JS` thread, from the principle of crash analysis, I still think it's because the read/write concurrency. Maybe the third parties write the `methodQueue` in some cases, seems he uses third parties, what's your opinion? 🤔

aleclarson:
He's only using [`react-native-svg`](https://github.com/react-native-community/react-native-svg) and [`react-native-charts-wrapper`](https://github.com/wuxudong/react-native-charts-wrapper), and neither of those have `methodQueue` anywhere in their codebases. Probably best to not change anything until we can get a solid repro.

zhongwuzw:
@aleclarson Yeah, we need to keep investigate.

cpojer:
Will close this for now, if you find a proper fix please send a new PR.

Issue #23561, opened 2019-02-20
☂️✅ Green CI: Re-enabling disabled tests


hramos:
We have a handful of tests that have been disabled on Circle CI over the years. It's time to get these back up and running so we can merge PRs and cut new releases with confidence.

This is an _umbrella_ task intended to track all the work necessary to re-enable these tests. This issue is being kept up to date so new items will be added and completed ones will be (re)moved. Some of the items will spawn separate issues to track work. If you don't see a name assigned to any of the below items and you would like to help out, please volunteer in the comments and reach out to me to get started!

## What needs to be re-enabled?

These tests are configured as aliases in the [Circle CI config](https://github.com/facebook/react-native/blob/master/.circleci/config.yml). I'm happy to help with the Circle CI side of things, but for now you can focus on making sure the command/script (e.g. `./scripts/run-ci-e2e-tests.js --ios`) succeeds locally.

| Circle CI Job | Test Name | Test Script | Fix? |
| - | - | - |  - |
| test_android | Android e2e | node ./scripts/run-ci-e2e-tests.js --android | |
| test_ios | iOS e2e | node ./scripts/run-ci-e2e-tests.js --ios | |
| test_js | JS e2e | node ./scripts/run-ci-e2e-tests.js --js | https://github.com/facebook/react-native/pull/28392 |
| test_ios | Podspecs | ./scripts/process-podspecs.sh | |

ericlewis:
> * CocoaPods Podspecs tests

@hramos is there more detail on what this should do?

hramos:
I haven’t looked into it recently. There’s a PR out right now that reworks how CocoaPods is supported in this repo, so it’s an open question whether this test is still necessary or sufficient. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gregoryfm:
@hramos I'd like to help with this  `./scripts/run-ci-e2e-tests.js --js`

hramos:
@gregoryfm it looks like @ericlewis is already taking a stab at this, through https://github.com/facebook/react-native/pull/23571.

ericlewis:
Any help with android e2e would be much appreciated tho!

ericlewis:
@hramos I think the iOS portion is complete too, it’s e2e test consists of setting up a project via `react-native init` and confirming that we are indeed using the master branch template & that the snapshot lines up. This should indicate a successful react native init on iOS.

ericlewis:
[...]

PR #23571, opened 2019-02-21
Enable js e2e tests

ericlewis:
## Summary

Part of #23561. Small refactor to use 1 marker for the different tests, as it doesn't matter where it is- so long as we can detect it.

## Changelog

[General] [Fixed] - turn on JS e2e tests

## Test Plan

run `node ./scripts/run-ci-e2e-tests.js --js`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,6 +249,10 @@ aliases:
     name: JavaScript End-to-End Test Suite
     command: node ./scripts/run-ci-e2e-tests.js --js --retries 3;
 
+  - &run-android-e2e-tests
+    name: Android End-to-End Test Suite
+    command: node ./scripts/run-ci-e2e-tests.js --android --retries 3;
+
   - &display-broken-tests-warning
     name: Running broken tests (Ignore any failures past this point)
     command: |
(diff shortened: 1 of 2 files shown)


ericlewis:
I removed the flow checks here, as they occur earlier on too. Is it important for us to flow check here too? (Or jest?)

hramos:
Hmm, given this script is meant to run on CI, and we already run Flow check elsewhere, I think it'd be fine to drop it from this script.

hramos:
(Also wanted to note, I'm immensely grateful you're tackling this!)

hramos:
That Switch test has been super flaky in the past. Feel free to remove the entire Switch-test.js file if you need to. See https://github.com/facebook/react-native/pull/23410 where I disabled one of the Switch tests just this past month.

ericlewis:
@hramos cool, I will take a look when I get back to keyboard.

.circleci/config.yml
```diff
@@ -245,6 +245,10 @@ aliases:
     name: iOS End-to-End Test Suite
     command: node ./scripts/run-ci-e2e-tests.js --ios --retries 3;
 
+  - &run-js-e2e-tests
+    name: JavaScript End-to-End Test Suite
+    command: node ./scripts/run-ci-e2e-tests.js --js --retries 3;
+
   - &display-broken-tests-warning
```
hramos: This is unrelated to your PR but if you can tackle it while you're touching this: can you remove this alias? It's no longer used.

PR #23572, opened 2019-02-21
enable symbolicate for sectioned sourcemaps

mbardauskas:
## Summary

`react-native-symbolicate` usage output claim to work with 4 arguments, though the code supports 3. Furthermore, `Symbolication.parseFileName` is getting a lineNumber argument instead of fileName. Described issues disable usage of sectioned source map symbolication, which was discussed in https://github.com/getsentry/sentry-cli/issues/408.
Duplicate of https://github.com/facebook/react-native/pull/23514 without modified commit history

## Changelog

[General] [Fixed] - enable symbolicate for sectioned/indexed/segmented source maps

## Test Plan

I've tried adding a test, but I couldn't make the suite run. Given that this package was just released, I would assume that there is still some implicit knowledge to running tests for it, which I'm not aware of. If so, could you please guide me how I could do it?
Also, I'm not entirely sure how to construct a sectioned source map for the test, thus I've used a react-native ram-bundle and stripped what I could. Any tips how I could improve it?


packages/react-native-symbolicate/__tests__/symbolicate-test.js
```diff
@@ -50,7 +50,7 @@ const execute = (args: Array<string>, stdin: string): Promise<string> =>
 afterAll(() => {
   try {
     fs.unlinkSync(resolve('testfile.temp.cpuprofile'));
-  } catch {}
+  } catch (e) {}
```
mbardauskas: jest was throwing syntax error because of this

Issue #8237, opened 2016-06-20
Fetch adds charset=utf-8 on Android but not on iOS


dstaver:
This fetch request sends "application/json" as ContentType on IOS but "application/json; charset=utf-8" on Android:

```
fetch(url, {
    method: 'POST',
    headers: {
        'Accept': 'application/json',
        'Content-Type': 'application/json',
    }
})
```

The server I'm working against requires application/json and fails when charset is added. I've filed a bug report with them too, but shouldn't fetch behave identically on ios and android?

dstaver:
http://stackoverflow.com/questions/25560601/how-to-suppress-charset-being-automatically-added-to-content-type-in-okhttp

israelidanny:
I'm also experiencing this. Having different headers between platforms doesn't make sense and introduces bugs

lacker:
What's the right thing to do here - should the charset be appended or not?

himelnagrana:
Having the same issue while working with Urban Airship APIs. Frustrating as UA supposed to be not changing at their end. What to do?

Negan1911:
Im also seeing this

ide:
> What's the right thing to do here - should the charset be appended or not?

Looking at Chrome, it doesn't modify the content-type header at all so we shouldn't be adding a charset.

alterx:
I'm also experiencing this issue, somehow `charset=utf-8` is added as a parameter to the `Content-Type` header. It only happens on Android.

In my specific case, I'm interacting with a JSON API (http://jsonapi.org/format/) and adding media type parameters causes the server to respond with a 415 error every time. This pretty much breaks the app :(

laurent22:
I'm also having this problem and, while doing some testing, I've tried setting an invalid charset "dropbox-cors-hack" (as I'm testing the Dropbox API) so the request was like this:

    apiRequest = request.post(getBaseURL(host) + path)
      //.type('application/octet-stream;')
      .type('text/plain; charset=dropbox-cors-hack')
      .set('Authorization', 'Bearer ' + accessToken)
      .set('Dropbox-API-Arg', httpHeaderSafeJson(args));

When doing so I got the following exception:

![](https://i.imgur.com/Ii4J0t7.png)

And that lead me to this file from the okhttp library:

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RequestBody.java#L53

[...]

PR #23580, opened 2019-02-21
Prevent okhttp from adding ;charset=utf8 to ContentType Header

nhunzaker:
## Summary

Before this commit, `fetch()` calls append `"charset=utf8"` to the `Content-Type` header on Android (and not on iOS). This is because of an implementation detail in the okhttp library. This means that you can make a call on the JavaScript side like:

```javascript
let body = JSON.stringify({ key: "value" });

let headers = {
  "Content-Type": "application/json"
};

fetch("http://10.0.2.2:3000", { method: "POST", body, headers });
```

However the resulting request appends the utf8 character:

```
POST - 13:34:32:
  content-type: application/json; charset=utf-8
  content-length: 15
  host: 10.0.2.2:3000
  connection: Keep-Alive
  accept-encoding: gzip
  user-agent: okhttp/3.12.1
```

Passing byte array into the RequestBody avoids this, as recommended by a maintainer of okhttp:

https://github.com/square/okhttp/issues/2099#issuecomment-366757161

Related issues: 
https://github.com/facebook/react-native/issues/8237

## Changelog

[Android][fixed] - Prevent fetch() POST requests on Android from appending `charset=utf-8` to `Content-Type` header.

## Test Plan

I setup a test project here: https://github.com/nhunzaker/rn-charset-issue. It includes a simple node server that logs header output from requests. With this change, the server no longer reports the addition of `charset=utf-8` in the `Content-Type` header.

### DIFF
--- a/ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
+++ b/ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
@@ -342,7 +342,7 @@ public class NetworkingModuleTest {
     ArgumentCaptor<Request> argumentCaptor = ArgumentCaptor.forClass(Request.class);
     verify(httpClient).newCall(argumentCaptor.capture());
 
-    // Verify okhttp does not append "charset=utf8
+    // Verify okhttp does not append "charset=utf-8"
     assertThat(argumentCaptor.getValue().body().contentType().toString()).isEqualTo("application/json");
   }
 
--- a/ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
+++ b/ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
@@ -311,10 +311,6 @@ public class NetworkingModuleTest {
 
   @Test
   public void testPostJsonContentTypeHeader() throws Exception {
-    RCTDeviceEventEmitter emitter = mock(RCTDeviceEventEmitter.class);
-    ReactApplicationContext context = mock(ReactApplicationContext.class);
-    when(context.getJSModule(any(Class.class))).thenReturn(emitter);
-
     OkHttpClient httpClient = mock(OkHttpClient.class);
     when(httpClient.newCall(any(Request.class))).thenAnswer(new Answer<Object>() {
       @Override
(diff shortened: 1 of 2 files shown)


ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
```diff
@@ -309,6 +309,55 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
     assertThat(requestHeaders.get("User-Agent")).isEqualTo("React test agent/1.0");
   }
 
+  @Test
+  public void testPostJsonContentTypeHeader() throws Exception {
```
nhunzaker: I copied this from previous tests. It's possible more of the mocks below can be removed.

hey99xx:
This is actually a pretty simple fix for a long going problem for many 👍
Can we still specify utf8 as the charset for the bytes though?

```java
import com.facebook.react.common.StandardCharsets;

body.getBytes(StandardCharsets.UTF_8)
```

For unit tests, you're heavily mocking OkHttp. Was the test actually failing as intended when you undo your fix?

nhunzaker:
@hey99xx good deal, :+1:. I wondered about that, or passing in the character set as defined by `MediaType.parse`.

When I reverse this change, I get the following failure:

```
org.junit.ComparisonFailure: expected:<null> but was:<UTF-8>
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
```

I'd love to actually assert the headers that OkHTTP is building. This was the closest I could manage to get.

cpojer:
@nhunzaker I'm not sure I understand your last comment. Can you confirm that @hey99xx's concern is addressed?

nhunzaker:
Regarding:

> For unit tests, you're heavily mocking OkHttp. Was the test actually failing as intended when you undo your fix?

Yes, I believe I have. While the test heavily mocks OkHttp, the important part to test is [the stringification of the MediaType class](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/MediaType.java#L151), which is not mocked.

I sent out another commit that I hope simplifies the test and makes that clearer (https://github.com/facebook/react-native/pull/23580/files#diff-0a0a7fd9969f1f5f68d80bf421fcc092R346).

Without my change, this fails, returning `"application/json; charset=utf-8"`

cpojer:
@nhunzaker I mean this part: "Can we still specify utf8 as the charset for the bytes though?" – I assume the answer is yes, just wanna hear the confirmation.

nhunzaker:
@cpojer Oh! Sorry, yep!

This line should do it:
https://github.com/facebook/react-native/pull/23580/files#diff-f8703d50cf5af4c06824e2928dddb500R375

cpojer:
Yes it definitely does matter ☺️

PR #23585, opened 2019-02-22
[iOS] Fix textAttributes not applied when typing text

zhongwuzw:
## Summary

Currently, if we has `defaultValue`, textAttributes like `letterSpacing` can works, but if textinput has not default text, when we typing the text, some attributes not applied.

## Changelog

[iOS] [Fixed] - Fix textAttributes not applied when typing text

## Test Plan

We can use code like below, `letterSpacing` not work previously.
```
<TextInput
  style={{ letterSpacing: 10, color: 'red' }}
  placeholder="I will not respect letterSpacing value on iOS"
/>
```

Before:

If we have `defaultValue`, `letterSpacing` works:
```
<TextInput
  style={{ letterSpacing: 10 , color: 'red'}}
  placeholder="I will not respect letterSpacing value on iOS"
  defaultValue="s"
/>
```

After:


zhongwuzw:
@janicduplessis Thanks for review, I scan the failed tests, seems the failed snapshot tests not related to `TextInput`, they don't use this module.

So I think it needs to fix by another `PR`? and TBO 😂 , I don't familiar how to do it, maybe I need remove all snapshot images and re-generate it?

janicduplessis:
👍 I just double checked and you are right it is for other components. I think you have to use xcode to update the snapshots but I've never done it either haha.

Libraries/Text/TextInput/Multiline/RCTUITextView.m
```diff
@@ -88,6 +91,20 @@ - (void)setPlaceholderColor:(UIColor *)placeholderColor
   _placeholderView.textColor = _placeholderColor ?: defaultPlaceholderColor();
 }
 
+- (void)setReactTextAttributes:(RCTTextAttributes *)reactTextAttributes
+{
+  if (!reactTextAttributes.effectiveTextAttributes || [reactTextAttributes isEqual:_reactTextAttributes]) {
```
janicduplessis: Can you add a variable for `effectiveTextAttributes` to avoid computing it twice, looks like a pretty big getter https://github.com/facebook/react-native/blob/9d16ddf639056b2fc3ce2194e20ab658bcc0397b/Libraries/Text/RCTTextAttributes.m#L82 (I hate properties :))

Libraries/Text/TextInput/Singleline/RCTUITextField.m
```diff
@@ -62,6 +65,20 @@ - (void)setPlaceholderColor:(UIColor *)placeholderColor
   [self _updatePlaceholder];
 }
 
+- (void)setReactTextAttributes:(RCTTextAttributes *)reactTextAttributes
+{
+  if (!reactTextAttributes.effectiveTextAttributes || [reactTextAttributes isEqual:_reactTextAttributes]) {
```
janicduplessis: same here

Libraries/Text/TextInput/Multiline/RCTUITextView.m
zhongwuzw: @janicduplessis 👍 I removed `!reactTextAttributes.effectiveTextAttributes` check because it cannot be `nil`.

cpojer:
@zhongwuzw can you update the screenshot examples?

zhongwuzw:
@cpojer I don't know why here failed 🤔 , seems all other `PR`s like #23608 #23625 passes. So I don't ensure wether the code has issues, or I just re-generate the snapshots inside this `PR`? any instructions I can find to re-generate snapshots? I checked the snapshot tests code, seems I only need to remove snapshots  https://github.com/facebook/react-native/blob/86d0611c5a0e9abfa80e9d2ff9efe69241fae82e/RNTester/RNTesterIntegrationTests/RNTesterIntegrationTests.m#L39 and set `_runner.recordMode = YES`?

cpojer:
Yeah I think that's it.

janicduplessis:
@zhongwuzw Looks like tests are passing on master, maybe you can just try rebasing see if it fixes it?

zhongwuzw:
@cpojer @janicduplessis CI passes! 🍻

PR #23591, opened 2019-02-22
[standard][perf] Convert CJS requires in react-native-implementation to ES imports/exports

ide:
(Note: this depends on a Metro PR landing first in order for the iOS E2E tests to pass. Everything else works.)

## Summary

This is the first step towards converting RN to use ECMAScript modules (ESM) instead of CommonJS modules. Most of RN can be converted in a straightfoward manner, but the react-native-implementation entry point uses getters containing lazy `require` calls. This commit is one of two parts to move to JS modules while preserving the lazy require behavior. The second part is in metro-react-native-babel-preset.

This commit changes all of the `require` calls in react-native-implementation.js to `export from` statements. Some of the old `require` calls had deprecation warnings in their enclosing getters -- these warnings are now in the deprecated modules themselves. (AlertIOS is one exception since is is a deprecated alias for Alert. We could add an intermediate AlertIOS.js file that prints the warning but I opted to keep the getter for this one module since we are removing it anyway.)

The Metro side of this conversion enables the `lazy` option for `@babel/plugin-transform-modules-commonjs` when any of the RN modules are required. This means that lazy `require` calls in react-native-implementation are preserved out of the box. (And if you specify `lazyImportExportTransform: true`, all imports are lazy and there's no change in behavior.)

## Changelog

[General] [Changed] - Converted public exports from CJS to ES exports. `import ReactNative from 'react-native'` is superseded by `import * as ReactNative from 'react-native'`.

Callers that use `require('react-native')` or `import { ... } from 'react-native'` don't need any changes.

## Test Plan

Opened RNTester and verified it loaded without an error. Tapped on several of the test screens. Also got yellow box warnings for the deprecated modules, since RNTester loads them.

Locally changed RNTester to be a very simple app that just displays a string. Verified the synthetic default export still works with `require()` and that no deprecation warnings were displayed because only non-deprecated components were used and the imports are lazy:
```js
const {AppRegistry, StyleSheet, Text, View} = ReactNative;
```

Also double-checked that the Metro Babel preset preserves lazy `require` calls in react-native-implementation.js by looking at the code in the bundle:

```js
__d(function (global, _$$_REQUIRE, _$$_IMPORT_DEFAULT, _$$_IMPORT_ALL, module, exports, _dependencyMap) {
  'use strict';

  var _interopRequireDefault = _$$_REQUIRE(_dependencyMap[0], "@babel/runtime/helpers/interopRequireDefault");

  Object.defineProperty(exports, "__esModule", {
    value: true
  });
  Object.defineProperty(exports, "AccessibilityInfo", {
    enumerable: true,
    get: function get() {
      return _AccessibilityInfo().default;
    }
  });

  // ... (one defineProperty call with a deferred getter for each export)

  var _extends2 = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[1], "@babel/runtime/helpers/extends"));

  function _AccessibilityInfo() {
    var data = _interopRequireDefault(_$$_REQUIRE(_dependencyMap[2], "AccessibilityInfo"));

    _AccessibilityInfo = function _AccessibilityInfo() {
      return data;
    };

    return data;
  }

  // ... (one lazy, memoized function for each import)

  (0, _extends2.default)(exports, 'AlertIOS', {
    get: function get() {
      var warnOnce = _$$_REQUIRE(_dependencyMap[94], "warnOnce");

      warnOnce('alert-ios', 'AlertIOS is deprecated. Use the `Alert` module directly instead.');
      return _$$_REQUIRE(_dependencyMap[46], "Alert");
    }
  });
},317,[1,13,318,319,357,367,451,452,454,413,432,455,457,458,459,461,462,468,470,511,513,355,515,493,420,437,517,519,521,417,523,524,528,368,530,537,373,482,380,381,442,341,538,414,539,542,264,382,543,307,312,314,553,554,555,324,323,409,277,477,556,390,423,424,316,249,557,526,268,322,558,559,560,561,321,183,562,563,315,280,500,564,491,250,273,187,229,299,344,565,327,370,566,483,313],"Libraries/react-native/react-native-implementation.js");
```

### DIFF
--- a/Libraries/react-native/react-native-implementation.js
+++ b/Libraries/react-native/react-native-implementation.js
@@ -10,365 +10,121 @@
 
 'use strict';
 
-const invariant = require('invariant');
-const warnOnce = require('warnOnce');
+// Components
+export {default as AccessibilityInfo} from 'AccessibilityInfo';
+export {default as ActivityIndicator} from 'ActivityIndicator';
+export {default as ART} from 'ReactNativeART';
+export {default as Button} from 'Button';
+export {default as CheckBox} from 'CheckBox';
+export {default as DatePickerIOS} from 'DatePickerIOS';
+export {default as DrawerLayoutAndroid} from 'DrawerLayoutAndroid';
+export {default as FlatList} from 'FlatList';
+export {default as Image} from 'Image';
+export {default as ImageBackground} from 'ImageBackground';
+export {default as ImageEditor} from 'ImageEditor';
+export {default as ImageStore} from 'ImageStore';
+export {default as InputAccessoryView} from 'InputAccessoryView';
+export {default as KeyboardAvoidingView} from 'KeyboardAvoidingView';
+export {default as MaskedViewIOS} from 'MaskedViewIOS';
+export {default as Modal} from 'Modal';
+export {default as Picker} from 'Picker';
+export {default as PickerIOS} from 'PickerIOS';
+export {default as ProgressBarAndroid} from 'ProgressBarAndroid';
+export {default as ProgressViewIOS} from 'ProgressViewIOS';
+export {default as SafeAreaView} from 'SafeAreaView';
+export {default as ScrollView} from 'ScrollView';
+export {default as SectionList} from 'SectionList';
+export {default as SegmentedControlIOS} from 'SegmentedControlIOS';
+export {default as Slider} from 'Slider';
+export {default as Switch} from 'Switch';
+export {default as RefreshControl} from 'RefreshControl';
+export {default as StatusBar} from 'StatusBar';
+export {default as SwipeableFlatList} from 'SwipeableFlatList';
+export {default as Text} from 'Text';
+export {default as TextInput} from 'TextInput';
+export {default as ToolbarAndroid} from 'ToolbarAndroid';
+export {default as Touchable} from 'Touchable';
+export {default as TouchableHighlight} from 'TouchableHighlight';
+export {default as TouchableNativeFeedback} from 'TouchableNativeFeedback';
+export {default as TouchableOpacity} from 'TouchableOpacity';
+export {default as TouchableWithoutFeedback} from 'TouchableWithoutFeedback';
+export {default as View} from 'View';
+export {default as ViewPagerAndroid} from 'ViewPagerAndroid';
+export {default as VirtualizedList} from 'VirtualizedList';
+export {default as WebView} from 'WebView';
 
-// Export React, plus some native additions.
-module.exports = {
-  // Components
-  get AccessibilityInfo() {
-    return require('AccessibilityInfo');
-  },
-  get ActivityIndicator() {
-    return require('ActivityIndicator');
-  },
-  get ART() {
-    return require('ReactNativeART');
-  },
-  get Button() {
-    return require('Button');
-  },
-  get CheckBox() {
-    return require('CheckBox');
-  },
-  get DatePickerIOS() {
-    return require('DatePickerIOS');
-  },
-  get DrawerLayoutAndroid() {
-    return require('DrawerLayoutAndroid');
-  },
-  get FlatList() {
-    return require('FlatList');
-  },
-  get Image() {
-    return require('Image');
-  },
-  get ImageBackground() {
-    return require('ImageBackground');
-  },
-  get ImageEditor() {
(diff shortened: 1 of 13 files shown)


Libraries/react-native/react-native-implementation.js
```diff
@@ -10,367 +10,117 @@
...
+export {default as EdgeInsetsPropType} from 'DeprecatedEdgeInsetsPropType';
+export {default as PointPropType} from 'DeprecatedPointPropType';
+export {default as ViewPropTypes} from 'DeprecatedViewPropTypes';
+
+// NOTE(2019-02-21): Legacy alias that is deprecated and not yet removed
+// $FlowFixMe: Flow always wants a "value" property: https://github.com/facebook/flow/issues/5380
+Object.defineProperty(exports, 'AlertIOS', {
+  enumerable: true,
+  get() {
+    const warnOnce = require('warnOnce');
```
gengjiawen: Why not inline `warnOnce` ?

ide: Don't think it matters either way.

cpojer:
@ide now that the Metro PR landed, I assume you can publish this PR, right? I assume before we can land this, we'll need to publish metro and update the CLI?

cpojer:
Oh, and also: could you show me what a compiled version of `react-native-implementation.js` looks like with this change?

motiz88:
> we'll need to publish metro and update the CLI

FYI @cpojer I'm publishing Metro v0.53.1 today.

ide:
@cpojer @motiz88 Before we can land this PR, we'll need to update RN's transitive dependencies on the Babel preset. It's more complicated than I expected:

```
[1/4] 🤔  Why do we have the module "metro-react-native-babel-preset"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "metro-react-native-babel-preset@0.52.0"
info Reasons this module exists
   - "metro-react-native-babel-transformer" depends on it
   - Hoisted from "metro-react-native-babel-transformer#metro-react-native-babel-preset"
info Disk size without dependencies: "72KB"
info Disk size with unique dependencies: "996KB"
info Disk size with transitive dependencies: "15.32MB"
info Number of shared dependencies: 71
=> Found "@react-native-community/cli#metro-react-native-babel-preset@0.51.0"
info Reasons this module exists
   - "@react-native-community#cli#metro-react-native-babel-transformer" depends on it
   - Hoisted from "@react-native-community#cli#metro-react-native-babel-transformer#metro-react-native-babel-preset"
info Disk size without dependencies: "40KB"
info Disk size with unique dependencies: "944KB"
info Disk size with transitive dependencies: "15.29MB"
info Number of shared dependencies: 71
=> Found "metro#metro-react-native-babel-preset@0.51.1"
info This module exists because "@react-native-community#cli#metro" depends on it.
info Disk size without dependencies: "40KB"
info Disk size with unique dependencies: "944KB"
info Disk size with transitive dependencies: "15.29MB"
info Number of shared dependencies: 71
```

One way to test whether things are working are to verify the tests pass (currently they don't).

> could you show me what a compiled version of react-native-implementation.js looks like with this change?

The larger code snippet in the PR description is the transformed react-native-implementation.js with the new Babel preset.

cpojer:
Currently working on upgrading Metro in the CLI, and then upgrading the CLI in the RN repo so that we can continue on this PR. Will get back to you when it's ready.

cpojer:
@ide Upgraded the CLI to 2.0 which should come with the latest Metro. I think you should be unblocked now.

ide:
@cpojer Thanks -- rebased and updated. Tests are now passing on iOS.

cpojer:
Update: I have this working internally at FB but for some reason, even though `lazyImportExportTransform` is always true, the lazy exports aren't being applied. It will take me a bit of time to figure out why that may be.

cpojer:
Hey @ide. I'm sorry it took so long to figure this one out. The problem is that our internal setup with Metro doesn't work with this change and I have been unable to figure it out at this time.

Given other priorities it doesn't look like I'll get back to this one within the next two months. I propose closing this and trying to solve the internal fb problem over the next couple of months by only converting a single import to start. I'll see if @rubennorte has some time to help us out on this :)

PR #23602, opened 2019-02-22
[RNTester] fix getter of result from Image query cache

zhongwuzw:
## Summary

We assume `map` is the type of `Map`, but actually it's not, so we would get type error.

## Changelog

[iOS] [Fixed] - [RNTester] fix getter of result from Image query cache

## Test Plan

Before:
image

After:
We can get cache info.
image


React/Modules/RCTStatusBarManager.m
```diff
@@ -108,8 +108,11 @@ - (void)applicationWillChangeStatusBarFrame:(NSNotification *)notification
     RCTLogError(@"RCTStatusBarManager module requires that the \
                 UIViewControllerBasedStatusBarAppearance key in the Info.plist is set to NO");
   } else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
zhongwuzw: Remove warning.

RNTester/js/ImageExample.js
```diff
@@ -93,7 +93,7 @@ class NetworkImageCallbackExample extends React.Component<
                     `✔ Prefetch OK (+${new Date() - mountTime}ms)`,
                   );
                   Image.queryCache([IMAGE_PREFETCH_URL]).then(map => {
-                    const result = map.get(IMAGE_PREFETCH_URL);
+                    const result = map[IMAGE_PREFETCH_URL];
```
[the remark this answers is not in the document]
cpojer: @zhongwuzw seems like Flow isn't happy about this change?

React/Modules/RCTStatusBarManager.m
cpojer: Can you move this change into a separate PR?

zhongwuzw: @cpojer Yes, please see #23625 .

zhongwuzw:
@cpojer Hi, CI passes, the warning is because`queryCache`'s return type is not `Map`, that's also the reason this `PR` fixes.

cpojer:
I'm assuming your changes to the Image files to `Promise<any>` are unintentional? Can you revert them?

zhongwuzw:
@cpojer I'm intentional :joy:, because I found the return type is not `Map`, this is also cause of  `TypeError: map.get is not a function` error. Am I miss something?

cpojer:
Ahhh.. now I see what you are trying to fix here. In that case, shouldn't the type be: `Promise<{string: 'memory' | 'disk' | 'disk/memory'}>`?

zhongwuzw:
@cpojer The type is `Promise<{string: 'memory' | 'disk' | 'disk/memory'}>`.

PR #23603, opened 2019-02-22
Use existing character set in POST body when possible

nhunzaker:
## Summary

This commit fixes a bug introduced in a previous attempt (https://github.com/facebook/react-native/pull/23580) to address an issue where okhttp appended `charset=utf-8` to the Content-Type header when otherwise not specified.

In that commit, I converted all characters to UTF-8, however it should instead use an existing encoding when possible.

Related issues:
https://github.com/facebook/react-native/issues/8237#issuecomment-466304854

## Changelog

[Android][fixed] - Respect existing character set when specified in fetch() POST request

## Test Plan

Updated my test case to include a button that posts UTF-16 JSON:
https://github.com/nhunzaker/rn-charset-issue

Without this change, sending UTF-16 JSON fails to decode on the node server.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -375,7 +375,9 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
         // Use getBytes() to convert the body into a byte[], preventing okhttp from
         // appending the character set to the Content-Type header when otherwise unspecified
         // https://github.com/facebook/react-native/issues/8237
-        Charset charset = contentMediaType.charset(StandardCharsets.UTF_8);
+        Charset charset = contentMediaType == null
+          ? StandardCharsets.UTF_8
+          : contentMediaType.charset(StandardCharsets.UTF_8);
         requestBody = RequestBody.create(contentMediaType, body.getBytes(charset));
       }
     } else if (data.hasKey(REQUEST_BODY_KEY_BASE64)) {
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -372,7 +372,11 @@ public final class NetworkingModule extends ReactContextBaseJavaModule {
           return;
         }
       } else {
-        requestBody = RequestBody.create(contentMediaType, body.getBytes(StandardCharsets.UTF_8));
+        // Use getBytes() to convert the body into a byte[], preventing okhttp from
+        // appending the character set to the Content-Type header when otherwise unspecified
+        // https://github.com/facebook/react-native/issues/8237
+        Charset charset = contentMediaType.charset(StandardCharsets.UTF_8);
+        requestBody = RequestBody.create(contentMediaType, body.getBytes(charset));
       }
     } else if (data.hasKey(REQUEST_BODY_KEY_BASE64)) {
       if (contentType == null) {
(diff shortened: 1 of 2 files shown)


nhunzaker:
@cpojer This PR fixes a bug I introduced with https://github.com/facebook/react-native/commit/4a807761a4aca9e551ff2cee8ca18a2450fb11ca. That commit makes it so that all bodies are converted to a byte array using utf-8 encoding. 

I'm happy to do the work to revert this commit on master if you'd like.

cpojer:
@nhunzaker can you clarify your last comment? Why would we revert on master when we can merge this PR instead?

nhunzaker:
@cpojer Reverting on master isn't necessary if this PR is merged. I just felt bad and wanted to offer if having a bug on master was a problem.

ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
```diff
@@ -372,7 +372,11 @@ public void onProgress(long bytesWritten, long contentLength, boolean done) {
           return;
         }
       } else {
-        requestBody = RequestBody.create(contentMediaType, body.getBytes(StandardCharsets.UTF_8));
+        // Use getBytes() to convert the body into a byte[], preventing okhttp from
+        // appending the character set to the Content-Type header when otherwise unspecified
+        // https://github.com/facebook/react-native/issues/8237
+        Charset charset = contentMediaType.charset(StandardCharsets.UTF_8);
```
hey99xx: Any chance for NPE if contentMediaType is null?

nhunzaker: Yes:

https://github.com/square/okhttp/blob/19466ca4c0d43ebc69cdbc219b0298a8a6db30b4/okhttp/src/main/java/okhttp3/MediaType.java#L104-L110

I'll make this change, :+1:

ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
```diff
@@ -372,7 +372,13 @@ public void onProgress(long bytesWritten, long contentLength, boolean done) {
...
         }
       } else {
-        requestBody = RequestBody.create(contentMediaType, body.getBytes(StandardCharsets.UTF_8));
+        // Use getBytes() to convert the body into a byte[], preventing okhttp from
+        // appending the character set to the Content-Type header when otherwise unspecified
+        // https://github.com/facebook/react-native/issues/8237
+        Charset charset = contentMediaType == null
+          ? StandardCharsets.UTF_8
+          : contentMediaType.charset(StandardCharsets.UTF_8);
+        requestBody = RequestBody.create(contentMediaType, body.getBytes(charset));
```
nhunzaker: @hey99xx I updated this line to include your suggestion for null safety. Just for posterity too, this matches how other parts of the code are handling this case:

https://github.com/facebook/react-native/blob/cda8171af30815edfa331e07d1bbf605f0926303/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java#L562-L563

ReactAndroid/src/test/java/com/facebook/react/modules/network/NetworkingModuleTest.java
```diff
@@ -346,6 +347,94 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
+
+    ArgumentCaptor<Request> argumentCaptor = ArgumentCaptor.forClass(Request.class);
+    verify(httpClient).newCall(argumentCaptor.capture());
+
+    Buffer contentBuffer = new Buffer();
+    argumentCaptor.getValue().body().writeTo(contentBuffer);
+
+    assertThat(contentBuffer.readString(StandardCharsets.UTF_8)).isEqualTo("test");
+    assertThat(argumentCaptor.getValue().header("Content-Type")).isEqualTo("invalid");
+  }
```
nhunzaker: @hey99xx I also added this test case to cover the null pointer exception instance on bad content types. It asserts that the content is still parseable as UTF8 and the Content Type header is preserved.

Issue #9145, opened 2016-08-01
Port is hard coded to 8081 on iOS, Android templates


jdmarshall:
The documentation on working around port conflict issues is no longer valid

https://facebook.github.io/react-native/docs/troubleshooting.html

Mentions the command line and the URL in AppDelegate.m (which has changed to a function call now)
but does not mention that 

node_modules/react-native/React/React.xcodeproj/project.pbxproj

has 8081 hard-coded

Also, to be perfectly blunt, telling people to kill their virus scanner is negligent behavior.  We need a solution that doesn't involve asking people to turn off safety features their organization mandated that they use.  Especially when you have a company name attached to that advice.

hramos:
Terminating the process that is using the port is one alternative that is provided, followed with instructions to use a different port which would be a better solution.

Can you edit https://github.com/facebook/react-native/blob/master/docs/Troubleshooting.md with the latest instructions and submit a PR?

jdmarshall:
I can't, because I can't get it to work.  Short of searching the entire node_modules directory for instances of 8081 and changing them, because there are appear to be new spots that were hard coded after the instructions were written.

jdmarshall:
react-native/local-cli/server/server.js

Takes an override, but at least most of these below seem to ignore it:

react-native/React/Base/RCTBundleURLProvider.m
react-native/Libraries/WebSocket/RCTWebSocketExecutor.m
react-native/React/React.xcodeproj/project.pbxproj

The others I could probably sort out, despite not being an Objective C programmer, but the last one seems to be new and I don't have the foggiest how to pass in the port as a parameter:

react-native/React/React.xcodeproj/project.pbxproj

>  shellScript = "if nc -w 5 -z localhost 8081 ; then...

perkola:
These files below also seem to have port `8081` hard-coded in some places.
```
react-native/ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
react-native/packager/react-packager/src/Server/__tests__/Server-test.js
react-native/local-cli/util/isPackagerRunning.js
react-native/Libraries/Image/__tests__/resolveAssetSource-test.js
```

perkola:
Does this classify as a bug? Or how do i set a custom port that actually works?

perkola:
Bump, this seems to be a problem with Android as well. Starting the packager on a different port with `react-native start --port 8082` worked on `v.0.26.*` but no longer works on `v.30.0`.

[...]

PR #23616, opened 2019-02-24
Android Fix for 9145: No longer hard code build port

nhunzaker:
### Problem

According to https://github.com/facebook/react-native/issues/9145, the `--port` setting is not respected when executing `react-native run-android`. The templates that report things like what port the dev server runs on are hard coded as well.

### Solution

This commit replaces the hardcoded instances of port 8081 on Android with a build configuration property. This allows setting of the port React Native Android connects to for the local build server.

For this change to work, there must also be an update to the react native CLI to pass along this setting:

https://github.com/react-native-community/react-native-cli/compare/master...nhunzaker:9145-android-no-port-hardcode-cli

To avoid some noise on their end, I figured I wouldn't submit a PR until it's this approach is deemed workable.

## Changelog

[Android][fixed] - `react-native run-android --port <x>` correctly connects to dev server and related error messages display the correct port

## Test Plan

Preparing a test case for this is pretty involved, but:

1. Clone https://github.com/nhunzaker/react-native-cli/tree/9145-android-no-port-hardcode-cli
2. Install dependencies for this project
4. Clone https://github.com/nhunzaker/react-native/tree/9145-android-no-port-hardcode
5. Install dependencies for this project
6. Create a new RN project
7. Link this branch to the new project following [The Build From Source Instructions](https://facebook.github.io/react-native/docs/building-from-source)

(for future PR's, I'd love figure out a better flow for reproducing fixes)

Then from the new project, execute:

```
yarn start --port 3000
```

And in another terminal window:

```
node ../react-native-cli/packages/cli/ run-android --port 3000
```

You can also do this from gradle via:

```
./gradlew :app:installDebug -PdebugHostPort=3001
```

image

Error messages are also updated, so you'll see the correct port if the server is down:

image

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -294,7 +294,8 @@ android {
 
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
-        buildConfigField("int", "DEBUG_SERVER_HOST_PORT", debugHostPort("8081"))
+
+        resValue "integer", "REACT_NATIVE_DEV_SERVER_PORT", "8081"
 
         testApplicationId("com.facebook.react.tests.gradle")
         testInstrumentationRunner("androidx.test.runner.AndroidJUnitRunner")
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -295,7 +295,7 @@ android {
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
 
-        resValue "integer", "REACT_NATIVE_DEV_SERVER_PORT", "8081"
+        resValue "integer", "react_native_dev_server_port", "8081"
 
         testApplicationId("com.facebook.react.tests.gradle")
         testInstrumentationRunner("androidx.test.runner.AndroidJUnitRunner")
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -203,6 +203,11 @@ def findNdkBuildFullPath() {
     return null
 }
 
+def debugHostPort(String defaultValue){
+    def value = project.getProperties().get("debugHostPort")
+    return value != null ? value : defaultValue
+}
+
 def getNdkBuildFullPath() {
     def ndkBuildFullPath = findNdkBuildFullPath()
     if (ndkBuildFullPath == null) {
@@ -289,6 +294,8 @@ android {
 
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
+        buildConfigField("int", "DEBUG_SERVER_HOST_PORT", debugHostPort("8081"))
+
         testApplicationId("com.facebook.react.tests.gradle")
         testInstrumentationRunner("androidx.test.runner.AndroidJUnitRunner")
     }
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -203,8 +203,8 @@ def findNdkBuildFullPath() {
     return null
 }
 
-def debugHostPort(String defaultValue){
-    def value = project.getProperties().get("debugHostPort")
+def reactNativeDevServerPort(String defaultValue) {
+    def value = project.getProperties().get("reactNativeDevServerPort")
     return value != null ? value : defaultValue
 }
 
@@ -295,7 +295,7 @@ android {
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
 
-        resValue "integer", "react_native_dev_server_port", "8081"
+        resValue "integer", "react_native_dev_server_port", reactNativeDevServerPort("8081")
 
         testApplicationId("com.facebook.react.tests.gradle")
         testInstrumentationRunner("androidx.test.runner.AndroidJUnitRunner")
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -203,9 +203,14 @@ def findNdkBuildFullPath() {
     return null
 }
 
-def reactNativeDevServerPort(String defaultValue) {
+def reactNativeDevServerPort() {
     def value = project.getProperties().get("reactNativeDevServerPort")
-    return value != null ? value : defaultValue
+    return value != null ? value : "8081"
+}
+
+def reactNativeInspectorProxyPort() {
+    def value = project.getProperties().get("reactNativeInspectorProxyPort")
+    return value != null ? value : reactNativeDevServerPort()
 }
 
 def getNdkBuildFullPath() {
@@ -295,7 +300,8 @@ android {
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
 
-        resValue "integer", "react_native_dev_server_port", reactNativeDevServerPort("8081")
+        resValue "integer", "react_native_dev_server_port", reactNativeDevServerPort()
+        resValue "integer", "react_native_inspector_proxy_port", reactNativeInspectorProxyPort()
 
         testApplicationId("com.facebook.react.tests.gradle")
         testInstrumentationRunner("androidx.test.runner.AndroidJUnitRunner")
--- a/ReactAndroid/src/main/java/com/facebook/react/common/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/common/BUCK
@@ -37,8 +37,7 @@ rn_android_build_config(
     name = "build_config",
     package = "com.facebook.react",
     values = [
-        "boolean IS_INTERNAL_BUILD = true",
-        "int DEBUG_SERVER_HOST_PORT = 8081",
+        "boolean IS_INTERNAL_BUILD = true"
     ],
     visibility = [
         "PUBLIC",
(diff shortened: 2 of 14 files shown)


ReactAndroid/src/main/java/com/facebook/react/common/BUCK
```diff
@@ -33,6 +33,7 @@ rn_android_build_config(
     package = "com.facebook.react",
     values = [
         "boolean IS_INTERNAL_BUILD = true",
+        "int DEBUG_SERVER_HOST_PORT = 8081",
```
nhunzaker: I had to do this to get unit tests to pass. Not sure if this is the best approach.

ReactAndroid/src/main/java/com/facebook/react/common/DebugServerException.java
```diff
@@ -28,8 +30,8 @@
       "\u2022 Ensure that the packager server is running\n" +
       "\u2022 Ensure that your device/emulator is connected to your machine and has USB debugging enabled - run 'adb devices' to see a list of connected devices\n" +
       "\u2022 Ensure Airplane Mode is disabled\n" +
-      "\u2022 If you're on a physical device connected to the same machine, run 'adb reverse tcp:8081 tcp:8081' to forward requests from your device\n" +
-      "\u2022 If your device is on the same Wi-Fi network, set 'Debug server host & port for device' in 'Dev settings' to your machine's IP address and the port of the local dev server - e.g. 10.0.1.1:8081\n\n";
+      "\u2022 If you're on a physical device connected to the same machine, run 'adb reverse tcp:" + DEBUG_SERVER_HOST_PORT + " tcp:" + DEBUG_SERVER_HOST_PORT + "' to forward requests from your device\n" +
```
nhunzaker: I really want to use `AndroidInfoHelpers.getAdbReverseTcpCommand()` here, but I wasn't sure if modules in `common` should depend on other modules.

Salakar:
@nhunzaker thanks for getting this started 🎉 

It looks like you've submitted your changes to `ReactAndroid/build.gradle`. This will only work when building an app from RN sources (as per your Test Plan) - RN releases for Android are distributed as an [Android Library](https://developer.android.com/studio/projects/android-library) (`aar` format). 

You'd need to make your changes in `/react.gradle` in the root of the repository instead for this to work in release versions. Probably a combination of changes to both files.

nhunzaker:
@Salakar ahh cool. Thanks! I think I finally got a flow down where I can build the archive and link it to a project. Sure enough, the port isn't passed through. I'll try to figure it out!

ReactAndroid/src/main/java/com/facebook/react/common/DebugServerException.java
```diff
@@ -28,8 +28,8 @@
       "\u2022 Ensure that the packager server is running\n" +
       "\u2022 Ensure that your device/emulator is connected to your machine and has USB debugging enabled - run 'adb devices' to see a list of connected devices\n" +
       "\u2022 Ensure Airplane Mode is disabled\n" +
-      "\u2022 If you're on a physical device connected to the same machine, run 'adb reverse tcp:8081 tcp:8081' to forward requests from your device\n" +
-      "\u2022 If your device is on the same Wi-Fi network, set 'Debug server host & port for device' in 'Dev settings' to your machine's IP address and the port of the local dev server - e.g. 10.0.1.1:8081\n\n";
+      "\u2022 If you're on a physical device connected to the same machine, run 'adb reverse tcp:<PORT> tcp:<PORT>' to forward requests from your device\n" +
```
nhunzaker: Leaving a TODO for me: 
This needs access to app context to get the port, since it's now defined as a string res. Ideally this would match the current port.

nhunzaker:
@Salakar I was able to make headway by using a configurable integer res value:

```
// ReactNative/build.gradle
 resValue "integer", "REACT_NATIVE_DEV_SERVER_PORT", "8081"
```

This makes it easy to add an overide in `react.gradle`:

```
// react.gradle
 def debugHostPort(String defaultValue) {
    def value = project.getProperties().get("debugHostPort")
    return value != null ? value : defaultValue
}

android {
    buildTypes.all {
        resValue "integer", "REACT_NATIVE_DEV_SERVER_PORT", debugHostPort("8081")
    }
}
```

I first looked at manifest placeholders, but they're annoying because they can easily be overridden by other use cases, like if you assign over them with `manifestPlaceholders = [....]`

Resource values were also reasonable to figure out in Buck, which I'm not very familiar with.

The drawback is that the utilities that access the port also need access to context, but I think it's manageable. It's only a problem right now for the error message the suggests `adb reverse tcp:<port>`.

What do you think about this approach?

ReactAndroid/src/main/java/com/facebook/react/common/DebugServerException.java
nhunzaker: Got it. DebugServerException now accepts a URL in the `makeGeneric` method which parses out the port.

nhunzaker:
I updated `DebugServerException` to report the correct port. I think this PR is ready for review again.

ReactAndroid/build.gradle
```diff
@@ -286,6 +291,9 @@ android {
 
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
+
+        resValue "integer", "react_native_dev_server_port", reactNativeDevServerPort("8081")
```
nhunzaker: This isn't strictly necessary, but useful if you link the ReactNative module directly.

nhunzaker:
Just following up here. Without putting pressure to review this, I wanted to make sure things are easy to confirm, and I really wanted an easier way to verify changes for future fixes, so I made a repo to help with that:

https://github.com/nhunzaker/ReactNativeTestPlan

Assuming you have buck installed already, I think all you have to do is:

```
git clone git@github.com:nhunzaker/ReactNativeTestPlan.git
cd ReactNativeTestPlan
cd react-native
git remote add nhunzaker git@github.com:nhunzaker/react-native.git
git fetch nhunzaker
git checkout nhunzaker/9145-android-no-port-hardcode
cd ../react-native-cli
git remote add nhunzaker git@github.com:nhunzaker/react-native-cli.git
git fetch nhunzaker
git checkout nhunzaker/9145-android-no-port-hardcode-cli
cd ..

yarn install
yarn build-react-native
```

Then confirm the change with:

```
yarn start --port 4000
yarn run-android --port 4000
```

The Android app should build and connect to port 4000

I'm not sure if you all have something like this already, but it was really helpful for me.

grabbou:
Thanks for sending this PR! I think it's really important and valuable that you have submitted a fix for that after the issue being open for so long.

cpojer:
What's the status of this PR?

nhunzaker:
This is ready for review, but I'm worried the complicated test plan is making review difficult. 

In the short term, I'd be happy to address any feedback about the code itself.

ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
```diff
@@ -23,7 +25,7 @@
 
   public static final String METRO_HOST_PROP_NAME = "metro.host";
 
-  private static final int DEBUG_SERVER_HOST_PORT = 8081;
+  private static final int DEBUG_SERVER_HOST_PORT = ReactBuildConfig.DEBUG_SERVER_HOST_PORT;
   private static final int INSPECTOR_PROXY_PORT = 8081;
```
nhunzaker: @cpojer I noticed these ports are now the same on master. I wonder if this should now be:

```
private static final int DEBUG_SERVER_HOST_PORT = ReactBuildConfig.DEBUG_SERVER_HOST_PORT;
private static final int INSPECTOR_PROXY_PORT = ReactBuildConfig.DEBUG_SERVER_HOST_PORT;
```

Any idea?

nhunzaker:
@cpojer Updated! Just had question about the update to the inspector proxy port (https://github.com/facebook/react-native/pull/23616/files#r290498124)

Also looks like a lot of CI checks are failing. I'll look into that next.

ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
cpojer: Can we use the same config by default but still allow it to be customized separately? I think that’s ideal.

cpojer:
Sounds good! Let me know once CI is passing (don’t worry if the same job fails on master) and then I can merge it. We can update the CLI in a separate commit later once you land those changes there.

ReactAndroid/src/main/java/com/facebook/react/modules/systeminfo/AndroidInfoHelpers.java
nhunzaker: Yep. That makes sense to me.

nhunzaker: Confused myself during the rebase. This actually was removed and replaced with a resource value.

I can definitely do the same with the inspector proxy port. Just need to trouble shoot some issues running tests from a fresh setup.

ReactAndroid/build.gradle
```diff
@@ -289,6 +299,10 @@ android {
 
         buildConfigField("boolean", "IS_INTERNAL_BUILD", "false")
         buildConfigField("int", "EXOPACKAGE_FLAGS", "0")
+
+        resValue "integer", "react_native_dev_server_port", reactNativeDevServerPort()
+        resValue "integer", "react_native_inspector_proxy_port", reactNativeInspectorProxyPort()
```
nhunzaker: Updated. This now lets you set the inspector port using:

```
./gradlew :app:installDebug -PreactNativeDevServerPort=3000
```

By default it uses the dev server port.

nhunzaker:
Just following up, I have sent along the accompanying PR for the CLI here:
https://github.com/react-native-community/cli/pull/421

Though I hit some trouble with some changes for the 2.0 release (described over there). I'll keep things moving along over on the cli repo!

PR #23627, opened 2019-02-25
[WIP] Bump @react-native-community/cli to 1.3.0

rajivshah3:
## Summary

This PR upgrades `@react-native-community/cli` to 1.3.0, fixing a ReDoS vulnerability in `plist` (see https://snyk.io/test/npm/@react-native-community/cli/1.1.0)

## Changelog

[General] [Security] - Bump `@react-native-community/cli` to 1.3.0 (fixes `plist` ReDoS vulnerability)

## Test Plan (in progress)

- Testing all CLI commands to ensure they still work properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@react-native-community/cli": "^1.1.0",
+    "@react-native-community/cli": "^1.3.0",
     "art": "^0.10.0",
     "base64-js": "^1.1.2",
     "connect": "^3.6.5",
--- a/rn-cli.config.js
+++ b/rn-cli.config.js
@@ -10,7 +10,6 @@
 'use strict';
 
 const {resolve} = require('path');
-const {mergeConfig} = require('metro-config');
 const getPolyfills = require('./rn-get-polyfills');
 
 const projectRoot = __dirname;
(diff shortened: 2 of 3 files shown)


cpojer:
Let me know once this is ready.

rajivshah3:
@cpojer sorry, this might take longer than I originally expected unfortunately. There seems to be a new parameter `reactNativePath`, and the old code that used to allow the CLI to operate in `react-native` was [removed](https://github.com/react-native-community/react-native-cli/commit/2a180e02313f048b02bb3aeb1d09ff039d09a84f#diff-41ba6c5f93ad7880f321089481f6fb9cL97). You can pass the parameter on the command line, but I'm trying to figure out if it's possible to just specify it in `rn-cli.config.js` so all of the test scripts don't need to be updated.

grabbou:
Oh, I actually did that already in 0.59-stable branch when doing release candidate. I was planning to send a follow-up PR.

Please look at this commit https://github.com/facebook/react-native/commit/5e1504b0fca99cad3bfe2339ac0e7862b2315f9c and then, check all the subsequent 5 or 6 commits that I just pushed to fix all the occurrences.

That should be everything needed.

I think for the time being, we can do that instead. I will be working on 0.60-stable next week, so if you feel like working on the configuration of the CLI itself in the meantime, we could wait with the proper solution too!

PR #23632, opened 2019-02-25
Improve inverted VirtualizedList implementation

Sharcoux:
Use flex instead of complex transform inversion. Better for performance, readability and provides natural behaviour for onEndReached.

## Summary

The current implementation of `inverted` in `VirtualizedList` introduces a lot of problems, especially combined with the `onEndReached` events. You can see it [here](https://codesandbox.io/s/p5373vlm).

The scroll is reverted in an unnatural way, the onEndReached is not emitted at the right moment, and the scroll bar seems to be teleported randomly.

## Changelog

[General] [Fixed] - Better implementation for `inverted` and `onEndReached` on `VirtualizedList`

## Test Plan

Would you mind helping me on this part? What should I do exactly?

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -661,6 +661,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       getItem,
       getItemCount,
       horizontal,
+      inverted,
       keyExtractor,
     } = this.props;
     const stickyOffset = this.props.ListHeaderComponent ? 1 : 0;
@@ -681,6 +682,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           cellKey={key}
           fillRateHelper={this._fillRateHelper}
           horizontal={horizontal}
+          inverted={inverted}
           index={ii}
           item={item}
           key={key}
@@ -1611,6 +1613,7 @@ class CellRenderer extends React.Component<
     cellKey: string,
     fillRateHelper: FillRateHelper,
     horizontal: ?boolean,
+    inverted: ?boolean,
     index: number,
     item: Item,
     onLayout: (event: Object) => void, // This is extracted by ScrollViewStickyHeader
@@ -1685,6 +1688,7 @@ class CellRenderer extends React.Component<
       ItemSeparatorComponent,
       fillRateHelper,
       horizontal,
+      inverted,
       item,
       index,
       parentProps,
@@ -1709,8 +1713,8 @@ class CellRenderer extends React.Component<
       <ItemSeparatorComponent {...this.state.separatorProps} />
     );
     const cellStyle = horizontal
-      ? {flexDirection: 'row'}
-      : {flexDirection: 'column'};
+      ? {flexDirection: inverted ? 'row-reverse' : 'row'}
+      : {flexDirection: inverted ? 'column-reverse' : 'column'};
     if (!CellRendererComponent) {
       return (
         /* $FlowFixMe(>=0.89.0 site=react_native_fb) This comment suppresses an
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -1759,10 +1759,10 @@ class VirtualizedCellWrapper extends React.Component<{
 
 const styles = StyleSheet.create({
   verticallyInverted: {
-    flexDirection: 'row-reverse',
+    flexDirection: 'column-reverse',
   },
   horizontallyInverted: {
-    flexDirection: 'column-reverse',
+    flexDirection: 'row-reverse',
   },
   debug: {
     flex: 1,
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -653,7 +653,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     stickyIndicesFromProps: Set<number>,
     first: number,
     last: number,
-    inversionStyle: ViewStyleProp,
   ) {
     const {
       CellRendererComponent,
@@ -683,7 +682,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           fillRateHelper={this._fillRateHelper}
           horizontal={horizontal}
           index={ii}
-          inversionStyle={inversionStyle}
           item={item}
           key={key}
           prevCellKey={prevCellKey}
@@ -758,10 +756,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           key="$header">
           <View
             onLayout={this._onLayoutHeader}
-            style={StyleSheet.compose(
-              inversionStyle,
-              this.props.ListHeaderComponentStyle,
-            )}>
+            style={this.props.ListHeaderComponentStyle}>
             {
               // $FlowFixMe - Typing ReactNativeComponent revealed errors
               element
@@ -785,7 +780,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
         stickyIndicesFromProps,
         0,
         lastInitialIndex,
-        inversionStyle,
       );
       const firstAfterInitial = Math.max(lastInitialIndex + 1, first);
       if (!isVirtualizationDisabled && first > lastInitialIndex + 1) {
@@ -810,7 +804,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
                 stickyIndicesFromProps,
                 ii,
                 ii,
-                inversionStyle,
               );
               const trailSpace =
                 this._getFrameMetricsApprox(first).offset -
@@ -839,7 +832,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
         stickyIndicesFromProps,
         firstAfterInitial,
         last,
-        inversionStyle,
       );
       if (!this._hasWarned.keys && _usedIndexForKey) {
         console.warn(
@@ -884,7 +876,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
               element.props.onLayout(event);
             }
           },
-          style: [element.props.style, inversionStyle],
+          style: element.props.style,
         }),
       );
     }
@@ -901,10 +893,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           key="$footer">
           <View
             onLayout={this._onLayoutFooter}
-            style={StyleSheet.compose(
-              inversionStyle,
-              this.props.ListFooterComponentStyle,
-            )}>
+            style={this.props.ListFooterComponentStyle}>
             {
               // $FlowFixMe - Typing ReactNativeComponent revealed errors
               element
@@ -921,6 +910,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       onScrollBeginDrag: this._onScrollBeginDrag,
       onScrollEndDrag: this._onScrollEndDrag,
       onMomentumScrollEnd: this._onMomentumScrollEnd,
+      contentContainerStyle: inversionStyle,
       scrollEventThrottle: this.props.scrollEventThrottle, // TODO: Android support
       invertStickyHeaders:
         this.props.invertStickyHeaders !== undefined
@@ -928,12 +918,6 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           : this.props.inverted,
       stickyHeaderIndices,
     };
-    if (inversionStyle && itemCount !== 0) {
-      /* $FlowFixMe(>=0.70.0 site=react_native_fb) This comment suppresses an
-       * error found when Flow v0.70 was deployed. To see the error delete
-       * this comment and run Flow. */
-      scrollProps.style = [inversionStyle, this.props.style];
-    }
 
     this._hasMore =
       this.state.last < this.props.getItemCount(this.props.data) - 1;
@@ -980,6 +964,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   _fillRateHelper: FillRateHelper;
   _frames = {};
   _footerLength = 0;
+  _distanceFromEnd = 0;
   _hasDataChangedSinceEndReached = true;
   _hasInteracted = false;
   _hasMore = false;
@@ -1243,9 +1228,12 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       getItemCount,
       onEndReached,
       onEndReachedThreshold,
+      inverted,
     } = this.props;
     const {contentLength, visibleLength, offset} = this._scrollMetrics;
-    const distanceFromEnd = contentLength - visibleLength - offset;
+    const distanceFromEnd = inverted
+      ? offset
+      : contentLength - visibleLength - offset;
     if (
       onEndReached &&
       this.state.last === getItemCount(data) - 1 &&
@@ -1253,12 +1241,15 @@ class VirtualizedList extends React.PureComponent<Props, State> {
        * error found when Flow v0.63 was deployed. To see the error delete this
        * comment and run Flow. */
       distanceFromEnd < onEndReachedThreshold * visibleLength &&
+      distanceFromEnd !== this._distanceFromEnd &&
       (this._hasDataChangedSinceEndReached ||
         this._scrollMetrics.contentLength !== this._sentEndForContentLength)
     ) {
       // Only call onEndReached once for a given dataset + content length.
       this._hasDataChangedSinceEndReached = false;
       this._sentEndForContentLength = this._scrollMetrics.contentLength;
+      //We record the value to prevent from calling twice, especially when distanceFromEnd===0 as the scroll will stick to the top
+      this._distanceFromEnd = distanceFromEnd;
       onEndReached({distanceFromEnd});
     }
   }
@@ -1447,7 +1438,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   };
 
   _updateCellsToRender = () => {
-    const {data, getItemCount, onEndReachedThreshold} = this.props;
+    const {data, getItemCount, onEndReachedThreshold, inverted} = this.props;
     const isVirtualizationDisabled = this._isVirtualizationDisabled();
     this._updateViewableItems(data);
     if (!data) {
@@ -1476,7 +1467,9 @@ class VirtualizedList extends React.PureComponent<Props, State> {
         }
       } else {
         const {contentLength, offset, visibleLength} = this._scrollMetrics;
-        const distanceFromEnd = contentLength - visibleLength - offset;
+        const distanceFromEnd = inverted
+          ? offset
+          : contentLength - visibleLength - offset;
         const renderAhead =
           /* $FlowFixMe(>=0.63.0 site=react_native_fb) This comment suppresses
            * an error found when Flow v0.63 was deployed. To see the error
@@ -1619,7 +1612,6 @@ class CellRenderer extends React.Component<
     fillRateHelper: FillRateHelper,
     horizontal: ?boolean,
     index: number,
-    inversionStyle: ViewStyleProp,
     item: Item,
     onLayout: (event: Object) => void, // This is extracted by ScrollViewStickyHeader
     onUnmount: (cellKey: string) => void,
@@ -1695,7 +1687,6 @@ class CellRenderer extends React.Component<
       horizontal,
       item,
       index,
-      inversionStyle,
       parentProps,
     } = this.props;
     const {renderItem, getItemLayout} = parentProps;
@@ -1717,13 +1708,9 @@ class CellRenderer extends React.Component<
     const itemSeparator = ItemSeparatorComponent && (
       <ItemSeparatorComponent {...this.state.separatorProps} />
     );
-    const cellStyle = inversionStyle
-      ? horizontal
-        ? [{flexDirection: 'row-reverse'}, inversionStyle]
-        : [{flexDirection: 'column-reverse'}, inversionStyle]
-      : horizontal
-        ? [{flexDirection: 'row'}, inversionStyle]
-        : inversionStyle;
+    const cellStyle = horizontal
+      ? {flexDirection: 'row'}
+      : {flexDirection: 'column'};
     if (!CellRendererComponent) {
       return (
         /* $FlowFixMe(>=0.89.0 site=react_native_fb) This comment suppresses an
@@ -1772,10 +1759,10 @@ class VirtualizedCellWrapper extends React.Component<{
 
 const styles = StyleSheet.create({
   verticallyInverted: {
-    transform: [{scaleY: -1}],
+    flexDirection: 'row-reverse',
   },
   horizontallyInverted: {
-    transform: [{scaleX: -1}],
+    flexDirection: 'column-reverse',
   },
   debug: {
     flex: 1,
--- a/Libraries/Lists/__tests__/__snapshots__/FlatList-test.js.snap
+++ b/Libraries/Lists/__tests__/__snapshots__/FlatList-test.js.snap
@@ -6,6 +6,7 @@ exports[`FlatList renders all the bells and whistles 1`] = `
   ListEmptyComponent={[Function]}
   ListFooterComponent={[Function]}
   ListHeaderComponent={[Function]}
+  contentContainerStyle={null}
   data={
     Array [
       Object {
@@ -65,7 +66,11 @@ exports[`FlatList renders all the bells and whistles 1`] = `
       <header />
     </View>
     <View
-      style={null}
+      style={
+        Object {
+          "flexDirection": "column",
+        }
+      }
     >
       <View
         style={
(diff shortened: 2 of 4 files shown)


cpojer:
@Sharcoux for the "Test Plan", can you fill out how exactly you verified that your PR fixes the described issue and how are you making sure it does not introduce regressions?

sahrens:
How does this actually invert the scroll direction? It should start at the bottom, like a chat, rather than at the top, like a feed.

Can you describe more specifically what’s wrong with onEndReached?

A screencast if walking through all the options in the FlatListExample would be a good part of a test plan.

Sharcoux:
The problem encountered is described [here](https://github.com/facebook/react-native/issues/23635). You can experiment it with [this sample](https://codesandbox.io/s/p5373vlm).

@sahrens The idea here is to use the `flex-direction : column-reverse` to make the elements display from bottom to top

@cpojer Actually, I had trouble finding out how to test my changes. What I ended up doing was to replicate my changes directly into the nodes_modules of the sample project provided above. What is the easiest way to import my changes into the project ? I don't understand the build process... Or should I create an account on npm and publish my own package ?

Sharcoux:
> A screencast if walking through all the options in the FlatListExample would be a good part of a test plan.

@sahrens I guess you are talking about this FlatListExample, but I don't understand how to import my changes...

Sharcoux:
Sorry for closing, I misclicked...

terrysahaidak:
Hi @Sharcoux, you can use the example for my issues about bugs with inverted lists for testing and fixing it. Let's work together to fix all of these.

https://github.com/terrysahaidak/ReactNative-SeactionList-Bug-Example

So far I've tested your changes with latest RN and my example here is the result:
image

Sharcoux:
Oh perfect @terrysahaidak, thanks. I told you, I manually reported my changes I made from the node_modules, so I probably messed up some part.

Still, I don't understand how I am supposed to import my changes within your project. I see that you used `react-native: '0.59.0-rc.3'`. What is this version?
Moreover, i made the changes, but I didn't build them. I guess it is part of the publication process? So should I publish a new version each time I need to test a change? I really don't understand the workflow to change something and test the changes...

Sharcoux:
@terrysahaidak I just inverted `row-reverse` and `column-reverse` at the very bottom of the file.

terrysahaidak:
@Sharcoux that's the latest one published on NPM. As for me, the easiest way will be making changes directly in the node_modules of that repo. That allows you to see them as soon as possible.

The other way is cloning RN's master somewhere, building from sources and linking the built version via `npm link`. That one is better but requires some time. Let me know if you'll choose a second one I'll help you to link RN to that example project because it's not a straight forward process (at least for now).

Sharcoux:
@terrysahaidak so there is no build process, ok.
I successfully replicated my changes into your project. I discovered some issues with the way separator components are inserted. Let me fix that.

Sharcoux:
Is it me, or are cells recreated during each call to `render`??

Sharcoux:
Ok, I fixed the insertion of separators. The current code for VirtualizedList is quite unsatisfying in my opinion. But I'm afraid it might be too late for a good clean up. I just kept things in phase with the current spirit to not break anything, but I'm wondering if it won't reveal performance issues for large lists.
Please, let me know your thoughts and if I missed something. For me the PR looks ok now.

terrysahaidak:
The main problem with your solution and why the previous implementation used `transform` – it should be bottom-to-top scroll instead of top-to-bottom.

Here is your solution:
kapture 2019-02-28 at 15 40 19

Here is the original one:
kapture 2019-02-28 at 15 43 27

Sharcoux:
I'm sorry, I'm not sure I get the difference. From your screencast, the only difference I see is that the scroll initial position is at the bottom, which I could probably fix if it's what is intended. Could you confirm?

Sharcoux:
By the way, if this really was your point, it means that there is an inconsistency between the implementation of column-reverse in browser and in mobile. Check [this sandbox](https://codesandbox.io/s/m52mjnyk98) for instance.
Thus, should we still fix this here?

terrysahaidak:
@Sharcoux that's not an only difference, that's the main problem. `scrollTo(listHeigth, false)` on did mount won't work for that.

Inconsistency between web and mobile because react-native-web is an experimental technology and it attempts to reimplement RN primitives (which are native on each platform) in the web.

But anything you're trying to fix should be tested and fixed in the "native" react native, not a web one.

Sharcoux:
You got it wrong. I'm not talking about how `react-native-web` handles the scroll, but how browsers handle `flex-direction: column-reverse` CSS property. When styling a component with CSS, I expect the implementation to behave the same, whether it is on mobile or web.

Sharcoux:
You got it wrong. I'm not talking about the way `react-native-web` handles the scroll, but the way browsers implement `flex-direction: column-reverse`. When styling a component with CSS, I just expect the implementation to follow the specs, whether it is on browser or mobile. And currently, the specs [state this](https://www.w3.org/TR/css-align-3/#overflow-scroll-position).

Sharcoux:
If you are using a browser that follows the specs, you can check [this sandbox](https://codesandbox.io/s/r70v5k78p4) that works without React-Native if you'd rather.

Sharcoux:
So? What should we do now?
Where is implemented the behaviour of `flex-direction: column-reverse` so that we can fix it?

cpojer:
That would be in Yoga which implements the flex layout spec. I’ll close this PR here because without the change to Yoga, we won’t be able to merge it. Let’s get back to making this change to RN once the issue is resolved there.

Sharcoux:
@cpojer Ok, I'll handle the problem directly with `yoga-layout`. I'll publish updates here when we have reached a PR there.

Sharcoux:
Issue in Yoga : [here](https://github.com/facebook/yoga/issues/866)

Sharcoux:
I got the answer from Yoga and I am ready to try and fix this once and for all, but I still don't understand how to easily test my changes... I forked react-native, I made my changes, I created a dummy project, I replaced the react-native version by my repo : "git@github.com:Sharcoux/react-native", but then I get:

`Couldn't find any versions for "react-native" that matches "git@github.com:Sharcoux/react-native"`

Sharcoux:
I rebased my changes on master and fixed the conflicts, I know how to fix things with Yoga, but I never succeeded to try them because I am totally unable to build any project.

I get:

```
/flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:5: error: cannot find symbol
import com.facebook.react.ReactApplication;
flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:6: error: cannot find symbol
import com.facebook.react.ReactNativeHost;
flatlist/MainApplication.java:14: error: cannot find symbol
public class MainApplication extends Application implements ReactApplication {
flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:16: error: cannot find symbol
  private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:36: error: cannot find symbol
  public ReactNativeHost getReactNativeHost() {
flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:16: error: cannot find symbol
  private final ReactNativeHost mReactNativeHost = new ReactNativeHost(this) {
flatlist/android/app/src/main/java/com/flatlist/MainApplication.java:35: error: method does not override or implement a method from a supertype
flatlist/android/app/src/main/java/com/flatlist/MainActivity.java:5: error: MainActivity is not abstract and does not override abstract method getPackages() in ReactActivity
public class MainActivity extends ReactActivity {
```

dblachut-adb:
> That would be in Yoga which implements the flex layout spec. I’ll close this PR here because without the change to Yoga, we won’t be able to merge it. Let’s get back to making this change to RN once the issue is resolved there.

@cpojer Does the yoga have any information about the (initial) scroll position or it "only" performs the layout? I am uncertain if it would be responsible for such setting.

sahrens:
I'm not sure the issue is in Yoga - it's more about the behavior of the native ScrollView component. Just setting the initial scroll position is also not sufficient - you also need to change the behavior when appending new content - if you're looking at the first items (first elements of the children array) and then you add more, without doing the transform and relying only on that initial scroll position, the new items will push the first item down off the bottom of the viewport, one of the main things reverse mode is supposed to avoid.

sahrens:
More fundamentally, I don't understand your concerns and motivations in the first place. What is the problem with performance? What are you trying to achieve? What's the problem with `onEndReached`? When you invert, the semantics change. If you need different semantics, just use `onScroll` to trigger whatever logic you want however you want - that's how onEndReached works under the hood.

sahrens:
And yes I agree the code is not in a great state - it has evolved a ton with a bunch of complex features getting tacked on over the years that have complex interactions with each other. You don't have to use it though - you can use ScrollView directly, or you can use https://github.com/Flipkart/recyclerlistview, or you can fork VirtualizedList and strip out the features you don't need to simplify it - it depends on what you need to accomplish.

sahrens:
And I agree that ScrollView with row-reverse should follow the w3c spec, but that's a whole project that is pretty low-pri at the moment. I think it can be done outside of Yoga though, but it will probably be pretty tricky.

dblachut-adb:
Thank you very much for your commitment to the topic.
> you also need to change the behavior when appending new content
[...]

I think that is already covered by `column-reverse` and `row-reverse` modes.

> I don't understand your concerns and motivations in the first place

I am not sure about others but I am facing some problems on AndroidTV where such transformation is causing the scroll to work in opposite direction (AFAIK it is not yet reported anywhere).

Correct me if I am wrong, but the current implementation is a bit hacky. It works by transforming whole view along one axis and transform **every** children (`Cell`) back to its original form. I am not sure how such transformations work under the hood, but surely this is not the most performant and elegant way to achieve the result.

IMO the implementation with `column-reverse` and `row-reverse` would be more straightforward.
The current implementation could possibly be simplified by transforming `ScrollComponent` through `style` prop, and transforming the content back through `contentContainerStyle`, but still this would not fix the problems that some of us are facing.

sahrens:
Using a transform should actually be very performant because it's a simply matrix operation performed by the GPU. We haven't seen any noticeable performance impact in our usage.

dblachut-adb:
I see, so it looks like the performance should not be our concern even on some low-end GPUs and when there are a lot of elements? 

> I think it can be done outside of Yoga though, but it will probably be pretty tricky.

How about introducing some new props on `ScrollView` like `initialScrollPosition` and/or `scrollMode` which could contain something like:
- _messageList_ - auto scroll to end on new element insertion (if the scroll is at the end)
- _initiallyEnd_ - scroll to end on initial layout
- _etc_...

Then it could be simply used in `FlatList`/`VirtualizedList` in `inverted` mode.
If I am not mistaken some of these features are natively supported in Android. I have found things like `stackFromBottom` `setReverseLayout` `setStackFromEnd`. Maybe the scroll component could be implemented on `ListView` which have `stackFromBottom` attribute support.
I am not sure if I am not over-complicating it a bit though...

dblachut-adb:
> The current implementation could possibly be simplified by transforming `ScrollComponent` through `style` prop, and transforming the content back through `contentContainerStyle`, but still this would not fix the problems that some of us are facing.

I think I was wrong on this one - this would only help for the initial scroll position, but the order of elements would be incorrect and they would get mirrored-back.

dblachut-adb:
> Can you elaborate more on the issues you're seeing on AndroidTV? Why are you using inverted in the first place? The whole point is to reverse the scroll direction.

Hmmm... So on mobile devices the user needs to swipe in reverse direction? I thought it was meant to reverse the item order and initial scroll position so that e.g. the element list grow upward.

I have something like a bottom navigation bar from which I am choosing the screen. If the screen contains list of tiles it is natural to start from the first one after pressing UP and scroll upward to further ones. The problem is that the directional navigation using DPAD on remote is working in opposite direction. Please see the description that I have added in my previous comment: https://github.com/facebook/react-native/pull/23632#issuecomment-523450494 (starting from _It happens_)

sahrens:
Oh I see, yeah the DPAD and touch gestures should all move the scroll in the same natural direction. Sounds like a bug in the AndroidTV ScrollView implementation or whatever primitive it's based on, especially if it works correctly on iOS, Android, and AppleTV. Who maintains AndroidTV? I'd focus on them.

sahrens:
Maybe ask @ - he added DPAD stuff in https://github.com/facebook/react-native/commit/b7bb2e5745f2bdbfeeccef8d97d469730942e01c

Maybe muck with stuff in here? https://github.com/facebook/react-native/blob/3b6f6ca4d5fcee6f1bc6d6242e3e2ef136e4d546/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java#L301-L309

dblachut-adb:
Thanks! :+1: I will look on it a bit deeper but I still can't help the feeling that it could be fixed by improved version of this PR. :-)

Sharcoux:
If you check [here](https://codesandbox.io/s/p5373vlm), you can notice the bug that @dblachut-adb was talking about, but there is worse. The onEndReached event is introducing weird behaviours when you scroll up and load more than 100 items.

This is caused by the current hack in the Flat-List. The only explaination I have to explain why the current code works on Android and iOS is that there should be a "counter-hack" in the native views that will fix the hack we are introducing here.

IMO, this is really not sane.

Sharcoux:
@sahrens  @dblachut-adb I really don't understand guys... Instead of fixing `yoga` implementation of `column-reverse` so that it can follow the specs, you'd rather keep a crazy hack in `VirtualizedList`, keep a hack somewhere in the Views, and keep a non standard implementation of `column-reverse`?

dblachut-adb:
@Sharcoux I don't have a very good understanding on the responsibility of each react-native component, but if I am not mistaken yoga is the library which calculates the measurements for views and defines how the views are laid out on the screen. It does not contain any information about the initial scroll position nor it is not able to set it.

That was my concern and I asked react-native team if the potential fix should really be there.

dblachut-adb:
BTW with your changes the initial scroll position would have to be changed on each platform. It will also have problems with the scroll position as reported by @terrysahaidak which means that in inverted mode:
- scrollToEnd will scroll to the bottom, which is the "beginning" in inverted mode
- all other scrollTo*(x) functions will have similar problem
- some events may report wrong scroll position
- maybe something else?

Unless I am wrong and this can really be fixed in yoga, which is common for all platforms, and hopefully the scroll position would then be reported/processed correctly.

Sharcoux:
But you need to separate the concerns. The objective of VirtualizedList is to provide a React-Native implementation of a list based with CSS/flex styling.
The concern of the scroll initial position is relative to the implementation of `row-reverse`, and this should not be addressed in `VirtualizedList`, but in Yoga as it seems.

"some events may report wrong scroll position" -> That's exactly the opposite don't you think? If you draw all components on a reversed axis but with reversed-reversed transformation, don't you think that it is very likely that the detected scroll position will be wrong hard to make sense of?

I mean, the specs says that a view with `row-reverse` direction should start the scroll at the bottom. Nobody seemed to report having troubles with incorrect scroll position being fired when using `row-reversed` on the web, don't you agree? So why should things be different with react-native implementation?

I'm not sure I'm being clear, so sorry if I'm confusing.

Sharcoux:
What `scrollToEnd` and `onEndReached` should be taking as reference for the "End" given if the list is `inverted` or not, this has no relation with the implementation of the View.

And with the current implementation, if I want to scroll to the start of an item in an inverted FlatList, what will it be? Because according to the VirtualizedList, the start of the item is actually the end of this item being drawn backwards... So there is 100% chance of what I'm trying to accomplish giving me headhache after 10 minutes.

dblachut-adb:
> The concern of the scroll initial position is relative to the implementation of column-reverse, and this should not be addressed in VirtualizedList, but in Yoga as it seems.

Yes, maybe the issue should be addressed in Yoga, but my concern was that the Yoga possibly does not maintain the position of scroll. In such situation it would need to send some information outside it or it should be implemented elsewhere.

> That's exactly the opposite don't you think? If you draw all components on a reversed axis but with reversed-reversed transformation, don't you think that it is very likely that the detected scroll position will be wrong, or at least hard to make sense of?

I was under impression that the goal is to change the underlying implementation without breaking existing API and applications. If the goal is to follow the spec of `row-reverse` and `column-reverse` then it should be properly documented as a breaking change and the doc should clearly say what position you should use and will receive in events.

It may have more sense to have common position calculation on both modes. I guess it basically depends what do we want from "inverted" mode. If it should simply reverse the item order and change initial scroll position (the same as `row/column-reverse` styles), then it possibly should just be removed as it could be simply applied through `contentContainerStyle` property.

Currently the "inverted" mode is something more than the web `row/column-reverse`. Also please keep in mind that react-native does not fully follow the web standards, e.g. the default `flex-direction` is different.

> What scrollToEnd and onEndReached should be taking as reference for the "End" given if the list is inverted or not, this has no relation with the implementation of the View.

If those methods are implemented in the View (`VirtualizedList`) itself then how it has no relation with it? What is your proposal on where it should be implemented then?

Don't get me wrong - I understand your concerns about the position calculation and the "badness" of current implementation. I also like your approach to implementing the "inverted" mode, but currently, it is a bit different in usage than previous solution. :)

Libraries/Lists/VirtualizedList.js
```diff
@@ -921,19 +912,14 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       onScrollBeginDrag: this._onScrollBeginDrag,
       onScrollEndDrag: this._onScrollEndDrag,
       onMomentumScrollEnd: this._onMomentumScrollEnd,
+      contentContainerStyle: inversionStyle,
```
dblachut-adb: `this.props.contentContainerStyle` is lost, please use array of styles or maybe `StyleSheet.compose` (I am not sure what is the difference).

dblachut-adb: `this.props.contentContainerStyle` is lost, please use array of styles or maybe `StyleSheet.compose` (I am not sure what is the difference).

Sharcoux: I probably got wrong what `inversionStyle` was supposed to be for. Ok, I should fix that indeed if it actually is the contentContainerStyle.

Sharcoux:
> Yes, maybe the issue should be addressed in Yoga, but my concern was that the Yoga possibly does not maintain the position of scroll. In such situation it would need to send some information outside it or it should be implemented elsewhere.

The position of the scroll is probably managed in the ScrollView, so everything related should stay there don't you think?

> I was under impression that the goal is to change the underlying implementation without breaking existing API and applications. If the goal is to follow the spec of `row-reverse` and `column-reverse` then it should be properly documented as a breaking change and the doc should clearly say what position you should use and will receive in events.

I think that there might be some minor braking changes if people already hacked their way through `scrollTo`. But most cases should still behave the same. Inverted FlatList will start at the bottom and `onEndReached` / `scrollToEnd` will be relative to the top.

> It may have more sense to have common position calculation on both modes. I guess it basically depends what do we want from "inverted" mode. If it should simply reverse the item order and change initial scroll position (the same as `row/column-reverse` styles), then it possibly should just be removed as it could be simply applied through `contentContainerStyle` property.

I think that as long as it is consistent it should be ok. We can reverse the axis or keep the same axis both ways. My concern is just that with current implementation, the offset inside one child goes on the opposite direction as the offset of the scroll view. Usually, `yOffsetInParentView = yOffsetInChild + yOffsetOfChildInParent`. Here, we have the very hacky and counter intuitive situation where `yOffsetInParentView = yOffsetOfChildInParent + childHeight - yOffsetInChild`. You see what I mean?

> Currently the "inverted" mode is something more than the web `row/column-reverse`. Also please keep in mind that react-native does not fully follow the web standards, e.g. the default `flex-direction` is different.

As long as you don't pretend that `View = div` I don't see why the defaults should be the same. Thus, I don't feel that RN doesn't follow the standards. Anyway, here we are currently following the standards in this VirtualizedList. The issue is that the implementation of Yoga doesn't follow the standards. But as I understand, the Yoga project aims at providing a native implementation of the Flex standards, so if it fails to do so, I feel more that it is a bug that should be fixed than a feature that should be accepted as is.

> > What scrollToEnd and onEndReached should be taking as reference for the "End" given if the list is inverted or not, this has no relation with the implementation of the View.
>
> If those methods are implemented in the View (`VirtualizedList`) itself then how it has no relation with it? What is your proposal on where it should be implemented then?

What I mean is that it is unrelated because deciding this doesn't mean we should change the axis direction. I mean that if we decide that the End is the top in inverted lists, we just need to make it behave this way in the implementation of these functions. Reversing the axis will indeed have an effect on the implementation but should be transparent as for the result. I'm not sure if I'm clear...
 
> Don't get me wrong - I understand your concerns about the position calculation and the "badness" of current implementation. I also like your approach to implementing the "inverted" mode, but currently, it is a bit different in usage than previous solution. :)

I think there could be a difference in the result of the `scrollTo` function which will be more accurate as I explained. It will also make a difference if other people used the `column-reversed` property implemented by Yoga and were fine with the fact that it didn't respect the specs, but I don't see what it will change for the high majority of cases... Do you have something in mind?

dblachut-adb:
> The position of the scroll is probably managed in the ScrollView, so everything related should stay there don't you think?

Yes, that was exactly my point.

> I think that there might be some minor braking changes if people already hacked their way through `scrollTo`. But most cases should still behave the same. Inverted FlatList will start at the bottom and `onEndReached` / `scrollToEnd` will be relative to the top.

With such proposal, to keep the same functionality, we would need something like `scrollToBeginning` and `onBeginningReached`. Actually I would strongly recommend to rename it to `scrollToTop` and `scrollToBottom` to avoid the confusion in inverted mode.

> I think that as long as it is consistent it should be ok. We can reverse the axis or keep the same axis both ways. My concern is just that with current implementation, the offset inside one child goes on the opposite direction as the offset of the scroll view. Usually, `yOffsetInParentView = yOffsetInChild + yOffsetOfChildInParent`. Here, we have the very hacky and counter intuitive situation where `yOffsetInParentView = yOffsetOfChildInParent + childHeight - yOffsetInChild`. You see what I mean?

Yes I understand you. For me the best solution would be if the functionality would be the same and the position would be consistent (independent from using inverted mode or not). The alternative is to keep the same functionality and maintain the reversed position in reversed mode.

> Yoga project aims at providing a native implementation of the Flex standards, so if it fails to do so, I feel more that it is a bug that should be fixed than a feature that should be accepted as is.

My point was that it may have more value to provide some functionality than follow the standards of different specification in this case. If we can have both - even better.

> What I mean is that it is unrelated because deciding this doesn't mean we should change the axis direction. I mean that if we decide that the End is the top in inverted lists, we just need to make it behave this way in the implementation of these functions. Reversing the axis will indeed have an effect on the implementation but should be transparent as for the result. I'm not sure if I'm clear...

Yes, I totally agree.

> I think there could be a difference in the result of the `scrollTo` function which will be more accurate as I explained. It will also make a difference if other people used the `column-reversed` property implemented by Yoga and were fine with the fact that it didn't respect the specs, but I don't see what it will change for the high majority of cases... Do you have something in mind?

What I meant was that if this commit breaks something - the doc should be updated. And it should not remove some functionality that previously was present.

dblachut-adb:
>     1. You seem to say that Yoga won't be able to communicate to the ScrollView the position it should start to. I don't see why [...]

I have just said that I wasn't sure if Yoga is able to do this. That was just the thing I was afraid of, nothing more (I don't have the knowledge in this area). ;-)

>        I propose to start with a correct CSS representation of what is a reverted FlatList, then a correct implementation of how to represent the CSS `reversed-column` and `reversed-row` in a ScrollView by updating Yoga (and/or the ScrollView if needed, but I don't think we do), and end up with a solution that will change nothing for users, except that things like offset and directions will now be consistant.

Yes, that would be great.

>     2. I don't see what functionality won't be present anymore. I don't see what you are referring to...

Right now e.g. `scrollToEnd` does not work as expected and `contentContainerStyle` cannot be used.

> I totally disagree on this. The End should always refer to the direction where the last item in the list is represented. Thus, the end should be the top on reversed lists, and the bottom on normal lists. I don't think either that we really need a `onBeginningReached` event, but why not.

You have understood me wrong, the rename would be needed in your current implementation, in which `scrollToEnd` always scrolls to bottom independently of the `inverted` mode or not.

> If you create `scrollToTop` and `scrollToBottom`, you will make it very confusing for horizontal lists. And creating `scrollToLeft` and `scrollToRight` functions will make it overcomplicated for nothing.

Ah, my bad, I have forgot about the horizontal mode.

Sharcoux:
> Right now e.g. scrollToEnd does not work as expected and contentContainerStyle cannot be used.

I agree on `contentContainerStyle`, it should be fixed.

Oh! My bad for scrollToEnd. I didn't see it being implemented here. Indeed, it should be fixed too.

mallenexpensify:
If anyone is looking for an incentive to fix this aspect `Selecting, copying, and pasting text on web and desktop are weird and in inverted order` We're willing to compensate.  Issue is posted here
https://github.com/Expensify/Expensify.cash/issues/1341

mallenexpensify:
We're still looking for help to "Fix Inverted prop on the FlatList component with a RN Patch". The current link is Expensify/Expensify.cash#3381 job amount is $8000. Any help and recommendations are greatly appreciated.

PR #23647, opened 2019-02-25
Call super if shouldDisableScrollInteraction false

ericlewis:
## Summary

Allow iOS to handle `touchesShouldCancelInContentView` on RCTScrollView if we aren't disabling the scroll interaction.

## Changelog

[iOS] [Changed] - RCTScrollView allows iOS to handle touchesShouldCancelInContentView

## Test Plan

N/A

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -280,8 +280,13 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
  */
 - (BOOL)touchesShouldCancelInContentView:(__unused UIView *)view
 {
-  //TODO: shouldn't this call super if _shouldDisableScrollInteraction returns NO?
-  return ![self _shouldDisableScrollInteraction];
+  BOOL shouldDisableScrollInteraction = [self _shouldDisableScrollInteraction];
+  
+  if (shouldDisableScrollInteraction == NO) {
+    [super touchesShouldCancelInContentView:view];
+  }
+  
+  return !shouldDisableScrollInteraction;
 }
 
 /*


React/Views/ScrollView/RCTScrollView.m
```diff
@@ -280,8 +280,13 @@ - (void)scrollRectToVisible:(CGRect)rect animated:(BOOL)animated
  */
 - (BOOL)touchesShouldCancelInContentView:(__unused UIView *)view
 {
-  //TODO: shouldn't this call super if _shouldDisableScrollInteraction returns NO?
-  return ![self _shouldDisableScrollInteraction];
+  BOOL shouldDisableScrollInteraction = [self _shouldDisableScrollInteraction];
+  
+  if (shouldDisableScrollInteraction == NO) {
+    [super touchesShouldCancelInContentView:view];
```
shergin: @ericlewis 
Eric,
I just discovered this change. Man, you are breaking my heart! 😂😂😂
(Happy 🦃 day!)

ericlewis: @shergin oh no! What’s wrong?!
Happy turkey day to you too!

sammy-SC: @ericlewis shouldn't this be `return [super touchesShouldCancelInContentView:view];`?

Because `[UIScrollView touchesShouldCancelInContentView:]` does nothing besides checking if view passed in as an argument is a UIControl.

If you remember by any chance, do you have anymore context on this PR? What does it fix?

ericlewis: @sammy-SC I want to say that there was some sort of issue that scroll would break if we didn't defer to iOS when we weren't overriding it ourselves. I think you're correct it should be passed back. Also, there is good chance I was simply correcting a TODO.

PR #23654, opened 2019-02-26
[iOS] Fixed placeholder font not consistent with text's font when in multiline mode

zhongwuzw:
## Summary

Placeholder's font not consistent with text's font, it leads to cursor dislocation. We need to keep consistent between placeholder and text.

## Changelog

[iOS] [Fixed] - Fixed placeholder font not consistent with text's font when in multiline mode

## Test Plan

Reproduce code:
```
        <TextInput 
  placeholder="Lorem ipsum dolor sit amet, risus nunc, mauris congue mus facilisis. Sit quis euismod. Nisl ad conubia vivamus hymenaeos massa, aliquam qui a sagittis et, eget etiam a ut sollicitudin suscipit, ut nisl cras tincidunt vestibulum sodales. Maecenas vivamus necessitatibus non, in sapien. Mauris ante"
  multiline={true}
/>
```

Before:
image

After:
image


hramos:
Ah, let me fix this. Sorry about that.

PR #23656, opened 2019-02-26
[iOS] Remove compatible system code for iOS8 and before

zhongwuzw:
## Summary

We already only support `iOS9+`, so we can remove all compatible codes now.

## Changelog

[iOS] [Fixed] - Remove compatible system code for iOS8 and before

## Test Plan

CicleCI passes.


ericlewis:
LGTM!

PR #23658, opened 2019-02-26
Properly validate JS->native method calls

aleclarson:
## Summary

Between invalidating a bridge and suspending its JS thread, native modules may have their methods called.

Only warn when a native module has been invalidated, which happens right before its JS thread is suspended.

Avoid initializing a native module's instance if its bridge is invalidated.

/cc @fkgozali https://github.com/facebook/react-native/commit/f94521244798f859648d1769f397102c06d763f0#commitcomment-32467567

## Changelog

[iOS] [Fixed] - Properly validate JS->native method calls

## Test Plan

N/A

### DIFF
--- a/React/Base/RCTModuleData.mm
+++ b/React/Base/RCTModuleData.mm
@@ -330,9 +330,11 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init);
 
 - (dispatch_queue_t)methodQueue
 {
-  (void)[self instance];
-  RCTAssert(_methodQueue != nullptr, @"Module %@ has no methodQueue (instance: %@, bridge.valid: %d)",
-            self, _instance, _bridge.valid);
+  if (_bridge.valid) {
+    id instance = self.instance;
+    RCTAssert(_methodQueue != nullptr, @"Module %@ has no methodQueue (instance: %@)",
+              self, instance);
+  }
   return _methodQueue;
 }
 
--- a/React/CxxModule/RCTNativeModule.mm
+++ b/React/CxxModule/RCTNativeModule.mm
@@ -71,17 +71,19 @@ void RCTNativeModule::invoke(unsigned int methodId, folly::dynamic &&params, int
     invokeInner(weakBridge, weakModuleData, methodId, std::move(params));
   };
 
-  if (m_bridge.valid) {
-    dispatch_queue_t queue = m_moduleData.methodQueue;
-    if (queue == RCTJSThread) {
-      block();
-    } else if (queue) {
-      dispatch_async(queue, block);
-    }
-  } else {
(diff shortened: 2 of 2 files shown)


React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -1060,20 +1060,17 @@ - (void)invalidate
     // wrappers) or we may have invalid pointers still in flight.
     dispatch_group_t moduleInvalidation = dispatch_group_create();
     for (RCTModuleData *moduleData in self->_moduleDataByID) {
+      [moduleData invalidate];
```
fkgozali: is the motivation for moving this call before calling the module's `invalidate` method to avoid the race condition?

React/Base/RCTModuleData.mm
```diff
@@ -330,15 +333,17 @@ - (void)gatherConstants
...
+    id instance = self.instance;
+    RCTAssert(_methodQueue != nullptr, @"Module %@ has no methodQueue (instance: %@)",
+              self, instance);
+  }
   return _methodQueue;
 }
 
 - (void)invalidate
 {
-  _methodQueue = nil;
```
fkgozali: Is there any reliable way to use `_methodQueue` as an indication that the module is "dead" instead of adding a new `_valid` ivar? No big deal, but would be better to re-use existing ivars if possible.

React/CxxBridge/RCTCxxBridge.mm
aleclarson: It *could* be moved back to where it was, as long as it's called whether or not `hasInstance` returns YES.

The race condition is fixed by using `m_moduleData.valid` (which is set in the JS thread) instead of `m_bridge.valid` (which is not).

React/Base/RCTModuleData.mm
aleclarson: The guarantee that `_valid` is always set on the JS thread is one I would like to keep. We can't guarantee that for `_methodQueue` since a native module could theoretically change its method queue from another thread.

aleclarson: Oh, it seems that `_methodQueue` cannot be mutated by the native module's `_instance`, so we could use it here.

aleclarson:
I've removed the warning entirely. We now rely on the `methodQueue` getter returning nil (without throwing) when the bridge is invalid and either (1) the `_instance` has not been created or (2) the `_methodQueue` was set to nil by `invalidate`.

matthargett:
I would prefer to have a warning, even just in dev mode. We’ve seen things like this happen when there’s memory corruption issues within JSC or CxxReact. Being able to catch the first symptoms earlier so a dump can be created and analyzed would help.

aleclarson:
@matthargett Okay, fixed.

Issue #22691, opened 2018-12-18
Auto correction doesn't call onChangeText


yoman07:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
```
React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
      Memory: 60.66 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.3 - /usr/local/bin/node
      Yarn: 1.12.3 - /usr/local/bin/yarn
      npm: 6.1.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      @storybook/react-native: ^3.3.12 => 3.4.5 
      @types/react: ^16.0.33 => 16.3.14 
      @types/react-native: ^0.51.8 => 0.51.10 
      react: 16.6.1 => 16.6.1 
      react-native: 0.57.7 => 0.57.5 
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-create-library: 3.1.2
      react-native-gesture-handler: 1.0.10
      react-native-git-upgrade: 0.2.7

```
## Description
Autocorrection for iOS doesn't call onChangeText for TextInput.

## Reproducible Demo

If I select text from autocorrection onChangeText doesn't work. 

Example: 
If I click "armed"  onChangeText won't be called.

![simulator screen shot - iphone xs max - 2018-12-18 at 16 20 59](https://user-images.githubusercontent.com/1101367/50163556-fa7df780-02e0-11e9-9858-714656a6bfdd.png)

imnapo:
Same issue here. it only fires onChangeText() function when it automatically adds an extra space after autocorrection.

PR #23666, opened 2019-02-27
Fix autocorrect onChangeText firing (#22691)

ericlewis:
## Summary

This fixes an issue where autocorrect changes may cause `onChangeText` to not fire. It subscribes to the underlying `textStorage` of the UITextView. It is possible this fixes a few other issues, due to the nature of listening to the underlying textStorage changes. 

## Changelog

[iOS] [Fixed] - iOS firing onChangeText when choosing autocorrect options

## Test Plan

Open RNTester, and interact with textfields.

### DIFF
--- a/Libraries/Text/TextInput/RCTBackedTextInputDelegateAdapter.m
+++ b/Libraries/Text/TextInput/RCTBackedTextInputDelegateAdapter.m
@@ -136,12 +136,11 @@ static void *TextFieldSelectionObservingContext = &TextFieldSelectionObservingCo
 
 #pragma mark - RCTBackedTextViewDelegateAdapter (for UITextView)
 
-@interface RCTBackedTextViewDelegateAdapter () <UITextViewDelegate>
+@interface RCTBackedTextViewDelegateAdapter () <UITextViewDelegate, NSTextStorageDelegate>
 @end
 
 @implementation RCTBackedTextViewDelegateAdapter {
   __weak UITextView<RCTBackedTextInputViewProtocol> *_backedTextInputView;
-  BOOL _textDidChangeIsComing;
   UITextRange *_previousSelectedTextRange;
 }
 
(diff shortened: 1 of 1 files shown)


ericlewis:
fixes #22691

ericlewis:
Would appreciate some extra eyes on this, it appears to work, though.

shergin:
cc @rigdern, @mandrigin 

That's really cool. I didn't know that the API that we use here exists! That also probably can help us to fix other Korean/Japanese/Chinese issues.

I will test it and hopefully land soon.

Libraries/Text/TextInput/RCTBackedTextInputDelegateAdapter.m
```diff
@@ -150,11 +136,21 @@ - (instancetype)initWithTextView:(UITextView<RCTBackedTextInputViewProtocol> *)b
...
 
   return self;
 }
 
+#pragma mark -
+- (void)textStorage:(NSTextStorage *)textStorage
+  didProcessEditing:(__unused NSTextStorageEditActions)editedMask
+              range:(__unused NSRange)editedRange
+     changeInLength:(__unused NSInteger)delta {
+  [_backedTextInputView.textInputDelegate textInputDidChange];
```
shergin: I am worried a bit about over-firing. So, e.g. if only text attributes were changed, we don't want to send an event.

ericlewis: I will check if this is the case & fix it if that does occur.

ericlewis:
@shergin I changed this PR- I originally removed code from the UITextField adapter implementation, this was a mistake I believe. So, listening to the NSTextStorage only applies to multiline TextInput components. I feel like it would make sense for us to use UITextView instead of UITextField for all TextInput's due to UITextView having lower level API access.

ericlewis:
@zhongwuzw ah yes, I forgot to set the delegate in my cleaned up version. I will fix that. Good point on the remaining issues.

PR #23687, opened 2019-02-28
[iOS] Fixed onChange prop of textinput not be called in some cases

zhongwuzw:
## Summary

TextInput not call `onChange` prop in some cases, the reason is `textInputDidChange` is not be called, for example select some text and click the text in keyboard predictive text.

I have the `PR` #23472 , it fix the correction of `_predictedText`, I think we need to fix that firstly 😂 .

## Changelog

[iOS] [Fixed] - Fixed onChange prop of textinput not be called in some cases

## Test Plan

`onChange` prop can be called when text changed.

image


ericlewis:
I think this is a duplicate of #23666, interesting way to solve it though!

zhongwuzw:
@ericlewis Oops, I'll see your `PR`. 😂

ericlewis:
@zhongwuzw my solution is only for multi line text inputs :(

zhongwuzw:
@ericlewis I give some my opinions in your `PR`, please check that! ❤️

zhongwuzw:
Seems #23666 has been closed by @ericlewis , so we can review this implementation? 🤔

zhongwuzw:
CC. @cpojer @shergin Needs some guy to review. 😂

zhongwuzw:
@shergin How about this way, to fix `textInputDidChange` not be called in some cases, we do it by ourself.

shergin:
Feel free to open a new one or reopen this one once we have a proper reliable fix for that,

zhongwuzw:
After some try, I think https://github.com/facebook/react-native/pull/23687/commits/f032e7afafda5f4683574fb5699d25ea364829d2 is more suitable to solve this issue.

zhongwuzw:
@shergin Hi :) , the reason I reopened is I think `onChangeText` may a common event for `TextInput`, can we fix it firstly? maybe we can optimize it later if anyone can find the better solution.

Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -392,6 +392,16 @@ - (BOOL)textInputShouldChangeTextInRange:(NSRange)range replacementText:(NSStrin
                                    notifyDelegate:YES];
 
         [self textInputDidChange];
+        
+        _nativeEventCount++;
+        
+        if (_onChange) {
```
shergin: I am not comfortable with calling `onChange` inside `textInputShouldChangeTextInRange`.

As I said previously, all quicky behavior should be implemented in delegate adapter classes.
I do want to fix this issue very very badly but we cannot sacrifice maintainability of the code for this.

PR #23688, opened 2019-02-28
Move eslint/prettier config to packages/eslint-config-react-native

dratwas:
## Summary

Basing on discussion from this issue - https://github.com/react-native-community/discussions-and-proposals/issues/65 we moved ESLint/Prettier config to its own package inside packages directory.
At this moment we had to left all `eslint` dependencies in root `package.json` since `eslint-config-react-native` is not published yet.

## Changelog

[General] [Changed] - Move eslint/prettier config to packages/eslint-config-react-native

## Test Plan
Lint returns same numbers as before
`yarn lint` output:  

✖ 80 problems (0 errors, 80 warnings)


michalchudziak:
Thanks, @dratwas! 

@matt-oakes @cpojer could you take a look? We'll probably need your help with `BUCK` file if it's required. 

Also, we were wondering if the new config should be included as a dependency in root `package.json`? For now, we just extended eslint config with a relative path to the package. We believe that it might not be the ideal solution. WDYT?

matt-oakes:
Looking good!

React Native has some custom prettier configuration in the root `package.json` (https://github.com/facebook/react-native/blob/master/package.json#L13-L20).

With this PR, the Prettier config for core React Native and users of this ESLint configuration would be different unless the users include the same Prettier configuration.

It is possible to specify the options in this ESLint config, however, it is [not recommended](https://github.com/prettier/eslint-plugin-prettier#options) because most editor plugins will not read the options and will give inconsistent results.

As I see it, there are two options:

1) Ensure that all users of this ESLint config also have the same PRettier config included.
2) Accept that the prettier config might be different across projects, but that it will at least be standard within the project.
3) Change the React Native core to just use prettier with the standard configuration and no customisation to ensure that it is consistent by default.

gengjiawen:
@matt-oakes We have `eslint-plugin-prettier`, there should be no format collision https://github.com/facebook/react-native/blob/fc91bccd5333af88e49ecb19f56c5d6bb0931b1e/package.json#L124

matt-oakes:
@gengjiawen The plugin is there, but I think that the core React Native repository has some custom settings, seen here in the `package.json` https://github.com/facebook/react-native/blob/master/package.json#L13-L20.

These will be correctly picked up when the core React Native repository runs ESLint, but if a community module uses this ESLint config then they will, by default, get the standard Prettier configuration. To fully match the React Native code style, they will need to include the same configuration in their repository.

I should say, this is all guesswork on my part, I've not actually tested that this is the case, however, from the Prettier and `eslint-plugin-prettier` documentation, that should be the case.

TheSavior:
We could also just duplicate the config into the eslint config and ensure that it stays consistent, right?

PR #23705, opened 2019-03-01
SysTraceSection.h: Improve 'unused' macro portability

empyrical:
## Summary

#23588 made use of the macro `__unused` to help suppress noisy compiler warnings. However, this is a GCC-only macro, so this causes compile errors in some environments.

This pull request introduces an `RCT_UNUSED` macro in `SysTraceSection.h` that will use `__unused` in GCC environments, and be a no-op in other environments. 

## Changelog

[General] [Fixed] - `SysTraceSection.h`: Improved portability

## Test Plan

No regressions have been noticed.

### DIFF
--- a/ReactCommon/cxxreact/SystraceSection.h
+++ b/ReactCommon/cxxreact/SystraceSection.h
@@ -9,8 +9,12 @@
 #include <fbsystrace.h>
 #endif
 
-#ifdef __OBJC__
-#define RCT_UNUSED ___unused
+#ifdef __unused
+#define RCT_UNUSED __unused
+#elif __cplusplus >= 201703L
+#define RCT_UNUSED [[maybe_unused]]
+#elif defined(__GNUC__)
+#define RCT_UNUSED __attribute__((unused))
 #else
 #define RCT_UNUSED
 #endif
--- a/ReactCommon/cxxreact/SystraceSection.h
+++ b/ReactCommon/cxxreact/SystraceSection.h
@@ -9,6 +9,12 @@
 #include <fbsystrace.h>
 #endif
 
+#ifdef __OBJC__
+#define RCT_UNUSED ___unused
+#else
+#define RCT_UNUSED
+#endif
+
 namespace facebook {
 namespace react {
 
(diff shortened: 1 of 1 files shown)


ReactCommon/cxxreact/SystraceSection.h
```diff
@@ -9,6 +9,12 @@
 #include <fbsystrace.h>
 #endif
 
+#ifdef __OBJC__
+#define RCT_UNUSED ___unused
```
zhongwuzw: Good to improve portability 👍 , just only `__OBJC__` may defined this macro? I think it's the compiler attribute, `GCC` or other compiler not support?

zhongwuzw: And it's `__unused`, not `____unused`.

empyrical: I messed that up, thanks for pointing that out! The fact that this didn't cause any CI failures means that `ifdef __OBJC__` wasn't even working on iOS, so I have to figure something else out.

As for other platforms, C++17 has the `[[maybe_unused]]` attribute and GCC has the `__attribute__((unused))` that could possibly be added in to this macro on available platforms

empyrical: `__unused` is simply a macro, so I tweaked it to just check for its definition instead.

ReactCommon/cxxreact/SystraceSection.h
```diff
@@ -26,7 +36,7 @@ struct ConcreteSystraceSection {
 public:
   template<typename... ConvertsToStringPiece>
   explicit
-  ConcreteSystraceSection(__unused const char* name, __unused ConvertsToStringPiece&&... args)
+  ConcreteSystraceSection(RCT_UNUSED const char* name, RCT_UNUSED ConvertsToStringPiece&&... args)
```
zhongwuzw: I think we need to change all `__unused` macro in `ReactCommon` if it has portable issue.

empyrical: Oh - I did not notice that it was in other files too! I am trying to just build the `fabric` stuff on MSVC right now, and this is the only `cxxreact` file directly depended on by Fabric (right now).

This tweak should be done to other files, but I would like to keep this pull request small for now - will make it a bit simpler to figure out if the macro I made will have issues when it is merged and ran through Facebook's internal tests

shergin:
As we discussed that with @empyrical in private conversation:
 * RCT_UNUSED thing seems generic, so it deserves the own file and probably should be in `/utils` BACK package.
 * Should be called `RN_UNUSED`. We use/will_use `RN_` as a prefix for new shiny C++ code from existing Objective-C code.
 * We have Systrace thing inside `fabric/debug` module. That should be consolidated first. I am not sure if `systrace` is more `debug` or `utils`, so up to the PR author.

PR #23711, opened 2019-03-01
[iOS] Fixed crash if event doesn't have body parameter

zhongwuzw:
## Summary

We need to check `body[@"target"]` wether equal to `nil`, if it is, return to prevent crash.

## Changelog

[iOS] [Fixed] - Fixed crash if event doesn't have body parameter

## Test Plan

If `body[@"target"]` is `nil`, not crash.


hramos:
Heads up @zhongwuzw, `sendInputEventWithName:body:` has been deprecated for a while, and will be removed when #15894 lands today.

zhongwuzw:
@hramos Oh, get it.

PR #23717, opened 2019-03-01
Fabric: Remove designated initializers in '/uimanager/'

empyrical:
## Summary

This pull request removes the designated initializers under the `fabric/uimanager/` folder. This will help improve the portability of these files.

## Changelog

[General] [Fixed] - Removed designated initializers under the `fabric/uimanager/` folder

## Test Plan

The Fabric test suite runs successfully without regressions.

### DIFF
--- a/ReactCommon/fabric/uimanager/ComponentDescriptorRegistry.cpp
+++ b/ReactCommon/fabric/uimanager/ComponentDescriptorRegistry.cpp
@@ -113,12 +113,13 @@ SharedShadowNode ComponentDescriptorRegistry::createNode(
   ComponentName componentName = componentNameByReactViewName(viewName);
   const SharedComponentDescriptor &componentDescriptor = (*this)[componentName];
 
-  SharedShadowNode shadowNode = componentDescriptor->createShadowNode(
-      {.tag = tag,
-       .rootTag = rootTag,
-       .eventEmitter =
-           componentDescriptor->createEventEmitter(std::move(eventTarget), tag),
-       .props = componentDescriptor->cloneProps(nullptr, RawProps(props))});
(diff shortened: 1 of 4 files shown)


sahrens:
Where are you seeing portability issues? Also, did you consider other resolutions? I'd rather not lose the static safety of named initializers if we can avoid it. For example, could use `:`, lambda's, or simply fix the initialization order instead, as suggested in https://stackoverflow.com/questions/31215971/non-trivial-designated-initializers-not-supported

empyrical:
On Windows, with MSVC 2017 - it currently doesn't have support for designated initializers.

empyrical:
That immediately initialized lambda solution looks like something that might work nicely! 

e.g. This chunk:

```cpp
  auto newShadowNode = shadowNode->clone({
      /* .tag = */ ShadowNodeFragment::tagPlaceholder(),
      /* .rootTag = */ ShadowNodeFragment::surfaceIdPlaceholder(),
      /* .props = */ ShadowNodeFragment::propsPlaceholder(),
      /* .eventEmitter = */ ShadowNodeFragment::eventEmitterPlaceholder(),
      /* .children = */ ShadowNodeFragment::childrenPlaceholder(),
      /* .localData = */ ShadowNodeFragment::localDataPlaceholder(),
      /* .state = */ state,
  });
```

could become:

```cpp
  auto newShadowNode = shadowNode->clone([](State::Shared state){
      ShadowNodeFragment node{};
      node.state = state,
      return node;
  }(state));
```

(could also be changed to a capturing lambda to make it less verbose - not sure how well the compiler would optimize that)

And would work fine on MSVC too.

Would you like to see the initializers ported over to something like that? Right now, the comment style I used on these initializers is using a style that was suggested by @Orvid because it will allow them to be easily code-modded back into designated initializers later on

PR #23721, opened 2019-03-01
[Lean Core] Move RCTTest & takeSnapshot to RNTester

ericlewis:
## Summary
Part of: #23313. 

This moves the `RCTTest` lib from `Libraries/RCTTest` to `RNTester/RCTTest`. This also removes `takeSnapshot` from React Native, and implements it as a standalone module in RNTester called `ScreenshotManager`.

## Changelog

[General] [Removed] - RCTTest & ReactNative.takeSnapshot

## Test Plan

Run the iOS RNTester tests, verified that the snapshot taker was still working when pointed at the new native module.


ericlewis:
I think the failing test is unrelated:
`cc1plus: all warnings being treated as errors` on android.

ericlewis:
@cpojer giving you a poke on this.

PR #23722, opened 2019-03-01
Create a cross-platform accessibility example module

marcmulcahy:
## Summary

This PR divides the accessibility tests into two activities-- a cross-platform activity for accessibility features which are expected to work on all platforms, and platform specific tests for Android and iOS.

We believe that most, if not all, accessibility features should be cross-platform, with fallback implementations where the underlying concept doesn't exist on a particular platform.

This division of the tests makes it clearer for developers which features are expected to work on all supported platforms, and which are platform-specific.

## Changelog

[CATEGORY] general, Android, iOS
[TYPE] change

Message

Refactor the RNTester accessibility activities to better represent where the features are expected to work. Moves all cross-platform tests to AccessibilityExample.js,  Android-specific tests to AccessibilityAndroidExample.android.js, and iOS-specific tests to AccessibilityIOsExample.ios.js.

## Test Plan

Verified that tests work on iOS with VoiceOver and Android with TalkBack.

### DIFF
--- a/RNTester/js/AccessibilityAndroidExample.android.js
+++ b/RNTester/js/AccessibilityAndroidExample.android.js
@@ -35,32 +35,6 @@ class AccessibilityAndroidExample extends React.Component {
     count: 0,
     backgroundImportantForAcc: 0,
     forgroundImportantForAcc: 0,
-    screenReaderEnabled: false,
-  };
-
-  componentDidMount() {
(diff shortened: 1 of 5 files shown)


marcmulcahy:
## changelog

[general] [change] - Re-factor accessibility tests into cross-platform and platform-specific tests.

PR #23724, opened 2019-03-01
Remove create-react-class dependency

ericlewis:
## Summary
Remove create-react-class from the core of RN. It is only used in a few components, and removes it from package.json as well. This saves us 4.19mb after install thanks to the removal of conflicting packages.

Note: it uses the library react-mixin, it is quite tiny. I am open to other ideas of course. 

## Changelog

[General] [Removed] - remove create-react-class dependency from core

## Test Plan

Interact with the RNTester app, let CI do its thang.


ericlewis:
the android error is unrelated.

TheSavior:
I think we would rather switch these from using mixins at all. This is what we did for all the other components in the codebase. See https://github.com/facebook/react-native/issues/21581 and https://github.com/facebook/react-native/issues/21485 for details. 

In order to do this we need to migrate off of TouchableMixin and convert it to something else. We've done this work internally and have been testing it. I believe moving it to open source is part of @cpojer's roadmap. 

I'm pretty hesitant to switching to use a personally scoped package in the mean time. However, I think moving connect to devDeps is reasonable to do in a standalone PR since it looks like nothing uses it as a regular dependency.

ericlewis:
@TheSavior the personally scoped package is based on another package, with the polyfills removed. I’m happy to use the “supported” package. But it is quite old, (and the polyfills that aren’t needed ofc). 

I’ll take a look at your examples and see if I can adapt to using them instead of the package. Will also open a sep PR for connect.

TheSavior:
Do you have measurements on what the bundle size impact would be if the Touchables (and thus createreactclass) were pulled out from React Native?

ericlewis:
@TheSavior ~5MB saved when installed with `—production` if we remove both the dep I introduced & the create-react-class. Honestly, it almost seems worth it to remove.. esp with fabrics touch events coming on views.

ericlewis:
I used cost-of-modules to determine this.

ericlewis:
Oh, I’m sorry- bundle size is a different thing yes?

TheSavior:
Saves 5 MB where? In the node_modules folder?

If you create a production bundle with metro for an app created with React-Native init both before and after this change, what is the size of the bundle? I doubt the before bundle is 5MB so I’m not sure what the win is here to the number that really matters.

I think we should fix these files and get them off createReactClass but the urgency is different depending on the bundle size impact.

cpojer:
I think saving 5 mb of dev dependencies itself is worth it. However, my hesitation here comes from changing these modules. I’m worried about introducing subtle breakages just to reduce size. I agree with Eli that the right move is to deprecate some stuff and remove these modules altogether.

That being said, I think it’s totally worth removing the mixin and createReactClass from TextInput. Could you try to find a way to do this safely and without introducing new dependencies?

PR #23725, opened 2019-03-02
YGJNI.cpp: fix build failure

empyrical:
## Summary

This pull request introduces a `YG_UNUSED` macro into `YGJNI.cpp`. This will suppress these errors that was causing this build failure:

```
ReactAndroid/src/main/jni/first-party/yogajni/jni/YGJNI.cpp:792:15: error: variable ‘edge’ set but not used [-Werror=unused-but-set-variable]
         float edge = *styleInputs++;
               ^~~~
```

## Changelog

[General] [Fixed] - Fixed CI failure in `YGJNI.cpp`

## Test Plan

File builds successfully

### DIFF
--- a/ReactAndroid/src/main/jni/first-party/yogajni/jni/YGJNI.cpp
+++ b/ReactAndroid/src/main/jni/first-party/yogajni/jni/YGJNI.cpp
@@ -11,6 +11,8 @@
 #include <iostream>
 #include <map>
 
+#define YG_UNUSED(x) (void) (x);
+
 using namespace facebook::jni;
 using namespace std;
 using facebook::yoga::detail::Log;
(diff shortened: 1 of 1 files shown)


dulmandakh:
how about showing warning for unused variables, instead of errors. this might obscure issues in the future.

empyrical:
I think that changing that behavior will be necessary, since this is a point where the OSS build-scripts differ from FB's internal scripts.

cpojer:
We actually sync Yoga from within fb to multiple repos, so I'm not sure what the right process is to land changes to Yoga.

cc @davidaurelio – this broke our CI a few days ago.

empyrical:
Should I close this and send this as a pull to @facebook/yoga instead?

cpojer:
Yeah, let's do that.

PR #23726, opened 2019-03-02
[iOS] Fixed method invoke when both nullable and __unused exist

zhongwuzw:
## Summary

We have parse code to try to support exported method parameter annotated `__unused` and `nullable`, but actually, seems we don't support it entirely, for example, we parse `__unused` firstly, then we parse `nullable`, actually, this case not exist, because `nullable` macro must appear right after `(`, before `__unused`, so we can't parse success, leads to error signature and crash.

## Changelog

[iOS] [Fixed] - Fixed method invoke when both nullable and __unused exist

## Test Plan

Support module export method which parameter support both `__unused` and `nullable`.

### DIFF
--- a/React/Base/RCTModuleMethod.mm
+++ b/React/Base/RCTModuleMethod.mm
@@ -90,8 +90,8 @@ static BOOL RCTParseSelectorPart(const char **input, NSMutableString *selector)
 
 static BOOL RCTParseUnused(const char **input)
 {
-  return RCTReadString(input, "__unused") ||
-         RCTReadString(input, "__attribute__((unused))");
+  return RCTReadString(input, "__attribute__((unused))") ||
+         RCTReadString(input, "__unused");
 }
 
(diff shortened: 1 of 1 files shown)


React/Base/RCTModuleMethod.mm
```diff
@@ -90,8 +90,8 @@ static BOOL RCTParseSelectorPart(const char **input, NSMutableString *selector)
 
 static BOOL RCTParseUnused(const char **input)
 {
-  return RCTReadString(input, "__unused") ||
-         RCTReadString(input, "__attribute__((unused))");
+  return RCTReadString(input, "__attribute__((unused))") ||
+         RCTReadString(input, "__unused");
```
zhongwuzw: Actually `__unused` may not happen, because `__unused` macro would be expanded to `__attribute__((unused))` by compiler automatically.

PR #23730, opened 2019-03-03
chore: update comment in DrawerLayoutAndroid

FLGMwt:
...referencing stale function component limitations

## Summary

While preparing a short tutorial on the DrawerLayoutAndroid using the v59 RC, I came across a comment that we can update that states refs as being available only in Class components. This updates that with both class and function components.

If we'd rather delete the comment entirely and move both examples to the website, I'm fine with that too. The current docs page doesn't address the imperative API. https://facebook.github.io/react-native/docs/drawerlayoutandroid

## Changelog

[General] [fixed] - Update stale comment to provide both class and functional component ref example for DrawerLayoutAndroid

## Test Plan

This code only affects source code comments in one file. I wouldn't expect any testing to be necessary.


FLGMwt:
Also, lemme know if this is an annoying PR, just trying some slight improvements 😬

PR #23780, opened 2019-03-06
use TraceCompat

dulmandakh:
## Summary

I was working to use more Android Support Library in RN core, but wasn't able to finish it. So this PR is for only Systrace and TraceCompat only.

## Changelog

[Android] [Changed] - Use TraceCompat in Systrace

## Test Plan

CI is green, and everything works as expected.


hramos:
This was reverted in 2a80a42d40a2930bee107187dcc121a661b245f5. Do you think this change can be made while making use of AndroidX?

Issue #22211, opened 2018-11-08
SafeAreaView doesn't respect `padding` property in style


dani-mp:
## Environment
```
React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
      Memory: 34.00 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 9.7.1 - ~/.nvm/versions/node/v9.7.1/bin/node
      Yarn: 1.2.1 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v9.7.1/bin/npm
      Watchman: 4.7.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
      Android SDK:
        Build Tools: 21.1.2, 22.0.1, 23.0.1, 23.0.3, 25.0.0, 25.0.2, 26.0.1, 26.0.2, 27.0.1, 27.0.3, 28.0.2, 28.0.3
        API Levels: 17, 18, 19, 22, 23, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.0/10A254a - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.0 => 16.6.0
      react-native: ^0.57.3 => 0.57.3
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description
Applying `padding` to SafeAreaView's style doesn't work.

## Reproducible Demo
https://snack.expo.io/@danielmartin/c2FmZW

bartolkaruza:
> renders nested content and automatically applies paddings

Seems like intended behaviour from the [docs](https://facebook.github.io/react-native/docs/safeareaview.html). Try adding padding to your view inside the `SafeAreaView` instead.

danielmartinprieto:
Thanks for the response, @bartolkaruza. I was wondering if it would be better if the padding applied by the consumer was added to the safe area one, instead of being overwritten. Thinking about it, the way the SafeAreaView component achieves its behaviour is more an implementation detail.

bartolkaruza:
[...]

PR #23784, opened 2019-03-06
[iOS] Add additionalSafeAreaInsets prop for SafeAreaView

zhongwuzw:
## Summary

Currently, if we use `SafeAreaView`, we lose the `padding` prop, https://github.com/facebook/react-native/blob/master/React/Views/SafeAreaView/RCTSafeAreaShadowView.m#L36, so the insets can only determined by system, but, actually, user may wants some additional insets based on system safe area, like #22211, and we can also see Apple document, https://developer.apple.com/documentation/uikit/uiviewcontroller/2902284-additionalsafeareainsets?language=objc#, it also has `additionalSafeAreaInsets` for `ViewController`, so I think we can add `additionalSafeAreaInsets` prop for user to do custom insets.

## Changelog

[iOS] [Added] - Add additionalSafeAreaInsets prop for SafeAreaView

## Test Plan

After we added `additionalSafeAreaInsets` prop, the final safe area equals to system safe area plus user custom safe area:

```
      <SafeAreaView style={styles.container} additionalSafeAreaInsets={{top: 50, left: 0, bottom: 0, right: 0}}>
        <View style={styles.content} />
      </SafeAreaView>

  container: {
    flex: 1,
    backgroundColor: 'red',
    padding: 50, // This padding is not respected
  },
  content: {
    flex: 1,
    backgroundColor: 'yellow',
  },
```

image

### DIFF
--- a/Libraries/Components/SafeAreaView/RCTSafeAreaViewNativeComponent.js
+++ b/Libraries/Components/SafeAreaView/RCTSafeAreaViewNativeComponent.js
@@ -12,10 +12,12 @@ const requireNativeComponent = require('requireNativeComponent');
 
 import type {ViewProps} from 'ViewPropTypes';
 import type {NativeComponent} from 'ReactNative';
+import type {EdgeInsetsProp} from 'EdgeInsetsPropType';
 
 type NativeProps = $ReadOnly<{|
   ...ViewProps,
   emulateUnlessSupported?: boolean,
+  additionalSafeAreaInsets?: ?EdgeInsetsProp,
 |}>;
 
 type RCTSafeAreaViewNativeType = Class<NativeComponent<NativeProps>>;
--- a/Libraries/Components/SafeAreaView/SafeAreaView.js
+++ b/Libraries/Components/SafeAreaView/SafeAreaView.js
@@ -13,10 +13,21 @@ const React = require('React');
 const View = require('View');
 
 import type {ViewProps} from 'ViewPropTypes';
+import type {EdgeInsetsProp} from 'EdgeInsetsPropType';
 
 type Props = $ReadOnly<{|
   ...ViewProps,
   emulateUnlessSupported?: boolean,
+|}>;
+
+type IOSProps = $ReadOnly<{|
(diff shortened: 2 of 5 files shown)


janicduplessis:
@zhongwuzw Would it be possible to just support padding instead of adding a new prop that basically does the same? I think we override the padding in the shadow node so we could just add safe area and the current padding.

zhongwuzw:
@janicduplessis Emm, my opinions based on two reasons:
1. We already removing support for `padding` in safe area view, please see https://github.com/facebook/react-native/blob/master/React/Views/SafeAreaView/RCTSafeAreaShadowView.m#L36, so at least, if we reuse `padding`, I think it's the API breaking?
2. I added the same prop as Apple does, it provide `additionalsafeareainsets` prop for custom insets, https://developer.apple.com/documentation/uikit/uiviewcontroller/2902284-additionalsafeareainsets?language=objc#.

So I did this `PR`. 😂

shergin:
* As Janic said, there is no need to introduce new prop names, we can theoretically reuse `padding`.
* On iOS, additionalSafeAreaInsets is for communicating safe area constrains to other view; RN does not have such concept at all, RN uses regular padding that applies only for SafeAreaView.
* I think there is no harm to use additional View to add additional padding.
* I think we should try to implement paddings for SafeAreView that **replace** the SafeArea-based ones if they are not zeros.

janicduplessis:
@shergin @zhongwuzw What about we add the padding with the safe area insets? There was a couple times I wanted to add padding to a SafeAreaView and the behaviour I was expecting was that.

A common case would be a footer bar:

```js
<SafeAreaView style={{padding: 16}}>...</SafeAreaView>
```

In this case the padding bottom would be safeAreaInsets.bottom + 16

zhongwuzw:
@janicduplessis Before I added a new prop `additionalSafeAreaInsets`, I considered re-enable support of `padding`, at that time, I worry about it has breaking issue, but if it's acceptable, I agree to re-enable it.

PR #23789, opened 2019-03-06
Use Backlog label to keep issues open

hramos:
## Summary

The Stale bot closes issues that have been inactive for 97 days (90 days until marked as stale, then 7 days of no activity before it is closed). This helps us focus on issues that affect new releases, and gets rid of drive-by issues where there has been no follow up. This process, however, does not acknowledge there's a type of issue that trascends releases and which the team might want to keep track of in order to fix in a future release.

The "Resolution: Backlog" label can be used to mark these types of issues, and keep the bot from closing them.

## Changelog

[General] [Added] - GitHub: Add Backlog label

## Test Plan

N/A

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -12,12 +12,13 @@ exemptLabels:
   - "Help Wanted :octocat:"
   - "Impact: Regression"
   - "Resolution: PR Submitted"
+  - "Resolution: Backlog"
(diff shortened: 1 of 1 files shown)


hramos:
@react-native-bot import

hramos:
I'm landing this, it's a trivial config change to a non-core bot mechanism.

PR #23796, opened 2019-03-06
Fix PickerAndroid breaking when child is null. 

a-c-sreedhar-reddy:
## Summary

On conditional rendering if child is null then the PickerAndroid breaks.
## Changelog
when conditional rendering is used then picker breaks when the child is null.

This conditional rendering inside Picker fails when a is 1, because child will be null in PickerAndroid.android.js.
```
{
   this.state.a === 2 && <Picker.Item label="value" value="value" />
}
```

[ANDROID] [FIXED] - Filter props.children.


cpojer:
Unfortunately, I don't think checking for `0` is correct. The right way of fixing this would likely be to filter for `null`. Feel free to send a PR with the proper fix (and make sure to run prettier using `yarn lint --fix`.

PR #23802, opened 2019-03-07
Fabric: add podspecs

ericlewis:
## Summary

This adds up to date podspecs for compiling fabric. Special thanks to @kmagiera which did the original libevent & folly header changes.

## Changelog

[iOS] [Added] - Fabric podspecs

## Test Plan

Added the following to RNTester and ran pod install then compiled.
```
pod 'React-Fabric', :path => '../ReactCommon'
pod 'React-graphics', :path => '../ReactCommon/fabric/graphics'
pod 'React-RCTFabric', :path => '../React'
pod 'libevent', :podspec => '../third-party-podspecs/libevent.podspec'
```


ericlewis:
This has a breaking change for Folly FYI, you will be required to include libevent in your podspec along side the other 3rd party libs.

ericlewis:
Note: this will have trouble compiling now, due to the fact I am pulling in codegen again. A prerequisite would be to properly run the expected codegen before pod install.

ericlewis:
Closing in favor of #23802.

PR #23803, opened 2019-03-07
Fabric: working podspecs & works in RNTester

ericlewis:
## Summary

This is the couple of hacks I used after I finished #23802 in order to get fabric working on RNTester. This is inspired from prior work by @kmagiera. 

The goal of this PR is to show others what I’m struggling with, and to eventually merge it sans hacks. 

### Directions to use
- Yarn Install
- Uncomment the commented out pods in RNTester's pod file
- Open RNTesterPods workspace
- Run App

### Notes
- this is only for pods, the non-pod RNTester will no longer work until updated with fabric too. 
- `SurfaceHostingView` & `SurfaceHostingProxyRootView` both try to start the surface immediately, this leads to a race condition due to the javascript not having loaded yet, the hack here is:
   1. Swizzle the `start` method on `RCTFabricSurface` to no-op when called.
   2. Add observer for `RCTJavaScriptDidLoadNotification`
   3. Call private method `_startAllSurfaces` on `_surfacePresenter` in AppDelegate when we receive `RCTJavaScriptDidLoadNotification`.

## Changelog

[General] [Added] - Use Fabric in RNTester

## Test Plan

Pod install, and run RNTester.


ericlewis:
The missing rncore headers, such as the one here: https://github.com/facebook/react-native/blob/cf3653baac43d208cff29b2aa9cb803856372ffc/React/Fabric/Mounting/ComponentViews/ActivityIndicator/RCTActivityIndicatorViewComponentView.mm#L10

Seem to be an easy fix, just missing headers from what appears to be codegen. Would be awesome to learn how to generate those files!

JoshuaGross:
Wow! Clearly a lot of work went into this. This is awesome, thanks for doing it.

ericlewis:
The `react-native` side of things is now hack free! RNTester implements the RCTSurface race condition hack itself, so no need to dirty react native code :)

The podspecs are complete, I am not sure if it is optimal to have a Folly Fabric podspec, but it is easiest without breaking existing users. We can converge once fabric is "done". I think it is fairly safe to import everything except for the changes to RNTester- as they are hacks.

ericlewis:
Note to self: simplify folly pods, should be able to eliminate libevent

package.json
```diff
@@ -127,6 +127,7 @@
     "eslint-plugin-react-native": "3.5.0",
     "eslint-plugin-relay": "0.0.28",
     "flow-bin": "^0.94.0",
+    "flow-remove-types": "^1.2.3",
```
fkgozali: what is this for? is it to run the codegen?

scripts/generate-rncore.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
```
fkgozali: can you put short comment explaining what this script does?

third-party-podspecs/libevent.podspec
```diff
@@ -0,0 +1,261 @@
+
+Pod::Spec.new do |s|
+
+  s.name         = "libevent"
```
fkgozali: so yeah, I'm not sure what pulled this in, but let's find a way to not depend on it

RNTester/RNTester/AppDelegate.m
```diff
@@ -18,8 +18,17 @@
...
-@interface AppDelegate() <RCTBridgeDelegate>
+#import <React/RCTSurfacePresenter.h>
+#import <React/RCTFabricSurfaceHostingProxyRootView.h>
 
+@interface AppDelegate() <RCTBridgeDelegate>{
+  RCTSurfacePresenter *_surfacePresenter;
+}
+@end
+
+// FIXME: this is a hack, remove when surfaces start correctly
```
fkgozali: Could you open an issue then link that issue here? We'll look into it.

RNTester/RNTester/AppDelegate.m
```diff
@@ -28,17 +37,24 @@ - (BOOL)application:(__unused UIApplication *)application didFinishLaunchingWith
...
+  NSDictionary *initProps = @{};
   NSString *_routeUri = [[NSUserDefaults standardUserDefaults] stringForKey:@"route"];
   if (_routeUri) {
     initProps = @{@"exampleFromAppetizeParams": [NSString stringWithFormat:@"rntester://example/%@Example", _routeUri]};
   }
 
-  RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:_bridge
-                                                   moduleName:@"RNTesterApp"
-                                            initialProperties:initProps];
+  UIView *rootView = [[RCTFabricSurfaceHostingProxyRootView alloc] initWithBridge:_bridge moduleName:@"RNTesterApp" initialProperties:initProps];
```
fkgozali: This needs to be configurable - it's ok to include fabric build in RNTester, but it shouldn't default to use Fabric yet (given it's not complete). It should default to the plain old RCTRootView, then via a toggle (a .plist, a global C flag, or something from NSUserDefault + simple UI to toggle, etc.), it should allow activating Fabric.

ericlewis: @fkgozali for sure- I don't think we should merge these "hacks", but things like the podspecs should be good to go.

package.json
ericlewis: Yes, for codegen.

scripts/generate-rncore.sh
ericlewis: Will do!

RNTester/RNTester/AppDelegate.m
ericlewis: Yes, will do that shortly.

RNTester/RNTester/AppDelegate.m
ericlewis: That said, feature flagging it is pretty easy I suppose.

ericlewis: What is your opinion on using a different scheme?

ericlewis: C Flag is easiest heh.

ericlewis: Added a new flag called `FABRIC_ENABLED`.

third-party-podspecs/libevent.podspec
ericlewis: removed it

RNTester/RNTester/AppDelegate.m
fkgozali: C #define'ed flag can work, that way the compiler/build system can influence it.

fkgozali:
One other note:

I think in general we want to make Fabric (including compiling it) optional at the moment, because if you include it the core React.xcodeproj or React podspec, then *every app has to pay the extra cost of having fabric* even if they don't use it (they shouldn't use it for now). This means extra app size or latent bugs for them. So let's split it in this PR.

Given that, I think the ideal place I'd like to go to is:
* Having fabric-only configuration (in podspec, or separate ReactFabric.xcodeproj, etc) that apps can pull in if they decide to activate fabric
* Having these configurations define a compiler flag, e.g. FABRIC_ENABLED, so that apps can choose to activate Fabric surface vs old RootView in their AppDelegate

Now, re: .xcodeproj vs podspec, I'd rather just maintain 1 thing going forward. So if the general community is moving towards CocoaPods, we can just drop support for the fabric in .xcodeproj setup.

Long-long term, I recognize it's tedious to manually "sync" these configurations across BUCK/podspec/gradle, so we can have a separate discussion about either utilizing `buck project`, or even simple parsing of BUCK files to generate the relevant podspec/gradle files.

RNTester/RNTesterPods.xcodeproj/project.pbxproj
```diff
@@ -17,7 +17,9 @@
 		3D13F84A1D6F6AFD00E69E0E /* OtherImages.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 3D13F8451D6F6AF200E69E0E /* OtherImages.xcassets */; };
```
fkgozali: unrelated: I thought we're not supposed to check in .xcodeproj/.xcworkspace file when using cocoapods? should probably remove this in different PR

ericlewis:
@fkgozali That makes sense to me, the only stuff touched here that could affect users currently is some additions to the Folly podspec, would it be ideal if I create an additional folly podspec for only the fabric bits? 

Also: flag is implemented, and disabled by default, the pods are also commented out by default.
Parsing buck would be ideal and probably pretty easy, I can look in to that. Folly podspec stuff is slightly harder to keep in sync, esp without public folly buck scripts.

RNTester/RNTesterPods.xcodeproj/project.pbxproj
ericlewis: Generally you check in the `xcodeproj` but not the `xcworkspace` when working with pods.

fkgozali: ok sounds good

fkgozali:
> Folly podspec stuff is slightly harder to keep in sync, esp without public folly buck scripts.

We're going to fork internal folly subset that we use in RN today to have the exact list that RN needs (Fabric + all existing stuffs). This will happen a bit later this half, but we can share that definition when it happens.

RNTester/Podfile.lock
```diff
@@ -2,6 +2,11 @@ PODS:
   - boost-for-react-native (1.63.0)
   - DoubleConversion (1.1.6)
   - Folly (2018.10.22.00):
+    - boost-for-react-native
+    - DoubleConversion
+    - Folly/Default (= 2018.10.22.00)
+    - glog
+  - Folly/Default (2018.10.22.00):
```
fkgozali: was this an extra alias? I don't see it used

RNTester/Swizzle_RCTFabricSurface.m
```diff
@@ -0,0 +1,43 @@
+#ifdef FABRIC_ENABLED
```
fkgozali: can you add the same copyright header like in other .m/.h?

RNTester/Podfile.lock
ericlewis: This is a Default subspec, so that Folly doesn't install the fabric subspec automatically as well. It is a no-op technically.

ericlewis: I can make it so that it's not a no-op but actually controls what is built. I decided against this originally to prevent any mistakes from developers.

ericlewis:
I can happily say: the failing detox test is unrelated! Will address the shell check momentarily.

RNTester/RNTester/AppDelegate.m
ericlewis: https://github.com/facebook/react-native/issues/23910

RNTester/Podfile.lock
ericlewis: It is better to keep default as a no-op, so user's can't accidentally forget to install Folly normally.

RNTester/ComponentRegistry.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <react/components/text/ParagraphComponentDescriptor.h>
+#include <react/components/text/RawTextComponentDescriptor.h>
+#include <react/components/text/TextComponentDescriptor.h>
+#include <react/components/view/ViewComponentDescriptor.h>
+#include <react/components/slider/SliderComponentDescriptor.h>
+
+namespace facebook {
+  namespace react {
+    
+    /**
```
fkgozali: styling nit, we usually don't indent for namespace stuffs, so it should just be:

```
namespace facebook {
namespace react {

/**
 ...
 */
ComponentRegistryFactory ... {
}

} // namespace react
} // namespace facebook

File: RNTester/RNTester/AppDelegate.m (hunk 4 above)
Comment (fkgozali) [reply RNTester/RNTester/AppDelegate.m]: let's put this link in the FIXME comment: `https://github.com/facebook/react-native/issues/23910`

File: ReactCommon/jsi/JSCRuntime.cpp [hunk 10]
```diff
@@ -819,12 +820,14 @@ jsi::Array JSCRuntime::getPropertyNames(const jsi::Object& obj) {
   return result;
 }
 
-jsi::WeakObject JSCRuntime::createWeakObject(const jsi::Object&) {
-  throw std::logic_error("Not implemented");
+jsi::WeakObject JSCRuntime::createWeakObject(const jsi::Object& obj) {
+  JSObjectRef objRef = objectRef(obj);
+  return make<jsi::WeakObject>(makeObjectValue(objRef));
```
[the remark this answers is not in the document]
fkgozali: Let's add `// TODO: revisit this implementation` for all WeakObject stuffs - we'll take care of it later

ericlewis:
@fkgozali addressed your feedback.

RNTester/ComponentRegistry.cpp
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#ifdef FABRIC_ENABLED
```
shergin: RN_FABRIC_ENABLED

RNTester/RNTesterPods.xcodeproj/xcshareddata/xcschemes/RNTester.xcscheme
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
shergin: Do we really need those xcscheme files to be checked in?

ericlewis: If I don’t check them in, there is no schema to run. This seems to be the case on master too I think, so I suspect something got messed up somewhere along the way *before* these changes.

ReactCommon/jsi/JSCRuntime.cpp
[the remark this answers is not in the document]
fkgozali: I consulted with some folks - let's actually guard this impl with `#ifdef FABRIC_ENABLED` and preserve the `throw std::logic_error(...)` when Fabric isn't enabled. Keep the comment though.

ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -184,7 +185,8 @@ class JSCRuntime : public jsi::Runtime {
   static JSStringRef stringRef(const jsi::String& str);
   static JSStringRef stringRef(const jsi::PropNameID& sym);
   static JSObjectRef objectRef(const jsi::Object& obj);
-
+  static JSObjectRef objectRef(const jsi::WeakObject& obj);
```
fkgozali: guard this defn with `#ifdef FABRIC_ENABLED`

ericlewis: okay cool- I will do that.

ericlewis:
@shergin & @fkgozali addressed both of your feedback

RNTester/RNTesterPods.xcodeproj/xcshareddata/xcschemes/RNTester.xcscheme
fkgozali: Hmm, I don't think this file should be checked in. We should update the .gitignore under RNTester to ignore .xccheme -- I think this scheme gets generated when you pod install?

msand:
Great work guys!

PR #23804, opened 2019-03-07
Resolve relative size rendering error in inspector

### LINKED ISSUE #17496: In-app inspector throws error when assigning percentage strings to padding(like this --> padding: '10%'), although it works in the UI.
### Is this a bug report?

YES

YES

### Environment

Environment:
  OS: macOS Sierra 10.12
  Node: 7.5.0
  Yarn: 0.20.3
  npm: 5.4.2
  Watchman: 4.7.0
  Xcode: Xcode 8.3.1 Build version 8E1000a
  Android Studio: 2.3 AI-162.4069837

  react: 16.2.0 => 16.2.0
  react-native: 0.52.0 => 0.52.0

  Target Platform: iOS (10.3)

### Steps to Reproduce

1. Use padding with percentage string as: `padding: '10%'`. 
2. Now open the inspector using in-app developer menu and tap on the **element with the given padding**
3. The inspector will error out saying that padding should be a number instead of a string, although padding as a string works in the UI.

### Expected Behavior
The inspector shouldn't have thrown an error. It should have work the same way if I have given a padding as a number.

### Actual Behavior
React-Native supports padding using percentage strings but the inspector is throwing error that the padding should be a number rather than a string.
simulator screen shot jan 9 2018 10 23 25 am

### Reproducible Demo
I have added a padding as a string in this app.js file. Just replace the default app.js file with this one.
```js
import {
  Platform,
  StyleSheet,
  Text,
  View
} from 'react-native';

const instructions = Platform.select({
  ios: 'Press Cmd+R to reload,\n' +
    'Cmd+D or shake for dev menu',
  android: 'Double tap R on your keyboard to reload,\n' +
    'Shake or press menu button for dev menu',
});

export default class App extends Component<{}> {
  render() {
    return (
        <View style={{backgroundColor: 'red', flex: 1}}>
        <Text style={styles.welcome}>
          Welcome to React Native!
        </Text>
        <Text style={styles.instructions}>
          To get started, edit App.js
        </Text>
        <Text style={styles.instructions}>
          {instructions}
        </Text>
        </View>
      </View>
    );
  }
}

  container: {
    flex: 1,
    padding:'10%',
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#F5FCFF',
  },
  welcome: {
    fontSize: 20,
    textAlign: 'center',
    margin: 10,
  },
  instructions: {
    textAlign: 'center',
    color: '#333333',
    marginBottom: 5,
  },
});
```

### PR opportunity
I think it is a good first bug. I would love to send a PR for this. 

Thank You 😄

gandreadis:
## Summary

Currently, when relative sizes are given in margin or padding stylings (be it a percentage or an auto measure), the inspector crashes, due to frame rendering not properly handling those kinds of measurements. This PR adds a resolution step for them:

* Percentages are evaluated relative to the window size. 
* I decided to simply not render `auto` margins/paddings, due to the complexities involved (e.g. when the margin is between multiple elements with relative sizes). 

Since the inspector does not crash anymore on relative sizes on paddings or margins, I believe that this addresses #17496.

Fixes #17496

## Changelog

[General] [Fixed] - Fix inspector rendering of relative margins and paddings

## Test Plan

I ran the RNTester app with my adapted version of the Inspector. Before running it, I modified the styling of example components in different ways. I tried the following combinations:

`paddingTop: 5%`
`paddingRight: 5%`
`paddingBottom: 5%`
`paddingLeft: 5%`
`marginTop: 5%`
`marginRight: 5%`
`marginBottom: 5%`
`marginLeft: 5%`
`padding: 5%`
`margin: 5%`
`marginLeft: auto`
`paddingLeft: auto`

They can be seen in that order in the screenshots below:

screenshots


radeno:
@gandreadis great work 👌 if you want automatically close addressed issues which will be fixed by your PR use any magical keyword https://help.github.com/en/articles/closing-issues-using-keywords in your description.
Handy feature.

gandreadis:
@radeno Thanks for the suggestion! I was a bit hesitant to do this, since I know that Facebook procedure is generally to close PRs and then to merge them manually (via a bot), so I was not sure if the GitHub keyword would still work. But I guess we'll see :)

Libraries/Inspector/ElementBox.js
```diff
@@ -31,6 +32,8 @@ class ElementBox extends React.Component<$FlowFixMeProps> {
     };
 
     if (margin != null) {
+      resolveRelativeSizes(margin);
```
cpojer: Could you change this API to return a new object, like `margin = resolveRelativeSizes(margin)`? I think that will be clearer.

Libraries/Inspector/resolveRelativeSizes.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+'use strict';
+
+const Dimensions = require('Dimensions');
+
+/**
+ * Resolves relative sizes (percentages and auto) in a style object.
+ */
+function resolveRelativeSizes(style: Object) {
```
cpojer: Could you change this flow type to have the correct object type information? Something like: 

```js
type Style = {
  height: number,
  top: number,
  …
}
```

?

gandreadis:
Hi @cpojer, thanks for your suggestions! I've tried to implement them, could you have another look?

Issue #23810, opened 2019-03-07
[Animated] crashes on Android when toValue is non Number and using NativeDrivers


andrewschenk-linx:
## 🐛 Bug Report
  The Animated library crashes on Android, RN v0.58.6 when using a non number toValue and native drivers.

## To Reproduce
  This Snack shows an example.  Run it on Android and adjust toValue from a number to a null or string.
[https://snack.expo.io/ry8tHCRLN](https://snack.expo.io/ry8tHCRLN)

## Expected Behavior
  Supplying a toValue of the improper type should fallback to toValue: 0 just like the javascript and iOS drivers do.

## Code Example
  This Snack shows an example.  Run it on Android and adjust toValue from a number to a null or string.
[https://snack.expo.io/ry8tHCRLN](https://snack.expo.io/ry8tHCRLN)

## Environment
   React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (4) x64 Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
      Memory: 20.31 GB / 48.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.0.0 - /usr/local/bin/node
      npm: 6.8.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 22, 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 23.0.2, 24.0.1, 25.0.1, 26.0.1, 26.0.2, 26.0.3, 27.0.2, 27.0.3, 28.0.3
        System Images: android-23 | Intel x86 Atom, android-23 | Intel x86 Atom_64, android-23 | Google APIs Intel x86 Atom_64, android-27 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.1 AI-173.4720617
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.58.6 => 0.58.6 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-create-library: 3.1.2
      react-native-git-upgrade: 0.2.7

TheSavior:
Looks like your PR for this got merged. Closing the issue! Thanks!

PR #23812, opened 2019-03-07
adds type checking to Animated toValue and iterations key values to m…

andrewschenk-linx:
…ake sure Android does not crash from bad params when using useNativeDriver

## Summary

Android apps crash when using Animated useNativeDriver: true and the toValue is not a number.  See issue here with test case.  [Issue](https://github.com/facebook/react-native/issues/23810)

## Changelog

[Android] [fixed] - Fix crash when using Animated with useNativeDriver and a non Number toValue

## Test Plan

Run [Snack](https://snack.expo.io/ry8tHCRLN)  code with patched React Native core.  Malformed parameters will no longer crash Android app.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/animated/FrameBasedAnimationDriver.java
@@ -41,8 +41,17 @@ class FrameBasedAnimationDriver extends AnimationDriver {
     for (int i = 0; i < numberOfFrames; i++) {
       mFrames[i] = frames.getDouble(i);
     }
-    mToValue = config.hasKey("toValue") ? config.getDouble("toValue") : 0;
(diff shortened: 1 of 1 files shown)


matthargett:
Should we show a warning in dev mode when this is the case? It a user has a bug in their JS, they may get confused when zeroes end up getting used silently.

andrewschenk-linx:
My motivation for fixing this is that the JS bug is in one of possibly 100 different dependency libraries.  Showing a warning might help focus attention on which library is the culprit or where in the internal project code the developer went awry.  On the other hand, it might be another yellow box warning that gets silenced because the 3rd party lib is unsupported and the developer is tired of getting nagged about it.

Personally, I would probably want to see a warning and have the option of silencing it or fixing the library.

AndrewSchenk:
The JavaScript animated version does not show a yellow box warning if you supply an invalid type.  i.e.

```
useNativeDriver: false
```

So to maintain parity between JS and native it would be best to not show a yellow box.

Issue #23645, opened 2019-02-25
[0.59rc2 blocking] Can't compile React Native on iOS using `-force_load` instead of `-ObjC` any more


tomduncalf:
## 🐛 Bug Report

Our project can't use the `-ObjC` linker flag to discover/load all symbols as it causes "duplicate symbol" errors for some of the static libraries we use (`ObjC` is all-or-nothing flag, so we can't opt these out of it, and the "duplicate symbols" are valid as we depend on the same shared library from multiple other libraries). As RN depends on this behaviour, the alternative is to use `-force_load` for each library that needs to be loaded in this way and specify the path to the compiled (`.a`) library.

With 0.58 this compiles fine, but with 0.59rc2 it results in "duplicate symbol" errors for some Yoga references in `libReact.a`:

![screenshot 2019-02-25 at 18 10 14](https://user-images.githubusercontent.com/5458070/53359194-1ed20200-392a-11e9-9484-a5c4f5fee8c4.png)

Please let me know if you'd like more help looking into this issue or reproducing it - this would be a blocking issue for our project as we can't compile with `-ObjC`.

## To Reproduce

1. Remove `-ObjC` from  and add `-force_load libReact.a` to the "Linker Flags" section of Xcode, with the correct path to `libReact.a` (see below)

2. Clean and build, error will be displayed

I've created a repro case: https://github.com/tomduncalf/rn0.59-force-load-bug - in the interest of keeping the project simple, the path to `libReact.a` needs to be manually changed in "Linker Flags", you can discover it by going to `~/Library/Developer/Xcode/DerivedData` and looking for the directory starting `forceLoad59rc2-....` then replacing the existing one with this path (feels like there should be a `$(..)` variable pointing to this location but I've never found it!).

There is also a `rn@0.58` branch in that repo to show it compiles OK in the current release version - you may need to "Clean Build Folder" and/or clear the DerivedData folder to see the change when switching branches. 

## Expected Behavior
Project compiles with no errors as in 0.58

## Code Example
https://github.com/tomduncalf/rn0.59-force-load-bug

## Environment
```  React Native Environment Info:
    System:
      OS: macOS 10.14.2
      CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 126.19 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.7.0 - ~/.nvm/versions/node/v10.7.0/bin/node
[...]

PR #23823, opened 2019-03-08
Remove duplicated Yoga compile sources to prevent "duplicate symbols" errors when linking using -force_load

tomduncalf:
## Summary

This change fixes https://github.com/facebook/react-native/issues/23645.

The issue is that the `YGMarker.cpp`, `YGValue.cpp`, `YGConfig.cpp` and `log.cpp` files are already included in the Yoga compilation unit, so including these files in React's compile sources too results in "duplicate symbols" errors when loading React Native with `-force_load` (which behaves slightly differently to `-ObjC` - according to a colleague, "ObjC will scan through each object file in each library and force linking of any object file that contains Objective C code, while force_load will link every object file in a particular staticlib (regardless of whether or not the object file contains Objective C code)").

These changes seemed to be introduced by a few PRs:
- https://github.com/facebook/react-native/commit/43601f1a1735c59c90a73ba2dc127ec4dcfa4fa6
- https://github.com/facebook/react-native/commit/b5c66a3fbe869daf56b1b29d0a69efa2d83e30ce
- https://github.com/facebook/react-native/commit/e8f95dc7a1a630d9e90d3e044a0ccf6c9d76f651
- https://github.com/facebook/react-native/commit/95f625e151c7abbb05efcd14404c8aa9583faa83

Perhaps we need to check with the original authors/any C++ experts to confirm if this fix is correct - it compiles for me but I'm not sure what the original intention of these changes was.

## Changelog

[iOS] [Fixed] - Remove duplicated Yoga compile sources to prevent "duplicate symbols" errors when linking using -force_load

## Test Plan

react-native init project compiles and links with `-force_load` flag as described in https://github.com/facebook/react-native/issues/23645

### DIFF
--- a/React/React.xcodeproj/project.pbxproj
+++ b/React/React.xcodeproj/project.pbxproj
@@ -990,7 +990,6 @@
 		59EDBCC61FDF4E55003573DE /* (null) in Copy Headers */ = {isa = PBXBuildFile; };
 		59EDBCC71FDF4E55003573DE /* RCTScrollView.h in Copy Headers */ = {isa = PBXBuildFile; fileRef = 59EDBCA31FDF4E0C003573DE /* RCTScrollView.h */; };
 		59EDBCC81FDF4E55003573DE /* RCTScrollViewManager.h in Copy Headers */ = {isa = PBXBuildFile; fileRef = 59EDBCA51FDF4E0C003573DE /* RCTScrollViewManager.h */; };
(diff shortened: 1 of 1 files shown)


kelset:
cc @SidharthGuglani & @davidaurelio too

hramos:
Thanks for the fix. I'll import this and ask for a review from the original authors.

Issue #23816, opened 2019-03-08
[Android] Image.getSize doesn't work with images from hybrid app's resources


hwth:
## 🐛 Bug Report
  I use Image.getSize to get width, height of images from hybrid app's resources (Drawable folder on Android and Images.xcassets on Xcode). It works well with iOS but I got failed with Android.

Can anyone fix it or suggest a way to get size of images from native.
Thanks
## Code Example
` Image.getSize('background', (width, height) => {
      // do something
});`

## Environment
 React Native Environment Info:

    System:
      OS: macOS 10.14.3
      CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 197.98 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh

    Binaries:
      Node: 11.5.0 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman

    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.2, 28.0.3

    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild

    npmPackages:
      react: 16.6.3 => 16.6.3
      react-native: 0.58.5 => 0.58.5

    npmGlobalPackages:
      react-native-cli: 2.0.1

hwth:
@react-native-bot done!

maxiee:
I was able to reproduce this issue with code:

```jsx
import React, {Component} from 'react';
import { View, Image } from 'react-native';

export default () => {
    console.log("maxieeee :" + JSON.stringify(Image.getSize('react_native')))

    return <View>
        <Image 
            source={{uri: 'react_native'}} 
            style={{
                width: 200, 
                height: 150,
                borderWidth: 1,
                borderColor: '#999999'
            }}/>
    </View>
}
```

The screenshot is:

[...]

PR #23832, opened 2019-03-09
[Android] Fix Image.getSize throws exception. issue #23816

maxiee:
## Summary

In the issue https://github.com/facebook/react-native/issues/23816, reporting Image.getSize doesn't work on the Android. I was able to reproduce this issue, too.

By having a deep look at the source, I finally figured it out.The bug is in the `com.facebook.react.modules.image.ImageLoaderModule#getSize` function.

As in the  issue https://github.com/facebook/react-native/issues/23816, I use this code to reproduce the bug:

```jsx
import { View, Image } from 'react-native';

export default () => {
    console.log("maxieeee :" + JSON.stringify(Image.getSize('react_native')))

    return <View></View>
}
```

In the code abrove, `'react_native'` is a image file in the  Android drawable folder (Hybrid App's Resource).

In the function `com.facebook.react.modules.image.ImageLoaderModule#getSize`:

```java
public void getSize(
    final String uriString,
    final Promise promise) {
```

The value of `uriString` is `'react_native'`. This value will pass into:

```
Uri uri = Uri.parse(uriString);
```

But Uri class can't handle `'react_native'`. It expects uri like `"android.resource://org.xyz.abc/drawable/react_native"`.

So in my fix commit, I use ImageSource class to parse uriString. It will handle the uri properly, both remote and local.

But one thing worries me: 

- The ImageSource class is in the package of `com.facebook.react.views.imagehelper`. 
- The ImageLoaderModule class is in the package of `com.facebook.react.modules.image`
- Can I import ImageSource from ImageLoaderModule?

## Changelog

[Android] [Fixed] - Fix Image.getSize throws exception. issue #23816

## Test Plan

Test code:

```jsx
import { View, Image } from 'react-native';

export default () => {
    console.log("maxieeee :" + JSON.stringify(Image.getSize('react_native')))

    return <View></View>
}
```

In the code abrove, `'react_native'` is a image file in the  Android drawable folder (Hybrid App's Resource).

If the bug has been fixed:

- The console will log the correct image size
- The "Failed to get size for image: " warning on the screen won't appear. (There's a screenshot of the warning  https://user-images.githubusercontent.com/3392007/54027135-574fc680-41db-11e9-8099-4a2f8e677898.png)

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java
@@ -32,6 +32,7 @@ import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.module.annotations.ReactModule;
+import com.facebook.react.views.imagehelper.ImageSource;
 
 @ReactModule(name = ImageLoaderModule.NAME)
 public class ImageLoaderModule extends ReactContextBaseJavaModule implements
@@ -77,8 +78,8 @@ public class ImageLoaderModule extends ReactContextBaseJavaModule implements
       return;
     }
 
-    Uri uri = Uri.parse(uriString);
-    ImageRequest request = ImageRequestBuilder.newBuilderWithSource(uri).build();
+    ImageSource imageSource = new ImageSource(getReactApplicationContext(), uriString);
+    ImageRequest request = ImageRequestBuilder.newBuilderWithSource(imageSource.getUri()).build();
 
     DataSource<CloseableReference<CloseableImage>> dataSource =
       Fresco.getImagePipeline().fetchDecodedImage(request, mCallerContext);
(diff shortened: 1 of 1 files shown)


maxiee:
I find ci/circleci: test_android failed with the error as following:

Build Android App:

```
BUILDING: FINISHED IN 3.6s (100%) 211/260 JOBS, 71 UPDATED
BUILD FAILED
Command failed with exit code 1.
stderr: /root/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/image/ImageLoaderModule.java:35: error: package com.facebook.react.views.imagehelper does not exist
import com.facebook.react.views.imagehelper.ImageSource;
                                           ^
Errors: 1. Warnings: 0.
```

So the question is: To fix this bug, I will use the ability of ImageSource. But ImageSource is in a different package. I can't import ImageSource from ImageLoaderModule.

I don't know how to solve this problem, anyone help?

hramos:
Add the dependency to the Image Buck file at https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/image/BUCK:

```
react_native_target("java/com/facebook/react/views/imagehelper:imagehelper"),
```

cpojer:
See #23961.

PR #23837, opened 2019-03-10
[iOS] Fix RCTJavaScriptLoader loadBundleAtURL error case

hcn1519:
## Summary

Change the function called when an error is returned from the completionHandler of the `RCTJavaScriptLoader(loadBundleAtURL:onProgress:onComplete:)`.

* AS-IS
    - `RCTLogError()`
    - This error method does not post any notification for users. So users cannot receive notification when `RCTJavaScriptLoader(loadBundleAtURL:onProgress:onComplete:)` complete with error.(i.e. bundleURL is is invalid but not nil, `http://1`, `http://localho:8081/index.bundle`)

* TO-BE
    - `handleError(error:)`
    -  Call this method will post notification for users when `completionHandler` fails with error.

## Changelog

[iOS] [Fixed] - Change the function called when an error is returned from the completionHandler of the `RCTJavaScriptLoader(loadBundleAtURL:onProgress:onComplete:)`.

## Test Plan

* Test Code - Set `sourceURL` of `RCTBridge` is invalid, but not nil.

```swift
import UIKit
import React

class ViewController: UIViewController {

    var bridge: RCTBridge?

    override func viewDidLoad() {
        super.viewDidLoad()
        NotificationCenter.default.addObserver(self, selector: #selector(bridgeDidFailToLoad),
                                               name: Notification.Name.RCTJavaScriptDidFailToLoad, object: nil)
        self.bridge = RCTBridge(delegate: self, launchOptions: nil)
    }

    @objc func bridgeDidFailToLoad(_ notification: Notification) {
        print("[RCT] RCTJavaScriptDidFailToLoad", notification)
    }
}

extension ViewController: RCTBridgeDelegate {
    func sourceURL(for bridge: RCTBridge!) -> URL! {
        // will not become nil, but invalid
        let url = URL(string: "http://localho:8081/index.bundle") 
        return url
    }
}
```

* `bridgeDidFailToLoad(_:)`will be called with notification like this.

```
[RCT] RCTJavaScriptDidFailToLoad name = RCTJavaScriptDidFailToLoadNotification, 
object = Optional(<RCTBridge: 0x600001370a10>), 
userInfo = Optional([AnyHashable("error"): Error Domain=RCTJavaScriptLoaderErrorDomain Code=3 "Could not connect to development server.
```

Previous(`RCTLogError()`) | Updated(`handleError(error:)`)
:---: | :---: | 
 |

### DIFF
--- a/React/CxxBridge/RCTCxxBridge.mm
+++ b/React/CxxBridge/RCTCxxBridge.mm
@@ -409,9 +409,10 @@ struct RCTInstanceCallback : public InstanceCallback {
                                          "server or have included a .jsbundle file in your application bundle.");
     onSourceLoad(error, nil);
   } else {
+    __weak RCTCxxBridge *weakSelf = self;
     [RCTJavaScriptLoader loadBundleAtURL:self.bundleURL onProgress:onProgress onComplete:^(NSError *error, RCTSource *source) {
       if (error) {
-        RCTLogError(@"Failed to load bundle(%@) with error:(%@ %@)", self.bundleURL, error.localizedDescription, error.localizedFailureReason);
+        [weakSelf handleError:error];
         return;
       }
       onSourceLoad(error, source);


hcn1519:
I tried to figure out AppVeyor build failed case. However, running same code in same environment has been succeeded from my origin branch.

https://ci.appveyor.com/project/hcn1519/react-native

shergin:
Thank you @cpojer for being so BOLD!!1 ❤️
Everyone believes that this change is for good but seems only Christoph is brave enough to land this. 😂

PR #23851, opened 2019-03-11
Fix broken perf marker logging for RN > 0.58 on Android

### LINKED ISSUE #23771: Performance markers logged from C++ no longer work in Android
## 🐛 Bug Report

From 0.58.0 onwards, the four most important perf markers used for measuring RN startup time on Android no longer work: 

`NATIVE_MODULE_SETUP_START`
`NATIVE_MODULE_SETUP_END`
`RUN_JS_BUNDLE_START`
`RUN_JS_BUNDLE_STOP`

This is the commit where the refactor from `JSCExecutor` to `JSIExecutor` took place: https://github.com/facebook/react-native/commit/749b18dbc944945532fc11dea41821010dcff5e4

---

On further investigation, the reason behind this is the [`hasLogger` boolean here](https://github.com/facebook/react-native/blob/50d095ce322a3355dca041aabab7d40692ea2f7f/ReactCommon/jsiexecutor/jsireact/JSIExecutor.cpp#L149) never gets a truthy value. This is because the address `&ReactMarker::logTaggedMarker` [here](https://github.com/facebook/react-native/blob/50d095ce322a3355dca041aabab7d40692ea2f7f/ReactCommon/jsiexecutor/jsireact/JSIExecutor.cpp#L150) is different from the one [here](https://github.com/facebook/react-native/blob/50d095ce322a3355dca041aabab7d40692ea2f7f/ReactAndroid/src/main/jni/react/jni/JReactMarker.cpp#L18). 

If we instead use `JReactMarker::logMarker` [here](https://github.com/facebook/react-native/blob/50d095ce322a3355dca041aabab7d40692ea2f7f/ReactCommon/jsiexecutor/jsireact/JSIExecutor.cpp#L150), then the metrics work fine.

cc-ing: @mhorowitz and @axe-fb who seem most involved with these changes. @leanmazzu for visibility.

## To Reproduce

1. Init a new RN project on any version above 0.58
2. Add the following code to start logging markers:

    ```java
    ReactMarker.addListener(new ReactMarker.MarkerListener() {
      @Override
      public void logMarker(
              ReactMarkerConstants name, @Nullable String tag, int instanceKey) {
        Log.d("ReactNativeEvent", "name: "+name.name() + " tag: "+tag);
      }
    });
    ```
3. Check logcat. The above mentioned 4 markers won't be logged.

## Expected Behavior

It should log the relevant markers.

## Code Example

N/A

## Environment

```
    System:
      OS: macOS 10.14.1
      CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
      Memory: 26.55 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
      npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.0, 25.0.2, 26.0.2, 26.0.3, 27.0.3, 28.0.3
        System Images: android-26 | Google APIs Intel x86 Atom_64, android-26 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5264788
    npmPackages:
      react: 16.8.3 => 16.8.3
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

karanjthakkar:
## Summary

Fixes #23771 

## Changelog

[Android] [Fixed] - Fix broken perf marker logging for RN > 0.58 on Android

## Test Plan

The relevant markers are now logged whenever you initialise RN bridge.

### DIFF
--- a/React/CxxBridge/JSCExecutorFactory.h
+++ b/React/CxxBridge/JSCExecutorFactory.h
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <jsireact/JSIExecutor.h>
+#include <cxxreact/ReactMarker.h>
 
 namespace facebook {
 namespace react {
@@ -15,8 +16,10 @@ namespace react {
 class JSCExecutorFactory : public JSExecutorFactory {
 public:
   explicit JSCExecutorFactory(
-      JSIExecutor::RuntimeInstaller runtimeInstaller)
-      : runtimeInstaller_(std::move(runtimeInstaller)) {}
+      JSIExecutor::RuntimeInstaller runtimeInstaller,
+      ReactMarker::LogTaggedMarker logTaggedMarker)
+      : runtimeInstaller_(std::move(runtimeInstaller)),
+      logTaggedMarker_(logTaggedMarker) {}
 
   std::unique_ptr<JSExecutor> createJSExecutor(
       std::shared_ptr<ExecutorDelegate> delegate,
@@ -24,6 +27,7 @@ public:
 
 private:
   JSIExecutor::RuntimeInstaller runtimeInstaller_;
+  ReactMarker::LogTaggedMarker logTaggedMarker_;
 };
 
 } // namespace react
--- a/React/CxxBridge/JSCExecutorFactory.h
+++ b/React/CxxBridge/JSCExecutorFactory.h
@@ -16,6 +16,10 @@ namespace react {
 class JSCExecutorFactory : public JSExecutorFactory {
 public:
   explicit JSCExecutorFactory(
+      JSIExecutor::RuntimeInstaller runtimeInstaller)
+      : runtimeInstaller_(std::move(runtimeInstaller)),
+      logTaggedMarker_(nullptr) {}
+  explicit JSCExecutorFactory(
       JSIExecutor::RuntimeInstaller runtimeInstaller,
       ReactMarker::LogTaggedMarker logTaggedMarker)
       : runtimeInstaller_(std::move(runtimeInstaller)),
--- a/React/CxxBridge/RCTCxxBridge.mm
+++ b/React/CxxBridge/RCTCxxBridge.mm
@@ -112,9 +112,9 @@ static void notifyAboutModuleSetup(RCTPerformanceLogger *performanceLogger, cons
   }
 }
 
-static void registerPerformanceLoggerHooks(RCTPerformanceLogger *performanceLogger) {
+static ReactMarker::LogTaggedMarker getPerformanceLoggerHooks(RCTPerformanceLogger *performanceLogger) {
   __weak RCTPerformanceLogger *weakPerformanceLogger = performanceLogger;
-  ReactMarker::logTaggedMarker = [weakPerformanceLogger](const ReactMarker::ReactMarkerId markerId, const char *__unused tag) {
+  return [weakPerformanceLogger](const ReactMarker::ReactMarkerId markerId, const char *__unused tag) {
     switch (markerId) {
       case ReactMarker::RUN_JS_BUNDLE_START:
         [weakPerformanceLogger markStartForTag:RCTPLScriptExecution];
@@ -230,8 +230,6 @@ struct RCTInstanceCallback : public InstanceCallback {
                         launchOptions:bridge.launchOptions])) {
     _parentBridge = bridge;
     _performanceLogger = [bridge performanceLogger];
-
-    registerPerformanceLoggerHooks(_performanceLogger);
 
     RCTLogInfo(@"Initializing %@ (parent: %@, executor: %@)", self, bridge, [self executorClass]);
 
@@ -349,7 +347,8 @@ struct RCTInstanceCallback : public InstanceCallback {
       executorFactory = [cxxDelegate jsExecutorFactoryForBridge:self];
     }
     if (!executorFactory) {
-      executorFactory = std::make_shared<JSCExecutorFactory>(nullptr);
+      executorFactory = std::make_shared<JSCExecutorFactory>(nullptr,
+        getPerformanceLoggerHooks(_performanceLogger));
     }
   } else {
     id<RCTJavaScriptExecutor> objcExecutor = [self moduleForClass:self.executorClass];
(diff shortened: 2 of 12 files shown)


karanjthakkar:
The intention of this PR is not to be a final solution but to allow someone from FB to import this internally to see if it causes any regressions.

karanjthakkar:
@axe-fb Do you have any updates from the result of importing this internally? There was a merge conflict that occurred from a recent change to master that I have now fixed as well.

axe-fb:
@karanjthakkar - I imported it internally, and looks like there are a ton of apps internally that use the old signature to initialize the JSC factory, causing compilation errors. Need to see if there is a way to change them all, or not add a new param to the constructor. 

A side question - can this instead be a C++ module, instead of passing it in the constructor ?

karanjthakkar:
> A side question - can this instead be a C++ module, instead of passing it in the constructor ?

@axe-fb I don't know what you mean by this. I'm not super well versed with the C++ architecture of RN. Do you have an example?

> Need to see if there is a way to change them all, or not add a new param to the constructor.

The constructor technique was something Ben suggested in the original issue. I did not know internally folks create different JSC Factory. What is the usecase for that? Would overloading the constructor to allow zero arguments help?

---

Essentially I just want the metrics to work like they did before. There are already plans in place for v0.60 whereas this has been broken since v0.58. I'm happy to do whatever changes you think are appropriate to get this resolved. As you can see this has been sitting for quite some time and at some point it would get too difficult for me to resolve the conflicts that come from continuous updates to master. Its super unfortunate that this bug wasn't caught because [they depend on different build systems for OSS and FB](https://github.com/facebook/react-native/issues/23771#issuecomment-470265326).

ReactAndroid/src/main/java/com/facebook/react/jscexecutor/OnLoad.cpp
```diff
@@ -19,6 +20,7 @@ namespace {
 
 class JSCExecutorFactory : public JSExecutorFactory {
  public:
+  JSCExecutorFactory(ReactMarker::LogTaggedMarker logTaggedMarker) : logTaggedMarker_(logTaggedMarker) {}
```
karanjthakkar:
```suggestion
  JSCExecutorFactory() { logTaggedMarker_ = nullptr; }
  JSCExecutorFactory(ReactMarker::LogTaggedMarker logTaggedMarker) : logTaggedMarker_(logTaggedMarker) {}
```

@axe-fb Maybe something like this? Assuming the other instances of JSC factory that you are talking about don't need marker logging.

karanjthakkar:
FYI, `test_android` job is failing on master as well. So I assume it's unrelated to this PR.

karanjthakkar:
@axe-fb @mhorowitz Do you folks have any update here? Anything I can do to get this to resolution faster?

ReactAndroid/src/main/java/com/facebook/react/jscexecutor/OnLoad.cpp
axe-fb: Yeah. I was thinking we could have a second constructor, which would be hte original signature and hence existing apps will not break.

React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -112,9 +112,9 @@ static void notifyAboutModuleSetup(RCTPerformanceLogger *performanceLogger, cons
   }
 }
 
-static void registerPerformanceLoggerHooks(RCTPerformanceLogger *performanceLogger) {
+static ReactMarker::LogTaggedMarker registerPerformanceLoggerHooks(RCTPerformanceLogger *performanceLogger) {
```
axe-fb: Comment from @mhorowitz 

This is also a strange name, since it's not "registering" anymore.

ReactAndroid/src/main/java/com/facebook/react/jscexecutor/OnLoad.cpp
karanjthakkar: @axe-fb I can make that change but I wonder if that would be enough. This change would not break those features but again it will swallow any marker logs from the features because the logger won't be available. Do you think that would create any problems?

React/CxxBridge/RCTCxxBridge.mm
karanjthakkar: I've made it `getPerformanceLoggerHooks` now. Let me know if that's okay.

React/CxxBridge/JSCExecutorFactory.h
```diff
@@ -15,6 +15,10 @@ namespace react {
 
 class JSCExecutorFactory : public JSExecutorFactory {
 public:
+  explicit JSCExecutorFactory(
```
karanjthakkar: This change is needed to make the [RNTester turbo module](https://github.com/facebook/react-native/blob/421ffb05aeb7ccf7013425bdac6b14372325bb26/RNTester/RNTester/AppDelegate.mm#L137) example work.

karanjthakkar:
@axe-fb Do you have any updates here? Did the internal tests pass?

axemclion:
Sorry, did not mean change the constructor of JSCExecutor. Meant change JSIExecutor. Never mind though. I have made the changes, and am now trying to get it building internally.

axe-fb:
So internal tests succeed, but OSS tests are failing !! 


```
xplat/js/react-native-github/RNTester/RNTester/AppDelegate.mm:139:15: note: in instantiation of function template specialization 'std::__1::make_unique<facebook::react::JSCExecutorFactory, (lambda at xplat/js/react-native-github/RNTester/RNTester/AppDelegate.mm:139:64)>' requested here
  return std::make_unique<facebook::react::JSCExecutorFactory>([weakSelf, bridge](facebook::jsi::Runtime &runtime) {
       
```


now working on fixing those !! 

This is indeed a game of whack-a-mole !! But atleast, we have good tests to catch these !!

karanjthakkar:
Yay! That's great! If you could push the commit here that you did to fix internal tests, I can try to look into why the oss tests are failing.

axe-fb:
Update: I fixed all internal and OSS tests, so I think we should be good. I am now waiting for internal code review now, since the commit now has grown to touch a lot more files now.

karanjthakkar:
@axe-fb Thanks for the update! Look forward to seeing this resolved!

karanjthakkar:
@axe-fb Any update here? 🙂

PR #23857, opened 2019-03-12
[iOS] Enhance search of directories when load local asset image

zhongwuzw:
## Summary

Currently, `RCTLocalAssetImageLoader` only support directory of `Library` and `App bundle`, actually, we need to support other directories like `tmp` or `Documents`. Otherwise, the local image load in `tmp` or `Documents` would be handled by `NSURLSession`, we don't need that.

image

## Changelog

[iOS] [Fixed] - Enhance search of directories when load local asset image

## Test Plan

For any image url which located in `Documents` or `tmp`, we can load it by `RCTLocalAssetImageLoader`.


zhongwuzw:
@shergin I removed unused methods, I think we can add it back if we need it later.

PR #23859, opened 2019-03-12
[iOS] Introduce Module Setup Metric

RCiesielczuk:
## Summary

The `RCTBridge` contains numerous definitions of notification names, which we can observe in order to get insights into the React Native performance.

The Android implementation is a little different, such that you can listen for any of the [following](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactMarkerConstants.java) marker constants, simply by including the following code:

```java
ReactMarker.addListener(new ReactMarker.MarkerListener() {
  @Override
  public void logMarker(ReactMarkerConstants name, @Nullable String tag, int instanceKey) {
    Log.d("ReactNativeEvent", "name: "+ name.name() + " tag: " + tag);
  }
});
```
This will allow you to perform the necessary processing, calculations as required.

---

This PR enables observing for the module setup event (`RCTDidSetupModuleNotification`) by including the respective module's name & setup time in milliseconds.

## Changelog

[iOS] [Added] - Gain insights on the module setup times by observing `RCTDidSetupModuleNotification`. The `userInfo` dictionary will contain the module name and setup time in milliseconds. These values can be extracted via `RCTDidSetupModuleNotificationModuleNameKey ` and `RCTDidSetupModuleNotificationSetupTimeKey`.

## Test Plan

Observe for `RCTDidSetupModuleNotification`

```objc
- (void)setup {
    [_notificationCenter addObserver:self
                            selector:@selector(didSetupModule:)
                                name:RCTDidSetupModuleNotification
                              object:nil];
}

- (void)didSetupModule:(NSNotification *)notification {
    NSDictionary *userInfo = notification.userInfo;
    if (userInfo) {
        NSString *moduleName = userInfo[RCTDidSetupModuleNotificationModuleNameKey];
        NSNumber *setupTime = userInfo[RCTDidSetupModuleNotificationSetupTimeKey];
        NSLog(@"Module %@ setup took %li ms", moduleName, [setupTime integerValue]);
    }
}
```

### Output

```
Module UIManager setup took 11 ms
Module Timing setup took 0 ms
Module DeviceInfo setup took 1 ms
Module Networking setup took 0 ms
Module BlobModule setup took 0 ms
Module I18nManager setup took 1 ms
Module NativeAnimatedModule setup took 1 ms
Module ImageViewManager setup took 0 ms
Module KeyboardObserver setup took 1 ms
Module ScrollViewManager setup took 1 ms
Module RNGestureHandlerModule setup took 0 ms
Module LinkingManager setup took 0 ms
Module WebViewManager setup took 0 ms
Module WKWebViewManager setup took 1 ms
```


React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -120,7 +133,12 @@ static void registerPerformanceLoggerHooks(RCTPerformanceLogger *performanceLogg
       case ReactMarker::JS_BUNDLE_STRING_CONVERT_START:
       case ReactMarker::JS_BUNDLE_STRING_CONVERT_STOP:
       case ReactMarker::NATIVE_MODULE_SETUP_START:
```
PeteTheHeat: You'll need to move around these cases.

`CREATE_REACT_CONTEXT_STOP`, `JS_BUNDLE_STRING_CONVERT_START` ... are still not used, they should not be executing `[weakPerformanceLogger markStartForTag:RCTPLNativeModuleSetup];`

RCiesielczuk: Ahhh yeah, nice catch!

I'll sort it out.

RCiesielczuk:
> Remind me, where are these timings coming from?

@PeteTheHeat,

Each module is setup during the following step:

https://github.com/facebook/react-native/blob/d752446b23e0a12171ca0c01d20f5be625f40455/ReactCommon/jsiexecutor/jsireact/JSINativeModules.cpp#L37

The `createModule` function contains the logging mechanism for `NATIVE_MODULE_SETUP_START` at the start and `NATIVE_MODULE_SETUP_STOP` before the module is returned. 

Markers for reference:

https://github.com/facebook/react-native/blob/d752446b23e0a12171ca0c01d20f5be625f40455/ReactCommon/jsiexecutor/jsireact/JSINativeModules.cpp#L60

https://github.com/facebook/react-native/blob/d752446b23e0a12171ca0c01d20f5be625f40455/ReactCommon/jsiexecutor/jsireact/JSINativeModules.cpp#L84

RCiesielczuk:
Hi @PeteTheHeat, were you able to review my changes since I last pushed them? 🙂

PeteTheHeat:
Looks good to me, merging 👍🏻

Issue #9566, opened 2016-08-24
[DatePickerIOS] minuteInterval not work with mode="time" minimumDate={min} maximumDate={max}


charlesLin:
render() {
  let {year, month, day} = new Date();
    let min = new Date(year, month, day, 8);
    let max = new Date(year, month, day, 22);
return (
<DatePickerIOS
              date={this.state.date}
              mode="time"
              minuteInterval={10}
              minimumDate={min} maximumDate={max}
              onDateChange={(date) => this.setState({date})}
              />);

I found that minuteInterval works on iOS emulator, but not work at real device both minuteInterval and date range feature.

ilove8:
I have a same problem, but not work also my simulator. I'm use version 0.31

olliebrennan:
Ditto. I am running 0.33.0 and this does not work. I am not using min/max date either. Keeping it simple.

ferndopolis:
Confirming same issue. 
Running Xcode 8.0, 
React Native UIExplorer (master branch), with 10 minute interval 
<img width="373" alt="screen shot 2016-09-30 at 9 22 01 am" src="https://cloud.githubusercontent.com/assets/764493/18999009/1c10c04a-86f0-11e6-9368-70a90a4e1b70.png">

CalabashSmokeTest Native UIDatePicker with 30 minute interval
https://github.com/calabash/ios-smoke-test-app
<img width="375" alt="screen shot 2016-09-30 at 9 19 36 am 1" src="https://cloud.githubusercontent.com/assets/764493/18999010/1fc62bf8-86f0-11e6-8f3d-53aa68389859.png">

dsibiski:
@nicklockwood Do you think that your latest commit to the DatePicker could be preventing the `minuteInterval` prop from being set somehow? I wonder if there is something else going on here...

https://github.com/facebook/react-native/commit/c1aff6b116dd8ca42dd81ab59a542a3a6d18e46e#diff-bb7f9a35dd5ae2a59fcff0292caf41f3

dsibiski:
See my comment here for a workaround: https://github.com/facebook/react-native/commit/002024cc456246761421f6e5c354466baa819843#commitcomment-19473525

ferndopolis:
@dsibiski yes, confirming that adding back ```width: 320``` to datePickerIOS style does fix this issue.

fr1n63:
Still an issue at RN 0.38 - minuteInterval isn't passed through in any combination of props.

[...]

PR #23861, opened 2019-03-12
Fix DatePickerIOS e2e tests

grabbou:
## Summary

DatePickerIOS tests stopped working after US changed their time zone on CircleCI. I decided to update the tests to be more reliable, by:
- removing "timezoneoffset" from the examples - I don't think it's needed. We don't demonstrate other props and it was causing us some troubles when timezones actually changed
- changing "interval" example from "time-only" to "datetime" because there's a known bug that prevented the previous example from working https://github.com/facebook/react-native/issues/9566
- splitting the label to two: one for date and other one for time, so that we can match the date only when we test "date" mode only

## Changelog

[IOS] [FIXED] - Improved reliability of DatePickerIOS e2e tests

## Test Plan

Tests pass on CircleCI (see 0.59-stable)


grabbou:
Detox tests are green, so this is ready to ship!

ericlewis:
Woot!

PR #23865, opened 2019-03-12
add support for native/downloadable fonts

dulmandakh:
## Summary

Android API 26 and Android Support Library 26 added support for font resource type and native/downloadable fonts. It allows apps to easily download fonts from online providers, but also use of various font weights other than normal and bold, like medium. So it deprecated APIs for asset fonts, and should be removed in the future.

Advantages:
- Just copy font files in res/font and use it specifying filename (without extension) in fontFamily
- Define custom font-family using XML file (in res/font) and font files, it may have many weights and styles. See PR for example.
- Define configuration to download fonts from online font providers, and use it.

See https://developer.android.com/guide/topics/ui/look-and-feel/fonts-in-xml and https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts

## Changelog

[Android] [Changed] - add support for custom/downloadable fonts

## Test Plan

CI is green and RNTester app's Text examples include Srisakdi font example.


dulmandakh:
@hey99xx please review

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +75,20 @@ public static ReactFontManager getInstance() {
     return typeface;
   }
 
+  public @Nullable Typeface getTypeface(
+    String fontFamilyName,
+    int style,
+    Context context) {
+    int fontId = context.getResources().getIdentifier(fontFamilyName, "font", context.getPackageName());
```
hey99xx: I don't know if the first priority should go to `mFontCache` (which has advantage of it can be changed programmatically) or to the static assets.

Also for static assets, I'd guess both `getIdentifier` and `Typeface.create` call can be slow, that's why there are caches in the first place. If you have 100s of textviews, you probably don't want to redo these computations that never change.

ResourcesCompat is documented as a blocking call. I don't know if that means it blocks for network or something else, but keep that in mind too. I don't have too much knowledge about the new font system, this is just looking up to javadoc, but I don't know if this would then work for downloadable fonts.

hey99xx:
Overall I really like the idea though. In our app we have fonts with 3+ different weights and have to hardcode asset names as fontFamily style. This brings us much closer to iOS that does not have this problem.

ReactAndroid/src/main/java/com/facebook/react/views/text/CustomStyleSpan.java
```diff
@@ -117,4 +142,43 @@ private static void apply(
     paint.setSubpixelText(true);
   }
 
+  private static void apply(
+    Paint paint,
+    int style,
+    int weight,
+    @Nullable String family,
+    Context context) {
```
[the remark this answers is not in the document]
hey99xx: Ah nevermind. Then you would call old `ReactFontManager.getTypeface` method again. Still, there might be a prettier way to refactor this. For example you can have a single method that takes both args:

```java
private static void apply(
    Paint paint,
    int style,
    int weight,
    @Nullable String family,
    Context context, 
    AssetManager assets) {

    // ...

    if (family != null) {
        ReactFontManager fontManager = ReactFontManager.getInstance();
        typeface = context != null 
              ? fontManager.getTypeface(family, want, context) 
              : fontManager.getTypeface(family, want, assets);
    } else if (typeface != null) {
```

dulmandakh:
@hey99xx yep, feature parity with iOS is the motivation for this PR. We have medium weights in our UI and I have to check platform and specify fontFamily on Android, but fontWeight on iOS. I have a concern that current API might be used somewhere in FB and changing it might break their apps, therefore created almost identical APIs.

dulmandakh:
@hey99xx thank you for the comments. I hope that I'll get someone from FB to review and maybe remove deprecated APIs if they're not used at FB.

hey99xx:
> remove deprecated APIs if they're not used at FB. 

That's why I didn't comment on `public CustomStyleSpan`, I learnt from previous PRs that FB is cautious on deprecating first then deleting. But for `private static void apply`, it makes less sense to keep it unless you think FB uses their own private methods by reflection.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
dulmandakh: Maybe adding cache would be a good idea, like you said it'll improve performance for 100s texts.

fkgozali:
Heads up: we're currently analyzing this commit and we suspect it's causing some start up regression for some RN surfaces at Facebook. I'll comment on the code that might be causing it, but in the short term we'd very likely need to revert this commit. I'll provide suggested "fix", but it's actually not that simple.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +76,33 @@ public static ReactFontManager getInstance() {
...
+    Typeface font = mTypeCache.get(fontFamilyName);
+
+    if (font != null) {
+      return Typeface.create(
+        font,
+        style
+      );
+    }
+
+    int fontId = context.getResources().getIdentifier(fontFamilyName, "font", context.getPackageName());
```
fkgozali: So `getIdentifier()` is problematic in a big app that has many resources. This could contribute to non-trivial msec, just trying to crack open the resources, then looking it up from the APK. Rather than calling this for all scenarios, we'd instead build a "provider" pattern where:
* each app decides if it wants to provide a custom fonts
* if no provider is given, avoid asking `getIdentifier()` completely
* if given, ask the provider for a resourceId (int), then use it

Now the implementation of the provider needs to be a switch statement for the best performance. Example:

```
int getCustomFontProvider(name) {
  switch(name) {
    case "mine":
       return R.font.mine;
    ...
```

Referencing `R.font.*` directly is fast and efficient, because it points exactly to an int resource Id. The only problem is, that each app needs to implement this code themselves, because each app has different set of custom fonts.

What do you think?

fkgozali: cc @cpojer @mdvacca

dulmandakh:
How about checking custom fonts the old way first, then the new way later, so that it won't affect FB. It's just a workaround for FB. I think that custom font feature parity with iOS is very important for general public.

fkgozali:
What do you mean by the “old way”? Regardless of whether parity is important, we’re trying to mitigate regression as the current priority

dulmandakh:
When someone tried to use a custom font it'll check
1. R.font.* or font resources. New way
2. Scan in assets. Old way.

I'm proposing to reverse it, so FB will find fonts in assets and won't cause regression.

fkgozali:
oh but I don’t think fb uses custom fonts (for RN) today, so we don’t want to look up any font

fkgozali:
to satisfy all use cases, the provider model should work (fb wont have any provider, so fb isn’t affected, and RNTester will provide a provider, so custom fonts will work)

dulmandakh:
Please feel free to revert the commit. I'll look for a generic solution and add you as reviewer.

dulmandakh:
@fkgozali I would like to have a TypedArray resource where you list custom fonts to make them usable in React Native, so it won't lookup for font resources and won't affect performance. iOS does the same, you have to list custom fonts in Info.plist.

fkgozali:
FYI the revert is https://github.com/facebook/react-native/commit/eb40b09bfd0e617912663fb056d0bdea85ce0c2f

betoharres:
Since this PR was reverted, is there any known issue/PR open about this bug? I'm currently having this problem and I'm not sure if I'm doing something wrong or if this bug still occurs.

```
info
    System:
      OS: Linux 5.1 Arch Linux undefined
      CPU: (6) x64 Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
      Memory: 3.17 GB / 15.59 GB
      Shell: 5.7.1 - /usr/bin/zsh
    Binaries:
      Node: 11.15.0 - /usr/bin/node
      Yarn: 1.16.0 - /usr/bin/yarn
      npm: 6.10.0 - /usr/bin/npm
    SDKs:
      Android SDK:
        API Levels: 26, 28, 29
        Build Tools: 26.0.3, 28.0.3, 29.0.0
        System Images: android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.8 => 0.59.8
    npmGlobalPackages:
      react-native-cli: 2.0.1
```
Thanks you =)

ScreamZ:
Any news on that thread ? Where to follow since revert

chrisbobbe:
> Any news on that thread ? Where to follow since revert

It landed with https://github.com/facebook/react-native/commit/fd6386a07eb75a8ec16b1384a3e5827dea520b64 on https://github.com/facebook/react-native/pull/24595! 🙂

PR #23869, opened 2019-03-12
Feature/gradle font sync

punksta:
## Summary

https://github.com/react-native-community/discussions-and-proposals/issues/89

## Changelog

[Android] [Changed] - added fonts linking in build time. 

## Test Plan

1) add some fonts in `./assets/fonts`
2) build android apk `gradlew assembleDebug`
3) find fonts in apk(you can unzip it or use android studio apk previewer)
image

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -14,7 +14,7 @@ def bundleCommand = config.bundleCommand ?: "bundle"
 def reactRoot = file(config.root ?: "../../")
 def inputExcludes = config.inputExcludes ?: ["android/**", "ios/**"]
 def bundleConfig = config.bundleConfig ? "${reactRoot}/${config.bundleConfig}" : null ;
-def fontsDir = file(config.fontsDir ?: "${reactRoot}/js/assets/fonts")
+def fontsDir = file(config.fontsDir ?: "${reactRoot}/assets/fonts")
 
 
 afterEvaluate {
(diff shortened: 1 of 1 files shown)


dulmandakh:
here is my PR to land downloadable fonts support https://github.com/facebook/react-native/pull/23865. I think that RN will deprecate assets fonts support in the future.

punksta:
@dulmandakh I think it's good alternative, but not replacement.

dulmandakh:
@punksta why not replace asset fonts?

punksta:
@dulmandakh 
About [downloadable fonts](https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts): I think we will always need ability to bundle fonts in app. First of all it's easy and doesn't require internet connection and handling "certs". So this can't replace fonts bundling.

About [fonts-in-xml ](https://developer.android.com/guide/topics/ui/look-and-feel/fonts-in-xml) I just don't see enough reasons of using that in react-native now. RN is not using native xml layouts so it's so important from where to load font file (asset or resource). 
The only reasons I can find are: 
- android resource system (for example different fonts in different system languages or screen sizes)
- supporting recent android standards.

maybe if you write motivation of pull request it would be more clear.

dulmandakh:
You can use fonts from res/font folder or custom font family defined in xml resource. Even you can ask to download files from the Internet to reduce APK size. But also you can use other font weights than normal and bold, like medium.

I should've added more summary to the PR, and will update it soon.

cpojer:
Closing because of @dulmandakh's PR.

Issue #23740, opened 2019-03-04
TouchableWithoutFeedback does not fire onPress like other Touchable Components


haddow777:
## 🐛 Bug Report

Same as closed bug #18611
TouchableWithoutFeedback does not call the onPress handler unless it's direct child is a View Component. This is counter to the fact that other touchable components will allow custom components as their child.

## To Reproduce

1. Create a FlatList
2. Create a ListItem that utilizes a TouchableWithoutFeedback component with an onPress handler
3. Add a non-View child Component, any custom component that renders content.
4. Test by tapping ListItem and find that no onPress event is fired.

## Expected Behavior

I would think that the expected behavior would be that the onPress event would be fired no matter what child component the TouchableWithoutFeedback component

## Code Example

[https://github.com/haddow777/tech_stack](https://github.com/haddow777/tech_stack)

Specifically look at the file ./src/components/ListItem.js. That is where the TouchableWithoutFeedback component is used. In it's current state, the event does not fire. If I encapsulate the current children in a View component tag, the event will fire. I changed the TouchableWithoutFeedback with a TouchableOpacity component without the View component child, and the onPress fires. It seems they should both work the same.

## Environment

```
React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (4) x64 Intel(R) Core(TM) i5-4300U CPU @ 1.90GHz
      Memory: 1.27 GB / 7.91 GB
    Binaries:
      Yarn: 1.13.0 - C:\Users\SURFACE\AppData\Roaming\npm\yarn.CMD
      npm: 6.8.0 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.2.0.0 AI-181.5540.7.32.5056338
```

`Android Emulator api 27
`
```
react-native-cli: 2.0.1
react-native: n/a - not inside a React Native project directory
```

CatapultJesse:
Repro Snack: https://snack.expo.io/@jkcooper/rn23740-touchablewithoutfeedback-wrapping-custom-component

radeno:
@haddow777 you are right. Other touchable components wraps child content to another `<View>` and pass all touchable props into it.
`TouchableWithoutFeedback` just clone child. And if child doesn't pass all RN props into nested `View` it doesn't work.

CatapultJesse:
Should there be a View inside a TouchableWithoutFeedback element? Im not sure what effect that would have on existing usages

amerllica:
[...]

PR #23871, opened 2019-03-12
Fix TouchableWithoutFeedback's Lack of View

RyanElliott10:
## Summary
This PR addresses the issue regarding `TouchableWithoutFeedback` components not correctly working without a `<View>` embedded within them. The most current [issue can be found here](https://github.com/facebook/react-native/issues/23740).

`render()` now returns a `View` -- rather than just cloning the children -- with all of the props passed into the TouchableWithoutFeedback component, as well as all children. This addresses the issue where the children's props weren't being properly passed down into a `View`.

## Changelog

[General] [Fixed] - `render()` now wraps all props and child props correctly in a view

## Test Plan

Pass all CircleCI Tests

### DIFF
--- a/Libraries/Components/Touchable/TouchableWithoutFeedback.js
+++ b/Libraries/Components/Touchable/TouchableWithoutFeedback.js
@@ -14,7 +14,9 @@ const React = require('React');
 const PropTypes = require('prop-types');
 const Touchable = require('Touchable');
 const View = require('View');
-const EdgeInsetsPropType = require('EdgeInsetsPropType');
+
+var EdgeInsetsPropType = require('EdgeInsetsPropType');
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks for your PR. We are planning on deprecating these components in the near future and will be open sourcing a new way of handling pressable elements. I don't think it makes sense at this point to introduce such a big change to a component like this so I'm going to close this PR but feel free to use this in your own code :)

mgcrea:
This issue keeps re-surfacing, would have been nice to properly fix it as it breaks a lot of code.

I haven't been able to make this PR work (encounters a `<View>` inside `<Text>` error).

@cpojer Any idea when the new touchables will be released?

Thanks!

PR #23872, opened 2019-03-12
[iOS] Fix semi-transparent backgrounds on Text components

### LINKED ISSUE #23849: Adding semi-transparent background color to Text component causing display issue on iOS
## 🐛 Bug Report
IMG_5858 (1) (1)
Adding semi-transparent background color to Text component causing display issue on iOS. 
It seems that background color is unnecessary reapplied under text characters inside Text component.

## To Reproduce
Apply semi-transparent `backgroundColor` to Text component (ex. `"rgba(255,0,0,.5)"`).

## Expected Behavior
Background color isn't reapplied under text characters.
(There is an easy workaround by adding `View` component wrapper to `Text` and applying background color to the wrapper)

## Code Example
https://snack.expo.io/@simek/rntext-semi-transparent-background-issue-ios

## Environment
```
    System:
      OS: macOS 10.14.3
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.10.1 - /usr/local/bin/node
      npm: 6.7.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react: 16.8.4 => 16.8.4 
      react-native: 0.58.6 => 0.58.6 
```

ericlewis:
## Summary
Fix #23849. When setting a semi-transparent background on text, it becomes obvious that we are drawing the background color twice. Since background color is handled by the view, we should not need to draw the glyph background color too.

| Before        | After   |
| ------------- |-------------|
|     |  

## Changelog

[iOS] [Fixed] - Semi-transparent backgrounds on text

## Test Plan

Added a new section to iOS RNTester Text Examples, CI passes. Also confirmed the TextExample snapshot tests passed with this change *before* I added the new test & snapshot.


ericlewis:
the test failing seems unrelated.

ericlewis:
I pushed changes to reflect (I think) what you said, but it breaks other places in the RNTester app.

#### If I understand you correctly, it should look like this:
image

#### But this has undesired effects when text doesn't have children:
image 2

So should the solution be to allow the background to be drawn so the after image works & so the nested backgrounds work?

Libraries/Text/Text/RCTTextShadowView.m
```diff
@@ -35,6 +35,11 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
   return self;
 }
 
+-(void)didSetProps:(NSArray<NSString*>*)changedProps {
```
shergin: Could you please add a comment here describing why we need this and apply code style?

ericlewis: Will do in the morning.

ericlewis:
@shergin fixing the snapshot now, changing the `TextExample.ios.js` dirtied it.

PR #23901, opened 2019-03-13
Add .eslintrc to RN project template

michalchudziak:
## Summary
The goal of this PR is to enable eslint checks in the projects generated by `react-native init` command. I added `template/_eslintrc` file, that would be replaced in an initialized project with `.eslintrc` file. This PR comes in parallel with https://github.com/react-native-community/react-native-cli/pull/229

## Changelog

[General] [Added] - Added `.eslintrc` file to generated template.

## Test Plan

1) Check out to this branch.

2) Follow setup from https://github.com/react-native-community/react-native-cli/blob/master/CONTRIBUTING.md

```bash
react-native init TestProject
cd TestProject && yarn lint
```

Expected result:
Eslint check ran with no issues found.

### DIFF
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -12,8 +12,6 @@ module.exports = {
 
   env: {
     es6: true,
-    jest: true,
-    'jest/globals': true,
   },
 
   plugins: [
@@ -29,6 +27,14 @@ module.exports = {
   settings: {
     react: {
       version: 'detect',
+    },
+  },
+
+  overrides: {
+    files: ['**/__tests__/**/*.js', '**/?(*.)(spec|test).js'],
+    env: {
+      jest: true,
+      'jest/globals': true,
     },
   },
 
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -12,6 +12,8 @@ module.exports = {
 
   env: {
     es6: true,
+    jest: true,
+    'jest/globals': true,
   },
 
   plugins: [
--- a/template/_eslintrc
+++ b/template/_eslintrc
@@ -1,0 +1,6 @@
+{
+  "env": {
+    "jest": true
+  },
+  "extends": "@react-native-community"
+}
--- a/template/_eslintrc
+++ b/template/_eslintrc
@@ -1,6 +1,9 @@
 {
+  "root": true,
+
+  "extends": "@react-native-community",
+  
   "env": {
     "jest": true
-  },
-  "extends": "@react-native-community"
+  }
 }
--- a/template/_eslintrc
+++ b/template/_eslintrc
@@ -1,9 +1,0 @@
-{
-  "root": true,
-
-  "extends": "@react-native-community",
-  
-  "env": {
-    "jest": true
-  }
-}
--- a/packages/eslint-config-react-native-community/package.json
+++ b/packages/eslint-config-react-native-community/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@react-native-community/eslint-config",
-  "version": "0.0.2",
+  "version": "0.0.3",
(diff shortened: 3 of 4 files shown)


template/_eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+  "root": true,
+
+  "extends": "@react-native-community",
+  
+  "env": {
+    "jest": true
```
satya164: I suggest something like this:

```js
  overrides: {
    files: ['**/__tests__/**/*.js', '**/?(*.)(spec|test).js'],
    env: {
      jest: true,
      'jest/globals': true,
    },
  },
```

michalchudziak: Thanks @satya164, I think it's a good point.

satya164: Maybe we should add it to community config itself?

michalchudziak: I think keeping eslint test runner configuration within the project is ok. Jest is integrated into RN template, but other projects might depend on a different test runner. I'm fine with both solutions tho.

template/_eslintrc
```diff
@@ -0,0 +1,9 @@
...
+  }
+}
```
cpojer: Please add a newline here.

template/_eslintrc
cpojer: @michalchudziak can we put it into the eslint config? I think that's better, especially since people can create a project without Jest still and it makes this a bit confusing.

matthargett:
The .eslintrc configuration file is deprecated according to eslint docs:
https://eslint.org/docs/user-guide/configuring#configuration-file-formats

My preference would be to use .eslintrc.js, so we can add comments, but .eslintrc.json would be fine if someone feels strongly otherwise.

cpojer:
Let's do `.js`.

Quick question: what happens to existing projects with an `.eslintrc.js` when they upgrade? Will this file be ignored or will it lead to merge conflicts?

kelset:
> Quick question: what happens to existing projects with an .eslintrc.js when they upgrade? Will this file be ignored or will it lead to merge conflicts? 

cc @pvinis do you know what will happen?

pvinis:
It will be a merge conflict, which is fine. Many people have their own, so it's good to see what they have and what the template suggests. The only weird case I see is if people have eslint json and we have eslint js, that might lead to some funkiness, but I don't think there are many things we can do here besides announce it to people.

packages/eslint-config-react-native-community/index.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
 
   env: {
     es6: true,
+    jest: true,
```
satya164: Shouldn't it be under `overrides` like:

```js
  overrides: {
    files: ['**/__tests__/**/*.js', '**/?(*.)(spec|test).js'],
    env: {
      jest: true,
      'jest/globals': true,
    },
  },
```

I think adding all jest globals to the normal env isn't great.

packages/eslint-config-react-native-community/index.js
```diff
@@ -24,6 +24,14 @@ module.exports = {
     'jest',
   ],
 
+  overrides: {
+    files: ['**/__tests__/**/*.js', '**/?(*.)(spec|test).js'],
```
thymikee: Should include TS and TSX files as well

michalchudziak: @thymikee Are you sure that's something we want to add? TSEslint configuration on the project side will be required anyway. It seems logical to keep it on the project side for the projects that use TS.

cpojer: Eslint now supports TS (or is planning to) so it makes sense.

satya164: Wouldn't we also need to add the parser as well? Otherwise, ESLint will fail to parse no?

Reference https://github.com/facebook/jest/blob/master/.eslintrc.js#L15

PR #23917, opened 2019-03-14
Fabric: add minimal ProgressView component for iOS

ericlewis:
## Summary

Codegen'd ProgressView, this brings RNTester’s example back to life!

## Changelog

[iOS] [Added] - Fabric: ProgressView

## Test Plan

View ProgressViewIOS example in RNTester with Fabric enabled.


React/Fabric/Mounting/ComponentViews/ProgressView/RCTProgressViewComponentView.h
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+
+#import <React/RCTViewComponentView.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * UIView class for root <ProgressView> component.
```
shergin: --root--

React/Fabric/Mounting/ComponentViews/ProgressView/RCTProgressViewComponentView.mm
```diff
@@ -0,0 +1,64 @@
...
+  [super updateProps:props oldProps:oldProps];
+
+  // `progress`
+  if (oldProgressViewProps.progress != newProgressViewProps.progress) {
+    _progressView.progress = newProgressViewProps.progress;
+  }
+
+  // `trackTintColor`
+  if (oldProgressViewProps.trackTintColor != newProgressViewProps.trackTintColor) {
+    _progressView.trackTintColor = [UIColor colorWithCGColor:newProgressViewProps.trackTintColor.get()];
```
shergin: `.get` is an implementation detail, we have to use special conversion function from RCTConvertions.h`.

React/Fabric/Mounting/ComponentViews/ProgressView/RCTProgressViewComponentView.h
ericlewis: @shergin all the components are like this lol
```
/**
 * UIView class for root <Slider> component.
 */
```

ericlewis:
@shergin I think that addresses your feedback, unless you want me to remove root- but other components have that too, so maybe I could do that in a diff PR..

fkgozali:
In the spirit of Lean Core, I'm not sure we want this in core? So I think we can hold off for now?

cpojer:
Closing based on @fkgozali's comment. Thanks for experimenting with all the new awesome features :)

PR #23923, opened 2019-03-14
[iOS] Fixed minuteInterval invalid in time mode

zhongwuzw:
## Summary

From https://github.com/facebook/react-native/pull/23861#issue-260337314
> changing "interval" example from "time-only" to "datetime" because there's a known bug that prevented the previous example from working

We need to ensure set minuteInterval after set datePickerMode, otherwise minuteInterval invalid in time mode.

cc. @grabbou @cpojer .

## Changelog

[iOS] [Fixed] - Fixed minuteInterval invalid in time mode

## Test Plan

minuteInterval can work in time mode.
```
            <DatePickerIOS
              testID="date-and-time-with-interval"
              date={state.date}
              minuteInterval={20}
              mode="time"
              onDateChange={onDateChange}
            />
```

### DIFF
--- a/RNTester/js/DatePickerIOSExample.js
+++ b/RNTester/js/DatePickerIOSExample.js
@@ -131,16 +131,16 @@ exports.examples = [
     },
   },
   {
-    title: 'Picker with 20-minute interval',
+    title: 'Time only picker, 20-minute interval',
     render: function(): React.Element<any> {
       return (
         <WithDatePickerData>
           {(state, onDateChange) => (
             <DatePickerIOS
-              testID="date-and-time-with-interval"
+              testID="time-with-interval"
               date={state.date}
               minuteInterval={20}
-              mode="datetime"
+              mode="time"
               onDateChange={onDateChange}
             />
           )}
--- a/React/Views/RCTDatePicker.m
+++ b/React/Views/RCTDatePicker.m
@@ -13,6 +13,7 @@
 @interface RCTDatePicker ()
 
 @property (nonatomic, copy) RCTBubblingEventBlock onChange;
+@property (nonatomic, assign) NSInteger reactMinuteInterval;
 
 @end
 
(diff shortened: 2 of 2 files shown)


radeno:
Should be implemented in extracted component too?
https://github.com/react-native-community/react-native-datetimepicker/tree/review

grabbou:
Didn't know we already extracted it! Let's definitely let the maintainers know to back port this.

@zhongwuzw, would you mind updating RNTester `DatePickerIOS` example (the last one) to actually feature "time" instead of "date time" as it is now? Would be good to revert the original example now that it's fixed!

PS. Awesome work, thanks for fixing this long-standing bug!

grabbou:
This fixes https://github.com/facebook/react-native/issues/9566

radeno:
@grabbou looks like `DateTimePicker` it is still in development or waiting for review.  But it is part of Lean Core and also PR in Proposals.
cc @Swaagie

Automobileengineeringservices:
#23923

zhongwuzw:
@grabbou Done! ❤️ 

@radeno Yes, DatePicker seems still in progress, I think we can fix this firstly, after some time, when DatePicker extracted from core, maybe we can let maintainers to cherry pick this. 🤔

zhongwuzw:
@cpojer 👌 I'll create a PR to its `review branch`.

BG687:
I am still having this issue on RN `0.59.3`, so maybe this PR didn't actually work? 

Here is my code: 

     `<DatePickerIOS
          date={dateTime}
          minuteInterval={interval}
          mode={mode}
          onDateChange={this.handleDateTimeChange}
          timeZoneOffsetInMinutes={timeZoneOffsetInMinutes}
        />`

zhongwuzw:
@BG687 This seems not be included in `0.59.3`, you can just cherry-pick this commit to test.

PR #23924, opened 2019-03-14
[iOS] Fix Info.plist templates to prevent parse error

friederbluemle:
## Summary

I ran into a build error when `xcodebuild` would invoke a custom script (in my case `sentry-cli`):

```
error: Could not parse Info.plist file
  caused by: invalid data
```

Upon further investigation it turned out the parse error was **caused by a comment**, and I found some other oddities with the default `Info.plist` template files:

- The problematic comment is a **dead link** (website does not exist)
- The main Info.plist contains a **duplicate entry** for `NSLocationWhenInUseUsageDescription`
- Some entries are _reordered_ and the comment is removed when modifying/saving the file in Xcode

This PR fixes these inconsistencies, unnecessary customizations, and potential sources of errors.

## Changelog

[iOS] [Fixed] - Fix Info.plist templates to prevent parse error

## Test Plan

In a custom app project the following command fails:

    xcodebuild "-workspace" "/path/ios/app.xcworkspace" "-scheme" "app" "clean" "archive" "-archivePath" /tmp/app

With the changes in this PR, the same command runs successfully.

### DIFF
--- a/template/ios/HelloWorld-tvOS/Info.plist
+++ b/template/ios/HelloWorld-tvOS/Info.plist
@@ -7,7 +7,7 @@
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)</string>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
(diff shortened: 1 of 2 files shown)


friederbluemle:
A couple related issues:

- https://github.com/getsentry/react-native-sentry/issues/333
- https://github.com/getsentry/sentry-cli/issues/226

Automobileengineeringservices:
#23924

PR #23927, opened 2019-03-14
TurboModules: add podspec

ericlewis:
## Summary

Add's a podspec to allow linking / building turbomodules.

## Changelog

[iOS] [Added] - podspec for turbomodules

## Test Plan

Included podspec for turbomodules, built RNTester.

### DIFF
--- a/ReactCommon/turbomodule/React-turbomodule.podspec
+++ b/ReactCommon/turbomodule/React-turbomodule.podspec
@@ -1,0 +1,49 @@
+# coding: utf-8
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+require "json"
+
+package = JSON.parse(File.read(File.join(__dir__, "..", "..", "package.json")))
+version = package['version']
+
+source = { :git => 'https://github.com/facebook/react-native.git' }
+if version == '1000.0.0'
+  # This is an unpublished version, use the latest commit hash of the react-native repo, which we’re presumably in.
+  source[:commit] = `git rev-parse HEAD`.strip
+else
+  source[:tag] = "v#{version}"
+end
+
+folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
+folly_version = '2018.10.22.00'
+boost_compiler_flags = '-Wno-documentation'
+
+Pod::Spec.new do |s|
+  s.name                   = "React-turbomodule"
+  s.version                = version
+  s.summary                = "-"  # TODO
+  s.homepage               = "http://facebook.github.io/react-native/"
+  s.license                = package["license"]
+  s.author                 = "Facebook, Inc. and its affiliates"
+  s.platforms              = { :ios => "9.0", :tvos => "9.2" }
+  s.source                 = source
+  s.source_files           = "**/*.{cpp,h}"
+  s.exclude_files          = "**/android/*"
+  s.compiler_flags         = folly_compiler_flags + ' ' + boost_compiler_flags
+  s.pod_target_xcconfig    = { "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/Folly\"" }
+  s.header_dir             = "jsireact"
+
+  s.dependency "React-Core", version
+  s.dependency "React-cxxreact", version
+  s.dependency "React-jsi", version
+  s.dependency "Folly", folly_version
+
+  s.subspec "RCTTurboModule" do |ss|
+    ss.source_files   = "**/*.{mm,h}"
+    ss.exclude_files  = "**/android/*"
+    ss.header_dir     = "React"
+  end
+end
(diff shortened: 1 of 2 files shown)


zhongwuzw:
@ericlewis Hey, good job first 👍 , just curious, we want to make every module to separate pod? and should integrate using CocoaPods?

ericlewis:
I’m not sure about the overall project, but for the bits that are still WIP / unstable, it’s nice to be able to integrate them separately from everything else. I would think as things get complete that we should see more convergence of podspecs (when it makes sense).

ReactCommon/turbomodule/React-turbomodule.podspec
```diff
@@ -0,0 +1,49 @@
...
+  s.pod_target_xcconfig    = { "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/Folly\"" }
+  s.header_dir             = "jsireact"
+
+  s.dependency "React-Core", version
+  s.dependency "React-cxxreact", version
+  s.dependency "React-jsi", version
+  s.dependency "Folly", folly_version
+
+  s.subspec "RCTTurboModule" do |ss|
+    ss.source_files   = "**/*.{mm,h}"
```
fkgozali: should probably include cpp here too (even if it doesn't exist today)

fkgozali:
> just curious, we want to make every module to separate pod? and should integrate using CocoaPods?

This is TBD, but if the modules are outside of the main repo, I'd expect a separate pod definition for each of them. We'll publish proper guide later when this whole system is ready.

fkgozali:
> I’m not sure about the overall project, but for the bits that are still WIP / unstable, it’s nice to be able to integrate them separately from everything else. I would think as things get complete that we should see more convergence of podspecs

There should be 1 podspec for the turbomodule core as described in this BUCK file: https://github.com/facebook/react-native/blob/e6f7d69428a4b3344ee5205c11bca58cbb4317fc/ReactCommon/turbomodule/core/BUCK

ReactCommon/turbomodule/React-turbomodule.podspec
fkgozali: Also, would it be possible to have 2 parts of this definition?
1. the shared part, under the turbomodule/core/*
2. the platform-specific part, under turbomodule/core/platform/ios/*

Sample:
1. https://github.com/facebook/react-native/blob/e6f7d69428a4b3344ee5205c11bca58cbb4317fc/ReactCommon/turbomodule/core/BUCK#L6-L13
2. https://github.com/facebook/react-native/blob/e6f7d69428a4b3344ee5205c11bca58cbb4317fc/ReactCommon/turbomodule/core/BUCK#L47-L61

You should probably move this podspec file to turbomodule/core/ subdir to mirror BUCK location

ericlewis: The shared part is the top level podspec, RCTTurboModule is the platform specific part. I can lock down the dirs being searched so it is clear.

ericlewis:
@fkgozali I think this addresses your feedback

ReactCommon/turbomodule/React-turbomodule.podspec
fkgozali: I see, let's then call the subspec `core` instead of `RCTTurboModule`.

ReactCommon/turbomodule/React-turbomodule.podspec
```diff
@@ -0,0 +1,49 @@
...
+  s.name                   = "React-turbomodule"
+  s.version                = version
+  s.summary                = "-"  # TODO
+  s.homepage               = "http://facebook.github.io/react-native/"
+  s.license                = package["license"]
+  s.author                 = "Facebook, Inc. and its affiliates"
+  s.platforms              = { :ios => "9.0", :tvos => "9.2" }
+  s.source                 = source
+  s.source_files           = "core/*.{cpp,h}"
+  s.exclude_files          = "**/android/*"
```
fkgozali: let's be explicit: `platform/android/*`

ReactCommon/turbomodule/React-turbomodule.podspec
```diff
@@ -0,0 +1,49 @@
...
+  s.header_dir             = "jsireact"
+  
+  s.dependency "React-Core", version
+  s.dependency "React-cxxreact", version
+  s.dependency "React-jsi", version
+  s.dependency "Folly", folly_version
+
+  s.subspec "RCTTurboModule" do |ss|
+    ss.source_files   = "core/platform/ios/*.{mm,cpp,h}"
+    ss.exclude_files  = "**/android/*"
```
fkgozali: let's be explicit `**/platform/android/*`

ReactCommon/turbomodule/React-turbomodule.podspec
```diff
@@ -0,0 +1,49 @@
...
+Pod::Spec.new do |s|
+  s.name                   = "React-turbomodule"
+  s.version                = version
+  s.summary                = "-"  # TODO
+  s.homepage               = "http://facebook.github.io/react-native/"
+  s.license                = package["license"]
+  s.author                 = "Facebook, Inc. and its affiliates"
+  s.platforms              = { :ios => "9.0", :tvos => "9.2" }
+  s.source                 = source
+  s.source_files           = "core/*.{cpp,h}"
```
fkgozali: btw, the source_files defn here doesn't intersect with the exclude_files, so do you still need the exclude_files?

ericlewis: We don't good point.

Automobileengineeringservices:
#23927

ericlewis:
@fkgozali addressed nits

PR #23933, opened 2019-03-14
[cleanup] Remove react-clone-referenced-element dependency

ide:
## Summary

This package was a dependency for ListView, which is no longer part of RN. Removed the dependency from package.json and yarn.lock.

Also removed its pattern from the Jest preset -- ListView is not in this repo and also react-clone-referenced-element was written at a time when Node did not support the same modern JS features as RN. The latest Node 8+ supports the features that react-clone-referenced-element uses so we don't need to transform it.

## Changelog

[General] [Removed] - Removed unused react-clone-referenced-element dependency 

## Test Plan

- `rg react-clone-referenced-element`
- Ensure CI passes

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -29,9 +29,7 @@ module.exports = {
       './jest/assetFileTransformer.js',
     ),
   },
-  transformIgnorePatterns: [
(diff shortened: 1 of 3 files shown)


Automobileengineeringservices:
#23933

PR #23939, opened 2019-03-14
Fabric: fix Switch crash

ericlewis:
## Summary

Switch isn’t subclassing UISwitch anymore, so this fills in the missing method that gets called.

## Changelog

[iOS] [Fixed] - Fabric: Switch crash

## Test Plan

Run RNTester with Fabric enabled, toggle switches. Observe no crash.


shergin:
That feature does not work not because we don't have this method implemented, it because Fabric does not have setNativeProps and command execution pipeline yet. This kinda change is hack and we don't want to have because it brings false feeling of solved problem.

ericlewis:
Indeed, I came to the same realization last night after pondering it somemore.

PR #23965, opened 2019-03-15
[General] [add] - Add openSettings method to Linking module

elucaswork:
## Summary

This will create a cross-platform and safe way to programmatically open the app's settings into the iOS /Android Settings app.

Right now it's possible to open the app's settings, but _**only for iOS**_ via `Linking.openURL("app-settings:")`

To do the same for Android, you need to either create NodeModule or install a dependency such as [react-native-open-settings](https://github.com/lunarmayor/react-native-open-settings).

Why this new method is useful: since Android 6, app permissions work similar to iOS. It's granular and it's requested in the app runtime. 

https://developer.android.com/guide/topics/permissions/overview#runtime_requests_android_60_and_higher

> If the device is running Android 6.0 (API level 23) or higher, and the app's targetSdkVersion is 23 or higher, the user isn't notified of any app permissions at install time. Your app must ask the user to grant the dangerous permissions at runtime. When your app requests permission, the user sees a system dialog telling the user which permission group your app is trying to access. The dialog includes a Deny and Allow button.

Thus, if the user checks the **"Never ask again box"** and taps **"Deny"**, for some specific permission, the only way to change the permission is going to the Android Setting app.

And that's where this new method becomes useful. It'll allow our apps to programmatically send the the user to  settings app.

Also, `openSettings()` doesn't receive a parameter to redirect to specific subsections of the Settings app because there's no public API to do it on iOS ([there's a way to have, via private API, but it causes the app to get rejected.](https://github.com/mauron85/cordova-plugin-background-geolocation/issues/394)) 

## Changelog

Create `Linking.openSettings()` for iOS and Android;

[General] [add ] - Add openSetting method to Linking module

## Test Plan

With `<Button onPress={() => Linking.openSettings()} title="Settings" color="#841584" />` it's possible to test this new method.

TODO: 
- [ ] In case of approval, create the PR updating the documentation.
- [x] Update `Linking` mock


elucaswork:
@cpojer I ended up overriding your commit by mistake with my force push. I just updated it including a commit doing the same you did. Thanks

cpojer:
That's alright, thanks for letting me know :)

PR #23966, opened 2019-03-15
Fix prop overrides of TouchableWithoutFeedback

aleclarson:
## Summary

Child props were being overridden by `<Touchable>` props even when the `<Touchable>` props were undefined.

## Changelog

[General] [Fixed] - Prevent prop override by TouchableWithoutFeedback when undefined

## Test Plan

None yet


cpojer:
Ah, thanks for the patch. I was just about to get back to you about that :D

PR #23967, opened 2019-03-15
Fix bat file

aliazizi:
Add reactNativePath to cli.js

## Summary

Add reactNativePath to cli.js

## Changelog

[GENERAL] [FIXED] - Internal - Update packagers to the latest CLI


hramos:
Thanks for the fix!

PR #23980, opened 2019-03-16
Remove unused argument in AppRegistry

gandreadis:
## Summary

The function `componentProviderInstrumentationHook` is defined in the same file and only accepts one argument, but is called with two. I've removed the second one, as it is not used (and might confuse readers).

For easier reference, this is the definition of that function:

```
let componentProviderInstrumentationHook: ComponentProviderInstrumentationHook = (
  component: ComponentProvider,
) => component();
```

## Changelog

Not necessary, as it does not change functionality.

## Test Plan

I would argue that this does not require testing: It removes a parameter that was not used before, so it should not change the behavior of the touched file in any way.

### DIFF
--- a/Libraries/ReactNative/AppRegistry.js
+++ b/Libraries/ReactNative/AppRegistry.js
@@ -109,10 +109,7 @@ const AppRegistry = {
       componentProvider,
       run: appParameters => {
         renderApplication(
-          componentProviderInstrumentationHook(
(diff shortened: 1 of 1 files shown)


gandreadis:
Nevermind, I made a mistake in reading the type annotation of this. While the function indeed only uses one argument, the type it conforms to has two :)

Issue #23151, opened 2019-01-24
Fetch fails on some https in android 4


seriiix:
## Environment
React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (4) x64 Intel(R) Core(TM) i5 CPU         760  @ 2.80GHz
      Memory: 2.19 GB / 7.99 GB
    Binaries:
      Yarn: 1.10.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.2.0.0 AI-181.5540.7.32.5014246

## Description
Some https URLs request fails on android 4 with no error while it's working on other androids (I've tested on some phones and it fails on 4.4.2 while successful on 5.0.2, 6.0.1, 9)

## Reproducible Demo
Create a project and paste the following code into your `App.js`. Make sure you are running on a device with android 4.4.2. It seems working on android 5+.
```
import React, {Component} from 'react';
import {Text, View} from 'react-native';

export default class App extends Component {

    componentDidMount() {
        /* this request fails on android 4 while its working on android 5+*/
        fetch('https://jsonplaceholder.typicode.com/posts/1')
            .then(resp => resp.json())
            .then(json => alert(`fail -> ${JSON.stringify(json)}`))
            .catch(err => {
                console.log(err)
            })


        /* this one is successful even on android 4 */
        fetch('https://www.json-generator.com/api/json/get/clfvuKZzOq?indent=2')
            .then(resp => resp.json())
            .then(json => alert(`success -> ${JSON.stringify(json)}`))
    }

    render() {
        return (
            <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
                <Text>https error on android 4</Text>
            </View>
        );
    }
}
```

seriiix:
I fixed this problem based on [this](https://github.com/facebook/react-native/issues/7192) issue and this [gist](https://gist.github.com/patrickhammond/0b13ec35160af758d98c). I don't know if this issue should remain open or not. It's clearly a problem. Anyway, [Here'](https://gist.github.com/seriiix/093289bb28747e26301432a5fba6b060)s RN fix to it.

dulmandakh:
it's fixed with https://github.com/facebook/react-native/commit/fab86ee4880f45ebfc0bd19d0e2fa1dbe15b7efc, please see commit description.

PR #23984, opened 2019-03-17
use Conscrypt as security provider if available

dulmandakh:
## Summary

This PR adds support to use Conscrypt as Security Provider if available runtime. Consscrypt supports TLS 1.2 on Android 4.x and TLS 1.3 on all Android versions. Fixes issues (ex https://github.com/facebook/react-native/issues/23151) with HTTPS connections on Android 4.x.

Just add below to your project build.gradle and it'll use it.

```gradle
implementation('org.conscrypt:conscrypt-android:2.0.0')
```

## Changelog

[Android] [Changed] - Add TLS 1.3 support to all Android versions using Conscrypt.

## Test Plan

CI is green and TLS 1.2 connections work on Android,

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -324,6 +324,7 @@ dependencies {
     api("com.facebook.fresco:imagepipeline-okhttp3:${FRESCO_VERSION}")
     api("com.facebook.soloader:soloader:${SO_LOADER_VERSION}")
     api("com.google.code.findbugs:jsr305:3.0.2")
+    implementation('org.conscrypt:conscrypt-android:2.0.0')
     api("com.squareup.okhttp3:okhttp:${OKHTTP_VERSION}")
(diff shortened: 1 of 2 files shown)


dulmandakh:
@hey99xx please review

sercand:
I think react-native shouldn't include TLS 1.3 support by default because:
- It adds 3mb to the apk file
- TLS 1.3 is not common enough to include a library
- Android Q adds TLS 1.3, ios 12.2 enables 1.3, why hurry?

gengjiawen:
I agree with @sercand . Maybe not a good time to add this.

dulmandakh:
Intention was to fix TLS on Android 4.x, and got TLS 1.3 as a bonus.

safciplak:
for anyone not resolved 
you can use that one

https://stackoverflow.com/questions/51902629/how-to-allow-all-network-connection-types-http-and-https-in-android-9-pie

PR #24015, opened 2019-03-18
There is a small gap in the SynchronizedWeakHashSet implementation. T…

dryganets:
## Summary
There is a small gap in the SynchronizedWeakHashSet implementation - the containsKey method of the WeakHashMap is modifying hence calling it during the iteration might cause ConcurrentModificationException. Added a command DO_IF_CONTAINS to safely handle this case.

## Changelog
[Android] [Bugfix] - Should fix a ConcurrentModificationException in onResume.

## Test Plan
It is hard to reproduce this problem as this is a race condition.
You rather need to have a good set of integration tests or check it in production.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java
@@ -176,14 +176,17 @@ public class ReactContext extends ContextWrapper {
               new Runnable() {
                 @Override
                 public void run() {
-                  if (!mLifecycleEventListeners.contains(listener)) {
-                    return;
-                  }
-                  try {
-                    listener.onHostResume();
-                  } catch (RuntimeException e) {
-                    handleException(e);
-                  }
+
+                  mLifecycleEventListeners.doIfContains(listener, new Runnable() {
+                    @Override
+                    public void run() {
+                      try {
+                        listener.onHostResume();
+                      } catch (RuntimeException e) {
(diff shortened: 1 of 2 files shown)


fkgozali:
Thanks for the fix! We'll update you if related crash still happens in the future. cc @cpojer

fkgozali:
@dryganets - looks like the same crash is still happening:

```
9Caused by: java.util.ConcurrentModificationException
10java.util.WeakHashMap$HashIterator.nextEntry(WeakHashMap.java:789)
11java.util.WeakHashMap$KeyIterator.next(WeakHashMap.java:822)
12+com.facebook.react.bridge.SynchronizedWeakHashSet.iterate(SynchronizedWeakHashSet.java:63)
13+com.facebook.react.bridge.ReactContext.onHostResume(ReactContext.java:218)
14+com.facebook.react.ReactInstanceManager.moveToResumedLifecycleState(ReactInstanceManager.java:660)
15+com.facebook.react.ReactInstanceManager.onHostResume(ReactInstanceManager.java:579)
16+com.facebook.react.ReactInstanceManager.onHostResume(ReactInstanceManager.java:535) 
```

Thoughts on further corner cases to handle? cc @cpojer

dryganets:
@fkgozali,
I don't see how this could happen now.
All the methods are synchronized.

The iterators are protected with the lock.
For the same thread ConcurrentModification, I had a boolean variable and it was protected with the same lock.

Issue #21006, opened 2018-09-07
Android 4.4.4 axios JSON.parse(response) error


wxcchdStar:
## Environment
    System:
      OS: macOS High Sierra 10.13.3
      CPU: x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
      Memory: 94.82 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.4 - ~/.nvm/versions/node/v8.11.4/bin/node
      npm: 5.6.0 - ~/.nvm/versions/node/v8.11.4/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
      Android SDK:
        Build Tools: 19.1.0, 20.0.0, 21.1.1, 21.1.2, 22.0.0, 22.0.1, 23.0.0, 23.0.1, 23.0.2, 23.0.3, 24.0.0, 24.0.1, 24.0.2, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.1, 27.0.2, 27.0.3, 28.0.2
        API Levels: 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27
    IDEs:
      Android Studio: 3.1 AI-173.4907809
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.4.1 => 16.4.1
      react-native: 0.56.0 => 0.56.0
    npmGlobalPackages:
      react-native-git-upgrade: 0.2.7

## Description
On Android 4.4.4，when response body is too big(~20kb), response body will cut the content and JSON.parse(response) will occur the `SyntaxError: Unexpected token i in JSON at position`. I print the request body, and this content like this: `{"id":1,"name":""isRequired":1}`, the content after `name` lose the `value",`. The complete content should be `{"id":1,"name":"value","isRequired":1}`.

I also set `maxContentLength: Number.MAX_VALUE,` but it's not work.

I using react native v0.56.0 and axios v0.18.0.

I also tried Fetch and it's work.

MarkDaleman:
Hallo! I've tried recreating this problem in React-Native version 0.56. I'm using Axios version 0.18.0.
[...]

PR #24038, opened 2019-03-19
android progressive decoding kitkat compatibility

dryganets:
## Summary
Turns out that Android KitKat has an issue with decoders.
This change fixes crash on KitKat devices.
See: https://github.com/facebook/react-native/issues/21006

Currently, only UTF-8 is supported.
UTF-16 and UTF-32 support will require different implementation, also there are different byte orders for those encodings and String in those encodings might contain BOM (Byte order mark).

## Changelog
[Android] [Fixed] - Fixes https://github.com/facebook/react-native/issues/21006

## Test Plan
In order to test this fix, you need to request the string larger than 20000 bytes over the network.
Ideally, byte 19999 should be beginning of the multi-byte symbol.


cpojer:
Thanks for this PR. I think I'm gonna close this for now given that Kitkat is relatively old by now and this PR introduces a fair bit of complexity. However, if people run into this issue and come upon this PR, they may be able to adapt it to their version of RN and use it in a fork.

biomancer:
@cpojer 
Kitkat is old but many devices have no way to upgrade to newer build. According to http://gs.statcounter.com/android-version-market-share/mobile-tablet/worldwide android 4.x share is  ~5%, meaning that every 20th android won't be able to use RN app if it requests chunks of data more than 20kb or so.

Issue #23868, opened 2019-03-12
PushNotificationIOS.removeEventListener method signature requires unused handler Function


schumannd:
https://github.com/facebook/react-native/blob/726d0012106316a40e36286f5eba0e881ef3bbba/Libraries/PushNotificationIOS/PushNotificationIOS.js#L236

This throws unhelpful flow errors. If this is an error, I can provide a PR removing this line.

ericlewis:
@schumannd it seems unused, you could try removing it.

grabbou:
I think we should actually fix the code to handle multiple listeners. Right now, if you call `addEventListener` two times, the latter one will silently overwrite the former one.

I believe this is kind of behaviour you wouldn't expect. 

If the API is designed to only handle single handler, it should be `setEventListener`.

cpojer:
This issue has been moved to react-native-community/react-native-push-notification-ios#10.

PR #24041, opened 2019-03-19
Fix PushNotificationIOS listeners

grabbou:
## Summary

Fixes https://github.com/facebook/react-native/issues/23868

Turns out if you set multiple PushNotificationIOS event handlers (for different types or two functions for same type), calling `removeEventHandler` will only remove the last one set.

This can cause some leaks and is just kind of behaviour that you wouldn't expect.

We also return listener itself from the `addEventListener` so that you can call `remove()`, just like for `BackButton`.

## Changelog

[iOS] [Fixed] - PushNotificationIOS properly remove listeners for registered events

## Test Plan

Set multiple listeners and then, remove each of them.

### DIFF
--- a/Libraries/EventEmitter/NativeEventEmitter.js
+++ b/Libraries/EventEmitter/NativeEventEmitter.js
@@ -16,7 +16,7 @@ const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
 
 const invariant = require('invariant');
 
-import type EmitterSubscription from 'EmitterSubscription';
+import type EmitterSubscription from '../vendor/emitter/EmitterSubscription';
 
 type NativeModule = {
   +addListener: (eventType: string) => void,
--- a/Libraries/EventEmitter/NativeEventEmitter.js
+++ b/Libraries/EventEmitter/NativeEventEmitter.js
@@ -16,7 +16,7 @@ const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');
 
 const invariant = require('invariant');
 
-import type EmitterSubscription from '../vendor/emitter/EmitterSubscription';
+import type EmitterSubscription from 'EmitterSubscription';
 
 type NativeModule = {
   +addListener: (eventType: string) => void,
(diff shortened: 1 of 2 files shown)


grabbou:
Let’s test this tomorrow

grabbou:
@cpojer tested, it's ready now.

cpojer:
I'll ship the fix in RN but would you mind also sending a PR to https://github.com/react-native-community/react-native-push-notification-ios as this module is being extracted at the moment? Sorry for the overhead.

cpojer:
One question: does anybody ever add the same handler to multiple event types? If yes, with this change, it would overwrite what is set in `_notifHandlers` and it would result in basically the same issue. I think we may need to store both the type and handler, no?

grabbou:
There's no good solution to that problem. 

I think we should just deprecate `removeListener` and educate people towards calling `.remove()` on a returned Subscription object. Otherwise, we are asking ourselves for troubles. To be honest, I can imagine a scenario where someone re-uses the same handler throughout different event types.

We should at least print a warning in such case (or even do `invariant` since this case is clearly not supported).

The convention of storing the listeners by handlers, not by type, seems to be used in other places [too](https://github.com/facebook/react-native/blob/aaa4a38fbcab9770b9731f96773e7357b2a6154f/Libraries/Network/NetInfo.js#L124), which means this aligns this module with others.

thymikee:
One solution would be to store each type in separate map to avoid removing same handler from different type of notification. 

However I'm onboard with removing `removeListener` as it's pretty common to do something like this:

```js
useEffect(() => {
  const unsubscribe = addEventListener('type', handler);
  return () => unsubscribe(); // or `unsubscribe.remove()` in this case
}, [handler]);
```

cpojer:
As @thymikee said, ideally we'll store all of these in a separate map. However, given that this module was moved out of RN I think we should just leave it as is here and move the corresponding issue to the [push-notification-ios](https://github.com/react-native-community/react-native-push-notification-ios) repo.

PR #24048, opened 2019-03-20
[Fabric] Fixed surface presenter observer

zhongwuzw:
## Summary

We missed create observer array, let's add it.
cc. @shergin .

## Changelog

[iOS] [Fixed] - Fixed surface presenter observer

## Test Plan

N/A.


shergin:
Wait a sec, how does work before?.. 🤔

PR #24052, opened 2019-03-20
[Fabric] Add lock for surface presenter observers

zhongwuzw:
## Summary

Add lock for observers when do enumeration.
cc. @shergin .

## Changelog

[iOS] [Fixed] - Add lock for surface presenter observers

## Test Plan

N/A.

### DIFF
--- a/React/Fabric/RCTSurfacePresenter.mm
+++ b/React/Fabric/RCTSurfacePresenter.mm
@@ -331,12 +331,8 @@ using namespace facebook::react;
 {
   RCTAssertMainQueue();
 
-  NSArray<id<RCTSurfacePresenterObserver>> *observers;
-  {
-    std::shared_lock<better::shared_mutex> lock(_observerListMutex);
-    observers = [_observers copy];
-  }
-  for (id<RCTSurfacePresenterObserver> observer in observers) {
+  std::shared_lock<better::shared_mutex> lock(_observerListMutex);
+  for (id<RCTSurfacePresenterObserver> observer in _observers) {
     if ([observer respondsToSelector:@selector(willMountComponentsWithRootTag:)]) {
       [observer willMountComponentsWithRootTag:rootTag];
     }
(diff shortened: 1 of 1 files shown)


React/Fabric/RCTSurfacePresenter.mm
```diff
@@ -331,7 +331,10 @@ - (void)mountingManager:(RCTMountingManager *)mountingManager willMountComponent
 {
   RCTAssertMainQueue();
 
-  for (id<RCTSurfacePresenterObserver> observer in _observers) {
+  std::unique_lock<std::mutex> lock(_observerListMutex);
```
shergin: We have to use `{ ... }` blocks to designate the scope covered by mutex. We must not call `.unlock` manually.
We also have to use better::shared_mutex and `std::shared_lock` here (and `std::unique_lock` where we modify the collection).

shergin:
The concert is totally valid. We must fix it.

React/Fabric/RCTSurfacePresenter.mm
zhongwuzw: Done, but seems may not give some benefits? 🤔  because reader seems always from main queue which mean only one reader. 😂

React/Fabric/RCTSurfacePresenter.mm
```diff
@@ -331,7 +331,12 @@ - (void)mountingManager:(RCTMountingManager *)mountingManager willMountComponent
 {
   RCTAssertMainQueue();
 
-  for (id<RCTSurfacePresenterObserver> observer in _observers) {
+  NSArray<id<RCTSurfacePresenterObserver>> *observers;
+  {
+    std::shared_lock<better::shared_mutex> lock(_observerListMutex);
+    observers = [_observers copy];
+  }
+  for (id<RCTSurfacePresenterObserver> observer in observers) {
```
shergin: Why not iterate over _observers while the lock is acquired?

ide: One thing to watch out for is deadlock if any of the observer delegate methods can add/remove observers unless better::shared_mutex is like recursive_mutex.

zhongwuzw: @shergin What I concerned is the same as @ide , we have no control what delegate would do. But if this deadlock situation shouldn't happened. I can remove copy.

shergin: `willMountComponentsWithRootTag:` is being called quite often and perf sensitive, so please do.

zhongwuzw: Done!

PR #24054, opened 2019-03-20
Picker Items should filter NULL child

Rahulkishanm:
## Summary

when conditional rendering is used then picker breaks when the child is null in iOS

This conditional rendering inside Picker fails when someBooleanValue variable is false
```
{
   this.state.someBooleanValue && <Picker.Item label="value" value="value" />
}
```

## Changelog

[iOS] [Fixed] - null check on children by using filter

## Test Plan

```
{
   this.state.someBooleanValue && <Picker.Item label="value" value="value" />
}
```


cpojer:
Duplicate of #24057.

PR #24066, opened 2019-03-20
allow HTTP in debug builds

dulmandakh:
## Summary

It allows HTTP connections in debug builds, and requires no configuration compared to previous/current solution where you need to edit config file to test on device.

Consulted with @Salakar about this.

## Changelog

[Android] [Changed] - Allow HTTP in debug builds

## Test Plan

CI is green


Salakar:
Much cleaner; thanks @dulmandakh 🎉

PR #24095, opened 2019-03-22
Additional Accessibility Roles and States

marcmulcahy:
## Summary

Assistive technologies use the accessibility role of a component to tell the disabled user what the component is, and provide hints about how to use it. Many important roles do not have analog AccessibilityTraits on iOS. This PR adds many critical roles, such as editabletext, checkbox, menu, and switch to name a few.

Accessibility states are used to convey the current state of a component. This PR adds several critical states such as checked, unchecked, on and off.

## Changelog

[general] [change] - Adds critical accessibility roles and states.

## Test Plan

* Added examples using all new roles. Ensure that both TalkBack on Android and VoiceOver on iOS properly identify the components using the new roles when touched. Add examples of new states, and ensure that state changes speak when the components are double tapped on both iOS and Android.


Libraries/Components/View/ViewAccessibility.js
```diff
@@ -51,7 +51,35 @@ export type AccessibilityRole =
   | 'adjustable'
   | 'imagebutton'
   | 'header'
-  | 'summary';
+  | 'summary'
+  | 'alert'
+  | 'checkbox'
+  | 'combobox'
+  | 'editabletext'
```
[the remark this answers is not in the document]
marcmulcahy: Good point. Would you prefer to remove this role for now and add "editabletext" or "textbox" as a separate PR?

RNTester/js/AccessibilityExample.js
```diff
@@ -133,6 +143,288 @@ class AccessibilityExample extends React.Component {
...
+        <View
+          accessibilityLabel="element 12"
+          accessibilityRole="spinbutton"
+          accessible={true}>
+          <Text>Spin button example</Text>
+        </View>
+        <View
+          accessibilityLabel="element 13"
+          accessibilityRole="switch"
+          accessibilityStates={['checked']}
```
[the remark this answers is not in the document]
marcmulcahy: Probably not, but semanticly, it's weird to say a switch is "checked". When describing the state of a switch verbally, we would usually say it is "on" or "off".

Libraries/DeprecatedPropTypes/DeprecatedViewAccessibility.js
```diff
@@ -49,7 +49,35 @@ module.exports = {
     'imagebutton',
     'header',
     'summary',
+    'alert',
+    'checkbox',
+    'combobox',
+    'editabletext',
```
jessebeach: Same issue as the `textbox` comment above.

Libraries/DeprecatedPropTypes/DeprecatedViewAccessibility.js
```diff
@@ -49,7 +49,35 @@ module.exports = {
...
+    'menu',
+    'menubar',
+    'menuitem',
+    'progressbar',
+    'radiobutton',
```
jessebeach: Same issue as the `radio` comment above.

RNTester/js/AccessibilityExample.js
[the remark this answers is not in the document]
jessebeach: We end up using 'on/off' quite a bit on native, especially for slide toggles.

RNTester/js/AccessibilityExample.js
```diff
@@ -133,6 +143,288 @@ class AccessibilityExample extends React.Component {
...
+    );
+  }
+}
+
+class AccessibilityRoleAndStateExample extends React.Component<{}> {
+  render() {
+    return (
+      <View>
+        <View
+          accessibilityLabel="element 1"
```
jessebeach: @marcmulcahy On web, we want to prefer using the element text content as the label. When a label is passed in as a prop, it's often assigned to `aria-label`, which overrides the text content. Will `accessibilityLabel` be assigned to `aria-label`? 

We've making a design decision here that we can't use `aria-labelledby` via React Native that targets web. I think we'll get by without it, but it might feel like a gap to some web developers.

necolas: I maintain the web implementation. accessibilityLabel is already mapped to aria-label. We discussed accessibility relationships in once of the other proposals, and that will provide a way to set labelled/described on web.

RNTester/js/AccessibilityExample.js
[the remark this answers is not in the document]
necolas: The question I have is whether on/off and checked/unchecked are redundant. If so, I think we don't need both. And the example here that uses a switch with checked/unchecked highlights the point of confusion

RNTester/js/AccessibilityExample.js
jessebeach: @necolas, good to know!

RNTester/js/AccessibilityExample.js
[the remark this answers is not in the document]
jessebeach: @necolas the restriction, as I understand it, is on iOS where the trait value is imply a string, not a token. So the announcement of state won't be automatically varied with the semantic role of the control. There is an argument to be made that once we introduce "on/off", we can't take it back. But it makes sense to me in the content of a `switch` role in contrast to a `checkbox` role.

React/Views/RCTView.m
```diff
@@ -184,6 +184,54 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
...
+                                      };
+  NSString *roleDescription = _accessibilityRole ? roleDescriptions[_accessibilityRole]: nil;
+  if (roleDescription) {
+    [value appendString:roleDescription];
+  }
+  for (NSString *state in _accessibilityStates) {
+    NSString *stateDescription =state ? stateDescriptions[state] : nil;
+    if (stateDescription) {
+        [value appendString:@" "];
+    [value appendString:stateDescription];
```
PeteTheHeat: Like you called out in your email, this doesn't localize well. We'll need to figure something out going forward.

React/Views/RCTViewManager.m
```diff
@@ -151,22 +168,47 @@ - (RCTShadowView *)shadowView
 
 RCT_CUSTOM_VIEW_PROPERTY(accessibilityRole, UIAccessibilityTraits, RCTView)
 {
-  // This mask must be kept in sync with the AccessibilityRoles enum defined in ViewAccessibility.js and DeprecatedViewAccessibility.js
   const UIAccessibilityTraits AccessibilityRolesMask = UIAccessibilityTraitNone | UIAccessibilityTraitButton | UIAccessibilityTraitLink | UIAccessibilityTraitSearchField | UIAccessibilityTraitImage | UIAccessibilityTraitKeyboardKey | UIAccessibilityTraitStaticText | UIAccessibilityTraitAdjustable | UIAccessibilityTraitHeader | UIAccessibilityTraitSummaryElement;
-
+  view.reactAccessibilityElement.accessibilityTraits = view.reactAccessibilityElement.accessibilityTraits & ~AccessibilityRolesMask;
   UIAccessibilityTraits newTraits = json ? [RCTConvert UIAccessibilityTraits:json] : defaultView.accessibilityTraits;
-  UIAccessibilityTraits maskedTraits = newTraits & AccessibilityRolesMask;
-  view.reactAccessibilityElement.accessibilityTraits = (view.reactAccessibilityElement.accessibilityTraits & ~AccessibilityRolesMask) | maskedTraits;
```
PeteTheHeat: Some of this convoluted logic existed to support state changes. (https://github.com/facebook/react-native/commit/5eaa2d29c0c4c633a40f7241408737c754edea84) 

Does this implementation still support that?
ie. Verify that a button that has accessibilityRole = button and accessibility state disabled can properly transition
to role = button and state = None

marcmulcahy: Yes, there's an additional test in RNTester to verify that this still works.

React/Views/RCTView.m
marcmulcahy: Any thoughts on how to do this? Since React Native is built as a static library, it wasn't clear to me how to ensure that string resources get bundled with apps that use it. This is already handled in Android.

Libraries/Components/View/ViewAccessibility.js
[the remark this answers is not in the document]
necolas: Feel free to move it to a separate PR if you prefer

React/Views/RCTView.m
PeteTheHeat: The classic iOS way is to use an `NSLocalizedString` here with translations in a `Localizable.strings` file in the bundle. Delivering the english only version of Localizable.strings seems reasonable, perhaps we can do that via CocoaPods?

marcmulcahy: This sounds promising. Any pointers at how to do this?

also, I'd like to use RNTester to validate this approach, and I don't think it's using CoCoPods yet, is it?

PeteTheHeat: This article explains `NSLocalizedString` pretty well https://nshipster.com/nslocalizedstring/

I'm not a CocoaPods expert, so that part I'm shaky on, perhaps ask in RN Discord (let me know if you need a link). If RNTester isn't using CocoaPods, it should be pretty easy to get translations working there. The trickier park will be integrating into apps that might already have a Localizable.strings file. 

Regardless, I think the localization bit should be a separate PR :)

marcmulcahy: Yep, I understand how to add localizable strings to an iOS project. We actually did that in some earlier internal revisions. The problem was getting those localizable strings into dependent projects, since React Native is built as a static library.

Unfortunately, there's no way to remove this code and still have the fallback paths functional. Do you propose we move forward with this PR as written, and then create a separate one to add the localized strings? If so, we'll need some support from Facebook to figure out a way to ensure that React Native apps get the localizable strings by default.

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -125,32 +138,53 @@ public void setAccessibilityComponentType(@Nonnull T view, String accessibilityC
...
-        view.setSelected(true);
-      } else if (state.equals("disabled")) {
-        view.setEnabled(false);
+      if (sStateDescription.containsKey(state)) {
+        updateViewContentDescription(view);
+      }
+    }
+  }
+
+  private void updateViewContentDescription(@Nonnull T view) {
```
chriscm2006: Background: Chris McMeeking, mobile accessibility expert, member WCAG Mobile Accessibility Task Force, etc. etc.

I love everything about this pull request except this function. Appending things to the content description is icky. Akin to Off Screen text in the web world. I'd rather see a feature request sent to Google. Programmatic association with Roles and States is crazy important. If TalkBack has not done their homework on how to communicate and keep track of those states, that is TalkBacks problem.

A solution like this will fragment the Android Accessibility Ecosystem even further than it already is. IMHO an effort like this should stick within the Operating Systems capabilities, and if those capabilities lead us to the conclusion that the platform is not Accessible, we should unitedly and vocally raise a fuss about this. But, appending a bunch of crud into Off Screen text and pretend like that fixes things, is a bad idea. We'd be better to embrace Android's features and refrain from fragmenting the Operating System even more than it already is.

marcmulcahy: I respectfully disagree. We're dealing with more than simply Android support with this pull request-- we're thinking about iOS and Web as well. And we can't afford to "dumb down" a toolkit like React Native to the least common denominator. If we do that, then we have no checkbox role, no checked state, etc. since iOS doesn't natively support those.

I wholeheartedly agree that this solution doesn't work well in braille, or for single switch access. But IMHO, improvements in speech access and support are better than no improvements at all.

The way this is architected, Javascript components expose the states and roles semantically, so if someday the situation on Android or iOS changes, we can re-implement the platform support layer to use native roles and states, which I think is the right approach.

chriscm2006: You can have roles and traits that you don't communicate. Off-Screen text is a known non-solution in the Web Sphere, and that is all that this is.

Having a trait in react native, and admitting that the Android operating system doesn't respect it, seems perfectly reasonable to me.

chriscm2006: > But IMHO, improvements in speech access and support are better than no improvements at all.

This is assuming that speaking roles is valuable. It frequently isn't. It is the other, programmatic behavioral responses to traits that matter much more.

chriscm2006: Quote from Android Accessibility Best Practices Page: Many accessibility services, such as TalkBack and BrailleBack, automatically announce an element's type after announcing its label, so you shouldn't include element types in your labels. For example, "submit" is a good label for a Button object, but "submitButton" isn't a good label.

If Android's best practices are wrong, we should ask them to change them. But a platform like React Native deciding that those best practices are wrong... bad plan.

marcmulcahy: Guess we'll agree to disagree on this. We are simply extending work Facebook began last summer, and I haven't seen from you an alternate  proposal which solves the problem(s) this PR addresses.

chriscm2006: Let me rephrase this for the TalkBack centric developer. Try this out:

1. Turn on TalkBack
2. Change it to `Controls` Navigation mode.
    1. Swipe Up until it says "Controls"
3. Note that most of your Pseudo Android roles are not marked up properly.

It is the ability to easily explore controls amidst a miriad of informative elements that makes a control a control NOT what it is announced as. Most of your pseudo roles are controls, and are lacking this behavior, which is significantly more important than appending that information to the "speakable text".

necolas: > Given this lack of clear purpose, such a role would be used very arbitrarily across the react-native ecosystem

The React Native ecosystem is not limited to mobile and includes desktop apps

blavalla: @chriscm2006, I think there may be some confusion here. This method only adds concatenated text for hints (which Android only supports if they are associated with an action, not just a generic string like iOS or web), and 5 different states (defined in the sStateDescription map). These states (on, off, expanded, collapsed, and busy) are not ones that allow users to change navigation modes to easily find.  Roles and States that can be semantically expressed in Android are handled correctly in the accessibility delegate here:

https://github.com/facebook/react-native/pull/24095/files/35a93f4f995bc849d90af0e8dd06aac873f66aff#diff-eee51b170ff20e1fc3aa718ca2d17bb1R116

I understand your concern though, as this definitely isn't a best practice approach when it can be avoided.  But if the choice was between not telling a user something is expanded (which is usually obvious visually), versus the user not having control to disable this announcement (as they can do by disabling state information), I'd err on the side of providing too much information over not enough. While this could be annoying for power users, the vast majority of users are not adjusting their settings from the default, and would simply be missing this state information which could be important to the understanding of the UI.

All that being said, I do think that on/off could be removed from this list, as semantically it used in the same way as "checked" and "unchecked".  In fact, when Talkback sees an element with the "Switch" role (the only one I know to use "on" and "off" as its labels), it uses the value of the checked property in the AccessibilityNodeInfo to append "on" or "off" to its readback.  Logic in the talkback compositor here:

https://github.com/google/talkback/blob/e69d4731fce02bb9e69613d0e48c29033cad4a98/compositor/src/main/res/raw/compositor.json#L998

If we want to keep the ReactNative API to use the state of "on" and "off", we can do so, but we should map this to setting the checkable and checked properties in the AccessibilityDelegate rather than appending them to the content description here.

chriscm2006: I think if I saw some of the items handled in the semantic way when it can be, I would feel better. It's one thing to create things that aren't there... I'll go quiet on that point, but it's another to ignore Operating System support that is there. The two examples that I can't get over.

- CollectionItemInfo for menu items.
- Ensuring that everything that is a control is either "mocked" as such, or throws exceptions or warnings when is isn't used properly.

chriscm2006: > > Given this lack of clear purpose, such a role would be used very arbitrarily across the react-native ecosystem
> 
> The React Native ecosystem is not limited to mobile and includes desktop apps

I'm not asking to drop those roles, I'm asking to ignore or mark up those roles properly for Android. Not all Reac Native apps need to behave the same way. In fact they shouldn't. They should behave the way the platform would normally behave.

React/Views/RCTView.m
```diff
@@ -184,6 +184,54 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
...
+                                     @"editabletext" : @"edit",
+                                     @"menu" : @"menu",
+                                     @"menubar" : @"menu bar",
+                                     @"menuitem" : @"menu item",
+                                     @"progressbar" : @"progress bar",
+                                     @"radiobutton" : @"radio button",
+                                     @"radiogroup" : @"radio group",
+                                     @"scrollbar" : @"scroll bar",
+                                     @"spinbutton" : @"spin button",
+                                     @"switch" : @"switch",
```
elucaswork: First of all, thanks for working on it.

1) What do you think of using `UISwitch().accessibilityTraits` for it? It would bring the default and **localizable** trait for switch role.

I was able to make it work (in my local react-native fork), based on this example https://github.com/jldailey13/24-Accessibility-Article-Examples/blob/master/24A11y%20Article/Name%20Examples/AccessibleSwitchView.swift

2) Swith component will benefit from this role https://github.com/facebook/react-native/blob/8774229f668adec94c1b76cc53e260f4a9354797/Libraries/Components/Switch/Switch.js#L169 👍

React/Views/RCTView.m
elucaswork: A feasible way to accomplish localizable roles and states would be having new props just override the default descriptions. 

The API would be like this:

```js
// src/index.js

// change descriptions globally from my app
if (currentLocale != 'en') {
  View.defaultProps.accessibilityRoleDescriptions = {
    radiogroup: translate('accessibility.role.radiogroup'),
  }

  View.defaultProps.accessibilityStateDescriptions = {
    expanded: translate('accessibility.states.radiogroup'),
  }
}
```

`accessibilityRoleDescriptions` and `accessibilityStateDescriptions` would only override prop keys passed from `stateDescriptions` and `roleDescriptions` arrays, for both iOS and Android.

Just to point out, IMO we could merge this PR without this localizable solution in case it demands too much. I'd rather have a non-localizable solution now than have nothing. And if case some help is needed, to have it in a follow-up PR, let me know.

Libraries/Components/View/ViewAccessibility.js
```diff
@@ -51,7 +51,35 @@ export type AccessibilityRole =
...
+  | 'menu'
+  | 'menubar'
+  | 'menuitem'
+  | 'progressbar'
+  | 'radiobutton'
```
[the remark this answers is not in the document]
marcmulcahy: Changed radiobutton -> radio.

React/Views/RCTView.m
marcmulcahy: Something like this could work, although it requires every app to maintain their own role and state descriptions. I'd love some mechanism that deploys localized descriptions into the new app template, so app developers don't have to think about it.

I'm in favor of moving forward with this PR as is, and creating an additional PR to sort out localization of the role and state descriptions.

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
marcmulcahy: To flip this argument around-- iOS doesn't have a checkbox trait or a checked trait. It's essential to know that something is a checkbox, and whether it is checked or unchedked. Android supports both a checkbox role convention (className == "android.widget.CheckBox" and the isChecked() method). So does that mean we shouldn't surface the checkbox role and associated checked/unchecked states on iOS because it doesn't support them?

We have used Android conventions where they exist to annotate the roles and states they represent, so setting class name as you suggest, as well as setting node metadata such as isChecked().

Happy to remove on/off in favor of checked/unchedked.

You can reasonably argue that menu and menuitem are not useful as roles on mobile, but that's a different argument than saying the way in which we support them is wrong. Using CollectionInfo and CollectionItemInfo only surfaces to the user that this is a list/grid, not that it's a menu. Knowing whether something is a menu or a standard list *could* be important IMHO.

We're using the roleDescription paradigm to annotate roles that don't have Android native analogs (equivalent Java class names). This is a paradigm introduced by Google for Chromium accessibility on Android where everything has a class name of "android.view.View".

Only states without AccessibilityNodeInfo analogs are concatenated to the content description, where we deemed such state information was useful.

I'm in favor of using Android-specific markup where it exists. If we've missed some standard markup, let us know and we'll fix it.

In summary:

* roles never get appended to the content description, they either cause the AccessibilityNodeInfo's class name to be set to the closest native analog, or a string description of the role to be placed in the roleDescription extra.
* States set the AccessibilityNodeInfo property where it exists, otherwise a string description of the state is appended to the content description.

React/Views/RCTView.m
marcmulcahy: Interestingly, this does seem correct, but we can't get it to work in our local trees. Switches are still spoken as buttons with a value of 1 or 0. Not sure what we're doing wrong...

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
chriscm2006: Checked/unchecked in favor of on/off I think is a good idea. Though you could also look into:

`setTextOn(), setTextOff()`

For scenarios when you want to be more expressive about the state of a given toggle-able trait.

I'd have to look closer at the roles where you add states, but the best Android Markup for times when you need to convey State AND Name, is to associate the stateful control and a `TextView` with a `labelFor` attribute. This may or may not be a solution in this case, I feel like that may turn this effort into something larger than it is...

React/Views/RCTView.m
chriscm2006: Hey @jldailey13, take a look @marcmulcahy has an interesting issue here!

elucaswork: @marcmulcahy

I was able to get the right trait by instantiating a new `UISwitch`

```
@synthesize a11ySwitch = _a11ySwitch;

....

_a11ySwitch = [UISwitch new];
view.reactAccessibilityElement.accessibilityTraits = [_a11ySwitch accessibilityTraits];
view.reactAccessibilityElement.accessibilityValue = isOn ? @"1" : @"0";

````

Comment (marcmulcahy) [reply React/Views/RCTView.m]: It looks like UISwitch is simply setting traits to Button. We tried something similar, but VoiceOver doesn't present the switch as a switch-- it says "button 1". Did you spin this up with VoiceOver and confirm that it properly describes the switch?

Comment (elucaswork) [reply React/Views/RCTView.m]: Weird. This is the full commit https://github.com/elucaswork/react-native/commit/80a304386101829baede42f2af7bf5813348dbb2. 
and yes, I was testing it with VoiceOver turned on. (I was very excited with the result)

Comment (marcmulcahy) [reply React/Views/RCTView.m]: That is exciting! I'll look closer at this full commit tomorrow, we must have missed something. It may be that VoiceOver is looking at more than the traits, and the class of the actual object is significant?

Comment (jldailey13) [reply React/Views/RCTView.m]: If you are making the accessibilityTraits of `RCTView` equal to `UISwitch().accessibilityTraits`, this should work, assuming that `RCTView` is inheriting from `UIView`.  I'm not sure what would be causing this to fail, given the information.

File: RNTester/js/AccessibilityExample.js [hunk 11]
```diff
@@ -133,6 +143,279 @@ class AccessibilityExample extends React.Component {
...
+          accessibilityRole="alert"
+          accessible={true}>
```
jessebeach: @marcmulcahy apologies if you've explained this in the past.

What would happen if the developer supplied `accessibleLabel` and `accessibleRole`, but set `accessible` to false? Is this a necessary orthogonal toggle? I can't think if a use case where one would supply a label and a role but not want the element to be "accessible".

marcmulcahy: @jessebeach I don't think this is a concern, as I think accessible=false would hide the component from accessibility services, so the label, roles, and states would be ignored.

React/Views/RCTView.m
marcmulcahy: I looked a bit closer at this. @elucaswork, how exactly did you test? I know you used VoiceOver-- which components did you play with?

The React Native switch component already behaved itself, since it uses a UISwitch underneath, so if that's what you used to test, you wouldn't see any change. It used to work before this change, and still does.

To validate the actual role, we created generic views (UIView underneath) and set the role to "switch". after applying what I think is a change similar to yours, VoiceOver says "label 1 button", which isn't what we want.

What I fear is going on is that VoiceOver is special case handling components based on their class names, not just their accessibility traits.


I hope I'm wrong, and we just missed a little detail. Let me know.

elucaswork: I've tested it with a `TouchableOpacity` (UIView as well):

```
<TouchableOpacity
  onPress={() => this.setState({switchOn: !this.state.switchOn})}
  accessibilityRole="switch"
  accessibilityStates={this.state.switchOn ? ['on'] : ['off']}>
  <Text>
    The switch value is {this.state.switchOn ? 'on' : 'off'}
  </Text>
</TouchableOpacity>
```

I'm gonna try later to play with this PR and see if I can replicate the same result that I got on my local branch. I let you know.

marcmulcahy: Interesting. Good to know. We can test with your patch here too... Although it conflicts with this PR, so we'll need to do a bit of surgery. :)

ikenwoo: There's actually a non-public trait that UISwitches use and it's `UIAccessibilityTraitToggle`. It's defined as `0x0020000000000000` ( 1 << 53 ). 

So you could do something like `switch.accessibilityTraits = UIAccessibilityTraitButton | 0x0020000000000000` (or define a constant) to get this behavior

React/Views/RCTViewManager.m
```diff
@@ -151,22 +167,47 @@ - (RCTShadowView *)shadowView
...
+  view.reactAccessibilityElement.accessibilityTraits = view.reactAccessibilityElement.accessibilityTraits & ~AccessibilityStatesMask;
 
-  UIAccessibilityTraits newTraits = json ? [RCTConvert UIAccessibilityTraits:json] : defaultView.accessibilityTraits;
-  UIAccessibilityTraits maskedTraits = newTraits & AccessibilityStatesMask;
-  view.reactAccessibilityElement.accessibilityTraits = (view.reactAccessibilityElement.accessibilityTraits & ~AccessibilityStatesMask) | maskedTraits;
+  for (NSString *state in states) {
+    if ([state isEqualToString:@"selected"]) {
+      view.reactAccessibilityElement.accessibilityTraits |= UIAccessibilityTraitSelected;
+    }
+    else if ([state isEqualToString:@"disabled"]) {
```
ikenwoo: nit: keep the `else` + `else if` in the same line as the previous `}`.

React/Views/RCTView.m
```diff
@@ -184,6 +184,53 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
   return YES;
 }
 
+- (NSString *)accessibilityValue
+{
+  NSMutableString *value = [NSMutableString stringWithString:@""];
+  NSDictionary *roleDescriptions = @{
```
ikenwoo: can you mark this and `stateDescriptions` as static so you're not creating these every time `accessibilityValue` is called? Or use a static constant defined outside of this method.

React/Views/RCTView.m
```diff
@@ -184,6 +184,53 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
   return YES;
 }
 
+- (NSString *)accessibilityValue
+{
+  NSMutableString *value = [NSMutableString stringWithString:@""];
```
ikenwoo: Rather than mutate a `NSMutableString`, build up a `NSMutableArray` and use `componentsJoinedByString:` at the end.

React/Views/RCTView.m
```diff
@@ -184,6 +184,53 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
...
+                                      @"collapsed" : @"collapsed",
+                                      };
+  NSString *roleDescription = _accessibilityRole ? roleDescriptions[_accessibilityRole]: nil;
+  if (roleDescription) {
+    [value appendString:roleDescription];
+  }
+  for (NSString *state in _accessibilityStates) {
+    NSString *stateDescription =state ? stateDescriptions[state] : nil;
+    if (stateDescription) {
+        [value appendString:@" "];
```
ikenwoo: separator here should be a "," since VoiceOver will make it sound more natural when reading a list of words.

React/Views/RCTView.m
ikenwoo: As an aside, while this constant isn't in the header, it's in the source so you can print out this symbol in lldb:

```
(lldb) po UIAccessibilityTraitToggle
0x0020000000000000
```

marcmulcahy:
I didn't use a comma for eventual localization concerns (this whole chunk of code needs a rethink to allow for role and state descriptions to be localized).

React/Views/RCTView.m
marcmulcahy: Inserted comment in the wrong field... I made this a space due to eventual localization concerns (this whole chunk of code needs a rethink to support localizing role and state descriptions). Seems like the right separator is likely different depending on language.

ikenwoo: The separator here doesn't need to be localized, it should be a comma regardless of localization. VoiceOver will insert pauses when it encounters commas, which will help this list of states sound more natural. The list of localized states should be comma separated in any translation. Otherwise, this list of states would be space separated and VoiceOver may read them out like a sentence.

e.g. the final string should be "combobox, radio, alert" instead of "combobox radio alert". Even when localized, it would be "<state A>, <state B>, <state C>"

marcmulcahy:
Thanks all for the great feedback. We're working on a final set of changes to add native iOS switch support, as well as make some of the suggested code cleanups recommended by @ikenwoo

I expect to have another revision posted by end of day.

marcmulcahy:
@ikenwoo could you please take another quick look to ensure we've adequately addressed your comments? Also, would love your feedback on how we can correctly support localization of the role and state description strings in iOS.

Thanks for all the  great feedback!

Libraries/Components/View/ViewAccessibility.js
```diff
@@ -51,7 +51,32 @@ export type AccessibilityRole =
...
+  | 'tab'
+  | 'tablist'
+  | 'timer'
+  | 'toolbar';
 
 // This must be kept in sync with the AccessibilityStatesMask in RCTViewManager.m
-export type AccessibilityStates = $ReadOnlyArray<'disabled' | 'selected'>;
+export type AccessibilityStates = $ReadOnlyArray<
+  | 'disabled'
+  | 'selected'
```
necolas: One issue with this `accessibilityStates` API in general is that to support states on web the `aria-*` property must always be set. For example, if a screenreader is to announce that a button is selectable then the DOM must include the `aria-selected` prop whether its value is `true` or `false`.

This means most states need a value equivalent to the `false` state (although there are tri-state states too), like we have with `checked` and `unchecked`. Or we could also support objects as values in the array of states, e.g., `[ 'checked', false ]` or `{ checked: 'mixed' }`.

marcmulcahy: @necolas, are you proposing a specific change to this PR?

necolas: I'm bringing up an issue with the `accessibilityStates` API / cc @jessebeach

marcmulcahy: The only alternate solution I can think of ATM would be to have the value of a state be a defined enumeration. So, define a state type:

BooleanState = {'true', 'false'}
TriState = {'true', 'false', 'mixed'}
... some others

then each state (checked, selected, enabled) defines its type.

But this seems like a significant complication. What benefit does it add for component developers and AT builders?

jessebeach: > But this seems like a significant complication. What benefit does it add for component developers and AT builders?

@necolas brings up a really important point. It would be impossible to build a tri-state checkbox with this API. @necolas might it be sufficient to introduce a 'mixedChecked' value to the state enum? 

>  if a screenreader is to announce that a button is selectable then the DOM must include the aria-selected prop whether its value is true or false.

@necolas the concept "selectableness" only applies to rows, tabs, options and cells. Could we address your concern by providing `unselected` and `selected` in the states enum? If `unselected` is passed to an element with a `row` class, then the web platform implementation would add `aria-selected="false"`. 

I think there are few enough of these exceptions that we can address them with the current state enum and some logic on the platform-specific target implementation.

necolas: (This is definitely not directly related to the PR but something to think about in terms of where we could steer RN accessibility APIs.)

I think we'd have to do it for quite a few states including `pressed`, `expanded`, etc. It's also a little inconvenient to work with strings in conditions (`isPressed ? 'pressed' : 'unpressed'`) compared to using booleans with ARIA attributes directly. From the perspective of multi-platform apps, it also seems like an enum makes it easier to accidentally not list "false" states and unwittingly produce poor AX for web (and perhaps other platforms). If `accessibilityStates` accepted an object some of these issues might be less likely, as the syntax encourages use of booleans and defining the attributes for every render.

```
const [ disabled, setDisabled ] = useState(false);
const [ checked, setChecked ] = useState(false);
// ...
return <View accessibilityStates={{ disabled, checked }} />
```

marcmulcahy: @necolas, I don't disagree, but am struggling to come up with a workable API. This maps reasonably for states like selected, enabled, pressed, checked, which are clearly booleans. Mixed checkboxes are a little trickier-- do they have three values (true, false, and mixed)? If so, the value isn't a boolean, is it a string?

If the value of a state isn't always a boolean, how do we define/enforce what the value of a state should be? And how does a developer know what type of value a particular state should be if they're all potentially different?

Seems like boolean states make sense in most of the cases I can think of-- maybe we just need to solve for mixed checkboxes? Are there other states that would need to be strings or some other value type?

necolas: Properties of the object that are of different or types can be annotated as such. I think most properties would also need to accept null/undefined values to cover the case where the platform property should be unset. Overall an API probably quite similar to some ideas we discussed for `accessibilityRelationships`.

```js
type AccessibilityStates = {
  checked?: ?boolean | 'mixed',
  disabled?: ?boolean,
  pressed?: ?boolean,
}

const [ disabled, setDisabled ] = useState(false);
const [ checked, setChecked ] = useState(false);
const [ pressed, setPressed ] = useState(false);

// ...do something that requires mixed checked state
setChecked('mixed');
// ...do something that means element should not be announced as pressable
setPressed(null);

return <View accessibilityStates={{ disabled, checked, pressed }} />
```

marcmulcahy:
Can someone help me understand the purpose of DeprecatedViewAccessibility.js and DeprecatedViewPropTypes.js. these are causing me problems when converting accessibilityStates from an array of strings to an object. Several components, including Button and touchableWithoutFeedback explicitly declare accessibilityStates as an array of deprecated types.

Can we safely remove DeprecatedAccessibility altogether and simply use the definitions in ViewAccessibility instead?

necolas:
> these are causing me problems when converting accessibilityStates from an array of strings to an object

@marcmulcahy I don't think that's necessary for this PR (as mentioned [here](https://github.com/facebook/react-native/pull/24095#discussion_r274226252)) or a change that has been requested yet. Changing the `accessibilityStates` API is something for a later patch if @jessebeach and others on the RN team think it sounds like a good idea and we have a migration/deprecation plan.

marcmulcahy:
OK, I was under the impression that this change was under consideration for this PR, and have the iOS implementation done, and am working on the Android implementation.

What outstanding issues remain on this PR for us to address?

cpojer:
What's the status of this PR? It is fairly large and I realize there is additional work for the future but is the current iteration shippable or are there still certain actions that need to be taken?

marcmulcahy:
To resolve conflicts, is it acceptable to simply rebase this code on top of master and force push this branch?

cpojer:
Yep, that’s the preferred way to do it.

________________________________

elucaswork:
@cpojer  It'd be awesome to have this PR and other accessibility improvements into `0.60`, such as: 
* Extended accessibility actions https://github.com/facebook/react-native/pull/24190, 
* `AccessibilityInfo` updates https://github.com/facebook/react-native/pull/23839, https://github.com/facebook/react-native/pull/23913.
* removal of `accessibilityComponentType` and `accessibilityTraits` props https://github.com/facebook/react-native/pull/24344

(I'm also working to submit another a11y PR implementing `onAccessibilityFocus` & `onAccessibilityBlur`)

marcmulcahy:
I've got the current code rebased, as well as an implementation which converts accessibilityStates from a list of strings to an object:

<accessibilityStates={['checked', 'collapsed']}> to
<accessibilityStates={{'checked" true, "expanded": false}}>


Should I include that in this PR, or should we consider that as a separate PR?

cpojer:
Let’s do that in a follow-up, this PR is already pretty big.

________________________________

marcmulcahy:
OK to delete the a11y-roles-states branch in my fork? Will the code and conversation history be preserved?

cpojer:
Yeah, go ahead!

elucaswork:
@marcmulcahy are you planning to make a documentation PR with those updates? (https://github.com/facebook/react-native-website) Please let me know if you need any help on it.

cpojer:
@marcmulcahy @elucaswork After merging I got an internal test failure that a `<Button disabled={true} />` was still clickable on Android. Is this something you could look into and send a fix for? I would prefer not to revert this PR.

If you can verify that this is still working as expected, then it may be an issue with our e2e framework, in which case I can look into it.

marcmulcahy:
@cpojer Sure, we'll take a look. I don't think we touched any non-a11y paths, but we'll double check.

marcmulcahy:
@elucaswork Happy to do that, but would like some feedback on #24608 first, as it changes the a11y states API, and is probably close to where we want the API to land longer term.

cpojer:
Thanks Marc,

the failure definitely bisected to a change in this PR but it’s unclear to me whether the behavior actually changed or the test framework needs a change.

________________________________

marcmulcahy:
@cpojer Is there a set of tests in the public repo we should be using, or is it as simple as creating a disabled button that shouldn't be clickable on Android?

cpojer:
It should be enough to check whether it works or not on master, yes!

________________________________

marcmulcahy:
@cpojer  We were unable to reproduce the problem when we did the following:

* OPen RNTester
* Select Button module
* Attempt to click the disabled button

The button doesn't appear to click either with or without this patch applied.

Happy to troubleshoot this if you can help us re-create the failing test.

cpojer:
Thank you so much! Can you confirm that the disabled state also works correctly on other elements?

I’ll check the internal test again tomorrow and see if we need to fix something up or if it’s the test.

________________________________

xuelgong:
@cpojer We have the test of disabled TouchableOpacity in AccessibilityExamples.js. Cannot repro the problem.

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -112,38 +122,60 @@ public void setNativeId(@Nonnull T view, String nativeId) {
...
       return;
     }
-
     view.setTag(R.id.accessibility_role, AccessibilityRole.fromValue(accessibilityRole));
   }
 
   @ReactProp(name = PROP_ACCESSIBILITY_STATES)
   public void setViewStates(@Nonnull T view, @Nullable ReadableArray accessibilityStates) {
-    view.setSelected(false);
-    view.setEnabled(true);
```
marcmulcahy: @Cpojer I suspect this is the culprit. This code is broken, which is why we proposed removing it. That is, setting accessibilityStates shouldn't change the view state like this. But you could imagine that if a test passed an empty state set, as I think Button does:

<Button accessibilityStates={[]}>

That causes the view to be enabled. Likewise, if you pass:

<Button <accessibilityStates={['disabled']}

This erroneously disables the view.

IMHO, setting accessibility states shouldn't affect the state of the underlying view.

In an ideal world, the accessibility "disabled" and "selected" states simply mirror the underlying view states.

Not sure what the right fix is here...

xuelgong: I was able to repro the bug in the AccessibilityExample.js by only setting the accessibility state to be disabled not the disable property itself. I guess the internal test of disabled button does the same thing.

blavalla: I have been digging into the test failure as well, and found that the test is only looking at the View's "enabled" property, and checking if it is false.  By setting the accessibilityState to disabled we are not longer setting the view's "enabled" property to false, which means the test now fails.

It seems to me like we should be keeping the accessibilityState's and the viewState's in sync wherever possible (selected, enabled, focusable, clickable, longClickable, etc.) otherwise you could have a view that is disabled visually, but not disabled in the AccessibilityNodeInfo, or disabled in the AccessibilityNodeInfo, but not disabled for users interacting without an AT.

The other option is to not even have these as specific accessibilityStates, but have them only exist as viewStates that we then set the accessibilityStates to automatically.  I'm not sure if this would work cross platform however, as not every view on iOS can be "Selected" or "enabled" like it can be on Android.

marcmulcahy: @blavalla I completely agree-- the goal is to have the view states and accessibility states in sync. But it seems to me that the test is broken if it relies on calling setAccessibilityStates with no states set to ensure that the view itself is disabled. Clearly, we wouldn't remove the view enabled property, and the accessibility enabled state should definitely match the view enabled state.

Can the test be modified to explicitly set the view state to disabled, which seems more correct, rather than relying on setting the accessibility states of the view to disable it?

I could perhaps be persuaded that setting the accessibility state of a view to disabled, should propagate to the view and disable the entire view as well, although this doesn't feel quite right to me...

cpojer: @marcmulcahy the current test is testing for `<View disabled={true} />` and that seems to trigger the failure.

necolas: disabled is a valid View prop?

cpojer: Sorry, it was late on Friday and I didn't verify. I meant `<Text disabled={true} />` is not working properly for testing after this PR. This prop is Android only and is only used for testing: https://github.com/facebook/react-native/blob/master/Libraries/Text/TextPropTypes.js#L136

@marcmulcahy would you be able to send a PR to restore the previous behavior on Android?

marcmulcahy: @cpojer Of course I'll submit a PR to fix it. I spent some time looking at this again this morning, and I'm not sure how to correctly fix it. As I see it, the issue is that we have three uses of enabled/disabled:

* This testing-only use in <Text>-- perhaps other places?
* the View state itself (Checkbox, Slider, etc.).
* The accessibility disable state.

As best I can understand this, the view's enabled state should be the master state-- I.E., all other states should derive from that. So however the Android view's enabled state gets set, both the testing only states, as well as the accessibility disabled state should reflect that.

Presumably, React Native has a way to specify that components are visible, but can't be interacted with, (disabled, greyed, etc.). When this state is set, the accessibility and testing disabled states should also reflect that change.

I think it would be reasonable for the testing disabled state to also serve as a master state-- that is, view disabled and accessibility disabled are synced to that state as well.

What I'm not convinced of, is that when the accessibility disabled state is set, the view state should be changed. This seems backwards.

From an API point of view, we don't really want components setting the accessibility disabled state to "disable" the component. This is what's happening, and the test code seems to be relying on this peculiarity of how things are implemented.

What I can't work out is how the test, by setting one of these "disabled" states, was causing the accessibilityStates['disabled'] to be set, which in turn caused the view.setEnabled(false) to be called.

If you can help me work that out, I think we can propose a fix that will both fix this problem and be correct.

We can revert the accessibility enabled state code, but I think this propagates undesirable uses of the accessibility API and apps relying on it to control main view state, which IMHO, isn't desirable.

cpojer: Regarding your proposed fix, I think that makes a lot of sense, if you can send a PR to that effect, I think that'll solve our issues.

I'm not entirely sure about how this all works, though. I'm sorry I'm not an expert. Our internal testing framework as I understand it, simply calls this code: https://github.com/selendroid/selendroid/blob/44118d68955047a0645c707a4e021c5de5d53b99/selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java#L654

Is that enough for you to reverse engineer the necessary change?

React/Views/RCTViewManager.m
```diff
@@ -150,12 +166,44 @@ - (RCTShadowView *)shadowView
...
+  for (NSString *state in states) {
+    if ([state isEqualToString:@"selected"]) {
+      view.reactAccessibilityElement.accessibilityTraits |= UIAccessibilityTraitSelected;
+    } else if ([state isEqualToString:@"disabled"]) {
+      view.reactAccessibilityElement.accessibilityTraits |= UIAccessibilityTraitNotEnabled;
+    } else {
+      [newStates addObject:state];
+    }
+  }
+  if (newStates.count > 0) {
```
janicduplessis: I'm getting a crash here with react-native-gesture-handler. Basically it uses `RCTViewManager` with a `UIControl` instead of a `RCTView` so the cast here is not safe. Maybe we should add a `isKindOfClass` check before casting here?

cpojer: Good idea. Do you wanna send a PR?

marcmulcahy:
PR to add additional a11y roles and states to the documentation is here:
https://github.com/facebook/react-native-website/pull/912

PR #24100, opened 2019-03-22
Add typescript support to the ESLint config

satya164:
## Summary

This adds TypeScript support to the community ESLint config. Our babel preset supports TypeScript by default, so it's nice to have TypeScript support pre-configured in the ESLint config too.

Note: Users need to install `typescript` in their project for linting to work for TypeScript files.

## Changelog

[General] [Added] - Add TypeScript support to the ESLint config

## Test Plan

Create files such as `Foo.ts` and `Foo.test.ts` and run `yarn eslint --ext .js,.ts,.tsx .` to see if the errors are reported correctly.

Note: In the react-native repo, ESLint fails to resolve `@typescript-eslint/plugin` when run from root, unless it's installed as a dev dependency, but it works in normal projects. I tried using `require.resolve`, but it doesn't work for plugins.

### DIFF
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -8,15 +8,16 @@
  */
 
 module.exports = {
-  parser: 'babel-eslint',
-
   env: {
     es6: true,
   },
 
+  parserOptions: {
+    sourceType: 'module',
+  },
+
   plugins: [
     'eslint-comments',
-    'flowtype',
     'prettier',
     'react',
     'react-hooks',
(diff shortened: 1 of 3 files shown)


packages/eslint-config-react-native-community/index.js
```diff
@@ -30,13 +31,39 @@ module.exports = {
...
-      'jest/globals': true,
+  overrides: [
+    {
+      files: ['*.js'],
+      parser: 'babel-eslint',
+      plugins: ['flowtype'],
+      rules: {
+        // Flow Plugin
+        // The following rules are made available via `eslint-plugin-flowtype`
+
```
thymikee: unnecessary newline :D

satya164: @thymikee just following the convention from rest of the file

packages/eslint-config-react-native-community/index.js
```diff
@@ -30,13 +31,39 @@ module.exports = {
...
   },
 
-  overrides: {
-    files: ['**/__tests__/**/*.js', '**/*.spec.js', '**/*.test.js'],
-    env: {
-      jest: true,
-      'jest/globals': true,
+  overrides: [
+    {
+      files: ['*.js'],
```
Titozzz:
```suggestion
      files: ['*.jsx?'],
```

Maybe that would be better?

packages/eslint-config-react-native-community/index.js
```diff
@@ -30,13 +31,39 @@ module.exports = {
...
+        // Flow Plugin
+        // The following rules are made available via `eslint-plugin-flowtype`
+
+        'flowtype/define-flow-type': 1,
+        'flowtype/use-flow-type': 1,
+      },
     },
-  },
+    {
+      files: ['*.ts', '*.tsx'],
```
Titozzz:
```suggestion
      files: ['*.tsx?'],
```

To match jsx? 😄

packages/eslint-config-react-native-community/index.js
satya164: Not sure about it coz the react or react native codebase dosen't use jsx extension or recommend it.

packages/eslint-config-react-native-community/index.js
satya164: It's not the correct glob syntax for this afaik. from bash docs:

> If you specified something at the command line like "hd?" GNU/Linux would look for hda, hdb, hdc and every other letter/number between a-z, 0-9.

so `.tsx?` will look for `.tsxa`, `.tsxb` etc.

Titozzz: sorry I meant 
files: ['*.ts?(x)'],
but it's not that important :)

PR #22463, opened 2018-11-30
Fixes #10964 TextInput maxLength cuts Unicode characters (Emojis) 

hzxiaosheng:
Fixes #10964
Fixes #10929

Summary:

1. When maxLength is set on TextInput, and allowedLength is 1, when the user picks an emoji character which length is 2 (or 3, 4), then maxLegnth will call `[@"😁" substringToIndex:1]` which will result in clearing all text already input (issue #10929 [iOS][TextInput] emoji after a new line make the line disappear).

2. Fix for this is to check the existence of emoji characters（NSStringEnumerationByComposedCharacterSequences）and shrink allowedLength, ensure that emoji character was rejected as an unit.

Test Plan:
----------
Use the RNTester app and navigate to TextInput test page and we can test input emojis in the "TextInput maxLength" section to replay the bug and verify this commit. 

Changelog:
----------
[IOS] [BUG] [TextInput] Fixes  #10964 TextInput maxLength cuts Unicode characters (Emojis).

PR #24109, opened 2019-03-23
[iOS] Fixed maxLength of TextInput based on glyph count

### LINKED ISSUE #10964: TextInput maxlength cuts Unicode caracters (Emojis)
Hello everyone,

yesterday i noticed that a TextInput has a problem displaying Emojis at the end of the fields maxlength. 
If i insert the "white happy man with black hair raising one hand" Emoji (iOS) at the position maxlength-1 it shows a square with an "A" in it instead.
If i insert the "white happy man with black hair raising one hand" Emoji (iOS) at the position maxlength-2 it shows a square with an "yellow happy woman with blonde hair raising one hand" in it instead.
If i insert the "white happy man with black hair raising one hand" Emoji (iOS) at the position maxlength-3 it shows a square with an "white happy woman with black hair raising one hand" in it instead.
If i insert the "white happy man with black hair raising one hand" Emoji (iOS) at the position maxlength-4 it shows a square with an "yellow happy woman with blonde hair raising one hand" and the square with an "A" in it instead.

Seems like the maxlength cuts the unicode sign at the end and causes to show different Emojis than inserted

* React Native version: 0.36
* Platform: iOS - tested on iOS 10.1.1

zhongwuzw:
## Summary

Fixes #10964 #22463 .
We need changed `maxLength` to count glyph instead. Because like emoji, some language，it represents using multiple bytes, so if we limit using characters, would leads some weird bugs, like clear the line or glyph not show correctly.

cc. @cpojer .

## Changelog

[iOS] [Fixed] - Fixed maxLength of TextInput based on glyph count

## Test Plan

If we input one emoji, it needs to show correctly.
```
<TextInput maxLength={1} />
```


Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -20,6 +20,25 @@
 #import "RCTTextAttributes.h"
 #import "RCTTextSelection.h"
 
+@interface NSString (RCTUtility)
```
shergin: We have to decouple that into separate file(s).

zhongwuzw:
@shegin Is this what we want? 🤔 I restructure the code and consolidate max length handler into separate function.

Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -619,6 +610,38 @@ - (void)handleInputAccessoryDoneButton
 
 #pragma mark - Helpers
 
+- (void)trimInputTextInRange:(NSRange)range
```
shergin: Not exactly.
This separate function is still a method (it mutates `self`) which does not solve the complexity problem.

zhongwuzw: @shergin Yeah, actually, Swift already solved this glyphs issue, but OC, we seems need to do these heavy things by ourself. Have no idea how to move on. 😂

shergin: Yeah, Swift's strings are character-aware but normal Objective-C are not. Yeah... but that's only String's properties, that is not related to UITextField and UITextView implementation, right? So, we already have two subclasses for those classes that enhance the behaviors and fix some issue. Maybe we can encapsulate that character-aware logic there?

Or, what if we move that `maxLength` feature implementation into those classes and try to make it totally transparent for all external logic?

zhongwuzw:
@shergin 
Hi :) I may need your some inputs, is this the right direction you wanted? 🤔

cpojer:
I thought about this some more and we have a ton of code that on the backend expects maxLength to adhere to the real length and not the human readable length. Changing the default now and only for iOS will likely cause a lot of issues for us at Facebook and others.

If we want to solve this I think we may want to add another prop that explicitly specifies the max length is based on the glyph count instead of the number of bytes and make people opt-in to use that prop in the cases where the human readable length matters to them.

zhongwuzw:
@cpojer 👌 We already improved TextInput so much, let's wait to see wether any user wants it, then we can add it. :)

Issue #21830, opened 2018-10-17
Wrong translation from react-native's accessibilityLabel to iOS name/label


melchior-by:
## Environment
  OS: macOS High Sierra 10.13.6
  Node: 10.8.0
  Yarn: Not Found
  npm: 6.4.1
  Watchman: Not Found
  Xcode: Xcode 10.0 Build version 10A255
  Android Studio: 3.1 AI-173.4907809

  Packages: (wanted => installed)
    react: ^16.4.1 => 16.4.1
    react-native: ^0.55.4 => 0.55.4

## Description
When accessibilityLabel set to the element in react-native code it translates to the iOS name/label, but all parent element's name/label accumulate accessibilityLabel value of a child element.

In Appium inspector it's looks like:
```
<e1 name ='label 2 label 1'>
  <e2 name ='label 2'>
    <e3 name ='label 1'>
      <e4 name ='label 1'>
```

But in native code:
```
<e1>
  <e2 accessibilityLabel ='label 2'>
    <e3>
      <e4 accessibilityLabel ='label 1'>
```

It completely ruins all efforts to automate testing of iOS app built from react-native code.
I investigated this issue and found it appears from work of RCTRecursiveAccessibilityLabel() function, which called in RCTView.m file in this code:
```
- (NSString *)accessibilityLabel
{
  NSString *label = super.accessibilityLabel;
  if (label) {
    return label;
  }
  return RCTRecursiveAccessibilityLabel(self);
}
```
Change `return RCTRecursiveAccessibilityLabel(self);` to `return nil;` - fix this issue, so I suggest this code changes to be implemented.

LorandP:
This problem is still persisting also on 0.57.7

lucaGazzola:
Is there a plan of fixing this behaviour in upcoming releases? Testing iOS applications using appium  is hell atm

ZacMonroe:
This issue persists for me on 58.6

PR #24118, opened 2019-03-24
Fix accessibility label concatenation on iOS

LorandP:
## Summary

When using Appium in order to provide automated tests a unique accessibility label has to be specified for a single element in order to be identified in the Appium inspector tool. However, on iOS, the accessibility label for a single element contained the labels of all the elements on the screen concatenated into one string.
I have provided a fix that would stop concatenating the accessibility labels of other elements from the screen.

Also, other issues opened on this:
https://github.com/facebook/react-native/issues/21830
https://github.com/appium/appium/issues/10654

## Changelog

[IOS][Changed] - Fix concatenation of accessibilityLabels for an element

## Test Plan

Before the fix, the accessibility labels were showing like this in the Appium inspector. https://ibb.co/Nxq4Cdh
I have run the following tests in order to make sure that the current modifications are working and that the old functionality is still working.
1. I have run the integration test locally from ./scripts/objc-test-ios.sh and the results have passed
2. I have also run the RNTester module on a simulator and all the components were working as expected
3. We have run automated tests on our local changes and it has been confirmed that the accessibility labels were no longer concatenated and that every element contained the label that was set in the project.

### DIFF
--- a/React/Views/RCTView.m
+++ b/React/Views/RCTView.m
@@ -151,7 +151,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
   if (label) {
     return label;
   }
-  return RCTRecursiveAccessibilityLabel(self);
+  return RCTRecursiveAccessibilityLabel(nil);
 }
 
 - (NSArray <UIAccessibilityCustomAction *> *)accessibilityCustomActions


shergin:
@rigdern What do you thing?

LorandP:
@shergin Hy is there a way in which I can see the entire test that is being run by circleci for the test_detox_end_to_end?

I have detox setup locally and I would like to reproduce the error in order to fix that failed test.

shergin:
@LorandP TBH, ¯\_(ツ)_/¯ .

shergin:
I am open to discussing this thing, but this implementation is incorrect because using `nil` as a parameter makes existence of `RCTRecursiveAccessibilityLabel` pointless.

LorandP:
I agree I will try to find a better solution for this and if you have any suggestions just let me know. Thanks for the feedback.

kat-max:
Any progress on this one? I am still experiencing the same issue :/

PR #24138, opened 2019-03-25
Improvement: Adjust template to match new init command

Esemesek:
## Summary

Adjusting template to new init command (https://github.com/react-native-community/react-native-cli/pull/241).

PR with new init command needs to be merged before we land this.

## Changelog

[General] [Changed] - Adjust template to match new init command

## Test Plan

A project is initializing successfully with the new init command.

cc @grabbou @thymikee


template/index.js
```diff
@@ -4,6 +4,5 @@
 
 import {AppRegistry} from 'react-native';
 import App from './App';
-import {name as appName} from './app.json';
 
-AppRegistry.registerComponent(appName, () => App);
+AppRegistry.registerComponent('HelloWorld', () => App);
```
thymikee: you could take `placeholderName` from `template.config.js`

thymikee: scratch that, forgot we aim for easy replacement of template name and don't want to tie template to the config anyhow

template/_watchmanconfig
```diff
@@ -1 +0,0 @@
-{}
```
cpojer: Can you bring this file back please?

template/index.js
cpojer: Can you elaborate? I don't think we should hardcode `HelloWorld` anywhere.

cpojer:
I don't think having dotfiles like this will work within FB. We have various tools that run based on the existence of, for example, a `.flowconfig` file. Landing this in the internal repo will try to typecheck the template and then fail. Can we keep the `_` prefix?

cpojer:
Also, is there a reason we are moving away from `app.json`? I just wanna make sure this is all completely thought through as we need to make sure this keeps working when we upgrade. Also cc @grabbou – wasn't the new auto linking feature going to make use of `app.json` in some form?

template/index.js
Esemesek: This is a placeholder, that will be globally replaced in the initialization process. This placeholder is used everywhere in the template: ios, android etc.

Esemesek:
@cpojer can you check if it will work with dotfiles? One of my main goals in this refactor was getting rid of this underscores prefixes.

template/_watchmanconfig
Esemesek: Sure C: Removed it by mistake :P

grabbou: Why do we need an empty watchman config in a regular React Native app?

grabbou:
> I don't think having dotfiles like this will work within FB. We have various tools that run based on the existence of, for example, a .flowconfig file. Landing this in the internal repo will try to typecheck the template and then fail. Can we keep the _ prefix?

One benefit is just being able to open the template as a root and work with these dot files to make sure everything passes (e.g. Flow works or Jest works). Not a deal breaker, but would be a good improvement.

> Also, is there a reason we are moving away from app.json? I just wanna make sure this is all completely thought through as we need to make sure this keeps working when we upgrade. Also cc @grabbou – wasn't the new auto linking feature going to make use of app.json in some form?

Not at all. We are going to use our own configuration. `app.json` was created in this commit https://github.com/facebook/react-native/commit/bc74dd1e14372ce2ff7c6d21401e67c946eb7595#diff-096cfc328f3353a1392c2094d8f3f604 and according to the description, it's only needed to run `react-native eject` (when you want to generate native files). I am not sure why we run it for the template, but you don't need it post `eject` phase, which is exactly where `react-native init` experience starts.

I would remove it.

template/index.js
cpojer: Got it!

template/_watchmanconfig
cpojer: If there is no watchmanconfig file like this, watchman will not be enabled for the project. We want it to be enabled by default if watchman is installed.

thymikee: Could you confirm this with somebody from Watchman team at FB? We don't need that file in projects that use Jest and it uses watchman without issues

cpojer: Good point, it may actually be a thing that we only need at FB given our settings here. In that case, I'm alright with you removing it.

cpojer:
Seems like we may be able to land this after all, with a few internal changes. Let me know once it is ready.

Esemesek:
Let me explain why there are so many changes:
To be backward compatible with all the packages that have custom templates and for people using global `react-native-cli` we will have 2 templates: old one and the new one. I've moved files to the `new` template, so to create `old` one I had to copy and paste old files.

package.json
```diff
@@ -50,7 +50,8 @@
     "README.md",
     "third-party-podspecs",
     "template",
-    "local-cli"
+    "local-cli",
+    "template.config.js"
```
thymikee: can you confirm with `npm pack` that `.gitignore` from `template/`is published as well?

gengjiawen:
I am -1 on this solution, maintaining two set of folder is too much work.

Esemesek:
@gengjiawen We will have one template. I am working on a solution in `cli` to have full backward compatibility. No worries there.

gengjiawen:
> I don't think having dotfiles like this will work within FB. We have various tools that run based on the existence of, for example, a `.flowconfig` file. Landing this in the internal repo will try to typecheck the template and then fail. Can we keep the `_` prefix?

This one maybe need to cope with more caution. Filename starts with `.` has pitfalls. They maybe not copied (maybe the original one change it to `_` due to this).
cc @grabbou

Esemesek:
@cpojer forgot about `package.json` which is necessary for new flow to work.

cpojer:
Note I removed the `yarn.lock` file, we don't want that in the template here.

PR #24152, opened 2019-03-26
[Fabric] Fixes dead-lock when enumerate surface

zhongwuzw:
## Summary

Import read-write lock to allow multiple read.

## Changelog

[iOS] [Fixed] - Fixes dead-lock when enumerate surface

## Test Plan

N/A.


shergin:
We are landing that manually right now.
@zhongwuzw Thank you so much!

sahrens:
Landed as https://github.com/facebook/react-native/commit/366fa2f1ef442bf30da17ec82131a59bb1b7906b

Issue #24139, opened 2019-03-25
[iOS] RCT_EXTERN_MODULE Swift modules broken in Xcode 10.2


kdawgwilk:
## 🐛 Bug Report

It appears Obj-C runtime got more strict with Swift ABI stability in Xcode 10.2 and now causes a crash at runtime when trying to load modules that have Swift implementations.  This is the crash error:

```
Swift class extensions and categories on Swift classes are not allowed to have +load methods
```

## To Reproduce

Create a RN NativeModule e.g.

`BadgeHandler.m`
```objc
#import <React/RCTBridgeModule.h>

@interface RCT_EXTERN_MODULE(BadgeHandler, NSObject)
    RCT_EXTERN_METHOD(setBadgeNumber:(int *)badgeNumber)
@end
```

`BadgeHandler.swift`
```swift
import UIKit

@objc(BadgeHandler)
class BadgeHandler: NSObject, RCTBridgeModule {
    static func moduleName() -> String! {
        return "BadgeHandler";
    }

    static func requiresMainQueueSetup() -> Bool {
        return true
    }
    
    @objc func setBadgeNumber(_ badgeNumber: Int) {
        DispatchQueue.main.async {
            UIApplication.shared.applicationIconBadgeNumber = badgeNumber
        }
    }
}
```
`AppName-Bridging-Header.h`
```objc
#ifndef AppName_Bridging_Header_h
#define AppName_Bridging_Header_h

#import <React/RCTBridgeModule.h>

#endif /* AppName_Bridging_Header_h */
```

## Expected Behavior

The NativeModule would load fine at runtime like it did in previous versions of Xcode

## Code Example

See above ^

## Environment

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
      Memory: 31.36 MB / 32.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 11.6.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.5.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 25, 26, 27, 28
[...]

PR #24155, opened 2019-03-26
Use constructor attribute instead of +load objc method

zienag:
## Summary

Xcode 10.2 forbids creating categories for swift class that uses `+load` method. In react-native categories like this are used to register swift classes as modules (macro `RCT_EXTERN_MODULE`) This PR changes it to use `__attribute__((constructor))` instead of objc `+load` method. 

I introduced new macro for this purpose, `RCT_EXPORT_MODULE_NO_LOAD`, it expands in something like:
```
void RCTRegisterModule(Class);

+ (NSString *)moduleName { 
  return @"jsNameFoo"; 
}

__attribute__((constructor)) static void initialize_ObjcClassFoo{ 
  RCTRegisterModule([ObjcClassFoo class]); 
}
```

Functions marked with `__attribute__((constructor))` are run before main and after all `+load` methods, so it seems like correct thing to do.

Fixes https://github.com/facebook/react-native/issues/24139
Doc about loading order https://developer.apple.com/documentation/objectivec/nsobject/1418815-load?language=objc

## Changelog

[iOS] [Fixed] - Fix runtime crash in xcode 10.2 when using RCT_EXTERN_MODULE for swift classes.

## Test Plan

Builds and runs successfully.

### DIFF
--- a/React/Base/RCTBridgeModule.h
+++ b/React/Base/RCTBridgeModule.h
@@ -74,6 +74,17 @@ RCT_EXTERN void RCTRegisterModule(Class); \
 + (void)load { RCTRegisterModule(self); }
 
 /**
+ * Same as RCT_EXPORT_MODULE, but uses __attribute__((constructor)) for module
+ * registration. Useful for registering swift classes that forbids use of load
+ * Used in RCT_EXTERN_REMAP_MODULE
+ */
+#define RCT_EXPORT_MODULE_NO_LOAD(js_name, objc_name) \
+RCT_EXTERN void RCTRegisterModule(Class); \
++ (NSString *)moduleName { return @#js_name; } \
+__attribute__((constructor)) static void \
+RCT_CONCAT(initialize_, objc_name)() { RCTRegisterModule([objc_name class]); }
+
+/**
  * To improve startup performance users may want to generate their module lists
  * at build time and hook the delegate to merge with the runtime list. This
  * macro takes the place of the above for those cases by omitting the +load
@@ -250,7 +261,7 @@ RCT_EXTERN void RCTRegisterModule(Class); \
   @interface objc_name (RCTExternModule) <RCTBridgeModule> \
   @end \
   @implementation objc_name (RCTExternModule) \
-  RCT_EXPORT_MODULE(js_name)
+  RCT_EXPORT_MODULE_NO_LOAD(js_name, objc_name)
 
 /**
  * Use this macro in accordance with RCT_EXTERN_MODULE to export methods


shergin:
Should we change all registration to use that?

LGTM, but I think Ramanpreet @RSNara should bless it.

zienag:
> Should we change all registration to use that?
New style requires `objc_name`, so it can be not so convenient as old style, and will require a lot of changes from users.

brenwell:
This fix worked well for me. Thanks!

zienag:
@shergin @RSNara test_android failed, and I don't think it is because of my changes, what can I do about it?

radex:
Is there a reason why +load must be used in some cases? Would it be better to simply change the old method to the new method? It would be best not to force changes on a lot of not just apps but library authors just to support new version of Xcode (while breaking compatibility with old version of React Native).

I get that since you can't do `RCTRegisterModule(self)`, you need objc_name, but in most cases it's the same as js_name. If it's not, we already have RCT_EXTERN_REMAP_MODULE

zienag:
> Is there a reason why +load must be used in some cases? Would it be better to simply change the old method to the new method? It would be best not to force changes on a lot of not just apps but library authors just to support new version of Xcode (while breaking compatibility with old version of React Native).
> 
> I get that since you can't do `RCTRegisterModule(self)`, you need objc_name, but in most cases it's the same as js_name. If it's not, we already have RCT_EXTERN_REMAP_MODULE

Change that I'm introducing is non-destructive, that means that no code will be broken. Changing the way `RCT_EXPORT_MODULE` works by adding the `objc_name` argument or assuming that it will be equal to `js_name` is destructive. There are dozens of places even in react-native code itself that calls `RCT_EXPORT_MODULE` without arguments (for example `RCTDataRequestHandler`, `RCTHTTPRequestHandler`), that means it will broken and should be fixed. In fact there are __much more__ uses of `RCT_EXPORT_MODULE` without arguments than with specified `js_name`. Also there are many places that `js_name` and `objc_name` are different (`RCTPlatform.m`).

So, as I'm seeing this, changing `RCT_EXPORT_MODULE` to use constructor attributed c functions instead of objc's +load is far more debatable thing to do. And as I'm not very familiar with react native, I'm not competent enough to make this change :)

radex:
> Also there are many places that js_name and objc_name are different

Alright, that convinces me 👍

renzhezizhi:
I found a similar [issue](https://github.com/ionic-team/capacitor/issues/1262) in one of the iconic-team repo. They just deleted all "load" methods in their code to avoid this problem.
I searched inside the react-native folder and found THREE files with +(void)load method for ios. They are:
React/Base/RCTBridgeModule.h
Libraries/WebSocket/RCTReconnectingWebSocket.m
RNTester/Swizzle_RCTFabricSurface.m

Hope it helps.

shergin:
@zienag I will land it on Monday.

benevbright:
Hm, actually `NativeModules.UserEventEmitter` becomes `undefined`.

benevbright:
No, my bad. Sorry guys.
It's working fine.

PR #24173, opened 2019-03-27
[Tests] Avoid failing Obj-C tests when xcpretty is not installed

hramos:
## Summary

Fixes an issue where `scripts/objc-test-ios.sh` would fail if `xcpretty` is not installed. As this tool is not bundled with macOS, and it's not explicitly called out in our docs on testing, the script should not fail if it's not present.

In a related change, JUnit reports are written to a more sensible location when the script is run outside of Circle CI.

## Changelog

[iOS] [Fixed] - Fixed test script failure when xcpretty is not present

## Test Plan

Ran the script prior to installing xcpretty, and confirmed tests run to completion without issues.

### DIFF
--- a/scripts/objc-test-ios.sh
+++ b/scripts/objc-test-ios.sh
@@ -11,7 +11,7 @@
 # also run the RNTester integration test (needs JS and packager):
 # ./objc-test-ios.sh test
 
-set -ex
+set -e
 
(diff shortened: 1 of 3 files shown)


hramos:
Will go ahead and merge this.

PR #24174, opened 2019-03-27
Fix unused var warning

hramos:
## Summary

Trivial. Removes a unused variable warning.

## Changelog

[iOS] [Fixed] - Fix JSCRuntime unused variable warning.

## Test Plan

Ran xcodebuild and confirm warning went away.

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -214,11 +214,11 @@ class JSCRuntime : public jsi::Runtime {
   static JSStringRef stringRef(const jsi::String& str);
   static JSStringRef stringRef(const jsi::PropNameID& sym);
   static JSObjectRef objectRef(const jsi::Object& obj);
-    
+
 #ifdef RN_FABRIC_ENABLED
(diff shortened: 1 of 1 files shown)


ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -928,7 +928,7 @@ jsi::WeakObject JSCRuntime::createWeakObject(const jsi::Object& obj) {
 #endif
 }
 
-jsi::Value JSCRuntime::lockWeakObject(const jsi::WeakObject& obj) {
+jsi::Value JSCRuntime::lockWeakObject(__unused const jsi::WeakObject& obj) {
```
zhongwuzw: @hramos Maybe we need to postpone to fix this warning, because `__unused` is the macro provided by Apple clang, but `JSCRuntime` is the common component run on all platforms, please see https://github.com/facebook/react-native/pull/23705, we discussed the details at there.

hramos: Thanks for bringing that up! In light of that, I'll go ahead and close this PR.

PR #24190, opened 2019-03-28
Extended accessibility actions support

marcmulcahy:
## Summary

Extend accessibility actions to include both  name (non-localized) and label (localized). this lets components provide programmatic access to custom actions, as well as standard actions such as slider adjustment and scrolling. This also means that when applications process actions, they need not compare against a localized name, which will be different for every language.

## Changelog

[general] [change] - Extend accessibility actions support.

## Test Plan

We have added tests in RNTester which exercise both custom actions, as well as standard actions.

### DIFF
--- a/Libraries/Components/View/ViewAccessibility.js
+++ b/Libraries/Components/View/ViewAccessibility.js
@@ -10,6 +10,8 @@
 
 'use strict';
 
+import type {SyntheticEvent} from 'CoreEventTypes';
+
 // This must be kept in sync with the AccessibilityRolesMask in RCTViewManager.m
 export type AccessibilityRole =
   | 'none'
@@ -50,4 +52,17 @@ export type AccessibilityStates = $ReadOnlyArray<
   | 'expanded'
   | 'collapsed'
   | 'hasPopup',
+>;
+
+// the info associated with an accessibility action
+export type AccessibilityActionInfo = $ReadOnly<{
+  name: string,
+  label?: string,
+}>;
+
+// The info included in the event sent to onAccessibilityAction
+export type AccessibilityActionEvent = SyntheticEvent<
+  $ReadOnly<{
+    actionName: string,
+  }>,
 >;
--- a/RNTester/js/AccessibilityExample.js
+++ b/RNTester/js/AccessibilityExample.js
@@ -410,6 +410,72 @@ class AccessibilityRoleAndStateExample extends React.Component<{}> {
   }
 }
 
+class AccessibilityActionsExample extends React.Component {
+  render() {
+    return (
+      <View>
+        <RNTesterBlock title="Non-touchable with activate action">
+          <View
+            accessible={true}
+            accessibilityActions={[{name: 'activate'}]}
+            onAccessibilityAction={event => {
+              switch (event.nativeEvent.actionName) {
+                case 'activate':
+                  Alert.alert('Alert', 'View is clicked');
+                  break;
+              }
+            }}>
+            <Text>Click me</Text>
+          </View>
+        </RNTesterBlock>
+
+        <RNTesterBlock title="View with multiple actions">
+          <View
+            accessible={true}
+            accessibilityActions={[
+              {name: 'cut', label: 'cut label'},
+              {name: 'copy', label: 'copy label'},
+              {name: 'paste', label: 'paste label'},
+            ]}
+            onAccessibilityAction={event => {
+              switch (event.nativeEvent.actionName) {
+                case 'cut':
+                  Alert.alert('Alert', 'cut action success');
+                  break;
+                case 'copy':
+                  Alert.alert('Alert', 'copy action success');
+                  break;
+                case 'paste':
+                  Alert.alert('Alert', 'paste action success');
+                  break;
+              }
+            }}>
+            <Text>This view supports many actions.</Text>
+          </View>
+        </RNTesterBlock>
+
+        <RNTesterBlock title="Adjustable with increment/decrement actions">
+          <View
+            accessible={true}
+            accessibilityRole="adjustable"
+            accessibilityActions={[{name: 'increment'}, {name: 'decrement'}]}
+            onAccessibilityAction={event => {
+              switch (event.nativeEvent.actionName) {
+                case 'increment':
+                  Alert.alert('Alert', 'increment action success');
+                  break;
+                case 'decrement':
+                  Alert.alert('Alert', 'decrement action success');
+                  break;
+              }
+            }}>
+            <Text>Slider</Text>
+          </View>
+        </RNTesterBlock>
+        </View>
+    );
+  }
+}
 class ScreenReaderStatusExample extends React.Component<{}> {
   state = {
     screenReaderEnabled: false,
@@ -465,6 +531,12 @@ exports.examples = [
     title: 'New accessibility roles and states',
     render(): React.Element<typeof AccessibilityRoleAndStateExamples> {
       return <AccessibilityRoleAndStateExample />;
+    },
+  },
+  {
+    title: 'Accessibility action examples',
+    render(): React.Element<typeof AccessibilityActionsExample> {
+      return <AccessibilityActionsExample />;
     },
   },
   {
--- a/RNTester/js/AccessibilityExample.js
+++ b/RNTester/js/AccessibilityExample.js
@@ -472,7 +472,7 @@ class AccessibilityActionsExample extends React.Component {
             <Text>Slider</Text>
           </View>
         </RNTesterBlock>
-        </View>
+      </View>
     );
   }
 }
--- a/React/Views/RCTView.h
+++ b/React/Views/RCTView.h
@@ -31,9 +31,10 @@ extern const UIAccessibilityTraits SwitchAccessibilityTrait;
 /**
  * Accessibility properties
  */
-@property (nonatomic, copy) NSArray <NSString *> *accessibilityActions;
 @property (nonatomic, copy) NSString *accessibilityRole;
 @property (nonatomic, copy) NSArray <NSString *> *accessibilityStates;
+@property (nonatomic, copy) NSArray <NSDictionary *> *accessibilityActions;
+@property (nonatomic, copy) NSDictionary *accessibilityActionsMap;
 
 /**
  * Used to control how touch events are processed.
--- a/React/Views/RCTView.h
+++ b/React/Views/RCTView.h
@@ -34,7 +34,6 @@ extern const UIAccessibilityTraits SwitchAccessibilityTrait;
 @property (nonatomic, copy) NSString *accessibilityRole;
 @property (nonatomic, copy) NSArray <NSString *> *accessibilityStates;
 @property (nonatomic, copy) NSArray <NSDictionary *> *accessibilityActions;
-@property (nonatomic, copy) NSDictionary *accessibilityActionsMap;
 
 /**
  * Used to control how touch events are processed.
--- a/React/Views/RCTView.m
+++ b/React/Views/RCTView.m
@@ -101,8 +101,8 @@ static NSString *RCTRecursiveAccessibilityLabel(UIView *view)
 @implementation RCTView
 {
   UIColor *_backgroundColor;
-  NSMutableDictionary *accessibilityActionsNameMap;
-  NSMutableDictionary *accessibilityActionsLabelMap;
+  NSMutableDictionary<NSString *, NSDictionary *> *accessibilityActionsNameMap;
+  NSMutableDictionary<NSString *, NSDictionary *> *accessibilityActionsLabelMap;
 }
 
 - (instancetype)initWithFrame:(CGRect)frame
@@ -160,10 +160,9 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
 
 -(void)setAccessibilityActions:(NSArray *)actions
 {
-  if (!actions || !actions.count) {
+  if (!actions) {
     return;
   }
-  _accessibilityActions = [NSMutableArray array];
   accessibilityActionsNameMap = [[NSMutableDictionary alloc] init];
   accessibilityActionsLabelMap = [[NSMutableDictionary alloc] init];
   for (NSDictionary *action in actions) {
@@ -179,7 +178,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
 
 - (NSArray <UIAccessibilityCustomAction *> *)accessibilityCustomActions
 {
-  if (!_accessibilityActions || !_accessibilityActions.count) {
+  if (!_accessibilityActions.count) {
     return nil;
   }
 
@@ -375,8 +374,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
   else if (_onAccessibilityTap) {
     _onAccessibilityTap(nil);
     return YES;
-  }
-       else {
+  } else {
     return NO;
        }
 }
@@ -385,12 +383,10 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
 {
   if ([self performAccessibilityAction:@"magicTap"]) {
     return YES;
-  }
-  else if (_onMagicTap) {
+  } else if (_onMagicTap) {
     _onMagicTap(nil);
     return YES;
-  }
-  else {
+  } else {
     return NO;
   }
 }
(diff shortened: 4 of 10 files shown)


marcmulcahy:
One of the challenges with the current design is the inability for action callback functions to return success or failure. So the implementation assumes that actions always succeed.

This generally works for most types of custom actions, but is problematic for actions like slider adjustment which might not always work.

Instead of specifying actions as a list, we could consider  adding API to component to allow actions to be added and removed dynamically. This would allow components to remove actions that aren't expected to work ("increment" when slider is at max, "cut" when no text on clipboard, etc.).

Short of that, is there a way for JavaScript code to return a value to native code?

ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
```diff
@@ -28,6 +38,17 @@
 
 public class AccessibilityDelegateUtil {
 
+  private static int sCounter = 0x3f000000;
+
+  public static final HashMap<String, Integer> sActionIdMap= new HashMap<>();
+  static {
+      sActionIdMap.put("activate", AccessibilityActionCompat.ACTION_CLICK.getId());
+      sActionIdMap.put("increment", AccessibilityActionCompat.ACTION_SCROLL_FORWARD.getId());
+      sActionIdMap.put("decrement", AccessibilityActionCompat.ACTION_SCROLL_BACKWARD.getId());
```
blavalla: Is the idea here that these are common actions that you might want to override the label for? For example instead of saying "Double tap to [activate]" you'd set your own label on the activate action and have it say "Double tap to [custom label]"?  If so, I think AccessibilityActionCompat.ACTION_LONG_CLICK should definitely be included on this list as well.

marcmulcahy: Sort of. The purpose of actions isn't to customize the usage hint ("double tap to...").

Android conflates the two ideas together, in that in order to add a usage hint to a component, you add a description to it's click action. And that's how usage hints are hooked up in our proposed React Native implementation of usage hints on Android.

Adding long click to the list of supported standard actions might be a good idea, but there wouldn't be any way to have that action force TalkBack to say something.

The other complication is that because of the way Android implements usage hints, all usage hints start with "double tap to...". On iOS, usage hints are simply strings, so can start with anything.

I suspect adding a "longclick" standard action name makes sense-- I'll look to see what the iOS equivalent would be.

React/Views/RCTView.h
```diff
@@ -29,7 +29,8 @@
 /**
  * Accessibility properties
  */
-@property (nonatomic, copy) NSArray <NSString *> *accessibilityActions;
+@property (nonatomic, copy) NSArray <NSDictionary *> *accessibilityActions;
+@property (nonatomic, copy) NSDictionary *accessibilityActionsMap;
```
ikenwoo: can we add the type information for the NSDictionary

React/Views/RCTView.m
```diff
@@ -99,6 +99,8 @@ - (UIView *)react_findClipView
 @implementation RCTView
 {
   UIColor *_backgroundColor;
+  NSMutableDictionary *accessibilityActionsNameMap;
+  NSMutableDictionary *accessibilityActionsLabelMap;
```
ikenwoo: annotate the type of the collections here as well

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
   return RCTRecursiveAccessibilityLabel(self);
 }
 
+-(void)setAccessibilityActions:(NSArray *)actions
+{
+  if (!actions || !actions.count) {
+    return;
```
ikenwoo: It doesn't look like the dictionaries get emptied out if someone tries to remove actions, e.g. `[view setAccessibilityActions:@[]]`

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
+  }
+  _accessibilityActions = [NSMutableArray array];
```
ikenwoo: the assignment here is not needed since you set this with a copy at L175.

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
+      accessibilityActionsLabelMap[action[@"label"]] = action;
+    }
+  }
+  _accessibilityActions = [actions copy];
+}
+
 - (NSArray <UIAccessibilityCustomAction *> *)accessibilityCustomActions
 {
-  if (!_accessibilityActions.count) {
+  if (!_accessibilityActions || !_accessibilityActions.count) {
```
ikenwoo: extra nil check isn't needed

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
 }
 
+-(void)setAccessibilityActions:(NSArray *)actions
+{
+  if (!actions || !actions.count) {
+    return;
+  }
+  _accessibilityActions = [NSMutableArray array];
+  accessibilityActionsNameMap = [[NSMutableDictionary alloc] init];
+  accessibilityActionsLabelMap = [[NSMutableDictionary alloc] init];
```
ikenwoo:
```suggestion
  accessibilityActionsLabelMap = [NSMutableDictionary dictionary];
```

let's go with this style since we're using `[NSMutableArray array]` consistently in this file.

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
   return RCTRecursiveAccessibilityLabel(self);
 }
 
+-(void)setAccessibilityActions:(NSArray *)actions
+{
+  if (!actions || !actions.count) {
+    return;
+  }
+  _accessibilityActions = [NSMutableArray array];
+  accessibilityActionsNameMap = [[NSMutableDictionary alloc] init];
```
ikenwoo:
```suggestion
  accessibilityActionsNameMap = [NSMutableDictionary dictionary];
```

Libraries/Components/View/ViewAccessibility.js
```diff
@@ -55,3 +57,16 @@ export type AccessibilityRole =
 
 // This must be kept in sync with the AccessibilityStatesMask in RCTViewManager.m
 export type AccessibilityStates = $ReadOnlyArray<'disabled' | 'selected'>;
+
+// the info associated with an accessibility action
+export type AccessibilityActionInfo = $ReadOnly<{
+  name: string,
```
ikenwoo: name sounds a little bit generic and quite similar to label which could cause some confusion. could we name this something like "type" or "actionIdentifier"?

also, could we give this a better type? it's pretty unclear that setting "name" here to "activate", "increment", "decrement", "escape" or "magicTap" has special behavior.

Libraries/Components/View/ViewAccessibility.js
```diff
@@ -55,3 +57,16 @@ export type AccessibilityRole =
...
+  label?: string,
+}>;
+
+// The info included in the event sent to onAccessibilityAction
+export type AccessibilityActionEvent = SyntheticEvent<
+  $ReadOnly<{
+    actionName: string,
```
ikenwoo: it would be great to align the "actionName" here with the "name" in `AccessibilityActionInfo`, so that it's clear these are the same thing.

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
+  if (!_accessibilityActions || !_accessibilityActions.count) {
     return nil;
   }
 
   NSMutableArray *actions = [NSMutableArray array];
-  for (NSString *action in _accessibilityActions) {
-    [actions addObject:[[UIAccessibilityCustomAction alloc] initWithName:action
-                                                                  target:self
-                                                                selector:@selector(didActivateAccessibilityCustomAction:)]];
+  for (NSDictionary *action in _accessibilityActions) {
```
ikenwoo: can we just loop through `accessibilityActionsLabelMap` here?

RNTester/js/AccessibilityExample.js
```diff
@@ -133,6 +133,73 @@ class AccessibilityExample extends React.Component {
...
+          </View>
+        </RNTesterBlock>
+
+        <RNTesterBlock title="View with multiple actions">
+          <View
+            accessible={true}
+            accessibilityActions={[
+              {name: 'cut', label: 'cut label'},
+              {name: 'copy', label: 'copy label'},
+              {name: 'paste', label: 'paste label'},
```
ikenwoo: I like this idea, but not a big fan of overloading the prop on the js side. It's a bit error prone, due to special 'name' definitions that don't appear outside the iOS client code, logic that could break if label is accidentally added/omitted when needed.

Specifically, from the js side, it's not clear what the behavior would be if someone defines one of the special case actions with a label accidentally or what if someone omits label from a custom action definition.

an example:
```
accessibilityActions={[
  {name: 'activate', label: 'tap'}, // this would add an extra custom action in addition to defining "accessibilityActivate" behavior on iOS
  {name: 'cut'}, // could we just assume that the label would also be `cut` here for their custom action? right now this would just be dropped completely on iOS without warning 
]}
```

it would be awesome if we could have a prop for the strongly typed events like "activate", "escape", etc, and a separate prop for custom actions specifically.

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
+    }
   }
 
   return [actions copy];
 }
 
 - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)action
 {
-  if (!_onAccessibilityAction) {
+  if (!_onAccessibilityAction || !accessibilityActionsLabelMap) {
```
ikenwoo:
```suggestion
  if (!accessibilityActionsLabelMap) {
```

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
     return NO;
   }
 
-  _onAccessibilityAction(@{
-    @"action": action.name,
-    @"target": self.reactTag
-  });
+  // iOS defines the name as the localized label, so use our map to convert this back to the non-localized action namne when passing to JS. This allows for standard action names across platforms.
```
ikenwoo:
```suggestion
  // iOS defines the name as the localized label, so use our map to convert this back to the non-localized action name when passing to JS. This allows for standard action names across platforms.
```
nit: spelling

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
   return NO;
 }
 
+- (BOOL)performAccessibilityAction:(NSString *) name
```
ikenwoo:
```suggestion
- (BOOL)performAccessibilityAction:(NSString *)name
```
nit

React/Views/RCTView.m
```diff
@@ -154,33 +156,58 @@ - (NSString *)accessibilityLabel
...
-    @"target": self.reactTag
-  });
+  // iOS defines the name as the localized label, so use our map to convert this back to the non-localized action namne when passing to JS. This allows for standard action names across platforms.
 
+  NSDictionary *actionObject = accessibilityActionsLabelMap[action.name];
+  if (actionObject) {
+   _onAccessibilityAction(@{
+    @"actionName": actionObject[@"name"],
+    @"actionTarget": self.reactTag
+  });
```
ikenwoo: we can call `performAccessibilityAction:` here instead to keep all the action logic at the same place

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
+}
+
 - (BOOL)accessibilityActivate
 {
-  if (_onAccessibilityTap) {
+  if ([self performAccessibilityAction:@"activate"]) {
+    return YES;
+    
+  }
+  else if (_onAccessibilityTap) {
```
ikenwoo: please keep the `else if` on the same line as the `}`

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
     _onAccessibilityTap(nil);
     return YES;
-  } else {
-    return NO;
   }
+       else {
```
ikenwoo: `} else {`

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
+       }
 }
 
 - (BOOL)accessibilityPerformMagicTap
 {
-  if (_onMagicTap) {
+  if ([self performAccessibilityAction:@"magicTap"]) {
+    return YES;
+  }
+  else if (_onMagicTap) {
```
ikenwoo: `} else if ...`

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
     _onMagicTap(nil);
     return YES;
-  } else {
+  }
+  else {
```
ikenwoo: `} else {`

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
-  if (_onAccessibilityEscape) {
+  if ([self performAccessibilityAction:@"escape"]) {
+    return YES;
+  }
+  else if (_onAccessibilityEscape) {
     _onAccessibilityEscape(nil);
     return YES;
-  } else {
+  }
+  else {
```
ikenwoo: `} else {`

React/Views/RCTView.m
```diff
@@ -264,36 +291,71 @@ - (BOOL)isAccessibilityElement
...
   }
 }
 
 - (BOOL)accessibilityPerformEscape
 {
-  if (_onAccessibilityEscape) {
+  if ([self performAccessibilityAction:@"escape"]) {
+    return YES;
+  }
+  else if (_onAccessibilityEscape) {
```
ikenwoo: `} else if ... {`

RNTester/js/AccessibilityExample.js
blavalla: I think keeping these as a single prop is nice, as it keeps all actions together, regardless if their implementation details on iOS/Android, especially considering that iOS and Android have different predefined actions. 

That being said, maybe we should have two different types allowed here.  One that is a "predefined action" such as "Activate", which doesn't let you set the label, and one that is a "custom action", which requires you to set both a name and a label.

blavalla:
> Short of that, is there a way for JavaScript code to return a value to native code?

@shergin, do you know if there is a way to do this?

shergin:
@marcmulcahy Totally unrelated to this diff: Feel free to implement the same for Fabric here: https://github.com/facebook/react-native/blob/master/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L522-L570

shergin:
> Instead of specifying actions as a list, we could consider adding API to component to allow actions to be added and removed dynamically. This would allow components to remove actions that aren't expected to work ("increment" when slider is at max, "cut" when no text on clipboard, etc.).


Do you mean some kinda imperative API (like `scrollTo(x,y)` or `.focus()`)? If so, IMHO, no. It's totally overkill. `Props` is a the most simple and idiomatic way to deal with it.

React/Views/RCTView.m
```diff
@@ -99,6 +99,8 @@ - (UIView *)react_findClipView
 @implementation RCTView
 {
   UIColor *_backgroundColor;
+  NSMutableDictionary<NSString *, NSDictionary *> *accessibilityActionsNameMap;
```
shergin: Not sure I understand why we need to have two maps here.

marcmulcahy:
This breaks double tap on Android 9.0. closing while I investigate and fix the issue, and will submit a new pull request once finished.

React/Views/RCTView.m
marcmulcahy: @shergin In this design, actions have a non-localized name and a localized label.

in iOS, custom accessibility actions only have a localized label, which is what is passed to didActivateAccessibilityCustomAction. The label map is used to find the action name to include with the event to JS.

For standard actions like increment, decrement, and activate, we get specific calls from iOS (such as didActivate). We use the name map to validate whether the JS component claims to support that standard action.

marcmulcahy:
@janicduplessis  Can you make sure I haven't  broken things again. I've moved actions back into RCTView, as we need to keep some class-specific maps around to match action names to labels and vice versa.

If you think UIView+React.* is the right place for actions, could you suggest how setAccessibilityActions (and the associated actionsN ameMNap and actionsLabelMap) could be factored out of RCTView?

PR #24195, opened 2019-03-29
[Fabric] Fix invalid CGContext when invalidate layer

zhongwuzw:
## Summary

If size is zero, the `CGContext` would invalid, so we need to filter zero things.
cc. @shergin 

## Changelog

[iOS] [Fixed] - Fix invalid CGContext when invalidate layer

## Test Plan

N/A

### DIFF
--- a/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
+++ b/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm
@@ -356,10 +356,12 @@ static RCTBorderStyle RCTBorderStyleFromBorderStyle(BorderStyle borderStyle)
 
 - (void)invalidateLayer
 {
+  CALayer *layer = self.layer;
+  if (CGSizeEqualToSize(layer.bounds.size, CGSizeZero)) {
+    return;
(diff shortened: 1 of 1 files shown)


shergin:
@zhongwuzw Will we have same problem if one of the sides (height or width) alone is zero? If so we should fix it as well.

Is this fixing a crash or incorrect rendering behavior?

zhongwuzw:
@shergin `CGContext` also would be `NULL` if one side of them is zero, `CGSizeZero` typically occurs when layout not applied, in our case, is `RCTUpdatePropsMountItem`, it would trigger redisplay,  and it applied before `RCTUpdateLayoutMetricsMountItem`. I don't know wether it's necessary to check one side zero case, only happens when user assign explicitly? we only check both zero case in RCTView either: https://github.com/facebook/react-native/blob/cd778873f0e7d79da5bf7b8204107702a87a4059/React/Views/RCTView.m#L608

> Is this fixing a crash or incorrect rendering behavior?
No crash, CG functions can accept NULL context, we add this check only just want to return quickly, don't need to do bunch of paint operations.

Issue #22186, opened 2018-11-06
Fix StrictMode warnings


brunolemos:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
  React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
      Memory: 33.30 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.4 - /usr/local/bin/node
      Yarn: 1.10.1 - ~/.yarn/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        Build Tools: 23.0.1, 23.0.3, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.2, 27.0.3, 28.0.0, 28.0.2, 28.0.3
        API Levels: 19, 22, 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5014246
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmGlobalPackages:
      react-native-cli: 2.0.1

## Description
When wrapping the app with `<StrictMode>`, there are warnings coming from internal react native components.

It seems to be the use of legacy context api on [VirtualizedList](https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js#L468)/[CellRenderer](https://github.com/facebook/react-native/blob/41eb2da33bafc450d5fcf03990d4dbeac5d5a5e6/Libraries/Lists/VirtualizedList.js#L1647) and [UNSAFE_componentWillReceiveProps](https://github.com/facebook/react-native/search?q=UNSAFE_componentWillReceiveProps&unscoped_q=UNSAFE_componentWillReceiveProps) on multiple other files.

<div>
[...]

PR #24214, opened 2019-03-30
Make ScrollView StrictMode compatible

Jyrno42:
## Summary

Converts ScrollView to use constructor instead of UNSAFE_componentWillMount and componentDidUpdate instead of UNSAFE_componentWillReceiveProps.

see https://github.com/facebook/react-native/issues/22186

## Changelog

[General] [Fixed] - Converted ScrollView to be compatible with StrictMode

## Test Plan

Tested manually via RNTester by opening StrictMode example and ensuring no strict-mode warnings are rendered. I also verified in a fresh project that stickyHeaderIndices still works.

### DIFF
--- a/Libraries/Components/ScrollResponder.js
+++ b/Libraries/Components/ScrollResponder.js
@@ -607,12 +607,11 @@ const ScrollResponderMixin = {
   },
 
   /**
-   * `componentWillMount` is the closest thing to a  standard "constructor" for
-   * React components.
+   * Constructor for this mixin, call from constructor or from deprecated UNSAFE_componentWillMount
    *
    * The `keyboardWillShow` is called before input focus.
    */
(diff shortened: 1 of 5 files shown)


Jyrno42:
Note: The ScrollView uses `ReactNative.findNodeHandle` in some places. I could not figure out how to remove those calls.

cpojer:
Similarly to the other PR, just moving lifecycle functionality to the constructor won't actually fix real issues. We'll need to change this component, possibly to use hooks. Would you be interested in reworking ScrollView with that in mind? Keep in mind that we have to be very careful and do a good amount of testing so we don't break this component.

Jyrno42:
Got it. I'll try to get animated component one done/merged first, and if anyone else has not grabbed scrollview conversion by then I'll take it as well.

Issue #24001, opened 2019-03-18
Long press broken for TouchableNativeFeedback


mjmasn:
## 🐛 Bug Report

After upgrading from React Native 0.58.6 to 0.59.1, any `<TouchableNativeFeedback />` with onLongPress is no longer working i.e. the handler is never fired.

## To Reproduce

Use TouchableNativeFeedback with onLongPress in a React Native 0.59.1 app.

## Expected Behavior

Long pressing the TouchableNativeFeedback triggers the onLongPress handler

## Code Example

https://github.com/mjmasn/rn59_tnf_longpress

## Environment

```
info 
  React Native Environment Info:
    System:
      OS: Linux 4.15 Ubuntu 16.04.6 LTS (Xenial Xerus)
      CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
      Memory: 856.03 MB / 15.54 GB
      Shell: 4.3.48 - /bin/bash
    Binaries:
      Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
      Yarn: 1.15.2 - /usr/bin/yarn
      npm: 5.6.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      Android SDK:
        API Levels: 16, 19, 22, 23, 26, 27, 28
        Build Tools: 23.0.1, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.3, 28.0.2, 28.0.3
        System Images: android-16 | Google APIs Intel x86 Atom, android-17 | Google APIs Intel x86 Atom, android-18 | Google APIs Intel x86 Atom, android-19 | Google APIs Intel x86 Atom, android-22 | Google APIs Intel x86 Atom, android-22 | Google APIs Intel x86 Atom_64, android-23 | Google APIs Intel x86 Atom, android-23 | Google APIs Intel x86 Atom_64, android-24 | Google APIs Intel x86 Atom, android-25 | Google Play Intel x86 Atom, android-26 | Google APIs Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom_64
    IDEs:
      Android Studio: 3.1 AI-173.4907809
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.1 => 0.59.1 
    npmGlobalPackages:
[...]

PR #24238, opened 2019-04-01
fixed touchable longpress

Kida007:
## Summary

This diff fixes a bug in TouchableNativeFeedback where a long press is not registered.

cause of the bug is _touchableHandleResponderMove_ being invoked **regardless** of a moving gesture ( even when movedDistance is 0) in some devices ( including OnePlus5t ), which was eventually clearing out  the long-press timeout.

## Changelog

[General] [Fixed] - Touchable onLongPress fix.


Kida007:
Fixes #24268

mjmasn:
Fixes #24001

PR #24264, opened 2019-04-02
add ram bundle to android template

gengjiawen:
## Summary

I plan to do this for a long time :)
Still a little surprised so much people not know it https://twitter.com/geekykaran/status/1112756026611257344

## Changelog

[Android] [Feature] - add ram bundle to android template 

## Test Plan

CI is green.

### DIFF
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -18,6 +18,9 @@ import com.android.build.OutputFile
  *   // the entry file for bundle generation
  *   entryFile: "index.android.js",
  *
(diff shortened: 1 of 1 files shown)


cpojer:
Thanks for this PR! We do not yet want to change the default and we'll probably want to make it opt-in soon. For now, let's keep it out of the template.

gengjiawen:
it's only in comment, user can uncomment it to enable it :)

cpojer:
Alright then, I'll land it but it may cause issues for people and they may need to do additional work to get it to run. It's also not useful without inline requires turned on.

PR #24276, opened 2019-04-03
Use node package dependency to manage JSC version

Kudo:
## Summary

In origin approach, we packed libjsc.so inside react-native.aar and it is difficult for user to choose different JSC variants. E.g., [the Intl supported version](https://github.com/react-native-community/jsc-android-buildscripts#international-variant).

This change list allows application to determine JSC versions or variants by npm/yarn package.

### To use Intl JSC
There is a |useIntlJsc| flag in build.gradle, it will use the same JSC version but with Intl support.

### To use latest JSC

`yarn add jsc-android@canary`

## Changelog

[Android] [Changed] - Allow application to select different JSC variants

**MIGRATION**
Note that there are some changes in build.gradle.
Existing application needs to change their android/build.gradle and android/app/build.gradle.
Hopefully, the rn-diff-purge should handle the case well.

## Test Plan

1. Test RNTester build and run successfully.
2. Create a new project by `react-native init Foo --version /somewhere/react-native-1000.0.0.tgz` and test that build and run successfully.
3. Turn on |useIntlJsc| and add `console.log('Intl type is ' + typeof Intl)` to App.js, verify if the adb log show "Intl type is object"
4. `yarn add jsc-android@canary` and rebuild. Verify adb log `adb logcat | grep 'JavaScriptCore.Version'` and it should return 241213.0.0 not 236355.1.0


ReactAndroid/build.gradle
```diff
@@ -255,8 +247,8 @@ task cleanReactNdkLib(type: Exec) {
 
 task packageReactNdkLibs(dependsOn: buildReactNdkLib, type: Copy) {
     from("$buildDir/react-ndk/all")
-    from("$thirdPartyNdkDir/jsc/jni")
     into("$buildDir/react-ndk/exported")
+    exclude("**/libjsc.so")
```
gengjiawen: Is this redundant ? By default gradle deps won't bundled in the aar.

Kudo: libjsc.so will be copied into `$buildDir/react-ndk/all` during ndkbuild.
If we don't exclude libjsc.so here, they will be copied to exported folder and packed in AAR.
This exclusion is pretty much same as what we did in old JSC, [e.g. 0.57](https://github.com/facebook/react-native/blob/0.57-stable/ReactAndroid/build.gradle#L239)

cpojer:
Update: just need to figure out the internal test system to actually land this but it should be good to go soon.

Kudo:
Thanks @cpojer for the update and revise wording for useIntlJsc comment.
Just a reminder that there is yet another same comment not revised in template/android/app/build.gradle.
Will you prefer to update that together?

ReactAndroid/build.gradle
```diff
@@ -133,21 +133,14 @@ task prepareGlog(dependsOn: dependenciesPath ? [] : [downloadGlog], type: Copy)
...
 
-task downloadJSC(dependsOn: createNativeDepsDirectories, type: Download) {
-    src("https://registry.npmjs.org/jsc-android/-/jsc-android-${JSC_VERSION}.tgz")
-    onlyIfNewer(true)
-    overwrite(false)
-    dest(new File(downloadsDir, "jsc-${JSC_VERSION}.tar.gz"))
-}
-
 // Create Android.mk library module based on jsc from npm
-task prepareJSC(dependsOn: downloadJSC) {
```
hey99xx: Bit confused about removing downloadJSC dependency.

What makes it appear jsc-android at `"$projectDir/../node_modules/jsc-android/dist"` location by the time this script runs? 

FYI I don't see build.gradle invoking npm at any point, so unless I'm missing it somewhere this may make builds without invoking `npm install` first always fail.

ReactAndroid/build.gradle
hey99xx: Can we also exclude libc++_shared.so too so that app packages don't have to do that `pickFirst` thing?

It's a dangerous thing in case there are other dependencies than RN that bring in libc++_shared.so as well.

ReactAndroid/build.gradle
cpojer: This is inside of `node_modules/react-native` when using React Native, so JSC should definitely be there.

cpojer:
@Kudo Do you want to send another PR for the remaining changes?

ReactAndroid/build.gradle
hey99xx: I meant when building RN from source, not when consuming it. While consuming none of ReactAndroid/build.gradle matters anyways.

The current instructions to build from source is simple download RN and run `./gradlew :ReactAndroid:installArchives` I am asking will this work?

cpojer: Good point. Is there a way we can invoke "yarn" from gradle just to make sure node modules are in sync?

hey99xx: A comment from when new JSC was checked in https://github.com/facebook/react-native/pull/22263#issuecomment-441437249

> Using npm is just not the right way to manage jsc as it's a native dependency. After fighting with it for a while I abandoned this approach. The correct approach (which will require some work in jsc-android-buildscripts) is to manage publications of jsc in maven. This way the user will be able to override / use other versions of jsc.

cpojer: cc @DanielZlotin

ReactAndroid/build.gradle
Kudo: That is a good point and I don't have a good solution right now.
The problem is that jsc-android-buildscripts packed AAR with a C++ runtime (libc++_shared.so).
[Old JSC excludes C++ runtime to AAR. ](https://github.com/facebook/android-jsc/blob/master/install.sh#L15-L18)
That's the difference why old JSC does not have to use pickFirst.

There is also an assumption that NDK maintains ABI compatibilities because RN and JSC may build from different NDK version, e.g. old JSC is built by NDK r10c.
Fortunately, we don't have problems from ABI incompatibilities during these years.

Possible solutions like:

1. Built JSC as static library. This will increase libjsc.so binary size and impact to application APK size.
2. Move jsc-android-buildscripts to RN core and build together. This is increase build time incredibly. (e.g. It takes about 2 hours to build a jsc-android-buildscripts job on CicleCI.)

IMO, excluding libc++_shared.so from jsc-android-buildscripts is better than pickFirst as pickFirst may impact to application level.
But this needs changes from jsc-android-buildscripts side.

ReactAndroid/build.gradle
Kudo: From my point of view, RN itself is tightly coupled with node_modules, i.e. you cannot simply get a AAR or APK to run without metro or node_modules stuffs.

I also prefer to maintain even native dependencies in single source of truth, the pakcage.json.
Otherwise, we currently have multiple place to declare dependency version.
E.g. folly version 2018.10.22.00 defines in multiple places like gradle, CocoaPods, and Xcodeproj.

Invoking yarn by gradle is good to me.
Or at least we could check if node_modules is not there and prompt user to run yarn manually.

Kudo:
@cpojer Sure, will have a PR after we have clarify which solution is better.

sahrens:
I think this broke the docker build because downloadJSC is still refrenced [here](https://github.com/facebook/react-native/blob/master/.circleci/Dockerfiles/Dockerfile.android#L51). cc @hramos
```

* What went wrong:
Task 'downloadJSC' not found in project ':ReactAndroid'.
```

ReactAndroid/build.gradle
sahrens: Note this also broke the docker build `yarn test-android-unit`. I tried changing the `Dockerfile.android` to `RUN yarn` before `./gradlew :ReactAndroid:packageReactNdkLibsForBuck` (and removed the dead reference to `downloadJSC`) but that didn't work, it still failed with
```
> Expected directory '/app/ReactAndroid/../node_modules/jsc-android/dist' to contain exactly one file, however, it contains no files.
```

hramos:
Fixed in https://github.com/facebook/react-native/pull/24360

Kudo:
Oops, my fault to miss this in Dockerfile. Thank you @sahrens @hramos

PR #24312, opened 2019-04-04
[General] [Fixed] - VirtualizedSectionList/SectionList: replace enableVirtualization prop annotation by correct underlying disableVirtualisation of VirtualizedList

MoOx:
## Summary

It seems (I used git history to confirm) that FlatList/VirtualizedList have ([since the begining](https://github.com/facebook/react-native/blame/c13f5d48cfe3e7c0f6c6d0b745b50a089d6993ef/Libraries/Lists/VirtualizedList.js#L79)) a `disableVirtualization` prop.
SectionList ([since it's begining](https://github.com/facebook/react-native/blame/abe737fe746406533798f9670e8e243cb18d5634/Libraries/Lists/VirtualizedSectionList.js#L98)) have a `enableVirtualization` prop, but since SectionList is VirtualizedSectionList which use VirtualizedList, this prop probably never did something. This fix just rename the prop properly so it can have an effect on the underlying VirtualizedList when you use a SectionList.

Since props are spread it's kind of working already, but the flow annotation are wrong (so it tells you it won't work/ you can't use it) which sucks.

(NB: I am doing this since I was trying to use a SectionList with react-native-web & server side rendering to get the all list, you can laugh).

## Changelog

[General] [Fixed] - VirtualizedSectionList/SectionList: replace enableVirtualization prop annotation by correct underlying disableVirtualisation of VirtualizedList

## Test Plan

I didn't really tested this to be honest, but I think the code speak by itself.

Here are also others proofs: 
- [disableVirtualization is being used](https://github.com/facebook/react-native/search?q=disableVirtualization&unscoped_q=disableVirtualization)
- [enableVirtualization isn't!](https://github.com/facebook/react-native/search?q=enableVirtualization&unscoped_q=enableVirtualization)

### DIFF
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -705,7 +705,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   };
 
   _isVirtualizationDisabled(): boolean {
-    return this.props.disableVirtualization;
+    return this.props.disableVirtualization || false;
   }
 
   _isNestedWithSameOrientation(): boolean {
--- a/Libraries/Lists/VirtualizedList.js
+++ b/Libraries/Lists/VirtualizedList.js
@@ -76,7 +76,7 @@ type OptionalProps = {
    * unmounts react instances that are outside of the render window. You should only need to disable
    * this for debugging purposes.
    */
-  disableVirtualization: boolean,
+  disableVirtualization?: ?boolean,
   /**
    * A marker property for telling the list to re-render (since it implements `PureComponent`). If
    * any of your `renderItem`, Header, Footer, etc. functions depend on anything outside of the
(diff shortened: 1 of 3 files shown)


MoOx:
Btw, this change show (see flow result) that `disableVirtualization` isn't optional in VirtualizedList...?

```
Libraries/Lists/VirtualizedList.js:79:26
    79|   disableVirtualization: boolean,
                                 ^^^^^^^ [2]
```

Flow detect defaultProps right? So should be set too as optional. Do you want me to adjust that too?

cpojer:
Oh yeah, seems like Flow is failing. Could you make Flow happy without any actual API changes?

MoOx:
I think so. disableVirtualization is already specified as optional in the doc https://facebook.github.io/react-native/docs/virtualizedlist#disablevirtualization so I will just have to adjust annotation accordingly.

MoOx:
ios & js are failing because of...

```
Error: Command failed: yarn add react@16.8.3
```

No idea what's happening.

Android is failing because of a end2end test. Not sure what I can do about it...

cpojer:
Yeah something is messed up about open source CI, let's see if it passes Flow at FB and lands :)

Issue #24316, opened 2019-04-04
[standards] Umbrella issue to track migration to standard path-based requires (remove Haste)


ide:
This issue tracks commits that rewrite `require()` calls to use standard module paths instead of Haste module names. This brings the RN repo closer to how the React repo is set up, and also opens the door for greater compatibility with the last 10 years of tooling that uses path-based module locations, and eliminates the entire class of issues related to the Haste map outside of the Facebook context.

All modules in the RN repository will still be requireable via Haste. That is, if there's Facebook application code that uses Haste, there will be no change.

This is a spiritual continuation of https://github.com/facebook/react-native/issues/18598, which has more technical details. 

- [x] Let Flow support Haste & Node simultaneously for now (pre-req): ~https://github.com/facebook/react-native/pull/24318~
- [x] RNTester: ~https://github.com/facebook/react-native/pull/24317~
- [x] RNTester (TurboModules): ~https://github.com/facebook/react-native/pull/24754~
- [x] Code under the Libraries directory: ~https://github.com/facebook/react-native/pull/24749~, ~https://github.com/facebook/react-native/pull/25001~
- [x] IntegrationTests: ~https://github.com/facebook/react-native/pull/24750~
- [x] Jest setup script: ~https://github.com/facebook/react-native/pull/24753~
- [x] androidTest JS: ~https://github.com/facebook/react-native/pull/24813~
- [x] Vendored libraries (~https://github.com/facebook/react-native/issues/24769~): ~https://github.com/facebook/react-native/pull/24807~
- [x] React renderers (~https://github.com/facebook/react-native/issues/24770~): ~https://github.com/facebook/react-native/pull/24782~, ~https://github.com/facebook/react/pull/15604~, ~https://github.com/facebook/react-native/pull/25100~, ~https://github.com/facebook/react/pull/15786~
- [x] react-native-implementation (~https://github.com/facebook/react-native/issues/24771~)
- [x] Lint rule to prevent more Haste: ~https://github.com/facebook/react-native/pull/25058~
[...]

PR #24317, opened 2019-04-04
[standards] Rewrite imports in RNTester to use standard paths

ide:
## Summary

See https://github.com/facebook/react-native/issues/24316 for the motivation. This commit rewrites the imports in the RNTester project.

## Changelog

[General] [Changed] - Replaced Haste-style imports with standard path-style imports for RNTester

## Test Plan

Run RNTester and verify that it loads without any issues. Tap through each screen to ensure the JS modules load (that is, were bundled) correctly.


cpojer:
Instead of the relative paths, what do you think about doing `require('react-native/Libraries/…')`?

ide:
- Updated requires to use the public interface where possible (`const {View} = require('react-native')`)
- Changed `const ReactNative = require('react-native'); const {...} = ReactNative;` to `const {...} = require('react-native');`, which sets us up to move to import/export more easily
- I left requires of internal modules (utilities, for example) alone, since React Native may want to seal them off entirely outside of the Facebook repo and not encourage `require('react-native/Libraries/Utilities/...')`. React in particular publishes itself pre-bundled to intentionally separate its public interface from its internals and prevent `require('react/...')`.

PR #24318, opened 2019-04-05
[standards][flow] Set up .flowconfig to support Haste and path-based imports

ide:
## Summary

See https://github.com/facebook/react-native/issues/24316 for the motivation.

By adding the .android.js and .ios.js extensions to the respective .flowconfig files, Flow is able to find files that either:

- Are required using Haste
- Are required using standard paths and have a .js extension
- Are required using standard paths and have a .platform.js subextension

## Changelog

[General] [Changed] - Adjusted .flowconfig to support both Haste and standard path-based requires

## Test Plan

After replacing some Haste-style requires with path-based requires, this change to the Flow config made Flow pass.


ide:
> I may need to make some internal changes.

Hopefully it is straightforward -- the concept here is that there were several files (namely those ending in `.platform.js`) that Flow was able to find only via Haste names. This commit teaches Flow about `.platform.js`, and when you specify `module.file_ext`, you override the defaults and need to re-specify all the extensions you use.

PR #24321, opened 2019-04-05
WIP: Add hooks variant of createAnimatedComponent 

Jyrno42:
## Summary

This is a continuation of https://github.com/facebook/react-native/pull/24218

As mentioned in that PR by maintainers, hooks based variant is preferred to make animated components strict-mode compatible. This PR adds that.

Note: The old variant still exists in the code and can be used by providing false as the second argument to createAnimatedComponent. 

related to https://github.com/facebook/react-native/issues/22186

Note: Will squash/cleanup history at the end

## Changelog

[General] [Fixed] - Converted createAnimatedComponent to be compatible with StrictMode. Old behaviour can be retained by providing false as the second argument.

## Test Plan

Tested all animation examples in RNTester to make sure they still work. Created a component that used refs to access `getNode` and it worked the same for both the old and new implementation. Finally I also ran some real-world apps that I have built at work with the code and did not notice any regressions.

### DIFF
--- a/Libraries/Animated/src/AnimatedOnAFeeling.js
+++ b/Libraries/Animated/src/AnimatedOnAFeeling.js
@@ -1,0 +1,190 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+'use strict';
+
+const {AnimatedEvent} = require('./AnimatedEvent');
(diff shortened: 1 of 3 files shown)


Jyrno42:
Going to be fixing the tests tomorrow.

cpojer:
Hey @Jyrno42, what's the status of this PR?

cpojer:
It seems like this PR got stale so I'm going to close it. I'm really curious to see whether we can build a better API for Animated though, or whether we should start recommending reanimated to people.

Jyrno42:
OK. Will look around for that before I start working on this again (once I have more time available).

PR #24323, opened 2019-04-05
Add ability to customise WebSocket OkHttp client and fallback to existing implementation

### LINKED ISSUE #18920: WebSocket https connection with self signed certificate on Android

jakoblorz:
## Summary
This is a follow-up PR to #19632 for which @cpojer [called upon](https://github.com/facebook/react-native/pull/19632#issuecomment-476693393). Disclaimer: I don't have much experience in the Android/Java field.

Quoting the original PR:
> Prior to 22efd95, users could customise the OkHttp client used by React Native on Android by calling setOkHttpClientFactory in OkHttpClientProvider.

> This functionality has a variety of legitimate applications from changing connection timeouts or pool size to Stetho integration. In this commit also provide the ability for WebSocket.

I believe I found a work-around but was not able validate the failing tests @cpojer experienced. Because of that I would label my PR as WIP.

Closes #18920

## Fix Description
[`getOkHttpClient`](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java) creates a client in-place, if a `sClient == null`.
```java
public static OkHttpClient getOkHttpClient() {
    if (sClient == null) {
      sClient = createClient();
    }
    return sClient;
  }
```
[`connect`](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java) of the `WebSocketModule` makes no use of this in-place client, it initalizes a client by itself (uses a "virgin" client).
```java
  @ReactMethod
  public void connect(
    final String url,
    @Nullable final ReadableArray protocols,
    @Nullable final ReadableMap options,
    final int id) {
    OkHttpClient client = new OkHttpClient.Builder()
      .connectTimeout(10, TimeUnit.SECONDS)
      .writeTimeout(10, TimeUnit.SECONDS)
      .readTimeout(0, TimeUnit.MINUTES) // Disable timeouts for read
      .build();

 //...
```
[The original fix](https://github.com/facebook/react-native/pull/19632/files) uses `getOkHttpClient` thus `WebSocketModule` implicitly receives a pre-loaded client, if no explicit `sClient` is set.

This PR adds a `hasOkHttpClient` Method to the `OkHttpClientProvider`, so that the `WebSocketModule` can select if it will use the explicitly set client from the provider or fall back to using the "virgin" client.

## Changelog

[ANDROID] [Changed] - Added custom OkHttpClient injection into WebSocketModule only if custom OkHttpClient is configured

## Test Plan

Quoting the original PR:
> Create React Native application and set a custom factory in the constructor, e.g. OkHttpClientProvider.setOkHttpClientFactory(new CustomNetworkModule());

> Where a custom factory would look like:
> ```java
> class CustomNetworkModule implements OkHttpClientFactory {
>    public OkHttpClient createNewNetworkModuleClient() {
>        return new OkHttpClient.Builder().build();
>    }
>}
>```

I was unable to locate the existing unit or integration test files for the `OkHttpClientProvider` and `WebSocketModule`, so I am unsure if unit tests are expected.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java
@@ -42,6 +42,10 @@ public class OkHttpClientProvider {
     sFactory = factory;
   }
 
+  public static boolean hasOkHttpClient() {
+    return sClient != null;
+  }
+
   public static OkHttpClient getOkHttpClient() {
     if (sClient == null) {
       sClient = createClient();
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
@@ -85,14 +85,14 @@ public final class WebSocketModule extends ReactContextBaseJavaModule {
     @Nullable final ReadableArray protocols,
     @Nullable final ReadableMap options,
     final int id) {
-    OkHttpClient client;
+    OkHttpClient.Builder builder;
     if (OkHttpClientProvider.hasOkHttpClient()) {
-      client = OkHttpClientProvider.getOkHttpClient().newBuilder();
+      builder = OkHttpClientProvider.getOkHttpClient().newBuilder();
     } else {
-      client = new OkHttpClient.Builder();
+      builder = new OkHttpClient.Builder();
     }
 
-    client
+    OkHttpClient client = builder
       .connectTimeout(10, TimeUnit.SECONDS)
       .writeTimeout(10, TimeUnit.SECONDS)
       .readTimeout(0, TimeUnit.MINUTES) // Disable timeouts for read
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java
@@ -84,7 +84,14 @@ public final class WebSocketModule extends ReactContextBaseJavaModule {
     @Nullable final ReadableArray protocols,
     @Nullable final ReadableMap options,
     final int id) {
-    OkHttpClient client = new OkHttpClient.Builder()
+    OkHttpClient client;
+    if (OkHttpClientProvider.hasOkHttpClient()) {
+      client = OkHttpClientProvider.getOkHttpClient().newBuilder();
+    } else {
+      client = new OkHttpClient.Builder();
+    }
+
+    client
       .connectTimeout(10, TimeUnit.SECONDS)
       .writeTimeout(10, TimeUnit.SECONDS)
       .readTimeout(0, TimeUnit.MINUTES) // Disable timeouts for read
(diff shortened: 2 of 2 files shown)


cpojer:
It seems like this PR changes the default behavior similarly to the other PR, and it makes our e2e tests fail :(

jakoblorz:
Hm, okay. There is one edge case in which my fix wont accomplish what we want. If at some point getOkHttpClient is called (without checking hasOkHttpClient before), a sClient instance is created and any further hasOkHttpClient calls will return true.

cpojer:
I guess the best way to understand the issue is if you could explain what `OkHttpClientProvider.getOkHttpClient().newBuilder()` does differently from the method we are using now. I assume it uses slightly different defaults – what are they?

jakoblorz:
So there are 2 cases I believe:
1. `getOkHttpClient()` is called after checking `hasOkHttpClient()` (`hasOkHttpClient() == true`):
The custom client is provided, with the defaults the developer selected
2. `getOkHttpClient()` is called without checking `hasOkHttpClient()` (maybe at some other point in the code):
The default client from the `createClient()` method is [created and the returned](https://github.com/facebook/react-native/blob/2be8dbc4d54a6f057c289436e0d4cce36c0e5068/ReactAndroid/src/main/java/com/facebook/react/modules/network/OkHttpClientProvider.java#L51). Any consecutive call to `hasOkHttpClient()` will return true, even if there is no custom client intended as in 1.

Besides these 2 cases (when `hasOkHttpClient()` returned false), the defaults selected are the ones already present before this PR.

cpojer:
Unfortunately this approach breaks internal use cases at FB so we cannot merge it at this time :(

arabold:
@cpojer do you have any advise how to improve or modify this PR so it can be merged eventually? We have a specific use case where we _have_ to customize the OkHttpProvider to accept a single custom root CA in order to establish a secure connection. While this works for HTTPS connections via React Native, it seems that this issue above is preventing us from connecting via WebSocket.

jakoblorz:
@arabold Well you could fork my fork and rebase all master commits onto it, then build react-native and use that version 🤷‍♂️
Though there would be no guarantees that facebook/react-native provides (explicit or implicit). It seems that there are edge-cases in the e2e tests, which cannot be published but prevent this PR to be merged.

PR #24335, opened 2019-04-06
Remove FB internal unused JSI code from OSS

zhongwuzw:
## Summary

Seems we expose some unnecessary declarations in JSI to OSS version. Let's remove it.
cc. @cpojer .

## Changelog

[General] [Changed] - Remove FB internal unused JSI code from OSS

## Test Plan

CI passes.


shergin:
Sorry, that file is dirsynced with internal FB infra, so removing that will probably break facebook apps.
We will work on refactoring this parts outside of this files/interface.

Issue #24346, opened 2019-04-08
[Android 8.0] com.facebook.react.uimanager.NativeViewHierarchyManager.dropView (NativeViewHierarchyManager.java:536)


7772:
## 🐛 Bug Report

![Screen Shot 2019-04-08 at 10 12 48 AM](https://user-images.githubusercontent.com/22615385/55692904-e6244e80-59e6-11e9-8aaf-0137caab633c.png)

This bug occurs unpredictably but continuously in Android 8.0. There is [#20288](https://github.com/facebook/react-native/issues/20288) related this bug but it is not perfect solution, So we have still faced this issue.

Since [the issue](https://github.com/facebook/react-native/issues/20288) is merged on react-native, `dropView` method is like below.

```
  protected synchronized void dropView(View view) {
    UiThreadUtil.assertOnUiThread();
    if (mTagsToViewManagers.get(view.getId()) == null) {
      // This view has already been dropped (likely due to a threading issue caused by async js
      // execution). Ignore this drop operation.
      return;
    }
    if (!mRootTags.get(view.getId())) {
      // For non-root views we notify viewmanager with {@link ViewManager#onDropInstance}
      resolveViewManager(view.getId()).onDropViewInstance(view);
    }
    ViewManager viewManager = mTagsToViewManagers.get(view.getId());
    if (view instanceof ViewGroup && viewManager instanceof ViewGroupManager) {
      ViewGroup viewGroup = (ViewGroup) view;
      ViewGroupManager viewGroupManager = (ViewGroupManager) viewManager;
      for (int i = viewGroupManager.getChildCount(viewGroup) - 1; i >= 0; i--) {
        View child = viewGroupManager.getChildAt(viewGroup, i);
        if (child == null) {
            FLog.e(TAG, "Unable to drop null child view");
        } else if (mTagsToViews.get(child.getId()) != null) {
          dropView(child);
        }
      }
      viewGroupManager.removeAllViews(viewGroup);
    }
    mTagsToViews.remove(view.getId());
    mTagsToViewManagers.remove(view.getId());
  }
```

The parameter, `View view` is passed on `dropView` method and it may be sometimes null. So `view` object can not invoke `getId` method. 

Simply add this codes below in the top of `dropView` method.

```
if (view == null) return;
```

## To Reproduce

I have tried to reproduce this bug, but It is difficult to make the circumstance that the bug occurs.

## Expected Behavior

The bug do not occur when `view` is null.

## Environment

[skip envinfo]

[...]

Issue #20288, opened 2018-07-19
[Android 8.0] Attempt to invoke virtual method 'int android.view.View.getId()' on a null object reference


Yusiro85:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [ ] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
Run `react-native info` in your terminal and paste its contents here.

## Description
Describe your issue in detail. Include screenshots if needed. If this is a regression, let us know.

## Reproducible Demo
Let us know how to reproduce the issue. Include a code sample, share a project, or     share an app that reproduces the issue using https://snack.expo.io/. Please follow the guidelines for providing a MCVE: https://stackoverflow.com/help/mcve

> java.lang.NullPointerException Attempt to invoke virtual method 'int android.view.View.getId()' on a null object reference 
>     NativeViewHierarchyManager.java:537 com.facebook.react.uimanager.NativeViewHierarchyManager.dropView
>     NativeViewHierarchyManager.java:538 com.facebook.react.uimanager.NativeViewHierarchyManager.dropView
>     NativeViewHierarchyManager.java:538 com.facebook.react.uimanager.NativeViewHierarchyManager.dropView
>     NativeViewHierarchyManager.java:538 com.facebook.react.uimanager.NativeViewHierarchyManager.dropView
>     NativeViewHierarchyManager.java:538 com.facebook.react.uimanager.NativeViewHierarchyManager.dropView
>     NativeViewHierarchyManager.java:431 com.facebook.react.uimanager.NativeViewHierarchyManager.manageChildren
>     UIViewOperationQueue.java:180 com.facebook.react.uimanager.UIViewOperationQueue$ManageChildrenOperation.execute
>     UIViewOperationQueue.java:819 com.facebook.react.uimanager.UIViewOperationQueue$1.run
>     UIViewOperationQueue.java:926 com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches
>     UIViewOperationQueue.java:47 com.facebook.react.uimanager.UIViewOperationQueue.access$2100
>     UIViewOperationQueue.java:986 com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.doFrameGuarded
>     GuardedFrameCallback.java:31 com.facebook.react.uimanager.GuardedFrameCallback.doFrame
[...]

PR #24347, opened 2019-04-08
ignore dropView method when view is null

7772:
## Summary

In #20288, we solved `com.facebook.react.uimanager.NativeViewHierarchyManager.dropView (NativeViewHierarchyManager.java:536)` by adding codes that prevent to run method on `view` object when `view.getId()` is null.

But if `view` is null, `view.getId()` can make error about NullPointerException. So, I think `dropView` method  needs to check if `view` is null.

If you need more information, Please read #24346. 

## Changelog

[Android] [Fixed] - Ignore dropView method when view is null. 

## Test Plan

This issue is unpredictable bug so please just read code. 
And You can see also #20288 related on this issue.


dulmandakh:
Please add description and changelog, so that release managers can prepare release notes easily. Thank you

cpojer:
Let me try this again…

PR #24359, opened 2019-04-08
[Android] Keyboard accessibility improvements

sweggersen:
## Summary
In order to meet our accessibility requirements we need to have full support for keyboard navigation. The Touchable components works with press/tap with a finger, but doesn't respond to 'enter' when using a keyboard. Navigation works fine. This PR adds an onClick listener to touchable views that have the onPress prop defined.

## Changelog
[Android] [Added] - Add View.OnClickListener to Touchable components when onPress is defined

## Test Plan
Add any Touchable component with an onPress callback. Use a keyboard and navigate to the touchable component and press the 'enter' key. Expected behavior is that the onPress callback is called. 

@acoates-ms

Sam Mathias Weggersen
Microsoft Corp.

### DIFF
--- a/Libraries/Components/View/ViewPropTypes.js
+++ b/Libraries/Components/View/ViewPropTypes.js
@@ -18,6 +18,7 @@ import type {TVViewProps} from 'TVViewPropTypes';
 import type {
   AccessibilityComponentType,
   AccessibilityTrait,
+  AccessibilityNodeInfoProp,
   AccessibilityRole,
   AccessibilityStates,
 } from 'ViewAccessibility';
--- a/Libraries/Components/View/ViewPropTypes.js
+++ b/Libraries/Components/View/ViewPropTypes.js
@@ -18,7 +18,6 @@ import type {TVViewProps} from 'TVViewPropTypes';
 import type {
   AccessibilityComponentType,
   AccessibilityTrait,
-  AccessibilityNodeInfoProp,
   AccessibilityRole,
   AccessibilityStates,
 } from 'ViewAccessibility';
--- a/Libraries/Components/View/ViewPropTypes.js
+++ b/Libraries/Components/View/ViewPropTypes.js
@@ -320,7 +320,6 @@ type AndroidViewProps = $ReadOnly<{|
    */
   nextFocusUp?: ?number,
 
-
   /**
    * Whether this `View` should be clickable with a non-touch click, eg. enter key on a hardware keyboard.
    *
--- a/Libraries/Components/View/ViewPropTypes.js
+++ b/Libraries/Components/View/ViewPropTypes.js
@@ -319,6 +319,21 @@ type AndroidViewProps = $ReadOnly<{|
    * @platform android
    */
   nextFocusUp?: ?number,
+
+
+  /**
+   * Whether this `View` should be clickable with a non-touch click, eg. enter key on a hardware keyboard.
+   *
+   * @platform android
+   */
+  clickable?: boolean,
+
+  /**
+   * The action to perform when this `View` is clicked on by a non-touch click, eg. enter key on a hardware keyboard.
+   *
+   * @platform android
+   */
+  onClick?: () => void,
 |}>;
 
 type IOSViewProps = $ReadOnly<{|
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -31,6 +31,7 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
   private static final String PROP_ELEVATION = "elevation";
   private static final String PROP_Z_INDEX = "zIndex";
   private static final String PROP_RENDER_TO_HARDWARE_TEXTURE = "renderToHardwareTextureAndroid";
+  private static final String PROP_ACCESSIBLE = "accessible";
   private static final String PROP_ACCESSIBILITY_LABEL = "accessibilityLabel";
   private static final String PROP_ACCESSIBILITY_COMPONENT_TYPE = "accessibilityComponentType";
   private static final String PROP_ACCESSIBILITY_HINT = "accessibilityHint";
@@ -110,6 +111,11 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
   public void setNativeId(@Nonnull T view, String nativeId) {
     view.setTag(R.id.view_tag_native_id, nativeId);
     ReactFindViewUtil.notifyViewRendered(view);
+  }
+
+  @ReactProp(name = PROP_ACCESSIBLE)
+  public void setAccessible(T view, boolean accessible) {
+    view.setImportantForAccessibility(accessible ? View.IMPORTANT_FOR_ACCESSIBILITY_YES : View.IMPORTANT_FOR_ACCESSIBILITY_NO);
   }
 
   @ReactProp(name = PROP_ACCESSIBILITY_LABEL)
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -6,10 +6,9 @@
 package com.facebook.react.uimanager;
 
 import android.graphics.Color;
-import android.os.Build;
-import androidx.core.view.ViewCompat;
 import android.view.View;
 import android.view.ViewParent;
+import androidx.core.view.ViewCompat;
 import com.facebook.react.R;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.uimanager.AccessibilityDelegateUtil.AccessibilityRole;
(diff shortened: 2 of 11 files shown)


cpojer:
Thanks so much for the PR! Seems like our bot is unhappy, do you mind fixing it up?

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -31,6 +31,7 @@
   private static final String PROP_ELEVATION = "elevation";
   private static final String PROP_Z_INDEX = "zIndex";
   private static final String PROP_RENDER_TO_HARDWARE_TEXTURE = "renderToHardwareTextureAndroid";
+  private static final String PROP_ACCESSIBLE = "accessible";
```
blavalla: What is this for?  It seems like its functionality could be handled with the existing IMPORTANT_FOR_ACCESSIBILITY prop, and it doesn't see, to be used in the rest of this PR.

ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -223,6 +226,29 @@ public void setCollapsable(ReactViewGroup view, boolean collapsable) {
...
+                }});
+
+      // Clickable elements are focusable. On API 26, this is taken care by setClickable.
+      // Explicitly calling setFocusable here for backward compatibility.
+      view.setFocusable(true /*isFocusable*/);
+    }
+    else {
+      view.setOnClickListener(null);
+      view.setClickable(false);
+      view.setFocusable(false);
```
blavalla: I think this may cause some problems. Just because something is set to not be clickable, doesn't mean it should explicitly be not focusable.  

For example, if you set accessible="true" (which sets view.setFocusable(true) and clickable="false", (which sets view.setFocusable(false), it would contradict here and I'm not sure which would actually win out.

cpojer:
Thanks @blavalla!

@sweggersen what do you think about making this a default?

sweggersen:
@blavalla @cpojer Thanks for great feedback. Yeah, I think that it makes sense to have it on by default.

cpojer:
Awesome, would you mind making those changes?

ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
sweggersen: My initial thought was that if you enable click on a view with setClickable(true) and then setClickable(false) later, you would want it back to the way it was. But I agree that this could introduce unwanted behavior. I will remove `view.setFocusable(false)`, but we'll need to keep setting it to true alongside `setOnClickListener`

Libraries/Components/View/ViewPropTypes.js
```diff
@@ -319,6 +319,20 @@ type AndroidViewProps = $ReadOnly<{|
    * @platform android
    */
   nextFocusUp?: ?number,
+
+  /**
+   * Whether this `View` should be clickable with a non-touch click, eg. enter key on a hardware keyboard.
+   *
+   * @platform android
+   */
+  clickable?: boolean,
```
necolas: I want to share some details about some of the work happening inside Facebook and how it relates to this change.

* Supporting multiple input types is a problem for all platforms.
* The web platform has designed APIs for this purpose.
* The concept of "click" on the web is essentially legacy, and touch events still trigger `onClick`.
* React Native for Web supports multiple input types but relies on many more events than `onClick` (i.e., `onMouseDown`, `onKeyPress`, etc.).
* Internally we're experimenting with a [new React event system](https://github.com/facebook/react/issues/15257). On the web that relies on Pointer Events and would eschew the need to use `onClick`. The general idea is to move away from input-specific event types and instead use `onPress`, etc., with the event object containing information about the pointer type that was used.
* Internally we're experimenting with web implementations of `View` and `Text` that shed all the event props found in React Native (in favour of the new event system) and introduce a `focusable` prop to mark a View as being able to receive (or not) user-initiated focus by any input type (all Views are programmatically focusable). In this way you could set `focusable={false}` and retain all the mouse/touch functionality but hide the View from inputs like keyboards that first need to focus an element before it can be interacted with. That's an important pattern on the web where you might want to make something like an avatar interactive to pointers, but hidden from keyboards if the functionality is redundant, i.e., if there's also an interactive user name that performs the same action.

Given all that I'd suggest we do something like rename `clickable` to `focusable` and decouple it from `onClick`. If `onClick` here is meant to be used for mouse as well as keyboard, that's probably a good enough prop for now to unblock other input types on Android. We'll hopefully address this properly in RN in the future with the new event system.

necolas:
I left a comment on the diff sharing some context about how we're looking to solve similar use cases for the web. It might be worth adjusting this change. Let me know your thoughts

Libraries/Components/View/ViewPropTypes.js
sweggersen: Thanks for the write up!

About the decouple of `focusable` from `onClick`, what do you suggest?

The Android default onClick listener works for all input types 'keyboard', 'touch', 'mouse', which is what we need support for. Before this change, only 'touch' and 'mouse' events would work.

blavalla: Naming this "focusable" could definitely lead to some confusion here due to Android's already confusing separate concepts of keyboard focusability and accessibility focusability.

Setting "focusable=true" on a standard Android view will set it to be both keyboard and accessibility focusable, but if this is property is meant to only trigger keyboard focusability, we should probably be more explicit in its naming.

necolas: > About the decouple of focusable from onClick, what do you suggest?

I was thinking that the `focusable` feature is something we need independent of how events are handled. So in the future we might not have event handlers on `View` but we'll still need to mark views as being able to receive user-initiated focus following a mouse, touch, or keyboard event.

> Setting "focusable=true" on a standard Android view will set it to be both keyboard and accessibility focusable, but if this is property is meant to only trigger keyboard focusability, we should probably be more explicit in its naming.

What is "accessibility focusable" on Android? The way we're using this on web is to allow the element to receive focus following any user action from any pointer device. It happens to also be required to support interactions on a view (with seems to overlap with this PR for Android).

blavalla: To answer what "accessibility focusable" is requires a bit of a history lesson, so bear with me. 

In the early days of Android many devices had trackballs or d-pads as an alternate non-touch way to interact with the device.  To the system, these keys worked much like external keyboards. To allow your UI to be interactive you had to make sure it was focusable in "non touch mode", and the best way to do this was to set focusable="true".  You would generally want all interactive elements to be focusable in this way, but wouldn't want non-interactive elements to be focusable, since it would be annoying to have to press keys to get past all of the non-interactive content.  

Then later on Google decided to add a real accessibility system to Android, and built the Talkback screen reader as well as other accessibility services like Switch Access and Select to Speak.  Some of these services (like Talkback) can only really be useful if you can focus on non-interactive content, like plain text.  Otherwise, this content could be totally hidden from a user with a vision impairment. This meant that they couldn't simply use the same focus system used for D-Pad's and keyboards, and they had to come up with a way that you could have an element be focusable for accessibility, but *not* focusable for keyboard users.  This is what I mean when I say "accessibility focusable".  

Unfortunately, they didn't simply add a property like accessibilityFocusable and instead derived whether or not something would be accessibility focusable by looking at the combination of multiple different properties (click handlers, the focusable property, if the view was a leaf node, if it had focusable ancestors, etc.).  It wasn't until Android Pie that they cleaned this up and added `setScreenReaderFocusable` to the View class.

necolas: Thank you! It sounds to me like the `focusable` concept is equivalent to its meaning on Android, as it's intended for interactive elements. Setting `focusable` to `false` would mean the element is no longer accessible to keyboards. On the web it would still be accessible to screen readers, and hiding it from them would need `accessibilityState.hidden` to be `true` as well.

necolas: Any more thoughts?

sweggersen: Thanks for all the info and great discussion above.

I think it makes sense to rename this prop to something more correct. What about `keyClickable` or `externalClickable`?

necolas: Neither of those names seem quite right. FWIW I noticed SwiftUI also uses `focusable` for essentially the same concept - https://developer.apple.com/documentation/swiftui/view/3278564-focusable - with the difference being that in React (at least on web, for now) if `focusable` is set to false on a View it is still able to receive programmatic focus for accessibility reasons. So `focusable` is about whether a view can receive user-initiated focus from a pointer or keyboard

matthargett: On our platforms, we also use focusable and that’s what @jayphelps (formerly of Netflix and ThisDot Labs) recommended in his draft RFC on this topic. Besides keyboard, and 5-way controls, this also maps well to gaze (AR/VR/XR) and 6DOF controls in my experience. It’s really awesome to see this ideas make it into React itself!

sweggersen: Thanks, I think it makes sense to call it `focusable`. Who should do the work? Do we just update this PR?

cpojer: Please send a new PR that changes everything. We can do it all in a single commit :)

necolas: If you'd like to write the patch that would be great! A new PR is best since this one has already been merged.

cpojer: Hey @sweggersen: just wanted to ping you and ask whether you could send a PR and rename this to `focusable`?

sweggersen: It's on my list. I'll do it this week.

sweggersen: It's on my list. I'll do it this week.

PR #24360, opened 2019-04-08
[CI] Fix Docker Android tests container issue related to the JSC

hramos:
## Summary

[Fixes an issue where the Docker Android tests container cannot be built.](https://github.com/facebook/react-native/pull/24276#issuecomment-480915232) The JSC is now pulled from the npm registry, so we need to run `yarn` prior to pulling the Gradle dependencies.

## Changelog

[Android] [Fixed] - Fixed React Native Android tests Docker container issue related to the JSC

## Test Plan

Ran `yarn docker-build-android` locally and verified the container was built successfully.

### DIFF
--- a/.circleci/Dockerfiles/Dockerfile.android
+++ b/.circleci/Dockerfiles/Dockerfile.android
@@ -8,10 +8,10 @@
 #
 # The base image is expected to remain relatively stable, and only
 # needs to be updated when major dependencies such as the Android
-# SDK or NDK are updated. 
+# SDK or NDK are updated.
 #
-# In this Android Test image, we download the latest dependencies 
-# and build a Android application that can be used to run the 
+# In this Android Test image, we download the latest dependencies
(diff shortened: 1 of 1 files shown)


hramos:
This won't fix the current `test_android` failures, because the test container is not used on Circle CI. We only use the base container.

This fix does allow anyone to easily repro the current failure on master like so:

1. Install Docker
2. Clone the RN repository
3. yarn setup-docker-android
3. yarn build-docker-android
4. yarn test-android-unit

Prior to this fix, step #3 would fail. Once this PR lands, you can proceed to step #4 and repro the current failure on `master` when Android unit tests are run.

sahrens:
Hmm, without this diff I was able to run step 3, but step 5 failed to build and couldn't run the tests. I didn't try running step 4 manually, maybe that was part of the problem?

gengjiawen:
I am thinking add this to circle ci to prevent future regression.

Issue #19453, opened 2018-05-25
<Text> Component with over ~500 lines won't render [iOS]


jackthias:
- [X] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [X] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

A `Text` component with about 500-600+ lines of text renders completely blank.

## Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 9.3.0
  Yarn: Not Found
  npm: 5.8.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.3.1 => 16.3.1
  react-native: 0.55.4 => 0.55.4

## Steps to Reproduce

```import React, { Component } from 'react';
import {
  Text,
  View,
  ScrollView
} from 'react-native';

const text = ".\n".repeat(600)

export default class App extends Component {
  render() {
    return (
      <View style={{
        flex: 1,
        justifyContent: 'flex-start',
        alignItems: 'center',
        backgroundColor: '#F5FCFF',
      }}>
        <ScrollView style={{flex: 1}}>
          <Text>{text}</Text>
        </ScrollView>
      </View>
    );
  }
}
```

At 600 lines there's no rendering of the text component, at 200 it renders fine.

## Expected Behavior

I expected the normal behavior of a `Text` component embedded in a `ScrollView`. For an example, render the above code with a repeat value of 200 instead of 600.

![screen shot 2018-05-25 at 2 38 33 pm](https://user-images.githubusercontent.com/2886316/40567297-95ddd2be-6029-11e8-89d8-e61596bdec22.png)

## Actual Behavior

Nothing is rendered, just a blank area where the text should be.

![screen shot 2018-05-25 at 2 37 58 pm](https://user-images.githubusercontent.com/2886316/40567302-9c2f79ba-6029-11e8-9855-eb3f6cf21389.png)

jackthias:
Sorry, I accidentally tapped the enter key while I was still typing the title.

jackthias:
This is tagged as no environment info, but there's clearly environment info in the post. Am I missing anything? That's the output of the command.

ericlewis:
The magic number appears to be 490 characters. Am taking a look at this.

ericlewis:
[...]

PR #24365, opened 2019-04-09
Workaround for long text rendering bug in YellowBox

janicduplessis:
## Summary

On iOS simulator there is an issue where very long text is not rendered properly. See https://github.com/facebook/react-native/issues/19453 for more context. A workaround is to use a non-editable TextInput instead of a Text component.

## Changelog

[iOS] [Fixed] - Workaround for long text rendering bug in YellowBox

## Test Plan

Tested that warnings shown in YellowBox display properly on both iOS and Android when using a very long text (15k+ chars)

### DIFF
--- a/Libraries/YellowBox/UI/YellowBoxInspector.js
+++ b/Libraries/YellowBox/UI/YellowBoxInspector.js
@@ -66,12 +66,16 @@ class YellowBoxInspector extends React.Component<Props, State> {
             <View style={styles.bodyHeading}>
               <Text style={styles.bodyHeadingText}>Warning</Text>
             </View>
-            <Text style={styles.bodyText}>
+            {/* There is a bug on iOS that causes very long text to
+                render incorrectly. As a workaround we can use a non-editable
+                multiline TextInput which does not have the bug.
+                See https://github.com/facebook/react-native/issues/19453 */}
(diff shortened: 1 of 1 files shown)


matthargett:
If I recall correctly, @fkgozali said that TextInput won't be in the core of fabric components. Does that mean a different workaround should be found for getting YellowBox to display properly?

cpojer:
We ran into the same issue at FB. The way we fixed it is by capping the warning message that was emitted because there is no point in printing such a long message in a YellowBox. I don't feel good about merging this workaround. I agree that an empty YellowBox isn't useful but 500+ lines of text in a warning box isn't useful either.

janicduplessis:
@cpojer Is that length fix OSS? Otherwise we should limit the length in `YellowBoxInspector.js`, which was my initial fix before investigating more and figuring out that `TextInput` actually renders long text properly.

cpojer:
That specific change was for Relay: https://github.com/facebook/relay/commit/f853e91df7c30f1c09bb43eea7b8673a668b14ea

I'm ok with capping the YellowBox error message. I wanted to do that initially but never had time to find a good way to calculate the max length for text.

janicduplessis:
Alright so I could help myself and found the root cause of the issue, see https://github.com/facebook/react-native/issues/19453#issuecomment-481503189

Issue #23910, opened 2019-03-13
Fabric: Race condition when using RCTFabricSurfaceHostingProxyRootView


ericlewis:
## 🐛 Bug Report

When attempting to use `RCTFabricSurfaceHostingProxyRootView` the application crashes when the surface is asked to start before the bridge has finished loading. Note, that I sometimes get a red screen instead, saying _fbBatchedBridge couldn't be found, it's "random". It is also pretty clear to me that the runtime pointer is null when this happens. Stack trace below:

![stack trace](https://imgur.com/dCjgQ4e.png)

## To Reproduce

- Follow the directions to run on https://github.com/facebook/react-native/pull/23803
- Comment out the code in `Swizzle_RCTFabricSurface.m` 
- remove the `RCTJavaScriptDidLoadNotification` observer in `AppDelegate.m`
- Open workspace, run app, observe error.

## Expected Behavior

`RCTFabricSurfaceHostingProxyRootView` to wait for the bridge to fully load before starting the surface.

## Code Example

https://github.com/facebook/react-native/pull/23803. Hopefully my swizzle / hack code can be of help in proper solution!

## Environment

N/A? This is RNTester.

ericlewis:
cc @fkgozali

fkgozali:
we’re tracking this internally and will investigate in a few weeks

fkgozali:
These 2 commits should fix the issues:
* null `Runtime`: https://github.com/facebook/react-native/commit/b312543d3cc005dc1548775d455848f19719c57a (this is actually a workaround, not an ideal fix)
* Missing `__fbBatchedBridge`:
https://github.com/facebook/react-native/commit/6aa896896f0cf6bf3f7617d980b105c5f8770f56

fkgozali:
Please re-open if the issue still exists.

PR #24372, opened 2019-04-09
 [Fabric] Make surface presenter manage start of surface

zhongwuzw:
## Summary

1. Start surface when first run, before, when we first load js bundle, `_startAllSurfaces` would not be called because `bridge == _batchedBridge`.
2. I think this fixes https://github.com/facebook/react-native/issues/23910, so I add the `return` in swizzle method to check. 

cc @shergin.

## Changelog

[iOS] [Fixed] - Make surface presenter manage start of surface

## Test Plan

N/A.


React/Fabric/Surface/RCTFabricSurface.mm
```diff
@@ -67,8 +65,7 @@ - (BOOL)start
   if (![self _setStage:RCTSurfaceStageStarted]) {
     return NO;
   }
-
-  [_surfacePresenter startSurface:self];
+  [_surfacePresenter registerSurface:self];
```
shergin: I definitely like this part. This is how it should be IMHO.

React/Fabric/Surface/RCTFabricSurface.mm
```diff
@@ -55,8 +55,6 @@ - (instancetype)initWithSurfacePresenter:(RCTSurfacePresenter *)surfacePresenter
     _touchHandler = [RCTSurfaceTouchHandler new];
 
     _stage = RCTSurfaceStageSurfaceDidInitialize;
```
shergin: 👍

RNTester/RNTester/AppDelegate.m
```diff
@@ -77,6 +77,7 @@ - (BOOL)application:(__unused UIApplication *)application didFinishLaunchingWith
 #ifdef RN_FABRIC_ENABLED
 // FIXME: remove when resolved https://github.com/facebook/react-native/issues/23910
 - (void)handleJavaScriptDidLoadNotification:(__unused NSNotification*)notification {
+  return;
```
shergin: And please change this.

zhongwuzw:
@shergin How about this, we add a stage check before start surface, not check bridge's equality.

RNTester/RNTester/AppDelegate.mm
```diff
@@ -98,15 +86,6 @@ - (BOOL)application:(__unused UIApplication *)application didFinishLaunchingWith
   return YES;
 }
 
-#ifdef RN_FABRIC_ENABLED
```
zhongwuzw: I removed all these hack code.

PR #24373, opened 2019-04-09
[Fabric] Add valid check for shadow tree and component view

zhongwuzw:
## Summary

The reason why we need add these checks is sometimes js bundle may not run successful, in these cases ,`rootView` or `shadowTree` may not created.

cc. @shergin 

## Changelog

[General] [Fixed] - Add valid check for shadow tree and component view

## Test Plan

N/A.

### DIFF
--- a/ReactCommon/fabric/uimanager/Scheduler.cpp
+++ b/ReactCommon/fabric/uimanager/Scheduler.cpp
@@ -147,26 +147,30 @@ void Scheduler::stopSurface(SurfaceId surfaceId) const {
   SystraceSection s("Scheduler::stopSurface");
 
   long commitStartTime = getTime();
-  shadowTreeRegistry_.visit(
-      surfaceId, [commitStartTime](const ShadowTree &shadowTree) {
-        // As part of stopping the Surface, we have to commit an empty tree.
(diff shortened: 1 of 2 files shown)


ReactCommon/fabric/uimanager/Scheduler.cpp
```diff
@@ -147,26 +147,30 @@ void Scheduler::stopSurface(SurfaceId surfaceId) const {
...
-                      /* .eventEmitter = */
-                      ShadowNodeFragment::eventEmitterPlaceholder(),
-                      /* .children = */
-                      ShadowNode::emptySharedShadowNodeSharedList(),
-                  });
-            },
-            commitStartTime);
-      });
+  
+  auto hasShadowTree = shadowTreeRegistry_.visit(
```
shergin: Early return in case of exception is NOT a valid solution (especially here where stopSurface does not have a way to communicate back about the problem).
The caller side must ensure that the surface with given id exists (so it should not call the method twice).

shergin:
(See my comment in the code.)

zhongwuzw:
@shergin Should we put this to surface presenter? and check the stage of surface before stop it?

PR #24375, opened 2019-04-09
add docker test to circle-ci

gengjiawen:
## Summary

add docker related test to ci (only build part), prevent future regression like https://github.com/facebook/react-native/pull/24360

## Changelog

[General] [Added] - add docker related test to ci

## Test Plan

pas current ci


hramos:
Thanks for taking care of this!

PR #24381, opened 2019-04-09
Install local react-native package in e2e

hramos:
## Summary

Following 6df2edeb2a33d529e4b13a5b6767f300d08aeb0a, the template for a new app no longer refers to a specific version of React Native, as the React Native CLI will set the correct version when `react-native init` is run. As it happens, the CLI cannot find a local react-native@1000 version when run as part of the JavaScript e2e tests. Therefore, I am working around using the CLI directly, and instead we're copying the contents of `template/` into a temporary directory for e2e tests. This is not dissimilar to what we already test internally at Facebook.

## Changelog

[General] [Fixed] - Fixed JavaScript e2e tests

## Test Plan

Local:
```
node ./scripts/run-ci-e2e-tests.js --js --retries 3 --skip-cli-install
echo $?
0
```

Circle CI: [✅ `test_javascript`](https://circleci.com/gh/hramos/react-native/5286)

### DIFF
--- a/scripts/run-ci-e2e-tests.js
+++ b/scripts/run-ci-e2e-tests.js
@@ -43,10 +43,14 @@ let exitCode;
 // Make sure we installed local version of react-native
 function checkMarker() {
   if (!test('-e', path.basename(MARKER))) {
-    echo('Marker was not found, react native init command failed?');
+    echo('Marker was not found, template copy failed?');
     exitCode = 1;
     throw Error(exitCode);
   }
+}
+
+function describe(message) {
+  echo(`\n\n>>>>> ${message}\n\n\n`);
 }
 
 try {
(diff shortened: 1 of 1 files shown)


hramos:
The fix in this PR is not specific to JS e2e but rather any test that uses the `run-ci-e2e-tests.js` file. I just noticed that `test_ios` has been broken for the same reason (react-native@1000.0.0 does not exist on npm), but this PR surfaces another failure further down the line when `test_ios` runs its e2e tests.

It looks like it's related to the renaming of the test app performed in this PR, so I'll make sure to fix that before landing this PR. Please do not merge until then.

hramos:
`test_ios` is legit broken due to 094f221a0c2d24f16a9540e4f176783843ec4fd9 having moved `jsi` around. Landing, with the intent to fix forward later.

Issue #17277, opened 2017-12-19
Text shadow doesn't render when blur is set but offset is (0,0)


Firehed:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment
Environment:
  OS:  macOS High Sierra 10.13.1
  Node:  8.9.0
  Yarn:  1.3.2
  npm:  5.5.1
  Watchman:  4.9.0
  Xcode:  Xcode 9.0.1 Build version 9A1004
  Android Studio:  Not Found

Packages: (wanted => installed)
  react: 16.0.0-beta.5 => 16.0.0-beta.5
  react-native: ^0.49.3 => 0.49.5

### Steps to Reproduce

```jsx
<Text style={{
    textShadowOffset: { height: 0, width: 0 },
    textShadowColor: 'red',
    textShadowRadius: 3,
}}>This text should have a shadow</Text>
```

### Expected Behavior

Text should have a blurred shadow with no offset, similar to having `text-shadow: 0 0 3px red` in CSS.

### Actual Behavior

No shadow exists at all

### Reproducible Demo

https://snack.expo.io/B1X1Lzvff

### Other

At least on iOS, I think the issue is an overly-grabby conditional in [`Libraries/Text/RCTShadowText.m`](https://github.com/facebook/react-native/blob/6ad1f0957a020cb57b177ab015c17aa883e1f0ad/Libraries/Text/RCTShadowText.m#L467), something like `if (!CGSizeEqualToSize(_textShadowOffset, CGSizeZero) || _textShadowRadius != 0.0) {`. Not too sure about the build process so I can't easily test that fix locally. Plus it seems to occur on Android as well.

mcmar:
@shergin @mdvacca @davidaurelio Can on of you please re-open this or otherwise respond? I can confirm the issue and it does not appear to be intended behavior. Compare to text-shadow prop on web browsers, where the following is perfectly valid and displays the intended behavior:
`text-shadow: 0 0 2px #FF0000;`

PR #24398, opened 2019-04-10
Fix: text shadow displays on iOS when textShadowOffset is {0,0}

woodpav:
## Summary

There is a problem rendering text shadows on iOS. If the offset of the text shadow is `{width:0,height:0}`, the shadow does not display. This prevents you from representing a light directly above the text. This occurs because a text shadow only renders if the offset is a non-zero CGRect `{width:0,height:0}`. 

My change checks `textShadowRadius` instead. If `textShadowRadius` is not nan then the user is rendering a text shadow. There are no situations to render a shadow without `textShadowRadius` making it a good variable to check. 

This PR fixes this stale issue: https://github.com/facebook/react-native/issues/17277

## Changelog

[iOS] [Fixed] - Text shadow now displays when the textShadowOffset is {width:0,height:0}

## Test Plan

    textShadowOffset: {width:0,height: 0},
    textShadowRadius: 10,
    textShadowColor: "black"

Before, without fix: 
Screen Shot 2019-04-10 at 2 12 16 PM

After with fix:
Screen Shot 2019-04-10 at 2 04 18 PM

### DIFF
--- a/Libraries/Text/RCTTextAttributes.m
+++ b/Libraries/Text/RCTTextAttributes.m
@@ -170,7 +170,7 @@ NSString *const RCTTextAttributesTagAttributeName = @"RCTTextAttributesTagAttrib
   }
 
   // Shadow
-  if (!CGSizeEqualToSize(_textShadowOffset, CGSizeZero)) {
+  if (!isnan(_textShadowRadius)) {
     NSShadow *shadow = [NSShadow new];
     shadow.shadowOffset = _textShadowOffset;
     shadow.shadowBlurRadius = _textShadowRadius;


cpojer:
Can you show a before and after screenshot of this behavior change and include it in the test plan section?

woodpav:
Top one is with the fix, bottom one is without.

willnorris:
in case anyone's interested... this got attributed to @a-googler because of the use of "noreply@google.com" as author email.  @woodpav, I'm just curious, do you intentionally have your git client configured that way?

woodpav:
I set it up without an email because I didn't have one I wanted made public. I guess it defaulted to that for some reason. I guess it's too late to change?

cpojer:
That was actually my doing. I only had a few minutes to land this in the morning today and it failed to import at fb because it had no email assigned. I amended and force pushed with a generic email that I thought was a black hole. I didn’t think anybody would have that one actually registered. Lol.


________________________________

willnorris:
given that this has already been merged, yes it's too late.  If you just want to keep your email private, see https://help.github.com/en/articles/about-commit-email-addresses for a special `@users.noreply.github.com` email you can use.

woodpav:
No worries @cpojer I cede the glory for `isnan(_textShadowRadius))` to you!

@willnorris I have decided to step out of the shadows and add a developer email to my future commits. I will not fade from the annals of history anymore!

P.S. Thank you for your work on rn. Words alone cannot express my gratitude.

Issue #23282, opened 2019-02-03
react-native run-ios can not find any simulator


eric-PearSports:
## Environment

  React Native Environment Info:
    System:
      OS: macOS 10.14.2
      CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
      Memory: 6.76 GB / 32.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
    IDEs:
      Android Studio: 3.1 AI-173.4819257
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.57.8 => 0.57.8 
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7

## Description

I have been facing an issue where 'react-native run-ios' can not start, regardless of the simulator I add to the --simulator argument.
XCode has the correct location for the 'command line tools'

I am always getting the error:
Could not find iPhone X simulator

    Error: Could not find iPhone X simulator
        at resolve (/Users/eric/.../swim/node_modules/react-native/local-cli/runIOS/runIOS.js:149:13)
        at new Promise (<anonymous>)
        at runOnSimulator (/Users/eric/.../swim/node_modules/react-native/local-cli/runIOS/runIOS.js:134:10)
        at Object.runIOS [as func] (/Users/eric/.../swim/node_modules/react-native/local-cli/runIOS/runIOS.js:106:12)
        at Promise.resolve.then (/Users/eric/.../swim/node_modules/react-native/local-cli/cliEntry.js:117:22)

## Reproducible Demo

Installed XCode from scratch.
Brand new 'react-native new app'

'react-native run-ios' always complain about not finding the simulator, regardless of the --simulator option.

My temporary fix is to change:
/node_modules/react-native/local-cli/runIOS/findMatchingSimulator.js

line 42, to:
    if (!version.startsWith('**com.apple.CoreSimulator.SimRuntime.**iOS') && !version.startsWith('tvOS')) {

[...]

PR #24400, opened 2019-04-10
Starts simulator when prefix is included

jgreen01su:
## Summary
Fixes issue https://github.com/facebook/react-native/issues/23282 for `0.57.x`. Where `react-native run-ios` fails with error `Could not find iPhone X simulator react native`.

## Changelog

[iOS] [Fixed] - Fix starting iOS simulator after xcode 10.2 update.

## Test Plan

Update xcode to version 10.2, run `react-native run-ios`, and see if simulator starts.


cpojer:
This was already fixed in React Native 0.59, please upgrade to that version.

Issue #20352, opened 2018-07-23
Fetching specific JSON causes massive memory leak


Ashoat:
## Environment
```
React Native Environment Info:
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 6.28 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.6.0 - ~/.nvm/versions/node/v10.6.0/bin/node
    Yarn: 1.7.0 - /usr/local/bin/yarn
    npm: 6.1.0 - ~/.nvm/versions/node/v10.6.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
    Android SDK:
      Build Tools: 23.0.1, 25.0.0, 26.0.1, 26.0.2, 26.0.3, 28.0.1
      API Levels: 23, 26, 28
  IDEs:
    Android Studio: 3.1 AI-173.4819257
    Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.4.1 => 16.4.1
    react-native: ^0.56.0 => 0.56.0
  npmGlobalPackages:
    react-native-git-upgrade: 0.2.7
```

## Description
Downloading [this <1 KiB JSON file](http://ashoat.com/rnleak.json) once a second, parsing the JSON, and appending the resultant JSON to an array, causes the demo app below to hit 500MiB of memory usage in less than 15 seconds. Each request permanently allocates ~35MiB of memory.

## Reproducible Demo
I have created a [repo](https://github.com/Ashoat/react-native-leak) to reproduce the issue.

corpsemxq:
Will this be fixed by https://github.com/facebook/react-native/pull/19333?

Ashoat:
Unfortunately, no. I just updated the test repo to 0.57 and the issue is still occurring. After downloading the tiny JSON file one a second for sixty seconds, memory usage of the Release build of the test project climbs to over 2 GiB.

dlajarretie:
[...]

PR #24405, opened 2019-04-11
[Blob] Release underlying resources when JS instance in GC'ed on iOS

### LINKED ISSUE #23801: Severe memory leak affecting fetch(), iOS Release Mode 0.59.0-rc.3, 0.58.6, 0.57.8
## 🐛 Bug Report
Every call to `fetch()` uses memory and does not release it. The larger the fetch response, the more memory leaked.

- I've confirmed that this is happening in 0.57.8, 0.58.6, and 0.59.0-rc.3.
- I've tried this with both `node-fetch` and `whatwg-fetch` and it happens using both libraries.
- It is not isolated to `response.blob()`. It affects all types of responses.
- It happens with both Release and Debug.
- It happens with raw `XMLHttpRequest` as well as `fetch`.
- I can confirm that it's happening on a (real) iPhone X with iOS 12.1.4 and the Simulator for an iPhone 8 with iOS 12.1.

Comparing memory snapshots in Safari found no difference in size even while the Chrome/Xcode memory profiles showed dramatic increases in memory. I'm not great at debugging memory leaks so I could be wrong but I suspect that means it might be on the native side and not on the JS side.

## To Reproduce
I've created an example repo to reproduce the error in 0.59.0-rc.3: https://github.com/cjroth/react-native-fetch-memory-leak. You can copy the App.js and see that it happens in previous versions as well.

## Expected Behavior
After each fetch() call that is no longer referenced, the memory should return to what it was previously.

## Code Example
https://github.com/cjroth/react-native-fetch-memory-leak

## Environment
```
    System:
      OS: macOS 10.14.1
      CPU: (4) x64 Intel(R) Core(TM) i7-4650U CPU @ 1.70GHz
      Memory: 16.96 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 11.10.0 - /usr/local/bin/node
    SDKs:
      iOS SDK:
    IDEs:
    npmPackages:
      react: 16.8.3 => 16.8.3 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-rename: 2.4.0
```

screen shot 2019-03-06 at 9 22 40 pm

janicduplessis:
## Summary

Our Blob implementation was very problematic because it didn't release its underlying resource when the JS instance was dealocated. The main issue is that the fetch polyfill uses blobs by default if the module is available, which causes large memory leaks.

This fixes it by using the new jsi infra to attach a `jsi::HostObject` (`BlobCollector`)  to `Blob` instances. This way when the `Blob` is collected, the `BlobCollector` also gets collected. Using the `jsi::HostObject` dtor we can schedule the cleanup of native resources. This is definitely not the ideal solution but otherwise it would require rewriting the whole module using TurboModules + jsi. 

Fixes #23801, #20352, #21092

## Changelog

[General] [Fixed] - [Blob] Release underlying resources when JS instance in GC'ed

## Test Plan

- Tested that creating a simple Blob instance results in resource allocation and deallocation. (`new Blob(['HI'])`)
- Tested that blob resources created by the fetch polyfill eventually get collected.

### DIFF
--- a/Libraries/Blob/RCTBlobCollector.mm
+++ b/Libraries/Blob/RCTBlobCollector.mm
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#import "RCTBlobCollector.h"
+
+#import <React/RCTBridge+Private.h>
+#import "RCTBlobManager.h"
+
+namespace facebook {
+namespace react {
+
+RCTBlobCollector::RCTBlobCollector(RCTBlobManager *blobManager, const std::string &blobId)
+: blobId_(blobId), blobManager_(blobManager) {}
+
+RCTBlobCollector::~RCTBlobCollector() {
+  RCTBlobManager *blobManager = blobManager_;
+  NSString *blobId = [NSString stringWithUTF8String:blobId_.c_str()];
+  dispatch_async([blobManager_ methodQueue], ^{
+    [blobManager remove:blobId];
+  });
+}
+
+void RCTBlobCollector::install(RCTBlobManager *blobManager) {
+  RCTCxxBridge *cxxBridge = (RCTCxxBridge *)blobManager.bridge;
+  if (cxxBridge.runtime == nullptr) {
+    return;
+  }
+  [cxxBridge dispatchBlock:^{
+    jsi::Runtime &runtime = *(jsi::Runtime *)cxxBridge.runtime;
+    runtime.global().setProperty(
+      runtime,
+      "__blobCollectorProvider",
+      jsi::Function::createFromHostFunction(
+        runtime,
+        jsi::PropNameID::forAscii(runtime, "__blobCollectorProvider"),
+        1,
+        [blobManager](jsi::Runtime& rt, const jsi::Value& thisVal, const jsi::Value* args, size_t count) {
+          auto blobId = args[0].asString(rt).utf8(rt);
+          auto blobCollector = std::make_shared<RCTBlobCollector>(blobManager, blobId);
+          return jsi::Object::createFromHostObject(rt, blobCollector);
+        }
+      )
+    );
+  } queue:RCTJSThread];
+}
+
+} // namespace react
+} // namespace facebook
--- a/Libraries/Blob/RCTBlobCollector.mm
+++ b/Libraries/Blob/RCTBlobCollector.mm
@@ -25,11 +25,11 @@ RCTBlobCollector::~RCTBlobCollector() {
 }
 
 void RCTBlobCollector::install(RCTBlobManager *blobManager) {
-  RCTCxxBridge *cxxBridge = (RCTCxxBridge *)blobManager.bridge;
-  if (cxxBridge.runtime == nullptr) {
-    return;
-  }
+  __weak RCTCxxBridge *cxxBridge = (RCTCxxBridge *)blobManager.bridge;
   [cxxBridge dispatchBlock:^{
+    if (!cxxBridge || cxxBridge.runtime == nullptr) {
+      return;
+    }
     jsi::Runtime &runtime = *(jsi::Runtime *)cxxBridge.runtime;
     runtime.global().setProperty(
       runtime,
--- a/Libraries/Blob/BlobManager.js
+++ b/Libraries/Blob/BlobManager.js
@@ -14,7 +14,7 @@ const Blob = require('Blob');
 const BlobRegistry = require('BlobRegistry');
 const {BlobModule} = require('NativeModules');
 
-import type {BlobData, BlobOptions} from 'BlobTypes';
+import type {BlobData, BlobOptions, BlobCollector} from 'BlobTypes';
 
 /*eslint-disable no-bitwise */
 /*eslint-disable eqeqeq */
@@ -29,6 +29,20 @@ function uuidv4(): string {
       v = c == 'x' ? r : (r & 0x3) | 0x8;
     return v.toString(16);
   });
+}
+
+// Blob collector is a jsi::HostObject that is used by native to know
+// when the a Blob instance is deallocated. This allows to free the
+// underlying native resources. This is a hack to workaround the fact
+// that the current bridge infra doesn't allow to track js objects
+// deallocation. Ideally this would use TurboModules and the whole Blob
+// object should be a jsi::HostObject.
+function createBlobCollector(blobId: string): BlobCollector | null {
+  if (global.__blobCollectorProvider == null) {
+    return null;
+  } else {
+    return global.__blobCollectorProvider(blobId);
+  }
 }
 
 /**
@@ -94,7 +108,18 @@ class BlobManager {
    */
   static createFromOptions(options: BlobData): Blob {
     BlobRegistry.register(options.blobId);
-    return Object.assign(Object.create(Blob.prototype), {data: options});
+    return Object.assign(Object.create(Blob.prototype), {
+      data:
+        // Reuse the collector instance when creating from an existing blob.
+        // This will make sure that the underlying resource is only deallocated
+        // when all blobs that refer to it are deallocated.
+        options.__collector == null
+          ? {
+              ...options,
+              __collector: createBlobCollector(options.blobId),
+            }
+          : options,
+    });
   }
 
   /**
--- a/Libraries/Blob/BlobManager.js
+++ b/Libraries/Blob/BlobManager.js
@@ -31,12 +31,13 @@ function uuidv4(): string {
   });
 }
 
+// **Temporary workaround**
+// TODO(#24654): Use turbomodules for the Blob module.
 // Blob collector is a jsi::HostObject that is used by native to know
 // when the a Blob instance is deallocated. This allows to free the
 // underlying native resources. This is a hack to workaround the fact
 // that the current bridge infra doesn't allow to track js objects
-// deallocation. Ideally this would use TurboModules and the whole Blob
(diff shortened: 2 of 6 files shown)


janicduplessis:
TODO:
- Android
- Investigate if code sharing is possible / worth it

Libraries/Blob/RCTBlobCollector.mm
```diff
@@ -0,0 +1,52 @@
...
+RCTBlobCollector::~RCTBlobCollector() {
+  RCTBlobManager *blobManager = blobManager_;
+  NSString *blobId = [NSString stringWithUTF8String:blobId_.c_str()];
+  dispatch_async([blobManager_ methodQueue], ^{
+    [blobManager remove:blobId];
+  });
+}
+
+void RCTBlobCollector::install(RCTBlobManager *blobManager) {
+  RCTCxxBridge *cxxBridge = (RCTCxxBridge *)blobManager.bridge;
```
janicduplessis: This is a bit hacky to say the least but it was the only way I found to use jsi from a old `RCTBridge` module

Libraries/Blob/RCTBlobCollector.mm
```diff
@@ -0,0 +1,52 @@
...
+namespace react {
+
+RCTBlobCollector::RCTBlobCollector(RCTBlobManager *blobManager, const std::string &blobId)
+: blobId_(blobId), blobManager_(blobManager) {}
+
+RCTBlobCollector::~RCTBlobCollector() {
+  RCTBlobManager *blobManager = blobManager_;
+  NSString *blobId = [NSString stringWithUTF8String:blobId_.c_str()];
+  dispatch_async([blobManager_ methodQueue], ^{
+    [blobManager remove:blobId];
```
fkgozali: Instead of doing this dance of using blobId as string, then asking nativemodule to clean it up, it's probably better long term to send down direct pointer to a blob (as jsi::Value) to JS, which can get cleaned up automatically by the JS runtime. See Fabric's UIManager method/impl as an example.

cpojer:
@janicduplessis mind addressing @fkgozali's comments? I'd really love for this fix to go in.

janicduplessis:
@fkgozali Do you have any recommendations on how to make this safe / less hacky?

What about adding a method on native modules a bit like `setBridge` but for the jsi::Runtime. I imagine that the turbomodule infra already does something like this.

fkgozali:
> What about adding a method on native modules a bit like setBridge but for the jsi::Runtime. I imagine that the turbomodule infra already does something like this.

TurboModule bypasses the bridge and interacts with Runtime executor directly during initialization. I don't think there's a safe way to get `jsi::Runtime` via the bridge to be accessed by a NativeModule. Further, the bridge will go away some time in the future after TurboModule/Fabric are done.

I think it's probably worth holding off on this change until TurboModule is ready? I think it'd be beneficial to refactor this module to take advantage of `jsi::HostObject` instead of keeping your own object via string IDs like today.

I'm closing this for now, but let's revisit in the future.

janicduplessis:
@fkgozali Do you have any timeframe for when turbomodules will be ready? I definitely agree that this module would be a lot cleaner with the new infra. My only concern is that the current blob implementation causes a significant memory leak (pretty much every response of `fetch` requests will leak).

I assume fb doesn't use the Blob module so it isn't an issue but it is currently included by default in OSS so I think we'd need to fix this ASAP or disable the blob module.

fkgozali:
> Do you have any timeframe for when turbomodules will be ready?

It'll be some time later this year, Q3/Q4, but if you're just targeting iOS only, we're currently testing in prod. So in theory you could start using it yourself (just need to write some of the "codegen" output by hand), it's just not enabled by default, yet.

> I assume fb doesn't use the Blob module so it isn't an issue but it is currently included by default in OSS so I think we'd need to fix this ASAP or disable the blob module.

We actually use blob module, but just not the full features I think. So the risk of getting random crash due to unsafe runtime access off bridge can affect FB as well.

> My only concern is that the current blob implementation causes a significant memory leak (pretty much every response of fetch requests will leak).

Do you know how bad the leak is? Is it like MBs/minute, or more like MBs/day?

janicduplessis:
@fkgozali

> It'll be some time later this year, Q3/Q4, but if you're just targeting iOS only, we're currently testing in prod. So in theory you could start using it yourself (just need to write some of the "codegen" output by hand), it's just not enabled by default, yet.

The issue affects both iOS and Android so the fix would have to work for both, I just implemented it on iOS initially to get feedback.

> We actually use blob module, but just not the full features I think. So the risk of getting random crash due to unsafe runtime access off bridge can affect FB as well.

I see, then you are probably affected by the leak. Note that the fix is safe, at worst the blob collector won't be created if for some reason the `jsi::Runtime` is not available, and `__blobCollectorProvider` is not injected.

> Do you know how bad the leak is? Is it like MBs/minute, or more like MBs/day?

I haven't measured personally but there are some info in this issue https://github.com/facebook/react-native/issues/23801, looks more like MBs / minute depending on network activity.

fkgozali:
> We actually use blob module, but just not the full features I think. So the risk of getting random crash due to unsafe runtime access off bridge can affect FB as well.

I just did an audit and it looks like we use it only for development purpose for now, so if this is causing bad leaks for apps today, maybe it's ok to have this fix for the short term.

> Note that the fix is safe, at worst the blob collector won't be created if for some reason the jsi::Runtime is not available, and __blobCollectorProvider is not injected.

If you're fine with this hack for now, I guess you can have this in, but let's just document via comments so that it won't bite us in the future. When TurboModule is ready, we should really refactor this entire module to take advantage of proper `jsi::HostObject` lifetime handling.

janicduplessis:
@fkgozali  The hack is more or less documented here https://github.com/facebook/react-native/pull/24405/files#diff-2bf13477704f772255c64e0beb652478R34, I can also create an issue to properly refactor the module when turbomodules are available.

I've shipped this to production for a couple weeks and haven't had issues so I'm confident that at least it doesn't causes crashes. I'll look at implementing it for android if you are fine with shipping this as a temporary solution.

fkgozali:
I'll let @cpojer pull this in.

> The hack is more or less documented

Could you add 1 sentence that "This is a **temporary** workaround."?

Libraries/Blob/RCTBlobCollector.mm
```diff
@@ -0,0 +1,52 @@
...
+  });
+}
+
+void RCTBlobCollector::install(RCTBlobManager *blobManager) {
+  RCTCxxBridge *cxxBridge = (RCTCxxBridge *)blobManager.bridge;
+  if (cxxBridge.runtime == nullptr) {
```
janicduplessis: Looking at this code again I can probably add a bit more safety: use a `__weak` ref for `cxxBrige` and move the `cxxBridge.runtime == nullptr` inside the dispatched block.

fkgozali:
> Could you add 1 sentence that "This is a temporary workaround."?

I guess you already hinted it, but let's be explicit with **temporary**

janicduplessis:
https://github.com/facebook/react-native/pull/24405/commits/92a9f028c6b66aa83f7269df1a5726994bb92214 tweaked the wording to make it clear and created #24654 to track it.

cpojer:
@janicduplessis let me know when this is ready to be pulled in. It still has [WIP] in the title and you mentioned Android support.

janicduplessis:
@cpojer I guess adding android in a separate PR would be better, just need to cleanup something then this will be ready.

ebollens:
As a React Native user that is being directly impacted by these leaks, we're very appreciative of the team's quick movement to get a temporary fix in place until turbomodules is ready.

ebollens:
As a user that is being directly impacted by these leaks, we're very appreciative of the team's quick movement to get a temporary fix in place until turbomodules is ready.

janicduplessis:
@cpojer Should be ready for import!

cpojer:
Seems like this failed with `react-native/Libraries/Blob/RCTBlobCollector.h:8:9: fatal error: 'jsi/jsi.h' file not found` and has to be reverted. Mind sending this again with a fix?

hramos:
The changes from this PR were reverted in 05baf6272143667694585a14fb59657fdc93c3b1 due to the issue Christoph mentioned.
The JSI headers were moved *right* before this PR was opened, in https://github.com/facebook/react-native/commit/094f221a0c2d24f16a9540e4f176783843ec4fd9#diff-e39171f096fb9caf46141fa6026d145e. If you rebase past that commit and apply a fix, we should be able to get these changes merged again.

janicduplessis:
@hramos @cpojer I rebased on master but it still builds properly in OSS. I checked the diff that moved jsi headers but it doesn't seem like it requires updating the way jsi is imported (I checked it is still imported as `<jsi/jsi.h>` elsewhere).

Do you know if it fails for a specific project or all RN projects internally? Maybe it would be better if someone at fb can debug the build issue since I can't repro in OSS.

tlupo:
Hello @janicduplessis !

As mentioned on the Android version of this PR (https://github.com/facebook/react-native/pull/24767), this fix seems to work on iOS. By profiling with Instruments (Allocations tool), I can see a stable memory usage (around 60 Mb).

Without this PR, I observed a progressive increase of the memory usage (Malloc etc. on the Network layer) until the crash of the application. 

However, I don't understand why but I still receive "Low memory warning" (after 2/3 hours of app usage, even with this stable memory usage) and an app crash (without crash log). 

With your investigation on this part, do you have any information (a better way to debug memory usage or something else) to help me ?

Thank you very much for your help.

janicduplessis:
@tlupo Sadly I don't know of other ways, maybe it could be something else on the device that is eating up memory, eventually causing your app to be killed? I only did limited memory profiling since the leak was very obvious.

PR #24411, opened 2019-04-11
Fix: RefreshControl in FlatList makes borderWidth not working

### LINKED ISSUE #22752: RefreshControl in FlatList makes borderWidth not working

## Environment
```
  System:
    OS: Windows 8.1
    CPU: (8) x64 Intel(R) Core(TM) i7-4710HQ CPU @ 2.50GHz
    Memory: 8.41 GB / 15.92 GB
  Binaries:
    Yarn: 1.12.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    Android Studio: Version  2.3.0.0 AI-162.3871768
```

## Description
Describe your issue in detail. Include screenshots if needed. If this is a regression, let us know.

## Reproducible Demo
1. `react-native init MyTestingProject`
2. `cd MyTestingProject`
3. `react-native run-android`
4. Create a normal `<FlatList>`:

```javascript
import { Text, FlatList, RefreshControl } from 'react-native';

  state = {isLoading: false};

  render() {
    return (
      <FlatList
        style={{
          flex: 1,
          alignSelf: "stretch",
          borderWidth: 2,
          borderColor: "red"
        }}
        data={["a", "b", "c"]}
        renderItem={({ item }) => (
          <Text>
            {item}
          </Text>
        )}
        refreshControl={
          <RefreshControl
            colors={["red", "yellow"]}
            refreshing={this.state.isLoading}
            onRefresh={() => this.reload()}
          />
        }
      />
    );
  }
}
```

5. We have no **borders**! Now remove the `refreshControl` to see the **borders**:

```javascript
<FlatList
  style={{
    flex: 1,
    alignSelf: "stretch",
    borderWidth: 2,
    borderColor: "red"
  }}
  data={["a", "b", "c"]}
  renderItem={({ item }) => (
    <Text>
      {item}
    </Text>
  )}
/>;
```

Nizarius:
## Summary

Fixes #22752

On line 1021 you are passing base style to props: 
`style: [baseStyle, this.props.style],`

Explicitly passing base style to ScrollView just overrides this line and doesn't let developers to customise style of any inheritors of ScrollView (not only FlatList) with custom RefreshControl.

So this line (1113) seems to be removed. 

## Changelog

[GENERAL] [Fixed] - fix of Android's bug that doesn't let override ScrollView's Style with custom RefreshControl. 

## Test Plan
Use this code snippet to test that passing RefreshController doesn't break styling of ScrollView on Android.
```
type Props = {};
  state = { isLoading: false };

  reload = () => {
    this.setState({ isLoading: true });

    setTimeout(() => {
      this.setState({ isLoading: false });
    }, 2000);
  };
  render() {
    return (
      <ScrollView
        style={{
          flex: 1,
          alignSelf: "stretch",
          borderWidth: 10,
          borderColor: "red"
        }}
        refreshControl={
          <RefreshControl
            colors={["red", "yellow"]}
            refreshing={this.state.isLoading}
            onRefresh={() => this.reload()}
          />
        }
      >
        <View
          style={{
            width: "100%",
            height: 300
          }}
        >
          <Text>3335</Text>
        </View>
        <View
          style={{
            width: "100%",
            height: 300
          }}
        />
        <View
          style={{
            width: "100%",
            height: 300
          }}
        />
      </ScrollView>
    );
  }
}
```

The same for FlatList: 
```
type Props = {};
  state = { isLoading: false };

  reload = () => {
    this.setState({ isLoading: true });

    setTimeout(() => {
      this.setState({ isLoading: false });
    }, 2000);
  };
  render() {
    return (
       <FlatList
        style={{
          flex: 1,
          alignSelf: "stretch",
          borderWidth: 10,
          borderColor: "red"
        }}
        data={["a", "b", "c"]}
        renderItem={({ item }) => (
          <View style={{ height: 500 }}>
            <Text>{item}</Text>
          </View>
        )}
        refreshControl={
          <RefreshControl
            colors={["red", "yellow"]}
            refreshing={this.state.isLoading}
            onRefresh={() => this.reload()}
          />
        }
      />
    );
  }
}
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "flow.pathToFlow": "${workspaceFolder}/node_modules/.bin/flow",
+  "javascript.validate.enable": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-  "flow.pathToFlow": "${workspaceFolder}/node_modules/.bin/flow",
-  "javascript.validate.enable": false
-}
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -1128,9 +1128,7 @@ class ScrollView extends React.Component<Props, State> {
         // AndroidSwipeRefreshLayout and use flex: 1 for the ScrollView.
         // Note: we should split props.style on the inner and outer props
         // however, the ScrollView still needs the baseStyle to be scrollable
-
         const {outer, inner} = splitLayoutProps(flattenStyle(props.style));
-
         return React.cloneElement(
           refreshControl,
           {style: [baseStyle, outer]},
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -25,6 +25,7 @@ const invariant = require('invariant');
 const processDecelerationRate = require('./processDecelerationRate');
 const requireNativeComponent = require('../../ReactNative/requireNativeComponent');
 const resolveAssetSource = require('../../Image/resolveAssetSource');
+const splitLayoutProps = require('../../StyleSheet/splitLayoutProps');
 
 import type {
   PressEvent,
@@ -1128,12 +1129,14 @@ class ScrollView extends React.Component<Props, State> {
         // Note: we should only apply props.style on the wrapper
         // however, the ScrollView still needs the baseStyle to be scrollable
 
+        const {outer, inner} = splitLayoutProps(flattenStyle(props.style));
+
         return React.cloneElement(
           refreshControl,
-          {style: props.style},
+          {style: outer},
           <ScrollViewClass
             {...props}
-            style={baseStyle}
+            style={inner}
             // $FlowFixMe
             ref={this._setScrollViewRef}>
             {contentContainer}
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -1126,17 +1126,17 @@ class ScrollView extends React.Component<Props, State> {
         // On Android wrap the ScrollView with a AndroidSwipeRefreshLayout.
         // Since the ScrollView is wrapped add the style props to the
         // AndroidSwipeRefreshLayout and use flex: 1 for the ScrollView.
-        // Note: we should only apply props.style on the wrapper
+        // Note: we should split props.style on the inner and outer props
         // however, the ScrollView still needs the baseStyle to be scrollable
 
         const {outer, inner} = splitLayoutProps(flattenStyle(props.style));
 
         return React.cloneElement(
           refreshControl,
-          {style: outer},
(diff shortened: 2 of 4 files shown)


Nizarius:
@janicduplessis yeah, you're right. In fact, margin properties only work on AndroidSwipeRefreshLayout and padding properties - only on ScrollView. So padding or margin doesn't get applied twice. But there is such a problem with opacity. 
So the ideal solution seems something like that:
1) Extract all margin properties ('margin', 'marginRight' etc) and opacity from this.props.style and apply it to AndroidSwipeRefreshLayout (with baseStyle).
2) Pass other style properties (padding, borders etc) to ScrollView (with baseStyle).
This can prevent unexpected behaviour in the future as we won't duplicate any properties to both views.
I don't know if you have some utility functions for extracting properties' family ("extract all margin properties" or something similar) to make code cleaner and more declarative.

What do you think?

janicduplessis:
I don't think we have anything currently to do this. I have something like this in an app though:

```js
const OUTER_PROPS = {
  margin: true,
  marginHorizontal: true,
  marginVertical: true,
  marginBottom: true,
  marginTop: true,
  marginLeft: true,
  marginRight: true,
  flex: true,
  flexGrow: true,
  flexShrink: true,
  flexBasis: true,
  alignSelf: true,
  height: true,
  minHeight: true,
  maxHeight: true,
  width: true,
  minWidth: true,
  maxWidth: true,
  position: true,
  left: true,
  right: true,
  bottom: true,
  top: true,
};

export function splitLayoutProps(props) {
  const inner = {};
  const outer = {};
  Object.keys(props).forEach(k => {
    if (OUTER_PROPS[k] === true) {
      outer[k] = props[k];
    } else {
      inner[k] = props[k];
    }
  });
  return { outer, inner };
}

```

 You could also use `StyleSheet.flatten` to make sure to have an object from the style prop.

@sahrens What do you think about this?

Nizarius:
@janicduplessis seems nice. We need to define where to place this utility function.

cpojer:
What's the status of this PR?

Nizarius:
@cpojer I suggest, we are waiting comments for the best code styling. The implementation itself will be fast.

sahrens:
Yeah let's add a `splitLayoutProps` export on `StyleSheet` and use that.

Nizarius:
@sahrens I've added splitLayoutProps to export and use it.
I'm also explicitly passing baseStyle to both wrapper and container style to be sure that we don't break anything:
```
return React.cloneElement(
          refreshControl,
          {style: [baseStyle, outer]},
          <ScrollViewClass
            {...props}
            style={[baseStyle, inner]}
            // $FlowFixMe
            ref={this._setScrollViewRef}>
            {contentContainer}
          </ScrollViewClass>,
        );
```
But as I check it works without it. It's on your side. So it's up to you.

Nizarius:
Any news, guys?

Nizarius:
@cpojer yep, will do it soon

Nizarius:
@cpojer done

Issue #14063, opened 2017-05-19
withCredentials flag in XHRs should default to "true"


DanielZlotin:
### Description

react-native 0.44 introduced [`withCredentials` flag in XHRs](https://github.com/facebook/react-native/commit/454ab8fc238da16f9c3986d9a64ea8e716ac0bac#diff-10faae6fe100efe7101c16fe93478084R122), which, if not specified in every `fetch` request, defaults to `false`.
This change conflicts with the default behavior in native. In the iOS native SDK and the Android native SDK, when making a native HTTP request, cookies are sent by default.
We rarely have agreement between the platforms, but for the last 10 years they both agree on this security model for apps.
This greatly affects projects relying on cookies with their requests.

### Additional Information

* React Native version: 0.44
* Platform: both
* Development Operating System: macOS

talkol:
According to the [commit](https://github.com/facebook/react-native/commit/454ab8fc238da16f9c3986d9a64ea8e716ac0bac#diff-10faae6fe100efe7101c16fe93478084R122) description, the reason for this breaking change is to be 
> "consistent with the default behavior of XHR on web for cross-site requests".

This doesn't make much sense to me. These are native apps. Native apps don't have cross-site concerns. The JS bundle is not served from a domain like the web.

The security model for native mobile apps has been established a long time ago. The standard native API's for making HTTP requests in iOS and Android send cookies by default. Changing this behavior to conform to websites just because we're using JavaScript is strange.

In addition, there's a big problem with the override mechanism. This is a breaking change, and now we have apps in production that we cannot release due to this change.

The override mechanism according to the commit is: 
> "Developers can restore the previous behavior by passing `true` for XHR's `withCredentials` argument". 

This makes the assumption that we can control the parameters for every request our app makes. Consider that we're using a 3rd party GraphQL client library that makes the `fetch` requests for us. This library is out of our control meaning we can't use the override mechanism. To support backwards compatibility for existing apps that are in production when introducing these types of changes, the minimum is to allow a global override when the app starts.

@rigdern

rigdern:
I'm sorry that my commit is causing issues for you.

> The standard native API's for making HTTP requests in iOS and Android send cookies by default.

Which Android API are you referring to?

I think there are several questions to think about here:

#### 1. What should RN's `withCredentials` default to?

The answer is not obvious to me. Ignoring the web, different APIs I'm familiar with have made different choices regarding the default for sending and saving cookies:
  - Libraries that disable cookies by default:
[...]

PR #24418, opened 2019-04-11
Remove vendored fetch polyfill, update to whatwg-fetch@3.0

janicduplessis:
## Summary

The original reason for vendoring the fetch polyfill was to remove the default blob response type but this was reverted.

Here's a little history around the fetch polyfill and the blob issue:

- Original commit introducing the vendored polyfill: #19333, the goal was to fix a memory leak because our blob implementation doesn't release resources automatically. Not an ideal fix but since the issue was pretty severe and the infra for a proper fix was not in place.
- This introduced an issue when downloading images using `fetch` which was fixed by #22063 which re-added the default blob content type. However that re-introduced the original fetch memory leak.
- We have better infra now with jsi and I was able to get blob deallocation working, see #24405

Currently the vendored fetch polyfill is useless since it was changed back to the original version. We can just use the npm version again. I also updated to 3.0 which brings better spec compliance and support for cancellation via `AbortController`, https://github.com/github/fetch/releases/tag/v3.0.0.

## Changelog

[General] [Changed] - Remove vendored fetch polyfill, update to whatwg-fetch@3.0

## Test Plan

Tested using fetch in RN tester to make sure everything still works.


Libraries/Core/setUpGlobals.js
```diff
@@ -21,6 +21,10 @@ if (global.window === undefined) {
   global.window = global;
 }
 
+if (global.self === undefined) {
```
janicduplessis: whatwg-fetch@3 relies on the `self` global

TheSavior: Maybe worth putting this comment in the code?

janicduplessis: I didn’t since it’s not specific to the fetch polyfill and is just a standard dom api like Window (https://developer.mozilla.org/en-US/docs/Web/API/Window/self). I don’t mind adding it though if you prefer.

cpojer:
Yay, thank you for doing this and removing code that we have to maintain! I'm going to wait until we merge the blob memory leak PR before landing this one.

janicduplessis:
@cpojer We don't actually have to wait for the blob memory leak PR to land, the leak happens with or without this. That's fine if you prefer to wait though :)

cpojer:
Oh good to know, I'll get to it soon then :)

janicduplessis:
@cpojer Fixed a conflict, wanna try landing this again?

zhongwuzw:
Ahha, after mentioned by @zhigang1992, I found fetch is to be managed by npm directly, I created a PR https://github.com/facebook/react-native/pull/24726 to fix fetch `withCredentials` issue, that is because security policy is different between web and iOS/Android, detailed discussion can see https://github.com/facebook/react-native/issues/14063. I don't know wether we have another differences. Any opinion about this?

hramos:
This might take a while longer to land. Targeting next week, at the earliest.

mislav:
Hello, I'm one of the authors and maintainers of whatwg-fetch 👋 

I'm happy to learn that adding `self` to React Native will make this problem go away https://github.com/github/fetch/issues/657

I have also found this puzzling comment:
https://github.com/facebook/react-native/blob/ddce74388a5662b5307582ca2e05710224c74ba0/Libraries/Core/Devtools/symbolicateStackTrace.js#L29-L41

Can anyone explain is this still relevant? Is there something we can tweak in whatwg-fetch to make this workaround obsolete?

Finally, what is the reason React Native bundles a `window.fetch` polyfill? Are any of the browser targets for React Native apps old browsers that have no native support for `window.fetch`?

Thank you!

janicduplessis:
Hi @mislav!

I don't have context on the comment so I can't help there.

We use the fetch polyfill because React Native does not run in a browser environment, it runs in JavaScriptCore (or possibly another JS engine) so it doesn't have any DOM apis built in. To avoid having to write a native implementation for both `XMLHttpRequest` and `fetch` we have one for `XMLHttpRequest` only and use a fetch polyfill. So this means the fetch polyfill will always be used in React Native.

PR #24419, opened 2019-04-11
Add support for cancelling fetch requests with AbortController

### LINKED ISSUE #18115: AbortController is missing
### Is this a bug report?

Yes. AbortController is a new spec (you can find info here: https://developers.google.com/web/updates/2017/09/abortable-fetch)

Yes

### Environment

```
Environment:
  OS: Linux 4.13
  Node: 8.9.4
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Not Found

  react: ^16.0.0 => 16.2.0
  react-native: ^0.53.0 => 0.53.0
```

### Steps to Reproduce

1. Istantiate an AbortController
2. Write a fetch request passing the signal
3. Try to abort the fetch request

### Expected Behavior

The requested behaviour is to abort the fetch request.

### Actual Behavior

Nothing happens

### Reproducible Demo

Something like this:

```
const controller = new AbortController();
const signal = controller.signal;

setTimeout(() => controller.abort(), 5000);

fetch(url, { signal }).then(response => {
  return response.text();
}).then(text => {
  console.log(text);
});
```

My question is a technical one: is the react native core that needs to be updated in order to support this? Or is it something that babel can patch with a new version?

janicduplessis:
## Summary

This adds https://github.com/mysticatea/abort-controller to polyfill [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController). This is used to cancel requests when using `fetch`.

This also updates `event-target-shim` to 5.0 to make sure we only have one version of this dependency. This updates required adding a polyfill for `console.assert` which is used by the new version. I made one based on https://github.com/gskinner/console-polyfill/blob/master/console.js#L74.

The polyfill is very small, especially since we already use `event-target-shim` so I think it makes sense to include in core.

Depends on #24418 so that the fetch polyfill supports the `signal` parameter.

Fixes #18115

## Changelog

[General] [Added] - Add support for cancelling fetch requests with AbortController

## Test Plan

Added a RNTester example of cancelling fetch requests.


matthargett:
once this (and its prerequisite PR) is merged, I'd like to propose it for backport to 0.59. @dysonpro fixed this in our apps by injecting/overriding RN's fetch with yetch (thanks @jayphelps !) and some other bits of magic.

we'd just need to make sure this doesn't introduce the same memory leak issues like the last time this abort controller support was attempted.

janicduplessis:
@matthargett What memory leak are you talking about? Is it the one I described in #24418 or something else I should be aware of?

cpojer:
@matthargett we aren't going to backport new features to 0.59, let's keep this work focused on 0.60+.

PR #24425, opened 2019-04-12
Update detox to 12.2.0

friederbluemle:
## Summary

This updates `detox` from `12.1.1` to `12.2.0`.
It includes (amongst other things) the following fixes, as described on the [12.1.4 release page](https://github.com/wix/Detox/releases/tag/12.1.4):

- hotfix: redundant logs in the folder
- fix: stabilize simulator log recording

It also updates the detox test script and config in `packages.json` (which was previously missed when detox was updated to 12.1.x). It will remove the following deprecation warnings:

```
detox[6416] WARN:  [test.js] Deprecation warning: "file" and "specs" support will be dropped in the next Detox version.
detox[6416] WARN:  [test.js] Please edit your package.json according to the migration guide: https://wix.to/I0DOAK0
```

Instead of extra preprocessing the test folder by `detox`, it will now be passed straight to `jest`.

More information in the [migration guide](https://github.com/wix/Detox/blob/master/docs/Guide.Migration.md).

## Changelog

[General] [Changed] - Update detox to 12.2.0

## Test Plan

Successfully ran:

```
yarn build-ios-e2e
yarn test-ios-e2e
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "test-android-unit": "yarn run docker-build-android && yarn run test-android-run-unit",
     "test-android-e2e": "yarn run docker-build-android && yarn run test-android-run-e2e",
     "build-ios-e2e": "detox build -c ios.sim.release",
-    "test-ios-e2e": "detox test -c ios.sim.release"
+    "test-ios-e2e": "detox test -c ios.sim.release RNTester/e2e"
   },
   "peerDependencies": {
     "react": "16.8.3"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -147,7 +147,7 @@
   "detox": {
     "test-runner": "jest",
     "runner-config": "RNTester/e2e/config.json",
-    "specs": "RNTester/e2e",
+    "specs": "",
```
gengjiawen: Can this be deleted since it's deprecated ?

friederbluemle: Good question, that's what I first thought as well. However, for backwards-compatibility reasons, they suggest to keep the field set to an empty string. We can only remove it once the library gets updated to `13.0.0` I assume.

There is a whole section with additional information in the [migration guide](https://github.com/wix/Detox/blob/master/docs/Guide.Migration.md#migrating-from-detox-120x-to-121x) about it:

> For the curious ones, who want to know more why we should use an empty string ("") instead of deleting "specs" [...]

friederbluemle:
Rebased and updated to `12.3.0`.

PR #24436, opened 2019-04-13
Specify correct terminal to start the packager

lucasbento:
## Summary

This PR specifies the correct terminal to start the Metro bundler as stated in https://github.com/react-native-community/cli/issues/303.

## Changelog

[iOS] [Changed] - Start the packager on the correct terminal.

## Test Plan

To make sure that this works just run it from within Xcode and it should go to your default terminal, you can also try setting the `REACT_TERMINAL` environment variable to make sure that the packager will open on the desired terminal.

### DIFF
--- a/React/React.xcodeproj/project.pbxproj
+++ b/React/React.xcodeproj/project.pbxproj
@@ -3975,7 +3975,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] && [ \"$CONFIGURATION\" == \"Debug\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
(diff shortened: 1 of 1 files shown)


thymikee:
I actually couldn't get this to work (legacy build, fish shell, latest macOS). Turns out that `printenv` inside Xcode results in completely different envs and `TERM_PROGRAM` is not there. Ideas?

lucasbento:
@thymikee: what happens in the end?

I'm on macOS 10.14, Hyper installed, legacy build it is working here, maybe Hyper did set it up.

I also trying running the app from fishshell.com and it has the `TERM_PROGRAM` on it, not entirely sure how I can reproduce your issue.

thymikee:
It doesn't open the terminal at all. I'm really not sure why that happens. I use iTerm. The `TERM_PROGRAM` is there on fish, zsh, bash and sh, just Xcode doesn't see it on my machine. If it works for you, I think it would be fine to default the `$terminal` to `Terminal.app` so it doesn't regress.

lucasbento:
Strange. That's a good idea though, I'll do it.

noahtallen:
Nifty! I'm curious - I just opened [this issue](https://github.com/react-native-community/cli/issues/317#issuecomment-483048597) on the CLI repo. Essentially, I have react imported through Pods, and _not_ at all through the Xcode project. Does this mean that this `shellScript` would not get run? Because it currently does not launch the packager for me on `react-native run-ios` for a simulator (though it appears that normal bundling during a release build works fine).

lucasbento:
@noahtallen: let's continue the conversation on https://github.com/react-native-community/cli/issues/317, I would like to fix that problem.

lucasbento:
@thymikee: I just added a check to verify if it found the default terminal or not, if not just open the packager without specifying it.

lucasbento:
@thymikee: sorry for the syntax error, thought I was testing this but in the end was testing another app.

I tried to check if I have `UseSanitizedBuildSystemEnvironment` set to `NO` but apparently I don't (I also don't recall setting it) and the Xcode still inherits the env variables.

If you define it as `NO`, does it work?

I'll try to check with a co-worker to see if I can reproduce that on their computer as well.

lucasbento:
Btw I tried to check with `defaults read-type com.apple.dt.Xcode UseSanitizedBuildSystemEnvironment` not sure if this is the correct syntax.

thymikee:
Just tried and still nope (inb4 I restarted Xcode)

lucasbento:
I'll check it here with some of my co-workers to see if I can reproduce that problem.

thymikee:
I've additionally tried this on a fresh project without pods, with `UseSanitizedBuildSystemEnvironment` disabled (and enabled as well). Still same results. Wonder if I set something weird years ago and now don't remember :D

lucasbento:
@thymikee: I tested on two other MacBooks and it's the same behaviour as yours, `TERM_PROGRAM` is not defined for some reason, so I guess I'm the one that may have done something weird in my machine to have it.

I'm wondering now if there's any way that we can pass the `TERM_PROGRAM` to the build phase script otherwise this will work only for a few people (I assume) and for the ones that use the CLI.

lucasbento:
@thymikee: the only way I see atm to improve this part is to have a way to customize those env variables through a file so that the script can read from it instead of trying to get the global ones, unfortunately this PR will not change the behaviour for people that use Xcode.

PR #24457, opened 2019-04-15
Fix exception when attempt to load image from file system

harry-nguyen-88:
## Summary

When trying to load image using method "RCTImageFromLocalAssetURL", if URL's path in file system representation return null (in my case, imageUrl is base64 format returned from server so it's always null) then method "stringWithUTF8String" will raise an exception.

## Changelog

[iOS] [Fixed] - Fix exception when attempt to load image from file system

## Test Plan

If we have base64 image with imageURL is something like "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAASABIAAD/4QBYRXhpZg...." and using RCTImageFromLocalAssetURL to load it, we can load the correct image without any exception.

### DIFF
--- a/React/Base/RCTUtils.m
+++ b/React/Base/RCTUtils.m
@@ -741,8 +741,8 @@ UIImage *__nullable RCTImageFromLocalAssetURL(NSURL *imageURL)
   if (!image) {
     // Attempt to load from the file system
     const char* fileSystemCString = [imageURL fileSystemRepresentation];
-    if (fileSystemCString != nil) {
-      NSString *filePath = [NSString stringWithUTF8String:[imageURL fileSystemRepresentation]];
+    if (fileSystemCString != NULL) {
+      NSString *filePath = [NSString stringWithUTF8String:fileSystemCString];
       if (filePath.pathExtension.length == 0) {
         filePath = [filePath stringByAppendingPathExtension:@"png"];
       }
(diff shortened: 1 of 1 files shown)


React/Base/RCTUtils.m
```diff
@@ -740,11 +740,14 @@ BOOL RCTIsLocalAssetURL(NSURL *__nullable imageURL)
 
   if (!image) {
     // Attempt to load from the file system
-    NSString *filePath = [NSString stringWithUTF8String:[imageURL fileSystemRepresentation]];
-    if (filePath.pathExtension.length == 0) {
-      filePath = [filePath stringByAppendingPathExtension:@"png"];
+    const char* fileSystemCString = [imageURL fileSystemRepresentation];
+    if (fileSystemCString != nil) {
```
shergin: It's not idiomatic to compare raw C pointer to nil, we should use null instead or rely on implicit conversion to bool.

React/Base/RCTUtils.m
```diff
@@ -740,11 +740,14 @@ BOOL RCTIsLocalAssetURL(NSURL *__nullable imageURL)
...
+      NSString *filePath = [NSString stringWithUTF8String:[imageURL fileSystemRepresentation]];
```
shergin: Why not use `fileSystemCString` here?

React/Base/RCTUtils.m
harry-nguyen-88: Agreed! I will fix it!

React/Base/RCTUtils.m
harry-nguyen-88: My bad! I forgot to review the code there.

PR #24463, opened 2019-04-16
bump android gradle plugin to 3.4.0

dulmandakh:
## Summary

This PR bumps Android Gradle Plugin to 3.4.0, which enables R8 shrinker by default and improves build performance significantly.

Disabled R8 for ReactAndroid because it'll strip out AndroidX and other libraries bundled in ReactAndroid.

## Changelog

[Android] [Changed] - bump Android Gradle plugin to 3.4.0

## Test Plan

CI is green, and everything works as normal.

### DIFF
--- a/ReactAndroid/gradle.properties
+++ b/ReactAndroid/gradle.properties
@@ -23,3 +23,4 @@ GLOG_VERSION=0.3.5
 
 android.useAndroidX=true
 android.enableJetifier=true
+android.enableR8=false
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:3.3.2")
+        classpath("com.android.tools.build:gradle:3.4.0")
         classpath("de.undercouch:gradle-download-task:3.4.3")
 
         // NOTE: Do not place your application dependencies here; they belong
--- a/react.gradle
+++ b/react.gradle
@@ -148,6 +148,11 @@ afterEvaluate {
                 into ("merged_assets/${variant.name}/merge${targetName}Assets/out") {
                     from(jsBundleDir)
                 }
+
+                // Workaround for Android Gradle Plugin 3.4+ new asset directory
+                into ("merged_assets/${variant.name}/out") {
+                    from(jsBundleDir)
+                }
             }
 
             // mergeAssets must run first, as it clears the intermediates directory
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -13,7 +13,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.0'
+        classpath("com.android.tools.build:gradle:3.4.0")
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -13,7 +13,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.2'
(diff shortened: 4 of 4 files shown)


dulmandakh:
@gengjiawen I found that RNTester app is crashing because JS is not bundled in a release APK. Could you please help me with investigation.

gengjiawen:
> @gengjiawen I found that RNTester app is crashing because JS is not bundled in a release APK. Could you please help me with investigation.

I may have some time on weekends :)
Also cc @Kudo

Kudo:
@dulmandakh The root cause is that the building merged assets path is changed.
This patch might works for you.

```diff
diff --git a/react.gradle b/react.gradle
index 142a21c8a2..f941d0c0b5 100644
--- a/react.gradle
+++ b/react.gradle
@@ -148,6 +148,10 @@ afterEvaluate {
                 into ("merged_assets/${variant.name}/merge${targetName}Assets/out") {
                     from(jsBundleDir)
                 }
+                // Workaround for Android Gradle Plugin 3.4+ new asset directory
+                into ("merged_assets/${variant.name}/out") {
+                    from(jsBundleDir)
+                }
             }

             // mergeAssets must run first, as it clears the intermediates directory
```

dulmandakh:
@Kudo thank you, it works.

dulmandakh:
@hramos @cpojer it would be great to have R8 and Android Studio 3.4 support in 0.60. Please review and merge.

mikehardy:
I realize this is closed and merged, but is there an issue tracking the need to alter code-shrink declarations so R8 may be enabled? We won't be able to disable it in the future and I see no linked issue and a search for android r8 in the tracker didn't return anything that seemed right

dulmandakh:
I disabled R8 for ReactAndroid because it bundles AndroidX appcompat,
okhttp and other libraries in it.

mikehardy:
@dulmandakh thanks for the reply

Your commit comment says `Disabled R8 for ReactAndroid because it'll strip out AndroidX and other libraries bundled in ReactAndroid.`

But here you say:

> I disabled R8 for ReactAndroid because it bundles AndroidX appcompat,
okhttp and other libraries in it.

Does it bundle them, or strip them? R8 is a shrinker, so I am going to assume you meant that it strips them, while ReactAndroid is *supposed* to bundle them? I just want to be clear so we can reason about it. 

If my assumption is correct (ReactAndroid should bundle them - and I think it should yes - while D8 is incorrectly stripping them) then the solution is to modify the shrinking declarations I believe?

That's what I am getting at - that there should be an issue to modify those declarations so that R8 is not disabled, because it is the future of the android build toolchain

I believe the work would take place here: https://github.com/facebook/react-native/blob/master/ReactAndroid/proguard-rules.pro vs disabling R8

Are there specific crashes associated with this, such that the rules can be modified as needed?

mikehardy:
associated reading (for interested parties) with context, link to FAQs and where to file R8 bugs if needed https://developer.android.com/studio/releases#r8-default

dulmandakh:
If enabled R8 will strip out unused classes, so I disabled it.

We bundle AppCompat and other libraries to be used in apps or libraries
that use ReactAndroid, so we don't know which classes will be used by them.
Thus, it's not possible to have proguard rules. So the safest way is to
disable R8 for ReactAndroid and deliver ReactAndroid, other libraries
intact.

mikehardy:
Ok - that makes sense - may I assume that at the app level - the consumers of ReactAndroid - if they are using R8 it will still shrink correctly? I think so, and I should test myself but you may know already.

dulmandakh:
RNTester was slightly smaller with R8 than Proguard. Also much smaller
compared to minification disabled. Please double check.

mikehardy:
Will do - I think this was all just me not fully understanding separation between R8 in ReactAndroid as a library, and an app. Thanks for responding

PR #24472, opened 2019-04-16
TM iOS: Add super class chain method search in RCTTurboModule

zhongwuzw:
## Summary

We need to search method lists of super class if TM has any subclasses.

## Changelog

[iOS] [Fixed] - TM iOS: Add super class chain method search in RCTTurboModule

## Test Plan

TM works when TM has subclass and super class has exported methods.

### DIFF
--- a/ReactCommon/turbomodule/core/platform/ios/RCTTurboModule.mm
+++ b/ReactCommon/turbomodule/core/platform/ios/RCTTurboModule.mm
@@ -304,19 +304,24 @@ SEL resolveMethodSelector(
     }
   } else {
     unsigned int numberOfMethods;
-    Method *methods = class_copyMethodList([module class], &numberOfMethods);
-    if (methods) {
-      NSString *methodPrefix = [NSString stringWithFormat:@"%s:", methodName.c_str()];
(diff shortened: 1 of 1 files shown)


fkgozali:
Thanks for the PR! However, the dynamic method lookup we have right now is not efficient, and I don't think we want to stick with dynamic lookup. The future plan is to have the codegen outputs exactly the method to call on the classes, so that there is no need to do method lookup during runtime anymore. This will give us build-time type safety as well.

So for the time being, let's keep the dynamic lookup as is (without going to the superclass).

zhongwuzw:
@fkgozali Thanks for review! Yeah, codegen is the future, can't wait to see it in TM (Fabric started to import), but just like the demo we are currently, we want to compatible legacy NativeModule system, in this case, maybe it's necessary to make runtime search robust, and in most situations, iteration would only do once(no module subclass).

fkgozali:
Technically having exposed methods on the super class is not an intended use case (e.g. Android doesn't let you do that). So we could revisit this in the future, pending our prod test/perf improvement work later this year, but I suspect at that time the codegen may be already usable.

PR #19274, opened 2018-05-15
Bump 0.49 to React 16

joeblynch:
Cherrypicks the update to React 16.0.0 from RN 0.50 into 0.49, so that 0.49 can be used with a non-rc/beta version of React. Bumps RN 0.49 to 0.49.6. Fixes https://github.com/react-native-community/react-native-releases/issues/18

We explored also cherrypicking the updates to prettier v1.7 and flow v0.56.0, however determined the related commits involved more changes than we'd like to back port.

Discussed with @hramos and @grabbou on contributor slack.

## Test Plan

Cherrypicked https://github.com/facebook/react-native/commit/678a7f3c39e495db602e90e994894051eb7aab10. Local run of tests, lint, and flow were successful.

## Related PRs

None

## Release Notes

[GENERAL] [MINOR] [-------------] - Updated React to v16.0.0

PR #24492, opened 2019-04-17
Revert "Bump 0.49 to React 16"

touqeerkhan11:
Reverts facebook/react-native#19274


cpojer:
I think you probably clicked a few wrong buttons to get here. This is not a change we are looking to make to a years old release.

PR #24496, opened 2019-04-17
Move Yargs to devDependencies

radeno:
## Summary

Yargs package is used only in test scripts and script for bumping version.
Based on this https://github.com/facebook/react-native/search?q=yargs&unscoped_q=yargs Yargs is not really required in production build and should be moved as development dependency.

## Changelog

[General] [Changed] - Moved Yargs to devDepencies

## Test Plan

CI Green


radeno:
You are welcome! Hope that failing test is general issue and not caused by PR.

PR #24497, opened 2019-04-17
iOS: Fix SafeAreaView in Modals

ericlewis:
## Summary

This is a fix for #18177.
## Changelog

[iOS] [Fixed] - SafeAreaView bottom not respected in Modal

## Test Plan

Poke around RNTester, open the SafeAreaView example and view the bottom spacing of the modal being correct.

### DIFF
--- a/React/Views/SafeAreaView/RCTSafeAreaView.m
+++ b/React/Views/SafeAreaView/RCTSafeAreaView.m
@@ -101,6 +101,10 @@ static BOOL UIEdgeInsetsEqualToEdgeInsetsWithThreshold(UIEdgeInsets insets1, UIE
 
 - (void)setSafeAreaInsets:(UIEdgeInsets)safeAreaInsets
 {
+  // HACK: enforces a bottom safe area inset when there isn't one being set.
+  // related issue: https://github.com/facebook/react-native/issues/18177
+  safeAreaInsets.bottom = safeAreaInsets.bottom == 0 && safeAreaInsets.left == 0 ? 34 : safeAreaInsets.bottom;
+  
   if (UIEdgeInsetsEqualToEdgeInsetsWithThreshold(safeAreaInsets, _currentSafeAreaInsets, 1.0 / RCTScreenScale())) {
     return;
   }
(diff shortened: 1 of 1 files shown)


ericlewis:
The reason this works is because setSafeAreaInsets seems to be called multiple times with the bottom safeAreaInset being 0 in the final call. A better solution might be to cache the inset value for bottom, so we don’t have to hardcode the value. It seems very weird that safe area inset would ever return 0 for bottom but a value for top.

ericlewis:
hmmm. this is not a fantastic way to solve it. after talking with Shergin, it seems the model is much more broken on apple's side than our own.

React/Views/SafeAreaView/RCTSafeAreaView.m
```diff
@@ -101,6 +101,17 @@ - (void)layoutSubviews
 
 - (void)setSafeAreaInsets:(UIEdgeInsets)safeAreaInsets
 {
+
+// Use the app's key window to determine Safe area insets.
+// fixes: https://github.com/facebook/react-native/issues/18177
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 /* __IPHONE_11_0 */
+  if (self.isSupportedByOS) {
+    if (@available(iOS 11.0, *)) {
+      safeAreaInsets = UIApplication.sharedApplication.keyWindow.safeAreaInsets;
```
zhongwuzw: @ericlewis 
I may have some worries about this, because we'd better let system to calculate and tell us the insets, and I may think the issue is caused by `Modal`. 🤔

JoshuaGross: @ericlewis I think this is reasonable, though I'm curious if what @zhongwuzw says is true / if there's a way of fixing this from Modal. If not, and regardless, can we move this logic from `setSafeAreaInsets` to `safeAreaInsetsIfSupportedAndEnabled` or `invalidateSafeAreaInsets `? Presumably `setSafeAreaInsets` is being called from `invalidateSafeAreaInsets`, so it seems weird to hijack the value in `setSafeAreaInsets` instead of just setting the correct value from `invalidateSafeAreaInsets`.

If it's not being called from `invalidateSafeAreaInsets`, where is it being called from, and could we fix it in that call site?

ericlewis: @JoshuaGross we could move it, but after messing with this more I noticed that when nesting SafeAreaViews that it will continually add the insets from window resulting in too much padding where it shouldn’t be! I might take a different approach here... maybe look at modal itself.

ericlewis:
closing this for now, will approach again with fresh eyes

zhongwuzw:
@ericlewis Maybe I'll fix this.

PR #24507, opened 2019-04-18
Fix RCTLinkingManager Swift representation (take 2!)

radex:
## Summary

The fix in https://github.com/facebook/react-native/pull/22764, unfortunately, does not work (at least not in Xcode 10.2 / Swift 5 mode). `__has_include(<UIKitCore/UIUserActivity.h>)` confuses the ObjC→Swift header converter and makes everything be `Any` (again, causing a crash at runtime).

This condition was added by @cpojer , so I assume it is important for Facebook tooling, but perhaps there's a different way to satisfy this?

## Changelog

[CATEGORY] [TYPE] - Message

### DIFF
--- a/Libraries/LinkingIOS/RCTLinkingManager.h
+++ b/Libraries/LinkingIOS/RCTLinkingManager.h
@@ -23,7 +23,7 @@
 + (BOOL)application:(nonnull UIApplication *)application
     continueUserActivity:(nonnull NSUserActivity *)userActivity
       restorationHandler:
-        #if __has_include(<UIKitCore/UIUserActivity.h>) && defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 12000) /* __IPHONE_12_0 */
+        #if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && (__IPHONE_OS_VERSION_MAX_ALLOWED >= 12000) /* __IPHONE_12_0 */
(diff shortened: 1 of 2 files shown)


radex:
Will include a changeling — but for now, I'm waiting for @cpojer's feedback – details above :)

cpojer:
Yeah I had to make this change at FB, otherwise it would break and not work. We use this check in other places of the codebase so I thought it is fine here too. I'm open to other suggestions.

radex:
OK, I have one idea in mind — will test it and let you know

cpojer:
@radex did you have a chance to get back to this?

radex:
Not quite :( I'm stuck with this — for some reason, the swift headers converter gets very confused by __has_include...

Can you give me a little more information about how the lack of __has_include is a problem with Facebook's internal tooling? Is it related to BUCK?

Perhaps there's some preprocessor flag that's passed in Facebook builds that are not applicable to open-source builds (e.g. `FACEBOOK`)? If so, maybe I could use a hack like this:

```
#ifdef FACEBOOK
#define xcode10_headers_available __has_include......
#elseif
#define xcode10_headers_available true
#endif
```

cpojer:
There is some build system stuff going on where I specifically need to test for that file to make something pass internally. I'm sorry I can't share more details than that :(

RubenSandwich:
@radex Can you share a working example of this crash? I have some time and want to help you out here.

cpojer:
We are trying to fix this internally, a commit should hopefully land and sync out to GitHub soon removing the `__has_include`.

PR #24399, opened 2019-04-10
[android] JSStackTrace -> Ensure lineNumber exists before consuming

Salakar:
## Summary

Fixes https://github.com/facebook/react-native/issues/24382

## Changelog

[ANDROID] [INTERNAL] - Fixed a `NoSuchKeyException` when parsing JS stack frames without line numbers.

## Test Plan

**TODO**: Will update tomorrow (sorry 🙈) just wanted to check CI.

PR #24508, opened 2019-04-18
Update JSStackTrace.java

tonmanayo:
## Summary

This change was recently added to the latest RN version 
https://github.com/facebook/react-native/pull/24399


cpojer:
Thanks! We are not currently merging things into older versions of React Native because it increases the maintenance burden for all of us. Please upgrade to the latest version of RN (0.60 will have this fix) as soon as it is out.

Issue #12159, opened 2017-02-01
EXC_BAD_ACCESS Attempted to dereference garbage pointer 0x167f4beb8.


dwilt:
### Description
So, this is going to be tough because I don't have exact reproduction steps for this error. We're getting reports into our Bugsnag application of crashes, which we've experienced ourselves, but as I said, are unable to reproduce. 

I've tried going through the stacktraces of the bugs, and a lot of them are consistent amongst themselves but there are also different once as well.

Here are the stracktraces:

Error 1:
```
EXC_BAD_ACCESS Attempted to dereference garbage pointer 0x18. 
    /usr/lib/libobjc.A.dylib objc_msgSend
    Frameworks/CoreFoundation.framework/CoreFoundation -[__NSDictionaryM objectForKey:]
    Frameworks/UIKit.framework/UIKit -[UIKBTree objectForProperty:]
    Frameworks/UIKit.framework/UIKit -[UIKBTree intForProperty:]
    Frameworks/UIKit.framework/UIKit -[UIKBTree displayType]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardLayoutStar setKeyplaneName:]
    AccessibilityBundles/UIKit.axbundle/UIKit -[UIKeyboardLayoutStarAccessibility setKeyplaneName:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardLayoutStar showKeyboardWithInputTraits:screenTraits:splitTraits:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardImpl finishLayoutChangeWithArguments:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardImpl updateLayout]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardImpl updateLayoutIfNecessary]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardImpl setFrame:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboard activate]
    AccessibilityBundles/UIKit.axbundle/UIKit -[UIKeyboardAccessibility activate]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardAutomatic activate]
    Frameworks/UIKit.framework/UIKit -[UIKeyboard setFrame:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboard _didChangeKeyplaneWithContext:]
    Frameworks/UIKit.framework/UIKit -[UIKeyboardImpl didMoveToSuperview]
[...]

PR #24516, opened 2019-04-18
Added default values for cachedLayouts and cachedMeasures

cvlehe:
## Summary

We were seeing this error when analyzing the project:
```The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage```.

We believe this is related to this issue: https://github.com/facebook/react-native/issues/12159. This PR adds default values for `cachedLayouts` and `cachedMeasures` to avoid this issue.

## Changelog

[iOS] [Fixed] - Message

## Test Plan

After adding the suggested changes, the analyzer no longer displays the attached error.
error

### DIFF
--- a/ReactCommon/yoga/yoga/YGMarker.h
+++ b/ReactCommon/yoga/yoga/YGMarker.h
@@ -23,8 +23,8 @@ typedef struct {
   int layouts;
   int measures;
   int maxMeasureCache;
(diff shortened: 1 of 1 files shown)


cpojer:
Hello! Please send Yoga specific PRs to the Yoga repository: https://github.com/facebook/yoga

PR #24506, opened 2019-04-18
[template] Implement changes to enable native modules auto linking

Salakar:
## Summary

Replaces #24099 (original PR became detached for some reason)

Implements the template changes required to enable native modules auto-linking for both Android & iOS. 

Requires the following to be merged first and an updated CLI to be published:

- [x] https://github.com/react-native-community/react-native-cli/pull/254 
- [x] https://github.com/react-native-community/react-native-cli/pull/256 
- [x] https://github.com/react-native-community/react-native-cli/pull/258

cc @grabbou @thymikee @orta for review

## Depends on

- [x] https://github.com/facebook/react-native/pull/24517 update CLI version)

## Changelog

[TEMPLATE] [FEATURE] - Enable auto-initialization/linking of react native modules for new projects

PR #24517, opened 2019-04-18
Use latest React Native CLI

grabbou:
## Summary

Updates React Native to use latest CLI.

Changes:
- No more `--reactNativePath`, define it once in the configuration file. This reverts the previous PR that added this flag
- Add `platforms` and `commands` - React Native now defines platform like any other package. There's no longer concept of "out-of-tree" platform. All are treated equally. If React Native works, any other platform will work too.
- Updates `jest/hasteImpl.js` to use public CLI interface (`loadConfig`) instead of `findPlugins` and removes a weird conditional that checks for CI presence.

## Changelog

[INTERNAL] - Update React Native CLI

## Test Plan

Run RNTester and it works


react-native.config.js
```diff
@@ -0,0 +1,33 @@
...
+} = require('@react-native-community/cli-platform-ios');
+const {
+  commands: commandsAndroid,
+  ...android
+} = require('@react-native-community/cli-platform-android');
+
+module.exports = {
+  reactNativePath: '.',
+  commands: [...commandsIOS, ...commandsAndroid],
+  platforms: {ios, android},
```
thymikee: maybe let's be explicit about what we put here, because:
1) it will serve as a documentation
2) if we ever export some extra helpers from these modules, they won't break the configuration

so, even though more verbose, I'd expect:
```js
platforms: {
  ios: {
    linkConfig: ios.linkConfig,
    projectConfig: ios.projectConfig,
    dependencyConfig: ios.dependencyConfig
  },
  android: {
    linkConfig: android.linkConfig,
    projectConfig: android.projectConfig,
    dependencyConfig: android.dependencyConfig
  }
}
```

grabbou: Yeah, good idea! Let me do that in a second.

grabbou:
Let's merge it so it unblocks https://github.com/facebook/react-native/pull/24506

package.json
```diff
@@ -87,7 +87,9 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@react-native-community/cli": "^2.0.0",
+    "@react-native-community/cli": "2.0.0-alpha.12",
+    "@react-native-community/cli-platform-ios": "2.0.0-alpha.9",
+    "@react-native-community/cli-platform-android": "2.0.0-alpha.9",
```
Salakar: These should probably be updated to alpha.14 - was a few bug fixes since.

grabbou:
Thanks @cpojer!

We will need to update the CLI dependency one more time, but RNTester works
w/o issues on master now, so that's good!

AkashBabu:
We are maintaining react-native project within android (the classical way) and we were soo much dependent on --reactNativePath for our build process. Now after migrating from 0.59 -> 0.61.5 we are not finding a way to provide --reactNativePath. 
So can you please suggest us an alternative ?

Thank you :)

PR #24518, opened 2019-04-18
Fix duplicate resource error in Android gradle build (#22234)

mikehardy:
## Summary

Issue #22234 includes a number of people (myself included) suffering with duplicate resource errors while building in Android. We have been collectively using a patch there but I don't believe any attempt has been made to PR it upstream.

I am not excellent at gradle, so I may have approached this in completely the wrong way, but it works for me in the standard init templates, as well as my current work project which has a complicated 2 buildType + 4 productFlavor configuration. If there is a better way to achieve the result I am happy to learn

The approach here is to determine the generated path the resources land in, then move them into the main build output tree. If they remain in the generated path android merging fails with duplicate resource errors, so that move (vs copy) is important.

## Changelog

[Android] [Fixed] - Fix duplicate resource error in Android build

## Test Plan

Reports of success on the linked issue via use of the patch + patch-package for a couple months, I personally use it full time with all gradle builds (`./gradlew clean assembleRelease` or if you have a 'staging' flavor, e.g. `./gradlew clean assembleStagingRelease`)

Related reading, also cross-links with the linked issue here:
https://stackoverflow.com/questions/53239705/react-native-error-duplicate-resources-android

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -48,6 +48,32 @@ afterEvaluate {
                 resourcesDir.mkdirs()
             }
 
+
+            // If there are flavors, remember the current flavor for use in the resource path we move from
+            def flavorPathSegment = ""
+            android.productFlavors.all { flavor ->
+                if (targetName.toLowerCase().contains(flavor.name)) {
+                    flavorPathSegment = flavor.name + "/"
+                }
+            }
+
+            // Address issue #22234 by moving generated resources into build dir so they are in one spot, not duplicated
+            doLast {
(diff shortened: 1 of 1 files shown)


mikehardy:
The failed results on test_android do not seem to be related to the PR in any way? In fact I'm not sure gradle was used at any point so I'm not sure the CI was relevant unfortunately.

turnrye:
Thank you for this fix @mikehardy !

PR #24533, opened 2019-04-19
Adds Logic To Catch MissingWebViewPackageException

thorbenprimke:
## Summary

We are seeing crash reports that the webview is missing. In this case
it should fail gracefully so that a missing webview does not block
from using an app built with React Native.

The contains could also be changed to check for "webview" in general
to catch all webview related exception. It's currently checking on
for the specific string that I'm seeing in our app's crashreporting tool.

Screen Shot 2019-04-19 at 11 26 19 AM

Screen Shot 2019-04-19 at 11 32 58 AM

## Changelog

[Android] [Fixed] - The ReactCookieJarContainer/ForwardingCookieHandler now handles the missing WebView  gracefully.

## Test Plan

No new tests have been added.
CI should pass.

This fix was made based on the exception stacktrace. I have not spent the time to acquire one of the devices that it is happening on.


thorbenprimke:
@hramos - mind taking a look at this PR?

PR #24535, opened 2019-04-19
[template] add jsc pickFirst packagingOptions

Salakar:
## Summary

Fixes the following build failures on Android Studio (happening when using the 0.60 template):

image

image

cc @dulmandakh 

## Changelog

[Android] [Fixed] - Fixed a `more than one file was found... lib/x86/libjsc.so` build failure with the template

## Test Plan

Build is now successful and the app starts with this change:

image

image

### DIFF
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -167,6 +167,8 @@ android {
         pickFirst '**/x86/libc++_shared.so'
         pickFirst '**/arm64-v8a/libc++_shared.so'
         pickFirst '**/x86_64/libc++_shared.so'
+        pickFirst '**/x86/libjsc.so'
+        pickFirst '**/armeabi-v7a/libjsc.so'
     }
 }
 


Kudo:
@Salakar I am sorry to introduce this issue.
Just doubt why there are multiple lib.jsc.
Could you please provide information below the screenshot you pasted?
IIRC there will have the path to the libjsc.so.

Salakar:
Hey @Kudo - this was just a fresh RN init of the template currently on 0.60-stable branch (no extra changes made) - there was also nothing extra after the screenshot sorry.

Reproducible example from the template is here: https://github.com/invertase/react-native-auto-link-example (npx command there to create a template - if you remove the 2 lines that I added as part of this PR - https://github.com/invertase/react-native-auto-link-example/commit/71d2cc7685394d96549fa8ee36648e5c4c77ce73 - the error will appear)

Kudo:
@Salakar Yes, the problem may related to 0.59.5 release.
The react-native-0.59.5.aar still introduced libjsc.so either by maven dependency or packed x86_64/arm64-v8a lib.jsc directly.
I am sorry that cannot pair 0.60 template with 0.59 AAR.
In the mean time that 0.60-rc not published yet, maybe to generate 0.60 AAR handy `./gradlew :ReactAndroid:installArchives`

Salakar:
@Kudo I agree that this is most likely down to 59.5 as well.

Maybe it's still worth merging this though, in case anyone downgrades their template for whatever reason (it happens)?

Kudo:
@Salakar IMHO, pickFirst is not reliable to choose correct JSC.
If user like to choose `android-jsc-intl` from 0.60 but gradle pickFirst may use the JSC from 0.59 (which is no Intl supported).
`Intl` not defined error happens implicitly.
I would prefer to clarify how duplicated libjsc.so happens instead of just pass the build issue. 

In fact, we are trying to remove the pickFirst for libc++_shared.so as well.
https://github.com/react-native-community/jsc-android-buildscripts/pull/95

Salakar:
@Kudo that's fair, how would you recommend we remove pickFirst behaviour as I can do another PR for this

Kudo:
@Salakar I still waiting for [the PR which exclude libc++_shared.so](https://github.com/react-native-community/jsc-android-buildscripts/pull/95) being merged and release a new version to npm.

PR #24538, opened 2019-04-20
Mock requireNativeComponent instead of individual components

bcarroll22:
## Summary

I believe this PR makes the native component mocks clearer in test output, and I wanted to start a dialog about it. I also believe this is a better approach because it should work for third-party native components out of the box as well, with no effort for the component maintainers.

The problem I believe it solves is that current state, snapshot output can be a little confusing. For instance, if I render a `TouchableOpacity`, my snapshot will be a `View`. That, to me, is confusing because:

1. I don't see my `TouchableOpacity`
1. I definitely didn't render a `View`

If, however, what I see is an `RCTView` with native props I maybe don't recognize, that, at least to me, more clearly communicates that what's in the snapshot is a native element--a primitive. It's similar to how react-dom developers wouldn't be confused if their `Link` component showed up in a shapshot as an `a` tag. It's implied that the snapshot output is the browser primitive that React rendered.

I think this makes it more clear that what shows up in a snapshot is a native primitive. I believe that ultimately, this will help users more clearly understand what's going on when components bridge from React to Native. 

As an added benefit, once it was refactored this way, I just didn't need individual component mocks. I really believe this angle will make the mocks more robust and resilient, even if it takes some iteration to get to the finish line.

## Changelog

[General] [Changed] - Changed `requireNativeComponent` mock
[General] [Removed] - Removed component mocks

## Test Plan

I know this creates a big snapshot diff, but I think it's great because they now (to me at least) more clearly convey what's being sent across the bridge. I think this makes the snapshots more clear and reflective of what's going on at the native level, which I will ultimately help developers debug and test their apps. 

The tests didn't really need changed, all I did was update the snapshot files now that the actual native components are being mocked. No warnings or console logs were introduced to the logs either, so I'm assuming we're peachy there 👍


bcarroll22:
I just wanted to clarify, I don't think this should be merged yet. I wanted to get some feedback on it before I spend too much more time on it. In order for this to be ready to merge, I'll need to go through and update the wording on some tests to accurately describe what they do now.

jest/requireNativeComponent.js
```diff
@@ -0,0 +1,34 @@
...
+  return viewName => {
+    class NativeComponent extends React.Component {
+      render() {
+        const {children, ...props} = this.props;
+
+        return React.createElement(viewName, props, children);
+      }
+    }
+
+    Object.assign(NativeComponent.prototype, {
```
TheSavior: Why is this needed?

TheSavior:
I don't think this is quite the approach we'd like to take. 

I'd like to avoid having the implementations of the components built into React Native leak into product's snapshot tests. For example, if we get to a semver world, if we change the nesting inside of TextInput or something, that shouldn't be a breaking change. This is why we mock the components exported by React Native, so you end up with `<TextInput ... />` in your snapshot test. 

I think that requireNativeComponent should definitely be mocked as well in case someone doesn't add a custom mock for that component, but it should be the fallback.

bcarroll22:
This explanation makes sense to me, but let me give you another example of why I think this is a good approach. I have a set of existing tests that I ran with these new mocks, and it caught multiple instances of props that were incorrect or invalid since the component being used in the test was the “real” component. These invalid things were missed with the current mocks because they’re, well, mocks. 

I could maybe see an argument that it’s leaking implementation details, but I think it might be the opposite; I think it’s communicating exactly what’s actually happening and _not_ leaking implementation, which I believe is more beneficial to people learning the framework at a more intermediate to advanced level. If this were mocking things like `Button` or Touchable, I would agree because those are definitely implementations, but this is only mocking the actual native components and communicating things in a more realistic way.  

To piggy back on your example of semver, I think that argument actually works in favor of this approach too — users might be relying on the core components working a certain way and their snapshots breaking could help them identify where issues lie. I think this would also help ease upgrades because it could point out where things may have broken at the native level between versions. 

I understand if you guys don’t want the jest preset bundled with the library to be this way, but I really do think it’s beneficial. It makes the test environment more like the actual native environment, which will ultimately give users more confidence in their apps (especially in cases like the example I gave in the first paragraph).

jest/requireNativeComponent.js
bcarroll22: This only mocks actual native components, so they will all need the native methods mocked out. And then for example, things like TextInput need _nativeTag and Animated (ScrollView) needs the viewConfig.

TheSavior:
Thanks for the additional info.

> I have a set of existing tests that I ran with these new mocks, and it caught multiple instances of props that were incorrect or invalid since the component being used in the test was the “real” component.

Do you have some concrete code examples of this? I’m not exactly sure what this means so it would be helpful to see some examples.

I think something we are missing today that we are kinda talking around is the need for these unlocked snapshot tests in core. We don’t currently have those which means our downstream users don’t have confidence that we didn’t break things. Ideally that confidence would exist and mocks for our exports would be reasonable. I’d prefer we strive towards that goal.

However, people will always want to mock at different levels based on their needs and trade offs. I think this conversation is mostly just revisiting where the defaults should be. I think I’d prefer to stick with the current defaults and see if we can sufficiently improve them rather than expose a much deeper level.

bcarroll22:
When I get home I’ll send some examples ☺️ 

I guess for me, you can remove a react developer from the DOM, but you can’t remove the DOM from the react developer. So what are those primitives and how do you very closely imitate them in tests? The core wrappers around requireNativeComponent or the native component itself? 

Regardless of which answer is right, I think 1) some significant work needs done to the bundled mocks 2) teams who aren’t on the latest version of RN should be able to take advantage of them. I’d love to help with that, whatever it ends up looking like 👍

I took this approach because I think it’s easier, just as good as building better individual mocks, and almost exactly resembles the way the app works at a native level. But I digress, count me in either way 💯

bcarroll22:
I had a big thing written but I think I convinced myself I'm wrong and I hate the way these snapshots are as well. Like I promised though, here's a good example to answer your question from earlier:

```
<Image source="/finding-nemo.png" />
```

In the current mocks, this was no problem. Once I changed it to the way it is in this PR, I got a propTypes warning because this isn't a valid value for an `Image` source. That was one of the things that had me super stoked about this change. So, this part is still better I think.

BUT. The way this PR makes the snapshots is ugly. Using my own argument against myself, if I render this in the React: `<input onChange={() => console.log('sup')} />` I'm going to get `<input />` in a DOM snapshot. This is much better because who cares about event listeners in a snapshot? Also they aren't going to literally be in the DOM. It's also not misleading and shows me something that makes sense and I would have expected. Let me give you an outrageous snapshot example:

Given that I render this:

```
<View>
  <TouchableOpacity onPress={this.fetch}>
    <Text>Fetch</Text>
  </TouchableOpacity>
  {data ? <Text>{data.greeting}</Text> : null}
</View>
```

this was my snapshot output (🤢):

```
<View>
  <View
    accessible={true}
    isTVSelectable={true}
    onResponderGrant={[Function]}
    onResponderMove={[Function]}
    onResponderRelease={[Function]}
    onResponderTerminate={[Function]}
    onResponderTerminationRequest={[Function]}
    onStartShouldSetResponder={[Function]}
    style={
      Object {
        "opacity": 1,
      }
    }
  >
    <Text>
      Fetch
    </Text>
  </View>
  <Text>
    hello there
  </Text>
</View>
```

This is sure to be an "uhhh wow okay, so where's my `onPress`?" moment. And this PR would make that even worse? Yeah, that'd be a no from me too 👎 You said some things that convinced me that the correct snapshot output for the above example would be:

```
<View>
  <TouchableOpacity>
    Fetch
  </TouchableOpacity>
  <Text>
    hello there
  </Text>
</View>
```

What are your thoughts on working toward something like this? It might be hard to get to a place where this output is good enough to give confidence and not be confusing, but does this seem right to you?

TheSavior:
Yay, thanks for providing some examples. Yeah, I agree those definitely don’t look pretty. For some other particularly nasty cases you might try a TextInput or ScrollView. Those both conditionally use 1 of something like 4 native components internally, and we hope with Fabric that will change to just one without breaking any functionality. 

Another situation might be a test with FlatList that ends up rendering some complex cells and scrollviews.

I think there is definitely room for each component in core to have some unmocked snapshot tests like the ones you touched for ActivityIndicator in this PR. If there were PRs for those tests for more components we’d merge them right away.

For the example you gave, I think I agree, but isn’t that also how it works today? I guess I’d expect to see the onPress there. For example, if you conditionally attach a function to a prop, I’d expect to see it in a snapshot test and for it to fail if it is passed incorrectly.

You make a good point about validation though, there might be some things we want to validate that we are validating when not using the mocks. I’m not sure propTypes is a good example though as we’ve been removing those from the codebase in favor of flow/typescript. Maybe a better example is something like how we sided to error when a View was nestled inside a Text. I wonder if there are more of those we want to be able to validate at test time.

bcarroll22:
@TheSavior I just force pushed an update based on our conversation, let me know what you think. I think this may be a better starting place for iterating.

The difference in approach here is that rather than focusing on mocking `requireNativeComponent` I just made sure that all components that call `requireNativeComponent` are mocked appropriately. To me this makes sense--rather than `requireNativeComponent` being the primitive, this makes any core component that calls it the primitive.

bcarroll22:
Sorry just saw this.

1 - Yep I can definitely help with that!

2 - I pushed a few commits that I think actually get us closer to where we wanna be with this step than anything I've done so far. I think I basically finished step 2 before I knew it would be step 2.

3- With every component mocked, the biggest changes are to Lists, which I expected, they're the most intense. The _coolest_ difference is that what shows up in snapshots will be what developers passed in (no leaky implementations 😉)

4 - The ScrollView/List snaps changed a lot with this PR, but it had more to do with other things being mocked differently than the lists themselves. The ScrollView is still mocked and the RefreshControl is mocked as well, just following the format of the others. `latestRef` doesn't show up in the repo as far as I could tell, so maybe you could tell me more about what that is/if it still matters?

TheSavior:
Awesome, so I think this PR can turn into #2, but I'd want to merge all of #1 first because this might change the snapshots in the test for #1. 

For #4, It looks like at Facebook we have one callsite that uses `latestRef`, doing this: 
```
RefreshControl.latestRef.props.onRefresh();
```

It doesn't seem like this is the right way to test that, so we'll need to update that test if we are going to change it.

bcarroll22:
@TheSavior you know, as I'm working on task 1, I'm realizing that it's going to make the diff for this really really big once it's completed.

I want to have a plan of attack because task 1 is important for the core team since it'll help you make sure updates aren't breaking components, but task 2 is important for the community because it'll make snapshots look like what they're expecting in their app tests.

This is a huge topic and this might not be the right place for it, but should the community have the same Jest config as the core team? 🤔 As I'm thinking about it, I actually could see (from the perspective of the core team) my very first PR being the most useful because it'll show you the native render result which would give you more confidence you aren't accidentally breaking the components.

But for the community, they don't care about that. We could give them a Jest preset that shows clean "high-level" output they're expecting to see with things like `onPress` for touchables, and repurpose this preset to be one just meant for super low-level snapshots like native components.

More confidence for the core team, more confidence for the community. Thoughts?

TheSavior:
Fun fact, the jest config in this repo is the same jest config that is used for products using RN at Facebook. My thinking for step one is that it includes snapshots for components the way products using RN would see them, as well as the unmocked version that we want to test the core. We don't want to unintentionally break either of those situations and thus would want tests for both. 

I do think we want the snapshot tests first, because that will give us confidence in the changes that step 2 is causing. Otherwise we are making changes to how people's snapshot tests will appear without knowing what those changes are and that we are getting them right. As we mentioned on one of your other PRs, we should feel free to break up that work so that it isn't one large impossible to review PR. :) 

Does that answer the questions?

bcarroll22:
Yep, I'm on the same page with all of that stuff, I think this was more of a future facing question. I'll continue down the same path with the other PRs to add current-state snapshots, just wanted to think about the implications it would have if we were to do something like that. No matter how much its broken up, it'll still be a big change.

I think for those building apps, it's beneficial to see:

```
<TouchableOpacity onPress={[Function]}>
```

because it reflects what they rendered. But for the internal library snapshots, my question is does it make more sense to see: 

```
<RCTView>
```` 
(with all of the `onGesture` events that I don't wanna type, lol). Because that would ensure the core components are rendering the native components correctly without them being "mocked".

TheSavior:
> But for the internal library snapshots, my question is does it make more sense to see...

Aren't the snapshot tests in the other PR showing us that? But also showing us what product developers will see? At least, that's the point of having a test with them mocked and not mocked.

cpojer:
Hey @bcarroll22 what's the status of this PR?

bcarroll22:
@cpojer lets close it for now, until I get some other stuff done per my conversations with @TheSavior this can't really move forward until some other things are wrapped up.

PR #24541, opened 2019-04-20
Make Alert not cancelable by default on Android

JonnyBurger:
## Summary

By default, an alert is `cancelable` on Android but not on iOS.
This PR changes the behavior so that the Alert is not dismissable on Android by default.

The motivation is that many developers develop on iOS and test on Android, and do forget to consider the case that the alert is dismissable.
Consistent behavior by default makes it easier to develop cross-platform apps in general.

---

For context and for your consideration, I have started a discussion here with the topic of whether React Native should try to use OS defaults or be consistent between platforms:

https://github.com/react-native-community/discussions-and-proposals/issues/121

---

If this PR gets merged, the docs should be updated as well:

https://github.com/facebook/react-native-website/blob/master/docs/alert.md#android

## Changelog

[Android] [Changed] - By default, alerts are not dismissable

## Test Plan

By testing out the change in `node_modules`, I have found that the change has the intended effect.

### DIFF
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -56,7 +56,7 @@ class Alert {
       let config = {
         title: title || '',
         message: message || '',
-        cancelable: false
+        cancelable: false,
       };
 
       if (options) {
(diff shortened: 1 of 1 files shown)


JonnyBurger:
Happy to do it! Thanks for taking the time to weigh in, review and merge.

I have also created a PR for docs to correctly reflect the new behavior: https://github.com/facebook/react-native-website/pull/899

PR #24547, opened 2019-04-21
Fix jsc-android not found if building from source

Kudo:
## Summary

If an app [builds from RN source](https://facebook.github.io/react-native/docs/building-from-source), there was an error for jsc-android not found.
It is a side effect of my previous [JSC as node dependency change](https://github.com/facebook/react-native/commit/8e375850de73d1b998a83279e6c4ba238867a453)
For building from RN source case, the jsc-android is located at `/path/to/app/node_modules/jsc-android`.
Original gradle task will try to find it at `/path/to/app/node_modules/react-native/ReactAndroid/../node_modules/jsc-android`, as ReactAndroid project path was being override inside node_modules.
The change fixes the building from source case.

## Changelog

N/A
This change does not need to publish into changelog, as it is a master branch building fix.

## Test Plan

1. Test building RN AAR: `./gradlew :ReactAndroid:installArchives`
2. Test an app to build from source as described at https://facebook.github.io/react-native/docs/building-from-source

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -136,11 +136,15 @@ task prepareGlog(dependsOn: dependenciesPath ? [] : [downloadGlog], type: Copy) 
 // Create Android.mk library module based on jsc from npm
 task prepareJSC {
     doLast {
-        def jscPackageRoot = fileTree("$projectDir/../node_modules/jsc-android/dist")
-        def jscAAR = jscPackageRoot.matching({ it.include "**/android-jsc/**/*.aar" }).singleFile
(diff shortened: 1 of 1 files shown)


janicduplessis:
@Kudo Would it be possible to make the path configurable? I'm using yarn workspaces and RN is not installed in `/path/to/app/node_modules`.

janicduplessis:
Maybe a better alternative would be to use `$projectDir/../../jsc-android/dist` instead. That way it should work with most monorepo configs, since it will look in the same directory where RN is installed.

Kudo:
@janicduplessis Good idea, please let me update accordingly.

PR #24555, opened 2019-04-22
Adds a ruby file in RN which reflects what Podspecs should be imported by default

orta:
## Summary

Simplifies the code anyone on iOS using RN has _to see_. In this case, React Native knows that everyone probably wants all these imports (unless they're using dev mode) and so we can auto-import the pod specs for a user from inside the lib.

Basically auto-link for the React side.

## Changelog

[iOS] [Added] - Adds a ruby function which imports the Pods for RN, so that users only have to include this function and it can change per RN version.

## Test Plan

I did a `pod install` with RNTester, and I don't think it makes sense for writing tests for the Ruby code

/cc @grabbou @alloy

### DIFF
--- a/RNTester/Podfile
+++ b/RNTester/Podfile
@@ -1,29 +1,21 @@
 platform :ios, '9.0'
 
+require_relative '../scripts/autolink-ios'
+
 target 'RNTester' do
   # Uncomment for Swift
   # use_frameworks!
 
   project 'RNTesterPods.xcodeproj'
+  use_react_native!(path: "..")
 
-  pod 'React', :path => '../'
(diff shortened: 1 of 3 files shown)


orta:
This will need changes in the 0.60 version of the RN-CLI default template to reference this ruby file in the same way the RNTester does.

RNTester/Podfile
```diff
@@ -1,48 +1,28 @@
 platform :ios, '9.0'
 
+require_relative '../scripts/autolink-ios'
```
orta: maybe this makes more sense (as a user of react native) as like `require_relative '../node_modules/react-native/scripts/cocoapods'`?

Issue #23712, opened 2019-03-01
Animated.Value does not have correct value after animation if component was re-mounted


karhatsu:
## 🐛 Bug Report
Animated.Value's internal value stops updating during animations (e.g. Animated.timing) if the component using it in styles have been unmounted.

## To Reproduce
1. Create Animated.Value `av` and add a listener for it.
2. Use `av` in styles for component `C` (e.g. `translateX`).
3. Run an animation (`useNativeDriver: true`) with `av` and see with the help of the listener how the value was changed correctly.
4. Unmount `C` (and all other components that utilize `av` in their styles).
5. Mount `C`
6. Run the same animation again and see how the value does not change (UI changes correctly though).
7. Unmount and mount `C` again.
8. `C` won't be positioned correctly since `av`'s internal value reflects the state after the phase 3 instead of the phase 6.

## Expected Behavior
The Animated.Value's internal value should change always to the value defined in the animation.

## Code Example
[https://github.com/karhatsu/react-native-animated-value-bug](https://github.com/karhatsu/react-native-animated-value-bug)

## Environment
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
      Memory: 48.68 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.9.4 - ~/.nvm/versions/node/v8.9.4/bin/node
      Yarn: 1.13.0 - ~/.yarn/bin/yarn
      npm: 6.7.0 - ~/.nvm/versions/node/v8.9.4/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.1, 25.0.2, 26.0.1, 26.0.2, 27.0.2, 27.0.3, 28.0.0, 28.0.3
        System Images: android-27 | Android TV Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5264788
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #24571, opened 2019-04-23
Fix Animated.Value value after animation if component was re-mounted

michalchudziak:
## Summary

Fixes https://github.com/facebook/react-native/issues/23712

Currently, It seems like `__nativeAnimatedValueListener` is not listening to the correct `onAnimatedValueUpdate` events if component was re-mounted. In this PR I'm attaching a new listener if the native view tag has changed.

## Changelog

[General] [Fixed] - Fixed Animated.Value value after animation if component was re-mounted

## Test Plan

Compare fresh react native app with this branch and current master. Following code from @karhatsu demonstrates issue very well:

`App.js`
```js
import React from 'react';
import { Animated, Button, ScrollView, StyleSheet, Text, View } from 'react-native';

const containerWidth = 200
const boxSize = 50

  constructor(props) {
    super(props)
    this.x = new Animated.Value(0)
    this.valueChangeCounter = 0
    this.x.addListener(({ value }) => {
      // do not log every change since there are so many of them
      if (this.valueChangeCounter % 5 === 0) {
        this.log(`x: ${value}`)
      }
      this.valueChangeCounter += 1
    })
    this.state = {
      boxVisible: true,
      logs: []
    }
  }

  render() {
    const { boxVisible } = this.state
    const toggleText = boxVisible ? 'Hide' : 'Show'
    return (
        {this.renderBox()}
        <View style={styles.buttonsContainer}>
          <Button style={styles.button} title="<-" onPress={this.moveTo(0)} disabled={!boxVisible}/>
          <Button style={styles.button} title={toggleText} onPress={this.toggleVisibility}/>
          <Button style={styles.button} title="->" onPress={this.moveTo(containerWidth - boxSize)} disabled={!boxVisible}/>
        </View>
        {this.renderFix(false)}
        <Text style={styles.instructions}>
          Instructions: Click arrow buttons to move the box and see how the animated value changes during the animation.
          Then hide the box and reveal it again. After that the arrow button still moves the box in the UI but the
          underlying value does not change during the animation. This means that when you hide and show the box again,
          the box location does not reflect the latest animation end value.
        </Text>
        <ScrollView style={styles.logs}>
          {this.state.logs.map(log => {
            return <Text key={log.id}>{log.text}</Text>
          })}
        </ScrollView>
      </View>
    );
  }

  // call this with true and the bug gets "fixed"
  renderFix = (fixIt) => {
    if (fixIt) {
      return <Animated.View style={{ transform: [{ translateX: this.x }] }}/>
    }
  }

  renderBox = () => {
    if (this.state.boxVisible) {
      const horizontalLocation = { transform: [{ translateX: this.x }] }
      return (
        <View style={styles.boxContainer}>
          <Animated.View style={[styles.box, horizontalLocation]}/>
        </View>
      )
    } else {
      return (
        <View style={styles.boxContainer}>
          <Text>The box view is not being rendered</Text>
        </View>
      )
    }
  }

  moveTo = x => {
    return () => {
      this.log(`Animate to ${x}`)
      Animated.timing(this.x, {
        toValue: x,
        duration: 1000,
        useNativeDriver: true
      }).start(({ finished }) => {
        this.log(`Animation finished (${finished}), x is now ${this.x._value}`)
      })
    }
  }

  toggleVisibility = () => {
    const { boxVisible } = this.state
    this.setState({ boxVisible: !boxVisible })
    const log = `${boxVisible ? 'Hide box' : 'Show box'}, x is now ${this.x._value}`
    this.log(log)
  }

  log = text => {
    const logs = [...this.state.logs]
    logs.unshift({ id: logs.length, text })
    this.setState({ logs })
  }
}

  container: {
    display: 'flex',
    flexDirection: 'column',
    backgroundColor: '#fff',
    padding: 30,
  },
  boxContainer: {
    backgroundColor: 'yellow',
    height: boxSize,
    width: containerWidth,
  },
  box: {
    width: boxSize,
    height: boxSize,
    backgroundColor: 'orange',
  },
  buttonsContainer: {
    marginTop: 20,
    display: 'flex',
    flexDirection: 'row',
    justifyContent: 'space-between',
    width: containerWidth,
  },
  button: {
    width: 80,
  },
  instructions: {
    color: 'black',
    fontSize: 10,
    marginVertical: 20,
  },
  log: {
    color: 'gray',
  },
});
```

### DIFF
--- a/Libraries/Animated/src/nodes/AnimatedNode.js
+++ b/Libraries/Animated/src/nodes/AnimatedNode.js
@@ -112,7 +112,11 @@ class AnimatedNode {
       return;
     }
 
-    this.__shouldUpdateListenersForNewNativeTag = false;
+    if (this.__shouldUpdateListenersForNewNativeTag) {
+      this.__shouldUpdateListenersForNewNativeTag = false;
+      this._stopListeningForNativeValueUpdates();
+    }
+
     NativeAnimatedAPI.startListeningToAnimatedNodeValue(this.__getNativeTag());
     this.__nativeAnimatedValueListener = NativeAnimatedHelper.nativeEventEmitter.addListener(
       'onAnimatedValueUpdate',
--- a/Libraries/Animated/src/nodes/AnimatedNode.js
+++ b/Libraries/Animated/src/nodes/AnimatedNode.js
@@ -43,6 +43,7 @@ class AnimatedNode {
   /* Methods and props used by native Animated impl */
   __isNative: boolean;
   __nativeTag: ?number;
+  __shouldUpdateListenersForNewNativeTag: boolean;
 
   constructor() {
     this._listeners = {};
@@ -104,10 +105,14 @@ class AnimatedNode {
   }
 
   _startListeningToNativeValueUpdates() {
-    if (this.__nativeAnimatedValueListener) {
+    if (
+      this.__nativeAnimatedValueListener &&
+      !this.__shouldUpdateListenersForNewNativeTag
+    ) {
       return;
     }
 
+    this.__shouldUpdateListenersForNewNativeTag = false;
     NativeAnimatedAPI.startListeningToAnimatedNodeValue(this.__getNativeTag());
     this.__nativeAnimatedValueListener = NativeAnimatedHelper.nativeEventEmitter.addListener(
       'onAnimatedValueUpdate',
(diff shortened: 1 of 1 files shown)


cpojer:
cc @janicduplessis could you review this PR?

michalchudziak:
@osdnk I found out that after re-mounting the element, it's `native-tag` changes (which seems quite logical). The origin of this issue was the listener observing the view with the previous `native-tag`, not updating the listener if the tag changes. In this implementation, I set the flag if the native tag changes and update the listeners if this flag is set to `true`.

PR #24583, opened 2019-04-24
Adjust test manual e2e script to work with new init

Esemesek:
## Summary

Since initialisation flow changed with default template, we need to adjust the script to make it work properly.

## Changelog

[General] [Fixed] - Adjust text manual e2e script to work with new init.

## Test Plan

I wasn't able to run whole script, but I've tested a part that I have written.

cc @grabbou Can you test this script in `0.60-stable` branch?

### DIFF
--- a/scripts/test-manual-e2e.sh
+++ b/scripts/test-manual-e2e.sh
@@ -83,7 +83,7 @@ project_name="RNTestProject"
 
 cd /tmp/
 rm -rf "$project_name"
-npx @react-native-community/cli init $project_name --template $repo_root
+npx @react-native-community/cli init "$project_name" --template "$repo_root"
 
 info "Double checking the versions in package.json are correct:"
 grep "\"react-native\": \".*react-native-$PACKAGE_VERSION.tgz\"" "/tmp/${project_name}/package.json" || error "Incorrect version number in /tmp/${project_name}/package.json"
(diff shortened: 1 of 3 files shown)


scripts/test-manual-e2e.sh
```diff
@@ -76,11 +76,14 @@ npm pack
...
 
+node set-rn-template-version.js "file://$PACKAGE"
+success "React Native version changed in the template"
+
 project_name="RNTestProject"
 
 cd /tmp/
 rm -rf "$project_name"
-react-native init "$project_name" --version $PACKAGE
+npx @react-native-community/cli init "$project_name" --template "$repo_root"
```
thymikee: This points to the `latest` which currently is 1.x and 0.60+ is going to be working with 2.x. Maybe it's worth to pass the version/tag from the arguments?

grabbou: We can just call $repo_root/cli.js? This will take whatever React Native has in source.

Esemesek: Calling `cli.js` would be the best solution. I'll change that.

grabbou:
@cpojer, please don't land this one yet! I encountered few issues while working with it. I already applied this onto 0.60-stable. 

Will send @Esemesek the files to update.

grabbou:
@Esemesek please cherry-pick this https://github.com/facebook/react-native/commit/a2646ef7a22774f7bd7fa8ddc60e2565907ca1ff onto your branch before we land it :)

I've redone it, but I feel like credits should go to you!

cpojer:
Stopped the land. Let me know when it's good to go. Fyi I made a small text change in the error message.

Esemesek:
@cpojer should be good to go

grabbou:
Yeah, let's go!

PR #24589, opened 2019-04-24
Add setNativeProps support for SafeAreaView

### LINKED ISSUE #24576: SafeAreaView does not have setNativeProps like View
## 🐛 Bug Report
SafeAreaView does not have setNativeProps like View

## To Reproduce

Call `ref.current.setNativeProps()`

## Expected Behavior
SafeAreaView should support this just like the View does.

## Code Example

https://snack.expo.io/@brunolemos/safeareaview-setnativeprops

```jsx
import { View, SafeAreaView } from 'react-native';

export default class App extends React.PureComponent {
  render() {
    return (
      <>
        <View
          ref={ref =>
            ref && alert(`View has setNativeProps: ${'setNativeProps' in ref}`)
          }
        />
        <SafeAreaView
          ref={ref =>
            ref &&
            alert(`SafeAreaView has setNativeProps: ${'setNativeProps' in ref}`)
          }
        />
      </>
    );
  }
}
```

## Environment
    System:
      OS: macOS 10.14.4
      CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
      Memory: 341.11 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.5.0 - /usr/local/bin/node
      Yarn: 1.13.0 - ~/.yarn/bin/yarn
      npm: 6.7.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 19, 22, 23, 24, 25, 27, 28
        Build Tools: 23.0.1, 23.0.3, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.2, 27.0.3, 28.0.0, 28.0.2, 28.0.3
        System Images: android-19 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
    IDEs:
    npmGlobalPackages:

zhongwuzw:
## Summary

Fixes #24576 .
cc. @cpojer .

## Changelog

[iOS] [Fixed] - Add setNativeProps support for SafeAreaView

## Test Plan

Ci passes.

### DIFF
--- a/Libraries/Components/SafeAreaView/SafeAreaView.js
+++ b/Libraries/Components/SafeAreaView/SafeAreaView.js
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @flow strict-local
+ * @flow
  * @format
  */
 
@@ -13,6 +13,7 @@ const React = require('React');
 const View = require('View');
 
 import type {ViewProps} from 'ViewPropTypes';
+import type {NativeComponent} from 'ReactNative';
 
 type Props = $ReadOnly<{|
   ...ViewProps,
@@ -31,24 +32,36 @@ let exported;
  * sensor housing area on iPhone X).
  */
 if (Platform.OS === 'android') {
-  exported = class SafeAreaView extends React.Component<Props> {
-    render(): React.Node {
-      const {emulateUnlessSupported, ...props} = this.props;
-      return <View {...props} />;
-    }
+  const SafeAreaView = (
+    props: Props,
+    forwardedRef?: ?React.Ref<typeof View>,
+  ) => {
+    const {emulateUnlessSupported, ...localProps} = props;
+    return <View {...localProps} ref={forwardedRef} />;
   };
+
+  const SafeAreaViewRef = React.forwardRef(SafeAreaView);
+  SafeAreaViewRef.displayName = 'SafeAreaView';
+  exported = ((SafeAreaViewRef: any): Class<React.Component<Props>>);
 } else {
   const RCTSafeAreaViewNativeComponent = require('RCTSafeAreaViewNativeComponent');
-  exported = class SafeAreaView extends React.Component<Props> {
-    render(): React.Node {
-      return (
-        <RCTSafeAreaViewNativeComponent
-          emulateUnlessSupported={true}
-          {...this.props}
-        />
-      );
-    }
+
+  const SafeAreaView = (
+    props: Props,
+    forwardedRef?: ?React.Ref<typeof RCTSafeAreaViewNativeComponent>,
+  ) => {
+    return (
+      <RCTSafeAreaViewNativeComponent
+        emulateUnlessSupported={true}
+        {...props}
+        ref={forwardedRef}
+      />
+    );
   };
+
+  const SafeAreaViewRef = React.forwardRef(SafeAreaView);
+  SafeAreaViewRef.displayName = 'SafeAreaView';
+  exported = ((SafeAreaViewRef: any): Class<NativeComponent<Props>>);
 }
 
 module.exports = exported;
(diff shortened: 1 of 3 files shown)


Libraries/Components/SafeAreaView/SafeAreaView.js
```diff
@@ -40,11 +40,22 @@ if (Platform.OS === 'android') {
 } else {
   const RCTSafeAreaViewNativeComponent = require('RCTSafeAreaViewNativeComponent');
   exported = class SafeAreaView extends React.Component<Props> {
```
TheSavior: This should use forwardRef instead and assign the ref to the RCTSafeAreaViewComponent. Search the repo for forwardRef to check out other examples

Libraries/Components/SafeAreaView/SafeAreaView.js
```diff
@@ -31,24 +29,33 @@ let exported;
  * sensor housing area on iPhone X).
  */
 if (Platform.OS === 'android') {
+  let exported;
   exported = class SafeAreaView extends React.Component<Props> {
```
TheSavior: Can you use forwardRef for this one too? We’d want to be consistent on the exported interface of the module.

Libraries/Components/SafeAreaView/SafeAreaView.js
```diff
@@ -31,24 +29,33 @@ let exported;
...
+        {...props}
+        ref={forwardedRef}
+      />
+    );
   };
-}
 
-module.exports = exported;
+  const SafeAreaViewRef = React.forwardRef(SafeAreaView);
+  /* $FlowFixMe(>=0.89.0 site=react_native_fb) This comment suppresses an error
```
TheSavior: If you cast through ‘any’ then you won’t need this. Search for “: any” in the codebase to find an example.

Libraries/Components/SafeAreaView/SafeAreaView.js
zhongwuzw: Emm, actually, we don't have `SafeAreaView` for Android.

TheSavior: Yeah. What I mean is that we render a View on Android, it seems like we should forward the ref onto the View that is rendered instead. What do you think?

Libraries/Components/SafeAreaView/SafeAreaView.js
TheSavior: I think it would be 

module.exports = ((SafeAreaViewRef: any): Class<NativeComponent<Props>>)

Libraries/Components/SafeAreaView/SafeAreaView.js
```diff
@@ -19,8 +19,6 @@ type Props = $ReadOnly<{|
   emulateUnlessSupported?: boolean,
 |}>;
 
-let exported;
```
TheSavior: Since we are following the same pattern in both branches we can still assign to this and then have the one module.exports at the bottom

zhongwuzw:
CC. @TheSavior .

Libraries/Components/SafeAreaView/__tests__/SafeAreaView-test.js
```diff
@@ -0,0 +1,20 @@
...
+ * @emails oncall+react_native
+ * @flow
+ */
+
+'use strict';
+
+const SafeAreaView = require('SafeAreaView');
+
+describe('<SafeAreaView />', () => {
+  it('should set displayName to prevent <Component /> regressions', () => {
```
TheSavior: Can you add all 4 of the tests that are in the other PR? They all catch slightly different but related issues. We want the same tests that also render to snapshots.

zhongwuzw: @TheSavior Added!

zhongwuzw:
@TheSavior Thanks!!! 💯

cpojer:
I'll merge it, and then Eli can monitor whether it works fine at FB later today :) #handoff

PR #24593, opened 2019-04-25
add snapshots for mocked and unmocked components part 2

bcarroll22:
## Summary

Per conversation with @TheSavior, in #24538, this adds snapshot tests for more components. Shallow and deep snapshots are included.

## Changelog

[General] [Added] - Snapshots

## Test Plan

This PR adds a lot of tests 😄


Libraries/Components/Picker/__tests__/Picker-test.js
```diff
@@ -0,0 +1,63 @@
...
+
+const React = require('React');
+const Picker = require('../Picker');
+
+const render = require('../../../../jest/renderer');
+
+describe('<Picker />', () => {
+  it('should render as <View> when mocked', () => {
+    const instance = render.create(
+      <Picker selectedValue="foo" onValueChange={jest.fn()}>
```
TheSavior: Should we be abstracting this out into a constant at the top? It’s pretty important that they are consistent across all of the tests, right? Probably should do that for all of the tests.

Libraries/Components/ProgressBarAndroid/__tests__/ProgressBarAndroid-test.js
```diff
@@ -0,0 +1,52 @@
...
+ * @format
+ * @emails oncall+react_native
+ * @flow
+ */
+
+'use strict';
+
+const React = require('React');
+// $FlowFixMe
+const ProgressBarAndroid = require('../ProgressBarAndroid.android');
```
TheSavior: @cpojer do you know how to fix this?

Libraries/Modal/__tests__/__snapshots__/Modal-test.js.snap
```diff
@@ -0,0 +1,79 @@
...
+      >
+        <View />
+      </View>
+    </View>
+  </View>
+</RCTModalHostView>
+`;
+
+exports[`<Modal /> should shallow render as <Component> when mocked 1`] = `
+<Component
```
TheSavior: This is an example of displayName being broken. We’ll need to fix this. Can you either fix it in this PR or change the name of this test to be consistent with the other component tests so it is clear this isn’t correct?

Libraries/Components/ProgressBarAndroid/__tests__/ProgressBarAndroid-test.js
cpojer: This is fine. The iOS flow config doesn't see these modules. I don't think it matters much.

Libraries/Modal/__tests__/__snapshots__/Modal-test.js.snap
bcarroll22: 🤔 hmm modal is a mocked component. Are you sure this line isn't what's causing this? 

https://github.com/facebook/react-native/blob/master/jest/mockComponent.js#L20

TheSavior: Yeah, that’s probably it but my understanding was the name line in the render function of that file is the thing that should be setting the component’s display name. Interestingly, that doesn’t seem to show up in this snapshot test at all.

bcarroll22: Sorry this took a bit. This goes back to my original PR, and this is related to what prompted all of this 😬 I tried explaining it but confused myself and deleted it all. I've been doing a bad job explaining it thus far so I think I need some time to put together a cohesive description.

TheSavior:
I'm gonna land this as is, we can continue to iterate and improve

PR #24595, opened 2019-04-25
custom fonts support The Android Way

dulmandakh:
## Summary

In https://github.com/facebook/react-native/pull/23865, RN introduced support for custom fonts the Android Way. But it introduced performance regression because it'll lookup for a font using getIdentifier() every time fontFamily changed. This PR fixes regression by requiring custom fonts to be listed in **fonts** array, and populating **mTypeCache** at first use using the list.

## Changelog

[Android] [Changed] - Require custom fonts to list in **fonts** array. Fixes performance regression. 

## Test Plan

CI is green and RNTester app works as expected.


ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -80,6 +83,27 @@ public static ReactFontManager getInstance() {
     String fontFamilyName,
     int style,
     Context context) {
+
+    if (!mTypeCacheLoaded) {
+      Resources resources = context.getResources();
+      int fontsId = resources.getIdentifier("fonts", "array", context.getPackageName());
```
fkgozali: Unfortunately, even 1 call to getIdentifier() would cause big regression. So I don't think we can do this without any ability to disable to lookup (e.g. via provider model I described in https://github.com/facebook/react-native/pull/23865/files#r278373255

dulmandakh:
This is only one for an instance, and it's singleton. IMHO, asking
developers to write a custom provider code just to have a custom font is a
bad DX.

fkgozali:
> This is only one for an instance, and it's singleton.

It's still gonna be in the RN startup path, so it would still affect startup perf.

> IMHO, asking developers to write a custom provider code just to have a custom font is a
bad DX.

You can always provide a default provider, and have the ability for apps to change or disable the default provider.

cpojer:
@dulmandakh to clarify, calling this method in RN's startup path means that it will literally add 10s of ms to the startup. FB scale is crazy.

dulmandakh:
ok, i'll figure out a solution for it, thank you.

dulmandakh:
@fkgozali changed it to look into meta-data in AndroidManifest.xml, thus not lookup resources by name. It requires some effort from developers, but it's much easier and clean way to add custom fonts to Android.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +81,49 @@ public static ReactFontManager getInstance() {
...
+        ApplicationInfo app = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+        Bundle bundle = app.metaData;
+        int fontsId = bundle.getInt("rn_fonts", 0);
+        if (fontsId != 0) {
+          Resources resources = context.getResources();
+          TypedArray fonts = resources.obtainTypedArray(fontsId);
+          for (int i = 0; i < fonts.length(); i++) {
+            int fontId = fonts.getResourceId(i, 0);
+            if (fontId != 0) {
+              Typeface font = ResourcesCompat.getFont(context, fontId);
```
hey99xx: How expensive is loading `Typeface` instances for every single font of the app into the cache and calling `resources.getResourceEntryName ` at startup?

If it is also taking too long (for FB scale), although it's verbose you might do something like:

```
<array name="fonts">
    
    <item>srisakdi</item>
    <item>@font/srisakdi</item>
    
    <item>anotherfont</item>
    <item>@font/anotherfont</item>

</array>
```

This way you can iterate the array in increments of 2, and get away with only caching `fontFamily`->`fontId` mapping at startup, and only inflate `fontId`->`Typeface` still lazily.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +81,49 @@ public static ReactFontManager getInstance() {
...
+  public @Nullable Typeface getTypeface(
+    String fontFamilyName,
+    int style,
+    Context context) {
+
+    if (!mTypeCacheLoaded) {
+      try {
+        ApplicationInfo app = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+        Bundle bundle = app.metaData;
+        int fontsId = bundle.getInt("rn_fonts", 0);
```
hey99xx: Can this be namespaced like `com.facebook.react.font.resources`?

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
hey99xx: Also this should allow font file names to not strictly match font family names, ex:

```
<item>Open Sans</item>
<item>@font/open_sans</item>
```

Native devs will be more familiar underscored naming `R.font.open_sans` whereas web developers will be more familiar with human readable naming `fontFamily: "Open Sans"`. 

This will allow sharing cross platform styling code with iOS, where font family name is already retrieved from font file content, not from file names.

dulmandakh: I tried to mimic iOS feature, where we only list font files, but also make typo-proof. Also I couldn't find anything about performance impact of getResourceEntryName, and hope it won't have impact on performance.

hey99xx: iOS works a bit differently, it reads font family name from the file content, Android in your implementation reads from the file name.
https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app
https://stackoverflow.com/questions/16788330/how-do-i-get-the-font-name-from-an-otf-or-ttf-file

IMO this will force developers to write strange JS code like: 
`{fontFamily: Platform.OS == 'ios' ? 'Open Sans' : 'open_sans'}`
I dont know why typo-proof is such an important thing, you only need to match with what you use in JS code, you could make a mistake in font file name too.

---
Just looking at getResourceEntryName implementation in AOSP repo, it's implemented in C++ so you will at least jump between JNI bridge for each call.

dulmandakh: Yep, I found that Android doesn't allow to use font family name, but resource filename without extension. So difference is inevitable.

With this change we can use various font weights like medium easily. Currently, you have to write something like which makes text style sharing difficult.
```js
{
   fontFamily: Platform.OS == 'ios' ? 'Open Sans' : 'open_sans_medium'
   fontWeight: Platform.OS == 'ios' ? '500' : 'normal'
}
```

once PR lands you will write

```js
{
   fontFamily: Platform.OS == 'ios' ? 'Open Sans' : 'open_sans'
   fontWeight: '500'
}
```

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
dulmandakh: removed all getIdentifier() calls

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +81,49 @@ public static ReactFontManager getInstance() {
...
+        if (fontsId != 0) {
+          Resources resources = context.getResources();
```
fkgozali: Actually, I'm not sure if accessing anything dynamically from Resources is fast, because it needs to crack them open from the APK.

Can we make this **not do any resource lookup** by default and only perform this lookup if the hosting app configures it? You don't really need to change it to use AndroidManifest, using font resource is fine, but I'd like:
* no lookup by default
* do custom lookup with any impl for now only if the hosting app enables it explicitly

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
hey99xx: You can name your file "Open Sans.ttf", "Open Sans_bold.ttf"... etc though, so the code can already be cross platform `fontFamily: "Open Sans"`. Can you leave a space in these font resource file names and reuse that resource in Java? IMO it's not good practice to enforce android resource naming conventions into JS land.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
hey99xx: I think @dulmandakh is trying to find a way where non-native developers don't have to touch java code to customize these things, that's why he switched to using AndroidManifest that's supposed to be a cheaper alternative to resource lookups.

If an app like FB does not provide "rn_fonts" in AndroidManifest.xml, there will never be any look up (besides looking to AndroidManifest itself), isn't that sufficient?

dulmandakh: got it. I'll figure out another solution :D

dulmandakh:
@fkgozali another implementation which addresses your concerns, but also requires less work from developers.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -7,15 +7,18 @@
...
 
+import android.content.Context;
 import android.content.res.AssetManager;
 import android.graphics.Typeface;
 import android.util.SparseArray;
 
+import androidx.annotation.Nullable;
+import androidx.core.content.res.ResourcesCompat;
+
+
```
fkgozali: remove extra newline

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -37,9 +40,12 @@
   private static ReactFontManager sReactFontManagerInstance;
 
   private Map<String, FontFamily> mFontCache;
+  private Map<String, Typeface> mTypeCache;
+  private boolean mTypeCacheLoaded = false;
```
fkgozali: I think a better set of names is:
* `mCustomFontCache`
* `mHasCustomFont`

fkgozali: or:
* `mCustomTypefaceCache`
* `mHasCustomTypeface`

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +83,15 @@ public static ReactFontManager getInstance() {
     return typeface;
   }
 
+  public void loadFont(Context context, int fontId) {
```
fkgozali: should rename to `addCustomFont()` to make it clearer

fkgozali: Can you also add comment for how to call this method? E.g. what you did for RNTester. Basically document how one can install a custom font support in their apps.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +83,15 @@ public static ReactFontManager getInstance() {
...
+    Typeface font = ResourcesCompat.getFont(context, fontId);
+    if (font != null) {
+      String fontFamily = context.getResources().getResourceEntryName(fontId);
```
fkgozali: This won't affect FB with this PR, but I wonder if this is efficient in practice. You don't have to address it in this PR, but I think allowing apps to provide a custom "typeface resolution" function can allow them to address slow lookup. For instance, if one day FB wants to use custom fonts, I'd actually implement the resolver via a switch statement, not via asking `getResourceEntryName()`. Most apps probably don't need such optimization, but having an option to do optimized thing is good.

fkgozali: > Most apps probably don't need such optimization, but having an option to do optimized thing is good.

An example pseudo solution:

```
public interface ReactTypefaceResolver {
  public String resolveTypefaceName(int fontId);
}
```

Then introduce a method in ReactFontManager:

```
public void setCustomTypefaceResolver(ReactTypefaceResolver resolver) {
  mCustomTypefaceResolver = resolver;
}
```

Then you can just use `mCustomTypefaceResolver != null` as the flag, no need to have the boolean flag you had above. Then you can just call it inside `loadFont()`.

dulmandakh:
Thank you for the detailed review. I'll make the changes you suggested.

dulmandakh:
@fkgozali renamed variables and removed resource lookup code, therefore resolved feature parity with iOS because we can assign any font family name to a font. Thank you

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -60,6 +65,13 @@ public static ReactFontManager getInstance() {
...
+    if(mHasCustomTypeface && mCustomTypefaceCache.containsKey(fontFamilyName)) {
+      return Typeface.create(
+        mCustomTypefaceCache.get(fontFamilyName),
+        style
+      );
+    }
+
     Typeface typeface = fontFamily.getTypeface(style);
     if (typeface == null) {
       typeface = createTypeface(fontFamilyName, style, assetManager);
```
hey99xx: Is there value caching invidual styled Typeface (normal/italic/bold/boldItalic) instances here like it's done for fonts in assets? Besides that all LGTM.

dulmandakh: to use styles or weights you have to define fonts using XML.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -60,6 +65,13 @@ public static ReactFontManager getInstance() {
       mFontCache.put(fontFamilyName, fontFamily);
     }
 
+    if(mHasCustomTypeface && mCustomTypefaceCache.containsKey(fontFamilyName)) {
```
fkgozali: nit: add space after `if`

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -37,9 +40,12 @@
   private static ReactFontManager sReactFontManagerInstance;
 
   private Map<String, FontFamily> mFontCache;
+  private Map<String, Typeface> mCustomTypefaceCache;
+  private boolean mHasCustomTypeface = false;
```
fkgozali: Technically you could just make `mCustomTypefaceCache` `@Nullable`, then default to `null`. If someone adds a custom typeface, you initialize it as a new HashMap, then you don't need `mHasCustomTypeface` boolean. You can just check if `mCustomTypefaceCache != null`

dulmandakh: Yes, my first version was like that but changed it to use booleans just before the commit because it looked cleaner.

dulmandakh:
@hramos it's not merged yet. Could you please investigate it. Thank you

fkgozali:
> it's not merged yet. Could you please investigate it. Thank you

No one kicked off the merging. I still need to run the simulated perf test to verify there's no regression.

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
fkgozali: from @mdvacca:

let's add final to these variables

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +83,21 @@ public static ReactFontManager getInstance() {
...
   }
 
+  /*
+   * This method allows you to load custom fonts from res/font folder as provided font family name.
+   * Fonts may be one of .ttf, .otf or XML (https://developer.android.com/guide/topics/ui/look-and-feel/fonts-in-xml).
+   * To support multiple font styles or weights, you must provide a font in XML format.
+   *
+   * ReactFontManager.getInstance().addCustomFont(this, "Srisakdi", R.font.srisakdi);
+   */
+  public void addCustomFont(Context context, @NonNull String fontFamily, int fontId) {
```
fkgozali: from @mdvacca:

should we add `@NonNull` to context?

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
```diff
@@ -71,6 +83,21 @@ public static ReactFontManager getInstance() {
...
+    Typeface font = ResourcesCompat.getFont(context, fontId);
+    if (font != null) {
+      mCustomTypefaceCache.put(fontFamily, font);
+    }
+    mHasCustomTypeface = true;
```
fkgozali: From @mdvacca:

doesn't this need to be inside the if (font != null) {
also, why not to just remove this variable and create a private method:

```
private boolean hasCustomTypeface()  {
    return mCustomTypefaceCache.isEmpty();
}
```
or maybe we don't even need this variable, it is just preventing a call to Map.containsKey?

fkgozali:
> I still need to run the simulated perf test to verify there's no regression.

seems like we can land this for now, just need to address the comments here before we land

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactFontManager.java
dulmandakh: we pass resource id to this method, but that ID might not be a font, but anything. So font != null is making sure that we have a font with provided id.

PR #24608, opened 2019-04-25
New Accessibility states API.

marcmulcahy:
## Summary

As currently defined, accessibilityStates is an array of strings, which represents the state of an object. The array of strings notion doesn't well encapsulate how various states are related, nor enforce any level of correctness.

This PR converts accessibilityStates to an object with a specific definition. So, rather than:

<View
...
accessibilityStates={['unchecked']}>

We have:

<View
accessibilityStates={{'checked': false}}>

And specifically define the checked state to either take a boolean or the "mixed" string (to represent mixed checkboxes).

We feel this API is easier to understand an implement, and provides better semantic definition of the states themselves, and how states are related to one another.

## Changelog

[general] [change] - Convert accessibilityStates to an object instead of an array of strings.

## Test Plan

Converted code in RNTester to use the new accessibilityStates API, and modified the checkbox example to support the "mixed" state, which was one of the limitations of the current API which prompted this work.


necolas:
For context, this is a change I proposed [here](https://github.com/facebook/react-native/pull/24095#discussion_r273179033) to provide an API that works for platforms where attributes must be present on host nodes to communicate their purpose to the native accessibility system

RNTester/js/AccessibilityExample.js
```diff
@@ -109,7 +109,7 @@ class AccessibilityExample extends React.Component {
           <TouchableOpacity
             onPress={() => Alert.alert('Button has been pressed!')}
             accessibilityRole="button"
-            accessibilityStates={['disabled']}
+            accessibilityStates={{enabled: false}}
```
necolas: Let's not change from `disabled` to `enabled`. This should be `disabled: true`

ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
```diff
@@ -99,6 +102,13 @@ public static AccessibilityRole fromValue(@Nullable String value) {
     }
   }
 
+  // State constants for states which have analogs in AccessibilityNodeInfo
+
+  private static final String STATE_ENABLED = "enabled";
+  private static final String STATE_SELECTED = "selected";
+  private static final String STATE_CHECKED = "checked";
+  private static final String STATE_HAS_POPUP = "hasPopup";
```
necolas: I think `haspopup` is probably not a state but a relationship. This property comes from web's ARIA spec, where the accepted value must match the role of the popup container - https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup

RNTester/js/AccessibilityExample.js
marcmulcahy: #24095  too.

marcmulcahy: @necolas Not sure about that. I think iOS is the only platform that defines this state in the negative. Wouldn't it be better to align with the rest of the world on this? Especially if we're moving to a new API.

marcmulcahy: @necolas Not sure about that. :) As far as I know, iOS is the only platform that defines this state in the negative (disabled instead of enabled). Shouldn't we take this opportunity to align with the majority of the platforms, especially given that we're proposing a braking API change anyway?

ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
marcmulcahy: I could support removing this as a state, and handling it in a accessible relationships PR, but then we should submit a PR to remove it from #24095 too.

marcmulcahy:
@necolas Any thoughts about how to deprecate the older API, while strongly encouraging developers to move to the new one?

RNTester/js/AccessibilityExample.js
necolas: Web uses `disabled` too. Let's leave this as `disabled`

ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
necolas: sounds good

necolas:
> Any thoughts about how to deprecate the older API, while strongly encouraging developers to move to the new one?

Hopefully someone from RN Core will be able to advise

RNTester/js/AccessibilityExample.js
cpojer: I agree with @necolas.

cpojer:
Ideally we make this a non-breaking change for now by supporting both arrays or objects for a while, then we’ll add a warning in the release after, and then release a codemod (I can help with that) and then remove it.

RNTester/js/AccessibilityExample.js
marcmulcahy: Done.

marcmulcahy:
@cpojer Not sure how to allow the native code to accept both an array of strings as well as an object. Any ideas?

cpojer:
Unfortunately we can't make a breaking change like that in a single pull request. We could wrap `<View />` in JS, which would be slow and which we can't do. It seems like the only option here is to introduce a new prop and deprecate the existing one (in a separate PR later) which I don't think we want to do. I'll ask at FB if there are other possibilities but unless we want to introduce a new prop name this may not easily be possible atm.

cpojer:
@marcmulcahy do you have any idea for a new prop that we could introduce so that we could deprecate the existing one? I know this is unfortunate but it seems like the only way. If not, please close this PR.

necolas:
Is there really no way to fix this without renaming the prop?

marcmulcahy:
@cpojer I suppose we could simply add a new prop called something like accessibilityStates2 that's an object instead of an array of strings. Switch all the RNTester samples and components to use accessibilityStates2, and then eventually deprecate accessibilityStates, or rename accessibilityStates2 to accessibilityStates.

I do think this object-based API is much clearer, in that the component author must enumerate the states that "matter", and it makes much clearer how states are related to each other (expanded/collapsed, checked/unchecked, etc.).

I'm also happy to close the PR if this isn't something we want to tackle ATM. Let me know what FB prefers to do.

cpojer:
Alright, yeah I think we can do this. Can you change this PR to do the following:
* Add deprecatedAccessibilityStates and make the native code work with that
* Make it so accessibilityStates still keeps working exactly the same after your PR lands.

In a month from now (sorry, this is some fb internal issue we have to deal with) we can then change how accessibilityStates works as long as we don't have internal users for it any longer. A month after that we can bring back the accessibilityStates prop with the new API and kill the deprecatedAccessibilityStates API.

I'm sorry for this slow multi process step but I think it's worth doing it if you are up for it :)

cpojer:
Alternatively, if we can think of a better name than `accessibilityStates`, we can also just add the new prop with a new name and then remove the old one in a month from now. That is way less involved and faster. The process is:

* Add a new prop with the new API
* I will migrate all internal callsites at FB
* Send a PR with the removal for the old `accessibilityStates` that can be merged in a month from now.

What would be a reasonable name for a new prop?

necolas:
> What would be a reasonable name for a new prop?

`accessibilityState`? hah

cpojer:
I think that’s a great idea. Let’s do it!

marcmulcahy:
As we discussed, this version moves the new API to a new property called "accessibilityState" and leaves "accessibilityStates" in tact.

To verify that the existing prop was still working, I applied all the patches which add the new property, and left RNTester unchanged, and verified that RNTester behaves as before on both Android and iOS. I then modified RNTester to use "accessibilityState" instead of "accessibilityStates" and re-tested.

necolas:
Looks good to me

Issue #24080, opened 2019-03-21
iOS: respect "withCredentials: false" in XMLHttpRequest


DimitryDushkin:
## 🐛 Bug Report

Even with `withCredentials: false` network requests still sending cookies.

## To Reproduce

1. Request yandex.ru (it will set `yandexuid` cookie in response)
2. Request anything else with `withCredentials: false` and you will see that cookies are sent

## Expected Behavior

No cookies with `withCredentials: false`.

## Code Example

Actually there is very easy fix to it
```diff
--- node_modules/react-native/Libraries/Network/RCTNetworking.mm	2019-03-18 17:28:57.000000000 +0300
+++ node_modules/react-native/Libraries/Network/RCTNetworking.mm	2019-03-18 17:29:36.000000000 +0300
@@ -245,10 +245,13 @@
   NSURL *URL = [RCTConvert NSURL:query[@"url"]]; // this is marked as nullable in JS, but should not be null
   NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:URL];
   request.HTTPMethod = [RCTConvert NSString:RCTNilIfNull(query[@"method"])].uppercaseString ?: @"GET";
+  request.HTTPShouldHandleCookies = [RCTConvert BOOL:query[@"withCredentials"]];

-  // Load and set the cookie header.
-  NSArray<NSHTTPCookie *> *cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookiesForURL:URL];
-  request.allHTTPHeaderFields = [NSHTTPCookie requestHeaderFieldsWithCookies:cookies];
+  if (request.HTTPShouldHandleCookies == YES) {
+      // Load and set the cookie header.
+      NSArray<NSHTTPCookie *> *cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookiesForURL:URL];
+      request.allHTTPHeaderFields = [NSHTTPCookie requestHeaderFieldsWithCookies:cookies];
+  }

   // Set supplied headers.
   NSDictionary *headers = [RCTConvert NSDictionary:query[@"headers"]];
@@ -259,7 +262,6 @@
   }];

   request.timeoutInterval = [RCTConvert NSTimeInterval:query[@"timeout"]];
-  request.HTTPShouldHandleCookies = [RCTConvert BOOL:query[@"withCredentials"]];
   NSDictionary<NSString *, id> *data = [RCTConvert NSDictionary:RCTNilIfNull(query[@"data"])];
[...]

PR #24629, opened 2019-04-26
Fix: respect "withCredentials: false" in RCTNetworking iOS

DimitryDushkin:
Fixes https://github.com/facebook/react-native/issues/24080.

## Summary
Even with `withCredentials: false` network requests still sending cookies. Fix this behaviour according to https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials.

## Changelog

[iOS] [Fixed] - Respect "withCredentials: false" in network requests

## Test Plan
1. GET request https://yandex.ru (it will set "yandexuid" cookie in response).
2. GET request https://yandex.ru again with `withCredentials: false`

Actual behaviour: "Cookies" header is present in request.
Expected behaviour: no "Cookies" header in request


DimitryDushkin:
@cpojer glad to help!
I think it's worth to warn about this in changelog, in case if someone rely on cookies in an app and should re-check this.

cpojer:
Yep, absolutely. It's a breaking changes but makes sense for consistency. `withCredentials` is `true` by default so hopefully this won't affect too many people.

cc @turnrye just to make sure this is on your radar for 0.61.

Issue #24626, opened 2019-04-26
☂️ Fix annoying Android and iOS issues


cpojer:
## 🐛 Bug Report

Through https://github.com/react-native-community/discussions-and-proposals/issues/104 and https://github.com/react-native-community/discussions-and-proposals/issues/69 we learned about a lot of issues that people pointed out about React Native (cc @slorber @noahtallen @alburdette619). While we addressed many, there are still many issues remaining that need to be validated and if they still persist, fixed.

If you'd like to work on any of these, please comment here to sign up for one of them and send a Pull Request with a fix!

* [x] Using `style={{width: "100"}}` and forgetting the %
* [x] Toggling secureTextEntry has been broken for years
* [x] https://twitter.com/estevao_lucas/status/1117572702083190785?s=215 
* [ ] On Android, you cannot name a property flex or the app will crash
* [ ] Animated translate errors: https://github.com/react-native-community/discussions-and-proposals/issues/69#issuecomment-450532964
* [ ] Issues with Animated.View: https://github.com/react-native-community/discussions-and-proposals/issues/69#issuecomment-482885975
* [ ] Unhelpful error message with ScrollViews. See https://twitter.com/jlongster/status/1123225841075392514

Thank you :)

guhungry:
One of the toggle secureTextEntry issue(s) was addressed in https://github.com/facebook/react-native/pull/23524

sasurau4:
I want to work on `style={{ width: "100" }}` and forgetting the % issue.
I have a plan that fix the issue by translate `style={{ width: "100"}}` to `style={{ width: "100%" }}` automatically with warning rather than error.
Is the plan good? If not, please tell me other plans :pray:

cpojer:
@sasurau4: give it a try! I’ll wait for your pull request. I think just making sure the app doesn’t crash and shows a useful error message is a good start!

ahmedu007:
I can give the 3rd issue a go. *On Android, you cannot name a property flex or the app will crash*

Can I get a bit of context, and then I can start digging into it

matheusmonte:
Can work in "On Android, you cannot name a property flex or the app will crash" ?

@cpojer

cpojer:
Hey @matheusmonte, give it a try! Unfortunately I don't have more context myself but maybe @alburdette619 (who mentioned this initially) can provide a repro?

zhongwuzw:
> https://twitter.com/estevao_lucas/status/1117572702083190785?s=215

[...]

PR #24630, opened 2019-04-27
[iOS] Add convert compatible of NSString for bridge message data

zhongwuzw:
## Summary

Fixes https://twitter.com/estevao_lucas/status/1117572702083190785?s=215 in #24626 . Now we try to convert any id to `NSString`, not throw error.

cc. @cpojer .

## Changelog

[iOS] [Fixed] - Add convert compatible of NSString for bridge message data

## Test Plan

If we set other types like number, we can convert it to `NSString` and not throw error.


cpojer:
@zhongwuzw seems like there is a test that is now failing. Do you mind fixing it up?

zhongwuzw:
@cpojer Ahah, fix it :)

PR #24631, opened 2019-04-27
Gradle KTS

dulmandakh:
## Summary

Convert root Gradle script to Kotlin DSL, and cleanup. Currently, there is not much benefit or advantage over Groovy scripts, except IDE support and it'll cache compiled KTS scripts on first run.

## Changelog

[Android] [Changed] - Convert root Gradle script to Kotlin DSL, and cleanup.

## Test Plan

CI is green, and RNTester builds, runs as expected


RNTester/android/app/build.gradle
```diff
@@ -4,10 +4,9 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
-apply plugin: 'com.android.application'
-
-import com.android.build.OutputFile
+plugins {
```
gengjiawen: File extension looks like not changed.

ReactAndroid/build.gradle
```diff
@@ -6,7 +6,7 @@
 plugins {
     id("com.android.library")
     id("maven")
-    id("de.undercouch.download")
+    id("de.undercouch.download") version "3.4.3"
```
gengjiawen: same.

RNTester/android/app/build.gradle
dulmandakh: I'm touching only top level Gradle scripts this time.

ReactAndroid/build.gradle
dulmandakh: Version is required because I removed it from top level build.gradle.

RNTester/android/app/build.gradle
gengjiawen: I think we better do it all.

ReactAndroid/build.gradle
gengjiawen: I mean why not `build.gradle.kts`

RNTester/android/app/build.gradle
dulmandakh: it'll require a lot of work, and it's not trivial to convert to Kotlin DSL due to static typing

dulmandakh:
I'm playing with Gradle and it's API's, trying to see what we can improve. I think that we can have a lot improvements if we develop Gradle React Plugin, but it'll take time.

Issue #24607, opened 2019-04-25
Regression: iOS upgrading 57 -> 59 lots of crashes "Exception in HostObject::get: <unknown>"


DimitryDushkin:
## 💥 Regression Report
We've released iOS app with upgrade RN from 0.57 to 0.59.4 and receiving a lot of crash reports (like 5% of users so far)

Crash:
[Crash file](https://github.com/facebook/react-native/files/3117452/crash_13553525293423706268_2.3.0.txt)

```
Application Specific Information:
*** Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception: Exception in HostObject::get: <unknown>', reason: 'Unhandled JS Exception: Exception in HostObject::get: <unknown>, stack:
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
d@2:868
global code@12:4
'

Thread 7 crashed with ARM-64 Thread State:

Extra Information:

Application Stats:
{
    "active_time_since_last_crash": 7.45225e+10,
    "active_time_since_launch": 0,
    "application_active": false,
    "application_in_foreground": true,
    "background_time_since_last_crash": 4.64477e+10,
    "background_time_since_launch": 0,
    "launches_since_last_crash": 419,
    "sessions_since_last_crash": 727,
    "sessions_since_launch": 1
}

CrashDoctor Diagnosis: Application threw exception RCTFatalException: Unhandled JS Exception: Exception in HostObject::get: <unknown>: Unhandled JS Exception: Exception in HostObject::get: <unknown>, stack:
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
h@2:1670
d@2:868
global code@12:4

Originated at or in a subcall of facebook::react::JSIExecutor::defaultTimeoutInvoker(std::__1::function<void ()> const&, std::__1::function<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > ()>)
```

## Last working version
0.57

Worked up to version:

Stopped working in version:
0.59.4

## To Reproduce
Random and rare crash on app startup.

## Environment

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
      Memory: 50.57 MB / 16.00 GB
      Shell: 3.0.2 - /usr/local/bin/fish
    Binaries:
[...]

PR #24633, opened 2019-04-27
Report native module name on crash when native module has failed to load

DimitryDushkin:
## Summary

After upgrading RN from 0.57 to 0.59.4 we've received a lot of crash reports like `Exception in HostObject::get: <unknown>`  with no clue what native module caused the crash. This commit adds native module name on crash in this situations. Related to https://github.com/facebook/react-native/issues/24607.

## Changelog

[General] [Added] - Report native module name on crash when native module has failed to load

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -214,11 +214,11 @@ class JSCRuntime : public jsi::Runtime {
   static JSStringRef stringRef(const jsi::String& str);
   static JSStringRef stringRef(const jsi::PropNameID& sym);
   static JSObjectRef objectRef(const jsi::Object& obj);
-    
+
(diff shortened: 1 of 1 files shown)


mhorowitz:
@DimitryDushkin  thanks for this change!

Sorry for the late feedback, but I have a couple more comments about it.  First, the wording is a little confusing, as <unknown> previously referred to the exception, not the property name, and now that's gone.  The same pattern occurs at lines 684, 721, and 728.  Can you make this change consistently throughout, so the exception always includes the property name?  Thanks!

DimitryDushkin:
Yeah, sure, I'll do it.

DimitryDushkin:
>  Can you make this change consistently throughout, so the exception always includes the property name? 
https://github.com/facebook/react-native/pull/24704

Issue #24564, opened 2019-04-23
@react-native-community/eslint-config "fb" should be object


telmen:
## 🐛 Bug Report

installing and running [@react-native-community/eslint-config](https://www.npmjs.com/package/@react-native-community/eslint-config) on git hooks gives an error

## Expected Behavior

`prettier/prettier` should be executed without errors

```sh
 Error: @react-native-community:
	Configuration for rule "prettier/prettier" is invalid:
	Value "fb" should be object.
```

## Additional info

I looked at https://github.com/facebook/react-native/blob/master/packages/eslint-config-react-native-community/index.js#L219 and it seems second item of passing array option is string and I think it should be object. 

## Environment

```  React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (4) x64 Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz
      Memory: 97.19 MB / 16.00 GB
      Shell: 5.6.2 - /usr/local/bin/zsh
    Binaries:
      Node: 10.11.0 - /usr/local/bin/node
      Yarn: 1.13.0 - ~/.yarn/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 19, 22, 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.2, 27.0.3, 28.0.1, 28.0.2, 28.0.3, 29.0.0
[...]

PR #24635, opened 2019-04-28
Set the Prettier config so it is not forced on users of @react-native-community/eslint-config

matt-oakes:
## Summary

This fixes an issue where the Prettier config was set to the `fb` (Facebook) values for all users of the `@react-native-community/eslint-config` package. This was due to [this line](https://github.com/facebook/react-native/blob/8f186b84aeeb2613bf6ae08f20a8547d40179007/packages/eslint-config-react-native-community/index.js#L219) in the config file.

It was causing issues like these:

* Errors when using newer versions of `eslint-plugin-prettier` (you had to use a version that was >1 year old): https://github.com/facebook/react-native/issues/24564
* Errors due to the Prettier parser being forced to be `flow` when using Typescript: https://github.com/typescript-eslint/typescript-eslint/issues/481

This PR:

* Changes that line to remove the explicit `fb` config so users can set their own.
* Moves the React Native Prettier config to `.prettierrc` so ESLint, Prettier, and code editors can all read from the same place.
* Upgrades both `prettier` and the `eslint-plugin-prettier` to the latest versions.

## Changelog

[General] [Fixed] - Stopped the Prettier config being set for all users of @react-native-community/eslint-config

## Test Plan

* ✅ Ensure there is no difference on this repo (no ESLint errors, same number of warnings, and no changes when running prettier.
* ✅ Test that the new config does not set the Prettier settings on projects which use it.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "requirePragma": true,
+  "singleQuote": true,
+  "trailingComma": "all",
+  "bracketSpacing": false,
+  "jsxBracketSameLine": true
+}
--- a/package.json
+++ b/package.json
@@ -11,14 +11,6 @@
   "engines": {
     "node": ">=8.3"
   },
-  "prettier": {
-    "requirePragma": true,
-    "singleQuote": true,
-    "trailingComma": "all",
-    "bracketSpacing": false,
-    "jsxBracketSameLine": true,
-    "parser": "flow"
-  },
   "jest-junit": {
     "outputDirectory": "reports/junit",
     "outputName": "js-test-results.xml"
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -216,7 +216,7 @@ module.exports = {
 
     // Prettier Plugin
     // https://github.com/prettier/eslint-plugin-prettier
-    'prettier/prettier': [2, 'fb', '@format'],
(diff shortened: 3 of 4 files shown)


cpojer:
Upgrading prettier can only be done at FB unfortunately. Any chance you could update your PR to do everything except upgrading prettier?

j-f1:
Does removing the `parser: flow` line cause any errors? If not, I’d recommend removing it since Prettier automatically infers which parser to use based on the file type.

matt-oakes:
@cpojer I've removed the two dependency upgrades 👍 

@j-f1 You are right, and as that was the cause of the issues before, I think it's best to remove it if possible. Prettier still correctly parses the `js` files as `flow` when removing that line and it won't break if we change to another format for some files.

matt-oakes:
@cpojer Thanks! Once this lands can you do a new release of the `@react-native-community/eslint-config` package?

cpojer:
@michalchudziak handles the releases for this package although I think you should also be able to.

matt-oakes:
@cpojer In that case I should have bumped the version number as well. I've submitted another PR to do that and then I'll try a release. Sorry about that!

https://github.com/facebook/react-native/pull/24650

nicovak:
Thanks for the update. Currently, I had to create a `.prettierrc` file, to get it configured as wanted.
I guess It will be part of next react native update ?

PR #24638, opened 2019-04-28
feat: Do not require `pod install` after `react-native init`

grabbou:
## Summary

This pull request pre-generates CocoaPods files by running `pod install` locally inside a `template` before we publish it.

## Rationale:

We have recently migrated default template from a regular iOS project to the one powered by CocoaPods. This was done to simplify set up required and make it easier to handle native modules in the future too.

Unfortunately, one drawback was that `pod install` was now going to be required to be run as an additional command, after `react-native init` has completed.

The detailed discussion and rationale can be found here: https://github.com/facebook/react-native/pull/23563#issuecomment-486003489

## Implementation:

We install Node dependencies inside a `template`. That also includes installing React Native, which at this point, points towards a previously generated `tgz` archive. Thanks to that, we are consuming an archive that will be published to the registry - same set of files.

Once dependencies are installed, we install CocoaPods dependencies (by running `pod install` command inside `ios`). Thanks to Node dependencies being installed, `../node_modules/react-native` paths are resolved correctly.

Once Pods folder and Pods.xcodeproj has been generated, we remove `node_modules` and proceed to the next step, which is running `react-native init` itself.

At this point, we have `Pods` folder with all files generated. No steps required after `react-native init` completed - just `react-native run-ios` and we are set.

Note: Once CLI supports passing path to a template as `.tgz` instead of a folder, we should `npm pack` the template too. That way, we can make sure that `Pods` is present in the archive (and on npm).

### DIFF
--- a/scripts/prepare-template.sh
+++ b/scripts/prepare-template.sh
@@ -22,5 +22,5 @@ npm install
 # Dependencies are installed on the client-side
 rm -rf node_modules
 
-# Set the React Native version to point to the version fron the registry
-node scripts/set-rn-template-version.js "$1"
+# Do not ship a package.json that points to a `tgz`
+git checkout template/package.json
--- a/scripts/prepare-template.sh
+++ b/scripts/prepare-template.sh
@@ -6,17 +6,12 @@
 
 PACKAGE_LOCATION=$(pwd)/react-native-$1.tgz
 
-# Part 1:
-#
 # Pack React Native into a `.tgz` file so we can run from source
 npm pack
 
 # Set the React Native version to point to the `.tgz` file
 node scripts/set-rn-template-version.js "file:$PACKAGE_LOCATION"
-success "React Native version changed in the template"
 
-# Part 2:
-#
 # We need to generate CocoaPods project. To do so, we install depdendencies
 # locally and manually run `pod install`
 cd template
@@ -26,3 +21,6 @@ npm install
 
 # Dependencies are installed on the client-side
 rm -rf node_modules
+
+# Set the React Native version to point to the version fron the registry
+node scripts/set-rn-template-version.js "$1"
--- a/scripts/prepare-template.sh
+++ b/scripts/prepare-template.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+PACKAGE_LOCATION=$(pwd)/react-native-$1.tgz
+
+# Part 1:
+#
+# Pack React Native into a `.tgz` file so we can run from source
+npm pack
+
+# Set the React Native version to point to the `.tgz` file
+node scripts/set-rn-template-version.js "file:$PACKAGE_LOCATION"
+success "React Native version changed in the template"
+
+# Part 2:
+#
+# We need to generate CocoaPods project. To do so, we install depdendencies
+# locally and manually run `pod install`
+cd template
+
+npm install
+(cd ios && pod install)
+
+# Dependencies are installed on the client-side
+rm -rf node_modules
--- a/scripts/prepare-template.sh
+++ b/scripts/prepare-template.sh
@@ -23,4 +23,4 @@ npm install
 rm -rf node_modules
 
 # Do not ship a package.json that points to a `tgz`
-git checkout template/package.json
+git checkout package.json
--- a/scripts/bump-oss-version.js
+++ b/scripts/bump-oss-version.js
@@ -113,7 +113,10 @@ if (
   exit(1);
 }
 
-// Change react-native version in the template's package.json
+// Generate template for this version
+exec(`scripts/prepare-template.sh ${version}`);
+
+// Change React Native version in the template's package.json
 exec(`node scripts/set-rn-template-version.js ${version}`);
 
 // Verify that files changed, we just do a git diff and check how many times version is added across files
--- a/scripts/bump-oss-version.js
+++ b/scripts/bump-oss-version.js
@@ -113,9 +113,6 @@ if (
   exit(1);
 }
 
-// Generate template for this version
-exec(`scripts/prepare-template.sh ${version}`);
-
 // Change React Native version in the template's package.json
 exec(`node scripts/set-rn-template-version.js ${version}`);
 
--- a/scripts/bump-oss-version.js
+++ b/scripts/bump-oss-version.js
@@ -113,7 +113,7 @@ if (
   exit(1);
 }
 
-// Change React Native version in the template's package.json
+// Change react-native version in the template's package.json
 exec(`node scripts/set-rn-template-version.js ${version}`);
 
 // Verify that files changed, we just do a git diff and check how many times version is added across files
--- a/scripts/publish-npm.js
+++ b/scripts/publish-npm.js
@@ -129,6 +129,9 @@ artifacts.forEach(name => {
   }
 });
 
+// Generate template for this version
+exec(`scripts/prepare-template.sh ${releaseVersion}`);
+
 // if version contains -rc, tag as prerelease
 const tagFlag = releaseVersion.indexOf('-rc') === -1 ? '' : '--tag next';
 
--- a/scripts/test-manual-e2e.sh
+++ b/scripts/test-manual-e2e.sh
@@ -35,10 +35,10 @@ success "Preparing version $PACKAGE_VERSION"
 
 repo_root=$(pwd)
 
-# rm -rf android
-# ./gradlew :ReactAndroid:installArchives || error "Couldn't generate artifacts"
+rm -rf android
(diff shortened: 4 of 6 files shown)


grabbou:
Big thanks to @orta for pointing me towards a solution - I would loose my mind if it wasn't for you <3

grabbou:
Note: Let's merge this PR once 0.60 is out. There might be some fixes needed in the meantime.

grabbou:
Going to close it and will re-open once 0.60 ships.

scripts/prepare-template.sh
```diff
@@ -0,0 +1,26 @@
...
+npm pack
+
+# Set the React Native version to point to the `.tgz` file
+node scripts/set-rn-template-version.js "file:$PACKAGE_LOCATION"
+
+# We need to generate CocoaPods project. To do so, we install depdendencies
+# locally and manually run `pod install`
+cd template
+
+npm install
```
thymikee: `npm install` creates `package-lock.json` (it's in gitignore so you don't see it). Can we use Yarn instead (it's faster) and checkout `yarn.lock` along with `package.json`?

grabbou: I didn't go with `yarn` for two reasons:
- it has issues with cache when installing from a tarball - it will quickly eat up available space unless you `yarn cache clean` yourself
- we use `npm install` and `npm pack` already in other `scripts/*`, so figured I'll just stick to the pattern

grabbou:
Now that https://github.com/react-native-community/cli/pull/362 is going to be merged, I'll update the PR to always init from a tarball.

That way, we will be able to make sure that we test contents of a real package, not the source. This is important, because we have an explicit list of `files` to publish - we recently added `Pods` and I want to make sure we actually ship it.

thymikee:
FYI putting `Pods` into the tarball increases its size by ~12MB (from ~3MB to ~15MB). Not cool, as it affects everybody, not only those who `init` new project. We'll need to move the template to a separate package in the monorepo.

scripts/prepare-template.sh
```diff
@@ -0,0 +1,26 @@
...
+(cd ios && pod install)
```
thymikee: I think we'll want to pass `--repo-update` flag so we publish up-to-date pods

grabbou:
Unfortunately, `Pods` cannot be `npm publish`'ed because Pods are using `symlinks` inside `Pods/Headers` and this will not be included in the package published.

There's no way right now other than requiring to run `pod install` after `init` that I am aware of. 

Our `tooling` doesn't support this at all (e.g. we don't prompt to run `pod install`) - we would need time to implement that properly. I feel like doing it right now, just in time for the 0.60 release, would be asking for potential troubles and degraded first-time experience. 

I think it's important to make sure that initing a new app is as smooth as possible and we should pay an extra time to a) think about the best scenario, b) how to implement it and c) test it.

The next steps are:
- hold with 0.60-release for another two weeks and do proper CocoaPods support in tooling for `init` experience (e.g. `run-ios` warn about missing `Pods`, suggest installing, prepare section in `docs` where we can explain this)
- revert CocoaPods PR from 0.60-release
- make default template a non-CocoaPods and add another one `react-native-cocoapods` to be enabled behind a flag (those who have CocoaPods can use it instead).

mikehardy:
As a very recent new user that commented on the initial RFC related to "podless install" I can say that Pods is absolutely, without a doubt the way to handle iOS things. The problem is just unfamiliarity and expectations (for me). If the tools even spit out a big message that said "Before running react-native run-ios you must run pod install in the ios directory http://docsurl/react-native-and-pods.html" and on that page there was a very quick intro to pods (either brew or gems to install, big first time download, always open .xcworkspace) This would be fine. In other words, two quick bits of documentation to start getting devs up to speed. They are not *that* complicated, it's just new knowledge - but useful enough that the community would/should rally around it as a big new feature I think

orta:
It might be possible to duplicate the files for the headers instead of having the symlinks?

grabbou:
It is possible. Question is whether this will affect CocoaPods at all? (I
assume it must have used symlinks for a reason)

orta:
My gut says no, CP (quite reasonably) probably didn't want to have to make duplicate files in the fs. I just tested it locally in one of my working CP projects, and replaced the links with the originals and it compiled file. Could be enough?

grabbou:
Ok. I will give it another go and hopefully that’s it!

grabbou:
Looks like this:
```
# Convert all symlinks inside Pods to real files
for f in $(find Pods -type l);
  do cp --remove-destination $(readlink $f) $f;
done;
```
will do! 

Updated and testing in `0.60-stable` branch. If it works, we are at home! I guess `pod install` down the road will convert them to symlinks anyway.

scripts/prepare-template.sh
grabbou: Hmm... mind elaborating a bit what that flag is doing? We have local pods (from local spec within the repo, so no need to update the whole CocoaPods repo) and we also rm -rf Pods anyway.

thymikee: forgot it's local, you're right

grabbou:
Unfortunately, converting symlinks to real files doesn't work as it produces "duplicated declaration" linker errors.

The only way to solve this would be to actually "restore symlinks" in `init` itself, but we would need to back in support to the CLI for that.

CC: @thymikee

matt-oakes:
I get the reasoning behind wanting to avoid users needing to manually run `pod install` after an `init`, however, I don't think trying to "fake it" is the way to go.

As the template uses Cocoapods now, users are going to have to use it at some point when they add any library with external dependencies. At that point they are going to need Cocoapods to be installed. Isn't it best to get them to install it along with the other required dependencies like Watchman and Node? All that is required is a call to `sudo gem install cocoapods` and they will have everything they need.

At that point, if we know (and can check) that they have Cocoapods installed, and we can just run `pod install` for them in the `init` command. It also means they are all setup and ready to go for when they add new dependencies.

scripts/prepare-template.sh
```diff
@@ -0,0 +1,26 @@
...
+
+PACKAGE_LOCATION=$(pwd)/react-native-$1.tgz
+
+# Pack React Native into a `.tgz` file so we can run from source
+npm pack
+
+# Set the React Native version to point to the `.tgz` file
+node scripts/set-rn-template-version.js "file:$PACKAGE_LOCATION"
+
+# We need to generate CocoaPods project. To do so, we install depdendencies
```
hramos:
```suggestion
# We need to generate CocoaPods project. To do so, we install dependencies
```

grabbou:
I agree with you @matt-oakes. Unfortunately, there are too many hacks that need to be done in order to support this reliably. 

I am going to consult it with the CLI team where and how we can add support for automatically running `pod install` and we will probably just notify that it's required to install if not present in order to run a React Native app.

benjarwar:
@grabbou we've found integrating CocoaPods to be very challenging, to the point that we've historically opted to stick with Carthage for iOS dependency management. We're going to try switching to CocoaPods again, but I'm nervous about the potential disruption to our development process. Sorry to cross-post, but [check my comment](https://github.com/facebook/react-native/pull/23563#issuecomment-489111207) in #23563.

thymikee:
@benjarwar this is for new projects only. For users who would like to always start with Carthage, there's an option to create and release to npm a custom template (since RN 0.60.0 or `@react-native-community/cli@next` currently). It's very straight-forward to create one, see the docs: https://github.com/react-native-community/cli/blob/master/docs/init.md#creating-custom-template

Once you create it, one can do:
```sh
npx @react-native-community/cli@next init ProjectName --template sample-carthage-template
```
or when RN 0.60.0 RC is up:
```sh
npx react-native@^0.60.0-rc.0 init ProjectName --template sample-carthage-template
```

benjarwar:
@thymikee thanks for the clarification. But there's mention in multiple comments in #23563 that `react-native link` is also going to be CocoaPods dependent.

thymikee:
We're moving away from `link` in favor of autolinking. See: https://github.com/react-native-community/cli/commit/76f079ee8a2ebf9843a47422bdbcb86651b2b086. I'm pretty sure something similar could be done in Carthage as well in userland :)

matt-oakes:
@grabbou What are the hacks which are needed and what is the issue with running it directly? It is because you're not sure if the user has Cocoapods installed through Bundler or a similar tool?

grabbou:
The hacks were needed because after running `pod install` pre-publish and
copying that to user directory after init has different paths which is
broken. Once you workaround this, you need to figure out symlinks -
CocoaPods uses symlinks but npm publish will always skip them. Converting
to real files works, but gives a duplicated file error.

Generally speaking, it's not worth it and we will just stick to running
`pod install` or installing CocoaPods on user side while initing.

matt-oakes:
@grabbou I think we're agreeing. I was suggesting that `pod install` is run on the user side while initing. The hacks you mention seem to be for running install before publishing the template, unless I'm getting confused.

grabbou:
Yes, exactly :) I agree with you and that's the approach we are taking
right now :)

PR #24641, opened 2019-04-28
Fix accessibility event properties for TextInput

elucaswork:
## Summary

When a `TextInput` receives any accessibility event prop (`onAccessibilityTap`, `onMagicTap`, `onAccessibilityEscape`, `onAccessibilityEscape`), causes a crash.

<p align=center></p>

## Changelog

[iOS] [Fixed] - Fix accessibility event properties for `TextInput`

## Test Plan

The following component shouldn't crash the app.
```js
<TextInput
  accessible={true}
  onMagicTap={() => {
    console.log("--onMagicTap works");
  }}
  onAccessibilityTap={() => {
    console.log("--onAccessibilityTap works");
  }}
  onAccessibilityEscape={() => {
    console.log("--onAccessibilityEscape works");
  }}
/>
```

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -283,16 +283,16 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
 - (void)setSecureTextEntry:(BOOL)secureTextEntry {
   UIView<RCTBackedTextInputViewProtocol> *textInputView = self.backedTextInputView;
-    
+
   if (textInputView.secureTextEntry != secureTextEntry) {
     textInputView.secureTextEntry = secureTextEntry;
-      
+
     // Fix #5859, see https://stackoverflow.com/questions/14220187/uitextfield-has-trailing-whitespace-after-securetextentry-toggle/22537788#22537788
     NSAttributedString *originalText = [textInputView.attributedText copy];
     self.backedTextInputView.attributedText = [NSAttributedString new];
     self.backedTextInputView.attributedText = originalText;
   }
-    
+
 }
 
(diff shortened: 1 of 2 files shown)


Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -505,13 +505,6 @@ - (CGSize)sizeThatFits:(CGSize)size
   return fittingSize;
 }
 
-#pragma mark - Accessibility
```
shergin: I am afraid this particular change is simply wrong/unreasonable.

shergin: Just to be clear: Removing this turns off all accessibility props of <TextInput> (by redirecting them to the wrapper view).

elucaswork: @shergin my idea of deleting this `reactAccessibilityElement` was to make `RCTBaseTextInputView` use [`reactAccessibilityElement` inherited from `RCTView`](https://github.com/facebook/react-native/blob/master/React/Views/RCTView.m#L316-L319) (the wrapper).

When I opened this PR I tested `TextField` passing those events set from [RCTViewManager](https://github.com/facebook/react-native/blob/master/React/Views/RCTViewManager.m#L125-L136), and it was working. But I might have missed something :-/.

Now I'll need to make more tests, especially because I this commit needed to be reverted (probably because of this detox failure https://circleci.com/gh/facebook/react-native/88586, right @grabbou?). I just don't know yet why this commit caused a regression into `Switch` component.

grabbou: Yes, I think Detox couldn't find a TextInput on the top of the app (by its accessibility label), to type `<switch />` and select that example.

So definitely related.

PR #24642, opened 2019-04-29
Add support for screen reader blur and focus (accessibility)

elucaswork:
## Summary

Add support for screen reader blur and focus events via [`UIAccessibilityFocus`](https://developer.apple.com/documentation/uikit/accessibility/uiaccessibilityfocus) for iOS and [`View.AccessibilityDelegate` with `AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED` and `TYPE_VIEW_ACCESSIBILITY_FOCUS_CLEARED`](https://developer.android.com/reference/android/view/accessibility/AccessibilityEvent) for Android.

When VoiceOver (iOS) or TalkBack (Android) enabled, `onAccessibilityFocus`  will be fired once the element gains the screen reader focus, and `onAccessibilityBlur` will be fired when it loses the screen reader focus.

## Changelog

[General] [Added] - Accessibility - Add support for screen reader blur and focus events

## Test Plan

```js 
  render() {
    const onFocus = e => {
      console.log("--focus", e.target);
    };
    const onBlur = e => {
      console.log("--blur", e.target);
    };
    return (
      <View onAccessibilityFocus={onFocus} onAccessibilityBlur={onBlur}>
        <TouchableOpacity
          onAccessibilityFocus={onFocus}
          onAccessibilityBlur={onBlur}
        >
          <Text>TouchableOpacity element</Text>
        </TouchableOpacity>
        <Text onAccessibilityFocus={onFocus} onAccessibilityBlur={onBlur}>
          Test element
        </Text>
        <TextInput
          accessible={true}
          onAccessibilityFocus={onFocus} onAccessibilityBlur={onBlur}
        />
      </View>
    );
  }
}
```

P.S.: To make it work for `TextInput` on iOS, this fix #24641 is necessary.


ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
```diff
@@ -108,17 +123,34 @@ public static void setDelegate(final View view) {
...
+        @Override
+        public void onInitializeAccessibilityEvent(View host, AccessibilityEvent accessibilityEvent) {
+          super.onInitializeAccessibilityEvent(host, accessibilityEvent);
+          ReactContext reactContext = (ReactContext)host.getContext();
+
+          WritableMap event = Arguments.createMap();
+          event.putInt("target", host.getId());
+
+          if (accessibilityEvent.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUS_CLEARED) {
+            reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(host.getId(), BLUR_EVENT_NAME, event);
```
blavalla: I'm not sure how React Native handles the call to receiveEvent here, but it's worth pointing out that not all events that get passed through `onInitializeAccessibilityEvent` will actually be fired.  Every ancestor of the view has the chance to cancel the event before it bubbles to the top.  Only if it makes it all the way to the top of the view hierarchy will it be passed off to a running accessibility service.

The AccessibilityEvent cycle goes like this:

1.) `View.sendAccessibilityEvent()` (intent to send event started)
2.) `View.onInitializeAccessibilityEvent()` (event populated with view info)
3.) `View.dispatchPopulateAccessibilityEvent()` (event text populated with additional info)
4.) `ViewParent.requestSendAccessibilityEvent()` (parent of View is free to modify or cancel the event, or pass it up to its parent)
5.) ViewRootImpl calls `sendAccessibilityEvent` on the `AccessibilityManager` (this is what actually passes the event to any running accessibility service).

The best way to actually handle this would probably be to override `performAccessibilityAction` instead.  When Talkback (or any other accessibility service) receives a focus event, it's not guaranteed to actually change focus.  For example, if the already focused view sent another `TYPE_VIEW_ACCESSIBILITY_FOCUSED` event, it would be ignored.  If it *does* decide to change focus, it will fire the `performAccessibilityAction` method on the View and pass in an `AccessibilityNodeInfo.ACTION_ACCESSIBILITY_FOCUS` action.

Here is the code in Talkback doing that for reference:
https://github.com/google/talkback/blob/e69d4731fce02bb9e69613d0e48c29033cad4a98/talkback/src/main/java/eventprocessor/ProcessorFocusAndSingleTap.java#L871

elucaswork: @bravalla Thanks for this detailed feedback.

I applied your suggestions, which works fine for focus event, but for the blur event it doesn't work as expected. Using `performAccessibilityAction`, blur is triggered just when the focus moves to a non react-native element (**my assumption**). When focus is moving around react-native elements, nothing happens.

In the video you can understand better what I say: blur just trigger when focus moves to status bar.


code: https://gist.github.com/elucaswork/242c098a78577ddcdef4d75a60d2a4a5

Do you have an idea of what can be happening?
Thank you.

Libraries/Components/TextInput/TextInput.js
```diff
@@ -1325,6 +1331,14 @@ const TextInput = createReactClass({
     }
   },
 
+  _onAccessibilityBlur: function(event: BlurEvent) {
+    this.props.oncAccessibilityBlur && this.props.oncAccessibilityBlur(event);
```
blavalla: typo oncAccessibilityBlur -> onAccessibilityBlur

Libraries/Components/TextInput/TextInput.js
```diff
@@ -1325,6 +1331,14 @@ const TextInput = createReactClass({
...
+  },
+
+  _onAccessibilityFocus: function(event: BlurEvent) {
+    this.props.oncAccessibilityFocus && this.props.oncAccessibilityFocus(event);
```
blavalla: typo oncAccessibilityFocus -> onAccessibilityFocus

Libraries/Components/TextInput/TextInput.js
```diff
@@ -1325,6 +1331,14 @@ const TextInput = createReactClass({
     }
   },
 
+  _onAccessibilityBlur: function(event: BlurEvent) {
```
blavalla: Shoudln't this be AccessibilityBlurEvent not BlurEvent?

Libraries/Components/TextInput/TextInput.js
```diff
@@ -1325,6 +1331,14 @@ const TextInput = createReactClass({
...
+    this.props.oncAccessibilityBlur && this.props.oncAccessibilityBlur(event);
+  },
+
+  _onAccessibilityFocus: function(event: BlurEvent) {
```
blavalla: Shoudln't this be AccessibilityFocusEvent not BlurEvent?

ReactAndroid/src/main/java/com/facebook/react/uimanager/AccessibilityDelegateUtil.java
blavalla: @elucaswork, The code here looks good to me.  I've verified the approach in a non-ReactNative app to make sure that Talkback was correctly calling performAccessibilityAction on both focus and blur, and it is.

It's possible that the issue is at the ReactNative side (I commented above on a few typos), but I am definitely not an expert on that layer so I'll defer to others if thats where the issue lies.

You could verify that the Android side is working correctly by adding some logging into the performAccessibilityAction method itself to see if the expected actions are ever being passed in.

elucaswork:
@blavalla good point. It makes sense to wait for @marcmulcahy's #24695 and then add blur and focus events on top of it. 

I'm going to close it for now and open a new one as soon as "extended accessibility actions" gets merged.

Thanks for your review!

chyamuna:
@marcmulcahy's [24695 ](https://github.com/facebook/react-native/pull/24695) PR has been merged. Is the functionality to add support for voice over focus and blur has bee added. or any active PR which is adding this functionality? Thank You [estevaolucas](https://github.com/estevaolucas)

blavalla:
@chyamuna , this functionality hasn't been added, but if you'd like to take up the mantle here and make a PR I'd be happy to review.

I think the API for this is pretty clear, since 24695 added an accessibilityActions prop with a few built-in actions like "activate" and "longpress". I think "focus" and "blur" could be added as additional built-in actions that work for both iOS and Android.

See the documentation here if you aren't familiar with this API - https://reactnative.dev/docs/accessibility#accessibility-actions

PR #24649, opened 2019-04-29
[iOS] Timing: Fixes timer when app get into background

zhongwuzw:
## Summary

Related #23674, in that PR, we imported background timer support, but it's not sufficient, I think the reason that works is because it enable the `Background Modes` and do some background tasks, for the users who don't enable it, timer would pause immediately before goes into background.

To fix it, we can mark a background task when goes into background, it can keep app active for minutes, try best to support timing when in background.

cc. @cpojer .

## Changelog

[iOS] [Fixed] - Timing: Fixes timer when app get into background

## Test Plan

Run code in background like the code below, console can log correctly.
```
    setInterval(() => {
      console.log("Timer run");
    }, 1000);
```

### DIFF
--- a/React/Modules/RCTTiming.m
+++ b/React/Modules/RCTTiming.m
@@ -185,10 +185,7 @@ RCT_EXPORT_MODULE()
 
 - (void)appDidMoveToForeground
 {
-  if (self->_backgroundTaskIdentifier != UIBackgroundTaskInvalid) {
-    [[UIApplication sharedApplication] endBackgroundTask:self->_backgroundTaskIdentifier];
-    _backgroundTaskIdentifier = UIBackgroundTaskInvalid;
-  }
+  [self markEndOfBackgroundTaskIfNeeded];
   _inBackground = NO;
   [self startTimers];
 }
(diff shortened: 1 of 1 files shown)


jacobp100:
Just a note that react-intl will set huge timeouts for its `<FormattedRelative>` components, so this could cause huge issues for some people's code

zhongwuzw:
@jacobp100 Eh, what do you mean? can you expand which issues you want to express? iOS can't run for days in background.

jacobp100:
I just wanted to highlight a popular library people use on React Native calls setTimeout with a delay that can last for days. If we wait for every set timeout by running the app in the background, it could, in some cases, be a drain on the battery.

Not saying it shouldn't be merged, I just wanted to highlight one potential issue!

zhongwuzw:
@jacobp100 I don't know wether I get which issue you want to express. You mean it may has battery issue? Actually, we already support background timer in #23674(Many people have expressed a desire for background timer support on iOS ), and background task can only last most a few minutes, so we only can support background timer for a short time, if app not return to foreground, it will be killed by system.
And if you have some other issues, please comment, and we'll try to fix them. Thanks.

jacobp100:
Ignore me then! 😄

zhongwuzw:
Fixed #25083.

craftzdog:
Cheers!

bharatidudhjiya:
I have updated RCTTiming.m file with lasted code but still js code not work when app in background.

thanks!

PR #24653, opened 2019-04-29
call onValueChange only when value changes

a-c-sreedhar-reddy:
## Summary
`OnValueChange` function of `Picker` is called when Picker is initialized.

## Changelog
[Android][fixed] - `OnValueChange` will be called only when the `selectedValue` changes.

## Test Plan
```
  constructor(props) {
    super(props);
    this.state = {language:"js",test:1}
  }
  render() {
    return (
        <Picker
          selectedValue={this.state.language}
          style={{height: 50, width: 100}}
          onValueChange={(itemValue, itemIndex) =>{
            this.setState({language: itemValue,test:2});
          }
        }>
          <Picker.Item label="Java" value="java" />
          <Picker.Item label="JavaScript" value="js" />
        </Picker>
        <Text>{this.state.test}</Text>
      </View>
    );
  }
}
```
Here `state.test` must be 1 when the app loads. But the `onValueChange` is called initially which changes `state.test` to 2. This pull request fixes the issue.

### DIFF
--- a/Libraries/Components/Picker/PickerAndroid.android.js
+++ b/Libraries/Components/Picker/PickerAndroid.android.js
@@ -119,7 +119,9 @@ class PickerAndroid extends React.Component<
         const value = children[position].props.value;
         /* $FlowFixMe(>=0.78.0 site=react_native_android_fb) This issue was
          * found when making Flow check .android.js files. */
-        this.props.onValueChange(value, position);
+        if (this.props.selectedValue !== value){
+          this.props.onValueChange(value, position);
+        }
       } else {
         this.props.onValueChange(null, position);
       }
--- a/Libraries/Components/Picker/PickerAndroid.android.js
+++ b/Libraries/Components/Picker/PickerAndroid.android.js
@@ -119,7 +119,7 @@ class PickerAndroid extends React.Component<
         const value = children[position].props.value;
         /* $FlowFixMe(>=0.78.0 site=react_native_android_fb) This issue was
          * found when making Flow check .android.js files. */
-        if (this.props.selectedValue !== value){
+        if (this.props.selectedValue !== value) {
           this.props.onValueChange(value, position);
         }
       } else {


cpojer:
cc @Kudo since you recently fixed an issue with `onValueChange`.

Kudo:
Thanks @cpojer.

Hey @a-c-sreedhar-reddy, just to clarify which RN version you used.
[My change](https://github.com/facebook/react-native/commit/6a3d9c06ce30433e4b1ded8f59f83697185282e9) landed after RN 0.59, so that if you tested on RN 0.58 or before, there might be another story.
I've tested your code, but seems works fine on RN 0.59.
Could you please double check your RN version and if RN 0.59 works for you?

a-c-sreedhar-reddy:
> Thanks @cpojer.
> 
> Hey @a-c-sreedhar-reddy, just to clarify which RN version you used.
> [My change](https://github.com/facebook/react-native/commit/6a3d9c06ce30433e4b1ded8f59f83697185282e9) landed after RN 0.59, so that if you tested on RN 0.58 or before, there might be another story.
> I've tested your code, but seems works fine on RN 0.59.
> Could you please double check your RN version and if RN 0.59 works for you?
```
    "react": "16.8.3",
    "react-native": "0.59.5"

```
This is what I tested with.

Kudo:
That's weird to me since I tested with the exact same version.
Supposedly this issue should not happen as it is handled by native side.
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPicker.java#L98-L103
```java
  @Override
  protected void onLayout(boolean changed, int left, int top, int right, int bottom) {
    super.onLayout(changed, left, top, right, bottom);

    // onItemSelected gets fired immediately after layout because checkSelectionChanged() in
    // AdapterView updates the selection position from the default INVALID_POSITION.
    // To match iOS behavior, which no onItemSelected during initial layout.
    // We setup the listener after layout.
    if (getOnItemSelectedListener() == null)
      setOnItemSelectedListener(mItemSelectedListener);
  }
```

a-c-sreedhar-reddy:
@Kudo  yes you are correct. I don't know why but its perfect now. May be I did not update node modules. Sorry for this.  To remove this commit should I again send a PR?

a-c-sreedhar-reddy:
@Kudo  You are right. May be I did not update node modules. Sorry for this. Should I make a PR removing this change?

Kudo:
@a-c-sreedhar-reddy Sorry missed this. Your fix is fine for me.
In fact, from the semantic of `onValueChange`, it is no hurt to double check for value inequality.
Just to clarify if I was doing wrong in Java native side.

PR #24659, opened 2019-04-30
Ignore unexpected style prop value on Android

elucaswork:
## Summary

Android crash the app when a style prop receives an unexpected value.

E.g.: the possible values for `width` or`height` is `"auto"`, `"100%"` or `100`. Passing a `"100"` (neight number nor percent string) throw an error - which crashes the app in release mode.

The propose in this PR is just to ignore any unexpected value. Which will be the same as not set any value for the style prop.

## Changelog

[Android] [Fix] - Ignore wrong style value on Android

## Test Plan

The following code shouldn't crash
```js
<View style={{ backgroundColor: "red", width: "50", height: "ignoreit" }} />
```

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/LayoutShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/LayoutShadowNode.java
@@ -60,7 +60,8 @@ public class LayoutShadowNode extends ReactShadowNodeImpl {
           unit = YogaUnit.PERCENT;
           value = Float.parseFloat(s.substring(0, s.length() - 1));
         } else {
-          throw new IllegalArgumentException("Unknown value: " + s);
+          unit = YogaUnit.UNDEFINED;
+          value = YogaConstants.UNDEFINED;
         }
       } else {
         unit = YogaUnit.POINT;


cpojer:
Thanks for this PR. I think this makes sense for release versions of the app for sure but while developing this can be very confusing. If we encounter a wrong valid for height or width, could we use a YellowBox warning that pretty much says what you said in your PR summary? I think that would be a good fix!

elucaswork:
@cpojer I was forgetting of the YellowBox. It’s a good suggestion. 
I don’t know it how to trigger a warning from java side, but I’ll take a look. 

Important info: this change affects not only width and height props, but any style props with unexpected value

cpojer:
I don't think we have a mechanism to do this on the native side. Could we do the same validation in JS and invoke the YellowBox warning there?

elucaswork:
@cpojer right now there's no such validation for style prop values. With the current RN architecture, those validation happens just on the native side. 

I don't have too much knowledge of the codebase but looks like It's possible to add something into [`ReactNativeRenderer-dev`](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-dev.js) (doing something similar to [warnForStyleProps](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-dev.js#L3712-L3726)). 

I might be wrong, but with the re-architecture, Codegen will resolve this kind problem when an unexpected value is used. Thus, this code would be just a temp solution.

If you think it worth add such validation, I try to do it. Just let me know.

cpojer:
I think that's a good idea! Note that this code lives in the React repo itself so you have to merge a change there first and then we'll sync it over here. If you could send a PR to React with the warnings in dev mode, then after that we can land this PR.

elucaswork:
@cpojer Ok, I'm going to work on it.

cpojer:
Let's close this PR here as the final solution will likely look somewhat different.

alishamae:
Unsubscribe all

PR #24662, opened 2019-04-30
[general][changed] cleanup RedBox message and stack output

thymikee:
## Summary

Cleanup RedBox messages and stack traces. This PR consists of 2 changes (I'm good with splitting them up if you'd like):

- [general] filter out some of the internal callsites from the symbolicated stack (I thought about using monospace font for title with code frame, but it looks weird)
- [ios][android] strip ANSI characters (coming from colored Babel code frame) from the error message

I think it's ok to strip it inside native handlers so we can still have a colorful code frame in the terminal output.

**JS Code frame:**

|before|after|
|--|--|
|Screenshot 2019-04-30 at 12 32 05|Screenshot 2019-04-30 at 12 52 43|

|before|after|
|--|--|
|image|image|

**Filtered stack traces:**

|before|after|
|--|--|
|Screenshot 2019-04-30 at 12 27 21Screenshot 2019-04-30 at 12 27 28|Screenshot 2019-04-30 at 12 26 55|

There's still a lot of places that are hard to read, but I think this is a good start towards more readable errors.

cc @cpojer 

## Changelog

[General][Changed] - Cleanup RedBox message and stack output

## Test Plan

Played around in RNTester

### DIFF
--- a/Libraries/Core/ExceptionsManager.js
+++ b/Libraries/Core/ExceptionsManager.js
@@ -12,6 +12,8 @@
 
 import type {ExtendedError} from 'parseErrorStack';
 
+const INTERNAL_CALLSITES_REGEX = /ReactNativeRenderer-dev\.js|MessageQueue\.js/;
+
 /**
  * Handles the developer-visible aspect of errors and exceptions
  */
@@ -38,9 +40,12 @@ function reportException(e: ExtendedError, isFatal: boolean) {
       symbolicateStackTrace(stack)
         .then(prettyStack => {
           if (prettyStack) {
+            const stackWithoutInternalCallsites = prettyStack.filter(
+              frame => frame.file.match(INTERNAL_CALLSITES_REGEX) === null,
+            );
             ExceptionsManager.updateExceptionMessage(
               e.message,
-              prettyStack,
+              stackWithoutInternalCallsites,
               currentExceptionID,
             );
           } else {
--- a/Libraries/Core/ExceptionsManager.js
+++ b/Libraries/Core/ExceptionsManager.js
@@ -26,7 +26,6 @@ function reportException(e: ExtendedError, isFatal: boolean) {
     const currentExceptionID = ++exceptionID;
     const message =
       e.jsEngine == null ? e.message : `${e.message}, js engine: ${e.jsEngine}`;
-    console.log('reportException', message);
     if (isFatal) {
       ExceptionsManager.reportFatalException(
         message,
@@ -73,7 +72,6 @@ function handleException(e: Error, isFatal: boolean) {
   // Unfortunately there is no way to figure out the stacktrace in this
   // case, so if you ended up here trying to trace an error, look for
   // `throw '<error message>'` somewhere in your codebase.
-  console.warn('abc', e.message);
   if (!e.message) {
     e = new Error(e);
   }
--- a/Libraries/Core/ExceptionsManager.js
+++ b/Libraries/Core/ExceptionsManager.js
@@ -26,6 +26,7 @@ function reportException(e: ExtendedError, isFatal: boolean) {
     const currentExceptionID = ++exceptionID;
     const message =
       e.jsEngine == null ? e.message : `${e.message}, js engine: ${e.jsEngine}`;
+    console.log('reportException', message);
     if (isFatal) {
       ExceptionsManager.reportFatalException(
         message,
@@ -72,6 +73,7 @@ function handleException(e: Error, isFatal: boolean) {
   // Unfortunately there is no way to figure out the stacktrace in this
   // case, so if you ended up here trying to trace an error, look for
   // `throw '<error message>'` somewhere in your codebase.
+  console.warn('abc', e.message);
   if (!e.message) {
     e = new Error(e);
   }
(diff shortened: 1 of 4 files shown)


React/Modules/RCTRedBox.m
```diff
@@ -176,14 +176,24 @@ - (void)dealloc
...
+{
+    NSError *error = nil;
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:@"\\x1b\\[[0-9;]*m" options:NSRegularExpressionCaseInsensitive error:&error];
+    return [regex stringByReplacingMatchesInString:text options:0 range:NSMakeRange(0, [text length]) withTemplate:@""];
+}
+
 - (void)showErrorMessage:(NSString *)message withStack:(NSArray<RCTJSStackFrame *> *)stack isUpdate:(BOOL)isUpdate
 {
+    // Remove ANSI color codes from the message
+    NSString * messageWithoutAnsi = [self stripAnsi:message];
```
cpojer: *messageWithoutAnsi

cpojer:
Thank you so much for doing this, this is a great start and makes stack traces so much cleaner. My only concern is about debugging React/React Native changes because when errors do happen, having them stripped out of those files may give a wrong impression to people. Do you think we could add a button at the bottom of the trace, using the same styling as a stack trace to be unobtrusive, that will show hidden parts of the stack when tapped?

thymikee:
> Do you think we could add a button at the bottom of the trace, using the same styling as a stack trace to be unobtrusive, that will show hidden parts of the stack when tapped?

Definitely. Thought about it, but since it involves writing stateful UI logic on both platforms, I decided to go with this as a PoC for now. Happy to give it a spin next week (I'm on vacation starting today.)

cpojer:
Enjoy your vacation! I think we can ship this for now as it is a good improvement and I'm gonna give others on the RN team a heads up in case they are doing a React sync soon so they are aware that a redbox is excluding some things.

cc @TheSavior @fkgozali @ejanzer

Libraries/Core/ExceptionsManager.js
```diff
@@ -12,6 +12,8 @@
 
 import type {ExtendedError} from 'parseErrorStack';
 
+const INTERNAL_CALLSITES_REGEX = /ReactNativeRenderer-dev\.js|MessageQueue\.js/;
```
cpojer: Actually should we make this an exact match with `^` and `$`?

Libraries/Core/ExceptionsManager.js
```diff
@@ -24,6 +26,7 @@ function reportException(e: ExtendedError, isFatal: boolean) {
     const currentExceptionID = ++exceptionID;
     const message =
       e.jsEngine == null ? e.message : `${e.message}, js engine: ${e.jsEngine}`;
+    console.log('reportException', message);
```
cpojer: oops please remove

Libraries/Core/ExceptionsManager.js
```diff
@@ -67,6 +73,7 @@ function handleException(e: Error, isFatal: boolean) {
   // Unfortunately there is no way to figure out the stacktrace in this
   // case, so if you ended up here trying to trace an error, look for
   // `throw '<error message>'` somewhere in your codebase.
+  console.warn('abc', e.message);
```
cpojer: another one here

Libraries/Core/ExceptionsManager.js
thymikee: For now I've added the `$` but I'm going to compare how the paths look in normal app vs RNTester, because they differ.

thymikee: Adjusted the paths: eb97ee3b2a27d0de3ddb8fe4f4988ce0172f8e85. They're now generated from an array, so it should be easier to add new ones (or generalize more, we'll see about that).

julioxavierr:
Awesome. Thanks

pvinis:
Could we do one more thing, and put the text in the red part monospace? At least the part with the line numbers like 30, 31, etc in the first example.

thymikee:
@pvinis to quote myself: "I thought about using monospace font for title with code frame, but it looks weird". I can play with that later though

pvinis:
Oops, I forgot about this parenthesis while looking at the rest :p 
If you tried and it's weird, ok, but it feels kinda silly to have all the `> | ^` to show things that should be aligned while using a non-monospace font.

TheSavior:
This will make things so much better, thanks!

As @cpojer mentioned, it will be pretty important to be able to see the hidden frames. Whenever we do react syncs, those frames are how we debit issues in the renderer. Hopefully you can do a quick follow up to add that.

pvinis:
Just for comparison:

| before | after |
| - | - |
| Screenshot 2019-04-30 at 12 52 43 | Screenshot 2a019-04-30 at 12 52 43 |

I think the right one looks way easier to understand what's up. If it was horizontally scrollable it could be even better, but already it feels nicer.

What do you think?

cpojer:
@pvinis I think it works for the code frame but not so well for the rest.

pvinis:
true. I'll try with only the code frame. it was a quick try. maybe I make a PR for it and we see there. thanks :)

aleclarson:
@pvinis If you could use a regular expression to detect the character range of the code frame (excluding the preceding error message), that would be great!

PR #24678, opened 2019-05-01
Update Android view state when accessibility states are changed.

marcmulcahy:
## Summary

#24095 introduced a regression in an internal test case. Without access to the test case code, my best guess is that the test instantiates a button, and sets its "disabled" state to true. In the original code, when the button is rendered, if "disabled" is true, "disabled" is added to the accessibility states of the button, which causes the Android view's enabled state to be set to false.

In #24095, we removed the code that changes the underlying Android view's enabled state to false when "disabled" is included in the accessibility states, which seems correct. The Android view's state shouldn't mirror the accessibility state, it should be the other way round-- the accessibility state should represent the state of the view.

It seems that the existing test is brokenly setting the disabled state in the Javascript object, and then querying the Android view's enabled state to confirm the change. If the Button implementation is actually the culprit, then IMHO, the correct fix would be to have the Button implementation manipulate the Android View's enabled state, not the accessibilityStates code.

## Changelog
[android] [fix] - Fix internal test case around disabled state of buttons

## Test Plan

Verified that existing RNTester cases still work as expected.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -146,11 +146,22 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
       return;
     }
     view.setTag(R.id.accessibility_states, accessibilityStates);
+    view.setSelected(false);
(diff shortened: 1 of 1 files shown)


cpojer:
I can confirm that this fixed the test internally. Thank you so much!

PR #24679, opened 2019-05-01
Fix crash when `reactAccessibilityElement` is not a `RCTView`

janicduplessis:
## Summary

React Native Gesture Handler uses a `RCTViewManager` subclass to manage `UIControl` so the cast there is not safe. Not sure if there is a better solution but this prevents the crash.

image

## Changelog

[General] [Fixed] - Fix crash when `reactAccessibilityElement` is not a `RCTView`

## Test Plan

Tested that it fixes the crash


janicduplessis:
cc @marcmulcahy @kmagiera

janicduplessis:
@shergin I think it makes sense

janicduplessis:
Can't force push to this, reopened as #24743

PR #24693, opened 2019-05-02
Fix event ordering when combining coalescable and non-coalescable events

janicduplessis:
## Summary

Fixes an issue introduced in https://github.com/facebook/react-native/pull/15894 that can cause events to be dispatched out of order.

Also reverted `viewTag` moved to optional property as it didn't actually work and makes code more complex.

## Changelog

[iOS] [Fixed] - Fix event ordering when combining coalescable and non-coalescable events

## Test Plan

Tested events still work, need to validate it fixes the crash seen @fb

### DIFF
--- a/React/Base/RCTEventDispatcher.h
+++ b/React/Base/RCTEventDispatcher.h
@@ -35,6 +35,7 @@ RCT_EXTERN NSString *RCTNormalizeInputEventName(NSString *eventName);
 @protocol RCTEvent <NSObject>
 @required
 
+@property (nonatomic, strong, readonly) NSNumber *viewTag;
 @property (nonatomic, copy, readonly) NSString *eventName;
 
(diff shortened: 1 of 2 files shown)


janicduplessis:
@shergin @sahrens Can you test if this fixes the issue you reported?

shergin:
AFAIK, this is hard to repro issue which happens only in prod at scale (one of our top crashes thought!).

Looks good and reasonable to me! Thank you so much Janic! ❤️

So, let's land and see what happened. 🤞

shergin:
AFAIK, this is hard to repro issue which happens only in prod at scale (one of our top crashes thought!).

Looks good and reasonable to me! Thank you so much Janic! ❤️

So, let's land and see what happened. 🤞

PR #24695, opened 2019-05-02
Extended Accessibility Actions Support

marcmulcahy:
## Summary

This is a reconstitution of #24190. It extends accessibility actions to include both a name and user facing label. These extensions support both standard and custom actions.

We've also added actions support on Android, and added examples to RNTester showing how both standard and custom accessibility actions are used.

## Changelog

[general] [changed] - Enhanced accessibility actions support

## Test Plan

Added RNTester examples of both standard and custom actions, and verified that all examples work properly on Android with TalkBack and iOS with VoiceOver.


ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactAccessibilityDelegate.java
```diff
@@ -0,0 +1,288 @@
...
+  public boolean performAccessibilityAction(View host, int action, Bundle args) {
+    if (mAccessibilityActionsMap.containsKey(action)) {
+      final WritableMap event = Arguments.createMap();
+      event.putString("actionName", mAccessibilityActionsMap.get(action));
+      ReactContext reactContext = (ReactContext)host.getContext();
+      reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(
+          host.getId(),
+          "performAction",
+          event);
+      return true;
```
blavalla: Since you are returning here, rather than calling super it means that if someone sets onAccessibilityAction "activate", and also sets a standard click event, *only* the "activate" event would be fired when a user of an accessibility service clicks on the element.

Is this an intentional design choice? It definitely removes the ambiguity of which should fire first, but I am not sure if people will expect setting an onAccessibilityAction handler to break their click handler for some users.  If it is by design, we should probably include an example showing this in the AccessibilityExample.js file.

Thoughts?

marcmulcahy: @blavalla great point. I'll take a look at the AOSP source code to see what the default behavior for CLICK does these days-- it used to be that double tapping on an object simply synthesized a single tap to it, but things have changed of late.

In any case, it seems like activate might be a special case where we *do* want to call the super class's action implementation. But I think in most cases, we wouldn't want to call the superclass implementation.

Thoughts?

marcmulcahy: @blavalla just to be clear, setting an action handler for activate won't affect the standard click behavior-- It's generally used as an alternative path to that same code. So the normal case is that the user taps an object to activate it for example, and the "activate" action handler generally would just call the same logic as gets called when the user taps on the component.

I honestly can't think of a great use case where you'd want to override the standard behavior, but both iOS and Android support it, so...

marcmulcahy:
In the latest patches, I've added iOS and Android "focus" and "blur" support. Although I couldn't think of an accessible way to build a test into RNTester for them. Any thoughts?

ReactAndroid/src/main/java/com/facebook/react/uimanager/ReactAccessibilityDelegate.java
blavalla: By default as of Android 8.0, double-tap will first attempt to call the ACTION_CLICK AccessibilityAction:

https://github.com/aosp-mirror/platform_frameworks_base/blob/master/services/accessibility/java/com/android/server/accessibility/TouchExplorer.java#L393

And only if that fails it will then generate a synthetic click on the element.

This was added in Feb 2017 in this commit:

https://github.com/aosp-mirror/platform_frameworks_base/commit/0adfbd33c884ce80ebe161428d7a8ae9e6aced03#diff-1da9a3224134e2b3bf4c5d6be9ed0e52

So in 8.0 and above *only* "activate" would be called if a user had both "activate" and a normal click handler on an element, and they clicked it via the double-tap gesture.  But before 8.0, both the standard click handler *and* the "activate" handler would be called, one after the other.

I am not sure of the best way to handle this discrepancy to be honest. I'm not familiar with how ReactNative handles standard click events, but if we are already manually managing that behavior, we might be able to normalize the differences between newer and older versions of Android.

marcmulcahy: We discussed this internally, and all things considered, I think the behavior as defined in this PR is appropriate.

With recent versions of Android, the only reason you would want to override the default view behavior is if it didn't work for some reason, in which case, you probably don't want to call the superclass implementation anyway.

In older versions of Android, which synthesize the single tap by default when the user double taps an item using a screen reader, you might want to override this behavior to allow braille display users to click your component, in which case, you'd want to trigger the same logic as is triggered by double tap, which I think you could do in Javascript code.

So, I think overriding the default behavior is probably expected, although as you point out, might be confusing at first. Perhaps we can clear things up with documentation.

marcmulcahy:
I have disabled focus/blur support for now. I tried them both using accessibility actions, as well as by trapping the appropriate accessibility events as was done in #24642. And in both cases, I couldn't get blur working properly. And I suspect focus without blur support isn't terribly useful.

I'd prefer to separate those into a separate PR and sort them out there.

Anything still needing addressing with this PR?

blavalla:
I don't have any other feedback on the Android side.

@shergin, @ikenwoo, do you guys have any more feedback on the iOS or Javascript sides?

cpojer:
@marcmulcahy it seems like one of the Android tests is failing. Do you mind fixing it?

marcmulcahy:
@cpojer Sure, I won't be able to look at it for another couple hours. Can you point me to the failing test?

cpojer:
Check out the link next to the failed `test_android` step on CI: https://circleci.com/gh/facebook/react-native/91273?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

cpojer:
Thank you for fixing the test!

PR #24704, opened 2019-05-03
Consistent reporting native module name on crash on native side

DimitryDushkin:
## Summary

PR https://github.com/facebook/react-native/pull/24633 introduced some inconsistency in crash messaging, this PR fix it. Asked by @mhorowitz 

## Changelog

[General] [Added] - Consistent reporting native module name on crash on native side

## Test Plan

Actually I didn't managed to reproduce this kind of crash, looks like something changed here.

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -664,11 +664,11 @@ jsi::Object JSCRuntime::createObject(std::shared_ptr<jsi::HostObject> ho) {
     static JSValueRef getProperty(
         JSContextRef ctx,
         JSObjectRef object,
-        JSStringRef propertyName,
+        JSStringRef propName,
         JSValueRef* exception) {
       auto proxy = static_cast<HostObjectProxy*>(JSObjectGetPrivate(object));
       auto& rt = proxy->runtime;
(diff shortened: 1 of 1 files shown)


mhorowitz:
Thanks!  The code change looks good to me.  I think I'm having some issue on my end accepting, so I'll let @cpojer do that.

cpojer:
Unfortunately this broke some stuff at FB and we had to revert it:

```
Summary: 
react-native/Libraries/Blob/RCTBlobCollector.h:8:9: fatal error: 'jsi/jsi.h' file not found
#import <jsi/jsi.h>
        ^~~~~~~~~~~
```

Do you mind sending the same PR again with the correct fix? Thank you!

(Not sure why this was able to land in the first place, oops!)

cpojer:
Oh actually nevermind, we had two PRs landing at the same time and the other one was causing this issue. Do you mind sending the same PR again so I can re-land it? Thank you!

DimitryDushkin:
> Oh actually nevermind, we had two PRs landing at the same time and the other one was causing this issue. Do you mind sending the same PR again so I can re-land it? Thank you!

Sure — https://github.com/facebook/react-native/pull/24741

Issue #21515, opened 2018-10-05
[iOS][TextInput] onSelectionChange is not called in single line mode


liamse:
## Environment
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i5-2435M CPU @ 2.40GHz
      Memory: 101.08 MB / 10.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.1 - /usr/local/bin/node
      Yarn: 1.5.1 - /usr/local/bin/yarn
      npm: 6.4.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
    IDEs:
      Xcode: 9.4/9F1027a - /usr/bin/xcodebuild
    npmPackages:
      @types/react: ^16.4.14 => 16.4.14
      @types/react-native: ^0.57.0 => 0.57.0
      react: 16.5.0 => 16.5.0
      react-native: 0.57.1 => 0.57.1
    npmGlobalPackages:
      react-native-cli: 2.0.1

## Description
`onSelectionChange` method of `TextInput` does not return proper value when you use TextInput in single mode (not set `multiline = {true}`).

## Reproducible Demo

```js
// App.js
import React, {Component} from 'react';
import {Platform, StyleSheet, Text, View, TextInput} from 'react-native';

type Props = {};
export default class App extends Component<Props> {

  constructor(props) {
    super(props)
    this.state = {
      selection: {start: 0, end: 0},
      text: 'This is a sample text',
    }
  }
  render() {
    return (
        <View style={styles.container}>
          <Text>{this.state.selection.start}, {this.state.selection.end}</Text>
          <TextInput
            style={{ width: 300, borderColor: 'gray', borderWidth: 1 }}
            onSelectionChange={({ nativeEvent: { selection } }) => {
              this.setState({ selection })
            }}
            onChangeText={(text) => this.setState({ text })}
            value={this.state.text}
          />
        </View>
    );
  }
}
```

Run the code, if you select the part of text in TextInput, you need to see the number of start and end of selection, but you don't see anything unless you start to type, or cut the selected part, if the text is change the selection numbers appear. 

After select the part of text, the result is like following image: 
[...]

PR #24714, opened 2019-05-06
[iOS] Fixes selection of single line text input

zhongwuzw:
## Summary

`UITextField` don't have delegate to observe the selection changes of users(multiline text input `UITextView` have it), so we can add a KVO to observe selection changes.

@cpojer @shergin 

## Changelog

[iOS] [Fixed] - Fixes selection of single line text input

## Test Plan

Create a single line textInput, then do some selection, `onSelectionChange` can be called.

### DIFF
--- a/Libraries/Text/TextInput/RCTBackedTextInputDelegateAdapter.m
+++ b/Libraries/Text/TextInput/RCTBackedTextInputDelegateAdapter.m
@@ -36,7 +36,7 @@ static void *TextFieldSelectionObservingContext = &TextFieldSelectionObservingCo
 
 - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary *)change context:(void *)context
 {
-  // UITextField don't have delegate like UITextView to get notification when user do selection, here we use KVO to observe changes.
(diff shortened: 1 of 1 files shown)


JoshuaGross:
Hi @zhongwuzw, I'm reverting this because it's causing the following crash:

`Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x7fb3453eda00 of class RCTUITextField was deallocated while key value observers were still registered with it.`

I believe this is because we're adding a KVO observer to the UITextField but only have a weak reference to it, so it can be deallocated before the observer, causing this crash. Feel free to resubmit the PR with a fix for this crash as well.

zhongwuzw:
@JoshuaGross Oops, I missing that weak declaration, I'll fix it :).

bankya:
Hey, @zhongwuzw I was wondering if this change will still be merged since it looks like you made a fix for the crashing problem. The issue from #21515 still exists

zhongwuzw:
@bankya https://github.com/facebook/react-native/commit/c38f167019a3c481847d4abc80a458f7784f1336, v0.61.2 fixes, please check.

PR #24721, opened 2019-05-06
Fix CameraRollManager putImageInfo uris with hash

lunaleaps:
Summary:
[Android] [Fix] CameraRollManager errors when there is a hashtag in the file name

Instead of using `DefaultFileNameMap`'s implementation to `guessContentTypeFromName`, just take last characters before extension period and use `MimeTypeMap` to get the mime type (which is what URLConnection's `guessContent...` did)

Reviewed By: mdvacca

Differential Revision: D15212557

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/camera/CameraRollManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/camera/CameraRollManager.java
@@ -16,12 +16,11 @@ import android.media.MediaMetadataRetriever;
 import android.media.MediaScannerConnection;
 import android.net.Uri;
 import android.os.AsyncTask;
-import android.os.Build;
 import android.os.Environment;
 import android.provider.MediaStore;
 import android.provider.MediaStore.Images;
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/modules/camera/CameraRollManager.java
```diff
@@ -355,6 +353,13 @@ protected void doInBackgroundGuarded(Void... params) {
     }
   }
 
+  private static String getExtension(String path) {
+    if (path.contains(".")) {
+      return path.substring(path.lastIndexOf(".") + 1);
```
zhigang1992: Do we need to check for hashtag as well in this implementation?

lunaleaps: This would not require a check because the extension.. at least as i'm aware, shouldn't have a hashtag.

lunaleaps:
Closing this in favor of https://github.com/react-native-community/react-native-cameraroll/pull/54

PR #24726, opened 2019-05-07
[General] fixes fetch module credentials

zhongwuzw:
## Summary

In https://github.com/facebook/react-native/issues/14063, we decided to set `withCredential` to `true` by default, now if we use `XMLHttpRequest`, it works, but for fetch polyfill, `withCredential` is be set to `false` because we use `Request`'s credentials https://github.com/facebook/react-native/blob/master/Libraries/vendor/core/whatwg-fetch.js#L515.

So we have inconsistency about  `withCrendential`. My fix is to change `Request` `credentials` default value to `same-origin`, it reasonable because refer to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials), the default value is `same-origin`.

cc. @cpojer.

## Changelog

[General] [Fixed] - fixes fetch module credentials

## Test Plan

For any request from either XMLHttpRequest or fetch, we set `withCredential` to true by default.

### DIFF
--- a/Libraries/vendor/core/whatwg-fetch.js
+++ b/Libraries/vendor/core/whatwg-fetch.js
@@ -384,7 +384,7 @@
       this.url = String(input);
     }
 
-    this.credentials = options.credentials || this.credentials || 'omit';
+    this.credentials = options.credentials || this.credentials || 'same-origin';
     if (options.headers || !this.headers) {
       this.headers = new Headers(options.headers);
     }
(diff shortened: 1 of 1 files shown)


zhigang1992:
Seems like this file is getting revert back to the npm version. https://github.com/facebook/react-native/pull/24418

cpojer:
I don't think we can change this as we are getting rid of the `fetch` fork. Maybe we can change something in fetch to do the right thing by default for RN?

zhongwuzw:
@cpojer Please see https://github.com/facebook/react-native/pull/24418#issuecomment-489988041 or https://github.com/facebook/react-native/issues/14063. We changed some behavior because react-native is not for web.

PR #24729, opened 2019-05-07
[IOS] [Changed] add nonnull/nullable for swift

0xd-cc:
## Summary

add nonnull/nullable for swift

## Changelog

[iOS] [Changed] - add nonnull/nullable for swift

### DIFF
--- a/React/Base/RCTRootView.h
+++ b/React/Base/RCTRootView.h
@@ -35,7 +35,10 @@ extern "C"
 #else
 extern
 #endif
-NSString *const _Nonnull RCTContentDidAppearNotification;
+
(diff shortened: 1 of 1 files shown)


0xd-cc:
this code in swift , xcode  don't give me a warining, and it will crash

```
let bridge: RCTBridge? = nil
let moduleName: String? = nil
        
let rootView = RCTRootView(bridge: bridge, moduleName: moduleName, initialProperties: nil)
```

0xd-cc:
Hi,@shergin 
updated with NS_ASSUME_NONNULL_BEGIN/END  macro .

PR #24732, opened 2019-05-07
[iOS] Introduce TurboModule Setup Metric

RCiesielczuk:
## Summary

With the introduction of TurboModules, it would be beneficial to measure the setup time of these modules, as we currently have it in place for NativeModules.

The instantiation of the TMs occurs in the `RCTTurboModuleManager`. In order to successfully measure the time it took to setup the module, we need to ensure that we don't take into account cached modules. As such, we need to:

1. Check if module is in `_turboModuleCache`
  a. Start mark for `RCTPLTurboModuleSetup` tag if not found
2. Get the TM via `[self provideTurboModule:]`
3. Check if module is in `_turboModuleCache`
   a. Stop mark for `RCTPLTurboModuleSetup` if we did not find module in cache prior to **step 2** and if it's now present in the cache.
   b. Notify about setup time if the above is true.
4. Return TM

## Changelog

[iOS] [Added] - Gain insights on the the turbo module setup times by observing `RCTDidSetupModuleNotification`. The userInfo dictionary will contain the module name and setup time in milliseconds. These values can be extracted via `RCTDidSetupModuleNotificationModuleNameKey` and `RCTDidSetupModuleNotificationSetupTimeKey`. 

## Test Plan

Observe for `RCTDidSetupModuleNotification `

```objc
- (void)setup {
    [_notificationCenter addObserver:self
                            selector:@selector(didSetupModule:)
                                name:RCTDidSetupModuleNotification
                              object:nil];
}

- (void)didSetupModule:(NSNotification *)notification {
    NSDictionary *userInfo = notification.userInfo;
    if (userInfo) {
        NSString *moduleName = userInfo[RCTDidSetupModuleNotificationModuleNameKey];
        NSNumber *setupTime = userInfo[RCTDidSetupModuleNotificationSetupTimeKey];
        NSLog(@"%@ setup took %li ms", moduleName, [setupTime integerValue]);
    }
}
```

### Output
```
TestTurboModule setup took 1 ms
```

### DIFF
--- a/ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm
+++ b/ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm
@@ -10,6 +10,7 @@
 #import <cassert>
 
 #import <React/RCTBridge+Private.h>
+#import <React/RCTPerformanceLogger.h>
 #import <React/RCTBridgeModule.h>
 #import <React/RCTCxxModule.h>
 #import <React/RCTLog.h>
@@ -67,17 +68,43 @@ static Class getFallbackClassFromName(const char *name) {
 
       __strong __typeof(self) strongSelf = weakSelf;
 
+      auto moduleName = name.c_str();
+      auto moduleWasNotInitialized = ![strongSelf moduleIsInitialized:moduleName];
+      if (moduleWasNotInitialized) {
+        [strongSelf->_bridge.performanceLogger markStartForTag:RCTPLTurboModuleSetup];
+      }
+
       /**
        * By default, all TurboModules are long-lived.
        * Additionally, if a TurboModule with the name `name` isn't found, then we
        * trigger an assertion failure.
        */
-      return [strongSelf provideTurboModule: name.c_str()];
+      auto turboModule = [strongSelf provideTurboModule:moduleName];
+
+      if (moduleWasNotInitialized && [strongSelf moduleIsInitialized:moduleName]) {
+        [strongSelf->_bridge.performanceLogger markStopForTag:RCTPLTurboModuleSetup];
+        [strongSelf notifyAboutTurboModuleSetup:moduleName];
+      }
+
+      return turboModule;
     };
 
     _binding = std::make_shared<react::TurboModuleBinding>(moduleProvider);
   }
   return self;
+}
+
+- (void)notifyAboutTurboModuleSetup:(const char*)name {
+    NSString *moduleName = [[NSString alloc] initWithUTF8String:name];
+    if (moduleName) {
+      int64_t setupTime = [self->_bridge.performanceLogger durationForTag:RCTPLTurboModuleSetup];
+      [[NSNotificationCenter defaultCenter] postNotificationName:RCTDidSetupModuleNotification
+                                                          object:nil
+                                                        userInfo:@{
+                                                                  RCTDidSetupModuleNotificationModuleNameKey: moduleName,
+                                                                  RCTDidSetupModuleNotificationSetupTimeKey: @(setupTime)
+                                                                  }];
+    }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


shergin:
cc @RSNara

RCiesielczuk:
Hello @shergin and @RSNara,

Have you managed to take a look at this PR?

ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm
```diff
@@ -67,19 +68,47 @@ - (instancetype)initWithRuntime:(jsi::Runtime *)runtime
 
       __strong __typeof(self) strongSelf = weakSelf;
 
+      auto moduleName = name.c_str();
+      bool didFindModuleInCacheBeforeCreation = strongSelf->_turboModuleCache.find(moduleName) != strongSelf->_turboModuleCache.end();
+      if (!didFindModuleInCacheBeforeCreation) {
+        [strongSelf->_bridge.performanceLogger markStartForTag:RCTPLTurboModuleSetup];
+      }
+
```
RSNara: nit: The variable name `didFindModuleInCacheBeforeCreation ` is a bit strange. What you're really checking here is: "Does this TurboModule exist?". So you could just use the `moduleIsInitialized` function defined further down in the file.

ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm
```diff
@@ -67,19 +68,47 @@ - (instancetype)initWithRuntime:(jsi::Runtime *)runtime
...
       /**
        * By default, all TurboModules are long-lived.
        * Additionally, if a TurboModule with the name `name` isn't found, then we
        * trigger an assertion failure.
        */
-      return [strongSelf provideTurboModule: name.c_str()];
+
+      std::shared_ptr<react::TurboModule> turboModule = [strongSelf provideTurboModule:moduleName];
+
+      bool didFindModuleInCacheAfterCreation = strongSelf->_turboModuleCache.find(moduleName) != strongSelf->_turboModuleCache.end();
```
RSNara: Same deal here.

React/Base/RCTBridge.m
```diff
@@ -27,7 +27,8 @@
 NSString *const RCTJavaScriptDidLoadNotification = @"RCTJavaScriptDidLoadNotification";
 NSString *const RCTJavaScriptDidFailToLoadNotification = @"RCTJavaScriptDidFailToLoadNotification";
 NSString *const RCTDidInitializeModuleNotification = @"RCTDidInitializeModuleNotification";
-NSString *const RCTDidSetupModuleNotification = @"RCTDidSetupModuleNotification";
+NSString *const RCTDidSetupNativeModuleNotification = @"RCTDidSetupNativeModuleNotification";
+NSString *const RCTDidSetupTurboModuleNotification = @"RCTDidSetupTurboModuleNotification";
```
RSNara: A module is either a NativeModule or a TurboModule (and you know which is which). So I'm not sure why it's necessary to to break this notification down into two. Eventually, we plan to remove the Bridge and its accompanying NativeModule system, so we'll have to undo this change anyway. Thoughts?

ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm
RCiesielczuk: Ahh, perfect - How did I miss that one out? 😅

React/Base/RCTBridge.m
RCiesielczuk: Good shout. It really didn't make sense to do so when I looked at this for the second time.

RCiesielczuk:
Hi @RSNara, I've implemented the changes thanks to your feedback and I think we've covered everything.

PR #24687, opened 2019-05-02
New Intro Screen Proposal

karanpratapsingh:
## Summary

Issue: [Polish the "new app screen"](https://github.com/react-native-community/discussions-and-proposals/issues/122)
This is the pull request for the new intro screen proposal in react native as directed by @cpojer 

## Changelog
[General][Added] - New Intro screen, Icons

Removed Lottie Integration
100% React Native 💥 

## Test Plan

```npm run ios``` should build the template app similar to this:

type 1

PR #24737, opened 2019-05-07
Polish the new app screen

cpojer:
## Summary

Continuation of #24687

> Issue: [Polish the "new app screen"](https://github.com/react-native-community/discussions-and-proposals/issues/122)
> This is the pull request for the new intro screen proposal in react native as directed by @cpojer 

This PR was created because the previous one could not be pushed to for some reason. I cleaned up a few small things and added the component as an example to RNTester so we can keep iterating. My plan is to land this, and then polish it and make it the default in a follow-up.

## Changelog
[General][Added] - New Intro screen, Icons

Removed Lottie Integration
100% React Native 💥 

## Test Plan

```npm run ios``` should build the template app similar to this:

type 1

### CODE REVIEW

**Comment by cpojer** [comment]
@eliperkins @elucaswork @gengjiawen do you have any current feedback before landing this? If not, please approve it. I can then land it, and if you like you can send more PRs to polish it.

@orta could you give this a spin and tell me whether it matches your mockup to the point where you feel comfortable shipping it?

**Comment by orta** [comment]
Yeah, this looks good to me.

I don't know if it's your plan, but I'd love it if the `index.js` from this _was_ the file in the template in your app, so you see some examples on how to do styling, platform checking, ScrollViews etc. ( e.g. not just `render() { <NewAppScreen/> }` in user-land code)

File: Libraries/NewAppScreen/index.js [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+};
+
+const App = () => {
+  return (
+    <Fragment>
+      <SafeAreaView style={styles.topSafeArea} />
+
+      <SafeAreaView style={styles.bottomSafeArea}>
+        <StatusBar barStyle="dark-content" />
+        <ScrollView bounces={false}>
```
elucaswork: is `bounces={false}` needed here? IMH it's unusual to remove it.

cpojer:
@orta I'm trying to figure out what a good middle ground is for the App template. On one hand I don't want to increase the app size a lot or bloat the template but on the other I do want there to be more than a single React component that is inlined. Do you have any recommendation on a good balance?

orta:
My opinion, based on the idea that this is also "sample code" for when you first see react native. Move anything too 'complex'  into `react-native` which is _privately_ exported (and not considered the API):

Making the `index.js`' header more like:

```
import {View, Text, StyleSheet, ImageBackground} from 'react-native';
import {Header, Section, LearnMoreLinks, colors} from 'react-native/components/new-app-screen';
```

Which would contain the components we see today:

- `Header` 
- `Section`
- `Learn More Links`
- `colors`
- An abstraction that combines the Fragment, and the two `SafeAreaView`s?
- I'm 60/40 on moving `ReloadInstructions` and `DebugInstructions` into it too.

If this is the sample code to give you a feel for what your React Native app will look like, then it's good to show:

- Using a `ScrollView`
- Using a `StyleSheet` (and how it hooks up to the `style` prop)
- Using a `Text` and understanding how the text is a child of it
- (_Ideally_) Using a `TouchableOpacity` to show how to do something which interacts (from reading the current code, this looks tricky)

As these are kinda the primitives for making an app at all.  Given that RN ships untranspiled, an editor should be able to jump to the 'hidden' components in RN to read.

Libraries/NewAppScreen/components/Header.js
```diff
@@ -0,0 +1,56 @@
...
+  container: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    paddingTop: 100,
+    paddingBottom: 40,
+    paddingHorizontal: 32,
+    backgroundColor: Colors.lighter,
+  },
+  backgroundLogo: {
+    position: 'absolute',
```
eliperkins: I'm fine landing this in for now, but I definitely fear other folks learning RN and styling for the first time picking up on using `absolute` to style things.

Libraries/NewAppScreen/components/Header.js
```diff
@@ -0,0 +1,56 @@
...
+    top: -20,
+    left: -200,
+    opacity: 0.2,
+    alignItems: 'center',
+    justifyContent: 'center',
+    height: 540,
```
eliperkins: Do these heights need to be explicit? What does this look like in landscape or on tablets?

Libraries/NewAppScreen/components/Colors.js
```diff
@@ -0,0 +1,20 @@
...
+'use strict';
+
+export default {
+  primary: '#1292B4',
+  white: '#FFF',
+  lighter: '#F3F3F3',
+  light: '#DAE1E7',
+  dark: '#444',
+  black: '#000',
+};
```
luisfmsouza: We are mixing color names (white and black), hierarchy pattern (primary) and color states (lighter, light, dark). Think would be better to use only one pattern to name colors here, it will help to understand the code.

Suggestion:

```js
export default {
  bondiBlue: '#1292B4',
  white: '#FFF',
  whiteSmoke: '#F3F3F3',
  pattensBlue: '#DAE1E7',
  charcoal: '#444',
  black: '#000',
};
```

I've used this [site as a reference](https://www.htmlcsscolor.com/hex/444444)

Libraries/NewAppScreen/index.js
```diff
@@ -0,0 +1,139 @@
...
+      <Text style={styles.highlight}>Shake</Text> your device to open the React
+      Native debug menu.
+    </Text>
+  );
+};
+
+const App = () => {
+  return (
+    <Fragment>
+      <SafeAreaView style={styles.topSafeArea} />
```
eliperkins: I'm still a bit confused why we need two `SafeAreaView`s. My understanding is that the `SafeAreaView` with the children should handle both top and bottom safe area insets.

glauberfc: @eliperkins I used two safe areas because I wanted the top safe area with the color of the header and the bottom with the color of the screen background. There is another way to do this?

eliperkins: Ah, interesting! I hadn't thought about that use-case. This totally makes more sense why we would reach for multiple `<SafeAreaView>`s now.

One option could be to use `ScrollView`'s `contentInsetAdjustmentBehavior`: https://facebook.github.io/react-native/docs/scrollview.html#contentinsetadjustmentbehavior

To do this, we would have to move the `ScrollView` up to be the root view (that is, outside of the `SafeAreaView` this is currently containing it) and this will let UIKit take over on adjusting the insets for the safe areas. Then, by setting the background of the `ScrollView` to be `Colors.lighter`, we could accomplish this all without `SafeAreaView` at all!

I put a Snack up of this here: https://snack.expo.io/@eliperkins/lonely-churros

The downside to this approach is that the end of our scroll view would then be the grey color as well. I think the trade off is better than over-prescribing `<SafeAreaView>`s!

eliperkins: This could also use `Animated` to change the color of the `<ScrollView>`'s `backgroundColor` while scrolling, but that seems a bit too complex for a sample app.

Libraries/NewAppScreen/components/Header.js
lucasbento: Definitely agree with this, is `position: absolute` required here? I hardly use it.

Libraries/NewAppScreen/index.js
```diff
@@ -0,0 +1,139 @@
...
+    flex: 1,
+    backgroundColor: Colors.white,
+  },
+  body: {
+    backgroundColor: Colors.white,
+  },
+  sectionTitle: {
+    fontSize: 24,
+    fontWeight: '600',
+    color: '#000',
```
lucasbento: Please use `Colors.black`.

PR #24739, opened 2019-05-07
Fix DatePickerAndroid with mode spinner on Android Nougat(7.0)

luancurti:
When mode is spinner in Android Nougat the DatePicker shows calendar

## Summary

I got it from https://gist.github.com/jeffdgr8/6bc5f990bf0c13a7334ce385d482af9f and
did some adjustments in order to work with `DatePicker`. Workaround for this bug: https://code.google.com/p/android/issues/detail?id=222208. In Android 7.0 Nougat, spinner mode for the DatePicker in DatePickerDialog is incorrectly displayed as calendar.

## Changelog

[Android][Fixed] Fix date picker with mode spinner on Android Nougat (7.0)

## Test Plan

Run RNTester with Simulator that has Android Nougat (7.0)
Search DatePickerAndroid
Click in Simple spinner date picker

datepicker_android

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/datepicker/DismissableDatePickerDialog.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/datepicker/DismissableDatePickerDialog.java
@@ -8,11 +8,15 @@
 package com.facebook.react.modules.datepicker;
 
 import android.app.DatePickerDialog;
-import javax.annotation.Nullable;
-
-import android.app.DatePickerDialog;
 import android.content.Context;
 import android.os.Build;
+import android.content.res.TypedArray;
+import android.util.AttributeSet;
+import android.widget.DatePicker;
+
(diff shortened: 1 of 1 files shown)


dulmandakh:
@luancurti I looked into the Android issue tracker and it seems that issue exists only on Android 7.0, so it might be best to use fixSpinner on Android 7.0. Could you please change and check that it works for other versions?

luancurti:
Hi @dulmandakh I tested only on Android Nougat but now I retested again in this Android versions 5.0, 7.0 and 9.0, and I changed this line `if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {` to only affect the Android Nougat because in another versions the `RNTester` crash because `DatePickerAndroid` thanks to remember!

Below are the test screenshots.

Android 5.0

Screenshot_1558058858

Android 7.0

Screenshot_1558059799

Android 9.0

Screenshot_1558059211

PR #24743, opened 2019-05-07
Move accessibility props to UIView+React

janicduplessis:
## Summary

React Native Gesture Handler uses a `RCTViewManager` subclass to manage `UIControl` so the cast to set accessibility props is not safe. This moves the accessibility props we set to `UIView+React` so they can be used safely on any `UIView`.

image

## Changelog

[General] [Fixed] - Move accessibility props to UIView+React

## Test Plan

Tested that it fixes the crash and that RNTester accessibility examples still work.


cpojer:
Actually nevermind, it seems to happen on other PRs as well.

PR #24745, opened 2019-05-07
[Blob] Release underlying resources when JS instance in GC'ed

### LINKED ISSUE #23801: Severe memory leak affecting fetch(), iOS Release Mode 0.59.0-rc.3, 0.58.6, 0.57.8

janicduplessis:
#24405 try 2

## Summary

Our Blob implementation was very problematic because it didn't release its underlying resource when the JS instance was dealocated. The main issue is that the fetch polyfill uses blobs by default if the module is available, which causes large memory leaks.

This fixes it by using the new jsi infra to attach a `jsi::HostObject` (`BlobCollector`)  to `Blob` instances. This way when the `Blob` is collected, the `BlobCollector` also gets collected. Using the `jsi::HostObject` dtor we can schedule the cleanup of native resources. This is definitely not the ideal solution but otherwise it would require rewriting the whole module using TurboModules + jsi. 

Fixes #23801, #20352, #21092

## Changelog

[General] [Fixed] - [Blob] Release underlying resources when JS instance in GC'ed

## Test Plan

- Tested that creating a simple Blob instance results in resource allocation and deallocation. (`new Blob(['HI'])`)
- Tested that blob resources created by the fetch polyfill eventually get collected.

### DIFF
--- a/Libraries/Blob/BlobManager.js
+++ b/Libraries/Blob/BlobManager.js
@@ -14,7 +14,7 @@ const Blob = require('Blob');
 const BlobRegistry = require('BlobRegistry');
 const {BlobModule} = require('NativeModules');
 
-import type {BlobData, BlobOptions} from 'BlobTypes';
+import type {BlobData, BlobOptions, BlobCollector} from 'BlobTypes';
 
 /*eslint-disable no-bitwise */
 /*eslint-disable eqeqeq */
@@ -29,6 +29,21 @@ function uuidv4(): string {
       v = c == 'x' ? r : (r & 0x3) | 0x8;
     return v.toString(16);
   });
+}
+
+// **Temporary workaround**
+// TODO(#24654): Use turbomodules for the Blob module.
+// Blob collector is a jsi::HostObject that is used by native to know
+// when the a Blob instance is deallocated. This allows to free the
+// underlying native resources. This is a hack to workaround the fact
+// that the current bridge infra doesn't allow to track js objects
+// deallocation. Ideally the whole Blob object should be a jsi::HostObject.
+function createBlobCollector(blobId: string): BlobCollector | null {
+  if (global.__blobCollectorProvider == null) {
+    return null;
+  } else {
+    return global.__blobCollectorProvider(blobId);
+  }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


hramos:
The error still occurs, but our current understanding is that this is specific to one of our internal Buck tests, and that we'll need to update our Buck definition, something that you won't be able to change as part of this PR.

Please hold off on adding any commits to this PR. I'll work on making the necessary changes over the internal version of this PR (where I can update filepaths outside of the React Native GitHub directory) and once the issue is resolved and the internal diff lands, this PR shall be marked as merged.

janicduplessis:
Awesome! Thanks @hramos

hramos:
I have a potential fix waiting for review, shouldn't be long.

janicduplessis:
@hramos There's a conflict because of 0ee5f68929610106ee6864baa04ea90be0fc5160, let me know if you need me to do anything.

hramos:
The conflict did hold this back, but I just resolved it internally and kicked off another land attempt.

Issue #15961, opened 2017-09-15
[ANDROID] Linking.getInitialURL() returns null after exiting the app with back button and using deep linking again (only in production!)


ferrannp:
### Is this a bug report?

Yes.

### Environment

1. `react-native -v`: 0.46.4
2. `node -v`: 8.4.0
3. `npm -v`: 5.3.0
4. `yarn --version`: 0.27.5

- Target Platform: Android
- Development Operating System: macOS

### Steps to Reproduce

Configure your app to receive deep linking links as explained here: https://facebook.github.io/react-native/docs/linking.html.

My native code looks like:

```xml
<intent-filter>
    <action android:name="android.intent.action.VIEW" />
    <category android:name="android.intent.category.DEFAULT" />
    <category android:name="android.intent.category.BROWSABLE" />
    <data android:scheme="https"
          android:host="some-url.com"
          android:pathPrefix="/some-prefix" />
</intent-filter>
```
I also set the Android Activity to `singleTask`.

On the JS code:

```js
componentDidMount() {
  Linking.getInitialURL().then((url: string) => {
    if (url) {
      this._handleLinking(url);
    }
  });
}
```

### Expected Behavior

1. URL is not null when handling deep linking. I am using the command:

```
$ adb shell am start
        -W -a android.intent.action.VIEW
        -d "example://gizmos" com.example.android
```

#### Development

- I kill the app (manually removing it from apps), I ran the adb command, url is not null and deep linking is ok
- I open the app, I kill the activity with the back button, I ran the adb command, url is not null and deep linking is ok

#### Production

- I kill the app (manually removing it from apps), I ran the adb command, url is not null and deep linking is ok
- I open the app, I kill the activity with the back button, I ran the adb command, **url is null so I cannot handle the deep linking**

From Logcat:

```
09-15 18:18:52.262 675-686/system_process I/ActivityManager: START u0 {act=android.intent.action.VIEW dat=https://some-url.com/... flg=0x10000000 pkg=com.myapp cmp=com.myapp/.MainActivity} from uid 0 on display 0
09-15 18:18:52.386 8191-10955/com.myapp I/ReactNativeJS: Running application "MyApp" with appParams: {"rootTag":11}. __DEV__ === false, development-level warning are OFF, performance optimizations are ON
09-15 18:18:52.393 8191-10955/com.myapp I/ReactNativeJS: getInitialURL null
```

As you can see Intent has data but the `console.log` inside `getInitialURL` shows `null` as url.

[...]

PR #24748, opened 2019-05-08
[Android] Make Linking.getInitialURL works properly even if current activity is not initialized

mu29:
## Summary

Solves #15961 - [ANDROID] Linking.getInitialURL() returns null after exiting the app with back button and using deep linking again (only in production!)

As the issue says, it seems that `currentActivity` is null in [IntentModule#getInitialURL](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/modules/intent/IntentModule.java#L50). So I used the `InteractionManager` to wait until current activity to finish initializing.

## Changelog

[Android] [Fixed] - In `Linking.getInitialURL` method, use the `InteractionManager` to wait for the current activity to finish initializing.

## Test Plan

Here is an error case. After several trial, my app cannot open deep link.
Video link: https://youtu.be/0qppEs4O1Ls

### DIFF
--- a/Libraries/Linking/Linking.js
+++ b/Libraries/Linking/Linking.js
@@ -13,6 +13,7 @@
 const NativeEventEmitter = require('NativeEventEmitter');
 const NativeModules = require('NativeModules');
 const Platform = require('Platform');
+const InteractionManager = require('InteractionManager');
 
 const invariant = require('invariant');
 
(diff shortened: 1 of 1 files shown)


satya164:
This change seems to be causing this issue https://github.com/facebook/react-native/issues/25675

Stephen2:
UPDATE: I don't believe this change is responsible (or at least not wholly responsible).  See update on #25675

PR #24749, opened 2019-05-08
[standards] Migrate "Libraries" from Haste to standard path-based requires (sans vendor & renderers)

ide:
## Summary

This is the next step in moving RN towards standard path-based requires. All the requires in `Libraries` have been rewritten to use relative requires with a few exceptions, namely, `vendor` and `Renderer/oss` since those need to be changed upstream. This commit uses relative requires instead of `react-native/...` so that if Facebook were to stop syncing out certain folders and therefore remove code from the react-native package, internal code at Facebook would not need to change.

See the umbrella issue at https://github.com/facebook/react-native/issues/24316 for more detail.

## Changelog

[General] [Changed] - Migrate "Libraries" from Haste to standard path-based requires

## Test Plan

Prettier/Lint, Flow, Jest tests, loaded RNTester

### DIFF
--- a/Libraries/ART/ReactNativeART.js
+++ b/Libraries/ART/ReactNativeART.js
@@ -10,15 +10,15 @@
 'use strict';
 
 const Color = require('art/core/color');
-const Path = require('ARTSerializablePath');
+const Path = require('./ARTSerializablePath');
 const Transform = require('art/core/transform');
 
-const React = require('React');
+const React = require('react');
 const PropTypes = require('prop-types');
-const ReactNativeViewAttributes = require('ReactNativeViewAttributes');
+const ReactNativeViewAttributes = require('../Components/View/ReactNativeViewAttributes');
 
-const createReactNativeComponentClass = require('createReactNativeComponentClass');
-const merge = require('merge');
+const createReactNativeComponentClass = require('../Renderer/shims/createReactNativeComponentClass');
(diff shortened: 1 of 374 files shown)


cpojer:
Current status: I'm trying to get this landed but it may require more work at FB. So far I'm reverting `react-native-implementation` for now as that causes a lot of issues with Jest mocking and the requires for ReactFiber/ReactNative because at FB they should be required from the `fb` folder and not `oss`. We use haste for that toggle atm.

I only have one more Flow issue in our integration tests remaining, will give another update later.

cpojer:
(Please don't push changes to this PR)

ide:
This PR was mostly a script, followed by two FlowIgnoreMes (it appears that strict-local doesn’t behave as the Flow docs describe), Prettier and manual reverts for non-Libraries parts of the repo.

cpojer:
Yay, it landed! Here are some follow-ups:
* I had to revert changes to `react-native-implementation`. We'll need a change for a ton of internal tests and add an API like `jest.mockReactNativeModule('<HasteModuleName>')` that works like `jest.mock` but mocks the full path and also patches the mock onto the exports of `react-native-implementation`. I'm not entirely sure what the issue was that I was seeing internally but it's gotta be something like this.
* Figure out a way how to pick the React renderer.
* Actually disable haste within this folder so it will throw when using haste.

PR #24758, opened 2019-05-08
[iOS] Optimize react children views management

zhongwuzw:
## Summary

No any logical changed, we have three parts to handle remove/move/add operations, most of the time, only one part may need to handle, I reorder these parts to break early if any of it doesn't need to handle, this is for performance boost.

cc. @cpojer @shergin.

## Changelog

[iOS] [Changed] - Optimize react children views management

## Test Plan

CI passes.


zhongwuzw:
@sammy-SC Thanks for review, I fixed the tests, please check again :).

shergin:
This is already pretty complex method. Adding more if-statements does not make it simpler and does not improve maintainability.

I do want to improve this code thought. So, I would love to land any changes that improve perf but only if they also make the code much more clear, easier to understand and maintain.

PR #24777, opened 2019-05-09
[iOS] optimize layout block iteration for root view

zhongwuzw:
## Summary

No logical changed, the goal is to eliminate iteration from O(2N) to O(N).
We iterate `affectedShadowViews` array repeated, it's unnecessary and we can consolidate them.

## Changelog

[iOS] [Changed] - optimize layout block iteration for root view

## Test Plan

CI passes.

### DIFF
--- a/React/Modules/RCTUIManager.m
+++ b/React/Modules/RCTUIManager.m
@@ -516,51 +516,53 @@ static NSDictionary *deviceOrientationEventBody(UIDeviceOrientation orientation)
   NSUInteger count = affectedShadowViews.count;
   NSMutableArray *reactTags = [[NSMutableArray alloc] initWithCapacity:count];
   NSMutableData *framesData = [[NSMutableData alloc] initWithLength:sizeof(RCTFrameData) * count];
-  NSUInteger index = 0;
-  RCTFrameData *frameDataMutableArray = (RCTFrameData *)framesData.mutableBytes;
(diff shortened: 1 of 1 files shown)


shergin:
Thank you for doing this, but before we can review this (and merge), could you please:
* Decouple *all* code-style/variable-renaming changes into separate PR.
* Describe exact problem that we are solving with this PR.

I am sure that we want to trade readability for such kinda of optimization (that should be performed by compiler for us).

zhongwuzw:
@shergin I updated the description. You mean compiler can optimized that two iterations?

Some explain that you saw the variable-renaming things, because I consolidate above iterations, and remove the bracket scope, compiler warning shadow variable.

zhongwuzw:
And seems we can also optimize views management, we have much places to index the view from array recursively, like remove views, it's linear. 

A simple case is https://github.com/facebook/react-native/blob/master/React/Views/UIView%2BReact.m#L98, I just log the `subviews.count`, run the RNTester, the biggest number I see it 61.

zhongwuzw:
@shergin I reopen because only that I can update the commit, and I expressed my thoughts, please see ^ . If you think I may wrong, feel free to close it. :)

osdnk:
@zhongwuzw This code actually doesn't make any vital optimisation.

It's true that not we have `N` iteration instead of `2N`, but each of them is now bigger so it sums to the same amount of code to be executed.

PR #24782, opened 2019-05-09
[standards] ReactPrivate, an explicit interface between the renderer and RN

ide:
## Summary

This introduces a new library named "ReactPrivate" that defines an explicit interface between the React renderers generated by the React repo and the code within RN. Previously, the React renderers would reach into RN internals via Haste wormholes. With this commit, there is now an explicit module (`ReactNativePrivateInterface`) that the renderers use to access RN internals.

Motivation: The main goal is to move one step closer to turning off Haste for RN (https://github.com/facebook/react-native/issues/24316). Since the generated renderers currently use Haste, this commit sets the foundation for giving them a path-based interface to access RN internals.

Additionally, this approach inverts abstraction control since RN needs to intentionally export its internals via the private interface instead of React reaching in via Haste.

There will also need to be a corresponding commit to the React repo to make the renderers use this new interface. This RN commit needs to land before the React commit.

## Changelog

[General] [Changed] - Add a private interface (do not use) between the renderer and RN

## Test Plan

Run unit tests, CI. This commit should be safe since it just introduces new modules.

Also tested with newly generated renderers (not in this commit; needs to happen in the React repo) that use ReactPrivate instead of Haste and verified that RNTester loads and that unit tests pass.


ide:
@cpojer -- This is ready for review (tests passing, conforms to the interface Sebastian suggested on the React PR).

PR #24786, opened 2019-05-09
Delete LICENSE

suny000:
## Changelog

[CATEGORY] [TYPE] - Message


cpojer:
Thanks but this doesn’t seem right. React Native uses the MIT license and we have to share this license file.

PR #24790, opened 2019-05-10
Add `Platform.nightMode` for Android

turnrye:
## Summary

Android Q is raising Dark Mode to prominence across the ecosystem. One of the expectations is to be able to respond to the system-wide dark theme setting. I've followed their documentation, added the flag to the InfoModule, and then presented this value to the RN user space via an Android-specific Platform value `nightMode`. There are rumors that similar functionality may come soon for iOS, and we may want to consider refactoring the JS side at that point for consistency between the two platforms.

## Changelog

[Android] [Added] - Add ability to check what system preference is set for dark mode

## Test Plan

I added `Platform.nightMode` to a sample app and saw it get "yes" and "no" based off of whether I'd selected light theme or dark theme on the Q beta.

### DIFF
--- a/Libraries/Utilities/Platform.android.js
+++ b/Libraries/Utilities/Platform.android.js
@@ -16,6 +16,7 @@ export type PlatformSelectSpec<A, D> = {
   android?: A,
   default?: D,
 };
+export type NightMode = "yes" | "no" | "auto" | "unknown" | false;
 
 const Platform = {
   OS: 'android',
@@ -33,6 +34,10 @@ const Platform = {
   get isTV(): boolean {
     const constants = NativeModules.PlatformConstants;
     return constants && constants.uiMode === 'tv';
+  },
+  get nightMode(): NightMode {
+    const constants = NativeModules.PlatformConstants;
+    return constants && constants.nightMode;
   },
   select: <A, D>(spec: PlatformSelectSpec<A, D>): A | D =>
     'android' in spec ? spec.android : spec.default,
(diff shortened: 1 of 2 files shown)


gengjiawen:
Also need to update docs, Can you post a PR to https://github.com/facebook/react-native-website too.

turnrye:
@gengjiawen thank you! Submitted (and referenced above).

gengjiawen:
@turnrye Can you rebase this PR, seems Circle CI not triggered.

necolas:
Multiple platforms (web, macOS) already support dark modes too. The web API is a new media-query `prefers-color-scheme` that is partially supported at the moment - https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme.

If we were to follow that path, we'd have an API more like `Platform.colorScheme === 'dark'` (or `light` or `none`). That could leave the door open for additional values like `high-contrast` etc.

mmmulani:
we should have a listener pattern here instead, since dark/night mode can change on the fly, which should cause a re-render.

cpojer:
We had a conversation about this at FB and the conclusion seems to be that this shouldn't be part of Platform because it isn't tied to the platform but rather a user setting that can change at anytime, unlike Platform.

My suggestion would be to make this a third-party native module first, and consider it for inclusion to React Native later.

necolas:
> My suggestion would be to make this a third-party native module first, and consider it for inclusion to React Native later.

This is going to be a very common feature to support, including at FB, so it might be worth getting ahead of the API design churn and suggesting something ourselves soon.

cpojer:
I'm fine considering this for inclusion into core if people agree but we should discuss a proposal in https://github.com/react-native-community/discussions-and-proposals first.

turnrye:
Opened https://github.com/react-native-community/discussions-and-proposals/issues/126

PR #24793, opened 2019-05-10
Fix PickerAndroid will reset selected value during items update.

### LINKED ISSUE #13351: Picker Component resets when items list changes
### Description

If the picker item list changes, then, in some cases, the picker will call the `onValueChange` callback with the first item in the list as it's argument.

This can be reproduced at [https://snack.expo.io/Sy1JClEag](https://snack.expo.io/Sy1JClEag).

Select 2 or 3, then click on exclude. The new value is 1. This does not happen if you select a higher value than the one being excluded

### Solution

Do not call the `onValueChange` callback when the items list changes, but the `selectedValue` is stil present in the list

### Additional Information

* React Native version: 0.42
* Platform: Android
* Development Operating System: Linux
* Dev tools: Android SDK version 23

Kudo:
## Summary
Fixes #13351

Two root causes:
1. Android Spinner will reset selection to undefined after setAdapter()
   which will trigger onValueChange().
   The behavior is not expected for RN.
   And the solution is to setSelection() explicitly

2. In original implementation, it setups `items` immediately,
   but delays the `selected` setup after update transaction.
   There will be some race condition and incosistency
   if update `items` only.
   The fix will do the setup all after update transaction.

## Changelog

[Android] [Fixed] - Fix #13351 PickerAndroid will reset selected value during items update.

## Test Plan

1. Test the test case as @stochris [reported](https://github.com/facebook/react-native/issues/13351#issue-219980433) in https://snack.expo.io/Sy1JClEag.
2. Test RNTester Picker sample

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPicker.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPicker.java
@@ -13,6 +13,7 @@ import android.util.AttributeSet;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.Spinner;
+import android.widget.SpinnerAdapter;
 
 import com.facebook.react.common.annotations.VisibleForTesting;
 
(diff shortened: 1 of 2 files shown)


axe-fb:
@Kudo - Looks like this PR is causing an increase in [ArrayOutOfBoundsException at getView](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPickerManager.java#L116). Do you have context on why this is happening ? 

I dont have context on this - do you think this is related ?

Kudo:
@axe-fb I've sent a PR #25276, hopefully that will fix your problem. Thanks for the reporting.

Issue #24760, opened 2019-05-08
New App Screen


cpojer:
This is the list of remaining things to do before we can ship the new app screen (see https://github.com/react-native-community/discussions-and-proposals/issues/122). The initial version of the code was landed via https://github.com/facebook/react-native/pull/24737:

- [x] Fix up the styling and color names a bit more.
- [x] Split up the components the way @orta proposed so we'll have some good sample code in the App.js template.
- [x] Make it so that links tapped will send an open-in-browser command via Metro on the host instead of in the emulator. See https://github.com/react-native-community/cli/blob/885f2c28d4e8ed66e334fd7bd0174173b2a394e1/packages/cli/src/commands/server/middleware/openStackFrameInEditorMiddleware.js for how we are doing this in the cli/metro (this may need some changes to the CLI and Metro) and https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/Core/Devtools/openFileInEditor.js (@lucasbento)
- [x] Add a "Follow us on Twitter" button 😄
- [x] Actually put this all in the template so it can be shipped as part of 0.61.

cc @eliperkins @lucasbento would be awesome if you could help out! :)

lucasbento:
I would love to help on this, specially the opening links on the browser part.

Would this be targeting 0.61? I'm currently busy with some stuff for the CLI but I would probably be able to do some stuff for this next week.

cpojer:
Yep, 0.61. Ideally we'll wrap this work up by the end of the month.

cpojer:
@eliperkins do you wanna take care of adding a follow on twitter link + putting it into the actual template? :)

eliperkins:
@cpojer sure thing. I'll tackle the moving to the template part now that the polish PR has landed.

As for the Twitter link, do you expect that to be a link to https://twitter.com/reactnative or is there another, more follow-specific link that we can put in there?

cpojer:
Yep, that links is good :)

eliperkins:
#24805 and #24806 are up now! 

I think there's still some work to do to rearrange components and tuck some complexity away before I'd call this template ready to ship, so I'd love some feedback.

lucasbento:
@ericlewis opened a PR for the open in browser functionality on CLI: react-native-community/cli#383.

cpojer:
[...]

PR #24805, opened 2019-05-10
Replace new app template with new app screen

eliperkins:
## Summary

This replaces the "new app screen" in the template with the new design from https://github.com/react-native-community/discussions-and-proposals/issues/122

This uses components that are shipped as part of the `react-native` module, but not necessarily as proper components exported by the main `react-native` module. To use these, we use absolute imports to those components.

Related to #24760

## Changelog

[General] [Changed] - Updated new app template design 💖

## Test Plan

1. Checkout this pull request (`git fetch origin pull/ID/head:BRANCHNAME && git checkout BRANCHNAME`, see also: https://help.github.com/en/articles/checking-out-pull-requests-locally).
1. Change the `react-nativ`e field in `react-native/template/package.json` to the absolute path of the react native repo, for example: `"react-native": "file:///Users/username/react-native"`.
1. Run `npx @react-native-community/cli@next init TestApp --template file:///Users/username/react-native` (If you don't have `npx`, run `brew install npm`).
1. Note that a new app built with this template looks like the designs from https://github.com/react-native-community/discussions-and-proposals/issues/122 !

Screen Shot 2019-05-10 at 12 29 25 PM

### DIFF
--- a/Libraries/NewAppScreen/index.js
+++ b/Libraries/NewAppScreen/index.js
@@ -10,88 +10,10 @@
 
 'use strict';
 
-import React, {Fragment} from 'react';
-import {StyleSheet, ScrollView, View, Text, StatusBar} from 'react-native';
-
 import Header from './components/Header';
 import LearnMoreLinks from './components/LearnMoreLinks';
 import Colors from './components/Colors';
 import DebugInstructions from './components/DebugInstructions';
 import ReloadInstructions from './components/ReloadInstructions';
 
-const Section = ({children}) => (
-  <View style={styles.sectionContainer}>{children}</View>
-);
-
-const App = () => {
-  return (
-    <Fragment>
-      <StatusBar barStyle="dark-content" />
-      <ScrollView
-        contentInsetAdjustmentBehavior="automatic"
-        style={styles.scrollView}>
-        <Header />
-        <View style={styles.body}>
-          <Section>
-            <Text style={styles.sectionTitle}>Step One</Text>
-            <Text style={styles.sectionDescription}>
-              Edit <Text style={styles.highlight}>App.js</Text> to change this
-              screen and then come back to see your edits.
-            </Text>
-          </Section>
-
-          <Section>
-            <Text style={styles.sectionTitle}>See Your Changes</Text>
-            <Text style={styles.sectionDescription}>
-              <ReloadInstructions />
-            </Text>
-          </Section>
-
-          <Section>
-            <Text style={styles.sectionTitle}>Debug</Text>
-            <Text style={styles.sectionDescription}>
-              <DebugInstructions />
-            </Text>
-          </Section>
-
-          <Section>
-            <Text style={styles.sectionTitle}>Learn More</Text>
-            <Text style={styles.sectionDescription}>
-              Read the docs on what to do once seen how to work in React Native.
-            </Text>
-          </Section>
-          <LearnMoreLinks />
-        </View>
-      </ScrollView>
-    </Fragment>
-  );
-};
-
-const styles = StyleSheet.create({
-  scrollView: {
-    backgroundColor: Colors.lighter,
-  },
-  body: {
-    backgroundColor: Colors.white,
-  },
-  sectionContainer: {
-    marginTop: 32,
-    paddingHorizontal: 24,
-  },
-  sectionTitle: {
-    fontSize: 24,
-    fontWeight: '600',
-    color: Colors.black,
-  },
-  sectionDescription: {
-    marginTop: 8,
-    fontSize: 18,
-    fontWeight: '400',
-    color: Colors.dark,
-  },
-  highlight: {
-    fontWeight: '700',
-  },
-});
-
-export default App;
+export {Header, LearnMoreLinks, Colors, DebugInstructions, ReloadInstructions};
--- a/Libraries/NewAppScreen/index.js
+++ b/Libraries/NewAppScreen/index.js
@@ -11,52 +11,17 @@
 'use strict';
 
 import React, {Fragment} from 'react';
-import {
-  StyleSheet,
-  ScrollView,
-  View,
-  Text,
-  Platform,
-  StatusBar,
-} from 'react-native';
+import {StyleSheet, ScrollView, View, Text, StatusBar} from 'react-native';
 
 import Header from './components/Header';
 import LearnMoreLinks from './components/LearnMoreLinks';
 import Colors from './components/Colors';
+import DebugInstructions from './components/DebugInstructions';
+import ReloadInstructions from './components/ReloadInstructions';
 
 const Section = ({children}) => (
   <View style={styles.sectionContainer}>{children}</View>
 );
-
-const ReloadInstructions = () => {
-  return Platform.OS === 'ios' ? (
-    <Text style={styles.sectionDescription}>
-      Press <Text style={styles.highlight}>Cmd+R</Text> in the simulator to
-      reload your app's code
-    </Text>
-  ) : (
-    <Text style={styles.sectionDescription}>
-      Double tap <Text style={styles.highlight}>R</Text> on your keyboard to
-      reload your app's code
-    </Text>
-  );
-};
-
-const DebugInstructions = () => {
-  return Platform.OS === 'ios' ? (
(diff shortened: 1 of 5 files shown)


eliperkins:
So this ends up clocking in at ~120 lines of code, which is over our ideal new app template of ~35-45 LoC.

One thought I have is to pull out the reload and debug instructions into the `react-native/Libraries/NewAppScreen` module itself, which should cut out ~25-30 LoC.

My second thought is to turn `Section` and it's subsequent `Text` usages into a more refined component, and tuck that into the `NewAppScreen` module as well. This seems to be where we introduce a lot of LoC. However, this is also the core usage of `<View>` and `<Text>`, so it'd be great to keep that around somehow.

I'm keen on some other ideas on how to reduce the lines of code here too!

orta:
I don't know where the number comes from but 30-40 LOC of the JSX in the render I think is totally fine 👍 

Whether there value in having the two platform-y functions is a good Q - but I'm pretty happy with the rest of the code.

eliperkins:
> I don't know where the number comes from but 30-40 LOC of the JSX in the render I think is totally fine 👍

@orta I think this was an arbitrary number we pulled out that seems like it could be a good amount of code to leave as a first impression to RN. For some context, create-react-app's template clocks in at 26LoC, granted their styling is imported from a CSS module: https://github.com/facebook/create-react-app/blob/e2f43a06a60646f2db00ff0558acb6d2a3355a36/packages/react-scripts/template/src/App.js

I'll move stuff around like @cpojer recommended and we'll see how that feels!

template/App.js
```diff
@@ -6,44 +6,85 @@
...
-    'Double tap R on your keyboard to reload,\n' +
-    'Shake or press menu button for dev menu',
-});
+import Header from 'react-native/Libraries/NewAppScreen/components/Header';
+import LearnMoreLinks from 'react-native/Libraries/NewAppScreen/components/LearnMoreLinks';
+import Colors from 'react-native/Libraries/NewAppScreen/components/Colors';
+import DebugInstructions from 'react-native/Libraries/NewAppScreen/components/DebugInstructions';
+import ReloadInstructions from 'react-native/Libraries/NewAppScreen/components/ReloadInstructions';
+
+const Section = ({children}) => (
```
eliperkins: We added this component in the original PR to add this functionality (see https://github.com/facebook/react-native/pull/24687#discussion_r280485176), but I'm starting to question whether this is a Good Component™ for us here or not.

Given that it doesn't handle anything but applying the style here, I'm not sure if it's worth the extra overhead or not? What thoughts do y'all have?

template/App.js
```diff
@@ -6,44 +6,85 @@
...
+import React, {Fragment} from 'react';
+import {StyleSheet, ScrollView, View, Text, StatusBar} from 'react-native';
 
-const instructions = Platform.select({
-  ios: 'Press Cmd+R to reload,\n' + 'Cmd+D or shake for dev menu',
-  android:
-    'Double tap R on your keyboard to reload,\n' +
-    'Shake or press menu button for dev menu',
-});
+import Header from 'react-native/Libraries/NewAppScreen/components/Header';
```
eliperkins: I'd like to pull these up into one destructed import like @orta mentioned, but I noticed that all the modules that RN exposes here are all default exports like we have here already. Is this worth doing as a part of this PR or is this something worth discussing elsewhere?

Libraries/NewAppScreen/index.js
```diff
@@ -11,53 +11,18 @@
 'use strict';
```
eliperkins: This file is now a duplicate of `template/App.js`. Is it worth maintaining both files? Their imports are different, given where each module lives, but it will create extra overhead in maintaining these.

template/App.js
cpojer: Kill it.

template/App.js
cpojer: I think I'd prefer importing this from "NewAppScreen" instead of reaching deep inside the folder.

Libraries/NewAppScreen/index.js
cpojer: No we should kill it and replace it with a module that just exports everything that is needed by App.js. The template should have only a single import that destructures into everything that's needed there.

eliperkins:
> Let's tuck ReloadInstructions and DebugInstructions into the NewAppScreen 

Done in b266a5c

> and also put those sections into the LeanMoreLinks component (possibly named differently?).

Hmmmm, I'm not sure how this would work, without losing all the `<Text>`, in the app template. Currently (as of this PR), the structure is as follows:

```jsx
<Fragment>
  <StatusBar/>
  <ScrollView>
    <Header />
    <View>{/* Sections with title & descriptions */}
      <Section>
        <Text />
        <Text />
      </Section>
      <Section>
        <Text />
        <Text>
          <ReloadInstructions />
        </Text>
      </Section>
      <Section>
        <Text />
        <Text>
          <DebugInstructions />
        </Text>
      </Section>
      <Section>
        <Text />
        <Text />
      </Section>
      <LearnMoreLinks />
    </View>
  </ScrollView>
</Fragment>;
```

In order to move `<ReloadInstructions>` and `<DebugInstructions>` into `<LearnMoreLinks>`, we'd likely have to move everything in the `Sections with title & descriptions` section into somewhere, leaving us with zero `<Text>` components.

@cpojer can you explain your thought process on how to move Reload and Debug into LearnMore for me?

Libraries/NewAppScreen/index.js
eliperkins: Should the usage of this in RNTester then become a rendering of each component from this module?

https://github.com/facebook/react-native/blob/2cd6c0c82e0d01201d97d887c4f79a019dbdaeea/RNTester/js/NewAppScreenExample.js#L14

cpojer:
Right now it looks like this:

```jsx
          <Section>
            <Text style={styles.sectionTitle}>Step One</Text>
            <Text style={styles.sectionDescription}>
              Edit <Text style={styles.highlight}>App.js</Text> to change this
              screen and then come back to see your edits.
            </Text>
          </Section>
          <Section>
            <Text style={styles.sectionTitle}>See Your Changes</Text>
            <Text style={styles.sectionDescription}>
              <ReloadInstructions />
            </Text>
          </Section>
          <Section>
            <Text style={styles.sectionTitle}>Debug</Text>
            <Text style={styles.sectionDescription}>
              <DebugInstructions />
            </Text>
          </Section>
          <Section>
            <Text style={styles.sectionTitle}>Learn More</Text>
            <Text style={styles.sectionDescription}>
              Read the docs on what to do once seen how to work in React Native.
            </Text>
          </Section>
          <LearnMoreLinks />
```

I think a bunch of this is superfluous. We only need a single `<Text>` in the template. Here is my suggestion:

```jsx
          <Section>
            <Text style={styles.sectionTitle}>Step One</Text>
            <Text style={styles.sectionDescription}>
              Edit <Text style={styles.highlight}>App.js</Text> to change this
              screen and then come back to see your edits.
            </Text>
          </Section>
          <LearnMoreLinks /> {/* Everything removed compared to above is inside this component know */}
```

It still uses the same type of components as before at least once and is much much shorter in App.js.

Libraries/NewAppScreen/index.js
cpojer: Yeah sounds good.

template/App.js
```diff
@@ -6,44 +6,83 @@
...
-    'Double tap R on your keyboard to reload,\n' +
-    'Shake or press menu button for dev menu',
-});
+import {
+  Header,
+  LearnMoreLinks,
+  Colors,
+  DebugInstructions,
+  ReloadInstructions,
+} from 'react-native/Libraries/NewAppScreen';
```
thymikee: this definitely needs an annotation that importing stuff from `react-native/Libraries` is only for demonstration purposes and one should never do it in a real app because it may break.

cpojer (approved):
Fantastic! Thank you so much for the hard work on this. I'm going to land this version but we can update the comment in a follow-up.

@thymikee instead of a comment, what do you think of moving the index file of NewAppScript to `react-native/NewAppScreen`? That way it should be clear not to reach into libraries as it special cases this file only.

thymikee:
@cpojer makes sense

PR #24811, opened 2019-05-10
[jest] Delete hasteImpl, providesModuleNodeModules, and modulePathNameMapper

ide:
## Summary

**Depends on https://github.com/facebook/react-native/pull/25100**

This commit depends on having migrated all of RN away from Haste. With 100% standard path-based requires, the key foundation is set and we no longer need `hasteImpl` and related settings in the Jest configuration.

This commit deletes the `hasteImpl` file and setting as well as `providesModuleNodeModules` and `modulePathNameMapper`, removing most of the dependency graph overriding performed by Jest.

## Changelog

[General] [Changed] - Delete hasteImpl, providesModuleNodeModules, and modulePathNameMapper from Jest config

## Test Plan

Run Jest tests after rewriting all imports to use paths instead of Haste names.


cpojer:
Had to bring back a stub for `hasteImpl` because the CLI requires it, see https://github.com/react-native-community/cli/pull/417

PR #24814, opened 2019-05-11
Add inline view examples to RNTester

rigdern:
## Summary

Now that inline views are supported on iOS and Android, we can add some examples to RNTester. I brought back examples from https://github.com/facebook/react-native/commit/03663491c6e68409f73d5c9bbc1a2e3c02ee0966.

I also added some new inline view examples in TextInlineView.js. Note that some examples are only supported on iOS because they rely on the inline view being able to size to content. Android's implementation requires that a width and height be specified on the inline view.

Here are the known bugs illustrated by these examples:
  - ClippedByText
    - Expected: The image/view wraps to the next line (because it is too wide) and gets clipped vertically (because it is too tall).
    - iOS bug: The image/view does not get wrapped to the next line
    - Android bug: The view gets wrapped to the next line but doesn't get clipped vertically. The image appears to be positioned too low.
  - ChangeImageSize/ChangeViewSize:
    - Expected: The "Change Image/View Width" button causes the image/view to toggle between a width of 50 and 100.
    - iOS bug: First update works. Subsequent updates don't get rendered.
    - Android bug: No updates get rendered.
  - ChangeInnerViewSize:
    - Expected: The "Change Pink View Width" button causes the pink inner view to toggle between a width of 50 and 100.
    - iOS bug: First update works but second update **CRASHES** the app.
    - Android bug: No updates get rendered.

## Changelog

[Internal] [Added] - Added inline view examples to RNTester

## Test Plan

Verified the new examples in RNTester on iOS and Android.

### iOS: New Examples

image

image

image

image

### Android: New Examples

image

image

Adam Comella
Microsoft Corp.

### DIFF
--- a/RNTester/js/Shared/TextInlineView.js
+++ b/RNTester/js/Shared/TextInlineView.js
@@ -10,7 +10,7 @@
 
 'use strict';
 
-const React = require('React');
+const React = require('react');
 const {Image, Text, TouchableHighlight, View} = require('react-native');
 
 function Basic() {
--- a/RNTester/js/Shared/TextInlineView.js
+++ b/RNTester/js/Shared/TextInlineView.js
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const React = require('React');
+const {Image, Text, TouchableHighlight, View} = require('react-native');
+
+function Basic() {
+  return (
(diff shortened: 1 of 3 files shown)


RNTester/js/Shared/TextInlineView.js
```diff
@@ -0,0 +1,205 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const React = require('React');
```
ide:
```suggestion
const React = require('react');
```
As part of a migration from Haste to standard imports, `React` (the Haste name) has been replaced with `react`.

Issue #24688, opened 2019-05-02
ScrollView with centerContent={true} renders non-deterministically on iPhone X simulator


m-pol:
## 🐛 Bug Report
When a ScrollView has property centerContent={true}, a contained item can sometimes appear not centered. I've only verified this on an iPhone X v12.2 simulator, but I think it happens on devices too. The orange square on the left sometimes is centered properly, and sometimes is centered on the right edge of the green rectangle. Refresh the simulator multiple times and it non-deterministically changes.

## To Reproduce
  - Start a fresh project with react-native init. I have react-native-cli: 2.0.1.
  - Replace your App.js code with the App.js code below
  - Run on an iPhone X simulator
  - Refresh (Command + R) a few times to see non-deterministic rendering behavior

## Expected Behavior
The render should not be non-deterministic. It should always be centered. And it should not play differently on iPhone X simulator vs. other simulators.

See video where I refresh the code and see the positioning sometimes changes:

![bug](https://user-images.githubusercontent.com/5598691/57085146-9ffe9a00-6cc9-11e9-8ef9-0dda2e60754a.gif)

## Code Example
Paste this into a fresh "react-native init" project's App.js:

```
import React, {Component} from 'react';
import {ScrollView, View} from 'react-native';

export default class App extends Component {
    render() {
        return (
            <View
                style={{
                    flex: 1,
                    backgroundColor: "blue",
                    justifyContent: "center",
                    alignItems: "center",
                    flexDirection: "row"
                }}
            >
                <View style={{width: 100,
                    backgroundColor: "red",
                    height: 400}}>
                    <ScrollView
                        contentContainerStyle={{justifyContent: "center", alignItems: "center"}}
                        style={{backgroundColor: "green"}}
                        centerContent={true}
                    >
                        <View style={{backgroundColor: "orange", width: 50, height: 50}} />
                    </ScrollView>
                </View>
                <View style={{width: 100,
                    backgroundColor: "indigo",
                    height: 400}}>
                    <ScrollView
                        contentContainerStyle={{justifyContent: "center", alignItems: "center"}}
                        style={{backgroundColor: "purple"}}
                        centerContent={true}
                    >
                        <View style={{backgroundColor: "violet", width: 50, height: 50}} />
                    </ScrollView>
                </View>
            </View>
        );
    }
}
```

## Environment
    React Native Environment Info:
    System:
      OS: macOS 10.14.4
[...]

PR #24817, opened 2019-05-11
[iOS] Fixes ScrollView centerContent not work in some cases

zhongwuzw:
## Summary

The bug description can see https://github.com/facebook/react-native/issues/24688, we add `contentView` size check before center the content, because in some cases, the contentView's size not yet be calculated, in those cases, we don't adjust the `contentOffset`.

cc. @cpojer .

## Changelog

[iOS] [Fixed] - Fixes ScrollView centerContent not work in some cases

## Test Plan

`centerContent` can works on iPhoneX when using the code like below:

```
import {ScrollView, View} from 'react-native';

    render() {
        return (
            <View
                style={{
                    flex: 1,
                    backgroundColor: "blue",
                    justifyContent: "center",
                    alignItems: "center",
                    flexDirection: "row"
                }}
            >
                <View style={{width: 100,
                    backgroundColor: "red",
                    height: 400}}>
                    <ScrollView
                        contentContainerStyle={{justifyContent: "center", alignItems: "center"}}
                        style={{backgroundColor: "green"}}
                        centerContent={true}
                    >
                        <View style={{backgroundColor: "orange", width: 50, height: 50}} />
                    </ScrollView>
                </View>
                <View style={{width: 100,
                    backgroundColor: "indigo",
                    height: 400}}>
                    <ScrollView
                        contentContainerStyle={{justifyContent: "center", alignItems: "center"}}
                        style={{backgroundColor: "purple"}}
                        centerContent={true}
                    >
                        <View style={{backgroundColor: "violet", width: 50, height: 50}} />
                    </ScrollView>
                </View>
            </View>
        );
    }
}
```

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -298,7 +298,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
 - (void)setContentOffset:(CGPoint)contentOffset
 {
   UIView *contentView = [self contentView];
-  if (contentView && _centerContent) {
+  if (contentView && _centerContent && !CGSizeEqualToSize(contentView.frame.size, CGSizeZero)) {
     CGSize subviewSize = contentView.frame.size;
     CGSize scrollViewSize = self.bounds.size;
     if (subviewSize.width <= scrollViewSize.width) {


sahrens:
Thanks!

Issue #24820, opened 2019-05-12
[iOS] Crash when tap on performance monitor


zhigang1992:
<img width="510" alt="Screen Shot 2019-05-12 at 4 32 13 PM" src="https://user-images.githubusercontent.com/1057756/57579782-9f3ce380-74d3-11e9-9ca7-d5b2a57f2929.png">

React Native version:

```
info 
  React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
      Memory: 2.06 GB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.15.1 - /var/folders/0b/k3zlcx0s3936lr99z43_6pxw0000gn/T/fnm-shell-9885663/bin/node
      Yarn: 1.15.2 - /usr/local/bin/yarn
      npm: 6.4.1 - /var/folders/0b/k3zlcx0s3936lr99z43_6pxw0000gn/T/fnm-shell-9885663/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 27, 28
        Build Tools: 27.0.3, 28.0.2, 28.0.3
        System Images: android-28 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.5429.30.34.5452501
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.8 => 0.59.8 
```

## Steps To Reproduce

1. Launch react native app on iOS simulator
2. Command + D to activate the performance monitor
3. Tap on the performance monitor

It should not crash.

<img width="1440" alt="Screen Shot 2019-05-12 at 3 48 33 PM" src="https://user-images.githubusercontent.com/1057756/57579409-d4dece00-74cd-11e9-9cf4-885f22002a63.png">

zhongwuzw:
Fixed in https://github.com/facebook/react-native/pull/24821.

PR #24821, opened 2019-05-12
[iOS][Bug] - fix crash on performance logger

zhigang1992:
## Summary

Fix Issue https://github.com/facebook/react-native/issues/24820

It's caused by `_labelsForTags` and `RCTPLTag` being out of sync, the crash might only be one of the issues that this bug was causing.

## Changelog

[iOS] [Fixed] - fix crash on performance logger

## Test Plan

- Launch iOS simulator
- Enable performance monitor
- Tap on performance monitor

It should not crash


React/Base/RCTPerformanceLogger.h
```diff
@@ -7,6 +7,7 @@
 
 #import <Foundation/Foundation.h>
 
+// Keep this in sync with _labelsForTags
```
zhongwuzw: I think we'd better to remove these missing labels in `_labelForTags`? because we don't use these enums at all.

PeteTheHeat: Yup, +1

cpojer:
@zhigang1992 could you make the change that was suggested by @zhongwuzw?

zhigang1992:
All the unused flags are now removed except `RCTPLJSCWrapperOpenLibrary`. Which was added in https://github.com/facebook/react-native/commit/abe0b349bcdb59155445889d9615c13294a37c91 Should I remove that as well?

zhongwuzw:
> All the unused flags are now removed except `RCTPLJSCWrapperOpenLibrary`. Which was added in [abe0b34](https://github.com/facebook/react-native/commit/abe0b349bcdb59155445889d9615c13294a37c91) Should I remove that as well?

We can keep it, don't break `RCTPLTag` enum.

PR #24829, opened 2019-05-13
[iOS] Change ReactNetworkForceWifiOnly from String to Boolean

zhongwuzw:
## Summary

We provided `ReactNetworkForceWifiOnly` in https://github.com/facebook/react-native/pull/24242, but it's a string type, actually the value only YES or NO, so let's change it to Boolean type, we can benefit from:
1. Users don't need to type string `YES`, just select bool YES or NO directly by click the button:
image

2. Don't need to do string compare.

3. More looks what it should looks, Boolean is the Boolean. :)

cc. @cpojer @ericlewis .

## Changelog

[iOS] [Changed] - Change ReactNetworkForceWifiOnly from String to Boolean

## Test Plan

CI passes.

### DIFF
--- a/Libraries/Network/RCTHTTPRequestHandler.mm
+++ b/Libraries/Network/RCTHTTPRequestHandler.mm
@@ -62,20 +62,19 @@ RCT_EXPORT_MODULE()
   if (!_session && [self isValid]) {
     // You can override default NSURLSession instance property allowsCellularAccess (default value YES)
     //  by providing the following key to your RN project (edit ios/project/Info.plist file in Xcode):
-    // <key>ReactNetworkForceWifiOnly</key>    <string>YES</string>
+    // <key>ReactNetworkForceWifiOnly</key>    <Boolean>YES</Boolean>
     // This will set allowsCellularAccess to NO and force Wifi only for all network calls on iOS
-    // If you do not want to override default behavior, do nothing or set key with value "NO"
+    // If you do not want to override default behavior, do nothing or set key with value NO
     NSDictionary *infoDictionary = [[NSBundle mainBundle] infoDictionary];
-    NSString *useWifiOnly = [infoDictionary objectForKey:@"ReactNetworkForceWifiOnly"];
+    NSNumber *useWifiOnly = [infoDictionary objectForKey:@"ReactNetworkForceWifiOnly"];
     
     NSOperationQueue *callbackQueue = [NSOperationQueue new];
     callbackQueue.maxConcurrentOperationCount = 1;
     callbackQueue.underlyingQueue = [[_bridge networking] methodQueue];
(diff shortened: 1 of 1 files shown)


cpojer:
cc @bondparkerbond

ericlewis:
if @zhongwuzw says it works it probably works ;)

bondparkerbond:
I initially tried that but kept getting errors. (Perhaps because I was trying to manually add it into the Info.plist file instead of creating it inside Xcode?) I saw references to user defined Bool values not being allowed in this [Stack Overflow Response](https://stackoverflow.com/questions/1832157/setting-a-boolean-property-in-info-plist-from-a-user-defined-setting). 
Specifically, the 3rd answer claimed "plist files containing booleans within tags are not valid any more", but I could find no other documentation specifically saying it was/wasn't allowed. Since I am relatively new to Obj C I decided to go with the way that I had tested and knew would work, but if @zhongwuzw was able to test it and it worked I see no reason not to merge this in. We may want to specify in the file to make sure to set this in Xcode before merging though.

bondparkerbond:
One thing I do want to double check because I'm not sure on the expected behavior: since bool values are apparently just 0/1, but are shown as YES/NO in Xcode and <true/>/<false/> when you examine the actual Info.plist values, are the following equivalent?:

configuration.allowsCellularAccess = !1 (or !YES)
configuration.allowsCellularAccess = 0
configuration.allowsCellularAccess = NO;

Maybe it would be better to do something like:
```
if(plist value exists AND is true/YES/1?) {
  configuration.allowsCellularAccess = NO;
}
```
(ie: literally writing NO like we are writing YES for setHTTPShouldSetCookies:YES on the next line)
This would be more explicit that we are setting a value to NO as we don't need to set = YES as that is the default if nothing is set.

zhongwuzw:
> One thing I do want to double check because I'm not sure on the expected behavior: since bool values are apparently just 0/1, but are shown as YES/NO in Xcode and `<true/>`/`<false/>` when you examine the actual Info.plist values, are the following equivalent?:
> 
> configuration.allowsCellularAccess = !1 (or !YES)
> configuration.allowsCellularAccess = 0
> configuration.allowsCellularAccess = NO;
> 
> Maybe it would be better to do something like:
> 
> ```
> if(plist value exists AND is true/YES/1?) {
>   configuration.allowsCellularAccess = NO;
> }
> ```
> 


The three cases are the same. YES/NO, true/false, 1/0 have no practical difference.

> (ie: literally writing NO like we are writing YES for setHTTPShouldSetCookies:YES on the next line)
> This would be more explicit that we are setting a value to NO as we don't need to set = YES as that is the default if nothing is set.

Emm, actually, based on standard OC coding style, we don't write `if (setHTTPShouldSetCookies == YES)`, we should write `if (setHTTPShouldSetCookies)`.

PR #24832, opened 2019-05-13
VirtualizedList.RenderItem throws when using function component with hooks

adkenyon:
## Summary

`<VirtualizedList />` will throw an error if the `renderItem` Prop component uses hooks. Function components without hooks and class components work without issue.

Super contrived Example
```{js}
function FlatListItem({ item }) {
  React.useEffect(() => console.log(item),[])

 return (<Text>{item}</Text>);
}

<FlatList data={[1, 2, 3]} renderItem={FlatListItem} />
```

Example Error:
```
Invariant Violation: Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)

This error is located at:
    in CellRenderer (at VirtualizedList.js:688)
    in RCTScrollContentView (at ScrollView.js:976)
    in RCTScrollView (at ScrollView.js:1115)
    in ScrollView (at VirtualizedList.js:1081)
    in VirtualizedList (at FlatList.js:632)
    in FlatList (at WithoutScrollbars.js:21)
    ...
```

## Changelog

[General] [Added] - VirtualizedList ListItemComponent. An alternative to renderItem that accepts function components with hooks.
[General][Added] - FlatList ListItemComponent. An alternative to renderItem that accepts function components with hooks.
[General][Added] - VirtualizedList and FlatList tests and updated RNTester example

## Test Plan

Updated JS tests for VirtualizedList and tested on local build.

### DIFF
--- a/Libraries/Lists/FlatList.js
+++ b/Libraries/Lists/FlatList.js
@@ -619,50 +619,71 @@ class FlatList<ItemT> extends React.PureComponent<Props<ItemT>, void> {
     };
   }
 
-  _renderItem = () => {
-    const {
-      renderItem,
-      ListItemComponent,
-      numColumns,
-      columnWrapperStyle,
-    } = this.props;
+  _renderItem = (info: RenderItemProps<ItemT>) => {
+    const {renderItem, numColumns, columnWrapperStyle} = this.props;
 
-    const renderer = (props: RenderItemProps<ItemT>) =>
-      ListItemComponent
-        ? React.createElement(ListItemComponent, props)
-        : renderItem
-        ? renderItem(props)
-        : () => undefined;
+    if (!renderItem) return null;
 
-    return (info: RenderItemProps<ItemT>) => {
-      if (numColumns > 1) {
-        const {item, index} = info;
-        invariant(
-          Array.isArray(item),
-          'Expected array of items with numColumns > 1',
-        );
-        return (
-          <View
-            style={StyleSheet.compose(
-              styles.row,
-              columnWrapperStyle,
-            )}>
-            {item.map((it, kk) => {
-              const element = renderer({
-                item: it,
-                index: index * numColumns + kk,
-                separators: info.separators,
-              });
-              return element != null ? (
-                <React.Fragment key={kk}>{element}</React.Fragment>
-              ) : null;
-            })}
-          </View>
-        );
-      } else {
-        return renderer(info);
-      }
-    };
+    if (numColumns > 1) {
+      const {item, index} = info;
+      invariant(
+        Array.isArray(item),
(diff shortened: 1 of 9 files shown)


cpojer:
I spent some time thinking about this and I'm worried about the performance implications of this breaking change because of the added wrapping for each cell. I think `renderItem` is also a misleading prop name.

How about this:
* Add a new prop called `item` which takes a React component
* If `item` is specified, use it with React.createElement
* If `item` isn't specified, fall back to `renderItem` (this ensures the default behavior stays the same)
* If both `item` and `renderItem` are specified, we should `console.warn` and explain that one takes precedence over the other.

Then in the future we can remove `renderItem` completely.

adkenyon:
@cpojer Implemented your suggestions.
- Used `ListItemComponent` rather than `Item`. I think it follows the existing naming convention.
- Minor flow type cleanup
- After adding `console.warn` for the condition that both `ListItemComponent` and `renderItem` are present I noticed that FlatList would constantly throw warnings as FlatList always returns an implementation of renderItem (to handle multiple columns). I was able to [fix](https://github.com/facebook/react-native/pull/24832/files#diff-85e8f96a713d23be94fbd08687332bffR702) this for FlatList, but VirtualizedSectionList and SectionList have the same issue. I'm thinking that I'll follow up with the ListItemComponent implementation for VirtualizedSectionList and SectionList on another PR.

Libraries/Lists/FlatList.js
```diff
@@ -629,16 +655,54 @@ class FlatList<ItemT> extends React.PureComponent<Props<ItemT>, void> {
...
   };
 
+  _listItemComponent = (info: RenderItemProps<ItemT>) => {
+    const {ListItemComponent, numColumns, columnWrapperStyle} = this.props;
+
+    if (!ListItemComponent) {
+      return null;
+    }
+
+    if (numColumns > 1) {
```
sahrens: Can you factor this shared code out into a separate function rather than copy-pasta?

sahrens: This is mostly copy-pasta - can you factor into a shared function?

adkenyon: @sahrens My original approach was to use one function. I opted to use two functions as it was easier to read and if the renderItem prop was ever removed it'd be easy to remove _renderItem from FlatList. I'm open rolling back to a single function. Having two functions with very similar functionality is bound to get out of sync down the road.

adkenyon: @sahrens I moved back to a single function

cpojer:
I applied the change internally and I'm trying to land it :) Thanks for going through all the iterations with us @adkenyon!

PR #24840, opened 2019-05-14
[iOS] Resubmit #24714 and fixes kvo crash

zhongwuzw:
## Summary

Resubmit #24714. 
From the https://github.com/facebook/react-native/pull/24714#issuecomment-491962489, seems it would crash sometimes. @JoshuaGross Hi :) I added a method to let `UITextField` call the cleanup explicitly. Please check again, thanks! 

## Changelog

[iOS] [Fixed] - Fixes selection of single line text input

## Test Plan

Create a single line textInput, then do some selection, `onSelectionChange` can be called.

### DIFF
--- a/Libraries/Text/TextInput/Singleline/RCTUITextField.m
+++ b/Libraries/Text/TextInput/Singleline/RCTUITextField.m
@@ -176,6 +176,16 @@
 
 #pragma mark - Overrides
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-implementations"
+// Overrides selectedTextRange setter to get notify when selectedTextRange changed.
+- (void)setSelectedTextRange:(UITextRange *)selectedTextRange
+{
+  [super setSelectedTextRange:selectedTextRange];
+  [_textInputDelegateAdapter selectedTextRangeWasSet];
+}
+#pragma clang diagnostic pop
+
 - (void)setSelectedTextRange:(UITextRange *)selectedTextRange notifyDelegate:(BOOL)notifyDelegate
 {
   if (!notifyDelegate) {


JoshuaGross:
Hi @zhongwuzw, thanks for working on this. Do you think it's possible to do this without KVO, or without KVO on unowned objects? KVO has a lot of pitfalls and this use-case is tricky. Could we, for example, have the `RCTUITextField` observe itself and call delegate methods when relevant?

zhongwuzw:
@JoshuaGross Agree! 👍 Here we go! no KVO and override `setSelectedTextRange` instead.

JoshuaGross:
Can you squash your commits?

Libraries/Text/TextInput/Singleline/RCTUITextField.m
```diff
@@ -176,6 +176,16 @@ - (CGRect)editingRectForBounds:(CGRect)bounds
 
 #pragma mark - Overrides
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-implementations"
```
JoshuaGross: Hmm, why do we need these pragmas? This seems like a sign that this might not be the right solution.

Libraries/Text/TextInput/Singleline/RCTUITextField.m
```diff
@@ -176,6 +176,16 @@ - (CGRect)editingRectForBounds:(CGRect)bounds
...
+// Overrides selectedTextRange setter to get notify when selectedTextRange changed
+- (void)setSelectedTextRange:(UITextRange *)selectedTextRange
+{
+  [super setSelectedTextRange:selectedTextRange];
+  [_textInputDelegateAdapter selectedTextRangeWasSet];
```
JoshuaGross: Should we try to call delegates in the same order? In the method below this one, we call the delegate method and then we call super; here, it's reversed.

JoshuaGross:
Thanks again for doing this! I have some specific concerns but I'm excited to get this in once we know we're doing everything the right way :)

Libraries/Text/TextInput/Singleline/RCTUITextField.m
zhongwuzw: Yeah, we need it to suppress warning, warning because we mark `setSelectedTextRange ` as unavailable, please see : https://github.com/facebook/react-native/blob/7a7eb11965a4126987fd0f4bfcd13dd872452e76/Libraries/Text/TextInput/RCTBackedTextInputViewProtocol.h#L30-L36

We disallow access `setSelectedTextRange ` directly, but Apple framework would call it, and this is how KVO works.

Libraries/Text/TextInput/Singleline/RCTUITextField.m
zhongwuzw: 🤔 Emm, may not have much difference IMO, but I can consolidate them.

JoshuaGross:
Hi @zhongwuzw, I'm testing this PR with the following example and seeing that selection changes are happening, but not really getting useful information about the selection changes that I can tell. Is this what you would expect? Is there a better test-case for this?

```
/**
 * Copyright 2004-present Facebook. All Rights Reserved.
 *
 * @flow strict-local
 * @format
 */

'use strict';
import {StyleSheet, Text, View} from 'react-native';
import { TextInput } from 'react-native';

class Playground extends Component {
  constructor(props) {
    super(props);
    this.state = { text: 'Useless Placeholder', selectionChange: null };
  }

  render() {
    return (
        <TextInput
          style={{height: 40, borderColor: 'gray', borderWidth: 1}}
          onChangeText={(text) => this.setState({...this.state, text}) && console.log('onChangeText: ', text)}
          onSelectionChange={(...args) => this.setState({...this.state, selectionChange: args}) && console.log('onSelectionChange: ', args)}
          value={this.state.text}
        />
        <Text>Text: {this.state.text}</Text>
        <Text>Selection: {JSON.stringify((this.state.selectionChange))}</Text>
      </View>
    );
  }
}

  container: {
    backgroundColor: 'white',
    padding: 10,
    paddingTop: 30,
  },
});

export default Playground;
```

Screen Shot 2019-05-20 at 11 33 53 AM

zhongwuzw:
@JoshuaGross I amended the code based yours, please see:

```class RNTesterApp extends React.Component<Props, RNTesterNavigationState> {
  constructor(props) {
    super(props);
    this.state = { text: 'Useless Placeholder', selectionChange: null };
  }

  onSelectionChange({nativeEvent: {selection}}) {
    this.setState({selection});
  }

  render() {
    return (
        <TextInput
          style={{height: 40, borderColor: 'gray', borderWidth: 1}}
          onChangeText={(text) => this.setState({...this.state, text}) && console.log('onChangeText: ', text)}
          onSelectionChange={this.onSelectionChange.bind(this)}
          value={this.state.text}
        />
        <Text>Text: {this.state.text}</Text>
        <Text>selection = {JSON.stringify(this.state.selection)}</Text>
      </View>
    );
  }
}
```

image

Libraries/Text/TextInput/Singleline/RCTUITextField.m
zhongwuzw: @JoshuaGross I reverted to this change, we need to call super first, otherwise, `onSelectionChange` would be called with last args, not latest args.

MrXyfir:
Would be nice to update the docs that `multiline={true}` is no longer required for `onSelectionChange`.

PR #24848, opened 2019-05-14
[Android] fix: set js delta bundle default value to false

wojteg1337:
## Summary

When running Android app for the first time, the packager is requesting delta bundles from metro instead of a bundle (in dev settings delta bundles are disabled by default and marked as experimental). UI of dev settings is not consistent with the current state, to turn off delta bundles you have to enable them and then disable.

## Changelog

[Android] [Fixed] - Disable delta bundles on the first app run

## Test Plan

Run android app and packager should use bundle instead of delta bundle.


cpojer:
Do you mind signing the CLA via https://code.facebook.com/cla so I can land this? It's only gonna take you a minute!

wojteg1337:
@cpojer, done

PR #24849, opened 2019-05-14
TM iOS: Remove retainArguments && do retain by us explicitly

zhongwuzw:
## Summary

This is a TODO, we met crash if we don't call `retainArguments` when return type is like `NSDictionary`, the reason is `getReturnValue` don't retain the return value, so we need to using `__bridge` to transfer ownership to OC type.
Also add `resolveBlock` and `rejectBlock` to `retainedObjectsForInvocation`.

cc. @cpojer 

## Changelog

[iOS] [Fixed] - Remove retainArguments && do retain by us explicitly

## Test Plan

TM example can run successfully.


zhongwuzw:
@matthargett Yeah, I'll see how I can do it. :)

RSNara:
This is awesome! 😁

zhongwuzw:
TM is still in progress, so I think we can't write unit test which integrated into CI? @RSNara 
I may need your help about plans how and where we can write tests? 😂

PR #24859, opened 2019-05-15
Remove Chakra specific conditional compilation hack from JSIExecutor

mganandraj:
## Summary
We introduced a hack in JSIExecutor to conditionally compile the creation of NativeModuleProxy as a JSI HostObject. This was needed in the days where the Chakra based JSI runtimes didn't support JSI HostObjects. We are now removing the conditional compilation as the Chakra based JSI runtimes do support HostObjects.

## Changelog
We introduced a hack in JSIExecutor to conditionally compile the creation of NativeModuleProxy as a JSI HostObject. This was needed in the days where the Chakra based JSI runtimes didn't support JSI HostObjects. We are now removing the conditional compilation as the Chakra based JSI runtimes do support HostObjects.

[CATEGORY] [TYPE] - Message

## Test Plan
Tested that the native module dispatch is happening with ChakraJSIRuntime without pushing the native modules config upfront.


mganandraj:
Sorry, this PR was made by mistake to the fb repo. This was meant for Microsoft fork.

PR #24867, opened 2019-05-15
Add Fabric/Turbo Modules flags to iOS autolinking script

ericlewis:
## Summary
This makes it easier to enable fabric / turbo modules with the new fancy autolinking script.

## Changelog

[iOS] [Added] - Add Fabric/Turbo Module flags to iOS autolinking script

## Test Plan

RNTester's TurboModules still work


orta:
Yep, makes sense 👍

PR #24873, opened 2019-05-15
[Fabric] use state in paragraph component

ericlewis:
## Summary
Updates the paragraph component to use State instead of Local Data, part of the path to a Fabric TextInput 💯 

## Changelog

[General] [Changed] - Fabric: Use State instead of Local Data for Paragraph

## Test Plan

Run RNTester with Fabric enabled, view that text still renders.


ReactCommon/fabric/components/text/paragraph/ParagraphState.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace facebook {
+namespace react {
+
+AttributedString ParagraphState::getAttributedString() const {
+  return attributedString_;
+}
+
+void ParagraphState::setAttributedString(const AttributedString &attributedString) const {
+  assert(!attributedStringIsInitialized_);
```
ericlewis: this assert gets called after pushing a new view in RNTester, so maybe I am not using this correctly?

React/Fabric/Mounting/ComponentViews/Text/RCTParagraphComponentView.mm
```diff
@@ -63,32 +63,32 @@ - (void)updateProps:(SharedProps)props oldProps:(SharedProps)oldProps
   _paragraphAttributes = paragraphProps->paragraphAttributes;
 }
 
-- (void)updateLocalData:(SharedLocalData)localData oldLocalData:(SharedLocalData)oldLocalData
+- (void)updateState:(State::Shared)state oldState:(State::Shared)oldState
 {
-  _paragraphLocalData = std::static_pointer_cast<const ParagraphLocalData>(localData);
-  assert(_paragraphLocalData);
+  _state = std::static_pointer_cast<ParagraphShadowNode::ConcreteState const>(state);
```
shergin: Special thanks for using East Const.

ReactCommon/fabric/components/text/paragraph/ParagraphShadowNode.cpp
```diff
@@ -38,21 +38,18 @@ void ParagraphShadowNode::setMeasureCache(
...
-void ParagraphShadowNode::updateLocalDataIfNeeded() {
+void ParagraphShadowNode::updateStateIfNeeded() {
   ensureUnsealed();
 
   auto attributedString = getAttributedString();
-  auto currentLocalData =
-      std::static_pointer_cast<const ParagraphLocalData>(getLocalData());
-  if (currentLocalData &&
-      currentLocalData->getAttributedString() == attributedString) {
+  auto state = getStateData();
```
shergin: It must be `auto const &state = getStateData();` (I added `&`), using a reference here will prevent expensive copying.

ReactCommon/fabric/components/text/paragraph/ParagraphShadowNode.cpp
```diff
@@ -38,21 +38,18 @@ void ParagraphShadowNode::setMeasureCache(
...
+  if (state.getAttributedString() == attributedString) {
     return;
   }
 
-  auto localData = std::make_shared<ParagraphLocalData>();
-  localData->setAttributedString(std::move(attributedString));
-  localData->setTextLayoutManager(textLayoutManager_);
-  setLocalData(localData);
+  state.setAttributedString(std::move(attributedString));
```
shergin: Following the previous comment, here we have to build a new state data object (only after we decided that we have to).

ReactCommon/fabric/components/text/paragraph/ParagraphState.h
```diff
@@ -8,50 +8,39 @@
...
-#if RN_DEBUG_STRING_CONVERTIBLE
-  std::string getDebugName() const override;
-  SharedDebugStringConvertibleList getDebugProps() const override;
+  folly::dynamic getDynamic() const;
 #endif
 
  private:
-  AttributedString attributedString_;
-  SharedTextLayoutManager textLayoutManager_;
+  mutable AttributedString attributedString_;
```
shergin: Doing that I mentioned before (constructing fresh Data Object) will allow to avoid all that complexity.
Using `mutable` is quite dangerous, so we should try to avoid that.

React/Fabric/Mounting/ComponentViews/Text/RCTParagraphComponentView.mm
```diff
@@ -8,7 +8,7 @@
 #import "RCTParagraphComponentView.h"
 
 #import <react/components/text/ParagraphComponentDescriptor.h>
-#import <react/components/text/ParagraphLocalData.h>
+#import <react/components/text/ParagraphState.h>
```
shergin: Those are not alphabetically sorted now.

If you can run clang format on that with config from the repo, that would be great. Otherwise I can do it for you.

ReactCommon/fabric/components/text/paragraph/ParagraphShadowNode.cpp
```diff
@@ -38,21 +38,20 @@ void ParagraphShadowNode::setMeasureCache(
...
+  auto const &state = getStateData();
+  if (state.attributedString == attributedString) {
     return;
   }
 
-  auto localData = std::make_shared<ParagraphLocalData>();
-  localData->setAttributedString(std::move(attributedString));
-  localData->setTextLayoutManager(textLayoutManager_);
-  setLocalData(localData);
+  auto newStateData = ParagraphState();
```
shergin: `setStateData({attributedString, textLayoutManager_});`

shergin: Also, I admit the confusion between State and StateData (names of variables and names of classes). Now it's inconsistent.

I am thinking that we should rename `State` to `StateContainer` (or `StateHolder`), and use `State` as `StateData`. This way we can improve product/component development experience.

That why some `*StateData` classes are named with `*State` suffix (e.g. `ParagraphState` here).

ericlewis: gotcha, I would agree with that too. will make the change to be more succinct too :)

React/Fabric/Mounting/ComponentViews/Text/RCTParagraphComponentView.mm
ericlewis: @shergin ran `yarn clang-format` (complains about a lot of files!) but weirdly, nothing to say about the ordering. manually fixed it.

shergin:
@ericlewis  Unfortunately, the change broke Fabric Android. Could you please figure out why?

Make sure you are using the changes from the actual commit (which is unlanding right now, btw), not from this PR because it was landed with some changes.

ericlewis:
Hadn’t forgotten about this. I have a feeling I may know what happened. 

Issue #24875, opened 2019-05-16
👻 Flow Typing NativeModules


RSNara:
# Flow Typing NativeModules

The TurboModule system is nearly feature-complete on both Android and iOS. Before we rollout the system, we'd like to migrate all our OSS NativeModules to use it. As a prerequisite for this migration, all the the NativeModules must have spec files. There are currently 41 NativeModules in total, so there is a bunch of work to do. We'd love to use this as an opportunity to help get more people involved by contributing to React Native.

## Instructions

1. Pick an unclaimed module and comment that you're working on it.
2. Create a Spec file called `NativeXYZ.js`, where `XYZ` is the name of the NativeModule.
3. Define flow types based on the NativeModule's native methods.
4. Change call-sites from `NativeModules.XYZ` to `NativeXYZ`
5. Make sure flow passes
6. Do one module per PR

## What is a Spec file?

A Spec file looks like this:

### NativeAnalytics.js

```typescript
/**
 * Copyright 2004-present Facebook. All Rights Reserved.
 *
 * @flow strict-local
 * @format
 */

'use strict';

import type {TurboModule} from 'RCTExport';
import * as TurboModuleRegistry from 'TurboModuleRegistry';

export interface Spec extends TurboModule {
  +getConstants: () => {|
    constant1: string,
    constant2: boolean,
  |};
  +logCounter: (key: string, value: number) => void;
  +logEvent: (eventName: string, data: Object, analyticsModule: ?string) => void;
  +logRealtimeEvent: (
    eventName: string,
    data: Object,
    analyticsModule: ?string,
  ) => void;
}

export default TurboModuleRegistry.getEnforcing<Spec>(
  'Analytics',
);
```

### Observations

* **We use `TurboModuleRegistry.getEnforcing<Spec>('Analytics')` as opposed to `NativeModules.Analytics` to require the NativeModule.** `TurboModuleRegistry.getEnforcing` is an indirection that allows us to require both NativeModules and TurboModules. It throws if the NativeModule isn't there. In the case that a NativeModule is platform-specific, please use Platform.OS and conditionally call `TurboModuleRegistry.getEnforcing`.
* **Each Spec file exports an `interface` called `Spec` that extends `TurboModule`.** This interface contains typings for methods exposed by the NativeModule/TurboModule.
[...]

PR #24877, opened 2019-05-16
[TM] Add spec for Linking

ericlewis:
## Summary
Part of #24875, adds a spec for linking.

## Changelog

[General] [Added] - TM Spec for Linking

## Test Plan

Flow passes.

### DIFF
--- a/Libraries/Linking/Linking.js
+++ b/Libraries/Linking/Linking.js
@@ -16,7 +16,7 @@ const Platform = require('../Utilities/Platform');
 
 const invariant = require('invariant');
 
-const LinkingManager = require('./NativeLinking').default;
+import NativeLinking from './NativeLinking';
 
 /**
  * `Linking` gives you a general interface to interact with both incoming
@@ -26,7 +26,7 @@ const LinkingManager = require('./NativeLinking').default;
  */
 class Linking extends NativeEventEmitter {
   constructor() {
-    super(LinkingManager);
+    super(NativeLinking);
   }
 
   /**
@@ -55,7 +55,7 @@ class Linking extends NativeEventEmitter {
    */
   openURL(url: string): Promise<any> {
     this._validateURL(url);
-    return LinkingManager.openURL(url);
+    return NativeLinking.openURL(url);
   }
 
   /**
@@ -65,7 +65,7 @@ class Linking extends NativeEventEmitter {
    */
   canOpenURL(url: string): Promise<boolean> {
     this._validateURL(url);
-    return LinkingManager.canOpenURL(url);
+    return NativeLinking.canOpenURL(url);
   }
 
   /**
@@ -74,7 +74,7 @@ class Linking extends NativeEventEmitter {
    * See https://facebook.github.io/react-native/docs/linking.html#opensettings
    */
   openSettings(): Promise<any> {
-    return LinkingManager.openSettings();
+    return NativeLinking.openSettings();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ericlewis:
I am not 100% sure how to handle the event emitters here. Any feedback is welcome!

Libraries/Linking/Linking.js
```diff
@@ -12,15 +12,11 @@
...
-const NativeModules = require('../BatchedBridge/NativeModules');
 const Platform = require('../Utilities/Platform');
 
 const invariant = require('invariant');
 
-const LinkingManager =
-  Platform.OS === 'android'
-    ? NativeModules.IntentAndroid
-    : NativeModules.LinkingManager;
+const LinkingManager = require('./NativeLinking').default;
```
fkgozali: @cpojer, @rickhanlonii -- is it a good time to just use import?

```
import NativeLinking from './NativeLinking';
```

btw @ericlewis, let's keep the variable name `NativeLinking`

Libraries/Linking/NativeLinking.js
```diff
@@ -0,0 +1,33 @@
...
+  // Android only
+  +sendIntent: (
+    action: string,
+    extras: ?Array<{key: string, value: string | number | boolean}>,
+  ) => Promise<void>;
+}
+
+const moduleName =
+  Platform.OS === 'ios' ? 'NativeLinkingManager' : 'NativeIntentAndroid';
+export default TurboModuleRegistry.getEnforcing<Spec>(moduleName);
```
fkgozali: Let's not use a variable for module name, so that scripting can collect these callsites (we have an internal script that find all JS callsites for a TurboModule).

```
if (Platform.OS === 'ios') {
  export default TurboModuleRegistry.getEnforcing<Spec>('NativeLinkingManager');
} else {
  export default TurboModuleRegistry.getEnforcing<Spec>('NativeIntentAndroid');
}
```

@RSNara: we probably should rename these modules in the native side to have the same name...

Libraries/Linking/Linking.js
ericlewis: Happy to use import, and will switch variable names.

Libraries/Linking/NativeLinking.js
ericlewis: Cool

ericlewis: @fkgozali btw, it is not legal to export that way.

ericlewis:
```

export default (Platform.OS === 'ios'
  ? TurboModuleRegistry.getEnforcing<Spec>('LinkingManager')
  : TurboModuleRegistry.getEnforcing<Spec>('IntentAndroid'));
```

^ is valid

Libraries/Linking/NativeLinking.js
```diff
@@ -0,0 +1,37 @@
...
+  +openURL: (url: string) => Promise<void>;
+  +openSettings: () => Promise<void>;
+
+  // Android only
+  +sendIntent: (
+    action: string,
+    extras: ?Array<{key: string, value: string | number | boolean}>,
+  ) => Promise<void>;
+
+  // RCTEventEmitter
```
rickhanlonii: Could we update this to say `// Events`, RCTEventEmmitter is iOS specific and we're also dropping the RCT convention

ericlewis: Yupp !

rickhanlonii: cc @fkgozali, i think this came from your comment on the thread, just want to double check that it wasn't for a purpose i'm missing

fkgozali: This is iOS specific and the exact super class name is RCTEventEmitter - so it's fine to leave it that way.

fkgozali:
lgtm, landing

PR #24878, opened 2019-05-16
[TM] Add spec for DeviceInfo

ericlewis:
## Summary
Part of #24875. Causes a flow issue in Dimensions.js, trying to access `DeviceInfo.Dimensions`.

## Changelog

[General] [Added] - Add TM spec for DeviceInfo

## Test Plan

Flow tests pass.


Libraries/Utilities/NativeDeviceInfo.js
```diff
@@ -0,0 +1,33 @@
...
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    Dimensions: {|
+      width: number,
+      height: number,
+      scale: number,
+      fontScale: number,
+
+      // Android specific
+      densityDpi?: number,
```
RSNara: Looking through the code, it looks there's a discrepancy here:

## iOS
```objc
- (NSDictionary<NSString *, id> *)getConstants
{
  return @{
    @"Dimensions": RCTExportedDimensions(_bridge),
    // Note:
    // This prop is deprecated and will be removed in a future release.
    // Please use this only for a quick and temporary solution.
    // Use <SafeAreaView> instead.
    @"isIPhoneX_deprecated": @(RCTIsIPhoneX()),
  };
}

static NSDictionary *RCTExportedDimensions(RCTBridge *bridge)
{
  RCTAssertMainQueue();

  RCTDimensions dimensions = RCTGetDimensions(bridge.accessibilityManager.multiplier);
  typeof (dimensions.window) window = dimensions.window;
  NSDictionary<NSString *, NSNumber *> *dimsWindow = @{
      @"width": @(window.width),
      @"height": @(window.height),
      @"scale": @(window.scale),
      @"fontScale": @(window.fontScale)
  };
  typeof (dimensions.screen) screen = dimensions.screen;
  NSDictionary<NSString *, NSNumber *> *dimsScreen = @{
      @"width": @(screen.width),
      @"height": @(screen.height),
      @"scale": @(screen.scale),
      @"fontScale": @(screen.fontScale)
  };
  return @{
      @"window": dimsWindow,
      @"screen": dimsScreen
  };
}
```

## Android

```Java
  @Override
  public @Nullable Map<String, Object> getConstants() {
    HashMap<String, Object> constants = new HashMap<>();
    constants.put(
        "Dimensions",
        DisplayMetricsHolder.getDisplayMetricsMap(mFontScale));
    return constants;
  }
```

```Java
  public static Map<String, Map<String, Object>> getDisplayMetricsMap(double fontScale) {
    Assertions.assertNotNull(
        sWindowDisplayMetrics != null || sScreenDisplayMetrics != null,
        "DisplayMetricsHolder must be initialized with initDisplayMetricsIfNotInitialized or initDisplayMetrics");
    final Map<String, Map<String, Object>> result = new HashMap<>();
    result.put("windowPhysicalPixels", getPhysicalPixelsMap(sWindowDisplayMetrics, fontScale));
    result.put("screenPhysicalPixels", getPhysicalPixelsMap(sScreenDisplayMetrics, fontScale));
    return result;
  }

  private static Map<String, Object> getPhysicalPixelsMap(DisplayMetrics displayMetrics, double fontScale) {
    final Map<String, Object> result = new HashMap<>();
    result.put("width", displayMetrics.widthPixels);
    result.put("height", displayMetrics.heightPixels);
    result.put("scale", displayMetrics.density);
    result.put("fontScale", fontScale);
    result.put("densityDpi", displayMetrics.densityDpi);
    return result;
  }
```

It looks like Dimensions should be a map of maps. Are they flattened somewhere?

ericlewis: you are correct, they should be. different names too: `windowPhysicalPixels` for android & `window` on iOS.

ericlewis:
@RSNara pinging you on this one, which should also be good to go!

fkgozali:
Hey @ericlewis - it looks like there's merge conflict in this PR. It also turned out that we have a lot of internal callsites to fix for this module, so we ended up writing an internal PR for it. So I'm going to close this one to speed up the land process.

Thanks for working on this one!

PR #24879, opened 2019-05-16
[TM] Add spec for Settings

ericlewis:
## Summary
part of #24875. I again, am not completely sure how the call site here works- appears settings can be directly accessed?
 

## Changelog

[General] [Added] - Add TM spec for Settings

## Test Plan

Flow passes.


Libraries/Settings/NativeSettingsManager.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    settings: Object,
```
retyui: You should avoid using Object

`Object` is `any` since 0.88

```js
settings: { +[key: string]: string },
```

ericlewis: Just as an FYI- Object is how it was recommended to do it for the exercise- I did update to your suggestion however.

Libraries/Settings/NativeSettingsManager.js
```diff
@@ -0,0 +1,24 @@
...
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    settings: {[key: string]: string | boolean | number},
+  |};
+  +setValues: (values: {+[key: string]: string | boolean | number}) => void;
```
RSNara: Looking through the code, 

```ObjC
/**
 * Set one or more values in the settings.
 * TODO: would it be useful to have a callback for when this has completed?
 */
RCT_EXPORT_METHOD(setValues:(NSDictionary *)values)
{
  _ignoringUpdates = YES;
  [values enumerateKeysAndObjectsUsingBlock:^(NSString *key, id json, BOOL *stop) {
    id plist = [RCTConvert NSPropertyList:json];
    if (plist) {
      [self->_defaults setObject:plist forKey:key];
    } else {
      [self->_defaults removeObjectForKey:key];
    }
  }];

  [_defaults synchronize];
  _ignoringUpdates = NO;
}

// RCTConvert.m
+ (NSPropertyList)NSPropertyList:(id)json
{
  return RCTConvertPropertyListValue(json);
}

static id RCTConvertPropertyListValue(id json)
{
  if (!json || json == (id)kCFNull) {
    return nil;
  }

  if ([json isKindOfClass:[NSDictionary class]]) {
    __block BOOL copy = NO;
    NSMutableDictionary *values = [[NSMutableDictionary alloc] initWithCapacity:[json count]];
    [json enumerateKeysAndObjectsUsingBlock:^(NSString *key, id jsonValue, __unused BOOL *stop) {
      id value = RCTConvertPropertyListValue(jsonValue);
      if (value) {
        values[key] = value;
      }
      copy |= value != jsonValue;
    }];
    return copy ? values : json;
  }

  if ([json isKindOfClass:[NSArray class]]) {
    __block BOOL copy = NO;
    __block NSArray *values = json;
    [json enumerateObjectsUsingBlock:^(id jsonValue, NSUInteger idx, __unused BOOL *stop) {
      id value = RCTConvertPropertyListValue(jsonValue);
      if (copy) {
        if (value) {
          [(NSMutableArray *)values addObject:value];
        }
      } else if (value != jsonValue) {
        // Converted value is different, so we'll need to copy the array
        values = [[NSMutableArray alloc] initWithCapacity:values.count];
        for (NSUInteger i = 0; i < idx; i++) {
          [(NSMutableArray *)values addObject:json[i]];
        }
        if (value) {
          [(NSMutableArray *)values addObject:value];
        }
        copy = YES;
      }
    }];
    return values;
  }

  // All other JSON types are supported by property lists
  return json;
}
```

So it seems that in addition to `string`, `boolean` and `number`, we also support `null`, and Arrays and Objects.  Null looks like it's important because it's being used to unset a setting. The Array and Object subtypes could be tricky because it looks like they might require recursive flow types.

Libraries/Settings/NativeSettingsManager.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    settings: {[key: string]: string | boolean | number},
```
RSNara: I think there's a similar problem here.

Libraries/Settings/NativeSettingsManager.js
ericlewis: Ahh indeed, will update that. and yes, the below has the same issue, will create a common type perhaps

RSNara: Oh wait. I hope you didn't make any changes. It seems like our codegen system doesn't understand this flow syntax:

```typescript
{+[key: string]: string | boolean | number}
```

So we should just change this to Object and call it a day.

ericlewis: ha. i just made changes. will fix this then

ericlewis: @RSNara switched to Object.

retyui: Omg... @RSNara  `Object` is `any` 

Why not `{ [string]: mixed }` ?

ericlewis:
@RSNara pinging you on this again, i think it is good to go.

PR #24880, opened 2019-05-16
[TM] Add spec for AppState

ericlewis:
## Summary
part of #24875. 

## Changelog

[General] [Added] - Add TM spec for AppState

## Test Plan

Flow passes

### DIFF
--- a/Libraries/AppState/AppState.js
+++ b/Libraries/AppState/AppState.js
@@ -12,8 +12,7 @@
 
 const EventEmitter = require('../vendor/emitter/EventEmitter');
 const NativeEventEmitter = require('../EventEmitter/NativeEventEmitter');
-const NativeModules = require('../BatchedBridge/NativeModules');
-const RCTAppState = NativeModules.AppState;
+import NativeAppState from './NativeAppState';
 
 const logError = require('../Utilities/logError');
 const invariant = require('invariant');
@@ -30,7 +29,7 @@ class AppState extends NativeEventEmitter {
   isAvailable: boolean;
 
   constructor() {
-    super(RCTAppState);
+    super(NativeAppState);
 
     this.isAvailable = true;
     this._eventHandlers = {
@@ -38,7 +37,7 @@ class AppState extends NativeEventEmitter {
       memoryWarning: new Map(),
     };
 
-    this.currentState = RCTAppState.initialAppState;
+    this.currentState = NativeAppState.getConstants().initialAppState;
 
     let eventUpdated = false;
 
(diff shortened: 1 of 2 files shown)


Libraries/AppState/NativeAppState.js
```diff
@@ -0,0 +1,30 @@
...
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    initialAppState: string,
+  |};
+  +getCurrentAppState: (
+    success: (appState: {|app_state: string|}) => void,
+    failure: (error: Object) => void,
```
retyui:
```
failure: (error: Error) => void,
```

Libraries/AppState/NativeAppState.js
```diff
@@ -0,0 +1,30 @@
...
+  ) => void;
+
+  // RCTEventEmitter
```
rickhanlonii: `// Events`

Libraries/AppState/NativeAppState.js
rickhanlonii: Since I don't think we're instantiating an actual Error object, the best is probably an object type with the fields specified but I defer to @RSNara

rickhanlonii: I checked and as far as I can tell we don't use Error in any of the specs we've created so far

RSNara: It looks like the native implementation of this method also never calls the error callback. We should just remove the error callback. It looks like we have three options:

1. Remove the argument from the error callback.
2. Remove the error callback.
3. Leave the argument in the error callback as an Object. 

Thoughts?

RSNara: Although it's a bit more work, I'd prefer that we clean up the native code and remove the error callback. But I'll leave that decision up to you @ericlewis.

rickhanlonii: Yeah I agree, I think we do at least 2 and preferably also remove it from native 👍

fkgozali: I think it's fine to keep the native side as is for this PR. You can send a follow up cleanup if you wish.

Libraries/AppState/AppState.js
```diff
@@ -152,7 +151,7 @@ class MissingNativeAppStateShim extends EventEmitter {
 // This module depends on the native `RCTAppState` module. If you don't include it,
 // `AppState.isAvailable` will return `false`, and any method calls will throw.
 // We reassign the class variable to keep the autodoc generator happy.
-if (RCTAppState) {
+if (NativeAppState) {
```
fkgozali: Btw, since you use `getEnforcing()`, this is always `true`-ish, and we get invariant() instead when it's not available.

This may be a breaking change, so if that's the case we could downgrade it to `get()` instead.

We can try landing this PR then adjust things if it breaks something.

ericlewis: oh wow, i didn't realize that. okay. this is more to satisfy flow i guess- so i should check null instead?

PR #24881, opened 2019-05-16
[TM] Add spec for KeyboardObserver

ericlewis:
## Summary

part of #24875. iOS only it appears, but not really used by RN itself. Should be fine? 

## Changelog

[General] [Added] - Add TM spec for KeyboardObserver

## Test Plan

Flow passes.


ericlewis:
Yep, this one is pretty weird- it doesn’t seem to have a reason to freak out that bad. Will address today though. 

Libraries/Components/Keyboard/NativeKeyboardObserver.js
```diff
@@ -0,0 +1,24 @@
...
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  +addListener: (eventName: string) => void;
+  +removeListeners: (count: number) => void;
+}
+
+export default (Platform.OS === 'ios'
+  ? TurboModuleRegistry.getEnforcing<Spec>('KeyboardObserver')
+  : null);
```
rickhanlonii: We can just use `.get<spec>` here without a platform check

PR #24888, opened 2019-05-16
[TM] Add spec for AndroidToast

wojteg1337:
## Summary

part of #24875 

## Changelog

[General] [Added] - Add TM spec for AndroidToast

## Test Plan

Flow passes

### DIFF
--- a/Libraries/Components/ToastAndroid/NativeToastAndroid.js
+++ b/Libraries/Components/ToastAndroid/NativeToastAndroid.js
@@ -12,7 +12,6 @@
 
 import type {TurboModule} from 'RCTExport';
 import * as TurboModuleRegistry from 'TurboModuleRegistry';
-import {Platform} from 'react-native';
 
 export interface Spec extends TurboModule {
   +getConstants: () => {|
@@ -38,6 +37,4 @@ export interface Spec extends TurboModule {
   ) => void;
 }
 
-export default (Platform.OS === 'android'
-  ? TurboModuleRegistry.getEnforcing<Spec>('ToastAndroid')
-  : null);
+export default TurboModuleRegistry.getEnforcing<Spec>('ToastAndroid');
--- a/Libraries/Components/ToastAndroid/NativeToastAndroid.js
+++ b/Libraries/Components/ToastAndroid/NativeToastAndroid.js
@@ -1,0 +1,44 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|
+    SHORT: number,
+    LONG: number,
+    TOP: number,
+    BOTTOM: number,
+    CENTER: number,
+  |};
+
+  +show: (message: string, duration: number) => void;
+  +showWithGravity: (
+    message: string,
+    duration: number,
+    gravity: number,
+  ) => void;
+  +showWithGravityAndOffset: (
+    message: string,
+    duration: number,
+    gravity: number,
+    xOffset: number,
+    yOffset: number,
+  ) => void;
+
+  // RCTEventEmitter
+  +addListener: (eventName: string) => void;
+  +removeListeners: (count: number) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('ToastAndroid');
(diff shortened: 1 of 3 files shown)


Libraries/Components/ToastAndroid/NativeToastAndroid.js
```diff
@@ -0,0 +1,44 @@
...
+    xOffset: number,
+    yOffset: number,
+  ) => void;
+
+  // RCTEventEmitter
+  +addListener: (eventName: string) => void;
+  +removeListeners: (count: number) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('ToastAndroid');
```
lunaleaps: I think we need to conditionally call `getEnforcing` here

See https://github.com/facebook/react-native/issues/24875
> In the case that a NativeModule is platform-specific, please use Platform.OS and conditionally call TurboModuleRegistry.getEnforcing.

Libraries/Components/ToastAndroid/NativeToastAndroid.js
```diff
@@ -0,0 +1,44 @@
...
+  +showWithGravityAndOffset: (
+    message: string,
+    duration: number,
+    gravity: number,
+    xOffset: number,
+    yOffset: number,
+  ) => void;
+
+  // RCTEventEmitter
+  +addListener: (eventName: string) => void;
```
lunaleaps: Is this an event emitter class?

Libraries/Components/ToastAndroid/NativeToastAndroid.js
ericlewis: ^ this is correct

Libraries/Components/ToastAndroid/NativeToastAndroid.js
ericlewis: Please remove these methods, it is not an event emitter.

wojteg1337:
@ericlewis i fixed comments you left, could you review it once again?

ericlewis:
@wojteg1337 there are tons of errors here still. Since the library has been changed to conditional you probably need to do invariant checks on wether or not the library exists at existing call sites. Once you get all the tests passing (except e2e, don’t worry about that) I’ll approve.

Libraries/Components/ToastAndroid/NativeToastAndroid.js
```diff
@@ -0,0 +1,43 @@
...
+    duration: number,
+    gravity: number,
+    xOffset: number,
+    yOffset: number,
+  ) => void;
+}
+
+export default (Platform.OS === 'android'
+  ? TurboModuleRegistry.getEnforcing<Spec>('ToastAndroid')
+  : null);
```
lunaleaps: Is it odd to `export default null`? I wonder if we shoudn't export at all if iOS

ericlewis:
you have to export something, undefined or null. i dont think there is a specific on witch. 

Sent via Superhuman ( https://sprh.mn/?vip=ericlewis777@gmail.com )

ericlewis:
its not possible to export *nothing* i.e., make the export itself conditional

Sent via Superhuman ( https://sprh.mn/?vip=ericlewis777@gmail.com )

wojteg1337:
@ericlewis, I removed the conditional export of turbo module since it's never used in `ios.js` files. In iOS files, I added stubs to return correct types.

wojteg1337:
also I'm not sure what causes `test_android` to fail, could you help me with it?

Libraries/Components/ToastAndroid/NativeToastAndroid.js
```diff
@@ -0,0 +1,44 @@
...
+  +getConstants: () => {|
+    SHORT: number,
+    LONG: number,
+    TOP: number,
+    BOTTOM: number,
+    CENTER: number,
+  |};
+
+  +show: (message: string, duration: number) => void;
+  +showWithGravity: (
```
uqmessias: Since this is android only, shouldn't it be optional?

Libraries/Components/ToastAndroid/NativeToastAndroid.js
```diff
@@ -0,0 +1,44 @@
...
+    message: string,
+    duration: number,
+    gravity: number,
+  ) => void;
+  +showWithGravityAndOffset: (
```
uqmessias: Ditto

lunaleaps:
@wojteg1337 do you have the verbose output of the failing test?

wojteg1337:
@lunaleaps when I add a verbose flag to the command from test `node cli.js bundle --max-workers 2 --platform android --dev true --entry-file ReactAndroid/src/androidTest/js/TestBundle.js --bundle-output ReactAndroid/src/androidTest/assets/AndroidTestBundle.js --verbose` the output seems to be the same as the one from CI

PR #24891, opened 2019-05-16
[TM] Add specs for AccessibilityInfo

michalchudziak:
## Summary

Part of https://github.com/facebook/react-native/issues/24875

## Changelog

[General] [Added] - Add TurboModule spec for AccessibilityInfo

## Test Plan

Run `yarn flow` on the root level

### DIFF
--- a/Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
+++ b/Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +isReduceMotionEnabled: (
+    resolve: (isReduceMotionEnabled: boolean) => void,
+  ) => void;
+  +isTouchExplorationEnabled: (
+    resolve: (isScreenReaderEnabled: boolean) => void,
+  ) => void;
+  +setAccessibilityFocus: (reactTag: number) => void;
+  +announceForAccessibility: (announcement: string) => void;
+
+  // RCTDeviceEventEmitter
+  +addListener: (eventName: string, handler: Function) => Object;
+  +removeListeners: (eventName: string, handler: Function) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('AccessibilityInfo');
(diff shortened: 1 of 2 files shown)


Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
```diff
@@ -0,0 +1,31 @@
...
+  ) => void;
+  +setAccessibilityFocus: (reactTag: number) => void;
+  +announceForAccessibility: (announcement: string) => void;
+
+  // RCTDeviceEventEmitter
+  +addListener: (eventName: string, handler: Function) => Object;
+  +removeListeners: (eventName: string, handler: Function) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('AccessibilityInfo');
```
RSNara: `AccessibilityInfo` is an Android-only module. This will throw an error when required on iOS. So, we should change this up to conditionally call `TurboModuleRegistry.getEnforcing<Spec>`.

```typescript
export default (Platform.OS === 'android'
  ? TurboModuleRegistry.getEnforcing<Spec>('AccessibilityInfo')
  : null);
```

Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +isReduceMotionEnabled: (
+    resolve: (isReduceMotionEnabled: boolean) => void,
```
RSNara: I think a more appropriate name for this would be `onSuccess`. The name `resolve` only makes sense if you're working with promises.

Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
```diff
@@ -0,0 +1,31 @@
...
+  ) => void;
+  +isTouchExplorationEnabled: (
+    resolve: (isScreenReaderEnabled: boolean) => void,
```
RSNara: We should also change this to `onSuccess`.

Libraries/Components/AccessibilityInfo/NativeAccessibilityInfo.js
```diff
@@ -0,0 +1,31 @@
...
+  ) => void;
+  +setAccessibilityFocus: (reactTag: number) => void;
+  +announceForAccessibility: (announcement: string) => void;
+
+  // RCTDeviceEventEmitter
+  +addListener: (eventName: string, handler: Function) => Object;
+  +removeListeners: (eventName: string, handler: Function) => void;
```
RSNara: NativeModules on Android don't expose inherited methods to JS. And I don't see these methods in `AccessibilityInfoMethod`. Are you sure these methods belong here?

michalchudziak: I wasn't sure about that, I'm going to remove it.

PR #24892, opened 2019-05-16
[TM] Add spec for Networking

thymikee:
## Summary

Part of #24875, adds a spec for Networking. Since `sendRequest` methods are different for both platforms, I had to create 2 spec files as Flow would merge their definitions even when I added `Platform.OS` check

## Changelog

[General] [Added] - TM spec for Networking

## Test Plan

Flow passes.


Libraries/Network/NativeNetworkingIOS.js
```diff
@@ -0,0 +1,36 @@
...
+      headers: {[key: string]: mixed},
+      responseType: NativeResponseType,
+      incrementalUpdates: boolean,
+      timeout: number,
+      withCredentials: boolean,
+    |},
+    callback: (requestId: number) => void,
+  ) => void;
+  +abortRequest: (requestId: number) => void;
+  +clearCookies: (callback: (result: boolean) => void) => void;
```
retyui: ```callback: (result: boolean) => mixed```

Libraries/Network/RCTNetworking.ios.js
```diff
@@ -53,11 +52,11 @@ class RCTNetworking extends NativeEventEmitter {
   }
 
   abortRequest(requestId: number) {
-    RCTNetworkingNative.abortRequest(requestId);
+    NativeNetworking.abortRequest(requestId);
   }
 
   clearCookies(callback: (result: boolean) => any) {
```
retyui: ```callback: (result: boolean) => mixed```

Libraries/Network/RCTNetworking.android.js
```diff
@@ -80,11 +79,11 @@ class RCTNetworking extends NativeEventEmitter {
   }
 
   abortRequest(requestId: number) {
-    RCTNetworkingNative.abortRequest(requestId);
+    NativeNetworking.abortRequest(requestId);
   }
 
   clearCookies(callback: (result: boolean) => any) {
```
retyui: ```callback: (result: boolean) => mixed```

Libraries/Network/NativeNetworkingAndroid.js
```diff
@@ -0,0 +1,34 @@
...
+    requestId: number,
+    headers: {[key: string]: mixed},
+    data: {[key: string]: mixed},
+    responseType: NativeResponseType,
+    useIncrementalUpdates: boolean,
+    timeout: number,
+    withCredentials: boolean,
+  ) => void;
+  +abortRequest: (requestId: number) => void;
+  +clearCookies: (callback: (result: boolean) => void) => void;
```
retyui: ```callback: (result: boolean) => mixed```

Libraries/Network/RCTNetworking.ios.js
thymikee: why mixed?

retyui: Here is the reason [flow fry ...](https://flow.org/try/#0FDAmFMGMBsEMCdwAIBmBXAdpALgSwPYZICCAFJLNNAEayQDWAXEqQJQC8AfAG766itmvfgG4QwMmyRckAbwC+rMRNKwAzgE8sLVtM5zFIoA)

thymikee: Fair call, changed

Libraries/Network/NativeNetworkingIOS.js
```diff
@@ -0,0 +1,36 @@
...
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {NativeResponseType} from './XMLHttpRequest';
+
+export interface Spec extends TurboModule {
+  +sendRequest: (
+    query: {|
+      method: string,
+      url: string,
+      data: {[key: string]: mixed},
```
ericlewis: use Object

Libraries/Network/NativeNetworkingIOS.js
```diff
@@ -0,0 +1,36 @@
...
+      headers: {[key: string]: mixed},
```
ericlewis: use Object

Libraries/Network/NativeNetworkingIOS.js
```diff
@@ -0,0 +1,36 @@
...
+      responseType: NativeResponseType,
+      incrementalUpdates: boolean,
+      timeout: number,
+      withCredentials: boolean,
+    |},
+    callback: (requestId: number) => mixed,
+  ) => void;
+  +abortRequest: (requestId: number) => void;
+  +clearCookies: (callback: (result: boolean) => mixed) => void;
```
ericlewis: not sure if mix is understood by the codegen, but i would err towards no.

Libraries/Network/NativeNetworkingAndroid.js
```diff
@@ -0,0 +1,36 @@
...
+
+type Header = [string, string];
+
+export interface Spec extends TurboModule {
+  +sendRequest: (
+    method: string,
+    url: string,
+    requestId: number,
+    headers: Array<Header>,
+    data: {[key: string]: mixed},
```
ericlewis: use Object

Libraries/Network/NativeNetworkingAndroid.js
```diff
@@ -0,0 +1,36 @@
...
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {NativeResponseType} from './XMLHttpRequest';
+
+type Header = [string, string];
```
ericlewis: not sure if Tuples are supported, should be interesting!

thymikee:
>mostly looks good, but are we supposed to create 2 Specs?

This is what we concluded with @rickhanlonii since there are 2 methods with different signatures, and Flow wasn't happy with `Platform.OS` branching. I'd be glad if someone can point a better solution :)

Libraries/Network/NativeNetworkingIOS.js
thymikee: @rickhanlonii we need your powers here

rickhanlonii: Why does a callback have a return value?

I don't see any other usages of mixed like this

rickhanlonii:
Probably the best we can do until we unify the native interface to use the same prop types

Libraries/Network/NativeNetworkingIOS.js
thymikee: changing to void right now

thymikee: Oh, this change. Please see this comment: https://github.com/facebook/react-native/pull/24892#discussion_r284665079 what do you think?

thymikee: @RSNara could you please advise here? Is using `mixed` for callbacks supported by codegen? Or if we want users to be able to pass async callbacks we should return `Promise<void> | void` instead of just `void`?

PR #24893, opened 2019-05-16
[TM] Add spec for WebSocketModule

jeanregisser:
## Summary

Part of #24875

## Changelog

[General] [Added] - Add TurboModule spec for WebSocketModule

## Test Plan

Run `yarn flow` at the root level


Libraries/WebSocket/NativeWebSocketModule.js
```diff
@@ -0,0 +1,38 @@
...
+    protocols: ?Array<string>,
+    options: ?{headers?: {origin?: string}},
+    socketID: number,
+  ) => void;
+  +send: (message: string, socketID: number) => void;
+  +sendBinary: (base64String: string, socketID: number) => void;
+  +ping: (socketID: number) => void;
+  +close: typeof close;
+
+  // RCTEventEmitter
```
ericlewis: change this to `// Events`

Libraries/WebSocket/NativeWebSocketModule.js
```diff
@@ -0,0 +1,38 @@
...
+  +connect: (
+    url: string,
+    protocols: ?Array<string>,
+    options: ?{headers?: {origin?: string}},
+    socketID: number,
+  ) => void;
+  +send: (message: string, socketID: number) => void;
+  +sendBinary: (base64String: string, socketID: number) => void;
+  +ping: (socketID: number) => void;
+  +close: typeof close;
```
RSNara: This is tricky. 

## Approaches
1. Create the `specs/{ios,android}` folders. Then, have a spec for ios and a spec for android. The only thing different between the two would be the close method. Then, you could create a NativeWebSocketModule.js file that returns the correct implementation based on Platform.OS. 
2. Refactor the API of close on Android to be `close(socketID: number, code: number, reason: string): void;`. Then type the second and third argument as optional. It seems like the only place where we use WebSocketModule is in WebSocket.js, so this wouldn't be difficult. Also, while we're working on WebSocketModule, it might be worthwhile to actually do some cleanup. Also, with approach number one, we have a lot of duplicate code. 

I'm leaning more towards option 2. Even though it's probably the more difficult option, and even though it is a breaking change, I think it's probably the better solution. cc @fkgozali

jeanregisser: Thanks @RSNara 

I was thinking about a 3rd approach, quite similar to approach 2:

3. Refactoring the iOS API and allowing closing the socket with the reason and code like on Android. I checked the underlying implementation and it seems like a viable approach: https://github.com/facebook/react-native/blob/7abfd23b90db08b426c3c91b0cb6d01d161a9b9e/Libraries/WebSocket/RCTSRWebSocket.m#L558

This would actually make the iOS implementation more correct. Right now the JS interface for closing the WebSocket supports the reason and code but ignores it on iOS.

If we go this way, should the refactor be done in a different PR? I assume yes, but just checking.

Let me know what you think.

ericlewis:
If there is a misalignment of features that is fine, for now we just want a Spec, if you do think there is a misalignment of feature that can be fixed open a new PR :) thank you!!

Libraries/WebSocket/NativeWebSocketModule.js
RSNara: That's also a good idea! But unfortunately, I just realized that our fb-internal infra doesn't allow refactoring NativeModule methods. So for the time being, I think we should just create two different spec files.

jeanregisser: I'm not too sure what you mean about the fb-internal infra refactoring limitation, but I've created 2 slightly different specs for now in https://github.com/facebook/react-native/pull/24893/commits/05d31c1f8ec74a7d35cebb1c693fa4cebe48af7a

Note that I had to suppress the flow error when calling the `close` method. Let me know if there's a better way to avoid this.

jeanregisser: Alright since my small refactor for the iOS close method [got merged](https://github.com/facebook/react-native/commit/0ac2171c549b389228c4a37ae645eb0d9813b82d), I'm gonna update the spec! There's no more a difference between the native iOS and Android implementations.

PR #24896, opened 2019-05-16
[TM] Add specs for ModalManager

abhiin1947:
## Summary
Part of #24875

## Changelog
[General] [Added] - Add TurboModule spec for ModalManager

## Test Plan
Flow and tests pass

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -12,8 +12,6 @@
 
 const AppContainer = require('../ReactNative/AppContainer');
 const I18nManager = require('../ReactNative/I18nManager');
-const NativeEventEmitter = require('../EventEmitter/NativeEventEmitter');
-const NativeModules = require('../BatchedBridge/NativeModules');
 const Platform = require('../Utilities/Platform');
 const React = require('react');
 const PropTypes = require('prop-types');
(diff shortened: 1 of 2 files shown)


Libraries/Modal/NativeModalManager.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  // RCTEventEmitter
+  +addListener: (eventName: string, handler: Function) => Object;
```
ericlewis: remove these listeners, since it is not an event emitter

Libraries/Modal/Modal.js
```diff
@@ -23,9 +21,7 @@ const View = require('../Components/View/View');
 const RCTModalHostView = require('./RCTModalHostViewNativeComponent');
 
 const ModalEventEmitter =
-  Platform.OS === 'ios' && NativeModules.ModalManager
-    ? new NativeEventEmitter(NativeModules.ModalManager)
-    : null;
+  Platform.OS === 'ios' ? require('./NativeModalManager').default : null;
```
ericlewis: change this to an import, also, do the condition in `NativeModalManager.js` not here.

Libraries/Modal/NativeModalManager.js
fkgozali: RCTModalManager actually extends RCTEventEmitter

abhiin1947: What @fkgozali said I couldn't get flow to run otherwise.

PR #24897, opened 2019-05-16
[TM] Add spec for TimePickerAndroid

jeanregisser:
## Summary

Part of #24875

## Changelog

[General] [Added] - Add TurboModule spec for TimePickerAndroid

## Test Plan

Run `yarn flow-check-ios` and `yarn flow-check-android` at the root level


Libraries/Components/TimePickerAndroid/NativeTimePickerAndroid.js
```diff
@@ -0,0 +1,25 @@
...
+import type {
+  TimePickerOptions,
+  TimePickerResult,
+} from './TimePickerAndroidTypes';
+
+export interface Spec extends TurboModule {
+  +open: (options: TimePickerOptions) => Promise<TimePickerResult>;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('TimePickerAndroid');
```
ericlewis: change this to only export if on android

Libraries/Components/TimePickerAndroid/TimePickerAndroid.android.js
```diff
@@ -59,7 +58,7 @@ class TimePickerAndroid {
    * being undefined. **Always** check whether the `action` before reading the values.
    */
   static async open(options: TimePickerOptions): Promise<TimePickerResult> {
-    return TimePickerModule.open(options);
+    return NativeTimePickerAndroid.open(options);
```
ericlewis: after the first change you will need to do this:
```
if (NativeTimePickerAndroid) {
  return NativeTimePickerAndroid.open(options);
}
```

jeanregisser: Thanks, I've consolidated both iOS and Android implementations in https://github.com/facebook/react-native/pull/24897/commits/b45ea6179e0f334dcb444bae082d818e4dfa0b88

Libraries/Components/TimePickerAndroid/NativeTimePickerAndroid.js
```diff
@@ -0,0 +1,28 @@
...
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import Platform from '../../Utilities/Platform';
+
+import type {
+  TimePickerOptions,
+  TimePickerResult,
+} from './TimePickerAndroidTypes';
```
RSNara: Unfortunately, all types used by the spec need to be declared in this file. Perhaps we should relocate these types to this file and have files that depend on TimePickerAndroidTypes instead depend on NativeTimePickerAndroid?

Libraries/Components/TimePickerAndroid/TimePickerAndroid.js
```diff
@@ -59,7 +58,13 @@ class TimePickerAndroid {
    * being undefined. **Always** check whether the `action` before reading the values.
    */
   static async open(options: TimePickerOptions): Promise<TimePickerResult> {
-    return TimePickerModule.open(options);
+    if (NativeTimePickerAndroid) {
+      return NativeTimePickerAndroid.open(options);
+    } else {
+      return Promise.reject({
```
uqmessias: :tada:

Libraries/Components/TimePickerAndroid/NativeTimePickerAndroid.js
jeanregisser: Sounds good, I did that in https://github.com/facebook/react-native/pull/24897/commits/03b99364f6f752c121fd502c116caee787cb7d4a

PR #24899, opened 2019-05-16
[TM] Add spec for HeapCapture

wojteg1337:
## Summary

Part of #24875, adds a spec for HeapCapture

## Changelog

[General] [Added] - TM Spec for HeapCapture

## Test Plan

Flow passes

### DIFF
--- a/Libraries/Core/setUpBatchedBridge.js
+++ b/Libraries/Core/setUpBatchedBridge.js
@@ -22,7 +22,7 @@ BatchedBridge.registerLazyCallableModule('JSTimers', () =>
   require('./Timers/JSTimers'),
 );
 BatchedBridge.registerLazyCallableModule('HeapCapture', () =>
-  require('../Utilities/HeapCapture'),
+  require('../HeapCapture/HeapCapture'),
 );
 BatchedBridge.registerLazyCallableModule('SamplingProfiler', () =>
   require('../Performance/SamplingProfiler'),
--- a/Libraries/HeapCapture/NativeHeapCapture.js
+++ b/Libraries/HeapCapture/NativeHeapCapture.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  // Common interface
+  +captureHeap: (path: string) => void;
+
+  // Android only
+  +captureComplete: (path: string, error: ?string) => void;
+
+  // Events
+  +addListener: (eventName: string, handler: Function) => Object;
+  +removeListeners: (eventName: string, handler: Function) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('HeapCapture');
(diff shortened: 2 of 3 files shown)


Libraries/Core/setUpBatchedBridge.js
```diff
@@ -22,7 +22,7 @@ BatchedBridge.registerLazyCallableModule('JSTimers', () =>
   require('./Timers/JSTimers'),
 );
 BatchedBridge.registerLazyCallableModule('HeapCapture', () =>
-  require('../Utilities/HeapCapture'),
+  require('../HeapCapture/HeapCapture'),
```
ericlewis: requiring like this means you should also grab `default` i think.

Libraries/HeapCapture/NativeHeapCapture.js
```diff
@@ -0,0 +1,28 @@
...
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  // Common interface
+  +captureHeap: (path: string) => void;
+
+  // Android only
+  +captureComplete: (path: string, error: ?string) => void;
```
ericlewis: i'm not 100% sure, but this may need to be optional.

Libraries/HeapCapture/NativeHeapCapture.js
```diff
@@ -0,0 +1,28 @@
...
+
+  // Events
+  +addListener: (eventName: string, handler: Function) => Object;
+  +removeListeners: (eventName: string, handler: Function) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('HeapCapture');
```
ericlewis: this is android only, so use platform to conditionally export

Libraries/HeapCapture/NativeHeapCapture.js
```diff
@@ -0,0 +1,28 @@
...
+
+export interface Spec extends TurboModule {
+  // Common interface
+  +captureHeap: (path: string) => void;
+
+  // Android only
+  +captureComplete: (path: string, error: ?string) => void;
+
+  // Events
+  +addListener: (eventName: string, handler: Function) => Object;
```
ericlewis: is HeapCapture an emitter? If not, remove these

Libraries/Core/setUpBatchedBridge.js
wojteg1337: i'm not sure if there should be default since export looks like this:

`module.exports = HeapCapture;`

Libraries/HeapCapture/NativeHeapCapture.js
wojteg1337: yes, sorry there is no event emitter in java class.

wojteg1337:
@ericlewis i fixed comments you left, could you review it once again?

Libraries/Core/setUpBatchedBridge.js
ericlewis: Ah gotcha, thought this was coming direct from NativeHeapCapture

Libraries/HeapCapture/NativeHeapCapture.js
```diff
@@ -0,0 +1,27 @@
...
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  // Common interface
+  +captureHeap: (path: string) => void;
+
+  // Android only
```
RSNara: nit: It looks like this module is only available on Android. So, this comment is confusing. Should it be here?

Libraries/HeapCapture/NativeHeapCapture.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  // Common interface
```
RSNara: nit: Should this be here? (see previous comment).

PR #24900, opened 2019-05-16
[TM] Add spec for ExceptionsManager

thymikee:
## Summary

Part of #24875, adds a spec for ExceptionsManager

## Changelog

[General] [Added] - TM Add spec for ExceptionsManager

## Test Plan

Flow passes.

### DIFF
--- a/Libraries/Core/NativeExceptionsManager.js
+++ b/Libraries/Core/NativeExceptionsManager.js
@@ -27,9 +27,7 @@ export interface Spec extends TurboModule {
     stack: Array<StackFrame>,
     exceptionId: number,
   ) => void;
-  /**
-   * Available only on Android
-   */
+  // Android only
   +dismissRedbox: () => void;
 }
 
--- a/Libraries/Core/NativeExceptionsManager.js
+++ b/Libraries/Core/NativeExceptionsManager.js
@@ -9,7 +9,13 @@
 
 import type {TurboModule} from 'RCTExport';
 import * as TurboModuleRegistry from 'TurboModuleRegistry';
-import type {StackFrame} from './Devtools/parseErrorStack';
+
+export type StackFrame = {|
+  column: ?number,
+  file: string,
+  lineNumber: number,
+  methodName: string,
+|};
 
 export interface Spec extends TurboModule {
   +reportFatalException: (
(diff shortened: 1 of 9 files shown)


Libraries/Core/NativeExceptionsManager.js
```diff
@@ -0,0 +1,36 @@
...
+  ) => void;
+  +updateExceptionMessage: (
+    message: string,
+    stack: Array<StackFrame>,
+    exceptionId: number,
+  ) => void;
+  /**
+   * Available only on Android
+   */
+  +dismissRedbox: () => void;
```
thymikee: what do we do about it @rickhanlonii?

zhongwuzw: Should we need to split it to Android and iOS? if not, seems we need to label it for Codegen.

rickhanlonii: For additional properties we have decided to generate it on both platforms and no-op on one, so for this you can update the comment to `// Android only`

Splitting out would be required for props/events that conflict

Libraries/Core/NativeExceptionsManager.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @flow strict-local
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {StackFrame} from './Devtools/parseErrorStack';
```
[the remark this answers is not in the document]
thymikee: Moved the type def to NativeExceptionsManager as there wasn't too many call-sites to change

Libraries/Core/NativeExceptionsManager.js
```diff
@@ -0,0 +1,40 @@
...
+    stack: Array<StackFrame>,
+    exceptionId: number,
+  ) => void;
+  +updateExceptionMessage: (
+    message: string,
+    stack: Array<StackFrame>,
+    exceptionId: number,
+  ) => void;
+  // Android only
+  +dismissRedbox?: () => void;
```
zhongwuzw: @RSNara  🤔 Seems `sentIntent` of `IntentAndroid` not marks optional
https://github.com/facebook/react-native/pull/24877/files#diff-7033e846872152e52446abb20da3efb8R25, should we also changed that?

RSNara: Yeah, nice catch. We should change that.

RSNara: Actually, no. I was wrong earlier about our codegen output. Turns out, regardless of if a method is optional or not, it'll generate a required method in native. So, making this optional doesn't do anything.

zhongwuzw: 👌 Get it.

PR #24902, opened 2019-05-16
[TM] Add spec for UIManager

ericlewis:
## Summary
part of #24875. Because some of the methods are rewriteable, I dropped the `+` from the signature, this doesn't feel right to me, but I am not sure if the codegen requires that. If it does, it will probably be better to extend the spec and allow those specific methods to be overriden in a UIManager.js interface. Thoughts on that @fkgozali or @RSNara?

## Changelog

[General] [Added] - Add TM spec for UIManager

## Test Plan

Flow passes.

### DIFF
--- a/Libraries/ReactNative/NativeUIManager.js
+++ b/Libraries/ReactNative/NativeUIManager.js
@@ -105,6 +105,7 @@ export interface Spec extends TurboModule {
 
   // android only
   +setLayoutAnimationEnabledExperimental: (enabled: boolean) => void;
+  +sendAccessibilityEvent: (tag: number, eventType: number) => void;
 }
 
 export default TurboModuleRegistry.getEnforcing<Spec>('UIManager');
--- a/Libraries/ReactNative/NativeUIManager.js
+++ b/Libraries/ReactNative/NativeUIManager.js
@@ -104,7 +104,7 @@ export interface Spec extends TurboModule {
   ) => void;
 
   // android only
-  +setLayoutAnimationEnabledExperimental?: (enabled: boolean) => void;
+  +setLayoutAnimationEnabledExperimental: (enabled: boolean) => void;
 }
 
 export default TurboModuleRegistry.getEnforcing<Spec>('UIManager');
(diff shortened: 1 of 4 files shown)


Libraries/ReactNative/NativeUIManager.js
```diff
@@ -0,0 +1,108 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
```
thymikee: Can you use strict-local and update references to Object which is implicit any?

ericlewis: @thymikee object is a desired return result according to the issue filed. In codegen, I think it has a different meaning from any.

thymikee: I tend to use `{[key: string]: mixed}` instead of Object. I'm not sure what assumptions codegen is making, but I would expect it (sooner or later) to get data from Flow Type AST

fkgozali: For now, let's stick with `Object`. The internal codegen isn't very smart yet, so we can improve it over time later.

fkgozali:
> If it does, it will probably be better to extend the spec and allow those specific methods to be overriden in a UIManager.js interface.

So, I think the Spec should focus on what exists in native classes. For props that are updated during runtime, you can introduce a wrapper object that provides those methods. For example:

```
interface Spec extends TurboModule {
  +foo: () => void;
}

const Module = TurboModuleRegistry.getEnforcing<Spec>('UIManager');

type WrapperSpec = {
// extend the spec above here somehow
};

const UIManager: WrapperSpec = {
  foo(): void {
    Module.foo();
  },
  updatedFunc(): void {
    // do something here
  },
}

export default UIManager;
```

ericlewis:
@fkgozali I wrapped the native module that I was originally exporting from UIManager.js in a type that extends the spec, the other overrides are in native code too- so im not 100% sure we want to remove them from the spec, they get modified later and call their originals. not sure if legal.

PR #24903, opened 2019-05-16
[TM] spec for JSCSamplingProfiler

ericlewis:
## Summary
part of #24875.

## Changelog

[General] [Added] - Add TM spec for JSCSamplingProfiler

## Test Plan

Flow passes.

### DIFF
--- a/Libraries/Performance/NativeJSCSamplingProfiler.js
+++ b/Libraries/Performance/NativeJSCSamplingProfiler.js
@@ -12,9 +12,12 @@
 
 import type {TurboModule} from 'RCTExport';
 import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import {Platform} from 'react-native';
 
 export interface Spec extends TurboModule {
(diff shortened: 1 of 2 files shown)


Libraries/Performance/NativeJSCSamplingProfiler.js
```diff
@@ -0,0 +1,23 @@
...
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  +operationComplete: (token: number, result: ?string, error: ?string) => void;
+}
+
+export default (Platform.OS === 'android'
+  ? TurboModuleRegistry.getEnforcing<Spec>('JSCSamplingProfiler')
```
fkgozali: Like in other PRs, this should call .get() instead, no need to check for platform

PR #24906, opened 2019-05-16
[TM] Add spec for AlertManager

sasurau4:
## Summary

Part of #24875 

## Changelog

[General] [Added] - Add TurboModule spec for AlertManager

## Test Plan
- `yarn flow`

### DIFF
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -15,8 +15,7 @@ import Platform from '../Utilities/Platform';
 import DialogManagerAndroid, {
   type DialogOptions,
 } from '../NativeModules/specs/NativeDialogManagerAndroid';
-
-const RCTAlertManager = NativeModules.AlertManager;
+import RCTAlertManager from './RCTAlertManager';
 
 export type Buttons = Array<{
   text?: string,
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -125,11 +125,14 @@ class Alert {
         );
 
         const callback = type;
+        if (!RCTAlertManager) {
+          return;
+        }
         RCTAlertManager.alertWithArgs(
           {
             title: title || '',
             type: 'plain-text',
-            defaultValue: message,
+            defaultValue: message || '',
           },
           (id, value) => {
             callback(value);
@@ -160,6 +163,9 @@ class Alert {
         });
       }
 
+      if (!RCTAlertManager) {
+        return;
+      }
       RCTAlertManager.alertWithArgs(
         {
           title: title || '',
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -16,23 +16,12 @@ import DialogManagerAndroid, {
   type DialogOptions,
 } from '../NativeModules/specs/NativeDialogManagerAndroid';
 import RCTAlertManager from './RCTAlertManager';
-
-export type Buttons = Array<{
-  text?: string,
-  onPress?: ?Function,
-  style?: AlertButtonStyle,
-}>;
-
-type Options = {
-  cancelable?: ?boolean,
-  onDismiss?: ?Function,
-};
-
-/* 'default' | plain-text' | 'secure-text' | 'login-password' */
-type AlertType = string;
-
-/* 'default' | 'cancel' | 'destructive' */
-export type AlertButtonStyle = string;
+import {
+  type Buttons,
+  type Options,
+  type AlertType,
+  type AlertButtonStyle,
+} from './NativeAlertManager';
 
 /**
  * Launches an alert dialog with the specified title and message.
@@ -172,6 +161,7 @@ class Alert {
         },
         (id, value) => {
           const cb = callbacks[id];
+          // $FlowFixMe
           cb && cb(value);
         },
       );
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -16,12 +16,7 @@ import DialogManagerAndroid, {
   type DialogOptions,
 } from '../NativeModules/specs/NativeDialogManagerAndroid';
 import RCTAlertManager from './RCTAlertManager';
-import {
-  type Buttons,
-  type Options,
-  type AlertType,
-  type AlertButtonStyle,
-} from './NativeAlertManager';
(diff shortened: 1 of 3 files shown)


Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,41 @@
...
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+type Button = {|
+  text: string,
+  onPress: () => void,
+  style: 'default' | 'cancel' | 'destructive',
```
RSNara: Unfortunately, our Codegen doesn't support unions. 😔

Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,41 @@
...
+|};
+
+type AlertType = $Enum<{
+  default: string,
+  'plain-text': string,
+  'secure-text': string,
+  'login-password': string,
+}>;
```
RSNara: It also doesn't support enums. 😔

sasurau4:
@RSNara Thanks for reviewing. :+1: 
I replaced enums and union to string. Please review again :pray:

Libraries/Alert/RCTAlertManager.ios.js
```diff
@@ -10,6 +10,6 @@
 
 'use strict';
 
-const RCTAlertManager = require('../BatchedBridge/NativeModules').AlertManager;
+const RCTAlertManager = require('./NativeAlertManager');
```
uqmessias: @sasurau4,
Could you update its call-sites inside [Alert.js](https://github.com/facebook/react-native/blob/master/Libraries/Alert/Alert.js#L13-L14), pls?

Also, this is a minor, but should we use `import RCTAlertManager from './NativeAlertManager'` here instead? I saw many PRs replacing it to this format.

sasurau4:
@uqmessias Thanks for reviewing 👍 
I fixed pointed out. Please review agian 🙏

Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,37 @@
...
+|};
+
+export interface Spec extends TurboModule {
+  +alertWithArgs: (
+    args: Args,
+    callback: (id: number, value: string) => void,
+  ) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('AlertManager');
```
uqmessias: Since, this is an iOS only feature, please export it conditionally.

Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,37 @@
...
+ * @flow strict-local
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+type Button = {|
```
uqmessias: Should we reuse these styles inside Alert.js instead of having them separately? Something like is done for [DialogOptions](https://github.com/facebook/react-native/blob/2b5b22f02fc5df253ea63c99d02c0adde028956a/Libraries/Alert/Alert.js#L16)

Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,37 @@
...
+  text?: string,
+  onPress?: () => void,
+  style?: string,
+|};
+
+type Args = {|
```
uqmessias: Ditto

Libraries/Alert/NativeAlertManager.js
```diff
@@ -0,0 +1,37 @@
...
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+type Button = {|
+  text?: string,
+  onPress?: () => void,
+  style?: string,
```
uqmessias: Do you mind to add some comments and already preparing this types for when the codegen starts to support enums and unions, like it's done in [NativeDialogManagerAndroid](https://github.com/facebook/react-native/blob/3ec1119e08ee214306add2174f3be7725fbcebb0/Libraries/NativeModules/specs/NativeDialogManagerAndroid.js#L18|L34)?

uqmessias:
It's almost there, just left a few comments

Libraries/Alert/NativeAlertManager.js
sasurau4: Should `Args` be located at Alert.js?

uqmessias: I'm not sure, but I would keep it in this file

Libraries/Alert/Alert.js
```diff
@@ -28,18 +28,11 @@ type Options = {
...
+/* 'default' | plain-text' | 'secure-text' | 'login-password' */
+type AlertType = string;
 
-export type AlertButtonStyle = $Keys<{
-  default: string,
-  cancel: string,
-  destructive: string,
-}>;
+/* 'default' | 'cancel' | 'destructive' */
+export type AlertButtonStyle = string;
```
uqmessias: Good job!
But it would be better to keep this inside the spec file, because there is where we're defining types. We're could still exporting it here too or updating all call-sites that uses these types from here to the spec path.

What do you think about it?

sasurau4: I agree with keeping types inside the spec file. I'll move types to spec file :+1:

sasurau4:
@uqmessias Thanks for comments! I worked on it! Plz review :pray:

Libraries/Alert/Alert.js
```diff
@@ -16,12 +16,7 @@ import DialogManagerAndroid, {
   type DialogOptions,
 } from '../NativeModules/specs/NativeDialogManagerAndroid';
 import RCTAlertManager from './RCTAlertManager';
-import {
-  type Buttons,
-  type Options,
-  type AlertType,
-  type AlertButtonStyle,
-} from './NativeAlertManager';
+import {type Buttons, type Options, type AlertType} from './NativeAlertManager';
```
uqmessias: This is a minor and no need to update it, but you also can use
```javascript
import type { Buttons, Options, AlertType } from './NativeAlertManager';
```

uqmessias:
We still need @RSNara's approval

PR #24907, opened 2019-05-16
[TM] add spec for HeadlessJsTaskSupport

ericlewis:
## Summary
part of #24875

## Changelog

[General] [Added] - add TM spec for HeadlessJsTaskSupport

## Test Plan

Flow passes

### DIFF
--- a/Libraries/ReactNative/AppRegistry.js
+++ b/Libraries/ReactNative/AppRegistry.js
@@ -262,14 +262,22 @@ const AppRegistry = {
     const taskProvider = taskProviders.get(taskKey);
     if (!taskProvider) {
       console.warn(`No task registered for key ${taskKey}`);
-      NativeHeadlessJsTaskSupport.notifyTaskFinished(taskId);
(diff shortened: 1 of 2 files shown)


Libraries/ReactNative/AppRegistry.js
```diff
@@ -261,16 +262,22 @@ const AppRegistry = {
     const taskProvider = taskProviders.get(taskKey);
     if (!taskProvider) {
       console.warn(`No task registered for key ${taskKey}`);
-      NativeModules.HeadlessJsTaskSupport.notifyTaskFinished(taskId);
+      if (NativeHeadlessJsTaskSupport) {
```
rickhanlonii: Are these ifs necessary?

ericlewis: Flow complains if I don’t. Android only lib. I could attempt to make less of them?

rickhanlonii: Ah ok makes sense

PR #24909, opened 2019-05-16
[TM] Add spec for BlobModule

ericlewis:
## Summary
Part of #24875. Not sure that the id’s types are necessarily correct here…

## Changelog

[General] [Added] - Add TM spec for BlobModule

## Test Plan

Flow passes

### DIFF
--- a/Libraries/Blob/NativeBlobModule.js
+++ b/Libraries/Blob/NativeBlobModule.js
@@ -16,11 +16,11 @@ import * as TurboModuleRegistry from 'TurboModuleRegistry';
 export interface Spec extends TurboModule {
   +getConstants: () => {|BLOB_URI_SCHEME: string, BLOB_URI_HOST: ?string|};
   +addNetworkingHandler: () => void;
-  +addWebSocketHandler: (id: number | string) => void;
-  +removeWebSocketHandler: (id: number | string) => void;
-  +sendOverSocket: (blob: Object, id: number | string) => void;
-  +createFromParts: (parts: Array<Object>, blobId: number | string) => void;
-  +release: (blobId: number | string) => void;
+  +addWebSocketHandler: (id: number) => void;
+  +removeWebSocketHandler: (id: number) => void;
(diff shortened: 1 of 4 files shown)


Libraries/Blob/NativeBlobModule.js
```diff
@@ -0,0 +1,26 @@
...
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +getConstants: () => {|BLOB_URI_SCHEME: string, BLOB_URI_HOST: ?string|};
+  +addNetworkingHandler: () => void;
+  +addWebSocketHandler: (id: number | string) => void;
+  +removeWebSocketHandler: (id: number | string) => void;
+  +sendOverSocket: (blob: Object, id: number | string) => void;
+  +createFromParts: (parts: Array<Object>, blobId: number | string) => void;
+  +release: (blobId: number | string) => void;
```
RSNara: It seems like there's a mistake here. 🤔

```Java
public void addWebSocketHandler(final int id);
public void removeWebSocketHandler(final int id);
public void sendOverSocket(ReadableMap blob, int id);
public void createFromParts(ReadableArray parts, String blobId);
public void release(String blobId)
```

```ObjC
RCT_EXPORT_METHOD(addWebSocketHandler:(nonnull NSNumber *)socketID)
RCT_EXPORT_METHOD(removeWebSocketHandler:(nonnull NSNumber *)socketID)
RCT_EXPORT_METHOD(sendOverSocket:(NSDictionary *)blob socketID:(nonnull NSNumber *)socketID)
RCT_EXPORT_METHOD(createFromParts:(NSArray<NSDictionary<NSString *, id> *> *)parts withId:(NSString *)blobId)
RCT_EXPORT_METHOD(release:(NSString *)blobId)
```

ericlewis: Indeed! Will fix

ericlewis:
@RSNara flow is quite angry, i am attempting to fix

PR #24915, opened 2019-05-16
[TM] Add spec for AnimationsDebugModule

ericlewis:
## Summary
Part of #24875.

## Changelog

[General] [Added] - Add TM spec for AnimationsDebugModule

## Test Plan

Flow passes

### DIFF
--- a/Libraries/NativeModules/specs/NativeAnimationsDebugModule.js
+++ b/Libraries/NativeModules/specs/NativeAnimationsDebugModule.js
@@ -1,0 +1,24 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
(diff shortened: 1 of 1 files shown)


Libraries/NativeModules/specs/NativeAnimationsDebugModule.js
```diff
@@ -0,0 +1,24 @@
...
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  +startRecordingFps: () => void;
+  +stopRecordingFps: (animationStopTimeMs: number) => void;
+}
+
+export default (Platform.OS === 'android'
+  ? TurboModuleRegistry.getEnforcing<Spec>('AnimationsDebugModule')
+  : null);
```
uqmessias: Shouldn't it be `undefined` instead?

fkgozali: Please use `TurboModuleRegistry.get<Spec>('AnimationsDebugModule')` instead and no need to check the Platform.

PR #24916, opened 2019-05-16
[TM] Add Spec for DatePickerAndroid

pedpess:
## Summary

This PR solves part of this issue: https://github.com/facebook/react-native/issues/24875

## Changelog

[General] [Added] - TM Spec for DatePickerAndroid

## Test Plan

Flow passes.

### DIFF
--- a/Libraries/Components/DatePickerAndroid/NativeDatePickerAndroid.js
+++ b/Libraries/Components/DatePickerAndroid/NativeDatePickerAndroid.js
@@ -9,9 +9,13 @@
 
 import type {TurboModule} from 'RCTExport';
 import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {Options, DatePickerOpenAction} from './DatePickerAndroidTypes';
+import Platform from 'Platform';
 
 export interface Spec extends TurboModule {
-  +open: () => Promise<void>;
+  +open: (options: ?Options) => Promise<DatePickerOpenAction>;
 }
 
-export default TurboModuleRegistry.getEnforcing<Spec>('DatePickerAndroid');
+export default (Platform.OS === 'android'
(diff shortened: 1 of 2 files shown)


pedpess:
Hi @fkgozali :), is it normal that the e2e test fails?

ericlewis:
@pedpess we aim for it not to be broken, but if it’s failing on iOS steps then that’s known.

pedpess:
@ericlewis It's failing in `RNTester/e2e/__tests__/Touchable-test.js ` for EarlGrey, which is the e2e I assume we are using here? Then I think I just iOS related.

ericlewis:
Yep, this is a known issue. You’re fine!

cpojer:
Hey @pedpess do you mind signing the FB CLA so we can merge your change? It only takes a minute:  code.facebook.com/cla

pedpess:
Hi @cpojer , just did it :)

Libraries/Components/DatePickerAndroid/NativeDatePickerAndroid.js
```diff
@@ -0,0 +1,21 @@
...
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {Options, DatePickerOpenAction} from './DatePickerAndroidTypes';
+import Platform from 'Platform';
+
+export interface Spec extends TurboModule {
+  +open: (options: ?Options) => Promise<DatePickerOpenAction>;
```
fkgozali: Unfortunately some of the types in DatePickerAndroidTypes like `Date` isn't compatible yet with our codegen system, so for now let's modify these types to use `Object` (I'll modify it after I import to FB)

PR #24921, opened 2019-05-16
[TM] Add Spec for ImageEditor

yjose:
## Summary
This PR solves part of this issue: #24875
## Changelog

[General] [Added] - TM Spec for ImageEditor

## Test Plan
Flow passes.

### DIFF
--- a/Libraries/Image/NativeImageEditor.js
+++ b/Libraries/Image/NativeImageEditor.js
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+type cropOptions = {
+  offset: {|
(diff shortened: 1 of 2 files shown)


Libraries/Image/NativeImageEditor.js
```diff
@@ -0,0 +1,48 @@
...
+export interface Spec extends TurboModule {
+  +cropImage: (
+    uri: string,
+    options: cropOptions,
+    success: (uri: string) => void,
+    error: (error: string) => void,
+  ) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('NativeImageEditor');
```
ericlewis: this should be `ImageEditingManager` instead

Libraries/Image/NativeImageEditor.js
```diff
@@ -0,0 +1,48 @@
...
+    width: number,
+    height: number,
+  |},
+
+  displaySize?: ?{|
+    width: number,
+    height: number,
+  |},
+
+  resizeMode?: ?$Enum<{
```
ericlewis: are these optional because android doesn't have them? if so, group and mark: `// iOS only`

Libraries/Image/NativeImageEditor.js
yjose: Got it ! do I need to update call-sites too?

PR #24922, opened 2019-05-16
[TM] add spec for RedBox

ericlewis:
## Summary
part of #24875.

## Changelog

[General] [Added] - add TM spec for RedBox

## Test Plan

Flow passes

### DIFF
--- a/Libraries/NativeModules/specs/NativeRedBox.js
+++ b/Libraries/NativeModules/specs/NativeRedBox.js
@@ -1,0 +1,24 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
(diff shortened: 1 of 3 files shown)


Libraries/NativeModules/specs/NativeRedBox.js
```diff
@@ -0,0 +1,24 @@
...
+import {Platform} from 'react-native';
+
+export interface Spec extends TurboModule {
+  +setExtraData: (extraData: Object, identifier: string) => void;
+  +dismiss: () => void;
+}
+
+export default (Platform.OS === 'ios'
+  ? TurboModuleRegistry.getEnforcing<Spec>('RedBox')
+  : null);
```
uqmessias: Shouldn't we use `undefined` here instead?

ericlewis: hmm good question, I didn't see anything about using it specifically. But it would be nice to have a consensus.

uqmessias: I saw some of the specific platforms PRs using `undefined` and others using `null`.
I don't have a strong opinion on that, but my preference is to go with `undefined` what denotes that this Native Module "was not defined" for this platform.

Anyway, I can update my PRs if we decide to go with `null`.

ericlewis: We should actually be using get not get enforcing

fkgozali: Yeah, I'll update it before I import/land to FB

PR #24928, opened 2019-05-17
[TM] add spec for PlatformConstants

sattaman:
## Summary

part of #24875.

## Changelog

[General] [Added] - add TM spec for PlatformConstants

## Test Plan

Tests + flow pass

### DIFF
--- a/Libraries/Core/ReactNativeVersionCheck.js
+++ b/Libraries/Core/ReactNativeVersionCheck.js
@@ -9,7 +9,7 @@
  */
 'use strict';
 
-const {PlatformConstants} = require('../BatchedBridge/NativeModules');
+import NativePlatformConstants from '../Utilities/NativePlatformConstants';
 const ReactNativeVersion = require('./ReactNativeVersion');
 
 /**
@@ -22,11 +22,12 @@ const ReactNativeVersion = require('./ReactNativeVersion');
  * and rely on its existence as a separate module.
  */
 exports.checkVersions = function checkVersions(): void {
-  if (!PlatformConstants) {
+  if (!NativePlatformConstants) {
     return;
   }
 
-  const nativeVersion = PlatformConstants.reactNativeVersion;
+  const nativeVersion = NativePlatformConstants.getConstants()
+    .reactNativeVersion;
   if (
     ReactNativeVersion.version.major !== nativeVersion.major ||
     ReactNativeVersion.version.minor !== nativeVersion.minor
@@ -46,6 +47,8 @@ exports.checkVersions = function checkVersions(): void {
 function _formatVersion(version): string {
   return (
     `${version.major}.${version.minor}.${version.patch}` +
-    (version.prerelease !== null ? `-${version.prerelease}` : '')
+    (version.prerelease !== null && version.prerelease !== undefined
+      ? `-${version.prerelease}`
+      : '')
   );
 }
--- a/Libraries/Core/ReactNativeVersionCheck.js
+++ b/Libraries/Core/ReactNativeVersionCheck.js
@@ -47,8 +47,7 @@ exports.checkVersions = function checkVersions(): void {
 function _formatVersion(version): string {
   return (
     `${version.major}.${version.minor}.${version.patch}` +
-    (version.prerelease !== null && version.prerelease !== undefined
-      ? `-${version.prerelease}`
-      : '')
+    // eslint-disable-next-line eqeqeq
+    (version.prerelease != undefined ? `-${version.prerelease}` : '')
   );
 }
--- a/Libraries/Utilities/NativePlatformConstants.js
+++ b/Libraries/Utilities/NativePlatformConstants.js
@@ -12,8 +12,10 @@
 
 import type {TurboModule} from 'RCTExport';
 import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import Platform from 'Platform';
 
 export interface Spec extends TurboModule {
+  // Exported methods.
   +getConstants: () => {|
     // Common interface
     isTesting: boolean,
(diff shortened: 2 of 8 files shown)


Libraries/Utilities/Platform.android.js
```diff
@@ -20,19 +19,17 @@ export type PlatformSelectSpec<A, D> = {
 const Platform = {
   OS: 'android',
   get Version() {
-    const constants = NativeModules.PlatformConstants;
-    return constants && constants.Version;
+    const { Version } = NativePlatformConstants.getConstants();
+    return Version || 0;
```
sattaman: I wasn't sure if there was a better was of forcing the value going from here to be non nullable

sattaman:
I didn't realise when picking this up @jarvisluong It looks like it might it might have been better if I'd picked up the ios part of PlatformConstants too since its all the same module .

jarvisluong:
@sattaman right, that is fine too!

Libraries/Core/ReactNativeVersionCheck.js
```diff
@@ -46,6 +47,8 @@ exports.checkVersions = function checkVersions(): void {
 function _formatVersion(version): string {
   return (
     `${version.major}.${version.minor}.${version.patch}` +
-    (version.prerelease !== null ? `-${version.prerelease}` : '')
+    (version.prerelease !== null && version.prerelease !== undefined
```
uqmessias: In order to keep it less verbose, what do you think of checking for `version.prerelease != undefined` instead? By doing that, we can check if it's different from `null` and `undefined`

sattaman: Sure, I guess `version.prerelease &&` will serve the same purpose.. I wasn't sure if this would have be frowned upon

uqmessias: Unfortunately no.
`version.prerelease` will be very different from simply `version.prerelease != undefined` because:

```javascript
const cond1 = version.prerelease;
const cond2 = !!version.prerelease;
const cond3 = (version.prerelease !== 0 && version.prerelease !== undefined && version.prerelease !== null && version.prerelease !== '' && version.prerelease !== false);
// cond1 === cond2 === cond3

const cond4 = version.prerelease != undefined;
const cond5 = version.prerelease != null;
const cond6 = version.prerelease !== undefined && version.prerelease !== null;
// cond4 === cond5 === cond6
```

uqmessias: This is a minor issue. In the end, it will be a simplification of `version.prerelease !== null && version.prerelease !== undefined`

RNTester/js/TouchableExample.js
```diff
@@ -18,17 +18,15 @@ const {
   Text,
   TouchableHighlight,
   TouchableOpacity,
-  NativeModules,
+  NativePlatformConstants,
```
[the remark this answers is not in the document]
sattaman: I'm wondering if I'm doing the right thing updating this here?

Libraries/Core/ReactNativeVersionCheck.js
sattaman: I've updated, many thanks for taking the time to explain. I'm finding I've got 152 tests failing locally, and I'm not sure why, also flow isn't seeming to like what I've done. Is there anything obvious wrong with what I've done?

RNTester/js/TouchableExample.js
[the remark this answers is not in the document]
sattaman: I've reverted this as presuming not

yarn.lock
```diff
@@ -1163,6 +1163,7 @@
 "@reactions/component@^2.0.2":
   version "2.0.2"
   resolved "https://registry.yarnpkg.com/@reactions/component/-/component-2.0.2.tgz#40f8c1c2c37baabe57a0c944edb9310dc1ec6642"
+  integrity sha1-QPjBwsN7qr5XoMlE7bkxDcHsZkI=
```
fkgozali: Can you undo this one? seems unrelated to the change

Libraries/Utilities/NativePlatformConstants.js
```diff
@@ -0,0 +1,51 @@
...
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import Platform from 'Platform';
+
+export interface Spec extends TurboModule {
+  // Exported methods.
+  +getConstants: () => {|
+    // Common interface
+    isTesting: boolean,
+    reactNativeVersion: {
```
fkgozali: let this be strict object as well:

```
reactNativeVersion: {|
  ...
|},
```

Libraries/Utilities/NativePlatformConstants.js
```diff
@@ -0,0 +1,51 @@
...
+
+    // iOS only
+    forceTouchAvailable?: boolean,
+    osVersion?: number,
+    systemName?: string,
+    interfaceIdiom?: 'phone' | 'pad' | 'tv' | 'carplay' | 'unknown',
+  |};
+
+  // Android only
+  +getAndroidID?: () => string;
```
fkgozali: make this required (no ?)

Libraries/Utilities/NativePlatformConstants.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+
+    // Android only
+    Version?: number,
+    Release?: string,
+    Serial?: string,
+    Fingerprint?: string,
+    Model?: string,
+    ServerHost?: string,
+    uiMode?: 'tv' | 'car' | 'desk' | 'watch' | 'normal' | 'unknown',
```
RSNara: Our Codegen doesn't support enums. Can you change this to `string`?

Libraries/Utilities/NativePlatformConstants.js
RSNara: Shouldn't this be optional because iOS doesn't expose this method?

fkgozali: Optional or not, codegen still generates the method, so the function existence check is now meaningless. Might as well make it required.

sattaman:
@fkgozali I've updated as requested

PR #24936, opened 2019-05-17
[CI] Consolidate test_end_to_end back into other jobs

hramos:
## Summary

Move the iOS end-to-end tests to `test_ios` and the JavaScript end-to-end tests to `test_javascript`. The remaining Detox iOS tests form part of the now-renamed `test_detox_ios` job.

## Changelog

[Internal] [Removed] - Consolidate jobs

## Test Plan

Circle CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -487,15 +487,15 @@ jobs:
       - save-cache: *save-yarn-cache
 
       # Xcode build
-      - run:
-          name: Build app for Detox iOS End-to-End Tests
-          command: yarn run build-ios-e2e
+      # - run:
+      #     name: Build app for Detox iOS End-to-End Tests
+      #     command: yarn run build-ios-e2e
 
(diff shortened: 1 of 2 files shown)


hramos:
https://github.com/facebook/react-native/pull/24953 is an effort to fix the Detox e2e tests. I've opened https://github.com/facebook/react-native/pull/24962 to disable the iOS e2e tests that keep `test_end_to_end` from going green.

I am not sure moving the e2e tests back into `test_ios` and `test_javascript` is the right move at this point. The problem with moving the iOS e2e tests into `test_ios` is that any JavaScript issue that prevents the bundle from being built will be reflected as a `test_ios` failure.

Once the two above PRs are merged, I might consider splitting the tests from `test_end_to_end` into `test_ios_e2e` and `test_javascript_e2e` and having those depend on `test_javascript`.

PR #24939, opened 2019-05-18
[Android][Keyboard] Return endCoordinates for keyboardDidHide keyboard event

nossbigg:
## Summary

This pull request enhances the Keyboard API event emitter for Android upon `keyboardDidHide` by returning a `KeyboardEvent` with a meaningful `endCoordinates` property (instead of emitting a `null` as of current implementation). This change standardizes the `keyboardDidHide` keyboard event emission across both iOS and Android, which makes it easier for developers to use the API.

In particular, the semantics of `endCoordinates` emitted during the `keyboardDidHide` event on Android will match nicely with semantics of the same event emitted on iOS:
- `screenY` will be height of the screen, as that the keyboard has collapsed to the bottom of the screen
- `screenX` will be 0, as the keyboard will always be flush to the sides of the screen
- `height` will be 0, as the keyboard has fully collapsed
- `width` will be the width of the screen, as the keyboard will always extend to the width of the screen

Also, the flowtypes for `KeyboardEvent` have been further improved and are more explicit to better highlight the different object shapes (see `IOSKeyboardEvent` and `AndroidKeyboardEvent`) depending on the platform.

## Changelog

[Android] [Added] - Return endCoordinates for keyboardDidHide keyboard event

## Test Plan

1. Listen to the `keyboardDidShow` and `keyboardDidHide` keyboard events, and log out the event details via `console.log`
1. Launch the App in both Android and iOS
1. Enter input in a text field
1. Close the keyboard
1. Compare the keyboard event details across both platforms. Most importantly, the `endCoordinates` property on both event details should share the same semantic meaning.

**`keyboardDidHide` event on iPhone 8**
keyboardDidHide ios

**`keyboardDidHide` event on Google Nexus 5X**
keyboardDidHide android

### DIFF
--- a/Libraries/Components/Keyboard/Keyboard.js
+++ b/Libraries/Components/Keyboard/Keyboard.js
@@ -33,19 +33,26 @@ export type KeyboardEventEasing =
   | 'linear'
   | 'keyboard';
 
-type ScreenRect = $ReadOnly<{|
+type KeyboardEventCoordinates = $ReadOnly<{|
   screenX: number,
   screenY: number,
   width: number,
   height: number,
 |}>;
 
-export type KeyboardEvent = $ReadOnly<{|
-  duration: number,
+export type KeyboardEvent = AndroidKeyboardEvent | IOSKeyboardEvent;
+
+export type AndroidKeyboardEvent = $ReadOnly<{|
+  easing: 'keyboard',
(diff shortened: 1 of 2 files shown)


nossbigg:
Closing pull request to fix type issues, will create new pull request once it's done.

PR #24947, opened 2019-05-19
[Android][Keyboard] Return endCoordinates for keyboardDidHide keyboard event

nossbigg:
## Summary

This pull request enhances the Keyboard API event emitter for Android upon `keyboardDidHide` by returning a `KeyboardEvent` with a meaningful `endCoordinates` property (instead of emitting a null as of current implementation). This change standardizes the `keyboardDidHide` keyboard event emission across both iOS and Android, which makes it easier for developers to use the API.

In particular, the semantics of `endCoordinates` emitted during the `keyboardDidHide` event on Android will match nicely with semantics of the same event emitted on iOS:
- `screenY` will be height of the screen, as that the keyboard has collapsed to the bottom of the screen
- `screenX` will be 0, as the keyboard will always be flush to the sides of the screen
- `height` will be 0, as the keyboard has fully collapsed
- `width` will be the width of the screen, as the keyboard will always extend to the width of the screen

Also, the flowtypes for `KeyboardEvent` have been further improved and are more explicit to better highlight the different object shapes (see `IOSKeyboardEvent` and `AndroidKeyboardEvent`) depending on the platform.

## Changelog

[Android] [Added] - Return endCoordinates for keyboardDidHide keyboard event

## Test Plan

1. Listen to the `keyboardDidShow` and `keyboardDidHide` keyboard events, and log out the event details via `console.log`
1. Launch the App in both Android and iOS
1. Enter input in a text field
1. Close the keyboard
1. Compare the keyboard event details across both platforms. The `endCoordinates` property on both event details should share the same semantic meaning.

**`keyboardDidHide` event on iPhone 8**
image

**`keyboardDidHide` event on Nexus 5X**
image

### DIFF
--- a/Libraries/Components/Keyboard/Keyboard.js
+++ b/Libraries/Components/Keyboard/Keyboard.js
@@ -33,18 +33,32 @@ export type KeyboardEventEasing =
   | 'linear'
   | 'keyboard';
 
-type ScreenRect = $ReadOnly<{|
+export type KeyboardEventCoordinates = $ReadOnly<{|
   screenX: number,
   screenY: number,
   width: number,
   height: number,
 |}>;
 
-export type KeyboardEvent = $ReadOnly<{|
-  duration: number,
+export type KeyboardEvent = AndroidKeyboardEvent | IOSKeyboardEvent;
+
+type BaseKeyboardEvent = {|
+  platform: 'android' | 'ios',
   easing: KeyboardEventEasing,
-  endCoordinates: ScreenRect,
-  startCoordinates: ScreenRect,
+  endCoordinates: KeyboardEventCoordinates,
+|};
+
+export type AndroidKeyboardEvent = $ReadOnly<{|
+  ...BaseKeyboardEvent,
+  platform: 'android',
+  easing: 'keyboard',
+|}>;
+
+export type IOSKeyboardEvent = $ReadOnly<{|
+  ...BaseKeyboardEvent,
+  platform: 'ios',
+  duration: number,
+  startCoordinates: KeyboardEventCoordinates,
   isEventFromThisApp: boolean,
 |}>;
 
--- a/RNTester/js/RNTesterList.android.js
+++ b/RNTester/js/RNTesterList.android.js
@@ -34,6 +34,10 @@ const ComponentExamples: Array<RNTesterExample> = [
     module: require('./ImageExample'),
   },
   {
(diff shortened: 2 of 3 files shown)


nossbigg:
@cpojer, i added the 'platform' property as a trade-off to having an elegant `KeyboardEvent` flow type definition. I chose to prioritize the following type definition (which is a disjoint union):
```javascript
type KeyboardEvent = IOSKeyboardEvent | AndroidKeyboardEvent
```
However, this change alone this results in some flowtypes issues, namely that you cannot destructure the 'duration' property of an event of type `KeyboardEvent`. Thus, I added the 'platform' property to refine the event type (ie. disambiguate the actual event type) at the `flow` level; this allows me to make minimal changes to existing code.

The alternative to this approach is to write code to do the event type disambiguation at runtime.

cpojer:
Thanks for the explanation. I prefer not merging it like it is because we shouldn't change what we send over the bridge just to make Flow happy. Do you mind getting rid of all the platform related stuff and pushing it to this PR? Then I can take a look at the Flow error :)

nossbigg:
@cpojer ok got rid of all the 'platform' property passing over the bridge. 

As it is, there are flowtypes errors at `Keyboard.js` and `KeyboardAvoidingView.js`. I tried forcibly reassigning the type of `event` to another type (eg. `IOSKeyboardEvent`) by doing so: 
```javascript
const evt: IOSKeyboardEvent = ((event: any):IOSKeyboardEvent);
```
but then I get a Flow(LintError-unclear-type).

Happy to hear alternatives to solving this flowtype error issue :) (thought about using `// $FlowFixMe`, but it probably isn't the best to reach for that as the first option haha 😅)

cpojer:
I had to revert the duration and for now I had to put it back on the main flow type again.

nossbigg:
@cpojer sure, any thing to do on my side to push this?

nossbigg:
@cpojer does it make sense to assign a default value to the `duration` property of an `AndroidKeyboardEvent`? Feels a bit weird to me that the type definition states that the property is supposed to exist when it doesn't exist when an android keyboard event is emitted.

A low hanging fruit would be to hardcode a 0 value for each `AndroidKeyboardEvent`, and it won't be a breaking change (at least for components relying on it within the codebase).

cpojer:
Yeah that works! Wanna send a PR for that?

Issue #23512, opened 2019-02-18
RCTFont maps UltraBold font to UIFontWeightBold font weight


dat-giggedin:
## 🐛 Bug Report
Cannot display correct GillSans-Bold font. It uses GillSans-UltraBold instead 
<img width="137" alt="screen shot 2019-02-18 at 10 15 06 pm" src="https://user-images.githubusercontent.com/47650220/52960138-af5c8f80-33ca-11e9-9ae9-06d4d7b7b2b4.png">

## To Reproduce
Add text to the screen `<Text style={{fontFamily: 'GillSans-Bold'}}></Text>`

## Expected Behavior
Display correct font 
<img width="163" alt="screen shot 2019-02-18 at 10 11 33 pm" src="https://user-images.githubusercontent.com/47650220/52959962-39582880-33ca-11e9-8ecd-d2351f4f4bcf.png">

## Code Example
`<Text style={{fontFamily: 'GillSans-Bold'}}></Text>`

## Environment
```
React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 1.18 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 11.6.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.7.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 25, 26, 28
        Build Tools: 28.0.3
        System Images: android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5199772
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.58.4 => 0.58.4 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
```

[...]

PR #24948, opened 2019-05-19
[iOS] Add ultrabold pairs for font weight

zhongwuzw:
## Summary

Add `ultrabold` map of font weight, if not, `ultrabold` would map to `bold`.
Fixes https://github.com/facebook/react-native/issues/23512.

## Changelog

[iOS] [Fixed] - Add ultrabold pairs for font weight

## Test Plan
After:

`<Text style={{fontFamily: 'GillSans-UltraBold'}}>Hello</Text>`
image

`<Text style={{fontFamily: 'GillSans-Bold'}}>Hello</Text>`
image

Before:

`<Text style={{fontFamily: 'GillSans-UltraBold'}}>Hello</Text>`
image

`<Text style={{fontFamily: 'GillSans-Bold'}}>Hello</Text>`
image

### DIFF
--- a/React/Views/RCTFont.mm
+++ b/React/Views/RCTFont.mm
@@ -32,6 +32,7 @@ static RCTFontWeight weightOfFont(UIFont *font)
       @"semibold",
       @"demibold",
       @"extrabold",
+      @"ultrabold",
       @"bold",
       @"heavy",
       @"black"
@@ -45,6 +46,7 @@ static RCTFontWeight weightOfFont(UIFont *font)
       @(UIFontWeightMedium),
       @(UIFontWeightSemibold),
       @(UIFontWeightSemibold),
+      @(UIFontWeightHeavy),
       @(UIFontWeightHeavy),
       @(UIFontWeightBold),
       @(UIFontWeightHeavy),


shergin:
Just curious, does it work in Fabric (without the fix)?
In Fabric we use:
```
static UIFontWeight RCTGetFontWeight(UIFont *font) {
  NSDictionary *traits =
      [font.fontDescriptor objectForKey:UIFontDescriptorTraitsAttribute];
  return [traits[UIFontWeightTrait] doubleValue];
}
```
Does it sufficient? Why should we parse the font name in the first place?

zhongwuzw:
> Just curious, does it work in Fabric (without the fix)?

🤔  After we fixes https://github.com/facebook/react-native/pull/24966, it works in this case.

> In Fabric we use:
> 
> ```
> static UIFontWeight RCTGetFontWeight(UIFont *font) {
>   NSDictionary *traits =
>       [font.fontDescriptor objectForKey:UIFontDescriptorTraitsAttribute];
>   return [traits[UIFontWeightTrait] doubleValue];
> }
> ```
> 
> Does it sufficient? Why should we parse the font name in the first place?

From the discussion in https://github.com/facebook/react-native/pull/15162, seems it's not sufficient, because there has inconsistency for some fonts to parse weight, so we parse the font name in the first place. Should we also add that in Fabric? 🤔

shergin:
> inconsistency for some fonts to parse weight

Do you mean inconsistency between how iOS interprets the fonts and how classic RN does it? Or which kind of inconsistency?

zhongwuzw:
@shergin Yeah, based on #15162, they said Roboto weight 200/300 both resolve to thin, but they should resolve to thin and light respectively.

PR #24949, opened 2019-05-19
Update Modal.js

Norfeldt:
Change default background color (when `transparent` prop is set to `false`) from `white` to semitransparent black.

#18398 - https://github.com/facebook/react-native/issues/18398

## Changelog

[CATEGORY] [TYPE] - Message

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -220,7 +220,7 @@ class Modal extends React.Component<Props> {
     }
 
     const containerStyles = {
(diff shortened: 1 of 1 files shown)


cpojer:
When the transparent prop is off, then the color used shouldn't be using transparency so I'm not sure whether the motivation for this change makes sense. If you'd like us to consider this, could you go into more detail about what prompted this change and what this looks like before/after on devices?

Norfeldt:
This PR was made in too much hast - sorry about that.


Yours Sincerely
*Lasse Nørfeldt*

Issue #18472, opened 2018-03-20
iOS main.jsbundle will not create automatically or update without manual react-native bundle


nickmccally:
- [x] I have reviewed the [documentation](https://facebook.github.io/react-native)
- [X] I have searched [existing issues](https://github.com/facebook/react-native/issues)
- [X] I am using the [latest React Native version](https://github.com/facebook/react-native/releases)

React native is not creating a main.jsbundle for ios automatically. One must manually be created, even though the scheme set as release.  The code used to create a bundle manually:
`react-native bundle --entry-file index.js --platform ios --dev false --bundle-output ios/main.jsbundle --assets-dest ios`

**Update** this exact issue is happening on android now for us as well during our Jenkins build process. We are able to manually bundle using: `react-native bundle --entry-file index.js --platform android --dev false --bundle-output android/main.jsbundle --assets-dest android`

No errors occur when this is run, and a bundle is created/updated. From this point forward no bundle updates occur. The following is in **Bundle React Native code and images**:
`export NODE_BINARY=node
../node_modules/react-native/scripts/react-native-xcode.sh`

## Environment

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 8.9.4
  Yarn: 1.3.2
  npm: 5.7.1
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.3.0-alpha.1 => 16.3.0-alpha.2
  react-native: 0.54.2 => 0.54.2

## Steps to Reproduce

I have been experiencing this with an existing project. I even created a new react native project, and opened the xcode project. It shows main.jsbundle as a missing file. I modified the project scheme to Release, and from there -> Build. No main.jsbundle was created in the iOS project folder. One can manually be created with the code above without an error message.

## Expected Behavior

A new main.jsbundle would update on build when the following is run: build from xcode OR `react-native run-ios --configuration Release` 

## Actual Behavior

No main.jsbundle is created/updated

nickmccally:
We still have to manually create our bundle for iOS, Android bundle is working fine.

gorjuspixels:
We had the same issue, the file was checked into the iOS project. We had to remove it & add a new build step `Bundle React Native code and images`:

```
export NODE_BINARY=node
../node_modules/react-native/scripts/react-native-xcode.sh
```

jsina:
[...]

Issue #15432, opened 2017-08-09
main.jsbundle does not exist. This must be a bug with + echo 'React Native


hanishsharma:
Hello team,

I am getting this error while running my xcode project.

main.jsbundle does not exist. This must be a bug with + echo 'React Native.

Please help.

Thanks.

IbraheemAlSaady:
+1

scottyantipa:
Not sure if this is the answer, but this issue went away for me when I moved my type declaration packages from `devDependencies` to `dependencies` in package.json.

The packages were @types/react, @types/react-native, @types/react-navigation.

lolobosse:
@scottyantipa Can you elaborate on that? What packages? and what `type declaration`? That would help me a lot! 😄

scottyantipa:
@lolobosse I updated my comment above

lolobosse:
@scottyantipa I found my error, it was a weird `\` in a string. 

I had to debug the bundler and I saw the file crashing.

@ReactNative Team: that would be great that if the `--verbose` option is activated, you deliver more logs than just `SyntaxError` on the bundler 😄

mstrchrstphr:
For this issue was caused because the `index.ios.js` file was not located in the root directory of the repo. I had to modify the `Build Phases` -> `Bundle React Native code and images` script in Xcode to point it to the correct directory.

Original (what react native generated):
```

xavieramoros:
I had a similar error after doing a pull of a repo upgraded to 0.48.4 (someone else did the RN upgrading). 

Doing 
`rm -rf ./node_modules && npm install
`
showed this error, but doing: 
`rm -rf ./node_modules && yarn
`
actually fixed it. So I guess there are subtle differences on how yarn and git handle install scripts.

johndanek:
@lolobosse What string was the / in? React Native code, or somewhere in XCode?

jpapillon:
@johndanek The `node_modules` directory is always at the root of your project (React Native code).

JamesDorrian:
Tried all of above, problem still persisting. Why was this closed @hanishsharma ?

YasirPanjsheri:
Hi , I am using react-native 0.49,  I get this error on device only , it works fine on Simulator. I tried many things but none worked. 

![img-0652](https://user-images.githubusercontent.com/12644122/32533706-934d8e6e-c46f-11e7-8b23-bd39b0a36ea3.PNG)

JamesDorrian:
[...]

PR #24953, opened 2019-05-19
[CI] Fix e2e detox build step

nossbigg:
## Summary

Fixes the e2e detox build step by manually overriding `PROJECT_ROOT` for the project's JS bundle build step. 

After seeing [quite](https://github.com/facebook/react-native/issues/18472#issuecomment-428996279) [a](https://github.com/facebook/react-native/issues/18472#issuecomment-450485004) [few](https://github.com/facebook/react-native/issues/15432#issuecomment-417310668) [comments](https://stackoverflow.com/a/49506072) suggesting running some variant of the `react-native bundle` manually on your own so as to build the jsbundle required as part of the build step for RNTester project...

The main issue I found was that the working directory in which `react-native-xcode.sh` executed the CLI bundle step was not correct. The `PROJECT_ROOT` was not resolving to the root of the `react-native` project directory, but instead to something to the effect of `/Users/gibson.cheng/IdeaProjects/react-native/../..` - of which of course the build step would not be able to find the `react-native` project to run the build against.

I'm not sure if new generated `react-native` projects require this manual override, so I only applied it to the RNTester project. Reviewers are welcome to correct my understanding and solutioning to this matter :)

@hramos, if this works, perhaps there would not be a need to push through with #24936. Also, this  contributes to #23561.

## Changelog

[Internal] [Fixed] - Fix build-ios-e2e build step

## Test Plan

1. `build-ios-e2e` should work now

### DIFF
--- a/RNTester/RNTester.xcodeproj/project.pbxproj
+++ b/RNTester/RNTester.xcodeproj/project.pbxproj
@@ -1107,7 +1107,7 @@
 				13B07F871A680F5B00A75B9A /* Sources */,
 				13B07F8C1A680F5B00A75B9A /* Frameworks */,
 				13B07F8E1A680F5B00A75B9A /* Resources */,
-				68CD48B71D2BCB2C007E06A9 /* Run Script */,
+				68CD48B71D2BCB2C007E06A9 /* Bundle React Native code and images */,
 				EDEBC7DE214C68E400DD5AC8 /* Embed Frameworks */,
 			);
 			buildRules = (
@@ -1707,19 +1707,19 @@
 			shellPath = /bin/sh;
 			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js";
 		};
-		68CD48B71D2BCB2C007E06A9 /* Run Script */ = {
+		68CD48B71D2BCB2C007E06A9 /* Bundle React Native code and images */ = {
 			isa = PBXShellScriptBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
 			);
 			inputPaths = (
 			);
-			name = "Run Script";
+			name = "Bundle React Native code and images";
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js $SRCROOT/..\n";
 		};
 /* End PBXShellScriptBuildPhase section */
 
--- a/RNTester/RNTester.xcodeproj/project.pbxproj
+++ b/RNTester/RNTester.xcodeproj/project.pbxproj
@@ -1719,7 +1719,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js $SRCROOT/..\n";
+			shellScript = "export NODE_BINARY=node\nPROJECT_ROOT_OVERRIDE=$SRCROOT/.. $SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
 		};
 /* End PBXShellScriptBuildPhase section */
 
--- a/RNTester/RNTester.xcodeproj/project.pbxproj
+++ b/RNTester/RNTester.xcodeproj/project.pbxproj
@@ -1719,7 +1719,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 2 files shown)


hramos:
Thanks for the PR! I'll update mine and avoid disabling the Detox tests.

RNTester/RNTester.xcodeproj/project.pbxproj
```diff
@@ -1707,19 +1707,19 @@
...
 			inputPaths = (
 			);
-			name = "Run Script";
+			name = "Bundle React Native code and images";
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js $SRCROOT/..\n";
```
hramos:
```suggestion
			shellScript = "export NODE_BINARY=node\nPROJECT_ROOT=$SRCROOT/.. $SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
```

hramos: Would this work, in place of adding a second parameter to the script? It would remove the need for $PROJECT_ROOT_OVERRIDE.

nossbigg: Supplying PROJECT_ROOT via env var whilst starting the `react-native-xcode.sh` won't override the PROJECT_ROOT var inside the script, as the var is constructed inside the shell script (with no existing logic to accept overrides externally).

https://github.com/facebook/react-native/blob/fb8a7c3cc51f2b68f5d82266ea5f6525ab396310/scripts/react-native-xcode.sh#L55-L60

I agree with you that supplying the PROJECT_ROOT_OVERRIDE via script arg is not as elegant as supplying it via env var. In the case of the latter approach, the initialization of the script will look something like:

```bash
PROJECT_ROOT_OVERRIDE=$SRCROOT/.. $SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js;
```

If this is workable, i'll go ahead with it 👍

RNTester/RNTester.xcodeproj/project.pbxproj
```diff
@@ -1707,19 +1707,19 @@
...
 			inputPaths = (
 			);
-			name = "Run Script";
+			name = "Bundle React Native code and images";
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\n$SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\nPROJECT_ROOT_OVERRIDE=$SRCROOT/.. $SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
```
hramos:
```suggestion
			shellScript = "export NODE_BINARY=node\nPROJECT_ROOT=$SRCROOT/.. $SRCROOT/../scripts/react-native-xcode.sh RNTester/js/RNTesterApp.ios.js\n";
```

scripts/react-native-xcode.sh
```diff
@@ -57,6 +57,10 @@ REACT_NATIVE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 # in node_modules.
 PROJECT_ROOT=${PROJECT_ROOT:-"$REACT_NATIVE_DIR/../.."}
 
+if [ "$PROJECT_ROOT_OVERRIDE" != "" ]; then
```
hramos:
```suggestion

```

scripts/react-native-xcode.sh
```diff
@@ -57,6 +57,10 @@ REACT_NATIVE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
...
+  PROJECT_ROOT=$PROJECT_ROOT_OVERRIDE
```
hramos:
```suggestion
```

scripts/react-native-xcode.sh
```diff
@@ -57,6 +57,10 @@ REACT_NATIVE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
...
+fi
```
hramos:
```suggestion
```

scripts/react-native-xcode.sh
```diff
@@ -57,6 +57,10 @@ REACT_NATIVE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
...
+
```
hramos:
```suggestion
```

RNTester/RNTester.xcodeproj/project.pbxproj
hramos: My understanding is it should be overridable. The following line sets PROJECT_ROOT to the current value of PROJECT_ROOT, unless it's nil, in which case it is set to "$REACT_NATIVE_DIR/../..".

```
PROJECT_ROOT=${PROJECT_ROOT:-"$REACT_NATIVE_DIR/../.."} 
```

I verified that's the case in https://circleci.com/gh/hramos/react-native/5739
. I'll apply those changes and land.

nossbigg:
Thanks for the merge, @hramos! :)

PR #24956, opened 2019-05-19
[Lean Core] Remove all Cameraroll references

bartolkaruza:
## Summary

This PR removes all references to the Cameraroll native module. The module has been moved to a react-native-community repo and is maintained there. The deprecation notice has been on the master branch for some time now and the path to including Cameraroll from the community repo versus the core is safe and clear.

This also removes the ImageEditing example, because it uses the Cameraroll and it is already moved to the ImageEditor community repo here: https://github.com/react-native-community/react-native-image-editor/blob/master/example/src/App.js

It also removes the Cameraroll part of the XHRExample, leaving only text form fields for that example.

## Changelog

[Lean Core] [cleanup] - Removed the Cameraroll native module from React Native (now lives at https://github.com/react-native-community/react-native-cameraroll).

## Test Plan

Tested by building and running the RNTester app for both Android and iOS to make sure the builds succeed and all the example references are cleanly removed; no issues found.


bartolkaruza:
@cpojer You said you need to do this from FB side, but this might help you find all occurrences. Feel free to close it if it doesn't help or you already got everything out of it.

cpojer:
Yap, exactly. Thanks for doing all this work, I'll mirror my internal work based on your PR :)

Issue #23463, opened 2019-02-14
Error closing multiples opened modals


BrodaNoel:
## 🐛 Bug Report
When a Modal A, opens a Modal B, if we close the Modal A, the Modal B is gonna be closed, but the Modal A still there, frozen. (live examples added).
This issue is only happening on iOS. Android is working well.

This issue was reported on Expo repo: https://github.com/expo/expo/issues/3473

## To Reproduce
1. Open https://snack.expo.io/HJQ_fn34N
2. Open the app in Android (we are gonna see it working well at first)
3. The App is gonna to open a Red Modal, automatically.
4. Quickly, Tap on "Show Modal Blue" (so, the Red Modal is gonna open a Blue Modal).
5. Wait for 10 seconds.
6. The parent "App" is gonna close the Red Modal. So, Blue modal is gonna be closed as well, due to the Blue Modal is a Red Modal children.

Now, let's see the error on iOS

1. Open the same Snack
2. Open the app in iOS
3. The App is gonna to open a Red Modal, automatically.
4. Quickly, Tap on "Show Modal Blue" (so, the Red Modal is gonna open a Blue Modal).
5. Wait for 10 seconds.
6. The parent "App" is gonna close the Red Modal. **But**, the Blue Modal is gonna be closed but the Red Modal still there, broken. Kind of froze.

## Expected Behavior
The Android behavior (both modals should be closed).

## Current Behavior
iOS only close the Blue Modal (the second one) and keeps the Red Modal (the first one) opened.

## Code Example
https://snack.expo.io/HJQ_fn34N

## Environment
**Expo environment**
```
Expo CLI 2.7.0 environment info:
    System:
      OS: macOS 10.14.2
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 9.10.1 - ~/.nvm/versions/node/v9.10.1/bin/node
      Yarn: 1.13.0 - ~/.yarn/bin/yarn
      npm: 6.7.0 - ~/.nvm/versions/node/v9.10.1/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    IDEs:
      Android Studio: 3.2 AI-181.5540.7.32.5056338
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^32.0.0 => 32.0.0
      react: 16.5.0 => 16.5.0
      react-native: https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz => 0.57.1
[...]

PR #24959, opened 2019-05-20
[iOS] Using presentingViewController to dismiss Modal

zhongwuzw:
## Summary

Fixes https://github.com/facebook/react-native/issues/23463.
Using `presentingViewController` to dismiss Modal VC directly, don't let Modal VC to find `presentingViewController`, then dismiss.

## Changelog

[iOS] [Fixed] - Using presentingViewController to dismiss Modal

## Test Plan

Sample code please see https://github.com/facebook/react-native/issues/23463, it should works.

### DIFF
--- a/React/Views/RCTModalHostViewManager.m
+++ b/React/Views/RCTModalHostViewManager.m
@@ -89,7 +89,7 @@ RCT_EXPORT_MODULE()
   if (_dismissalBlock) {
     _dismissalBlock([modalHostView reactViewController], viewController, animated, completionBlock);
   } else {
-    [viewController dismissViewControllerAnimated:animated completion:completionBlock];
+    [viewController.presentingViewController dismissViewControllerAnimated:animated completion:completionBlock];
   }
 }
 


zhongwuzw:
> Can we use `[modalHostView reactViewController]` instead? It would be more constant with the rest of the code.

🤔 Originally, I want to change all `[modalHostView reactViewController]` to `viewController.presentingViewController` for consolidation, but I didn't because they may not related about this fix.

And seems we can't use `[modalHostView reactViewController]` here, because `[modalHostView reactViewController]` is not always equal to `viewController.presentingViewController`. `reactViewController` is only next VC in responder chain( In the demo, when dismiss, `reactViewController` is `nil`), [`presentingViewController`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621430-presentingviewcontroller?language=objc#) is the VC which present presented VC. So I think we may need to use `viewController.presentingViewController`. :)

shergin:
@janicduplessis @zhongwuzw I fell that my idiomatic iOS skills are now rusty. :( I don't remember exactly how it should work but I have to admit fixed version looks more correct to me.

`reactViewController` returns the closest VC that powers RCTRootView. I don't think it should be related to this issue at all because when we remove `Modal` we should ask the particular VC that DID present already presented VC that we are dismissing. `reactViewController` can be even `nil` in cases where we don't use VC for RCTRootView.

React/Views/RCTModalHostViewManager.m
```diff
@@ -89,7 +89,7 @@ - (void)dismissModalHostView:(RCTModalHostView *)modalHostView withViewControlle
   if (_dismissalBlock) {
     _dismissalBlock([modalHostView reactViewController], viewController, animated, completionBlock);
   } else {
-    [viewController dismissViewControllerAnimated:animated completion:completionBlock];
+    [viewController.presentingViewController dismissViewControllerAnimated:animated completion:completionBlock];
```
shergin: The documentation is clear and confirms my recollections:
```
dismissViewControllerAnimated:completion:

Dismisses the view controller that was presented modally by the view controller.
SDKs
```

So, the fix looks perfectly legit to me, the only Q that prevents me from landing it:
I don't understand how/why it worked previously. Did it? (Until we find answer to this question we cannot be sure that our conclusions are correct.)

janicduplessis: @shergin Could be why it works, but maybe it dismisses the wrong vc sometimes.

```
 If you call this method on the presented view controller itself, UIKit asks the presenting view controller to handle the dismissal.
```

Thanks for clarifiying what `reactViewController` is. LGTM

zhongwuzw: If we just have one Modal, it works whatever we passed presented VC or presenting VC, but if we have multiple Modals, bug may happen, for example, VC present Modal A, ModalA present Modal B, if we call [ModalA dismissVC], what this should perform? because `ModalA` is the presenting VC to present `ModalB`,but it also the presented VC which presenting VC is the closest VC when present ModalA, I think this is why bug happened. :)

Issue #16037, opened 2017-09-20
Refreshing when multiple Modals open will keep the first one open after refresh


aforty:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.3.0
  Yarn:  1.0.2
  npm:  2.15.12
  Watchman:  4.9.0
  Xcode:  Xcode 8.3.3 Build version 8E3004b
  Android Studio:  2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0-alpha.12 => 16.0.0-alpha.12
  react-native: ^0.48.1 => 0.48.2

  Target Platform: iOS (10.3)

### Steps to Reproduce

See demo app below. 

Initial background color is `blue`, first modal is `red`, second modal `black`.

1. Press "Launch Modal" to launch the first `red` modal.
2. Press "Launch Modal" again to launch second `black` modal.
3. Now press ⌘R to refresh. 

### Expected Behavior

I would expect all modals to close and not to reopen after refreshing the app.

### Actual Behavior

You'll notice that immediately after refreshing the app the second `black` modal becomes active again (while the first `red` modal does not) and the entire screen is unresponsive. The only way out of this state is to recompile the app completely.

I haven't confirmed whether this also happens on Android.

### Reproducible Demo

```
import React from 'react';
import { Modal, StyleSheet, Text, TouchableOpacity, View } from 'react-native';

export default function App () {
  return <ComponentThatLaunchesModal color1="blue" color2="red" color3="black" />;
}

class ComponentThatLaunchesModal extends React.Component {
  state = {
    showModal: false,
  };

  render () {
    return (
      <View style={[styles.container, { backgroundColor: this.props.color1 }]}>
        <TouchableOpacity onPress={() => this.setState({ showModal: true })}>
          <Text style={styles.label}>Launch Modal</Text>
        </TouchableOpacity>

        <Modal animationType="slide" visible={this.state.showModal}>
          <ComponentThatLaunchesModal color1={this.props.color2} color2={this.props.color3} />
        </Modal>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'center',
  },

  label: {
    color: 'white',
  },
});

```

aforty:
[...]

PR #24961, opened 2019-05-20
[iOS] Keep the order of Modals that we can dismiss in sequence

zhongwuzw:
## Summary

Fixes https://github.com/facebook/react-native/issues/16037.
We need to keep the order of Modals, if we dismiss Modal randomly(before we use hash table), some Modals may not dismiss successfully.

This PR should based on https://github.com/facebook/react-native/pull/24959.

## Changelog

[iOS] [Fixed] - Keep the order of Modals that we can dismiss in sequence

## Test Plan

Code example, please see https://github.com/facebook/react-native/issues/16037, it should work. :)

### DIFF
--- a/React/Views/RCTModalHostViewManager.m
+++ b/React/Views/RCTModalHostViewManager.m
@@ -49,7 +49,7 @@ RCT_ENUM_CONVERTER(UIModalPresentationStyle, (@{
 
 @implementation RCTModalHostViewManager
 {
-  NSHashTable *_hostViews;
+  NSPointerArray *_hostViews;
 }
 
 RCT_EXPORT_MODULE()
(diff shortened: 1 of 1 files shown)


React/Views/RCTModalHostViewManager.m
```diff
@@ -104,7 +104,7 @@ - (void)invalidate
   for (RCTModalHostView *hostView in _hostViews) {
     [hostView invalidate];
   }
-  [_hostViews removeAllObjects];
+  _hostViews = [NSPointerArray weakObjectsPointerArray];
```
shergin: Can it be just `_hostViews = nil;`?

zhongwuzw: Currently, after invalidation, module should be deallocated, seems we can just nil it. :)

PR #24966, opened 2019-05-20
[Fabric] Add NAN check for text font sizeMultiplier

zhongwuzw:
## Summary

Set `sizeMultiplier` to `1.0` if default value is `NAN`, otherwise, text cannot show properly.

## Changelog

[iOS] [Fixed] - Add NAN check for text font sizeMultiplier

## Test Plan

Code like `<Text style={{fontFamily: 'GillSans-Bold'}}>Hello</Text>`, can run properly.

### DIFF
--- a/ReactCommon/fabric/textlayoutmanager/platform/ios/RCTFontUtils.mm
+++ b/ReactCommon/fabric/textlayoutmanager/platform/ios/RCTFontUtils.mm
@@ -137,8 +137,9 @@ UIFont *RCTFontWithFontProperties(RCTFontProperties fontProperties) {
   RCTFontProperties defaultFontProperties = RCTDefaultFontProperties();
   fontProperties = RCTResolveFontProperties(fontProperties);
 
-  CGFloat sizeMultiplier = !isnan(fontProperties.sizeMultiplier) ? fontProperties.sizeMultiplier : 1.0;
-  CGFloat effectiveFontSize = sizeMultiplier * fontProperties.size;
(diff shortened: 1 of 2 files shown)


ReactCommon/fabric/textlayoutmanager/platform/ios/RCTFontUtils.mm
```diff
@@ -137,8 +137,8 @@ static UIFontWeight RCTUIFontWeightFromFloat(CGFloat fontWeight) {
   RCTFontProperties defaultFontProperties = RCTDefaultFontProperties();
   fontProperties = RCTResolveFontProperties(fontProperties);
 
-  CGFloat effectiveFontSize =
-      fontProperties.sizeMultiplier * fontProperties.size;
+  CGFloat sizeMultiplier = !isnan(fontProperties.sizeMultiplier) ? fontProperties.sizeMultiplier : 1.0;
```
shergin: That's not a proper fix of a problem.
We must have an `assert(!isnan(fontProperties.sizeMultiplier)` here, but not defaulting to `1.0`.

Instead, we have to figure out why it's happening (the default text attributes are probably wrong) and fix it there.

zhongwuzw: Done. :)

shergin: Thank you, Wu!!1

PR #24972, opened 2019-05-20
[CI] Split test_end_to_end into platform-specific e2e tests

hramos:
## Summary

Split up `test_end_to_end` into `test_js_e2e` and `test_ios_e2e` to increase visibility into failing tests. Should either break, we can easily go back through the Circle CI history and observe which platform's e2e tests were broken.

Removing the `checkout_code` top-level job might introduce a bit of delay to all downstream jobs in the `tests` workflow, but doing so ensures each platform gets a clean checkout + yarn run.

**Before:**
<pre>┌────────────────────┐     ┌──────────────────────────┐
│   checkout_code    ├──┬──▶         test_javascript  │
└────────────────────┘  │  └──────────────────────────┘
┌────────────────────┐  │  ┌──────────────────────────┐
│   test_node_lts    │  ├──▶         test_ios         │
└────────────────────┘  │  └──────────────────────────┘
┌────────────────────┐  │  ┌──────────────────────────┐
│   test_docker      │  ├──▶         test_android     │
└────────────────────┘  │  └──────────────────────────┘
                        │  ┌──────────────────────────┐
                        ├──▶         test_end_to_end  │
                        │  └──────────────────────────┘
                        │  ┌──────────────────────────┐
                        └──▶         analyze          │
                           └──────────────────────────┘</pre>

**After:**
<pre>┌──────────────────────────┐         ┌──────────────────────────────┐
│         test_js          ├────┬────▶         test_js_e2e          │
└──────────────────────────┘    │    └──────────────────────────────┘
┌──────────────────────────┐    │    ┌──────────────────────────────┐
│         test_ios         │    └────▶         test_ios_e2e         │
└──────────────────────────┘         └──────────────────────────────┘
┌──────────────────────────┐
│         test_android     │
└──────────────────────────┘
┌──────────────────────────┐
│         test_docker      │
└──────────────────────────┘
┌──────────────────────────┐
│         analyze          │
└──────────────────────────┘
┌──────────────────────────┐
│         test_node_lts    │
└──────────────────────────┘</pre>

| job | time (before) | time (after) | 
| - | - | - |
| checkout_code | 01:16 | - |
| test_docker_build | 17:59 | 17:47 |
| test_node_lts | 01:44 | 01:40 |
| test_android | 10:28 | 15:22 * |
| test_javascript | 01:03 | 01:52 |
| test_end_to_end* | 27:38 | 03:23 (js) + 10:53 (ios) |
| analyze | 01:28 | 02:21 |
[...]

PR #24974, opened 2019-05-21
[CI] Fix ios/android e2e tests

nossbigg:
## Summary

Fixes broken ios/android e2e tests for CI.

Changes:
1. Use `npm pack` instead of `yarn pack` during e2e test
2. Update test string assertions in ios/android e2e tests
3. Use `Step One` as candidate for source code replacement to test for page changes (since the "Welcome to React Native" string exists in the `Header` component instead of being on `App.js`)

## Changelog

[Internal] [Fixed] - Fix ios/android e2e tests

## Test Plan
1. ios/android e2e tests in CI stage should pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -507,16 +507,15 @@ jobs:
             node ./scripts/run-ci-e2e-tests.js --js --retries 3
           when: always
 
-      # - run:
-      #     name: Run iOS End-to-End Tests
-      #     command: |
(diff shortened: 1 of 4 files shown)


nossbigg:
I also verified that android e2e passes locally on my box. Let's see if we can re-enable android e2e on the CI :)

nossbigg:
@hramos i know you're trying to land #24972, should be safe to reinstate the ios e2e test on the CI 👍

Issue #24970, opened 2019-05-20
numberOfLines property of Text is invalid in iPhoneX


xutm:
## Bug
This is OK when only set `numberOfLines = {2}` in `Text`.
But when lineHeight is set, only one row is shown.

The style on iphonex is as follows：
![image](https://user-images.githubusercontent.com/11706751/58019937-486d8480-7b39-11e9-8ea6-36ac10248822.png)

The style on iphone 6 is as follows：
![image](https://user-images.githubusercontent.com/11706751/58019947-4f949280-7b39-11e9-803f-db29b5bf5261.png)

## React Native version:
  React Native Environment Info:
    System:
      OS: macOS 10.14.3
      CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
      Memory: 34.22 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.15.3 - /usr/local/bin/node
      npm: 6.4.1 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
    IDEs:
      Android Studio: 3.4 AI-183.5429.30.34.5452501
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native: 0.59.4

## code example
- react 16.8.3
- react-native 0.59.4
- react-navigation 3.5.1
```js
import React, { Component } from 'react'
import { View, Text, AppRegistry } from "react-native";
import { createStackNavigator, createAppContainer } from "react-navigation";

class HomeScreen extends React.Component {
  render() {
    return (
      <View style={{flex: 1}}>
        <Text style={{
          lineHeight: 20,
          }}
          numberOfLines={2}
          >#完美日记晒单@perfect 'Diary完美日记首先快递盒子挺好看，包装得挺不错，打开包装看到小金盖很满意，摸着看着都很有质感 试了下很服帖，第一次在某书上买东西，感觉还不错，</Text>
      </View>
    );
  }
}

const AppNavigator = createStackNavigator({
  Home: {
    screen: HomeScreen
  }
});

const AppContainer = createAppContainer(AppNavigator)

AppRegistry.registerComponent('rn-event', () => AppContainer)
```

PR #24975, opened 2019-05-21
[iOS] Fixes text incorrectly display on iPhoneX

zhongwuzw:
## Summary

Fixes https://github.com/facebook/react-native/issues/24970, `[NSLayoutManager usedRectForTextContainer]` returns different size on different platforms, so we need to ceil the size to guarantee enough size to render text.

## Changelog

[iOS] [Fixed] - Fixes text incorrectly display on iPhoneX

## Test Plan

Code like below can render text on iPhoneX:

```
        <Text style={{
          lineHeight: 20,
          }}
          numberOfLines={2}
          >#完美日记晒单@perfect 'Diary完美日记首先快递盒子挺好看，包装得挺不错，打开包装看到小金盖很满意，摸着看着都很有质感 试了下很服帖，第一次在某书上买东西，感觉还不错，</Text>
```


shergin:
Why `ceil` exactly?

IMHO, that works only because it's practically randomly adds `0.5`, `0.3` or `0.7` to the value. We have `epsilon` here for the same purpose.

The issue does exists, and it's quite horrible, and we still don't really know why it happens sometimes. Increasing threshold/epsilon helps in some cases but it's just an ugly workaround.

shergin:
Why `ceil` exactly?

IMHO, that works only because it's practically randomly adds `0.5`, `0.3` or `0.7` to the value. We have `epsilon` here for the same purpose.

The issue does exists, and it's quite horrible, and we still don't really know why it happens sometimes. Increasing threshold/epsilon helps in some cases but it's just an ugly workaround.

shergin:
BTW, `RCTCeilPixelValue already does kinda `ceil`.

zhongwuzw:
> BTW, `RCTCeilPixelValue already does kinda `ceil`.

Seems nope, that's not ceil to nearest integer.


> Why `ceil` exactly?
> 
> IMHO, that works only because it's practically randomly adds `0.5`, `0.3` or `0.7` to the value. We have `epsilon` here for the same purpose.
> 
> The issue does exists, and it's quite horrible, and we still don't really know why it happens sometimes. Increasing threshold/epsilon helps in some cases but it's just an ugly workaround.

From docs from `usedRectForTextContainer`, we can't get info that we should ceil to nearest integer, but maybe we can infer from [boundingrectwithsize](https://developer.apple.com/documentation/foundation/nsattributedstring/1529154-boundingrectwithsize?language=objc).

Base my experience in the past, we need to ceil it directly. We can also see how other framework deals.
Textture:
https://github.com/TextureGroup/Texture/blob/6e1600571c3710c69879c2cb3f3eceb5ca180ea0/Source/ASEditableTextNode.mm#L247-L249

shergin:
@zhongwuzw All that are very good points!

You probably mean this quote from Apple's doc (which I haven't seen before):

> In iOS 7 and later, this method returns fractional sizes (in the size component of the returned rectangle); to use a returned size to size views, you must use raise its value to the nearest higher integer using the ceil function.

That's very interesting but my previous experiments clearly showed that the problem is not in iOS, it's in Yoga. Maybe we have two different issues here.

Can you repro this issue locally? If so, can you log exact value (width) that measuring function returns and which size is bing assigned to the view?

zhongwuzw:
@shergin Yeah, I can. Seems it's 100 reproducible.

iPhone 8, can show correctly.
```
<RCTTextView: 0x7fd08a546500; reactTag: 133; text: #完美日记晒单@perfect 'Diary完美日记首先快递盒子挺好看，包装得挺不错，打开包装看到小金盖很满意，摸着看着都很有质感 试了下很服帖，第一次在某书上买东西，感觉还不错， frame = (0 60.5; 375 40.5); opaque = NO; layer = <CALayer: 0x6000038b4d40>>
```

iPhoneX, not correctly:
```
<RCTTextView: 0x7fc68f720440; reactTag: 133; text: #完美日记晒单@perfect 'Diary完美日记首先快递盒子挺好看，包装得挺不错，打开包装看到小金盖很满意，摸着看着都很有质感 试了下很服帖，第一次在某书上买东西，感觉还不错， frame = (0 84.3333; 375 40.3333); opaque = NO; layer = <CALayer: 0x600001de57a0>>
```

xutm:
Any progress?

zhongwuzw:
@xutm Hey, is this works well for you?

xutm:
I want to wait for this fix to be merged and try it out.

Because I'm not sure if there's something else going on.

shergin:
@zhongwuzw 
Can you log exact value (width) that measuring function returns and which size is bing assigned to the view?

shergin:
@zhongwuzw That's not what I am asking, sorry.
Could you please repro the issue on the master and log:
 1. What `usedRectForTextContainer` (whatever system API for text measure we use) returns.
 2. Exact size of UIView subclass on the screen (that cropped the text).
 3. (Bonus) Exact size of the view in `YGLayout` struct.

The results might surprise you (as they surprised me when I tried to solve the issue).

zhongwuzw:
Summary: after talking to @shergin , we decided to separate the measuring and drawing size to different parts, add one point to height when we draw text.

Libraries/Text/Text/RCTTextShadowView.m
```diff
@@ -218,7 +218,8 @@ - (NSTextStorage *)textStorageAndLayoutManagerThatFitsSize:(CGSize)size
     return cachedTextStorage;
   }
 
-  NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:size];
+  // `[NSLayoutManager usedRectForTextContainer]` not calculated correctly in some cases(For example, https://github.com/facebook/react-native/issues/24970 ), so here we add one point to the height of text conatiner.
+  NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:CGSizeMake(size.width, size.height + 1.0)];
```
shergin: Wu, this method is being used by `measure` function as well, so it creates positive feedback loop between drawing and measuring. Event if it works for some cases, I don't think it's correct.

The idea we discussed was about providing additional pixel during drawing, not measuring (otherwise it will just increase the height of all blocks by one pixel).

zhongwuzw: Ahha, I don't realize that it also be called by measure function, I'll adjust it. :)

zhongwuzw:
@shergin Any feedback? :)

sota000:
@zhongwuzw, I realize this is a pretty old PR, but I am wondering if you are still willing to work on this. If so, would you mind posting screenshots of before and after the fix?

zhongwuzw:
@sota000 Yeah，it's a long time passed.I'll find some time to see this again.

zhongwuzw:
@sota000 The issue is already fixed on latest master branch. I'll close this PR.

Issue #24770, opened 2019-05-08
[standards] Migrate React renderers from Haste to path-based imports


ide:
The renderer modules under `Libraries/Renderer/oss/` use Haste. Since these modules are auto-generated, we need to make the renderer-generator rewrite the imports. This work mostly needs to happen in the React repo.

The code in the React repo makes assumptions about the React Native repo, like the existence of the `InitializeCore` module. We need to decide on an interface between the renderer in the React repo and the other modules in the React Native repo (ex: path-based imports or exposing `InitializeCore` through a module like `react-native/Libraries/ReactPrivate` that only React is allowed to import).

As part of implementing the interface, we will also need to update React's Rollup scripts and some Flow hooks to work with the new interface.

I'll take a look at the private interface approach, I think it may work out nicely and tighten up the interface in the sense that cross-repo Haste imports are an abstraction violation.

RN PR: https://github.com/facebook/react-native/pull/24782
React PR: https://github.com/facebook/react/pull/15604
Sync PR: https://github.com/facebook/react-native/pull/25010

ide:
This is complete in React and RN.

PR #25010, opened 2019-05-23
Sync React with Haste-style imports rewritten to use path-based imports instead

ide:
## Summary

This commit uses the same base commit of React that's currently in RN (React ec6691a68716bc59291746fc62f374a56fb435c9) plus a commit in the React repo that removes Haste-style imports (React 61f62246c8cfb76a4a19d1661eeaa5822ec37b36). The resulting commit hash is 13d0561d30a4da10002e56ad1519d1dfc4d2e1ba.

I built React in the React repo with `yarn build`, copied over the `oss` and `shims` directories, and updated `REVISION` to the new commit hash (13d0561d30a4da10002e56ad1519d1dfc4d2e1ba). 

This is part of https://github.com/facebook/react-native/issues/24770.

## Changelog

[General] [Changed] - Sync React with Haste-style imports rewritten to use path-based imports instead

## Test Plan

Test plan: run unit tests, CI


ide:
Closing this since we need to figure out how to deal with the shims. The issue is that the shims still rely on Haste so that FB can internally use a different React renderer than the OSS repo does.

PR #25019, opened 2019-05-24
[Fabric] Break retain cycle between view component and observer

zhongwuzw:
## Summary

Fixes potential memory leaks.

## Changelog

[iOS] [Fixed] - [Fabric] Break retain cycle between view component and observer

## Test Plan

N/A.


shergin:
For anyone who wonder how this can even work here is an article that I found:
http://www.informit.com/articles/article.aspx?p=1745876&seqNum=3

PR #25034, opened 2019-05-24
debug crash the ios

liangguohun:
when use saveToCameraRoll，we should check Permission of photo liberary, if haved even deneid we should tell the user

## Changelog

[CATEGORY] [TYPE] - Message


cpojer:
Thank you for your Pull Request. This module is deprecated from RN and will be removed soon. It was extracted into https://github.com/react-native-community/react-native-cameraroll – please send a Pull Request there :)

Issue #24149, opened 2019-03-26
App state not working in android when I pull down the notifications bar on android


akhilsanker:
## 🐛 Bug Report

App state not working in android when I pull down the notifications bar.

## To Reproduce

The app state listerner not working when I  pull down the notifications bar in android.

## Expected Behavior

To emit the app state event listener when I pull down the notifications bar and changes any settings, such as location on/off, wifi on/off etc..

## Environment
    "react": "16.6.3",
    "react-native": "0.57.8",
    "react-native-cli": '2.0.1"

kelset:
Hello there 👋 this issue has been reported for an old version of React Native. Ideally we'd like everyone to be using 0.59 (see the [awesome changes](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059) it brought) but we know updating can be a pain. To upgrade please refer to the documentation -> facebook.github.io/react-native/docs/upgrading#react-native-projects (if you are upgrading from an earlier version than 0.59.0 refer to pvinis/rn-diff-purge)

But please, if it's actually still an issue with 0.59 please comment below with a repro and we can reopen it. Even better, please send us a pull request with a fix 😊

LaurenceRolandJames:
Hi there. I am experiencing this issue using RN 59.4:

```
React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
      Memory: 22.87 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 8.15.1 - ~/.nvm/versions/node/v8.15.1/bin/node
      npm: 6.4.1 - ~/.nvm/versions/node/v8.15.1/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 26, 27, 28
        Build Tools: 28.0.2, 28.0.3
        System Images: android-28 | Intel x86 Atom_64, android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5314842
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.4 => 0.59.4
[...]

PR #25039, opened 2019-05-25
Feature to listen on window focus events

Krizzu:
## Summary

Addressed issue: https://github.com/facebook/react-native/issues/24149

On Android, activity's lifecycle events are not triggered when the user pulls down the Status Bar (opening Notification Drawer). In order to know that, you need to override [onWindowFocusChanged method](https://developer.android.com/reference/android/app/Activity.html#onWindowFocusChanged(boolean)).

## Changelog

[Android] [Added] - Adds a new listener for `onWindowFocusChanged`
[JavaScript] [Added] - New event, `focusChanged`, to listen on focus gain/loss

## Test Plan

```javascript

AppState.addEventListener("focusChanged", hasFocus => {
   if(hasFocus) {
       // do something while we have focus
   } else {
       // unfocused
   }
});
```

appfocusg

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java
@@ -36,6 +36,8 @@ public class ReactContext extends ContextWrapper {
       new CopyOnWriteArraySet<>();
   private final CopyOnWriteArraySet<ActivityEventListener> mActivityEventListeners =
       new CopyOnWriteArraySet<>();
+  private final CopyOnWriteArraySet<WindowFocusChangeListener> mWindowFocusEventListeners =
+    new CopyOnWriteArraySet<>();
 
   private LifecycleState mLifecycleState = LifecycleState.BEFORE_CREATE;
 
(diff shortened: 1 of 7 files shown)


ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java
```diff
@@ -196,6 +198,14 @@ public void removeActivityEventListener(ActivityEventListener listener) {
     mActivityEventListeners.remove(listener);
   }
 
+  public void addWindowFocusChangeListener(WindowFocusChangeListener listener) {
+    mWindowFocusEventListeners.add(listener);
+  }
+
+  public void removeWindowFocusChangeListener(WindowFocusChangeListener listener) {
```
thymikee: `removeWindowFocusChangeListener` is never invoked

Krizzu: That's right, I left it there in case I've missed a place where I should remove all listeners

matthargett:
can this be exposed as an onfocus and onblur event, to match web?
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onblur

Krizzu:
@matthargett Do you mean to have two separate events?

cpojer:
@Krizzu yes, let's mirror web.

Krizzu:
@cpojer @matthargett yup, done!

Issue #24393, opened 2019-04-10
Can't open NFC url with Linking


LeoIannacone:
## 🐛 Bug Report
I'm trying to use a NFC tag to open a give URL with my application, but Linking.getInitialUrl returns always a null url. However the system recognises the intent registered (so I get asked if I want to open the URL with my app)

## To Reproduce
1. Create a new app
2. Add the following to your AndroidManifest.json:
```
    <uses-permission android:name="android.permission.NFC" />
...
        <intent-filter>
          <action android:name="android.intent.action.VIEW" />
          <category android:name="android.intent.category.DEFAULT" />
          <category android:name="android.intent.category.BROWSABLE" />
          <data android:scheme="https" android:host="www.example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="https" android:host="example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="http" android:host="www.example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="http" android:host="example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="app" android:host="example" android:pathPrefix="/some-path/" />
          <data android:scheme="app" android:host="com.example" android:pathPrefix="/some-path/" />
        </intent-filter>
        <intent-filter>
          <action android:name="android.nfc.action.NDEF_DISCOVERED" />
          <category android:name="android.intent.category.DEFAULT" />
          <category android:name="android.intent.category.BROWSABLE" />
          <data android:scheme="https" android:host="www.example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="https" android:host="example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="http" android:host="www.example.com" android:pathPrefix="/some-path/" />
          <data android:scheme="http" android:host="example.com" android:pathPrefix="/some-path/" />
        </intent-filter>
```
3. create NFC tag with `http://example.com/some-path/this-is-an-url`
4. Add this to your App.js
```
  public componentDidMount (): void {
      Linking.getInitialURL().then(url => {
        console.warn('getInitialURL', url)
      })
  }
```
5. Read the NFC tag with the mobile NFC reader

6. You get asked if you want to open the URL with the app

7. Click your app

The console.warn prints `null`

## Expected Behavior
Linking.getInitialUrl should return the url found in the NFC tag

The `Linking.addEventListener('url', ({url}) => console.warn('event url': url))` should also work

## Environment
```
  React Native Environment Info:
    System:
[...]

PR #25055, opened 2019-05-27
Linking.getInitialURL() to work with NFC tags on Android

cimitan:
## Summary
This PR solves bug https://github.com/facebook/react-native/issues/24393 for Android. Allows an app to be opened with an NFC tag and getting the url trough Linking.getInitialURL()

## Changelog
[Android] [Fixed] - This branch checks also for `ACTION_NDEF_DISCOVERED` intent matches to set the initialURL

## Test Plan
Tested the code multiple times with both NFC tags and normal links

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/intent/IntentModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/intent/IntentModule.java
@@ -12,6 +12,7 @@ import android.content.ComponentName;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.nfc.NfcAdapter;
 import android.provider.Settings;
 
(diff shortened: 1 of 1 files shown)


amirhsobhi:
hi everyone,
Linking.getInitialURL() is working very well on both android and ios, when nfc is readed,
but Linking.addEventListener only read urls on ios and there is nothing on android.
actually when app is on foreground/background and we scan the nfc tag which holds universal link, app comes to foreground but the Linking.addEventListener doesn't return anything 
i think the bug is still there.

Paul-Lategan:
Hi @cimitan — thanks for your work on this.  Were you able to confirm this was working for you?  I'm experiencing the same issue as @amirhsobhi.

cimitan:
I think it was working for me... but I haven't checked in more than a year... It's a separate bug though, this is about getInitialURL not the event listener

PR #25058, opened 2019-05-27
[lint] Add a lint rule to disallow Haste imports

ide:
## Summary

This is an ESLint plugin that infers whether an import looks like a Haste module name. To keep the linter fast and simple, it does not look in the Haste map. Instead, it looks for uppercase characters in single-name import paths, since npm has disallowed uppercase letters in package names for a long time. There are some false negatives (e.g. "merge" is a Haste module and this linter rule would not pick it up) but those are about 1.1% of the module names in the RN repo, and unit tests and integration tests will fail anyway once Haste is turned off.

You can disable the lint rule on varying granular levels with ESLint's normal disabling/enabling mechanisms.

Also rewrote more Haste imports so that the linter passes (i.e. fixed lint errors as part of this PR).

## Changelog

[General] [Changed] - Add a lint rule to disallow Haste imports

## Test Plan

Run `yarn lint`, `yarn flow`, and `yarn test` as a sanity check. Modify an import statement to use a Haste module name and verify the linter reports an error. Do the same with a require call.

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedModule.js
+++ b/Libraries/Animated/src/NativeAnimatedModule.js
@@ -10,8 +10,8 @@
 
 'use strict';
 
-import type {TurboModule} from 'RCTExport';
-import * as TurboModuleRegistry from 'TurboModuleRegistry';
+import type {TurboModule} from '../../TurboModule/RCTExport';
+import * as TurboModuleRegistry from '../../TurboModule/TurboModuleRegistry';
 
 type EndResult = {finished: boolean};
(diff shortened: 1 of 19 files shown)


cpojer:
Doing some internal fiddling right now to make this land. It's gonna look a little bit different from the version of the PR here in terms of how it's integrated into the repo.

ide:
It probably makes sense to publish `@react-native-community/eslint-plugin` to npm as well. I wasn't able to do that because the permission system seems broken but you might have better luck.

PR #25082, opened 2019-05-30
ReactActivityDelegate implements LifecycleObserver

dulmandakh:
## Summary

Android introduced Lifecycles and LifecycleObserver to make it easy to respond to component lifecycle events. This PR changes ReactActivityDelegate to implement LifecycleObserver and handle lifecycle.

## Changelog

[Android] [Changed] - ReactActivityDelegate implements LifecycleObserver

## Test Plan

CI is green and RNTester is working just fine


ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java
```diff
@@ -94,12 +93,14 @@ protected void loadApp(String appKey) {
...
   }
 
+  @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
   protected void onPause() {
     if (getReactNativeHost().hasInstance()) {
       getReactNativeHost().getReactInstanceManager().onHostPause(getPlainActivity());
     }
   }
 
+  @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
```
benoitdion: `ON_RESUME`?

cpojer:
@dulmandakh: What's the status of this PR?

ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java
```diff
@@ -13,37 +13,36 @@
...
   private @Nullable ReactRootView mReactRootView;
   private @Nullable DoubleTapReloadRecognizer mDoubleTapReloadRecognizer;
   private @Nullable PermissionListener mPermissionListener;
   private @Nullable Callback mPermissionsCallback;
 
-  @Deprecated
-  public ReactActivityDelegate(Activity activity, @Nullable String mainComponentName) {
-    mActivity = activity;
-    mMainComponentName = mainComponentName;
-  }
```
makovkastar: Is it safe to delete this constructor now? It caused a few compile errors in the FB codebase, but I fixed them. How about other people's code?

dulmandakh: it's been deprecated for a while, also activities must extend ReactActivity and everything will work fine.

makovkastar:
After looking at this PR once again, I'm a bit concerned that now we have ReactActivityDelegate lifecycle events sent from two different places: we still call `mDelegate.onCreate(savedInstanceState)` from ReactActivity and we rely on the lifecycle events reported by the lifecycle owner. @dulmandakh what do you think?

makovkastar:
@dulmandakh could you please check my previous comment? I would like to know what you think, because I'm leaning towards keeping the current implementation and not using `LifecycleObserver` for `ReactActivityDelegate`.

dulmandakh:
@makovkastar ok. I'm closing this PR

PR #25084, opened 2019-05-30
[TM] Add spec for DevSettings

jarvisluong:
## Summary

Part of #24875, adds a spec for DevSettings.

## Changelog

[General] [Added] - TM spec for DevSettings

## Test Plan

Flow passes


Libraries/NativeModules/specs/NativeDevSettings.js
```diff
@@ -0,0 +1,26 @@
...
+  +reload: () => void;
+  +setHotLoadingEnabled: (isHotLoadingEnabled: boolean) => void;
+  +setIsDebuggingRemotely: (isDebugginRemotelyEnabled: boolean) => void;
+  +setLiveReloadEnabled: (isLiveReloadEnabled: boolean) => void;
+  +setProfilingEnabled: (isProfilingEnabled: boolean) => void;
+  +toggleElementInspector: () => void;
+  +setIsShakeToShowDevMenuEnabled?: (enabled: boolean) => void;
+}
+
+export default TurboModuleRegistry.get<Spec>('DevSettings');
```
fkgozali: could you use `getEnforcing` instead of `get`?

jarvisluong: yes, my bad that I forgot this.

Libraries/NativeModules/specs/NativeDevSettings.js
```diff
@@ -0,0 +1,26 @@
...
+import * as TurboModuleRegistry from '../../TurboModule/TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +reload: () => void;
+  +setHotLoadingEnabled: (isHotLoadingEnabled: boolean) => void;
+  +setIsDebuggingRemotely: (isDebuggingRemotelyEnabled: boolean) => void;
+  +setLiveReloadEnabled: (isLiveReloadEnabled: boolean) => void;
+  +setProfilingEnabled: (isProfilingEnabled: boolean) => void;
+  +toggleElementInspector: () => void;
+  +setIsShakeToShowDevMenuEnabled?: (enabled: boolean) => void;
```
thymikee: Should we mark this as iOS only with a comment `// iOS only` like we do for Android (`// Android only`)?

fkgozali: Yeah, and this should just be required. I'll adjust it when I import this.

PR #25088, opened 2019-05-30
Don't reference null android.ndkDirectory in build.gradle

### LINKED ISSUE #25087: Improve build failure diagnostics when NDK is missing or incorrectly set up
If you (try to) build React Native for Android without having the NDK properly installed, you get the following error:

>A problem occurred evaluating project ':ReactAndroid'.
\> Cannot get property 'absolutePath' on null object

This is not an overly helpful diagnostic; what's wanted is this one:

>ndk-build binary cannot be found, check if you've set $ANDROID_NDK environment variable correctly or if ndk.dir is setup in local.properties

I have a one-line fix and will submit a PR shortly.

React Native version: **0.59.8**
>System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
    Memory: 1.83 GB / 15.93 GB
  Binaries:
    Node: 10.15.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    Android Studio: Version  3.4.0.0 AI-183.6156.11.34.5522156

## Steps To Reproduce

1. Clone react-native
2. Run yarn
3. Import into Android Studio (no NDK installed)
4. Build

I would like the build failure to properly diagnose the missing NDK rather than throwing an NRE

petterh:
## Summary

If you (try to) build React Native for Android without having the NDK properly installed and referenced, you get the following error:

>A problem occurred evaluating project ':ReactAndroid'.
\> Cannot get property 'absolutePath' on null object

This is not an overly helpful diagnostic. This PR results in this message instead:

>ndk-build binary cannot be found, check if you've set $ANDROID_NDK environment variable correctly or if ndk.dir is setup in local.properties

Fixes #25087

## Changelog

[Android] [Fixed] - Show proper error message instead of throwing a NullReferenceException if Gradle cannot find the NDK

## Test Plan

Manual testing. Ensure that the Android NDK is improperly referenced (it doesn't matter whether it's actually installed somewhere), then try to sync Android Studio's project with the Gradle files. This will fail with or without this change, but **with** the change, the developer gets a proper error message instead of an NRE.

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -193,7 +193,7 @@ def findNdkBuildFullPath() {
     def ndkDir = android.hasProperty("plugin") ? android.plugin.ndkFolder :
             plugins.getPlugin("com.android.library").hasProperty("sdkHandler") ?
                     plugins.getPlugin("com.android.library").sdkHandler.getNdkFolder() :
-                    android.ndkDirectory.absolutePath
+                    android.ndkDirectory ? android.ndkDirectory.absolutePath : null
     if (ndkDir) {
         return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
     }


dulmandakh:
Thank you for the PR. I have experienced this issue just yesterday and had
to dig into Gradle script to find the cause. And I think that it would be
best to show error or warning rather than silencing it. Could you please
update the PR.

PR #25101, opened 2019-05-31
[TM] add spec for ImageStore

mitulsavani:
## Summary
This PR solves part of this issue: #24875

## Changelog

[General] [Added] - add TM spec for ImageStore

## Test Plan

Flow passes

### DIFF
--- a/Libraries/Image/NativeImageStore.js
+++ b/Libraries/Image/NativeImageStore.js
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {TurboModule} from 'RCTExport';
+import * as TurboModuleRegistry from 'TurboModuleRegistry';
+
+export interface Spec extends TurboModule {
+  +hasImageForTag: (
+    uri: string, 
+    callback: (hasImage: boolean) => void,
+  ) => void;
+  +removeImageForTag: (
+    uri: string,
+  ) => void;
+  +addImageFromBase64: (
+    base64ImageData: string,
+    succees: (uri: string) => void,
+    failure: (error: string) => void;
+  ) => void;
+  +getBase64ForTag: (
+    uri: string,
+    succees: (base64ImageData: string) => void,
+    failure: (error: string) => void,
+  ) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('ImageStoringManager');
--- a/Libraries/Image/ImageStore.js
+++ b/Libraries/Image/ImageStore.js
@@ -9,8 +9,7 @@
  */
 'use strict';
 
-const RCTImageStoreManager = require('../BatchedBridge/NativeModules')
-  .ImageStoreManager;
+import NativeImageStore from './NativeImageStore';
 
 const Platform = require('../Utilities/Platform');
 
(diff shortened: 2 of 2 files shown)


Libraries/Image/NativeImageStore.js
```diff
@@ -0,0 +1,36 @@
...
+  +hasImageForTag: (
+    uri: string, 
+    callback: (hasImage: boolean) => void,
+  ) => void;
+  +removeImageForTag: (
+    uri: string,
+  ) => void;
+  +addImageFromBase64: (
+    base64ImageData: string,
+    succees: (uri: string) => void,
```
ericlewis: Small typo here, should be `success`

Libraries/Image/NativeImageStore.js
```diff
@@ -0,0 +1,36 @@
...
+    failure: (error: string) => void;
+  ) => void;
+  +getBase64ForTag: (
+    uri: string,
+    succees: (base64ImageData: string) => void,
```
ericlewis: Same here. `success`

Libraries/Image/NativeImageStore.js
```diff
@@ -0,0 +1,36 @@
...
+    failure: (error: string) => void,
+  ) => void;
+}
+
+export default TurboModuleRegistry.getEnforcing<Spec>('ImageStoringManager');
```
ericlewis: This should be `ImageStoreManager`

Libraries/Image/NativeImageStore.js
```diff
@@ -0,0 +1,36 @@
...
+    uri: string, 
+    callback: (hasImage: boolean) => void,
+  ) => void;
+  +removeImageForTag: (
+    uri: string,
+  ) => void;
+  +addImageFromBase64: (
+    base64ImageData: string,
+    succees: (uri: string) => void,
+    failure: (error: string) => void;
```
ericlewis: I don’t think failure callback’s pass back a string arg. I think they are ‘Object’ types.

Libraries/Image/NativeImageStore.js
```diff
@@ -0,0 +1,36 @@
...
+  ) => void;
+  +getBase64ForTag: (
+    uri: string,
+    succees: (base64ImageData: string) => void,
+    failure: (error: string) => void,
```
ericlewis: Same as above

Libraries/Image/NativeImageStore.js
mitulsavani: Would it then look something like this?
```
failure: Object;
```
OR
```
failure: (error: Object) => void;
```

mitulsavani:
> Great work! Just a few minor changes. I also suggest that you group the iOS only methods and comment that they are iOS only

What I have found is `getBase64ForTag` method is present in both iOS and Android. How should I comment particularly that method? 

PS: Please correct me if I am wrong here
> Example
```
// iOS only
+hasImageForTag: (
    uri: string, 
    callback: (hasImage: boolean) => void,
  ) => void;
  +removeImageForTag: (
    uri: string,
  ) => void;
  +addImageFromBase64: (
    base64ImageData: string,
    success: (uri: string) => void,
    failure: (error: string) => void;
  ) => void;

// iOS and Android 
  +getBase64ForTag: (
    uri: string,
    success: (base64ImageData: string) => void,
    failure: (error: string) => void,
  ) => void;
```

ericlewis:
@mitulsavani switch the common one to be at the top & remove its comment basically. We also need to address all of these test failures.

fkgozali:
@mitulsavani are you going to make some updates to this PR some time today? If not, I can just import this and update the code per @ericlewis comments accordingly.

Libraries/Image/NativeImageStore.js
fkgozali: `failure: (error: Object) => void,`

fkgozali:
Actually, since the changes are minor, I'll just edit them when importing, so we can save some time :). We're almost done!

mitulsavani:
@fkgozali yes I will definitely make changes by tonight!

PR #25108, opened 2019-05-31
mock Animated components with versions that skipSetNativeProps

rdy:
## Summary

Using Animated components like View and Image do not get created with __skipSetNativeProps_FOR_TESTS_ONLY = true since they get created before the jest mock can be applied to createAnimatedComponent. For these components mock getters to create versions that properly skip set native props.

Jest tests that render these components get warnings the setNativeProps gets called even though using a testRenderer this should not happen.

## Changelog

[Javascript] [Fixed] - Define Animated Components for react-native/jest/setup that properly skip setNativeProps

## Test Plan

Run tests

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -127,6 +127,12 @@ jest
       '../Libraries/Components/ActivityIndicator/ActivityIndicator',
     ),
   )
+  .mock('../Libraries/Animated/src/Animated', () => {
+    const Animated = jest.requireActual('../Libraries/Animated/src/Animated');
+    Animated.Text.__skipSetNativeProps_FOR_TESTS_ONLY = true;
+    Animated.View.__skipSetNativeProps_FOR_TESTS_ONLY = true;
+    return Animated;
+  })
   .mock('../Libraries/Animated/src/AnimatedImplementation', () => {
     const AnimatedImplementation = jest.requireActual(
       '../Libraries/Animated/src/AnimatedImplementation',


rdy:
@grabbou @matthargett this is a change i needed to suppress warnings i was seeing from animated components being rendered in jest in 0.59, it looks like createAnimatedComponent gets mocked too late so setNativeProps gets called on these components and throws a bunch of warnings.

RubenSandwich:
@rdy 

I ran into this problem myself and I solved it a bit differently. I used jest.spyOn and change the params at runtime:

```js
import { Animated } from 'react-native';
const mockedAnimatedTiming = Animated.timing;
jest.spyOn(Animated, 'timing').mockImplementation((value, options) => {
  return mockedAnimatedTiming(value, {
    ...options,
    useNativeDriver: false,
  });
});
```

I'd suggest we adopt this for the RN jest/setup.js because it removes the whole need for `__skipSetNativeProps_FOR_TESTS_ONLY`. However, I can imagine jest.spyOn has a higher overhead cost then your approach.

rdy:
@cpojer i rebased against master as you requested.

I wasn't sure about your comment on applying this to the `AnimatedImplementation`. The problem with the way the current mock worked was that `Animated.View`, `Animated.Image`, .etc are constructed using a version prior to the mock applied by `jest/setup.js`. I couldn't think of a better way to intercept those components except by directly mocking them with versions that were not called setNativeProps in test, I duped the code because of how jest doesn't like accessing stuff through closure scope.

If there is a better way to apply the `AnimatedImplementation` mock at the right time I'd be happy to apply that to this commit.

rdy:
@RubenSandwich I tried your snippet with my tests cases but I still saw the warning. 

Maybe I didn't apply it correctly?

In my case I have `Animated.View` that end up getting constructed with the original implementation of `createAnimatedComponent`. In the `componentDidMount` it eventually calls `_animatedPropsCalback` but that version doesn't have the flag that tells it to skip the `setNativeProps` calls which spits out a bunch of those warnings.

I'd still like the animation to run in my test as I have something that tracks the animation frames so I can prove that those values are getting interpolated the way i expect over time.

RubenSandwich:
@rdy 

So I put that code at the start of my own app's native module mocks. To use it in jest/setup.js `.mock('AnimatedImplementation', () => {` It has to be modified slightly:
```js
...
  .mock('AnimatedImplementation', () => {
    const AnimatedImplementation = require.requireActual('AnimatedImplementation');

    const oldTiming = AnimatedImplementation.timing;
    AnimatedImplementation.timing = function(value, options) {
      return oldTiming(value, {
        ...options,
        useNativeDriver: false,
      });
    };

    return AnimatedImplementation;
  })
...
```

I ran my own apps tests and that change worked just fine without any changes to the snapshots or console warnings. No idea if it would pass the current tests in the RN repo though. Another important point is that I'm on RN 52, but it looks like the Animation mock has not been touched since then.

jest/setup.js
```diff
@@ -127,6 +127,45 @@ jest
...
+    const Animated = jest.requireActual('../Libraries/Animated/src/Animated');
+    function createAnimatedComponent(Component, defaultProps) {
+      const Wrapped = oldCreate(Component, defaultProps);
+      Wrapped.__skipSetNativeProps_FOR_TESTS_ONLY = true;
+      return Wrapped;
+    }
+    const FlatList = createAnimatedComponent(
+      jest.requireActual('../Libraries/Lists/FlatList'),
+    );
+    const Image = createAnimatedComponent(mockComponent('Image'));
```
cpojer: this needs a relative require

cpojer:
It looks like this is breaking all of the Animated tests in React Native. Can you fix them up?

matthargett:
Note that the underlying issue appears to be a regression from sometime between 0.55 and 0.59, and was discovered during an upgrade. cc @kelset

cpojer:
@rdy could you take a look at the test failures? This version is not mergeable and breaks all the JS tests.

rdy:
@cpojer yes, sorry was wrapped up in a few other tasks this week but I’ll look into the failures.

PR #25137, opened 2019-06-04
[Fabric] Add check when convert transform prop

zhongwuzw:
## Summary

`ART` doesn't ported to Fabric currently, but `ART` view has its own prop `transform` which collide with `transform` of [`ViewProps`](https://github.com/facebook/react-native/blob/184073813e9220af270d13c8f537b4cfbfcb65b4/ReactCommon/fabric/components/view/ViewProps.cpp#L60), so in order to prevent issue when do conversion, we can add a filter before it that users can do custom conversion by themself. 

## Changelog

[iOS] [Fixed] - Add check when convert transform prop

## Test Plan

Just run `ART` example to verify.

### DIFF
--- a/ReactCommon/fabric/components/view/conversions.h
+++ b/ReactCommon/fabric/components/view/conversions.h
@@ -402,6 +402,9 @@ inline void fromRawValue(const RawValue &value, Transform &result) {
   auto configurations = (std::vector<RawValue>)value;
 
   for (const auto &configuration : configurations) {
+    if(!configuration.hasType<better::map<std::string, RawValue>>()){
+      return;
+    }
     auto configurationPair = (better::map<std::string, RawValue>)configuration;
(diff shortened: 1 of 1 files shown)


shergin:
@zhongwuzw
Are you working on ART? If so, I would love to discuss it and help you with it.

I believe ART components which has own `transform` prop should *not* be subclasses of ViewShadowNode.

cpojer:
ART is currently being removed from RN and split out into a separate repo here: https://github.com/react-native-community/react-native-art – maybe we can make the changes to ART there (and mirror them here while they are still in the main repo?).

zhongwuzw:
@cpojer Ahha, but actually I think this is the enhancement for fabric when convert props? if @shergin think we don't need to do it for performance, we can just ignore it. :)

Issue #20466, opened 2018-07-31
[Document Request] Include Comperhensive Documentation on how to change PORT


meehanman:
There are lots of different methods online on how to change the port used by React-Native which is a breaking feature if port 8081 is blocked. There should be a tool created or good documentation that is updated with the steps in order to run React-Native, not on port 8081. 

McAfee uses this port and is required on many corporate servers and closing the process is not recommended.

I'd like to see this included in Getting Started // Troubleshooting.

Nkzn:
This docs helps you?

```
$ react-native start --port=8088
```
https://facebook.github.io/react-native/docs/troubleshooting#using-a-port-other-than-8081

PR #25144, opened 2019-06-04
Ensure that iOS uses RCT_METRO_PORT instead of hardcoded 8081 when appropriate

jimberlage:
## Summary

In environments with McAfee EPro software, port 8081 is used and cannot be unbound.  (See [#20466](https://github.com/facebook/react-native/issues/20466) for a recent example issue.)  Most issues can be worked around by passing a `--port` option or setting `RCT_METRO_PORT`, but there are a couple places where 8081 is hardcoded that block adoption.  Right now the workaround I've seen pushed on Stack Overflow and elsewhere is to modify node_modules after a `yarn install`, so I'd like to fix it at the source.

This PR changes the react "Start Packager" build step and some code in the iOS DevSupport library to read from the `RCT_METRO_PORT` variable.

## Changelog

[General] [Changed] - Removed hardcoded references to port 8081 and replaced them by reading from RCT_METRO_PORT (w/ fallback to 8081)

## Test Plan

I am going to check the CI output - my first priority is to avoid breaking existing builds.  Everywhere I changed the code, I made the default port 8081.

### DIFF
--- a/React/DevSupport/RCTInspectorDevServerHelper.mm
+++ b/React/DevSupport/RCTInspectorDevServerHelper.mm
@@ -34,7 +34,7 @@ static NSURL *getInspectorDeviceUrl(NSURL *bundleURL)
 {
   NSNumber *inspectorProxyPort = @8081;
   NSString *inspectorProxyPortStr = [[[NSProcessInfo processInfo] environment] objectForKey:@"RCT_METRO_PORT"];
-  if (inspectorProxyPortStr) {
+  if (inspectorProxyPortStr && [inspectorProxyPortStr length] > 0) {
     inspectorProxyPort = [NSNumber numberWithInt:[inspectorProxyPortStr intValue]];
   }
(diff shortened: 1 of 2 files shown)


React/DevSupport/RCTInspectorDevServerHelper.mm
```diff
@@ -44,6 +48,10 @@
 static NSURL *getAttachDeviceUrl(NSURL *bundleURL, NSString *title)
 {
   NSNumber *metroBundlerPort = @8081;
+  NSString *metroBundlerPortStr = [[[NSProcessInfo processInfo] environment] objectForKey:@"RCT_METRO_PORT"];
+  if (!metroBundlerPortStr) {
```
cpojer: This doesn't look correct to me?

Also, should we make sure that the string is non-empty? If it is `""`, then we don't want to apply the string.

jimberlage: Ah, you're right.  I'll fix that up.

PR #25147, opened 2019-06-04
Add debug build support for Android native code

Kudo:
## Summary

With JSI based architecture, there will be more and more C++ native code involved.
Original NDK builder in RN only supports release build and that's not reasonable for native debugging.
This change introduces a way to build native code in debuggable version.

Simply add `NATIVE_BUILD_TYPE=Debug` environment variable during gradle build,
e.g.
`NATIVE_BUILD_TYPE=Debug ./gradlew clean :ReactAndroid:assembleDebug`

## Changelog

[Android] [Added] - Add native debug build support to improve debugging DX

## Test Plan

1. Check original gradle build works:
`./gradlew clean :ReactAndroid:installArchives`

2. Check debug build works:
`NATIVE_BUILD_TYPE=Debug ./gradlew clean :ReactAndroid:assembleDebug` and verify *.so is not stripped
`file ReactAndroid/build/react-ndk/all/armeabi-v7a/lib*.so`

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -33,6 +33,9 @@ def dependenciesPath = System.getenv("REACT_NATIVE_DEPENDENCIES")
 // and the build will use that.
 def boostPath = dependenciesPath ?: System.getenv("REACT_NATIVE_BOOST_PATH")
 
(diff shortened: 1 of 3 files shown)


Kudo:
@cpojer That's great. Thanks for your help.

PR #25146, opened 2019-06-04
Fix Xcode 11 build

ericlewis:
## Summary

Fixes build in Xcode 11 beta, the signature for `__unused` was changed. This adds a new check for the new style.

## Changelog

[iOS] [Fixed] - Xcode 11 beta build

## Test Plan

Build & run RNTester

PR #25148, opened 2019-06-04
Fix unused attribute parsing for Xcode 11 beta

ex3ndr:
Fixes crash when any native method with unused attribute is called

## Summary

In latest Xcode 11 beta format of method signatures was changed a little and causes a crash for any method with unused attribute. This PR fixes the issue

## Changelog

[iOS] [Fixed] - Xcode 11 beta runtime crash for methods with unused attribute.


cpojer:
Thanks! We are already landing the same fix in this PR: https://github.com/facebook/react-native/pull/25146

Issue #23254, opened 2019-02-01
Remote Debug JS - works only DEV mode after 0.58 , but in 0.57 all good


in4core:
## Environment
React Native Environment Info:
System:
OS: Windows 10
CPU: (4) x64 Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz
Memory: 1.68 GB / 7.95 GB
Binaries:
npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
IDEs:
Android Studio: Version 3.3.0.0 AI-182.5107.16.33.5199772

Dependencies :
"dependencies": {
"react": "16.6.3",
"react-native": "0.58.3",
"react-native-gesture-handler": "^1.0.15",
"react-native-img-cache": "^1.6.0",
"react-native-section-list-get-item-layout": "^2.2.3",
"react-native-vector-icons": "^6.2.0",
"react-navigation": "^3.0.9",
"rn-fetch-blob": "^0.10.15"
},

## Description
Remote Debug JS - works only DEV mode after 0.58 , but in 0.57 all good.
When starts it in Prod mode we have some errors
Cannot read property 'Constants' of null
Module AppRegistry is not a registered callable module (calling runApplication)

Abuitime:
Same

darekg11:
Just leaving comment, it is also happening to me.  
But seeing that not so many people reported this, do you think that this might be bound to specific 3rd module having problems with newest React-Native (0.58)?

in4core:
> Just leaving comment, it is also happening to me.
> But seeing that not so many people reported this, do you think that this might be bound to specific 3rd module having problems with newest React-Native (0.58)?

I dont know,  but i will try use clear project, without dependencies - today.  But, if i think true, clear project also fail

in4core:
"dependencies": {
    "js-md5": "^0.7.3",
    "react": "16.6.3",
    "react-native": "^0.58.5",
    "react-native-device-info": "^0.26.2",
    "react-native-gesture-handler": "^1.0.15",
    "react-native-image-picker": "^0.28.0",
    "react-native-img-cache": "^1.6.0",
    "react-native-push-notification": "^3.1.2",
    "react-native-section-list-get-item-layout": "^2.2.3",
    "react-native-vector-icons": "^6.2.0",
    "react-native-webview": "^5.0.5",
    "react-navigation": "^3.0.9",
    "rn-fetch-blob": "^0.10.15"
  },

abs-zero:
[...]

PR #25162, opened 2019-06-06
Correctly bypass sync calls in UIManager during remote debugging

statm:
## Summary

Remote debugging stopped working (since 0.58, according to #23254). See https://github.com/facebook/react-native/issues/23254#issuecomment-474692753 for repro steps.

The root cause is incorrect checks for Chrome debugging environment in `UIManager.js`.
- In one place where sync function `lazilyLoadView` should be avoided, we effectively use `if (__DEV__ && !global.nativeCallSyncHook)` to check remote debugging, which misses ship flavor (i.e. `__DEV__` is false).
- In another place where we want to pre-populate view managers' constants to avoid calling sync function `getConstantsForViewManager`, `if (__DEV__)` is used, also missing ship flavor.

This PR fixes both checks, only using the absense of `global.nativeCallSyncHook` to determine remote debugging environments.

## Changelog

[JavaScript] [Fixed] - Correctly bypass sync calls in UIManager during remote debugging

## Test Plan

(Copying repro steps from https://github.com/facebook/react-native/issues/23254#issuecomment-474692753)

On a freshly created project by `react-native-cli` in android:

1. Shake phone -> Dev Settings -> Uncheck JS Dev Mode and Use JS Deltas. Then back.
2. Shake phone -> Turn on Debug JS Remotely. Then back
3. Reload

**Current behavior**: Invariant Violation: requireNativeComponent: "RCTView" was not found in the UIManager.

**Fixed (expected) behavior**: No errors, app executes correctly.

### DIFF
--- a/Libraries/ReactNative/UIManager.js
+++ b/Libraries/ReactNative/UIManager.js
@@ -78,10 +78,8 @@ const UIManagerJS: UIManagerJSInterface = {
 
     // If we're in the Chrome Debugger, let's not even try calling the sync
     // method.
-    if (__DEV__) {
-      if (!global.nativeCallSyncHook) {
-        return config;
-      }
+    if (!global.nativeCallSyncHook) {
+      return config;
     }
 
     if (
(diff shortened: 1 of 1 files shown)


cpojer:
Something is wrong with our Facebook infra, retrying…

PR #21211, opened 2018-09-19
Support background timers on iOS via NSTimer

jamesreggio:
This PR is a minimalistic change to RCTTiming that causes it to switch exclusively to NSTimer (i.e., the 'sleep timer') in order to continue triggering timers when the app has moved to the background.

Motivation:
----------

Many people have expressed a desire for background timer support on iOS. (See #1282, #167, and #16493). In our app — a podcast/audio player — we use background timers to ensure that we never lose track of the user's playback position, should the app crash or be terminated by the OS.

The RCTTimer module uses a RN-managed CADisplayLink if the next requested timer is less than a second away; otherwise, it switches to an NSTimer (which is refers to as a 'sleep timer' in source). The RN-managed CADisplayLink is always disabled when the app goes to the background (and thus cannot be used); however, the NSTimer will still issue its callbacks in the background.

This PR adds a flag to track whether the app is in the background, and if so, all timers are routed through NSTimer until the app returns to the foreground. @vishnevskiy at Discord opened a similar PR (#16493) that implements a drop-in for CADisplayLink which falls back to NSTimer, but I decided to incorporate the background-NSTimer logic directly into RCTTimer, since NSTimer is already in use.

~It's worth noting that the background NSTimer may not fire as often as requested — it may give the appearance of lagging depending upon your app's priority in the background. For our audio app, NSTimer fires exactly on schedule if there's an open AVAudioSession and audio is playing; if audio is not playing, it fires about half as often as requested, which is still adequate for networking polling and other tasks.~

It's worth noting that background timers only function as long as an app is actually running in the background. Apple offers a variety of Background Modes (which can be toggled in the Capabilities section of the target inspector in Xcode), and the app will need to be legitimately using one of these modes in order for this change to provide any value — otherwise it will be terminated within a couple of seconds of moving to the background.

The good thing about this change is that for apps that do perform essential computation in support of their Background Mode, they can now use `setTimeout` and `setInterval` without problem — whereas in the past, neither would ever trigger their callback until the app returns to the foreground.

Open Question:
---------

~Despite the demand for this timer behavior, it's not clear whether this should be specifically opt-in. There's nothing specifically breaking about this change, but an app that abuses timers may lead to increased battery consumption if they now fire in the background.~

~On the other hand, if we were to make this an opt-in feature, it's not clear how to do it other than to expose a top-level function to toggle it (and that feels hacky and undiscoverable). RN's timers are based upon standard JS functions which cannot have their parameters modified to take additional flags.~

[...]

PR #25164, opened 2019-06-06
[iOS] Removed locks && ensure everything runs on JS Thread

zhongwuzw:
## Summary

Before, we add the lock to prevent race condition, it would happen because there has two threads can access concurrently, main thread and JS Thread, actually, most of the time, only JS Thread would call it, and main thread would access it only when app state changed, like in/out background.

Now, we only need to dispatch operations to JS Thread when app state changed, then we can remove all locks. Timer module is time sensitive, so I think remove locks is better. :)

## Changelog

[iOS] [Enhancement] - Removed locks && ensure everything runs on JS Thread

## Test Plan

CI passed.

### DIFF
--- a/React/Modules/RCTTiming.m
+++ b/React/Modules/RCTTiming.m
@@ -178,6 +178,7 @@ RCT_EXPORT_MODULE()
   _bridge = nil;
 }
 
+// Called from main thread
 - (void)appDidMoveToBackground
 {
   [_bridge dispatchBlock:^{
@@ -191,6 +192,7 @@ RCT_EXPORT_MODULE()
   } queue:RCTJSThread];
 }
 
+// Called from main thread
 - (void)appDidMoveToForeground
 {
   [_bridge dispatchBlock:^{
--- a/React/Modules/RCTTiming.m
+++ b/React/Modules/RCTTiming.m
@@ -171,14 +171,14 @@ RCT_EXPORT_MODULE()
   return RCTJSThread;
 }
 
-// Called from JS Thread.
 - (void)invalidate
 {
-  [self stopTimers];
-  _bridge = nil;
+  [_bridge dispatchBlock:^{
+    [self stopTimers];
+    self->_bridge = nil;
+  } queue:RCTJSThread];
 }
 
-// Called from main thread
 - (void)appDidMoveToBackground
 {
   [_bridge dispatchBlock:^{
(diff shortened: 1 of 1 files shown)


React/Modules/RCTTiming.m
```diff
@@ -171,28 +171,35 @@ - (dispatch_queue_t)methodQueue
...
+    [self stopTimers];
+    self->_inBackground = YES;
+    
+    // Issue one final timer callback, which will schedule a
+    // background NSTimer, if needed.
+    [self didUpdateFrame:nil];
+  } queue:RCTJSThread];
 }
 
+// Called from main thread
```
shergin: Could you please use `RCTAssertMainQueue()` (and co) instead of comments?

React/Modules/RCTTiming.m
```diff
@@ -171,28 +171,35 @@ - (dispatch_queue_t)methodQueue
   return RCTJSThread;
 }
 
+// Called from JS Thread.
```
shergin: I am concerned that it's always true.

React/Modules/RCTTiming.m
zhongwuzw: @shergin 🤔 Emm, actually, these two method can be called on any threads, because we always dispatch work to JS Thread via `dispatchBlock:queue`. Should we need to add it?

React/Modules/RCTTiming.m
zhongwuzw: 🤔 For the old module system, seems we use this rule to do things, for example : https://github.com/facebook/react-native/blob/master/React/Modules/RCTUIManager.m#L97-L101, https://github.com/facebook/react-native/blob/ac7ec4602fd4d8ec033a084df6a92437ea2d3a8a/React/Modules/RCTAlertManager.m#L42 .... , but yeah TM may change it which call `invalidate` always on `JS Thread`. I'll add an `dispatch` here. :)

React/Modules/RCTTiming.m
zhongwuzw: I removed the comments. We allow it to be called on any thread. :)

zhongwuzw:
@shergin Yes, it has, more info please see https://github.com/facebook/react-native/pull/23674, https://github.com/facebook/react-native/pull/21211#issuecomment-443461199, the thread issue may happened when app go into background, open a dialog or something, the reason is we add the notification to observe app state, and its handler would be called on main thread, but we need to put all things to JS thread. Scheduling the block to JS thread only happened in cases like app goes into background, it's the rare case, most of the time, for the app running in foreground, we always timing on JS thread, not need any locks.

PR #25175, opened 2019-06-06
Bump CLI to 2.0.0-rc.0

thymikee:
## Summary

Upgrading the CLI to the latest with a bunch of fixes and features included.

## Changelog

[General] [Changed] - Bump CLI to 2.0.0-rc.0

## Test Plan

CI passes


jest/hasteImpl.js
```diff
@@ -9,6 +9,7 @@
 
 'use strict';
 
+// TODO: remove this file entirely
```
cpojer: Can you remove the whole file in this update? Why not?

thymikee:
@cpojer you said something? 😛 

image

jest/hasteImpl.js
thymikee: Yep, trashed it!

Issue #25191, opened 2019-06-07
[iOS] RCTRootView.cancelTouches() does not immediately dispatch cancellation to JS


iain-merrick-fanduel:
The [`RCTRootView.cancelTouches()`](https://github.com/iain-merrick-fanduel/react-native/blob/2b262ec78b48a7aaa164125cadc4f0593b71fdf3/React/Base/RCTRootView.h#L129) function is provided to handle cases such as a React view embedded in a native `UIScrollView`, where React gestures should be cancelled during scrolling.

This correctly prevents in-flight React gestures from completing. However, it does not cancel in-flight gestures immediately (on the JS side, [`View.props.onResponderTerminate()`](https://facebook.github.io/react-native/docs/view#onresponderterminate) is not invoked immediately). The cancellation doesn't take effect until all touches have ended.

React Native version:

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
      Memory: 18.11 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.15.0 - ~/.nvm/versions/node/v8.15.0/bin/node
      Yarn: 1.15.2 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v8.15.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        Build Tools: 23.0.1, 26.0.3, 27.0.3, 28.0.1, 28.0.2, 28.0.3, 29.0.0
        API Levels: 21, 22, 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      @storybook/react: ~5.0.0 => 5.0.11 
      @types/react: ^16.8.2 => 16.8.2 
      react: ^16.8.1 => 16.8.1 
      react-native: 0.59.0 => 0.59.0 
```

## Steps To Reproduce

1. Create a RN `View` with an `onResponderTerminate` callback.
2. Embed it in a native `UIScrollView`.
3. Add on `onScroll` delegate that calls `RCTRootView.cancelTouches()`.

[...]

PR #25193, opened 2019-06-07
Ensure RCTRootView.cancelTouches() immediately sends termination event to JS

iain-merrick-fanduel:
Addresses https://github.com/facebook/react-native/issues/25191

## Summary

When this function is called, we force our gesture handler to cancel. However, we don't send a termination event to JS until `reset()` is called, and iOS does not automatically call `reset()` until all touches have ended. This means the JS code is unable to detect or respond to the cancellation immediately.

This PR fixes the problem by calling `reset()` manually after the gesture is cancelled. This is safe to do as `reset()` is idempotent.

## Changelog

[iOS] [Changed] - RCTRootView.cancelTouches() now takes effect immediately

## Test Plan

Tested manually.

### DIFF
--- a/React/Base/RCTTouchHandler.m
+++ b/React/Base/RCTTouchHandler.m
@@ -368,6 +368,7 @@ static BOOL RCTAnyTouchesChanged(NSSet<UITouch *> *touches)
 {
   self.enabled = NO;
   self.enabled = YES;
+  [self reset];
 }
 
 #pragma mark - UIGestureRecognizerDelegate


shergin:
I am afraid, `[RCTRootView cancelTouches]` is deprecated and should be deleted. I am surprised that it's not noted in the code. That's my fault and I will add the deprecation warning. 😳

Instead of using `cancelTouches` method, a standard `UIGestureRecognizer` conflict resolution process should be used.

shergin:
To more context this is internal related note from the year of 2017:

> Previously we had lots of super weird bugs where React Native would inaccurately recognize touch gestures which were meant to be addressed by the native environment. Usually, these bugs occurred as unintentional taps happening right after swipe gestures. In all of these cases, we had to manually call method `cancelTouches` as part of external gesture recognition process which prevented touch delivery to React Native. Furthermore, we had to delay touch delivery to React Native to wait for these cancellations. That code always looked like a hack (in the bad meaning of this word), like in some random place something dispatch event to another random place where something finally calls `cancelTouches`, yak. It was super annoying because it required adding this hack to all existing apps and screens, and because sometimes it was even too late to cancel touches.
>  
> We fixed that. Instead of delaying touch delivery and waiting for calls to `cancelTouches`, we set up the React Native gesture recognizer in such way that it always agrees to fail in favor of any native gestures (from non-RN-based and served views which are placed higher in a hierarchy). React Native will now cancel all active touches itself so that we no longer need to call `cancelTouches` manually. We already removed all these calls and supported code from Facebook and Instagram.
>

Issue #10088, opened 2016-09-24
“Error watching file for changes: EMFILE” when run the examples from facebook/react-native/Examples


aircloud:
I want to run the Examples from facebook/react-native/Examples in OS X 10.12.

First I use npm install and then use npm start:

There is an error:
```
Error watching file for changes: EMFILE
{"code":"EMFILE","errno":"EMFILE","syscall":"Error watching file for changes:","filename":null}
Error: Error watching file for changes: EMFILE
    at exports._errnoException (util.js:1008:11)
    at FSEvent.FSWatcher._handle.onchange (fs.js:1406:11)
```
Then I use Xcode run the code but there is the same error.
Need help or advice. Thanks very much.

p.s.  When I init a new project and then use `react-native run-ios` or Xcode , and there is the same error

iamthomasbishop:
Fwiw, I was running into the same issue, and this solved it: https://github.com/facebook/react-native/issues/910

aircloud:
Yeah, it was already solved

cbcye:
i use this command to resolve my problem
 sudo react-native start

brennanMKE:
You really do not want to use `sudo` regularly, especially with software which is under development. It gives that process unrestricted access to your computer and files which opens you up to major security risks. With all of the npm modules in your Node.js you cannot know that one of them does not contain a bug which could accidentally damage files on your computer or even do this maliciously.

What you want to do is increase the maximum allowed number of file descriptors. This command will show you the current maximum.

`ulimit -n`

Adding a number will set the maximum to the new value. This will allow you to run the `react-native` command without `sudo`.

`ulimit -n 2560`

parassahu:
If you have homebrew earlier and you have upgrade to Sierra, then it can solve your problem. Run this in your terminal, then again run react-native build.

`sudo chown -R $(whoami) /usr/local`

zthornto:
If anyone else comes across this, installing watchman through brew ended up fixing things for me. Can't explain why, just glad it's working.

SupriyaKalghatgi:
None of the mentioned solutions helped me

brennanMKE:
@SupriyaKalghatgi This may help. It appears to have helped other people already.

https://gist.github.com/brennanMKE/f6aa55b452ecda2f4c7a379e21647c88

## React Native Trouble

[...]

PR #25241, opened 2019-06-12
[Internal] [Fix] - Bump CLI to 2.0.0-rc.2

thymikee:
## Summary

Fixing Metro validation issue. Hope that helps with fixing the CI 🤞.

## Changelog

[Internal] [Fix] - Bump CLI to 2.0.0-rc.2

## Test Plan

Fixed CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,9 +82,9 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@react-native-community/cli": "2.0.0-rc.1",
-    "@react-native-community/cli-platform-android": "2.0.0-rc.1",
-    "@react-native-community/cli-platform-ios": "2.0.0-rc.1",
+    "@react-native-community/cli": "2.0.0-rc.2",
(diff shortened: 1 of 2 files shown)


thymikee:
Looks like this is a regression in `jest-haste-map` introduced in this PR: https://github.com/facebook/jest/pull/8056/files#diff-5c5055968e3661a18f280d5e3f10c969L738

The code was wrong earlier as well but looks like we just didn't hit the edge case. We should delete the `mapper` from `crawlerOptions` when invoking `nodeCrawl`.

cc @scotthovestadt

thymikee:
Now we hit 
```
Error: Error watching file for changes: EMFILE
    at FSEvent.FSWatcher._handle.onchange (fs.js:1359:9)
```

We have at least 2 options now:
1. Use watchman on CI 
2. Increase the watcher limit: https://github.com/facebook/react-native/issues/10088#issuecomment-265242851

PR #25274, opened 2019-06-16
[Android] Keyboard accessibility follow up

sweggersen:
## Summary

This is a follow up PR to #24359. There's a good thread in the mentioned PR for more background for why I'm doing this change. Essentially `focusable` makes more sense since it is about whether a view can receive user-initiated focus from a pointer or keyboard.


sweggersen:
@cpojer @necolas FYI

PR #25286, opened 2019-06-17
>=16.8.3 version of react

a-c-sreedhar-reddy:
## Summary

Showing warn while npm install even though latest version of react is installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "test-ios": "./scripts/objc-test-ios.sh test"
   },
   "peerDependencies": {
-    "react": "16.8.6"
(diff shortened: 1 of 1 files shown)


ide:
In the past, using different patch versions of React has led to bugs with React Native. Unless the React-to-React Native interface also strictly follows semver, the most correct approach is to use the exact version of React in your project i.e. fix the version warning on your end.

a-c-sreedhar-reddy:
So you mean to say that even though react upgrades react-native should use the same version specified in deps?

ide:
Yes, make sure your package.json is compatible the peerDep semver range.

Issue #25297, opened 2019-06-18
react-native 0.59 Some language wrapped texts are cut off on Android


soh335:
React Native version:

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
      Memory: 550.93 MB / 16.00 GB
      Shell: 5.5.1 - /usr/local/bin/zsh
    Binaries:
      Node: 10.10.0 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 25, 26, 27, 28
        Build Tools: 23.0.1, 27.0.3, 28.0.1, 28.0.3
        System Images: android-23 | Intel x86 Atom_64, android-23 | Google APIs Intel x86 Atom_64, android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.8 => 0.59.8
```

## Steps To Reproduce

1. Add text component with wrapped text (Japanese or Chinese, not Enlgish).
2. Wrapped texts are cut off.

<img width="672" alt="スクリーンショット 2019-06-18 18 55 06" src="https://user-images.githubusercontent.com/31937/59672869-4875a880-91fb-11e9-8dcd-b8a229c7f3fb.png">

Describe what you expected to happen:

Wrapped texts are not cut off.

This problem may be related to https://github.com/facebook/react-native/issues/24837. But happen not only RTL language.

After changing to target sdk 27, this problem is not happen.(https://github.com/facebook/react-native/issues/24837#issuecomment-492061717) 

<img width="672" alt="スクリーンショット 2019-06-18 18 54 38" src="https://user-images.githubusercontent.com/31937/59673600-82937a00-91fc-11e9-97f9-00062e92f260.png">

Snack, code example, or link to a repository:

[...]

Issue #25275, opened 2019-06-17
In Android pie React Native text component not showing fully text in malayalam language, It breaks and does not show the bottom lines


sahadpareed:
In my case long text above 25 lines Malayalam language text content not fully showing in android pie devices, There is no any problem in ios devices also below pie api's versions android.
![Screenshot_20190616-115302](https://user-images.githubusercontent.com/42888039/59572878-b3c15c80-90cd-11e9-9dba-26880c080eaa.jpg)

linjson:
```
StaticLayout.Builder layoutBuild =
                                    StaticLayout.Builder.obtain(text, 0, text.length(), textPaint, (int) width)
                                            .setAlignment(alignment)
                                            .setLineSpacing(0.f, 1.f)
                                            .setIncludePad(mIncludeFontPadding)
                                            .setBreakStrategy(mTextBreakStrategy)
                                            .setHyphenationFrequency(Layout.HYPHENATION_FREQUENCY_NORMAL);
                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
                                //add this `setUseLineSpacingFromFallbacks`
                                layoutBuild.setUseLineSpacingFromFallbacks(true);
                            }
```

Issue #25155, opened 2019-06-05
Text Cut Off Unicode Android


dharmen1901:
Text gets cut off when writing in Hindi(Unicode) whereas goes well with English.
The same was working well with version 0.49.2

React Native Version :- 0.59.8

**React-Native version** 
info
  React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
      Memory: 916.05 MB / 3.78 GB
    Binaries:
      npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.3.0.0 AI-182.5107.16.33.5264788

Phone:- Samsung On6 , Samsung M30

```
render() {
        data=[
            {
                text:"ज. गु. श्री रामान्दाचार्य जी",
                style:{color:"#ff0000",textAlign:"center",fontSize:25}
            },
            {
                text:"Google LLC[5] is an American multinational technology company that specializes in Internet-related services and products, which include online advertising technologies, search engine, cloud computing, software, and hardware. It is considered one of the Big Four technology companies, alongside Amazon, Apple and Facebook.Google LLC[5] is an American multinational technology company that specializes in Internet-related services and products, which include online advertising technologies, search engine, cloud computing, software, and hardware. It is considered one of the Big Four technology companies, alongside Amazon, Apple and Facebook.",
                style:{color:"#0000ff",fontSize:25,margin:10,fontFamily: 'lucida grande',}
            },
            {
[...]

Issue #24837, opened 2019-05-14
RTL languages text gets cut-off vertically on Android


iRayan7:
RTL languages text gets cut-off vertically on Android (Pixel 2 and OnePlus 6). This is a new app just initialized to reproduce the issue. 

Arabic text, last line vertically cut off:
<img width="322" alt="Screen Shot 2019-05-14 at 3 27 46 AM" src="https://user-images.githubusercontent.com/24809389/57663012-cd691300-75fa-11e9-9e5c-cc06ad9c9caa.png">
Where it should be rendered like this:
<img width="322" alt="Screen Shot 2019-05-14 at 3 32 13 AM" src="https://user-images.githubusercontent.com/24809389/57663055-05705600-75fb-11e9-9a5d-8eb9e4621112.png">

Persian text, last line vertically cut off:
<img width="325" alt="Screen Shot 2019-05-14 at 3 33 20 AM" src="https://user-images.githubusercontent.com/24809389/57663081-25077e80-75fb-11e9-91d2-2ea86c4bef5d.png">
Where it should be rendered like this:
<img width="328" alt="Screen Shot 2019-05-14 at 3 33 13 AM" src="https://user-images.githubusercontent.com/24809389/57663090-2c2e8c80-75fb-11e9-9487-3adad2a1f1d7.png">

English text is rendered fine:
<img width="331" alt="Screen Shot 2019-05-14 at 3 36 13 AM" src="https://user-images.githubusercontent.com/24809389/57663113-4cf6e200-75fb-11e9-9bf0-1920801b1641.png">

Emulator: Pixel 2
Device: OnePlus 6
React Native version: 0.59.8

```
React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 422.23 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.1 - /usr/local/bin/node
      Yarn: 1.13.0 - ~/n/bin/yarn
      npm: 6.4.1 - ~/n/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 26, 27, 28
[...]

PR #25306, opened 2019-06-19
Fix some languages wrapped texts are cut off on android

soh335:
## Summary

Fix wrapped some languages (like Japanese, Chinese) texts are cut off on android. This p-r is based on @linjson [patch](https://github.com/facebook/react-native/issues/25275#issuecomment-502548807).

- related (maybe)
    - https://github.com/facebook/react-native/issues/25297
    - https://github.com/facebook/react-native/issues/25275
    - https://github.com/facebook/react-native/issues/24837
    - https://github.com/facebook/react-native/issues/25155

`setUseLineSpacingFromFallbacks` is recommended to set true on [document](https://developer.android.com/reference/android/text/StaticLayout.Builder#setUseLineSpacingFromFallbacks(boolean))

>For backward compatibility reasons, the default is false, but setting this to true is strongly recommended. It is required to be true if text could be in languages like Burmese or Tibetan where text is typically much taller or deeper than Latin text.

## Changelog

[Android] [Fixed] - Fix some languages wrapped texts are cut off.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextShadowNode.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextShadowNode.java
@@ -115,6 +115,9 @@ public class ReactTextShadowNode extends ReactBaseTextShadowNode {
               if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                 builder.setJustificationMode(mJustificationMode);
               }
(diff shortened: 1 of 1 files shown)


makovkastar:
Hello, thank you for working on this. Do you know what backward compatibility reasons prevented this flag to be set to true by default?

soh335:
@makovkastar Sorry, I don't know why default is false...

PR #25314, opened 2019-06-19
[iOS] [Added] – Bring back autolinking to the iOS template

thymikee:
## Summary

Bringing back iOS autolinking to the template after reverting: https://github.com/facebook/react-native/commit/261197d85705dad1a362cc4637aa308c0382dcbe#commitcomment-33479829

## Changelog

[iOS] [Added] – Bring back autolinking to the iOS template

## Test Plan

`test_ios_e2e` should still pass


thymikee:
Yay, iOS e2e is passing again, cc @hramos

Issue #25339, opened 2019-06-20
[iOS] Inconsistent behaviour when converting homogeneous arrays as native call arguments


gavrix:
On iOS, when calling native method that accepts array as argument, automatic conversion of values inside array is different in Release and Debug configuration. What works in Debug crashes in Release

React Native version:
  React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
      Memory: 663.40 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.11.0 - ~/.nvm/versions/node/v10.11.0/bin/node
      Yarn: 1.13.0 - ~/.dev/yarn/1.13.0/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v10.11.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    IDEs:
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.8 => 0.59.8 

## Steps To Reproduce

Call `AsyncStorage.set('key', 123)`
Above works in Debug and crashes in Release

## Describe what you expected to happen:

Above works in debug as `number` (although types suggest it explicitly expects `string`) gets automatically converted to string via sequence of calls ending with `RCTConvertArrayValue` which converts converts anything to expected type (which is `NSString`) https://github.com/facebook/react-native/blob/ffa3b0d4d601fe6788319a7cfd4185b8e4bf462f/React/Base/RCTConvert.m#L558

in Release sequence of macro calls is different though, never reaching `RCTConvertArrayValue` as can be seen here https://github.com/facebook/react-native/blob/ffa3b0d4d601fe6788319a7cfd4185b8e4bf462f/React/Base/RCTConvert.m#L591-L596

What I think is expected behaviour is that in Debug it should log error descriptions of types mismatches but never convert anything to NSString automatically.

Were there any considerations as to why automatic conversion to string was added that I am missing?

cpojer:
[...]

PR #25340, opened 2019-06-20
[iOS] Remove debug version of RCT_JSON_ARRAY_CONVERTER_NAMED

gavrix:
## Summary
This is attempt to fix issue described in https://github.com/facebook/react-native/issues/25339
In short, value conversion of `NSArray<String>` should work the same way in Release and Debug configurations.

## Changelog

[iOS] [Fixed] - Array of numbers (`NSNumber`) is automatically (mistakenly) converted to `NSString` but only in Debug.

## Test Plan

`AsyncAtorage.setItem('key', 123)` behaviour:
Release: crash
Debug: red screen with exception that reads:
```
Exception '-[__NSCFNumber length]: unrecognized selector sent to instance 0xa8e477b3c1a00abf' was thrown while invoking multiSet on target AsyncLocalStorage with params (
        (
                (
            "key",
            123
        )
    ),
    96099
)

```

### DIFF
--- a/React/Base/RCTConvert.m
+++ b/React/Base/RCTConvert.m
@@ -588,11 +588,7 @@ RCT_ARRAY_CONVERTER(UIColor)
  * This macro is used for creating converter functions for directly
  * representable json array values that require no conversion.
  */
-#if RCT_DEBUG
-#define RCT_JSON_ARRAY_CONVERTER_NAMED(type, name) RCT_ARRAY_CONVERTER_NAMED(type, name)
-#else
 #define RCT_JSON_ARRAY_CONVERTER_NAMED(type, name) + (NSArray *)name##Array:(id)json { return json; }
-#endif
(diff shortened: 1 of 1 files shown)


sammy-SC:
Hey @gavrix 

According to documentation `AsyncStorage.setItem` is only meant to accept string as value

> https://facebook.github.io/react-native/docs/0.59/asyncstorage#setitem

cpojer:
AsyncStorage was extracted from RN and now lives in a separate repository here: https://github.com/react-native-community/async-storage

gavrix:
@sammy-SC please read carefully. I know that AsyncStorage expects array of strings and I'm sending it number. That is not an issue. The issue is that React Native converts number behind the scenes to string but only on Debug configuration. 

We made a bug which only surfaced on users device — that is an issue in ReactNative runtime. AsyncStorage was used _only_ to illustrate it, could have been any other native module with method that accepts array of strings. I attempted to make Release and Debug work the same way.

sammy-SC:
hey @gavrix, 

You were right that I misunderstood the problem :) sorry.
We are looking at the issue and possible solutions.

Issue #6480, opened 2016-03-16
TouchableNativeFeedback's ripples aren't affected by borderRadius on Android 9


jeffchienzabinet:
This is a bit of a problem when creating buttons with borderRadius and TouchableNativeFeedback, and the ripples extend out. I'm not an expert but I suppose setNativeBackground needs to get the corner radius when making the mask for the RippleDrawable and setBorderRadius may need to remake the background drawable somehow.

jonathanpalma:
Attached the following images to represent graphically what is proposed

![before](https://cloud.githubusercontent.com/assets/12414771/15800623/623d4082-2a3b-11e6-9891-e57bbe4a4c5c.gif)
![after](https://cloud.githubusercontent.com/assets/12414771/15800622/623aebb6-2a3b-11e6-8c6e-589f3bbc7df8.gif)

Images were taken from <a href="http://michaelevans.org/blog/2015/05/07/android-ripples-with-rounded-corners/">Michael Evans Blog - Android Ripples With Rounded Corners</a>

antoinerousseau:
It's supposed to be implemented in https://github.com/facebook/react-native/pull/6515 but I can't make it work :/

antoinerousseau:
I realize that PR got reverted

reyesr:
FWIW, as of 0.30, the ripple effect works with respect to the borderRadius of its parent <View>, but it requires its borderless effect to be set to true.
Eg:

`<View style={{borderRadius: 24, backgroundColor: '#F66', width: 48,height: 48}}>
    <TouchableNativeFeedback background={TouchableNativeFeedback.Ripple('#AAF', true)}>
[snip]
    </TouchableNativeFeedback>
</View>
`
If this is the intented effect, it should probably be made explicit, and added to the documentation. It it's a side effect (no pun intented), #6515 still makes sense.

tituswoo:
@reyesr OMG thank you so much for this!  👍

jonathanpalma:
@reyesr IT WORKS!!! Thank you so much buddy!! :raised_hands:

ericvicenti:
Looks like this is resolved. Feel free to open a new issue if you're still having troubles with it

satya164:
Re-opening since this issue has re-surfaced on Android P.

ferrannp:
Just to illustrate from Android P:

[...]

PR #25342, opened 2019-06-21
Set rounded rectangle mask on TouchableNativeFeedback's ripples

nhunzaker:
## Summary

This commit fixes an issue where ripple touch feedback extends beyond the border radius of a view.

### The fix

It achieves this by adding a mask to the RippleDrawable background, collecting that information from two new methods on ReactViewGroup:

1. getBorderRadiusMask() returns a drawable rounded rectangle matching the view's border radius properties
2. getBorderRadius() produces a float[] with the border radius information required to build a RoundedRectShape in getBorderRadiusMask()

Additionally, this commit updates setBorderRadius in ReactViewManager to re-apply the background whenever it is set, which is necessary to update the mask on the RippleDrawable background image as the border radius changes.

Related issues:
https://github.com/facebook/react-native/issues/6480

## Changelog

[Android][fixed] - Adding border radius styles to TouchableNative react-native run-android --port <x> correctly connects to dev server and related error messages display the correct port

## Test Plan

Link this branch to a new React native project with the following App.js class:

``` 
import React, { Component } from "react";
import { StyleSheet, Text, View, TouchableNativeFeedback } from "react-native";

  render() {
    const ripple = TouchableNativeFeedback.Ripple("#ff0000");

    return (
        <TouchableNativeFeedback background={ripple}>
          <View
            style={{
              width: 96,
              borderRadius: 12,
              borderTopLeftRadius: 10,
              borderBottomRightRadius: 37,
              height: 96,
              alignItems: "center",
              justifyContent: "center",
              borderColor: "black",
              borderWidth: 2
            }}
          >
            <Text>{"CLICK CLICK"}</Text>
          </View>
        </TouchableNativeFeedback>
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: "center",
    alignItems: "center",
    backgroundColor: "#F5FCFF"
  }
});
```

It's important to ensure that updates to border radius are accounted for. I did this by enabling hot module reloading and updating the border radius styles to verify that the ripple remains correct.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
@@ -52,6 +52,9 @@ public class ReactViewManager extends ViewGroupManager<ReactViewGroup> {
   private static final int CMD_SET_PRESSED = 2;
   private static final String HOTSPOT_UPDATE_KEY = "hotspotUpdate";
 
+  private ReadableMap mBg = null;
+  private ReadableMap mFg = null;
+
   @ReactProp(name = "accessible")
   public void setAccessible(ReactViewGroup view, boolean accessible) {
     view.setFocusable(accessible);
@@ -118,6 +121,14 @@ public class ReactViewManager extends ViewGroupManager<ReactViewGroup> {
     } else {
       view.setBorderRadius(borderRadius, index - 1);
     }
+
+    if (mBg != null) {
+      setNativeBackground(view, mBg);
+    }
+
+    if (mFg != null) {
+      setNativeForeground(view, mFg);
+    }
   }
 
   @ReactProp(name = "borderStyle")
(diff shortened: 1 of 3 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java
```diff
@@ -154,16 +165,18 @@ public void setPointerEvents(ReactViewGroup view, @Nullable String pointerEvents
 
   @ReactProp(name = "nativeBackgroundAndroid")
   public void setNativeBackground(ReactViewGroup view, @Nullable ReadableMap bg) {
+    mBg = bg;
```
nhunzaker: Is there a better way to track the current background style object passed into the React Component? I need to keep track of the last background style to reapply the ripple drawable with new border radius styles.

I wasn't sure if  assigning this to a variable was the best approach, but I couldn't find another way to do it.

nhunzaker:
CI Failing with:

> /opt/ndk/android-ndk-r20/toolchains/llvm/prebuilt/linux-x86_64/lib64/clang/8.0.2 doesn't exist.

Huh. Maybe a fluke, upstreaming to retrigger the build.

makovkastar:
Thanks a lot for working on it, looks good! I've made a few changes to add the RTL support for the `borderRadius` prop (so that `borderBottom[Top]StartRadius` or `borderBottom[Top]EndRadius` are not ignored when creating the RippleDrawable mask).

Regarding storing the background and foreground ReadableMaps in class member variables, I don't know another way to do it, so let's keep it.

nhunzaker:
> borderBottom[Top]StartRadius or borderBottom[Top]EndRadius

Totally forgot about these. Great!

grabbou:
Heads up: This PR introduced regression: https://github.com/facebook/react-native/issues/26544. We are reverting it.

Does the regression look anything similar that could be caused by the changes in this commit?

mjmasn:
Aside from the background/elevation regressions this would be a really nice improvement. Not sure if this helps but I found the following in my testing:

With `TouchableNativeFeedback.Ripple('#ff0000')` if `overflow: 'hidden'` is not added to the view then the red ripple is masked correctly but what I assume is the standard greyish ripple is also shown and is visible in the corners outside the border radius.

With `TouchableNativeFeedback.Ripple('#ff0000', true)` even with `overflow: 'hidden'` it just renders the entire ripple circle and no mask is applied.

`TouchableNativeFeedback.SelectableBackground()` also doesn't mask correctly even with `overflow: 'hidden'` (although maybe this is expected behaviour?)

For `TouchableNativeFeedback.SelectableBackgroundBorderless()` with `overflow: 'hidden'` the ripple is masked correctly but both with or without the `overflow: 'hidden'` the View's background colour disappears (or changes to white?).

Plus if the view or any (later?) sibling views have elevation, then the elevation shadow is rendered in front of rather than behind the view.

satya164:
This is definitely a much needed fix. But the background related regression is much worse. So it's not possible to keep this commit right now. A new PR based on this which addresses the regression is very welcome.

PR #25358, opened 2019-06-22
bump fresco to 2.0.0, supports AndroidX

dulmandakh:
## Summary

Bump Fresco to 2.0.0, which supports AndroidX. We should cherry-pick to 0.60 release, to support brown field apps but also native components.

## Changelog

[Android] [Changed] - Bump Fresco to 2.0.0, supports AndroidX

## Test Plan

CI is green, and RNTester builds and runs as expected.


dulmandakh:
@cpojer @kelset I would like to include this change in 0.60 release, as it uses Fresco with AndroidX support.

Issue #25318, opened 2019-06-19
Android: Animated.View opacity does not animate from 1 when using the native driver


guiccbr:
Hello

For some reason, it seems that an `Animated.View` does not animate opacity when the animated value starts from 1 when using the native driver. When using 1.1 or 0.9 (instead of exactly 1) - or when not using the native driver - the animation works fine.

### Steps To Reproduce

1. Have an animated value that goes from 1 to 0, as in the code below:
```
...
this.anim = new Animated.Value(1);
...
Animated.timing(this.anim, {
    duration: Constants.animation.duration,
    toValue: 0,
    useNativeDriver: true,
}).start();
```

2. Use the animated value as the opacity of an `Animated.View`. Something like:
```
<Animated.View style={{ opacity: this.anim }}>
  <Text>{text}</Text>
</Animated.View>
```

I'd expect the Animated View to fade out when the animation runs, but no opacity change happens.

If a number larger or smaller than 1 is used, the animation works. Not reproducible on iOS.

### Snack:
https://snack.expo.io/H1jsre_kB

### React Native version:

```
info
  React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
      Memory: 91.55 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 25, 26, 27, 28
        Build Tools: 26.0.0, 26.0.2, 26.0.3, 27.0.3, 28.0.0, 28.0.2, 28.0.3
        System Images: android-27 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.8 => 0.59.8
    npmGlobalPackages:
[...]

PR #25361, opened 2019-06-22
Set collapse property before starting an animation that uses the native driver

cabelitos:
## Summary

Depending on the style props of an Animated.View it may be optimised away
by the NativeViewHierarchyOptimizer, which will make the animation to
fail, because the native view is virtual (it does not exists
in the native view hierarchy).
Although the createAnimatedComponent already sets the collapsable property
based on the this._propsAnimated.__isNative flag, it won't work on all
cases, since the __isNative flag is only set when one starts the animation.
Which won't cause a re-render to occuor, thus not setting the collapsable
property to false.
In order to prevent this issue the HOC will just set the collapsable property
to false.

## Changelog
[Javascript] [Fixed] - Properly set collapsable to false before starting a nativeDriver animation

### **Without this patch:**
Run the following App on an Android device without this patch and click start.
Outcome: The animation **will not** make the text invisible.

### **With this patch:**

Run the following App on an Android device with this patch and click start.
Outcome: The animation **will** make the text invisible.

```javascript
import React, { Component, ReactNode } from 'react';
import { View, Text, TouchableOpacity, Animated, StyleSheet, Easing } from 'react-native';

interface Props { }

const Constants = {
  animation: {
    duration: 500,
  },
};

const text =
  'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla sed orci erat. Suspendisse feugiat elit gravida elit consequat ultrices. Sed sollicitudin ullamcorper molestie. Mauris a diam neque. Vivamus in lectus.';

class App extends Component<Props> {
  anim: any;

  constructor(props: Props) {
    super(props);
    this.anim = new Animated.Value(0);
  }

  handleStartPress = () => {
    this.anim.setValue(0);
    console.log('start');
    Animated.timing(this.anim, {
      duration: Constants.animation.duration,
      toValue: 1,
      easing: Easing.linear(),
      useNativeDriver: true,
    }).start();
  };

  render(): ReactNode {
    return (
        <Animated.View
          style={{
            opacity: this.anim.interpolate({
              inputRange: [0, 1],
              outputRange: [1, 0],
            }),
          }}>
          <Text>{text}</Text>
        </Animated.View>
        <TouchableOpacity
          style={styles.startButton}
          onPress={this.handleStartPress}>
          <Text style={styles.startButtonText}>START</Text>
        </TouchableOpacity>
      </View>
    );
  }
}


  container: {
    alignItems: 'center',
    backgroundColor: 'white',
    flex: 1,
  },
  description: {
    marginTop: 20,
    paddingHorizontal: 10,
  },
  startButton: {
    alignItems: 'center',
    aspectRatio: 1,
    backgroundColor: 'yellow',
    borderRadius: 100,
    height: 50,
    justifyContent: 'center',
  },
  startButtonText: {
    fontSize: 10,
    fontWeight: 'bold',
  },
});

export default App;
```

Closes https://github.com/facebook/react-native/issues/25318

### DIFF
--- a/Libraries/Animated/src/createAnimatedComponent.js
+++ b/Libraries/Animated/src/createAnimatedComponent.js
@@ -156,9 +156,7 @@ function createAnimatedComponent(Component: any, defaultProps: any): any {
           // have to make sure the view doesn't get optimized away because it cannot
           // go through the NativeViewHierarchyManager since it operates on the shadow
           // thread.
-          collapsable={
-            this._propsAnimated.__isNative ? false : props.collapsable
-          }
+          collapsable={false}
         />
       );
     }


janicduplessis:
@cabelitos I think we can make this simpler and just always set `collapsable` to `true` instead of checking `__isNative` https://github.com/facebook/react-native/blob/da903c1547f3f6517c8458d8f690a5613a698724/Libraries/Animated/src/createAnimatedComponent.js#L159.

Animated views should never really be collapsable since they will end up with some props that make them non-collapsable when an animation starts.

cabelitos:
Hello and thanks for you review @janicduplessis.

At first I tought about harding coding the  `collapsable` to `false` which avoids this `NativeViewHierarchyOptimizer` optimization. However I did not want to break the existing behavior or making performance worse in some cases. But I'm happy to just hard code the `collapsable` to `false` if you guys prefer.

janicduplessis:
I don't think it will have a perf impact since usually animated views will have a non-collapsable prop. Also there will be an overhead anyway when the animation starts when the view gets marked as non-collapsable and we have to update the view again.

cabelitos:
@janicduplessis ok then. I'll update the PR hardcoding the `collapsable` prop

cabelitos:
@janicduplessis updated the PR, please let me know what do you think.
thank you.

cpojer:
Just want to double check here: This changes `collapsible` to always be `false` but @janicduplessis was saying in his original comment that we should set it to `true`. I'm a bit confused here, please help.

janicduplessis:
@cpojer Oups my bad I meant false

cabelitos:
Hey guys, thank you for the approval.
@cpojer about the `collapsible` being hardcoded to `false`. If we set it to `true` the `NativeViewHierarchyOptimizer` would not create the backing native view, which would cause the native to won't work with the native driver. So I assumed that @janicduplessis did a typo in his comment and he meant to say `false` and not `true`.

Thank you very much for the support!

cpojer:
Unfortunately this did cause issues in production on Android where things didn't appear properly so I had to revert this change for now :(

sintylapse:
@cpojer could you please specify in which circumstances this fix cause issues? maybe not everyone will have those issues but animation will be fixed for them

cabelitos:
Good morning @cpojer,

if possible could you please share what's the problem and how I can trigger it. I'm glad to send a fix for the problem that is affecting you.

Thank you very much and sorry for any problem that I have caused.

vdmtrv:
@cpojer as @cabelitos has mentioned, can you elaborate on the problem, please? This still seems to be an issue on 0.60.4

PR #5223, opened 2016-01-09
Implement countdown mode for DatePickerIOS

dapetcu21:
I needed this for my project and I thought I'd share the implementation. I'm open to shotgun fire :smile:

PR #25373, opened 2019-06-24
[iOS] Add workaround for countdown mode where first event would not fire

jesperjohansson:
## Summary
The change event would not fire the first time the value is updated if you use the mode "countdown" (`UIDatePickerModeCountDownTimer`) because of an iOS bug. This is a workaround that asynchronously sets `countDownDuration` in the `setDatePickerMode` function if its in countdown mode and makes the change event run.

Links describing the issue:
https://github.com/romaonthego/RETableViewManager/issues/161
https://github.com/facebook/react-native/pull/5223#issuecomment-351510842

## Changelog

[iOS] [Fixed] - Fix first change event not firing for RCTDatePicker in countdown mode

## Test Plan

Below are GIF:s demonstrating the change. I temporarily added a datepicker with countdown mode in the example file for this demonstration, as countdown mode doesn't seem to be fully supported yet.

[Before (date doesn't change first time)](https://user-images.githubusercontent.com/21033162/60018599-17c6c080-968c-11e9-8a09-156acb333dda.gif)
[After (date changes both times)](https://user-images.githubusercontent.com/21033162/60018600-17c6c080-968c-11e9-9a6c-114908159470.gif)

Below is the code I added for the demonstration (in [DatePickerIOSExample.js](https://github.com/jesperjohansson/react-native/blob/fix/datepicker-countdown-mode-onchange/RNTester/js/examples/DatePicker/DatePickerIOSExample.js#L149)):
```javascript
  {
    title: 'Countdown picker',
    render: function(): React.Element<any> {
      return (
        <WithDatePickerData>
          {(state, onDateChange) => (
            <DatePickerIOS
              testID="countdown"
              date={state.date}
              mode="countdown"
              onDateChange={onDateChange}
            />
          )}
        </WithDatePickerData>
      );
    },
  },
```

### DIFF
--- a/React/Views/RCTDatePicker.m
+++ b/React/Views/RCTDatePicker.m
@@ -43,6 +43,13 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
   [super setDatePickerMode:datePickerMode];
   // We need to set minuteInterval after setting datePickerMode, otherwise minuteInterval is invalid in time mode.
   self.minuteInterval = _reactMinuteInterval;
+
+  // Workaround for iOS bug where the first UIControlEventValueChanged doesn't fire in countdown mode.
+  if (datePickerMode == UIDatePickerModeCountDownTimer) {
+    dispatch_async(dispatch_get_main_queue(), ^{
+      self.countDownDuration = self.countDownDuration;
+    });
+  }
 }
 
 - (void)setMinuteInterval:(NSInteger)minuteInterval


cpojer:
Thank you for the PR! We are deprecating the DatePicker in favor of a new implementation here: https://github.com/react-native-community/react-native-datetimepicker

I would prefer not making any more changes to the current implementation and instead will defer to the new one.

jesperjohansson:
> Thank you for the PR! We are deprecating the DatePicker in favor of a new implementation here: https://github.com/react-native-community/react-native-datetimepicker
> 
> I would prefer not making any more changes to the current implementation and instead will defer to the new one.

Ah, that implementation looks good! I use Expo right now, so does this mean I'll have to eject if I want to use the new implementation and not have this bug? 😔

PR #25382, opened 2019-06-24
[iOS] Remove asl_log (Fixes #25380)

### LINKED ISSUE #25380: iOS: Everything get's logged twice
React Native version:
0.59.9

```
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
      Memory: 305.17 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.13.0 - /usr/local/bin/node
      Yarn: 1.15.2 - ~/.yarn/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.3, 27.0.2, 27.0.3, 28.0.3
        System Images: android-25 | Google APIs Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
    IDEs:
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.9 => 0.59.9
    npmGlobalPackages:
      create-react-native-app: 2.0.2
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.4
```

## Steps To Reproduce

1. Create a RN project using `react-native init`.

image

When you start the project, everything is logged twice.

JonnyBurger:
## Summary

By default absolutely everything gets logged twice (#25380).
This was introduced over 4 years ago here:

[React Native Log to ASL · facebook/react-native@d1b14ef · GitHub](https://github.com/facebook/react-native/commit/d1b14ef0627dc04847500faa7fef75fd3c22c900)

with the reason to support ASL.
With this PR the support for that will be removed but I believe this is justified because:

- the benefit of not having every log message twice far outweighs the benefit of having Apple System Log.
- ASL was superseded by "unified logging" (yellow box on https://developer.apple.com/documentation/os/logging).
- I assume that people who use asl_log is very small. There is also very little information about it on the internet

## Changelog

[iOS] [Fixed] - Logs would get printed twice

### DIFF
--- a/React/Base/RCTLog.mm
+++ b/React/Base/RCTLog.mm
@@ -7,7 +7,6 @@
 
 #import "RCTLog.h"
 
-#include <asl.h>
 #include <cxxabi.h>
 
 #import "RCTAssert.h"
--- a/React/Base/RCTLog.mm
+++ b/React/Base/RCTLog.mm
@@ -56,26 +56,6 @@ RCTLogFunction RCTDefaultLogFunction = ^(
   NSString *log = RCTFormatLog([NSDate date], level, fileName, lineNumber, message);
   fprintf(stderr, "%s\n", log.UTF8String);
   fflush(stderr);
-
-  int aslLevel;
-  switch(level) {
-    case RCTLogLevelTrace:
-      aslLevel = ASL_LEVEL_DEBUG;
-      break;
-    case RCTLogLevelInfo:
-      aslLevel = ASL_LEVEL_NOTICE;
-      break;
-    case RCTLogLevelWarning:
-      aslLevel = ASL_LEVEL_WARNING;
-      break;
-    case RCTLogLevelError:
-      aslLevel = ASL_LEVEL_ERR;
-      break;
-    case RCTLogLevelFatal:
-      aslLevel = ASL_LEVEL_CRIT;
-      break;
-  }
-  asl_log(NULL, NULL, aslLevel, "%s", message.UTF8String);
 };
 
 void RCTSetLogFunction(RCTLogFunction logFunction)


cpojer:
@JonnyBurger it seems we have tests for this. Can you check out `RCTLoggingTests` and remove the associated tests?

JonnyBurger:
I don't see immediately which tests have to be changed D:

Running out of time this week, I asked @Kida007 if he can take a look tomorrow otherwise will try to get it merge ready next week!

cpojer:
I *think* this is actually a CI issue and not an issue with your change. Let me try to land it.

aleclarson:
Printing to stderr is **not** good enough. We need to use `os_log` for the messages to be "captured" (and thus be viewable from Console.app).

aleclarson:
Hmmm, are we sure that `os_log` isn't necessary? I'm not seeing `fprintf` logs in Console.app, but it could just be a problem on my end.

JonnyBurger:
Sorry for causing trouble. 
Could we alternatively remove the other log statement instead?

So that it writes out and also prints to Console.app.

Maybe we can use 'unified logging' (see above) instead?

aleclarson:
IMO, we should ditch `fprintf` entirely for "unified logging" (eg: `os_log`), because it supports Console.app and [`log stream`](https://www.dssw.co.uk/reference/log.html), and it stores logs in binary format (so less disk space, yet accessible in Release mode).

Here's an implementation of `os_log` in React Native, if you want to submit a PR:
https://gist.github.com/aleclarson/03e9b1eb6ed6c6fae8539d4500360bf5

aleclarson:
The only issue is that it replicates the timestamp, so we should remove the `timestamp` argument from `RCTFormatLog` too.

Example:
```
2019-07-04 11:43:09.892 [info][tid:main][RCTCxxBridge.mm:216] Initializing <RCTCxxBridge: 0x101c01440> (parent: <LiteTimer.LTBridge: 0x600002908bd0>, executor: (null))
2019-07-04 11:43:09.892224-0400 LiteTimer[5928:31582] 2019-07-04 11:43:09.892 [info][tid:main][RCTCxxBridge.mm:216] Initializing <RCTCxxBridge: 0x101c01440> (parent: <LiteTimer.LTBridge: 0x600002908bd0>, executor: (null))
```

damandeepsingh93:
Now it doesnt Log in Console App. How to view logs in Release build now ?

JonnyBurger:
Sorry @damandeepsingh93 for breaking this.
Now I understand why this was there in the first place.

I think the correct behavior would then be to do asl_log in production and normal log in development?

PR #25383, opened 2019-06-24
[iOS] Make logs shorter

JonnyBurger:
## Summary

Currently all logs include current time, including timezone and nanoseconds, as well as the thread ID and the log level.

While it is detailed, I find that it is also very noisy. I rarely need those things.

- Most of the time timing is irrelevant, so I removed it completely. Developers can also add the current time to the logs themselves easily. I would at least remove nanaoseconds and timezone.

- Thread ID is often irrelevant (Prior to this PR I did not even know what `[tid:123]` meant). The only interesting case is when the Thread ID is com.facebook.react.JavaScript. Then I add [JS] in front of it.

- Log level I have also removed because as a developer I have very little control over it.

## Changelog

[iOS] [Changed] - Log format

## Test Plan

Tested it manually.

### DIFF
--- a/React/Base/RCTLog.h
+++ b/React/Base/RCTLog.h
@@ -71,8 +71,6 @@ typedef void (^RCTLogFunction)(
  * particular data from the log, just pass nil or zero for the argument.
  */
 RCT_EXTERN NSString *RCTFormatLog(
-  NSDate *timestamp,
-  RCTLogLevel level,
   NSString *fileName,
   NSNumber *lineNumber,
   NSString *message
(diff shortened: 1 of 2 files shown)


cpojer:
@JonnyBurger could you add a screenshot to compare the before/after?

JonnyBurger:
0.59.9 Log:

image


With this PR applied:

image

With this PR applied and #25382 applied:

image

With this PR, #25382, #25381 applied:

image

The last one missing would be to kill these nw_socket messages!

akshetpandey:
At least to me, this makes it harder to read the logs since the indent levels at which the important parts of the log start are now significantly different

cpojer:
I discussed this with others and it seems that people are preferring to keep the date/time stamp as is because it aligns with other logging on iOS. Since the logs already point out that it's a JS log, I let's keep it as is.

Note that the next version of RN + CLI will log `console.log` calls to the terminal where Metro is running.

PR #25392, opened 2019-06-25
[Fabric]: Remove cxx from iOS project

zhongwuzw:
## Summary

Remove cxx from iOS project, because we have the corresponding implementation in iOS. Otherwise, it would break the logical handling of iOS part. 

## Changelog

[iOS] [Fixed] - Remove cxx from iOS project

## Test Plan

Fabric enabled Tester can work like loading image.


shergin:
Those path fragments in `exclude_files` (sic!) sections.

zhongwuzw:
@shergin ? emm, I don't get what your point.

shergin:
Oops, seems I was very wrong about that. The changes actually make a lot of sense: we don't need `cxx` and `android` folders to compile those targets on iOS.

PR #25407, opened 2019-06-26
add `passwordRules` for `textContentType` `newPassword`

pvinis:
## Summary

On `textContentType` `newPassword` on ios, there is another property called `passwordRules` on ios 12 that can give hints to the os to generate a password with specific requirements like [here](https://developer.apple.com/password-rules/).
This is useful for apps that have a "register" screen with `emailAddress`/`username` and a `newPassword` fields, to let ios make a password that will satisfy the requirements and not one that might be not accepted after the user presses "register".

## Changelog

[iOS] [Added] - PasswordRules for new password textContentType input fields

## Test Plan

This is a bit harder, but to test you need to make an app that has associated domains with an apple-app-site-association file on that domain, enable iCloud Keychain on the test device, and then iOS will suggest a password, otherwise you will just get a warning on Xcode saying "Couldn't suggest password because of: blabla".

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -208,6 +208,8 @@ export type TextContentType =
   | 'newPassword'
   | 'oneTimeCode';
 
+type PasswordRules = string;
+
 type IOSProps = $ReadOnly<{|
   spellCheck?: ?boolean,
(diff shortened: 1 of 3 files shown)


pvinis:
```
Test Suite RNTesterIntegrationTests.xctest started
RCTLoggingTests
Too long with no output (exceeded 10m0s)
```
🤔

PR #25416, opened 2019-06-27
Use CocoaPods-based RNTesterPods for iOS tests

hramos:
Summary:
Use CocoaPods-based RNTesterPods workspace to run iOS unit tests and integration tests.

This is necessary as new iOS projects now use CocoaPods by default. CocoaPods also powers the new package auto-linking feature.

In order to provide test coverage for this new default configuration, our iOS tests are being migrated to use a CocoaPods-managed RNTester workspace. This applies to both Circle CI, and Sandcastle.

Reviewed By: fkgozali

Differential Revision: D15958209


scripts/packager.sh
```diff
@@ -10,11 +10,8 @@ REACT_NATIVE_ROOT="$THIS_DIR/.."
 # Application root directory - General use case: react-native is a dependency
 PROJECT_ROOT="$THIS_DIR/../../.."
 
-# shellcheck source=/dev/null
-source "${THIS_DIR}/.packager.env"
```
dmitryskorbovenko, line 14: @hramos what is a reason to remove it?

Issue #25181, opened 2019-06-06
☂️ Supporting Xcode 11 and iOS 13 Beta


hramos:
This task tracks issues with Xcode 11 and iOS 13. Both are currently in Beta, with an expected release date sometime in September 2019.

## Related Issues

- [Supporting Project Catalyst: Running iPad apps on macOS](https://github.com/react-native-community/discussions-and-proposals/issues/131)
- [Initial UIKitForMac support](https://github.com/facebook/react-native/pull/25427)
- [Initial UIWindowScene support](https://github.com/facebook/react-native/pull/25425)

## Known Issues

- [ ] [Compatibility with new UIScene, UISceneDelegate APIs](https://github.com/facebook/react-native/issues/25181#issuecomment-502081954)

**Fixed Issues**

- [x] [EXC_BAD_ACCESS in fishhook.c (iOS 13/Xcode 11) ](https://github.com/facebook/react-native/issues/25182)
- [X] [`__unused` signature changed](https://github.com/facebook/react-native/issues/25138)

lucasbento:
#25182 is apparently related.

chrisspankroy:
I've opened a PR for #25182 at #25212

radex:
- [ ] A lot of things in UIApplication / UIApplicationDelegate are deprecated and you're supposed to move to UIScene / UISceneDelegate (this is groundwork for supporting multiple windows on iPadOS and macOS -- and even if you don't, this is where the APIs will evolve in the future). I might be interested in sending some pull requests with the necessary changes (necessary for those willing to implement multiple window support in their own apps -- with full backwards compatibility so that most people don't have to do anything)

radex:
Copying over from a thread on discord (sorry about formatting):

I'm doing some digging around React Native core to prepare it for support for new UIScene based APIs — that's necessary to be able to support multiple windows on iPadOS (and therefore, multiple windows on macOS via Catalyst)

a problem I'm seeing is that there are many APIs that are RCTBridge-global — and assume the app has one window
for example, status bar is a NativeModule - https://github.com/facebook/react-native/blob/master/React/Modules/RCTStatusBarManager.m

this is a problem, because we need to be able to target a specific UIScene, not the whole UIApplication

I'm interested in helping out on this front as I'd like @Nozbe 4 to work nicely with multiple windows
[...]

PR #25425, opened 2019-06-28
[iOS] Moving towards UIWindowScene support

radex:
## Summary

I'm taking the first step towards supporting iOS 13 UIScene APIs and modernizing React Native not to assume an app only has a single window. See discussion here: https://github.com/facebook/react-native/issues/25181#issuecomment-505612941

The approach I'm taking is to take advantage of `RootTagContext` and passing it to NativeModules so that they can identify correctly which window they refer to. Here I'm just laying groundwork.

- [x] `Alert` and `ActionSheetIOS` take an optional `rootTag` argument that will cause them to appear on the correct window
- [x] `StatusBar` methods also have `rootTag` argument added, but it's not fully hooked up on the native side — this turns out to require some more work, see: https://github.com/facebook/react-native/issues/25181#issuecomment-506690818
- [x] `setNetworkActivityIndicatorVisible` is deprecated in iOS 13
- [x] `RCTPerfMonitor`, `RCTProfile` no longer assume `UIApplicationDelegate` has a `window` property (no longer the best practice) — they now just render on the key window

Next steps: Add VC-based status bar management (if I get the OK on https://github.com/facebook/react-native/issues/25181#issuecomment-506690818 ), add multiple window demo to RNTester, deprecate Dimensions in favor of a layout context, consider adding hook-based APIs for native modules such as Alert that automatically know which rootTag to pass

## Changelog

[Internal] [Changed] - Modernize Modal to use RootTagContext
[iOS] [Changed] - `Alert`, `ActionSheetIOS`, `StatusBar` methods now take an optional `surface` argument (for future iPadOS 13 support)
[Internal] [Changed] - Do not assume `UIApplicationDelegate` has a `window` property

## Test Plan

- Open RNTester and:
- go to Modal and check if it still works
- Alert → see if works
- ACtionSheetIOS → see if it works
- StatusBar → see if it works
- Share → see if it works


shergin:
I believe,
1. We should land this.
2. We should decouple all of this components from the core.
3. We should figure out a better API for this and migrate over it.

Speaking about the better API, I think we have to use some kind of reference to *any* view to specify a concrete surface, not a root one (so, the root one can be inferred from a view). Now we can use `reactTag` for classic react, but later we have to migrate to some non-numerical reference to the view (compare implementation of UIManager.measure in Paper and Fabric for reference).

@radex Radek, I would prefer if we fix this PR to use a view reference (not a root tag), and then land. But I am also okay with landing this as is now. LMK which approach you prefer.

hramos:
Thanks for taking a look, Valentin. @radex, I can help land this after the feedback is taken into account.

radex:
> Speaking about the better API, I think we have to use some kind of reference to any view to specify a concrete surface, not a root one (so, the root one can be inferred from a view). Now we can use reactTag for classic react, but later we have to migrate to some non-numerical reference to the view (compare implementation of UIManager.measure in Paper and Fabric for reference).

That makes a lot of sense to me! Let me make those changes :)

Libraries/Alert/Alert.js
```diff
@@ -31,6 +32,11 @@ export type Buttons = Array<{
 type Options = {
   cancelable?: ?boolean,
   onDismiss?: ?() => void,
+  surface?: mixed,
```
radex: I struggle to come up with the right name for this argument (also in other modified APIs).

 Taking @shergin's suggestions, it's not `rootTag` anymore, since: a) we don't really need the root, any component/view will do, and b) it shouldn't be a numeric tag for fabric forward compatibility.

Instead, I'm passing this to `ReactNative.findNodeHandle()` -- so a `ref.current` to any component will do, but you _can_ also pass a numeric tag if you prefer.

radex:
@hramos I took Valentin's feedback into account and revised the PR.

Also added some examples to RNTester & fixed up Flow types where I noticed they were wrong

radex:
@hramos did you have a chance to take a look at the PR? Let me know what I can do to get it landed

radex:
@hramos hey, this seems stuck on import for 18 days… is there something I can do to push this along?

jpshelley:
Whats the status of this PR? Its now quite out of date, I see its landed internally but not been released yet. What can we do to help? With the iOS requirements approaching, this seems vital

hramos:
I apologize for the huge delay here. I'm working on getting this merged.

hramos:
I resolved all conflicts and fixed some Flow issues. I still need to go through our internal apps and update some call sites. You can find an early copy of the internal diff with conflicts resolved in https://github.com/facebook/react-native/pull/28058. When the internal diff gets merged, it should attribute the change back to PR #25425.

hramos:
💔 We had to back 80e6d672f32fdc860c73eabcc63763dcab3c6269 out. Some of our own applications ran into issues with the changes; nothing that can't be fixed. I'll work on getting this change into the codebase again ASAP.

PR #25430, opened 2019-06-28
Remove incorrect constexpr specifier in YGLayout

selat:
Summary:
* According to C++ standard values which produce undefined behavior cannot be declared as `constexpr`.
* Expressions which evaluate to `nan, infinity, etc` are undefined behavior.

For more details you can checkout this Stackoverflow answer: https://stackoverflow.com/a/46373136

Differential Revision: D16055108


cpojer:
Thank you for your PR. Please send Yoga related PRs to the Yoga repository here: https://github.com/facebook/yoga

PR #25451, opened 2019-07-01
Add generated `xcshareddata` folder to gitignore

karanjthakkar:
## Summary

A new project created with `0.60.0-rc.3` generates an `xcshareddata` when you open the workspace in Xcode at `ios/ProjectName.xcworkspace/xcshareddata`. This does not need to be tracked in git.

## Changelog

[IOS] [ADDED] - Add generated `xcshareddata` folder to gitignore

## Test Plan

N/A

### DIFF
--- a/template/_gitignore
+++ b/template/_gitignore
@@ -21,6 +21,7 @@ DerivedData
 *.ipa
 *.xcuserstate
 project.xcworkspace
+xcshareddata
 
 # Android/IntelliJ
 #
--- a/template/_gitignore
+++ b/template/_gitignore
@@ -20,7 +20,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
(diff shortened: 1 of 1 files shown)


template/_gitignore
```diff
@@ -21,6 +21,7 @@ DerivedData
 *.ipa
 *.xcuserstate
 project.xcworkspace
```
satya164: I guess we can remove this line too since we're changing this file :D

karanjthakkar: @satya164 @kelset done! 🙂

vovkasm:
@karanjthakkar 

> This does not need to be tracked in git.

This phrase does not describe change. It is incorrect statement or should be explained. Why exactly this change?

https://stackoverflow.com/a/53039267/2266462

PR #25496, opened 2019-07-04
CocoaPods frameworks compatibility: Step 1

jtreanor:
## Summary

This is the first step towards fixing https://github.com/facebook/react-native/issues/25349. These are the changes to the podspec to correctly update dependencies and build config that will cause any breaking change for users or libraries.

I am breaking these changes out from https://github.com/facebook/react-native/pull/25393 as suggested by @fkgozali in https://github.com/facebook/react-native/pull/25393#issuecomment-508322884.

These are the changes:

- Made C++ headers in `React-Core` private by default so that ObjC files can import the module without failures.
- Reduced the number of `yoga` headers that are exposed for the same reason as above. As far as I can see this doesn't cause issues but we can find another solution if it does.
- Adding some missing dependencies to fix undefined symbols errors.
- Added `DoubleConversion` to `HEADER_SEARCH_PATHS` where it was missing.

## Changelog

[iOS] [Fixed] - Updated podspecs for improved compatibility with different install types.

## Test Plan

Everything should work exactly as before. I have a branch on my [sample project](https://github.com/jtreanor/react-native-cocoapods-frameworks) here which points at this branch to show that it is still working `Podfile` to demonstrate this is fixed.

2. `git checkout podspec-updates`
3. `cd ios && pod install`

The sample app will build and run successfully.


Zyphrax:
Is this going to be merged soon? It would be great to have this fix for all of us Swift users.

Andreyco:
Good work, I will test this branch agains my Swift package

TemaTian:
Very good, I really hope to merge soon.

boliveira:
Sorry to bother, but I applied these changes to my project and I get the error <React/RCTDefines.h> file not found in RCTWebSocketExecutor.h and use_frameworks! in Podfile. Any help someone can give? :)

jtreanor:
Hi @boliveira. That is expected at the moment. This PR only fixes the first part of the `use_frameworks!` issue. The full fix is being tracked in https://github.com/facebook/react-native/pull/25393.

fkgozali:
Btw, with this PR as is, there is a build failure (without any change) when using Podfile without use_frameworks:

```
..../ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm:16:9: 'React/RCTCxxModule.h' file not found
```

jtreanor:
Oh, thanks for the heads up. That should be an easy fix.

React/React-Core.podspec
```diff
@@ -41,14 +41,27 @@ Pod::Spec.new do |s|
...
   s.pod_target_xcconfig    = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/ReactCommon\" \"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/DoubleConversion\" \"$(PODS_ROOT)/Folly\"" }
+  s.default_subspec         = "Default"
+
+  s.subspec "Default" do |ss|
+    # no-op
+  end
+
+  s.subspec "CxxBridge" do |ss|
+    # Make the C++ headers visible if they are needed
+    ss.public_header_files   = "**/*.{h}"
```
fkgozali: According to https://guides.cocoapods.org/syntax/podspec.html#private_header_files, it looks like `private_header_files` is computed vs `public_header_files`, so doing it this way ended up still excluding the private_header_files?

> These patterns are matched against the public headers (or all the headers if no public headers have been specified) to exclude those headers which should not be exposed to the user project and which should not be used to generate the documentation. When the library is built, these headers will appear in the build directory.

fkgozali:
> Btw, with this PR as is, there is a build failure (without any change) when using Podfile without use_frameworks:

To clarify, I was building RNTesterPods, which includes turbomodules by default already. Let's make sure RNTesterPods works with this PR as well.

fkgozali:
> Oh, thanks for the heads up. That should be an easy fix.

Looks like I had a wrong file opened, so it was indeed a simple fix. I'll just fix it within FB import so we can land this sooner.

React/React-Core.podspec
fkgozali: Nvm, this seems to work. I edited the wrong file when verifying.

fkgozali:
I think Fabric related podspecs aren't compatible with this change (when TM is installed), but let's deal with that later, since Fabric isn't enabled by default yet.

jtreanor:
@fkgozali On merging `master` into my branch, I saw the same error you posted above (`'React/RCTCxxModule.h' file not found`), so it seems it was a legitimate error.

I have fixed it by updating `React-turbomodule-core.podspec` to include `React-Core/CxxBridge`.

> I think Fabric related podspecs aren't compatible with this change (when TM is installed), but let's deal with that later, since Fabric isn't enabled by default yet.

Sure. I'm not sure what has changed there but I'm happy to follow up as needed.

vivekhumnabad:
When will my fix make it into a release?

iOSUser110:
When will get the update ??

PR #25497, opened 2019-07-04
Fix (tvos): On press and focus missing for the buttons in redbox

nbennink:
## Summary
The buttons in the new Redbox are not working on tvOS. You can see the hover effect of the button but onPress function is not triggered. `UIControlEventTouchUpInside` was used to handle it but this is not working with the tvOS remote. I believe that `UIControlEventPrimaryActionTriggered` is a better alternative to use both.

I have also added a focus state that makes it possible to see the selected button.

## Changelog
[iOS] [Fixed] - Fixed the press event of the buttons in redbox.

## Test Plan
It's actually quite easy to test. Create an error in debug mode and try to use the buttons on the bottom. These were not working before and it was not possible to see the selected button.

### DIFF
--- a/React/Modules/RCTRedBox.m
+++ b/React/Modules/RCTRedBox.m
@@ -94,7 +94,8 @@
         [dismissButton setTitle:dismissText forState:UIControlStateNormal];
         [dismissButton setTitleColor:[UIColor whiteColor] forState:UIControlStateNormal];
         [dismissButton setTitleColor:[UIColor colorWithWhite:1 alpha:0.5] forState:UIControlStateHighlighted];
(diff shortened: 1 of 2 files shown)


cpojer:
@nbennink what's the status of this pull request?

nbennink:
@cpojer I can to merge master to  this pull request to confirm that it's working with the newest version of React Native. Although I'm not sure if this fix is still necessary for this codebase (with the release of [react-native-tvos](https://github.com/react-native-community/react-native-tvos)).

cpojer:
Yeah let's put it there! Thanks.

PR #25498, opened 2019-07-04
Set minimum width and height for views

M-i-k-e-l:
## Summary
This can help when creating custom ViewGroups (in native modules).
For example: images need to have size in order to be shown, and if the minimum size (width\height) is not set their size is 0.

## Changelog
[Android] [Added] - add `setMinWidth` and `setMinHeight` to `BaseViewManager`.

[CATEGORY] [TYPE] - Message

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -38,6 +38,8 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
   private static final String PROP_TRANSFORM = "transform";
   private static final String PROP_ELEVATION = "elevation";
   private static final String PROP_Z_INDEX = "zIndex";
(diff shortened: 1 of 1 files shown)


osdnk:
@M-i-k-e-l  Shouldn't we drop in from `ViewManager` then?

M-i-k-e-l:
@osdnk What do you mean by "drop in from `ViewManager`"? Do you mean to move the code to the `ViewManager`? Because I think the `BaseViewManager` is more suitable; the `ViewManager` looks more like an interface.

makovkastar:
Hello @M-i-k-e-l, thank you for the contribution. Could you please provide a test plan with an example when this might be useful? As far as I can see we already have [minWidth](https://facebook.github.io/react-native/docs/layout-props.html#minwidth) and [minHeight](https://facebook.github.io/react-native/docs/layout-props.html#minheight) properties supported.

osdnk:
Frankly speaking, I don't understand this PR. Can you add some text / gifs / screenshots / snack?

PR #25549, opened 2019-07-08
[Android][TextInput] Update types for autoComplete prop.

drtangible:
## Summary

I believe there's a mismatch between the type definitions and the expected prop in Android for `TextInput`'s `autoComplete` prop.

* Android is expecting `autoComplete`.
* JS types are expecting `autoCompleteType`.
* Latest documentation documents `autoCompleteType`.

Prop added here: https://github.com/facebook/react-native/commit/179d490607620a988a53aacb01031ed300d4ac66

This change updates the JS types to match what Android is expecting (`autoComplete`). Can update documentation if this is the approach we'd prefer (rather than updating Android to expect `autoCompleteType`).

## Changelog

[Javascript] [Fixed] - Update types for `TextInput`'s `autoComplete` prop.

## Test Plan

Before:

* Pass invalid value to `TextInput`'s `autoComplete` prop, see no type errors on JS side, and Android blows up with:

```sh
Invalid autocomplete option: foobar
updateViewProp
    ViewManagersPropertyCache.java:95
setProperty
    ViewManagerPropertyUpdater.java:132
updateProps
    ViewManagerPropertyUpdater.java:51
updateProperties
    ViewManager.java:37
```

After:

* Pass invalid value to `TextInput`'s `autoComplete` prop, see PropType warning for `autoComplete` prop.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -227,7 +227,7 @@ type IOSProps = $ReadOnly<{|
 |}>;
 
 type AndroidProps = $ReadOnly<{|
-  autoCompleteType?: ?(
+  autoComplete?: ?(
     | 'cc-csc'
     | 'cc-exp'
     | 'cc-exp-month'
@@ -464,7 +464,7 @@ const TextInput = createReactClass({
      *
      * @platform android
      */
-    autoCompleteType: PropTypes.oneOf([
+    autoComplete: PropTypes.oneOf([
       'cc-csc',
       'cc-exp',
       'cc-exp-month',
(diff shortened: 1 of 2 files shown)


rickhanlonii:
Hey @drtangible, thanks for submitting your first RN PR!

This is a great find! This change would be a breaking change for the TextInput component API which we should avoid if possible. There are two ways to do that:

- Update android to change the prop as you mentioned
- Change the prop as it's passed to native

The first option is ideal, but the second is a quick fix. To do the second option you can change TextInput [here](https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L1205) to be something like:

```js
const textContainer = (
  <AndroidTextInput
    {...props}
    autoComplete={props.autoCompleteType}
    // ...
  />
);
```

This will keep the prop we use in JS as `autoCompleteType` but send `autoComplete` to Android. You can choose either route 👍

drtangible:
@rickhanlonii Sounds good! Pushed a change on the Android side, but can try to take a look this weekend about anything else that's required for native-side changes.

rickhanlonii:
@yungsters did you mean to re-open this?

yungsters:
Whoah, what. No.

PR #25554, opened 2019-07-08
Export Modal’s Flow types

swashcap:
## Summary

I'd like to use `Modal`’s flow types in my application to make a reusable component.

## Changelog

[JavaScript] [Added] - Exported `Modal`’s types

## Test Plan

n/a

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -42,13 +42,13 @@ import type {SyntheticEvent} from '../Types/CoreEventTypes';
 // destroyed before the callback is fired.
 let uniqueModalIdentifier = 0;
 
-type ModalOrientationChangeEvent = SyntheticEvent<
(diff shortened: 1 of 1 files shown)


swashcap:
> Do you mind keeping the type names the same as they are now and just adding export type for them?

Yeah, totally! I just noticed that `View`’s props [are exported as `ViewProps`](https://github.com/facebook/react-native/blob/a162554f5dc36fa0647b5bf52119a62bd20046e3/Libraries/Components/View/ViewPropTypes.js#L367), but I'll change it.

swashcap:
Reverted the names in 1d08c25571ff6f727aa0044a4e3e638b37d20d69.

cpojer:
Thank you! I’ll land it tomorrow.

We should probably change the ViewProps file then :D

PR #25559, opened 2019-07-09
Android: Log basic initialization performance numbers.

tido64:
## Summary

I need to measure React Native initialization performance on Android. On iOS, I simply pulled numbers from [RCTPerformanceLogger](https://github.com/facebook/react-native/blob/master/React/Base/RCTPerformanceLogger.h) when needed. The Android equivalent is missing.

## Changelog

[Android] [Added] - Basic initialization performance logging.

## Test Plan

Tests were added.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactNativeHost.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactNativeHost.java
@@ -26,6 +26,7 @@ public abstract class ReactNativeHost {
 
   private final Application mApplication;
   private @Nullable ReactInstanceManager mReactInstanceManager;
+  private @Nullable ReactPerformanceLogger mReactPerformanceLogger;
 
   protected ReactNativeHost(Application application) {
     mApplication = application;
@@ -34,11 +35,21 @@ public abstract class ReactNativeHost {
   /** Get the current {@link ReactInstanceManager} instance, or create one. */
   public ReactInstanceManager getReactInstanceManager() {
     if (mReactInstanceManager == null) {
+      // Ensure {@link ReactPerformanceLogger} exists before we create {@link ReactInstanceManager}.
+      getReactPerformanceLogger();
       ReactMarker.logMarker(ReactMarkerConstants.GET_REACT_INSTANCE_MANAGER_START);
       mReactInstanceManager = createReactInstanceManager();
       ReactMarker.logMarker(ReactMarkerConstants.GET_REACT_INSTANCE_MANAGER_END);
     }
     return mReactInstanceManager;
+  }
+
+  /** Get the current {@link ReactPerformanceLogger} instance, or create one. */
+  public ReactPerformanceLogger getReactPerformanceLogger() {
+    if (mReactPerformanceLogger == null) {
+      mReactPerformanceLogger = new ReactPerformanceLogger();
+    }
+    return mReactPerformanceLogger;
   }
 
   /**
@@ -57,6 +68,10 @@ public abstract class ReactNativeHost {
     if (mReactInstanceManager != null) {
       mReactInstanceManager.destroy();
       mReactInstanceManager = null;
+    }
+    if (mReactPerformanceLogger != null) {
+      mReactPerformanceLogger.release();
+      mReactPerformanceLogger = null;
     }
   }
 
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactPerformanceLogger.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactPerformanceLogger.java
@@ -1,0 +1,93 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * <p>This source code is licensed under the MIT license found in the LICENSE file in the root
+ * directory of this source tree.
+ */
+package com.facebook.react;
+
+import android.util.Log;
+
+import com.facebook.react.bridge.ReactMarker;
+import com.facebook.react.bridge.ReactMarkerConstants;
+import com.facebook.react.common.annotations.VisibleForTesting;
+
+import javax.annotation.Nullable;
+
+import static com.facebook.react.bridge.ReactMarkerConstants.RUN_JS_BUNDLE_END;
+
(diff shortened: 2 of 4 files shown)


hey99xx:
Do we need this as part of RN core? I've found creating a custom `ReactMarker.MarkerListener` implementation sufficient, and I can save all events, times, thread ids etc to JSON to disk and analyze with Chrome tracing as explained in http://blog.nparashuram.com/2018/11/react-native-performance-playbook-part-i.html

If you want to go ahead with what's on iOS, one difference there is `RCTBridge` owns `RCTPerformanceLogger`, so the events are focused. On Android if you have multiple `ReactInstanceManager`s your events will be all over the place as all `ReactPerformanceLogger`s would be receiving events for `ReactInstanceManager` they don't care about. That's where `String tag, int instanceKey` could be handy.

tido64:
I do believe we need it in core. For context, we have high-value components that get integrated into numerous purely native apps. Some of the apps initialize React Native on app startup, when the app is idle, and some on demand. We don't want to make intrusive changes to all of these apps (we're talking ~20 apps we would have to update), I think it makes sense for React Native to log like it is done on iOS.

I'm not entirely sure about your comment about `instanceKey`. Can you please elaborate? When I look at the code, it seems to always pass 0 unless the marker comes from a native module.

```java
  @DoNotStrip
  public static void logMarker(ReactMarkerConstants name) {
    logMarker(name, null, 0);
  }
```

hey99xx:
> We don't want to make intrusive changes to all of these apps

Won't you have to call something like `reactNativeHost.getReactPerformanceLogger().getDuration()` in all those apps? Is that ok?

For markers, it uses `instanceKey`/`tag` for only some of them. For example in your code you use `RUN_JS_BUNDLE_END` to detect script execution end, but that marker is tagged with script name: https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/ReactAndroid/src/main/jni/react/jni/JReactMarker.cpp#L38-L39
https://github.com/facebook/react-native/blob/664646055a3eee2a6ad031294eac4b2ec18a2cf4/ReactCommon/jsiexecutor/jsireact/JSIExecutor.cpp#L132

Maybe it does not matter if you have a single bundle and don't need to match pairs, not sure. I'll leave it up to you and maintainers. This code does not break custom ReactMarker listeners so I don't mind it.

axe-fb:
IMHO, we dont need this in core, especially since this can be extracted in the app itself.

tido64:
@axe-fb The reason I think it makes sense to have it in core is because every app will have to implement this by themselves, and all will do it a bit differently. I think it's especially valueable if you own a component that is integrated in many hybrid apps without any guarantees for how or when React Native initializes. This means you can grab the numbers yourself without asking the apps themselves to make any intrusive changes. We've had issues with this in the past in that people have been complaining about perf issues, and increases in ANR. Without these numbers, we're basically flying blind and can't help them address issues.

That said, I'm open to suggestions on any improvements or alternative solutions you might have.

ReactAndroid/src/main/java/com/facebook/react/ReactPerformanceLogger.java
```diff
@@ -0,0 +1,93 @@
...
+public class ReactPerformanceLogger implements ReactMarker.MarkerListener {
+
+  public enum Tag {
+    BRIDGE_STARTUP,
+    SCRIPT_EXECUTION,
+  }
+
+  private static final String TAG = ReactPerformanceLogger.class.getSimpleName();
+
+  private long[][] mData = new long[Tag.values().length][2];
```
axe-fb: How is mData exposed to the apps ? Whats the API to consume this ?

ReactAndroid/src/main/java/com/facebook/react/ReactNativeHost.java
```diff
@@ -58,6 +69,10 @@ public void clear() {
       mReactInstanceManager.destroy();
       mReactInstanceManager = null;
     }
+    if (mReactPerformanceLogger != null) {
+      mReactPerformanceLogger.release();
```
axe-fb: What happens when app is reloaded in DEV mode ? Will the markers continue to aggregate ?

ReactAndroid/src/main/java/com/facebook/react/ReactNativeHost.java
```diff
@@ -34,13 +35,23 @@ protected ReactNativeHost(Application application) {
...
       mReactInstanceManager = createReactInstanceManager();
       ReactMarker.logMarker(ReactMarkerConstants.GET_REACT_INSTANCE_MANAGER_END);
     }
     return mReactInstanceManager;
   }
 
+  /** Get the current {@link ReactPerformanceLogger} instance, or create one. */
+  public ReactPerformanceLogger getReactPerformanceLogger() {
+    if (mReactPerformanceLogger == null) {
+      mReactPerformanceLogger = new ReactPerformanceLogger();
```
axe-fb: Why can't this be a singleton ? I am not sure if we are using this instance variable anywhere else.

ReactAndroid/src/main/java/com/facebook/react/ReactPerformanceLogger.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  public void release() {
+    ReactMarker.removeListener(this);
+  }
+
+  @Override
+  public void logMarker(ReactMarkerConstants name, @Nullable String tag, int instanceKey) {
+    switch (name) {
+      case GET_REACT_INSTANCE_MANAGER_START:
```
axe-fb: Why are we only marking these ? Why not just pick up all markers that RN gives ?

axe-fb:
I think that this can be a library ? I still don't understand how existing apps could use this data - could you please explain that ?

mdvacca:
@tido64 Thank you for working on this PR!
I agree with @axe-fb that in this case it would be better to not include this PR into the core of RN.

React Native already exposes all the required APIs to measure performance in RN Android and our objective is to keep the Core of the framework as small as possible, exposing clear APIs that should be used or extended by ReactNative apps.
In this case recommend you to create a library that allow you to reuse this code across multiple React Native Android apps.

PR #25569, opened 2019-07-10
Update build.gradle

### LINKED ISSUE #25145: Adding android gradle plugin as an extra property of project template
React Native version:
0.58.8

## Steps To Reproduce
This is an enhancement (or feature request)

having an extra property in `ext` that specify android gradle plugin version, so libraries developer can read and set the main project android gradle plugin version to their library and so whole project build with the same gradle plugin. this is an important and useful property when users using CI server with code repository for building their product. for example, I use a docker with previous installed all necessary resources for building an android app in my CI server, but some of libraries I used have different gradle plugin version, so my docker has to download the different version of gradle plugins to build a final product.

I mean something like this:

I wanted to submit a pull request but I'm not expert in gradle build system and not familiar with this repository and I couldn't find android gradle template file, So I suggest my changes in here

```
buildscript {
    ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 16
        compileSdkVersion = 28
        targetSdkVersion = 28
        supportLibVersion = "28.0.0"
        gradlePluginVersion = "3.4.1" <----- add this property to react native android template
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:$ext.gradlePluginVersion'
    }
}
```

By adding the property other library developers can do something like this in their library's gradle file:

```
def safeExtGet(prop, fallback) {
    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
}

buildscript {
  repositories {
    ...
  }

  dependencies {
    classpath 'com.android.tools.build:gradle:${safeExtGet('gradlePluginVersion', '3.4.0')}'
  }
}
```

I don't familiar with gradle syntax, so it might be wrong in the syntax of what I wrote, I just want to suggest my idea.

Also, it would be great if make a standard mechanism that libraries could read version of gradle wrapper version of main project from `gradle-wrapper.properties` , because I have seen some libraries that have `gradle-wrapper.properties` file in their directory. for example, you can take a look at [react-native-keychain library](https://github.com/oblador/react-native-keychain/tree/master/android/gradle/wrapper). I have no idea right now how can we add this feature, but I think it's worth discussing about it

SaeedZhiany:
fixes #25145

## Summary
Adding android Gradle plugin version and react-native version to `ext`. By that, the library developers can automatically read and use the root project's setting in the library's gradle and so, the whole project has been built with the same android gradle plugin and its dependency libraries rely on the same version of react-native. Also, putting these properties can cause a naming convention among library developers to avoid using an arbitrary name for reading these properties from the root project. (I currently using two libraries that use different names for getting android gradle plugin version just because of the lack of convention)

**Note for library developers:**
change your android `build.gradle` like this:

```
buildscript {
   +ext.safeExtGet = {prop, fallback ->
   +    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
   +}
    repositories {
        jcenter()
        google()
    }

    dependencies {
        -classpath 'com.android.tools.build:gradle:3.4.1'
        +classpath("com.android.tools.build:gradle:${safeExtGet('gradlePluginVersion', '3.4.1')}")
    }
}

android {
    dependencies {
       -implementation 'com.facebook.react:react-native:+'
       +implementation "com.facebook.react:react-native:${safeExtGet('reactnativeVersion', '+')}"
    }
}
```

## Changelog

[Android] [Added] - root project's android gradle plugin version and react-native version is now accessible for use in libraries' `build.gradle` files.


dulmandakh:
Usually it's not recommended to re-import Android Gradle Plugin to submodule, because it's already imported into root project, or RN app. So, I think it's not a prober solution to your problem. I would like to recommend you to change a native module to not use Android Gradle Plugin directly, but from root project.

SaeedZhiany:
@dulmandakh
It seems there is a misunderstanding here.
I don't want to re-import the tool in sub module. actually I want to prevent to import and using multiple Android Gradle plugin version in one project. If you take a look at some RN libraries, you'll see some of them strictly using Android Gradle tool version lower than 3.4.1 (some of them using 2.x!!!) because it is not actively maintained by its maintainer. with multiple Gradle tool version, CI servers have to build project with multiple tools that I think it's not proper way.
(As a workaround I already wrote a script to replace the Gradle tool versions of project dependencies with project's Gradle tool version in my current project).
By this PR we can solve the problem in long term and users don't need to handle such cases.

dulmandakh:
I understand the purpose of this PR, but you still need to make changes to libraries to make use of above changes. Instead, I would like to recommend to fix build.gradle to not use Android Gradle Plugin.

dulmandakh:
About a year ago, I created a PR to RN to export Android SDK tools version configurations to be used in other libraries, and it helped a lot. I strongly believe that sub-projects or modules shouldn't be using Android Gradle Plugin.

SaeedZhiany:
@dulmandakh 
sorry but your statements couldn't convince me, please re-open this PR to having more discussion about it.

dulmandakh:
could you please list of modules you use and having problem?

dulmandakh:
I created a PR to react-native-keychain https://github.com/oblador/react-native-keychain/pull/227

SaeedZhiany:
> could you please list modules you use and having problem?

[react-native-restart](https://github.com/avishayil/react-native-restart)
[react-native-svg](https://github.com/react-native-community/react-native-svg)
[react-native-keychain](https://github.com/oblador/react-native-keychain)
[react-native-community/async-storage](https://github.com/react-native-community/async-storage)
[react-native-firebase](https://github.com/invertase/react-native-firebase)
[react-native-gesture-handler](https://github.com/kmagiera/react-native-gesture-handler)

SaeedZhiany:
Also, [create-react-native-module](https://github.com/brodybits/create-react-native-module) by @brodybits and [react-native-create-library](https://github.com/frostney/react-native-create-library) are library template boilerplates, we can apply your suggestion in there too for fixing the problem for future RN libraries.

cpojer:
What's the status of this PR?

dulmandakh:
IMHO, Android Gradle Plugin should be loaded only once. Please correct me
if wrong.

This change will allow load AGP many times easily, or hide it. So I don't
like the approach. It's best to fix third-party modules instead.

SaeedZhiany:
> This change will allow **load AGP many times easily**

Are you sure about that?!
I don't know how exactly the Gradle system works, but I think it's reasonable to consider AGP loaded once and in next refers, it uses the first cached loaded one.

However, are you have any other idea to solve the problem (prevent sub-modules from using different versions of AGP) your PR on react-native-keychain does not seem to solve this issue. (at least I didn't understand it)

dulmandakh:
IIRC, an issue that you are trying to solve is AGP version mismatch between
a project and its modules, and you want to share it between them with the
change. Therefore, it means AGP can be loaded multiple times.

What I'm trying to say is that your proposed change might seem fix the
issue, but IMHO it's not a proper fix. So it's better be fixed in modules
instead, or remove AGP usage in modules.

SaeedZhiany:
> Therefore, it means AGP can be loaded multiple times.

just for clarification, you said "**can be loaded**" not "**will** be loaded", does it means you're not sure about that? this is all I'm asking. have you a reference (to AGP documentation) that confirms using same AGP version causes the AGP loaded multiple times? because it does not make sense to me why it has not been cached in the first place and used multiple times. maybe it's a problem with Gradle system.

dulmandakh:
I don't want to argue with you. But what you are trying to do is share AGP
version and load multiple copies of it without conflict. I don't need to
prove it, because we all experienced it.

IMHO, a PR creator should make sure that he/she is doing right thing and
not causing issues, but mistakes happen.

SaeedZhiany:
I don't want either. I just wanted to make sure.

However, my purpose is to prevent from loading different AGP version that causes some conflicts, not to prevent from loading multiple AGP, it is another problem I think (that already exists with current gradle configs) that and my change is not the reason for it. I'm not expert in Gradle system and don't know how exactly it works and this is all I could to do to solve the problem, but, with respect, you continue to oppose me without having a better solution.

SaeedZhiany:
I don't want either. I just wanted to make sure.

However, my purpose is to prevent from loading different AGP version that causes some conflicts, not to prevent from loading same AGP multiple times, it is another problem I think (that already exists with current gradle configs) that and my change is not the reason for it. I'm not expert in Gradle system and don't know how exactly it works and this is all I could to do to solve the problem, but, with respect, you continue to oppose me without having a better solution.

dulmandakh:
I propose that modules shouldn’t include AGP in it, so please help them and create PRs to remove it.

SaeedZhiany:
As I said, I don't know how I should do what you proposed. I've read your PR on react-native-keychain and couldn't find any clue. not just me, even @mikehardy has left a [comment](https://github.com/oblador/react-native-keychain/pull/227#issuecomment-521381092) for you there and we wainting for your response.

dulmandakh:
Sorry, I didn’t see the comment. I commented on the PR to keychain.

dulmandakh:
Created a PR to async-storage https://github.com/react-native-community/async-storage/pull/207

dulmandakh:
another PR https://github.com/FormidableLabs/react-native-app-auth/pull/381

mikehardy:
I looked through the linked react-native-keychain issue and reviewed one of the other PRs. I could see this working. I've never made a gradle project using kotlin definitions so I didn't think of it as an option. My only remaining question is if it's possible in the future for react-native to adopt kotlin for gradle, and would that then collide with this new style? I think with the project defined in the root directory of the module (instead of in android/) the answer is no, but I'm checking

dulmandakh:
@mikehardy some RN Gradle scripts are already converted to use Kotlin DSL, which makes editing easy and has better IDE support. You can have and mix gradle scripts no matter which language is used.

friederbluemle:
@SaeedZhiany - Could you please provide a link to an app project that has problems when including a dependency with an outdated Android Gradle plugin? It is my understanding that the `buildscripts` section is not used when a Gradle project is in a _subfolder_ and **included as a module**. That way, the version that the dependency specifies should not matter at all. I could be wrong though. I'm sure you did plenty of tests, and I want to see it for myself.

SaeedZhiany:
@friederbluemle 
I've faced the problem in a real project in the company I'm working, so, unfortunately, I can't share the project. but I can explain what was wrong in our project.

we using CI for our project and every time a workflow has been started, the CI server downloads multiple AGP versions, so the first problem is it downloads lots of stuff and causes taking more time to finish the builing process (even we don't consider that it wastes our internet bandwidth), the second and main problem was that it seems the CI server built each module with the module's AGP and sometimes it causes the build failed unexpectedly, the behavior is nondeterministic in our project and it builds successfully in next try. so I had to write a script that reads the main project AGP version and re-write all sub-modules used in the project. 

Recently I've thought maybe we can automate this procedure and this is the motivate I raise this issue #25145 and this PR and [this discussion](https://github.com/react-native-community/discussions-and-proposals/issues/151).

Although I didn't figure out @dulmandakh's approach yet, but it seems his solution is better than mine and it's enough for me (I just want the problem solved), so if you and @cpojer prefer his solution, we can close this PR.

SaeedZhiany:
@friederbluemle 
I've faced the problem in a real project in the company I'm working, so, unfortunately, I can't share the project. but I can explain what was wrong in our project.

we using CI for our project. when a workflow starts, the CI server downloads multiple AGP versions, so the first problem is it downloads lots of stuff and causes taking more time to finish the building process (even we don't consider that it wastes our internet bandwidth), the second and main problem was that it seems the CI server built each module with the module's AGP and sometimes it causes the build failed unexpectedly, the behavior is nondeterministic in our project and it may builds successfully in next try. so I had to write a script that reads the main project AGP version and re-write all sub-modules used in the project. 

Recently I've thought maybe we can automate this procedure and this is the motivate I raise this issue #25145 and this PR and [this discussion](https://github.com/react-native-community/discussions-and-proposals/issues/151).

Although I didn't figure out @dulmandakh's approach yet, it seems his solution is better than mine and it's enough for me (I just want the problem solved), so if you and @cpojer prefer his solution, we can close this PR.

Just it may worth If someone that has experiences in Gradle system (and also free time) writes a blog in [react-native blogs](https://facebook.github.io/react-native/blog/) to motive maintainers to change their library using @dulmandakh's solution and show them how they should do it. I think @dulmandakh is the best one who can do it. :)

friederbluemle:
Actually, I want it fixed as much as you do :)

From my experience, there is not _that_ much wrong with the prevailing current setup of most projects (although there is certainly a LOT to simplify and consolidate). I definitely agree with you that it would be great to have some kind of "best practices" guide for RN module creation.

What would be a super strong point for your approach is if you could create an open source sample project that showcases the issues you mention in reproducible way.

SaeedZhiany:
Additional to mentioned issues:

according to @dulmandakh's [comment](https://github.com/facebook/react-native/pull/25569#issuecomment-531430488)

> Therefore, it means AGP can be loaded multiple times.

If it means that there may be a situation where multiple AGPs loaded at the same time into RAM, so it means CI server waste more resources for building too. then his approach has another advantage to improve building process performance.

SaeedZhiany:
Additional to mentioned issues:

according to @dulmandakh's [comment](https://github.com/facebook/react-native/pull/25569#issuecomment-531430488)

> Therefore, it means AGP can be loaded multiple times.

If it means that there may be a situation where multiple AGPs loaded at the same time into RAM, so it means CI server waste more resources for building too (which it should not). then his approach has another advantage to improve building process performance.

SaeedZhiany:
> What would be a super strong point for your approach is if you could create an open source sample project that showcases the issues you mention in reproducible way.

As I mentioned before, it's happened nondeterministically, so I'm not sure how can I reproduce it, and because it needs to have CI server too it will take more time from me to make such repository and I haven't much free time these days to work on it, sorry about that. but from all discussion we have in few issues and PRs, I concluded, it seems it is a bad practice that almost all module maintainers put gradle files into `android` directory from the beginning. even the nondeterminism is only my project's problem and any other one hadn't such problem before, but we have some other good reasons why it's necessary to removing gradle configs from submodules' android directory.

mikehardy:
For me, simply having multiple gradle distributions download is a major pain point. If you haven't experienced what it's like to be a multi-platform developer in a developing country with low bandwidth and small hard drives, much less run a crew with independent copies of dev environments, it's incredibly painful, and the gradle distributions weigh in at 100+MB. For me, that is at the level of "bug" on it's own.

I believe the proposed solution from @dulmandakh solves that neatly by making sure your project only ever downloads one version for the project, while still having one specified (in Kotlin) for maintenance on the module itself.

The issue of multiple AGPs loading at the same time seems legit but I've not seen problems, I'm interested in single-point version control of the gradle + plugin's in use, and ability to have a single copy at a time.

SaeedZhiany:
> For me, simply having multiple gradle distributions download is a major pain point. If you haven't experienced what it's like to be a multi-platform developer in a developing country with low bandwidth and small hard drives, much less run a crew with independent copies of dev environments, it's incredibly painful, and the gradle distributions weigh in at 100+MB. For me, that is at the level of "bug" on it's own.

Yeah, this is exactly what I'm said and I'm experiencing it every day :D

> The issue of multiple AGPs loading at the same time seems legit but I've not seen problems

it is a problem too I think, just like small hard drives it may users have limited resources (including RAM) for local CI servers, avoid loading multiple AGP can reduce RAM usage, correct me if I'm wrong.

friederbluemle:
@mikehardy @SaeedZhiany - I fully agree with you, the download of multiple Android Gradle plugin distributions is a major pain point, even with a slightly faster internet connection.

I actually found an alternate solution, without moving any Gradle related files into the root of a project.

Essentially, it's just two lines, wrapping the Android Gradle plugin dependency in the `buildscripts` section:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

With this, the dependency is only evaluated (and possibly downloaded) when the project is opened **stand-alone** (opening the `android` folder in Android Studio).

When a library project is included as a dependency, then `rootProject` will be the main application project. -> the statement will evaluate to false, skipping the potentially outdated Android Gradle plugin download.

I opened a bunch of PRs:

https://github.com/react-native-community/react-native-localize/pull/69
https://github.com/react-native-community/react-native-share/pull/588
https://github.com/react-native-community/async-storage/pull/209
https://github.com/FormidableLabs/react-native-app-auth/pull/384

It would be great if you could test this and let me know if it works/what you think. Thanks you 🙏

SaeedZhiany:
@friederbluemle 
That's so interesting and much easier to understand for me. 
just one question:
in @dulmandakh's approach the file `gradle-wrapper.properties` moved to root directory, so it will not be downloaded in the main project too (in case of versions mismatch between the main project's `gradle-wrapper.properties` and the module's `gradle-wrapper.properties`) and this is so good.

does your approach consider it too?

friederbluemle:
Hi @SaeedZhiany - Good question! As the name implies, the file `gradle-wrapper.properties` is part of the Gradle wrapper (together with the three other files `gradle-wrapper.jar`, `gradlew`, and `gradlew.bat`). For an application, the **only wrapper** that will be used is the one that is distributed with the application project itself (in the `android` folder). Any other Gradle wrappers that might be contained in `node_modules` will **not** be used. In fact, these files should _not_ be distributed by library projects in the first place, since it would only result in an increased artifact file size.
By not moving anything into the root directory, it will keep everything related to Android contained in the `android` folder, and everything related to iOS in the `ios` folder (just like in a regular RN application project).
What is distributed in the end is controlled by other means. Have a look at the [files section of package.json](https://docs.npmjs.com/files/package.json#files), and [npmignore](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package).

SaeedZhiany:
> In fact, these files should not be distributed by library projects in the first place, since it would only result in an increased artifact file size.

Yeah, you're right. the only RN module that has such problem is react-native-keychain, the gradle folder has been distributed within the android folder. the other dependencies I'm using in my project has no such problem. Although with @dulmandakh's PR the gradle file currently moved to root directory. if your approach comes to as an agreement, maybe we need to revert the PR's change and apply your approach and then simply add the gradle directory path of react-native-keychain in .gitignore file.

image

I like your suggestion because it's simple and can easily apply. even we can write a simple script that automatically wraps the `repositories` and  `dependencies` with the `if (project == rootProject)` statement.

@mikehardy and @dulmandakh what do you think?

friederbluemle:
There are definitely many other library projects that incorrectly distribute the Gradle wrapper (an others) files. Even with the updated `react-native-keychain` project, the wrapper is _still_ distributed - It's just outside the android folder. In either case, it should not result in any conflicts (just increased download size).
Regarding a quick script that locally fixes up build.gradle files from library projects that have not been updated yet: Great idea! It does not even have to be as complex as wrapping something. Simply removing the whole `buildscripts {}` block should be fine too. As explained, it's only needed when building/opening the project stand-alone, and can safely be removed from dependencies inside of `node_modules` (to avoid the extra downloads).

SaeedZhiany:
> Simply removing the whole `buildscripts {}` block should be fine too.

Yeah, it would be great too. even we can add it to CLI Autolink procedure that simply remove the block in `postinstall` procedure, so it's not necessary to module consumers deal with that.

friederbluemle:
@SaeedZhiany - Quick update to npmignore and files in package.json over at `react-native-keychain`: https://github.com/oblador/react-native-keychain/pull/248

mikehardy:
The `(if project == rootProject)` idea seems much more "android-native" to me, and conceptually simpler, while apparently meeting the same goals. I like it

mikehardy:
Side note: always fun hammering out a solution as a group. The first few idea iterations of jetifier were similarly nowhere near the final, sometimes it just takes a few people a few attempts before you hit on the right thing. Cool

friederbluemle:
Yes, thanks a lot for the input and suggestions everyone (especially @SaeedZhiany @dulmandakh @mikehardy) 👍

SaeedZhiany:
I'm closing this PR in favor of @friederbluemle [solution.](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277) 
Thank you all you guys for contributing.

PR #25582, opened 2019-07-10
Added button with accessibility action example and support for Touchables

AmazonKevin:
## Summary
We added a test to make sure button and accessibility actions would not have unwanted behavior. Additionally we added support for accessibility actions for all touchables. However we discovered that RCTTextView and possibly anything else that does not derive from RCTView does not support accessibility actions and need to be children off a component that does support it for it to not crash in ios. This became noticeable when TouchableWithoutFeedback only worked if text is a child of view on ios.

In a local branch we where able to modify RCTTextView to support accessibility actions and text no longer needed to be a child of view for it to work.

## Changelog

[General] [Added] - Button test with accessibility actions
[General] [Added] - Support for accessibility actions to all Touchables. With TouchableWithoutFeedback being a special case where text must be a child of view. (See AccessibilityExample.js for an example)

## Test Plan

## Open Question
What would you say is the best practice for adding accessibility action support for all the components that do not extended from RCTView?


osdnk:
Thanks!

PR #25583, opened 2019-07-10
deprecate iOS .xcodeproj

fkgozali:
Summary: We now use CocoaPods for better maintainability.

Differential Revision: D16193719


fkgozali:
@thymikee - do you mind double checking if this change is compatible with the RN CLI? I updated the reference to the RNTester.xcodeproj to use RNTesterPods.xcworkspace instead

react-native.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   reactNativePath: '.',
   project: {
     ios: {
-      project: './RNTester/RNTester.xcodeproj',
+      project: './RNTester/RNTesterPods.xcworkspace',
```
fkgozali: cc @thymikee

thymikee: The config resolution works fine, so this is good to go. You can run from cli using:
```
yarn react-native run-ios --project-path RNTester --scheme RNTester
```
However I can see that the `warnAboutManuallyLinkedLibs` fn we call internally to discover and warn about manually linked projects takes very long in this repo, we might want to introduce a temporary flag to disable it.

Btw, how about using the name `RNTester.xcworkspace` instead of `RNTesterPods` now that it's the only one there? If changed, we could run it with just 

```
yarn react-native run-ios --project-path RNTester
```

fkgozali: sounds good!

> Btw, how about using the name RNTester.xcworkspace instead of RNTesterPods now that it's the only one there? If changed, we could run it with just

We could, but let's keep the name as is to avoid confusion, at least until many ppl move over to this new setup.

PR #25611, opened 2019-07-12
Add jetifier to the template to ease migration to Android X

satya164:
Currently the ecosystem is going through a lot of churn due to the Android X upgrade as lots of third party modules are not updated to use Android X yet. Adding `jetifier` to the default template makes this automatic during the migration period.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -4,7 +4,8 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
-    "test": "jest"
+    "test": "jest",
+    "postinstall": "jetify"
   },
   "dependencies": {
     "react": "16.8.1",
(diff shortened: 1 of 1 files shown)


template/package.json
```diff
@@ -4,7 +4,10 @@
   "private": true,
   "scripts": {
     "start": "react-native start",
-    "test": "jest"
+    "run-android": "react-native run-android",
+    "run-ios": "react-native run-ios",
+    "test": "jest",
+    "postinstall": "jetify"
```
thymikee: why not:
```suggestion
    "postinstall": "npx jetifier"
```
so it's easier to remove once everything is migrated?

satya164: @thymikee is `npx` needed? `npm` will run binaries under node_modules fine

thymikee: `npx` would avoid adding `jetifier` as a dependency

satya164: @thymikee yea but we already add it as a dep, and when the user removes it from dep, they can remove the postinstall script too. imo it doesn't change anything

thymikee: It's just my preference, but yea, may feel weird for some.

PR #25623, opened 2019-07-12
update bug_report.md template

ecreeth:
## Summary
It would be of great help a **screenshot** of the error.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -29,7 +29,7 @@ React Native version:
 Describe what you expected to happen:
 
 
-Snack, code example, or link to a repository:
+[Snack](https://snack.expo.io/), screenshot, code example, or link to a repository:
 
 
 <!--
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -29,7 +29,7 @@ React Native version:
 Describe what you expected to happen:
 
 
-Snack, screenshot, code example, or link to a repository:
+Snack, code example, screenshot, or link to a repository:
 
 <!--
  If you think that a screenshot could be useful, you should verify that the errors are of a visual nature, that is, that they give us a clue as to where or why this problem has happened.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,12 +32,9 @@ Describe what you expected to happen:
 Snack, code example, screenshot, or link to a repository:
 
 <!--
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -29,7 +29,7 @@ React Native version:
 Describe what you expected to happen:
 
 
-Snack, code example, or link to a repository:
+[Snack](https://snack.expo.io/), screenshot, code example, or link to a repository:
```
hramos: This should be covered by the comment below (Line 36). I did not use an inline link intentionally, as the bug report template is commonly seen in plain text within the "new issue" text area, and I want to make sure this one-liner is easy to read. Any additional details can go in the commented block of text that follows.

hramos: I am ok with adding "screenshot" to the list here, if it comes with a clarification in the comment block below that states the screenshot should be used for bugs of a visual nature. The situation I would like to avoid is getting bug reports with screenshots of errors / log spew.

ecreeth: Thank you very much, I did not realize that I had already been added in a comment. LGTC!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -29,8 +29,11 @@ React Native version:
 Describe what you expected to happen:
 
 
-[Snack](https://snack.expo.io/), screenshot, code example, or link to a repository:
+Snack, screenshot, code example, or link to a repository:
 
+<!--
```
hramos: Less is better when it comes to issue templates. How about we expand the existing comment block below instead of adding a new one? For example:

```
<!--
  Please provide a Snack (https://snack.expo.io/), a link to a repository on GitHub, or
  provide a minimal code example that reproduces the problem.
  You may provide a screenshot of the application if you think it is relevant to your bug report. 
  Here are some tips for providing a minimal example: https://stackoverflow.com/help/mcve.	  --	
-->
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -29,8 +29,11 @@ React Native version:
 Describe what you expected to happen:
 
 
-Snack, code example, or link to a repository:
+Snack, screenshot, code example, or link to a repository:
```
hramos: Place code example first, to encourage actual code to be included before screenshots.

```suggestion
Snack, code example, screenshot, or link to a repository:
```

.github/ISSUE_TEMPLATE/bug_report.md
ecreeth: Thank you very much for helping me with this!

PR #25636, opened 2019-07-14
[iOS] Remove old GIF code

zhongwuzw:
## Summary

[After we migrated to new GIF implementation](https://github.com/facebook/react-native/pull/24822), we can remove old implementation now.

## Changelog

[iOS] [Deprecated] - Remove old GIF code

## Test Plan

GIF still works.


osdnk:
Let's give as try!

PR #25639, opened 2019-07-14
Fix Android CI

mdvacca:
Summary: This diff adds support for is_androidx param into the buck rule rn_java_library

Differential Revision: D16247319


mdvacca:
Closing PR as I confirmed that this PR fixes the CI. This is being reviewed internally at FB

PR #25363, opened 2019-06-23
Fix regression of improper assets copy (revert #24518 #24778)

robertying:
## Summary

Pull requests #24518 #24778 make Gradle copy all **generated** assets and resources into `android/app/src/res`, which is a bad behavior, because `src/res` goes into version control and should hold only those **original** resource files.

These changes in #24518 #24778 were merged into 0.60.0-rc release and cause regression.

This pull request will:

- Revert pull requests #24518 #24778
- Close #25325 

## Changelog

[Android] [Fixed] - Fix regression of improper assets copy (revert #24518 #24778)

## Test Plan

It is a revert pull request and the reverted script should work the same as it has in 0.59.x.

PR #25642, opened 2019-07-14
Duplicate drawable resources issue

mjm918:
Duplicate drawable resources issue when trying to assembleRelease


cpojer:
Please see https://github.com/facebook/react-native/pull/25363

PR #25719, opened 2019-07-18
Implement URL standard and URLSearchParams spec compliant

### LINKED ISSUE #16434: Creating an instance of URL like: `new URL('http://facebook.com')` throws an exception.
### Is this a bug report?

Yes

Yes

### Environment

  Environment:
  OS:  macOS Sierra 10.12.6
  Node:  8.6.0
  Yarn:  0.24.6
  npm:  5.3.0
  Watchman:  4.7.0
  Xcode:  Xcode 8.3 Build version 8E162

  Target Platform: iOS (10.3)

### Steps to Reproduce

1. Create a ReactNative project, which uses React Native 16
2. Try to create an instance of `URL` in any .js file, like this:
    ```javascript
    const url = new URL('http://facebook.com');
    ```
3. Run the project

### Expected Behavior

The URL to be the native URL, [this one](https://developer.mozilla.org/en-US/docs/Web/API/URL). **It was working fine in React 15, but in React 16 it was replaced with something [internal](https://github.com/facebook/react-native/blob/91372e8f9aafcad2b5594175b1a3ce6f9c8f3b8e/Libraries/Blob/URL.js).** We use URL to parse URLs and extract information about the hostname, port, etc. so we need this behaviour back.

### Actual Behavior
An exception is thrown: `Creating BlobURL objects is not supported yet.`

screen shot 2017-10-17 at 18 15 19

### Reproducible Demo

https://snack.expo.io/SJUA5i7TZ

### LINKED ISSUE #24428: URL incorrectly adds trailing slash
## 🐛 Bug Report

`URL` polyfill in `React Native` 0.59 is broken because it's not compliant to WHATWG URL Standard. And it's broken even for the simplest use cases.

## To Reproduce

```js
const url = new URL('https://facebook.github.io/react-native/img/header_logo.png');
console.log(url);
```

## Expected Behavior

```
https://facebook.github.io/react-native/img/header_logo.png
```

## Actual Behavior

```
https://facebook.github.io/react-native/img/header_logo.png/
```

Note the trailing slash causes the URL points to a 404 page.

## Environment
```
    System:
      OS: macOS 10.14.4
      CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
      Memory: 823.64 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.0
      Yarn: 1.15.2
      npm: 6.5.0
      Watchman: 4.9.0
    SDKs:
      iOS SDK:
      Android SDK:
        API Levels: 28
        Build Tools: 28.0.3
        System Images: android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom
    IDEs:
      Xcode: 10.2/10E125 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.4 => 0.59.4 
```

## Possible Cause

https://github.com/facebook/react-native/blob/65033d7078654bae2a87e4791dbb725110ae6bad/Libraries/Blob/URL.js#L146-L149

This code was introduced in https://github.com/facebook/react-native/pull/22901 .

### LINKED ISSUE #25717: URL implementation should add a trailing slash to the base
React Native version:

```
System:
    OS: macOS 10.14.5
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 1002.17 MB / 32.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  SDKs:
    iOS SDK:
    Android SDK:
      Build Tools: 27.0.3, 28.0.2, 28.0.3, 29.0.0
      System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom
      Android NDK: 17.2.4988734
  IDEs:
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.3 => 0.60.3
```

The current custom implementation of [URL Standard](https://url.spec.whatwg.org/) incorrectly handles conjunctions between URL and base.

https://github.com/facebook/react-native/blob/45a4bec60cd330b9cecf5727c5ccf4203359c86e/Libraries/Blob/URL.js#L131-L155

## Steps To Reproduce

```js
const url = new URL('about', 'https://www.mozilla.org');
console.log(url.href);
```
## Expected Behavior

```
https://www.mozilla.org/about
```

## Actual Behavior

```
https://www.mozilla.orgabout
```

charpeni:
Fixes #25717 
Fixes #24428
Fixes #16434

## Summary

Currently, our implementation of [`URL`](https://github.com/facebook/react-native/blob/master/Libraries/Blob/URL.js) is not spec-compliant and this is causing headaches to RN users.

Tests were added to `URL`, to be sure we're spec compliant with https://url.spec.whatwg.org/.

As a solution to those issues, we could do one of the following:

- Implement [`whatwg-url`](https://github.com/jsdom/whatwg-url) instead of our custom implementation. ⚠️ This would add [significant weight](https://bundlephobia.com/result?p=whatwg-url@7.0.0) to React Native.
- Fork `whatwg-url` and remove the support of unicode from there, because `tr46` dependency is like ~80% of `whatwg-url` size.
- Implement a lightweight polyfill. Those available on npm are not really trustworthy or depend heavily on the DOM.
- Fix RN's URL issues, case by case. IMO, this is not a viable solution. 🤷‍♂ 

I'll tackle an implementation of `whatwg-url` and compare what will be the increase of RN size with and without unicode support.

Let me know if you have any thoughts, concerns or ideas.

___

After looking at this, our best approach is to use `whatwg-url` to cover all the spec. But we can't use `whatwg-url` directly, because it's coming with a dependency to `tr46` to handle Unicode cases. I think we can live without Unicode, so I forked `whatwg-url`, and I removed the unicode support, edited tests, and published this as `whatwg-url-without-unicode`.

Here are some stats:

📦`whatwg-url` is taking 352.5 KB.
📦`whatwg-url-without-unicode` is taking 53.5 KB.

Then, `whatwg-url-without-unicode` requires `Buffer`, which is 142 B.

As `URLSearchParams` is bundled in `whatwg-url`, I replaced our custom implementation with the spec-compliant one and some tests to cover this.

## Changelog

[General] [Changed] - Implement spec compliant URL and URLSearchParams

## Test Plan

Added tests, ran RNTester, and called `new URL()`.

### DIFF
--- a/Libraries/Blob/URL.js
+++ b/Libraries/Blob/URL.js
@@ -9,6 +9,8 @@
 
 'use strict';
 
+const {URL: whatwgUrl} = require('whatwg-url-without-unicode');
+
 const Blob = require('./Blob');
 
 import NativeBlobModule from './NativeBlobModule';
@@ -105,115 +107,17 @@ export class URLSearchParams {
   }
 }
 
-function validateBaseUrl(url: string) {
-  // from this MIT-licensed gist: https://gist.github.com/dperini/729294
-  return /^(?:(?:(?:https?|ftp):)?\/\/)(?:(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))(?::\d{2,5})?(?:[/?#]\S*)?$/i.test(
-    url,
-  );
-}
-
-export class URL {
-  _searchParamsInstance = null;
-
-  static createObjectURL(blob: Blob) {
-    if (BLOB_URL_PREFIX === null) {
-      throw new Error('Cannot create URL for blob!');
-    }
-    return `${BLOB_URL_PREFIX}${blob.data.blobId}?offset=${
-      blob.data.offset
-    }&size=${blob.size}`;
+whatwgUrl.createObjectURL = function createObjectURL(blob: Blob) {
+  if (BLOB_URL_PREFIX === null) {
+    throw new Error('Cannot create URL for blob!');
   }
+  return `${BLOB_URL_PREFIX}${blob.data.blobId}?offset=${
+    blob.data.offset
+  }&size=${blob.size}`;
+};
 
-  static revokeObjectURL(url: string) {
-    // Do nothing.
-  }
+whatwgUrl.revokeObjectURL = function revokeObjectURL(url: string) {
+  // Do nothing.
+};
 
-  constructor(url: string, base: string) {
-    let baseUrl = null;
-    if (base) {
-      if (typeof base === 'string') {
-        baseUrl = base;
-        if (!validateBaseUrl(baseUrl)) {
-          throw new TypeError(`Invalid base URL: ${baseUrl}`);
-        }
-      } else if (typeof base === 'object') {
-        baseUrl = base.toString();
-      }
-      if (baseUrl.endsWith('/') && url.startsWith('/')) {
-        baseUrl = baseUrl.slice(0, baseUrl.length - 1);
-      }
-      if (baseUrl.endsWith(url)) {
-        url = '';
-      }
-      this._url = `${baseUrl}${url}`;
-    } else {
-      this._url = url;
-      if (!this._url.endsWith('/')) {
-        this._url += '/';
-      }
-    }
-  }
-
-  get hash() {
-    throw new Error('not implemented');
-  }
-
-  get host() {
-    throw new Error('not implemented');
-  }
-
-  get hostname() {
-    throw new Error('not implemented');
-  }
-
-  get href(): string {
-    return this.toString();
-  }
-
-  get origin() {
-    throw new Error('not implemented');
-  }
-
-  get password() {
-    throw new Error('not implemented');
-  }
-
-  get pathname() {
-    throw new Error('not implemented');
-  }
-
-  get port() {
-    throw new Error('not implemented');
-  }
-
-  get protocol() {
-    throw new Error('not implemented');
-  }
-
-  get search() {
-    throw new Error('not implemented');
-  }
-
-  get searchParams(): URLSearchParams {
-    if (this._searchParamsInstance == null) {
-      this._searchParamsInstance = new URLSearchParams();
-    }
-    return this._searchParamsInstance;
-  }
-
-  toJSON(): string {
-    return this.toString();
-  }
-
-  toString(): string {
-    if (this._searchParamsInstance === null) {
-      return this._url;
-    }
-    const separator = this._url.indexOf('?') > -1 ? '&' : '?';
-    return this._url + separator + this._searchParamsInstance.toString();
-  }
-
-  get username() {
-    throw new Error('not implemented');
-  }
-}
+export const URL = whatwgUrl;
--- a/Libraries/Blob/URL.js
+++ b/Libraries/Blob/URL.js
@@ -53,60 +53,6 @@ if (
  * ```
  */
 
-// Small subset from whatwg-url: https://github.com/jsdom/whatwg-url/tree/master/lib
-// The reference code bloat comes from Unicode issues with URLs, so those won't work here.
-export class URLSearchParams {
-  _searchParams = [];
-
-  constructor(params: any) {
-    if (typeof params === 'object') {
-      Object.keys(params).forEach(key => this.append(key, params[key]));
-    }
-  }
-
-  append(key: string, value: string) {
-    this._searchParams.push([key, value]);
-  }
-
-  delete(name) {
-    throw new Error('not implemented');
-  }
-
-  get(name) {
-    throw new Error('not implemented');
-  }
-
-  getAll(name) {
-    throw new Error('not implemented');
-  }
-
-  has(name) {
-    throw new Error('not implemented');
-  }
-
-  set(name, value) {
-    throw new Error('not implemented');
-  }
-
-  sort() {
-    throw new Error('not implemented');
-  }
-
-  [Symbol.iterator]() {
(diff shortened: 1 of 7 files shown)


cpojer:
@charpeni did you make any progress on this?

charpeni:
@cpojer Yes! Jungling with other priorities at the moment, but I should be able to push changes soon.

charpeni:
After looking at this, our best approach is to use `whatwg-url` to cover all the spec. But we can't use `whatwg-url` directly, because it's coming with a dependency to `tr46` to handle Unicode cases. I think we can live without Unicode, so I forked `whatwg-url`, and I removed the unicode support, edited tests, and published this as `whatwg-url-without-unicode`.

Here are some stats:

📦`whatwg-url` is tacking 352.5 KB.
📦`whatwg-url-without-unicode` is tacking 53.5 KB.

Then, `whatwg-url-without-unicode` requires `Buffer`, which is 142 B.

As `URLSearchParams` is bundled in `whatwg-url`, I replaced our custom implementation with the spec-compliant one and some tests to cover this.

This PR is ready to review!

cpojer:
Thanks for doing this exploration. Adding 60k to bundle size for something that not everyone needs seems excessive. At Facebook we are trying to reduce the size of our apps and it will be hard to justify adding this much code at this time.

What if we made the URL implementation injectable? That way people could modify the URL class that is used, like `require('react-native').setURLImplementation(…)`. What do you think?

charpeni:
Thanks, @cpojer. I would have loved getting that concern during the draft PR 😅.

I totally agree that adding something to the bundle size for something that not everyone needs is excessive.

Adding the initial 352 KB **is excessive**, but adding the forked version of 53 KB doesn't seem that excessive considering the fact that `URL` may not be as trivial as we think. We can't really opt-in or opt-out of `URL`, because it's part of the core of JavaScript, at least in the Web Standard and Node spec, right? Sure, we can limit our usage of `URL` within our app (let's say from a require), but can we control what's going on in third-party dependencies?

I'd be afraid that React Native consumers will still have weird behavior without acknowledging that it could come from our `URL` implementation. I'd prefer to not provide an `URL` implementation by default at all than providing a minimal one that doesn't respect the spec.

⚠️ Note that RN requires an `URL` implementation that implement `static createObjectURL(blob)` to support blob.

If we keep our current implementation with a possibility to inject one at the choice of users, it would be the same frustration as now:

1. Try to use `new URL()`
1. Something is unexpected in the behavior
1. Doubt on ourself
1. Doubt on React Native
1. Be sad about React Native
1. [Implement it ourself in our app](https://github.com/facebook/react-native/issues/16434#issuecomment-392147974)
1. Unexpected behaviors with blob?

I'd be more prompt to fix our implementation even regarding the additional 53 KB to the bundle size than shipping something that is not polished/usable to React Native consumers. Or, do a breaking change by removing it. The initial 352 KB was excessive, but 53 KB to solve those issues seems to worth a discussion.

If we want to focus on reducing the bundle size, I think that focusing our efforts on tree shaking within Metro would be beneficial for every dependency that not everyone needs. Like by removing `whatwg-fetch` and `abort-controller` for people that are not using `fetch`, or even `URL`, or any other dependencies.

Otherwise, making them all injectable could be helpful? Not sure. 🤷‍♂ 

What do y'all think? Would love to hear other feedback on the bundle size in general and `URL` usage.

charpeni:
My previous numbers were the minified size from [bundlephobia](https://bundlephobia.com).

I did some tests while playing with 📦 RNTester's bundle, here are my observations:

- `fetch` is adding 10 KB
- Legacy implementation of `URL` and `URLSearchParams` are adding 4 KB
- `Blob` and `File` are adding 3 KB
- Websocket is adding 4 KB
- XMLHTTPRequest is adding 21 KB
- `AbortController` is adding 2 KB
- `whatwg-url` would add 455 KB
- `whatwg-url-without-unicode` is adding 63 KB (_NEW_)
- `Buffer` is adding 23 KB (_NEW_)

cpojer:
@charpeni sorry for not raising it earlier. I didn't realize that after my comment you would immediately get back to iterate on this and I wasn't sure which solution you were picking. That's on me.

I am pointing out that adding 50k of compressed app size to fix hypothetical use cases is a very very hard sell at Facebook right now. The reason I suggested the approach with injection is because we could document it and only the people who need this will have to deal with a bigger bundle size. In that case we can recommend people to use the official whatwg-url solution as well.

charpeni:
@cpojer if we make it injectable, what do you think about doing a breaking change by completely removing `URL` and `URLSearchParams`? 🤔

cpojer:
Interesting question, I think that could work. I unfortunately don't have a good idea on how much URL is used within RN itself and how many users depend on it. How can we assess the impact of such a breaking change?

leethree:
@cpojer I'm sorry but I cannot agree with your assessment that proper URL handling is "hypothetical use case" when you cannot even handle the most basic use case like `new URL('http://facebook.com')` (see #16434) 

The React Native team should either make `Blob` compatible with the standard `URL`, or rename the broken `URL` to something else.

charpeni:
What we all want is a way to have polyfill for missing classes, right?

If we split the apple in half, we could say that we all want to configure the polyfills we actually need without adding unnecessary code to the bundle. But, we won't have to deal with installing and applying polyfills ourself.

What do y'all think if we expose something like this from React Native:

```js
configurePolyfills({
  URL: true,
  URLSearchParams: true,
  fetch: false,
});
```

Then, the corresponding polyfill will be added to global or not, based on this or on the RN's default value.

React Native will provide expected polyfills within its dependencies`, but may not include them by default, which won't affect the bundle size, right?

https://github.com/facebook/react-native/blob/a5ea7200da7740dda0d266fa08fba8d4def7072a/Libraries/Core/setUpXHR.js#L31-L32

___

> How can we assess the impact of such a breaking change?

We could let the legacy implementation of `URL` and `URLSearchParams` for the next release, but show a warning message saying that implementation is deprecated, thus consumers while have to enable the polyfill with `configurePolyfills` or move away of `URL` and `URLSearchParams`.

charpeni:
Closing in favor of https://github.com/charpeni/react-native-url-polyfill.

PR #25720, opened 2019-07-18
Remove jsc dep from blob native library

janicduplessis:
## Summary

The blob native lib should not include JSC, otherwise it causes a crash when using Hermes.

This actually doesn't depend on the Hermes PR so we can land it now.

## Changelog

[Android] [Fixed] - Remove jsc dep from blob native library

## Test Plan

Tested that is no crash with hermes enabled and disabled

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/blob/jni/Android.mk
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/blob/jni/Android.mk
@@ -15,7 +15,7 @@ LOCAL_C_INCLUDES := $(LOCAL_PATH)
 
 LOCAL_CFLAGS += -fvisibility=hidden -fexceptions -frtti
 
-LOCAL_STATIC_LIBRARIES := libjsi libjsireact jscruntime
+LOCAL_STATIC_LIBRARIES := libjsi libjsireact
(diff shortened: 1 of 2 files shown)


mdvacca:
@janicduplessis The merge failed internally because BlobCollector.cpp depends on JSI:

java/com/facebook/react/modules/blob/jni/BlobCollector.cpp:22: error: undefined reference to 'facebook::jsi::HostObject::~HostObject()'

janicduplessis:
@mdvacca This is when building with BUCK right? Weird that it doesn't happen on oss CI. I tried updating the dep to jsi instead of jsiexecutor. Works when building locally with buck.

mdvacca:
Yes, this is when building with BUCK, I will take a deeper look later today and I will let you know if we need any change in the PR

cpojer:
@osdnk I think the re-import failed. Could you import it again, try to fix up the BUCK file in phabricator and then land this PR? We need it before cutting 0.61.

gengjiawen:
@janicduplessis Need to resolve conflict.

janicduplessis:
Looks like the blob collector pr was reverted yesterday so we’ll have to figure that out first.

janicduplessis:
Going to work on re-landing the blob pr with this fix included instead.

PR #25723, opened 2019-07-19
add support for stringifying error object messages to stringifySafe

danReynolds:
## Summary

Error objects logged as part of the arguments to `console.error` such as from [rejected es6 promises](https://github.com/zloirock/core-js/blob/v2/modules/es6.promise.js#L110) contain the error that the user would want to see as the error object's message, but is not captured by `stringifySafe`. Here we modify it to if the logged value is an error object print the error similar to chrome:

```
const error = new Error('error');
stringifySafe(error); // Error: error
```

Versus the current behavior which does not recognize the error type and instead tries to stringify the it as an object:

```
JSON.stringify(new Error('error')) // "{}"
```

## Changelog

[JavaScript] [Changed] - Add support for stringifying error object messages to safeStringify

## Test Plan

Tests:
Screen Shot 2019-07-18 at 8 39 52 PM

Lint:
Screen Shot 2019-07-18 at 8 43 35 PM

### DIFF
--- a/Libraries/Utilities/__tests__/stringifySafe-test.js
+++ b/Libraries/Utilities/__tests__/stringifySafe-test.js
@@ -47,4 +47,10 @@ describe('stringifySafe', () => {
     const result = stringifySafe(arg);
     expect(result).toEqual('["object" failed to stringify]');
   });
+
+  it('stringifySafe stringifies error messages', () => {
+    const error = new Error('error');
+    const result = stringifySafe(error);
+    expect(result).toEqual('Error: error');
+  });
 });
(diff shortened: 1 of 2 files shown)


janicduplessis:
Chrome formats it like this

image

Maybe we could do something similar (no sure if we want the stack tho).

danReynolds:
@janicduplessis So put `ErrorClassName:` in front? Can do. I think the stack would be taking it too far, `stringifySafe` is really just for converting whatever object into a stringified version of it, which for an error makes sense to be its message but not including a stack.

danReynolds:
Updated to prefix it by error type.

PR #25736, opened 2019-07-19
Fix building on tvOS when using CocoaPods

hsousa:
## Summary

This PR makes it possible to build on tvOS again, when using CocoaPods.

When RN moved from a subspec to multiple podspecs (#23559) all dependencies were added to `React` as "generic", but two of them as actually iOS only (`React-RCTActionSheet` and `React-RCTLinking`).

This change meant that `pod install` still works, since all dependencies are satisfied, but tvOS targets would fail to build since CocoaPods would create static libraries for both platform variants.

Previously this wasn't a problem since subspecs can be included individually for each platform.

## Changelog

[iOS] [Fixed] - Fix building on tvOS when using CocoaPods

## Test Plan

1. Create sample project with iOS and tvOS targets
2. Use CocoaPods to inject a React Native dependency
3. Run `pod install`
4. Open the workspace and build both targets
5. Smile


cpojer:
@hsousa could you rebase this PR?

cpojer:
This is now outdated. If it is still a problem, please send a new PR :)

PR #25737, opened 2019-07-19
fix VirtualizedSectionList.scrollToSection ignoring viewOffset param

jsamr:
## Summary

PR #24735 introduced commit d376a44 to handle sticky headers gracefully. But it missed taking into account user-provided `viewOffset` parameter, overwriting it.

This is a straightforward change, adding the sticky header offset to the existing user provided `viewOffset` value. Also, I added `viewOffset` which was missing from the method signature [while described on the official documentation](https://facebook.github.io/react-native/docs/sectionlist#scrolltolocation).

## Changelog

[Javascript] [Fixed] - Fix VirtualizedSectionList.scrollToSection ignoring viewOffset param

## Test Plan

I believe it is too trivial to require screenshots. Just look at the 3 changed lines.

### DIFF
--- a/Libraries/Lists/VirtualizedSectionList.js
+++ b/Libraries/Lists/VirtualizedSectionList.js
@@ -143,17 +143,18 @@ class VirtualizedSectionList<
     itemIndex: number,
     sectionIndex: number,
     viewPosition?: number,
+    viewOffset?: number,
   }) {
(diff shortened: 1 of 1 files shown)


cpojer:
See https://github.com/facebook/react-native/pull/25997 with the proper long-term fix.

PR #25738, opened 2019-07-19
Allow jest globals in __mocks__ directories

artdent:
## Summary

The `__mocks__` directories should be treated the same as `__tests__`
for the purpose of linting. See https://jestjs.io/docs/en/manual-mocks

## Changelog

[Internal] [Changed] - eslint: allow jest globals in `__mocks__` directories

## Test Plan

I installed the updated package locally in my project and verified
that the `no-undef` rule no longer fires for references to `jest` in
`__mocks__` directories.

### DIFF
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -58,7 +58,7 @@ module.exports = {
       },
     },
     {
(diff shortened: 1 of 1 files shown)


osdnk:
Thanks!

PR #25746, opened 2019-07-20
Fix/revamp rn tester

satheeshwaran:
## Summary

Revamp of RNTester.
Removed non-core components for Android that will be moved to react-native-community soon.
Fixed status bar style in iOS for the first launch.
Removed non-core components for iOS that will be moved to react-native-community soon.

## Changelog

[General] [Deprecated] - Removed non-core components from RNTester for Android & iOS

## Test Plan

- Tested the RNTester app after changes. 
- Fixed all unit tests that might fail after this change.

#### iOS
Simulator Screen Shot - iPhone X - 2019-07-20 at 20 49 45

#### Android
Screenshot_1563628168


cpojer:
Thank you for your PR. I would prefer to keep these examples around until we remove each individual component which we can only do from Facebook's side.

Would you mind changing your PR to only include the status bar change? Thank you!

cpojer:
Closing due to inactivity. Please create a new PR if you have time to get back to this with the feedback from me.

Issue #25594, opened 2019-07-11
TextInput prevent scroll on ScrollView


Aiiros:
In a too long form, scroll is only available if the touch start outside the `TextInput` fields

React Native version:
```
info Fetching system and libraries information...
System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
    Memory: 1.92 GB / 7.67 GB
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 8.10.0 - /usr/bin/node
    npm: 3.5.2 - /usr/bin/npm
  SDKs:
    Android SDK:
      API Levels: 22, 23, 26, 27, 28
      Build Tools: 23.0.1, 27.0.3, 28.0.3
      System Images: android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.0 => 0.60.0 
  npmGlobalPackages:
    react-native-cli: 2.0.1
    react-native-debugger: 1.1.0
    react-native: 0.59.3
```
## Steps To Reproduce

1. create a ScrollView with `flex: 1`
2. put more TextInputs than your screen can handle
3. try to scroll with your touch starting on a field

Describe what you expected to happen:

I expect the page to scroll

Snack, code example, or link to a repository:

```
     <ScrollView style={{ flex: 1, }}>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <View style={{ height: 150, }} />
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
            <TextInput placeholder="test" textAlign='center'/>
        </ScrollView>
```
snack: https://snack.expo.io/@aiiros/react-native-issue-25594

this is a duplicate of https://github.com/facebook/react-native/issues/15962 which has not yet been answered.

Edit:
Turns out `textAlign='center'` is causing all of this

RobinCsl:
Hi @Aiiros,

[...]

PR #25749, opened 2019-07-20
[Android] Fix scrolling issues with TextInputs in ScrollViews

RobinCsl:
_This is my first PR to this repository and I would appreciate any feedback. Thanks_

## Summary

This change fixes https://github.com/facebook/react-native/issues/25594

The motivation is to allow `ScrollView` components to be correctly scrollable on Android when the touch falls on a `TextInput` component.
When a given `TextInput` component has `textAlign` property set to `center` or `right`, or when its value is long enough that the `TextInput` becomes horizontally scrollable, it is no longer possible to use it as an anchor for a touch aiming at scrolling the parent `ScrollView`.

Here is a [Snack](https://snack.expo.io/@markuscosinus/https:-github.com-facebook-react-native-issues-25594) showing this behaviour (run it for Android, iOS is fine).

Here is a video (in GIF form) to show the behaviour in an example in RNTester (sadly, the touches don't really show, but I'm trying to take each input field as an anchor point to scroll vertically in the first half of the video, and horizontally in the second half of the video).

<p align="center">
No Scrolling from specific TextInputs
</p>

Even though the original implementation seems completely reasonable, the way Android seems to compute `canScrollHorizontally` or `canScrollVertically` seems non-functional when the text is centered or aligned to the right.

I first tried to relax the condition to be "can't scroll vertically OR can't scroll horizontally", which yielded the desired behaviour, and commenting out the whole condition yielded the same. Upon inspection of the commit history, this change is more or less reverting this commit https://github.com/facebook/react-native/commit/372d001a5d5e24b68d194e01c356d973642fac86

## Changelog

[Android] [Fixed] - TextInput components can be valid anchors to scroll within a ScrollView component

## Test Plan

I ran my changes on RNTester on the Android emulator (Nexus_5X_API_27) but could not test on an Android device since I don't have one handy.

I added an example called "ScrollView with TextInputs".

Here is a video (in GIF form) to show the new behaviour: `TextInput` components can be taken as anchor points to scroll down the view, or sideways in the second example. (Sadly once again, touches aren't visible and the same comment as above applies)

<p align="center">
Can scroll from TextInputs
</p>

The only "downside" to this fix is that it is no longer possible to scroll horizontally single-line `TextInput` components' content if it is within a horizontal `ScrollView`.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -23,7 +23,6 @@ import android.text.method.QwertyKeyListener;
 import android.util.TypedValue;
 import android.view.Gravity;
 import android.view.KeyEvent;
-import android.view.MotionEvent;
 import android.view.View;
 import android.view.accessibility.AccessibilityNodeInfo;
 import android.view.inputmethod.EditorInfo;
@@ -80,7 +79,6 @@ public class ReactEditText extends EditText {
   private @Nullable ContentSizeWatcher mContentSizeWatcher;
   private @Nullable ScrollWatcher mScrollWatcher;
   private final InternalKeyListener mKeyListener;
-  private boolean mDetectScrollMovement = false;
   private boolean mOnKeyPress = false;
   private TextAttributes mTextAttributes;
 
@@ -151,31 +149,6 @@ public class ReactEditText extends EditText {
   @Override
   protected void onLayout(boolean changed, int left, int top, int right, int bottom) {
     onContentSizeChange();
-  }
-
-  @Override
-  public boolean onTouchEvent(MotionEvent ev) {
-    switch (ev.getAction()) {
-      case MotionEvent.ACTION_DOWN:
-        mDetectScrollMovement = true;
-        // Disallow parent views to intercept touch events, until we can detect if we should be
-        // capturing these touches or not.
-        this.getParent().requestDisallowInterceptTouchEvent(true);
-        break;
-      case MotionEvent.ACTION_MOVE:
-        if (mDetectScrollMovement) {
-          if (!canScrollVertically(-1)
-              && !canScrollVertically(1)
-              && !canScrollHorizontally(-1)
-              && !canScrollHorizontally(1)) {
-            // We cannot scroll, let parent views take care of these touches.
-            this.getParent().requestDisallowInterceptTouchEvent(false);
-          }
-          mDetectScrollMovement = false;
-        }
-        break;
-    }
-    return super.onTouchEvent(ev);
   }
 
   // Consume 'Enter' key events: TextView tries to give focus to the next TextInput, but it can't
--- a/RNTester/js/examples/ScrollView/ScrollViewTextInputsExample.js
+++ b/RNTester/js/examples/ScrollView/ScrollViewTextInputsExample.js
@@ -1,0 +1,130 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const React = require('react');
+const {ScrollView, TextInput} = require('react-native');
+
+exports.title = 'ScrollView with TextInputs';
+exports.description = 'Displays a list of TextInputs in a ScrollView';
+exports.examples = [
+  {
+    title: 'ScrollView with TextInputs - vertical',
+    description:
+      'Displays a list of scrollable TextInputs in a vertical ScrollView',
+    render(): React.Element<any> {
+      return (
+        <ScrollView style={{flex: 1}}>
+          <TextInput placeholder="default" />
+          <TextInput placeholder="center" style={{textAlign: 'center'}} />
+          <TextInput placeholder="right" style={{textAlign: 'right'}} />
+          <TextInput placeholder="left" style={{textAlign: 'left'}} />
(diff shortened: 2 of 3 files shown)


Aiiros:
Any updates ?

RobinCsl:
I tried to run the E2E test suite on Android but cannot seem to make it work, it seems the docker container is missing `watchman`. But I did manage to run the E2E test suite on the iOS simulator.

I can add regression tests from the example, but I cannot really test whether they failed prior to my fix on Android as they should be passing on iOS either way. Is there something else I could do to test it in a better way?

JoshuaGross:
@RobinCsl If you could add passing e2e tests that'd be great; that way if others have time, someone else can verify that it's broken without your fix on Android and is resolved after.

JoshuaGross:
Digging into this more, I'm a bit concerned about just reverting the original changes; I assume that we'll just cause the original bug to reappear if we aren't careful. I'll try to get more context about the original issue. In the meantime, is there a way to not delete this logic but instead take your specific repro cases into account in the logic to fix the bug? 

One more tip: in Android system settings you can enable showing touches, which will make your demonstration video more clear. It's pretty hard to tell in the first video what's wrong without being able to see touches. https://medium.theuxblog.com/enabling-show-touches-in-android-screen-recordings-for-user-research-cc968563fcb9

RobinCsl:
>  If you could add passing e2e tests that'd be great

I'll do that this weekend hopefully!

> In the meantime, is there a way to not delete this logic but instead take your specific repro cases into account in the logic to fix the bug?

If I remember well, I had tried to play with the logic I offer to delete and I did not manage to fix the issue other than by deleting it. I can try again though.

> It's pretty hard to tell in the first video what's wrong without being able to see touches.

Absolutely, thank you for the Medium article, I had a look at it and could get my emulator to show touches. I'll re-record the examples this weekend and update my PR description.

Thank you for your feedback!

RobinCsl:
I tried to add passing e2e tests and I ran into the issue that it's difficult to reliably target where the scrolling takes place in Detox. There's meant to be a way to specify the start point (cf. `startPositionX`, `startPositionY` in [`scroll` method](https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md#scrollpixels-direction-startpositionxnan-startpositionynan)), but I could not see it work on my iOS emulator when e2e tests were running.

I then had to modify my example to display the content of the ScrollView almost like a Picker component, in which case I was sure that the scrolling performed by Detox would be done "over" the TextInput I wanted to test.

Finally, I ran into another issue that, on iOS, multiline TextInput's do capture the scroll event sometimes and that's probably what the original code that I offer to delete is solving (meaning, for the behaviour to be the same on Android). What's more, one-line TextInput don't scroll on iOS
when they do on Android, so that's another difference the e2e tests might need to take into account if we really want to cover everything.

Sorry I don't have something more substantial to share just yet, I'll try and continue on it this week.

daveyjones:
@RobinCsl Thanks for all your work so far on this issue! Definitely an important one to solve.

jschenkDD:
Any news on that?

RobinCsl:
Hi @jschenkDD,
Feel free to continue where I left off. These last few months have been really busy.

ffastym:
Still waiting for merge...

fabriziobertoglio1987:
https://github.com/facebook/react-native/commit/372d001a5d5e24b68d194e01c356d973642fac86

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -153,31 +151,6 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
...
-        mDetectScrollMovement = true;
-        // Disallow parent views to intercept touch events, until we can detect if we should be
-        // capturing these touches or not.
-        this.getParent().requestDisallowInterceptTouchEvent(true);
-        break;
-      case MotionEvent.ACTION_MOVE:
-        if (mDetectScrollMovement) {
-          if (!canScrollVertically(-1)
-              && !canScrollVertically(1)
-              && !canScrollHorizontally(-1)
```
fabriziobertoglio1987, line 169: https://github.com/facebook/react-native/commit/372d001a5d5e24b68d194e01c356d973642fac86 does not detect direction of the scroll

>Note: this might change in the future to also detect the direction of the scroll, and
only block the scroll if the component can scroll in that direction. This is however not
trivial, since the scroll needs to be above some threshold of pixels. Blocking the
parent view from scrolling until that threshold is passed might cause incorrect behavior
in the parent component.



The checks [`canScroll`](https://developer.android.com/reference/android/view/View#canScrollVertically(int)) are performed for all directions (for ex. to scroll down, we check that TextInput can scroll Up, Left and Right), if the `TextInput` can not scroll in any direction, then the scroll `requestDisallowInterceptTouchEvent` is disabled and scrolling goes back to the parent. 



I would have to build a logic to detect the scroll direction and handle it correctly.

fabriziobertoglio1987: The issue is caused by `canScrollHorizontally` returns allways `true` for:

```javascript
<TextInput multiline={false} textAlign="center" />
```

 

The method definition of [`canScrollHorizontally`](https://github.com/aosp-mirror/platform_frameworks_base/blob/e80b45506501815061b079dcb10bf87443bd385d/core/java/com/android/internal/widget/ViewPager.java#L2380-L2394).

I'll prepare a workaround

shmkane:
any update on this?

AhteshamKhanCoder:
@RobinCsl Thanks for this work really appreciable

shoxter:
Thanks for trying to solve this...

Issue #25456, opened 2019-07-02
React Native Picker shows ghost picked value


Aiiros:
I'm using the default react native picker on Android, and some values's color are randomly changed as i scroll. 

Note that this bug is only visual, selected values are working fine.

React Native version:

```
System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
    Memory: 1.11 GB / 7.67 GB
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 8.10.0 - /usr/bin/node
    npm: 3.5.2 - /usr/bin/npm
  SDKs:
    Android SDK:
      API Levels: 22, 23, 26, 27, 28
      Build Tools: 23.0.1, 27.0.3, 28.0.3
      System Images: android-27 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.0 => 0.60.0 
  npmGlobalPackages:
    react-native-cli: 2.0.1
    react-native-debugger: 1.1.0
    react-native: 0.59.3
```

## Steps To Reproduce
use the `Picker` component from react native on android,
put more items in it than your screen can contain,
scroll back and forth a few times and some values should be greyed out

Describe what you expected to happen:

I need the component not to randomly change some colors as i scroll

Snack, code example, or link to a repository:

Gif demo: https://i.imgur.com/bDQov0K.gif
Snack exemple: https://snack.expo.io/rJNN1JKeB
Stack issue: https://stackoverflow.com/questions/56853193/react-native-picker-show-ghost-picked-value

cabelitos:
Hello @Aiiros I submitted a fix for this problem. You and others are welcome to take a look and add comments/suggestions.

Please see it in here: https://github.com/facebook/react-native/pull/25750

PR #25750, opened 2019-07-20
[Android][Fix] Revert Picker item original color

cabelitos:
## Summary

Since the Android's Picker implementation uses an ArrayAdapter,
it means that the views that were created may be reused for other items
in order to save memory. With this in mind, if one sets the Picker.Item
prop color for only certain items there might be an state that
some items that does not have the color set will end up appearing
with the wrong color. This happens because, this new item is
reusing a view of an item that had the color prop set.
In order to avoid this problem, once a new view is created
the ReactPickerAdapter will save the original color and
re-apply if the item does not have the color prop.

## Changelog

[Android] [Fixed] - Picker.Item displays wrong colors

## Test Plan
On android execute the code below. Only the FIRST item should be red.

```javascript
import React from 'react';
import { StyleSheet, View, Picker } from 'react-native';

const values = new Array(100);

for (let i = 0; i < values.length; i += 1) {
  values[i] = (i * i).toString();
}

const App = () => {
  const [selected, setSelected] = React.useState(0);
  const onValueChange = React.useCallback((_, idx) => {
    setSelected(idx);
  }, []);
  return (
      <Picker onValueChange={onValueChange} selectedValue={values[selected]}>
        {values.map((v, i) => (
          <Picker.Item
            key={v}
            value={v}
            label={v}
            {...(!i ? { color: 'red' } : {})}
          />
        ))}
      </Picker>
    </View>
  );
};

export default App;

  container: {
    flex: 1,
    justifyContent: 'center',
    paddingHorizontal: 20,
  },
});

```

### Without the patch

You should see various items with the red color (when only the first one should be red)
picker-not-working

### With the patch
Only the first item is red.

picker-working

Closes https://github.com/facebook/react-native/issues/25456

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPickerAdapter.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPickerAdapter.java
@@ -6,6 +6,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
 import android.widget.TextView;
+import android.content.res.ColorStateList;
 import androidx.annotation.Nullable;
 import com.facebook.infer.annotation.Assertions;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPickerAdapter.java
```diff
@@ -50,6 +55,9 @@ private View getView(int position, View convertView, ViewGroup parent, boolean i
       textView.setTextColor(mPrimaryTextColor);
     } else if (item.color != null) {
       textView.setTextColor(item.color);
+    } else if (textView.getTag() != null && !isNew) {
+      // In case the new item does not set the color prop, go back to the default one
+      textView.setTextColor((ColorStateList) textView.getTag());
     }
 
     return convertView;
```
mdvacca: Since you are touching this file, can you return textView instead?

cabelitos:
Thank you for your review @mdvacca. I'm doing the change right now!

cabelitos:
@mdvacca did the change, please let me know what you think.
And once again thank you and your colleagues for the support.

mdvacca:
Looks great, thanks!

I'm landing the PR

Issue #25745, opened 2019-07-20
_RCTSetLogFunction build failure in release mode fixed with Dead Code Stripping = No


TheSavior:
I created a new project with 0.60.4. I was able to build and deploy the app to my physical iOS device with Debug scheme. I wanted to build for release by following the instructions here: https://facebook.github.io/react-native/docs/running-on-device

Specifically this line "go to Product → Scheme → Edit Scheme. Select the Run tab in the sidebar, then set the Build Configuration dropdown to Release."

After making that change I tried building a release mode version of the app to my device and the build failed with:
```
Undefined symbol: _RCTSetLogFunction
```

I found [this stack overflow article](https://stackoverflow.com/a/52541765) that says 
> in the "Linking" section, find"Dead Code Stripping" section and change it "No"

This change made the build pass successfully.

It looks like this is a common problem that has been around for a long time. It would be great to either change the default or figure out how to keep this code from being dead code eliminated. 

System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
    Memory: 273.86 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
    Watchman: 4.7.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
  IDEs:
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.4 => 0.60.4

PR #25751, opened 2019-07-21
[iOS] Fixes template build failed in release mode

zhongwuzw:
## Summary

Fixes https://github.com/facebook/react-native/issues/25745, Xcode stripped dead code in Release mode, and in template test code, it should only run in Debug mode, so we can use a macro to fix this issue.

## Changelog

[iOS] [Fixed] - Fixes template build failed in release mode

## Test Plan

1. Create a new project using `react-native init AwesomProject`.
2. Change project target scheme to `Release`.
3. Build and it should success.

### DIFF
--- a/template/ios/HelloWorldTests/HelloWorldTests.m
+++ b/template/ios/HelloWorldTests/HelloWorldTests.m
@@ -40,11 +40,13 @@
   BOOL foundElement = NO;
 
   __block NSString *redboxError = nil;
+#ifdef DEBUG
   RCTSetLogFunction(^(RCTLogLevel level, RCTLogSource source, NSString *fileName, NSNumber *lineNumber, NSString *message) {
     if (level >= RCTLogLevelError) {
       redboxError = message;
     }
   });
+#endif
 
   while ([date timeIntervalSinceNow] > 0 && !foundElement && !redboxError) {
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for this. I wouldn’t have expected a tests file to cause the app to fail to build. Is that normal? 

Also, would it make sense to validate that release builds build successfully in CI to keep this kind of issue from sneaking back in? 

Should we also have a release scheme in the default template? I can’t imagine that anyone starts an RN project without wanting to build for release so they have to go through all those steps on their own.

zhongwuzw:
> Thanks for this. I wouldn’t have expected a tests file to cause the app to fail to build. Is that normal?

Yeah, we should avoid.

> Also, would it make sense to validate that release builds build successfully in CI to keep this kind of issue from sneaking back in?

We don't use template project in main `react-native`'s CI, I think only [cli](https://github.com/react-native-community/cli) uses it, so more suitable for cli to add it in CI?

> Should we also have a release scheme in the default template? I can’t imagine that anyone starts an RN project without wanting to build for release so they have to go through all those steps on their own.

Emm, we can left users to create, users don't use release mode to develop in most of the time, only when profile or publish app.

mmazzarolo:
Applying this change to `/ios/{appname}Tests/{appname}Tests.m` fixes the issue.  
Thanks!

PR #25774, opened 2019-07-23
Fix building process

holmeshe:
## Summary

The `react-native init AwesomeProject` does not work out of the box, which is majorly caused by a miss configured glog podspecs.

## Changelog

[iOS] [Fixed] - Fixed various issues in third-party-podspecs/glog.podspec.
[iOS] [Fixed] - Fixed scripts/ios-configure-glog.sh by correcting the public header directory.

## Test Plan

In the `AwesomeProject`, change the react-native dependencies to the branch, and `AwesomeProject` is runnable.


third-party-podspecs/glog.podspec
```diff
@@ -11,19 +11,33 @@ Pod::Spec.new do |spec|
...
                       'src/utilities.cc',
-                      'src/vlog_is_on.cc'
+                      'src/vlog_is_on.cc',
+                      'src/config.h.in',
+                      'config.guess',
+                      'config.sub',
+                      'configure',
+                      'configure.ac',
+                      'Makefile.in',
+                      'install-sh',
```
osdnk: do we really need these files. Have you been verifying each of them?

holmeshe: Yes they are individually added and verified. Basically, `configure` is required because the target architecture only becomes available in compile time. (whilst during pod installation, the architecture cannot be determined). And `install.sh`, `config.sub`, `config.guess` and `Makefile.in` are required by `configure` and its subroutines. The other are source files that are required to the compile the pod.

cpojer:
@osdnk what's the status of this one?

holmeshe:
Just verified that the `react-native init` flow is working. Hence closed.
Thanks for reviewing and following up @cpojer  @osdnk

PR #25793, opened 2019-07-23
Allow Animation EndResult callback to return Promise

draperunner:
## Summary

I am sending an asynchronous function as callback to the `.start` method of `Animation.parallel([...]).start(callback)`. Flow does not like this, as the `EndCallback` type is saying that these callbacks must return `void`. Since my callback returns `Promise<void>` this results in an error.

Does it really matter what the callback returns?

## Changelog

[General] [Changed] - Make Animation EndCallback type allow any return value

## Test Plan

I have run `yarn flow`, which reported no errors.

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedModule.js
+++ b/Libraries/Animated/src/NativeAnimatedModule.js
@@ -14,7 +14,7 @@ import type {TurboModule} from '../../TurboModule/RCTExport';
 import * as TurboModuleRegistry from '../../TurboModule/TurboModuleRegistry';
 
 type EndResult = {finished: boolean};
-type EndCallback = (result: EndResult) => void;
+type EndCallback = (result: EndResult) => any;
 
 export type EventMapping = {|
   nativeEventPath: Array<string>,
--- a/Libraries/Animated/src/NativeAnimatedModule.js
+++ b/Libraries/Animated/src/NativeAnimatedModule.js
@@ -14,7 +14,7 @@ import type {TurboModule} from '../../TurboModule/RCTExport';
 import * as TurboModuleRegistry from '../../TurboModule/TurboModuleRegistry';
 
 type EndResult = {finished: boolean};
-type EndCallback = (result: EndResult) => any;
+type EndCallback = (result: EndResult) => void | Promise<void>;
 
 export type EventMapping = {|
   nativeEventPath: Array<string>,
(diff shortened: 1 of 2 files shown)


Libraries/Animated/src/NativeAnimatedModule.js
```diff
@@ -14,7 +14,7 @@ import type {TurboModule} from '../../TurboModule/RCTExport';
 import * as TurboModuleRegistry from '../../TurboModule/TurboModuleRegistry';
 
 type EndResult = {finished: boolean};
-type EndCallback = (result: EndResult) => void;
+type EndCallback = (result: EndResult) => any;
```
osdnk:
```suggestion
type EndCallback = (result: EndResult) => Promise<void> | void;
```

Libraries/Animated/src/animations/Animation.js
```diff
@@ -14,7 +14,7 @@ const NativeAnimatedHelper = require('../NativeAnimatedHelper');
 import type AnimatedValue from '../nodes/AnimatedValue';
 
 export type EndResult = {finished: boolean};
-export type EndCallback = (result: EndResult) => void;
+export type EndCallback = (result: EndResult) => any;
```
osdnk:
```suggestion
export type EndCallback = (result: EndResult) => Promise<void> | void;
```

draperunner:
I see. I have updated the PR with `void | Promise<void>` now.

cpojer:
I have thought about this change for the past two days and I'm not sure we want to merge this, even with `Promise<void>`. As @osdnk said, we want to ensure that there is no expectation of any return value.

What if you wrap your async operation like this:
```
.start(() => {
  asyncCallback();
});
```

That may be a bit more code to type but it explains the intent without having to change types to accommodate for a slightly shorter way of writing this.

draperunner:
I understand. The workaround with a wrapping function is OK for me. Feel free to close this PR if you want.

osdnk:
@cpojer I think we currently are following strategy to allow having `async` methods as a result of native callback:
https://github.com/facebook/react-native/blob/fb7b2d353356f67a3149c19ad733d46ec6842767/Libraries/Types/CodegenTypes.js#L21
I believe it's fine to consistently allow for using promises is these cases.

osdnk:
I'm importing it. If you, @cpojer, don't want to merge it, feel free to close this PR!

cpojer:
@osdnk I didn't realize we were doing this in some parts of RN already. I guess.. shipit then!

osdnk:
Sorry, I'm afraid we need to revert this part of code. It's not compatible with our codegen. I hope we will be able to restore it soon.

PR #25803, opened 2019-07-24
[iOS] Fixes iOS packager connection not work

zhongwuzw:
## Summary

Macro `ENABLE_PACKAGER_CONNECTION` invalid because of `__has_include` can't find the header now. Leads to packager connection not work anymore.

## Changelog

[iOS] [Fixed] - Fixes iOS packager connection not work

## Test Plan

1. Init a new project.
2. Run and input `curl http://localhost:8081/reload` in terminal. Reload operation can execute.

### DIFF
--- a/React/Base/RCTDefines.h
+++ b/React/Base/RCTDefines.h
@@ -55,7 +55,7 @@
 #endif
 
 #ifndef ENABLE_PACKAGER_CONNECTION
-#if RCT_DEV && __has_include(<React/RCTPackagerConnection.h>) && !TARGET_OS_UIKITFORMAC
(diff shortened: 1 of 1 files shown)


cpojer:
It seems like the existing definition is the one that works for internal use cases at FB, and the change you made for open source. I'm not sure how to resolve, cc @sammy-SC.

zhongwuzw:
@cpojer This works in internal at FB like below? add an `OR` statement.
```
#if RCT_DEV && (__has_include("RCTPackagerConnection.h") || __has_include(<React/RCTPackagerConnection.h>)) && !TARGET_OS_UIKITFORMAC
```

TheSavior:
Thanks for tracking this down and fixing it!

Issue #25629, opened 2019-07-13
Keystore file '/Project-Folder/android/app/debug.keystore' not found for signing config 'debug' in react-native 0.60


harikanammi:
I created a sample application with react-native version 0.60. I am trying to run application in android. But i am getting this error `Keystore file '/Project-Folder/android/app/debug.keystore' not found for signing config 'debug' `. I don't know how to resolve this error. Please give any suggestions for this issue.

React Native version:
>     "react": "16.8.6",
>     "react-native": "0.60.0"

## Steps To Reproduce
1. Trying to run application in android.
2. But getting Keystore file error in android.

kiranjd:
Which commands are you using?

cmorbitzer:
You can generate the debug keystore by running this command in the `android/app/` directory: `keytool -genkey -v -keystore debug.keystore -storepass android -alias androiddebugkey -keypass android -keyalg RSA -keysize 2048 -validity 10000`

harikanammi:
hi @kiranjd and @cmorbitzer thank you for your response. I missed the file called **debug.keystore**. So i placed this **debug.keystore** file in my project folder it solves the error i.e., ` Keystore file '/Project-Folder/android/app/debug.keystore' not found for signing config 'debug'`.

jeancabral:
> You can generate the debug keystore by running this command in the `android/app/` directory: `keytool -genkey -v -keystore debug.keystore -storepass android -alias androiddebugkey -keypass android -keyalg RSA -keysize 2048 -validity 10000`

Works fine to me Thx

bondehagen:
Or just download from official template https://raw.githubusercontent.com/facebook/react-native/master/template/android/app/debug.keystore

priyajainSE:
@bondehagen Where are we supposed to store this file in our react native project?

JohnnyFl:
> @bondehagen Where are we supposed to store this file in our react native project?

put `debug.keystore` in the `android/app/` directory

tuantvk:
Ubuntu:

 Running this command in the **android/app/** directory: *keytool -genkey -v -keystore debug.keystore -storepass android -alias androiddebugkey -keypass android -keyalg RSA -keysize 2048 -validity 10000*

In folder **/home/** *Ctrl + H* show all file hidden

Copy file **debug.keystore** in */home/.android/* to *android/app/* directory your project, and rebuild

zfing:
[...]

PR #25807, opened 2019-07-24
Adding the debug.keystore file back

bondehagen:
Fixes https://github.com/facebook/react-native/issues/25629 that happened after v60 upgrade when using the diff tool https://react-native-community.github.io/upgrade-helper/?from=0.59.10&to=0.60.4

## Summary

When following the diff for upgrading to react native 60. The debug.keystore is missing. It's added in the repository but ignored in the .gitignore, so it does not show. This adds an exception for this file.

## Changelog

[Android] [Fixed] - Add exception in .gitignore for `debug.keystore` to the android template.

## Test Plan

Create a new project from template and check that debug.keystore will be checked in when committing.

### DIFF
--- a/template/_gitignore
+++ b/template/_gitignore
@@ -40,6 +40,7 @@ yarn-error.log
 buck-out/
 \.buckd/
 *.keystore
+!debug.keystore
 
 # fastlane
 #


mdvacca:
Importing for review

PR #25816, opened 2019-07-25
Remove 's.static_framework = true' requirement for podspec

jtreanor:
## Summary

As part of the fix for https://github.com/facebook/react-native/issues/25349 I added `s.static_framework = true` to each podspec in repo (see https://github.com/facebook/react-native/pull/25619#discussion_r306993309 for more context).

This was required to ensure the existing conditional compilation with `#if RCT_DEV` and `__has_include` still worked correctly when `use_frameworks!` is enabled.

However, @fkgozali pointed out that it would be ideal if we didn't have this requirement as it could make life difficult for third-party libraries.

This removes the requirement by moving `React-DevSupport.podspec` and `React-RCTWebSocket.podspec` into `React-Core.podspec` as subspecs. This means the symbols are present when `React-Core.podspec` is built dynamically so `s.static_framework = true` isn't required.

This means that any `Podfile` that refers to `React-DevSupport` or `React-RCTWebSocket` will need to be updated to avoid errors.

## Changelog

I don't think this needs a changelog entry since its just a refinement of https://github.com/facebook/react-native/pull/25619.

## Test Plan

Check `RNTesterPods` still works both with and without `use_frameworks!`:

1. Go to the `RNTester` directory and run `pod install`.
2. Run the tests in `RNTesterPods.xcworkspace` to see that everything still works fine.
3. Uncomment the `use_frameworks!` line at the top of `RNTester/Podfile` and run `pod install` again.
4. Run the tests again and see that it still works with frameworks enabled.


fkgozali:
Nice, thank you! I'll merge this.

fkgozali:
Looks like we've fully fixed the issue in master then? (except Fabric, which is separate, of course)

PR #25818, opened 2019-07-25
Test RNTesterPods on CI with use_frameworks! enabled

jtreanor:
## Summary

This adds a `test_ios_frameworks` job to CircleCI to test the `RNTesterPods` project with `use_frameworks!` enabled. It will ensure the issue in https://github.com/facebook/react-native/issues/25349 is not reintroduced as suggested in https://github.com/facebook/react-native/pull/25619#issuecomment-514380653.

## Changelog

[iOS] [Internal] - Added CircleCI job for testing `RNTesterPods` with `use_frameworks!` enabled.

## Test Plan

Tests seem to be failing on `master` at the moment but you can see that the new job builds successfully [here](https://circleci.com/gh/facebook/react-native/103929?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). You can confirm it installs the pods with `use_frameworks!` by seeing that `Installing pods with use_frameworks!` is at the start of the log for the `Generate RNTesterPods Workspace` step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,15 +122,19 @@ commands:
       steps:
         type: steps
     steps:
+      - run:
+          name: Setup CocoaPods cache
+          # Copy RNTester/Podfile.lock since it can be changed by pod install
+          command: cp RNTester/Podfile.lock RNTester/Podfile.lock.bak
       - restore_cache:
           keys:
-            - v1-pods-{{ checksum "RNTester/Podfile.lock" }}
-            - v1-pods-
+            - v1-pods-{{ .Environment.CIRCLE_JOB }}-{{ checksum "RNTester/Podfile.lock.bak" }}
+            - v1-pods-{{ .Environment.CIRCLE_JOB }}-
       - steps: << parameters.steps >>
       - save_cache:
           paths:
             - RNTester/Pods
-          key: v1-pods-{{ checksum "RNTester/Podfile.lock" }}
+          key: v1-pods-{{ .Environment.CIRCLE_JOB }}-{{ checksum "RNTester/Podfile.lock.bak" }}
 
   download_gradle_dependencies:
     steps:
@@ -301,6 +305,10 @@ jobs:
   # Runs unit tests on iOS devices
   test_ios:
     executor: reactnativeios
+    parameters:
+      use_frameworks:
+        type: boolean
+        default: false
     environment:
       - REPORTS_DIR: "./reports"
     steps:
@@ -320,6 +328,13 @@ jobs:
           name: Fetch CocoaPods Specs
           command: |
             curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
+
+      - when:
+          condition: << parameters.use_frameworks >>
+          steps:
+            - run:
+                name: Set USE_FRAMEWORKS=1
+                command: echo "export USE_FRAMEWORKS=1" >> $BASH_ENV
 
       - with_pods_cache_span:
           steps:
@@ -607,6 +622,11 @@ workflows:
           requires:
             - setup_android
       - test_ios:
+          requires:
+            - setup_ios
+      - test_ios:
+          name: test_ios_frameworks
+          use_frameworks: true
           requires:
             - setup_ios
       - test_ios_e2e:
--- a/RNTester/Podfile
+++ b/RNTester/Podfile
@@ -2,10 +2,12 @@ platform :ios, '9.0'
 
 require_relative '../scripts/autolink-ios'
 
-def pods()
-  # Uncomment for Swift
-  # use_frameworks!
+if ENV['USE_FRAMEWORKS'] == '1'
+  puts "Installing pods with use_frameworks!"
+  use_frameworks!
+end
 
+def pods()
   project 'RNTesterPods.xcodeproj'
 
   # Enable TurboModule
--- a/RNTester/Podfile.lock
+++ b/RNTester/Podfile.lock
@@ -354,36 +354,36 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
-  FBLazyVector: 34431b7e61740bed29b082ff81500b0ffafaffa0
-  FBReactNativeSpec: 0c434153f44d4a1cbd111fbde1502580475e9132
+  FBLazyVector: e38af664187698e33919cd2a190970acbca68630
+  FBReactNativeSpec: e734dea0614646e9b863785cf236c465fbcaf72d
   Folly: 30e7936e1c45c08d884aa59369ed951a8e68cf51
   glog: 1f3da668190260b06b429bb211bfbee5cd790c28
-  RCTRequired: 770d733d8b6881e0ecd1085d0eb7f57dfacc5d93
-  RCTTypeSafety: 8540a73bafd032be534cb683471f985dc656bfb6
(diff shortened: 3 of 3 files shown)


.circleci/config.yml
```diff
@@ -334,6 +334,15 @@ jobs:
...
 
       - run: yarn test-ios
+
+      - run:
+          name:  Generate RNTesterPods Workspace with frameworks
+          command: cd RNTester && pod install --verbose
+          environment:
+            USE_FRAMEWORKS: '1'
+      
+      - run: yarn test-ios
```
jtreanor: This just adds a second test run with frameworks enabled to the `test_ios` job. @fkgozali would you prefer this to be a separate parallel job (such as `test_ios_frameworks`) or is it better here?

jtreanor: Actually, I think a separate job probably is better. Let me know if you would like me to change it back.

jtreanor:
Its worth pointing out that fixing https://github.com/facebook/react-native/issues/23561 could potentially make this redundant. `process-podspecs.sh` validates the podspecs with `pod lib lint` both for frameworks and libraries. When that is working again its arguable that a full test run of `RNTesterPods` for `use_frameworks!` would not be needed.

A quick look suggests it shouldn't be too hard to fix `process-podspecs.sh`. I'll take a look next week.

fkgozali:
> A quick look suggests it shouldn't be too hard to fix process-podspecs.sh.

@hramos: do we still use that script? I think ios CI builds are fully using cocoapods already?

.circleci/config.yml
fkgozali: I'll have to defer to @hramos for the actual tradeoff discussion. To me, putting it in one place is easier to maintain (because it's basically testing exactly the same thing, just with framework output).

So for now, let's get this in as is and we can adjust it if needed in the future.

jtreanor:
I think the suggestion in that issue is to get it working again. Even if the repo uses CocoaPods extensively it could be useful to validate the podspecs.

fkgozali:
The CI builds are failing, do you know if it's related to this change?

https://circleci.com/gh/facebook/react-native/103930?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://circleci.com/gh/facebook/react-native/103932?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

jtreanor:
@fkgozali Not related. master is currently failing with the same error.

grabbou:
The tests are now failing on 0.61-stable branch (https://circleci.com/gh/facebook/react-native/107763) with the following:

> :x:  Undefined symbols for architecture x86_64
> Symbol: _OBJCCLASS$__RCTTypedModuleConstants
> Referenced from: objc-class-ref in RCTPlatform.o

I believe this may be related to the CocoaPods integration - does anyone happen to know where to dig this deeper?

jtreanor:
I can take a look @grabbou. I think this should be relatively straightforward to fix.

jtreanor:
Okay, so I did a `git bisect` and it seems that the test failure was first introduced by https://github.com/facebook/react-native/commit/2956eb266f1edad09165baf7c5db832c170ad41c.

It seems these tests have been failing on master for a while 😞 I'm looking into how much effort it will take to fix.

grabbou:
Great, thanks! Make sure to check 0.61-stable too, I've reverted one commit
there to make other lanes work as well.

In the worst case scenario, I think what we can do is we can just revert
this commit until a better solution is found.

fkgozali:
cc @osdnk 

@grabbou / @jtreanor that commit seems likely culprit. To revert it, you'll have to revert https://github.com/facebook/react-native/commit/28e18e41058a612290e0cf35c974297522c3ba1f beforehand (they're related change)

jtreanor:
> Great, thanks! Make sure to check 0.61-stable too, I've reverted one commit
there to make other lanes work as well.

Yeah, I've confirmed the problem on 0.61-stable.

> @grabbou / @jtreanor that commit seems likely culprit. To revert it, you'll have to revert 28e18e4 beforehand (they're related change)

Thanks @fkgozali! I'll try that out. The alternative approach which has worked for me is to add `s.static_framework = true` to all of the podspecs as we discussed in https://github.com/facebook/react-native/pull/25619#discussion_r306993309.

jtreanor:
Reverting those commits fixed the RCTPlatform build error but uncovered a number of other build errors (see below) that seem to be caused by https://github.com/facebook/react-native/commit/bf78d7969a48b63f7d57bf7cdb1b1d8af2f46e5c. That didn't revert cleanly.

```
Undefined symbols for architecture x86_64:
  "_RCTGetImageData", referenced from:
      ___45-[RCTImageStoreManager storeImage:withBlock:]_block_invoke in RCTImageStoreManager.o
      ___47-[RCTImageStoreManager(Deprecated) storeImage:]_block_invoke in RCTImageStoreManager.o
  "_OBJC_CLASS_$_RCTImageCache", referenced from:
      objc-class-ref in RCTImageLoader.o
  "_RCTImageHasAlpha", referenced from:
      ___43-[RCTImageLoader sendRequest:withDelegate:]_block_invoke in RCTImageLoader.o
  "_RCTTransformImage", referenced from:
      ___75-[RCTImageEditingManager cropImage:cropData:successCallback:errorCallback:]_block_invoke in RCTImageEditingManager.o
      RCTResizeImageIfNeeded(UIImage*, CGSize, double, RCTResizeMode) in RCTImageLoader.o
  "_RCTGetImageMetadata", referenced from:
      ___50-[RCTImageLoader getImageSizeForURLRequest:block:]_block_invoke in RCTImageLoader.o
  "_RCTDecodeImageWithData", referenced from:
      ___80-[RCTImageLoader decodeImageData:size:scale:clipped:resizeMode:completionBlock:]_block_invoke_3.264 in RCTImageLoader.o
  "_RCTTransformFromTargetRect", referenced from:
      ___75-[RCTImageEditingManager cropImage:cropData:successCallback:errorCallback:]_block_invoke in RCTImageEditingManager.o
      RCTResizeImageIfNeeded(UIImage*, CGSize, double, RCTResizeMode) in RCTImageLoader.o
  "_RCTTargetRect", referenced from:
      ___75-[RCTImageEditingManager cropImage:cropData:successCallback:errorCallback:]_block_invoke in RCTImageEditingManager.o
      RCTResizeImageIfNeeded(UIImage*, CGSize, double, RCTResizeMode) in RCTImageLoader.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

@fkgozali It seems that this problem will continue to occur with CoreModules. What are your thoughts on using `s.static_framework = true` in the podspecs to resolve it?

fkgozali:
> What are your thoughts on using s.static_framework = true in the podspecs to resolve it?

I think this is fine to do to unblock. I won't have cycle to debug further for the next week or so, so let's do this quick fix.

grabbou:
I have reverted https://github.com/facebook/react-native/pull/25816 on the 0.61-stable branch. However, this may still provide a lot of issues for existing applications, that depend on React podspec, as I am pretty sure most of them don't have `static_framework = true` turned on.

jtreanor:
Thanks @grabbou, you beat me to it.

> However, from what I understand from the linked PR, the issue may still occur within existing apps that depend on React podspecs and have use_frameworks turned on.

The issue will likely be for third party libraries which have podspecs. It would now be required to have `s.static_framework = true` for `use_frameworks!` to work.

> I think it makes sense to note this explicitly in the first release candidate and that we are working on providing a better solution to this problem.

That makes sense.

jtreanor:
Okay, I have found a simple solution that does not require `s.static_framework = true`. We just needed to update the CoreModules podspec. I've opened a PR in https://github.com/facebook/react-native/pull/26151 to target the 0.61-stable branch.

tamxomdao (commented):
Số tài khoản ngân hàng của mình

PR #25836, opened 2019-07-26
Fixed android bounding box

### LINKED ISSUE #19637: View bounds and touchable area incorrect after a rotate or a scale transform
  - [x] Use the latest React Native version: https://github.com/facebook/react-native/releases
  - [x] Run `react-native info` in your terminal and paste its contents under "Environment"
  - [x] Let us know how to reproduce the issue. Include a code sample, share a project, or
    share an app that reproduces the issue using https://snack.expo.io/

## Environment

    System:
      OS: macOS High Sierra 10.13.5
      CPU: x64 Intel(R) Core(TM) i7-6660U CPU @ 2.40GHz
      Memory: 4.38 GB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.5.0 - /usr/local/bin/node
      Yarn: 1.7.0 - ~/.yarn/bin/yarn
      npm: 6.1.0 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
      Android SDK:
        Build Tools: 23.0.1, 25.0.0, 25.0.1, 25.0.2, 26.0.0, 26.0.1, 26.0.2, 26.0.3
        API Levels: 19, 21, 22, 23, 24, 25, 26
    IDEs:
    npmPackages:
      react: 16.4.1 => 16.4.1 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-rename: 2.2.2

## Description

The bounds of transformed views, as reported by the inspector or as detected by the Touchable components, do not match what you seen on the screen. The touch area is not correctly transformed with the view.

**Screenshot of a { scale: 2 } transform**
The real view (scaled) is in pink, the incorrect touchable area is in blue.

**Screenshot of a { rotateZ: "-45deg" } transform**
The real view (rotated) is in pink, the incorrect touchable area is in blue.

*This issue has [already been reported](https://github.com/facebook/react-native/issues/6734) in 2016 but has been closed because it was missing reproduction steps. This bug has been here for a long time.*

## Steps to Reproduce

- **Use a real device** (I haven't been able to reproduce the bug on emulator, I have on a real Galaxy A5 2017)
  **UPDATE**: See [this comment](https://github.com/facebook/react-native/issues/19637#issuecomment-396045764) for an emulator reproduction.
- Use the inspector to display the view bounds
- See that the reported bounds (blue square) do not match the real view (pink square)
- Without the inspector, press somewhere on the real view **where the blue square didn't overlap**: the press is not correctly detected. onPressOut is triggered immediately after onPressIn, without onPress being triggered. onPressOut is immediately triggered even if you don't release the touch. Sometimes on the very first press, onPress is logged but it doesn't wait for touch release.

**Demo code with on-screen logging**
I've included some on-screen logging code to easily see the press events. 
This example shows the bug with the scale transform but it can be reproduced with a rotate transform as well.

```js
import React from "react";
import { AppRegistry, View, TouchableOpacity, Text } from "react-native";

let i = 1;

class Demo extends React.Component {

  state = {
    logs: []
  };

  log(str) {
    this.setState(state => ({
      logs: [...state.logs.slice(-5), `${i++}  - ${str}`]
    }));
  }

  render() {
    return (
      <View style={{ flex: 1 }}>

        <View style={{
          position: "absolute",
          top: 100,
          left: 100,
          height: 100,
          width: 100,
          transform: [{
            scale: 2 // you can also try rotateZ: "-45deg"
          }]
        }}>
          <TouchableOpacity
            onPress={() => this.log("pressed")}
            onPressIn={() => this.log("pressed in")}
            onPressOut={() => this.log("pressed out")}
            style={{
              flex: 1,
              backgroundColor: "pink"
            }}
          />
        </View>

        <View style={{ position: "absolute", bottom: 0, height: 200 }}>
          <Text style={{ fontWeight: "bold", fontSize: 20 }}>Logs</Text>
          <Text>{this.state.logs.join("\n")}</Text>
        </View>

      </View>
    );
  }
}

AppRegistry.registerComponent("Demo", () => Demo);
```

**Minimal example without on-screen logging**

```js
import React from "react";
import { AppRegistry, View, TouchableOpacity } from "react-native";

class Demo extends React.Component {

  render() {
    return (
      <View style={{ flex: 1 }}>

        <View style={{
          position: "absolute",
          top: 100,
          left: 100,
          height: 100,
          width: 100,
          transform: [{
            scale: 2 // you can also try rotateZ: "-45deg"
          }]
        }}>
          <TouchableOpacity
            onPress={() => alert("pressed")}
            style={{
              flex: 1,
              backgroundColor: "pink"
            }}
          />
        </View>

      </View>
    );
  }
}

AppRegistry.registerComponent("Demo", () => Demo);
```

## Expected Behavior

The bounds should match the view even when transformed. 
The touch area should match what you see.   

## Actual Behavior

The bounds are not correctly transformed. 
The touch area is not correctly transformed and press behaviour is broken where the real view and the incorrect bounds do not overlap.

floriancargoet:
## Summary

This PR fixes #19637. 

Summary of the issue: on Android, transformed touchables have press issues because the touchable's measurements are incorrect in the release phase. `UIManager.measure()` returns an incorrect size and position as explained [here](https://github.com/facebook/react-native/issues/19637#issuecomment-396065914)

This is easily seen with the inspector : 

**Screenshot of a { scale: 2 } transform**
The real view (scaled) is in pink, the incorrect touchable area is in blue.

**Screenshot of a { rotateZ: "-45deg" } transform**
The real view (rotated) is in pink, the incorrect touchable area is in blue.

## Changelog

[Android] [Fixed] - Fix UIManager.measure()

## Test Plan

Android now produces the same results as iOS as seen on these screenshots

| Android without fix | Android with fix | iOS |
| --- | --- | --- |
| Screenshot_1564133103 | Screenshot_1564130198 | Simulator Screen Shot - iPhone X - 2019-07-26 at 11 19 34 |

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -660,18 +660,18 @@ public class NativeViewHierarchyManager {
     outputBuffer[1] -= rootY;
   }
 
-  void computeBoundingBox(View v, int[] outputBuffer) {
+  private void computeBoundingBox(View v, int[] outputBuffer) {
     mBoundingBox.set(0, 0, v.getWidth(), v.getHeight());
     mapRectFromViewToWindowCoords(v, mBoundingBox);
 
-    outputBuffer[0] = (int) mBoundingBox.left;
-    outputBuffer[1] = (int) mBoundingBox.top;
-    outputBuffer[2] = (int) (mBoundingBox.right - mBoundingBox.left);
-    outputBuffer[3] = (int) (mBoundingBox.bottom - mBoundingBox.top);
+    outputBuffer[0] = Math.round(mBoundingBox.left);
+    outputBuffer[1] = Math.round(mBoundingBox.top);
+    outputBuffer[2] = Math.round(mBoundingBox.right - mBoundingBox.left);
+    outputBuffer[3] = Math.round(mBoundingBox.bottom - mBoundingBox.top);
   }
 
   // simplified version of the hidden Android method View.mapRectFromViewToScreenCoords()
-  void mapRectFromViewToWindowCoords(View v, RectF rect) {
+  private void mapRectFromViewToWindowCoords(View v, RectF rect) {
     Matrix m = v.getMatrix();
     if (!m.isIdentity()) {
       m.mapRect(rect);
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -660,6 +660,12 @@ public class NativeViewHierarchyManager {
     outputBuffer[1] -= rootY;
   }
 
+  /**
+   * Fills the outputBuffer with the view's bounding box [x, y, width, height]
+   *
+   * @param v
+   * @param outputBuffer
+   */
   private void computeBoundingBox(View v, int[] outputBuffer) {
     mBoundingBox.set(0, 0, v.getWidth(), v.getHeight());
     mapRectFromViewToWindowCoords(v, mBoundingBox);
@@ -670,7 +676,14 @@ public class NativeViewHierarchyManager {
     outputBuffer[3] = Math.round(mBoundingBox.bottom - mBoundingBox.top);
   }
 
-  // simplified version of the hidden Android method View.mapRectFromViewToScreenCoords()
+  /**
+   * Map a rectangle from view-relative coordinates to root-view-relative coordinates.
+   * This is a simplified version of the hidden Android method View.mapRectFromViewToScreenCoords()
+   * @see {@link View#mapRectFromViewToScreenCoords}
+   *
+   * @param v The view the coordinates are relative to
+   * @param rect The rectangle to be mapped
+   */
   private void mapRectFromViewToWindowCoords(View v, RectF rect) {
     Matrix m = v.getMatrix();
     if (!m.isIdentity()) {
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
@@ -7,6 +7,8 @@
 package com.facebook.react.uimanager;
 
 import android.content.res.Resources;
+import android.graphics.Matrix;
+import android.graphics.RectF;
 import android.util.SparseArray;
 import android.util.SparseBooleanArray;
 import android.util.SparseIntArray;
@@ -74,6 +76,7 @@ public class NativeViewHierarchyManager {
   private final LayoutAnimationController mLayoutAnimator = new LayoutAnimationController();
   private final SparseArray<SparseIntArray> mTagsToPendingIndicesToDelete = new SparseArray<>();
   private final int[] mDroppedViewArray = new int[100];
+  private final RectF mBoundingBox = new RectF();
 
   private boolean mLayoutAnimationEnabled;
   private PopupMenu mPopupMenu;
@@ -649,16 +652,47 @@ public class NativeViewHierarchyManager {
     if (rootView == null) {
       throw new NoSuchNativeViewException("Native view " + tag + " is no longer on screen");
     }
-    rootView.getLocationInWindow(outputBuffer);
+    computeBoundingBox(rootView, outputBuffer);
     int rootX = outputBuffer[0];
     int rootY = outputBuffer[1];
+    computeBoundingBox(v, outputBuffer);
+    outputBuffer[0] -= rootX;
+    outputBuffer[1] -= rootY;
+  }
 
-    v.getLocationInWindow(outputBuffer);
+  void computeBoundingBox(View v, int[] outputBuffer) {
+    mBoundingBox.set(0, 0, v.getWidth(), v.getHeight());
+    mapRectFromViewToWindowCoords(v, mBoundingBox);
 
-    outputBuffer[0] = outputBuffer[0] - rootX;
-    outputBuffer[1] = outputBuffer[1] - rootY;
-    outputBuffer[2] = v.getWidth();
-    outputBuffer[3] = v.getHeight();
+    outputBuffer[0] = (int) mBoundingBox.left;
+    outputBuffer[1] = (int) mBoundingBox.top;
+    outputBuffer[2] = (int) (mBoundingBox.right - mBoundingBox.left);
+    outputBuffer[3] = (int) (mBoundingBox.bottom - mBoundingBox.top);
+  }
+
+  // simplified version of the hidden Android method View.mapRectFromViewToScreenCoords()
+  void mapRectFromViewToWindowCoords(View v, RectF rect) {
+    Matrix m = v.getMatrix();
+    if (!m.isIdentity()) {
+      m.mapRect(rect);
+    }
+
+    rect.offset(v.getLeft(), v.getTop());
+
+    ViewParent parent = v.getParent();
+    while (parent instanceof View) {
+      View parentView = (View) parent;
+
+      rect.offset(-parentView.getScrollX(), -parentView.getScrollY());
+      m = parentView.getMatrix();
+      if (!m.isIdentity()) {
+        m.mapRect(rect);
+      }
+
+      rect.offset(parentView.getLeft(), parentView.getTop());
+
+      parent = parentView.getParent();
+    }
   }
 
   /**


osdnk:
Thanks for this contribution! Are you aware that on this blue area on iOS does not mean that this area is touchable? 
This is how we virtualise touchable areas on iOS:
image
image

So clickable area is limited to bounds on component. Can you confirm that your code works like these?
Please consider using `touchables` button in an inspector while attaching screenshot :)

floriancargoet:
Yes, I know the blue rectangle represents the bounds, but it's the bounds that are used to detect the LEAVE_PRESS_RECT event. 

1. You press in the touchable area, pressIn is triggered.
2. You move a bit. Since the bounds are incorrect, a LEAVE_PRESS_RECT is triggered.
3. You release your touch. Since you are considered out of the "press rect", pressOut in NOT triggered.

You can see that the bounds are used for step 2 : 
 - [`_remeasureMetricsOnActivation()`](https://github.com/facebook/react-native/blob/master/Libraries/Components/Touchable/Touchable.js#L848) is called
 - [`UIManager.measure()`](https://github.com/facebook/react-native/blob/master/Libraries/Components/Touchable/Touchable.js#L685) is called, the results are store in `positionOnActivate` and `dimensionsOnActivate` which are used to compute [`isTouchWithinActive` ](https://github.com/facebook/react-native/blob/master/Libraries/Components/Touchable/Touchable.js#L551) and to [trigger `LEAVE_PRESS_RECT`](https://github.com/facebook/react-native/blob/master/Libraries/Components/Touchable/Touchable.js#L575).

Does this clarify things?

Sorry if this wasn't included in the PR description but when I investigated this issue a year ago the `Touchable.js` code was very different and I haven't a good understanding of this new code.

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
```diff
@@ -649,16 +652,47 @@ public synchronized void measure(int tag, int[] outputBuffer) {
...
+      rect.offset(-parentView.getScrollX(), -parentView.getScrollY());
+      m = parentView.getMatrix();
+      if (!m.isIdentity()) {
+        m.mapRect(rect);
+      }
+
+      rect.offset(parentView.getLeft(), parentView.getTop());
+
+      parent = parentView.getParent();
+    }
```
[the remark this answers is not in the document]
floriancargoet: I've commented the original Android code, explaining each change. 
Most of the changes are about replacing private access with a public method.
The most important comment is the last one.

```java
public void mapRectFromViewToScreenCoords(RectF rect, boolean clipToParent) {
    // hasIdentityMatrix is not public, instead I used getMatrix().isIdentity()
    if (!hasIdentityMatrix()) {
        getMatrix().mapRect(rect);
    }
    // mLeft & mTop are not public, instead I used getLeft() & getTop()
    rect.offset(mLeft, mTop);

    ViewParent parent = mParent;
    while (parent instanceof View) {
        View parentView = (View) parent;
        
        // mScrollX & mScrollY are not public, instead I used getScrollX() & getScrollY()
        rect.offset(-parentView.mScrollX, -parentView.mScrollY);
        
        // We don't need clipping to measure, so I removed this block entirely
        if (clipToParent) {
            rect.left = Math.max(rect.left, 0);
            rect.top = Math.max(rect.top, 0);
            rect.right = Math.min(rect.right, parentView.getWidth());
            rect.bottom = Math.min(rect.bottom, parentView.getHeight());
        }

        // hasIdentityMatrix is not public, instead I used getMatrix().isIdentity()
        if (!parentView.hasIdentityMatrix()) {
            parentView.getMatrix().mapRect(rect);
        }

        // mLeft & mTop are not public, instead I used getLeft() & getTop()
        rect.offset(parentView.mLeft, parentView.mTop);

        // mParent is not public, instead I used getParent()
        parent = parentView.mParent;
    }
    // I think the 2 following offsets would be the same for the measured view and 
    // the React root view. Since we subtract the root view position from the 
    // measured view position in measure(), these offsets are cancelled.
    // In short: (a + offsets) - (b + offsets) = a - b
    // 
    // class ViewRootImpl is hidden, I could not find a way to rewrite this code.
    // Since we don't need it for our purpose, I removed it.
    if (parent instanceof ViewRootImpl) {
        ViewRootImpl viewRootImpl = (ViewRootImpl) parent;
        rect.offset(0, -viewRootImpl.mCurScrollY);
    }
    rect.offset(mAttachInfo.mWindowLeft, mAttachInfo.mWindowTop);
}
```

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
```diff
@@ -649,16 +652,47 @@ public synchronized void measure(int tag, int[] outputBuffer) {
...
+    computeBoundingBox(rootView, outputBuffer);
     int rootX = outputBuffer[0];
     int rootY = outputBuffer[1];
+    computeBoundingBox(v, outputBuffer);
+    outputBuffer[0] -= rootX;
+    outputBuffer[1] -= rootY;
+  }
 
-    v.getLocationInWindow(outputBuffer);
+  void computeBoundingBox(View v, int[] outputBuffer) {
```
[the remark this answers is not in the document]
floriancargoet: OK, I guess this comment is valid for the two added methods, not just this one.

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
```diff
@@ -649,16 +652,47 @@ public synchronized void measure(int tag, int[] outputBuffer) {
...
+    mBoundingBox.set(0, 0, v.getWidth(), v.getHeight());
+    mapRectFromViewToWindowCoords(v, mBoundingBox);
 
-    outputBuffer[0] = outputBuffer[0] - rootX;
-    outputBuffer[1] = outputBuffer[1] - rootY;
-    outputBuffer[2] = v.getWidth();
-    outputBuffer[3] = v.getHeight();
+    outputBuffer[0] = (int) mBoundingBox.left;
```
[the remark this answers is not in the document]
floriancargoet: I can't see a Math.round() in `View.mapRectFromViewToScreenCoords` but I've updated the PR.

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
```diff
@@ -649,16 +652,47 @@ public synchronized void measure(int tag, int[] outputBuffer) {
     if (rootView == null) {
       throw new NoSuchNativeViewException("Native view " + tag + " is no longer on screen");
     }
-    rootView.getLocationInWindow(outputBuffer);
+    computeBoundingBox(rootView, outputBuffer);
```
makovkastar: Here we could still use `rootView.getLocationInWindow(outputBuffer)`, but I assume you used `computeBoundingBox` for consistency? We don't need the width and the height (only x and y) of the root view and it's not supposed to be transformed, so `getLocationInWindow` should still be sufficient.

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
```diff
@@ -649,16 +652,47 @@ public synchronized void measure(int tag, int[] outputBuffer) {
...
-    outputBuffer[1] = outputBuffer[1] - rootY;
-    outputBuffer[2] = v.getWidth();
-    outputBuffer[3] = v.getHeight();
+    outputBuffer[0] = Math.round(mBoundingBox.left);
+    outputBuffer[1] = Math.round(mBoundingBox.top);
+    outputBuffer[2] = Math.round(mBoundingBox.right - mBoundingBox.left);
+    outputBuffer[3] = Math.round(mBoundingBox.bottom - mBoundingBox.top);
+  }
+
+  // simplified version of the hidden Android method View.mapRectFromViewToScreenCoords()
```
makovkastar: Could you replace this comment with the proper Javadoc (perhaps borrow it from `View.mapRectFromViewToScreenCoords`) and add the `@see` annotation referring to that method?

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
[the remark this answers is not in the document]
makovkastar: We definitely don't need `rect.offset(mAttachInfo.mWindowLeft, mAttachInfo.mWindowTop);` here since it actually offsets the coordinates relative to the screen coordinates, which isn't what we want (we want the coordinates relative to the window coordinates).

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
floriancargoet: Yes consistency and cancellation of the missing offsets in my version of mapRectFromViewToScreenCoords. Have you read my annotations on the original code in response to mdvacca? By using computeBoundingBox twice, offsets cancel each other and I don't need to port this part of the function (hard to do since everything is private).

```js
    // I think the 2 following offsets would be the same for the measured view and 
    // the React root view. Since we subtract the root view position from the 
    // measured view position in measure(), these offsets are cancelled.
    // In short: (a + offsets) - (b + offsets) = a - b
    // 
    // class ViewRootImpl is hidden, I could not find a way to rewrite this code.
    // Since we don't need it for our purpose, I removed it.
    if (parent instanceof ViewRootImpl) {
        ViewRootImpl viewRootImpl = (ViewRootImpl) parent;
        rect.offset(0, -viewRootImpl.mCurScrollY);
    }
    rect.offset(mAttachInfo.mWindowLeft, mAttachInfo.mWindowTop);
```

makovkastar: Thank you for the explanation. I think I got the idea now: if we call `rootView.getLocationInWindow(outputBuffer)` instead of `computeBoundingBox(rootView, outputBuffer)`, `getLocationInWindow` internally will calculate the offset for the rootView:

```
if (parent instanceof ViewRootImpl) {
    ViewRootImpl viewRootImpl = (ViewRootImpl) parent;
    rect.offset(0, -viewRootImpl.mCurScrollY);
}
```

And if mCurScrollY != 0 for our root view, the bounding box rect of the child view will be missing this offset since we don't have access to the private API of ViewRootImpl to calculate it accordingly.

makovkastar:
@floriancargoet could you please address my comment related to the missing Javadoc and we should be good to merge this PR. @mdvacca what do you think?

ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyManager.java
floriancargoet: @makovkastar Done.
However the @link in the @see annotation cannot resolve the hidden method. Is that an issue?
image

makovkastar: The Javadoc is actually misleading since the method doesn't map coordinates relative to the root view, but rather to the window. Nevermind, I just realised that this method is private and the name is actually quite self-descriptive, so I will remove it and merge the PR now.

PR #25839, opened 2019-07-26
Use Metro support for auto-collapsing internal stack frames

motiz88:
Summary:
Changes `ExceptionsManager` to respect the `collapse` field in each symbolicated stack frame returned from Metro, in preference to the client-side regex. This is ultimately driven by a Metro config option (now also set in the RN new project template). Note that the client-side regex is intentionally ignored for any frame where `collapse` is present ( = all of them if using Metro with https://github.com/facebook/metro/pull/435).

Also updates the client-side regex to match the new renderer paths.

This is part of a redesign of work done originally in https://github.com/facebook/react-native/pull/24662; I intend to eventually remove the client-side regex from RN.

Reviewed By: cpojer

Differential Revision: D16500277

### DIFF
--- a/Libraries/Core/ExceptionsManager.js
+++ b/Libraries/Core/ExceptionsManager.js
@@ -12,13 +12,6 @@
 
 import type {ExtendedError} from './Devtools/parseErrorStack';
 
-const INTERNAL_CALLSITES_REGEX = new RegExp(
-  [
-    '/Libraries/Renderer/oss/ReactNativeRenderer-dev\\.js$',
-    '/Libraries/BatchedBridge/MessageQueue\\.js$',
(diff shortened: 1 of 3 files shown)


template/metro.config.js
```diff
@@ -5,7 +5,22 @@
  * @format
  */
 
+const INTERNAL_CALLSITES_REGEX = new RegExp(
```
TheSavior: Should we be hiding this by default but let people override it here?

motiz88: Yes. There was some discussion on the RN Discord but I didn't get around to making the changes today. The basic idea is to put the defaults in the CLI, where there's precedent for this kind of thing.

Issue #25319, opened 2019-06-19
ActivityIndicator breaks when adding a testID


Twinbird24:
This is the exact same issue as reported here by another user: https://github.com/facebook/react-native/issues/21563.
The previous ticket was not resolved and was closed, so I am opening a new one to address the issue. I am using the latest React Native release, `0.59.9`.

```
 React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
      Memory: 26.00 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 25, 26, 27, 28
        Build Tools: 23.0.1, 26.0.2, 27.0.3, 28.0.3
        System Images: android-23 | Intel x86 Atom_64, android-23 | Google APIs Intel x86 Atom_64, android-25 | Android TV Intel x86 Atom, android-25 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom_64, android-25 | Google Play Intel x86 Atom, android-26 | Android TV Intel x86 Atom, android-26 | Google APIs Intel x86 Atom_64, android-26 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: ^16.8.6 => 16.8.6
      react-native: ^0.59.9 => 0.59.9
```

## Steps To Reproduce

1) Import `ActivityIndicator` anywhere into your project and have it render, i.e.
`<ActivityIndicator color="blue" />`
2) You'll notice the color of the indicator is blue. Now, add a testID, i.e.
`<ActivityIndicator color="blue" testID="activity-indicator" />`
You'll notice that the color changes to white (appears to disappear on a white background) and you cannot change the color to anything else.

[...]

PR #25849, opened 2019-07-27
[iOS][Fix] Do not override ActivityIndicator color when setting its size

cabelitos:
## Summary

The activityIndicatorViewStyle property overrides the previous set color
if it's changed. Depending on the property set order you may end in a state
that the color property will never be respected since it first sets
the color and then the activityIndicatorViewStyle property (which overrides
the color property). In order to prevent this problem
before setting the new activityIndicatorViewStyle save the old
color and override it after activityIndicatorViewStyle is set. Thus
always respecting the user's color.

## Changelog

[iOS] [Fixed] - Do not override ActivityIndicator color when setting its size

## Test Plan
Using the code below on iOS notice that the last ActivityIndicator will always have its color set to white while te testID is provided

### Without the patch
Notice the white -> blue transition when disabling the testID

broken

### With the patch
Color remains unchanged 

working

```javascript
import React from "react";
import { View, StyleSheet, ActivityIndicator, Button } from "react-native";

const App = () => {
  const [enableTestID, onSetEnableTestID] = React.useState(true);
  const onPress = React.useCallback(() => {
    onSetEnableTestID(!enableTestID);
  }, [enableTestID]);
  return (
      <ActivityIndicator size="large" color="red" />
      <ActivityIndicator size="small" color="red" />
      <ActivityIndicator size="small" />
      <ActivityIndicator color="green" />
      <ActivityIndicator
        key={enableTestID.toString()}
        size="large"
        color="blue"
        testID={enableTestID ? 'please work' : undefined}
      />
      <Button
        title={enableTestID ? 'Disable testID' : 'enable testID'}
        onPress={onPress}
      />
    </View>
  );
};

export default App;

  container: {
    flex: 1,
    alignItems: "center",
    justifyContent: "center",
    backgroundColor: "black"
  },
});

```

Closes https://github.com/facebook/react-native/issues/25319

### DIFF
--- a/React/Views/RCTActivityIndicatorViewManager.m
+++ b/React/Views/RCTActivityIndicatorViewManager.m
@@ -33,7 +33,17 @@ RCT_EXPORT_MODULE()
 
 RCT_EXPORT_VIEW_PROPERTY(color, UIColor)
 RCT_EXPORT_VIEW_PROPERTY(hidesWhenStopped, BOOL)
-RCT_REMAP_VIEW_PROPERTY(size, activityIndicatorViewStyle, UIActivityIndicatorViewStyle)
+RCT_CUSTOM_VIEW_PROPERTY(size, UIActivityIndicatorViewStyle, UIActivityIndicatorView)
+{
+  /*
+    Setting activityIndicatorViewStyle overrides the color, so restore the original color
+    after setting the indicator style.
+  */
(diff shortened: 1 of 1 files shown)


sammy-SC:
@cabelitos thank you for fixing this!

cabelitos:
@sammy-SC thank you very much for your approval! glad to help;

PR #25853, opened 2019-07-29
Explain @format and @flow to freshmen

sunnylqm:
Related https://github.com/facebook/react-native-website/issues/1065


cpojer:
Thank you for the PR! I think instead of adding this to the template we should probably explain this in the documentation.

PR #25872, opened 2019-07-30
Add missing hermes include

janicduplessis:
## Summary

The Hermes react executor c++ code is not currently getting built. The include was probably forgotten when merging hermes on master. The 0.60 branch does have it https://github.com/facebook/react-native/blob/0.60-stable/ReactAndroid/src/main/jni/react/jni/Android.mk#L72

## Changelog

[Android] [Fixed] - Add missing hermes include

## Test Plan

Run an app with hermes enabled on RN master


janicduplessis:
cc @cpojer

Issue #25664, opened 2019-07-15
Memory issues after upgrading to version 0.59.2


bhandarijiwan:
After upgrading to react native 0.59.2 from 0.56 we noticed big memory issues in iOS.  I don't  have a reproducible scenario yet that I can post right now, I am working on creating one. Using git bisect I was able to  track it down to [this commit](https://github.com/facebook/react-native/commit/a23fb1a22c58b20fdbe7c8ff11f6eab70288d465).   Unfortunately that's all the info I have at the minute I just wanted to post this to see if others have seen noticed anything similar. 

React Native version:

React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 38.09 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 8.12.0 - ~/.nvm/versions/node/v8.12.0/bin/node
      Yarn: 1.7.0 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v8.12.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.3
        System Images: android-19 | Intel x86 Atom, android-19 | Google APIs Intel x86 Atom, android-23 | Intel x86 Atom, android-23 | Intel x86 Atom_64, android-23 | Google APIs ARM EABI v7a, android-23 | Google APIs Intel x86 Atom, android-23 | Google APIs Intel x86 Atom_64, android-25 | Google APIs ARM EABI v7a, android-25 | Google APIs Intel x86 Atom, android-26 | Google APIs Intel x86 Atom, android-28 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.2 => 0.59.2 
[...]

PR #25884, opened 2019-07-31
Memory Leak due to JSStringRelease not called in multiple places in JSCRuntime.cpp

sachinservicemax:
## Summary
Memory Leak due to JSStringRelease not called in multiple places in JSCRuntime.cpp
Issue: https://github.com/facebook/react-native/issues/25664
Reproducible repo: https://github.com/bhandarijiwan/memory_issue_repro

## Changelog
[JSC] [JSCRuntime.cpp] - Added missing JSStringRelease calls in missing places

## Test Plan
Tested that is no memory leak with various NativeModule to JS call flows

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -630,7 +630,9 @@ jsi::String JSCRuntime::createStringFromUtf8(
     size_t length) {
   std::string tmp(reinterpret_cast<const char*>(str), length);
   JSStringRef stringRef = JSStringCreateWithUTF8CString(tmp.c_str());
-  return createString(stringRef);
+  auto result = createString(stringRef);
+  JSStringRelease(stringRef);
+  return result;
 }
 
 std::string JSCRuntime::utf8(const jsi::String& str) {


ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -605,11 +605,19 @@ jsi::PropNameID JSCRuntime::createPropNameIDFromString(const jsi::String& str) {
 }
 
 std::string JSCRuntime::utf8(const jsi::PropNameID& sym) {
-  return JSStringToSTLString(stringRef(sym));
+  JSStringRef strRef = stringRef(sym);
+  std::string result = JSStringToSTLString(strRef);
+  JSStringRelease(strRef);
```
cabelitos: I'm not really sure if all these string releases are correct. since stringRef() does not create a new ref. It simple returns a pointer to the string ref.

sachinservicemax: I agree, however without these JSStringRef releases, we see huge memory leak in our application, we are able to demonstrate similar behavior in the sample application. 
Secondly, there are places in the JSCRuntime.cpp which are performing JSStringRelease for similar kind pointers e.g. createValue method.

cabelitos: The other JSStringRelease() usages in this files matches a String creation like this:
```cpp
jsi::PropNameID JSCRuntime::createPropNameIDFromAscii(
    const char* str,
    size_t length) {
  std::string tmp(str, length);
  JSStringRef strRef = JSStringCreateWithUTF8CString(tmp.c_str()); // Here a new string ref is created
  auto res = createPropNameID(strRef);
  JSStringRelease(strRef); // then released.
  return res;

```

cabelitos: I think that the problem lays in another somewhere else and we're "fixing" it in the wrong place.

sachinservicemax: You are right, I am not fully aware of the flow, so we made this fix based on the analysis of memory profiling, putting these fixes solved memory leaks we were seeing in our application.

sachinservicemax:
Attaching  the ios memory profile trace generated from instruments
[ios-memory-profile.trace.zip](https://github.com/facebook/react-native/files/3459127/ios-memory-profile.trace.zip)

ReactCommon/jsi/JSCRuntime.cpp
sachinservicemax: Attaching  the ios memory profile trace generated from instruments
[ios-memory-profile.trace.zip](https://github.com/facebook/react-native/files/3459127/ios-memory-profile.trace.zip)

TheSavior:
@sachinservicemax @cabelitos Is this ready for review? From reading the conversation it seems like you two aren't sure yet which approach to take of where to put JSStringRelease. If you two are happy with an approach then we can take a closer look and try to get this in. I think this confusion is why you haven't heard much from us on this PR yet. Sorry about that.

sachinservicemax:
> @sachinservicemax @cabelitos Is this ready for review? From reading the conversation it seems like you two aren't sure yet which approach to take of where to put JSStringRelease. If you two are happy with an approach then we can take a closer look and try to get this in. I think this confusion is why you haven't heard much from us on this PR yet. Sorry about that.

I have updated the PR with the latest changes, after thorough testing internally, I have made only one change to clear the JSStringRef in createStringFromUtf8. Releasing of the JSStringRef in this method seems to be missed and adding the release helps in fixing out memory leak issue.

TheSavior:
Thanks for the PR! Sorry for the miscommunication that made this take so long to get to.

guohonghanusa:
@TheSavior thanks for taking care of this PR, and looking forward for 60.6 to include this fix. thanks @sachinservicemax and @bhandarijiwan! what a relief!!

PR #25909, opened 2019-08-01
Add proguard rules for Hermes

theweavrs:
## Summary
This adds Proguard rules for retaining the unicode classes of Hermes (if R8/Proguard is enabled). This basically fixes a `ClassNotFound` exception in release builds.

## Changelog

[Internal] [Added] - Added Proguard rules for Hermes

## Test Plan

Tested on RN 0.60.4 project with Hermes 0.1.1 enabled.

### DIFF
--- a/template/android/app/proguard-rules.pro
+++ b/template/android/app/proguard-rules.pro
@@ -8,3 +8,6 @@
 #   http://developer.android.com/guide/developing/tools/proguard.html
 
 # Add any project specific keep options here:
+
+# Hermes specific rules (uncomment only if you have hermes enabled):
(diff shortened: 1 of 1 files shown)


janicduplessis:
Hey @theweavrs ! Thanks for fixing this. We actually already have a fix for it in master by adding DoNotStrip annotations (https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/hermes/unicode/AndroidUnicodeUtils.java#L18). However this has not made it to the 0.60 release. I suggest opening a PR against the 0.60 branch to add the annotations from the file I linked.

theweavrs:
@janicduplessis will do.

PR #25912, opened 2019-08-01
Fix `ClassNotFound` exception in Android during Release builds

theweavrs:
## Summary

This adds the `@DoNotStrip` annotation to disallow Proguard/R8 from stripping the `AndroidUnicodeUtils` class during release build and causing a `ClassNotFound` crash.

## Changelog

[Internal] [Fixed] - Fixed `ClassNotFound` exception in Android during Release builds

## Test Plan


theweavrs:
@janicduplessis done

janicduplessis:
@mdvacca This actually needs to go in the 0.60 branch, I think we can just merge it directly on github. This is because we landed 2 different versions of the hermes commit in 0.60 and master.

mdvacca:
Thanks @janicduplessis, that's why I was having problems in the import.

PR #25919, opened 2019-08-02
[iOS] UIViewController-based status bar management

radex:
⚠️ This is a follow up to #25425 -- which isn't merged yet… See https://github.com/facebook/react-native/pull/25919/files/2a286257a6553a80a34e2b1f1ad94fc7bae36ea3..d3f63dd6d40cd2ece0726768c416d76938fbe023 for actual diff

## Summary

Currently, StatusBar native module manages the status bar on iOS globally, using `UIApplication.` APIs. This is bad because:

- those APIs have been deprecated for 4 years
- Apple really, really wants you to have an explicitly defined view controller, and control the status bar there
- it [breaks external native components](https://github.com/facebook/react-native/issues/25181#issuecomment-506792819)
- it's [not compatible with iPadOS 13 multi window support](https://github.com/facebook/react-native/issues/25181#issuecomment-506690818)

for those reasons I we should transition towards view controller-based status bar management.

With that, there is a need to introduce a default React Native root view controller, so I added `RCTRootViewController`. Using it is completely opt-in and there is no breaking change here. However I believe this should be a part of the template for new RN iOS apps.

Additionally, I added `RCTRootViewControllerProtocol` with hooks needed for RCTStatusBarManager to control the status bar. This means apps that want to have total control over their view controller can still opt in to react native VC-based status bar by conforming their root view controller to this protocol.

## Changelog

[iOS] [Added] - Added `RCTRootViewController` and `RCTRootViewControllerProtocol`
[iOS] [Fixed] - `UIViewControllerBasedStatusBarAppearance=YES` no longer triggers an error as long as you use `RCTRootViewController`
[iOS] [Fixed] - Status bar style is now correctly changed in multi-window iPadOS 13 apps if you use `RCTRootViewController` and set `UIViewControllerBasedStatusBarAppearance=YES`

## Test Plan

- Open RNTester → StatusBar → and check that no features broke


radex:
cc @hramos @shergin

radex:
also cc @steipete, re: https://github.com/facebook/react-native/issues/25181#issuecomment-506792819 — do you mind taking a look at this PR? You mentioned using `childViewControllerForStatusBarStyle`… Can you elaborate why this is necessary? Don't presented view controllers already control status bar by default if they are full screen?

React/Base/RCTRootViewController.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * The root view used by the view controller.
+ */
+@property (nonatomic, strong, readonly) RCTRootView *rootView;
+
+/**
+ * See: RCTRootViewControllerProtocol
+ */
+- (void)updateStatusBarStyle:(UIStatusBarStyle)style
```
steipete: Do we really need both `animation` and `animated`? With `UIStatusBarAnimation` we already can set it to `UIStatusBarAnimationNone`. Choosing fade or slide should automatically infer that we indeed animate the change.

steipete:
@radex Thank you for writing this PR! This looks great and will resolve a long-standing compatibility issue with PSPDFKit and React Native, next to many other libraries.

We use `childViewControllerForStatusBarStyle` only for special cases such as a container view controller hat hosts a child full-screen; I do not see this relevant for the implementation here.

ccorcos:
Any ETA on this? I'm trying to prepare my app for iOS 13!

hramos:
Both this, and the base PR it is based off (#25425), are getting closer to being merged. I've resolved all conflicts and addressed issues raised by our internal suite so far (mostly related to RCTPresentedViewController's new signature, and running the codegen as the modules touched here are now TurboModules). Hold tight.

radex:
@hramos thank you ❤️

adbl:
I see this was reverted in https://github.com/facebook/react-native/commit/f4538e577734d9448deec007a156905ae3c2527b @hramos @logandaniels could you explain what happened?

Issue #23465, opened 2019-02-14
iOS crash, only release from testflight or archive build


mtostenson:
## 🐛 Bug Report

Crash on launch. Here is the relevant log:
```
Thread 6 Crashed:
0   libsystem_kernel.dylib        	0x00000002388a5104 __pthread_kill + 8
1   libsystem_pthread.dylib       	0x0000000238924a00 pthread_kill$VARIANT$armv81 + 296
2   libsystem_c.dylib             	0x00000002387fcd78 abort + 140
3   libsystem_malloc.dylib        	0x00000002388f9768 _malloc_put + 0
4   libsystem_malloc.dylib        	0x00000002388f9924 malloc_report + 64
5   libsystem_malloc.dylib        	0x00000002388ec2d0 free + 376
6   libc++.1.dylib                	0x0000000237ea7120 std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::~basic_string+ 258336 () + 32
7   ExampleApp                 	0x00000001053cd494 std::__1::__vector_base<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >::~__vector_base() + 5264532 (vector:451)
8   ExampleApp                 	0x00000001059c8f04 facebook::react::ModuleRegistry::getConfig+ 11538180 (std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 100
9   ExampleApp                 	0x00000001059d7e58 facebook::react::JSCNativeModules::createModule+ 11599448 (std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, OpaqueJSContext const*) + 292
10  ExampleApp                 	0x00000001059d79b8 facebook::react::JSCNativeModules::getModule+ 11598264 (OpaqueJSContext const*, OpaqueJSString*) + 188
[...]

PR #25946, opened 2019-08-05
Fix launch crash on iOS when using together with `react-native-unity-view`

yuweipei:
## Summary
The issue was reported [here](https://github.com/facebook/react-native/issues/23465) and discussed. As @mtostenson found

> I have been able to narrow it down to this function: ModuleRegistry::moduleNames
When the vector that is returned from this function goes out of scope and the memory is freed, the app inexplicably crashes.

Adding a moduleNamesVec_ property to hold `moduleNames()` will address the issue by keeping a reference to the return results.
This will fix launch crash when using together with `react-native-unity-view`

## Changelog

[iOS] [Fixed] - Fix launch crash on iOS when integrated together with `react-native-unity-view`

### DIFF
--- a/ReactCommon/cxxreact/ModuleRegistry.cpp
+++ b/ReactCommon/cxxreact/ModuleRegistry.cpp
@@ -83,7 +83,7 @@ folly::Optional<ModuleConfig> ModuleRegistry::getConfig(const std::string& name)
 
   // Initialize modulesByName_
   if (modulesByName_.empty() && !modules_.empty()) {
-    moduleNames();
+    moduleNamesVec_ = moduleNames();
   }
 
   auto it = modulesByName_.find(name);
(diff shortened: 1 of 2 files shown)


ReactCommon/cxxreact/ModuleRegistry.cpp
```diff
@@ -83,7 +83,7 @@ folly::Optional<ModuleConfig> ModuleRegistry::getConfig(const std::string& name)
 
   // Initialize modulesByName_
   if (modulesByName_.empty() && !modules_.empty()) {
-    moduleNames();
+    moduleNamesVec_ = moduleNames();
```
shergin: That's truly terrible that app crashes on the start. :(

However I don't see how this change can make any difference.

Calling `moduleNames();` here in the first place is very strange thing; I think the original intention was to initialize/warm `modulesByName_` but that's very obscure behavior which needs to be refactored anyway.

@yuweipei  Why do you believe that retaining a vector of strings can fix anything?

cpojer:
@yuweipei could you reply to @shergin's comments?

sota000:
@yuweipei, I am going to close this PR since it's outdated, but please reopen if it's relevant. Thanks for the contribution!

PR #25969, opened 2019-08-07
fix display problems when image fails to load

jsfu:
## Summary

When the image resource is changed and the new image resource fails to load, we expect the display image to fail to load, but the image still shows the image that was successfully loaded last time.

## Changelog

[iOS] [Fixed] - fix display problems when image fails to load

## Test Plan
I found that the old image resource of ImageView was not cleaned up when the image failed to load, so I added the cleanup of the old resource in the logic of image loading failure.
```
- (void)imageLoaderLoadedImage:(UIImage *)loadedImage error:(NSError *)error forImageSource:(RCTImageSource *)source partial:(BOOL)isPartialLoad
{
  if (![source isEqual:_pendingImageSource]) {
    // Bail out if source has changed since we started loading
    return;
  }

  if (error) {
    if (_onError) {
      _onError(@{ @"error": error.localizedDescription });
    }
    if (_onLoadEnd) {
      _onLoadEnd(nil);
    }
    return;
  }
  ……
}
```

Before fix：
before
After fix:
after

### DIFF
--- a/Libraries/Image/RCTImageView.m
+++ b/Libraries/Image/RCTImageView.m
@@ -347,6 +347,14 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
   }
 
   if (error) {
+    if (_pendingImageSource.request && _imageSource.request && ![_pendingImageSource.request.URL.absoluteString isEqual: _imageSource.request.URL.absoluteString]) {
+        RCTExecuteOnMainQueue(^{
+            [self.layer removeAnimationForKey:@"contents"];
+            self.image = loadedImage;
+            self->_imageSource = source;
+            self->_pendingImageSource = nil;
+        });
+    }
     if (_onError) {
       _onError(@{ @"error": error.localizedDescription });
     }


Libraries/Image/RCTImageView.m
```diff
@@ -347,6 +347,14 @@ - (void)imageLoaderLoadedImage:(UIImage *)loadedImage error:(NSError *)error for
   }
 
   if (error) {
+    if (_pendingImageSource.request && _imageSource.request && ![_pendingImageSource.request.URL.absoluteString isEqual: _imageSource.request.URL.absoluteString]) {
```
sammy-SC: Could you explain me why this condition is necessary? I am having hard time understanding why it is needed.

Otherwise
You can get rid of `. absoluteString` as `NSURL` has `isEqual` defined (https://developer.apple.com/documentation/foundation/nsurl/1806686-isequal?language=objc).

Also I don't think we have to check whether `_pendingImageSource.request` and `_imageSource.request` are nils.

sammy-SC:
Thank you for taking your time to work on this!

Libraries/Image/RCTImageView.m
jsfu: I think the `Image` should really show whether the image is loaded successfully. If the loading fails and the URL is not equal to the previous one, then the previous image should not be displayed again, because the developer expects to display the new image, But the new image fails to load. . Usually developers can use `onError`'s callbacks to handle failures, but when the developer doesn't handle the error, the image displayed is actually wrong.

Also, thank you for your suggestion. I got a new skill  : )

PR #25989, opened 2019-08-08
Android: Allow touches on clipped children when `clipChildren==false`

danielgindi:
## Summary

Currently `overflow: visible` allows showing child views that are outside the parent's bounds (well, `overflow: visible` on the grandparent allows that, not the parent).
But those non-clipped children do not receive touches.

This is an attempt to solve it.

## Changelog

[Android] [Fixed] - `overflow: visible` now allows interaction with children that are rendered outside their parents' bounds.

## Test Plan

Added `overflow: visible`, tapped on a button.


danielgindi:
CI is broken. Tests pass and it still fails.

PR #25990, opened 2019-08-08
Fix forceUpdate method on useWindowDimensions

brunolemos:
## Summary

useState won't trigger re-renders if the value passed is the same.

## Changelog

[Internal] [Fixed] - Fix forceUpdate method on useWindowDimensions

## Test Plan

Codesandbox: https://codesandbox.io/embed/elegant-cori-0ixbx

### DIFF
--- a/Libraries/Utilities/useWindowDimensions.js
+++ b/Libraries/Utilities/useWindowDimensions.js
@@ -16,7 +16,7 @@ import * as React from 'react';
 
 export default function useWindowDimensions(): DisplayMetrics {
   const dims = Dimensions.get('window'); // always read the latest value
-  const forceUpdate = React.useState()[1];
+  const forceUpdate = React.useState(false)[1].bind(null, v => !v);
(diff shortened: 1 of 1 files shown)


Libraries/Utilities/useWindowDimensions.js
```diff
@@ -16,7 +16,7 @@ import * as React from 'react';
 
 export default function useWindowDimensions(): DisplayMetrics {
   const dims = Dimensions.get('window'); // always read the latest value
-  const forceUpdate = React.useState()[1];
+  const forceUpdate = React.useState()[1].bind(this, v => !v);
```
sahrens: Could bind `null` instead of `this`? Also probably want to pass `false` explicitly as the initial value for cleaner typing.

brunolemos:
I just noticed forceUpdate is at the useEffect dependencies, it shouldn’t be there. Might be causing #26008.

PR #26005, opened 2019-08-09
Consolidate OkHttp instances to only one instance

mrjoema:
## Summary

According to the official documentation of OkHttp (https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html), 

OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

## Changelog

[Android] [Fixed] - Consolidate OkHttp instances to only one instance for performance concerns

## Test Plan
Build the Android app with the change to make sure there's no regression related to any networking code

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/fresco/FrescoModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/fresco/FrescoModule.java
@@ -149,7 +149,7 @@ public class FrescoModule extends ReactContextBaseJavaModule implements
     HashSet<RequestListener> requestListeners = new HashSet<>();
     requestListeners.add(new SystraceRequestListener());
 
-    OkHttpClient client = OkHttpClientProvider.createClient();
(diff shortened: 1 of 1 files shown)


RubenSandwich:
Hello @mrjoema,

It looks like `OkHttpClientProvider.createClient` is used in the networking module as well:
https://github.com/facebook/react-native/blob/3b6f6ca4d5fcee6f1bc6d6242e3e2ef136e4d546/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java

Is there a reason why you didn't update those to use `getOkHttpClient`? Does the context passed in those instances of OkHttpClientProvider prevent the use of `getOkHttpClient`?

mrjoema:
Decide to close this PR as dev can define their OkHttp instance through the interface of OkHttpClientFactory

PR #26008, opened 2019-08-10
fix useWindowDimensions hook

dulmandakh:
## Summary

I found useWindowDimensions hook, and thought that it would solve many issues for me, but implementation looked interesting. So I created a new project, copy pasted the code and console.log dimentions returned from the hook, then rotated iOS simulator and it's printing dimentions non-stop.

Below is my fix for useWindowDimensions, and it works for me.

## Changelog

[GENERAL] [CHANGED] - fix useWindowDimensions hook

## Test Plan

Copy paste the code to RN project and console.log the return value, and rotate screen it'll log new dimensions.

### DIFF
--- a/Libraries/Utilities/useWindowDimensions.js
+++ b/Libraries/Utilities/useWindowDimensions.js
@@ -15,10 +15,10 @@ import {type DisplayMetrics} from './NativeDeviceInfo';
 import * as React from 'react';
 
 export default function useWindowDimensions(): DisplayMetrics {
-  const [dims, setDims] =  React.useState(Dimensions.get('window')); // set initial value
+  const [dims, setDims] = React.useState(Dimensions.get('window')); // set initial value
   React.useEffect(() => {
     function handleChange({window}) {
-      setDims(window)
+      setDims(window);
     }
     Dimensions.addEventListener('change', handleChange);
     return () => {
--- a/Libraries/Utilities/useWindowDimensions.js
+++ b/Libraries/Utilities/useWindowDimensions.js
@@ -17,7 +17,7 @@ import * as React from 'react';
 export default function useWindowDimensions(): DisplayMetrics {
   const [dims, setDims] =  React.useState(Dimensions.get('window')); // set initial value
   React.useEffect(() => {
-    function handleChange({ window}) {
+    function handleChange({window}) {
       setDims(window)
     }
     Dimensions.addEventListener('change', handleChange);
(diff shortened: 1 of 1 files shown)


dulmandakh:
please review @brunolemos

brunolemos:
Yes it looks more similar to what I would write myself. But the hook was written by @sahrens so I’ll let this to him.

dulmandakh:
I don't think that objects will be the same, even if dimensions are same.

sahrens:
Ah yes, the issue is that when we added the `bind` to `forceUpdate`, now the dependencies for the `useEffect` change on every render, which means it also does the comparison with initialDims, which means we always re-render if you change the dims once.

Your implementation here is the simple obvious one, but risks missing an event and never rendering with the correct dims. I'm pretty sure if we just pass an updater function when calling forceUpdate rather than binding it on every render, everything should work.

sahrens:
I'm going to import this but with some changes to handle lost updates.

dulmandakh:
I'm not getting your concerns about missing events, thus re-renders. This
hook sets initial value and handles events as they occur, and components
will re-render itself if value changes.

Libraries/Utilities/useWindowDimensions.js
```diff
@@ -15,21 +15,15 @@ import {type DisplayMetrics} from './NativeDeviceInfo';
...
 export default function useWindowDimensions(): DisplayMetrics {
-  const dims = Dimensions.get('window'); // always read the latest value
-  const forceUpdate = React.useState(false)[1].bind(null, v => !v);
-  const initialDims = React.useState(dims)[0];
+  const [dims, setDims] = React.useState(Dimensions.get('window')); // set initial value
   React.useEffect(() => {
-    Dimensions.addEventListener('change', forceUpdate);
-
-    const latestDims = Dimensions.get('window');
-    if (latestDims !== initialDims) {
```
aleclarson: You can't remove this check. Otherwise, there will be a chance of missing a change that occurs between render and effect.

ScreamZ:
@grabbou Can you confirm that mate? :) Thanks

jfrolich:
I also ran into this. Would be great if it would be included in the next react-native release.

Naturalclar:
@jfrolich This change is indeed included in the 0.62 stable branch, so it should be included :)
https://github.com/facebook/react-native/blob/0.62-stable/Libraries/Utilities/useWindowDimensions.js

Issue #26006, opened 2019-08-10
URL with base is always used, even when absolute URL is provided


phallguy:
MDN [documentation for URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) states that the `base` should only be used when `url` is a relative url. However the [current implementation of URL](https://github.com/facebook/react-native/blob/0ccedf3964b1ebff43e4631d1e60b3e733096e56/Libraries/Blob/URL.js#L131) (as of 0.60.4) in react-native _always_ prepends the base if given.

React Native version:

System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 447.55 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.6.0 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.9.0 => 16.9.0
    react-native: 0.60.4 => 0.60.4
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce

```javascript
const url = new URL("http://github.com", "http://google.com")
console.log(url.toString()) // => "http://google.comhttp://github.com"
```

Describe what you expected to happen:

```javascript
const url = new URL("http://github.com", "http://google.com")
console.log(url.toString()) // => "http://github.com"
```

Snack, code example, screenshot, or link to a repository:

https://snack.expo.io/@phallguy/url-bug-report

Notice that if you view the snack with the "Web" preview, it is handled correctly.

phallguy:
Related to https://github.com/facebook/react-native/issues/16434

jeswinsimon:
Would [this](https://github.com/jeswinsimon/react-native/blob/update_url_impl/Libraries/Blob/URL.js) be a valid fix for this?

phallguy:
Yes that looks like it would probably do the trick.

jeswinsimon:
[...]

PR #26009, opened 2019-08-10
URL: Do not prepend baseUrl if the URL is not a relative URL

jeswinsimon:
## Summary

Fix for bug #26006 URL with base is always used, even when absolute URL is provided

## Changelog

[Javascript] [Fixed] - `URL`: Base url value is ignored when the input url is not a relative url.

## Test Plan

`new URL('http://github.com', 'http://google.com')` now returns `http://github.com/`
Added a test case to `URL-test.js` to verify the same.

### DIFF
--- a/Libraries/Blob/URL.js
+++ b/Libraries/Blob/URL.js
@@ -130,13 +130,12 @@ export class URL {
 
   constructor(url: string, base: string) {
     let baseUrl = null;
-    if(!base || validateBaseUrl(url)) {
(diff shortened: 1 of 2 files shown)


jeswinsimon:
@cpojer Done.

mattfysh:
Did this break something? I'm trying `new URL('/a', 'http://xy.com/b/c/d/e').toString()` and getting the wrong output

Currently getting: `http://xy.com/b/c/d/e/a`

Should be getting: `http://xy.com/a`

PR #26053, opened 2019-08-13
Use yoga from cocoapods instead of local podspec

janicduplessis:
## Summary

I think yoga is stable enough to be used from cocoapods instead of relying on the internal vendored version. This only affects install with cocoapods.

The main motivation behind this is better interop with other libraries that use yoga also from cocoapods (in my case Flipper). Right now it causes an error because of multiple copies of libyoga.a. By doing this they will share the same copy of yoga.

## Changelog

[iOS] [Changed] - Use yoga from cocoapods instead of local podspec

## Test Plan

Tested locally in an app and made sure it builds still.
Tested in RNTester


dulmandakh:
I find that Yoga releases lag behind the master, while React Native use Yoga from the master. I'm afraid that there might be breakages due to the mismatch. We had a case where Buck release delay broke Android CI.

grabbou:
Hm, @janicduplessis, does it mean we can't debug the layout with Flipper right now, without this PR?

janicduplessis:
@grabbou Yea, we can't include the layout plugin because it depends on Yoga from pods and conflicts with the one from RN.

ajanuar:
Waiting for this PR to be merged, I have case when we want to use `YogaKit` in our native page, there is conflict between Yoga in ReactCommon and Yoga as dependency of YogaKit.

dulmandakh:
I copied YogaKit files into the project and used it for my project. IMHO, no need to wait for this to be merged.

axemclion:
@janicduplessis  - I spoke about this internally, and with the Flipper people. I think the biggest risk of moving Yoga out of React Native would be that we would need to ensure that Yoga and React Native are released in lock-step. 

Instead of moving Yoga out, I have a PR, where i change the Yoga version in the yoga.podspec file to match that of the real yoga. This should solve the flipper issue, since both pods will point to the same version. IMHO, doing this will be a smaller change. What do you think ?

janicduplessis:
I'm not sure how cocoapods resolve dependencies but wouldn't that be dangerous since the implementation can diverge for the same version? Are we sure it would resolve to the local podspec included in RN?

I'm fine with landing some compromise if it unblocks flipper. Ideally we'd have regular yoga releases whenever important fixes or improvements are made and the version is bumped in RN at the same time, like we do with metro.

axemclion:
Here are the diffs I am talking about 
- https://github.com/facebook/react-native/commit/82a8080f0704e83079d0429e4e367f5131052e64
- https://github.com/facebook/react-native/commit/a58dd9683da6135bca119a8b0ae5d0ea5382cf4c

Can you apply these changes, and see if Flipper is able to work ?

janicduplessis:
@axemclion I'm getting this error

```
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `YogaKit` depends upon `Yoga`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

Adding 

```
spec.pod_target_xcconfig = {
      'DEFINES_MODULE' => 'YES'
  }
```

in the yoga podspec fixes the issue (taken this from the one in the yoga repo)

janicduplessis:
This PR is no longer needed

PR #26054, opened 2019-08-13
Add documentation to TextInput's Flow types

empyrical:
## Summary

The documentation from the Flow types' respective proptypes have been copied over to `TextInput`.

## Changelog

[Internal] [Changed] - Added documentation to TextInput's Flow types

## Test Plan

`yarn flow-check-ios` and `yarn flow-check-android` both pass.


TheSavior:
I think I actually feel fine leaving these in the original TextInput.js file

empyrical:
I mainly did that because it's how some of the other components converted over to classes were laid out - but if that's not important, I can change this pull request to just add the docstrings to the flow types

TheSavior:
Yeah, let's do that. We might need to move them later as we create spec files for these components, but I'm not sure we know what that will look like for TextInput yet, so keeping it in the one file is probably good. 👍

empyrical:
Done :+1:

PR #26064, opened 2019-08-14
Remove Hermes React Executor from BUCK file.

cpojer:
## Summary

Currently CI is broken because of a missing BUCK file. We don't actually need to build these things with buck right now so I'm just gonna remove it until somebody figures out how to build the proper BUCK file.

## Test Plan

* buck test works again


janicduplessis:
@cpojer Looks like the build is still failing because of a missing jni rule in `ReactAndroid/src/main/java/com/facebook:yoga`

empyrical:
If it's still needed internally, then an OSS-only conditional like this could be used:

https://github.com/facebook/react-native/commit/cb87c3f4100c1f1dde2ad2b4ada522abe8d38547#diff-eb9398d8c631ce0cd12c55e5121870bf

cpojer:
@mhorowitz has a proper fix, closing this PR.

PR #26079, opened 2019-08-16
Update Gradle wrapper to 5.6

friederbluemle:
## Summary

```
Welcome to Gradle 5.6!

Here are the highlights of this release:
 - Incremental Groovy compilation
 - Groovy compile avoidance
 - Test fixtures for Java projects
 - Manage plugin versions via settings script

For more details see https://docs.gradle.org/5.6/release-notes.html
```

## Changelog

[Android] [Changed] - Gradle wrapper 5.6

## Test Plan

Ran build and tests locally.

### DIFF
--- a/template/android/gradle.properties
+++ b/template/android/gradle.properties
@@ -17,5 +17,9 @@
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app's APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
 android.useAndroidX=true
+# Automatically convert third-party libraries to use AndroidX
 android.enableJetifier=true
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/gradlew
+++ b/gradlew
@@ -7,7 +7,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#      https://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -5,7 +5,7 @@
 @rem you may not use this file except in compliance with the License.
 @rem You may obtain a copy of the License at
 @rem
-@rem      http://www.apache.org/licenses/LICENSE-2.0
+@rem      https://www.apache.org/licenses/LICENSE-2.0
 @rem
 @rem Unless required by applicable law or agreed to in writing, software
 @rem distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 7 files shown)


.appveyor/config.yml
```diff
@@ -4,7 +4,7 @@ environment:
   ANDROID_BUILD_VERSION: 28
   ANDROID_TOOLS_VERSION: 28.0.3
 
-  GRADLE_OPTS: -Dorg.gradle.daemon=false
```
gengjiawen: I don't see this necessary, can you explain why ?

template/android/gradle.properties
```diff
@@ -9,13 +9,16 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
```
gengjiawen: same here

.appveyor/config.yml
friederbluemle: Hi @gengjiawen - Thanks for taking time to review the PR!
Manually increasing the heap space was necessary because of the way the build is run on AppVeyor. There is no `gradle.properties` file in the root directory of the repo (even though it contains the Gradle wrapper scripts and Gradle build files). You can see it _directly_ runs the `assembleRelease` task in `RNTester/android` - It might not immediately be obvious, but when called that way it will *not* pick up the Gradle settings from that directory, and instead use the [default heap space of only 64MB](https://github.com/facebook/react-native/blob/master/gradlew.bat#L33)! Previously it failed because it was running out of heap space. However, it looks like the AppVeyor CI build was disabled in 382cb05ae6ccf96355d297bcf54055305c2a7bac, so this line modification might not be needed anymore (at least for now). Let me try to remove it.

template/android/gradle.properties
friederbluemle: Actually, the update to this file has been missing for a very long time. The **default** Android project templates as generated by Android Studio include the setting of the jvmargs to `-Xmx1536m`. Running with lower heap space might degrade performance or fail the build. You can check by just generating any new project in Android Studio and open the `gradle.properties` file.

.appveyor/config.yml
gengjiawen: It's removed because android failed to run due to low memory. We intend to re-enable it later.
So maybe just leave it be.

template/android/gradle.properties
gengjiawen: Those default is fit for project like react native, it's likely you can't run on machine memory like than 8GB.

dulmandakh:
@friederbluemle could you please revert some changes @gengjiawen requested?

friederbluemle:
@dulmandakh - Thanks for your comment. I reverted the line. Do you also want me to undo the change in the .appveyor config file?

dulmandakh:
@friederbluemle usually it's best to change one thing in a PR, or version bump in this case. So please revert changes that are not related to version change.

friederbluemle:
Okay, I also reverted the change to `.appveyor/config.yml`.

friederbluemle:
FYI, the changes to the wrapper scripts were done _automatically_ by the wrapper task (of a newer Gradle version) - I did not manually modify these files. Since they are directly related to the Gradle wrapper, I think it makes sense to include them here too?

dulmandakh:
how about template/android/gradle.properties?

friederbluemle:
These are just a few comments, in a file directly related to the Gradle wrapper, to bring it in sync with the default template. I agree they _could_ be independent from the version update. Having a separate PR seemed a bit excessive to me though. Of course, if you prefer, I will exclude them here as well.

dulmandakh:
@cpojer @hramos  please merge 👍

friederbluemle:
Thanks @dulmandakh @gengjiawen

dulmandakh:
@friederbluemle you don't need to sync to master unless there is a conflict. thanks, let's wait for reviewers to merge the change.

friederbluemle:
@dulmandakh - Okay sure, I won't push the branch anymore. I just like to keep branches fully rebased to make sure everything works with the latest base branch. Related to that - Do you happen to know when the CI will be fully working again? I assume work is already being done?

friederbluemle:
@grabbou - Any reason this was reverted (6830ec87ff) on the `0.61-stable` branch, after being cherry-picked (afd35296f2)?

dulmandakh:
it because I forgot to mention that https://github.com/facebook/react-native/commit/7c8e266e390de71ce39233b53a369b2c5e8a1c14#diff-4a7ba4beeeab016621279eed94a6270b needs to be cherry-picked too

friederbluemle:
No, this is about the Gradle wrapper update.. There is no cross dependency on the Android Gradle plugin update. Those are two separate things. Both `3.4.1` and `3.5.0` work fine with Gradle `5.6`.

PR #26102, opened 2019-08-19
Prevent usage of Array index in keys

isaaclem:
To ensure source code conform with no-array-index-key rules: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

## Changelog

[General] [Fixed] - To avoid using index as key


isaaclem:
For the changes I've made, it simply doesn't makes any sense for failing the ci tests

Issue #25653, opened 2019-07-15
Upgrade to 0.58: Break Animated Snapshot


joan-saum:
I've recently upgrade my RN version from 0.57.8 to 0.59.10 (Will look soon to upgrade to 0.60)
But many of my snapshot test (using Jest) using an animation are failing. I've notice something broke them since the 0.58.0. but i can't figure it out

React Native version:
React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
      Memory: 24.96 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 25, 27, 28
        Build Tools: 23.0.1, 27.0.3, 28.0.2, 28.0.3, 29.0.0
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.10 => 0.59.10
    npmGlobalPackages:
      react-native-git-upgrade: 0.2.7

## Steps To Reproduce

An exemple of my animated Component :
I have a circle which do a translateY from 0 to 5 every 1000 ms all the time : 
```
Animated.loop(
      Animated.sequence([
        Animated.timing(move, {
          toValue: 100,
          duration: ANIM_DURATION,
          useNativeDriver: true,
        }),
        Animated.timing(move, {
          toValue: 0,
          duration: ANIM_DURATION,
          useNativeDriver: true,
        }),
      ]),
    ).start()
```
(It is called in the componentDidMount)

My test about it is like : 

```
global.useFakeAnimation();

expect(tree.root.instance.state.move.__getValue()).toStrictEqual(0);
global.advanceAnimationByTime(DELAY);
expect(tree.root.instance.state.move.__getValue()).toStrictEqual(100);

[...]

PR #26109, opened 2019-08-19
Mock createAnimatedComponent

joan-saum:
## Summary

Since react-native 0.58, I encountered some issues about snapshot with animated components. I opened an issue here : https://github.com/facebook/react-native/issues/25653

After hours of debugging, I finally found the problem:

In RN 0.57, an Animated View was created like this : 
`View: AnimatedImplementation.createAnimatedComponent(View)`

And `AnimatedImplementation` was mock like this : 
```
mock('../Libraries/Animated/src/AnimatedImplementation', () => {
  const AnimatedImplementation = jest.requireActual('../Libraries/Animated/src/AnimatedImplementation');
  const oldCreate = AnimatedImplementation.createAnimatedComponent;
    AnimatedImplementation.createAnimatedComponent = function(
      Component,
      defaultProps,
    ) {
      const Wrapped = oldCreate(Component, defaultProps);
      Wrapped.__skipSetNativeProps_FOR_TESTS_ONLY = true;
      return Wrapped;
    };
    return AnimatedImplementation;
  })
```

So thanks to this mock, the animated component had a props `__skipSetNativeProps_FOR_TESTS_ONLY` set to `true` and this was used to forceUpdate the component
```

if (AnimatedComponent.__skipSetNativeProps_FOR_TESTS_ONLY || typeof this._component.setNativeProps !== 'function') {
  this.forceUpdate();
}
```

But since RN 0.58, the way we create an animated component as changed into : 

```
const View = require('View');

const createAnimatedComponent = require('createAnimatedComponent');

module.exports = createAnimatedComponent(View);
```

As you can see, we directly use `createAnimatedComponent`, we don't use it through `AnimatedComponent` like before.
This caused the animated component had not anymore the props `__skipSetNativeProps_FOR_TESTS_ONLY`, so the component doesn't forceUpdate during the animation and breaks the snapshot.

Mocking `createAnimatedComponent` fix the problem

## Changelog

[General] [Fixed] - Mock createAnimatedComponent to fix snapshot with animated component

## Test Plan

See the issue

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -131,20 +131,18 @@ jest
     Animated.View.__skipSetNativeProps_FOR_TESTS_ONLY = true;
     return Animated;
   })
-  .mock('../Libraries/Animated/src/AnimatedImplementation', () => {
-    const AnimatedImplementation = jest.requireActual(
-      '../Libraries/Animated/src/AnimatedImplementation',
+  .mock('../Libraries/Animated/src/createAnimatedComponent', () => {
+    const createAnimatedComponent = jest.requireActual(
+      '../Libraries/Animated/src/createAnimatedComponent',
     );
-    const oldCreate = AnimatedImplementation.createAnimatedComponent;
-    AnimatedImplementation.createAnimatedComponent = function(
-      Component,
-      defaultProps,
-    ) {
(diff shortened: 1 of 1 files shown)


joan-saum:
Up

This is a regression and can be easily fix

PR #26115, opened 2019-08-19
fix: check "sendIdleEvents" before calling "stopTimers"

aleclarson:
## Summary

Avoid calling `stopTimers` when `sendIdleEvents` is true, since that's how `setSendIdleEvents` works.

## Changelog

[iOS] [Fixed] - avoid stopTimers when sendIdleEvents is true

## Test Plan

None


cpojer:
@aleclarson could you go into more detail on why this change is necessary and spec out the test plan to verify that the change is having the intended effect? It's hard to say why this change is necessary with the little information you provided. Thank you.

aleclarson:
Ah, I didn't notice the `_sendIdleEvents` check in `hasPendingTimers` method. My bad!

PR #26143, opened 2019-08-21
iOS: Add Appearance TurboModule, useColorScheme hook [DO NOT MERGE]

hramos:
This is a preview, and should not be merged.

### DIFF
--- a/Libraries/Utilities/Appearance.js
+++ b/Libraries/Utilities/Appearance.js
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+import EventEmitter from '../vendor/emitter/EventEmitter';
+import NativeEventEmitter from '../EventEmitter/NativeEventEmitter';
+import NativeAppearance, {type AppearancePreferences} from './NativeAppearance';
+import invariant from 'invariant';
+
+const COLOR_SCHEME_NAME = {
+  light: 'light',
+  dark: 'dark',
+  noPreference: 'no-preference',
+};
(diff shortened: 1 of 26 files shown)


hramos:
This is an exported version of the internal diff stack that starts at D16699954, to test changes in open source.

Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,104 @@
...
+   * Add an event handler that is fired when appearance preferences change.
+   */
+  static addChangeListener(listener: AppearanceListener): void {
+    eventEmitter.addListener('change', listener);
+  }
+
+  /**
+   * Remove an event handler.
+   */
+  static removeChangeListener(listener: AppearanceListener): void {
```
brentvatne: we should get rid of this function

Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,104 @@
...
+    } else {
+      appearancePreferencesInitialized = true;
+    }
+  }
+
+  /**
+   * Add an event handler that is fired when appearance preferences change.
+   */
+  static addChangeListener(listener: AppearanceListener): void {
+    eventEmitter.addListener('change', listener);
```
brentvatne: we should return the result of this, an `EventSubscription` (which people can just call `.remove()` on) rather than depend on `removeChangeListener`

Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param {object} appearancePreferences Simple string-keyed object of
+   * appearance preferences to set.
+   */
+  static set(preferences: AppearancePreferences): void {
+    let {colorScheme} = preferences;
+    appearancePreferences = {colorScheme};
+
+    if (appearancePreferencesInitialized) {
+      // Don't fire 'change' the first time the dimensions are set.
```
brentvatne: s/dimensions/appearance preferences/

PR #26163, opened 2019-08-23
[Android] Implement fading edges for ScrollView and it's dependent FlatList

akrger:
## Summary
This should add props for enabling horizontal and vertical fading
edges for Scrollview and FlatList.
These fading edges are used to communicate to the user that there is more content to see.

## Changelog

[Android] [Added] - fading edges props to the FlatList and ScrollView components

## Test Plan

Open the React Native test app and navigate to the FlatList section.
Enable the `useFadingEdges` switch and insert a number into `Fading edge length`.

device-2019-08-23-123745
device-2019-08-23-123844

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -335,6 +335,18 @@ type AndroidProps = $ReadOnly<{|
    * @platform android
    */
   persistentScrollbar?: ?boolean,
+  /**
+   * Fades out the edges of the the scroll content.
+   *
+   * If the value is greater than 0, the fading edges will be set accordingly
+   * to the current scroll direction and position,
+   * indicating if there is more content to show.
+   *
+   * The default value is 0.
+   *
+   * @platform android
+   */
+  fadingEdgeLength?: ?number,
 |}>;
 
 type VRProps = $ReadOnly<{|
--- a/RNTester/js/examples/FlatList/FlatListExample.js
+++ b/RNTester/js/examples/FlatList/FlatListExample.js
@@ -29,7 +29,14 @@ const {
   pressItem,
   renderSmallSwitchOption,
 } = require('../../components/ListExampleShared');
-const {Alert, Animated, StyleSheet, View} = require('react-native');
+const {
+  Alert,
+  Animated,
+  Platform,
+  StyleSheet,
+  TextInput,
+  View,
+} = require('react-native');
 
 import type {Item} from '../../components/ListExampleShared';
 
@@ -51,6 +58,7 @@ type State = {|
   virtualized: boolean,
   empty: boolean,
   useFlatListItemComponent: boolean,
+  fadingEdgeLength: number,
 |};
 
 class FlatListExample extends React.PureComponent<Props, State> {
@@ -65,6 +73,7 @@ class FlatListExample extends React.PureComponent<Props, State> {
     virtualized: true,
     empty: false,
     useFlatListItemComponent: false,
+    fadingEdgeLength: 0,
   };
 
   _onChangeFilterText = filterText => {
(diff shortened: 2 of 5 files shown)


sahrens:
Oh man that options list is getting pretty large...can you make it slightly larger and add "android-only" to this new option? Also, could you or someone you work with implement support for iOS as well? It would be awesome to have platform parity on as many features as possible. This is a bit old, but is probably still relevant: https://www.cocoanetics.com/2011/08/adding-fading-gradients-to-uitableview/

RNTester/js/examples/FlatList/FlatListExample.js
```diff
@@ -124,11 +135,29 @@ class FlatListExample extends React.PureComponent<Props, State> {
               {renderSmallSwitchOption(this, 'empty')}
               {renderSmallSwitchOption(this, 'debug')}
               {renderSmallSwitchOption(this, 'useFlatListItemComponent')}
+              {Platform.OS === 'android' && (
```
sahrens: Oh I see you added an android check - that's fine.

TheSavior:
Is there a need to have a distinction between `horizontalFadingEdgesEnabled` and `verticalFadingEdgesEnabled`? Could we just have `fadingEdgesEnabled` that fades the edges depending on which direction the scrollview is? And if we do that we only need `fadingEdgeLength`, since that defaults to 0 which is off that handles the option being enabled or disabled too.

That would make this go from needing three props, to only one. :)

sahrens:
I agree, let's only have a single edge length prop. I'm reverting now.

akrger:
I did combine the props to a single one, but unfortunately, I don't feel too confident to also implement the feature for iOS (I might be able to do that in a follow-up, though).

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -335,6 +335,20 @@ type AndroidProps = $ReadOnly<{|
    * @platform android
    */
   persistentScrollbar?: ?boolean,
+  /**
+   * Fades out the edges.
+   * The default value is false.
+   *
+   * @platform android
+   */
+  fadingEdgesEnabled?: ?boolean,
```
TheSavior: Do we need this Boolean? Or can we just check whether the length is 0 for disabled, non zero for enabled?

akrger: Yeah, you're right. Fixed.

RNTester/js/examples/FlatList/FlatListExample.js
```diff
@@ -51,6 +58,8 @@ type State = {|
   virtualized: boolean,
   empty: boolean,
   useFlatListItemComponent: boolean,
+  useFadingEdges: boolean,
```
TheSavior: This isn’t needed here anymore, right?

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -335,6 +335,13 @@ type AndroidProps = $ReadOnly<{|
    * @platform android
    */
   persistentScrollbar?: ?boolean,
+  /**
+   * The amount of fading.
```
TheSavior: Can you add a bit more description here? Assume that what you put here will be the sentence describing what this prop does in our websites api documentation.

RNTester/js/examples/FlatList/FlatListExample.js
akrger: Yes, fixed.

Libraries/Components/ScrollView/ScrollView.js
akrger: Thanks for the feedback. I added more description.

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollViewManager.java
```diff
@@ -276,4 +276,12 @@ public void setOverflow(ReactHorizontalScrollView view, @Nullable String overflo
   public void setPersistentScrollbar(ReactHorizontalScrollView view, boolean value) {
     view.setScrollbarFadingEnabled(!value);
   }
+
+  @ReactProp(name = "fadingEdgeLength")
+  public void setFadingEdgeLength(ReactHorizontalScrollView view, int value) {
+    if (value > 0) {
+      view.setHorizontalFadingEdgeEnabled(true);
+      view.setFadingEdgeLength(value);
+    }
```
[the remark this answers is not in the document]
akrger: Yeah, you're right, this might apply to dev environments where the prop is changed quite often. Maybe the fading edge length should also be reset?

sahrens:
Did you verify that toggling the fade switch in the example correctly shows and hides the fade?

akrger:
Yes, it works correctly. I'm now also resetting the fading edge length to zero if the fading edges are disabled, just in case.

sahrens:
Ok great.

PR #26169, opened 2019-08-23
Add accessibilityValueDescription support.

marcmulcahy:
## Summary

React Native components need a mechanism to specify their value to assistive technologies. This PR adds the notion of accessibilityValueDescription-- a property which either contains a textual description of a component's value, or for range-based components, such as sliders and progress bars, it contains range information (minimum, current, and maximum).

On iOS, the range-based info if present is converted into a percentage and added to the accessibilityValue property of the UIView. If text is present as part of the accessibilityValueDescription, it is used instead of the range-based information.

On Android, any range-based information in accessibilityValueDescription is exposed in the AccessibilityNodeInfo's RangeInfo. Text which is part of accessibilityValueDescription is appended to the content description.

## Changelog

[GENERAL] [Change] - add accessibilityValuedescription property.

## Test Plan

Added two new accessibility examples to RNTester, one which uses text and another which uses range-based info in accessibilityValueDescription. Verified that they both behave correctly on both Android and iOS.


necolas:
Internally, the equivalent we have is:

```
export type AccessibilityValue = $ReadOnly<{|
  max?: number,
  min?: number,
  now?: number,
  text?: Stringish,
|}>;
```

marcmulcahy:
I would have done something like that, but I didn't know how to workaround the fact that iOS also has a property named accessibilityValue, which we're using to cover lots of metadata that doesn't fall logically somewhere else (such as some roles nd states). Can you scribble me a quick code example of how to have a property named accessibiltyValue which is of type AccessibliityValue and also have the iOS implementation properly return a string to iOS for accessibilityValue?

necolas:
> I didn't know how to workaround the fact that iOS also has a property named accessibilityValue, which we're using to cover lots of metadata that doesn't fall logically somewhere else

Is "we" Amazon? Could you expand on what you're using iOS `accessibilityValue` for and what its role is on the iOS platform?

marcmulcahy:
accessibilityValue is used to present the value of a component in iOS-- something that's obviously not its label, but should be presented to the user. It's used, for example, to present the current value of a slider. on iOS, accessibilityValue is a string.

As part of the addition of numerous roles and states in PR #24095, we use accessibilityValue to present a few things:

* It's required to make VoiceOver present the state of switches correctly (accessibilityValue must be set to "0" for off or "1" for on)
* Any roles which do not have an associated iOS trait (including checkbox)
* Any states which do not have an associated iOS trait (such as checked)

So we'll want the implementation of accessibilityValue on iOS to continue behaving as it does, but also retrieve the accessibilityValue property from the underlying component (which isn't a string), and present its elements as well.

Incidentally, for range-based values, iOS does not expose a structure like the one you described or the one we're proposing, so we convert the min, current, max range info into a string and place it into accessibliityValue when presenting the range on iOS.

necolas:
Thanks, I see. In which case it might make sense to use a name other than `accessibilityValue`

cpojer:
@marcmulcahy could you address @blavalla's comments so we can merge this? :)

marcmulcahy:
Sorry, I assumed they were already addressed based on @blavalla 's last comment. Let me know what changes are suggested.

cpojer:
Ohh sorry my mistake. In that case do you mind rebasing so we can merge it?

PR #26172, opened 2019-08-23
[DO NOT MERGE] Preview: Appearance Native Module

hramos:
Implements the Appearance native module as discussed in https://github.com/react-native-community/discussions-and-proposals/issues/126.

The purpose of the Appearance native module is to expose the user's appearance preferences. It provides a basic API that returns the user's preferred color scheme on iOS 13 and Android 10devices, also known as Dark Mode or Night theme. It also provides the ability to subscribe to events whenever an appearance preference changes.

The new useColorScheme hook is provided as the preferred way of accessing the user's preferred color scheme.

The name, "Appearance", was chosen purposefully to allow for future expansion to cover other appearance preferences such as reduced motion, reduced transparency, or high contrast modes.

Changelog:

[iOS] [Added] - The Appearance native module can be used to prepare your app for Dark Mode on iOS 13.
[Android] [Added] - The Appearance native module can be used to prepare your app for Night theme on Android 10.

## Android: Configuration Changes

The AppearanceModule now emits a `appearanceChanged` event when the current uiMode configuration changes.

To make your app handle Night mode changes, make sure to do the following:

* Declare your Activity can handle uiMode configuration changes (https://developer.android.com/preview/features/darktheme#java):
```
android:configChanges="uiMode"
```
* Make sure to pass the configuration changed activity lifecycle callback from your ReactActivity:
```
Override
protected void onConfigurationChanged(Configuration newConfig) {
    super.onConfigurationChanged();

    if (mReactInstanceManager != null) {
        mReactInstanceManager.onConfigurationChanged(newConfig);
    }
}
```

## iOS: Dark Mode changes

If your app does not use RCTRootView, you'll need to override one of the lifecycle view/view controller methods that get called when the trait collection changes (see https://developer.apple.com/documentation/appkit/supporting_dark_mode_in_your_interface). Then, post a notification with name "RCTUserInterfaceStyleDidChangeNotification" and the new color scheme in the payload. Refer to the contents of this PR for more details.

### DIFF
--- a/Libraries/Utilities/Appearance.js
+++ b/Libraries/Utilities/Appearance.js
@@ -1,0 +1,79 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+import EventEmitter from '../vendor/emitter/EventEmitter';
+import NativeEventEmitter from '../EventEmitter/NativeEventEmitter';
+import NativeAppearance, {
+  type AppearancePreferences,
+  type ColorSchemeName,
+} from './NativeAppearance';
+import invariant from 'invariant';
+
+type AppearanceListener = (preferences: AppearancePreferences) => void;
+const eventEmitter = new EventEmitter();
+
+const nativeColorScheme: ?string =
+  NativeAppearance == null ? null : NativeAppearance.getColorScheme();
+invariant(
+  nativeColorScheme === 'dark' ||
+    nativeColorScheme === 'light' ||
+    nativeColorScheme == null,
+  "Unrecognized color scheme. Did you mean 'dark' or 'light'?",
+);
+
+let currentColorScheme: ?ColorSchemeName = nativeColorScheme;
+
+if (NativeAppearance) {
+  const nativeEventEmitter = new NativeEventEmitter(NativeAppearance);
+  nativeEventEmitter.addListener(
+    'appearanceChanged',
+    (newAppearance: AppearancePreferences) => {
+      const {colorScheme} = newAppearance;
+      invariant(
+        colorScheme === 'dark' ||
+          colorScheme === 'light' ||
+          colorScheme == null,
+        "Unrecognized color scheme. Did you mean 'dark' or 'light'?",
+      );
+      currentColorScheme = colorScheme;
+      eventEmitter.emit('change', {colorScheme});
+    },
+  );
+}
+
+module.exports = {
+  /**
+   * Note: Although color scheme is available immediately, it may change at any
+   * time. Any rendering logic or styles that depend on this should try to call
+   * this function on every render, rather than caching the value (for example,
+   * using inline styles rather than setting a value in a `StyleSheet`).
+   *
+   * Example: `const colorScheme = Appearance.getColorScheme();`
+   *
+   * @returns {?ColorSchemeName} Value for the color scheme preference.
+   */
+  getColorScheme(): ?ColorSchemeName {
+    return currentColorScheme;
+  },
+  /**
+   * Add an event handler that is fired when appearance preferences change.
+   */
+  addChangeListener(listener: AppearanceListener): void {
+    eventEmitter.addListener('change', listener);
+  },
+  /**
+   * Remove an event handler.
+   */
+  removeChangeListener(listener: AppearanceListener): void {
+    eventEmitter.removeListener('change', listener);
+  },
+};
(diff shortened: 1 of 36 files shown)


Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,79 @@
...
+   * @returns {?ColorSchemeName} Value for the color scheme preference.
+   */
+  getColorScheme(): ?ColorSchemeName {
+    return currentColorScheme;
+  },
+  /**
+   * Add an event handler that is fired when appearance preferences change.
+   */
+  addChangeListener(listener: AppearanceListener): void {
+    eventEmitter.addListener('change', listener);
```
brentvatne:
```suggestion
    return eventEmitter.addListener('change', listener);
```

Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @returns {?ColorSchemeName} Value for the color scheme preference.
+   */
+  getColorScheme(): ?ColorSchemeName {
+    return currentColorScheme;
+  },
+  /**
+   * Add an event handler that is fired when appearance preferences change.
+   */
+  addChangeListener(listener: AppearanceListener): void {
```
brentvatne:
```suggestion
  addChangeListener(listener: AppearanceListener): EventSubscription {
```

Libraries/Utilities/Appearance.js
```diff
@@ -0,0 +1,79 @@
...
+    eventEmitter.addListener('change', listener);
+  },
+  /**
+   * Remove an event handler.
+   */
+  removeChangeListener(listener: AppearanceListener): void {
```
brentvatne: imo we should remove this function and have people just call `subscription.remove()` on the returned value from `addChangeListener`

Libraries/Utilities/useColorScheme.js
```diff
@@ -0,0 +1,31 @@
...
+import {useSubscription} from 'use-subscription';
+import Appearance from './Appearance';
+import type {ColorSchemeName} from './NativeAppearance';
+
+export default function useColorScheme(): ?ColorSchemeName {
+  const subscription = useMemo(
+    () => ({
+      getCurrentValue: () => Appearance.getColorScheme(),
+      subscribe: callback => {
+        Appearance.addChangeListener(callback);
```
brentvatne:
```suggestion
        let eventSubscription = Appearance.addChangeListener(callback);
```

Libraries/Utilities/useColorScheme.js
```diff
@@ -0,0 +1,31 @@
...
+        return () => Appearance.removeChangeListener(callback);
```
brentvatne:
```suggestion
        return () => eventSubscription.remove();
```

React/CoreModules/RCTAppearance.mm
```diff
@@ -0,0 +1,107 @@
...
+{
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
+  if (@available(iOS 13.0, *)) {
+    static NSDictionary *appearances;
+    static dispatch_once_t onceToken;
+
+    dispatch_once(&onceToken, ^{
+      appearances = @{
+                      @(UIUserInterfaceStyleLight): RCTAppearanceColorSchemeLight,
+                      @(UIUserInterfaceStyleDark): RCTAppearanceColorSchemeDark
```
brentvatne: we should add back `@(UIUserInterfaceStyleUnspecified): RNCAppearanceColorSchemeNoPreference` here to better match related [web spec](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). see [related discussion](https://github.com/react-native-community/discussions-and-proposals/issues/126#issuecomment-521070504)

jfrolich:
Awesome! Would there be a declarative way to add colors with a light and dark appearance in stylesheets that implement two values for dark-mode and light-mode? (creating the appropriate NSColor and something equivalent in Android).

This allows to declare colors in views that are not declarative (or directly managed by react). For instance in 'react-native-navigator', where you can add styles of the appearance of a modal view once (when opening), but the modal view is not in the react view hierarchy.

This also implements the dark-mode more directly as you would in native apps. Including the fade to the new color when the mode is  changing.

I do see that this needs changes in the Stylesheet internals. If that is not happening it would also be great to be able to use native "Color Sets" and system colors in react-native. Perhaps by prefixing the color name with a symbol?

jfrolich:
> @jfrolich - no, that's not part of this PR. see [react-native-community/discussions-and-proposals#126](https://github.com/react-native-community/discussions-and-proposals/issues/126) for more discussion

Thanks for referencing! Definitely good to have a hook to consume the appearance preference, and a way to do dark mode before dynamic colors are available in react-native.

hramos:
Closing as this was a preview of the internal stack, of which the first commit just landed: https://github.com/facebook/react-native/commit/63fa3f21c5ab308def450bffb22054241a8842ef

PR #26175, opened 2019-08-24
rename useWindowDimensions -> useDimensions

dulmandakh:
## Summary

Dimensions module has support for both window and screen dimensions, but useWindowDimensions omitted screen. This PR renamed useWindowDimensions to useDimensions, and supports both window and screen dimensions, while defaulting to window.

## Changelog

[General] [Changed] - renamed hook useWindowDimensions -> useDimensions

## Test Plan

Everything works as expected.


sahrens:
I'm not sure we want to make this change. It's very rare that any app would want to use screen instead of window, and `useDimensions` is such a generic name...dimensions of what?

Why do you think this is an important change to make?

dulmandakh:
@sahrens I think that people still use Dimensions module, and we should have matching hook for that.

sahrens:
That's not really a reason. The Dimensions module is quite flawed - I don't see a need to perpetuate its flaws by replacing it as close to 1:1 as possible. If people really want to access screen, they can easily copy-pasta their own custom hook, but I don't think we should encourage it in the core.

cpojer:
I agree with @sahrens here. Let's close this PR, hope that's ok @dulmandakh.

Issue #25127, opened 2019-06-03
useEffect hooks are not flushed when app is backgrounded


benoitdion:
React Native version:
```
  React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
      Memory: 161.89 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 8.12.0 - ~/.nvm/versions/node/v8.12.0/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.5.0 - ~/.nvm/versions/node/v8.12.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 16, 18, 19, 20, 22, 23, 24, 25, 26, 27, 28
        Build Tools: 19.1.0, 21.1.2, 23.0.1, 23.0.3, 24.0.2, 24.0.3, 25.0.0, 25.0.1, 25.0.2, 26.0.1, 26.0.2, 27.0.1, 27.0.2, 27.0.3, 28.0.3
        System Images: a... | Android TV Intel x86 Atom Syst..., a...ear | Android Wear ARM EABI v7a Syst..., a...ndroid-wear | Android Wear Intel x86 Atom Sy..., a...fault | ARM 64 v8a, a...ult | ARM EABI v7a, a...-24 | Intel x86 Atom, a... | Intel x86 Atom_64, a..._apis | Google APIs ARM 64 v8a System ..., a...pis | Google APIs ARM EABI v7a Syste..., a...google_apis | Google APIs Intel x86 Atom Sys..., a...gle_apis | Google APIs Intel x86 Atom_64 ..., a...s_playstore | Google Play Intel x86 Atom Sys..., a... | Android TV Intel x86 Atom Syst..., a...google_apis | Google APIs Intel x86 Atom Sys..., a...gle_apis | Google APIs Intel x86 Atom_64 ..., a...google_apis | Google APIs Intel x86 Atom Sys...
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.6 => 16.8.6
[...]

PR #26176, opened 2019-08-24
useAppState hook

dulmandakh:
## Summary

This PR introduces useAppState hook.

## Changelog

[GENERAL] [Added] - new useAppState hook

## Test Plan

It works 👍

### DIFF
--- a/Libraries/AppState/AppState.js
+++ b/Libraries/AppState/AppState.js
@@ -18,6 +18,8 @@ const logError = require('../Utilities/logError');
 const invariant = require('invariant');
 
 /**
+ * NOTE: `useAppState` is the preffered API for React components.
+ *
  * `AppState` can tell you if the app is in the foreground or background,
  * and notify you when the state changes.
  *
--- a/Libraries/AppState/useAppState.js
+++ b/Libraries/AppState/useAppState.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow strict-local
+ */
+
+'use strict';
+
+import AppState from './AppState';
+import {useEffect, useState} from 'react';
+
+export default function useAppState(): string {
+  const [state, setState] = useState(AppState.currentState);
+  useEffect(() => {
+    function handleChange(state) {
+      setState(state)
+    }
+
+    AppState.addEventListener('change', handleChange)
+    setState(AppState.currentState)
+    return () => {
+      AppState.removeEventListener('change', handleChange)
+    }
+  }, [])
+  return state
+}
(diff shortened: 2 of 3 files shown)


brunolemos:
There’s this repo with some more react native hooks: https://github.com/react-native-community/react-native-hooks

Are we gonna have most of them inside the official react-native package? 

Given the Lean Core initiative I was not sure, but it would make sense, because it could be used internally by react-native once it starts to use less classes.

dulmandakh:
@brunolemos yep, that's the reason.

TheSavior:
This might be a good opportunity, can you try adding a jest test for this hook?

benoitdion:
related: https://github.com/facebook/react-native/issues/25127

Libraries/AppState/AppState.js
```diff
@@ -18,6 +18,8 @@ const logError = require('../Utilities/logError');
 const invariant = require('invariant');
 
 /**
+ * NOTE: `useAppState` is the preffered API for React components.
```
cpojer: *preferred.

Libraries/AppState/useAppState.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    AppState.addEventListener('change', handleChange)
+    setState(AppState.currentState)
+    return () => {
+      AppState.removeEventListener('change', handleChange)
+    }
+  }, [])
+  return state
+}
```
cpojer: Needs a newline

benoitdion:
To add a bit more color to the issue I linked to above, it means that a `useAppState` hook isn't going to work properly with `useEffect`. 

You should be able to reproduce with something like the following:

```javascript
const AppStateView = () => {
  const appState = useAppState();

  useEffect(() => {
    console.log(">>> [useEffect] new app state:", appState);
  }, [appState]);

  useLayoutEffect(() => {
    console.log(">>> [useLayoutEffect] new app state:", appState);
  }, [appState]);

  return <Text>{appState}</Text>;
};
```

radeno:
@cpojer @dulmandakh 
There is community module which has similar behavior https://github.com/react-native-community/react-native-hooks#useappstate
I think this "utility" functions should be in community module if they are not used internally in RN core.
Even RN core utility `useWindowDimensions` is reimplemented it as `useDimensions`.
I think, core should be lean and prefer that module instead implementing same features into core.

dulmandakh:
IMHO, RN should include hooks for core modules

radeno:
@dulmandakh i agree with you about using hooks. But it depends how to propagate it.
There are two ways:
1. Add to documentation that community module exists. It handles more that appState and screenDimensions. It is good.
2. Add hooks into every core module. It should be done for all core modules and published in one version because consistency. Later, it will be easier to extract some core modules as independed react-native-community module with same functionality.

First option is low-hanging fruit. Second option should be future-proof, but takes a lot time.

cpojer:
We should provide hook based APIs for React Native and possibly integrate the hooks from the community project. That will open a path to possibly deprecate older APIs in the future.

Libraries/AppState/useAppState.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default function useAppState(): string {
+  const [state, setState] = useState(AppState.currentState);
+  useEffect(() => {
+    function handleChange(state) {
+      setState(state)
+    }
+
+    AppState.addEventListener('change', handleChange)
+    setState(AppState.currentState)
```
radeno: Is it necessary to set state again after mounting?

cpojer:
@dulmandakh what's the status of this PR?

PR #26178, opened 2019-08-24
use Keep annotation instead of DoNotStrip

dulmandakh:
## Summary

This PR replaces custom DoNotStrip annotation uses with Keep annotation from AndroidX, because in ReactAndroid we use nullability annotations from AndroidX and AndroidX annotations are already in dependencies.

## Changelog

[Android] [Changed] - use Keep annotation instead of DoNotStrip

## Test Plan

RNTester builds and runs as expected.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/hermes/instrumentation/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/hermes/instrumentation/BUCK
@@ -13,7 +13,7 @@ rn_android_library(
     srcs = ["HermesSamplingProfiler.java"],
     visibility = ["PUBLIC"],
     deps = [
-        react_native_dep("java/com/facebook/proguard/annotations:annotations"),
+        react_native_dep("third-party/android/androidx:annotation"),
         react_native_dep("libraries/soloader/java/com/facebook/soloader:soloader"),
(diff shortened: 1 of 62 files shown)


dulmandakh:
Yep, Yoga ones and we need to discuss with FB guys about using AndroidX
annotations in it. Previously, they suggested not to use it when I migrated
nullable annotations.

bartolkaruza:
Just posting this for reference: https://github.com/facebook/react-native/commit/ffdf3f22c68583fe77517f78dd97bd2e97ff1b9e

yungsters:
By the way, I have been in discussions with the team at Facebook working on [Redex](https://github.com/facebook/redex) to understand if we can safely migrate to `@Keep` instead of `@DoNotStrip`.

If I can get a consensus that we can safely do this, @dulmandakh — would you be willing to rebase this pull request?

yungsters:
Closing this because it was opened a long time ago and would require a rebase / rewrite.

If we want to pursue this in the future, I recommend only converting a single annotation (or a single class) so that we can verify it does not cause problems with people's Proguard configurations (e.g. if people are not composing the default configurations). If we can develop confidence with one call site, then we can go with a broader migration.

PR #26197, opened 2019-08-26
Add some missing dependencies to the fabric podspec

janicduplessis:
## Summary

Managed to get fabric building in my app. Here are the changes needed in the podspec.

There were other issues with codegen but it would be too involved to fix for now. This is a good first spec to get it working better in OSS.

## Changelog

[iOS] [Fixed] - Add some missing dependencies to the fabric podspec

## Test Plan

Managed to get fabric building in a project.


fkgozali:
this seems fine cc @jtreanor

janicduplessis:
ping @osdnk @fkgozali

PR #26200, opened 2019-08-26
fix: add "transformsDiffer" for CATransform3D props

aleclarson:
## Summary

<s>Native props with `CATransform3D` type were not being processed before being diffed, which resulted in `matricesDiffer` always returning `false` after first render.</s>

Replace `matricesDiffer` comparison (for `CATransform3D` props) with new `transformsDiffer` function. The native `[RCTConvert CATransform3D:]` function has deprecated matrix arrays.

## Changelog

[iOS] [Fixed] - add "transformsDiffer" for CATransform3D props

## Test Plan

None

### DIFF
--- a/Libraries/ReactNative/getNativeComponentAttributes.js
+++ b/Libraries/ReactNative/getNativeComponentAttributes.js
@@ -15,13 +15,13 @@ const UIManager = require('./UIManager');
 
 const insetsDiffer = require('../Utilities/differ/insetsDiffer');
 const invariant = require('invariant');
-const matricesDiffer = require('../Utilities/differ/matricesDiffer');
 const pointsDiffer = require('../Utilities/differ/pointsDiffer');
 const processColor = require('../StyleSheet/processColor');
 const processColorArray = require('../StyleSheet/processColorArray');
 const processTransform = require('../StyleSheet/processTransform');
 const resolveAssetSource = require('../Image/resolveAssetSource');
 const sizesDiffer = require('../Utilities/differ/sizesDiffer');
+const transformsDiffer = require('../Utilities/differ/transformsDiffer');
 const warning = require('fbjs/lib/warning');
 
 function getNativeComponentAttributes(uiViewClassName: string): any {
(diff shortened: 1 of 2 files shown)


osdnk:
Let's see

Libraries/Utilities/differ/transformsDiffer.js
```diff
@@ -0,0 +1,19 @@
...
+ * @format
+ * @flow
+ */
+'use strict';
+
+const transformsDiffer = function(one, two) {
+  return one !== two ||
+    !one || !two ||
+    one.length !== two.length ||
+    JSON.stringify(one) !== JSON.stringify(two);
```
shergin: Why do we need to use `JSON.stringify` here?
Which kinda values we diff here?

aleclarson: An array of objects (eg: `[{ scaleX: 0.5 }, { translateX: 100 }]`) where order matters.

Libraries/ReactNative/getNativeComponentAttributes.js
```diff
@@ -15,12 +15,13 @@ const UIManager = require('./UIManager');
 
 const insetsDiffer = require('../Utilities/differ/insetsDiffer');
 const invariant = require('invariant');
-const matricesDiffer = require('../Utilities/differ/matricesDiffer');
 const pointsDiffer = require('../Utilities/differ/pointsDiffer');
 const processColor = require('../StyleSheet/processColor');
 const processColorArray = require('../StyleSheet/processColorArray');
+const processTransform = require('../StyleSheet/processTransform');
```
philIip, line 21: why do we need this import?

Libraries/Utilities/differ/transformsDiffer.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+'use strict';
+
+const transformsDiffer = function(one, two) {
```
philIip, line 12: if both one and two are null, this would return true right? that seems not totally correct

philIip (changes requested):
thanks for working on this! i have a couple of comments.

aleclarson:
bump

PR #26265, opened 2019-08-30
Add RNTester Pods to gitignore

friederbluemle:
## Summary

Pods directory showed up as untracked files after running `pod install`
in the `RNTester` directory. With this small change to `.gitignore`, it
will no longer show up.

## Changelog

Small repo-level change, no changelog needed (?)

## Test Plan

Run `pod install` in `/RNTester`, verify that "Pods" no longer show up
as untracked files in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,5 +80,4 @@ RNTester/build
 # CocoaPods
 /template/ios/Pods/
 /template/ios/Podfile.lock
-RNTester/Pods/Local Podspecs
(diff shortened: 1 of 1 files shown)


fkgozali:
Hello! It looks like this broke some continuous build for FB internal system. I'm reverting this change for now, and we'll put it back in after we resolve the issue. Sorry for the inconvenience.

PR #26294, opened 2019-09-02
iOS13 status bar has now 3 styles

gaodeng:
iOS13 status bar has now 3 styles
UIStatusBarStyleDefault, UIStatusBarStyleLightContent, UIStatusBarStyleDarkContent

## Summary
UIStatusBarStyleDefault now acts as an automatic style which will set it’s value dynamically based on the the userinterfacestyle(One of the traits) of the viewcontroller that controls the status bar appearance.

## Changelog

[iOS] [Fixed] - iOS13 new status bar style UIStatusBarStyleDarkContent

### DIFF
--- a/React/Modules/RCTStatusBarManager.m
+++ b/React/Modules/RCTStatusBarManager.m
@@ -19,35 +19,39 @@
   static NSDictionary *mapping;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
-      if(@available(iOS 13.0, *)){
-        mapping=@{
-          @"default": @(UIStatusBarStyleDefault),
-          @"light-content": @(UIStatusBarStyleLightContent),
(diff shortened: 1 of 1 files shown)


React/Modules/RCTStatusBarManager.m
```diff
@@ -14,12 +14,25 @@
 #if !TARGET_OS_TV
 @implementation RCTConvert (UIStatusBar)
 
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_13_0) && \
```
janicduplessis: Is if possible to have this #ifdef just around the dark-content enum value instead of duplicating the whole thing?

gaodeng: Fixed

Issue #22703, opened 2018-12-19
'glog/logging.h' file not found


MohamadKh75:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
```
React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: (4) x64 Intel(R) Core(TM) i7-2620M CPU @ 2.70GHz
      Memory: 6.74 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.13.0 - /usr/local/bin/node
      Yarn: 1.12.3 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
    IDEs:
      Xcode: 10.0/10A255 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3
      react-native: 0.58.0-rc.1 => 0.58.0-rc.1
    npmGlobalPackages:
      react-native-cli: 2.0.1
```

## Description
I just tried to create a new project using `react-native init --version=react-native@0.58.0-rc.1` for testings about [this issue](https://github.com/facebook/react-native/issues/22267) but i face this problem! I guess this is a regression!

## Reproducible Demo
1. `react-native init MyTestingProject --version=react-native@0.58.0-rc.1`
2. `cd MyTestingProject/ios`
3. open `MyTestingProject.xcodeproj`
4. File / Project Settings... / Build System => Legacy Build System / Done
5. Product / Clean
6. Product / Clean Build Folder
7. Build

EdwardDrapkin:
Did you try the ole `rm -rf ~/.rncache node_modules/react-native/third-party/*`?

MohamadKh75:
> Did you try the ole `rm -rf ~/.rncache node_modules/react-native/third-party/*`?

Yep but no result!

MohamadKh75:
I can confirm that this issue is resolved in Xcode 10.1 but exist in Xcode 10

kelset:
thanks for reporting, I feel that we can close this then.

thecodecafe:
[...]

PR #26300, opened 2019-09-02
fixed wrong path on install script - #22703

thisisablock:
## Summary

On installing Pods on Mac OS, the script fails on downloading the dependencies.

```
cd "$dir"
eval "${cmd:-true}" && touch .installed)
```

will fail, because the workdir is wrong here. 
cd "$dir" is accessing the extracted folder of the source but the script call on

```
fetch_and_unpack glog-0.3.5.tar.gz https://github.com/google/glog/archive/v0.3.5.tar.gz 61067502c5f9769d111ea1ee3f74e6ddf0a5f9cc "\"$SCRIPTDIR/ios-configure-glog.sh\""
```

is referencing on ./ios-configure-glog.sh

this is fixing the 'glog/logging.h' file not found error on build.
on the most users the error will not occur, because glog was compiled before 
or the ~/.rncache already has all files.

## Changelog

- changed relative path to a correct absolute to fix non fetching third-party libs on pod install

iOS Changed - pod build script for third party libs

## Test Plan

- clone fresh repo
- install node dependencies
- pod install (to fetch glog, etc.)
- build

### DIFF
--- a/scripts/ios-install-third-party.sh
+++ b/scripts/ios-install-third-party.sh
@@ -69,7 +69,7 @@ function fetch_and_unpack () {
 
 mkdir -p third-party
 
-SCRIPTDIR=$(dirname "$0")
+SCRIPTDIR=$(pwd)
 
 fetch_and_unpack glog-0.3.5.tar.gz https://github.com/google/glog/archive/v0.3.5.tar.gz 61067502c5f9769d111ea1ee3f74e6ddf0a5f9cc "\"$SCRIPTDIR/ios-configure-glog.sh\""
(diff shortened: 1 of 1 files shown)


cpojer:
I don't think `pwd` is a good idea because it's where the script was invoked from but this script wants to have the directory of the scripts folder itself so I don't think that's the correct fix. I'm struggling to understand your summary. Could you explain the problem and the before and after state with your fix applied? That way I can try to reproduce this myself and confirm the fix is working.

alloy:
I just came across this PR while digging through the history of this script. This script was removed in #27810, so going to close this now. Thanks for the contribution, though!

PR #26317, opened 2019-09-04
Added prefix to node_modules into Podfile template

jMarinhoX:
## Summary

The Podfile used for HelloWorld template has verbose pod's paths. Adding a prefix variable for `"../node_modules/react-native"` improves readability and base path modification.

## Changelog

[iOS] [Added] - Added prefix to node_modules into Podfile template

## Test Plan

Run pod install in a 0.60 project with this Podfile (changing some paths to obey version) and verify that all pods are located and installed correctly.

### DIFF
--- a/template/ios/Podfile
+++ b/template/ios/Podfile
@@ -1,53 +1,14 @@
 platform :ios, '9.0'
 require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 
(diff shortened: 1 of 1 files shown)


orta:
I'm not sure this is worth it (even outside of it not working) - having the full path in each one does make it obvious where they are coming from.

yungsters:

PR #26359, opened 2019-09-06
Make Yoga lowercase in Podfile

hassanbazzi:
## Summary

When doing RN init from scratch for 0.61, I noticed pod install was failing. This was because "yoga" was mis-spelled to "Yoga" in the podfile. Thus, it couldn't find the podspec file.

```
[!] No podspec found for `Yoga` in `../node_modules/react-native/ReactCommon/yoga
```
## Changelog

iOS [Fixed] - Fix yoga case sensitivity in PodFile to match PodSpec


hassanbazzi:
Whoops. Wrong change.

PR #26360, opened 2019-09-06
Rename yoga.podspec to Yoga.podspec

hassanbazzi:
## Summary

When doing RN init from scratch for 0.61, I noticed pod install was failing. This was because "Yoga.podspec" was mis-spelled to "yoga.podspec" in the yoga ReactCommon folder. Thus, `pod install` was failing

```
[!] No podspec found for `Yoga` in `../node_modules/react-native/ReactCommon/yoga
```
## Changelog

[iOS] [Fixed] - Fix Yoga.podspec case sensitivity

## Test Plan

`pod install` now runs successfully with a clean build of master.


jerson:
@hassanbazzi When yoga is renamed Yoga, many imports need to be replaced, wouldn't it be better to change the pod to yoga?

grabbou:
Introduced via https://github.com/facebook/react-native/commit/8d02e1723fd972d55eceb7de7c25dacf691d83df CC: @axe-fb @axemclion (not sure which one is "the right" now)

grabbou:
I think what happened here is that Git is not tracking case changes in the filename and you need to run `git mv` yourself instead of `mv` to rename the files. 

At least that's what I had to do when applying this change in the 0.61-stable branch. I need this fix right now, but let's ship this to `master` instead, so that @hassanbazzi can get its first contribution right there!

(btw, thanks for sending this!)

cpojer:
Let's see if this works with out syncing at FB, otherwise we may have to do something funky here ;)

axe-fb:
The reason I had to make `Yoga` upper case is that the Yoga Cocoapods externally are used as such. Also, i believe that the convention in cocoapods is to use upper case names.

PR #26379, opened 2019-09-09
Petterh/support parcelable array args

petterh:
## Summary

`ReactRootView.startReactApplication` takes a `Bundle` argument called `initialProperties`. This is translated to a `ReadableMap` via `Arguments.fromBundle`. If the bundle contains an array of bundles, this gets translated by `Arguments.fromArray`.

If the bundle was passed from one activity to another via intent extras, however, there is a problem. After the bundle has been marshaled and unmarshaled, the array of `Bundle`s come out the other end as an arrap of `Parcelable`s, although each array element remains a `Bundle`. This results in an "Unknown array type" exception.

This PR fixes this by adding support for `Parcelable` arrays &ndash; provided they contain only members of type `Bundle`.

## Changelog

[Android] [Fixed] - Don't throw "Unknown array type" exception when passing serialized bundle arrays in ReactRootView.startReactApplication's initialProperties parameter

## Test Plan

Added test class `ArgumentsTest`. The test method `testFromMarshaledBundle` fails when used on the old version of `Arguments`.


petterh:
I don't understand what's up with the failing tests -- none of them appear related to these changes. This doesn't seem to be the only PR thus affected...

PR #26413, opened 2019-09-12
fix: prevent activity recreation on theme change

Esemesek:
## Summary

On Android 10 whole activity is recreated when the user changes system theme. We should prevent it by adding `uiMode` in `android:configChanges` key in the default template.

## Changelog

[Android] [Fixed] - Android 10: Prevent activity recreation on theme change

## Test Plan

NOOP

### DIFF
--- a/template/android/app/src/main/AndroidManifest.xml
+++ b/template/android/app/src/main/AndroidManifest.xml
@@ -13,7 +13,7 @@
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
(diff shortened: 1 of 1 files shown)


template/android/app/src/main/AndroidManifest.xml
```diff
@@ -13,7 +13,7 @@
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
-        android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
```
bishop65: Works better

PR #26423, opened 2019-09-13
remove flow type annotation from template

dulmandakh:
## Summary

This PR removes flow type annotation from App.js in the template, so it won't break non-flow setup.

## Changelog

[General] [Changed] - remove flow type annotation from template 

## Test Plan

new project created from master will work even without flow.

### DIFF
--- a/template/App.js
+++ b/template/App.js
@@ -24,7 +24,7 @@ import {
   ReloadInstructions,
 } from 'react-native/Libraries/NewAppScreen';
 
-const App: () => React$Node = () => {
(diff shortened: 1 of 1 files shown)


cpojer:
It turns out we actually use this to verify the Flow setup in the template is working so I'm not sure how we'd remove this. Given that the flow-strip-types transform is not very costly and invasive, I'm feeling like we can keep this with little harm. Is there a concrete use-case that I'm missing that makes this necessary?

PR #26429, opened 2019-09-13
Fix ShareSheet crash on iOS 13

tomtargosz:
## Summary

Currently on iOS 13 the app will crash if you:
- Open the share sheet
- Tap something like messages or photos
- Cancel the dialog
- Perform any other action

This is because `shareController.completionWithItemsHandler` is called when the dialog box is canceled and currently `failureCallback` or `successCallback` will always be called. In the situation above, `activityError` is `nil` so `successCallback` will be called even though `completed` is false. This leaves us in a state where the callback has been invoked but the ShareSheet is still active, meaning the success or error callback will be invoked again, leading to the crash.

This PR adds a check to make sure `completed` is true before calling `successCallback`. This way `successCallback` will only be called when the user has successfully completed an action and the ShareSheet is closed.

## Changelog

[iOS] [Fixed] - Fix crash in RCTActionSheetManager.m on iOS 13

## Test Plan

- Saved an image successfully
- Opened and dismissed the `Photos` dialog multiple times without crashing


PeteTheHeat:
Awesome, thanks for the fix :)

PR #26431, opened 2019-09-13
Fix printing of false boolean values in the style inspector

janicduplessis:
## Summary

React can't print false, we can just use `JSON.stringify` for any values except string to avoid adding quotes.

## Changelog

[General] [Fixed] - Fix printing of false boolean values in the style inspector

## Test Plan

Before

image

After

image

### DIFF
--- a/Libraries/Inspector/StyleInspector.js
+++ b/Libraries/Inspector/StyleInspector.js
@@ -33,13 +33,12 @@ class StyleInspector extends React.Component<$FlowFixMeProps> {
 
         <View>
           {names.map(name => {
-            const value =
-              typeof this.props.style[name] !== 'string'
-                ? JSON.stringify(this.props.style[name])
-                : this.props.style[name];
+            const value = this.props.style[name];
             return (
               <Text key={name} style={styles.value}>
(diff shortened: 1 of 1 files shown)


JoshuaGross:
If `typeof` is `number`, do you think it's preferable to JSON.stringify or just concatenate that? I would imagine results could be different in some cases?

For example, one pretty important case:

`JSON.stringify(NaN) == "null"`
`""+NaN == "NaN"`

janicduplessis:
AFAIK there are no style values that support NaN, but yeah we could pass the value directly to react for numbers. Do you know if react would print NaN in that case (not on a computer to test)?

JoshuaGross:
@janicduplessis I'm not sure! It's a good question.

Can you see what happens with `NaN` and `Infinity`? I think that probably covers all the interesting cases, even if those are uncommon.

janicduplessis:
@JoshuaGross Ok so React prints NaN and Infinity correctly (as NaN and Infinity string) while JSON.stringify returns null. Lets just pass numbers directly so we handle these cases better.

PR #26432, opened 2019-09-13
Fix `includeFontPadding` for `TextInput` placeholder

janicduplessis:
## Summary

The custom font I'm using requires using `includeFontPadding={false}` to be correctly centered vertically. The only case where this is not working is with the placeholder of `TextInput`. To fix it we call `setIncludeFontPadding` on the `EditText` instance, like we do for `Text`.

## Changelog

[Android] [Fixed] - Fix `includeFontPadding` for `TextInput` placeholder

## Test Plan

Tested the fix in an app.

Before

image

After

image

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -291,6 +291,11 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
     }
   }
 
+  @ReactProp(name = ViewProps.INCLUDE_FONT_PADDING, defaultBoolean = true)
+  public void setIncludeFontPadding(ReactEditText view, boolean includepad) {
+    view.setIncludeFontPadding(includepad);
+  }
+
   @ReactProp(name = "importantForAutofill")
   public void setImportantForAutofill(ReactEditText view, @Nullable String value) {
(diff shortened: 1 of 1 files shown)


JoshuaGross:
Add flow type to AndroidTextInputNativeComponent.js

janicduplessis:
The flow type seems to already be there since it is used by the shadow node https://github.com/facebook/react-native/blob/janicduplessis-patch-13/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js#L528

JoshuaGross:
Can you retest this with borders around the TextInput? We're curious if this needs to be accounted for in the measure function, if this will mess up layout slightly.

janicduplessis:
Looks like borders are accounted for the view size properly but the text is not offset.

image

```js
    <TextInput
        placeholder="Hello"
        style={{
          borderColor: 'red',
          borderWidth: 15,
          fontSize: 32,
          margin: 16,
        }}
      />
      <TextInput
        placeholder="Hello"
        style={{
          borderColor: 'red',
          borderWidth: 1,
          fontSize: 32,
          margin: 16,
        }}
      />
      <TextInput
        placeholder="Hello"
        style={{
          borderColor: 'red',
          borderWidth: 15,
          fontSize: 32,
          margin: 16,
          paddingLeft: 15,
        }}
      />
```

Not sure if this was related to this PR but setting `includeFontPadding` doesn't affect this.

janicduplessis:
Uneven borders also do not affect the text vertical positioning properly.

image

PR #26433, opened 2019-09-14
Fix `cursorColor` on Android Q+

janicduplessis:
## Summary

`cursorColor` is broken since android P restricted access to private APIs. Luckily Android added a new API in Android Q to set the cursor drawable programmatically. Sadly it will stay broken for Android P and O but at least it will work going forward. I still kept the reflection implementation for now, we might want to drop it at some point.

Before landing this we need to bump the compile sdk version to SDK 29. This should not be done in this PR.

## Changelog

[Android] [Fixed] - Fix `cursorColor` on Android Q+

## Test Plan

Added an example in RNTester and checked that it works on an Android Q simulator.


janicduplessis:
Now that we compile with sdk 29 this should be mergeable.

janicduplessis:
Buck wasn't compiling using sdk 29 like gradle. Updated https://github.com/facebook/react-native/pull/26433/commits/7e4a71b2b140b2e640d05983b1673f5a8f4d7976 and it seems to build fine on CI now. We probably want to land this in a separate commit and idk if this will affect fb.

janicduplessis:
cc @mdvacca @dulmandakh

dulmandakh:
LGTM. I would bump target to android-29.

mdvacca:
LGTM, importing to review and test internally

janicduplessis:
ping @mdvacca

janicduplessis:
e7a14b803fdc8840bbcde51d4bfa9cf9a85a8472

PR #26451, opened 2019-09-16
Bump hermes to v0.2.1

sunnylqm:
## Summary

Bump hermes to v0.2.1 

## Changelog

See https://github.com/facebook/hermes/releases/tag/v0.2.1

[Android] [Changed] - Bump hermes to v0.2.1

## Test Plan

RNTester builds and runs as expected

### DIFF
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -151,12 +151,6 @@ android {
             signingConfig signingConfigs.release
         }
     }
-    packagingOptions {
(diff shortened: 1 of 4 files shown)


sunnylqm:
@cpojer @willholen  According to the release note, maybe we need to remove `pickFirst`?

willholen:
You definitely could. @Kudo , what do you think?

Kudo:
Yes, this is the intention to separate C++ runtime from single AAR.
Please remove those `pickFirst`.

PR #26466, opened 2019-09-16
Bump flipper version to fix build

janicduplessis:
## Summary

The Flipper version in the podfile doesn't include the podspec for the React plugin, it was added in version 0.23.7.

Bumped the version and ran `bundle exec pod install`

Fixes the CI failure in iOS tests

image

## Changelog

[iOS] [Fixed] - Bump flipper version to fix build

## Test Plan

Build on iOS

### DIFF
--- a/RNTester/Podfile
+++ b/RNTester/Podfile
@@ -9,7 +9,7 @@ end
 
 # Add Flipper Poods
 def flipper_pods()
-  flipperkit_version = '0.23.4'
+  flipperkit_version = '0.23.7'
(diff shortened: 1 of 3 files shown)


janicduplessis:
cc @axe-fb

PeteTheHeat:
Flipper in RNTester is causing some problems internally. We've decided to remove the Flipper deps from RNTester for now, which should solve this build issue, as well as some others we're facing. I'm landing that removal right now.

PR #26469, opened 2019-09-17
AndroidTextInputNativeComponent: Make event properties ReadOnly

empyrical:
## Summary

This pull request makes properties of events' Flow types in `AndroidTextInputNativeComponent` be `$ReadOnly`.

This will make them more compatible with the callback types in `TextInput`.

## Changelog

[Internal] [Changed] - Made properties of events' Flow types in `AndroidTextInputNativeComponent` readonly

## Test Plan

`yarn flow-check-ios` and `yarn flow-check-android` both pass.

No regressions to running `scripts/generate-rncore.sh` have been noted.

### DIFF
--- a/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js
+++ b/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js
@@ -324,8 +324,7 @@ export type NativeProps = $ReadOnly<{|
   onContentSizeChange?: ?DirectEventHandler<
     $ReadOnly<{|
       target: Int32,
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks! Can you run the codegen script as part of your test plan? Just checking that flow passes isn’t sufficient to know that the codegen supports these readonlys.

empyrical:
That script needs some modification to work (missing the positional argument "module spec name" to `generate-tests.js`) but I notice no difference in behaviour before or after this modification to `AndroidTextInputNativeComponent`.

PR #26471, opened 2019-09-17
Remove ignored iOS schemes from template

friederbluemle:
## Summary

The merge of #25451 added `xcshareddata` to the .gitignore file in `template`. Two xcscheme files in the `xcschemes` subfolder were still left in the template and should have been removed at the same time.

The project opens and builds fine without the xcscheme files both from command line and in Xcode.

## Changelog

[iOS] [Changed] - Remove ignored iOS schemes from template

## Test Plan

Generate new project, remove files, verify that project builds, and opens correctly in Xcode.

Please review @karanjthakkar @satya164 @kelset


kelset:
cc @cpojer I think that I saw that in 0.61 we are cleaning up the xcodeproj files so prob it would be good to have this merged too

PR #26474, opened 2019-09-17
Fix onPress prop for Touchable Components being called twice on Android Tv

dbarr33:
## Summary

Due to an update to react-native on the  android tv platform tapping the select button on a remote calls the onPress prop twice for `TouchableHighlight`, `TouchableOpacity`, and `TouchableWithoutFeedback`. This is happening because touchableHandlePress gets called from two places. First from the onClick prop in the touchable component and second from the TVEventHandler in the TouchableMixin.

## Changelog

[Android] [Fixed] - Adds a not android check to the select case of the TVEventHandler callback in the TouchableMixin.

## Test Plan
Confirmed on Android Tv and Apple Tv

1) Add a TouchableOpacity to a screen with an onPress callback
2) Run app
3) Focus the TouchableOpacity
4) Press the Select Button on the Remote

**Expected Results**
onPress is called once

### DIFF
--- a/Libraries/Components/Touchable/Touchable.js
+++ b/Libraries/Components/Touchable/Touchable.js
@@ -380,7 +380,7 @@ const TouchableMixin = {
           cmp.touchableHandleFocus(evt);
         } else if (evt.eventType === 'blur') {
           cmp.touchableHandleBlur(evt);
-        } else if (evt.eventType === 'select') {
+        } else if (evt.eventType === 'select' && Platform.OS !== 'android') {
           cmp.touchableHandlePress &&
             !cmp.props.disabled &&
             cmp.touchableHandlePress(evt);


TheSavior:
Adding a check for android here feels like it probably isn't the right fix. You mention that the issue is from "the onClick prop in the touchable component and second from the TVEventHandler in the TouchableMixin". Should this fix instead be a part of TVEventHandler?

dbarr33:
@TheSavior I thought about updating TVEventHandler to not fire on android when the select button is pressed but I was afraid that would break other places that expect TVEventHandler trigger when select is pressed.

TheSavior:
I’m a little nervous to merge this as is. Since this problem only occurs with tv, making a change to the code that effects all platforms feels a little strange to me. Also, having the solution be a check for Android when the problem impacts android tv but not android mobile also seems a bit scary. 

Can you help me understand this change a bit more, why this is safe for all platforms,  and why this change shouldn’t live in tv specific code?

dbarr33:
@TheSavior There is a platform check with an early return if the platform isn't tv above the code that this pr is modifying. So this code will only run on tv platforms.

dblachut-adb:
> First from the onClick prop in the touchable component 
@dbarr33 Could you please point out the firs place where the `onPress` is called? What is the call stack of the `onClick` call? Is it called from native somewhere? Thank you in advance.

Issue #26473, opened 2019-09-17
Modal's onDismiss callback is wrongly implemented in 0.61.0-rc.3


wddwycc:
Issue:

With [This commit](https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43), `onDismiss` on `Modal` will never be called when dismissed, code below as an example.

```jsx
const App = () => {
  const [visible, setVisible] = useState(true)
  const onClosePress = useCallback(() => setVisible(false), [])
  const onDismiss = useCallback(() => { console.log('dismiss') }, [])
  return (
    <View>
      <TouchableOpacity onPress={onClosePress}>
        <Text>Close</Text>
      </TouchableOpacity>
      <Modal visible={visible} onDismiss={onDismiss} />
    </View>
  )
}
```

This is not the intended behavior of this callback according to [the document](https://facebook.github.io/react-native/docs/modal#ondismiss)

---

React Native version:

System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 1.32 GB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.9.0 - /usr/local/bin/node
    npm: 6.2.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
  IDEs:
    Xcode: 11.0/11A419c - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.0-rc.3 => 0.61.0-rc.3

xiao99xiao:
The commit https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43 is a disaster. It actually introduced the problem which it claims to solve in its summary.

```
# Problem:
`onDismiss` prop isn't being called once the modal is dismissed, this diff fixes it.
```

bartolkaruza:
Fix for this issue is on the way, I'll try to get it cherry-picked in the next RC for 0.61 as well. https://github.com/facebook/react-native/pull/26490
Thanks for reporting this!

wddwycc:
[...]

PR #26490, opened 2019-09-18
fix(modal): fix onDismiss logic for the modal component on iOS

bartolkaruza:
## Summary

This PR fixes https://github.com/facebook/react-native/issues/26473 which was introduced by https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43

Basically ```componentWillUnmount``` for a modal is never called because it internally returns 'null' and remains mounted on iOS. For Android, this may have worked because of empty nodes being removed by the OS from the view hierarchy, but I'm not sure.

## Changelog

[iOS] [Fixed] - fix onDismiss logic for the modal component on iOS

## Test Plan

Before this change onDismiss is never called. After the change it is called on prop change `visible`  from true to false.

```
const App = () => {
  const [isOpen, setOpen] = useState(true);

  function onDismiss() {
    console.log('dismissed');
  }

  return (
    <>
      <StatusBar barStyle="dark-content" />
      <SafeAreaView>
        <Modal visible={isOpen} animationType="slide" onDismiss={onDismiss}>
          <TouchableOpacity onPress={() => setOpen(false)}>
            <View style={{height: 500, backgroundColor: 'blue'}}>
              <Text>1</Text>
            </View>
          </TouchableOpacity>
        </Modal>
      </SafeAreaView>
    </>
  );
};
```

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -169,13 +169,10 @@ class Modal extends React.Component<Props> {
     };
   }
 
-  componentWillUnmount() {
-    if (this.props.onDismiss != null) {
+  UNSAFE_componentWillReceiveProps(nextProps: Props) {
+    if (!nextProps.visible && this.props.visible && this.props.onDismiss) {
       this.props.onDismiss();
     }
-  }
-
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
     Modal._confirmProps(nextProps);
   }
 


xiao99xiao:
I guess we need to test it on Android too, since https://github.com/facebook/react-native/issues/26473 also exists on Android, at least for me.

wddwycc:
This callback is initially designed to represent `UIViewController.dismiss`'s onCompletion on iOS. If it is changed as the timing when visible is changed from `true` to `false`, why we even need this callback, we can just inspect visible itself.

I'd prefer we revert [this commit](https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43) and keep everything as before.

wddwycc:
Because iOS won't allow you to present any `UIViewController` when dismissing `UIViewController`, so this callback is the only reliable timing we can present the next `UIViewController` and dismissing the modal. [This commit](https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43) breaks it all.

bartolkaruza:
> This callback is initially designed to represent UIViewController.dismiss's onCompletion on iOS. If it is changed as the timing when visible is changed from true to false, why we even need this callback, we can just inspect visible itself.

Is the timing actually different?

neilkimmett:
Yep that commit broke some of our flows too, because we rely on the `UIViewController` being dismissed before presenting another. If we could revert that would be amazing.

bartolkaruza:
So the commit has been reverted by @grabbou and picked for the next 0.61 RC which should solve all the immediate problems for you.

https://github.com/facebook/react-native/commit/a5353c0361f330b2b8aebc5de8dc2d006887081b

Maybe @sammy-SC could chime in here for where you wanted to take this change? I think a working solution for both iOS and Android (keeping the mechanism in tact for iOS, as it is right now) should be easy to do.

neilkimmett:
Amazing, thanks for the quick fix @bartolkaruza @grabbou, really appreciate it.

bartolkaruza:
The fix is in this PR causes new issues that are worse than the original issue. A new solution to this problem needs to be found, in another PR.

Issue #25704, opened 2019-07-18
Blank screen for Android. RN Version 60.0.3


vyshakh:
Getting blank screen for android on Mac OS.  Looks fine on Windows OS.

OS: macOS 10.14.4
React Native version:
react: 16.8.6 => 16.8.6 
react-native: 0.60.3 => 0.60.3 

## Steps To Reproduce

1. react-native init testapp
2. react-native run-android

Expected screen
New intro screen

**Workaround**
Run **react-native start** before running **react-native run-android**

iamacup:
does running react-native start followed by react-native run-android fix this issue for you?

vyshakh:
> does running react-native start followed by react-native run-android fix this issue for you?

@iamacup Yup. It works.  Thank you 👍
So another workaround is 
Run **react-native start** before running **react-native run-android**

iamacup:
This is todo with some behavior change where when we run `react-native run-android` on mac after 0.60.0 the bundler does not start automatically - and then if you start the bundler AFTER running `react-native start`, no matter how many times you reload the app it will not move on from white screen.

I do not know if this is intended behavior or a bug, but it did make me pull my hair out for a while ;-)

hannanstd:
hi
i have this problem. i use UBUNTU. when i use react-native run-android i will get red page. and then i use react-native start i will get blank white screen.

williamprt:
Hi
The same for me. I use Ubuntu and tried run react-native on a physical device and Genymotion android emulator. Nothing works

alitsvin:
I'm faced this issue on Linux Mint.
Run `react-native start` before `react-native run-android` helped

danielnora:
This happens to me as well on Ubuntu 19.04.
On a brand new project:
**react-native info**
```
Binaries:
    Node: 10.16.2 - ~/.nvm/versions/node/v10.16.2/bin/node
    Yarn: 1.17.3 - /usr/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.2/bin/npm
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.5 => 0.60.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1
```
`react-native run-android` apparently starts the Metro bundler and, after installing the app on the phone, the Metro server is closed and the app displays the message:
```
Unable to load script. Make sure you are either running a Metro server or that your bundle 'index.android.bundle' is packaged correctly for release.
[...]

PR #26495, opened 2019-09-18
Allow again for injecting custom root view via ReactActivityDelegate

kmagiera:
## Summary

This change restores the possibility of injecting custom root views via ReactAcitivtyDelegate. It has been used by react-native-gesture-handler library in order to replace default root view with a one that'd route touch events to gesture-handler internal pipeline.

The regression happened in d0792d4b8ac42711dfd9fccb782f16e72ce3e335 where new `ReactDelegate` was introduced to provide support for rendering react native views in both Android fragments and activities. As a part of that change the logic responsible for creating root view has been moved from `ReactActivityDelegate` to `ReactDelegate` rendering `ReactActivityDelegate.createRootView` unused – that is there is no code path that leads to this method being called. Instead `ReactDelegate.createRootView` method has been added which now plays the same role. The custom root view injection was relying on overwriting that method and hence the method is no longer referenced overwriting it has no effect. Following the logic migration out of `ReactActivityDelegate` into `ReactDelegate` we could potentially now start overwriting methods of `ReactDelegate`. However when working with Android's activities in React Native we can only provide an instance of `ReactActivityDelegate` and in my opinion it does not make too much sense to expose also a way to provide own instance of `ReactDelegate`.

The proposed fix was to route `ReactDelegate.createRootView` to call `ReactActivityDelegate.createRootView` and this way regaining control over root view creation to `ReactActivityDelgate`. The change of the behavior has been implemented by subclassing `ReactDelegate` directly from `ReactActivityDelegate` and hooking the aforementioned methods together. Thanks to this approach, the change has no effect on `ReactDelegate` being used directly from fragments or in other scenarios where it is not being instantiated from `ReactActivityDelegate`.

This fixes an issue reported in https://github.com/kmagiera/react-native-gesture-handler/issues/745 and discussed on 0.61 release thread: https://github.com/react-native-community/releases/issues/140#issuecomment-532235945

## Changelog

[Internal] [Fixed] - Allow for custom root view to be injected via ReactActivityDelegate

## Test Plan

1. Run RNTester, take layout snapshot, see the react root view being on the top of view hierarchy. 
2. Run gesture-handler Example app (or some other app that overwrites ReactActivityDelegate.createRootView method), on layout snapshot see custom root view being used.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java
@@ -47,7 +47,7 @@ public class ReactActivityDelegate {
   }
 
   protected ReactRootView createRootView() {
-    return mReactDelegate.createRootView();
+    return new ReactRootView(getContext());
   }
 
   /**
@@ -73,7 +73,12 @@ public class ReactActivityDelegate {
     String mainComponentName = getMainComponentName();
     mReactDelegate =
         new ReactDelegate(
-            getPlainActivity(), getReactNativeHost(), mainComponentName, getLaunchOptions());
+            getPlainActivity(), getReactNativeHost(), mainComponentName, getLaunchOptions()) {
+          @Override
+          protected ReactRootView createRootView() {
+            return ReactActivityDelegate.this.createRootView();
+          }
+        };
     if (mMainComponentName != null) {
       loadApp(mainComponentName);
     }


flyskywhy:
This commit will cause Blank screen for Android after just `react-native run-android`as described in https://github.com/facebook/react-native/issues/25704 , it will be nice if you improve the compatibility between this commit and the Metro server :stuck_out_tongue:

PR #26497, opened 2019-09-19
cleanup DrawerLayoutAndroid

dulmandakh:
## Summary

This PR removes Java reflection use in ReactDrawerLayoutManager.java because we all use AndroidX and setDrawerElevation is available. Also adds NonNull annotations

## Changelog

[Android] [Changed] - cleanup DrawerLayoutAndroid

## Test Plan

RNTester is working as expected.


TheSavior:
cc @makovkastar since you have a diff changing this file too

PR #26503, opened 2019-09-19
Bump SoLoader to v0.8.0

sweggersen:
## Summary

Use latest version of SoLoader

## Changelog

See https://github.com/facebook/SoLoader/releases/tag/v0.8.0


sweggersen:
We shouldn't bump this now, as I've looked at recent user data with this change on: https://github.com/facebook/SoLoader/issues/47

PR #26556, opened 2019-09-25
fix build with hermes on windows

B27:
## Summary

On the Windows platform, with hermes-engine enabled, the assembly crashes with an error:
image
The problem lies in calling hermes command without the leading arguments `"cmd", "/c"` ([react.gradle, line: 152](https://github.com/facebook/react-native/blob/e028ac7af2d5b48860f01055f3bbacf91f6b6956/react.gradle#L152) )

## Changelog

[General] [Fixed] - Added a platform check and running commandLine with the corresponding arguments

## Test Plan

Under Windows, enable hermes-engine in _build.gradle_ and run the `gradlew assembleRelease` or `gradlew bundleRelease` command
Also check assembly on other available platforms

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -149,7 +149,12 @@ afterEvaluate {
                             hermesFlags = config.hermesFlagsDebug
                             if (hermesFlags == null) hermesFlags = []
                         }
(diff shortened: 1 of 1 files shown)


zenz:
this was commited, but don't know why it is not in the release: https://github.com/facebook/react-native/commit/ee82182a70e5aa1841ca2418d0798c26c3ba93a1

Issue #26541, opened 2019-09-23
(Feature request) ACCESS_BACKGROUND_LOCATION in PermissionsAndroid


Sakarisson:
Since Android 10, the `ACCESS_BACKGROUND_LOCATION` permission is now necessary to track a user's location while the app is in the background.

https://developer.android.com/about/versions/10/privacy/changes

Similarly to iOS, Android users now have the option to allow location access only when the app is active. 

It would be very useful to be able to access this data from `PermissionsAndroid`. I can create the PR myself, but I feel like it's best to create an issue first.

PR #26562, opened 2019-09-25
add ACCESS_BACKGROUND_LOCATION to PermissionsAndroid

dulmandakh:
## Summary

This PR adds ACCESS_BACKGROUND_LOCATION to PermissionsAndroid. Fixes https://github.com/facebook/react-native/issues/26541

## Changelog

[Android] [Changed] - add ACCESS_BACKGROUND_LOCATION to PermissionsAndroid

## Test Plan

```javascript
PermissionsAndroid.ACCESS_BACKGROUND_LOCATION === 'android.permission.ACCESS_BACKGROUND_LOCATION'
```


mikehardy:
for this reason I use react-native-permissions - it's up to date more or less, and unifies (as much as possible) the ios vs android permissions request experience, if you haven't seen it...

thymikee:
Maybe it's worth to deprecate this module in favor of the community one? cc @cpojer

mikehardy:
I think that module is another that is largely single-maintainer and just today released 2.0.0 (meaning: will likely need some more effort short-term) so it might be nice to offer help at the same time, but I do like that module and found it more complete than the built-in APIs, $0.02)

PR #26602, opened 2019-09-26
Resolve signature issue causing crashes on iOS 13 under RN version 0.56.1

zachary95:
## Summary

Tried to build an app under RN 0.56.1 but it crashed immediately. A fix was provided on recent version with https://github.com/facebook/react-native/pull/25146, I'm just bringing this fix to 0.56.

## Changelog

[iOS] [Fixed] - Resolve signature issue causing crashes on iOS 13 under RN version 0.56.1.

## Test Plan

Just built an app and it worked. No more crashes!


cpojer:
Thank you for your PR! We are not actively maintaining stable branches that are more than a year old. In case you cannot upgrade the version of React Native you are using I recommend forking from the stable point and building RN from source.

Issue #26181, opened 2019-08-25
[Android] Regression after 0.59: FlatList in inverted mode on Android doesn't invert pull-to-refresh gesture.


avinidiktov:
## Description:
When inverted prop is set to true for FlatList on Android pulling from bottom to top does nothing. But pulling from top to bottom causing refresh indicator to show up on top and calling onRefresh.

iOS works as expected.

The problem doesn't exist for React Native 0.59.8.

## React Native version:
System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 1.80 GB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.16.0 - /usr/local/opt/node@10/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/opt/node@10/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 27, 28
      Build Tools: 27.0.3, 28.0.3
      System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.5 => 0.60.5
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce

1. Use FlatList on Android;
2. Set inverted prop;
2. Pull the list up. The indicator will not be displayed;
3. Pull the list down. The indicator will be displayed at the top of the screen.

## Describe what you expected to happen:
The indicator must be displayed at the bottom of the screen.

## Snack, code example, screenshot, or link to a repository:
**The same code works differently.**

Wrong behavior demo (react-native: 0.60.5):
![wrong](https://user-images.githubusercontent.com/15199362/63640273-4e3ead00-c6c8-11e9-87a1-bc2e4dd08dc0.gif)
https://github.com/avinidiktov/flatlist-bug

Expected behavior demo (react-native: 0.59.8):
[...]

PR #26611, opened 2019-09-27
Include transform in OUTER_PROPS

migbot:
## Summary
Without `transform` in `OUTER_PROPS`, the refresh control component would not include `transform: {scaleY: -1}` in its style and so pulling down, rather than up, on a scroll view would trigger a refresh.

Fixes https://github.com/facebook/react-native/issues/26181

## Changelog
[Android] [Fixed] - Fixed issue with refresh control not working properly on an inverted ScrollView

## Test Plan
Updated unit test in splitLayoutProps-test.js.

### DIFF
--- a/Libraries/StyleSheet/__tests__/splitLayoutProps-test.js
+++ b/Libraries/StyleSheet/__tests__/splitLayoutProps-test.js
@@ -13,11 +13,14 @@
 const splitLayoutProps = require('../splitLayoutProps');
 
 test('splits style objects', () => {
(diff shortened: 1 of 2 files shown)


xxRockOnxx:
is this actually the intended behavior of `inverted`?

I've been using it and had to upgrade to `0.61.2` this day for a fix i needed but now my Chat screen when loading earlier messages shows the RefreshControl at the bottom.

I was expecting it to be the "wrong behavior" of #26181. Is there anyway to opt-out at least?

PR #26624, opened 2019-09-27
Announce accessibility state changes happening in the background

xuelgong:
## Summary
Currently the react native framework doesn't handle the accessibility state changes of the focused item that happen not upon double tapping. Screen reader doesn't get notified when the state of the focused item changes in the background.
To fix this problem, post a layout change notification for every state changes on iOS.
On Android, send a click event whenever state "checked", "selected" or "disabled" changes. In the case that such states changes upon user's clicking, the duplicated click event will be skipped by Talkback.

## Changelog
[General][Fixed] - Announce accessibility state changes happening in the background

## Test Plan

Add a nested checkbox example which state changes after a delay in the AccessibilityExample.


RNTester/js/examples/Accessibility/AccessibilityExample.js
```diff
@@ -314,6 +298,114 @@ class ExpandableElementExample extends React.Component {
   }
 }
 
+class NestedCheckBox extends React.Component {
```
blavalla: While this example is functioning how you intended it to, I found it pretty confusing at first why it was checking/unchecking multiple boxes when I selected one. I don't see how adding these additional checkboxes helps to showcase this feature either, as it could easily be shown with a single checkbox that gets checked on a delay.

blavalla:
Overall the implementation looks good, but I added a few comments inline that I think will help clarify exactly what is being added here.

RNTester/js/examples/Accessibility/AccessibilityExample.js
xuelgong: This example mimics ordering a pizza. When "meat" is chosen, all meats are checked. If only "beef" or "bacon" is selected, the state of the checkbox "meat" becomes "mixed". The reason I created nested checkboxes is to make sure that the state change of the unfocused item won't be announced. That is, if all meats are checked previously and then deselect beef, only the state change of "beef" checkbox will be announced. Even though the "meat" checkbox state also changes, screen reader shouldn't announce it.

ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -170,6 +171,11 @@ public void setViewState(@NonNull T view, @Nullable ReadableMap accessibilitySta
               && accessibilityState.getType(STATE_CHECKED) == ReadableType.String)) {
         updateViewContentDescription(view);
         break;
+      } else if (view.isAccessibilityFocused()) {
+        // Send a click event to make sure Talkback get notified for the state
+        // changes that don't happen upon users' click. For the state changes
+        // that happens immediatly, Talkback will skip the duplicated click event.
+        view.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_CLICKED);
```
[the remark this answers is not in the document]
xuelgong: Will do.

Issue #26577, opened 2019-09-25
View within text returns strange image - iOS 13


djorkaeffalexandre:
React Native version: 0.61.1 | 0.60.5

## Steps To Reproduce

1. Any `<View>` within `<Text>` generate a weird behavior..
```
<Text>
  <View style={{width: 200, height: 200}} />
</Text>
```

2. Same to `<Text><Image /></Text>` like a:
```
<Text>
  <Image
    style={{ width: 30, height: 30 }}
    source={{
      uri:
      'https://upload.wikimedia.org/wikipedia/commons/5/58/Ponto_Amarelo.png',
    }}
   />
</Text>
```

It returns a strange image behind the image after upgrade to iOS 13.

### Screenshots
![Captura de Tela 2019-09-25 às 16 23 54](https://user-images.githubusercontent.com/29778115/65632871-1a012980-dfb1-11e9-9976-cf62d28789ca.png)

KarenQiao:
yes ,i also have this problem

partiellkorrekt:
I searched a little through the react-native code and found that the strange image is the placeholder icon for an attachment in an NSAttributedString. RN seems to add attachments to the string to reserve some space and lay the real view on top of the rendered string.

In RCTTextShadowView.m there is a function "attributedTextWithMeasuredAttachmentsThatFitSize" which creates the attachments by using the following code:
```
NSTextAttachment *attachment = [NSTextAttachment new];
attachment.bounds = (CGRect){CGPointZero, fittingSize};
[attributedText addAttribute:NSAttachmentAttributeName value:attachment range:range];
```

To verify that the image really is the placeholder image, I modified the code to add an empty image as the attachment image:
```
NSTextAttachment *attachment = [NSTextAttachment new];
attachment.bounds = (CGRect){CGPointZero, fittingSize};

// This creates an empty image on the fly (just for testing purposes)
UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), NO, 0);
UIImage *img = UIGraphicsGetImageFromCurrentImageContext();
UIGraphicsEndImageContext();

// Set this empty image as the attachment image
[attachment setImage:img];
      
[attributedText addAttribute:NSAttachmentAttributeName value:attachment range:range];
```

Using this code, RN stops rendering the document icons and the output looks as expected.

**Warning**
This code might create a lot of unnecessary images and I don't know how it will affect memory usage. So don't use it in production.

partiellkorrekt:
[...]

PR #26653, opened 2019-09-30
iOS: Possible fix for #26577

partiellkorrekt:
See https://github.com/facebook/react-native/issues/26577 for the idea behind this change

## Summary

When rendering a view nested inside a text on iOS 13 there will be a document icon behind this view. That's because RN uses attachments to make the text flow around the child-views.

This code replaces the attachment image with an empty image so that the rendered output looks like expected.

## Changelog

[iOS] [Fixed] - Remove the document-icon behind childviews of text

## Test Plan

See https://github.com/facebook/react-native/issues/26577 for an example where the old code would render incorrectly. The new code renders just like it's expected.

I also added many of childviews to different texts and added NSLog-statements to verify that only one empty placeholder-image gets created.

Since I'm not normally developing in Objective-C it might be a good idea for someone else to have a look at it and see if this is a good way to do it.

### DIFF
--- a/Libraries/Text/Text/RCTTextShadowView.m
+++ b/Libraries/Text/Text/RCTTextShadowView.m
@@ -175,6 +175,17 @@
                          range:NSMakeRange(0, attributedText.length)];
 }
 
+static UIImage *emptyImg = NULL;
+- (UIImage*)getEmptyImg
+{
+  if (emptyImg == NULL) {
+    UIGraphicsBeginImageContextWithOptions(CGSizeMake(1, 1), NO, 0);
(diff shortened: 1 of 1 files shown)


kristerkari:
I ran the code from this PR using [patch-pacakge](https://github.com/ds300/patch-package)  agaist the test/repro repository that I created. 

It does fix the iOS 13 bug:
https://github.com/kristerkari/react-native-ios-13-view-or-image-inside-text-bug/pull/1

PeteTheHeat:
Thanks for submitting this! I ended up using a similar strategy and fixing this bug in 

https://github.com/facebook/react-native/commit/06599b3e594355a1d5062ede049ff3e333285516

Issue #25265, opened 2019-06-14
Changing TextInput selection after setting text crash


Vaintti:
When TextInput selection is updated after changing the text (using InteractionManager.runAfterInteractions to wait for the change to apply) RN crashes. You get error like "setSpan(4 ... 4) ends beyond length" which indicates that the text field text has not changed like you would expect when using runAfterInteractions.
React Native version:

info
  React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
      Memory: 34.23 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
      Yarn: 1.12.3 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.0, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.3
        System Images: android-25 | Google Play Intel x86 Atom, android-26 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.9 => 0.59.9
## Steps To Reproduce

1. Change controlled TextInput value in code
2. Change TextInput selection to be after the inserted text using InteractionManager.runAfterInteractions

or just run the example in repo I linked and press the button.

Describe what you expected to happen:
Selection changes to place after inserted text

Snack, code example, or link to a repository:
[https://github.com/Vaintti/RNTextInputBug](url)

blohamen:
Have similar issue on RN 0.60.4

blohamen:
Issue still reproduced on React Native 0.61.2

callcter:
same issue in 0.61.2

cayodonatti:
[...]

PR #22723, opened 2018-12-20
Fix crash if set text and set selection at the same time.

shenyute:
Since text and selection has dependency, handle text selection in
updateExtraData as well.

The root cause is due to setText is handled on extra data update but setSelection is handled on set property. And extra data update will be handled after all properties are handled. Since selection and text has dependency, move selection to extra data update as well.

Changelog:
----------
[Android] [Fixed] - Fix crash when set text and selection on textinput at the same time

Test Plan:
----------
input.setNativeProps({ text: "xxx", selection: {"begin": 0, "end": 3}});

PR #26680, opened 2019-10-01
Cap selection indices when text changes

MarcoPolo:
## Summary
This PR https://github.com/facebook/react-native/pull/22723 cached selections, so if you had a cached selection indicies, but updated the text to be an empty string, then this would crash.

As reported in https://github.com/facebook/react-native/issues/25265 and other issues of `setSpan(4 ... 4) ends beyond length`

## Changelog

[Android] [fixed] - Crash in TextInput

## Test Plan
```
input.setNativeProps({ text: "xxx", selection: {"begin": 0, "end": 3}});
input.setNativeProps({ text: ""});
```


vforvasile:
hey @cpojer! 
When can we use this? I tried 0.61.5 and it's still an issue.

Anjy123:
I'm also facing same issue..

PR #26689, opened 2019-10-02
Cpp crash branch do not merge

emmajam:
[TEST BRANCH]
[DO NOT MERGE]

Created cpp crash for testing NDK symbolication in Crashlytics.


cpojer:
I don't think this PR is the best possible way to point out an issue, especially since it isn't properly based on master or a stable release. I recommend having this discussion in an issue with links to various branches.

PR #26690, opened 2019-10-02
Hornet/cpp crash [DO NOT MERGE]

emmajam:
DO NOT MERGE
CPP CRASH TEST


ecreeth:
:thinking: What are the differences between this PR and #26689

PR #26709, opened 2019-10-03
refactor prop type of TouchableWithoutFeedback

ma96o:
## Summary

I refactored prop type of `TouchableWithoutFeedback`

## Changelog

[Internal] [Fixed] - refactor prop type of TouchableWithoutFeedback


cpojer:
Thanks! As a convention we always use `Props` as a type and recommend people to import those in their apps and rename them locally. This makes it easier to guess for each file which prop name to use.

ma96o:
Thanks for your review!
I learned it💪

PR #26711, opened 2019-10-03
Moved babel runtime to dependencies

jdmg94:
## Summary
Moved babel's runtime dependency to regular `dependencies` as described in [babel's official docs](https://babeljs.io/docs/en/babel-runtime#usage) and [NPM page](https://www.npmjs.com/package/@babel/runtime)

## Changelog

[internal] [changed] - `@babel/runtime` from `devDependencies` to `dependencies`

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -10,11 +10,11 @@
   },
   "dependencies": {
     "react": "16.9.0",
+    "@babel/runtime": "^7.6.2",
     "react-native": "1000.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


cpojer:
This is not necessary for React Native because Metro bundles all dependencies up into a single bundle for production releases. Because React Native apps aren't deployed to npm as packages but to App Stores as binaries, we do not need to change this.

jdmg94:
@cpojer say I have jest, eslint, etc as devDependencies, are those bundled up as well? just curious

PR #26715, opened 2019-10-04
Added missing metro dev dependency

saketkumar:
## Summary
Added the Metro dev dependency which was missing earlier.

## Changelog

[Internal] [Fixed] - Added the Metro dev dependency which was missing earlier.


cpojer:
Metro is not a dev dependency of React Native. Metro is part of the cli so it will already be part of the dependency tree.

PR #26730, opened 2019-10-05
Replace folly::make_unique with std::make_unique

JoeLoser:
## Summary

There is a mixed usage of `folly::make_unique` and `std::make_unique`. Soon, `folly::make_unique` may be removed (see [this PR](https://github.com/facebook/folly/pull/1150)). Since `react-native` only supports C++14-compilers and later, switch to always using `std::make_unique`.

## Changelog

[Internal] [Removed] - Replace folly::make_unique with std::make_unique

## Test Plan

Running the existing test suite. No change in behavior is expected.


JoshuaGross:
Thanks for the PR @JoeLoser!

PR #26769, opened 2019-10-08
Android template: Allow overriding default "index.js" entry file

rumax:
## Summary

- Using "System.getenv" allows to specify any entry file using environment variables and without modifying gradle file. Example:

    export ENTRY_FILE="another_entry_file.js"
    ./gradlew assembleDebug

- This functionality is also more align with iOS implementation that uses 'if [[ "$ENTRY_FILE" ]]; then'. See https://github.com/facebook/react-native/pull/23667 for more details.

- Possibility to define entry file on CI without modifying sources (Example: project like [pixels-catcher](https://www.npmjs.com/package/pixels-catcher) requires different entry file)

## Changelog

[Android] [Added]  - Custom entry file on android using `ENTRY_FILE` environment variable

## Test Plan

- Create a project from template

- Define `ENTRY_FILE` environment variable

```
export ENTRY_FILE="anotherIndexFile.js"
```

- Build android 

```
./gradlew assembleDebug
```

Expected result: App contains bundle file that starts from `anotherIndexFile.js` file.

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -7,10 +7,22 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 def config = project.hasProperty("react") ? project.react : [];
 
+def detectEntryFile(config) {
+    if (System.getenv('ENTRY_FILE')) {
+        return System.getenv('ENTRY_FILE')
+    } else if (config.entryFile) {
+        return config.entryFile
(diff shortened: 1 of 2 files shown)


dulmandakh:
I think it's best to do it in react.build file.

rumax:
> I think it's best to do it in react.build file.

Good point, I'll update PR with this change.

PR #26773, opened 2019-10-09
[android][gradle] Use Node's module resolution algo to find JSC & Hermes

ide:
## Summary

The Gradle build file looks up jsc-android and hermes-engine using hard-coded paths. Rather than assuming the location of these packages, which are distributed and installed as npm packages, this commit makes the Gradle file use Node's standard module resolution algorithm. It looks up the file hierarchy until it finds a matching npm package or reaches the root directory.

## Changelog

[Android] [Changed] - ReactAndroid's Gradle file uses Node's module resolution algorithm to find JSC & Hermes

## Test Plan

Ensure that CI tests pass, and that `./gradlew :ReactAndroid:installArchives` works. Printed out the paths that the Gradle script found for jsc-android and hermes-engine (both were `<my stuff>/react-native/node_modules/jsc-android|hermes-engine`).

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -94,23 +94,16 @@ task prepareFolly(dependsOn: dependenciesPath ? [] : [downloadFolly], type: Copy
 }
 
 task prepareHermes() {
-    def hermesAAR = null
     def hermesPackagePath = findNodeModulePath(projectDir, "hermes-engine")
-    if (hermesPackagePath) {
-        hermesAAR = file("$hermesPackagePath/android/hermes-debug.aar")
-        if (!hermesAAR.exists()) {
-            throw new GradleScriptException("The hermes-engine npm package is missing \"android/hermes-debug.aar\"")
-        }
-    } else {
-        // For an app to build from RN source, hermes-engine is located at /path/to/app/node_modules
-        // and $projectDir is located at /path/to/app/node_modules/react-native/ReactAndroid
-        hermesAAR = file("$projectDir/../../hermes-engine/android/hermes-debug.aar")
-
-        if (!hermesAAR.exists()) {
-            // At Facebook, this file is in a different folder
-            hermesAAR = file("$projectDir/../../node_modules/hermes-engine/android/hermes-debug.aar")
-        }
+    if (!hermesPackagePath) {
+        throw new GradleScriptException("Could not find the hermes-engine npm package")
     }
+
+    def hermesAAR = file("$hermesPackagePath/android/hermes-debug.aar")
+    if (!hermesAAR.exists()) {
+        throw new GradleScriptException("The hermes-engine npm package is missing \"android/hermes-debug.aar\"")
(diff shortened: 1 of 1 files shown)


ReactAndroid/build.gradle
```diff
@@ -92,8 +94,14 @@ task prepareFolly(dependsOn: dependenciesPath ? [] : [downloadFolly], type: Copy
...
-    def hermesAAR = file("$projectDir/../node_modules/hermes-engine/android/hermes-debug.aar")
-    if (!hermesAAR.exists()) {
+    def hermesAAR = null
+    def hermesPackagePath = findNodeModulePath(projectDir, "hermes-engine")
+    if (hermesPackagePath) {
+        hermesAAR = file("$hermesPackagePath/android/hermes-debug.aar")
+        if (!hermesAAR.exists()) {
+            throw new GradleScriptException("The hermes-engine npm package is missing \"android/hermes-debug.aar\"")
+        }
+    } else {
```
janicduplessis: I don’t think we need this else case anymore since it should get picked by the node algorithm.

ReactAndroid/build.gradle
```diff
@@ -160,16 +168,22 @@ task prepareGlog(dependsOn: dependenciesPath ? [] : [downloadGlog], type: Copy)
...
-        def jscPackageRoot = file("$projectDir/../node_modules/jsc-android/dist")
-        if (!jscPackageRoot.exists()) {
+        def jscDist = null
+        def jscPackagePath = findNodeModulePath(projectDir, "jsc-android")
+        if (jscPackagePath) {
+            jscDist = file("$jscPackagePath/dist")
+            if (!jscDist.exists()) {
+                throw new GradleScriptException("The jsc-android npm package is missing its \"dist\" directory")
+            }
+        } else {
```
janicduplessis: Same here

ReactAndroid/build.gradle
```diff
@@ -192,6 +206,20 @@ task downloadNdkBuildDependencies {
     dependsOn(downloadGlog)
 }
 
+def findNodeModulePath(baseDir, packageName) {
```
janicduplessis: Have you considered invoking the actual node binary to resolve the path instead? This seems good enough but I’m pretty sure there are a lot of edge cases in the node resolution algorithm

ide: Many of the edge cases in Node's module resolution algorithm are around finding the main module (needing to look at package.json["main"] and defaulting to index.js, for example) and handling `@scoped/packages`, neither of which we care about here. The algorithm for finding a package is actually pretty simple if you're not trying to super-optimize it - https://github.com/nodejs/node/blob/6db45bf7def22eedfd95dac95713e850c366b169/lib/internal/modules/cjs/loader.js#L584-L618

ide:
Removed the hard-coded edge cases and explained in a comment the various repo structures this new approach automatically handles. Error handling is better too.

Issue #26790, opened 2019-10-09
Starting a sequence animation for the second time throws an error


lbrdar:
## Env info
React Native Environment Info:
    System:
      OS: macOS 10.14.6
      CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
      Memory: 635.29 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.1 - ~/.nvm/versions/node/v10.15.1/bin/node
      Yarn: 1.19.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v10.15.1/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 13.1, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    IDEs:
      Android Studio: 3.1 AI-173.4819257
      Xcode: 11.1/11A1027 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.9.0 => 16.9.0 
      react-native: 0.61.2 => 0.61.2 

## Steps To Reproduce

1. Write an animation with  `Animated.seqence([ /* your animations */ ])`
2. Call `.start()` on that animation more than once

## Describe what you expected to happen:

I would expect that I can start an animation more than once without an error. The first iteration of the animation plays normally, but the second one fails and throws `undefined is not an object (evaluating 'animations[current].start')` pointing to the line 302:28 of `AnimatedImplementation.js`.

## Snack, code example, screenshot, or link to a repository:

* **Reproducible demo**: https://snack.expo.io/@leobrdar/rn-sequence-animation

![Screenshot 2019-10-09 at 19 05 11](https://user-images.githubusercontent.com/15323137/66503520-bce49800-eac7-11e9-98d6-eb753123f504.png)

PR #26791, opened 2019-10-09
Add support for calling .start() multiple times on animated sequence

lbrdar:
## Summary

This PR solves #26790 

## Changelog

[General] [Fixed] - Fix error thrown when sequence animation was started for the second time

## Test Plan

I've added a test to cover this case

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -280,6 +280,10 @@ const sequence = function(
   let current = 0;
   return {
     start: function(callback?: ?EndCallback) {
+      if (current > 0) {
+        throw new Error("Can't start a new iteration until previous one ends");
+      }
+
       const onComplete = function(result) {
         if (!result.finished) {
           current = 0;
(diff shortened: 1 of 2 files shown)


TheSavior:
What happens if start is called again before the first animation sequence finishes?

lbrdar:
@TheSavior good question. It won't break but the animation won't play properly either. 

I don't think other types of animation (e.g. parallel) handle that case. I guess the user should take care of that? What do you expect to happen in that case?

TheSavior:
I'm not sure what the "correct" behavior is here, I just wanted to flag it as it seems like a common edge case of being able to call start multiple times. 

Can parallel be called multiple times? What happens for parallel? Naively, I'd hope that they'd be consistent with respect to each other.

cpojer:
I was going to bring up the same concern that @eliwhite did. I think it makes more sense to me to throw an error if `start` is called more than once. What do you think?

lbrdar:
@cpojer are you suggesting that RN always throw an error when start() is called again or just if it's called before the previous iteration is complete? 

I'd say it makes sense to throw if called while the previous iteration is still running. Otherwise no.  Because I don't see a reason to block a user from starting the same animation multiple times.

cpojer:
Yeah I think that makes sense.

lbrdar:
@cpojer @TheSavior updated per your request. 

Could you help me with failing checks here?

cpojer:
I took an extended break, sorry I didn't get back to your PR earlier. I thought about this some more and do not feel comfortable changing the behavior right now. While I do not know what the original authors intended, I recommend creating new instances for new animations instead of re-using the same ones.

PR #26842, opened 2019-10-13
Fix Share dialog not resolving promise when dismissed on iOS

### LINKED ISSUE #26809: Promise returned by share dialog won't resolve if cancelled on iOS
On iOS, the share text dialog won't resolve the returned promise when the dialog is dismissed by pressing the Cancel button or by pressing outside the dialog. Other share options resolve the promise as expected.

```typescript
shareLink = async () => {
  await Share.share({ message: 'message-to-share' }) // Hangs if the dialog is dismissed by the Cancel button or by pressing outside the dialog
}
```

React Native version: 0.61.2

## Steps To Reproduce

1. Open share dialog with `Share.share(content, options)`
2. Dismiss the dialog by pressing Cancel or pressing outside the dialog.

Describe what you expected to happen: The promise returned by `Share.share(content, options)` should be resolved when the dialog is dismissed.

v-fernandez:
## Summary
On iOS the promised returned by `Share.share(content, options)` isn't resolved if the user dismisses the dialog by either pressing "Cancel" or pressing outside the shared dialog. This PR fixes this issue. 

This fixes #26809.

## Changelog

[iOS] [Fixed] - Fix promised returned by `Share.share(content, options)` not resolving if share dialog dismissed

## Test Plan
1. on iOS, open a share dialog with: 

```typescript
const onShare = async () => {
  const result = await Share.share({ message: 'example message' });
}
```

2. Dismiss the opened share dialog and the returned promised should resolve.

### DIFF
--- a/React/CoreModules/RCTActionSheetManager.mm
+++ b/React/CoreModules/RCTActionSheetManager.mm
@@ -183,7 +183,7 @@ RCT_EXPORT_METHOD(showShareActionSheetWithOptions:(JS::NativeActionSheetManager:
   shareController.completionWithItemsHandler = ^(NSString *activityType, BOOL completed, __unused NSArray *returnedItems, NSError *activityError) {
     if (activityError) {
       failureCallback(@[RCTJSErrorFromNSError(activityError)]);
-    } else if (completed) {
+    } else if (completed || activityType == nil) {
       successCallback(@[@(completed), RCTNullIfNil(activityType)]);
     }
   };


tomtargosz:
So it looks like this is because of my PR (https://github.com/facebook/react-native/pull/26429). I added this check because we need to have some sort of guard against the success callback being called multiple times. Is there a way we can know that the ActionSheet was closed so we can do `if (completed || dismissed)`?

tomtargosz:
@v-fernandez so I think we can change the check to:
`if (completed || activityType == nil)`

It looks like in `completionWithItemsHandler` the activityType will only be `nil` if the user closed the ShareSheet themselves, like so:

Screen Shot 2019-10-14 at 1 41 22 PM

Here's when they close the reminders modal:
Screen Shot 2019-10-14 at 1 43 02 PM

And the photos modal:
Screen Shot 2019-10-14 at 1 47 03 PM

This should fix the current issue without reintroducing the crash.

v-fernandez:
@tomtargosz makes sense. I'll update the PR with this. Thanks for the help!

franciscomorais:
When is this PR going to be picked?

dpolkovnikov:
> When is this PR going to be picked?

https://github.com/facebook/react-native/releases/tag/v0.62.0-rc.0

PR #26847, opened 2019-10-14
fix(eslint-config): Clashes with prettier

iRoachie:
## Summary

Currently, @react-native-community config package extends from prettier/recommended which comes with default settings from prettier. However there are still some eslint rules in the config that either clash or duplicate the settings from prettier.

This results in eslint fixing the formatting and then prettier undoing it. This PR removes the style specific rules from eslint and place them in the prettier section.
## Changelog

[General] [Fixed] - Remove style rules from eslint config for prettier options

## Test Plan

I created a repo for you to test with https://github.com/iRoachie/eslint-bug-replicate. You can see that running `yarn lint --fix` will never fix the issue. Eslint will complain about double quotes and subsequently after fixing it will complain about single quotes. 

Here's a gif of the behaviour (vscode eslint plugin `"eslint.autoFixOnSave": true`):

Kapture 2019-10-13 at 23 34 15

### DIFF
--- a/packages/eslint-config-react-native-community/index.js
+++ b/packages/eslint-config-react-native-community/index.js
@@ -112,7 +112,7 @@ module.exports = {
 
   rules: {
     // General
-    'comma-dangle': [1, 'always-multiline'], // allow or disallow trailing commas
+    'comma-dangle': 0, // allow or disallow trailing commas (off by default)
     'no-cond-assign': 1, // disallow assignment in conditional expressions
     'no-console': 0, // disallow use of console (off by default in the node environment)
     'no-const-assign': 2, // disallow assignment to const-declared variables
@@ -250,9 +250,9 @@ module.exports = {
     'no-trailing-spaces': 1, // disallow trailing whitespace at the end of lines
     'no-underscore-dangle': 0, // disallow dangling underscores in identifiers
     'no-mixed-spaces-and-tabs': 1, // disallow mixed spaces and tabs for indentation
-    quotes: [1, 'single', 'avoid-escape'], // specify whether double or single quotes should be used
+    quotes: 0, // specify whether double or single quotes should be used (off by default)
     'quote-props': 0, // require quotes around object literal property names (off by default)
-    semi: 1, // require or disallow use of semicolons instead of ASI
+    semi: 0, // require or disallow use of semicolons instead of ASI (off by default)
     'sort-vars': 0, // sort variables within the same declaration block (off by default)
     'space-in-brackets': 0, // require or disallow spaces inside brackets (off by default)
     'space-in-parens': 0, // require or disallow spaces inside parentheses (off by default)
@@ -309,5 +309,11 @@ module.exports = {
     'jest/no-focused-tests': 1,
     'jest/no-identical-title': 1,
     'jest/valid-expect': 1,
+
+    // Prettier Plugin
+    'prettier/prettier': [
+      1,
+      {singleQuote: true, trailingComma: 'es5', semi: true},
+    ],
   },
 };


vikrantnegi:
@iRoachie You are a lifesaver. I've been struggling with this issue ever since react-native came with the eslint config. 

Thanks for the fix.

packages/eslint-config-react-native-community/index.js
```diff
@@ -112,7 +112,6 @@ module.exports = {
 
   rules: {
     // General
-    'comma-dangle': [1, 'always-multiline'], // allow or disallow trailing commas
```
TheSavior: It seems like we prefer to leave these, but just change them to 0

iRoachie: Cool np

iRoachie: @TheSavior Changes made

packages/eslint-config-react-native-community/index.js
```diff
@@ -309,5 +309,11 @@ module.exports = {
     'jest/no-focused-tests': 1,
     'jest/no-identical-title': 1,
     'jest/valid-expect': 1,
+
+    // Prettier Plugin
+    'prettier/prettier': [
+      1,
+      {singleQuote: true, trailingComma: 'es5', semi: true},
```
friederbluemle: @iRoachie - I wonder how this section is related to the default content of `.prettierrc.js`: 

```
module.exports = {
  bracketSpacing: false,
  jsxBracketSameLine: true,
  singleQuote: true,
  trailingComma: 'all',
};
```

Does it override and/or make it (partially) obsolete?

iRoachie: Not sure exactly, what i do know is that the contents of that file doesn’t get passed onto user projects and previously you had to have a .prettierrc of your own

friederbluemle: Okay thanks for your reply. If (part of?) that file is obsolete (after updating the [`@react-native-community/eslint-config` dependency](https://github.com/facebook/react-native/blob/master/template/package.json#L18)), maybe we should update the [template file](https://github.com/facebook/react-native/blob/master/template/_prettierrc.js) as well?

friederbluemle: @satya164 - Thanks for your comment! Could you give an example (or a link to one)?

satya164: Prettier can format a wide-range of files such as markdown files (including code inside markdown files). The ESLint config only applies to TypeScript and JS files. Not sure what kind of example you're expecting?

cpojer:
Actually I have to revert this change because it's causing trouble with our setup at Facebook.

packages/eslint-config-react-native-community/index.js
friederbluemle: Okay never mind, I understand the reason. Thanks!

friederbluemle:
Basically, without the changes in this PR, you're _required_ to have a `.prettierrc.js` with the following config as well:

```
module.exports = {
  bracketSpacing: false,
  jsxBracketSameLine: true,
  singleQuote: true,
  trailingComma: 'all',
};
```

Otherwise, you'd run into the problem as stated by @iRoachie - Is that assumption correct?

satya164:
You're not required to have that file. It'd just use prettier's default config.

iRoachie:
@cpojer Sigh 😔 thanks everyone for your time. Will just use another config

friederbluemle:
> You're not required to have that file. It'd just use prettier's default config.

What I meant was that if you don't have it, you'd run into the problem described by @iRoachie 

Let's try to find a solution that works - A couple of weeks ago I also tried to used `@react-native-community/eslint-config` and had linter conflicts related to the quotes.

@cpojer - Could you elaborate a bit on the trouble it caused?

satya164:
The issue that @iRoachie described was because of conflicting eslint rules being enabled, not presence of the file.

Issue #24549, opened 2019-04-21
unable to find module for uimanager (lazilyloadIfNeccessary)


nomi9995:
in react native 0.58 they added lazilyloadIfNeccessary and remove uimanager. but i am using gl-react-native library and this library is using uimanger. I face this error when I reload app like cmd+r. Therefore I am facing an error that has attached below. what is the solution. how can i change code in gl-react-native to work without giving me uimanager error

you can see changelog for react native 0.58
https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#058
![Screenshot 2019-04-19 at 2 43 48 AM](https://user-images.githubusercontent.com/36044436/56467991-8e063700-643f-11e9-9f77-be25ac0ab5bd.png)

Issue #23235, opened 2019-01-31
[iOS] Simulator error message `Unable to find module for DevLoadingView`


superguineapig:
## Environment
 React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
      Memory: 2.85 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.10.0 - ~/.nvm/versions/node/v8.10.0/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 5.6.0 - ~/.nvm/versions/node/v8.10.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        Build Tools: 23.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.1, 27.0.3, 28.0.2, 28.0.3
        API Levels: 23, 25, 26, 27
    IDEs:
      Android Studio: 3.1 AI-173.4907809
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.58.3 => 0.58.3 
    npmGlobalPackages:
      react-native-git-upgrade: 0.2.7

## Description
Recently upgraded to RN 0.58.3 (from 0.57.8) via `react-native-git-upgrade`.  Everything compiles just fine.

However, invoking a debug run via `react-native run-ios` will result in an immediate error message:

<img width="545" alt="error" src="https://user-images.githubusercontent.com/4808737/52081971-343d5180-2561-11e9-945e-a885fd9363aa.png">

Notes:
- Reloads (cmd-R) of a running app clear the error. (App runs as expected)
- Subsequent reloads (cmd-R) of already running app *do not* trigger the error. (App runs as expected)
- Closing either the simulator itself and *restarting*, OR closing the running app within the simulator session and *relaunching* will trigger the error once again.
- Each of the 0.58.0, 0.58.1, 0.58.2, 0.58.3 releases give the same results.

## Reproducible Demo
[...]

PR #26861, opened 2019-10-14
Fix Unable to find module for UIManager issue

kevinvangelder:
## Summary
A race condition sometimes causes redbox errors when reloading the app after previously opening the Chrome debugger. In the project I was working on, we were regularly getting `Unable to find module for UIManager` and I was able to track it down to `_moduleDataByName` sometimes being empty or nil. This change safely early-exits when that is the case and the app loads successfully. Related issues:
#24549
#23235

## Changelog

[iOS] [Fixed] - Race condition causing redbox errors when reloading app after using Chrome debugger

## Test Plan
As this is a difficult-to-reproduce race condition, I'm not entirely certain what the test plan should be. I created this change as a patch to React Native in the closed-source project I've been working on and it works flawlessly, however I haven't been able to reproduce the issue in a clean project. This is a fairly safe change as the early exit only ever occurs if the variable is empty and it seems to be resilient enough to load the module again after the race condition has been defeated, but I'm open to suggestions on potential options to test this.

### DIFF
--- a/React/CxxBridge/RCTCxxBridge.mm
+++ b/React/CxxBridge/RCTCxxBridge.mm
@@ -463,6 +463,9 @@ struct RCTInstanceCallback : public InstanceCallback {
 
 - (id)moduleForName:(NSString *)moduleName lazilyLoadIfNecessary:(BOOL)lazilyLoad
 {
+  if (!_moduleDataByName) {
+    return nil;
+  }
   if (RCTTurboModuleEnabled() && _turboModuleLookupDelegate) {
     const char* moduleNameCStr = [moduleName UTF8String];
     if (lazilyLoad || [_turboModuleLookupDelegate moduleIsInitialized:moduleNameCStr]) {


mmmulani:
rather than change the module loading logic, I would change uiManager's getter here: https://github.com/facebook/react-native/blob/master/React/Modules/RCTUIManager.m#L1616

that seems to make more sense as it looks like the dealloc is throwing the error

kevinvangelder:
Is having an early exit when `_moduleDataByName` is nil a bad thing? We can certainly also address other weaknesses, but I doubt there is any situation where we would want to receive a redbox error instead of having it gracefully handled.

mmmulani:
> Is having an early exit when `_moduleDataByName` is nil a bad thing?

yes. you're essentially checking to see if RCTUIManager has been invalidated and if so skip some checks. But we can instead harden when the RCTUIManager is accessed off RCTCxxBridge

> We can certainly also address other weaknesses, but I doubt there is any situation where we would want to receive a redbox error instead of having it gracefully handled.

yea I definitely agree that we do not want to show a redbox in this case but I think that we should solve the problem a different way.

kevinvangelder:
@mmmulani Okay. Unfortunately I have no C++ experience, so I'm not sure what the alternative solution would look like. If it's a simple change would you be willing to open a PR and I'll close this one in favor of it?

spacesuitdiver:
image

Just helping document the issue on the project I use it happens every time in the scenario @kevinvangelder described.

kevinvangelder:
I'm curious if @mhorowitz has any thoughts for us.

satbirkira:
Any update on this?

Issue #26788, opened 2019-10-09
console.log commands are not shown their code line locations


ezranbayantemur:
After 0.61.x releases on the debug mode ```console.log``` commands are not showing their code line locations. They always shown at  ```... Project Directioy ...\projectName\node_modules\react-native\Libraries\Core\setUpDeveloperTools.js:73``` directory. 

For example;
**Before update:**
![1_Ot8O3BcmnRF_JkeerVYWfA](https://user-images.githubusercontent.com/42299721/66490643-29f13100-eaba-11e9-8eb5-3b0180064cc6.jpg)

**After update:**
![t1 (1)](https://user-images.githubusercontent.com/42299721/66490654-2eb5e500-eaba-11e9-830a-2cfac52005e2.jpg)

Is this a bug? If it's not previous version was way more useful for variable, data or etc.. tracking.

**Versions:**
```
    "react": "16.9.0",
    "react-native": "0.61.1"
```
**Environment:**
`[skip envinfo]`

TheSavior:
@motiz88, does this look familiar to any recent changes? Maybe something like framesToPop on console.log?

motiz88:
This is probably due to the way we wrap console methods to add React component stacks. I'm not sure whether Chrome exposes a way to manipulate the call site attribution for logs. cc @bvaughn

ezranbayantemur:
I'm not sure why and how it's happening but it would be more better to work like other version.

JulianCurrie:
I'm getting the same result in my console with the newer versions of RN.

acoutts:
I'm also experiencing this after upgrading to React Native 0.61.2. Worked fine in 0.60.x

bvaughn:
To be clear, **this is 100% expected to happen for v61+ and not for v60 (or earlier)** because v61 embeds the new backend from [`react-devtools-core`](https://www.npmjs.com/package/react-devtools-core) v4.

I also made a judgement call for the override to be **on by default** (so it's opt-out rather than opt-in) because (1) people might not be running the DevTools frontend at all and (2) even if they were, there's an initialization race during which any logged errors would not get the appended callstack.

This decision is somewhat controversial. If folks feel that I made the wrong decision by defaulting on, we can talk about it.

acoutts:
Thanks @bvaughn  - for those of us on v61+ now, what's the recommended toolset to continue debugging to see line numbers? I'm willing to switch tools if it means I can see the line numbers again.

[...]

PR #26883, opened 2019-10-16
Don't wrap console methods for metro logging in Chrome debugger

rickhanlonii:
Summary:
This diff fixes an issue reported in https://github.com/facebook/react-native/issues/26788 where logs in the Chrome console were showing a different location than previous versions.

In the change here, we stop wrapping the console functions to attach the metro websocket in any environment that isn't a native app environment. We do this by checking `global.nativeLoggingHook` which is bound only by native apps and not environments like the Chrome DevTools.

Differential Revision: D17951707


rickhanlonii:
Closes https://github.com/facebook/react-native/issues/26788

PR #26902, opened 2019-10-17
Fix ActionSheet completion callback not called when dialog is dismissed

mzehngut:
## Summary

https://github.com/facebook/react-native/pull/26429 introduced this problem: when the action sheet is dismissed without having selected and then canceled an action, neither of the success or failure callbacks is called.  This causes the Promise that opens the action sheet in the Javascript code to remain unfulfilled.  This fix checks that if no activity was chosen, in other words the dismiss button on the share dialog was pressed and the dialog was closed, then the success callback will still be called.

## Changelog

[iOS] [Fixed] - Fix ActionSheet completion callback not called when dialog is dismissed

## Test Plan

I'm using the react-native-share component to invoke the action sheet.  Add these lines to the Javascript code:
`import Share from 'react-native-share';`
`Share.open({`
`    url: 'http://www.facebook.com'`
`}).finally(() => {`
`    console.log('finally called');`
`});`

In the ActionSheet dialog, press the dismiss button without having selected an action.  Without this fix, the finally function is not called.  With the fix, it is called.  All other behavior is the same as before.


mzehngut:
Sorry, just noticed this problem has already been reported and fixed in https://github.com/facebook/react-native/pull/26842.

Issue #26755, opened 2019-10-08
KeyboardAvoidingView styles don't compose nicely


garrettgu10:
https://github.com/facebook/react-native/blob/92073d4a71d50a1ed80cf9cb063a6144fcc8cf19/Libraries/Components/Keyboard/KeyboardAvoidingView.js

In the render function of KeyboardAvoidingView, the calculation of the styles of the container view are not calculated with regards to the given props.style. For example, on line 194, bottomHeight is used for the bottom style attribute rather than (bottomHeight + contentContainerStyle.bottom).

React Native version: 0.61

## Steps To Reproduce

1. Render a KeyboardAvoidingView with behavior "padding" and set the paddingBottom of its style. 
2. Toggle the keyboard. The paddingBottom attribute is reset.

Describe what you expected to happen:
I expect the extra keyboard padding to be added to the given paddingBottom rather than replace it. 

Snack, code example, screenshot, or link to a repository:
https://snack.expo.io/ryk4ftKOS
Notice that the bottom padding starts off being 20, but after toggling the keyboard, it is reset to 0.

lunaleaps:
Line 194 is only for behavior "position", you'd want to update the appropriate "padding" handling in KeyboardAvoidingView.

lunaleaps:
Added a comment on the attempted fix: #26948. I challenge whether this is a good idea and rather discourage styling KeyboardAvoidingView with paddingBottom, bottom values.
We should guide users to instead be styling an inner view. @garrettgu10, @jeswinsimon let me know if there's a use-case I'm not understanding.

lunaleaps:
As discussed on #26948, let's update the docs to discourage padding/height styles on KeyboardAvoidingView. @garrettgu10 assigning this to you so we have a clear owner. Feel free to assign to me if you want to pass.

PR #26948, opened 2019-10-22
Update bottom padding calculation

jeswinsimon:
## Summary

This is a fix for issue **KeyboardAvoidingView styles don't compose nicely** #26755.

## Changelog

[General] [Fixed] - Updated calculation of `bottom` style when the behaviour is `position`.

## Test Plan

The expected behaviour as per #26755 is obtained.

### DIFF
--- a/Libraries/Components/Keyboard/KeyboardAvoidingView.js
+++ b/Libraries/Components/Keyboard/KeyboardAvoidingView.js
@@ -191,7 +191,7 @@ class KeyboardAvoidingView extends React.Component<Props, State> {
               style={StyleSheet.compose(
                 contentContainerStyle,
                 {
-                  bottom: bottomHeight,
+                  bottom: bottomHeight + contentContainerStyle.bottom,
                 },
               )}>
               {children}


Libraries/Components/Keyboard/KeyboardAvoidingView.js
```diff
@@ -191,7 +191,7 @@ class KeyboardAvoidingView extends React.Component<Props, State> {
               style={StyleSheet.compose(
                 contentContainerStyle,
                 {
-                  bottom: bottomHeight,
+                  bottom: bottomHeight + contentContainerStyle.bottom,
```
lunaleaps: Tests are failing because this isn't accounting for the null case for `contentContainerStyle` and `bottom` property

lunaleaps: Also this needs to verify  that `contentContainerStyle.bottom` is a number value before adding

lunaleaps: Should also add: 
`bottomHeight + style != null ? style.paddingBottom : 0,`
to line 208

lunaleaps:
Looking at this more closely, I think the best solution here is to not place paddingBottom or bottom styles on KeyboardAvoidingView.

Either that or providing a specific prop that takes only a number value as `style`, `paddingBottom`, and `bottom` are very generally typed and it'd be a bit roundabout to get a number value (which is the only type that would be useful here) and make it quite murky about lack of support for percentages or Animated values. 

Let me know what you think. Will close if there are no further comments in a week.

garrettgu10:
I suppose that makes sense. Grabbing the bottom styles would require also supporting style arrays and IDs for completeness which would incur a performance penalty for a relatively rare edge case. Would you recommend updating the documentation to explicitly discourage adding these styles?

jeswinsimon:
> Looking at this more closely, I think the best solution here is to discourage styling KeyboardAvoidingView with paddingBottom or bottom styles.
> 
> Either that or providing a specific prop on KeyboardAvoidingView that takes only a number value. As `style`, `paddingBottom`, and `bottom` are very generally typed it's be a bit roundabout to get a number value (which is the only type that would be useful here). It muddies up the expectation of what KeyboardAvoidingView supports and doesn't support regarding percentages or Animated values.
> 
> Let me know what you think. Will close if there are no further comments in a week.

Sounds good to me.

lunaleaps:
@garrettgu10 yea definitely. Is that something you'd be interested in updating?

PR #26966, opened 2019-10-23
Add support for native pseudo-OS to Platform.select

koke:
## Summary

When you write platform-specific code using [file extensions](https://facebook.github.io/react-native/docs/platform-specific-code#platform-specific-extensions), you can specify `.ios.js`, `.android.js`, or the catch-all `.native.js` when you are sharing code with a web project.

This `native` shortcut is missing for the `Platform.select` method, and this PR is adding support for that.

## Changelog

[General] [Added] - Platform.select now supports native as an option.

## Test Plan

Added relevant passing unit tests for Platform module.

### DIFF
--- a/Libraries/Utilities/Platform.android.js
+++ b/Libraries/Utilities/Platform.android.js
@@ -55,7 +55,11 @@ const Platform = {
     return this.constants.uiMode === 'tv';
   },
   select: <A, N, D>(spec: PlatformSelectSpec<A, N, D>): A | N | D =>
-    'android' in spec ? spec.android : 'native' in spec ? spec.native : spec.default,
+    'android' in spec
+      ? spec.android
+      : 'native' in spec
+      ? spec.native
+      : spec.default,
 };
 
 module.exports = Platform;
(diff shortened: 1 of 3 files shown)


koke:
Note that running `yarn lint --fix` seems to change all the double quotes into single quotes in docs/generatedComponentApiDocs.js, but I haven't committed that change, since it looks unrelated to this PR.

cpojer:
Hey @koke! Thanks for this PR. I think it makes sense to support this via `Platform.select()`. However, before we merge this, we need to add support for this in Metro as well. Metro inlines `Platform.select()` calls in order to not increase the bundle size with unnecessary code. For this, we should change it so that it inlines `native` calls if there is no platform item in the object and drop it if the platform is web. Once you submit a PR, please tag me and I'll review and merge it, and then get back to the PR here.

* The code for the plugin is here: https://github.com/facebook/metro/blob/master/packages/metro/src/JSTransformer/worker/inline-platform.js
* The tests should go here: https://github.com/facebook/metro/blob/master/packages/metro/src/JSTransformer/worker/__tests__/inline-plugin-test.js

PR #26978, opened 2019-10-24
[NativeAnimated] Don't restore default values when components unmount

janicduplessis:
## Summary

There are some cases where restoring default values on component unmount is not desirable. For example in react-native-screens we want to keep the native view displayed after react has unmounted them. Restoring default values causes an issue there because it will change props controlled my native animated back to their default value instead of keeping whatever value they had been animated to.

Restoring default values is only needed for updates anyway, where removing a prop controlled by native animated need to be reset to its default value since react no longer tracks its value.

This splits restoring default values and disconnecting from views in 2 separate native methods, this way we can restore default values only on component update and not on unmount. This takes care of being backwards compatible for OTA JS updates.

## Changelog

[General] [Fixed] - NativeAnimated - Don't restore default values when components unmount

## Test Plan

- Tested in an app using react-native-screens to make sure native views that are kept after their underlying component has been unmount don't change. Also tested in RNTester animated example.

- Tested that new JS works with old native code

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -116,6 +116,13 @@ const API = {
     invariant(NativeAnimatedModule, 'Native animated module is not available');
     NativeAnimatedModule.disconnectAnimatedNodeFromView(nodeTag, viewTag);
   },
+  restoreDefaultValues: function(nodeTag: number): void {
+    invariant(NativeAnimatedModule, 'Native animated module is not available');
+    // Backwards compat with older native runtimes, can be removed later.
+    if (NativeAnimatedModule.restoreDefaultValues != null) {
+      NativeAnimatedModule.restoreDefaultValues(nodeTag);
+    }
+  },
   dropAnimatedNode: function(tag: number): void {
     invariant(NativeAnimatedModule, 'Native animated module is not available');
     NativeAnimatedModule.dropAnimatedNode(tag);
--- a/Libraries/Animated/src/NativeAnimatedModule.js
+++ b/Libraries/Animated/src/NativeAnimatedModule.js
@@ -45,6 +45,7 @@ export interface Spec extends TurboModule {
   +extractAnimatedNodeOffset: (nodeTag: number) => void;
   +connectAnimatedNodeToView: (nodeTag: number, viewTag: number) => void;
   +disconnectAnimatedNodeFromView: (nodeTag: number, viewTag: number) => void;
+  +restoreDefaultValues: (nodeTag: number) => void;
   +dropAnimatedNode: (tag: number) => void;
   +addAnimatedEventToView: (
     viewTag: number,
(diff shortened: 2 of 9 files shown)


janicduplessis:
Note: This also requires updating the generated spec to work, I did it manually to test but I'm not sure if codegen is runnable in OSS so I didn't include this change.

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -116,6 +116,13 @@ const API = {
     invariant(NativeAnimatedModule, 'Native animated module is not available');
     NativeAnimatedModule.disconnectAnimatedNodeFromView(nodeTag, viewTag);
   },
+  restoreDefaultValues: function(nodeTag: number): void {
+    invariant(NativeAnimatedModule, 'Native animated module is not available');
+    // Backwards compat with older native runtimes, can be removed later.
+    if (NativeAnimatedModule.restoreDefaultValues != null) {
```
JoshuaGross: If this is running against an older runtime, does this check actually work?

JoshuaGross:
If appropriate could you attach before/after videos?

Libraries/Animated/src/NativeAnimatedHelper.js
janicduplessis: I did test running just the JS without the native changes and it worked fine. On older runtime the function doesn't exist and the disconnectFromView method handles both disconnection and restoring default values.

janicduplessis:
You can see the bug at the end of this video, you can see the navbar flicker white during the screen back transition:

https://imgur.com/cPHdnxT

How react-native-screen works for screen animation is when a screen component gets unmounted the native view does not get removed from the native view hierarchy and gets animated. In my app I have a navbar with an opacity value controlled by Animated initially at 0. When the screen gets unmounted the animated native driver would restore the default value for opacity (so 1 in this case) and case the header to show during the transition.

Restoring default values is not needed on component unmount since the view will usually get removed and if it doesn't we want it to keep its current state.

Here's a video after the fix:

https://imgur.com/a/bAZ9O54

janicduplessis:
@JoshuaGross Rebased and added a JS test

JoshuaGross:
@janicduplessis one of the tests is failing to pass locally:

```
stderr: /tmp/react-native-github/ReactAndroid/src/test/java/com/facebook/react/animated/NativeAnimatedNodeTraversalTest.java:984: error: method restoreDefaultValues in class com.facebook.react.animated.NativeAnimatedNodesManager cannot be applied to given types;
    mNativeAnimatedNodesManager.restoreDefaultValues(propsNodeTag, viewTag);
                               ^
  required: int
  found: int,int
  reason: actual and formal argument lists differ in length
```

janicduplessis:
@JoshuaGross Fixed, I forgot to update android tests after changing the signature of restoreDefaultValues (The 2nd parameter was unused).

janicduplessis:
@JoshuaGross Also do you know what the strategy is to land changes to native modules signature with codegen for OSS PRs? I did not include generated file changes in my PR since AFAIK running codegen in OSS is broken. I did test by editing the codegened files manually.

JoshuaGross:
@janicduplessis Not really sure - longterm we have to fix codegen in OSS. I can't land your diff without running codegen manually, so for now our land-blocking tests will prevent us from merging in anything that breaks trunk.

PR #27006, opened 2019-10-25
Add Android support for fontVariant prop

mcuelenaere:
## Summary

Android was missing support for the `fontVariant` prop in TextViews, this PR adds that.

## Changelog

[Android] [Added] - Add Android support for fontVariant prop

## Test Plan

Since I can't get RNTester to work locally (it crashes when loading `libyoga.so` on `No implementation found for long com.facebook.yoga.YogaNative.jni_YGConfigNew()`), I'll post some screenshots below of our app showing the difference.

We are using a slightly different [version](https://github.com/getdelta/react-native/commit/10cafcaa0798e5dbe8b56d461885fa84c6953739) of this commit, since we're still on 0.60, but the gist remains the same when rebased on master.

Before:
Screenshot_20191025-130325__01

After:
Screenshot_20191025-130444__01

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomStyleSpan.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/text/CustomStyleSpan.java
@@ -10,6 +10,7 @@ package com.facebook.react.views.text;
 import android.content.res.AssetManager;
 import android.graphics.Paint;
 import android.graphics.Typeface;
+import android.os.Build;
 import android.text.TextPaint;
 import android.text.style.MetricAffectingSpan;
 import androidx.annotation.NonNull;
@@ -32,27 +33,30 @@ public class CustomStyleSpan extends MetricAffectingSpan implements ReactSpan {
 
   private final int mStyle;
   private final int mWeight;
+  private final @Nullable String mFeatureSettings;
   private final @Nullable String mFontFamily;
 
   public CustomStyleSpan(
       int fontStyle,
       int fontWeight,
+      String fontFeatureSettings,
       @Nullable String fontFamily,
       @NonNull AssetManager assetManager) {
     mStyle = fontStyle;
     mWeight = fontWeight;
+    mFeatureSettings = fontFeatureSettings;
     mFontFamily = fontFamily;
     mAssetManager = assetManager;
   }
 
   @Override
   public void updateDrawState(TextPaint ds) {
-    apply(ds, mStyle, mWeight, mFontFamily, mAssetManager);
+    apply(ds, mStyle, mWeight, mFeatureSettings, mFontFamily, mAssetManager);
   }
 
   @Override
   public void updateMeasureState(@NonNull TextPaint paint) {
-    apply(paint, mStyle, mWeight, mFontFamily, mAssetManager);
+    apply(paint, mStyle, mWeight, mFeatureSettings, mFontFamily, mAssetManager);
   }
 
   /** Returns {@link Typeface#NORMAL} or {@link Typeface#ITALIC}. */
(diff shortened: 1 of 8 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTypefaceUtils.java
```diff
@@ -38,6 +43,36 @@ public static int parseFontStyle(@Nullable String fontStyleString) {
     return fontStyle;
   }
 
+  public static String parseFontVariant(ReadableArray fontVariantArray) {
```
janicduplessis: `@Nullable String` and `@Nullable ReadableArray`

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -483,6 +491,16 @@ public void setFontWeight(@Nullable String fontWeightString) {
     }
   }
 
+  @ReactProp(name = ViewProps.FONT_VARIANT)
+  public void setFontVariant(ReadableArray fontVariantArray) {
```
janicduplessis: `@Nullable ReadableArray`

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
```diff
@@ -483,6 +491,16 @@ public void setFontWeight(@Nullable String fontWeightString) {
...
+    String fontFeatureSettings = ReactTypefaceUtils.parseFontVariant(fontVariantArray);
+
+    if (!Objects.equals(fontFeatureSettings, mFontFeatureSettings)) {
```
janicduplessis: I'm not familiar with `Object.equals`, does it compare all values for an array?

ReactAndroid/src/main/java/com/facebook/react/views/text/TextAttributeProps.java
```diff
@@ -155,6 +162,14 @@ private float getFloatProp(String name, float defaultvalue) {
     }
   }
 
+  private ReadableArray getArrayProp(String name) {
```
janicduplessis: `@Nullable ReadableArray `

ReactAndroid/src/main/java/com/facebook/react/views/text/TextAttributeProps.java
```diff
@@ -280,6 +295,10 @@ public void setFontFamily(@Nullable String fontFamily) {
     mFontFamily = fontFamily;
   }
 
+  public void setFontVariant(ReadableArray fontVariant) {
```
janicduplessis: `@Nullable ReadableArray`

ReactAndroid/src/main/java/com/facebook/react/views/text/CustomStyleSpan.java
```diff
@@ -32,27 +33,30 @@
 
   private final int mStyle;
   private final int mWeight;
+  private final @Nullable String mFeatureSettings;
   private final @Nullable String mFontFamily;
 
   public CustomStyleSpan(
       int fontStyle,
       int fontWeight,
+      String fontFeatureSettings,
```
janicduplessis: `@Nullable String`

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
mcuelenaere: This isn't comparing the `ReadableArray`, but the `String` returned from `ReactTypefaceUtils.parseFontVariant()` call with `this.mFontFeatureSettings`.

https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

TBH, I stumbled upon this via a Stackoverflow post when looking for a shorter way to write this check so it was new to me as well ;)

janicduplessis:
Cc @mdvacca @JoshuaGross

ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java
gaykov: I might be wrong here, but looks like `java.util.Objects` and it's static methods was introduced in API level 19 (https://developer.android.com/reference/java/util/Objects). So I think it can possibly crash on Android 4 devices with `NoClassDefFoundError: java.util.Objects`, when `Object.equals` is called. We're investigating the similar crash in react-native-camera here: https://github.com/react-native-community/react-native-camera/issues/2528.

mcuelenaere: I can replace this with https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/Objects.java#L57 if required

mcuelenaere:
> As a follow up, can you help updating the documentation? https://facebook.github.io/react-native/docs/text-style-props#fontvariant
> Thank you!

https://github.com/facebook/react-native-website/pull/1443

PR #27012, opened 2019-10-25
Expose [RCTBundleURLProvider isPackagerRunning] publicly

afoxman-microsoft:
## Summary

The private API isPackagerRunning is useful. Rather than duplicating it, I am exposing it here so that I can make use of it in future PRs, and so others can do the same.

## Changelog

[iOS] [Changed] - Expose the isPackagerRunning methods on RCTBundleURLProvider

## Test Plan

This change doesn't impact any runtime code, though it does impact build. I used a vanilla test app (react-native init) and built it using xcode as well as react-native run-ios.


afoxman-microsoft:
The CI loop failures don't seem related to my change. This is my first PR, so I'm not sure what happens in this situation.

PR #27016, opened 2019-10-26
Make JSCRuntime::createValue 35% faster

radex:
## Summary

JSC does some sort of thread safety locking on every single JSC API call, which makes them ridiculously expensive for a large number of calls (such as when passing a large array over the RN bridge). It would be great if we could lock and unlock once for an entire sequence of JSC calls… but in the meantime, the less we call JSC the better.

unknown

In my benchmark environment (https://github.com/Nozbe/WatermelonDB/pull/541), the time spent in JSCRuntime::createValue went down from 1.07s to 0.69s by changing JSValueIsXXXX calls to a single JSValueGetType call.

The underlying implementation does the same thing: https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/API/JSValueRef.cpp#L58

## Changelog

[General] [Fixed] - Make JSCRuntime::createValue faster

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -1314,27 +1314,31 @@ jsi::Object JSCRuntime::createObject(JSObjectRef obj) const {
 }
 
 jsi::Value JSCRuntime::createValue(JSValueRef value) const {
-  if (JSValueIsNumber(ctx_, value)) {
-    return jsi::Value(JSValueToNumber(ctx_, value, nullptr));
-  } else if (JSValueIsBoolean(ctx_, value)) {
-    return jsi::Value(JSValueToBoolean(ctx_, value));
-  } else if (JSValueIsNull(ctx_, value)) {
-    return jsi::Value(nullptr);
-  } else if (JSValueIsUndefined(ctx_, value)) {
-    return jsi::Value();
-  } else if (JSValueIsString(ctx_, value)) {
-    JSStringRef str = JSValueToStringCopy(ctx_, value, nullptr);
-    auto result = jsi::Value(createString(str));
-    JSStringRelease(str);
-    return result;
-  } else if (JSValueIsObject(ctx_, value)) {
-    JSObjectRef objRef = JSValueToObject(ctx_, value, nullptr);
-    return jsi::Value(createObject(objRef));
-  } else if (smellsLikeES6Symbol(ctx_, value)) {
-    return jsi::Value(createSymbol(value));
-  } else {
-    // WHAT ARE YOU
-    abort();
+  JSType type = JSValueGetType(ctx_, value);
+
+  switch (type) {
+    case kJSTypeNumber:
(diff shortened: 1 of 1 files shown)


radex:
Doh, I accidentally broke Symbols.

btw. @mhorowitz -- https://github.com/facebook/react-native/commit/dcc40a6267b429c9f0151287752922a5c8c4ea26#diff-44cf263c203a5949e25721f41394e9aeR418 — I haven't tested yet, but I think this was broken already on iOS 13 / macOS Catalina, and community Android builds of JSC, where there _is_ a kJSTypeSymbol:

https://github.com/WebKit/webkit/commit/76b4e1bf6f4f47a92bb9a5520eb999cd742c70c1#diff-7cc5e4afb4c7092e07c92db27a5c328eR89-R90

ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -1314,27 +1314,31 @@ jsi::Object JSCRuntime::createObject(JSObjectRef obj) const {
...
+      JSStringRelease(str);
+      return result;
+    }
+    case kJSTypeObject: {
+      JSObjectRef objRef = JSValueToObject(ctx_, value, nullptr);
+      return jsi::Value(createObject(objRef));
+    }
+//    case kJSTypeSymbol:
+    default:
+      // WHAT ARE YOU
```
jacobp100: This is neat! But do you need to check `smellsLikeES6Symbol` here?

mhorowitz: There are tests for this in https://github.com/facebook/react-native/blob/master/ReactCommon/jsi/jsi/test/testlib.cpp#L1160 but I am not sure if they actually run in CI.  But you should be able to compile/run them by hand.

If we want RN to build and pass these tests against JSC libraries with and without this symbol, it may be tricky.  Under Xcode, there are macros to test library versions, but I don't know if that works when using community JSC on Android.

ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -417,10 +417,18 @@ bool JSCRuntime::isInspectable() {
...
+  // Since iOS 13, JSValueGetType will return kJSTypeSymbol
+  // Before: Empirically, an es6 Symbol is not an object, but its type is
   // object.  This makes no sense, but we'll run with it.
-  return (!JSValueIsObject(ctx, ref) &&
-          JSValueGetType(ctx, ref) == kJSTypeObject);
+  // https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/API/JSValueRef.cpp#L79-L82
+
+  JSType type = JSValueGetType(ctx, ref);
+
+  if (type == /* kJSTypeSymbol */ 6) {
```
radex: I think this _should_ work, no?

radex:
@jacobp100 @mhorowitz fixed!

radex:
@mhorowitz any update? I've been running this patch in production for a week now, haven't seen any issues.

radex:
@mhorowitz what's holding up this PR?

PR #27038, opened 2019-10-28
Fix for: You cannot blur a text-input if you tab to it.

fat:
## Summary

I've been working on a new iOS experience with lots of text inputs and this has been driving me a bit nuts… 

If you're in a scrollview with `keyboardShouldPersistTaps="handled"` and you tab through your text-inputs, you aren't able to tap outside of a given text-input to blur it (and dismiss the keyboard).

I wrote up a quick explanation and some repo steps here: https://snack.expo.io/BJBcKgrqB

The patch i came up with, after poking around for a little bit seems terrifying - so almost certainly not it. But if it's helpful at all - decided to just got ahead and submit it. 

## Changelog

[iOS] [Fixed] - TextInput blur when tabbing in iOS simulator. 

## Test Plan
I tried to think of a way to test this in jest… but i didn't get very far sorry 😢

I did create a snack here so you can demo the issue: https://snack.expo.io/BJBcKgrqB

I also created two videos…

**Here's the text input not working when i try to blur it after tabbing in simulator**
ezgif-1-dc85b405c760

**Here's it working after I applied this patch**
ezgif-1-ed9f6b19653d

Thanks!

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -1098,6 +1098,7 @@ const TextInput = createReactClass({
     if (this.props.onFocus) {
       this.props.onFocus(event);
     }
+    this.focus();
   },
 
   _onPress: function(event: PressEvent) {
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -1095,10 +1095,10 @@ const TextInput = createReactClass({
   },
 
   _onFocus: function(event: FocusEvent) {
+    this.focus();
     if (this.props.onFocus) {
       this.props.onFocus(event);
     }
-    this.focus();
   },
 
   _onPress: function(event: PressEvent) {


fat:
Signed, thanks 🙏

shergin:
I would love to see the explanation of why it works and why it should be implemented this way.

fat:
I originally was trying to figure out why `_onPress` (https://github.com/facebook/react-native/blob/96333d09938d1a3118589fd5a0f390303d19390e/Libraries/Components/TextInput/TextInput.js#L1106) was working, while `_onFocus` wasn't (specifically when `_onFocus` is triggered by something like a tab event. That lead me to `_onPress` calling the `this.focus` method exposed by the `NativeMethodsMIxin`.

In trying to figure out what `focus()` did - I got as far as this code: (https://github.com/facebook/react-native/blob/96333d09938d1a3118589fd5a0f390303d19390e/Libraries/Renderer/shims/NativeMethodsMixin.js#L14) and then kinda through my hands up (unclear to me what is happening under the hood by calling `focus`? I might just not be familiar enough with the project to know where that code lives, but from my quick look - seemed like it was something obfuscated. 

Because of that, the first thing I thought to do then was… "what if i just called this focus mixin method when focus is triggered" (such as by a tab event). This seems to work for my usecase - although again - unclear if this is the way this issue should be resolved.

fat:
Hope that's helpful 🤷‍♀

TheSavior:
Thanks for highlighting this bug...and sending a PR too!

Okay, here is what I think is happening. For context, let's start with this diagram I have of how focus and blur propagates through the system.

graphviz (12)

ScrollView's scrollResponder is responsible for blurring text inputs when a touch occurs in the ScrollView but outside of the currently focused TextInput. The code for that is here: 
https://github.com/facebook/react-native/blob/6ba2769f0f92ca75fb0eb60ccb8337920a9c31eb/Libraries/Components/ScrollResponder.js#L301-L314

This happens on `scrollResponderHandleResponderRelease` aka, touch up.

It checks for what the currently focused textinput is by calling `TextInputState.currentlyFocusedField()`.

That function is a JS variable that is being updated by calls to `TextInputState.focusTextInput` and `TextInputState.blurTextInput`:

https://github.com/facebook/react-native/blob/6ba2769f0f92ca75fb0eb60ccb8337920a9c31eb/Libraries/Components/TextInput/TextInputState.js#L36-L71

I added some console logs to those methods to see which ones are being called when running your repro (thanks for the repro!). **This is without your fix**

Click on and off:
```
// Click on input 1
focusTextInput input1
TextInput's _onFocus called

// Click on blank space
scrollResponderHandleResponderRelease blur input1
blurTextInput input1
TextInput's _onBlur called
```

Click on input1, then input 2, then off
```
// Click on input 1
focusTextInput input1
TextInput's _onFocus called for input1

// Click on input 2
focusTextInput input2
TextInput's _onBlur called for input1 
TextInput's _onFocus called for input2

// Click on blank space
scrollResponderHandleResponderRelease blur input2
blurTextInput input2
TextInput's _onBlur called for input2
```

And now for the bug. Click on input 1, tab to 2, then off
```
// Click on input 1
focusTextInput input1
TextInput's _onFocus called for input1

// Tab to input 2
TextInput's _onBlur called for input1 
TextInput's _onFocus called for input2

// Click on blank space
scrollResponderHandleResponderRelease blur input1
blurTextInput input1
```

Notice how `focusTextInput` was never called with input2 in the last example. Since this is the function that sets the `currentlyFocusedField` when we click on the blank space RN is trying to blur the first input instead of the second.

# The root cause
We are tracking the state of which field is focused in JS which has to stay in sync with what native knows is focused. We [listen to _onPress](https://github.com/facebook/react-native/blob/6ba2769f0f92ca75fb0eb60ccb8337920a9c31eb/Libraries/Components/TextInput/TextInput.js#L1103-L1107) and call `TextInputState.focusTextInput` in that handler. However, we don't currently have anything listening to other ways for an input to become focused (like tabbing) so it doesn't end up updating the `currentlyFocusedField`.

We have the same problem with blur that we actually fixed the same way you did here in this PR:
https://github.com/facebook/react-native/blob/6ba2769f0f92ca75fb0eb60ccb8337920a9c31eb/Libraries/Components/TextInput/TextInput.js#L1182-L1189

The reason this solution works is because this function **is** the notification from native that an input was focused or blurred. This solution is *fine* because this updates the `currentlyFocusedID` but isn't great because it both sets that value and **calls the native code to focus or blur again**. Luckily the native code doesn't send an event back to JS if you try to blur an already blurred TextInput otherwise we'd have an infinite loop.

# The correct solution
The correct thing would probably be to have all of this tracking in native code and not in JavaScript code. That's a pretty big change though and very out of scope. Something for our team to keep in mind for the future.

A shorterm term solution would be to refactor `focusTextInput` and `blurTextInput` to pull out the part that sets the `currentlyFocusedID` that we could call from `TextInput` directly from `_onFocus` and `_onBlur`

# Thank you
Thank you for highlighting this bug and giving a really solid repro for me to investigate. I don't think you need to take any action on these "correct" solutions in this PR but I wanted to share all the context back with you that I can. This is also helpful documentation for our team as we work on these components in the future.

Libraries/Components/TextInput/TextInput.js
```diff
@@ -1098,6 +1098,7 @@ const TextInput = createReactClass({
     if (this.props.onFocus) {
       this.props.onFocus(event);
     }
+    this.focus();
```
TheSavior: Can you flip the order and put this above the if statement to be consistent with `_onBlur`?

https://github.com/facebook/react-native/blob/6ba2769f0f92ca75fb0eb60ccb8337920a9c31eb/Libraries/Components/TextInput/TextInput.js#L1182-L1189

I think it's important because if someone checks if they are the currently focused input in the `onFocus` prop they pass in it should be true.

TheSavior: Actually, I'm able to just edit the branch on your remote through the GitHub UI. Neat!

RSNara:
So inside a `ScrollView`, when you click outside the `TextInput`, we grab the currently focused element and unfocus it. All that code is in JS, and we keep a track of which `TextInput` is focused inside `TextInputState`. In `TextInput`, that JS state is updated only by calls to `this.blur()` and `this.focus()`. When Native changes focus or blur, it calls `TextInput._onBlur` and `TextInput._onFocus`. So, we have to call `this.blur()` and `this.focus()`. Otherwise, we won't update `TextInputState`, and the ScrollView (and JS in general) won't know which element is currently focused/blurred.

Is that right, @TheSavior?

TheSavior:
Yes! Exactly! 

Well, kinda exactly. The only part that I'd want to clarify from your comment is:

> In TextInput, that JS state is updated only by calls to this.blur() and this.focus()

All NativeComponents (host component, component using NativeMethodsMixin, or component extending ReactNative.NativeComponent) have a `.focus` and `.blur` method which updates `TextInputState`. 

That means that if you have a `<Switch>` you can call `.focus` on it, which will set it to be the currently focused field in `TextInputState` and then when you click somewhere else in a ScrollView, ScrollResponder will tell `TextInputState` to blur that `<Switch>`. 

Even though it isn't a TextInput it is still gets focus and blur set on it. But to make it worse, since we don't listen to `onFocus` and `onBlur` in the Switch JS code to update TextInputState as well then if native focuses off a Switch, the JS might still think it is focused. That's the same problem this PR is fixing for TextInput.

Doesn't this whole thing just make you feel terrible? 😁

PR #27060, opened 2019-10-30
Bump react to 16.9.0

pvinis:
## Summary

When using `npx react-native init Bla`, the react version that is installed there is 16.8.1 instead of the recommended 16.9.0.

This PR is to fix that.

## Changelog

[General] [Fixed] - Updated React version to 16.9.0

## Test Plan

Run `npx react-native init Bla` using this PR somehow, and see if the correct react version is pulled in.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -18,6 +18,7 @@
     "@babel/runtime": "^7.6.2",
     "@react-native-community/eslint-config": "^0.0.5",
     "babel-jest": "^24.9.0",
+    "eslint": "^6.5.1",
     "jest": "^24.9.0",
     "metro-react-native-babel-preset": "^0.56.0",
     "react-test-renderer": "16.9.0"
--- a/template/package.json
+++ b/template/package.json
@@ -9,7 +9,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "react": "16.8.1",
+    "react": "16.9.0",
     "react-native": "0.61.3"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


pvinis:
🤦‍♂ doing it now

pvinis:
should I also update the other deps too?

thymikee:
Please update the `metro-react-native-babel-preset` to `^0.56.0` as well. Not sure if anything changed since then, because `metro-*` packages are kept in sync, but it's worth to be on the version used by CLI (which is 0.56)

pvinis:
Alright, I upgraded the rest of the deps too, to the versions they should be, and I added the `lint` script since I'm here already.

template/package.json
```diff
@@ -6,20 +6,21 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "test": "jest"
+    "test": "jest",
+    "lint": "eslint ."
```
thymikee: we need to add eslint to dev deps in such case

cpojer:
Hey everyone! React inside of RN can only be updated by the React Native team itself because it has some dependencies with FB infra but also because the React specific code in React Native needs to be updated at the same time.

In this case, the release was cut more than two months ago when 16.8.1 was still current. Updating React without updating the renderers will lead to problems and making such a large change in a patch release of a current stable release is very risky.

We will soon update master to React 16.9.0 and the next stable release will ship with that.

I'm curious however, why did this PR receive so many accepts? Is there a blocking bug in the current version of React in React Native that is causing issues?

gaearon:
OK, that was the missing context. Makes sense then. Thanks for explaining!

cpojer:
Thanks for the thorough explanation.

Issue #26696, opened 2019-10-02
React Native 0.61.1 and iOS 13.1.2 leads to lots of unneeded relaunches of our app


bobber205:
Last week we upgraded to the latest version of React. Around the same time my boss updated his phone to iOS 13.1.2 (got a new phone)

I have a iPhone 5S on my desk with iOS 12.x (whatever the latest in that branch is I updated it today)

I also run the app on my Android 10 phone.

After almost no time at all with the app not focused my boss's phone relaunches the app as if the phone just rebooted. Splash screen and everything. This doesn't happen on the very old iPhone 5.  Not on Android either. Of course this relaunch can happen under certain circumstances that are reasonable. For instance my boss will open our app, open up IG and TW, then go back to our app and it relaunches.

We narrowed it down to our upgrade of .60.x to 0.61.1 last Thursday.

This is a very jarring UX issue. I'm not sure how to produce a reproducible use case since it depends on the latest version of iOS and I'm really busy trying to fix other bugs atm related to the new version of iOS.  Has anyone else experienced this?

radko93:
Are you sure your app is not crashing in the background? Do you have any crash reporting service enabled?

bobber205:
Pretty sure it's not crashing. I've narrowed it down to a commit that was responsible for the 0.61 upgrade -- before that commit doesn't matter how long you put the app in the background (say 10 minutes) it's still active when you come back.

We've noticed as little as 1 minute timed of launching the app going over to Instagram and doing literally nothing for a minute then back to our app, relaunch.

We have no background processes for this app. And while we don't have a ton of devices to test on at the moment we have only noticed it on the latest iOS.

oleksandr-dziuban:
@bobber205 Have almost same issue
I have described it in https://github.com/facebook/react-native/issues/26746

Just app is really crashing on iOS 13.1.2 when device as sleeping or app staying in background
Stracktrace crash report is available in my issue

bobber205:
@oleksandr-dziuban 
![image](https://user-images.githubusercontent.com/5841107/66344894-b9081880-e903-11e9-84b1-c92924f28e84.png)

radko93:
Are you using Websockets @bobber205 @oleksandr-dziuban ?

oleksandr-dziuban:
> Are you using Websockets @bobber205 @oleksandr-dziuban ?

No, I don't use Websockets.

For data transport I use GraphQL based on Apollo Client without Websockets.
[...]

PR #27065, opened 2019-10-30
Reverts "Timing: Fixes timer when app get into background (#24649)"

radko93:
This reverts commit 338298417f8077dee177057c57b38671b4ec8c75

## Summary
This is related #26696 - an issue where the app would be closed immediately after going to background on iOS 13.1/13.2 and was investigated by @minhtc https://github.com/facebook/react-native/issues/26696#issuecomment-548056694. The commit that is being reverted is apparently causing the app to be closed immediately. This has to be reverted in master separately as the file differs there.

## Changelog

[iOS] [Fixed] - Reverts commit causing a crash in the background

## Test Plan

Create an empty app on 0.61.x with a timer, put it in the background.

### DIFF
--- a/React/Modules/RCTTiming.m
+++ b/React/Modules/RCTTiming.m
@@ -97,7 +97,6 @@ static const NSTimeInterval kIdleCallbackFrameDeadline = 0.001;
   NSTimer *_sleepTimer;
   BOOL _sendIdleEvents;
   BOOL _inBackground;
-  UIBackgroundTaskIdentifier _backgroundTaskIdentifier;
 }
 
 @synthesize bridge = _bridge;
@@ -113,7 +112,6 @@ RCT_EXPORT_MODULE()
   _paused = YES;
   _timers = [NSMutableDictionary new];
   _inBackground = NO;
-  _backgroundTaskIdentifier = UIBackgroundTaskInvalid;
 
   for (NSString *name in @[UIApplicationWillResignActiveNotification,
                            UIApplicationDidEnterBackgroundNotification,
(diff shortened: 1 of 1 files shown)


zhongwuzw:
@radko93 Hey, seems iOS13 killed background apps much more aggressively, because we just use `UIApplication background task`, is this means we can't use it anymore? If we create a new iOS project(not integrate RN), just use `UIApplication background task`, is it also crashed?

radko93:
@zhongwuzw I think iOS 13.1 put some stricter limits on how this works. https://forums.developer.apple.com/thread/85066:
> Failing to end a background task is the #1 cause of background task problems on iOS.  This usually involves some easy-to-overlook error in bookkeeping that results in the app beginning a background task and not ending it. 

Maybe we don't stop the background task in time, I'm not sure at this time, but we should investigate later on. You can try to experiment with iOS 13.1.

ex3ndr:
Woah, starting background task and never stop it? Of course iOS will kill your app otherwise. It was the same logic like forever.

radko93:
@cpojer please look here and at #27073

zhongwuzw:
> Woah, starting background task and never stop it? Of course iOS will kill your app otherwise. It was the same logic like forever.

@ex3ndr You can check the commit to see wether we stop it :).

zhongwuzw:
> @zhongwuzw I think iOS 13.1 put some stricter limits on how this works. https://forums.developer.apple.com/thread/85066:
> 
> > Failing to end a background task is the #1 cause of background task problems on iOS.  This usually involves some easy-to-overlook error in bookkeeping that results in the app beginning a background task and not ending it.
> 
> Maybe we don't stop the background task in time, I'm not sure at this time, but we should investigate later on. You can try to experiment with iOS 13.1.

🤔 I don't have iOS13.0 or 13.1 now :), is this not a crash, right? it terminated by system after seconds, and I think we stop the background task in time, we can have two options to fix this: 
1. Just revert it, disable the background timer feature for all iOS system users.
2. Provide an option for Timer module, developer can choose wether to enable/disable the background timer on different iOS version.

@radko93 What do you think?

radko93:
@zhongwuzw we should make the users safe now, then we can bring this back improved.

grabbou:
FYI let's keep the discussion under #27073 - that's the PR against master and will be easier to manage for other maintainers. 

Meanwhile, I think it's fair to revert something that causes a major regression. We can always re-iterate again.

Issue #26280, opened 2019-08-31
Multiple 'Set-Cookie' header values in HTTP response are not respected on Android RN 0.60.5


pinstripe-potatohead:
Multiple 'Set-Cookie' header values in HTTP response are not respected on Android RN 0.60.4 - only the last one shows up in JS. Multiple 'Set-Cookie' headers are concatenated into a string (correctly) on iOS.
 
React Native version: 0.60.5

chenyanfei-m:
Same problem in v0.59.8

ronyv:
any updates on this ? is this a bug with react native or is it not a good practice to set multiple 'Set-Cookie' headers ?

pinstripe-potatohead:
@ronyv Please look at #23005 and ##21795 . AFAIK, multiple `Set-Cookie` headers with is allowed by the spec. This is a bug with React Native Android. Downgrading to v0.59 will allow you to use the `userNativeAccessor` workaround

ronyv:
@chenyanfei-m : what you suggested is just a workaround for RN versions < 0.60. . I am using 0.60.5 and this is not working. I just don't want to downgrade my RN version for resolving this issue. Rather i will consider some other efficient alternative. Hope RN will fix this fundamental issue in the upcoming releases.

ozican:
@ronyv Hello brother. Did you solve the issue? I still fighting with it, and any help would be great.

Issue #23185, opened 2019-01-28
Cookie based authentication issue aggregation + Issues got closed without resolution.


Return-1:
## Environment
Provided in corresponding issues
## Reproducible Demo
Provided in corresponding issues

## Description
Issues closed while still open. Cookie based authentication is at this moment not usable. This is partially due to the following issues:

- https://github.com/facebook/react-native/issues/929
- https://github.com/facebook/react-native/issues/23005

These issues have been closed even though they are still open and very relevant.
There's more but an issue needs to officially be made for them. Unofficially speaking, one of the two platforms, i believe iOS , wont store cookies after app restart.

## Conclusion
In general cookie based authentication is very problematic on multiple levels. If cookie based authentication is claimed to be supported on React Native these issues need attention. Otherwise people need to know before implementing a project using such an authentication mechanism or even before attempting to do as dozens of hours could be spend working on an architecture that is inevitably simply not supported.

This is not a matter of pointing fingers. But it is misleading and very damaging as it has been for me to leave people unaware of all these limitations putting them in a place where they cant explain to the rest of the team why the whole authentication workflow suddenly needs refactoring.

cirediew:
I have the same issue as mentioned in #23005.
Using the fetch functionality to do an authentication post returns a header with two cookies with the same name. On iOS these are concatenated. On Android only the second cookie is returned.

Example using postman:

`set-cookie →token=XXXXXXXXXXXXXXXXXXXXXX; expires=Mon, 27-Jan-2020 12:45:43 GMT; Max-Age=31449600; Path=/`

`set-cookie →session=YYYYYYYYYYYYYYYYYYYYYYYYY; httponly; Path=/`

Example RN iOS:
`"set-cookie": "token=XXXXXXXXXXXXXXXXXXXXXX; expires=Mon, 27-Jan-2020 12:47:58 GMT; Max-Age=31449600; Path=/, session=YYYYYYYYYYYYYYYYYYYYYYYYY; httponly; Path=/"`

Example RN Android:
`"set-cookie": "session=YYYYYYYYYYYYYYYYYYYYYYYYY; httponly; Path=/"`



  >React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (4) x64 Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
      Memory: 4.83 GB / 15.68 GB
    Binaries:
      npm: 6.0.0 - C:\Program Files\nodejs\npm.CMD

[...]

Issue #21795, opened 2018-10-15
HTTP response headers with the same key are not processed properly. [The expected behaviour is that the values of the identical keys are combined to a string separated by a comma]


mdloselo:
## Environment

  OS: macOS High Sierra 10.13.6
  Node: 8.12.0
  Yarn: Not Found
  npm: 6.4.1
  Watchman: 4.9.0
  Xcode: Xcode 10.0 Build version 10A255
  Android Studio: 3.2 AI-181.5540.7.32.5014246

Packages: (wanted => installed)
  react: 16.3.1 => 16.3.1
  react-native: https://github.com/expo/react-native/archive/sdk-30.0.0.tar.gz => 0.55.4

## Description

I have an issue supporting multiple cookies passed on the response header on Android. I tracked the issue to `line 607` in `translateHeaders` function in `NetworkingModule.java` file from the link below.

https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java

I debugged and realised that the code that checks if the headerMap of type `WritableMap` has the header key already in it is always failing and defaulting to the "Else" part which always replaces the previous key and value every time instead of combining previous key's value string separated by a comma as it is supposed to. Due to this, only one cookie is always passed to React side.

Furthermore, the code I mentioned above use `ReadableNativeMap` class which holds the actual key-value map as an instance variable. `getLocalMap` function in `ReadableNativeMap` class is used to retrieve the "mLocalMap" variable but always return an object that doesn't have values hence the code inside `NetworkingModule` is not doing what it is supposed to.

Lastly, I'm aware that my environment info state that I'm using react-native from Expo https://github.com/expo/react-native, however the code that I found an issue on, Expo also use it as is from https://github.com/facebook/react-native hence I created an issue here.

## Reproducible Demo
Pass multiple headers with the same key from a network request.

In my case I'm getting response headers with multiple cookies with key "Set-Cookie" from a network request, but after it has been processed by the code I mentioned above, there is always one.

Sample response header with multiple cookies similar to what get from the server.

`Server: Example
Date: Tue, 16 Oct 2018 10:35:21 GMT
Content-Type: application/json;charset=UTF-8
Connection: keep-alive
Expires: 0
X-B3-TraceId: d6745404c1d2a7fb
[...]

PR #27066, opened 2019-10-31
Fix multiple `set-cookie` not aggregated correctly in response headers

vinceplusplus:
## Summary

Multiple `set-cookie` headers should be aggregated as one `set-cookie` with values in a comma separated list. It is working as expected on iOS but not on Android. On Android, only the last one is preserved

The problem arises because `NetworkingModule.translateHeaders()` uses `WritableNativeMap` as the translated headers but uses both native and non-native methods. The mixup causes out of sync data that both sets of methods do no agree. A simple fix is to use `Bundle` as the storage and only convert it to `WritableMap` at the end in one go

Related issues: #26280, #21795, #23185

## Changelog

[Android] [Fixed] - Fix multiple headers of the same name (e.g. `set-cookie`) not aggregated correctly

## Test Plan

A mock api, https://demo6524373.mockable.io/, will return 2 `set-cookie` as follows:
```
set-cookie: cookie1=value1
set-cookie: cookie2=value2
```

Verify the following will print the `set-cookie` with a value `cookie1=value1, cookie2=value2`
```javascript
  fetch('https://demo6524373.mockable.io/')
    .then(response => {
      console.log(response.headers);
    });
```
On iOS, `set-cookie` will have `cookie1=value1, cookie2=value2` while on Android it will have `cookie2=value2` (preserving only the last one)

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/NetworkingModule.java
@@ -8,6 +8,7 @@
 package com.facebook.react.modules.network;
 
 import android.net.Uri;
+import android.os.Bundle;
 import android.util.Base64;
 import androidx.annotation.Nullable;
 import com.facebook.common.logging.FLog;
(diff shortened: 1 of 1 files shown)


pinstripe-potatohead:
@mdvacca Can this be merged? This issue has been preventing cookie based auth since the changes made in v0.60, and a fix would be much appreciated.

pinstripe-potatohead:
@mdvacca Can this be merged? This issues has been breaking cookie based auth ever since the changes in v0.60, and a fix would be greatly appreciated.

aristech:
Will this fix be available in the next release?

hugomaurer-dd:
Will this be available soon?

ozican:
not working for me

Issue #26995, opened 2019-10-25
【iOS】setInterval/setTimeout causes the app to crash, May be the problem with RCTTiming.m


bashen1:
React Native version:
System:
    OS: macOS 10.15
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 63.55 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.10.0 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.12.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29
      Build Tools: 25.0.2, 27.0.3, 28.0.3, 29.0.1, 29.0.2
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.2 => 0.61.2 
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce

1. react-native init demo
2. modify demo/App.js

```javascript
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 * @flow
 */

import React, {PureComponent} from 'react';
import {
    SafeAreaView,
    ScrollView,
    View,
    Text,
    StatusBar,
} from 'react-native';

class App extends PureComponent {
    constructor(props) {
        super(props);
        this.state = {
            num: 0
        }
    }

    componentDidMount(): * {
        this.timer = setInterval(() => {
            this.setState({
                num: this.state.num + 1
            })
        }, 1000)
    }

    render(): boolean | number | string | React$Element<*> | React$Portal | Iterable | null {
        return (
            <>
                <StatusBar barStyle="dark-content"/>
                <SafeAreaView>
                    <ScrollView
                        contentInsetAdjustmentBehavior="automatic"
                       >
                      <View>
                        <Text>{this.state.num}</Text>
                      </View>
                    </ScrollView>
                </SafeAreaView>
            </>
        )
    }
}

export default App;
```
[...]

Issue #25083, opened 2019-05-30
[iOS] JS doesn't keep running in background if it's production build


craftzdog:
I have an editor app that syncs with server. I would like to keep it running for a few seconds to finish syncing after it becomes inactive in background.
I managed to do that by adding following code to `AppDelegate.m`:

```objective-c
- (void)applicationDidEnterBackground:(UIApplication *)application
{
  UIBackgroundTaskIdentifier taskId = [application beginBackgroundTaskWithExpirationHandler:nil];
  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
    // do nothing for 4.7 seconds
    [NSThread sleepForTimeInterval:4.7f];
    [application endBackgroundTask:taskId];
  });
}
```

While it works well in debug build, I encountered that it does not work in production build (TestFlight).
Seems like JS stops immediately after becoming inactive whereas the background task runs as expected on native side.
I suspect it's a bug and tried to fix but couldn't.
Could you help me out? Where should I look into?
I'm happy to contribute.

React Native version: 0.59.2

```
react-native info
info
  React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
      Memory: 21.46 GB / 56.00 GB
      Shell: 3.0.2 - /usr/local/bin/fish
    Binaries:
      Node: 11.12.0 - ~/.nvm/versions/node/v11.12.0/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.nvm/versions/node/v11.12.0/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 22, 23, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 26.0.3, 27.0.3, 28.0.3
        System Images: android-22 | Google APIs Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-Q | Google Play Intel x86 Atom
    IDEs:
[...]

PR #27073, opened 2019-10-31
Reverts "Timing: Fixes timer when app get into background (#24649)"

radko93:
## Summary
This PR reverts commit https://github.com/facebook/react-native/commit/338298417f8077dee177057c57b38671b4ec8c75 that is causing #26696 #26995.
> app would be closed immediately after going to background on iOS 13.1/13.2 and was investigated by @minhtc https://github.com/facebook/react-native/issues/26696#issuecomment-548056694. The commit that is being reverted is apparently causing the app to be closed immediately. This has to be reverted in master separately as the file differs there.

Similar PR for 0.61. branch https://github.com/facebook/react-native/pull/27065

## Changelog

[iOS] [Fixed] - Fix apps crashing on iOS 13.x when running timer in the background

## Test Plan

Try [this](https://github.com/facebook/react-native/commit/338298417f8077dee177057c57b38671b4ec8c75#commitcomment-35745287) snippet on iOS 13.1/13.2, the app should not crash anymore

### DIFF
--- a/React/CoreModules/RCTTiming.mm
+++ b/React/CoreModules/RCTTiming.mm
@@ -105,7 +105,6 @@ static const NSTimeInterval kIdleCallbackFrameDeadline = 0.001;
   NSTimer *_sleepTimer;
   BOOL _sendIdleEvents;
   BOOL _inBackground;
-  UIBackgroundTaskIdentifier _backgroundTaskIdentifier;
   id<RCTTimingDelegate> _timingDelegate;
 }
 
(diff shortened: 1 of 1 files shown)


cpojer:
Duplicate of https://github.com/facebook/react-native/pull/27065

grabbou:
Not really. The #27065 is towards a 0.61-stable branch (so we can have easier time cherry-picking). Here, it's something that should be imported if we revert from master.

cpojer:
Thanks for the explanation, I didn't realize they were meant for different branches, my mistake.

cristianoccazinsp:
Sorry to bring this up, does this bug only affect RN 0.61? I'm seeing lots of strange "no background task exists with identifier ..." logs with RN 0.60 and iOS 13.x. These logs were not here before. I wonder if these background timers have anything to do with it.

craftzdog:
Hi, I'm from https://github.com/facebook/react-native/issues/25083.
@zhongwuzw I upgraded my RN project and found that the patch got reverted. Now, how do you let JS keep running for a few seconds in background on iOS? Thanks!

PR #27110, opened 2019-11-04
Added AccessibilityLabel support to Picker

AmazonKevin:
## Summary

With a Picker we would like to allow accessibility labels to be passed as a prop for situations where we want go give more detail. For example if we have a number picker that will be used for a timer instead of just saying 3, we might want to say 3 hours.

## Changelog

[CATEGORY] [TYPE] - Message

## Test Plan


jamidwyer:
Hi @AmazonKevin, I need this on the project I'm working on as well, and I'm interested in why it was closed without a merge. We can work around the issue in my project, but it would be great for accessibility to be built into RN as you've done here.

I'd be happy to open a similar PR if the issue for you was time to follow up.

PR #27115, opened 2019-11-05
Experimental VirtualList Optimization

### LINKED ISSUE #20174: FlatList component inefficient function passing for CellRenderer

azizhk:
This is VirtualList Optimization done by @christianbach in #21163 enabled behind a prop: `experimentalVirtualizedListOpt` with backward compatibility.

## Summary

Fixes #20174 based of @jasekiw pull request #20208

## Changelog

// TODO:

[CATEGORY] [TYPE] - Message

## Test Plan

// TODO:
Add tests related to backward compatibility. (Will need help)


christianbach:
@cpojer this seems like a more reasonable approach. What can we do to move this along?

dominiczaq:
Is there a way to help get this merged?

abmantis:
This is open for 9 months already. Any way to get some attention to this?

ksitko:
Is there any reason why this has been left open?

Libraries/Components/ScrollView/ScrollViewStickyHeader.js
```diff
@@ -161,4 +167,24 @@ const styles = StyleSheet.create({
...
 
-module.exports = ScrollViewStickyHeader;
+class ScrollViewStickyHeaderWrapped extends React.Component<Props, State> {
+  _childRef: ?React.ElementRef<typeof ScrollViewStickyHeader>;
+  setNextHeaderY(y: number) {
+    this._childRef && this._childRef.setNextHeaderY(y);
+  }
+  render(): React.Node {
+    return (
+      <ExperimentalVirtualizedListOptContext.Consumer>
```
lunaleaps, line 177: Some thoughts about how we do experimentation with this. 

I'm hesitant about this approach because we're wrapping a Context around everyone's sticky header. 
I'm thinking we approach this like the component re-factors we've been doing where we create an injection that is up to product teams to include in their bundle. The injection is read then instead of a new prop `experimentalVirtualizedListOpt`. 

See how we're doing Animated refactor: 
* Injection: https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Animated/createAnimatedComponentInjection.js
Then in our Bundle initialization we have:
```
import createAnimatedComponentInjection from 'createAnimatedComponentInjection';
import createAnimatedComponent_EXPERIMENTAL from 'createAnimatedComponent_EXPERIMENTAL';
...

  if (<our experiment framework says you're in this test>) {
    createAnimatedComponentInjection.inject(
      createAnimatedComponent_EXPERIMENTAL,
    );
  }
```

What do you think about this approach? I'm happy to import and make these changes

lunaleaps (changes requested):

Libraries/Components/ScrollView/ScrollViewStickyHeader.js
azizhk: > What do you think about this approach? I'm happy to import and make these changes

I didn't quite get you, but am happy to go through your PR.

lunaleaps:
@azizhk What kind of tests do you need help writing for this?

Libraries/Lists/VirtualizedList.js
```diff
@@ -819,27 +831,46 @@ class VirtualizedList extends React.PureComponent<Props, State> {
...
+        fillRateHelper: this._fillRateHelper,
+        horizontal: horizontal,
+        index: ii,
+        inversionStyle: inversionStyle,
+        item: item,
+        key: key,
+        prevCellKey: prevCellKey,
+        onUpdateSeparators: this._onUpdateSeparators,
+        onUnmount: this._onCellUnmount,
+        extraData: extraData,
```
lunaleaps, line 847: I believe this would be a new prop we're passing through in the "non-experiment" path. Can we update this to ensure we're not changing original behavior?

lunaleaps (changes requested):

lunaleaps:
@azizhk Here is what I meant by using an injection: https://github.com/facebook/react-native/pull/31947/files

Somewhere in your bundle initialization, you can enable the experiment by 
```
import {setUseVLOptimization} from 'react-native/Libraries/Lists/VirtualizedListInjection';
...
if (yourExperimentFramework.isInExperiment) {
   setUseVLOptimization();
}
```

azizhk:
> @azizhk What kind of tests do you need help writing for this?

Basically I need to add tests which ensure that we are not re-rendering the rows. So something like re-render the virtual list, don't expect the cells to be rendered again.

lunaleaps:
@azizhk's comment on https://github.com/facebook/react-native/pull/31947/files (a demonstration PR)
> Ah, I see. I love this approach. 😍 Less Nesting, Less Props.
> Just that now in the same tree, you can't have two versions of VirtualList one with optimization and one without. (Which I am totally fine with)

Mm that's a good point. I think since this is a question of optimizing overall that's a fine tradeoff. We'll want to run an experiment internally.. to be honest I'm not sure I understand the optimization... might have to scan the issue history a bit more closely or if you could give me TL;DR?

PR #27217, opened 2019-11-13
Remove requestAnimationFrame when focusing input on mount

janicduplessis:
## Summary

When using `react-native-screen` which uses native view controller animations for navigation `TextInput` with `autoFocus` causes a weird animation glitch.

Removing the requestAnimationFrame will cause the focus command to be sent in the same batch as starting screen transitions which fixes the issue.

It is unclear why the rAF was added in the first place as it was part of the initial RN open source commit. If someone at facebook has more context that would be great to make sure it doesn't cause unintended side effects.

Credits to @kmagiera for figuring out this

## Changelog

[General] [Fixed] - Remove requestAnimationFrame when focusing input on mount

## Test Plan

- Tested in an app using react-native-screen to make sure the animation glitch is fixed
- Tested in RNTester to make sure it doesn't cause other issues when not using react-native-screens

Before:

After:

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -702,14 +702,25 @@ function useFocusOnMount(
     // Since initialAutoFocusValue and inputRef will never change
     // this should match the expected behavior
     if (initialAutoFocusValue.current) {
-      const rafId = requestAnimationFrame(() => {
+      const focus = () => {
         if (inputRef.current != null) {
           inputRef.current.focus();
         }
-      });
+      };
+
+      let rafId;
+      if (Platform.OS === 'android') {
+        // On Android this needs to be executed in a rAF callback
+        // otherwise the keyboard opens then closes immediately.
+        rafId = requestAnimationFrame(focus);
+      } else {
+        focus();
+      }
 
       return () => {
-        cancelAnimationFrame(rafId);
+        if (rafId != null) {
+          cancelAnimationFrame(rafId);
+        }
       };
     }
   }, [initialAutoFocusValue, inputRef]);


TheSavior:
I am pretty sure I remember this being needed in order to work in RNTester, although you mention explicitly that it worked there.

It very likely could be a race condition resolved with rAF as calling focus goes into native which has to cross the async bridge which might require certain things on the native side to have run before t he OS can give the input focus.

Im not exactly sure how we get confidence in this in the meantime. Maybe we just merge and see if we get any issues? I would like to take a closer look though when I get a chance, likely on Friday.

I’m surprised that this has such a negative effect on Screens. From my uninformed view it kinda seems like the fixes for these problems should be separate

janicduplessis:
@TheSavior I actually did not test android, will do that too.

Looking at the iOS code I don't see how a race could happen since it calls UIManager.focus which then uses `addUIBlock` to focus the input. All the operation should happen in order so the new input will always be created and attached before focus is called. On android it uses dispatchViewManagerCommand which uses UIViewOperationQueue so it will also be properly synced with view creating operations.

Might have been a workaround for an old race condition bug that was since then fixed.

It seems like a bug with UIKit, It must assume keyboard focus will happen in the same event loop as the transition. Normally you'd call `becomeFirstResponder` in `viewWillAppear`. This will cause the keyboard to slide in from right to left smoothly with the new screen animation. When removing the raF it will cause pretty much the same thing to happen.

TheSavior:
Did you get a chance to test this on Android?

janicduplessis:
@TheSavior Just tested on Android and it causes the keyboard to open then close immediately :(

This explains why this was added. What about we document the need for this rAF and add a platform check so we only do it on Android?

janicduplessis:
Updated the diff, let me know if that seems like a reasonable fix.

TheSavior:
Since this is adding a platform check we probably need to check if this change is valid for React-Native Web and Windows / MacOS. Can you try Web and I'll see if someone from Microsoft can check Windows / Mac?

janicduplessis:
@TheSavior hahaha yeah you can land this on top of your text input changes.

TheSavior:
I'd like to keep this open for tracking if you don't mind so we don't forget about this

TheSavior:
Alright, my refactor to hooks has landed and we haven't heard reports yet of a problem internally. 

Would you like to reapply these changes on top? I'm fine with the platform check

janicduplessis:
@TheSavior 👍 Updated the PR

PR #27218, opened 2019-11-13
Fix RNTester networking conversion to TurboModule

janicduplessis:
## Summary

Now that the networking module is a TurboModule request handlers need to be passed at module creation. The code was missing for RNTester.

## Changelog

[Internal] [Fixed] - Fix RNTester networking conversion to TurboModule

## Test Plan

Tested that RNTester XMLHttpRequest example now works on iOS


janicduplessis:
cc @RSNara

RSNara:
Thank you for fixing this, @janicduplessis! 😁

PR #27268, opened 2019-11-19
Summary: Added accessibility label prop to Picker component 

AmazonKevin:
## Summary

With a Picker we would like to allow accessibility labels to be passed as a prop for situations where we want go give more detail. For example if we have a number picker that will be used as a timer instead of just saying 3, we might want to say 3 hours.

## Changelog

## Test Plan


jamidwyer:
You are awesome for re-opening this! I'm gonna see if I can do a PR against your branch in the next hour to get you the lint fixes and whatnot.

AmazonKevin:
> You are awesome for re-opening this! I'm gonna see if I can do a PR against your branch in the next hour to get you the lint fixes and whatnot.

Thank you! Just wanted to check in, did you ever do the PR? No rush if you have not had the time.

PR #27280, opened 2019-11-20
do not start packager when building in release configuration

mo22:
## Summary

Right now building a project in release (Xcode/Build/Archive or fastlane release builds) also spawn a terminal window running metro packager.

This is not required as a release build does not even try to connect to the packager.

This can also cause problems with headless CI builds.

Note: The Start Packager script is getting quite long and is duplicated between the normal and the tvOS targets. It might make sense to move it next to ./scripts/react-native-xcode.sh - like ./scripts/start-packager-xcode.sh. Let me know if I should prepare this.

## Changelog

[CATEGORY] [TYPE] - Message
[IOS] [Changed] - Start Packager Run Script phase: do not spawn metro bundler when building release builds.

## Test Plan

- Test release build in Xcode
- Test debug build in Xcode

### DIFF
--- a/template/ios/HelloWorld.xcodeproj/project.pbxproj
+++ b/template/ios/HelloWorld.xcodeproj/project.pbxproj
@@ -193,7 +193,7 @@
 			dependencies = (
 			);
 			name = HelloWorld;
-			productName = "HelloWorld";
+			productName = HelloWorld;
 			productReference = 13B07F961A680F5B00A75B9A /* HelloWorld.app */;
 			productType = "com.apple.product-type.application";
 		};
@@ -358,7 +358,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
+			shellScript = "if [[ \"$CONFIGURATION\" == *Release* ]]; then\n  exit 0\nfi\nexport RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
 			showEnvVarsInLog = 0;
 		};
 		FD10A7F122414F3F0027D42C /* Start Packager */ = {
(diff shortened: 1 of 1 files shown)


TheSavior:
@mo22, can you sign the CLA?

mo22:
> @mo22, can you sign the CLA?

done!

template/ios/HelloWorld.xcodeproj/project.pbxproj
```diff
@@ -377,7 +377,7 @@
...
+			shellScript = "if [[ \"$CONFIGURATION\" == *Release* ]]; then\n  exit 0\nfi\nexport RCT_METRO_PORT=\"${RCT_METRO_PORT:=8081}\"\necho \"export RCT_METRO_PORT=${RCT_METRO_PORT}\" > \"${SRCROOT}/../node_modules/react-native/scripts/.packager.env\"\nif [ -z \"${RCT_NO_LAUNCH_PACKAGER+xxx}\" ] ; then\n  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then\n    if ! curl -s \"http://localhost:${RCT_METRO_PORT}/status\" | grep -q \"packager-status:running\" ; then\n      echo \"Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly\"\n      exit 2\n    fi\n  else\n    open \"$SRCROOT/../node_modules/react-native/scripts/launchPackager.command\" || echo \"Can't start packager automatically\"\n  fi\nfi\n";
```
thymikee: This will be painful to upgrade for users. Can we at least move these scripts to files that are only linked through Xcode? This way we can modify them in future with no upgrade cost.

kelset: Yeah I agree about having as separate script 👍

TheSavior:
I’m gonna close this PR then in favor of moving them to a script in core that can be easily modified later. @thymikee, thanks for thinking about the upgrade experience!

kelset:
> I’m gonna close this PR then in favor of moving them to a script in core that can be easily modified later

sorry, probably my brain not working atm - but this means @mo22 should open a new PR to do that?

TheSavior:
Yeah, @mo22 or anyone else

Krizzu:
Going to pick this up and open PR

mo22:
> Going to pick this up and open PR

Thank you! I have the PR half-finished

scripts/start-packager-xcode.sh:
```
#!/bin/sh

if [[ "$CONFIGURATION" == *Release* ]]; then
  exit 0
fi

RCT_SCRIPT_DIR="$( dirname "$0" )"
export RCT_METRO_PORT="${RCT_METRO_PORT:=8081}"

echo "export RCT_METRO_PORT=${RCT_METRO_PORT}" > "${RCT_SCRIPT_DIR}/.packager.env"

if [ -z "${RCT_NO_LAUNCH_PACKAGER+xxx}" ] ; then
  if nc -w 5 -z localhost ${RCT_METRO_PORT} ; then
    if ! curl -s "http://localhost:${RCT_METRO_PORT}/status" | grep -q "packager-status:running" ; then
      echo "Port ${RCT_METRO_PORT} already in use, packager is either not running or not running correctly"
      exit 2
    fi
  else
    open "${RCT_SCRIPT_DIR}/launchPackager.command" || echo "Can't start packager automatically"
  fi
fi
```

and the appropriate change in the Xcode template.

In scripts/react-native-xcode.sh I added:
```
[...] fi

if [[ "$CONFIGURATION" == *Debug* ]]; then
  if [ -e "$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH/main.jsbundle" ]; then
    echo "Debug configuration and main.jsbundle already exists; skipping."
    exit 0;
  fi
fi

[...] if [[ "$SKIP_BUNDLING" ]]; then
```

That will not run the bundler each time a development build is put on the phone, as the development build will only work when connected the the packager.

Does it make sense to open the packager from gradle also?

Krizzu:
> That will not run the bundler each time a development build is put on the phone, as the development build will only work when connected the the packager.

Wouldn't that conflict with re-installation after changing native code?

> Does it make sense to open the packager from gradle also?

IMO, that'd be a nice feature.

mo22:
> > That will not run the bundler each time a development build is put on the phone, as the development build will only work when connected the the packager.
> 
> Wouldn't that conflict with re-installation after changing native code?

Nope, the native code is compiled by Xcode when changed. But in development builds the device always pulls the latest javascript version from the bundler running on the host, so rebuilding the main.jsbundle on each compilation does not make sense. The cleanest solution would be if the main.jsbundle would only contain the necessary code to pull the real code from the bundler.

Krizzu:
Right, that make sense, so I'll hold with my PR for this.

mo22:
I could create a quick PR today (skip bundling on debug builds; do not spawn bundler on release build).

For the gradle integration I probably need a while.

On a different topic - I always use a small run code phase and gradle snippet to set the build number / version etc. from package.json. Does it make sense to include that in the default template?

bitcrumb:
@mo22 Did you ever follow up on this with a separate PR? This seems like a really nice feature.

Also, it seems that the logic with regards to the custom RCT_METRO_PORT (which is part of the script) is broken. This environment variabel used to be passed on via the .package.env file, but somehow this got stripped out in this commit:
https://github.com/facebook/react-native/commit/9ece5bda9bd035c31c3b975d1a8b66574dd43c4e#diff-6ca7c99209bdf630550bb9e2946ce8611948c5a23b32ffb25028792ef5d48b8dL14

Would be nice if this could be fixed as well.

PR #27293, opened 2019-11-21
Add missing accessibility props to Touchables

xuelgong:
## Summary

The following accessibility properties was added for view but not for Touchables - importantForAccessibility, accessibilityLiveRegion, accessibilityViewIsModal and accessibilityElementsHidden. This PR is to extend the support for all touchables.

## Changelog

[General] [Added] - Add missing accessibility props on Touchables

## Test Plan

Tested in RNTester app.


TheSavior:
This is failing internally due to some other files that need to get updated with the same changes. @yungsters will be doing that and then landing this. Thanks for contributing!

PR #27295, opened 2019-11-21
use iPhone X instead of iPhone 6s

TheSavior:
Summary: Changelog: [Testing] [iOS] Use iPhone X in end to end tests since newer versions of iOS do not run on iPhone 6s

Reviewed By: TheSavior

Differential Revision: D18645860


hramos:
Closing as we will likely make changes to this as part of the update to use Xcode 11 in our tooling.

PR #27311, opened 2019-11-22
Add `scripts/generate-rncore.sh` to npm module

sandfox:
the `react-native/ReactCommon/React-Fabric.podspec` references this file
https://github.com/facebook/react-native/blob/master/ReactCommon/React-Fabric.podspec#L33
which causes various cocoapods commands to crash when trying to read the `podspec`

```
[!] Invalid `React-Fabric.podspec` file: No such file or directory @ rb_sysopen - ../scripts/generate-rncore.sh.
```

## Changelog

[CATEGORY] [TYPE] - Message

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "README.md",
     "rn-get-polyfills.js",
     "scripts/compose-source-maps.js",
+    "scripts/generate-rncore.sh",
     "scripts/ios-configure-glog.sh",
     "scripts/ios-install-third-party.sh",
     "scripts/launchPackager.bat",


TheSavior:
It doesn't look like you have listed a test plan in your PR. Can you document how you know that this change fixes the problem? From a brief glance I'm not sure I understand how it could work because the `generate-rncore.sh` file references some other scripts in `packages` which are also not included in the npm package and probably shouldn't be as those will be separate npm packages once Fabric is ready for open source.

sandfox:
I don't have a test plan for this, and I'm not sure how to go about here.

At a very basic level, running the following command
```
pod ipc spec ../node_modules/react-native/ReactCommon/React-Fabric.podspec
```
gives the following error
```

[!] Invalid `React-Fabric.podspec` file: No such file or directory @ rb_sysopen - ../scripts/generate-rncore.sh.

 #  from /Users/sandfox/code/some-project/node_modules/react-native/ReactCommon/React-Fabric.podspec:34
 #  -------------------------------------------
 #    s.source                 = source
 >    s.prepare_command        = File.read("../scripts/generate-rncore.sh")
 #    s.source_files           = "dummyFile.cpp"
 #  -------------------------------------------
```
Copying the `generate-rncore.sh` into the expected location allows the command to run successfully. 
`generate-rncore.sh` itself doesn't actually get executed by cocoapods (or possibly anything), but it is read as string when cocoapods "runs" the  podspec and used to calculate the checksum of the dependency, so it's just necessary that it, or something exists at `../scripts/generate-rncore.sh`. I'm assuming that internally at least the prepare command is used and therefore needs to exist in the podspec.

Totally open to other suggestions or ideas here.

PR #27316, opened 2019-11-23
Removed type-check in App.js file

pdlanl:
## Summary

'types' can only be used in a .ts file. App.js file inside templates was type checking for ReactNode

## Changelog

[General] [Fixed] - Removed type-check in App.js file

### DIFF
--- a/template/App.js
+++ b/template/App.js
@@ -24,7 +24,7 @@ import {
   ReloadInstructions,
 } from 'react-native/Libraries/NewAppScreen';
 
-const App: () => React$Node = () => {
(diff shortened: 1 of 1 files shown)


TheSavior:
Try running `yarn flow`

pdlanl:
@TheSavior Yeah, removing the type check from `template/App.js` and running `yarn flow` produced no errors

TheSavior:
I tried this change internally at Facebook and Flow errors because it expects all file exports to be annotated with types. 

I'm not sure what we should do here...

PR #27328, opened 2019-11-25
Replcae componentWillMount with UNSAFE_componentWillMount in AnimatedComponent

retyui:
## Changelog

[CATEGORY] [TYPE] - Message


TheSavior:
This changed it from componentWillUnmount to UNSAFE_componentWillMount, unmount -> mount. Was that intentional? I don’t think that’s what we want. 


componentWillUnmount does not need a prefix as it is still supported.

PR #27342, opened 2019-11-26
Support for accessibility Label prop to the Picker component

KevinGVargas:
## Summary
With a Picker we would like to allow accessibility labels to be passed as a prop for situations where we want go give more detail. For example if we have a number picker that will be used for a timer instead of just saying 3, we might want to say 3 hours.

## Changelog

## Test Plan

### DIFF
--- a/Libraries/Components/Picker/Picker.js
+++ b/Libraries/Components/Picker/Picker.js
@@ -106,6 +106,10 @@ type PickerProps = $ReadOnly<{|
    * Used to locate this view in end-to-end tests.
    */
   testID?: ?string,
+  /**
+   * The string used for the accessibility label. Will be read once focused on the picker but not on change.
+   */
+  accessibilityLabel?: ?string,
 |}>;
 
 /**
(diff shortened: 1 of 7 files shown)


AmazonKevin:
Made a new PR. Old one is closed. Link https://github.com/facebook/react-native/pull/27268

ReactAndroid/src/main/java/com/facebook/react/views/picker/ReactPickerManager.java
```diff
@@ -44,6 +44,11 @@ public void setPrompt(ReactPicker view, @Nullable String prompt) {
     view.setPrompt(prompt);
   }
 
+  @ReactProp(name = "accessibilityLabel")
```
hramos: One of the internal test failures might be related to this line. Here's what we're seeing internally when I tried to import your changes:

```
stderr: : error: Module com.facebook.react.views.picker.ReactDialogPickerManager has already registered a property named "accessibilityLabel". If you want to override a property, don't add the @ReactProp annotation to the property in the subclass
```

KevinGVargas: I will take a look. I think removing the @ReactProp would fix this according to the log. Thanks for pointing it out!

PR #27366, opened 2019-11-29
Add typescript entry file auto detection to the xcode script

Norris1z:
## Summary
Archiving a project with a `.ts` entry file on Xcode crashes because the `react-native-xcode.sh` script does not auto-detect entry files with the `.ts` extension. 
Screenshot 2019-11-29 at 05 33 51
This PR adds support for `.ts` and `.ios.ts` entry file auto-detection

## Changelog

[General] [Fixed] - Auto detect entry file for projects with a `.ts` index file when Archiving iOS projects

## Test Plan
Archive an iOS project with a `.ts` entry file


dulmandakh:
You can write code in .ts and it just works, because babel will convert .ts to .js. Also RN doesn't support TypeScript, but plain JavaScript.

PR #27372, opened 2019-11-30
moved normalizeColor and changed dependencies

isa3bel:
## Summary

Originally, normalizeColor.js was in Library/Color/ however, I noticed that its tests were in a completely different directly (Library/StyleSheet/__tests__) which was confusing. The other files such as processColor.js, setNormalizedAlphaColor.js had their tests in Library/StyleSheet/__tests__ as well. 

## Changelog

[Internal] [Changed] - Moved normalizeColor.js to a more appropriate directory where its tests live.

## Test Plan

I simply moved a file and changed dependencies. The code should still function as is.


isa3bel:
I need to fix the builds before you can merge! thanks though for approving

isa3bel:
some tests are failing but I feel like they aren't due to my change....?

isa3bel:
was this merged or just closed?

yungsters:
It was merged. 😄

PR #27395, opened 2019-12-03
[iOS] Add autorelease pool for each run loop for JS Thread

### LINKED ISSUE #27327: Obj-C++ TurboModules leak memory – 600 bytes per a method call
Calling TurboModules methods results in an increased memory usage. Increased memory usage may cause the app to crash.

Platform: iOS

React Native version:
```
System:
    OS: macOS 10.15.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Memory: 340.77 MB / 16.00 GB
    Shell: 5.6.2 - /usr/local/bin/zsh
  Binaries:
    Node: 12.7.0 - ~/.asdf/installs/nodejs/12.7.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.10.0 - ~/.asdf/installs/nodejs/12.7.0/bin/npm
  SDKs:
    iOS SDK:
    Android SDK:
      API Levels: 26, 28
      Build Tools: 23.0.1, 26.0.1, 26.0.3, 27.0.3, 28.0.2, 28.0.3
      System Images: android-24 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.2.1/11B500 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5
```

## Steps To Reproduce

Executing the following code block
```js
const SampleTurboModule = TurboModuleRegistry.get('SampleTurboModule');

let i = 0;
while (i < 100000) {
  SampleTurboModule.getNumber(i);
  i += 1;
}
```
results in a 60 MB memory usage increase.

memoryusage

Link to a repository with a reproducing example:

https://github.com/sjchmiela/TurboModulesLeakExample

zhongwuzw:
## Summary

Fixes #27327 , we need to create autorelease pool for each run loop in secondary thread, otherwise application may have memory issues. More details can refer to [CreatingThreads](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/CreatingThreads/CreatingThreads.html)
image

## Changelog

[iOS] [Fixed] - Add autorelease pool for each run loop for JS Thread

## Test Plan

Example can be found in https://github.com/facebook/react-native/issues/27327. No memory spikes any more.

### DIFF
--- a/React/CxxBridge/RCTMessageThread.mm
+++ b/React/CxxBridge/RCTMessageThread.mm
@@ -36,7 +36,12 @@ RCTMessageThread::~RCTMessageThread() {
 
 // This is analogous to dispatch_async
 void RCTMessageThread::runAsync(std::function<void()> func) {
-  CFRunLoopPerformBlock(m_cfRunLoop, kCFRunLoopCommonModes, ^{ func(); });
+  CFRunLoopPerformBlock(m_cfRunLoop, kCFRunLoopCommonModes, ^{
+    // Create an autorelease pool each run loop to prevent memory footprint from growing too large, which can lead to performance problems.
+    @autoreleasepool {
+      func();
+    }
+  });
   CFRunLoopWakeUp(m_cfRunLoop);
 }
 


React/CxxBridge/RCTMessageThread.mm
```diff
@@ -36,7 +36,12 @@
 
 // This is analogous to dispatch_async
 void RCTMessageThread::runAsync(std::function<void()> func) {
-  CFRunLoopPerformBlock(m_cfRunLoop, kCFRunLoopCommonModes, ^{ func(); });
+  CFRunLoopPerformBlock(m_cfRunLoop, kCFRunLoopCommonModes, ^{
+    // Create an autorelease pool each run loop to prevent memory footprint from growing too large, which can lead to performance problems.
+    @autoreleasepool {
+      func();
+    }
+  });
```
fkgozali: This will impact pre-TurboModule system as well. The OOM is only when TM is used right? So I think the fix should be done somewhere around the caller of `runAsync` that's specific to TM?

In other words, this code doesn't leak memory pre-TM, why is that? If we can find the logic in the old system that prevents the leak, we should apply the same fix for TM. So this PR might not address the TM-specific reason for the leak.

zhongwuzw: @fkgozali Actually, pre-TurboModule also has the issue, the reason why pre-TurboModule seems not have is it not creates giant autorelease objects, for example, TurboModule create [one `NSInvocation` (autorelease object) for each method call](https://github.com/facebook/react-native/blob/8797a5cfcb8a3dbd233408fb4ac86d78239f5e3f/ReactCommon/turbomodule/core/platform/ios/RCTTurboModule.mm#L474), but pre-TurboModule [only create one by cache](https://github.com/facebook/react-native/blob/aee88b6843cea63d6aa0b5879ad6ef9da4701846/React/Base/RCTModuleMethod.mm#L508). So pre-TurboModule just hidden the issue in @sjchmiela 's demo, but actually it also has if module method create many autorelease objects.

fkgozali: > but pre-TurboModule only create one by cache

I see, good catch. I think we should aim to get the caching ideally, but that may be out of scope for this issue. If you're up for some refactoring, that would be helpful :).

PR #27425, opened 2019-12-05
Avoid throwing exceptions when the host activity is not FragmentActivity

mganandraj:
## Summary

There are some code in native modules which currently throws exception when the host activity is not a FragmentActivity, even if the native module is not used. This change avoids the throw and fail the promise instead.

## Changelog

There are some code in native modules which currently throws exception when the host activity is not a FragmentActivity, even if the native module is not used. This change avoids the throw and fail the promise instead.

[CATEGORY] [TYPE] - Message

## Test Plan

We've tested the change on MS Office applications, which currently don't use FragmentActivity.


mganandraj:
> @mdvacca is landing this pull request. If you are a Facebook employee, you can view this diff on Phabricator.

Thanks David @mdvacca  !

PR #27426, opened 2019-12-05
Fix Flipper integration on iOS and update it

charpeni:
## Summary

The integration of Flipper on iOS was crashing the template, and `test_ios_e2e` were failing because of it.

Flipper integration must be wrapped inside `FB_SONARKIT_ENABLED`.

Flipper's issues with Yoga and YogaKit were solved in Flipper 0.28.0, so let's update our integration to it, and we'll need to use Swift.

We will be able to ship this in 0.62 (cc @grabbou, @axe-fb, @priteshrnandgaonkar), and it should fix `test_ios_e2e` (cc @hramos) 🤞 .
## Changelog

[iOS] [Fixed] - Fix Flipper integration on and update Flipper to 0.30.0

## Test Plan

- Initialized an app with my branch as source.
- Added `[AppDelegate initializeFlipper:application];` to `didFinishLaunchingWithOptions`.
- Launched Flipper.
- Ran the app (debug).

image

- Ran the app (release).
- Everything succeeds, no Flipper connection available in release.

### DIFF
--- a/template/ios/HelloWorld.xcodeproj/project.pbxproj
+++ b/template/ios/HelloWorld.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		2D02E4BD1E0B4A84006451C7 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		2D02E4BF1E0B4AB3006451C7 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
 		2DCD954D1E0B4F2C00145EB5 /* HelloWorldTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 00E356F21AD99517003FC87E /* HelloWorldTests.m */; };
+		37E9F8DD2399A8DD00F82F3E /* Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37E9F8DC2399A8DD00F82F3E /* Swift.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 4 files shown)


axe-fb:
Thanks for working on this. I have imported this, and should be able to land this soon.

axe-fb:
Should we separate the version update into a separate PR, just to keep this things clean ?

charpeni:
I'm not sure if it will work fine with the original Flipper version as I'm on Xcode 10. 😬

axe-fb:
Thats find them. Just imported it, once it is  reviewed, we can land it.

dulmandakh:
why Swift.swift file?

charpeni:
Flipper requires Swift compiler for YogaKit, and maybe Yoga too as we're using modular header. In order to get it compiled correctly, we need to have at least one Swift file in our project.

Maybe there's a way without this additional Swift file, but I didn't succeed, even with `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES`.

friederbluemle:
👋 Thanks for the fixes - Just a few quick questions related to this:

- How will the now included Swift standard libraries impact binary size (if at all)?
- If the Swift file is removed from an app project (that does not use Flipper), is the compiler smart enough to also exclude the Swift standard libraries, or are additional settings required?
- `SWIFT_VERSION = 5.0` and `SWIFT_OPTIMIZATION_LEVEL = "-Onone"` - Are these the defaults? Can it safely be set to something else?
- `Swift.swift` is the **first and only** Swift file in the _entire_ `react-native` repository (almost 4000 files) - And it sounds like the only reason it's there is to "trick" the compiler. Are there any other potential solutions that achieve the same without the dummy file and embedding the Swift standard libraries?

Sorry for all the questions - Swift newbie here 😇

dulmandakh:
I have same questions too 😇

friederbluemle:
@charpeni @axe-fb - I know there has already been some discussion on Discord, but to close the loop it would be great if you could post a quick reply here too.

dulmandakh:
ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES might add 5mb to an app size

charpeni:
@friederbluemle never be sorry for asking questions!

> * How will the now included Swift standard libraries impact binary size (if at all)?

@dulmandakh is right. We depend on `YogaKit`, which includes [YGLayoutExtensions.swift](https://github.com/facebook/yoga/blob/master/YogaKit/Source/YGLayoutExtensions.swift), and then there's not much we can do for the binary size with Swift, I'm afraid.

But, since Flipper is only running in Debug, we should be able to set `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to `NO` in Release. I'll have a look at this.

___

> * If the Swift file is removed from an app project (that does not use Flipper), is the compiler smart enough also to exclude the Swift standard libraries, or are additional settings required?

Yes, Xcode is smart enough the change all settings related to this. The same thing applies when you add a swift file to a project. I wasn't able to keep that configuration, because the moment I removed the Swift file, Xcode removed everything related to Swift, even if we still have Swift pods. The template folder is not the most pleasant way to deal with that, I'll retry with a sample app and compare the diff with git, it may be more helpful. 😄 

___

> * `SWIFT_VERSION = 5.0` and `SWIFT_OPTIMIZATION_LEVEL = "-Onone"` - Are these the defaults? Can it safely be set to something else?

Yes, it can be safely set to something else. This Podfile script will set the `SWIFT_VERSION` for YogaKit to 4.1 anyway, but I wouldn't go lower than that.

https://github.com/facebook/react-native/blob/438891f35577a7cf3d82e4a10af3f9b000264a33/template/ios/Podfile#L16-L22

___

> * `Swift.swift` is the **first and only** Swift file in the _entire_ `react-native` repository (almost 4000 files) - And it sounds like the only reason it's there is to "trick" the compiler. Are there any other potential solutions that achieve the same without the dummy file and embedding the Swift standard libraries?

Well, some say the only reason I did that was to introduce a Swift file into a Facebook repo silently. But this is not completely true.

___

> it sounds like the only reason it's there is to "trick" the compiler.

Exactly, I wasn't able to keep a Swift compiler, or in that case, standard libraries without a Swift file. Weird, right? I can't believe this is not possible, because having a placeholder Swift file seems really odd to me. I didn't look more than a few minutes into that as the goal was to start testing Flipper with React Native and fix tests. As I said earlier, I'll retry that, and use git to compare changes in the project with and without a Swift file. That way, we may be able to trick the compiler without a Swift file.

Thanks for raising those concerns, I'll have a closer look to make sure we release something enjoyable in RN 0.62.

I hope this answers your questions, feel free to reach out again. Otherwise, I'll let you know or cc you into upcoming pull requests.

rickhanlonii:
@charpeni have you looked into "ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES to NO in Release"? Seems like that would smooth over the app size increase. As for upgrades, how would that flag impact people who already use swift?

PR #25427, opened 2019-06-28
[iOS] Initial UIKitForMac support

radex:
## Summary

This PR adds initial support for Project Catalyst a.k.a. UIKitForMac. This is not yet meant for production, but this is enough for RNTester to successfully compile and mostly work :)

Some APIs are not supported on the Mac -- e.g. telephony, and deprecated APIs are removed on Mac -- those had to be ifdef'd out via platform checks.

The biggest limitation right now is that I couldn't get Web Socket code to successfully compile, and so there are a lot of temporary platform checks  for that , and the RCTWebSocket.xcodeproj is marked as not supporting UIKitForMac. Again -- temporary, until someone with more knowledge knows how to fix this.

https://github.com/react-native-community/discussions-and-proposals/issues/131

## Changelog

[iOS] [Added] - Fixed compilation for macOS (Project Catalyst) -- not meant for production use yet

## Test Plan

- Open RNTester/RNTester.xcodeproj with Xcode 10.2, run it like a normal iOS app -- make sure it compiles and runs correctly (no regression)
- Open the same project with Xcode 11 beta 2 (or higher) on macOS Catalina beta, select "My Mac" as device target, and run -- see that it actually compiles and runs. **Note** there are unfortunately some required steps:
   - change build configuration to Release (because packager doesn't work correctly yet)
   - change development team to yours if Xcode tells you to
   - go to RNTester project → Build phases → Link binary with libraries, and change `platforms` for `libRCTWebSocket.a` to `iOS` (without Mac compatibility). I can't commit that change because it breaks compatibility with earlier Xcode versions

The two extra steps for successful compile will disappear once web socket compilation for Catalyst is fixed

PR #27469, opened 2019-12-10
[iOS] Enabling RCTWebSocket on UIKitForMac (macOS Catalyst)

andymatuschak:
## Summary

In #25427, @radex added initial support for running React Native projects on macOS via Catalyst. However, `RCTWebSocket` was disabled for that target because of some compilation issues. This meant that running projects via a connection to the packager wasn't possible: no live reload, and projects must be run in "Release" mode. It also meant making manual changes to Xcode projects deploying to macOS and scattering a number of conditional checks throughout the codebase.

In this change, I've implemented support for `RCTWebSocket` on the macOS target and re-enabled the affected features. Live reload and the inspector now work for macOS targets. Manual modifications of Xcode build settings are no longer necessary for react-native projects running on macOS.

Screen Shot 2019-12-10 at 8 36 38 AM

### Limitations

There's no binding which displays the developer menu (since there's no shake event on macOS). We'll probably want to add one, perhaps to the menu bar.

I've chosen not to commit the modifications to RNTester which enable macOS support, since that would imply more "official" support for this target than I suspect you all would like to convey. I'm happy to add those chunks if it would be helpful.

## Changelog

[iOS] [Added] - Added web socket support for macOS (Catalyst), enabling debug builds and live reload

## Test Plan

* Open RNTester/RNTester.xcodeproj with Xcode 11.2.1, run it like a normal iOS app -- make sure it compiles and runs correctly (no regression)
* Select "My Mac" as device target, and run. You may need to configure a valid development team to make signing work.
* RNTester should run fine with no additional configuration. Modify a file in RNTester, note that live reload is now working.
* Test the developer inspector. To display the developer menu, you'll need to manually show it; here's an example diff which does that:
```
diff --git a/RNTester/js/RNTesterApp.ios.js b/RNTester/js/RNTesterApp.ios.js
index 8245a68d12..a447ad3b1b 100644
--- a/RNTester/js/RNTesterApp.ios.js
+++ b/RNTester/js/RNTesterApp.ios.js
@@ -19,6 +19,8 @@ const React = require('react');
 const SnapshotViewIOS = require('./examples/Snapshot/SnapshotViewIOS.ios');
 const URIActionMap = require('./utils/URIActionMap');
 
+import NativeDevMenu from '../../Libraries/NativeModules/specs/NativeDevMenu';
+
 const {
   AppRegistry,
   AsyncStorage,
@@ -143,6 +145,7 @@ class RNTesterApp extends React.Component<Props, RNTesterNavigationState> {
 
   UNSAFE_componentWillMount() {
     BackHandler.addEventListener('hardwareBackPress', this._handleBack);
+    NativeDevMenu.show();
   }
 
   componentDidMount() {
```

### DIFF
--- a/Libraries/WebSocket/RCTSRWebSocket.m
+++ b/Libraries/WebSocket/RCTSRWebSocket.m
@@ -14,12 +14,9 @@
 //   limitations under the License.
 //
 
-#if !TARGET_OS_UIKITFORMAC
-
 #import <React/RCTSRWebSocket.h>
 
 #import <Availability.h>
-#import <Endian.h>
 
 #import <Security/SecRandom.h>
 
@@ -580,7 +577,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
     NSMutableData *mutablePayload = [[NSMutableData alloc] initWithLength:sizeof(uint16_t) + maxMsgSize];
     NSData *payload = mutablePayload;
 
-    ((uint16_t *)mutablePayload.mutableBytes)[0] = EndianU16_BtoN(code);
+    ((uint16_t *)mutablePayload.mutableBytes)[0] = NSSwapBigShortToHost(code);
 
     if (reason) {
       NSRange remainingRange = {0};
@@ -749,7 +746,7 @@ static inline BOOL closeCodeIsValid(int closeCode)
     return;
   } else if (dataSize >= 2) {
     [data getBytes:&closeCode length:sizeof(closeCode)];
-    _closeCode = EndianU16_BtoN(closeCode);
+    _closeCode = NSSwapBigShortToHost(closeCode);
     if (!closeCodeIsValid(_closeCode)) {
       [self _closeWithProtocolError:[NSString stringWithFormat:@"Cannot have close code of %d", _closeCode]];
       return;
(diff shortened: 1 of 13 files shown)


andymatuschak:
`ci/circleci: js_coverage` failed with this error:

```
Test Suites: 127 passed, 127 total
Tests:       1 skipped, 1192 passed, 1193 total
Snapshots:   555 passed, 555 total
Time:        106.245s
Ran all test suites.
Done in 107.11s.

/home/circleci/react-native/node_modules/coveralls/bin/coveralls.js:18
        throw err;
        ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}

Exited with code exit status 1
```

That's mysterious to me and strikes me as possibly spurious. If there's something actually broken, a pointer would be helpful!

Libraries/WebSocket/RCTSRWebSocket.m
```diff
@@ -749,7 +746,7 @@ - (void)handleCloseWithData:(NSData *)data
     return;
   } else if (dataSize >= 2) {
     [data getBytes:&closeCode length:sizeof(closeCode)];
-    _closeCode = EndianU16_BtoN(closeCode);
+    _closeCode = NSSwapBigShortToHost(closeCode);
```
radex: It's so simple! I wish I spent 10 more minutes on the problem ;)

radex:
Thank you @andymatuschak for this!

TheSavior:
This is awesome!

Does `CMD+D` on the app bring up the dev menu?

fungilation:
Awesome work indeed!

charpeni:
@TheSavior `CMD+D` doesn't bring up the dev menu. I know what's up, sending a PR in a few minutes.

hramos:
I'm going to import these in order to get a head start with our internal test suite, as well as to get it onto @shergin's review queue.

PR #27482, opened 2019-12-11
[flipper] Fix RNTester integration with Flipper

passy:
Summary:
There was a reflective call to a non-existent class. It did, however,
exist in the template, so I copied it over from there and updated
the references accordingly.

Test Plan:
Built it and ran it. Works again with the latest Flipper desktop app.

Screenshot 2019-12-11 16 02 40

### DIFF
--- a/RNTester/android/app/src/debug/java/com/facebook/react/uiapp/ReactNativeFlipper.java
+++ b/RNTester/android/app/src/debug/java/com/facebook/react/uiapp/ReactNativeFlipper.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+package com.facebook.react.uiapp;
+
+import android.content.Context;
+import com.facebook.flipper.android.AndroidFlipperClient;
+import com.facebook.flipper.android.utils.FlipperUtils;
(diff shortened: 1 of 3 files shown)


RNTester/android/app/src/debug/java/com/facebook/react/uiapp/ReactNativeFlipper.java
```diff
@@ -0,0 +1,73 @@
...
+import com.facebook.flipper.plugins.react.ReactFlipperPlugin;
+import com.facebook.flipper.plugins.sharedpreferences.SharedPreferencesFlipperPlugin;
+import com.facebook.react.ReactInstanceManager;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.modules.network.NetworkingModule;
+import okhttp3.OkHttpClient;
+
+public class ReactNativeFlipper {
+  public static void initializeFlipper(Context context, ReactInstanceManager reactInstanceManager) {
+    if (FlipperUtils.shouldEnableFlipper(context)) {
```
mdvacca: Is it documented how to enable flipper? By default it will not work in debug mode, right?

passy: It should work by default in debug builds. :) This is just an additional check to ensure that this is a) an `IS_INTERNAL_BUILD` b) not an end-to-end test (but that's FB-specific) and c) is on the main process as apps with multiple processes may not map native libraries into their address space.

PR #27484, opened 2019-12-11
macOS: Implement Platform.isMac

charpeni:
## Summary

Implementing `Platform.isMac` to know if we're running on macOS with Catalyst.

⚠️ Note that `Platform.OS` is still returning `ios`, even on macOS.

We'll be able to use `Platform.isMac` to target macOS.

```javascript
import { Platform } from 'react-native';

if (Platform.isMac) {
  // Do something specific to macOS
} else {
 // Do something else for other platforms.
}
```

## Changelog

[iOS] [Added] - Implement Platform.isMac to know if we're running on macOS with Catalyst.

## Test Plan

```javascript
// On iOS
Platform.isMac // false

// On macOS
Platform.isMac // true
```

### DIFF
--- a/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
+++ b/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
@@ -2062,6 +2062,7 @@ namespace JS {
           RCTRequired<NSString *> osVersion;
           RCTRequired<NSString *> systemName;
           RCTRequired<NSString *> interfaceIdiom;
+          RCTRequired<bool> isMac;
         };
 
         /** Initialize with a set of values */
@@ -3671,6 +3672,8 @@ inline JS::NativePlatformConstantsIOS::Constants::Builder::Builder(const Input i
   d[@"systemName"] = systemName;
   auto interfaceIdiom = i.interfaceIdiom.get();
   d[@"interfaceIdiom"] = interfaceIdiom;
+  auto isMac = i.isMac.get();
+  d[@"isMac"] = @(isMac);
   return d;
 }) {}
 inline JS::NativePlatformConstantsIOS::Constants::Builder::Builder(Constants i) : _factory(^{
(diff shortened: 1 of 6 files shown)


dulmandakh:
how about returning false in Platform.android.js

charpeni:
@dulmandakh should we? I think it will be great to compare boolean values instead of falsy/truthy, but we're not doing it for other properties like `isTV`.

Is still working with like this, because `undefined` is falsy.

```javascript
if (!Platform.isMac) {
  // Reached on Android
}
```

dulmandakh:
sure 😇

charpeni:
Thanks for the feedback, @dulmandakh!

Added `Platform.isMac` to `Platform.android.js` in 2765151! Other properties could be handled in other pull requests.

xiao99xiao:
Any update about this PR? I believe this is a must-have for building Catalyst apps.

exotexot:
Would appreciate having this feature!

safaiyeh:
Should this be isCatalyst? Given we have react-native-macos now, this might cause some confusion.

losh11:
Would appreciate this, but as @safaiyeh said it makes much more sense to change Platform.isMac to Platform.isCatalyst.

charpeni:
Thanks for bumping this in my notifications.

I submitted a new PR addressing the concern around [`react-native-macos`](https://github.com/microsoft/react-native-macos): https://github.com/facebook/react-native/pull/29937.

In the meantime, if you need to detect if the environment is Catalyst, I extracted it into an external library available as [`react-native-is-catalyst`](https://github.com/charpeni/react-native-is-catalyst).

sota000:
Hi @charpeni, could you please resolve the merge conflict? After that I will send this to an internal review. Thanks!

sota000:
Actually, it sounds like this PR should be closed, and the new PR should be reviewed. Is that right?

xiao99xiao:
I think so.

sota000:
Closing this PR in favor of #29937. Please reopen if you think otherwise.

PR #27491, opened 2019-12-12
$ENTRY_FILE with ts extension

expouic:
## Summary
Add possible to rename index.js to index.ts when you are using only typescript files on your project

## Changelog

[iOS] [Changed] - Add support to index.ts and index.ios.ts files

## Test Plan
None


cpojer:
We don't currently support this for Android, or for any other language extensions. We recommend keeping the main entry point in JS, and using that to include your entry point written in another language.

PR #27492, opened 2019-12-12
setting this.state.bottom to 0 when keyboard is hidden in the Keyboar…

fredpinon:
## Summary
In the KeyboardAvoidingView, `this._relativeKeyboardHeight` returns the height of the keyboard on keyboard change events. This method seems to return a non-zero value while the keyboard is hidden. This PR ensures that if the keyboard is hidden, the returned value of `this._relativeKeyboardHeight` is dismissed to instead apply `0`. 

## Changelog

[General] [Fixed] - setting this.state.bottom to 0 when keyboard is hidden in the KeyboardAvoindingView

## Test Plan

All tests pass when running `npm test`. This does not affect native components.


yungsters:

PR #27498, opened 2019-12-12
move benchmark from xplat/third-party to third-party

igorsugak:
Reviewed By: pixelb

Differential Revision: D16869800


cpojer:
I'll close this PR because I don't think we'll need it. Once it lands, the change syncs out to GitHub.

PR #27506, opened 2019-12-13
Add support for animating all properties using native driver

chrfalch:
## Summary

Native animation only supports properties that can be updated through the shadow view tree, limiting it to a small subset of properties and styles available for animation. Whenever a developer wants to animate something like size, color etc. they revert to JS based animation leading to complaints about stutters in the UX. 

Inspired by the wide usage of Reanimated I've created a solution that uses the same principles as Reanimated. This is basically two separate paths to update a property - one is for shadow tree properties that can be synchronously updated through `UIManager .synchronouslyUpdateViewOnUIThread`, and one that uses a batched update queue that is updated through `UIManager.updateView`.

This might have some performance issues, but the result will be much smoother and faster than using JS based animation. 

I'm aware that a Fabric based animation library is in the works, but believe that this PR can add value for developers until Fabric is available.

## Changelog

- [iOS][Added] Added support for animating unsupported props with useNativeDriver 
- [General][Added] Added support for animating all props using native driver
- [Android][Added] Added support for animating all props with useNativeDriver 

## Test Plan

- Added examples that shows native animations of size and color in RNTester
- Removed unit tests for illegal native properties


Libraries/Animated/src/__tests__/AnimatedNative-test.js
```diff
@@ -642,19 +642,6 @@ describe('Native Animated', () => {
...
-
-      TestRenderer.create(<Animated.View style={{left}} />);
-
-      const animation = Animated.timing(left, {
-        toValue: 10,
-        duration: 50,
-        useNativeDriver: true,
-      });
-      expect(animation.start).toThrowError(/left/);
-    });
```
JoshuaGross: can we turn this into an affirmative test that animating these properties works now?

Libraries/Animated/src/colorToRgba.js
```diff
@@ -0,0 +1,33 @@
...
+  int32Color = int32Color || 0;
+
+  const r = (int32Color & 0xff000000) >>> 24;
+  const g = (int32Color & 0x00ff0000) >>> 16;
+  const b = (int32Color & 0x0000ff00) >>> 8;
+  const a = (int32Color & 0x000000ff) / 255;
+
+  return `rgba(${r}, ${g}, ${b}, ${a})`;
+}
+
```
JoshuaGross: let's add some basic tests for this?

JoshuaGross: Ah I see that you're just moving this, but if tests don't already exist, it'd be nice to add a few

JoshuaGross:
Hey @chrfalch, thanks for doing this! In general I think this is something I'm supportive of, and seems like a fine thing to do. I'll need to spend some time with this to see if it has implications on native driver animations in Fabric, since they share most code currently. I'll try to set aside some time this week to play around with it.

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -243,13 +253,7 @@ function validateTransform(
 }
 
 function validateStyles(styles: {[key: string]: ?number, ...}): void {
```
janicduplessis: I guess we can just remove this now.

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -159,7 +160,7 @@ const API = {
  * In general native animated implementation should support any numeric property that doesn't need
  * to be updated through the shadow view hierarchy (all non-layout properties).
  */
-const STYLES_WHITELIST = {
+const SHADOW_VIEW_PROPS = {
```
janicduplessis: I think this name is wrong, it is the other way around. These props do not require to be update through the shadow view. Maybe call it something like `NON_LAYOUT_PROPS` or `UI_THREAD_PROPS`.

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -312,10 +316,18 @@ function transformDataType(value: number | string): number | string {
     const radians = (degrees * Math.PI) / 180.0;
     return radians;
   } else {
-    return value;
+    return colorToRgba(value);
   }
 }
 
+function configureProps() {
```
janicduplessis: Is there a reason for configuring these from JS instead of just having the list live in native implementation of each platform?

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -74,6 +101,15 @@ - (BOOL)isNodeManagedByFabric:(nonnull NSNumber *)tag
   return false;
 }
 
+- (void)enqueueUpdateViewOnNativeThread:(nonnull NSNumber *)reactTag
```
janicduplessis: What about `enqueueUpdateViewOnUIManager`? As in use the ui manager to update this view props.

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
```diff
@@ -386,6 +409,14 @@ public void removeAnimatedEventFromView(int viewTag, String eventName, int anima
     }
   }
 
+  public void configureProps(Set<String> shadowViewProps) {
+    this.shadowViewProps = shadowViewProps;
+  }
+
+  public void enqueueUpdateViewOnNativeThread(int viewTag, WritableMap nativeProps) {
```
janicduplessis: `enqueueUpdateViewOnUIManager` see prev comment

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -61,10 +76,22 @@ - (instancetype)initWithBridge:(nonnull RCTBridge *)bridge
     _animationNodes = [NSMutableDictionary new];
     _eventDrivers = [NSMutableDictionary new];
     _activeAnimations = [NSMutableSet new];
+    _shadowViewProps = [[NSArray alloc] init];
+    _operationsInBatch = [NSMutableArray new];
```
janicduplessis: `_uiManagerOperationQueue` to go with `enqueueUpdateViewOnUIManager`

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
```diff
@@ -58,10 +64,27 @@
...
+
+  private final class NativeUpdateOperation {
+    public int mViewTag;
+    public WritableMap mNativeProps;
+    public NativeUpdateOperation(int viewTag, WritableMap nativeProps) {
+      mViewTag = viewTag;
+      mNativeProps = nativeProps;
+    }
+  }
+  private Queue<NativeUpdateOperation> mOperationsInBatch = new LinkedList<>();
```
janicduplessis: `mUIManagerOperationQueue`

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
```diff
@@ -58,10 +64,27 @@
   // Used to avoid allocating a new array on every frame in `runUpdates` and `onEventDispatch`.
   private final List<AnimatedNode> mRunUpdateNodeList = new LinkedList<>();
 
-  public NativeAnimatedNodesManager(UIManagerModule uiManager) {
+  private final ReactContext mContext;
+  private final UIManagerModule mUIManager;
+
+  public Set<String> shadowViewProps = Collections.emptySet();
+
+  private final class NativeUpdateOperation {
```
janicduplessis: `UIManagerUpdateOperation`

chrfalch:
Thanks so much for the feedback. I’m currently unavailable for a few days :( will set a side time as soon as possible to fix the issues you pointed out.

Libraries/Animated/src/NativeAnimatedHelper.js
chrfalch: Removed.

Libraries/Animated/src/NativeAnimatedHelper.js
chrfalch: Moved to native side. Added deprecation warnings when trying to add extra props to these lists.

Libraries/Animated/src/NativeAnimatedHelper.js
chrfalch: See prev. comment

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
chrfalch: Done!

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
chrfalch: Done!

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
chrfalch: Done

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
chrfalch: Done

ReactAndroid/src/main/java/com/facebook/react/animated/NativeAnimatedNodesManager.java
chrfalch: Done

Libraries/Animated/src/__tests__/AnimatedNative-test.js
chrfalch: Done

chrfalch: Done

chrfalch:
Hi, @JoshuaGross and @janicduplessis - how should we continue with this PR? I see that some of the tests are failing, but I think the test_ios_frameworks test is failing for everyone these days, and the test_android fail seems to be an infrastructure failure?

janicduplessis:
@chrfalch Looks like we got a part wrong. In the original implementation only style props are validated, this means that any regular prop should still go through `synchronouslyUpdateViewOnUIThread`. However we don't have this context anymore in `RCTPropsAnimatedNode`.

I suggest we invert our list and instead of having uiProps which are props that can be updated with `synchronouslyUpdateViewOnUIThread` we go with something like `layoutProps` which are props that need to use the regular UIManager pipeline (`enqueueUpdateViewOnUIManager `). The list of layout props can be found here https://github.com/facebook/react-native/blob/master/React/Views/RCTViewManager.m#L339.

chrfalch:
> @chrfalch Looks like we got a part wrong. In the original implementation only style props are validated, this means that any regular prop should still go through `synchronouslyUpdateViewOnUIThread`. However we don't have this context anymore in `RCTPropsAnimatedNode`.
> 
> I suggest we invert our list and instead of having uiProps which are props that can be updated with `synchronouslyUpdateViewOnUIThread` we go with something like `layoutProps` which are props that need to use the regular UIManager pipeline (`enqueueUpdateViewOnUIManager `). The list of layout props can be found here https://github.com/facebook/react-native/blob/master/React/Views/RCTViewManager.m#L339.
> 
> An example of this is animating the value of a slider https://github.com/facebook/react-native/blob/master/RNTester/js/examples/NativeAnimation/NativeAnimationsExample.js#L661. With the current changes it would go through `enqueueUpdateViewOnUIManager` instead of `synchronouslyUpdateViewOnUIThread`.

Great! Thanks - will see if I can make the required changes. Thanks for the pointers as well.

chrfalch:
> > @chrfalch Looks like we got a part wrong. In the original implementation only style props are validated, this means that any regular prop should still go through `synchronouslyUpdateViewOnUIThread`. However we don't have this context anymore in `RCTPropsAnimatedNode`.
> > I suggest we invert our list and instead of having uiProps which are props that can be updated with `synchronouslyUpdateViewOnUIThread` we go with something like `layoutProps` which are props that need to use the regular UIManager pipeline (`enqueueUpdateViewOnUIManager `). The list of layout props can be found here https://github.com/facebook/react-native/blob/master/React/Views/RCTViewManager.m#L339.
> > An example of this is animating the value of a slider https://github.com/facebook/react-native/blob/master/RNTester/js/examples/NativeAnimation/NativeAnimationsExample.js#L661. With the current changes it would go through `enqueueUpdateViewOnUIManager` instead of `synchronouslyUpdateViewOnUIThread`.
> 
> Great! Thanks - will see if I can make the required changes. Thanks for the pointers as well.

Should be fixed now, @janicduplessis :)

janicduplessis:
Hmm maybe we could just expose the layoutProps array to make it possible to add to it from objc / java by getting the native animated module instance with bridge.moduleForClass. It can also be possible to add a new method on the native animated module like addLayoutProp. We might want to address this is a separate PR though.

janicduplessis:
As for SVG, afaik it already works with the current implementation using synchronouslyUpdateViewOnUIThread.

chrfalch:
> Hmm maybe we could just expose the layoutProps array to make it possible to add to it from objc / java by getting the native animated module instance with bridge.moduleForClass. It can also be possible to add a new method on the native animated module like addLayoutProp. We might want to address this is a separate PR though.

What about adding support for updating it from JavaScript to make it easier for devs to use it?

Libraries/Animated/src/NativeAnimatedHelper.js
```diff
@@ -243,13 +219,11 @@ function validateTransform(
 }
 
 function validateStyles(styles: {[key: string]: ?number, ...}): void {
```
janicduplessis: I think we can delete this completely, I highly doubt there are other external callsites, kind of a private api anyway.

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -61,10 +76,70 @@ - (instancetype)initWithBridge:(nonnull RCTBridge *)bridge
...
+      @"flexWrap",
+      @"justifyContent",
+      @"alignItems",
+      @"alignSelf",
+      @"alignContent",
+      @"position",
+      @"aspectRatio",
+      @"overflow",
+      @"display",
+      @"onLayout",
```
janicduplessis: I don't think we need onLayout here.

Libraries/Animated/src/NativeAnimatedHelper.js
chrfalch: I already added a deprecation warning so that we can remove this one safely at a later time - would that be ok?

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
chrfalch: Done

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -61,10 +76,69 @@ - (instancetype)initWithBridge:(nonnull RCTBridge *)bridge
     _animationNodes = [NSMutableDictionary new];
     _eventDrivers = [NSMutableDictionary new];
     _activeAnimations = [NSMutableSet new];
+    _uiManagerOperationQueue = [NSMutableArray new];
+    _layoutProps = @[
```
osdnk: I'm wondering if I can extend this set. E.g. in my app (cc. @christianbaroni, @jinchung) I wish to animate `blurAmount` from 'react-native-blur-view`. I managed to do it with reanimated, bc we exported the method for handling this.

osdnk: Basically I'd love to have sth like `configureProps` 😄

Libraries/Animated/src/NativeAnimatedHelper.js
janicduplessis: I think we can delete it now since I don't see why it would be called from user code. Is there a case for it I don't see?

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
chrfalch: We're discussing this - all properties should be animated automatically, but a method for adding extra props that needs to go through the `enqueueUpdateViewOnUIManager` should be available. Will try to add this.

osdnk: Thanks!

chrfalch: Added addEnqueuedUpdateProp method :)

janicduplessis: Just to clarify this, we want to use `synchronouslyUpdateViewOnUIThread` when possible since it will result in the best UX / performance and fallback to `enqueueUpdateViewOnUIManager` for props that cannot be updated synchronously from the UI thread because they require to go through the shadow tree (usually because of layout).

Libraries/Animated/src/NativeAnimatedHelper.js
chrfalch: I'm not sure - but it is possible to call this from user land - so we don't know if anyone is using this or not. That's why I suggest we go the deprecation route by giving a warning and then remove in a later release.

chrfalch: Do you want me to remove or can we keep with deprecation warning, @janicduplessis? If we can keep it I think the branch is ready to be merged.

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
```diff
@@ -61,10 +76,69 @@ - (instancetype)initWithBridge:(nonnull RCTBridge *)bridge
     _animationNodes = [NSMutableDictionary new];
     _eventDrivers = [NSMutableDictionary new];
     _activeAnimations = [NSMutableSet new];
+    _uiManagerOperationQueue = [NSMutableArray new];
+    _layoutProps = [NSMutableArray arrayWithArray:@[
```
janicduplessis: Let's make this a `NSSet` so we can remove logic to avoid duplicates in `addEnqueuedUpdateProp`. Might also be slightly more performant (although array and set performance is very similar for lookups with a small amount of items).

Can probably also change the android implementation to use a Set too.

Libraries/Animated/src/NativeAnimatedHelper.js
janicduplessis: Let's remove it

chrfalch: Done!

Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m
chrfalch: Done. Android already implemented this with Set.

chrfalch:
> Left a few more comments, after those are resolved it is good to ship for me, I looked mostly at the iOS implementation. I'll let @JoshuaGross pick it up from here.

Thanks so much for helping out - @JoshuaGross: Ping me on discord or here if you have questions.

JoshuaGross:
@shergin or @sammy-SC do you want to test or review the iOS side of this PR?

ReactAndroid/src/main/java/com/facebook/react/animated/InterpolationAnimatedNode.java
```diff
@@ -219,31 +220,29 @@ public void update() {
...
           if (mShouldRound) {
             // rgba requires that the r,g,b are integers.... so we want to round them, but we *dont*
             // want to
             // round the opacity (4th column).
-            boolean isAlpha = i == 4;
-            int rounded = (int) Math.round(isAlpha ? val * 1000 : val);
-            String num =
-                isAlpha ? Double.toString((double) rounded / 1000) : Integer.toString(rounded);
-            mSOutputMatcher.appendReplacement(sb, num);
+            boolean isAlpha = i == 3;
```
JoshuaGross: Why is `i==4` changing to `i==3`?

chrfalch:
> @shergin or @sammy-SC do you want to test or review the iOS side of this PR?

@janicduplessis reviewed the iOS code - but feel free to review this part of the implementation if you want. :)

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
```diff
@@ -133,32 +133,27 @@ - (void)performUpdate
...
                                                           _extrapolateLeft,
                                                           _extrapolateRight);
-        NSString *str;
         if (_shouldRound) {
           // rgba requires that the r,g,b are integers.... so we want to round them, but we *dont* want to
           // round the opacity (4th column).
           bool isAlpha = i == 3;
-          CGFloat rounded = isAlpha ? round(val * 1000) / 1000 : round(val);
-          str = isAlpha ? [NSString stringWithFormat:@"%1.3f", rounded] : [NSString stringWithFormat:@"%1.0f", rounded];
+          CGFloat rounded = isAlpha ? round(val * 1000) / 1000 : round(val) / 255;
```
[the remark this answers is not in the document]
chrfalch: Why `255.99`? I can't see this value used anywhere else in the source code when converting colors - in `RCTConvert.m` the value `255.0` is used - maybe that would be better?

Libraries/Animated/src/colorToRgba.js
```diff
@@ -0,0 +1,33 @@
+/**
```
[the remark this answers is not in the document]
chrfalch: It is not a public API and just used internally in the Animated Library so I saw no reason to complicate the process by adding a separate dependency on another PR.

chrfalch:
> @chrfalch Is this change implied that we run layout during updating the layout props? If so, which exact change enables that?

I've added calls through an operation queue in `RCTNativeAnimatedNodesManager.m` that ends up calling `enqueueUpdateViewOnUIManager` for properties that need to use the layout pipeline.

Not sure if that was the answer you were looking for?

Libraries/NativeAnimation/Nodes/RCTInterpolationAnimatedNode.m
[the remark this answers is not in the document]
shergin: Oh, my bad, you are right, it should be 255 here.

shergin:
@chrfalch Thanks!
Looks good to me!

chrfalch:
> @chrfalch Thanks!
> Looks good to me!

Thanks so much! Anyone specific I just poke to get it merged? I also discussed briefly with @JoshuaGross about having to run some internal codegen tools to update the API before merging. @shergin

terrysahaidak:
After some testing realized fontSize, letterSpacing and lineHeight are missing in `layoutProps` array. They cannot be updated on UIThread, so animating them makes no effect right now.

chrfalch:
> After some testing realized fontSize, letterSpacing and lineHeight are missing in `layoutProps` array. They cannot be updated on UIThread, so animating them makes no effect right now.

We might also consider adding 'text' as a property to be passed through the layout cycle (It can be used in Animated version of `TextInput`).

JoshuaGross:
Had a chat with @chrfalch offline, will summarize here:

While I see the value in this, this native Animated library may not be the right solution for animations longterm. It (and LayoutAnimations) have a number of issues in their current incarnation; aside from the APIs not being the nicest to use, it's very easy to get an app into a bad state (visual glitches or crashing) using these libraries, in ways that aren't super trivial to solve.

At the same time, as we're preparing for Fabric to be released, any major breaking changes to the core need to be supported in Fabric, and this increases the workload for the core team substantially and does actually push out the ship date of Fabric with every change we need to port to Fabric. With this change in particular, changing layout props requires a Yoga relayout, which in Fabric requires a C++ commit of a new shadow node - simply not possible at all with the current version of native Animated. This isn't a matter of "it would take a few weeks", we would have to completely rewrite Animated for Fabric to support this.

So our options are:
1) Merge this and live with a big inconsistency between this version of RN and Fabric... forever
2) Rewrite LayoutAnimations and Animated for Fabric
3) and/or, create a new animation API for Fabric

Since the current version of native Animated **is** compatible with Fabric, and because I want the migration to Fabric to be as trivial as possible, I'm very hesitant to merge this. At the same time, if we did rewrite Animated to support this in Fabric, it's not clear that we would even want that with the current API; if we're spending that amount of effort it makes sense to build a new, modern animation system with an API we can put more thought into. (I plan to do just that, but can't promise anything wrt timing)

I realize this is probably disappointing for some people, so I figured it would be helpful to see our rationale.

PR #27509, opened 2019-12-13
Move cli tools to devDependencies

safaiyeh:
## Summary
I noticed the cli tools are in `dependencies`, which means the cli tools make it to production builds. I moved it to `devDependencies`  to start a discussion. Is there a reason for these packages to be in the `dependencies`?

## Changelog

[General] [Fixed] - Moved react-native-cli to devDependencies 

## Test Plan
CLI Tools resolve correctly in local development, removed in production.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,9 +86,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@react-native-community/cli": "^3.0.0",
-    "@react-native-community/cli-platform-android": "^3.0.0",
-    "@react-native-community/cli-platform-ios": "^3.0.0",
     "abort-controller": "^3.0.0",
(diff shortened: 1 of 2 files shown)


thymikee:
Um, the CLI is bundled with RN and this effectively breaks official usage. So no, we can't do that.

safaiyeh:
Thanks for the reply @thymikee.

How are React Native and the CLI coupled? I would like to understand how the CLI tools operate with this project. I only understand all the development/build commands.

thymikee:
Anytime you run `react-native <command>` the binary, whether global or local, reach out to local react-native bin, which calls the cli. It's a crucial development flow for users.

thymikee:
As a part of the lean core efforts we extracted the cli out of the core for faster development

Issue #27504, opened 2019-12-13
[Android][Flatlist][Horizontal] Last item delete in horizontal flatlist does not adjust scroll position.


jayshah123:
React Native version: 0.59.10

![output](https://user-images.githubusercontent.com/659202/70800493-6dc22880-1dd2-11ea-8a43-66e87ba867d9.gif)

## Steps To Reproduce

```
export default class App extends Component<Props> {

  constructor(props) {
    super(props);
    this.dataItems = [{
      id: 1,
      text: 'one'
    },{
      id: 2,
      text: 'two'
    },{
      id: 3,
      text: 'three'
    }, {
      id: 4,
      text: 'four',
    }, {
      id: 5,
      text: 'five'
    }];
  }

  removeItem = (id) => {
    this.dataItems = this.dataItems.filter((item) => item.id !== id);
    this.setState({});
  };

  renderItem = ({item}) => {
    return (
      <View style={{ width: 140, height: 140, backgroundColor: 'yellow', margin: 10}}>
        <Text>{item.text}</Text>
        <TouchableOpacity onPress={() => {
          this.removeItem(item.id);
        }}>
          <Text>{'Remove me'}</Text>
        </TouchableOpacity>
      </View>
    )
  };

  render() {
    return (
      <View style={styles.container}>
        <View style={{height: 140, width: '100%'}}>
          <FlatList
            horizontal
            data={this.dataItems}
            renderItem={this.renderItem}
            keyExtractor={item => `${item.id}`}
          />
        </View>
      </View>
    );
  }
}

```
Describe what you expected to happen:
After removing last item, scroll position should be adjusted.

Snack, code example, screenshot, or link to a repository:

jayshah123:
This problem will most likely also be present on the latest version from the code it seems.

jayshah123:
The fix can be simple: add a hierarchy change + layout change listener, which does a `scrollTo` whenever we have scrolled beyond capacity post an item removal. This is already present in ReactScrollView.

Peeeep:
Thank you, kind little bot, but this still needs a fix :/

alexeyvax:
I would suggest to use scrollToEnd(), but anyway you need to know screen width and each item width. In case you have two or only one item in your list, you will need to use scrollToIndex({ index: 0 }) instead

jayshah123:
[...]

PR #27514, opened 2019-12-15
[Android][HorizontalScrollView] last item removal scroll position restore

jayshah123:
## Summary

Fixes https://github.com/facebook/react-native/issues/27504
Which is: Whenever a last list item in a horizontal scroll view in android is removed, the position is not reset and shows an empty/blank space in the end (gif attached in the issue).

## Changelog

Similar to ReactScrollView(https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java#L774): 
We add a listener which triggers whenever a list item is removed/added, and it checks if we have a scroll position that is more than the max possible scrollposition, it will reset it to max possible scroll position.

[Android] [Fixed] - last item removal scroll position restore in ReactHorizontalScrollView

## Test Plan

Code should not be a problem since there is already a similar solution in ReactScrollView.
Attaching gif for the fixed version.
output_fixed


ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
```diff
@@ -537,6 +541,51 @@ protected void onOverScrolled(int scrollX, int scrollY, boolean clampedX, boolea
...
+    mContentView = null;
+  }
+
+  /**
+   * Called when a mContentView's layout has changed. Fixes the scroll position if it's too large
+   * after the content resizes. Without this, the user would see a blank ScrollView when the scroll
+   * position is larger than the ScrollView's max scroll position after the content shrinks.
+   */
+  @Override
+  public void onLayoutChange(
```
safaiyeh: Would setting `fillViewport` achieve the same behavior? If Android already handles it, the manually calculations may not needed. 

https://developer.android.com/reference/android/widget/HorizontalScrollView.html#setFillViewport(boolean)

jayshah123: `fillViewPort` will change how the child views are being measured, typically it might result in different behaviour in a case where content size is less than viewport size.
I will try to come up with a case where fillViewport might break existing behaviours.

Peeeep:
Hey, great you are tackling this issue which has been bugging me for a couple of days now! Unfortunately, I wasn't able to get this to work @0.61.5... Is there any way I can help debugging this?

velmar85:
that can be possible , all deppends how tou want fix it

Sent from my iPhone

On Dec 31, 2019, at 5:48 PM, Alexander Kurnikowski <notifications@github.com<mailto:notifications@github.com>> wrote:


Hey, great you are tackling this issue which has been bugging me for a couple of days now! Unfortunately, I wasn't able to get this to work @0.61.5... Is there any way I can help debugging this?

JoshuaGross:
@jayshah123 it sounds like you're still working on finding a case where setting `fillViewport` alone would not work, right? Is this ready for review yet?

jayshah123:
I will check this weekend. and update.
Also, My current solution is just an imitation of what is already present in the regular ScrollView.

ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
jayshah123: Even with fillViewPort set to true, the issue does not go away.
Screen Shot 2020-02-16 at 12 20 10 PM

hence the original solution

jayshah123:
Even with fillViewPort set to true, the issue does not go away.
Screen Shot 2020-02-16 at 12 20 10 PM

hence the original solution

jayshah123:
having some merging issues. will cleanup and raise

Issue #27510, opened 2019-12-13
TextInput autoCapitalize="words" not working when keyboardType prop is passed on Android


WouterFlorijn:
When adding `autoCapitalize="words"` and `keyboardType` with any value, on Android, nothing is auto-capitalized. There have been other issues regarding this, but they have been closed, locked and since ignored ([#11776](https://github.com/facebook/react-native/issues/11776) [#8932](https://github.com/facebook/react-native/issues/8932)).

In older issues/comments it's stated that this happens only when margin or padding is applied through style on the `TextInput`, but this is not the case. It also happens without.

If this has been fixed in React Native 0.60+, please let me know. I'm only able to test on 0.59.

React Native version:
info
  React Native Environment Info:
    System:
      OS: Windows 10
      CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
      Memory: 4.79 GB / 15.89 GB
    Binaries:
      Yarn: 1.1.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
      npm: 6.10.3 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: Version  3.5.0.0 AI-191.8026.42.35.5791312

## Steps To Reproduce

1. Add `autoCapitalize="words"` to a `TextInput`.
2. Also add `keyboardType=...` with any value. Even `undefined`, even though `undefined` props are supposed to be ignored in React.

Describe what you expected to happen:

The first letter of each word is capitalized.

safaiyeh:
I investigated this issue on master, it works correctly with the physical keyboard; however, not with soft input (inputs from your computer). 

With soft input notice how the capitalize on the keyboard turns on with every space. This is a dev environment issue, not a bug to a shipped app.

![inputTextCapitalization](https://user-images.githubusercontent.com/8675043/70853754-289c0600-1e67-11ea-9912-c28726efc87e.gif)

WouterFlorijn:
@safaiyeh It's not. I've confirmed this issue on Android devices in development and production. It occured on every Android device I tested on. The only workaround was to remove the `keyboardType` prop. If it's only happening with soft inputs for you, there might be other factors (props) that also impact this behavior, but it's definitely not a PC-only bug.

safaiyeh:
@WouterFlorijn I understand now.

[...]

PR #27523, opened 2019-12-16
Fix setting keyboardType from breaking autoCapitalize

safaiyeh:
## Summary
Fix for https://github.com/facebook/react-native/issues/27510.

Setting the `InputType.TYPE_CLASS_TEXT` flag when `keyboardType` is null or default breaks autoCapitalize. Handle the case when `keyboardType` is null, default, or invalid type.

## Changelog

[Android] [Fixed] - Fix setting keyboardType from breaking autoCapitalize

## Test Plan

Added keyboardType prop to RNTester as so
```
<TextInput autoCapitalize="words" keyboardType="default" style={styles.default} />
```
fixedKeyboardType

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
@@ -67,6 +67,7 @@ import com.facebook.yoga.YogaConstants;
 import java.lang.reflect.Field;
 import java.util.LinkedList;
 import java.util.Map;
+import java.util.Arrays;
 
 /** Manages instances of TextInput. */
 @ReactModule(name = ReactTextInputManager.REACT_CLASS)
@@ -104,6 +105,15 @@ public class ReactTextInputManager extends BaseViewManager<ReactEditText, Layout
   private static final String KEYBOARD_TYPE_NUMBER_PAD = "number-pad";
   private static final String KEYBOARD_TYPE_PHONE_PAD = "phone-pad";
   private static final String KEYBOARD_TYPE_VISIBLE_PASSWORD = "visible-password";
+  private static final String[] KEYBOARD_TYPES = new String[]{ 
+    KEYBOARD_TYPE_DEFAULT,
+    KEYBOARD_TYPE_EMAIL_ADDRESS,
+    KEYBOARD_TYPE_NUMERIC,
+    KEYBOARD_TYPE_DECIMAL_PAD,
+    KEYBOARD_TYPE_NUMBER_PAD,
+    KEYBOARD_TYPE_PHONE_PAD,
+    KEYBOARD_TYPE_VISIBLE_PASSWORD
+  };
   private static final InputFilter[] EMPTY_FILTERS = new InputFilter[0];
   private static final int UNSET = -1;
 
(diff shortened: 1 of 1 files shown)


safaiyeh:
Any action items from me for the Facebook Internal - Linter?

TheSavior:
I think this needs someone internal (likely @JoshuaGross) to run the formatter internally and reland.

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -103,6 +106,14 @@
   private static final String KEYBOARD_TYPE_NUMBER_PAD = "number-pad";
   private static final String KEYBOARD_TYPE_PHONE_PAD = "phone-pad";
   private static final String KEYBOARD_TYPE_VISIBLE_PASSWORD = "visible-password";
+  private static final List<String> KEYBOARD_TYPES = new ArrayList<>(Arrays.asList(
```
makovkastar: Can we use a Set instead, so that the check on line 731 is done in a constant time?

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java
```diff
@@ -717,7 +728,9 @@ public void setAutoCapitalize(ReactEditText view, Dynamic autoCapitalize) {
   @ReactProp(name = "keyboardType")
   public void setKeyboardType(ReactEditText view, @Nullable String keyboardType) {
     int flagsToSet = InputType.TYPE_CLASS_TEXT;
-    if (KEYBOARD_TYPE_NUMERIC.equalsIgnoreCase(keyboardType)) {
+    if (keyboardType == null || !KEYBOARD_TYPES.contains(keyboardType)) {
```
makovkastar: Why do we need this condition? Does it mean that `InputType.TYPE_CLASS_TEXT` will never be set? Can we just remove this condition and add a single else at the end?

safaiyeh: The `else` makes more sense. removes the need for the List check.

By default type is `InputType.TYPE_CLASS_TEXT`. Setting the flag again breaks the autoCapitalize flags.

safaiyeh: I cleaned up the logic and added a comment of why it is needed.

@makovkastar

JoshuaGross:
@safaiyeh, please see this commit and verify that autoCapitalize still works for your test case here.

https://github.com/facebook/react-native/commit/26d5faf379c6fc13678f697ec22c1cc6e96a949d

safaiyeh:
I'll check this out tomorrow @JoshuaGross, would adding an E2E test for it be helpful?

safaiyeh:
@JoshuaGross Just tested and confirm it works for https://github.com/facebook/react-native/commit/26d5faf379c6fc13678f697ec22c1cc6e96a949d.

JoshuaGross:
@safaiyeh E2E tests would be great!

vongohren:
@safaiyeh @JoshuaGross when is it worth thinking this fix will make it into the codebase? I dont see it with the 0.62.x RC

safaiyeh:
@vongohren @alloy is managing the release, the RC discussion will determine that.

Also if this is commit: https://github.com/facebook/react-native/commit/26d5faf379c6fc13678f697ec22c1cc6e96a949d also must be picked due to a regression this caused.

vongohren:
@safaiyeh I tried asking in the RC discussion, but I was ignored. So just asking here to find some more insight

alloy:
https://github.com/facebook/react-native/commit/233fdfc014bb4b919c7624c90e5dac614479076f was identified by https://github.com/react-native-community/releases/issues/157#issuecomment-592542297, which was picked.

@safaiyeh Are you saying that https://github.com/facebook/react-native/commit/26d5faf379c6fc13678f697ec22c1cc6e96a949d is essential for your change [that I picked] to for right?

safaiyeh:
@alloy yup! It fixes a regression caused by this PR

alloy:
Sigh, so that means RC4 is broken :( Thanks for the info!

PR #27539, opened 2019-12-16
Add an API for Detox to check if there are any timers expiring in a certain range

ejanzer:
## Summary
Detox currently relies on reflection to inspect the private timers queue in the Timing module to check if React Native is idle or not. This broke when I renamed TimingModule and moved the queue to JavaTimerManager in D17260848 and D17282187. A better solution to this problem is for us to expose a public API for checking the timers queue from TimingModule, so that Detox doesn't need to access private fields.

Using similar logic to Detox's TimersIdlingResource: https://github.com/wix/Detox/blob/4f81a77baee4e4542a693922bcbde621d9d8c1a5/detox/android/detox/src/main/java/com/wix/detox/reactnative/idlingresources/TimersIdlingResource.kt#L95
based on discussion here: https://github.com/wix/detox/issues/907

## Changelog
[Android] [Added] Added an API for checking if there are busy timers to TimingModule

## Test Plan
Added tests to TimingModuleTest.java

Differential Revision: D19128786

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/core/JavaTimerManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/core/JavaTimerManager.java
@@ -373,4 +373,23 @@ public class JavaTimerManager {
           }
         });
   }
+
+  /**
+   * Returns a bool representing whether there are any active timers that will be fired within a
+   * certain period of time. Disregards repeating timers (setInterval). Used for testing to
+   * determine if RN is idle.
+   *
+   * @param rangeMs The time range, in ms, to check
+   * @return True if there are pending timers within the given range; false otherwise
+   */
+  /* package */ boolean hasActiveTimersInRange(long rangeMs) {
+    synchronized (mTimerGuard) {
+      for (Timer timer : mTimers) {
+        if (!timer.mRepeat && timer.mInterval < rangeMs) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
 }
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/core/TimingModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/core/TimingModule.java
@@ -11,6 +11,7 @@ import com.facebook.fbreact.specs.NativeTimingSpec;
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.WritableArray;
+import com.facebook.react.common.annotations.VisibleForTesting;
 import com.facebook.react.devsupport.interfaces.DevSupportManager;
 import com.facebook.react.jstasks.HeadlessJsTaskContext;
 import com.facebook.react.jstasks.HeadlessJsTaskEventListener;
@@ -133,5 +134,10 @@ public final class TimingModule extends NativeTimingSpec
         HeadlessJsTaskContext.getInstance(getReactApplicationContext());
     headlessJsTaskContext.removeTaskEventListener(this);
     mJavaTimerManager.onInstanceDestroy();
+  }
+
+  @VisibleForTesting
+  public boolean hasActiveTimersInRange(long rangeMs) {
+    return mJavaTimerManager.hasActiveTimersInRange(rangeMs);
   }
 }
--- a/ReactAndroid/src/test/java/com/facebook/react/modules/BUCK
+++ b/ReactAndroid/src/test/java/com/facebook/react/modules/BUCK
@@ -28,6 +28,7 @@ rn_robolectric_test(
         react_native_target("java/com/facebook/react/common/network:network"),
         react_native_target("java/com/facebook/react/devsupport:interfaces"),
         react_native_target("java/com/facebook/react/jstasks:jstasks"),
+        react_native_target("java/com/facebook/react/module/annotations:annotations"),
         react_native_target("java/com/facebook/react/modules/blob:blob"),
         react_native_target("java/com/facebook/react/modules/camera:camera"),
         react_native_target("java/com/facebook/react/modules/clipboard:clipboard"),
--- a/ReactAndroid/src/test/java/com/facebook/react/modules/timing/TimingModuleTest.java
+++ b/ReactAndroid/src/test/java/com/facebook/react/modules/timing/TimingModuleTest.java
@@ -7,6 +7,7 @@
 
 package com.facebook.react.modules.timing;
 
+import static org.fest.assertions.api.Assertions.assertThat;
 import static org.mockito.Mockito.*;
 
 import com.facebook.react.bridge.Arguments;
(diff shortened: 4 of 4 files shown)


ejanzer:
Hey @d4vidi, mind taking a look at this PR? I want to make sure this is something that we'd actually want to use for Detox before I land it. Some additional discussion/context with @rotemmiz here: https://github.com/wix/Detox/issues/1801#issuecomment-566289820

LeoNatan:
Hello,
Another thing to consider, if possible, is to create an event-based API, rather than a polling API. 

On iOS, we are rewriting Detox to be fully event-based, and have thus dropped Earl Grey. On iOS, whenever a new timer (that is to be accepted) is started, we increment a counter and once timer expires, we decrement the counter. Once counter is zero, the sync resource is considered free. Otherwise busy. 

On Android, Espresso already has support for event based idle resources, and some resources already are such. In the future, we may consider converting our polling idle resources to event based, and it would be great to have such an API available for use. Perhaps this can be the first resource to convert. An observer patter over timers state would work great here. 

Thanks 🙂

rotemmiz:
In addition to what @LeoNatan said, event based IdlingResources will greatly decrease the CPU utilization of Detox, and will enable more test concurrency using the same hardware. My assumption is that the new implementation will open the possibility to use twice the worker count when running Detox (currently we use 3 workers on 8-vcore VMs, I believe the new implementation will give us the ability to run 5-6 workers with the same hardware).

ejanzer:
@LeoNatan / @rotemmiz : Correct me if I'm wrong, but it sounds like the reason why Detox needs the number of active timers is just to tell whether JS is idle, right? If we're just looking for a proxy for whether JS is idle, or whether the UI is done updating, then I think we could expose that information more directly - we could create an idling resource that would notify you when the JS runtime itself is free, for example. This is something that I'm going to need as I'm trying to write tests for RN running in bridgeless mode, so I'd probably want to do this anyway. If that would meet Detox's needs, then I probably wouldn't want to rewrite timers to use an idling resource right now, since I'd just be replacing it soon anyway.

If you do actually need the timers themselves for some reason, then I think it makes sense to use an event-based resource like you described. I don't really have time to do it right now, but I'm happy to look at any PRs that make the change, if anyone else wants to jump in! (The motivation for this PR was just to unbreak Detox on the current RN RC, because I broke it by refactoring the timers module).

LeoNatan:
We’ve found creative uses for the information. While the Detox iOS rewrite is ongoing, we’re now using the same sync management system in Detox Instruments to provide app activity information to the user: 

https://github.com/wix/DetoxSync
https://github.com/wix/DetoxInstruments/blob/master/Documentation/Instrument_Activity.md

It’s an unintended consequence, but very useful indeed. The sync manager I’ve developed has lit up a lot of useful information, all from the same sync points as Detox uses, including the timers.

d4vidi:
@ejanzer first of all, thanks for this!
Second, in Detox, indeed we _do_ have interest in telling whether there are timers waiting in queue, so as to consider the system 'busy' even though the JS itself is idle (namely, whenever there are non-repeating timers within a 1.5sec look-ahead window - hence the logic in our implementation).
The reasoning behind this is to avoid moving to the next test-step (e.g. tapping on a button), as it can trigger a time-consuming operation that would be interrupted by the expiring pending timeout.

So to answer your question, once we decide to switch to the event-based world, we would have to have an event-based idling-resource interrogator implementation _specific_ to timers. As @LeoNatan explain, it shouldn't be hard to achieve as we have the concept already working on iOS, and also Espresso is ready for this as a framework.

ejanzer:
@d4vidi Ah, I see, yeah that makes sense. So what would you like to do with this PR? Should I abandon it since we want to move to an event-based world soon anyway? Or land it as a temporary measure?

ejanzer:
Nevermind, just caught up with the discussion over in https://github.com/wix/Detox/issues/1801 - sounds like we're good to go with this PR for now. I'll land it.

ReactAndroid/src/main/java/com/facebook/react/modules/core/JavaTimerManager.java
```diff
@@ -373,4 +373,23 @@ public void run() {
...
+   * Returns a bool representing whether there are any active timers that will be fired within a
+   * certain period of time. Disregards repeating timers (setInterval). Used for testing to
+   * determine if RN is idle.
+   *
+   * @param rangeMs The time range, in ms, to check
+   * @return True if there are pending timers within the given range; false otherwise
+   */
+  /* package */ boolean hasActiveTimersInRange(long rangeMs) {
+    synchronized (mTimerGuard) {
+      for (Timer timer : mTimers) {
```
d4vidi: Hey @ejanzer! I hope I'm not too late - seeing this has already been merged, I'd like to request some changes nevertheless.

Here, in order to evaluate whether _active_ timers exist, the timers collection gets iterated over. If you take a closer look into detox' implementation, we _do_ eventually run the same logic, but in addition we [first attempt to inspect the first (i.e. _least_) timer in the collection](https://github.com/wix/Detox/blob/4f81a77baee4e4542a693922bcbde621d9d8c1a5/detox/android/detox/src/main/java/com/wix/detox/reactnative/idlingresources/TimersIdlingResource.kt#L87). That is done for a reason:

As the collection is in fact a `PriorityQueue`, these facts hold true:
1. There's no indexed access to items in it - only access to the _least_ item using `peek()` and `poll()`.
2. (To quote from PriorityQueue.iterator())  _The iterator does not return the elements in any particular order_.

These 2 combined evidently mean that most of the time, the least item is in fact the one we're looking for, as its interval is the lowest. We therefore resolve to this order-less iteration only "if we have to", namely, if the least item happens to be a repeating one.
> Actually, I see now that an additional optimization can be applied: if the least item is a nonrepeating one that has a long interval, the iteration can be avoided as well, as all other timers' intervals are longer by-design.

So, given this tedious explanation..... do you think you could apply these existing optimizations to your implementation? 😄

ejanzer: Ah I see, I should have looked at that a little more carefully! So to reiterate, we're going to skip iterating over the entire timers queue if the least timer (timers.peek()) is nonrepeating and its interval is greater than the specified threshold.

That's the only change to make, right? Or were there any other optimizations that I missed?

d4vidi: Yes! expect for:
> interval is greater than the specified threshold
interval should be lower (i.e. within the time-window's bounds)

In any case, I'm gonna go ahead and merge + release the Detox changes so you could start working with an actual Detox that does what you need!

d4vidi: Yes! except for:
> interval is greater than the specified threshold

interval should be lower (i.e. within the time-window's bounds)

In any case, I'm gonna go ahead and merge + release the Detox changes so you could start working with an actual Detox that does what you need!

d4vidi: Yes! except for:

> interval is greater than the specified threshold

interval should be lower (i.e. within the time-window's bounds)

In any case, I'm gonna go ahead and merge + release the Detox changes so you could start working with an actual Detox that does what you need!

ejanzer: Thanks @d4vidi! I'll send a PR today with this change.

ejanzer: Adding in #27841

PR #27558, opened 2019-12-18
Add AnimatedInterpolation as possible type for toValue

nabati:
## Summary

AnimatedInterpolation should be a valid type to pass to `toValue` according to the documentation. 

https://facebook.github.io/react-native/docs/animations#tracking-dynamic-values

## Changelog

[General] [Fixed] - Add AnimationInterpolation as possible type for toValue

## Test Plan

`yarn flow` works without errors.

### DIFF
--- a/Libraries/Animated/src/animations/SpringAnimation.js
+++ b/Libraries/Animated/src/animations/SpringAnimation.js
@@ -12,6 +12,7 @@
 
 const AnimatedValue = require('../nodes/AnimatedValue');
 const AnimatedValueXY = require('../nodes/AnimatedValueXY');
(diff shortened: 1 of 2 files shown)


nabati:
Any idea why the tests fail? At first glance it doesn't look related to the changes, as they are happening with the native build and I've only changed the flow-types. @GummyGod

GummyGod:
To be honest I am not sure, but, I think it might be a problem with the Polyfills. Someone from the core team might be helpful here.

chrfalch:
If you look at all the recent pull requests they all fail the test_ios_framworks test :( So probably not something you've done.

PR #27563, opened 2019-12-18
remove `xcshareddata` from gitignore

pvinis:
## Summary

Files that are inside `xcshareddata` are meant to be "shared", and that means committed to git. These files are for example `ios/RnDiffApp.xcodeproj/xcshareddata/xcschemes/RnDiffApp.xcscheme`, and if ignored that means they are removed from the repo or are missing when cloning. These should be present. They are the shared schemes so Xcode knows what and how to build, on everyone's machine.

I noticed it being there in a new project on 0.62.0-rc.0.

## Changelog

[iOS] [Changed] - Remove the xcshareddata from .gitignore

### DIFF
--- a/template/_gitignore
+++ b/template/_gitignore
@@ -20,7 +20,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
-xcshareddata
 
 # Android/IntelliJ
 #


TheSavior:
Looks like this was added in https://github.com/facebook/react-native/pull/25451

pvinis:
Ah, good spotting. I remember writing about this, but I couldn't find it. Anyway, like I said there, `xcshareddata` should be committed, not ignored.

Issue #21126, opened 2018-09-14
Inverted scrollsToTop scrolls to bottom


raphaelrk:
## Environment
```
Environment:
  OS: macOS High Sierra 10.13.5
  Node: 10.4.1
  Yarn: 1.7.0
  npm: 6.4.1
  Watchman: 4.9.0
  Xcode: Xcode 9.4.1 Build version 9F2000
  Android Studio: Not Found

Packages: (wanted => installed)
  react: ^16.4.2 => 16.4.2
  react-native: https://github.com/expo/react-native/archive/sdk-29.0.0.tar.gz => 0.55.4
```

## Description
When a FlatList has both `scrollsToTop` and `inverted` enabled, tapping on the iOS status bar scrolls to the bottom of the list, which isn't a behavior I've seen in any native apps. It would be great if this was either changed or it was possible to do e.g. `scrollsToBottom={true}` or `scrollsToTopDirection={'inverted'}`.

## Reproducible Demo
[Expo Snack](https://snack.expo.io/@raphaelrk/inverted-scrollstotop)

raphaelrk:
This has been an annoying UX issue since I started using React Native ~10 months ago, would be a really great fix for being able to quickly scroll to the top of a chat screen or other `inverted` screens!

kelset:
It seems to me that you are using Expo, so I suggest you open a dedicated issue there 🤷‍♂️

Our can you reproduce on a new `react-native init` project with `0.57.2`?

raphaelrk:
Tested with the same code from the snack on a new `react-native init` (`0.57.4`) and still have the issue that scrollsToTop scrolls to the bottom

MartinMasek:
Have you found any workaround while it's getting fixed?

emilioicai:
This PR should fix it: https://github.com/facebook/react-native/pull/27574

PR #27574, opened 2019-12-19
tapping on iOS status bar should scroll inverted ScrollViews to their top

emilioicai:
## Summary

React Native ScrollViews are flipped upside down when the prop inverted is set to true. This is the root of a bug: tapping on the status bar in iOS should scroll the Flatlist up to the top but currently it does to the bottom.

The solution proposed is to detect natively if the ScrollView is inverted, on such case, prevent it from scrolling it to the beginning of the ScrollView (as a non-inverted ScrollView would do) and force a scroll to the end of it.

I've been careful enough not to force the scroll if the user explicitly selected not to do it or if it's happening in a nested ScrollView, as it is the default behaviour in iOS.

Fixes https://github.com/facebook/react-native/issues/21126

## Changelog

[iOS] [Fixed] - Inverted ScrollViews scroll to their bottom when the status bar is pressed

## Test Plan

- on iOS, add a ScrollView and put enough content to overflow the screen size so it can be scrolled
- add the prop `inverted={true}` to the ScrollView
- go to the screen the Scrollview is in and press the status bar
- it should scroll to top (previously it scrolled to the bottom)

v

### DIFF
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -829,6 +829,10 @@ RCT_SCROLL_EVENT_HANDLER(scrollViewDidScrollToTop, onScrollToTop)
       return NO;
     }
   }
+  if(self.inverted) {
+    [self scrollToEnd:YES];
+    return NO;
+  }
   return YES;
 }
 
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -829,10 +829,12 @@ RCT_SCROLL_EVENT_HANDLER(scrollViewDidScrollToTop, onScrollToTop)
       return NO;
     }
   }
-  if(self.inverted) {
+
+  if (self.inverted) {
     [self scrollToEnd:YES];
     return NO;
   }
+
   return YES;
 }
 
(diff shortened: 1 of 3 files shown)


React/Views/ScrollView/RCTScrollView.m
```diff
@@ -829,6 +829,10 @@ - (BOOL)scrollViewShouldScrollToTop:(UIScrollView *)scrollView
       return NO;
     }
   }
+  if(self.inverted) {
```
hramos:
```suggestion

  if (self.inverted) {
```

React/Views/ScrollView/RCTScrollView.m
```diff
@@ -829,6 +829,10 @@ - (BOOL)scrollViewShouldScrollToTop:(UIScrollView *)scrollView
...
+    [self scrollToEnd:YES];
+    return NO;
+  }
```
hramos:
```suggestion
  }

```

Hirbod:
@emilioicai sorry for commenting an "old" ticket, but Expo SDK just released with RN 0.62.2 and that fix is included there so its the first time I can test it. Your change works fine so far, the direction works, but only when I scroll "up" at least 1px. If my chat scroll position is "0" a tap on the statusbar won't do anything for me now. If I just scroll up a bit, scroll to top (or here: to bottom since its inverted) works.

Hirbod:
Here is a video (listen to audio please)
https://streamable.com/5pkkko

PAJoeInc:
@Hirbod were you ever able to find a solution to this issue? I am also experiencing the same thing in RN 0.63.2

Hirbod:
@PAJoeInc no - unfortunately not, since nobody responded to my message here.

Hirbod:
I think I will have to write a new ticket

jake-stewart:
very annoying. i want the old behaviour as how discord handles it.

PR #27598, opened 2019-12-24
Fix hermes cannot pick up on custom variants

booker0108:
## Summary

To fix issue of Hermes not being pick up correctly on custom variants

## Changelog

[Android] [Fixed] - Pick up Hermes correctly with custom build variants

## Test Plan

- Add a custom release type, e.g: stagingRelease
- Generate APK with stagingRelease
- `releaseImplementation` will not be successfully pickup hermes

### DIFF
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -167,8 +167,12 @@ dependencies {
     implementation project(':ReactAndroid')
 
     def hermesPath = '$projectDir/../../../../node_modules/hermes-engine/android/'
-    debugImplementation files(hermesPath + "hermes-debug.aar")
-    releaseImplementation files(hermesPath + "hermes-release.aar")
+
+    if(gradle.startParameter.taskNames.contains("Release")) {
+        implementation files(hermesPath + "hermes-release.aar")
+    } else {
+        implementation files(hermesPath + "hermes-debug.aar")
+    }
 
     debugImplementation("com.facebook.flipper:flipper:0.23.4") {
         exclude group:'com.facebook.yoga'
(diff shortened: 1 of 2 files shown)


willholen:
@booker0108 Hi, I'm the one who originally set this up at some point. Sorry if I'm misunderstanding you here.

Are you saying that alongside "Debug" and "Release", you may have a "StagingRelease"? And you want this treated the same way as "Release" based on the name?

Given that the template project doesn't have a "StagingRelease", wouldn't it make more sense to make this change in the project where you introduce such a variant?

booker0108:
> @booker0108 Hi, I'm the one who originally set this up at some point. Sorry if I'm misunderstanding you here.
> 
> Are you saying that alongside "Debug" and "Release", you may have a "StagingRelease"? And you want this treated the same way as "Release" based on the name?
> 
> Given that the template project doesn't have a "StagingRelease", wouldn't it make more sense to make this change in the project where you introduce such a variant?

Hi @willholen ,

`StagingRelease` should and have to be treated same as `Release` for enabling hermes

To differentiate whether a build is `Release`, the variant name will be used
In `react.gradle`, whether the build `isRelease` is determined by checking the target name (variant)
https://github.com/facebook/react-native/blob/master/react.gradle#L302 

So, if a variant name contains `release`, it will be treated as a release app

`releaseImplement` will only be picked by the build variant named exactly `Release` but not other variant targeting a "release" app

If I have a custom release variant and hermes enabled, both `debugImplement` and `releaseImplement` will not be picked up and thus causes hermes engine not linked in the app
When the app launches, it will throw `Fatal Exception: java.lang.UnsatisfiedLinkError: couldn't find DSO to load: libhermes.so`

PR #27612, opened 2019-12-25
Fix race condition in timers in Android

ex3ndr:
## Summary

Fixes race condition when Headless Task and React Host can cause clearing timers callbacks and stop all "setTimeout" timers in the app even when app is visible.

## Changelog

Android Fixed - Race condition when using HeadlessJS that cause death of timers


ex3ndr:
Unfortunatelly it is not fixing my problem

PR #27618, opened 2019-12-26
Implement onRequestClose for iOS 13+ modals

### LINKED ISSUE #26892: [0.61][iOS 13] pageSheet/formSheet dismissal from swipe not propagated
When the user dismisses the modal by a swipe gesture using a custom presentation style,
the event isn't caught by onDismiss.

React Native version: 0.61.0

Sample code :
```js
import {
  StyleSheet,
  Text,
  Button,
  View,
  Modal as RNModal,
} from 'react-native';

export default class Example extends Component {
  state = {
    visible: false,
  };
  
  render() {
    return (
        <Button
          onPress={() => this.setState({ visible: true })}
          title="Default"
        />
        <RNModal
          visible={this.state.visible}
          onDismiss={() => console.log("on dismiss")}
          onRequestClose={() => console.log("on dismiss")}
          presentationStyle={"pageSheet"}>
          <View style={styles.content}>
            <Text style={styles.contentTitle}>Open</Text>
            <Button
              onPress={() => this.setState({ visible: false })}
              title="Close"
            />
          </View>
        </RNModal>
      </View>
    );
  }
}

  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: 'white',
  },
  content: {
    backgroundColor: 'white',
    padding: 22,
    justifyContent: 'center',
    alignItems: 'center',
    borderRadius: 4,
    borderColor: 'rgba(0, 0, 0, 0.1)',
  },
  contentTitle: {
    fontSize: 20,
    marginBottom: 12,
  },
});
```

I'm no expert in iOS, but [this article](https://medium.com/@hacknicity/view-controller-presentation-changes-in-ios-13-ac8c901ebc4e) might give a hint. 

I can fill in a PR with some help.

koke:
## Summary

Starting on iOS 13, a View Controller presented modally will have a "bottom sheet" style unless it's explicitly presented full screen.

Before this, modals on iOS were only being dismissed programatically by setting `visible={false}`. However, now that the dismissal can happen on the OS side, we need a callback to be able to update the state.

This PR reuses the `onRequestClose` prop already available for tvOS and Android, and makes it work on iOS for this use case.

Should fix #26892

## Changelog

[iOS] [Added] - Add support for onRequestClose prop to Modal on iOS 13+

## Test Plan

I tested this using the RNTester app with the Modal example:

1. Select any presentation style other than the full screen ones
2. Tap Present and the modal is presented
3. Swipe down on the presented modal until dismissed
4. Tap Present again and a second modal should be presented

Screen Recording 2019-12-26 at 14 05 33

### DIFF
--- a/React/Fabric/Mounting/ComponentViews/Modal/RCTModalHostViewComponentView.mm
+++ b/React/Fabric/Mounting/ComponentViews/Modal/RCTModalHostViewComponentView.mm
@@ -111,6 +111,7 @@ static ModalHostViewOnOrientationChangeStruct onOrientationChangeStruct(CGRect r
     _viewController = [RCTFabricModalHostViewController new];
     _viewController.modalTransitionStyle = UIModalTransitionStyleCoverVertical;
     _viewController.delegate = self;
+    _viewController.presentationController.delegate = self;
   }
 
   return self;
--- a/React/Views/RCTModalHostView.m
+++ b/React/Views/RCTModalHostView.m
@@ -20,6 +20,10 @@
 #import "RCTTVRemoteHandler.h"
 #endif
 
+@interface RCTModalHostView () <UIAdaptivePresentationControllerDelegate>
+
+@end
+
 @implementation RCTModalHostView
 {
   __weak RCTBridge *_bridge;
@@ -46,6 +50,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:coder)
     UIView *containerView = [UIView new];
     containerView.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
     _modalViewController.view = containerView;
+    _modalViewController.presentationController.delegate = self;
     _touchHandler = [[RCTTouchHandler alloc] initWithBridge:bridge];
 #if TARGET_OS_TV
     _menuButtonGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(menuButtonPressed:)];
@@ -70,10 +75,12 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:coder)
         _onRequestClose(nil);
     }
 }
+#endif
 
 - (void)setOnRequestClose:(RCTDirectEventBlock)onRequestClose
 {
   _onRequestClose = onRequestClose;
+  #if TARGET_OS_TV
   if (_reactSubview) {
     if (_onRequestClose && _menuButtonGestureRecognizer) {
       [_reactSubview addGestureRecognizer:_menuButtonGestureRecognizer];
@@ -81,8 +88,8 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:coder)
       [_reactSubview removeGestureRecognizer:_menuButtonGestureRecognizer];
     }
   }
+  #endif
 }
-#endif
 
 - (void)notifyForBoundsChange:(CGRect)newBounds
 {
@@ -256,5 +263,12 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:coder)
   return supportedOrientations;
 }
 #endif
+
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController
+{
+  if (_onRequestClose) {
+    _onRequestClose(nil);
+  }
+}
 
 @end
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -95,7 +95,8 @@ export type Props = $ReadOnly<{|
 
   /**
    * The `onRequestClose` callback is called when the user taps the hardware
-   * back button on Android or the menu button on Apple TV.
+   * back button on Android, the menu button on Apple TV, or a modal is dismissed
+   * with a gesture on iOS 13+.
    *
    * This is required on Apple TV and Android.
    *
--- a/Libraries/Modal/RCTModalHostViewNativeComponent.js
+++ b/Libraries/Modal/RCTModalHostViewNativeComponent.js
@@ -70,7 +70,8 @@ type NativeProps = $ReadOnly<{|
 
   /**
    * The `onRequestClose` callback is called when the user taps the hardware
-   * back button on Android or the menu button on Apple TV.
+   * back button on Android, the menu button on Apple TV, or a modal is dismissed
+   * with a gesture on iOS 13+.
    *
    * This is required on Apple TV and Android.
    *
--- a/React/Views/RCTModalHostView.h
+++ b/React/Views/RCTModalHostView.h
@@ -32,8 +32,9 @@
 @property (nonatomic, copy) NSArray<NSString *> *supportedOrientations;
 @property (nonatomic, copy) RCTDirectEventBlock onOrientationChange;
 
-#if TARGET_OS_TV
 @property (nonatomic, copy) RCTDirectEventBlock onRequestClose;
+
+#if TARGET_OS_TV
 @property (nonatomic, strong) RCTTVRemoteHandler *tvRemoteHandler;
 #endif
 
--- a/React/Views/RCTModalHostViewManager.m
+++ b/React/Views/RCTModalHostViewManager.m
@@ -108,9 +108,6 @@ RCT_EXPORT_VIEW_PROPERTY(onShow, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(identifier, NSNumber)
 RCT_EXPORT_VIEW_PROPERTY(supportedOrientations, NSArray)
 RCT_EXPORT_VIEW_PROPERTY(onOrientationChange, RCTDirectEventBlock)
-
-#if TARGET_OS_TV
 RCT_EXPORT_VIEW_PROPERTY(onRequestClose, RCTDirectEventBlock)
-#endif
 
 @end


janicduplessis:
Cc @hramos

cpojer:
Getting this failure at FB:

```
xplat/js/react-native-github/React/Fabric/Mounting/ComponentViews/Modal/RCTModalHostViewComponentView.mm:114:55: error: assigning to 'id<UIAdaptivePresentationControllerDelegate> _Nullable' from incompatible type 'RCTModalHostViewComponentView *__strong'
    _viewController.presentationController.delegate = self;
                                                      ^~~~
```

koke:
Oh, I forgot to mention, https://github.com/facebook/react-native/blob/master/React/Fabric/Mounting/ComponentViews/Modal/RCTModalHostViewComponentView.mm probably needs updating as well, but I tried to enable fabric in the RNTester Podfile and was stuck with errors:

```
koke:RNTester/ (fix/26892-ios13-modal-dismiss*) $ git --no-pager diff  -- Podfile
diff --git a/RNTester/Podfile b/RNTester/Podfile
index d7fa6188ad..8e364a5879 100644
--- a/RNTester/Podfile
+++ b/RNTester/Podfile
@@ -11,7 +11,7 @@ def pods()
   project 'RNTesterPods.xcodeproj'
 
   # Enable TurboModule
-  use_react_native!(path: "..")
+  use_react_native!(path: "..", fabric_enabled: true)
   pod 'ReactCommon/turbomodule/samples', :path => '../ReactCommon'
 
   # Additional Pods which aren't included in the default Podfile
koke:RNTester/ (fix/26892-ios13-modal-dismiss*) $ pod install                    
📦 Using pod from bundler
📦 Installing dependencies with bundler...
Using CFPropertyList 3.0.2
Using concurrent-ruby 1.1.5
Using i18n 0.9.5
Using minitest 5.13.0
Using thread_safe 0.3.6
Using tzinfo 1.2.6
Using activesupport 4.2.11.1
Using httpclient 2.8.3
Using json 2.3.0
Using algoliasearch 1.27.1
Using atomos 0.1.3
Using bundler 2.1.2
Using claide 1.0.3
Using fuzzy_match 2.0.4
Using nap 1.1.0
Using cocoapods-core 1.8.4
Using cocoapods-deintegrate 1.0.4
Using cocoapods-downloader 1.3.0
Using cocoapods-plugins 1.0.0
Using cocoapods-search 1.0.0
Using cocoapods-stats 1.1.0
Using netrc 0.11.0
Using cocoapods-trunk 1.4.1
Using cocoapods-try 1.1.0
Using colored2 3.1.2
Using escape 0.0.4
Using fourflusher 2.3.1
Using gh_inspector 1.1.3
Using molinillo 0.6.6
Using ruby-macho 1.4.0
Using nanaimo 0.2.6
Using xcodeproj 1.14.0
Using cocoapods 1.8.4
Bundle complete! 1 Gemfile dependency, 33 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
Analyzing dependencies
Fetching podspec for `Folly` from `../third-party-podspecs/Folly.podspec`
Downloading dependencies
Installing Folly 2018.10.22.00
Installing React-Fabric (1000.0.0)
[!] /bin/bash -c 
set -e
#!/bin/bash
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

# This script collects the "core" component schemas used by fabric
# then uses react-native-codegen to generate the component headers
# to a location that the podspecs expect.

# shellcheck disable=SC2038

find "$PWD/../Libraries" -name "*NativeComponent.js" -print | xargs yarn flow-node packages/react-native-codegen/src/cli/combine/combine-js-to-schema-cli.js schema-rncore.json
yarn flow-node packages/react-native-codegen/buck_tests/generate-tests.js schema-rncore.json rncore ReactCommon/fabric/components/rncore rncore

yarn run v1.21.1
$ /Users/koke/src/react-native/node_modules/.bin/flow-node packages/react-native-codegen/src/cli/combine/combine-js-to-schema-cli.js schema-rncore.json /Users/koke/src/react-native/ReactCommon/../Libraries/Image/ImageViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Image/TextInlineImageNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Utilities/codegenNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/SafeAreaView/RCTSafeAreaViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ScrollView/AndroidHorizontalScrollViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ScrollView/AndroidHorizontalScrollContentViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ScrollView/ScrollViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ScrollView/ScrollContentViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Picker/AndroidDialogPickerNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Picker/AndroidDropdownPickerNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Picker/RCTPickerNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ProgressViewIOS/RCTProgressViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ProgressBarAndroid/ProgressBarAndroidNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/RefreshControl/AndroidSwipeRefreshLayoutNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/RefreshControl/PullToRefreshViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/TextInput/AndroidTextInputNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/TextInput/RCTInputAccessoryViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/DatePicker/RCTDatePickerNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/CheckBox/AndroidCheckBoxNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Slider/SliderNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/DrawerAndroid/AndroidDrawerLayoutNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/SegmentedControlIOS/RCTSegmentedControlNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/MaskedView/RCTMaskedViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/ActivityIndicator/ActivityIndicatorViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/UnimplementedViews/UnimplementedNativeViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/View/ViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Switch/SwitchNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Components/Switch/AndroidSwitchNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/Modal/RCTModalHostViewNativeComponent.js /Users/koke/src/react-native/ReactCommon/../Libraries/ReactNative/requireNativeComponent.js
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
/Users/koke/src/react-native/node_modules/babylon/lib/index.js:4457
  throw err;
  ^

SyntaxError: Unexpected token (215:5)
    at Parser.pp$5.raise (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:4454:13)
    at Parser.pp.unexpected (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:1761:8)
    at Parser.pp$8.flowParsePrimaryType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5772:8)
    at Parser.pp$8.flowParsePostfixType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5778:19)
    at Parser.pp$8.flowParsePrefixType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5795:17)
    at Parser.pp$8.flowParseAnonFunctionWithoutParens (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5800:20)
    at Parser.pp$8.flowParseIntersectionType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5815:19)
    at Parser.pp$8.flowParseUnionType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5826:19)
    at Parser.pp$8.flowParseType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5837:19)
    at Parser.pp$8.flowParseObjectType (/Users/koke/src/react-native/node_modules/babylon/lib/index.js:5476:30)
error Command failed with exit code 1.
```

janicduplessis:
Yea I guess we can remove the change from the fabric implementation and file a task to fix it later when it can compile properly in OSS. @cpojer what do you think?

mattvagni:
Is there anything I can do to help land this (I doubt it since it sounds like internal FB failures but thought I'd ask) ?

jpamarohorta:
It fixes the dismiss not being propagates issue in pageSheet modals but it seems to make all other modals be pageSheet regardless or the presentationStyle prop used. Anyone can reproduce the same issue with this PR?

scarlac:
@jpamarohorta I believe it's not the PR itself that does it but the fact that iOS detects that the new API is being used, thus it auto-upgrades the entire app to the new iOS 13 default. I think it's by design.

HarishMurru:
How can I get this fix into my project.? please help me, someone.

AdrianMrn:
@HarishMurru You can use the steps @scarlac detailed in this comment: https://github.com/facebook/react-native/issues/26892#issuecomment-581198937

Otherwise, you'll probably have to wait until the next release.

jpamarohorta:
@scarlac so with this PR will we be able to use a sheet modal and a fullscreen one in the same app?

shergin:
@koke Seems, it's breaks `presentationStyle` prop, so I have to unland this. Could you please investigate that?

scarlac:
I looked into it. The root cause is that we are changing the delegate which implies that some methods are called. If they are unimplemented, default presentation style is used instead.
Place this delegate method on the delegate object (the one assigned with `_viewController.presentationController.delegate = self`):
```objective-c
// Method must be implemented, otherwise iOS defaults to 'automatic' (pageSheet on >= iOS 13.0)
- (UIModalPresentationStyle)adaptivePresentationStyleForPresentationController:(UIPresentationController *)controller traitCollection:(UITraitCollection *)traitCollection
{
  return self.presentationStyle;
}
```
You won't normally find this via StackOverflow answers because most examples don't change the delegate.

I will also modify my gist to include this change.

React/Views/RCTModalHostView.m
```diff
@@ -257,4 +264,11 @@ - (UIInterfaceOrientationMask)supportedOrientationsMask
 }
 #endif
```
scarlac: We'll need to add this to support changing the presentation style after setting the delegate
```objective-c
// Method must be implemented, otherwise iOS defaults to 'automatic' (pageSheet on >= iOS 13.0)
- (UIModalPresentationStyle)adaptivePresentationStyleForPresentationController:(UIPresentationController *)controller traitCollection:(UITraitCollection *)traitCollection
{
  return self.presentationStyle;
}
```

React/Fabric/Mounting/ComponentViews/Modal/RCTModalHostViewComponentView.mm
```diff
@@ -111,6 +111,7 @@ - (instancetype)initWithFrame:(CGRect)frame
     _viewController = [RCTFabricModalHostViewController new];
     _viewController.modalTransitionStyle = UIModalTransitionStyleCoverVertical;
     _viewController.delegate = self;
+    _viewController.presentationController.delegate = self;
```
scarlac: See below. We may need to add similar code in this file as well.

scarlac:
I know I'm not the author but I'm trying to solve this as well. This PR will likely suffer from the same issue as my Gist: ViewControllers underneath the modal will be unmounted after the modal finishes its transition. The net effect is that transparent modals are transparent while animating but once the animation finishes the background turns fully black (despite using `overlayFullScreen`). This effect is only seen on iOS 12.X devices and is also caused by setting the delegate. I solved it by adding an iOS 13 wrapper:
```objective-c
      _modalViewController.presentationController.delegate = self;
    }
```
This fixes the transparency issue to some extent but transparent iOS 13 modals will cast a tiny shadow which will most likely result in new issues being reported to the repo. The reason for that is that iOS 13 now renders an extra shadow view behind the new modal that are out of our control. In most cases users won't notice but a keen eye will notice a very faint shadow while they drag down on the page sheet modals. Issue is described here:
https://stackoverflow.com/a/59524857/2135705

> [...] we have inserted some private views in the view hierarchy in between UIWindow and its rootViewController's view

image

The overall conclusion is that the transition to iOS 13-style modals is troublesome and will likely cause reports like "transparency doesn't work on iOS 13" with screenshots of vague shadows (where you see "Private UIKit View(s)") despite using `backgroundColor: 'transparent'`. This is an iOS 13 novelty that requires custom modals to get rid of.

jacobp100:
@scarlac Are the issues you mentioned only for the page sheet modals, or does the change also cause issues in other modal styles too?

scarlac:
@jacobp100 Only for pageSheet modals. The shadow that appears is an iOS bug for page and form sheets aka “cards”. 

Regular old full screen modals are fine AFAIK.

noambonnie:
The issue related to this PR (https://github.com/facebook/react-native/issues/26892) is still not resolved, see discussion on the closed issue. Is there a way to reopen the issue?

Foveluy:
@noambonnie the reason why doesn't solve the problem is that RN - team doesn't release a version that contain this merge.

I have checked the code v0.62.0 - v0.62.1 , it doesn't contain this change at all.

so i have to change the native code by myself and it worked perfectly

mrousavy:
@Foveluy what changes did you make, to "enable" this functionality for 0.62?

alloy:
> @noambonnie the reason why doesn't solve the problem is that RN - team doesn't release a version that contain this merge.
> 
> I have checked the code v0.62.0 - v0.62.1 , it doesn't contain this change at all.
> 
> so i have to change the native code by myself and it worked perfectly

This change went into v0.62.0, see [the CHANGELOG](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#ios-specific-5). The reason you couldn’t find it is perhaps because it was cherry-picked into the v0.62-stable branch and thus its commit hash changed: https://github.com/facebook/react-native/commit/6d00239e4970354ad4dc39e1c0f9099da6ea11e8

alloy:
> The issue related to this PR (#26892) is still not resolved, see discussion on the closed issue. Is there a way to reopen the issue?

Done.

mrousavy:
@koke @alloy since `react-native` 0.63 was released the fix is broken. Dragging down doesn't work at all, so `onRequestClose` doesn't get called anymore either..

I'm using patch-package to fix this, so I've updated my patch: https://gist.github.com/mrousavy/b4a4ce2df5ff61d12f46829edb132c7c

Did I miss something? 
Screen Recording 2020-07-09 at 13 40 49 mov

koke:
Sorry, I'm not working with React Native any longer

alekseyanikin:
> @koke @alloy since `react-native` 0.63 was released the fix is broken. Dragging down doesn't work at all, so `onRequestClose` doesn't get called anymore either..
> 
> I'm using patch-package to fix this, so I've updated my patch: https://gist.github.com/mrousavy/b4a4ce2df5ff61d12f46829edb132c7c
> 
> Did I miss something?
> Screen Recording 2020-07-09 at 13 40 49 mov

https://github.com/facebook/react-native/commit/1e7ed81d16dda4188352e0ccdc0f0bd3ad4741f3

you can undo this change if you want this behavior

PR #27290, opened 2019-11-21
bump gradle to 6.0.1

dulmandakh:
## Summary

Bump Gradle to 6.0.1, which has Java 13 support and many improvements. But also it'll help us to clean up deprecation warnings.

## Changelog

[Android] [Changed] - bump Gradle to 6.0.1

## Test Plan

RNTester app builds and runs as expected

PR #27639, opened 2019-12-30
Applied missing changes from bumping Gradle wrapper to 6.0.1

SaeedZhiany:
## Summary

This PR is related to #27290.
I just upgraded my project's Gradle wrapper version to 6.0.1 and I realized some files have some differences with the files in react-native `template` folder. so I create this PR to apply differences.
the main difference is in the `gradlew` file. I'm not familiar with Linux shell scripts but it seems there was a syntax error in `case` items syntax. `(` should not be used in declaring case's items. it may has building error in Linux OS.

## Changelog

[Android] [Fixed] - Applied missing changes from bumping Gradle wrapper to 6.0.1

## Test Plan

I have no Linux OS right now, so I can't directly test these changes, but because the changes have made by running `gradlew wrapper` command, it should not break CI. (I hope :) )


SaeedZhiany:
In `ci/circleci: test_android`'s log, it seems it can build Android RNTester App successfully and it failed in **Collect Test Results** phase because some directory has not been found.

friederbluemle:
> In `ci/circleci: test_android`'s log, it seems it can build Android RNTester App successfully and it failed in **Collect Test Results** phase because some directory has not been found.

Please check again after this PR has been merged. The error is most likely caused by something else.

friederbluemle:
> LGTM. I use gradle wrapper to upgrade versions, like ./gradlew wrapper --gradle-version=6.0.1 --distribution-type=all, and it's strange that it didn't produce this change. Or I made a mistake. Sorry.

No worries :)

Tip for the future: When updating between major versions (or when updates to the wrapper scripts are expected), run the command **twice**: The first time `./gradlew` it _still the old version_ (unaware of the new scripts).

PR #27651, opened 2020-01-02
Build release android

yanniski:
## Summary

There is an issue when you try to build apk image for Android. (https://github.com/facebook/react-native/issues/22234). 
## Changelog

[Android] [Fixed] - Fixed build release for Android issue #22234

## Test Plan
 On any project do "cd android && ./gradlew bundleRelease"

The build release without issue


cpojer:
I think you did something wrong here when making a diff, it's listing 160 commits. Could you try again with only your changes?

PR #27660, opened 2020-01-03
Fix build release Android

yanniski:
issues #22234

## Summary

There is an issue when you try to build apk image for Android. (#22234).
## Changelog

[Android] [Fixed] - Fixed build release for Android issue #22234

## Test Plan

On any project do "cd android && ./gradlew bundleRelease"

The build release without issue


cpojer:
We have had multiple PRs like this, and it isn't the right fix at this time.

cc @dulmandakh who's been involved in this discussion – do you know how we can fix this once and for all?

PR #27685, opened 2020-01-06
Updated Modal.js

ashertoufeeq:
## Summary
Edited Modal.js to make its height adjustable.
## Changelog
[CATEGORY]
-JavaScript
 [TYPE] 
Changed, for changes in existing functionality.

MESSAGE
Added PositionFromTop prop.
PositionFromTop takes a number and sets its position from top accordingly.
In order to change its height

## Test Plan
Yes, Tested it on android.
WhatsApp Image 2020-01-06 at 4 39 05 PM


cpojer:
Thank you for your first PR! Unfortunately we normally don't use a functional style like this to add props. Could you instead of adjusting the positioning of the modal view just position another View inside of your Modal that is positioned like you want to?

(Happy to reopen if there is a reason why that wouldn't work.)

PR #27701, opened 2020-01-07
[iOS] Custom NSURLSession configuration

hakonk:
## Summary

While it is possible in the React Native implementation for Android to provide a custom configuration for HTTP requests, the iOS implementation does not allow for the same customization. As the NSURLSession used for HTTP requests on iOS is configured internally, one may for instance not supply an ephemeral configuration for HTTP requests. Other concerns related to the given problem have been addressed in the community: https://github.com/react-native-community/discussions-and-proposals/issues/166. I did make a PR with an RFC in the community repo, but after some discussion in the said repo, I figured I might as well make a PR with a suggestion :)

## Changelog

[iOS] [Added] - Allow for configuring the NSURLSessionConfiguration

Implement a C function `RCTSetCustomNSURLSessionConfigurationProvider` which gives the app programmer the ability to provide a block which provides an NSURLSessionConfiguration that will be used for all HTTP requests instead of the default configuration. The provided block will be called when the session configuration is needed.

## Test Plan

Unsure if this can be tested in any other way than uncommenting the example code in `RNTester/RNTester/AppDelegate.mm`.


akshetpandey:
+1 On this, this will greatly simplify implementations that need to use their own NSURLSessionConfiguration, ex. [RCTCronetHTTPRequestHandler.mm](https://github.com/akshetpandey/react-native-cronet/blob/master/ios/RCTCronetHTTPRequestHandler.mm)

zsiegel:
Any movement on this? I am looking for the ability to customize the iOS networking client like you do on Android.

hakonk:
@zsiegel I haven't heard from the review team. I assume they've had their hands full. Not sure if I should tag someone in the discussion, and in that case, who.

safaiyeh:
Hi @hakonk thanks for the PR, I'll see if I can ping someone for this.

hakonk:
Thanks, @safaiyeh 😊

hakonk:
Hey @safaiyeh, did you have luck finding anyone?

safaiyeh:
Hi @TheSavior, could someone check this out?

namchuai:
Nearly a year already..

hakonk:
Happy new year!

I guess there are more important PRs in the pipe still. Any news @safaiyeh ?

pke:
Interesting that this important (for SSL Pinning that can't use sizzling) PR is still not merged.

packages/rn-tester/RNTester/AppDelegate.mm
```diff
@@ -82,6 +82,12 @@ @implementation AppDelegate
 - (BOOL)application:(__unused UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
   RCTEnableTurboModule(YES);
+  // uncomment to configure NSURLSession
```
TheSavior, line 85: We probably don’t need to add commented code here as this is an uncommon use case. Perhaps this explanation of usage should just live in the documentation.

hakonk: Alright, I can add a section about that. Would it be appropriate to add a section [here](https://reactnative.dev/docs/network) for instance?

TheSavior: I think so

hakonk: I have created a PR that contains the needed documentation now: https://github.com/facebook/react-native-website/pull/2629

hakonk: Also, the commented code has been removed in: 1634c87

hakonk: @TheSavior Please see messages above 🙂

taylorkline:
@pke you mentioned this is useful for SSL pinning.

Can you elaborate on how? From what I can tell, [Apple's docs on performing manual server trust authentication](https://developer.apple.com/documentation/foundation/url_loading_system/handling_an_authentication_challenge/performing_manual_server_trust_authentication) state that we should provide an `NSURLSessionDelegate` that evaluates sever trust within `urlSession(_:didReceive:completionHandler:)`.

This PR does not allow us to set our own `NSURLSessionDelegate`, so how would this be useful in evaluating server trust?

@hakonk definitely open to your input as well. Thank you for this PR.

hakonk:
@taylorkline I believe you are right in that you would have to implement `NSURLSessionDelegate` as a part of implementing certificate or public key pinning, and this PR does not address that particular problem. In `RCTHTTPRequestHandler.mm:99`, you can see that `self` is assigned as the delegate for the `NSURLSession` instance. Thus, there is currently no way of supplying the delegate implementation from outside of React Native.

You can make use of swizzling when you do not have the means to implement `NSURLSessionDelegate` yourself. While I guess this would work most of the time, this mandates potential use of swizzling in other dependencies or your own code. Not really a problem in itself, it just isn't so transparent given that funky stuff happens at runtime. https://github.com/datatheorem/TrustKit does actually allow for implementing public key pinning with swizzling. That could possibly be useful if you want to achieve certificate or public key pinning in a React Native app.

taylorkline:
@hakonk Thank you for confirming and thank you for your time.

pke:
@taylorkline iirc one can only sizzle functions once. So if one react native module sizzles the NSURLSession (for certificate pinning) then no other module could sizzle the session. The ability to provide the pinning configuration for the sessions directly removes the need to use sizzling for it. So thanks for merging this!

taylorkline:
Is this in v0.65.0? The commit does not appear to be tagged with v0.65.0.

TheSavior:
0.65 was cut on April 28th, before this PR was merged. This PR will be included in 0.66

hakonk:
@TheSavior I'm a bit confused. The [changelog](https://github.com/facebook/react-native/pull/27701#issuecomment-900823660) for 0.65.0 lists this PR.

TheSavior:
Ah, I was looking at the tags on the commit as @taylorkline mentioned. It looks like this commit was cherry picked and thus doesn't have the same hash in the release branch. This commit is in 0.65.0 here: https://github.com/facebook/react-native/commit/c299694c87e9912d4c884b5acec04c708ed0f841

Sorry for the confusion! 😅

hakonk:
Alright, thanks for answering so quickly! 😄

PR #27729, opened 2020-01-10
[WIP] Migrate to FBJNI

passy:
## Summary

This is an incomplete effort to migrate from libfb to libfbjni. This is needed to restore the compatibility with Flipper and other FB Android projects that make use of FBJNI. Effectively, the outcome is that `fbjni` no longer has a checked-in copy here, but instead relies on the public artifacts published at github.com/facebookincubator/fbjni that can be deduplicated at build-time.

**A non-exhaustive list of tasks:**

* [X] Gradle builds the SDK and RNTester for Android.
* [X] Buck build for rntester works in OSS.
* [ ] Move from `java-only` release to full `fbjni` release. This requires finding a solution for stripping out `.so` files that the old `Android.mk` insists on including in the final artifacts and will clash with the full distribution.
* [ ] Import this and fix potential internal build issues.
* [ ] Verify that the changes made to the Hermes integration don't have any unintended consequences.

## Changelog

[Android] [Changed] - Migrated from libfb to libfbjni for JNI calls

## Test Plan

- CI is already passing again for Gradle and Buck in OSS.
- After applying the following patch, RNTester builds and works with the latest Flipper SDK:

```
diff --git a/RNTester/android/app/build.gradle b/RNTester/android/app/build.gradle
index b8a6437d7..eac942104 100644
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -170,10 +170,19 @@ dependencies {
     debugImplementation files(hermesPath + "hermes-debug.aar")
     releaseImplementation files(hermesPath + "hermes-release.aar")
 
-    debugImplementation("com.facebook.flipper:flipper:0.23.4") {
+    debugImplementation("com.facebook.flipper:flipper:+") {
         exclude group:'com.facebook.yoga'
-        exclude group:'com.facebook.flipper', module: 'fbjni'
-        exclude group:'com.facebook.litho', module: 'litho-annotations'
+        exclude group:'com.facebook.fbjni'
+    }
+
+    debugImplementation("com.facebook.flipper:flipper-network-plugin:+") {
+        exclude group:'com.facebook.yoga'
+        exclude group:'com.facebook.fbjni'
+    }
+
+    debugImplementation("com.facebook.flipper:flipper-fresco-plugin:+") {
+        exclude group:'com.facebook.yoga'
+        exclude group:'com.facebook.fbjni'
     }
 
     if (useIntlJsc) {
```


gengjiawen:
Looks like internal tests failed. cc @passy

passy:
@gengjiawen Yep, there's still a lot of internal build infra that needs to be adjusted before this can get landed.

chuganzy:
This approximately increased the binary size by 1mb I think. Is there any way to reduce the size?

rishabh876:
Is this required in release builds? libfbjni.so seems to be increasing size of release builds as well even though Flipper is only in debug builds.

PR #27730, opened 2020-01-10
[RCTImageLoader] Make C++ requirement opt-in

alloy:
## Summary

When building as a framework these headers get automatically added to the framework umbrella header for React-Core. Instead of converting all the React sources to ObjC++ files and still forcing external users that build native source (and link against a framework build) to also compile as ObjC++, this makes the attribution related methods that were added in https://github.com/facebook/react-native/commit/fdcdca4 opt-in to ObjC++ builds.

This is also the reason for the current failure of the CI `test_ios_frameworks` run.

## Changelog

I’m unsure if this change really warrants an entry in the CHANGELOG, as it’s more of an amendment of the (afaik) unreleased [change](https://github.com/facebook/react-native/commit/fdcdca4).

[iOS] [Fixed] - Make framework builds work again by making `RCTImageLoader` C++ requirement opt-in

## Test Plan

I tested static and dynamic (framework) builds and ran the test suite.

This change should make the `test_ios_frameworks` CI run _build_ again, ~~but it may still fail overall as in my local testing one of the tests leads to a segfault (which I will try to address separately)~~.

### DIFF
--- a/Libraries/Image/RCTImageLoaderWithAttributionProtocol.h
+++ b/Libraries/Image/RCTImageLoaderWithAttributionProtocol.h
@@ -17,6 +17,7 @@ RCT_EXTERN void RCTEnableImageLoadingPerfInstrumentation(BOOL enabled);
 
 @protocol RCTImageLoaderWithAttributionProtocol<RCTImageLoaderProtocol>
 
+#ifdef __cplusplus
 /**
  * Same as the variant in RCTImageURLLoaderProtocol, but allows passing attribution
  * information that each image URL loader can process.
@@ -30,6 +31,8 @@ RCT_EXTERN void RCTEnableImageLoadingPerfInstrumentation(BOOL enabled);
                                         progressBlock:(RCTImageLoaderProgressBlock)progressBlock
                                      partialLoadBlock:(RCTImageLoaderPartialLoadBlock)partialLoadBlock
                                       completionBlock:(RCTImageLoaderCompletionBlock)completionBlock;
+#endif
+
 /**
  * Image instrumentation - notify that the image content (UIImage) has been set on the native view.
  */
--- a/Libraries/Image/RCTImageURLLoaderWithAttribution.h
+++ b/Libraries/Image/RCTImageURLLoaderWithAttribution.h
@@ -7,6 +7,7 @@
 
 #import <React/RCTImageURLLoader.h>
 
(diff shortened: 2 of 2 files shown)


alloy:
A back ported version for the 0.62 branch is here https://github.com/facebook/react-native/pull/27731

alloy:
> but it may still fail overall as in my local testing one of the tests leads to a segfault (which I will try to address separately)

Looks like this is only a problem when running on iOS 13, CI runs iOS 12.4. Nonetheless a legit segfault, for when somebody on iOS 13 would try to serialise strings that can’t be utf8 encoded. Will do a bit of time-boxed sleuthing.

fkgozali:
Hmm, I don't think I intended `RCTImageURLLoaderWithAttribution.h` and `RCTImageURLLoaderWithAttribution.h` to be exported outside of the RCTImage lib/spec, so perhaps we can blacklist these 2 files in the podspec exported headers?

Maybe this import should do local header import instead: https://github.com/facebook/react-native/blob/eb95b2f8557ca9e763d2399f7f59ffa764fe95c3/Libraries/Image/RCTImageLoaderWithAttributionProtocol.h#L11

alloy:
Ah, I see. That should work, I’ll update it next week 👍

alloy:
@fkgozali I took a stab at this, but found that while it works well for framework builds, it will fail a static build. The reason is that e.g. [RCTImageView.mm](https://github.com/facebook/react-native/blob/8259a9d369c310f4c31e50e2ad5157f5c5a8dc30/Libraries/Image/RCTImageView.mm#L16) imports `React/RCTImageLoaderWithAttributionProtocol.h` rather than `RCTImage/RCTImageLoaderWithAttributionProtocol.h` and thus the header is always expected to be exported by the `React` namespace. I think that the current version of my PR is the simpler solution to this issue, what do you think?

fkgozali:
> The reason is that e.g. RCTImageView.mm imports React/RCTImageLoaderWithAttributionProtocol.h rather than RCTImage/RCTImageLoaderWithAttributionProtocol.h and thus the header is always expected to be exported by the React namespace.

I see the complication now. RCTImageView.mm is pre-Fabric, so I'm hoping with Fabric rolled out, this gets easier to solve (because Fabric is ObjC++ everywhere). It seems like your current change is the most practical to unblock RC atm. I'll merge.

Issue #11068, opened 2016-11-22
Duplicated letters when autoCapitalize="characters" on android


joncrocks:
### Description

When trying to implement an upper case only input, adding `autoCapitalize="characters"` seems to work on Android by setting the keyboard to be upper case by default. One can hit the shift key and then type a lower case letter.

To ensure that we only let the user enter (and see) upper case letters, I thought that I might be able to handle it by ensuring that when we update the react state of the component we capture the text in upper case.

By using a toUpperCase on top of what is a pretty standard state update cycle (i.e. a very similar update method ala the examples at https://facebook.github.io/react-native/releases/next/docs/textinput.html ), this saves the input into the state, uppercased, ready for the next render cycle. (I'm not concerned about the dangers of toUpperCase at this point.)

Unfortunately, the behaviour is a bit strange when you start typing both upper and lowercase letters, where you start getting repeated letters, e.g. if I type _AbC_, I will end up with _ABABC_, _AbcdE_ I get _ABABCDABABCDE_.

### Reproduction

I created an example app here: https://rnplay.org/apps/t-gBOA
Note that the behaviour seems fine on the iOS simulator, but is 'wrong' on the android simulator.

or see the component below:

```
import React, {Component} from "react";
import {View, TextInput} from "react-native";

export class UpperCaseTextEntry extends Component {
    
    constructor() {
        super();
        this.state = {
            text: ""
        }
    }
    
    upperCaseIt(text) {
        var textUpperCase = text.toUpperCase();

        this.setState({text: textUpperCase});
    }
    
    render() {
        var text = this.state.text;
        return (
            <View>
                <TextInput value={text} autoCapitalize="characters"
                           onChangeText={(text) => {this.upperCaseIt(text)}}
                />
            </View>
        )
        
    }
}
```

### Solution

I suspect that there's something going awry with the syncing of state between the react state and the state of the underlying components, quite possibly some case-ignoring checks are being in some places, but not others.

### Additional Information

* React Native version: 0.35.0
* Platform: Android
* Operating System: Windows (building/testing on windows)

I've also noted during that it's more than possible to fire multiple onChangeTexts before a render is triggered, which could lead to unexpected app behaviour. This could be expected unexpected behaviour though :-)

kenmasters:
Are there any updates on these issue already? , I am also experiencing this.

magneticz:
[...]

Issue #20119, opened 2018-07-09
TextInput becomes slow after lots of typing


gnprice:
## Environment

```
$ react-native info
Environment:
  OS: Linux 4.9
  Node: 8.11.3
  Yarn: 1.7.0
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Not Found

Packages: (wanted => installed)
  react: 16.3.1 => 16.3.1
  react-native: 0.56.0 => 0.56.0 (*)
```

(*) In my test app, actually v0.55.4 plus the patch from #19645 . Others on #19126 report the same symptoms using v0.56.

## Description

After typing a lot of text into a controlled TextInput, like ~500 char on a recent fast phone, we start dropping frames. It gets worse the more you type; at ~1000 char, we very frequently drop frames and the app looks noticeably laggy.

## Reproducible Demo

The original repro is by @s-nel: https://github.com/s-nel/rn19126

Originally reported at https://github.com/facebook/react-native/issues/19126#issuecomment-402904164; more discussion in subsequent comments in that thread.

In particular, here's the video from that repro:
![video](https://github.com/s-nel/rn19126/raw/master/frameloss.gif)

Here's a video by @s-nel on [my test app](https://github.com/gnprice/rn-textinput-test), with this description:
> Yes I can reproduce from [that repo] after typing for a couple minutes straight. I don't have to clear anything as the original bug description [in #19126] indicates. Here you can see my JS framerate drop to single digits from adding a few characters

![image](https://user-images.githubusercontent.com/504326/42430009-adf31f14-8390-11e8-9c2e-527533484d35.gif)

Here's my description of replicating that repro, again in my test app:

> I just tried again with my test app, on RN 0.55.4 + patch. If I type continuously for about 60 seconds -- maybe 400-500 characters, at a rough estimate (just gibberish, both thumbs constantly typing letters) -- then the perf overlay gets to about "10 dropped so far", and zero stutters. If I continue and get up to about 90 seconds, the "dropped" figure climbs faster, and the framerate (in "JS: __ fps") noticeably drops. At about 150 seconds (~1000 characters?), the "dropped" figure rapidly climbs past 200, and the framerate hangs out around 30.
>
> Even at that point, it's still "0 stutters". And if I ignore the perf overlay and try typing normal text and watching it like I were using an app for real: it's a bit laggy, but doesn't make things feel unusable or outright broken, and I think most of the time I wouldn't even notice.

[...]

PR #27757, opened 2020-01-14
[TextInput]: Fixed duplicated letters being added when changing text through onChangeText

safaiyeh:
## Summary
Fix for https://github.com/facebook/react-native/issues/11068.
` getText().replace` caused a bug where letters would be duplicated when the text was being set through onChangeText:

```
<TextInput ref={(el) => this.input = el} onChangeText={(val) => { this.input.setNativeProps({text: val.toUpperCase()}); }}/>
```

Debugged and tried a bunch of different ways to solve it explained here: https://github.com/facebook/react-native/issues/11068#issuecomment-574013816

## Changelog

[Android] [Fixed] - Fixed duplicated letters being added when changing text through onChangeText

## Test Plan

Added this to RNTester:
```
<TextInput ref={(el) => this.input = el} onChangeText={(val) => { this.input.setNativeProps({text: val.toUpperCase()}); }}/>
```
Verified it works properly and there were no regressions in the other examples.

Note: There is some performance issues when typing for a long time, probably a lag between clearing and appending. Would love to see if someone has suggestions in handling that.

**Edit**: Seems like the perf is a known issue https://github.com/facebook/react-native/issues/20119

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -503,7 +503,8 @@ public class ReactEditText extends AppCompatEditText {
       // When we update text, we trigger onChangeText code that will
       // try to update state if the wrapper is available. Temporarily disable
       // to prevent an infinite loop.
-      getText().replace(0, length(), spannableStringBuilder);
+      getText().clear();
+      append(spannableStringBuilder);
     }
     mDisableTextDiffing = false;
 


JoshuaGross:
Does this retain selections when typing in the middle of text?

Could you add some videos?

safaiyeh:
That is true @JoshuaGross. Append would always put the cursor to the end. Do you have any ideas I could try for this? I debugged this for a while and can't find a sold answer.

Maybe a diff check and find the appropriate place to put the cursor?

martianatwork:
@safaiyeh @JoshuaGross I’m not expert at this but here are some suggestions, can we introduce a function that transforms the text at java instead of letting keyboard transform text? 

Ex. We have autocapitalise option then on changetext we can apply the transformation and forward the updated text to view. Currently user can type in lowercase even if the uppercase is set.

Or we create a transformation prop which will update the text at JavaScript and the complete state will be managed at JavaScript. 

I’m not sure about the performance impact.

safaiyeh:
@martianatwork I like those solutions. I will give them a shot!

One flaw to having options for on change is custom text transforms. I like the second solution because it can cover both.

vonovak:
textinput is _very_ tricky. We wrote our own component for rich text input, and just want to add that if this is to be merged, it should be very thoroughly tested. I think some issues will only get resolved once we have sync communication between JS and native (fabric stable).

safaiyeh:
Is that component open sourced? @vonovak

Ive working on a text input heavy app and the performance kills UX at some point.

vonovak:
@safaiyeh no, it's not open sourced, and it's highly unlikely that it would be...

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -503,7 +503,8 @@ public void maybeSetText(ReactTextUpdate reactTextUpdate) {
       // When we update text, we trigger onChangeText code that will
       // try to update state if the wrapper is available. Temporarily disable
       // to prevent an infinite loop.
-      getText().replace(0, length(), spannableStringBuilder);
+      getText().clear();
+      append(spannableStringBuilder);
     }
```
fabriziobertoglio1987, line 508: Pull request https://github.com/safaiyeh/react-native/pull/2

```java
Integer currentPosition = getSelectionStart();
getText().clear();
append(spannableStringBuilder);
setSelection(currentPosition);
```

`setSelection` sets the cursor position to the previous one after append operation. 

This is some quick testing of the functionality with the react-native master branch. Let me know your feedback, I can publish a separate Pull Request if you are not available otherwise I can futher help you to review your pull request and test the changes.

| **BEFORE** | **AFTER** | 
|  |  |

Thanks a lot
Fabrizio Bertoglio

fabriziobertoglio1987: @safaiyeh I'm preparing a pull request with this solution. Let me know if I should stop. I wish you a good weekend. Thanks. Fabrizio.

fabriziobertoglio1987: https://github.com/facebook/react-native/pull/29070

SrBrahma:
I have the same issue when using toUpperCase method on the input string.

sdandois:
How can I apply this as a patch?

lunaleaps:
@safaiyeh, @fabriziobertoglio1987 Is this PR still relevant with https://github.com/facebook/react-native/pull/29070 open? They seem like they both fix issue #11068.

If so, can I close this PR?

safaiyeh:
Yup that PR takes care of it. I’ll close this @lunaleaps

PR #27762, opened 2020-01-14
Integrate Flipper in RNTester on iOS

priteshrnandgaonkar:
Summary: Integrates Flipper in RNTester application and fixes the integration issues with `use_frameworks!` keyword.

Differential Revision: D19391739

### DIFF
--- a/RNTester/Podfile
+++ b/RNTester/Podfile
@@ -17,6 +17,7 @@ def pods()
   # Additional Pods which aren't included in the default Podfile
   pod 'React-ART', :path => '../Libraries/ART'
   pod 'React-RCTPushNotification', :path => '../Libraries/PushNotificationIOS'
+  pod 'Yoga',:path => '../ReactCommon/yoga', :modular_headers => true
 
   # Additional Pods which are classed as unstable
   #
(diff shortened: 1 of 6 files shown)


RNTester/Podfile
```diff
@@ -17,15 +17,69 @@ def pods()
...
+      Pod::Installer::Xcode::TargetValidator.send(:define_method, :verify_no_static_framework_transitive_dependencies) {}
+      installer.pod_targets.each do |pod|
+          if $static_framework.include?(pod.name)
+            def pod.build_type;
+              Pod::Target::BuildType.static_library
+            end
+          end
+        end
+    end
+  end
```
alloy: @priteshrnandgaonkar Could you link me to a write-up, if one exists, on the need for these deps to be linked statically?

pritnandgaonkar: Go through the iOS setup mentioned [here](https://fbflipper.com/docs/getting-started.html)

priteshrnandgaonkar: Go through the iOS setup mentioned [here](https://fbflipper.com/docs/getting-started.html)

alloy: Thanks. I’ve seen those before, but they do not describe _why_ these dependencies need to be linked statically and why a monkey-patch of CocoaPods is required. I would like to understand those technical reasons, as I would like to see if there’s a way to avoid the need to monkey-patch.

priteshrnandgaonkar: The reason being, Flipper relies on CocoaLibEvent which has bunch of statically linked binaries libevent.a, libevevnt_core.a etc. due to which the SubPods relying on it has to be built as a Static Framework, which is Flipper-Folly. So ideally speaking building just CocoaLibEvent, Flipper-Folly as a static framework should work, but for some reason the dynamic framework failed to link Flipper-Folly, which I was not able to solve. Thus I built all the Flipper and Flipper related deps as a Static Framework.

Even I feel that we shouldn't have a monkey-patch, if you can solve that problem, then it would be great.

alloy: Ok, when I’m back from vacation I’ll take a look-see 👍

alloy:
@priteshrnandgaonkar This has already landed in master, right? cb80e3b599d0a781af8ad2999c4f1b58a11de05f

priteshrnandgaonkar:
Yes, I will close the PR.

PR #27766, opened 2020-01-15
[CI] Use correct location for buck_to_junit

hramos:
## Summary

`test_android`'s Test Collection step fails when it attempts to convert buck's output to the junit-compatible format that Circle CI expects. Executing the script from the `okbuck` location should help here.

## Changelog

[Internal] - Use correct location for buck_to_junit in test_android CI step

## Test Plan

Circle CI


hramos:
Merging as this fixes test_android.

PR #27790, opened 2020-01-16
Bring back support for Dialogs over non-FragmentActivity host

mganandraj:
## Summary

[This](https://github.com/microsoft/react-native/commit/243070afe2873c09c1149ca3ede3386abb88a529#diff-8e3b74ad18aa49187acd4240321bc9b9) commit had earlier (intentionally) stopped supporting DialogModule over non-FragmentActivity hosts. 

The change was done for a genuine reason, but unfortunately, we have a number of brownfield clients within Microsoft who are finding it hard to change their host host activities in time, due to various reasons. As we expect the transition to take a while (Another 6 months), we would like to put back the support for legacy FragmentManager in DialogModule.

This change is an essential revert of the above mentioned commit.

## Changelog

Bring back the support for legacy FragmentManager in DialogModule. This is required for brownfield applications who don't have ReactActivity as the host.

[CATEGORY] [TYPE] - Message

## Test Plan

1. Tested Alert dialogs over FragmentActivity and vanilla Activity.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/DialogModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/DialogModule.java
@@ -15,7 +15,6 @@ import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.FragmentActivity;
-import androidx.fragment.app.FragmentManager;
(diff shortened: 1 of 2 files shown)


dulmandakh:
I'm not sure about FB opinion about supporting non-Fragment activities, @mdvacca please review. Thanks

manavsmittal:
poopoopoo

PR #27809, opened 2020-01-19
Add Expo dot folder to .gitignore

GeeksdelaCave:
Hello,

## Summary

When using expo for React native instant verification, I saw that the configuration folder wasn't in actual .gitignore.

## Changelog

[GENERAL] [CHANGED] - Adding .expo folder to .gitignore

## Test Plan

Now .expo folder is ignored by tracking.
Capture d’écran 2020-01-19 à 22 08 44

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,6 @@ RNTester/Pods/*
 # Visual studio
 .vscode
 .vs
+
+# Expo
+.expo/


cpojer:
Thank you for your PR. I don't think this needs to be in the main RN repo. What were you trying to do?

cc @notbrent to help me understand whether I'm wrong.

GeeksdelaCave:
Hello,

> Thank you for your PR. I don't think this needs to be in the main RN repo. What were you trying to do?

I used since a long time the .gitignore from this repository for my projects. Recently, I discovered [expo](https://github.com/expo/expo), a useful tool that help in testing applications.

But the folder it created doesn't appears in this .gitignore. I was hoping that it can be handy for others.

cpojer:
That makes sense but the gitignore here is fore React Native itself, which does not use Expo.

PR #27838, opened 2020-01-22
Add missing NativeEventListener methods to NativeDevSettings

janicduplessis:
## Summary

Since migrating to Turbomodules (8fe04cf) the addMenuItem method crashes because the NativeEventListener methods are missing from the codegen flow type. Added the same methods based on what we do in AppState which is another native module that extends NativeEventListener.

## Changelog

[Internal] [Fixed] - Add missing NativeEventListener methods to NativeDevSettings

## Test Plan

TODO: Need someone at fb to run codegen and run the RNTester DevSettings example

### DIFF
--- a/Libraries/NativeModules/specs/NativeDevSettings.js
+++ b/Libraries/NativeModules/specs/NativeDevSettings.js
@@ -23,6 +23,10 @@ export interface Spec extends TurboModule {
   +toggleElementInspector: () => void;
   +addMenuItem: (title: string) => void;
 
+  // Events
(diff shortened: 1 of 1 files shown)


janicduplessis:
Cc @PeteTheHeat

PeteTheHeat:
Nice catch, wonder how this passed internal land, I'll run RNTester.

janicduplessis:
The crash only triggers when calling the addMenuItem function since listeners are added lazily. I dont think this api is used at fb, I added it recently.

PR #27841, opened 2020-01-22
Add in the optimization to check the first/least timer in the queue before iterating

ejanzer:
## Summary
Follow up from https://github.com/facebook/react-native/pull/27539 - adding back in the optimization that Detox has in TimersIdlingResource to avoid iterating over the entire timers queue if the next timer is already within the specified range.

## Changelog
[Internal]

## Test Plan
TimingModuleTest

Differential Revision: D19522346

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/core/JavaTimerManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/core/JavaTimerManager.java
@@ -384,12 +384,26 @@ public class JavaTimerManager {
    */
   /* package */ boolean hasActiveTimersInRange(long rangeMs) {
     synchronized (mTimerGuard) {
+      Timer nextTimer = mTimers.peek();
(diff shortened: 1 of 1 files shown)


ejanzer:
cc @d4vidi

d4vidi:
LGTM, except for maybe these concerns:
1. Some (very minimal) code duplication (timer "active" or not)
2. Not sure what the policy here but I'd add test coverage using unit tests.

ejanzer:
@d4vidi Agree about the repetition, I'll make a change. I did actually add unit tests for this function in the previous PR: https://github.com/facebook/react-native/blob/master/ReactAndroid/src/test/java/com/facebook/react/modules/timing/TimingModuleTest.java#L259

I haven't added any tests explicitly for this optimization, because that feels like an implementation detail that probably doesn't make sense to test (the functionality hasn't changed).

PR #27844, opened 2020-01-23
Split loadApplicationScript into initializeRuntime and loadBundle

simka:
## Summary

This is the first of three PRs related to enabling multi-bundle support in React Native. More details, motivation and reasoning behind it can be found in RFC [here](https://github.com/react-native-community/discussions-and-proposals/issues/152).

Logic responsible for installing globals was pulled out from `loadApplicationScript` to `initializeRuntime` since it should be ran only once, what was left was renamed to `loadBundle`.

It's based on @dratwas work from [here](https://github.com/callstack/react-native/tree/feat/multibundle/split-load-application), but applied to current `master` to avoid rebasing 3-months old branch and issues that come with that.

## Changelog

[Internal] [Changed] - split `loadApplicationScript` into `initializeRuntime` and `loadBundle` to enable multi-bundle support in the future

## Test Plan


zamotany:
@ejanzer 
`initializeRuntime` is only empty for the specific executor - `RCTObjcExecutor`. For other executors like `JSIExecutor` and `ProxyExecutor`, `initializeRuntime` will contain a logic, that should ideally be run only once. We've tried to more some of the logic (that made sense) out from `RCTObjcExecutor` constructors to it's `initializeRuntime`, but the app was crashing.

joeblynch:
Hey @ejanzer, thanks for taking a look at this PR! Based on @zamotany's comment, do you think this PR is still needed? Also just a heads up, the other PRs that split out the work from the original RFC are up [here](https://github.com/facebook/react-native/pull/27855) and [here](https://github.com/facebook/react-native/pull/27875).

ReactCommon/cxxreact/Instance.cpp
```diff
@@ -48,7 +48,8 @@ void Instance::initializeBridge(
   jsQueue->runOnQueueSync([this, &jsef, jsQueue]() mutable {
     nativeToJsBridge_ = std::make_unique<NativeToJsBridge>(
         jsef.get(), moduleRegistry_, jsQueue, callback_);
-
+    // TODO investigate why it has to be async
```
ejanzer: Why what has to be async? InitializeRuntime?

ReactCommon/cxxreact/JSExecutor.h
```diff
@@ -53,10 +53,14 @@ class JSExecutorFactory {
 
 class RN_EXPORT JSExecutor {
 public:
+  /**
+   * Sets all globals in the JS context.
```
ejanzer: Maybe something like: "Prepares the JS runtime for React Native by installing global variables. Called once before any JS is evaluated."

ReactAndroid/src/main/java/com/facebook/react/bridge/ReactMarkerConstants.java
```diff
@@ -59,8 +59,8 @@
   UNPACKING_JS_BUNDLE_LOADER_CHECK_END,
   UNPACKING_JS_BUNDLE_LOADER_EXTRACTED,
   UNPACKING_JS_BUNDLE_LOADER_BLOCKED,
-  loadApplicationScript_startStringConvert,
-  loadApplicationScript_endStringConvert,
+  loadBundle_startStringConvert,
```
ejanzer: This is the only thing that I'm a little bit concerned about - renaming our perf markers could be a breaking change for our perf instrumentation. I think it might be better to keep these the same, even though they won't match the function name anymore.

joeblynch:
Thank you @ejanzer for reviewing these PRs!

ReactAndroid/src/main/java/com/facebook/react/bridge/ReactMarkerConstants.java
zamotany: Totally valid concert, we'll revert that change.

ReactCommon/cxxreact/Instance.cpp
zamotany: I think this is the old leftover, where we make the initial drafts of changes for RN 0.60. When implementing it, we've noticed that the code in `NativeToJsBridge::initializeRuntime` has to be run using `runOnExecutorQueue`. We haven't investigated more why we have to do it like that, since the code was working as expected and was crashing if `runOnExecutorQueue` was not used. This comment can be discarded and we'll remove it.

mhorowitz: With some versions of JSC, if you create the runtime on one thread, and then use it on another, GC bugs will cause crashes.  That's why this has to be done this way.  It's not pretty, but there's not much else to be done if you want to use older versions of iOS.  I'm not sure if the android JSC has this bug or not.

grabbou:
Hey @ejanzer, is there any update on this? Seems to be taking a while, could be due to some internal merge conflicts.

PR #27851, opened 2020-01-24
fix: resolve Require Cycle warning

Naturalclar:
## Summary

When I was testing React Native 0.62-rc.1, I noticed that console was showing a warning for Require Cycle as shown in image below

Screen Shot 2020-01-24 at 12 22 38

This is because ScrollResponder was importing `ScrollView` to get the `typeof ScrollView`.
I've made an export for ScrollView type in `ScrollView` so that Require cycle warning will not show up.

## Changelog

[General] [Fixed] - Remove Require cycle warning.

### DIFF
--- a/Libraries/Components/ScrollResponder.js
+++ b/Libraries/Components/ScrollResponder.js
@@ -18,14 +18,13 @@ const ReactNative = require('../Renderer/shims/ReactNative');
 const TextInputState = require('./TextInput/TextInputState');
 const UIManager = require('../ReactNative/UIManager');
 const Platform = require('../Utilities/Platform');
-const ScrollView = require('./ScrollView/ScrollView');
 import Commands from './ScrollView/ScrollViewCommands';
 
 const invariant = require('invariant');
 const performanceNow = require('fbjs/lib/performanceNow');
 
 import type {PressEvent, ScrollEvent} from '../Types/CoreEventTypes';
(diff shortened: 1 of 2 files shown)


Libraries/Components/ScrollResponder.js
```diff
@@ -18,14 +18,16 @@ const ReactNative = require('../Renderer/shims/ReactNative');
...
 
 const invariant = require('invariant');
 const performanceNow = require('fbjs/lib/performanceNow');
 
 import type {PressEvent, ScrollEvent} from '../Types/CoreEventTypes';
-import type {Props as ScrollViewProps} from './ScrollView/ScrollView';
+import type {
+  Props as ScrollViewProps,
+  ScrollViewType,
+} from './ScrollView/ScrollView';
```
TheSavior: Based on https://flow.org/en/docs/types/modules/#toc-importing-and-exporting-types

I think you can change this to something like:

```
import type ScrollView, {Props as ScrollViewProps} from './ScrollView/ScrollView';
```

And then you don't need to export the type

Naturalclar: @TheSavior Thank you!
The syntax was a little different, but I was able to use
```jsx
import typeof ScrollView from './ScrollView/ScrollView
import type {Props as ScrollViewProps } from './ScrollView/ScrollView'
```

PR #27855, opened 2020-01-24
Create bundle registry in Executor and pass bundle to register bundle (follow-up to #27844)

simka:
## Summary

This is the second of three PRs related to enabling multi-bundle support in React Native. First one can be found [here](https://github.com/facebook/react-native/pull/27844). More details, motivation and reasoning behind multi-bundle support can be found in RFC [here](https://github.com/react-native-community/discussions-and-proposals/issues/152).

As the first one, it's based on @dratwas work from [here](https://github.com/callstack/react-native/tree/feat/multibundle/ram-bundles), but applied to current `master` to avoid rebasing 3-months old branch and issues that come with that.

## Changelog

[Internal] [Changed] - Create bundle registry in Executor and pass bundle to register bundle to enable multi-bundle support in the future

## Test Plan


React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -1345,7 +1344,7 @@ - (void)executeApplicationScript:(NSData *)script
 - (void)registerSegmentWithId:(NSUInteger)segmentId path:(NSString *)path
 {
   if (_reactInstance) {
-    _reactInstance->registerBundle(static_cast<uint32_t>(segmentId), path.UTF8String);
+    /* _reactInstance->registerBundle(static_cast<uint32_t>(segmentId), path.UTF8String); */
```
ejanzer: Did you mean to leave this commented out?

ReactCommon/cxxreact/Instance.cpp
```diff
@@ -201,7 +195,7 @@ void Instance::callJSCallback(uint64_t callbackId, folly::dynamic &&params) {
 void Instance::registerBundle(
     uint32_t bundleId,
     const std::string &bundlePath) {
-  nativeToJsBridge_->registerBundle(bundleId, bundlePath);
+  /* nativeToJsBridge_->registerBundle(bundleId, bundlePath); */
```
ejanzer: And here

React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -1328,25 +1328,42 @@ - (void)executeApplicationScript:(NSData *)script
...
+    NSData *sourceCode = source.data;
+    __strong RCTCxxBridge *strongSelf = weakSelf;
+    if (strongSelf->_reactInstance) {
+      [strongSelf executeApplicationScript:sourceCode url:pathURL async:NO];
+    }
+    dispatch_group_leave(group);
+        
+    }];
+    
+    dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
```
zamotany: @ejanzer We had to modify the code in `registerSegmentWithId` and replace it with this one. 2nd argument of `registerBundle` function on `_reactInstance` is no longer a string but the `JSModulesUnbundle`, so in `registerSegmentWithId` we now use `RCTJavaScriptLoader` to load ther bundle either from packager server or filesystem and then `executeApplicationScript` which now can be called multiple times. All of that is wrapped in a `dispatch_group` to make it synchronous to keep compatibility with the old behaviour. The same code was applied in the 3rd PR as well: #27875

cpojer:
This diff adds OTA-like functionality to React Native core which we cannot currently support. From looking at the code it will also break a bunch of stuff at FB so it'll likely be very hard to integrate. It's confusing to add a "bundleId" to so many callsites, I don't think it is necessary to increase complexity like this.

I think we should take a different approach. I landed a diff that exposes a new loader on the bridge, which will allow you to build your own TurboModule to load bundles outside of React Native. It also comes with an example TurboModule that should only be used for development purposes (ie. we will not be accepting changes to this module to make it work in prod). This should give you an idea of what it would take to build your own OTA client including downloading, verification, storage and evaluation of code outside of RN core.

See https://github.com/facebook/react-native/commit/ad879e50bcd51caca76b1073720f2b63df485ff1

PR #27875, opened 2020-01-27
Add bundleId parameter (follow-up to #27844 and #27855)

simka:
## Summary

This is the third (and last) of three PRs related to enabling multi-bundle support in React Native. First one can be found [here](https://github.com/facebook/react-native/pull/27844) and the second one [here](https://github.com/facebook/react-native/pull/27855). More details, motivation and reasoning behind multi-bundle support can be found in RFC [here](https://github.com/react-native-community/discussions-and-proposals/issues/152).

As the first and second one, it's based on @dratwas work from [here](https://github.com/callstack/react-native/tree/feat/multibundle/bundleId), but applied to current `master` to avoid rebasing 3-months old branch and issues that come with that.

## Changelog

[Internal] [Changed] - Added `bundleId` param to enable multi-bundle support in the future

## Test Plan


KingAmo:
any updates ?

Issue #27274, opened 2019-11-20
missing global.nativePerformanceNow


parth3724:
In React native android app, using react native version 0.59.8, global.nativePerformanceNow is not defined.

React Native version: 0.59.8

## Steps To Reproduce

1. Try to get time value in any js file using global.nativePerformanceNow()

Describe what you expected to happen:
As it was described in RN version 55.3 AndroidJSCFactory.cpp,
  // Since SystemClock.uptimeMillis() is commonly used for performance measurement in Java
  // and uptimeMillis() internally uses clock_gettime(CLOCK_MONOTONIC),
  // we use the same API here.
  // We need that to make sure we use the same time system on both JS and Java sides.

We cannot use systrace in Android (since only stub is open sourced). We invested so much in building tooling to measure performance, and now when we change the version, such basic api vanishes.

https://github.com/facebook/react-native/blob/0.53-stable/ReactAndroid/src/main/jni/react/jni/AndroidJSCFactory.cpp

emilisb:
I have encountered very same issue. After Profiler API (https://reactjs.org/docs/profiler.html) was released, this feature is even more important than before. Currently it is impossible to get precise execution time without the real performance.now() (and without using a remote debugger). Without performanceNow we are forced to use local date, which is not good enough. 

It seems like this feature was removed during transition from JSC to JSI and it was never reimplemented.

I found commits that removed this feature:

Android: https://github.com/facebook/react-native/commit/7ffb406517e70a2ac299d20020d3798a5b7dc612#diff-c2a333229249aaff833b2a50be32a489

iOS: https://github.com/facebook/react-native/commit/816d41718998868f276d83b0c21e17d11ad392a2

I don't see any reason to not reimplement it again. 

In addition to that, could we implement a performance polyfill with at least most popular functions? Performance docs: https://developer.mozilla.org/en-US/docs/Web/API/Performance

PR #27885, opened 2020-01-28
Implement nativePerformanceNow to improve Profiler API results

emilisb:
## Summary

When experimenting with React Profiler API (https://reactjs.org/docs/profiler.html), I noticed that durations are integers without a debugger, but they are doubles with higher precision when debugger is attached. After digging into React Profiler code, I found out that it's using `performance.now()` to accumulate execution times of individual units of work. Since this method does not exist in React Native, it falls back to Javascript `Date`, leading to imprecise results.

This PR introduces `global.nativePerformanceNow` function which returns precise native time, and a very basic `performance` polyfill with `now` function.

This will greatly improve React Profiler API results, which is essential for profiling and benchmark tools.

Solves #27274

## Changelog

[General] [Added] - Implement `nativePerformanceNow` and `performance.now()`

## Test Plan

```
const initialTime = global.performance.now();
setTimeout(() => {
  const newTime = global.performance.now();
  console.warn('duration', newTime - initialTime);
}, 1000);
```

### Android + Hermes

Screenshot_1580198068

### Android + JSC

Screenshot_1580199089

### iOS

Simulator Screen Shot - iPhone 8 - 2020-01-28 at 10 06 49

### DIFF
--- a/Libraries/polyfills/performance.js
+++ b/Libraries/polyfills/performance.js
@@ -1,0 +1,18 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow strict
+ * @polyfill
+ */
+
+if (!global.performance) {
+  global.performance = {
+    now: function() {
+      return global.nativePerformanceNow();
+    },
+  };
+}
--- a/Libraries/polyfills/performance.js
+++ b/Libraries/polyfills/performance.js
@@ -10,9 +10,15 @@
  */
 
 if (!global.performance) {
-  global.performance = {
-    now: function() {
-      return global.nativePerformanceNow();
-    },
+  global.performance = {};
+}
+
+/**
+ * Returns a double, measured in milliseconds.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+ */
+if (typeof global.performance !== 'function') {
+  global.performance.now = function() {
+    return global.nativePerformanceNow();
   };
 }
--- a/Libraries/polyfills/performance.js
+++ b/Libraries/polyfills/performance.js
@@ -17,8 +17,9 @@ if (!global.performance) {
  * Returns a double, measured in milliseconds.
  * https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
  */
-if (typeof global.performance !== 'function') {
+if (typeof global.performance.now !== 'function') {
   global.performance.now = function() {
-    return global.nativePerformanceNow();
+    const performanceNow = global.nativePerformanceNow || Date.now;
+    return performanceNow();
   };
 }
--- a/React/CxxBridge/JSCExecutorFactory.mm
+++ b/React/CxxBridge/JSCExecutorFactory.mm
@@ -25,6 +25,13 @@ std::unique_ptr<JSExecutor> JSCExecutorFactory::createJSExecutor(
         [NSString stringWithUTF8String:message.c_str()]);
     };
     react::bindNativeLogger(runtime, iosLoggingBinder);
+
+    react::PerformanceNow iosPerformanceNowBinder =[]() {
+        // CACurrentMediaTime() returns the current absolute time, in seconds
+        return CACurrentMediaTime() * 1000;
+    };
+    react::bindNativePerformanceNow(runtime, iosPerformanceNowBinder);
+
     // Wrap over the original runtimeInstaller
     if (runtimeInstaller) {
       runtimeInstaller(runtime);
--- a/ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
+++ b/ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <time.h>
+#include "NativeTime.h"
+
+namespace facebook {
+namespace react {
+
+double reactAndroidNativePerformanceNowHook() {
+  struct timespec now;
+  clock_gettime(CLOCK_MONOTONIC, &now);
+
+  return (now.tv_sec * 1000000000LL + now.tv_nsec) / 1000000.0;
+}
+
+} // namespace react
+} // namespace facebook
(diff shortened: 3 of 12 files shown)


rn-get-polyfills.js
```diff
@@ -13,4 +13,5 @@ module.exports = () => [
   require.resolve('./Libraries/polyfills/console.js'),
   require.resolve('./Libraries/polyfills/error-guard.js'),
   require.resolve('./Libraries/polyfills/Object.es7.js'),
+  require.resolve('./Libraries/polyfills/performance.js'),
```
ejanzer: Do we need to have the polyfill? I think we really want to remove polyfills instead of adding more... what if we just added a JS module that people could require normally if they want to use it?

ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "NativeTime.h"
+
+namespace facebook {
+namespace react {
+
+double reactAndroidNativePerformanceNowHook() {
+  struct timespec now;
+  clock_gettime(CLOCK_MONOTONIC, &now);
+
+  return (now.tv_sec * 1000000000LL + now.tv_nsec) / 1000000.0;
```
ejanzer: nit: I think it'd be preferable to create more semantic consts for these values (e.g. NANOSECONDS_IN_SECOND) so that it's easier for people to understand exactly what this is doing

Libraries/polyfills/performance.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @format
+ * @flow strict
+ * @polyfill
+ */
+
+if (!global.performance) {
+  global.performance = {
+    now: function() {
+      return global.nativePerformanceNow();
```
ejanzer: Is it possible for `nativePerformanceNow` to not exist after this change? If so, it'd be nice to fall back to the less-precise method in that case.

ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <time.h>
+#include "NativeTime.h"
+
+namespace facebook {
+namespace react {
+
+double reactAndroidNativePerformanceNowHook() {
+  struct timespec now;
+  clock_gettime(CLOCK_MONOTONIC, &now);
```
mganandraj: Any reason for not using std::steady_clock/high_resolution_clock ? They usually end up calling clock_gettime on POSIX based platforms .. That would allow this fix to apply for all engines/platforms (including windows !)

emilisb: Good point, thanks! I changed it to high_resolution_clock, I can confirm it still works on Android.

ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
emilisb: Sure, I moved it to a constant, thanks!

rn-get-polyfills.js
emilisb: I looked through the code, and I noticed that if Systrace is enabled, `Libraries/Performance/Systrace.js` would override `global.performance`. What if we load performance module in `InitializeCore.js`, and combine `performance.now` with all the methods from Systrace into one module that would define `performance`?

ReactAndroid/src/main/jni/react/jni/NativeTime.cpp
alexeylang: The standard says that: "The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided. It is often just an alias for std::chrono::steady_clock or std::chrono::system_clock, but which one it is depends on the library or configuration."
For performance measurements we need a monotonic clock so you should probably use steady_clock here 🙂

emilisb: Thanks!

emilisb:
Could someone review this please?

rn-get-polyfills.js
ejanzer: Why can't we access `nativePerformanceNow`? It should be installed already, right?

Forgive the ignorant question, but - what would happen if, instead of installing `global.nativePerformanceNow`, we just actually installed `global.performance.now` directly instead? Would that be a disaster?

Libraries/polyfills/performance.js
```diff
@@ -0,0 +1,25 @@
...
+  global.performance = {};
+}
+
+/**
+ * Returns a double, measured in milliseconds.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
+ */
+if (typeof global.performance.now !== 'function') {
+  global.performance.now = function() {
+    const performanceNow = global.nativePerformanceNow || Date.now;
```
ejanzer: Is it possible for `global.nativeePerformanceNow` not to exist, after this change? It seems like you're adding this to all the executors, right?

React/CxxBridge/JSCExecutorFactory.mm
```diff
@@ -25,6 +25,13 @@
         [NSString stringWithUTF8String:message.c_str()]);
     };
     react::bindNativeLogger(runtime, iosLoggingBinder);
+
+    react::PerformanceNow iosPerformanceNowBinder =[]() {
+        // CACurrentMediaTime() returns the current absolute time, in seconds
+        return CACurrentMediaTime() * 1000;
+    };
+    react::bindNativePerformanceNow(runtime, iosPerformanceNowBinder);
```
ejanzer: Is there a reason to bind this here (in each executor factory) instead of in somewhere more central like Instance::bindBridge? (I guess that would probably involve changing the API for bindBridge to add another parameter, which would be a breaking change for any other executors that aren't part of RN core). I'm just wondering because it seems like the difference between how we bind nativePerformanceNow is based on the platform we're running on, not the JS VM, so we don't necessarily need this to be in the JS executor factory.

rn-get-polyfills.js
emilisb: I have no idea why I initially thought we cannot access `nativePerformanceNow` there. I moved the polyfill to Core instead.

The rest of `performance` implementation is in JS (it's part of `Systrace`), so I think it would be better to keep `now` in JS as well. Also, it's nice to have a fallback in case the executor has no `performance.now` implementation.

Libraries/polyfills/performance.js
emilisb: I think we should have a fallback for unofficial executors.

React/CxxBridge/JSCExecutorFactory.mm
emilisb: It seemed like a good place to use without making any breaking changes. Logging binder is there too.

pie6k:
Is it possible to somehow use it in regular app code?

I've got some complex computations in my app. They're highly optimized, but they're called very often (eg. 30k times on app boot). If I try to measure it - most of the calls are under 1ms so JS measures it as 0ms which is not really useful.

Also, it's hard in my case to group them together, as they're not called together, but in various scopes.

PR #27906, opened 2020-01-30
Fixed Grammar Issue

iqbalfaisal:
## Summary
Fixed Grammar Issue in ImageProps File


emilisb:
I don't think this is a grammar mistake, Stringish is a declared flow type: https://github.com/facebook/react-native/blob/master/flow/Stringish.js

TheSavior:
Yeah, this is meant to be Stringish to enable overriding that type for internationalization. Thanks anyways!

PR #27915, opened 2020-01-30
iOS: Fix dev menu sometimes not showing

tido64:
## Summary

RCTDevMenu can get into a state where a UIAlertController is created but
never presented. When in this state, we will never be able to show the
dev menu unless we restart the app.

## Changelog

[iOS] [Fixed] - Fix dev menu sometimes not showing

## Test Plan

This is a bit difficult to test as I cannot consistently get the
module into this state. It seems to trigger if one randomly shakes the
device while navigating.

### DIFF
--- a/React/CoreModules/RCTDevMenu.mm
+++ b/React/CoreModules/RCTDevMenu.mm
@@ -183,7 +183,7 @@ RCT_EXPORT_MODULE()
 
 - (BOOL)isActionSheetShown
 {
-  return _actionSheet != nil;
+  return _actionSheet.beingPresented;
 }
 
 - (void)addItem:(NSString *)title handler:(void (^)(void))handler
@@ -418,7 +418,7 @@ RCT_EXPORT_MODULE()
 
 RCT_EXPORT_METHOD(show)
 {
-  if (_actionSheet || !_bridge || RCTRunningInAppExtension()) {
+  if ([self isActionSheetShown] || !_bridge || RCTRunningInAppExtension()) {
     return;
   }
 
--- a/React/CoreModules/RCTDevMenu.mm
+++ b/React/CoreModules/RCTDevMenu.mm
@@ -149,6 +149,11 @@ RCT_EXPORT_MODULE()
   return self;
 }
 
+- (UIAlertController *)alertController
+{
+  return _actionSheet;
+}
+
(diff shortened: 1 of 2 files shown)


tido64:
cc @hramos

hramos:
A unit test failed while trying to land this:

```
	RNTesterUnitTests:
		-[RCTDevMenuTests testShowCreatingActionSheet]
```

I can now see that this test also failed here on Circle CI, in the `test_ios` job: https://circleci.com/gh/facebook/react-native/130989?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

tido64:
Thanks, @hramos. Also discovered that `-testClosingActionSheetAfterAction` was always succeeding because it's looping over an empty array 😄

hramos:
This is failing a Unit Test, `testViewManagerNotInitializedBeforeSetBridgeModule`.

tido64:
> @hramos: This is failing a Unit Test, `testViewManagerNotInitializedBeforeSetBridgeModule`.

I don't think that's caused by this change. It passes locally after I rebased on latest `master`.

hramos:
This one is blocked internally because of this failing unit test:

```
Failing tests:
	RNTesterUnitTests:
		-[RCTModuleInitTests testCustomInitModuleInitializedAtBridgeStartup]
```

I've rebased on top of master, but the failure is still there. I don't see why this would fail with these changes. I'll need to investigate a bit more.

tido64:
> This one is blocked internally because of this failing unit test:
> 
> ```
> Failing tests:
> 	RNTesterUnitTests:
> 		-[RCTModuleInitTests testCustomInitModuleInitializedAtBridgeStartup]
> ```
> 
> I've rebased on top of master, but the failure is still there. I don't see why this would fail with these changes. I'll need to investigate a bit more.

@hramos: This is a different test from before. The previous one was `testViewManagerNotInitializedBeforeSetBridgeModule`.

tido64:
@hramos: I'm trying to revive this PR and I still can't figure out why `-[RCTModuleInitTests testCustomInitModuleInitializedAtBridgeStartup]` fails. I can't reproduce this locally, and I keep seeing other randomly tests fail. They just all seem very flaky to me 😞

Have you had any luck with this?

Issue #27565, opened 2019-12-18
Flipper for React Native 0.62 


rickhanlonii:
# Overview

This issue is to track items to get Flipper enabled for 0.62.

## RNTester Integration

- [x] Android works with Flipper in RNTester
  - This should be working as of #27482
  - May also need https://github.com/facebook/react-native/pull/27631
- [x] iOS works with Flipper in RNTester
  - Requires https://github.com/facebook/react-native/pull/27762

## Default Template Integration

- [x] Android - Can install app from template and use Flipper by default.
  - Requires https://github.com/facebook/react-native/pull/27569
- [x] iOS - Can install app from template and use Flipper by default
  - Requires https://github.com/facebook/react-native/pull/27569

## Upgrade Flipper SDK Version
- [x] Android - Can use latest Flipper version
  - [FBJNI migration](https://github.com/facebook/react-native/pull/27729)
  - [BUCK fix](https://github.com/facebook/react-native/pull/27826)
  - [RNTester Upgrade](https://github.com/facebook/react-native/pull/27833)
  - [Templade upgrade](https://github.com/facebook/react-native/pull/27837) 
  - [Upgrade Flipper version in template](https://github.com/facebook/react-native/pull/27837)
- [X] iOS - Can use latest Flipper version
  - Should be done by #27426

## Upgrading
- [x] Address pod install speed with [openssl](https://github.com/facebook/react-native/issues/27565#issuecomment-567285529) Edit: addressed [here](https://github.com/facebook/react-native/issues/27565#issuecomment-567330597)
- [x] Address swift inclusion
  - See: https://github.com/facebook/react-native/pull/27426
   - Removed in https://github.com/facebook/react-native/commit/8f93dedc6a5653edd2220c65ccb4ff8736ee060c
- [ ] Clean up upgrade requirements
    - Using Flipper with default RN template
    - See https://github.com/facebook/react-native/issues/27565#issuecomment-580950480
- [x] Address React Devtools issue
  - React devtools plugin for Flipper is blank
- [ ] Make sure the upgrade tool works well for this release #27879
- [ ] Figure out use_framworks support

## Documentation

- [ ] Add React Native the Flipper website
- [ ] Add documentation to the website
    -  Adding Flipper to new projects (mention included by default after 0.62)
    -  Adding Flipper to existing projects
- [ ] In [Flipper Docs](https://fbflipper.com/docs/getting-started.html#setup-your-react-native-app)

## Commits to Cherry Pick
[...]

PR #27922, opened 2020-01-31
[Template] Remove empty Swift file for Flipper

alloy:
Related to #27426 & #27565.

## Summary

The wish to not include an empty Swift file to trigger the correct Xcode settings was voiced in #27426 & #27565.

## Changelog

[iOS] [Fixed] - Remove need for Swift file in the user’s project in order to use Flipper

## Test Plan

An application created with the template, like so:

```bash
react-native init --template=~/Code/React/react-native TestFlipper
```

…will successfully build, launch, and have Flipper connect.

### DIFF
--- a/template/ios/HelloWorld.xcodeproj/project.pbxproj
+++ b/template/ios/HelloWorld.xcodeproj/project.pbxproj
@@ -16,7 +16,6 @@
 		2D02E4BD1E0B4A84006451C7 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		2D02E4BF1E0B4AB3006451C7 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
 		2DCD954D1E0B4F2C00145EB5 /* HelloWorldTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 00E356F21AD99517003FC87E /* HelloWorldTests.m */; };
-		37E9F8DD2399A8DD00F82F3E /* Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37E9F8DC2399A8DD00F82F3E /* Swift.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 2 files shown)


template/ios/HelloWorld.xcodeproj/project.pbxproj
```diff
@@ -501,7 +495,6 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
-				DEAD_CODE_STRIPPING = NO;
```
alloy: This setting was introduced with [the initial import](https://github.com/facebook/react-native/commit/b74da13a75baa1345f72aebceee226f23b8ba9fe#diff-065bc63a1a0283f31962d4ab2316087bR655) of this project. The commit doesn’t provide any context as to whether this setting is needed for a FB/RN specific reason, but this is not the default setting in default Apple templates. (I tried this with both new Swift and ObjC apps.)

template/ios/HelloWorld.xcodeproj/project.pbxproj
```diff
@@ -690,6 +683,12 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				LD_RUNPATH_SEARCH_PATHS = "/usr/lib/swift $(inherited)";
```
charpeni: Thanks for looking into that!

So, by using those two lines, we don't have to "force" Swift? Interesting!

alloy: Yup! 👍 

(And the above dead code stripping, but that’s only because it was override from the default.)

friederbluemle:
Nice work @alloy 👍

buranmert:
hi @alloy and everyone 👋 
this fix doesn't work when a framework that contains Swift symbols without a Obj-C counterpart (eg: [`FileHandle.readToEnd()`](https://developer.apple.com/documentation/foundation/filehandle/3516318-readtoend)) is added to the ReactNative app target.

Repro steps:
1. Create a Swift framework/pod, use `FileHandle.readToEnd()` in it
2. Create a new Obj-C app project from Xcode (not necessarily a React Native app)
3. Add your pod to `Podfile` **without `use_frameworks!` flag**
4. Set your app target's build settings: 
```
LIBRARY_SEARCH_PATHS = (
  "\"$(TOOLCHAIN_DIR)/usr/lib/swift/$(PLATFORM_NAME)\"",
  "\"$(TOOLCHAIN_DIR)/usr/lib/swift-5.0/$(PLATFORM_NAME)\"",
  "\"$(inherited)\"",
);
```
Expected result: it should build
Actual result: it fails with `undefined symbol: readToEnd()` error ❌ When I remove this build setting and add a `Dummy.swift` then it works though.

Maybe it could help to know how this fix works in the first place?
Are symbols like `readToEnd()` in a different location than the rest of the `Foundation` symbols?

PR #27943, opened 2020-02-04
Capitalizing FormData header field name

iamdhj:
## Summary

In general, the header field name should be Capitalizing. 
Lowercase FormData field name will cause `Stream ended unexpectedly` in same server.

## Changelog

[General] [Fixed] Capitalizing FormData header field name

### DIFF
--- a/Libraries/Network/FormData.js
+++ b/Libraries/Network/FormData.js
@@ -68,7 +68,7 @@ class FormData {
     return this._parts.map(([name, value]) => {
       const contentDisposition = 'form-data; name="' + name + '"';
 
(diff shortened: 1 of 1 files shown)


shergin:
@iamdhj Could you please update failing tests?

TheSavior:
Thanks for the PR. However, http header names are case-insensitive per the spec. https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive

Your server should properly handle the spec and case-insensitive headers. If we changed this we would surely get people complaining about it not being the other way.

PR #27951, opened 2020-02-05
Make Vibration.vibrate compatible with TurboModules

brunobar79:
## Summary

This PR fixes a compatibility issue with the Vibration module and TurboModules.
The TurboModules spec doesn't allow nullable arguments of type Number, causing the following problem: 

IMG_3758

## Changelog

[iOS] [Fixed] - Make Vibration library compatible with TurboModules.

## Test Plan

Just submitted a PR to my own app to fix the issue [here](https://github.com/rainbow-me/rainbow/pull/340) 

The problem should be reproducible on RNTester due to this line: https://github.com/facebook/react-native/blob/91f139b94118fe8db29728ea8ad855fc4a13f743/RNTester/js/examples/Vibration/VibrationExample.js#L66  and should be working on this branch.

### DIFF
--- a/Libraries/Vibration/NativeVibration.js
+++ b/Libraries/Vibration/NativeVibration.js
@@ -15,7 +15,7 @@ import * as TurboModuleRegistry from '../TurboModule/TurboModuleRegistry';
 
 export interface Spec extends TurboModule {
   +getConstants: () => {||};
-  +vibrate: (pattern?: ?number) => void;
+  +vibrate: (pattern: number) => void;
 
   // Android only
   +vibrateByPattern: (pattern: Array<number>, repeat: number) => void;
--- a/Libraries/Vibration/RCTVibration.mm
+++ b/Libraries/Vibration/RCTVibration.mm
@@ -25,7 +25,7 @@ RCT_EXPORT_MODULE()
   AudioServicesPlaySystemSound(kSystemSoundID_Vibrate);
 }
 
-RCT_EXPORT_METHOD(vibrate:(NSNumber *)pattern)
+RCT_EXPORT_METHOD(vibrate:(nonnull NSNumber *)pattern)
 {
   [self vibrate];
 }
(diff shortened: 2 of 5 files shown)


hramos:
This is a great PR. Thanks for fixing this!

hramos:
Ran into some issues landing this, which only surfaced internally due to TurboModules not being fully supported in open source yet.

1. Had to regenerate the native Java spec for the Vibration TurboModule, since the JS spec changed
2. In turn, this surfaced the following error, related to the generated method signatures having changed on Android:

```
stderr: ReactAndroid/src/main/java/com/facebook/react/modules/vibration/VibrationModule.java:20: error: com.facebook.react.modules.vibration.VibrationModule is not abstract and does not override abstract method vibrate(double) in com.facebook.fbreact.specs.NativeVibrationSpec
public class VibrationModule extends NativeVibrationSpec {
       ^
ReactAndroid/src/main/java/com/facebook/react/modules/vibration/VibrationModule.java:33: error: method does not override or implement a method from a supertype
  @Override
  ^
```

Do you think you can fix that up in your PR?

brunobar79:
@hramos  Sure, np. I'll take a look some time this week.

alloy:
@brunobar79 bump, if you want this to make it into v0.62.0-rc.3, which we'll cut over the next few days.

brunobar79:
@hramos Can you give it a try now?

hramos:
Imported, but it ran into some issues. I'll look at this again sometime this week.

Frans-L:
@hramos The same issue exists on 0.62.0. Is there any schedule when this pr could be merged? :)

maxencehenneron:
Note: This error only exists on debug builds, it seems to work fine in release mode, for some reasons.. This PR fixes the issue.

hramos:
This is still on my todo list, and the past month kind of threw things off. I'll try to take another look soon.

hramos:
We're actively working on landing this change.

hramos:
This is landing now. Thanks to @TheSavior for helping push this through.

PR #27983, opened 2020-02-06
chore: add eslint back to template

Naturalclar:
## Summary

This PR is part of https://github.com/react-native-community/releases/issues/157#issuecomment-582504146
Adding eslint back to the template package.json

## Changelog

[Internal] [Fixed] - Add eslint back in template

## Test Plan

(edited)
- npx react-native init RN062 --version 0.62.0-rc.1
- yarn add -d eslint
- add npm script for lint
- yarn lint

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -6,7 +6,8 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "test": "jest"
+    "test": "jest",
+    "lint": "eslint ."
   },
   "dependencies": {
     "react": "16.12.0",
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for doing this!

As a heads up, the purpose of the Test Plan section is so that the reviewer knows what you did to validate that this change was correct. If you just brought back the lines that were there before, are you sure that these versions still work? Does ESLint pass on a newly `init`ed project still? 

FWIW, I didn't really get the purpose of it until I joined Facebook :)

Naturalclar:
@TheSavior 
Thanks for the information about the Test Plan! 🤗
I've written the process I've done to verify the change.
From my understanding, there are no change in the JS file from 0.62-rc.1 to the current 0.62-stable branch, so the lint check result should be the same.

TheSavior:
Great, thanks!

PR #27984, opened 2020-02-07
Use autolink script in template on iOS

janicduplessis:
## Summary

We use this for RN tester but it is not currently included in npm releases. It cleans up the podfile a lot and will make adding / removing pods easier in the future since it won't require users to update their project.

## Changelog

[iOS] [Added] - Use autolink script in template on iOS

## Test Plan

Not sure if there is a way to end to end test this but I've been using this autolink script in my fork for a while.

### DIFF
--- a/scripts/autolink-ios.rb
+++ b/scripts/autolink-ios.rb
@@ -43,7 +43,7 @@ def use_react_native! (options={})
   pod 'React-jsinspector', :path => "#{prefix}/ReactCommon/jsinspector"
   pod 'ReactCommon/callinvoker', :path => "#{prefix}/ReactCommon"
   pod 'ReactCommon/turbomodule/core', :path => "#{prefix}/ReactCommon"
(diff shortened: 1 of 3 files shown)


scripts/autolink-ios.rb
```diff
@@ -43,7 +43,7 @@ def use_react_native! (options={})
   pod 'React-jsinspector', :path => "#{prefix}/ReactCommon/jsinspector"
   pod 'ReactCommon/callinvoker', :path => "#{prefix}/ReactCommon"
   pod 'ReactCommon/turbomodule/core', :path => "#{prefix}/ReactCommon"
-  pod 'Yoga', :path => "#{prefix}/ReactCommon/yoga"
+  pod 'Yoga', :path => "#{prefix}/ReactCommon/yoga", :modular_headers => true
```
janicduplessis: Updated this to match what was in the template before.

PR #27995, opened 2020-02-07
Add localized accessibility strings to React Core pod

xuelgong:
## Summary

The accessibility roles and states description strings are not able to be localized on iOS platform. Those strings are exposed to the end users so it should be localized. This PR is to add localized strings as a resource bundle to the React Core Pod so that any React Native app integrating the React Native dependencies using CocoaPods can get the localized accessibility roles and states description.

## Changelog

[iOS] [Added] - Add localized accessibility strings to React Core pod

## Test Plan

Verified with RNTester app.

### DIFF
--- a/React/Views/RCTView.m
+++ b/React/Views/RCTView.m
@@ -211,7 +211,12 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:unused)
 
     if (bundle) {
       NSURL *url = [bundle URLForResource:@"Localizable" withExtension:@"strings"];
-      rolesAndStatesDescription = [NSDictionary dictionaryWithContentsOfURL:url];
+      if (@available(iOS 11.0, *)) {
+        rolesAndStatesDescription = [NSDictionary dictionaryWithContentsOfURL:url error:nil];
+      } else {
+        // Fallback on earlier versions
+        rolesAndStatesDescription = [NSDictionary dictionaryWithContentsOfURL:url];
+      }
     }
     if (rolesAndStatesDescription == nil) {
       NSLog(@"Cannot load localized accessibility strings.");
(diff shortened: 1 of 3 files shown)


React/Views/RCTView.m
```diff
@@ -202,6 +202,46 @@ - (BOOL)didActivateAccessibilityCustomAction:(UIAccessibilityCustomAction *)acti
...
+  static dispatch_once_t onceToken;
+  static NSDictionary<NSString *, NSString *> *rolesAndStatesDescription = nil;
+
+  dispatch_once(&onceToken, ^{
+    NSString *bundlePath = [[NSBundle mainBundle]pathForResource:@"AccessibilityResources" ofType:@"bundle"];
+    NSBundle *bundle = [NSBundle bundleWithPath:bundlePath];
+
+    if (bundle) {
+      NSURL *url = [bundle URLForResource:@"Localizable" withExtension:@"strings"];
+      rolesAndStatesDescription = [NSDictionary dictionaryWithContentsOfURL:url];
```
PeteTheHeat: It looks like this method is deprecated according to https://developer.apple.com/documentation/foundation/nsdictionary/1574185-dictionarywithcontentsofurl

Can you update to use the suggested one?

Also, what is the flow for when the strings don't exist for a language? I imagine the `AccessibilityResources` bundle will exist, but the url will not. That `NSDictionary` method is safe to use with a nil url, correct?

PeteTheHeat:
Also, don't worry about the failing tests, the failures seem unrelated.

React/Views/RCTView.m
xuelgong: I'll update the method.
When the strings don't exist for a language. It will fallback to the English strings in the AccessibilityResources.bundle. The url is the one for the Localizable.strings in en.lproj. Also, if something bad happens and results in a nil url, the NSDictionary method is still safe to use.

xuelgong: We'll leave the localization for the other languages to the community. The accessibility strings on the Android side have been marked as localized for some reason. https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/res/views/uimanager/values/strings_unlocalized.xml
So I think it's ok to only have the english strings for iOS for now.

PR #27996, opened 2020-02-07
[GitHub] Issues without any template will get closed

hramos:
## Summary

Update the default issue template (e.g. the one shown when no custom template is chosen at https://github.com/facebook/react-native/issues/new/choose, and instead a new issue is opened via https://github.com/facebook/react-native/issues/new) to clarify that moving forward without a template will result in the issue getting closed.

The bot will aggressively close issues that have no labels, and only using a custom template will guarantee your issue has a label.

## Changelog

[Internal] - GitHub-only change.

## Test Plan

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
-👉 Please follow one of these issue templates:
-- https://github.com/facebook/react-native/issues/new/choose
+✋ To keep the backlog clean and actionable, issues will be
+🚫 closed if they do not follow one of the issue templates:
+👉 https://github.com/facebook/react-native/issues/new/choose
 
-Note: to keep the backlog clean and actionable, issues may be immediately closed if they do not follow one of the above issue templates.


hramos:
Will land this as is, but I may come back to this next week with some adjustments:

* Modify @react-native-bot so it nags with "Needs: Issue Template" instead of closing. It's not ideal, because I've found that people rarely go back and edit the original issue when they skip using a template altogether. It's a high friction process to go hunt for the template. But closing their issue is also going to cause them a lot of inconvenience.

* Reword the basic no-template message to be something like this, so that contributors are aware how they can opt-out of getting their issues closed:

```
To keep the backlog clean and actionable, issues may be closed if they do not follow one of the issue templates:
👉 https://github.com/facebook/react-native/issues/new/choose

> Note to contributors: 
> You may create an issue without the template, but please self-triage by adding a label to your issue.
```

...the idea being that if we're creating issues ourselves, we should at least put them in the right issue bucket.

PR #27999, opened 2020-02-08
fix: prop name for passwordRules

Naturalclar:
## Summary

There was a typo in passwordRules prop for TextInput.

ref: 
https://github.com/facebook/react-native/blob/master/ReactCommon/fabric/components/textinput/iostextinput/propsConversions.h#L106-L110
https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/RCTSinglelineTextInputViewConfig.js#L126

## Changelog

[iOS] [Fixed] - Fix prop name of passwordRules in TextInput

## Test Plan

run `flow` type check with TextInput that has `passwordRules` in the prop.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -267,7 +267,13 @@ type IOSProps = $ReadOnly<{|
    */
   textContentType?: ?TextContentType,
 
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for adding the documentation too

PR #28036, opened 2020-02-12
Stamp React Native Version Into C++ Code

NickGerleman:
## Summary

The PlatformConstants native module exposes the ability to query the React
Native version used to build native code. This is managed on iOS and Android by
a version bumping script, which replaces module code based on a template.

It is currently difficult to accurately determine this version for out-of-tree C++
platforms (I.e. React Native Windows). The version of upstream react-native we resolve to is ultimately
dependent on the version of react-native chosen a peer dependency, which is not
neccesarily constant given a build of react-native-windows.

We could try to hack around this, and make our native build try to reason about
the resolved pacakge for react-native using a lockfile, but a much cleaner
solution is to embed version into C++ code, similar to what is done for Android
and iOS. This change does that, adding a header with React Native version and
updating the build stamping script to write to it.

Usage sample:

```c++
  constants["reactNativeVersion"] = folly::dynamic::object();
  constants["reactNativeVersion"]["major"] = ReactNativeVersion.Major;
  constants["reactNativeVersion"]["minor"] = ReactNativeVersion.Minor;
  constants["reactNativeVersion"]["patch"] = ReactNativeVersion.Patch;
```
## Changelog

[General] [Added] - Stamp React Native Version Into C++ Code

## Test Plan

Validated that the bumping script will accurately update the header, can compile under both MSVC and Clang.


NickGerleman:
@hramos do you know what failed for Facebook Internal CI?

hramos:
@NickGerleman this patch no longer applies on top of master. Looks like `ReactCommon/cxxreact/ReactNativeVersion.h` no longer exists.

hramos:
Oh, my bad, that file is added in this PR. In any case, we failed to apply the patch internally, usually it's either due to a file that moved or because the PR is based off an older revision and the process timed out (can happen if the base commit is over 3 months old). I'm applying the changes manually on top of HEAD and will push them out. Hold tight.

PR #28050, opened 2020-02-13
Address @octokit/rest deprecation warnings

tido64:
## Summary

Addresses deprecation warnings from `@octokit/rest`. This is a follow up to #28019.

## Changelog

[Internal] [Fixed] - Address deprecation warnings from `@octokit/rest`

## Test Plan

PRs should still get app bundle sizes report, but the warnings in the build logs should be gone.

### DIFF
--- a/bots/code-analysis-bot.js
+++ b/bots/code-analysis-bot.js
@@ -18,9 +18,6 @@ if (!process.env.GITHUB_REPO) {
   process.exit(1);
 }
 
-// https://octokit.github.io/rest.js/
-const octokit = require('@octokit/rest')();
-
 const path = require('path');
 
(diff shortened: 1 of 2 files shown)


tido64:
@hramos Will this also address #28043?

tido64:
No more warnings:
image

tido64:
@hramos It looks like `test_ios_e2e` didn't get triggered?

hramos:
It won’t run if test_js fails. I’m looking at that failure today.

hramos:
BTW, https://github.com/facebook/react-native/pull/28048 fixes test_js, next up I need to see if the test_js_e2e failure already exists in master or not. I'm following up on this later today.

Issue #28060, opened 2020-02-14
Image component crashes on iOS when uri is null


mlazari:
Description:

Rendering an `Image` with `source={{ uri: null }}` crashes on iOS with a `JSON value '<null>' of type NSNull cannot be converted to NSString` error. It seems to have started doing so in version `0.61.0` after this commit: https://github.com/facebook/react-native/commit/df17ebfcfaaf2cf5984e836bdb70ccf8d9f66307
Seems to be a regression of https://github.com/facebook/react-native/commit/f940e7c4a61e6c29d88cf01b91f8fff1afe68b85

React Native version:

```
info Fetching system and libraries information...
System:
    OS: macOS 10.15
    CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
    Memory: 285.80 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.15.1 - ~/.nvm/versions/node/v10.15.1/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.13.0 - ~/.nvm/versions/node/v10.15.1/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.2, 28.0.3
      System Images: android-27 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.6010548
    Xcode: 11.0/11A420a - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.10.2 => 16.12.0 
    react-native: ^0.61.5 => 0.61.5
```

## Steps To Reproduce

1. Create a simple app that renders `<Image source={{ uri: null }} />`
2. Run the app on iOS
3. The app crashes with a `JSON value '<null>' of type NSNull cannot be converted to NSString` error.

## Expected Results

    The app should not crash.

## Snack, code example, screenshot, or link to a repository:

    <Image source={{ uri: null }} />
![error](https://user-images.githubusercontent.com/4928274/74491245-bf228d00-4ed3-11ea-810c-fb3a2aa643d7.png)

PR #28061, opened 2020-02-14
Fix Image component crashing when uri is null

mlazari:
## Summary

This fixes https://github.com/facebook/react-native/issues/28060 with a fix similar to https://github.com/facebook/react-native/commit/f940e7c4a61e6c29d88cf01b91f8fff1afe68b85

## Changelog

[iOS] [Fixed] - Fix Image component crashing when uri is null

## Test Plan

Run an app on iOS that renders `<Image source={{ uri: null }} />`. It should not crash.
error

### DIFF
--- a/React/Base/RCTConvert.m
+++ b/React/Base/RCTConvert.m
@@ -146,7 +146,7 @@ RCT_ENUM_CONVERTER(NSURLRequestCachePolicy, (@{
       URLString = [NSString stringWithFormat:@"%@.bundle/%@", bundleName, URLString];
     }
 
-    URL = [self NSURL:URLString];
+    URL = [self NSURL:RCTNilIfNull(URLString)];
     if (!URL) {
       return nil;
     }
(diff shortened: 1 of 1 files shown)


React/Base/RCTConvert.m
```diff
@@ -146,7 +146,7 @@ + (NSURLRequest *)NSURLRequest:(id)json
       URLString = [NSString stringWithFormat:@"%@.bundle/%@", bundleName, URLString];
     }
 
-    URL = [self NSURL:URLString];
+    URL = [self NSURL:RCTNilIfNull(URLString)];
```
sammy-SC: `[self NSURL]` is called from multiple places, wouldn't it be better to handle `NSNULL` inside of `NSURL`?

mlazari: Yes, I think you're right. Moved the fix in `NSURL:`.

PR #28067, opened 2020-02-14
Add window to jest setup

timomeh:
## Summary

`window` exists in the React Native runtime, but not inside the test environment. Many libraries use `typeof window === 'undefined'` to check if code is running in SSR. Because of the difference in the real environment and test environment, tests can behave different than the real app, which is an unwanted behavior.

## Background

I'm using https://github.com/tannerlinsley/react-query in my React Native Project, which works really well. When writing tests, they wouldn't work: jest started and then seemingly did nothing. While debugging I noticed the render was stuck in an infinite loop. Then I noticed the following line inside `react-query`:

```js
const isServer = typeof window === 'undefined'
```

I didn't know that the React Native runtime has a global `window`, and thought it's a bug inside react-query. But it does have a `window`, which is not defined inside the test environment.

The infinite loop was caused by react-query thinking it is running on the server, which doesn't fetch any data. If the react-query hook mounts, it re-executes because then it should be mounted inside the client. But `isServer` was still `true`. This repeats forever.

## Changelog

[General] [Fixed] - Fix `window` not existing in jest setup

## Test Plan

Are there tests to check if the test environment is setup correctly? 🤔

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -23,6 +23,7 @@ global.performance = {
 
 global.Promise = jest.requireActual('promise');
 global.regeneratorRuntime = jest.requireActual('regenerator-runtime/runtime');
+global.window = global;
 
 global.requestAnimationFrame = function(callback) {
   return setTimeout(callback, 0);


charlesbdudley:
Thanks for this @timomeh, do you mind rebasing to get these tests to pass and we'll import for review?

timomeh:
@charlesbdudley Done! The test_ios_unit_hermes check is failing, but it also fails on the main branch.

Issue #27463, opened 2019-12-10
[RCTWeakProxy displayDidRefresh:]: unrecognized selector sent to instance 0x28048f800


DongDongDongDong:
small probability of the following error：

-[RCTWeakProxy displayDidRefresh:]: unrecognized selector sent to instance 0x28048f800
NSInvalidArgumentException

React Native version:0.61

DongDongDongDong:
Has anyone encountered a similar situation?

letscms:
Same issue with me.
Anyone could you please provide solution of this

letscms:
![react-issue](https://user-images.githubusercontent.com/29188998/70522302-e15ffd80-1b66-11ea-83e5-2c8e3a3a206e.png)

letscms:
Please help of this issue

ovy9086:
same here. started seeing this since 0.61 in our user crashlogs 🤕

ovy9086:
here's the full stacktrace 
```NSInvalidArgumentException -[RCTWeakProxy displayDidRefresh:]: unrecognized selector sent to instance 0x2827a68c0 
    Frameworks/CoreFoundation.framework/CoreFoundation ___exceptionPreprocess
    /usr/lib/libobjc.A.dylib _objc_exception_throw
    Frameworks/CoreFoundation.framework/CoreFoundation -[NSObject(NSObject) doesNotRecognizeSelector:]
    Frameworks/CoreFoundation.framework/CoreFoundation ____forwarding___
    Frameworks/CoreFoundation.framework/CoreFoundation ___forwarding_prep_0___
    Frameworks/QuartzCore.framework/QuartzCore CA::Display::DisplayLink::dispatch_items(unsigned long long, unsigned long long, unsigned long long)
    Frameworks/IOKit.framework/Versions/A/IOKit _IODispatchCalloutFromCFMessage
    Frameworks/CoreFoundation.framework/CoreFoundation ___CFMachPortPerform
    Frameworks/CoreFoundation.framework/CoreFoundation ___CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__
    Frameworks/CoreFoundation.framework/CoreFoundation ___CFRunLoopDoSource1
    Frameworks/CoreFoundation.framework/CoreFoundation ___CFRunLoopRun
    Frameworks/CoreFoundation.framework/CoreFoundation _CFRunLoopRunSpecific
[...]

PR #28070, opened 2020-02-14
fix [RCTWeakProxy displayDidRefresh:] crash

zlpcloud:
## Summary
Fix small probability crash: -[RCTWeakProxy displayDidRefresh:]: unrecognized selector sent to instance.   
#27463 can be fixed then. 

## Changelog

[iOS] [Fixed] - Fix crash -[RCTWeakProxy displayDidRefresh:]

## Test Plan

This crash occurred randomly in our app of previous version. This crash has not occurred again after this commit.

### DIFF
--- a/React/Base/RCTWeakProxy.m
+++ b/React/Base/RCTWeakProxy.m
@@ -27,4 +27,16 @@
   return _target;
 }
 
+- (void)forwardInvocation:(NSInvocation *)invocation
+{
+    void *nullPointer = NULL;
+    [invocation setReturnValue:&nullPointer];
+}
+
+
+- (NSMethodSignature *)methodSignatureForSelector:(SEL)selector
+{
+    return [NSObject instanceMethodSignatureForSelector:@selector(init)];
+}
+
 @end


alicancelik:
please merge it :D @shergin

alicancelik:
is there any update ?

shergin:
@alicancelik Why do you think it fixes the issue?

alicancelik:
@shergin we got some errors without this fixes, after that we forked your solution to our app which is in live, it fixed

zlpcloud:
> @zipclould Thank you for the PR! ❤️
> _Why_ do you think it works? Why the implementation of new methods does the right thing? Why does the crash is happening and how does this change fixes that?
> 
> I would recommend taking look at how `RCTGenericDelegateSplitter` is implemented and borrow some ideas from there.

The crash is happening when `forwardingTargetForSelector:` returns nil. In that case, the weak target has been reclaimed. So we return a dummy method signature in `methodSignatureForSelector:` and just return NULL/nil in `forwardInvocation:` to keep `doesNotRecognizeSelector:` from firing.

shergin:
@zlpcloud This is still a very over-complicated solution. Even if it works, I don't think it would be easy to find someone who would take responsibility for landing that.

Here is why it's bad concretely:
* It uses runtime introspection without actual need;
* It uses runtime introspection (which is slower than regular code) in a very perf-critical part (DisplayLink connection);
* Seems there is a problem when weak pointers meet the introspection.

So, here is my proposal: We just need to replace `RCTWeakProxy` with a concrete trivial class that has a an ivar, a method that calls the same method on the ivar, and use that in Image component instead of `RCTWeakProxy`.

It's trivial to implement, trivial to review, it will benefit the performance and maintainability. Profit.

dephiros:
Maybe we can close this issue since the fix is already in `master` through https://github.com/facebook/react-native/commit/e5a6655e71d41a58ce0e51d37aa9fb8792e37dd5 ?

Issue #28103, opened 2020-02-18
Migrate off of old context APIs


TheSavior:
We want to get rid of our usage of the `prop-types` package. The last remaining usage of this is a couple of old components that use [the old React context API](https://reactjs.org/docs/context.html#classcontexttype). We need to migrate these components to use the modern `React.createContext` API.

The components using the old context API are:
- ~~[VirtualizedList](https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js#L1979)~~
- ~~[Modal](https://github.com/facebook/react-native/blob/master/Libraries/Modal/Modal.js#L162-L166) (@satya164)~~
- [AppContainer](https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/AppContainer.js#L49-L59) (@safaiyeh)
- ~~[Incremental](https://github.com/facebook/react-native/blob/master/Libraries/Experimental/Incremental.js#L109-L117) (@venits)~~

For more information about the old API and how to migrate to the new API you can check out the React docs for Context: https://reactjs.org/docs/context.html#classcontexttype

If we are able to remove the usage of prop-types from React Native we should be able to substantially lower the bundle size of React Native. We'd love your help!

Naturalclar:
I'd love to help out! Could I work on `Modal`?

venits:
I would love to take care of `Incremental` :)

mikaoelitiana:
hey there, I can check virtualizedlist

thymikee:
@Naturalclar @mikaoelitiana they're all yours to take :)

mikaoelitiana:
@thymikee thanks, I am already working on `VirtualizedList`, the module is quite big and I try to make sure I am doing things the right way

TheSavior:
It looks like Modal actually only uses the VirtualizedList Context and AppContainer Context, it doesn't have its own context. So we need to get the VirtualizedList and AppContainer ones done first.

mikaoelitiana:
So this means these contexts need to be exported, right?

TheSavior:
Yeah, the root tag context is already exported, unstable_RootTagContext, the VirtualizedList one can be exported as a static property on VirtualizedList: VirtualizedList.Context

thymikee:
[...]

PR #28104, opened 2020-02-18
Migrate AppContainer to new Context API

safaiyeh:
## Summary
1/4 of https://github.com/facebook/react-native/issues/28103

## Changelog

[General] [Fixed] - Migrated AppContainer to new Context API.

## Test Plan

No flow errors, built RNTester successfully.


TheSavior:
Thanks for working on this!

It looks like it is pretty common to consume this rootTag context downstream. Class components access this.context.rootTag to pass to native modules for example. 

I think this would be a fairly substantial breaking change as is, do you have any ideas of how to make this backwards compatible and make it easy to migrate?

For an example, take a look at Modal.js, I think one of the usages of context there is to receive this context type. 

I'd love to see your test plan include the RNTester examples for Modal. 
Ideally, first without any changes to the JS code and a warning being printed recommending to use the new way to consume this API, and second changing the Modal code to consume the context in the new way. 

This will hopefully make sure the necessary pieces are exported for downstream consumers to use the context, and for it to be backwards compatible for now.

safaiyeh:
I see, thank you for the context @TheSavior! 

Not sure if there is a way to make it backwards compatible. With this change downstreams need to access the rootTag from the AppContainer's Consumer. Is a warning sufficient enough to address this breaking change?

TheSavior:
To make it backwards compatible could you leave the existing code:

```
static childContextTypes:
    | any
    | {|rootTag: React$PropType$Primitive<number>|} = {
    rootTag: PropTypes.number,
  };

  getChildContext(): Context {
    return {
      rootTag: this.props.rootTag,
    };
  }
```

but make `rootTag` in here a getter?
```
    return {
      rootTag: this.props.rootTag,
    };
```

And then *also* render `<Context.Provider value={{rootTag: this.props.rootTag}}>`?

safaiyeh:
Makes sense, since keeping the legacy Context API is not an issue, no need to break usage over it.

Just need to migrate the flow typing to not use PropTypes with this.

TheSavior:
I think it is fine to leave the usage of prop-types for now, as long as we know we are making progress towards removing it. I'm not sure if the React API requires the usage of prop-types for the old context API. If we can avoid it, great! Otherwise, I'm not so worried as long as we have a plan.

safaiyeh:
Great! Sounds like a plan I’ll make the changes & update the Test Plan.

Learned a new mindset here. Should be thinking about backwards capability with any migration.

safaiyeh:
I made the changes to warn with the getter.

Unfortunately, `getChildContext` gets called with any prop or state change: https://reactjs.org/docs/legacy-context.html#updating-context

So even using the new Context API, the warning still gets thrown. 

cc @TheSavior

Libraries/ReactNative/AppContainer.js
```diff
@@ -46,17 +48,23 @@ class AppContainer extends React.Component<Props, State> {
...
+     | {|rootTag: React$PropType$Primitive<number>|} = {		
+     rootTag: PropTypes.number,		
+   };		
+
+    getChildContext(): ContextType {
+     console.warn(
+        'AppContainer has been migrated to the new Context API. ' +
+        'It is recommended to use AppContainer.Context.Consumer to consume the context.'
+      );
+     return {
```
TheSavior: What if you make this an object with a getter for rootTag that includes the warning. So the warning would hopefully only fire when the rootTag is read, not on every call to this function.

Libraries/ReactNative/AppContainer.js
```diff
@@ -35,6 +33,10 @@ type State = {|
   hasError: boolean,
 |};
 
+type ContextType = {rootTag: number, ...};
```
TheSavior: You should remove the ... since we know this is exact now since it’s new

TheSavior: Also, should probably wrap the object in $ReadOnly<> for good measure, if flow is okay with it.

Libraries/ReactNative/AppContainer.js
```diff
@@ -11,15 +11,13 @@
 'use strict';
 
 const EmitterSubscription = require('../vendor/emitter/EmitterSubscription');
-const PropTypes = require('prop-types');
 const RCTDeviceEventEmitter = require('../EventEmitter/RCTDeviceEventEmitter');
+const PropTypes = require('prop-types');
 const React = require('react');
```
TheSavior: If you change this to 

Import * as React from ‘react’;

Then you can use React.Context below instead of React$Context.

Libraries/ReactNative/AppContainer.js
```diff
@@ -46,17 +48,23 @@ class AppContainer extends React.Component<Props, State> {
...
+  static childContextTypes:
+     | any		
+     | {|rootTag: React$PropType$Primitive<number>|} = {		
+     rootTag: PropTypes.number,		
+   };		
+
+    getChildContext(): ContextType {
+     console.warn(
+        'AppContainer has been migrated to the new Context API. ' +
+        'It is recommended to use AppContainer.Context.Consumer to consume the context.'
```
TheSavior: We can specifically mention rootTag since that’s the only thing in the context

Libraries/ReactNative/AppContainer.js
TheSavior: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

TheSavior:
Any progress on this one?

safaiyeh:
I'll finish this up tonight! @TheSavior

safaiyeh:
@TheSavior made the updates per code review.

With the getter it seems to still fire when the function is called.

The only other usage of `AppContainer` is in `renderApplication.js` which receives `rootTag` from the `renderApplication()` function which does not access through the Legacy API.
```jsx
<AppContainer
  rootTag={rootTag}
  fabric={fabric}
  showArchitectureIndicator={showArchitectureIndicator}
  WrapperComponent={WrapperComponent}
  internal_excludeLogBox={isLogBox}>
  <RootComponent {...initialProps} rootTag={rootTag} />
</AppContainer>
```

Is it warranted to introduce the breaking change given this?

TheSavior:
Let’s avoid the breaking change. We have a couple hundred callsites internally at FB that we will need to migrate so I imagine other people are using it too.

TheSavior:
*sigh*. It looks like the new context was already added in July of last year https://github.com/facebook/react-native/commit/f9ecce7440e712782f4fd760a46e54b3102914d8

i'm so sorry for not realizing this earlier. You have helped me learn about the core.

safaiyeh:
hmm I see, so Modal should be migrated to use the rootTag provider instead it being consumed from AppContainer, then remove the legacy context from AppContainer?

TheSavior:
Modal needs to use `RootTagContext`, but we still can't get rid of the legacy context api in AppContainer yet because we have hundreds of internal callsites still using it that need to get migrated to RootTagContext as well.

safaiyeh:
Got it, makes sense!

thymikee:
@safaiyeh still interested? :) Recently Modal was migrated in #29002

safaiyeh:
Alrighty, I’ll check it out! @thymikee

PR #28107, opened 2020-02-18
[JavaScript][Removed] - Remove leftover `Incremental` component

venits:
## Summary

1/4 of https://github.com/facebook/react-native/issues/28103  

## Changelog

[JavaScript][Removed] - Remove leftover `Incremental` component.

## Test Plan

In RNTester `Incremental` component is no longer used so I modified one of existing examples to use it.


Libraries/Experimental/Incremental.js
```diff
@@ -94,6 +94,11 @@ export type Props = {
 };
 type State = {doIncrementalRender: boolean, ...};
 
+const IncrementalContext = React.createContext<Context>({
```
thymikee: This component is not a context provider, but a consumer, so there's no need to create a new context here. 
@TheSavior I guess this component is used internally at Facebook or it's some dead code that haven't been removed yet. Can you shed some information on the topic? If it can be removed, that's great. If not, we need to get the context from somewhere.

venits:
In previous versions of **react-native** we had components like `IncrementalGroup.js`. Now with only `Incremental` component is hard to decide how it should behave.

Here is how it used to look like:
https://github.com/facebook/react-native/tree/2ee11711e0351f695ec1d5e591ac904e5c9187c4/Libraries/Experimental

thymikee:
Looks like a leftover from https://github.com/facebook/react-native/commit/0ef0d3167e291f31ce01ceb729df77cc679d2330. I assume you're safe to remove it

TheSavior:
Yeah, this is unused! Yay. Thanks for deleting.

janicduplessis:
I made a package with this if anyone is interested https://github.com/th3rdwave/react-native-incremental

PR #28113, opened 2020-02-19
[CI] Disable "Report size of RNTester.apk" on test_android

safaiyeh:
## Summary
test_android currently fails because it relies on a PR number.
Screen Shot 2020-02-18 at 5 39 27 PM

## Changelog

[Internal] [Fixed] - Disable "Report size of RNTester.apk" on test_android

## Test Plan

Should report APK size to this PR


safaiyeh:
Back to the drawing board 🙇🙇

PR #28115, opened 2020-02-19
Add feature: Modal custom background color

gabriel1536:
## Summary

This is some kind of syntactic sugar to avoid adding a view with some particular styling when trying to add some custom (non-white) overlay over the 'non-reached' parts of the modal.

Works when transparency={false}. So it replaces (`optional`) the hardcoded 'white' color for a custom valid color.

## Changelog

[General] [Added] - Add backgroundColor property on Modal component

## Test Plan

- [x] npm test
- [x] npm run flow
- [x] npm run lint

### DIFF
--- a/Libraries/Modal/__tests__/__snapshots__/Modal-test.js.snap
+++ b/Libraries/Modal/__tests__/__snapshots__/Modal-test.js.snap
@@ -31,7 +31,7 @@ exports[`<Modal /> should render as <RCTModalHostView> when not mocked 1`] = `
           "top": 0,
         },
         Object {
-          "backgroundColor": "white",
+          "backgroundColor": undefined,
         },
       ]
     }
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -211,7 +211,9 @@ class Modal extends React.Component<Props> {
     }
 
     const containerStyles = {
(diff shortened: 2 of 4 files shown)


TheSavior:
Can you provide an example of the current approach required to get this result? 

I'm not a big fan of having two props that are conditional on each other so I am curious how difficult the current approach is to be worth this change in API.

Libraries/Modal/__tests__/__snapshots__/Modal-test.js.snap
```diff
@@ -31,7 +31,7 @@ exports[`<Modal /> should render as <RCTModalHostView> when not mocked 1`] = `
           "top": 0,
         },
         Object {
-          "backgroundColor": "white",
+          "backgroundColor": undefined,
```
Patys: I think it's also worth to fix and tests for it 👍

robberfree:
@TheSavior 

The modal has different `backgroundColor ` (not only `transparent` or `white`) is very common in design. In our project we must build a custom modal component only because it has a different color(for example `rgba(0,0,0,0.4)`). 

1. add a new `backgroundColor` prop and keep `transparent` can give us a smooth path to use new prop.
2. we can keep `backgroundColor` prop only, but that's a big change maybe.

please have a look at this PR again.

thanks a lot.

PR #28130, opened 2020-02-19
Dark mode support for RCTPerfMonitor

gorhom:
## Summary

Hi There 👋,

While I'm developing on iOS with dark appearance enabled, i notice that `Pref Monitor` view doesn't support dark mode yet, so here is the PR to fix it :)  

## Changelog

[iOS] [Fixed] - Fix Pref Monitor in dark appearance 

## Test Plan
Run any React Native app on iOS > Turn on dark appearance
### Before
Before
### After
After

### DIFF
--- a/React/CoreModules/RCTFPSGraph.m
+++ b/React/CoreModules/RCTFPSGraph.m
@@ -79,6 +79,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithCoder:(NSCoder *)aDecoder)
     _label = [[UILabel alloc] initWithFrame:self.bounds];
     _label.font = [UIFont boldSystemFontOfSize:13];
     _label.textAlignment = NSTextAlignmentCenter;
+    _label.textColor = UIColor.blackColor;
   }
 
   return _label;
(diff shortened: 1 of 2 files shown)


gorhom:
hi @hramos , thanks for taking a time to review this.

by reverting my commits and add `systemBackgroundColor` to the root container 
```Objective-C
  _container.backgroundColor = UIColor.systemBackgroundColor;
}
```
i got this :
Dark

is this okay ?

gorhom:
hi @hramos , i have reverted my previous commits and only set the root container of `RCTPerfMonitor` background to `UIColor.systemBackgroundColor`, and this is how it looks in

**Light** 
light
**Dark**
dark

PeteTheHeat:
This looks good! Thanks for the PR.

PeteTheHeat:
Although this probably breaks anything pre iOS13. When I land it, I'll add another background color for that.

gorhom:
@PeteTheHeat thanks for taking a time to review this :)

about worrying about breaking pre iOS 13, i think it should be fine since wrapped the background colour with `available(iOS 13.0, *)` annotation if-condition. Also pre iOS 13 has only light mode.

PeteTheHeat:
But that line adding white background color was important. After this diff, on iOS 12, the background color will be transparent. I'm thinking:

` 
      _container.backgroundColor = UIColor.systemBackgroundColor;
    } else {
      _container.backgroundColor = UIColor.whiteColor;
    }`

gorhom:
@PeteTheHeat , ohh yeah sure i will push it right now

PR #28141, opened 2020-02-20
iOS: Add UIScene support to RCTImageView

tido64:
## Summary

Apps implementing `UISceneDelegate` no longer clear out images when backgrounded because `UIApplicationDidEnterBackgroundNotification` no longer gets fired.

## Changelog

[iOS] [Added] - UIScene support for RCTImageView

## Test Plan

1. Create a new iOS app implementing `UISceneDelegate` or modify an existing one
2. Open a React view with some images
3. Switch to another app, or background the current app
4. Observe that `-[RCTImageView clearImageIfDetached]` gets called

### DIFF
--- a/Libraries/Image/RCTImageView.mm
+++ b/Libraries/Image/RCTImageView.mm
@@ -86,6 +86,10 @@ static NSDictionary *onLoadParamsForSource(RCTImageSource *source)
 {
   if ((self = [super initWithFrame:CGRectZero])) {
     _bridge = bridge;
+    _imageView = [[RCTUIImageViewAnimated alloc] init];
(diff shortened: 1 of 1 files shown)


Libraries/Image/RCTImageView.mm
```diff
@@ -86,6 +86,10 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
 {
   if ((self = [super initWithFrame:CGRectZero])) {
     _bridge = bridge;
+    _imageView = [[RCTUIImageViewAnimated alloc] init];
+    _imageView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
AshishKapoor: What's happening on this line?

tido64: I don't know. I just moved some lines up to group things together.

PR #28149, opened 2020-02-21
Add Fragment support to Modal component on Android

grahammendick:
## Summary

Displayed the Modal by showing a `DialogFragment` instead of a `Dailog`. Made this Fragment available to Modal descendants so they can add their own View Fragments.

### Example Use-case - Native Stack Navigation

Native Navigation libraries use Fragments to manage the stack on Android. Adding Fragment support to the Modal component means you can render a navigation stack inside of a Modal. iOS already supports this use-case because the Modal component renders as a UIViewController. 

## Changelog

[Android] [Added] - Add Fragment support to Modal component

## Test Plan

After building React Native from my branch by following [the guidelines](https://github.com/facebook/react-native/wiki/Building-from-source), I ran the following tests:

1. I created a project with a Modal and checked that it opened and closed. I checked that the Fragment was removed when the Modal was closed.
2. I created a project with a Modal containing a native navigation stack from [the Navigation router](https://grahammendick.github.io/navigation/documentation/native/hello-world.html). I checked the scenes as I pushed them onto the stack and then popped them off again.
3. I created a project with a Modal opening from another Modal on Android. Each Modal contained a native navigation stack. I checked both Modals opened and closed and that the stacks operated independently.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
@@ -26,6 +26,8 @@ import android.view.ViewTreeObserver;
 import android.view.WindowManager;
 import android.widget.FrameLayout;
 import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+
 import com.facebook.common.logging.FLog;
 import com.facebook.infer.annotation.Assertions;
 import com.facebook.infer.annotation.ThreadConfined;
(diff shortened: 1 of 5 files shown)


grahammendick:
I worked out a way to do it. I created a containerless fragment that returns the native stack from `onCreateView`. Then the native stack uses the child fragment manager of that fragment instead of the fragment activity

PR #28160, opened 2020-02-23
Move App and test to src folder in template

codler:
## Summary
Make it more similar and familiar file structure as Create React App project, https://create-react-app.dev/docs/folder-structure

## Changelog

[General] [Changed] - Move App.js and App-test.js to src folder in template


TheSavior:
This change is reasonable, however I think it is likely to make upgrades harder and confusing since these files are moving.

Since we are currently optimizing for making upgrades easier I think I am going to avoid merging this.

codler:
huh? the upgrade are still a manual process, so its optional for the upgrader to move the test file.

TheSavior:
But it will show up as a big change in the upgrade helper. We try to avoid optional changes in the upgrade. If users don’t make that change then when there are additional changes to those files in future releases it will be difficult to see the diff and know what to change because there project no longer matches. Eventually it will get confusing enough that they’ll feel the need to create a new project and copy their files over. We want to avoid that.

Also, if they use the automatic upgrade tool it will create new files in their project that would be confusing.

I think someone needs to think through what the upgrade tooling experience is like for file moves before we should make a change like this. Maybe you’d like to work with @pvinis on that.

codler:
It wont look like a big change, it will just be a git diff, exact same as changes as you see here https://github.com/facebook/react-native/pull/28160/files
and from my experience React native have always had some changes in every release so i dont see any difference with changes now. Better now than later.

pvinis:
Just throwing my opinion out here: I think it's fine to do this move from the upgrade side. I agree that upgrades should be kept small, until we have a better way to handle the native parts upgrading, as they are the ones that cause the "fatigue" for users. At the same time, if the files have no changes when moving, the diff should be easy, and most probably we won't have many changes soon in the moves files.

If the decision was completely up to me, and if I was the only user, I would merge this, just because it's small and simple. **But** I would probably not merge it for all the other users for the specific reasons: It's one file, no big deal. By now the `src` is kinda "standard", so a lot of users would start like that right after init. Some people use `app` instead of `src`, but I wouldn't care about that, nothing breaks. Lastly, if users have already a `src/App`, they might be confused.

Basically, for me, it comes down to this. I like this change. I want this change. I am sure many people would understand and care/not care about it, but a lot of people might not understand. And while normally this is not a thing a smaller project should care about, I feel like fb is not too happy with that. That's what I feel from the outside, and currently.

I hope this makes sense, and I hope this is helpful. 🤷‍♂️

PR #28168, opened 2020-02-24
Change iOS default simulator to iPhone 11 Pro Max

codler:
## Summary

Changing from iPhone 11 to iPhone 11 Pro Max due to screen resolution. When uploading screenshots to app store you are required to open iPhone 11 Pro Max and take screenshots. So it is better to use iPhone 11 Pro Max from beginning.

## Changelog

[General] [Changed] - Change iOS default simulator to iPhone 11 Pro Max

## Test Plan

Run `npm run ios` and it should open iPhone 11 Pro Max

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "android": "react-native run-android",
-    "ios": "react-native run-ios",
+    "ios": "react-native run-ios --simulator \"iPhone 11 Pro Max\"",
     "start": "react-native start",
     "test": "jest",
     "lint": "eslint ."


template/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "android": "react-native run-android",
-    "ios": "react-native run-ios",
+    "ios": "react-native run-ios --simulator \"iPhone 11 Pro Max\"",
```
TheSavior: It seems like run-ios internally should use this simulator as a default if this is desired. Putting it here seems like it is just a hack around the symptom. Also, since it is changing the template it will show up for everyone when they are upgrading.

todorone:
This PR should belong to cli repo, otherwise, it's a dirty cli hack... 😸

codler:
Do you guys agree that cli should change simulator? I tried to submit a PR to CLI repo also.

TheSavior:
If this change is valid and should be changed for the entire ecosystem then it would belong in the CLI repo. If the CLI maintainers have concerns, they are probably valid and it wouldn't make sense to go around their concerns by putting it here.

PR #28174, opened 2020-02-25
Update engines to node >=10

safaiyeh:
## Summary
v8 dropped out of LTS support, updating the package.json engines to node v10.

## Changelog

[General] [Changed] - Bump supported Node engines to >= 10


tido64:
I think this is now failing builds:
image

Issue #28065, opened 2020-02-14
React Native Picker component does not have an onTouch property


donatas-vindzigalskis:
Description:

The current picker does not have any event listeners for onBlur, onTouch events. So it is impossible in the app to know if the user has interacted with the picker component. The only way to know if a user has made contact with the component is through the onValueChange property. However, if the user opens the picker on Android, but doesn't select a value and/or closes the picker, the app cannot know if the user has interacted with the picker. 

This is needed for form validations, to display warnings or error messages after the user has "touched" the picker component.

React Native version:
System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 6.30 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.13.1 - /usr/local/bin/node
    npm: 6.13.4 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.6010548
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: https://github.com/expo/react-native/archive/sdk-36.0.0.tar.gz => 0.61.4 
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    react-native: 0.61.5

## Steps To Reproduce

1. Import Picker component from react-native library,
2. Check the props of Picker component.

## Expected Results

    There should be a listener event of touch gestures or focus changes.

b0nbon1:
I'd love to work on this one as my first contribution to react-native.

maschad:
Hey @b0nbon1 I have actually submitted a PR [here](https://github.com/facebook/react-native/pull/28180) maybe you could take a look and let me know what you think :)

maschad:
Following @TheSavior 's [comment](https://github.com/facebook/react-native/pull/28180#issuecomment-593568371), this issue should be moved to React community's [Picker](https://github.com/react-native-community/react-native-picker/issues)

donatas-vindzigalskis:
@maschad I've created an [issue](https://github.com/react-native-community/react-native-picker/issues/54) on react community's picker repo

PR #28180, opened 2020-02-26
added onTouch event to picker

maschad:
## Summary

As mentioned in [https://github.com/facebook/react-native/issues/28065](url) , the current picker does not have any event listeners for onTouch events. So it is impossible in the app to know if the user has interacted with the picker component. This pull request uses a `SyntheticEvent` type to return a boolean i.e. touched is true or false.

## Changelog

[Android] [Added]

## Test Plan

Check the Props of the Picker component.


TheSavior:
Thanks for your PR! Picker has been pulled out of core and now lives in a community repo. [react-native-community/picker](https://github.com/react-native-community/react-native-picker#readme)

You should make a PR to that repo instead

maschad:
Thanks @TheSavior

PR #28183, opened 2020-02-27
Add support for uppercase named colors

usrbowe:
## Summary

Sometimes by accident, we can specify color in uppercase: `RED, WHITE`, which is currently not supported by React Native. So basically this small change will add support for those. And it also goes along with the web spec here: https://www.w3.org/TR/css-color-3/#html4
**TLDR**: `The color names are case-insensitive.`

## Changelog

[General] [Added] - Support uppercase named colors (RED, WHITE,...)

## Test Plan

Simply test with: `<Text style={{ color: 'RED' }}>I'm red!</Text>`

### DIFF
--- a/Libraries/StyleSheet/normalizeColor.js
+++ b/Libraries/StyleSheet/normalizeColor.js
@@ -28,8 +28,8 @@ function normalizeColor(color: string | number): ?number {
     return parseInt(match[1] + 'ff', 16) >>> 0;
   }
 
-  if (names.hasOwnProperty(color)) {
-    return names[color];
+  if (names.hasOwnProperty(color.toLowerCase())) {
+    return names[color.toLowerCase()];
   }
 
   if ((match = matchers.rgb.exec(color))) {


TheSavior:
While this change would make it consistent with the web spec, I'm not excited about getting less strict here. If we were parsing css files directly then it would be more compelling, but the only way I can see people using the upper case versions in RN code would be as a typo. I'd rather not make things more loose to support typos.

usrbowe:
@TheSavior Yeah, make sense from this point of view. I don't consider this myself as big deal. Maybe at least we can change the documentation, so it doesn't says:
>You can also use color names as values. React Native follows the [CSS3 specification](https://www.w3.org/TR/css3-color/#svg-color):
[Source: Color Reference - React Native](https://reactnative.dev/docs/colors#named-colors)

- Another way to prevent confusion, for developers from web, would be to add some Yellow.warning?

TheSavior:
Can you make a PR to the docs? Maybe something like “unlike on the web, named colors must be lower case”.

usrbowe:
@TheSavior alright, here is the PR: https://github.com/facebook/react-native-website/pull/1696

PR #28193, opened 2020-02-27
Modal: replace childContext with new Context api

Naturalclar:
## Summary

This PR is part of #28103 

Replaced getChildContext in the Modal component with `createContext` and get rid of the `prop-types` usage.

## Changelog

[General] [Changed] - Migrated Modal to use the new Context API

## Test Plan
- Flow check passes.
- RNTest built successfully.
- Having nested modal in RNTester App doesn't cause an error

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -12,12 +12,11 @@
 
 const AppContainer = require('../ReactNative/AppContainer');
 const I18nManager = require('../ReactNative/I18nManager');
-const PropTypes = require('prop-types');
-const React = require('react');
 const ScrollView = require('../Components/ScrollView/ScrollView');
 const StyleSheet = require('../StyleSheet/StyleSheet');
 const View = require('../Components/View/View');
 
+import * as React from 'react';
 import type {ViewProps} from '../Components/View/ViewPropTypes';
 import type {DirectEventHandler} from '../Types/CodegenTypes';
 import type EmitterSubscription from '../vendor/emitter/EmitterSubscription';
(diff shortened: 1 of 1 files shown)


Libraries/Modal/Modal.js
```diff
@@ -218,7 +205,12 @@ class Modal extends React.Component<Props> {
 
     const innerChildren = __DEV__ ? (
       <AppContainer rootTag={this.context.rootTag}>
```
TheSavior: I think this will be broken now because `this.context.rootTag` will be undefined because you removed 

```
 static contextTypes: any | {|rootTag: React$PropType$Primitive<number>|} = {
    rootTag: PropTypes.number,
  };
```
I think you'll need to leave that part until https://github.com/facebook/react-native/pull/28104 is landed.

Libraries/Modal/Modal.js
```diff
@@ -159,19 +158,7 @@ class Modal extends React.Component<Props> {
     this._identifier = uniqueModalIdentifier++;
   }
 
-  static childContextTypes:
-    | any
-    | {|virtualizedList: React$PropType$Primitive<any>|} = {
-    virtualizedList: PropTypes.object,
```
TheSavior: I think this is resetting the usage of the context that is created in virtualizedlist, so pulling this into a separate context won't work because they'll be different values. I think this needs to reset whatever provider is created from VirtualizedList

Libraries/Modal/Modal.js
TheSavior: You actually don't need to wait, there is already a RootTagContext you can use: https://github.com/facebook/react-native/commit/f9ecce7440e712782f4fd760a46e54b3102914d8

thymikee:
Closing in favor of https://github.com/facebook/react-native/pull/29002. Thanks @Naturalclar!

PR #28198, opened 2020-02-28
Add xcscheme files back in the template

pvinis:
## Summary

Adding `xcscheme` files for iOS template back in. They were removed in #26471 after #25451. We have reverted the change on `.gitignore`, as this directory is supposed to be in git, but we didn't revert the removed files. Now they are back as well.

You can also see here that the two files are still missing, they got removed on rc.0 and are still missing on rc.3.
https://react-native-community.github.io/upgrade-helper/?from=0.61.5&to=0.62.0-rc.3

## Changelog

[iOS] [Changed] - Add `xcscheme` files for iOS template back in.

## Test Plan

`init` a nre project, and check if the files are there or not. The paths are `ios/MyApp.xcodeproj/xcshareddata/xcschemes/MyApp.xcscheme` and `ios/MyApp.xcodeproj/xcshareddata/xcschemes/MyApp-tvOS.xcscheme`

### DIFF
--- a/template/ios/HelloWorld.xcodeproj/xcshareddata/xcschemes/HelloWorld-tvOS.xcscheme
+++ b/template/ios/HelloWorld.xcodeproj/xcshareddata/xcschemes/HelloWorld-tvOS.xcscheme
@@ -1,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1130"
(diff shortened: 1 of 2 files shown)


TheSavior:
Can you make the PR title clearer? I’d edit it myself but I can’t figure out how to do that from mobile...

pvinis:
Done, I missed the title when submitting 🤦‍♂

PR #28199, opened 2020-02-28
iOS: Install Node 10 as 8 got dropped

tido64:
## Summary

We recently bumped our engine requirement in #28174 but are still installing an older Node version.

## Changelog

[Internal] [Fixed] - Install Node 10 as is required for `test_ios_e2e`

## Test Plan

`test_ios_e2e` should start building again.


tido64:
cc @safaiyeh, @hramos

PR #28221, opened 2020-03-03
Changing blob[@"type"] to blobType

chaira19:
## Issue and Motivation:
A crash occurring in the react-native application due to a code defect in this line of the library.

## Root Cause:
The issue is being caused because blob[@'type"] is assigned to contentType and whenever it will be nil, there would be a crash.

## Fix
 Variable blobType is created from blob[@"type"] and that should be used to be assigned to contentType. Also, the nil check was already done for blobType only.

## Changelog
Crash Fix in RCTBlobManager.mm

## Test Plan
Tested in the RN application, the crash is not occurring anymore.

### DIFF
--- a/Libraries/Blob/RCTBlobManager.mm
+++ b/Libraries/Blob/RCTBlobManager.mm
@@ -260,7 +260,7 @@ RCT_EXPORT_METHOD(release:(NSString *)blobId)
   NSString *contentType = @"application/octet-stream";
   NSString *blobType = [RCTConvert NSString:blob[@"type"]];
   if (blobType != nil && blobType.length > 0) {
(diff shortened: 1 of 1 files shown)


hramos:
Based on your explanation, this seems like a good change. I'd like to understand more about the underlying crash, however. Do you have a repro you can share that demonstrates the issue? This would help us understand the test plan better, and ascertain we are not introducing another issue with this change.

charlesbdudley:
Hi @chaira19, I'm going to close this so that we can focus on active PRs. If you'd like to take another stab at it, let me know and I'll be happy to reopen it. Thanks

PR #28224, opened 2020-03-03
[TextInput] Call onPress handler

retyui:
## Summary

I don't have any ideas how handle onPress events on `TextInput` (for web compatibility)
```js
// do not work
<TouchableWithoutFeedback onPress={() => console.log('it works')}>
  <TextInput editable={false} onTouchStart={() => console.log('it works')} />
</TouchableWithoutFeedback>;

// https://github.com/facebook/react-native/issues/14958#issuecomment-324237317
// do not work 
<TouchableWithoutFeedback onPress={()=>{console.log('press')}}>
   <View pointerEvents='none'>
      <TextInput editable={false} />
   </View>
</TouchableWithoutFeedback>
```

## Changelog

[General] [Added] - Add `onPress` prop to TextInput component

## Test Plan

```js
<TextInput
  onPress={() => console.log('it works')}
/>
```

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -982,6 +982,10 @@ function InternalTextInput(props: Props): React.Node {
     if (props.editable || props.editable === undefined) {
       nullthrows(inputRef.current).focus();
     }
+
+    if (props.onPress) {
+      props.onPress(event);
+    }
   };
 
   const _onChange = (event: ChangeEvent) => {


TheSavior:
Why do you want to listen for onPress? Are you sure you shouldn't be listening to onFocus and onBlur instead?

retyui:
@TheBerg 
I need to get coords onPress event

TheSavior:
Got it. I don't think this is the right approach. The callsite you modified is only the Android code path. This won't work on iOS. I don't think iOS has a way to make this same change. I'm going to close this as we won't merge this approach.

It feels like wrapping this in a touchable component is the correct approach. I'm not sure off the top of my head what that looks like, but Stack Overflow would likely be the right place to ask. 

Thanks for trying to fix this problem and sending your first PR though! I hope you find more opportunities to continue contributing!

PR #28267, opened 2020-03-09
E2E for keyboardType breaking autoCapitalize

safaiyeh:
## Summary

E2E for when keyboardType is set to default, autoCapitalize should work correctly. Mentioned here: https://github.com/facebook/react-native/pull/27523#issuecomment-577960694

## Changelog

[Internal] [Added] - E2E for when keyboardType is set to default, autoCapitalize should work correctly

## Test Plan

Added tests; however, noticed CircleCI Android tests do not run the E2E tests.


safaiyeh:
cc @JoshuaGross

TheSavior:
I actually tried this, but detox actually doesn't intend for this to work. See my issue here: https://github.com/wix/Detox/issues/1876

I wish we could have these tests...

PR #28268, opened 2020-03-09
fix(android): Ensure the MultiPostprocessor cache key is set correctly

matt-oakes:
## Summary

If the `MultiPostprocessor` wrapped postprocessors which returned a `null` cache key, it would return a `MultiCacheKey` with `mull` elements in the list. When compared these would match in the cache, causing issues where cached values were used where they shouldn't be.

This was causing issues when an `Image` component was using `resizeMode="repeat"` once, and then drawing another component with the same image source but with a different size. Because the source matched and the MultiPostprocessor cache key was incorrectly matching, it would return the repeated image with the wrong size, causing issues.

With this fix cache is now correctly implemented.

## Changelog

[Android] [Fixed] - Ensure the MultiPostprocessor cache key is set correctly to avoid issues when redisplaying images with a different size with resizeMode="repeat"

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/image/MultiPostprocessor.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/image/MultiPostprocessor.java
@@ -8,6 +8,7 @@
 package com.facebook.react.views.image;
 
 import android.graphics.Bitmap;
+import androidx.annotation.Nullable;
 import com.facebook.cache.common.CacheKey;
 import com.facebook.cache.common.MultiCacheKey;
 import com.facebook.common.references.CloseableReference;
@@ -49,6 +50,7 @@ public class MultiPostprocessor implements Postprocessor {
   }
 
   @Override
+  @Nullable
   public CacheKey getPostprocessorCacheKey() {
     LinkedList<CacheKey> keys = new LinkedList<>();
     for (Postprocessor p : mPostprocessors) {
@@ -58,7 +60,7 @@ public class MultiPostprocessor implements Postprocessor {
       }
     }
 
-    switch (postprocessors.size()) {
+    switch (keys.size()) {
       case 0:
         return null;
       case 1:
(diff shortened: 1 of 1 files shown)


ReactAndroid/src/main/java/com/facebook/react/views/image/MultiPostprocessor.java
```diff
@@ -52,9 +52,20 @@ public String getName() {
   public CacheKey getPostprocessorCacheKey() {
```
mdvacca: have you check what are the side effects of this method returning null?

ReactAndroid/src/main/java/com/facebook/react/views/image/MultiPostprocessor.java
```diff
@@ -52,9 +52,20 @@ public String getName() {
...
     LinkedList<CacheKey> keys = new LinkedList<>();
     for (Postprocessor p : mPostprocessors) {
-      keys.push(p.getPostprocessorCacheKey());
+      CacheKey cacheKey = p.getPostprocessorCacheKey();
+      if (cacheKey != null) {
+        keys.push(cacheKey);
+      }
+    }
+
+    switch (postprocessors.size()) {
```
mdvacca: can you double check this method?
where is postprocessors defined? maybe you need to use keys instead?

ReactAndroid/src/main/java/com/facebook/react/views/image/MultiPostprocessor.java
mdvacca: can you add the @Nullable annotation?

matt-oakes: Looking at the uses inside Fresco, `null` is an accepted and expected value:

https://github.com/facebook/fresco/search?p=2&q=getPostprocessorCacheKey&unscoped_q=getPostprocessorCacheKey

matt-oakes:
@mdvacca Thanks for the review. I've made those changes now

PR #28272, opened 2020-03-09
Publish nightly builds from master

acoates-ms:
## Summary

Publish builds of react-native from master on a nightly basis.

Nightly builds will publish versions of `react-native` as: `0.0.0-<first 9-digits-of-latest-commit>`.  This is inline with how `react` publishes nightly builds.  These builds will be tagged as `@nightly`.

This should make it easier for developers to try latest builds of react-native before a stable release, which may improve early reporting of bugs.  It is also a prereq to enable `react-native-windows`, and any other community platform or community module to be able to more aggressively track react-native master and be up to date with RN releases.

## Changelog

[Internal] [Added] - Publish nightly builds from master

## Test Plan

I attempted to run publish-npm.js and bump-oss-version.js manually, but due to the nature of the scripts its a little hard to do end to end testing.

### DIFF
--- a/scripts/bump-oss-version.js
+++ b/scripts/bump-oss-version.js
@@ -23,29 +23,42 @@ const yargs = require('yargs');
 let argv = yargs.option('r', {
   alias: 'remote',
   default: 'origin',
+})
+.option('n', {
+  alias: 'nightly',
+  type: 'boolean',
+  default: false,
 }).argv;
 
-// Check we are in release branch, e.g. 0.33-stable
-let branch = exec('git symbolic-ref --short HEAD', {
-  silent: true,
-}).stdout.trim();
+const nightlyBuild = argv.nightly;
 
-if (branch.indexOf('-stable') === -1) {
-  echo('You must be in 0.XX-stable branch to bump a version');
-  exit(1);
-}
+let version, branch;
+if (nightlyBuild) {
+  const currentCommit = exec('git rev-parse HEAD', {silent: true}).stdout.trim();
+  version = `0.0.0-${currentCommit.slice(0, 9)}`;
+} else {
+  // Check we are in release branch, e.g. 0.33-stable
+  branch = exec('git symbolic-ref --short HEAD', {
+    silent: true,
+  }).stdout.trim();
 
-// e.g. 0.33
-let versionMajor = branch.slice(0, branch.indexOf('-stable'));
+  if (branch.indexOf('-stable') === -1) {
+    echo('You must be in 0.XX-stable branch to bump a version');
+    exit(1);
+  }
 
-// - check that argument version matches branch
-// e.g. 0.33.1 or 0.33.0-rc4
-let version = argv._[0];
-if (!version || version.indexOf(versionMajor) !== 0) {
-  echo(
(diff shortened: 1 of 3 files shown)


alloy:
The publish step on CI needs to be fixed to accept GitHub as host automatically. Currently this is being done by connecting to a CI VM through SSH and initiating the publish step interactively.

Did you run into that?

acoates-ms:
Really?  I dont remember having to do that on the last time I published RN.  -- Although at this point I'm pretty sure you are more familiar with the process than I am.

Like I said in the testing note, its kind of hard to really test it end to end. -- Part of why I'm setting this up now is to give us some time on ironing out issues with the nightly before we are quite ready to start using it from `react-native-windows`

alloy:
It obviously shouldn’t work like that, so I’ll be fixing that shortly.

TheSavior:
This is exciting!

One thing I want to check since I'm not super familiar with this integration, will this cause [the releases tab](https://github.com/facebook/react-native/tags) to be spammed with a new release or tag every night? I'd like to avoid that. The React Repo doesn't seem to show the nightlies in the releases tab: https://github.com/facebook/react/tags

acoates-ms:
It shouldn't mess with the releases tab, since we are not adding any git tags. (No point since the version number already encodes a commit number)

hramos:
@alloy can you tell me more about this step? While I've had to SSH into a Circle CI job in the past, it should not be a requirement for publishing a release now.

TheSavior:
@hramos, For at least some additional info, see step 6 of the doc @alloy just added: https://github.com/react-native-community/releases/blob/master/docs/publish-release.md

hramos:
That sounds like the issue we ran into in the past. If I recall correctly, it was due to the use of a Docker container whose image didn't inherit from one of the stock Circle CI images. I thought we had fixed this, as it was a major friction point when cutting releases.

scripts/bump-oss-version.js
```diff
@@ -23,29 +23,42 @@ const yargs = require('yargs');
...
-  echo(
-    `You must pass a tag like 0.${versionMajor}.[X]-rc[Y] to bump a version`,
-  );
-  exit(1);
+const nightlyBuild = argv.nightly;
+
+let version, branch;
+if (nightlyBuild) {
+  const currentCommit = exec('git rev-parse HEAD', {silent: true}).stdout.trim();
+  version = `0.0.0-${currentCommit.slice(0, 9)}`;
```
alloy: Currently the shortest a hash can safely be in the RN repo is 8 characters, so, while it will probably take a while before 9 becomes an issue, in the changelog-generator I did settle on 10 chars for display purposes and no shortening where displaying it isn’t a concern.

Just my 2¢

acoates-ms: I picked 9 to align with what react is using.

alloy:
I expect this to fix the publish step from CI https://github.com/facebook/react-native/pull/28304

hramos:
These changes landed in 07def553968f88f65a2a07c3aea8e757e49530dd.

alloy:
@hramos I noticed a few issues re the nightly tasks:

* They seem to run every hour, rather than once every 24 hour.
* The publish task fails in a Buck related build step: https://circleci.com/gh/facebook/react-native/140387

PR #28274, opened 2020-03-10
Don't mess with Windows line endings

pvinis:
## Summary

During some rc releases we have seen the file `gradlew.bat` get diffs from release to release, and sometimes the diff is nothing, except some script messing with the line endings of that file.

According to [this](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#example) and my own experiments using `unix2dos` and `dos2unix` to swap the line endings of that file, this change fixes this problem.
We specify all the `.bat` files should have the Windows line endings.

## Changelog

[General] [Fixed] - Specify line endings of `.bat` files for Windows in `.gitattributes`

## Test Plan

To assist with the test, you can use `brew install unix2dos`.

- Init a new project or use an existing one.
- Make sure in your `.gitattributes` of the project and of the user (home dir or `.config/git/attributes` don't have anything about `.bat` files, or `* text=auto`.
- Run `unix2dos ./android/gradlew.bat`.
- Check git status.
- Run `dos2unix ./android/gradlew.bat`.
- Check git status.
- You will notice that the changes when swapping around are the whole file changed.

- Using the changes in this PR, and doing the above, the git status will be no changes, as it should be. The line changes stay Windows-style.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
-# Force LF line endings for Bash scripts.   On Windows the rest of the source
-# files will typically have CR+LF endings (Git default on Windows), but Bash
-# scripts need to have LF endings to work (under Cygwin), thus override to force
-# that.
+# Force correct line endings.
+# On Windows the git default is CR+LF and on macOS/*nix the git default is LF.
+# Regardless of OS, bash scripts should have LF endings (handled by Cygwin on Windows),
+# and bat files should have CR+LF (only used on Windows).
 gradlew text eol=lf
 *.sh text eol=lf
+*.bat text eol=crlf
--- a/template/_gitattributes
+++ b/template/_gitattributes
@@ -1,1 +1,4 @@
 *.pbxproj -text
+
+# specific for windows script files
+*.bat text eol=crlf
--- a/template/_gitattributes
+++ b/template/_gitattributes
@@ -1,4 +1,1 @@
 *.pbxproj -text
-
-# specific for windows script files
-*.bat text eol=crlf


TheSavior:
Thanks, I’m excited to have this not be a problem anymore. However, this change will still require a change to everyone’s project to add this. Could we avoid that? Could we make this change in React Native core to make sure that PRs modifying the gradle file don’t change the line endings? Once people init a project I don’t care so much if they cause the file’s line endings to be swapped.

pvinis:
I guess we could, but to be honest I think this change is the better way.

I haven't found where and when this broke (I haven't looked too hard though), but I think this could happen again by just having some formatter or linter that is a tiny more too zealous.

I also don't know if Windows is fine with that file having unix line endings, no matter if the change came from a new project, an upgrade, or a change in a user's computer.

As an extra, I have this rule on some of my machines at my user dir git config, when I know I might touch files like that there. This is a good/sensible default that is not there on many dev machines, unfortunately.

So based on these reasons, I think adding it here is the better choice. But if you think we should find where this broke (rn or cli or elsewhere) and add it there, I would be ok to add it there **too**, not **only** there. Does that make sense?

TheSavior:
The bumps look like they were made as part of a Facebook employee doing a codemod across the internal repo: https://github.com/facebook/react-native/commits/master/template/android/gradlew.bat

When people upgrade and see this change to the gitattributes in their project they'll wonder why it matters. I'd rather avoid changes to the template that we can avoid. 

I'd feel more comfortable [adding it to here](https://github.com/facebook/react-native/blob/master/.gitattributes#L5) to cover the existing use case and punt on the decision to add it to the template in the case we run into problems later.

pvinis:
Alright, as you think it's best. 👍

I can make a PR to change it for the rn repo and not the template. Let's close this one then?

TheSavior:
Do you want to change this PR to modify the other file?

I think it's reasonable from our side to make the case that they are suggestions not requirements, but I'm not sure if that's how other people see them. Personally, I would think I should make all the changes in the upgrade helper because if I don't future upgrades will be less clear, or I'll realize I actually need a step that I skipped a couple releases back and now have problems with my project. I'm not sure how we fix that or communicate the intent better which is why I'm trying to be fairly protective of the template. 

I don't actually know much about these line endings and don't frequently work on open source projects with React Native, so it is possible that putting it in the template is the right choice and I'm happy to be overruled with some more voices. In the mean time though, it's *definitely* reasonable to add it to the core `.gitattributes` so we should at least do that and revisit adding it to the template if we hear more opinions. Does that sound okay?

pvinis:
Fair, I'll edit the PR here to change the other file, and I will revert the current changes. Then we either tag some people that could give opinions, or we leave it and revisit if this comes up again.

Issue #27696, opened 2020-01-07
[iOS] TextInput default text color change to white when running on real device


thormengkheang:
React Native version:

System:
    OS: macOS 10.15.2
    CPU: (4) x64 Intel(R) Core(TM) i5-4570R CPU @ 2.70GHz
    Memory: 69.59 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.5.0 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 23.0.2, 25.0.0, 25.0.1, 25.0.2, 26.0.2, 26.0.3, 27.0.3, 28.0.2, 28.0.3
      System Images: android-23 | Google APIs ARM EABI v7a, android-23 | Google APIs Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Xcode: 11.3/11C29 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: 0.61.5 => 0.61.5

## Steps To Reproduce

1. git clone git@github.com:james1888/RnTextInputBug.git
2. cd RnTextInputBug
3. npm start
4. npm ios
5. open xcode
6. connect real device
7. run on real device

On simulator

![ezgif com-video-to-gif (1)](https://user-images.githubusercontent.com/8617506/71880857-c41d4f80-3163-11ea-8b9b-fc093429cebe.gif)

On real device

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/8617506/71880873-cda6b780-3163-11ea-8577-f85032d4488e.gif)

Describe what you expected to happen:

Default TextInput text color should be black on iOS

link to a repository:
https://github.com/james1888/RnTextInputBug

AlexKitCoder:
me too; ios iphone XR device;TextInput  default color was white,can't see it any more! simulator was ok,android was ok

james1888:
[...]

Issue #27856, opened 2020-01-24
TextInput Text is not shown on IOS 13.3


appify-waheed:
I am using react-native `0.61.2` and while entering text into textInput didn't show anything but text entered successfully, this issues only happens with IOS `v13.3` on the other versions and on android it works fine.

React Native version:
  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
    Memory: 23.05 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.2 - /usr/local/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.11.3 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 22, 23, 24, 25, 26, 27, 28, 29
      Build Tools: 23.0.1, 25.0.3, 26.0.2, 27.0.3, 28.0.3
[...]

Issue #27437, opened 2019-12-07
TextInput: no placeholder and white font in testflight


wkibbler:
Issue with text input not showing placeholder and the value font being while when the project is compiled and released testflight

IOS Version: 13
React Native Version: 0.60.5
React Native CLI version: 2.0.1
Xcode version: 11.2.1
`[skip envinfo]`

Here is the code behind the textinput
`<TextInput style={styles.input} keyboardType="numeric" onChangeText={(value) => this.afunction(value)} value={this.state.value} placeholder="placeholder"/>`

This is what it looks like when in a development simulator:
https://github.com/wkibbler/rn-issue-screenshots/blob/master/in-simualator.png

This is what it looks like when downloaded into an iPhoneX using testflight and ios 13:
https://github.com/wkibbler/rn-issue-screenshots/blob/master/on-testflight.jpeg
(ignore the red circle thats just a tester illustrating the issue)

markwillis82:
We are also seeing this on a single iPhone 7 plus. IOS 13.2.2. It's fine in the simulator and on other devices (iPhone XS. IOS 13.1.3)

Our react-native info is:

> System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
    Memory: 630.31 MB / 32.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    npm: 6.4.1 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 27.0.3, 28.0.2, 28.0.3
      System Images: android-23 | Google APIs Intel x86 Atom, android-26 | Google APIs Intel x86 Atom, android-26 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 11.2.1/11B53 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.4 => 0.60.4
[...]

PR #28280, opened 2020-03-10
Fix TextInput in dark mode: Unconditionally apply foreground color

JonnyBurger:
## Summary

This PR should fix a bug which fixes TextInput color being white in dark mode on iOS.
It seems like on Line 217, the intention is that if no `_foregroundColor` is set, it falls back to `[UIColor blackColor]`.

However, the `effectiveForegroundColor` is only being applied if `_foregroundColor` is not nil. Therefore the fallback is never triggered in the first place. This means that `NSForegroundColorAttributeName` is unintentionally being set to nil, which still meant that by default the text is being rendered black. However in iOS 13, the text would be rendered in white if you have enabled dark mode.

Having a fallback to black that is never triggered makes me believe that this is a bug and that the current behavior is not intended. I am fixing the bug by removing the `if` statement completely as it seems to me that the `effectiveForegroundColor` is already sufficiently performing null checks.

## Changelog

[iOS] [Fixed] - TextInput color has the same default (#000) on iOS whether in light or dark mode

## Test Plan

I have applied this patch in my project and it has solved the bug that the TextInput turns white when in dark mode. I have no further plans on how to perform tests.

### DIFF
--- a/Libraries/Text/RCTTextAttributes.m
+++ b/Libraries/Text/RCTTextAttributes.m
@@ -131,9 +131,7 @@ NSString *const RCTTextAttributesTagAttributeName = @"RCTTextAttributesTagAttrib
   // Colors
   UIColor *effectiveForegroundColor = self.effectiveForegroundColor;
 
-  if (_foregroundColor || !isnan(_opacity)) {
-    attributes[NSForegroundColorAttributeName] = effectiveForegroundColor;
-  }
+  attributes[NSForegroundColorAttributeName] = effectiveForegroundColor;
 
   if (_backgroundColor || !isnan(_opacity)) {
     attributes[NSBackgroundColorAttributeName] = self.effectiveBackgroundColor;


JonnyBurger:
See Twitter for video of the bug and for further context: https://twitter.com/JNYBGR/status/1237440821445767171

This PR should also fix #27696 #27437 #27856

JonnyBurger:
@shergin When the `if` statement is removed, it will always use the value of `effectiveForegroundColor` which can be seen above in the same file:
https://github.com/facebook/react-native/blob/c49e8d34350c5fb58c05040a1b9707dba9b55a0d/Libraries/Text/RCTTextAttributes.m#L216-L225

> Why do you think so?

Because in the effectiveForegroundColor getter, a fallback is defined: `_foregroundColor ?: [UIColor blackColor];`

> Why unintentionally?

I believe so because there is logic for a fallback to `[UIColor blackColor]` defined, however it is never being used because `effectiveForegroundColor` is guarded by an if statement that checks if `_foregroundColor` is truthy (or `opacity` but in this case I think it doesn't matter).

> Why do you think all props are gated with ifs but this particular prop should not?

Because both `_foregroundColor` being falsy and `_opacity` being not NaN are already handled in the getter of `effectiveForegroundColor`.

---
Currently the logic is:
```
-> foregroundColor truthy _or_ opacity set
    true:
         use *effectiveForegroundColor*
         --> foregroundColor set?
                if false:
                        --> set to black
         --> opacity set?
                if true:
                        --> change the color to match opacity
    false:
         *use foregroundColor* (this is the buggy case, in this case it does not default to black)
```

I want to remove the if statement so that the logic is just:

```
use *effectiveForegroundColor*
 --> foregroundColor set?
        if false:
            --> set to black
--> opacity set?
        if true:
            --> change the color to match opacity
```

So the difference is that in the case that neither `foregroundColor` nor `opacity` is set, now the color is being set to `[UIColor black]` rather than `foregroundColor` which is nil.

Hope I did a good job explaining!

shergin:
@JonnyBurger I am truly sorry, but I still cannot understand why it's **the right approach** to fix this problem. I understand your logic but you described how it would work in your case but not why it will work as desired in all other cases.

I see several problems here (not in your approach, in the whole problem).
1. Seems we have incorrect implementations of `effectiveForegroundColor` and `effectiveBackgroundColor`, these method assumes that the default is always black on white (which is clearly not the case in dark mode).
2. The background of the text input stays white in dark mode. Not sure why.

Note that all text attributes are applied cascadingly, they are being applied several times (directly and indirectly) to the attributed text and because of that we never should override anything there without a good reason.

I would suggest just try to explicitly set `self.textColor = [UIColor blackColor];` right inside `RCTUITextField` constructor (exactly as we do in RCTUITextView) to override iOS defaults (which are seems change with dark/light mode).

JonnyBurger:
@shergin I understand!
I will take this into consideration, go back to the drawing board and come back with a more straightforward PR solving the problem more explicitly.

thomasbird1984:
Did this get resolved? I'm still seeing this issue.

PR #28284, opened 2020-03-11
Add diffing to app bundle size reports

tido64:
## Summary

Add diffing to app bundle size reports.

TODO:
- [x] `master` CI should commit data to the database
- [ ] Get help storing the API key for the app with write permissions

## Changelog

[Internal] [Changed] - Add diffing to app bundle size reports

## Test Plan

- App bundle size reports should now display a diff where available
  - Right now, the database contains only one entry for the last known good iOS build
- Triggering a new build should not create additional comments

### DIFF
--- a/bots/datastore.js
+++ b/bots/datastore.js
@@ -16,9 +16,16 @@ require('firebase/firestore');
 function initializeStore() {
   const PROJECT_ID = 'react-native-1583841384889';
   const {FIREBASE_APP_ID} = process.env;
+  const apiKey = [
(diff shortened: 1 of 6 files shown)


hramos:
>  Get help storing the API key for the app with write permissions

Would this be the key used to write to the store on commits to master? If so, I can help get it configured as an environment variable on Circle CI itself. We have Circle configured as to never expose these envvars to forks and pull requests, but they will be available to jobs that are invoked by commits to master. You may message me on Teams.

PR #28297, opened 2020-03-12
Make Lambda function called in NativeModule mutable to improve performance

### LINKED ISSUE #28271: CxxNativeModule::invoke is copying parameters that are meant to be moved
### Description:

The `CxxNativeModule::invoke` function [makes use of a lambda](https://github.com/facebook/react-native/blob/0.62-stable/ReactCommon/cxxreact/CxxNativeModule.cpp#L136-L159) to run a native module function on the message queue thread. However the lambda isn't marked as `mutable` -- this prevents the `params` capture from being movable [when calling the method itself](https://github.com/facebook/react-native/blob/0.62-stable/ReactCommon/cxxreact/CxxNativeModule.cpp#L146). If the `params` argument is expensive to copy, this can impact performance.

### React Native version:

This can be seen in v0.62.

Steps to reproduce (and expected results) are not applicable (unless running with a profiler).

maschad:
## Summary

Fixes #28271
As explained by @mmallet-youilabs , if the parameters passed to the `move` function are too expensive, this can have an impact on performance. Thus making these parameters captured by value mutable, the parameters are not movable.

## Changelog

[iOS] [Fixed] - Message

## Test Plan

Steps to reproduce (and expected results) are not applicable (unless running with a profiler).

### DIFF
--- a/ReactCommon/cxxreact/CxxNativeModule.cpp
+++ b/ReactCommon/cxxreact/CxxNativeModule.cpp
@@ -148,7 +148,7 @@ void CxxNativeModule::invoke(
   // mhorowitz #7128529: convert C++ exceptions to Java
 
   messageQueueThread_->runOnQueue(
-      [method, params = std::move(params), first, second, callId]() {
+      [method, params = std::move(params), first, second, callId]() mutable {
 #ifdef WITH_FBSYSTRACE
         if (callId != -1) {
           fbsystrace_end_async_flow(TRACE_TAG_REACT_APPS, "native", callId);


maschad:
Hey @shergin , those are valid points but I am not sure I understand what you mean by point 4. Isn't the [move](https://github.com/facebook/react-native/blob/f6a8452e7699842c0ef9fcfb58a33a36209a68f9/ReactCommon/cxxreact/CxxNativeModule.cpp#L146) moving the data members?

NickGerleman:
I think this does avoid a copy. Pay attention to the call of `method.func(...)`.

```C++
messageQueueThread_->runOnQueue(
      [method, params = std::move(params), first, second, callId]() mutable {
        ...
        try {
          method.func(std::move(params), first, second);
        } catch (const facebook::xplat::JsArgumentException &ex) {
          ...
        }
      });
}
```

Even before, we would do a move construct of the `dynamic` in the lambda capture. Before `mutable` though, `method.func(std::move(params), first, second)` does a sneaky copy. The `std::move` will cast `params` from l-value to r-value but does not remove constness. That means we try to call the `folly::dynamic` constructor using an `const folly::dynamic&&`. Overload resolution will then choose the copy constructor taking a const l-value, sine the move contructor requires a non-const r-value reference.

shergin:
@maschad  @NickGerleman Thank you! You are absolutely right.
I am sorry, I missed the `method.func(std::move(params), first, second);` line.

maschad:
Thanks for explaining @NickGerleman I'm not very strong in C++ so that clarified a lot for me 👌🏾

Issue #28200, opened 2020-02-28
Missing NSLocationAlwaysUsageDescription in Info.plist without any GeoLocation Usage in 0.61.5 (ITMS-90683)


felix-voicehub:
Description:

I am getting E-Mail from Apple after submitting to App Store Connect regarding ITMS-90683. It basically says that I have to include `NSLocationAlwaysUsageDescription` and `NSLocationWhenInUseUsageDescription` in my Info.plist because they say that my

> app's code references one or more APIs that access sensitive user data.

Now, I do not use any of react natives geolocation functions. I have scanned my whole repo for anything regarding these APIs, and the only thing which came up was `RCTConvert+CoreLocation.h` (https://github.com/facebook/react-native/blob/master/React/Views/RCTConvert%2BCoreLocation.h) and it's ".m" file. These do import "CoreLocation", which could be enough to trigger this E-Mail, as suggested [here](https://github.com/facebook/react-native/issues/20879#issuecomment-592505530).

See this issue for explaining the whole background: https://github.com/facebook/react-native/issues/20879.

React Native version:
```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory: 22.33 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.12.0 - /usr/local/bin/node
    npm: 6.13.6 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2
      System Images: android-28 | Intel x86 Atom_64, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5977832
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-eject: 0.1.2
```

## Steps To Reproduce

1. Bootstrap a clean react native app.
2. Submit it to App Store connect.

## Expected Results

[...]

PR #28300, opened 2020-03-13
Removed core RCTConvert CoreLocation Libraries as well as reference i…

maschad:
## Summary

This PR is to address #28200

Even though GeoLocation was extracted from the Core, there are still libraries i.e. the CoreLocation.h library included which cause Apple to reject Apps, as mentioned here.

This removed all references to the CoreLocation library that where remaining.

## Changelog

[iOS] [Removed] - Message

## Test Plan

- Flow check passes.
- RNTest built successfully.


shergin:
This file is still being used by some deps. One of them (which is blocking landing internally) is this one: https://github.com/react-native-community/react-native-maps/blob/72b0f77e31d18959e91acb036d5f91ab46e11ef2/lib/ios/AirMaps/AIRMapCircleManager.m#L14

I think we should copy the functionally/code there. And then try to remove it from here one more time.


@maschad Can you do it?

maschad:
Thanks for the recommendation @shergin I re-added them, but my only concern is that Apple's static analyzer may still pick up the `CoreLocation` files and then trigger a warning.

shergin:
@maschad Nonono, I meant we need to move/copy those files to that library to break the dependency the core in this regard. After we do that, we can safely remove them from the core.

I will land this one as is. But I would appreciate if you can fix that library and then open another PR that will finally delete the files from the core. 🙏

maschad:
@shergin silly me, I should have realized that 😅. Thanks for clarifying anyways, I can definitely do that 👍🏿

PR #28311, opened 2020-03-15
Do not explicitly include ".js" in Library imports

NickGerleman:
## Summary

A few recent imports have explicitly added ".js" to the end of their path. This prevents Metro from resolving platform-specific JS files, e.g. "Foo.android.js" or "Foo.windows.js" instead of "Foo.js".

React Native Windows provides its own implementation of files in a few cases where stock React Native will share them between Android and iOS. We hope to reduce/eliminate these long term, but requiring explicit ".js" files currently breaks us in a couple of places where we have custom implementations.

This change is a quick regex replace of ES6 and CommonJS imports in 'Libraries/" to eliminate ".js".  

## Changelog

[General] [Fixed] - Do not explicitly include ".js" in Library imports

## Test Plan

I haven't done any manual validation of this, but `flow-check` should catch any issues with this during CI.

### DIFF
--- a/Libraries/Components/Pressable/Pressable.js
+++ b/Libraries/Components/Pressable/Pressable.js
@@ -12,18 +12,18 @@
 
 import * as React from 'react';
 import {useMemo, useState, useRef, useImperativeHandle} from 'react';
-import useAndroidRippleForView from './useAndroidRippleForView.js';
+import useAndroidRippleForView from './useAndroidRippleForView';
 import type {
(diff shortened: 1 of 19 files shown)


NickGerleman:
The only way I think this could cause breakage was if there was anywhere where we had both shared JS and platform specific JS, and were intentionally including the shared instead of platform-specific, but I don't think we're doing that anywhere based on some quick inspection.

PR #28326, opened 2020-03-16
Removed CoreLocations libraries in accordance with Lean core

maschad:
## Summary

As mentioned in #23313 as well as [here](https://github.com/facebook/react-native/pull/28300#issuecomment-599344895) , the Geolocation libraries are being moved out of react-native core. Further more these particular libraries are being moved to[ react-native-maps](https://github.com/react-native-community/react-native-maps) so [this PR](https://github.com/react-native-community/react-native-maps/pull/3342) will have to be merged first.

## Changelog

[iOS] [Removed] - RCTConvert+CoreLocation removed.

## Test Plan
All dependencies on these libraries should still work.


hramos:
Will wait for https://github.com/react-native-community/react-native-maps/pull/3342 to be merged before landing this.

I saw that RNTesterPods's `project.pbxproj` has a reference to `/Headers/Public/React-RCTGeolocation` in `HEADER_SEARCH_PATHS`. Does that need to be removed as well?

maschad:
@hramos I would think so. And I suppose removing that would require a similar level of scrutiny regarding the dependencies it may have, do you think that should be a separate PR?

PR #28332, opened 2020-03-17
Fix mock for TextInput

SergioEstevao:
## Summary

This PR adds the `isFocused` method to the mock of the TextInput component. My understanding some of the latest changes on the TextInput to make it use a forwardRef change the way this method is mock giving an error when trying to use in on a mock.
The change suggested here fixes the issue.

## Changelog

[JavaScript] [Fixed] - Fix the mock for TextInput to support the `isFocused` method

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -92,6 +92,8 @@ jest
     mockComponent('../Libraries/Components/TextInput/TextInput', {
       ...MockNativeMethods,
       isFocused: jest.fn(),
+      clear: jest.fn(),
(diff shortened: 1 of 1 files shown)


TheSavior:
Good catch! Can you add the other two methods as well? https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L694-L698

SergioEstevao:
@TheSavior just update the PR with the changes you requested.

PR #28338, opened 2020-03-17
[CI] Enable 'Needs Attention' action

hramos:
## Summary

The ['Needs Attention'](https://github.com/hramos/needs-attention) action will remove the "Needs: Author Feedback" label and replace it with "Needs: Attention" whenever the original author of an issue adds a comment.

Removes the `no-response` GitHub integration. The 'close after 21 days' functionality from `no-response` is not present in the 'Needs Attention' action. We'll need to add this back some other way. Considering we were not closing issues that were in the Needs Response state until recently, I think it's OK to proceed here.

## Changelog

[Internal] [CI] - Enable 'Needs Attention' action

## Test Plan

https://github.com/hramos/needs-attention/issues/1


.github/workflows/needs-attention.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Apply Needs Attention Label
+        uses: hramos/needs-attention@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          response-required-label: "Needs: Author Feedback"
+          needs-attention-label: "Needs: Attention"
+        id: needs-attention
```
TheSavior: What is the id for?

hramos: Assigning an id to a step lets us refer back to it later. In this case, it's used in L21 to get the output of this step via `steps.needs-attention`.

PR #28349, opened 2020-03-20
Fix Animated Value initialized with undefined in ScrollView

janicduplessis:
## Summary

When passing an object to contentOffset that doesn't have `y` prop set it causes the following error:

```
 Error: AnimatedValue: Attempting to set value to undefined

This error is located at:
    in ScrollView (at src/index.js:638)
...
```

This happens since a runtime check was added to the `AnimatedValue` constructor. (a3aaa471eca58b31597b9a0669f7ade385ccb175)

According to flow types the object passed to contentOffset should always contain both x and y props but since it worked before when y is undefined I think its fine to patch the runtime behaviour defensively, especially since the code change is simple.

## Changelog

[General] [Fixed] - Fix Animated Value initialized with undefined in ScrollView

## Test Plan

Tested that the crash no longer reproduces when passing an empty object to contentOffset.

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -716,11 +716,9 @@ class ScrollView extends React.Component<Props, State> {
   UNSAFE_componentWillMount() {
     this._scrollResponder.UNSAFE_componentWillMount();
     this._scrollAnimatedValue = new AnimatedImplementation.Value(
-      this.props.contentOffset ? this.props.contentOffset.y : 0,
+      this.props.contentOffset?.y ?? 0,
     );
-    this._scrollAnimatedValue.setOffset(
-      this.props.contentInset ? this.props.contentInset.top || 0 : 0,
-    );
+    this._scrollAnimatedValue.setOffset(this.props.contentInset?.top ?? 0);
     this._stickyHeaderRefs = new Map();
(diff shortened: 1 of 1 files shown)


janicduplessis:
cc @lunaleaps @sahrens

Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -716,11 +716,9 @@ class ScrollView extends React.Component<Props, State> {
   UNSAFE_componentWillMount() {
     this._scrollResponder.UNSAFE_componentWillMount();
     this._scrollAnimatedValue = new AnimatedImplementation.Value(
-      this.props.contentOffset ? this.props.contentOffset.y : 0,
-    );
-    this._scrollAnimatedValue.setOffset(
-      this.props.contentInset ? this.props.contentInset.top || 0 : 0,
+      this.props.contentOffset?.y ?? 0,
```
lunaleaps: Since there is no FlowFixMe here, my guess is there is a FlowFixMe somewhere in its usage and I'd rather we fix it there than mask the error by providing a default?

janicduplessis: The crash was happening for me in a third party library that did not use flow / typescript. I'm also fine keeping the behavior as is and fixing the library only, the goal of this is mostly to keep the same runtime behavior as before.

PR #28358, opened 2020-03-20
Fix rounded border drawing when border-radius is smaller than border-width

IjzerenHein:
## Summary

This PR fixes the drawing of the border rounded edges when the border-radius is small than the border-width. The current implementation capped the possible border-radius making it impossible to set smaller border-radii when using thicker borders. After inspection it was found that the rounded-rect calculation is incorrect.

## Changelog

`[Android] [Fixed] - Fix rounded border-drawing when border-radius is smaller than border-width`

## Test Plan

**Faulty situation:**

As you can see, when the border-radius becomes very low, the border is stuck at a minimum value. Only after setting the border-radius fully to 0 is it again rendered correctly.

ezgif com-video-to-gif (2)

**After the fix:**

ezgif com-video-to-gif (3)

### DIFF
--- a/RNTester/js/examples/View/ViewExample.js
+++ b/RNTester/js/examples/View/ViewExample.js
@@ -148,12 +148,53 @@ exports.examples = [
     },
   },
   {
-    title: 'Circle with Border Radius',
+    title: 'Rounded Borders',
     render(): React.Node {
       return (
-        <View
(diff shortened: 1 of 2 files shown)


jmunozDevsu:
@shergin This introduces a new bug when border width is smaller than radius.

PR #28382, opened 2020-03-24
Improve test manual e2e script

alloy:
## Summary

Cleans up and improves the manual e2e test script:

* Use `yarn` instead of `npm` to install dependencies that have previously been recorded in `yarn.lock`
* Fix incorrectly opening the Xcode project, rather than workspace, of the iOS `RNTestProject`
* Use RN CLI to build/run both Android and iOS test apps
* Launch and prepare Android emulator and iOS simulators instead of requiring manual intervention:
  - Delete `RNTester` app
  - Delete `RNTestProject` app
* Launch packager for `RNTester` automatically in background instead of requiring manual intervention
* Refactor and organise differently to allow for easier future maintenance

## Changelog

[Internal] [Fixed] - Fix and smoothen usage of `test-manual-e2e.sh` script

## Test Plan

Ran it manually many times.

### DIFF
--- a/scripts/test-manual-e2e.sh
+++ b/scripts/test-manual-e2e.sh
@@ -126,7 +126,7 @@ PACKAGE_VERSION=$(cat package.json \
             info "Press any key to run the sample in Android emulator/device"
             info ""
             read -n 1
-            npx react-native run-android
+            node "$CLI_PATH" run-android
         }
 
         # iOS
@@ -142,7 +142,7 @@ PACKAGE_VERSION=$(cat package.json \
             info "Press any key to open the project in Xcode"
             info ""
             read -n 1
-            open "ios/${project_name}.xcodeproj"
+            node "$CLI_PATH" run-ios
         }
     }
     popd
--- a/scripts/test-manual-e2e.sh
+++ b/scripts/test-manual-e2e.sh
@@ -24,6 +24,15 @@ info() {
     echo -e "$BLUE""$*""$ENDCOLOR"
 }
 
+kill_packagers() {
+    success "Killing any running packagers"
+    lsof -i :8081 | grep LISTEN
+    lsof -i :8081 | grep LISTEN | /usr/bin/awk '{print $2}' | xargs kill
+}
+
+REPO_ROOT=$(pwd)
+CLI_PATH="$REPO_ROOT/cli.js"
+
 PACKAGE_VERSION=$(cat package.json \
   | grep version \
   | head -1 \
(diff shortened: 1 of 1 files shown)


alloy:
Actually still need to make the iOS side of launching/cleaning work, but need to do some work in the CLI package for that first.

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
+{
+    info "Starting the packager in the background"
+    node "$CLI_PATH" start &
+
+    # Android
+    {
+        ./gradlew :RNTester:android:app:installJscDebug || error "Couldn't build RNTester Android"
+
+        info "Deleting previously installed Android RNTester app"
+        adb uninstall com.facebook.react.uiapp
```
kelset: nice

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
-read -n 1
-
-./gradlew :RNTester:android:app:installJscDebug || error "Couldn't build RNTester Android"
+ANDROID_EMULATOR_NAME=${ANDROID_EMULATOR_NAME:-$(emulator -list-avds | awk '{print $1}')}
 
-info "Press any key to run RNTester in an already running Android emulator/device"
-info ""
-read -n 1
-adb shell am start -n com.facebook.react.uiapp/.RNTesterActivity
+# Prepare
```
kelset: I would probably change a bit this section to 'clean up the house' more. For ex. I'd extract (similar to what happens to `kill_packagers`) all the cleanup part to its own method and in it, something along the lines of:
```
clean_folders() {
    rm -rf android
	rm react-native-*.tgz
    rm -rf /tmp/RNTestProject
	rm -rf node_modules 
}
```

and call it after `    info "Preparing version $PACKAGE_VERSION"`

What do you think?

PS: maybe we could even add some more scripts to that like
`rm -rf ~/Library/Developer/Xcode/DerivedData/*`

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
+        info "Press any key to run RNTester in an already running Android emulator/device"
+        info ""
+        read -n 1
+        adb shell am start -n com.facebook.react.uiapp/.RNTesterActivity
+    }
+
+    # iOS
+    pushd RNTester
+    {
+        info "Press any key to open the workspace in Xcode, then build and test manually."
```
kelset: not sure if it should live here or not, but I was wondering if it was useful to run `rm -rf ios/build`...

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
+        info ""
+        read -n 1
+        success "Installing CocoaPods dependencies"
+        rm -rf Pods && pod install
```
kelset: ...and `pod cache clean --all` here.
Too much?

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
+    info ""
+    read -n 1
+
+    kill_packagers
+}
 
-cd "$repo_root"
+# Test new app from template
+{
+    npm pack
```
kelset:
```suggestion
    yarn pack
```
(or is it a npm specific command?)

scripts/test-manual-e2e.sh
```diff
@@ -24,98 +24,139 @@ info() {
...
+            info "Press any key to open the project in Xcode"
+            info ""
+            read -n 1
+            node "$CLI_PATH" run-ios
+        }
+    }
+    popd
+}
 
 info "Next steps:"
```
kelset: tbh I think that these last 2 lines could be even be removed, I don't think they provide any actual value. Or maybe we could change the link to point to https://github.com/react-native-community/releases/blob/master/README.md or even https://github.com/react-native-community/releases/blob/master/docs/release-process.md?

scripts/test-manual-e2e.sh
alloy: Hmm, yeah that seems a bit much. What is your experience with this?

scripts/test-manual-e2e.sh
alloy: Not really, but it also doesn’t do anything with version resolving, so I’m less worried about it. I can take a look, though, and make the whole script uniform 👍

scripts/test-manual-e2e.sh
kelset: tbh just that I am a bad person and I do tabula rasa on my projects when I need to ensure full clean 😂 and you know way better than me how Pods work so I'm fully ok with not including this command 👍

kelset:
closing this as it's stale + we moved to a new script https://github.com/facebook/react-native/pull/34513

PR #28390, opened 2020-03-24
Yoga: add option to call measure callback on all nodes

sagokhal:
## Summary

When using Yoga in combination with external layout systems (brown field scenarios) Yoga leaf nodes can be subtrees w.r.t. the external layout system.
During the measure phase, these subtrees need to be measured by the external layout system - even if the Yoga node's mode is YGMeasureModeExactly.

This change adds a preprocessor conditional that can be set by out-of-tree platforms to always call measure for all nodes that have the measure callback.

## Changelog

[General] [Added] - Yoga: add preprocessor option to always call measure callback.

## Test Plan

Verified that this behaves as intended.

### DIFF
--- a/ReactCommon/yoga/yoga/Yoga.cpp
+++ b/ReactCommon/yoga/yoga/Yoga.cpp
@@ -1695,6 +1695,15 @@ static void YGNodeWithMeasureFuncSetMeasuredDimensions(
       : YGFloatMax(
             0, availableHeight - marginAxisColumn - paddingAndBorderAxisColumn);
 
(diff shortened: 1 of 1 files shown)


NickGerleman:
For some more context, this change was needed for the Win32 version of react-native. @acoates-ms has some more context as well.

shergin:
@sagokhal To merge the change into Yoga you need to merge that into Yoga repository. After it merged, it will be automatically sync-ed with React Native repo.
Could you please open another PR here: https://github.com/facebook/yoga/ ?

PR #28399, opened 2020-03-26
Fix RequestBodyUtil for older version of Android (M or before)

daisy1754:
## Summary:

This is bugfix for following code in RequestBodyUtil.java.

```
stream.getChannel().transferFrom(channel, 0, Long.MAX_VALUE);
```

This throws IllegalArgumentException in Android M or before. It seems in old version of JVM it internally casts third argument to integer so when you pass value that is larger than Integer.MAX_VALUE app would crash. 

See:
https://bugs.openjdk.java.net/browse/JDK-5105464
https://stackoverflow.com/questions/55696035/thrown-illegalargumentexception-when-using-java-nio

## Changelog:

[ANDROID] [FIXED] - Fix issue downloading request body via remote URL

## Test Plan
Run following code on Android M or before. It would crash w/o this PR but it won't with this PR.

```
const body = new FormData();
const image = { uri: "https://reactnative.dev/img/showcase/facebook.png", path: null };
body.append('user[img]', fileBody(image));	
fetch("https://example.com", {
    method: 'POST',
    headers: {
    Accept: 'application/json',
    'Content-Type': 'multipart/form-data;',
    },
    body: body
});
```

### DIFF
--- a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/network/RequestBodyUtil.java
+++ b/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/network/RequestBodyUtil.java
@@ -11,6 +11,7 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.net.Uri;
+import android.os.Build;
 import android.util.Base64;
 import androidx.annotation.Nullable;
 import com.facebook.common.logging.FLog;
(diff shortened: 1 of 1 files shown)


daisy1754:
This 7 line PR has been open for more than a month with no response. can someone review please? @mdvacca @fkgozali (randomly picked you two since you seemed to touch android native code recently)

daisy1754:
@axe-fb @RSNara can either of you help? (wanted to clean up my open PRs)

daisy1754:
:/ it's sad that this is being ignored, IMO such a simple fix that fixes easy-to-reproduce crash

Issue #28002, opened 2020-02-09
Cannot run program "npx": error=2, No such file or directory


Pratap22:
Description:

    Please provide a clear and concise description of what the bug is. Include screenshots if needed.
    Please test using the latest React Native release to make sure your issue has not already been fixed: http://facebook.github.io/react-native/docs/upgrading.html

Steps taken to start new fresh project.

- installed nvm
-  installed node using nvm
- initiated a new project : npx react-native init projectname
- Opened the project from Android-Studio and Got the following error

```
org.gradle.api.GradleScriptException:` A problem occurred evaluating settings 'ganasansaar'.
	at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory.java:92)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl$2.run(DefaultScriptPluginFactory.java:221)
	at org.gradle.configuration.DefaultScriptTarget.addConfiguration(DefaultScriptTarget.java:74)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.apply(DefaultScriptPluginFactory.java:226)
	at org.gradle.configuration.BuildOperationScriptPlugin$1$1.run(BuildOperationScriptPlugin.java:69)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:402)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:394)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:165)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:250)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:158)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:92)
	at org.gradle.internal.operations.DelegatingBuildOperationExecutor.run(DelegatingBuildOperationExecutor.java:31)
	at org.gradle.configuration.BuildOperationScriptPlugin$1.execute(BuildOperationScriptPlugin.java:66)
[...]

PR #28415, opened 2020-03-27
fix: Android gradle config when bundling for release

vdmtrv:
## Summary
This fix aims to address the issue when bundling an Android app for release and getting the error exhibited in #28002 which I also encountered myself.

The config was changed sometime in November 2019 (as part of #26940, commit https://github.com/facebook/react-native/commit/a3b08048674e324dbe1f0ca816f35607e9e06a2f) to be very opinionated when it comes to the use of `npx` which Gradle itself cannot find anyway (I have `npx` installed globally and it didn't pick it up).

Another issue that the use of `npx` creates is that Gradle should only ever use the currently installed react-native cli rather than a (possibly) higher version which may not always have backward compatibility.

The proposed change simply throws a more descriptive error rather than defaulting to a tool which may or may not exist on the machine, be it CI or a development environment. I've also modified the RNTester app to reflect the correct config implementation relative to the RNTester app itself.

In real projects, the config inside `android/app/build.gradle` should look similar to the following snippet:

```
project.ext.react = [
  cliPath: "$rootDir/../node_modules/react-native/cli.js",
  entryFile: "index.js"
];
```

## Changelog
[Android] [Fixed] - Gradle release config

## Test Plan
- [x] Successfully bundled an Android release build with correct config
- [x] Works with RNTester app

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -111,7 +111,7 @@ afterEvaluate {
             }
         } else {
             throw new Exception("Missing cliPath or nodeExecutableAndArgs from build config. " +
-                "Please set set project.ext.react.cliPath to the path of the react-native cli.js");
+                "Please set project.ext.react.cliPath to the path of the react-native cli.js");
         }
 
         def enableHermes = enableHermesForVariant(variant)
--- a/react.gradle
+++ b/react.gradle
@@ -110,7 +110,8 @@ afterEvaluate {
                 execCommand.addAll([*nodeExecutableAndArgs, cliPath])
             }
         } else {
-            execCommand.addAll([npx, "react-native"])
+            throw new Exception("Missing cliPath or nodeExecutableAndArgs from build config. " +
+                "Please set set project.ext.react.cliPath to the path of the react-native cli.js");
         }
 
         def enableHermes = enableHermesForVariant(variant)
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -65,6 +65,7 @@ plugins {
  */
 
 project.ext.react = [
+    cliPath: "../../../node_modules/react-native/cli.js",
     bundleAssetName: "RNTesterApp.android.bundle",
     entryFile: file("../../js/RNTesterApp.android.js"),
     root: "$rootDir",
--- a/RNTester/android/app/build.gradle
+++ b/RNTester/android/app/build.gradle
@@ -65,7 +65,7 @@ plugins {
  */
 
 project.ext.react = [
-    cliPath: "../../../node_modules/react-native/cli.js",
+    cliPath: "../../../cli.js",
     bundleAssetName: "RNTesterApp.android.bundle",
     entryFile: file("../../js/RNTesterApp.android.js"),
     root: "$rootDir",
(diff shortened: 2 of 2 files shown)


react.gradle
```diff
@@ -110,7 +110,8 @@ afterEvaluate {
                 execCommand.addAll([*nodeExecutableAndArgs, cliPath])
             }
         } else {
-            execCommand.addAll([npx, "react-native"])
+            throw new Exception("Missing cliPath or nodeExecutableAndArgs from build config. " +
+                "Please set set project.ext.react.cliPath to the path of the react-native cli.js");
```
hramos: Thank you for taking your time to improve this experience! Do you think we should mention `android/app/build.gradle` explicitly in the error message so it is clear where the config is located?

vdmtrv: You're correct. It's probably worth adding. I was simply following the convention on line 49:

```
throw new Exception("OS not recognized. Please set project.ext.react.hermesCommand " +
                        "to the path of a working Hermes compiler.");
```

If we include the path in this exception message then it needs to be added to the one on line 49.

hramos: I'm OK merging this as is, since it follows convention. I'm not as familiar with Gradle, so my main worry is whether these error messages are clear to the average Android developer. Would they know where to go to fix this config? Does the config location vary across workspaces or is it generally expected at the same path?

In other words, if the config is always at `android/app/build.gradle` then perhaps we can explicitly point to that file.

grabbou:
Thank you @vdmtrv and @hramos for working on this PR! In the future, please CC either me or @Esemesek on the CLI/Gradle related changes, so that we can help coordinate various work.

We have shipped a follow-up PR a few days after that which completely removed the usage of `npx` to fix this and a few other issues. We rely on pure `node` resolution mechanism, which works in all environments and is already used by the CLI to resolve additional modules from `node_modules`.

In this particular case, I'd consider this a downgrade in terms of the DX - we will now need to propagate `cliPath` to be added back to all React Native projects, including non-standard setups (such as mono-repositories) where this may be confusing. We never required this configuration, at least not in [`0.59`](https://github.com/facebook/react-native/blob/0.59-stable/template/android/app/build.gradle#L76).

That said, I am going to submit a follow-up PR which reverts the requirement of `cliPath` and aligns this with the rest of the CLI work. In other words, it will resolve the CLI path, but without using `npx` to do so.

grabbou:
I am going to send a follow-up PR as well to update code owners to better detect changes around that.

vdmtrv:
Thank you for providing more context, @grabbou!

I'll keep the points you raised in mind should I come across something similar in the future and CC you as requested. I was not aware that this was being worked on as it wasn't mentioned in any of the open issues related to the use of `npx`.

I did see `config.cliPath`, but the behaviour I was able to observe was that the default CLI path was not being respected. Because of this, it fell into the `else`, where it was using `npx`, which it couldn't find, hence the rationale behind the commit.

> We have shipped a follow-up PR a few days after that which completely removed the usage of npx to fix this and a few other issues.

Can we have a link to the commit/PR for reference, please?

Lastly, I think it's worth linking to the PR which is aiming to improve the above for anyone's future reference - https://github.com/facebook/react-native/pull/28625

grabbou:
No worries @vdmtrv, hard to stay on top of everything - we should update the code owners to better reflect that state. Appreciate time spent on debugging - we wouldn't find a fix if it wasn't for you.

Some additional context: https://github.com/react-native-community/cli/issues/886

PR #28435, opened 2020-03-29
Use GradleScriptException constructor with two parameters

IvanShafran:
## Summary

GradleScriptException used to have a constructor with one string parameter. But in the latest 6.3 version, it has a constructor with two parameters. [GradleScriptException  docs](https://docs.gradle.org/current/javadoc/org/gradle/api/GradleScriptException.html)

It led to Gradle sync errors in Android Studio.

`caused by: groovy.lang.groovyruntimeexception: could not find matching constructor for: org.gradle.api.gradlescriptexception(string)`

## Changelog

[Internal] [Fixed] - Fix gradle sync error. Use GradleScriptException constructor with two parameters

## Test Plan

Open root folder in Android Studio without calling `yarn install`. During the sync phase, it can't find Hermes and it tries to throw GradleScriptException with one parameter. Android Studio shows:
`caused by: groovy.lang.groovyruntimeexception: could not find matching constructor for: org.gradle.api.gradlescriptexception(string)`.

Thank you :)

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -98,12 +98,14 @@ task prepareFolly(dependsOn: dependenciesPath ? [] : [downloadFolly], type: Copy
 task prepareHermes() {
     def hermesPackagePath = findNodeModulePath(projectDir, "hermes-engine")
     if (!hermesPackagePath) {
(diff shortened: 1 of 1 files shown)


sota000:
The change has been made in the files. Closing the PR.

Issue #28446, opened 2020-03-30
iOS: Compile fails when gflags is installed


KDederichs:
Please provide all the information requested. Issues that do not follow this format are likely to stall.

## Description
When trying to build an iOS version of a react native app, the compilation of the `glog` (and `Flipp-er-Glog`) pods will fail if the machine has the gflags package installed (`brew install gflags`).
This seems to be due to the configure script picking up the `gflags.h` file and in turn enabling this line in logging.h by setting the if precompile flag to true:

```
#if 0
#include <gflags/gflags.h>
#endif
```

## React Native version:

```
System:
    OS: macOS 10.15.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 16.84 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.14.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 24, 25, 26, 27, 28, 29
      Build Tools: 28.0.3, 29.0.2
      System Images: android-29 | Google Play Intel x86 Atom
  IDEs:
    Xcode: 11.4/11E146 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5
```

This bug should hit all existing RN versions that relay on glog and starting RN 62 also the Flipper integration.

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.

1. install gflags via `brew install gflags`
2. try to compile a fresh react native project that was setup AFTER! you installed `gflags`

## Expected Results
The app should still compile regardless if the user has setup gflags or not.

## Snack, code example, screenshot, or link to a repository:
Not needed

KDederichs:
Still in the Backlog

KDederichs:
I'd assume it's still an issue

waqas19921:
[...]

PR #28451, opened 2020-03-30
Disable gflags include

KDederichs:
## Summary

Fixes the issue explained in https://github.com/facebook/react-native/issues/28446
It basically disabled the gflags include before configure can detect the header on the users system.

## Changelog

[iOS] [Fixed] - Fixed inability to build apps when gflags is installed

## Test Plan

Tested by installing gflags `brew install gflags` and verifying that apps build after.


scripts/ios-configure-glog.sh
```diff
@@ -28,6 +28,10 @@ if [ -h "test-driver" ]; then
     rm test-driver
 fi
 
+# Manually disable gflags include.
```
sota000, line 31: Since this is a bit hacky fix, can you please add a link to the issue to the comment to ensure the context is shared?

sota000 (changes requested):
Hi @KDederichs, apologies for taking the time to review. Could you please address the comment I made inline and solve the merge conflict? Thank you for the contribution!

scripts/ios-configure-glog.sh
KDederichs: Done.

KDederichs:
No worries, projects of this size get a lot of contribution so I understand that small stuff like this lies around for a while.
I added the issue link to the comment, I didn't see any conflicts though when merging upstream into the fix branch, so I assume that's good?

sota000 (approved):
Thanks for updating!

PR #28491, opened 2020-04-02
Make XMLHttpRequest timeout behavior consistent on Android and iOS

jsying13:
## Summary

Use callTimeout instead of connectTimeout

### Problem: 
the timeout setting of XMLHttpRequest doesn't work.
### Demo code:

```
const log = (s) => {
  let ts = new Date().toISOString()
  console.log("[" + ts + "] " + s)
}

const App: () => React$Node = () => {

  useEffect(() => {
    let request = new XMLHttpRequest()
    request.timeout = 3210
    request.onload = (e) => {
      log("onload")
      log("response: " + e.target.responseText)
    }
    request.ontimeout = () => {
      log("request ontimeout")
    }
    request.onerror = (e) => {
      log("request onerror")
    }
    request.open("GET", "https://httpbin.org/delay/6")
    request.send()
    log("request sent")
  })
```

### On Android
Screen Shot 2020-03-22 at 12 14 44 AM

### On iOS
Screen Shot 2020-03-22 at 12 18 22 AM

## Changelog

[Android] [Fixed] - Fix XMLHttpRequest timeout

## Test Plan

Run the same demo code and check that the timeout event is fired on Android

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/network/ResponseUtil.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/network/ResponseUtil.java
@@ -12,7 +12,7 @@ import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.modules.core.DeviceEventManagerModule.RCTDeviceEventEmitter;
-import java.net.SocketTimeoutException;
+import java.io.InterruptedIOException;
 
 /** Util methods to send network responses to JS. */
 public class ResponseUtil {
@@ -84,7 +84,7 @@ public class ResponseUtil {
     args.pushInt(requestId);
     args.pushString(error);
 
-    if ((e != null) && (e.getClass() == SocketTimeoutException.class)) {
+    if ((e != null) && (e.getClass() == InterruptedIOException.class)) {
       args.pushBoolean(true); // last argument is a time out boolean
     }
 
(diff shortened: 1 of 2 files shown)


jnurkka:
@shadow-tw thanks for this PR. We've noticed exactly the same issue, I wrote a stackoverflow question about it some time ago [here](https://stackoverflow.com/questions/64120002/correctly-timing-out-xhr-requests-in-react-native-on-android). 

I think your description of this issue undersells the importance of this change, though. In my opinion, just using connectTimeout on android is not only inconsistent but quite frankly a critical bug. As a developer, you'd never intuitively expect that the timeout you specify for your API call only applies to connecting to the server, but to the whole call being completed. In our case, this bug has lead to "broken connections" staying open for hours, leading to consequent API calls to the same server also taking forever before failing.

We actually had to work around this by keeping track of the duration of HTTP requests and aborting them if we notice that they have taken longer than allowed. But because our app also makes HTTP requests in the background state, where setTimeout is not available, this can lead to requests staying open until the next time we can run our scheduled background task.

So I'm hoping this PR can be merged ASAP before even more developers are tricked into believing that timeout works how they would expect it to work. I can really see this happening even more when [this](https://github.com/facebook/react-native/pull/28938) (or some other) PR gets merged to allow specifying timeouts more easily

ReactAndroid/src/main/java/com/facebook/react/modules/network/ResponseUtil.java
```diff
@@ -84,7 +84,7 @@ public static void onRequestError(
     args.pushInt(requestId);
     args.pushString(error);
 
-    if ((e != null) && (e.getClass() == SocketTimeoutException.class)) {
+    if ((e != null) && (e.getClass() == InterruptedIOException.class)) {
```
jnurkka, line 87: this will return false on timeout, should use `java.io.IOException`

PR #28504, opened 2020-04-03
Exclude all FlipperKit transitive dependencies from iOS Release builds

javiercr:
## Summary

The `:configuration` option from `pod` only affects the specified pod and not its dependencies [1]. Therefore in order to avoid all transitive dependencies being linked in the resulting Release IPA we need to list them in the `Podfile`.

Note that this will still build Flipper's pods when doing a Release, but it won't link it in the resulting IPA.

[1] https://guides.cocoapods.org/syntax/podfile.html#pod

Fixes https://github.com/react-native-community/upgrade-support/issues/28
Related https://github.com/CocoaPods/CocoaPods/issues/9658

## Changelog

* [iOS] [Fixed] - Exclude Flipper from iOS Release builds

## Test Plan

Create a new React Native 0.62 project, run `pod install`, then diff:
```
ProjectName/ios/Pods/Target Support Files/Pods-ProjectName/Pods-ProjectName.debug.xcconfig`
```
and 
```
ProjectName/ios/Pods/Target Support Files/Pods-ProjectName/Pods-ProjectName.relaese.xcconfig
```

image

### DIFF
--- a/scripts/react_native_pods.rb
+++ b/scripts/react_native_pods.rb
@@ -63,6 +63,23 @@ def use_flipper!(version = '~> 0.33.1')
   pod 'FlipperKit/SKIOSNetworkPlugin', version, :configuration => 'Debug'
   pod 'FlipperKit/FlipperKitUserDefaultsPlugin', version, :configuration => 'Debug'
   pod 'FlipperKit/FlipperKitReactPlugin', version, :configuration => 'Debug'
+
+  # List all transitive dependencies for FlipperKit pods
+  # to avoid them being linked in Release builds
+  pod 'Flipper', version, :configuration => 'Debug'
+  pod 'Flipper-DoubleConversion', '1.1.7', :configuration => 'Debug'
+  pod 'Flipper-Folly', '~> 2.1', :configuration => 'Debug'
+  pod 'Flipper-Glog', '0.3.6', :configuration => 'Debug'
+  pod 'Flipper-PeerTalk', '~> 0.0.4', :configuration => 'Debug'
+  pod 'Flipper-RSocket', '~> 1.0', :configuration => 'Debug'
+  pod 'FlipperKit/Core', version, :configuration => 'Debug'
+  pod 'FlipperKit/CppBridge', version, :configuration => 'Debug'
+  pod 'FlipperKit/FBCxxFollyDynamicConvert', version, :configuration => 'Debug'
+  pod 'FlipperKit/FBDefines', version, :configuration => 'Debug'
+  pod 'FlipperKit/FKPortForwarding', version, :configuration => 'Debug'
+  pod 'FlipperKit/FlipperKitHighlightOverlay', version, :configuration => 'Debug'
+  pod 'FlipperKit/FlipperKitLayoutTextSearchable', version, :configuration => 'Debug'
+  pod 'FlipperKit/FlipperKitNetworkPlugin', version, :configuration => 'Debug'
 end
 
 # Post Install processing for Flipper
--- a/scripts/react_native_pods.rb
+++ b/scripts/react_native_pods.rb
@@ -57,29 +57,36 @@ def use_react_native! (options={})
   end
 end
 
-def use_flipper!(version = '~> 0.33.1')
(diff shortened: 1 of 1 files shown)


scripts/react_native_pods.rb
```diff
@@ -63,6 +63,23 @@ def use_flipper!(version = '~> 0.33.1')
...
   pod 'FlipperKit/FlipperKitReactPlugin', version, :configuration => 'Debug'
+
+  # List all transitive dependencies for FlipperKit pods
+  # to avoid them being linked in Release builds
+  pod 'Flipper', version, :configuration => 'Debug'
+  pod 'Flipper-DoubleConversion', '1.1.7', :configuration => 'Debug'
+  pod 'Flipper-Folly', '~> 2.1', :configuration => 'Debug'
+  pod 'Flipper-Glog', '0.3.6', :configuration => 'Debug'
+  pod 'Flipper-PeerTalk', '~> 0.0.4', :configuration => 'Debug'
+  pod 'Flipper-RSocket', '~> 1.0', :configuration => 'Debug'
```
javiercr, line 74: I wasn't sure what to do about version numbers for the transitive dependencies. If the FlipperKit pods change its dependencies to a different version number, we'll need to update these.

I exposed this problem in the cocoapods repo:
https://github.com/CocoaPods/CocoaPods/issues/9658#issuecomment-608260488

For the time being I can't find another solution to this.

alloy: I guess the parameter to the `use_flipper!` method should take a single options hash and then apply defaults. Something like:

```ruby
def use_flipper!(versions = {})
  versions['Flipper'] ||= '~> 0.33.1'
  versions['DoubleConversion'] ||= '1.1.7'
  # etc

  pod 'Flipper', versions['Flipper'], :configuration => 'Debug'
  pod 'Flipper-DoubleConversion', versions['DoubleConversion'], :configuration => 'Debug'
  # etc
end
```

alloy:
I think I’ll also port this to the `0.62-stable` branch and get it in today’s `0.62.1` release.

scripts/react_native_pods.rb
javiercr: @alloy I just made those changes, let me know if there is anything else you would like to see :)

alloy: Ace, thanks! 👌

alloy:
@passy Can you get this in?

Ashoat:
Is there a reason this doesn't cover the following transitive dependencies:

- `CocoaAsyncSocket`
- `CocoaLibEvent`
- `OpenSSL-Universal`
- `YogaKit`

alloy:
@Ashoat Not as far as I know. Can you make a PR for it?

Ashoat:
My only worry is that in contrast with the other excluded transitive dependencies, which are definitely Flipper-specific, these four could be required by other pods. It appears that applying `:configuration => 'Debug'` will affect the given dependency wherever it appears in the dependency tree (otherwise this PR would not seem to work), which makes me worry that if another pod depends on these dependencies in release builds, it would end up breaking. So I'm reluctant to put up a PR that simply excludes them without understanding a bit more.

Ideally we would have a mechanism to exclude them from release builds only if they don't appear anywhere else in the dependency tree, but I'm not familiar enough with CocoaPods to know if something like that is possible.

javiercr:
> My only worry is that in contrast with the other excluded transitive dependencies, which are definitely Flipper-specific, these four could be required by other pods. It appears that applying `:configuration => 'Debug'` will affect the given dependency wherever it appears in the dependency tree (otherwise this PR would not seem to work), which makes me worry that if another pod depends on these dependencies in release builds, it would end up breaking. 

I think that's why I didn't include them in this PR. I could only exclude (actually include) the ones that I knew for sure were dependencies only for Flipper (basically the ones with Flipper in the name).

alloy:
The best solution would be to take this CocoaPods PR to completion https://github.com/CocoaPods/CocoaPods/pull/9066. It’s on my TODO list, but I have no idea how long it would take before I get to that, so if there are other takers please do take a shot.

PR #28506, opened 2020-04-03
eslint-config: add version badge and add homepage for eslint-config

Naturalclar:
## Summary
Add version badge to README of eslint-config, and add specific url for the homepage so people looking at the npm package can find out where the package is from.

## Changelog

[Internal] [Changed] - Add version badge to README of eslint-config

## Test Plan
Not required as the only changes are made in README and homepage prop of package.json

### DIFF
--- a/packages/eslint-config-react-native-community/README.md
+++ b/packages/eslint-config-react-native-community/README.md
@@ -1,5 +1,7 @@
 # eslint-config-react-native-community
 
+[![Version][version-badge]][package]
+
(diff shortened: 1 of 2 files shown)


Naturalclar:
@matt-oakes 
>  I presume that NPM will only pick up the new homepage when a new version is published?

Yes, npm page only updates when a new version is published.
not sure if it's worth publishing a new version with just this change, I'll leave that up to you :)

PR #28523, opened 2020-04-04
Make facebook::react::Instance methods overridable

JunielKatarn:
### Public methods of class `facebook::react::Instance` that interface with NativeNoJsBridge become virtual, making the class mockable

Currently, custom types (i.e. deriving `facebook::xplat::module::CxxModule` or `facebook::react::JSExcecutor`) that need to interact with the `facebook::react::Instance` can't be fully unit tested without bootstrapping a full ReactNative app.

This change aims to make some of the class's public methods virtual so mock implementations of `facebook::react::Instance` can be written and any interaction with this class can be simulated in isolation, which will allow for complete and lightweight unit tests, increasing code coverage.

## Changelog: 

[General] [Changed] - Allow facebook::react::Instance to be extended for broader JSExecutor and Module unit testing / mocking.

## Test Plan:

- Build
- Run unit and integration tests.


shergin:
I am sorry, but, IMHO, a particular approach to testability is not a good reason to make methods `virtual`.

JunielKatarn:
> I am sorry, but, IMHO, a particular approach to testability is not a good reason to make methods `virtual`.

@shergin What is the drawback of making these methods virtual?
Testability helps writing better modules, and, to my knowledge, there is no noticeable risk or performance implication.

Can you share your concerns?

shergin:
The performance implications are real (virtual methods are being called using dynamic dispatch (vtable) whereas normal ones are just direct calls; calling virtual functions cannot be inlined) but, besides that, the question is more about using idiomatic C++.

In C++, if a method is marked `virtual`, it gives a strong signal to a reader that the method is being called on a pointer to a base class, that the method is being called from a parent class in order to delegate the execution to a super class, and/or that the method is part of some interface (abstract class). All that begs questions like "why is this done this way?", "what was the intent?", "what am I missing reading this?". In short, making a method `virtual` is a very important design decision.

And usually the testability is not a good answer to that because we usually want to test everything, and that implies that we should add `virtual` to everything?.. If we do so, that will not be real C++ anymore.

I think we should consider another approach to mocking (e.g. replacing real classes with mocks using build tools).

JunielKatarn:
> And usually the testability is not a good answer to that because we usually want to test everything, and that implies that we should add `virtual` to everything?.. If we do so, that will not be real C++ anymore.

For this purposes, struct interfaces are used.
I'd argue it would be better to have a supertype of `Instance`,s say, `IInstance`.
I do wonder if the `vtable` performance impact is relevant for this case, but haven't tried any measurements.

matthargett:
Using virtual methods to mock for unit-testability is fairly common, but I understand the concern about performance impacts as well.

Using virtual methods does not automatically mean an indirect call via viable ends up in the generated code. Both LLVM and GCC can both devirtualize these calls when they have visibility of the concrete type being passed into other parts of the program via function/method/ctor arguments. 

Can you show a man objdump of the dispatches to these now-virtual  methods to see if if the compiler has correctly applied this optimization? If it has, then that should allay Valentin’s concerns. FWIW, when I was tuning folly and C++ RN on my last project, I found that LLVM correctly devirtualized but I was using Link Time Optimization and Whole Program Optimization, so YMMV when using the BUCK build and whatever compiler version is included in the NDKs.

shergin:
I am way more concerned about the readability and simplicity of the code than actual runtime performance *in this particular case* (which will probably be good even if the calls are not devirtualized because it's not a hot path).

The framework has many places where virtualization is not allowed because of actual performance overhead though. To test those places we need to figure out something different anyway. That's why I lean towards compile-time approaches. 

If we believe that using virtual dispatch is the right approach here, we should start from designing meaningful interfaces to which Instance would conform (improving code to make it testable). I mean, now it's 11 methods, even if we will introduce `Interface` interface, `ConcreteInterface` implementation and `MockInterface` mock, It's hard to believe that it's a path to the right direction (towards achieving confidence in the code) because we are increasing complexity which is already pretty high.

Issue #18990, opened 2018-04-23
TextInput Chinese and English  length limit


RoarRain:
## Environment

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.9.3
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.7.0
  Xcode: Xcode 9.0 Build version 9A235
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.0.0-alpha.12 => 16.0.0-alpha.12
  react-native: 0.48.0 => 0.48.0

## Steps to Reproduce

I want to limit  Chinese and English length at the same time，see https://blog.csdn.net/m372897500/article/details/25820353  and want  to get keyBoard type for TextInput。
maxLength props is not best，when i typing Pinyin，it is limit.

## Expected Behavior

can get keyBoard type

PR #28539, opened 2020-04-06
fix: Chinese or other nonalphabetic language maxLength bug on iOS

aoarashi1988:
## Summary
When input Chinese or Japanese even other nonalphabetic language word , the calculation of text length is wrong, you use the alphabet in TextInput instead of the real word they want to input。
It's an old issue, but still remain in the up-to-date version of react-native, just like the issue  #18990. 

## Changelog
Take markedTextRange into consideration.

[iOS] [fixed] - Chinese or other nonalphabetic language maxLength limit bug

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -384,6 +384,12 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
   if (_maxLength) {
     NSInteger allowedLength = MAX(_maxLength.integerValue - (NSInteger)backedTextInputView.attributedText.string.length + (NSInteger)range.length, 0);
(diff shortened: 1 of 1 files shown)


TheSavior:
Does this issue repro with multi character length emojis as well? If so, can you provide a small repro code?

Having a repro case for this that works when the device language is english would be super helpful as we can put it in the RNTester test suite and make sure it doesn't break in the future.

sota000:
@aoarashi1988, I know this is a year old pull request, but would you be able to check on the CLA and @TheSavior's comment so that we can review this internally?

sota000:
Hi @aoarashi1988, I am going to close this for now to focus on other active PRs, but please reopen if you are able to address the comments above. Thank you for your contribution!

KingAmo:
related PR https://github.com/facebook/react-native/pull/31376
any updates please ?

PR #28545, opened 2020-04-06
Upgrade Flipper to 0.37.0

sunnylqm:
## Summary

Bump flipper to 0.37 for both iOS and Android

## Changelog

[Android] [Changed] - Upgrade Flipper to 0.37.0
[iOS] [Changed] - Upgrade Flipper to 0.37.0

## Test Plan

RNTester build pass

### DIFF
--- a/RNTester/Podfile.lock
+++ b/RNTester/Podfile.lock
@@ -11,11 +11,11 @@ PODS:
     - React-Core (= 1000.0.0)
     - React-jsi (= 1000.0.0)
     - ReactCommon/turbomodule/core (= 1000.0.0)
-  - Flipper (0.36.0):
-    - Flipper-Folly (~> 2.1)
-    - Flipper-RSocket (~> 1.0)
+  - Flipper (0.37.0):
+    - Flipper-Folly (~> 2.2)
+    - Flipper-RSocket (~> 1.1)
   - Flipper-DoubleConversion (1.1.7)
-  - Flipper-Folly (2.1.1):
+  - Flipper-Folly (2.2.0):
     - boost-for-react-native
     - CocoaLibEvent (~> 1.0)
     - Flipper-DoubleConversion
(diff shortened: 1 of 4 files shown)


hramos:
@sunnylqm FYI, I'm currently looking into the `test_ios_detox_frameworks` and `test_ios_unit_frameworks` failures. They're both failing on the master branch so your changes should be unrelated to this failure.

hramos:
@sunnylqm let's fix the conflict (introduced when https://github.com/facebook/react-native/pull/28504 was merged) and then I'll pull this in.

sunnylqm:
@hramos Rebased

sunnylqm:
Updated to 0.37.0

guhungry:
This will fix the crash in ios when use base64 image https://github.com/facebook/react-native/issues/28583

hramos:
I've been looking at this on and off, but haven't been able to get past the following when running `pod install` in the `RNTester/` directory:

```
[!] CocoaPods could not find compatible versions for pod "FlipperKit/FlipperKitReactPlugin":
  In Podfile:
    FlipperKit/FlipperKitReactPlugin (~> 0.37.0)

None of your spec sources contain a spec satisfying the dependency: `FlipperKit/FlipperKitReactPlugin (~> 0.37.0)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
```

Any ideas?

sunnylqm:
pod install --repo-update

hramos:
I could swear I had tried that already, as it's in the error message, but alas, it did the trick.

We have a copy of Pods checked in to the internal monorepo, so I suspect I was running into issues updating the internal Pods mirror. Trying again...

PR #28566, opened 2020-04-09
StickyHeaders can throw error when layoutY is smaller than 0

jfrolich:
Small fix, input ranges have to be monotonically increasing, this code path didn't test for this possible condition.

## Changelog
Fix a bug that section list can throw an error in certain conditions related to StickyHeaders.

### DIFF
--- a/Libraries/Components/ScrollView/ScrollViewStickyHeader.js
+++ b/Libraries/Components/ScrollView/ScrollViewStickyHeader.js
@@ -119,7 +119,7 @@ class ScrollViewStickyHeader extends React.Component<Props, State> {
         //   header to continue scrolling up and make room for the next sticky header.
         //   In the case that there is no next header just translate equally to
         //   scroll indefinitely.
-        inputRange.push(layoutY);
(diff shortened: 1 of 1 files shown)


hramos:
@jfrolich thanks for the fix! Since I'm not familiar with the issue that is being fixed, would you be able to add a Test Plan to your PR description? It would help us understand the behavior before and after the fix is applied.

jfrolich:
I encounter this scenario with a `ListHeaderComponent` where the height of the element is animated. In that scenario the value of `layoutY` becomes negative.  I am not sure what you mean by a test plan, perhaps make sure that a negative value of `layoutY` case a crash?

sota000:
Hi @jfrolich, I know this is a year old PR, but I'll need some clarification before bringing this to an internal review. As for the test plan, I'd like to see how I (or anyone reading it) can verify that this change fixes the bug and doesn't break the intended behavior. This can be a combination of steps to repro, a video, or/and unit test. Can you think of a good way to verify this change?

PR #28579, opened 2020-04-10
Upgrade flow-remove-types version to 2.122.0

empyrical:
## Summary

This pull request upgrades `flow-remove-types` to version  `2.122.0`.
 
`flow-node` in the older version was not able to work with the Codegen files that `scripts/generate-rncore.sh` depended on.

## Changelog

[Internal] [Changed] - Upgraded flow-remove-types version to 2.122.0

## Test Plan

`scripts/generate-rncore.sh` now runs without issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "eslint-plugin-react-native": "3.8.1",
     "eslint-plugin-relay": "1.7.0",
     "flow-bin": "^0.122.0",
-    "flow-remove-types": "1.2.3",
+    "flow-remove-types": "^2.122.0",
     "jest": "^24.9.0",
     "jest-junit": "^6.3.0",
     "jscodeshift": "^0.7.0",


TheSavior:
How is this library currently used? I don't see any imports to `flow-remove-types`

empyrical:
The CLI tool `flow-node`, which is used by `scripts/generate-rncore.sh`, is a part of `flow-remove-types`

TheSavior:
Ah. got it. Just a heads up that there is some tangles with this package and the facebook monorepo that will have to get resolved to be able to land this. I'm not planning on working on that maybe @dvacca will, so just letting you know that this might take a bit to land.

empyrical:
Understandable. Are you able to say what they are? Perhaps I can figure out another solution to this.

hramos:
Do we still need this change, now that the codegen package uses Babel and a build script to work around the need for flow-remove-types?

empyrical:
Nope, I forgot to close this

PR #28603, opened 2020-04-12
Adds an error message in the Alert.alert method if buttons is not an array

lodi-g:
## Summary

I've found that developers (especially beginners) will often provide an object as the third argument to `Alert.alert` if they only want one button. This provides a cleaner error message instead of making the application crash a few lines later (in the `buttons.slice` method for Android).

## Changelog

[General] [Changed] - Alert.alert will now throw an error if the third argument is not an array.

## Test Plan

Prettier + eslint + jest. RNTester.

### DIFF
--- a/Libraries/Alert/Alert.js
+++ b/Libraries/Alert/Alert.js
@@ -47,6 +47,10 @@ class Alert {
     buttons?: Buttons,
     options?: Options,
   ): void {
+    if (buttons && !Array.isArray(buttons)) {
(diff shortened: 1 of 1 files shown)


TheSavior:
@lodi-g thanks for the PR. There are a lot of APIs in React Native that could have validation like this. However, we've been moving away from this runtime validation as it negatively impacts performance. We've been recommending people use and rely on TypeScript or Flow to ensure this type of validation instead.

PR #28625, opened 2020-04-14
fix: do not throw on missing `cliPath`, use the default value

grabbou:
## Summary

The `cliPath` has always been optional value and in fact, even had its default value hardcoded in the React gradle file.

In this PR, I am just taking use of it and remove throwing an error, which is going to be a really annoying breaking change.

## Changelog

[ANDROID] [INTERNAL] - Don't require `cliPath`

## Test Plan

Run Android project, everything works. 
Provide custom `cliPath`, it gets respected

### DIFF
--- a/react.gradle
+++ b/react.gradle
@@ -21,7 +21,22 @@ def detectEntryFile(config) {
     return "index.js";
 }
 
-def defaultCliPath = "node_modules/react-native/cli.js"
+/**
+ * Detects CLI location in a similar fashion to the React Native CLI
+ */
+def detectCliPath(config) {
+    def cliPath = ["node", "-e", "console.log(require('react-native/cli').bin);"].execute([], projectDir).text
+
+    if (cliPath) {
+        return cliPath
+    } else if (new File("node_modules/react-native/cli.js").exists()) {
+        return "node_modules/react-native/cli.js"
+    } else {
+        throw new Exception("Couldn't determine CLI location. " +
+            "Please set `project.ext.react.cliPath` to the path of the react-native cli.js");
+    }
+}
+
 def composeSourceMapsPath = config.composeSourceMapsPath ?: "node_modules/react-native/scripts/compose-source-maps.js"
 def bundleAssetName = config.bundleAssetName ?: "index.android.bundle"
 def entryFile = detectEntryFile(config)
@@ -109,7 +124,7 @@ afterEvaluate {
                 execCommand.addAll([*nodeExecutableAndArgs, config.cliPath])
             }
         } else {
-            execCommand.add(defaultCliPath)
+            execCommand.add(defaultCliPath())
         }
 
         def enableHermes = enableHermesForVariant(variant)
--- a/react.gradle
+++ b/react.gradle
@@ -26,10 +26,10 @@ def detectEntryFile(config) {
  */
 def detectCliPath(config) {
     def cliPath = ["node", "-e", "console.log(require('react-native/cli').bin);"].execute([], projectDir).text
-
+    
     if (cliPath) {
         return cliPath
-    } else if (new File("node_modules/react-native/cli.js").exists()) {
+    } else if (new File("${projectDir}/../../node_modules/react-native/cli.js").exists()) {
         return "node_modules/react-native/cli.js"
     } else {
         throw new Exception("Couldn't determine CLI location. " +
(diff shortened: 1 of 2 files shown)


grabbou:
CC: @hramos @vdmtrv @Esemesek @thymikee

react.gradle
```diff
@@ -104,13 +104,12 @@ afterEvaluate {
...
-                execCommand.addAll(["cmd", "/c", *nodeExecutableAndArgs, cliPath])
+                execCommand.addAll(["cmd", "/c", *nodeExecutableAndArgs, config.cliPath])
             } else {
-                execCommand.addAll([*nodeExecutableAndArgs, cliPath])
+                execCommand.addAll([*nodeExecutableAndArgs, config.cliPath])
             }
         } else {
-            throw new Exception("Missing cliPath or nodeExecutableAndArgs from build config. " +
-                "Please set project.ext.react.cliPath to the path of the react-native cli.js");
+            execCommand.add(defaultCliPath)
```
thymikee, line 112: shouldn't we distinguish between Windows and the rest here as well?

grabbou:
Next step would be to replace `defaultCliPath` with what we have done for React Native CLI already:
https://github.com/react-native-community/cli/blob/master/packages/platform-android/native_modules.gradle#L220-L222

In other words, run `console.log(require('react-native/cli').bin);` from the Gradle project root, which will point us to the closest `node_modules` location containing the CLI.

We already use this logic in other parts of Gradle build process, so it's proven to work.

react.gradle
grabbou: Yeah, probably, we need a bit more work here. @Esemesek

grabbou:
If `rootProject.projectDir` is defined here as well, then, we can use the same mechanism to learn about the path. This will be better than using hardcoded version, which isn't going to work in most cases.

react.gradle
vdmtrv: Don't we also need `nodeExecutableAndArgs` in here?

grabbou:
Shipped a simple function that tries to detect the CLI path by running CLI command. If it fails (it will not work in React Native source because there's no `react-native` package there), it will try to use the default path (which was present from the very beginning and covers most of the cases). Lastly, it will throw the error that we throw right now, asking to provide the path explicitly.

Just want to point out that in Groovy, if the process fails with non-zero exit code, the `text`/`getText()` will be empty.

grabbou:
Wanted to point out that we use the same logic here as in `detectEntryPoint` https://github.com/facebook/react-native/blob/1b5d53a2fa0101fd9a5db98410fc5ea89354836d/react.gradle#L12.

grabbou:
@vdmtrv would you mind checking if this works for you? Should be a matter of copying-pasting to your `node_modules`.

react.gradle
grabbou: I have rewrote the logic a bit to support both cases

grabbou:
I have confirmed this works on a brand new project (limited with 0.63 branch), the `cliPath` has the following location:
> /private/tmp/RNTestProject/node_modules/@react-native-community/cli/build/bin.js
resolved by the `node` script.

The fallback location is:
> /private/tmp/RNTestProject/android/app/../../node_modules/react-native/cli.js

which also works in case `node` script would fail.

react.gradle
```diff
@@ -21,7 +21,26 @@ def detectEntryFile(config) {
...
-def cliPath = config.cliPath ?: "node_modules/react-native/cli.js"
+/**
+ * Detects CLI location in a similar fashion to the React Native CLI
+ */
+def detectCliPath(config) {
+    if (config.cliPath) {
+        return config.cliPath
+    }
+
+    def cliPath = ["node", "-e", "console.log(require('react-native/cli').bin);"].execute([], projectDir).text
```
vdmtrv, line 32: Does this mean that we have to have `react-native` installed as a global dependency?

grabbou: Nope,  we're running this from `projectDir` context, which is a directory inside your project. Will resolve to local `node_moduls`.

grabbou: Can we connect somewhere in a bit more "direct" way to quickly resolve this? If you're on Twitter or Discord, we can exchange DMs, put few console.logs here and there and see what's going on.

I'm @grabbu or grabbou@gmail.com - will be online for next two hours. Let's try tackling that!

PR #28626, opened 2020-04-14
chore: remove Kotlin version from the default template

grabbou:
## Summary

The default application requires Kotlin version that is not supported by the Gradle plugin (should be at least `1.3.10`). However, instead of upgrading, we should remove it entirely. Here's why.

This commit https://github.com/facebook/react-native/commit/29d3dfbd196176af98c9727c82ff2668e697d78e introduced Detox for RNTester Android application.

Since the commit doesn't mention Detox for the default application and there are no Detox tests present by default in the default application, I believe that this addition was performed by a mistake.

The best way is to remove Kotlin from the default template. This step is described in the Detox documentation and all users that are integrating Detox will be asked to perform it anyway. No need to do it for them.

## Changelog

[ANDROID] [INTERNAL] - remove Kotlin from the default template

## Test Plan

Building a brand new project with `master` should work

### DIFF
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         google()
         jcenter()
     }
-    ext.kotlinVersion = '1.3.0'
+    ext.kotlinVersion = '1.3.10'
     dependencies {
         classpath("com.android.tools.build:gradle:3.5.3")
         // for Detox
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -11,11 +11,8 @@ buildscript {
         google()
         jcenter()
     }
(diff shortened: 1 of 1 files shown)


grabbou:
CC: @cs01 please correct me if I am wrong :)

cs01:
As long as `yarn run detox build -c android.emu.debug` and `yarn run detox build -c android.emu.debug` still succeed I am fine with this.

grabbou:
Yes, I have only removed that from the default template. No changes to RNTester.

I also run the commands you suggested:

> As long as yarn run detox build -c android.emu.debug and yarn run detox build -c android.emu.release still succeed I am fine with this.

and all finished without Gradle build error (BUILD SUCCEEDED).

PR #28634, opened 2020-04-15
enable ArrayBuffer API in JSCRuntime

wkozyra95:
## Summary

Currently, there is no way to interact with TypedArrays and ArrayBuffers using C++ API.

Comment `T23270523 - This would fail on builds that use our custom JSC` is no longer valid
- on android old JSC was replaced with support for 64-bit builds
- on iOS old JSC was replaced when support for iOS 9 was dropped

## Changelog

[General] [Added] - Add/Enable implementation for ArrayBuffer jsi API

## Test Plan

- I didn't test ArrayBuffer API with this PR in runtime, but I used those exact changes and few others [here](https://github.com/expo/expo/blob/%40wkozyra95/jsi-expo-gl-part5/android/ReactCommon/jsi/JSCRuntime.cpp)

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -922,24 +922,18 @@ bool JSCRuntime::isArray(const jsi::Object &obj) const {
 #endif
 }
 
-bool JSCRuntime::isArrayBuffer(const jsi::Object & /*obj*/) const {
-  // TODO: T23270523 - This would fail on builds that use our custom JSC
-  // auto typedArrayType = JSValueGetTypedArrayType(ctx_, objectRef(obj),
-  // nullptr);  return typedArrayType == kJSTypedArrayTypeArrayBuffer;
-  throw std::runtime_error("Unsupported");
+bool JSCRuntime::isArrayBuffer(const jsi::Object &obj) const {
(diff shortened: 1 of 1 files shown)


shergin:
@@wkozyra95 Could you please sign the CLA?

wkozyra95:
Hi @shergin
I need to be added to my company CLA but, website does not support adding people if CLA was signed already, so my employer messaged cla@fb.com to add me. It was a few days ago(before I created this PR) and I'm still waiting for a response.

wkozyra95:
@shergin CLA is signed now, sorry it took so long

wkozyra95:
Hi
Can you share what is the actual error? I'm modifying functions that threw exceptions, so I don't suppose there is a bug there that existing tests could detect?

We need that feature for WebGL implementation inside Expo Client, without that API there is no way to use TypedArray on c++ side efficiently.

tsapeta:
Hey @shergin, we at Expo would love to be able to add Hermes support to `expo-gl-cpp` module, but in order to move forward with JSC -> JSI migration we will need these changes to land. Is there anything that we can do to get an approval and make it merged? Thanks in advance 🙂

wkozyra95:
Fixed in https://github.com/facebook/react-native/pull/28961

PR #28641, opened 2020-04-15
[DO NOT MERGE] Fix CIRCLE_PR_NUMBER may not always be set

TheSavior:
Testing that https://github.com/facebook/react-native/pull/28640 fixes the jobs for PRs opened from branches on the same repo (not a fork).


TheSavior:
Looks like this worked. Yay!

PR #28645, opened 2020-04-15
Codegen: Add prepublish script to build Flow files

empyrical:
## Summary

This pull request adds a build step to `react-native-codegen` that builds the Flow-annotated JS files so that users of the NPM module `react-native-codegen` do not need to use require hooks to be able to import it.

The `flow-remove-types` CLI script is used to build the JS files in `src/` and it outputs them into `lib/`. Then, the CLI tool `flow-copy-source` copies over the original Flow annotated files to `lib/` with a `.js.flow` extension, so users of `react-native-codegen` can still typecheck against it using Flow.

## Changelog

[General] [Added] - Codegen: Add prepublish script to build Flow files

## Test Plan

I am able to make use of the Codegen scripts without needing to use the `flow-node` CLI or the `flow-remove-types/register` require hook.

### DIFF
--- a/packages/react-native-codegen/package.json
+++ b/packages/react-native-codegen/package.json
@@ -7,13 +7,22 @@
     "type": "git",
     "url": "git@github.com:facebook/react-native.git"
   },
+  "scripts": {
+    "build": "yarn clean && yarn flow-remove-types src/ -d lib/ -q && yarn flow-copy-source src/ lib/",
+    "clean": "rm -rf lib",
+    "prepublish": "yarn run build"
+  },
   "license": "MIT",
   "files": [
-    "src"
+    "lib"
   ],
   "dependencies": {
     "flow-parser": "^0.121.0",
     "jscodeshift": "^0.7.0",
     "nullthrows": "^1.1.1"
+  },
+  "devDependencies": {
+    "flow-copy-source": "^2.0.9",
+    "flow-remove-types": "^2.122.0"
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ RNTester/Pods/*
 # react-native-codegen
 /ReactCommon/fabric/components/rncore/
 /schema-rncore.json
+/packages/react-native-codegen/lib
 
 # Visual studio
 .vscode


packages/react-native-codegen/package.json
```diff
@@ -7,13 +7,22 @@
...
+    "lib"
   ],
   "dependencies": {
     "flow-parser": "^0.121.0",
     "jscodeshift": "^0.7.0",
     "nullthrows": "^1.1.1"
+  },
+  "devDependencies": {
+    "flow-copy-source": "^2.0.9",
+    "flow-remove-types": "^2.122.0"
```
cpojer, line 26: Unfortunately these two packages pull a lot of dependencies into our monorepo at FB that we don't really use internally so it would slow down our product developers quite a bit.

We use `flow-remove-types` at version `1.2.3`. Can we use that for now and possibly not add `flow-copy-source` as it pulls in chokidar and a new version of Yargs etc.?

empyrical: `flow-remove-types@1.2.3` runs into syntax errors with the files in Codegen:

```
src/CodegenSchema.js
 ↳ Syntax Error: Unexpected token (319:70)
   318:        components?: $ReadOnly<{[component: string]: ComponentShape, ...}>,
```

I could make a build script that does all this, and not need to use `flow-remove-types` or `flow-copy-source` at all, hopefully eliminating the version bump of `flow-remove-types` as I had no idea about the issues the version bump it had would have internally.

Metro, for example, has a nice script that could be adapted to this (builds the files with Babel, and copies over Flow sources as necessary) 

https://github.com/facebook/metro/blob/b651e535cd0fc5df6c0803b9aa647d664cb9a6c3/scripts/build.js

If this sounds could, I could do this instead 👍

cpojer: @hramos has another diff/PR to just upgrade flow-remove-types. If you could rebase on top of that and then do some custom copying like Metro, I think that would be best.

hramos:
We'll be upgrading flow-remove-types from the other PR.

cpojer:
Just for more context: I believe Babel will be more sustainable as we'll eventually start using new syntax features that we need to compile away. You can copy-paste most of Metro's code and it should be able to produce the proper output.

empyrical:
That sounds good to me. I am actually going to close this pull request, and the build script based off of the one from Metro will get submitted in another pull request :+1:

PR #28646, opened 2020-04-16
Add support for icons in ActionSheetIOS

matt-oakes:
## Summary

This PR adds support for icons next to each option in the iOS Action Sheet Picker. It allows users to specify an array of `icons` which match the indexes of the `options` which are provided. It will show the icon next to each option.

There is also a `tintIcons` option which allows switching between tinting the icon with the given color (the default behaviour) and leaving the icons as they were provided.

These props and defaults were deliberately chosen to match the [`@expo/react-native-action-sheet`](https://github.com/expo/react-native-action-sheet) library. This library uses `ActionSheetIOS` internally, so added support to React Native core would allow the props  in this library to work across Android, Web, and iOS.

## Changelog

[iOS] [Added] - Add support for icons in ActionSheetIOS

## Test Plan

I have added three new cases to RNTester and tested it manually. The screenshots below show it with icons that were already present in RNTester. It would look nicer with smaller icons, but I didn't want to find  suitable icons were had the correct licencing to include in the repo.

**Icons for first two options with default tint**

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-16 at 09 31 40

**Icons for first two options with custom tint**

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-16 at 09 31 44

**Icons for first two options with tinting turned off**

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-16 at 09 31 48

# TODO

- [x] I have manually modified the `FBReactNativeSpec.h` file to support the new options. This is actually a file which is generated by a Facebook internal version of codegen. Can someone at Facebook generate the new version of this file and commit it to this PR so it is correct?

### DIFF
--- a/Libraries/ActionSheetIOS/NativeActionSheetManager.js
+++ b/Libraries/ActionSheetIOS/NativeActionSheetManager.js
@@ -10,6 +10,7 @@
 
 import type {TurboModule} from '../TurboModule/RCTExport';
 import * as TurboModuleRegistry from '../TurboModule/TurboModuleRegistry';
+import type {ResolvedAssetSource} from '../Image/AssetSourceResolver';
 
 export interface Spec extends TurboModule {
   +getConstants: () => {||};
@@ -25,6 +26,8 @@ export interface Spec extends TurboModule {
       +cancelButtonTintColor?: ?number,
       +userInterfaceStyle?: ?string,
       +disabledButtonIndices?: Array<number>,
+      +icons?: ?Array<ResolvedAssetSource>,
+      +shouldTintIcons?: ?boolean,
     |},
     callback: (buttonIndex: number) => void,
   ) => void;
--- a/React/CoreModules/RCTActionSheetManager.mm
+++ b/React/CoreModules/RCTActionSheetManager.mm
@@ -97,6 +97,11 @@ RCT_EXPORT_METHOD(showActionSheetWithOptions
   UIColor *cancelButtonTintColor =
       [RCTConvert UIColor:options.cancelButtonTintColor() ? @(*options.cancelButtonTintColor()) : nil];
 
+  NSArray<NSDictionary *> *icons = RCTConvertOptionalVecToArray(options.icons(), ^id(NSObject *element) {
+    return element;
+  });
+  BOOL shouldTintIcons = options.shouldTintIcons() ? [RCTConvert BOOL:@(*options.shouldTintIcons())] : YES;
+
   if (controller == nil) {
     RCTLogError(
         @"Tried to display action sheet but there is no application window. options: %@", @{
@@ -109,6 +114,8 @@ RCT_EXPORT_METHOD(showActionSheetWithOptions
           @"tintColor" : tintColor,
           @"cancelButtonTintColor" : cancelButtonTintColor,
           @"disabledButtonIndices" : disabledButtonIndices,
+          @"icons" : icons,
+          @"shouldTintIcons" : @(shouldTintIcons),
         });
     return;
   }
@@ -144,6 +151,25 @@ RCT_EXPORT_METHOD(showActionSheetWithOptions
     if (isCancelButtonIndex) {
       [actionButton setValue:cancelButtonTintColor forKey:@"titleTextColor"];
     }
+      
+    if (index < icons.count) {
+      NSDictionary *imageDictionary = icons[index];
+
+      if (imageDictionary != nil && ![imageDictionary isEqual:[NSNull null]]) {
+        UIImage *iconImage = [self iconImageFromDictionary:icons[index]];
+        if (iconImage) {
+          if (!shouldTintIcons) {
+            iconImage = [iconImage imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
+          }
+          [actionButton setValue:iconImage forKey:@"image"];
+        } else {
+          RCTLogWarn(
+              @"ActionSheetIOS: Index %@ from `icons` has an invalid icon image",
+              @(index));
+        }
+      }
+    }
+
     [alertController addAction:actionButton];
 
     index++;
@@ -266,6 +292,23 @@ RCT_EXPORT_METHOD(showShareActionSheetWithOptions
 - (std::shared_ptr<TurboModule>)getTurboModule:(const ObjCTurboModule::InitParams &)params
 {
   return std::make_shared<NativeActionSheetManagerSpecJSI>(params);
+}
+
+- (nullable UIImage *)iconImageFromDictionary:(NSDictionary *)dictionary
+{
+  NSString *iconURLString = dictionary[@"uri"];
+  if (!iconURLString) {
+    return nil;
+  }
+  NSURL *iconURL = [NSURL URLWithString:iconURLString];
+  if (!iconURL) {
+    return nil;
+  }
+  NSData *iconData = [NSData dataWithContentsOfURL:iconURL];
+  if (!iconData) {
+    return nil;
+  }
+  return [UIImage imageWithData:iconData];
 }
 
 @end
--- a/Libraries/ActionSheetIOS/ActionSheetIOS.js
+++ b/Libraries/ActionSheetIOS/ActionSheetIOS.js
@@ -12,6 +12,8 @@ import RCTActionSheetManager from './NativeActionSheetManager';
 
 const invariant = require('invariant');
 const processColor = require('../StyleSheet/processColor');
+import type {ImageSource} from '../Image/ImageSource';
+import resolveAssetSource from '../Image/resolveAssetSource';
 import type {ColorValue} from '../StyleSheet/StyleSheet';
 import type {ProcessedColorValue} from '../StyleSheet/processColor';
 
@@ -50,6 +52,8 @@ const ActionSheetIOS = {
       +cancelButtonTintColor?: ColorValue | ProcessedColorValue,
       +userInterfaceStyle?: string,
       +disabledButtonIndices?: Array<number>,
+      +icons?: ?Array<ImageSource>,
+      +shouldTintIcons?: boolean,
     |},
     callback: (buttonIndex: number) => void,
   ) {
@@ -64,6 +68,7 @@ const ActionSheetIOS = {
       tintColor,
       cancelButtonTintColor,
       destructiveButtonIndex,
+      icons,
       ...remainingOptions
     } = options;
     let destructiveButtonIndices = null;
@@ -85,12 +90,20 @@ const ActionSheetIOS = {
         typeof processedCancelButtonTintColor === 'number',
       'Unexpected color given for ActionSheetIOS.showActionSheetWithOptions cancelButtonTintColor',
     );
+
+    let processedIcons = null;
+    if (icons) {
+      invariant(Array.isArray(icons), 'Must provide a valid array of icons');
+      processedIcons = icons.map(icon => resolveAssetSource(icon));
+    }
+
     RCTActionSheetManager.showActionSheetWithOptions(
       {
         ...remainingOptions,
         tintColor: processedTintColor,
         cancelButtonTintColor: processedCancelButtonTintColor,
         destructiveButtonIndices,
+        icons: processedIcons,
       },
       callback,
     );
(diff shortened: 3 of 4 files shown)


Libraries/ActionSheetIOS/NativeActionSheetManager.js
```diff
@@ -25,6 +25,8 @@ export interface Spec extends TurboModule {
       +anchor?: ?number,
       +tintColor?: ?number,
       +userInterfaceStyle?: ?string,
+      +icons: ?Array<Object>, // TODO: This should be a ResolvedAssetSource
+      +tintIcons: boolean,
```
hramos, line 29:
```suggestion
      +icons?: ?Array<Object>, // TODO: This should be a ResolvedAssetSource
      +tintIcons?: boolean,
```

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -123,11 +130,21 @@ - (void)presentViewController:(UIViewController *)alertController
...
     NSInteger localIndex = index;
-    [alertController addAction:[UIAlertAction actionWithTitle:option
-                                                        style:style
-                                                      handler:^(__unused UIAlertAction *action) {
-                                                        callback(@[ @(localIndex) ]);
-                                                      }]];
+    UIAlertAction *action = [UIAlertAction actionWithTitle:option style:style handler:^(__unused UIAlertAction *action) {
+      callback(@[ @(localIndex) ]);
+    }];
+
```
hramos, line 136:
```suggestion
    UIAlertAction *action = [UIAlertAction actionWithTitle:option
                                                     style:style
                                                   handler:^(__unused UIAlertAction *action) {
                                                     callback(@[ @(localIndex) ]);
                                                   }];
      
```

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -239,6 +256,26 @@ - (void)presentViewController:(UIViewController *)alertController
   return std::make_shared<NativeActionSheetManagerSpecJSI>(self, jsInvoker, nativeInvoker, perfLogger);
 }
 
+- (nullable UIImage *)iconImageFromDictionary:(NSDictionary *)dictionary
+{
+  NSString *iconURLString = dictionary[@"uri"];
+  if (!iconURLString) {
+    return nil;
+  }
+
```
hramos, line 265:
```suggestion

```

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -239,6 +256,26 @@ - (void)presentViewController:(UIViewController *)alertController
...
+  NSString *iconURLString = dictionary[@"uri"];
+  if (!iconURLString) {
+    return nil;
+  }
+  
+  NSURL *iconURL = [NSURL URLWithString:iconURLString];
+  if (!iconURL) {
+    return nil;
+  }
+
```
hramos, line 270:
```suggestion

```

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -239,6 +256,26 @@ - (void)presentViewController:(UIViewController *)alertController
...
+  NSURL *iconURL = [NSURL URLWithString:iconURLString];
+  if (!iconURL) {
+    return nil;
+  }
+  
+  NSData *iconData = [NSData dataWithContentsOfURL:iconURL];
+  if (!iconData) {
+    return nil;
+  }
+
```
hramos, line 275:
```suggestion

```

Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
```diff
@@ -118,6 +118,8 @@ namespace JS {
       folly::Optional<double> anchor() const;
       folly::Optional<double> tintColor() const;
       NSString *userInterfaceStyle() const;
+      folly::Optional<facebook::react::LazyVector<id<NSObject>>> icons() const;
+      bool tintIcons() const;
```
hramos, line 122:
```suggestion
      folly::Optional<bool> tintIcons() const;
```

Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
```diff
@@ -2942,6 +2944,16 @@ inline NSString *JS::NativeActionSheetManager::SpecShowActionSheetWithOptionsOpt
...
+inline folly::Optional<facebook::react::LazyVector<id<NSObject>>> JS::NativeActionSheetManager::SpecShowActionSheetWithOptionsOptions::icons() const
+{
+  id const p = _v[@"icons"];
+  return RCTBridgingToOptionalVec(p, ^id<NSObject>(id itemValue_0) { return itemValue_0; });
+}
+inline bool JS::NativeActionSheetManager::SpecShowActionSheetWithOptionsOptions::tintIcons() const
+{
+  id const p = _v[@"tintIcons"];
+  return RCTBridgingToBool(p);
+}
```
hramos, line 2956:
```suggestion
inline folly::Optional<bool> JS::NativeActionSheetManager::SpecShowActionSheetWithOptionsOptions::tintIcons() const
{
  id const p = _v[@"tintIcons"];
  return RCTBridgingToOptionalBool(p);
}
```

matt-oakes:
@hramos Thanks for the review. I've addressed all of those issues now. I've also made the `tintOptions` optional for the native module and moved the "default to `true`" logic into native code rather than JavaScript.

sota000:
@matt-oakes, I know this is 1+ year old PR, but would you be wiling to rebase and solve the merge conflict? After that, I'll triage it internally. I'll check back in a few days.

matt-oakes:
@sota000 Thanks for looking into this. I'd actually given up hope that it was going to be looked at.

I have rebased on `main`, fixed the conflicts, and tested that everything still works in RNTester. Everything should be good to merge in now 👍

sota000:
Thanks, @matt-oakes! Sorry again for the delay (missed the notification). I've imported this to review this internally. Thanks!

Libraries/ActionSheetIOS/NativeActionSheetManager.js
```diff
@@ -24,6 +24,8 @@ export interface Spec extends TurboModule {
       +tintColor?: ?number,
       +userInterfaceStyle?: ?string,
       +disabledButtonIndices?: Array<number>,
+      +icons?: ?Array<Object>, // TODO: This should be a ResolvedAssetSource
+      +tintIcons?: ?boolean,
```
philIip, line 28: i think it's more clear to make this a verb, i.e. `shouldTintIcons`. otherwise, i would think this is an array of icon metadata, and since we lose the type in the obj-c layer, that could be very misleading.

also, this conforms more to obj-c naming standards

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -131,11 +138,23 @@ - (void)presentViewController:(UIViewController *)alertController
...
+                                                   }];
+      
+    if (index < icons.count) {
+      UIImage *iconImage = [self iconImageFromDictionary:icons[index]];
+      if (iconImage) {
+        if (!tintIcons) {
+          iconImage = [iconImage imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal];
+        }
+        [action setValue:iconImage forKey:@"image"];
+      }
```
philIip, line 154: i feel it is probably better to have an error here instead of failing silently, if the consumer is passing down a list of icons then presumably iconImage should never be nil

Libraries/ActionSheetIOS/NativeActionSheetManager.js
```diff
@@ -24,6 +24,8 @@ export interface Spec extends TurboModule {
       +tintColor?: ?number,
       +userInterfaceStyle?: ?string,
       +disabledButtonIndices?: Array<number>,
+      +icons?: ?Array<Object>, // TODO: This should be a ResolvedAssetSource
```
philIip, line 27: do you plan to land this code with this comment? it's not totally clear here

philIip (changes requested):
thanks a lot for working on this! i have some questions / recommendations.

React/CoreModules/RCTActionSheetManager.mm
```diff
@@ -131,11 +138,23 @@ - (void)presentViewController:(UIViewController *)alertController
...
-                                                      handler:^(__unused UIAlertAction *action) {
-                                                        callback(@[ @(localIndex) ]);
-                                                      }]];
+    UIAlertAction *action = [UIAlertAction actionWithTitle:option
+                                                     style:style
+                                                   handler:^(__unused UIAlertAction *action) {
+                                                     callback(@[ @(localIndex) ]);
+                                                   }];
+      
+    if (index < icons.count) {
```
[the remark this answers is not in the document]
matt-oakes: I implemented it this way so it matches the implementation of the existing Android & Web only options in the Expo Action Sheet. Details here: https://github.com/expo/react-native-action-sheet#androidweb-only-props

The intention of this pull request was to add support for these options into React Native code as the Expo library relies on ActionSheetIOS for it's iOs implementation, while providing its own implementation for Android and web.

Changing the options in the core would just complicate usage for existing users of the Expo library.

However, to support the use case you mentioned, I have added the ability to pass in `null` values into the array to signify that an icon should not be used for that index.

React/CoreModules/RCTActionSheetManager.mm
matt-oakes: This was origionally done to allow passing in a `null` value for some of the icons, as described above. However, I think you are correct that it is better to show an error when a value has been passed in.

I have changed the code so it will ignore `null` values, but will show an error when something has been passed in.

Libraries/ActionSheetIOS/NativeActionSheetManager.js
matt-oakes: I do, because I wasn't sure how to correctly type this and there is no documentation for Turbo Modules that's publicly available yet. If there is a better way to do this, then feel free to either let me know or add a commit to this branch.

React/CoreModules/RCTActionSheetManager.mm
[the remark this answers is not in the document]
matt-oakes: I have added a test case to RNTester which explicitly tests this use case by supplying a `null` for the first value and an icon for the 2nd

Libraries/ActionSheetIOS/NativeActionSheetManager.js
matt-oakes: As noted in the PR description, I chose these option names to match the existing ones from the Expo action sheet library, which uses ActionSheetIOS internally. The options are named this way for Android and web and changing them would introduce some complexity and incompatabilites. Details here: https://github.com/expo/react-native-action-sheet#androidweb-only-props

Libraries/ActionSheetIOS/NativeActionSheetManager.js
RSNara: Hey @matt-oakes!

In the TurboModule codegen, you can use any Flow types that you declare in the Spec. So, if you know the shape of ResolvedAssetSource, you can do this:

```
type ResolvedAssetSource = {
  foo: string,
};

icons: ?Array<ResolvedAssetSource>
```

sota000:
@matt-oakes, mind taking a look at the comment?

Libraries/ActionSheetIOS/NativeActionSheetManager.js
matt-oakes: @RSNara Can you point me to a place in the code where the shape of this object is defined, or another place where it's passed through JSI?

At the time I wrote this, the other places that were passing asset references were doing what I did, however, that might have changed now.

matt-oakes:
@sota000 I have replied to the comment.

I don't know how to fix this issue because it's something related to the internal structure of the assets which is not documented anywhere. That is why I left this as a TODO for someone with more knowledge to add in, rather than trying to guess myself.

Personally, I would say to merge this in as-is as it works and the last remaining issue is a small code quality thing that can be tidied up by someone with more knowledge in the future. The type saftey won't really add much at this point.

sota000:
I'll see what we can do. Thanks for being patient!

RSNara:
Hey @matt-oakes, I just saw your comment. It looks like ResolvedAssetSource is defined here: https://github.com/facebook/react-native/blob/77ecc7ede1da8fc590d7bc238a2fc02daa736746/Libraries/Image/AssetSourceResolver.js#L13-L19. 

I found this with a quick search through the repository. Did you encounter this type earlier, but rule it out, because it was invalid/not applicable in this scenario for some reason?

matt-oakes:
@RSNara I have made that change now and rebased on `main`

PR #28650, opened 2020-04-16
Add support for shadowColor on Android (API >= 28)

IjzerenHein:
## Summary

This PR adds support for the `shadowColor` style on Android. 

This is possible as of Android P using the `setOutlineAmbientShadowColor` and `setOutlineSpotShadowColor` View methods. The actual rendered color is a multiplication of the color-alpha, shadow-effect and elevation-value.

## Changelog

`[Android] [Added] - Add support for shadowColor on API level >= 28`

## Test Plan

- Only execute code on Android P
- Added Android `BoxShadow` tests to RNTester app

image

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
@@ -96,6 +96,18 @@ public abstract class BaseViewManager<T extends View, C extends LayoutShadowNode
   }
 
   @Override
+  @ReactProp(
+      name = ViewProps.SHADOW_COLOR,
+      defaultInt = Color.BLACK,
+      customType = "Color")
+  public void setShadowColor(@NonNull T view, int shadowColor) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      view.setOutlineAmbientShadowColor(shadowColor);
+      view.setOutlineSpotShadowColor(shadowColor);
+    }
+  }
+
+  @Override
   @ReactProp(name = ViewProps.Z_INDEX)
   public void setZIndex(@NonNull T view, float zIndex) {
     int integerZIndex = Math.round(zIndex);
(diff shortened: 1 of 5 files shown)


ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java
```diff
@@ -95,6 +95,18 @@ public void setElevation(@NonNull T view, float elevation) {
...
 
+  @Override
+  @ReactProp(
+      name = ViewProps.SHADOW_COLOR,
+      defaultInt = Color.BLACK,
+      customType = "Color")
+  public void setShadowColor(@NonNull T view, int shadowColor) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      view.setOutlineAmbientShadowColor(shadowColor);
+      view.setOutlineSpotShadowColor(shadowColor);
```
shergin, line 106: Why do we need both?

IjzerenHein: The rendered shadows are a combination of both an ambient shadow (soft) and a spot (hard) shadow which produce the Material design elevation effect.

> In the Material Design environment, virtual lights illuminate the UI. Key lights create sharper, directional shadows, called key shadows. Ambient light appears from all angles to create diffused, soft shadows, called ambient shadows.

image

It's best described here:
https://material.io/design/environment/light-shadows.html#light

shergin:
@IjzerenHein Why do we need a separate Android-specific `BoxShadowExample.android.js`  example?

IjzerenHein:
@shergin I did this because Android has a different shadows API (elevation) vs the other platforms. And I wanted to provide a full test-suite for that.

Frans-L:
Awesome work @IjzerenHein!  This feature has been highly needed, thanks :)

By the way, does the shadow color prop support animation for android? 
So, is it possible to animate shadow color by using ```<Animated.View>```?

PR #28662, opened 2020-04-17
chore: upgrade eslint-config in template

Naturalclar:
## Summary
Upgraded the eslint-config in template

## Changelog

[General] [Changed] - Upgrade eslint-config in template to prepare for 0.63 release https://github.com/react-native-community/releases/issues/186

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -16,7 +16,7 @@
   "devDependencies": {
     "@babel/core": "^7.8.4",
     "@babel/runtime": "^7.8.4",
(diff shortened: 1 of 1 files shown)


TheSavior:
Ah, I thought I did this already, whoops. 

Changing packages in `package.json`s requires more work for us internally so we can't land the diffs that do this directly. I've put up this change internally though, so it should go through after review.

TheSavior:
This change landed in https://github.com/facebook/react-native/commit/780f06cd477f34da48646a949bd25dd3f883a9a2

PR #28703, opened 2020-04-20
Allow iOS PlatformColor strings to be ObjC or Swift UIColor selectors

tom-un:
## Summary
Per discussion in https://github.com/react-native-community/releases/issues/186 the iOS `PlatformColor()` function is documented to use the semantic color names provided by the system.   The referenced HIG documentation itself links to the `UIColor` documentation for semantic colors names.   However, these names differ depending on if you are viewing the new Swift API docs or the Objective C docs.   The current Objective C implementation in react-native assumes Objective C UIColor selector names that are suffixed 'Color'.   But in Swift, Apple provides a Swift Extension on UIColor that makes aliases without the the 'Color' suffix and then makes the original selectors invalid presumably via `NS_UNAVAILABLE_SWIFT`.

Since both selector names are valid depending on if you are using Objective C or Swift, let's make both forms be legal for `PlatformColor()`.   In `RCTConvert.m` there is a dictionary of legal selector names.   The code already supports the ability to have names be aliases of other selectors via a RCTSelector metadata key.   The change adds code to the initialization of the map: it iterates over the keys in the map, which are all ObjC style UIColor selectors, and creates aliases by duplicating the entries, creating key names by stripping off the ObjC "Color" suffix, adds the RCTSelector key referring to the original and then appends these new Swift aliases to the map.

## Changelog

[iOS] [Changed] - Allow iOS PlatformColor strings to be ObjC or Swift UIColor selectors

## Test Plan

The PlatformColorExample.js is updated to use the new, shorter Swift selector names.   There are still other examples in the same file and in unit tests that exercise the ObjC selector names.

PlatformColor

### DIFF
--- a/RNTester/js/examples/PlatformColor/PlatformColorExample.js
+++ b/RNTester/js/examples/PlatformColor/PlatformColorExample.js
@@ -22,94 +22,94 @@ function PlatformColorsExample() {
       colors = [
         // https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors
         // Label Colors
-        {label: 'labelColor', color: PlatformColor('labelColor')},
+        {label: 'label', color: PlatformColor('label')},
         {
-          label: 'secondaryLabelColor',
-          color: PlatformColor('secondaryLabelColor'),
+          label: 'secondaryLabel',
+          color: PlatformColor('secondaryLabel'),
         },
         {
-          label: 'tertiaryLabelColor',
-          color: PlatformColor('tertiaryLabelColor'),
+          label: 'tertiaryLabel',
+          color: PlatformColor('tertiaryLabel'),
         },
         {
-          label: 'quaternaryLabelColor',
-          color: PlatformColor('quaternaryLabelColor'),
+          label: 'quaternaryLabel',
+          color: PlatformColor('quaternaryLabel'),
         },
         // Fill Colors
-        {label: 'systemFillColor', color: PlatformColor('systemFillColor')},
+        {label: 'systemFill', color: PlatformColor('systemFill')},
         {
-          label: 'secondarySystemFillColor',
-          color: PlatformColor('secondarySystemFillColor'),
+          label: 'secondarySystemFill',
+          color: PlatformColor('secondarySystemFill'),
         },
         {
-          label: 'tertiarySystemFillColor',
-          color: PlatformColor('tertiarySystemFillColor'),
+          label: 'tertiarySystemFill',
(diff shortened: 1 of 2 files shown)


RNTester/js/examples/PlatformColor/PlatformColorExample.js
```diff
@@ -22,94 +22,94 @@ function PlatformColorsExample() {
       colors = [
         // https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors
         // Label Colors
-        {label: 'labelColor', color: PlatformColor('labelColor')},
```
TheSavior, line 25: Do you want to keep any examples that include the Color suffix? And / or maybe have an automated test that verifies both ways work?

tom-un: As I said in the Test notes, there still are ObjC examples later in this file and in unit tests

TheSavior: Ah, whoops. I should read better :D

React/Base/RCTConvert.m
```diff
@@ -729,7 +729,22 @@ +(type)type : (id)json                                      \
...
+    } mutableCopy];
+    // The color names are the Objective-C UIColor selector names,
+    // but Swift selector names are valid as well, so make aliases.
+    static NSString *const RCTColorSuffix = @"Color";
+    NSMutableDictionary<NSString *, NSDictionary *> *aliases = [NSMutableDictionary new];
+    for (NSString *objcSelector in map) {
+      RCTAssert([objcSelector hasSuffix:RCTColorSuffix], @"A selector in the color map did not end with the suffix Color.");
+      NSMutableDictionary *entry = [map[objcSelector] mutableCopy];
+      RCTAssert([entry objectForKey:RCTSelector] == nil, @"Entry should not already have an RCTSelector");
+      NSString *swiftSelector = [objcSelector substringToIndex:[objcSelector length] - [RCTColorSuffix length]];
```
alloy, line 741: Seems like this code could have been slightly simpler/faster if it had been reversed, i.e. manually add them all without the `Color` suffix and then alias them by appending the suffix. Not a huge deal [on its own], though, as this map is cached.

tom-un: If reversed, it would mean having to add an explicit RCTSelector for each entry -- which is more error prone for maintenance.   I like it this way because the table is more truthy to ObjC which is the native language of the RN implementation itself.

PR #28711, opened 2020-04-21
Add Dark Mode support to the App template and  NewAppScreen components

Simek:
## Summary

This PR adds support for the dark mode and dynamic theme changing to the default App template and to the related `NewAppScreen` components. Using `useColorScheme` hook forced me to refactor a bit main `App.js` file, but I think those changes are step in the right direction according to way in which React Native is used in larger apps, so new `Section` component has been extracted to reduce code redundancy/repetition inside `App`. 

Additional color `darker` has been added to the `Colors` statics from `NewAppScreen` because `dark` was too bright for the Dark Mode backgrounds.

Also main `StoryBoard` on iOS has been updated to use theme based colors instead of static or hardcoded ones. There was also an unused, empty `Label` which I have removed.

~~I'm not so much experienced with Android. If someone could also update Android splash screen (if Android requires such change) it will be nice. I want to look at this later using simulator.~~
> I have updated the Android splash screen and tested this change on the Android emulator.

If you have any comment or corrections feel free to post them out, I would like to put more work into this PR if it's needed. Dark Mode this days is a part of near every OS, so it could be considered as a standard feature. I hope those changes helps people which struggle with the basic theming implementation (+ there is now an example of hook and `children` prop usage in the template). 

## Changelog

[Internal] [Added] - Add dark mode support to the default app template

## Test Plan

I have tested the App from the template on the iOS device and in Android emulator with RN `0.63.0-rc`. 

Screen recording on iOS (demonstarates both modes, both splash screens and transition):
ezgif-6-e24ee8e839c9

Screenshot of iOS app in Dark Mode:
IMG_6542

Screenshot of iOS app splash screen in Dark Mode:
IMG_6544

Screenshot of Android app in the emulator:
Screenshot_1587566100

### DIFF
--- a/Libraries/NewAppScreen/components/Colors.js
+++ b/Libraries/NewAppScreen/components/Colors.js
@@ -16,5 +16,6 @@ export default {
   lighter: '#F3F3F3',
   light: '#DAE1E7',
   dark: '#444',
+  darker: '#222',
   black: '#000',
 };
--- a/Libraries/NewAppScreen/components/Header.js
+++ b/Libraries/NewAppScreen/components/Header.js
@@ -11,25 +11,40 @@
 'use strict';
 import Colors from './Colors';
 import type {Node} from 'react';
-import {Text, StyleSheet, ImageBackground} from 'react-native';
+import {ImageBackground, StyleSheet, Text, useColorScheme} from 'react-native';
 import React from 'react';
 
-const Header = (): Node => (
-  <ImageBackground
-    accessibilityRole={'image'}
-    source={require('./logo.png')}
-    style={styles.background}
-    imageStyle={styles.logo}>
-    <Text style={styles.text}>Welcome to React</Text>
-  </ImageBackground>
-);
+const Header = (): Node => {
+  const isDarkMode = useColorScheme() === 'dark';
+  return (
+    <ImageBackground
+      accessibilityRole="image"
+      source={require('./logo.png')}
+      style={[
+        styles.background,
+        {
+          backgroundColor: isDarkMode ? Colors.darker : Colors.lighter,
+        },
+      ]}
+      imageStyle={styles.logo}>
+      <Text
+        style={[
+          styles.text,
+          {
+            color: isDarkMode ? Colors.white : Colors.black,
+          },
+        ]}>
+        Welcome to React
+      </Text>
+    </ImageBackground>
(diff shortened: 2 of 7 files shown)


Simek:
GitHub Internal Server Errors (`500`) are responsible for the failed CIs. More information [here](https://www.githubstatus.com/incidents/dsf2qtzh4jpz).

jacobp100:
You might want to hold off doing this until PlatformColor has landed - then you can support dark mode inside StylesSheet.create

Simek:
Thank you for your comment. I have mentioned using `Platform Color` API in my [initial comment](https://github.com/react-native-community/releases/issues/186#issuecomment-615904754) in the `0.63` release discussion thread. 

But after thinking and researching a while it looks like, at least for now, there is no possibility to remove `Colors` statics from the `NewAppScreen` due to compatibility with out-of-tree platforms. I'm also not sure if using `Platform Color` simplify or make more complex the code, because you need to define color for at least 3 platforms. It's also brand new API so I think is wise to wait with this improvement a bit.

Of course if such changes will be request I'm happy to implement it.

Simek:
BTW. I have noticed that after updating Android `AppTheme` even JS bundle loading view has dark background, while on iOS Dark Mode do not affect the background color and in both modes it's white.

jacobp100:
You can change LaunchScreen.xib for iOS - if you set the background to system background and the text to label, it'll be dark mode compatible

jacobp100:
In AppDelegate.m, you could change the root background colour code to something like

```obcj
    rootView.backgroundColor = [UIColor systemBackgroundColor];
} else {
    rootView.backgroundColor = [UIColor whiteColor];
}
```

Simek:
@jacobp100 React Native uses `StoryBoard`s not `XIB`s as I mentioned in the PR description. Splash screen for the iOS has been updated in the initial commit.

Simek:
@jacobp100 Awesome tip, I forget about updating RootView. This change fixes the issue described above, thank you!

shergin:
Looks so pretty! I can't resist!
Thank you!

Simek:
@shergin One more thing that came to my mind: 
* Can I move Hermes badge to the `NewAppScreen` components or it is important to expose `global.HermesInternal` for the template users?

PR #28795, opened 2020-04-30
Support changing position of LogBox

osdnk:
## Summary

LogBox is extremely useful! Thanks for it.
I observed that sometimes it's blocking my interactions with the app, but I'd like to keep it visible.
image

So I've added a small button for switching position to top or bottom of the screen

image
image
image
image

Tested on iOS and Android.

## Changelog

[General] [Added] - Possibility to change the position of log box.

## Test Plan

Warnings are immediately visible after running RNTester app.


rickhanlonii:
Hey @osdnk, thanks for submitting this!

I agree we should solve this problem. My thinking in the original designs is that we would allow you to drag and move the notifications, like you can do with [FLEX](https://github.com/Flipboard/FLEX).

What do you think?

FilipMaslovaric:
Can we please move this forward soon, please? This is infuriating if you're developing with something like React Navigation, and using a Tab Navigator.

rickhanlonii:
@FilipMaslovaric sorry for the poor experience. Note that this isn't a regression with LogBox. Previously, YellowBox and RedBox both covered _more_ of the screen than LogBox.

But we should make the dragable, instead of using a configuration, so I'm closing this.

PR #28817, opened 2020-05-03
Allow passing partial contentOffset to ScrollView on Android

janicduplessis:
## Summary

Since support for contentOffset was added to horizontal ScrollView on android (30cc158a875a0414cf53d4d5155410eea5d5aeea) I'm seeing a crash in my app because of a library. What happens is that it passes a partial object for contentOffset so something like `{x: 1}` which causes a crash on Android.

According to the flow types the object should always contain both x and y but I think we should preserve the runtime behaviour and just use 0 like iOS does. 

## Changelog

[Android] [Fixed] - Allow passing partial contentOffset to ScrollView on Android

## Test Plan

Tested that passing partial object for contentOffset does not crash.


janicduplessis:
cc @JoshuaGross

shergin:
@janicduplessis Do we need to land #28760 as well?

janicduplessis:
@shergin I haven't had issues with what #28760 fixes but it seems to make sense.

PR #28827, opened 2020-05-05
Codegen: Add prepublish script to build Flow files

empyrical:
## Summary

*This is a follow-up to #28645, redone using a build script based off of Metro's build script instead of using `flow-remove-types` and `flow-copy-source`.*

This pull request adds a build step to `react-native-codegen` that builds the Flow-annotated JS files so that users of the NPM module `react-native-codegen` do not need to use require hooks to be able to import it.

A new build script, `scripts/build.js` is added that builds every JS file in `src/` into a `lib/` folder, and also copies over the original Flow annotated files to `lib/` with a `.js.flow` extension, so users of `react-native-codegen` can still typecheck against it using Flow. The shell scripts in `src` are also copied over. It is based off of the [build script from Metro](https://github.com/facebook/metro/blob/00867816eb9b2f69c8af9cebb523e9e4d280671a/scripts/build.js)

## Changelog

[General] [Added] - Codegen: Add prepublish script to build Flow files

## Test Plan

I am able to make use of the Codegen scripts without needing to use the `flow-node` CLI or the `flow-remove-types/register` require hook.

### DIFF
--- a/packages/react-native-codegen/package.json
+++ b/packages/react-native-codegen/package.json
@@ -7,13 +7,28 @@
     "type": "git",
     "url": "git@github.com:facebook/react-native.git"
   },
+  "scripts": {
+    "build": "yarn clean && node scripts/build.js --verbose",
+    "clean": "rm -rf lib",
+    "prepublish": "yarn run build"
+  },
   "license": "MIT",
   "files": [
-    "src"
+    "lib"
   ],
   "dependencies": {
     "flow-parser": "^0.121.0",
     "jscodeshift": "^0.9.0",
     "nullthrows": "^1.1.1"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.2.0",
+    "@babel/preset-env": "^7.2.0",
+    "@babel/preset-flow": "^7.2.0",
+    "chalk": "^2.4.1",
+    "glob": "^7.1.1",
+    "micromatch": "^2.3.11",
+    "mkdirp": "^0.5.1",
+    "prettier": "1.19.1"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/react-native-codegen/package.json
```diff
@@ -7,13 +7,28 @@
...
+  },
+  "devDependencies": {
+    "@babel/core": "^7.2.0",
+    "@babel/preset-env": "^7.2.0",
+    "@babel/preset-flow": "^7.2.0",
+    "chalk": "^2.4.1",
+    "glob": "^7.1.1",
+    "micromatch": "^2.3.11",
+    "mkdirp": "^0.5.1",
+    "prettier": "1.19.1"
```
hramos, line 32: Can you use these versions for the following packages? 
They would match what we are using internally, and help us avoid pulling multiple copies of the same library.

```
    "@babel/core": "^7.0.0",
    "@babel/preset-flow": "^7.0.0",
    "chalk": "^4.0.0",
    "micromatch": "^3.0.4",
    "mkdirp": "^0.5.1",
```

empyrical:
Thanks for the feedback, I'll make those changes! For Chalk and the like, I did initially use the versions used in the metro repo but looks like they got upgraded since I first wrote this patch 🤔

hramos:
Yes, sorry about that! Christoph has been working on reducing the size of our node_modules directory, so there was some work that landed while you were working on the updates to the PR. Thanks for bearing with us!

empyrical:
No problem, being conscientious about dependencies brought in is a good thing 😁 

I changed those versions around, and swapped out the Babel presets for the plugins used by Metro's Babel config minus the Babel plugins in it that were marked as probably not being necessary. Codegen appears to work fine without those plugins added.

PR #28841, opened 2020-05-06
Fix: Preserve native animated value after animated component unmount

### LINKED ISSUE #28114: Animated.Value resets if component was re-mounted
## Description:

If Animated.Value is used for a natively driven animation and then the component is unmounted, when the component is mounted back, the rendered value resets to the initial state.

## React Native version: 
0.61.5

## Steps To Reproduce

1. Create Animated.Value av.
2. Use av in styles for component C (e.g. translateX). Don't attach any listeners to the value.
3. Run an animation (useNativeDriver: true) with av.
4. Unmount C (and all other components that utilize av in their styles).
5. Mount C.
6. C won't be positioned correctly since av's internal value resets to its state on step 2.

## Expected Results

Animated.Value should hold the last value it had before the component was unmounted, and the component should return to the same state on being re-mounted as a result.

## Code Example

https://github.com/djonin/react-native-animated-value-bug

This is a duplicate of issues #23712 and #23621, which were incorrectly closed as fixed, due to using a listener in the example, which avoids this defect. The stated fix #24571 only fixed the problem with the listener not correctly re-binding to the new value, however the broader case issue mentioned in #23621 is still there.

This issue is due to the fact the native node, which has the up to date value, is destroyed after a component is dismounted. The JS animated value object doesn't receive any native value updates unless there is a listener attached explicitly by the user. Therefore the next time the component is mounted, it uses the value as it was before the native animation.

Currently the only way to preserve the value is to attach a listener, meaning the value updates will be sent over the bridge often during animations. In addition, it's not obvious to users that attaching a listener is required for correct re-mounting behavior.

## Proposed solution

One possible fix would be to make a new event to call back to JS with the value just once when the native node is being detached/destroyed, so that we can preserve whatever the final native value was in our JS object, and for the JS animated value object to always listen to this new native event, even with no listeners attached explicitly. We need some place to store our value, and since the native side is quite correctly being cleaned up after unmount, JS side seems like the best place to do it.

TMaszko:
## Summary
After animation has been finished using Native driver there is no final value passed from the native to JS side. This causes a bug from #28114.

This PR solves this problem in the same way as `react-native-reanimated` library. When detaching it is calling native side to get the last value from Animated node and stores it on the JS side.

Preserving animated value even if animation was using `useNativeDriver: true`
Fixes #28114 

## Changelog

[Internal] [Fixed] - Save native Animated node value on JS side in detach phase

## Test Plan
Unit tests for added getValue method passed. Green CI

### DIFF
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -43,7 +43,9 @@ const API = {
     saveValueCallback: (value: number) => void,
   ): void {
     invariant(NativeAnimatedModule, 'Native animated module is not available');
-    NativeAnimatedModule.getValue(tag, saveValueCallback);
+    if (NativeAnimatedModule.getValue) {
+      NativeAnimatedModule.getValue(tag, saveValueCallback);
+    }
   },
   disableQueue: function(): void {
     invariant(NativeAnimatedModule, 'Native animated module is not available');
@@ -242,7 +244,9 @@ function validateTransform(
   configs.forEach(config => {
     if (!TRANSFORM_WHITELIST.hasOwnProperty(config.property)) {
       throw new Error(
-        `Property '${config.property}' is not supported by native animated module`,
+        `Property '${
+          config.property
+        }' is not supported by native animated module`,
       );
     }
   });
--- a/Libraries/Animated/src/NativeAnimatedHelper.js
+++ b/Libraries/Animated/src/NativeAnimatedHelper.js
@@ -38,6 +38,9 @@ const API = {
   enableQueue: function(): void {
     queueConnections = true;
   },
+  getValue: function(tag: number, callback: (value: number) => void): void {
+    NativeAnimatedModule.getValue(tag, callback);
+  },
   disableQueue: function(): void {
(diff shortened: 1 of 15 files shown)


thymikee:
cc @wcandillon @shergin

zamotany:
Hey @janicduplessis, we would really appreciate if you could take a look at it

JoshuaGross:
Hi @TMaszko, this looks reasonable to me at first glance, though I need to do a deeper code-review.

Is it possible to create an RNTester example for this, maybe to one of the existing Animation examples? That would make it a lot easier to test and to prevent regressions in the future.

TMaszko:
Hi @JoshuaGross , I added this example to the Animated examples as you have asked for :)

zamotany:
@JoshuaGross any update/feedback??

JoshuaGross:
I'll test this internally.

JoshuaGross:
I have minor concerns about perf implications of this, since it adds a native module/TM call to every animated view detach - and those calls aren't free. However, this seems pretty reasonable overall; and as for perf implications, users should strive to re-render or detach Animated values/Views as little as possible anyway, so it doesn't really change any constraints. Given that caveat I think it's reasonable to merge and it seems like there are no ill effects within the FB app. :)

PR #28853, opened 2020-05-08
Remove duplicate local-cli/cli.js file

ecreeth:
# Summary
This PR remove unused local-cli/cli.js file and removes it from package.json.
The same file can be found in [ROOT_DIRECTORY/cli.js](https://github.com/facebook/react-native/blob/master/cli.js) and it is exported in the package.json file too

## Changelog

[JavaScript] [Removed] - Unused and duplicate file


cpojer:
Thank you for the PR! When we removed this the last time some external tooling broke because it was requiring this file from RN. Since it’s a very small file, we decided to keep it.

PR #28862, opened 2020-05-09
Fix skewX on Android and in the JS decomposition

### LINKED ISSUE #27649: [Android] skew transform not working
## Steps To Reproduce

1. Try to set `skewX` for View component: `transforms: [{ skewX: '45deg' }}`
3. Doesn't change the skew of View on Android (works on iOS)

### Describe what you expected to happen:
Basically `skewX` doesn't apply any skew on the view on Android.
Here is comparison between iOS/Android:
> also `skewY` doesn't render same as on iOS (which looks more correct)
Screenshot 2020-01-02 at 2 13 00 PM

### React Native version:
```
System:
    OS: macOS Mojave 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.50 GB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.19.1 - ~/.yarn/bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
  SDKs:
    iOS SDK:
    Android SDK:
      API Levels: 25, 28, 29
      Build Tools: 28.0.3, 29.0.2
      System Images: android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5977832
    Xcode: 11.3/11C29 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5
```

### Related issues
- [x] https://github.com/facebook/react-native/issues/12212

### Snack demo

https://snack.expo.io/@usrbowe2/e9b920

### Resources
In case anyone interested, here is the source code for skew code:
1. This is called for SkewX: `MatrixMathHelper.applySkewX(helperMatrix, convertToRadians(transform, transformType));` in file: `Libraries/StyleSheet/processTransform.js`
2. Which laters uses this matrix helper to apply transformation:
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/MatrixMathHelper.java

wcandillon:
This issue fixes #27649.

## Summary

By using 2d decomposition that transforms a skewX into a rotate/scale/rotate, the skewX issue on Android was still there which made me suspect that the issue came from the decomposition algorithm. Then I noticed that the bug existed in the JavaScript decomposition as well which led me to a fix on the JS and therefore on the Android side most likely. 

## Changelog

[Android] [Fixed] skewX transforms

## Test Plan

Check that skewX works on Android.
On JS, making sure that processTransform() doesn't skip, you can try the following sequence:

```tsx
  const matrix = processTransform([{ skewX: `${Math.PI / 3}rad` }]);
  const result = MatrixMath.decomposeMatrix(matrix);
  console.log({ result });
```

### DIFF
--- a/Libraries/Utilities/MatrixMath.js
+++ b/Libraries/Utilities/MatrixMath.js
@@ -651,10 +651,6 @@ const MatrixMath = {
     skew[0] = MatrixMath.v3Dot(row[0], row[1]);
     row[1] = MatrixMath.v3Combine(row[1], row[0], 1.0, -skew[0]);
 
-    // Compute XY shear factor and make 2nd row orthogonal to 1st.
-    skew[0] = MatrixMath.v3Dot(row[0], row[1]);
-    row[1] = MatrixMath.v3Combine(row[1], row[0], 1.0, -skew[0]);
-
     // Now, compute Y scale and normalize 2nd row.
     scale[1] = MatrixMath.v3Length(row[1]);
     row[1] = MatrixMath.v3Normalize(row[1], scale[1]);
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/MatrixMathHelper.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/MatrixMathHelper.java
@@ -169,10 +169,6 @@ public class MatrixMathHelper {
     skew[0] = v3Dot(row[0], row[1]);
     row[1] = v3Combine(row[1], row[0], 1.0, -skew[0]);
 
-    // Compute XY shear factor and make 2nd row orthogonal to 1st.
-    skew[0] = v3Dot(row[0], row[1]);
-    row[1] = v3Combine(row[1], row[0], 1.0, -skew[0]);
-
     // Now, compute Y scale and normalize 2nd row.
     scale[1] = v3Length(row[1]);
     row[1] = v3Normalize(row[1], scale[1]);


wcandillon:
@janicduplessis This PR should fix skewX on Android.

shergin:
😯Importing to run internal tests.

shergin:
@wcandillon Thank you so much for doing that!
This is amazing catch! I am pretty sure it's correct but I will land this on Monday just to be safe.

Meanwhile, do you mind to add examples/demos which illustrate the problem to RNTester app? We do have a bunch of matrix transformation logic in Fabric/C++ and... non of us is a real scientist, so it would be super helpful just to compare how it looks on different platforms w/ and w/o Fabric. We don't use decomposition in Fabric though.

wcandillon:
Thank you @shergin for looking into it. Could you provide some pointers on how to add examples to the RNTester app?

shergin:
@wcandillon Here: https://github.com/facebook/react-native/blob/master/RNTester/js/examples/Transform/TransformExample.js

janicduplessis:
@wcandillon Nice find! Thanks for working on this.

There are also info in the readme here on how to run the test app if you have issues https://github.com/facebook/react-native/blob/master/RNTester/README.md.

wcandillon:
Is there anything I can do on my side to get this fix shipped into React Native or will it happen eventually?

alejandrorangel:
this was shipped with 0.63.3?

DingSoung:
> this was shipped with 0.63.3?

haven't,


I have changed my local code to test this fix, It seems not work

Podlipny:
was this issue shiped in version 0.63.4? Because it doesn't work for me.
`<View style={{ backgroundColor: 'red', width: 60, height: 60, transform: [{ skewX: '20rad' }] }} />`

PR #28868, opened 2020-05-10
Fixed italic style not being removed on TextInput on Android

hsource:
## Summary

When a user removes `{ fontStyle: 'italic' }` from a TextInput on Android, the text remains italic. This is because the `fontStyle` is set to UNSET, which just retains the previous style.

The user would rather expect that the `fontStyle` be set back to 'normal'

## Changelog

[Android] [Fixed] - Fixed so that removing `fontStyle: 'italic'` style from TextInput correctly resets to font-style to be non-italic

## Test Plan

Added a new TextInput test in RNTesterApp called "fontWeight and fontStyle with toggles"

See behaviour:

| Before | After |
| --- | --- |
| Before | Before |

### DIFF
--- a/RNTester/js/examples/TextInput/TextInputSharedExamples.js
+++ b/RNTester/js/examples/TextInput/TextInputSharedExamples.js
@@ -627,7 +627,7 @@ module.exports = ([
     },
   },
   {
-    title: 'fontFamily and fontWeight',
+    title: 'fontWeight and fontStyle with toggles',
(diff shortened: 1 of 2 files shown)


hsource:
@JoshuaGross - it looks like you've added a bunch of code before for the TextInput on Android - would you be able to give this a quick review? Thanks!

hsource:
Looks fixed!

PR #28880, opened 2020-05-12
Added the TextPropTypes and ImagePropTypes to the react native properties

whiteHatCip:
## Summary

The types added allow for PropTypes prop validation when passing textStyles to a custom component, or to pass an imported image as a source for an image contained inside a custom component. This edit is needed when using TypeScript.

## Changelog

[General] [Added] - Added the TextPropTypes and ImagePropTypes to the react native properties.

## Test Plan

Create a View component that contains an Image and/or a Text component. The View component takes in as props the textStyle and the image source. When you need to specify the propTypes for the View component you can type:
`MyView.propTypes = {textStyle: TextPropTypes.style, icon: ImagePropTypes.source}`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -496,7 +496,7 @@ module.exports = {
   },
   get ImagePropTypes(): DeprecatedImagePropTypes {
     return require('./Libraries/DeprecatedPropTypes/DeprecatedImagePropType');
-  }
+  },
 };
 
 if (__DEV__) {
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for the PR! Unfortunately we are currently working on getting rid of these props (hence the Deprecated) name, and don't want to add more of them to our public API.

whiteHatCip:
Thanks for the response, but since I need to use these kind of types to type definition for component props, what am I supposed to use? I made this pull request only because I found nowhere any clue on how to manage this need. Please provide me with the current alternative to my PR, thanks a lot

whiteHatCip:
Thanks for the response, but since I need to use these kind of types to type definition for component props, what am I supposed to use? I made this pull request only because I found nowhere any clue on how to manage this need. Please provide me with the current alternative to my PR, thanks a lot

TheSavior:
For now, these types are exported off of Text and Image:

```
import {Text} from 'react-native';

// This is what you are looking for
const TextPropTypes = Text.propTypes; 
```

But beware that these prop types will stop being exported off of the components in a future release. We are moving away from providing prop-types for React Native in favor of typed JavaScript like TypeScript or Flow.

whiteHatCip:
Thanks for the quick response. I understand that the goal is to go through the path you pointed at in the response, but isn’t the purpose of proptypes different from the flow/typescript type checking one? I mean, proptypes provide runtime typechecking in a debug environment, which could be very useful when working with web services, if I am right, whilst typescript and flow only provide a compile time type checking. I mean, they offer a different service to the developer. Do I miss something?

PR #28885, opened 2020-05-13
set`NSAllowsArbitraryLoads` to false by default in template

wddwycc:
## Summary

Since we already have `localhost` as`Exception Domains` in NSAppTransportSecurity to allow connect to dev server, the template should set `NSAllowsArbitraryLoads` to false by default, as exaplained in [Apple's document](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowsarbitraryloads) as a good practice.

## Changelog

[iOS] [enhancement] - set `NSAllowsArbitraryLoads` to `false` by default in template

### DIFF
--- a/template/ios/HelloWorld/Info.plist
+++ b/template/ios/HelloWorld/Info.plist
@@ -26,8 +26,6 @@
 	<true/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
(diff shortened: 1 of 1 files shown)


shergin:
Even if it can complicate the setup of new apps, I think it's the right change to make. We should not ease platform-wise standard security restrictions by default.

PR #28887, opened 2020-05-13
Add `Animated.useValue` hook

iamolegga:
## Summary

When working with functional components this is easy to make bugs with statement, that I already saw several times from different people:

```js
function MyComponent() {
  const value = new Value(...);
}
```

This will lead to creating new value instance for every re-render. And components, that use internal value of previous instance will be reset to the new one, and this will lead to visual bugs. So one should always keep the same instance wrapping it either in `useRef(new Value(...))` or `useMemo(() => new Value(...), [])`.

This PR adds hook, that could be used in very often simple case of creating instance of Value without extra wrap:

```js
function MyComponent() {
  const value = useValue(...);
}
```

## Changelog

[General] [Added] - Animated.useValue hook

### DIFF
--- a/Libraries/Animated/src/AnimatedImplementation.js
+++ b/Libraries/Animated/src/AnimatedImplementation.js
@@ -29,6 +29,7 @@ const SpringAnimation = require('./animations/SpringAnimation');
 const TimingAnimation = require('./animations/TimingAnimation');
 
 const createAnimatedComponent = require('./createAnimatedComponent');
+const useValue = require('./useValue');
 
 import type {
   AnimationConfig,
@@ -707,6 +708,12 @@ module.exports = {
    * Expose Event class, so it can be used as a type for type checkers.
    */
   Event: AnimatedEvent,
+
+  /**
+   * React hook for keeping the same instance of AnimatedValue in functional
+   * components
+   */
+  useValue,
 
   __PropsOnlyForTests: AnimatedProps,
 };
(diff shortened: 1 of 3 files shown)


vamshi9666:
How is it different from

 `const animValue = useRef( new  Animated.Value(0)).current'`

iamolegga:
If one day someone will review this (😆) I just want you to know that same API is already [merged in react-native-reanimated](https://github.com/software-mansion/react-native-reanimated/pull/786), so it would be convenient for end-users to switch between libs when they need to, and only changing `import` statement

iamolegga:
Btw, CI errors are not correlated to current changes

iamolegga:
I think CI should be just triggered again by someone to pass
image

fabriziobertoglio1987:
I think that you can git force push to trigger again the tests. No change required.

sbearben:
Potentially this PR is better suited for: https://github.com/react-native-community/hooks ?

iamolegga:
@sbearben maybe, but as I've said before, it would be convenient to have same API for `Animated` from `react-native` and `react-native-reanimated`

javache:
This now exists as `useAnimatedValue` https://github.com/facebook/react-native/blob/main/Libraries/Animated/useAnimatedValue.js

iamolegga:
@javache this would be done 2.5 years ago if anyone could just [press one button](https://github.com/facebook/react-native/pull/28887#issuecomment-633029399) 🤦‍♂️

PR #28895, opened 2020-05-14
Enabling [-Werror,-Wunused-property-ivar]

aditya7fb:
Summary:
Changelog:
Enabling [-Werror,-Wunused-property-ivar]

Reviewed By: jdthomas

Differential Revision: D20961613

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -304,6 +304,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
 - (void)setShowSoftInputOnFocus:(BOOL)showSoftInputOnFocus
 {
+  (void)_showSoftInputOnFocus;
   if (showSoftInputOnFocus) {
     // Resets to default keyboard.
     self.backedTextInputView.inputView = nil;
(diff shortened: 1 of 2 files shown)


Libraries/Text/TextInput/RCTBaseTextInputView.m
```diff
@@ -304,6 +304,7 @@ - (void)setKeyboardType:(UIKeyboardType)keyboardType
 
 - (void)setShowSoftInputOnFocus:(BOOL)showSoftInputOnFocus
 {
+  (void)_showSoftInputOnFocus;
```
anandrajeswaran, line 307: Should we instead have the compiler avoid generating an ivar by adding an explicit getter? As it stands now if someone tries to access `showSoftInputOnFocus` they will always get `NO` regardless of what has been set.
```
- (BOOL)showSoftInputOnFocus
{
  return [[self backedTextInputView] inputView] == nil;
}
```
Alternatively, just declare a setter and not a property if it is not meant to be gotten

aditya7fb: I'm not familiar with the source code. I'm enabling the warning across the entire codebase for safeguard against future errors. Using `(void)decl;` seems like a safe approach. Could you suggest a patch i can use? I can also disable the warning in this build target as in the React/CoreModules below.

PR #28923, opened 2020-05-19
Add a test that loads all RNTester pages and improve reliability of RCTTestRunner

tom-un:
## Summary

This change adds an iOS integration test that loads each of the RNTester example pages ensuring that all the tests are free of RedBox errors on load.   Several of these test pages and some of the existing integration tests suffer intermittent failures due to unfinished RCTImageLoader tasks when the test loader deallocs.  

The `RNTesterLoadAllPages` is an `XCTestCase` sublcass that overrides the `defaultTestSuite` class method. In this method it run uses the `RCTTestRunner` to run a new JavaScript "test" in `RNTester.ios.js` called `EnumerateExamplePages`. This JavaScript test simply writes to console the key names of each test page in the `RNTesterList.ios.js` map.  In `defaultTestSuite` a bridge log block function captures the console and builds up an array of the test names. It creates an `XCTestCase` for each test and returns the tests. This method for dynamically creating the tests at runtime was borrowed from GTMGoogleTestRunner.

When each dynamically created `XCTestCase` is executed, it's selector IMP also uses `RCTTestRunner` to run an RNTester test named `"LoadPageTest_<key>"`. In `RNTesterApp.ios.js` instances of a new component LoadPageTest is created for each RNTester page and registered using the scheme `"LoadPageTest_<key>"` in a way similar to how the existing Snapshot tests are registered.

Some RNTester pages cannot be tested in this manner: some pages require native modules that only function in the full RNTester app (these pages have their own dedicated integration test code), don't function in the iOS Simulator, or have other issues. A new skipTest key is added the `RNTesterTypes` to allow individual tests to be skipped by providing a "reason" string.

`RCTImageLoader` is modified to expose a `activeTasks` readonly property that returns the value of the `_activeTasks` ivar.   `RCTTestRunnder`'s while loop that waits for the test to complete is modified to wait until activeTasks to be zero.

This contribution was implemented in the https://github.com/microsoft/react-native-macos fork and has proven to be very useful in catching regressions in RNTester examples and in the general reliability of the iOS and macOS integration tests.  https://github.com/microsoft/react-native-macos/issues/368

## Changelog

[iOS] [Fixed] - Add a test that loads all RNTester pages and improve reliability of RCTTestRunner

## Test Plan

Ran RNTester integration tests locally.

### DIFF
--- a/RNTester/RCTTest/RCTTestRunner.m
+++ b/RNTester/RCTTest/RCTTestRunner.m
@@ -14,6 +14,8 @@
 #import <React/RCTRootView.h>
 #import <React/RCTUIManager.h>
 #import <React/RCTUtils.h>
+#import <React/RCTBundleURLProvider.h>
+#import <React/RCTImageLoader.h>
 
 #import "FBSnapshotTestController.h"
 #import "RCTTestModule.h"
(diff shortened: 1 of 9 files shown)


RNTester/RCTTest/RCTTestRunner.m
```diff
@@ -199,8 +201,10 @@ - (void)runTest:(SEL)test module:(NSString *)moduleName
         configurationBlock(rootView);
       }
 
+      RCTImageLoader *imageLoader = [bridge moduleForClass:[RCTImageLoader class]];
```
tom-un, line 204: The Circle CI test `test_ios_unit_frameworks` is failing with a link failure Undefined symbol: `_OBJC_CLASS_$_RCTImageLoader`.   It would appear that either RCTTest.framework needs to have RCTImage added as a dep, or the change here in RCTTestRunner could be refactored to use `NSClassFromString(@"RCTImageLoader")` and use performSelector on activeTasks.

PR #28938, opened 2020-05-20
Add timeout support to `Networking`

ifsnow:
## Summary

As we all know, `fetch` has no built-in timeout option. But, there are many workarounds such as using `XMLHttpRequest API`, `AbortController` and `setTimeout + Promise`

Since it is rare to set a timeout for each API function, I thought I would like to have an easy way to set global settings without these workarounds.

Looking at the internal implementation, the final function called `RCTNetworking.sendRequest` has already a timeout parameter. Also, `Networking` module are exported externally.

I guess it would be nice to have a global setting function as shown below.

```javascript
// App.js
import { Networking } from 'react-native';

// Setting default timeout
Networking.setTimeout(5000);
```

If timeout is set in XMLHttpRequest, it does not affect and has backward compatibility. We can set a default timeout with `Networking.setTimeout` and have individual timeouts for each function.

```javascript
// Component.js
import axios from 'axios';

// Its timeout is 1 second, not 5 seconds.
const instance = axios.create({
  baseURL: 'https://some-domain.com/api/',
  timeout: 1000,
});
```

Please comment on a better implementation.

## Changelog

[General] [Added] - Add timeout support to `Networking`

## Test Plan

Tested in my development environment.

### DIFF
--- a/Libraries/Network/RCTNetworking.android.js
+++ b/Libraries/Network/RCTNetworking.android.js
@@ -41,6 +41,8 @@ function generateRequestId(): number {
  * requestId to each network request that can be used to abort that request later on.
  */
 class RCTNetworking extends NativeEventEmitter {
+  _timeout: number = 0;
+
   constructor() {
     super(NativeNetworkingAndroid);
   }
@@ -73,7 +75,7 @@ class RCTNetworking extends NativeEventEmitter {
       {...body, trackingName},
       responseType,
       incrementalUpdates,
-      timeout,
+      timeout || this._timeout,
       withCredentials,
     );
     callback(requestId);
@@ -85,6 +87,10 @@ class RCTNetworking extends NativeEventEmitter {
 
   clearCookies(callback: (result: boolean) => any) {
     NativeNetworkingAndroid.clearCookies(callback);
+  }
+
+  setTimeout(timeout: number) {
+    this._timeout = timeout;
   }
 }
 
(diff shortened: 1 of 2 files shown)


Eyesonly88:
Is there a global timeout handler? I feel like it could be useful to define a global timeout handler instead of doing it for each request. Something like `Networking.onTimeout = () => LogSomewhere()`. What do you think?

ifsnow:
@Eyesonly88 To make global handler what you said, we have to change implementation of `XMLHttpRequest` or `fetch(whatwg-fetch)`. I'm not sure if this is the best way everyone wants. I can do it if everyone wants. :)

Eyesonly88:
@ifsnow yep, good point. After investigation the native code more, I don't think there is a way to provide a timeout handler. We would have to add it to the native code first. So yeah, I agree with you. Let's keep it simple.

arcollector:
why not just create a race

```js
try {
await Promise.race([
  new Promise((_, reject) => setTimeout(() => reject('timeout'), 5000),
 fetch(...)
 ])
} catch(e) {
}
```

ifsnow:
@arcollector Thanks for your comments. As mentioned above, my approach would be to have a simpler way than implement different workarounds for each person. I think someone thinks your code is good and some are not.

1280103995:
How long do we have to wait?

Issue #27107, opened 2019-11-04
Android: Text component not selectable when removeClippedSubviews


cristianoccazinsp:
React Native version:

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 22.99 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    npm: 6.12.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 28
      Build Tools: 28.0.3
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 11.2/11B52 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.11.0 => 16.11.0 
    react-native: 0.61.3 => 0.61.3
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    create-react-native-module: 0.10.2
    react-native-cli: 2.0.1
```

## Steps To Reproduce

1. Render a `<Text selectable={true}>test</Text>` component inside a ScrollView with `removeClippedSubviews={true}` on Android
2. Try to copy paste text. Text is not selectable by long pressing

Describe what you expected to happen:
Text should be selectable regardless of removeClippedSubviews

Snack, code example, screenshot, or link to a repository:

xutm:
Any process？

cristianoccazinsp:
Still an issue

cristianoccazinsp:
Still an issue

fabriziobertoglio1987:
Thanks for the issue report. I tested your used case by adding `removeClippedSubviews={true}` to the `Animated.FlatList` component in my branch https://github.com/facebook/react-native/pull/28952

This issue is fixed with https://github.com/facebook/react-native/pull/28952 Thanks a lot

**<details><summary>CLICK TO OPEN TESTS RESULTS</summary>**
<p>

the onPressHandler is triggered when we set `onPressDisabled: false`

| **AFTER** | 
|:-------------------------:|
[...]

Issue #26264, opened 2019-08-30
Text in FlatList is not selectable on Android


maltoze:
## React Native version: 0.60.5
<details>

```
System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 317.94 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.3.1 - /usr/local/opt/nvm/versions/node/v12.3.1/bin/node
    Yarn: 1.16.0 - ~/.yarn/bin/yarn
    npm: 6.9.0 - /usr/local/opt/nvm/versions/node/v12.3.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    Android SDK:
      API Levels: 23, 28, 29
      Build Tools: 28.0.3, 29.0.1
      System Images: android-28 | Intel x86 Atom, android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom_64, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.5 => 0.60.5
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

</details>

## Steps To Reproduce
```javascript
export default class App extends React.Component {
  _renderItem = ({item}) => {
    return (
      <View>
        <Text selectable>{item}</Text>
      </View>
    )
  }

  _keyExtractor = (item, index) => index.toString();

  render() {
    return (
      <View style={styles.container}>
        <Text selectable>This is selectable Text...</Text>
        <FlatList
          data={['not selectable text', 'not selectable text']}
          renderItem={this._renderItem}
          keyExtractor={this._keyExtractor}
        />
      </View>
    );
  }
}
```

## Describe what you expected to happen:
Need to be able to select Text in FlatList.

## Snack, code example, screenshot, or link to a repository:
Snack: https://snack.expo.io/@maltoze/selectable-bug-in-flatlist

[...]

PR #28952, opened 2020-05-22
[Android] Fix non selectable Text in FlatList

fabriziobertoglio1987:
## Summary

This issue fixes https://github.com/facebook/react-native/issues/26264 fixes https://github.com/facebook/react-native/issues/27107
Text is not selectable inside a FlatList on Android. The solution is to invalidate the ReactTextView after a change of the selectable prop. If the view is visible, onDraw(android.graphics.Canvas) will be called at some point in the future and make the Text selectable.

## Changelog

[Android] [Fixed] - Fix non selectable Text in FlatList

## Test Plan

**<details><summary>CLICK TO OPEN TESTS RESULTS</summary>**
<p>

The issue was demonstrated in the following [snack](https://snack.expo.io/@fabrizio.bertoglio/selectable-bug-in-flatlist) (more info in issue https://github.com/facebook/react-native/issues/26264). 

The solution is:

1) Calling `invalidate()` from [setSelectableText][1] after changing the `selectable` prop and `mSelectableText` value. [`invalidate()`](https://developer.android.com/reference/android/view/View#invalidate()) triggers the `onDraw` callback.

[1]: https://github.com/fabriziobertoglio1987/react-native/blob/8027524947cafd5cbdc492e4ef9c92b346fe23fc/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextView.java#L427-L430

2) calling `setTextIsSelectable(mSelectableText);` from the [`onDraw`][2] callback

[2]: https://github.com/fabriziobertoglio1987/react-native/blob/8027524947cafd5cbdc492e4ef9c92b346fe23fc/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextView.java#L456-L460

The example below is availabe in RNTester FlatList example. Two options (`onPressDisabled` and `textSelectable`) have been added to test the functionality inside a FlatList. 

<p align="center">

</p>

</p>
</details>


fabriziobertoglio1987:
https://github.com/facebook/react-native/commit/372d001a5d5e24b68d194e01c356d973642fac86

ShikaSD:
Hey, just noticed this PR in the internal review queue :)

Not sure doing this in each onDraw here is a correct solution (for performance and otherwise).
If the underlying reason for this change was to delay the update, doing it in `postOnAnimation()/post()` (whatever works) is more correct approach. If it was caused by some other issues (e.g. cleared on view removal), maybe we can use `onAttachedToWindow` or some other callbacks to re-apply the value?
E.g.:

```
public void setSelectableText(boolean value) {
  postOnAnimation(new Runnable() {
    @Override
    public void run() {
      setTextSelectable(value)    
    }
  });
}
```

fabriziobertoglio1987:
after some investigation I notice that

The change was already applied with commit https://github.com/facebook/react-native/commit/89340f1620295c7ca6baecccc6b6d3befdf45cdb and then reverted later with commit https://github.com/facebook/react-native/commit/862136a398decff8ad5d4509ca51bd6b893ff851

https://github.com/facebook/react-native/commits/main?after=363a8fb08cef8326bba53cd3ea250bde02444126+34&branch=main&path%5B%5D=ReactAndroid&path%5B%5D=src&path%5B%5D=main&path%5B%5D=java&path%5B%5D=com&path%5B%5D=facebook&path%5B%5D=react&path%5B%5D=views&path%5B%5D=text&path%5B%5D=ReactTextView.java

cjhines:
Is this broken again on 0.68?

Zmwang622:
I'm on version 0.64.3, and Text still isn't selectable in a FlatList. Any ideas? @fabriziobertoglio1987 

This is my code:

```tsx
const MyClass = ({data}) => {
  const renderItem = useCallback(
      (item: key) => {
        return (
          <View>
              <Text selectable> {key} </Text>
              <Text selectable> {data[key]} </Text>
          </View>
        )
      }
   ,  [data])
  return <FlatList data={Object.keys(data)} renderItem={renderItem}
} 
```

fabriziobertoglio1987:
@Zmwang622 I think it is part of 0.67 https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0670

Zmwang622:
@fabriziobertoglio1987 Ah okay got it, thanks for the quick response!

markedwards:
I'm using react-native 0.72.4 and still need `removeClippedSubviews={false}` to enable text selection on Android.

Code-Victor:
Please, this issue isn't fixed yet. 
I just spent over an hour trying to fix this issue, thank God I found this.
https://github.com/facebook/react-native/issues/26264

LA-Johan:
Reopened the issue here: https://github.com/facebook/react-native/issues/46999

PR #28959, opened 2020-05-23
Improve XMLHttpRequest/Networking API performance

ifsnow:
## Summary

Probably, there're no apps that do not use `fetch`. `XMLHtttpRequest API`, implemented as `RCTNetworking` native module, uses 6 event listeners to process networking events. For each request, these 6 listeners are registered and removed after processing. I thought this was quite inefficient.

My implementation uses only one event listener. I worked with the following purpose.

### * Increases performance
When tested in my environments(iOS Simulator, Android device), there was about 20% improvement in performance. Here is the simple [benchmark code](https://gist.github.com/ifsnow/670fa52e1707b072722431b53dc4cb13) I used.

### * Keeps backward compatibility
In 0.62 version, `RCTNetworking` was exposed as a public `Networking` API. There may not be many, but I didn't want to affect the modules that use this API. So, I was working on using the improved event option only on the RN core.

------------

Please let me know if there is anything else I have to consider.

## Changelog

[General] [Added] - Improve XMLHttpRequest/Networking API performance

## Test Plan
I confirmed that `fetch` works well in my modified version.

### DIFF
--- a/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec-generated.mm
+++ b/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec-generated.mm
@@ -1741,7 +1741,7 @@ namespace facebook {
 
     
     static facebook::jsi::Value __hostFunction_NativeNetworkingAndroidSpecJSI_sendRequest(facebook::jsi::Runtime& rt, TurboModule &turboModule, const facebook::jsi::Value* args, size_t count) {
-      return static_cast<ObjCTurboModule&>(turboModule).invokeObjCMethod(rt, VoidKind, "sendRequest", @selector(sendRequest:url:requestId:headers:data:responseType:useIncrementalUpdates:timeout:withCredentials:), args, count);
+      return static_cast<ObjCTurboModule&>(turboModule).invokeObjCMethod(rt, VoidKind, "sendRequest", @selector(sendRequest:url:requestId:headers:data:responseType:useIncrementalUpdates:timeout:withCredentials:useImprovedEvent:), args, count);
     }
 
     static facebook::jsi::Value __hostFunction_NativeNetworkingAndroidSpecJSI_abortRequest(facebook::jsi::Runtime& rt, TurboModule &turboModule, const facebook::jsi::Value* args, size_t count) {
--- a/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
+++ b/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
@@ -1823,7 +1823,8 @@ namespace facebook {
        responseType:(NSString *)responseType
 useIncrementalUpdates:(BOOL)useIncrementalUpdates
             timeout:(double)timeout
-    withCredentials:(BOOL)withCredentials;
+    withCredentials:(BOOL)withCredentials
+   useImprovedEvent:(BOOL)useImprovedEvent;
 - (void)abortRequest:(double)requestId;
 - (void)clearCookies:(RCTResponseSenderBlock)callback;
 - (void)addListener:(NSString *)eventName;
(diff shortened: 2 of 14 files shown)


hey99xx:
> there was about 20% improvement in performance. Here is the simple benchmark code I used.

What are the results in terms of absolute numbers? Id expect networking to be more affected by network latency and maybe sending data over Native-JS bridge, rather than registering JS event listeners. So curious to see if there was significant wins here. 
20% of 10ms may not matter too much, 20% of 100ms could be actually important.

Btw not a FB engineer but couple notes for your code though:

1. Better to move the static hashset in `ResponseUtil` to a member variable of `NetworkingModule`. An app can have more than one RN instances, and network request ids per RN instance will always start from 0 as it's increased one by one on JS side. So multiple calls from different RN instances will share the same request id, and one request with id:1 may used improved events, and another request id:1 may not use improved events, but `ResponseUtil` maintaining a single data structure could break this. That's probably why `mRequestIds` is similarly in `NetworkingModule` atm.
2. Why keep both implementations anyway? If proved to be better, RN could default to this improved event registration and remove the previous one. For FB and others to test results in production, it might make sense to have a global flag that can be enabled from JS instead of making new default immediately.
3. Few minor typos using "Imporved" instead of "Improved". For example  in `removeImporvedEvent`.

ifsnow:
@hey99xx First of all, Thank you for your comment.

### Performance improvement points

You can see why it is efficient by using `MessageQueue.spy`.
```javascript
import MessageQueue from 'react-native/Libraries/BatchedBridge/MessageQueue';
MessageQueue.spy(true);
```

The following bridge communications occur for each `fetch` request.
```
JS->N : Networking.addListener(["didSendNetworkData"])
JS->N : Networking.addListener(["didReceiveNetworkResponse"])
JS->N : Networking.addListener(["didReceiveNetworkData"])
JS->N : Networking.addListener(["didReceiveNetworkIncrementalData"])
JS->N : Networking.addListener(["didReceiveNetworkDataProgress"])
JS->N : Networking.addListener(["didCompleteNetworkResponse"])
JS->N : Networking.sendRequest([{"method":"GET","url":"...."....])
N->JS : <callback for Networking.sendRequest>([2])
N->JS : RCTDeviceEventEmitter.emit(["didReceiveNetworkResponse", ....])
N->JS : RCTDeviceEventEmitter.emit(["didReceiveNetworkData", ...])
N->JS : RCTDeviceEventEmitter.emit(["didCompleteNetworkResponse", ...])
JS->N : Networking.removeListeners([1])
JS->N : Networking.removeListeners([1])
JS->N : Networking.removeListeners([1])
JS->N : Networking.removeListeners([1])
JS->N : Networking.removeListeners([1])
JS->N : Networking.removeListeners([1])
```

The improved version looks like this.
```
JS->N : Networking.addListener(["events"])
JS->N : Networking.sendRequest([{"method":"GET","url":"...."....])
N->JS : <callback for Networking.sendRequest>([2])
N->JS : RCTDeviceEventEmitter.emit(["events", ....])
N->JS : RCTDeviceEventEmitter.emit(["events", ...])
N->JS : RCTDeviceEventEmitter.emit(["events", ...])
JS->N : Networking.removeListeners([1])
```

Minor but also includes code optimization. 
```javascript
this._didReceiveResponse(args);
// original code: this._didReceiveResponse(...args);
```
The above code is compiled by `Babel` as follows. In general, The apply call is less effective than the direct call.
```javascript
this._didReceiveResponse.apply(this, args);
```

As you said, the difference in performance may be smaller than expected. But, I think there is no reason not to use it even though there is a more efficient way.

### ResponseUtil Implementation

Networking API is exposed as a class instance. It's a singletone.
```javascript
module.exports = (new RCTNetworking(): RCTNetworking);
```
This is the reason for communication and processing based on `requestId`.

### Why backward compatibility is needed
As mentioned above, there are modules using `Netwokring` API. I'm not sure if we can make a breaking change. If possible, I want to maintain backward compatibility.

### Typos
Fixed it. Thanks.

PR #28961, opened 2020-05-25
Enable array buffers in JSCRuntime.cpp

ryantrem:
## Summary

The JavaScriptCore implementation of JSI [does not currently support array buffers](https://github.com/facebook/react-native/blob/master/ReactCommon/jsi/JSCRuntime.cpp#L925-L943). The comments in the code suggest the JSC version used by React Native does not work with array buffers, but this seems to be out of date since the current version of JSC used by React Native does indeed support array buffers. This change just enables array buffers in JSCRuntime.cpp.

NOTE: See https://github.com/react-native-community/discussions-and-proposals/issues/91#issuecomment-632371219 for more background on this change.

## Changelog

[General] [Added] - Support for array buffers in the JavaScriptCore implementation of JSI

## Test Plan

To test these changes, I just made some temporary changes to RNTester to use JSI to inject a test function into the JS runtime that reads from and writes to an array buffer, and call that function from the JS of the RNTester app (see https://github.com/ryantrem/react-native/commit/28152ce3f4ae0fa906557415d106399b3f072118).

For the JS side of this, specifically look at https://github.com/ryantrem/react-native/blob/28152ce3f4ae0fa906557415d106399b3f072118/RNTester/js/RNTesterApp.android.js#L13-L18

For the native side of this, specifically look at https://github.com/ryantrem/react-native/blob/28152ce3f4ae0fa906557415d106399b3f072118/RNTester/android/app/src/main/cpp/JSITest.cpp#L22-L38

### DIFF
--- a/ReactCommon/jsi/JSCRuntime.cpp
+++ b/ReactCommon/jsi/JSCRuntime.cpp
@@ -922,24 +922,17 @@ bool JSCRuntime::isArray(const jsi::Object &obj) const {
 #endif
 }
 
-bool JSCRuntime::isArrayBuffer(const jsi::Object & /*obj*/) const {
-  // TODO: T23270523 - This would fail on builds that use our custom JSC
-  // auto typedArrayType = JSValueGetTypedArrayType(ctx_, objectRef(obj),
-  // nullptr);  return typedArrayType == kJSTypedArrayTypeArrayBuffer;
-  throw std::runtime_error("Unsupported");
+bool JSCRuntime::isArrayBuffer(const jsi::Object &obj) const {
+  auto typedArrayType = JSValueGetTypedArrayType(ctx_, objectRef(obj), nullptr);
+  return typedArrayType == kJSTypedArrayTypeArrayBuffer;
 }
 
-uint8_t *JSCRuntime::data(const jsi::ArrayBuffer & /*obj*/) {
-  // TODO: T23270523 - This would fail on builds that use our custom JSC
-  // return static_cast<uint8_t*>(
-  //    JSObjectGetArrayBufferBytesPtr(ctx_, objectRef(obj), nullptr));
-  throw std::runtime_error("Unsupported");
+uint8_t *JSCRuntime::data(const jsi::ArrayBuffer &obj) {
+  return static_cast<uint8_t*>(JSObjectGetArrayBufferBytesPtr(ctx_, objectRef(obj), nullptr));
 }
 
-size_t JSCRuntime::size(const jsi::ArrayBuffer & /*obj*/) {
-  // TODO: T23270523 - This would fail on builds that use our custom JSC
-  // return JSObjectGetArrayBufferByteLength(ctx_, objectRef(obj), nullptr);
-  throw std::runtime_error("Unsupported");
+size_t JSCRuntime::size(const jsi::ArrayBuffer &obj) {
+  return JSObjectGetArrayBufferByteLength(ctx_, objectRef(obj), nullptr);
 }
 
 bool JSCRuntime::isFunction(const jsi::Object &obj) const {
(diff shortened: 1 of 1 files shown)


ryantrem:
FYI @mhorowitz, @tmikov, @bghgary.

ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -922,24 +922,17 @@ bool JSCRuntime::isArray(const jsi::Object &obj) const {
...
+uint8_t *JSCRuntime::data(const jsi::ArrayBuffer &obj) {
+  return static_cast<uint8_t*>(JSObjectGetArrayBufferBytesPtr(ctx_, objectRef(obj), nullptr));
 }
 
-size_t JSCRuntime::size(const jsi::ArrayBuffer & /*obj*/) {
-  // TODO: T23270523 - This would fail on builds that use our custom JSC
-  // return JSObjectGetArrayBufferByteLength(ctx_, objectRef(obj), nullptr);
-  throw std::runtime_error("Unsupported");
+size_t JSCRuntime::size(const jsi::ArrayBuffer &obj) {
+  return JSObjectGetArrayBufferByteLength(ctx_, objectRef(obj), nullptr);
```
shergin, line 935: These functions are unavailable on iOS 9, can we gate those lines?

```
xplat/js/react-native-github/ReactCommon/jsi/JSCRuntime.cpp:935:10: error: 'JSObjectGetArrayBufferByteLength' is only available on iOS 10.0 or newer [-Werror,-Wunguarded-availability]
  return JSObjectGetArrayBufferByteLength(ctx_, objectRef(obj), nullptr);
```

ryantrem: Can you point me to any existing examples in the React Native code base? How does the build work where this error is coming from? Are there different builds of the React Native code for different iOS versions?

ryantrem: I found an example already in this file, so basically followed the same pattern.

ReactCommon/jsi/JSCRuntime.cpp
```diff
@@ -276,6 +276,9 @@ class JSCRuntime : public jsi::Runtime {
 #if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_9_0
```
ryantrem, line 276: Worth noting, this existing define looks wrong to me. I believe `_JSC_FAST_IS_ARRAY` will never be defined on Android, so will always use the slow version. Presumably this should be fixed, though I didn't really want to overload this PR.

ReactCommon/jsi/JSCRuntime.cpp
shergin: Awesome! Thank you so much!

ryantrem:
From what I can tell, there is no way for me to see what the warning is that is causing the "Facebook Internal - Builds & Tests" check to fail, is that right? Assuming yes, can someone (@shergin?) provide some details if there is something more I need to address for this?

ryantrem:
Ping! @shergin, @mhorowitz, @tmikov - can anyone help here? I'd like to see this PR merged, but I don't know what the build warning is. I don't get any build warnings in either XCode or Android Studio when building locally.

tmikov:
AFAICT, it is a clang-format warning on line 944. The line is too long and needs to be reformated.

ryantrem:
@shergin or @tmikov - thanks for your help so far on this PR. Can one of you re-"import" this PR after my latest change (which I believe should address the issue @tmikov mentioned above)?

tmikov:
@ryantrem the original importer "owns" the diff internally, so ordinarily only they can re-import them. I will reach out to @shergin internally.

ryantrem:
Hi again @tmikov and @shergin - sorry to nag, but since another week has passed I wanted to follow up on this and see if there has been any progress, or if you need anything else from me.

tmikov:
@ryantrem sorry for the delay. We discussed it with @shergin  and we will proceed to land it shortly.

PR #28977, opened 2020-05-26
[Android][Flatlist] Fix last item removal scroll position

jayshah123:
## Summary

Currently on Android, when last item in a horizontal flatlist is removed, scroll position does not update.

Fixes https://github.com/facebook/react-native/issues/27504

## Changelog

[Android] [Changed] - HorizontalScrollView last item removal will adjust scroll position

## Test Plan

This is the same code which is being used and already proven for vertical scrollview.


sanduluca (approved):
Looks very good to me.

jayshah123:
will re-raise this after some due diligence.

PR #28984, opened 2020-05-26
Make sure LogBox is not included in production bundles

janicduplessis:
## Summary

I was doing some bundle inspection and noticed the LogBox module was included. It does amount to around 15kb so I think it is worth making sure it is not there.

To fix it I moved some imports to require inside __DEV__ blocks to make sure metro is able to remove these imports.

## Changelog

[General] [Fixed] - Make sure LogBox is not included in production bundles

## Test Plan

Tested using react-native-bundle-visualizer and made sure nothing from LogBox was included in the bundle after these changes.

### DIFF
--- a/Libraries/LogBox/LogBox.js
+++ b/Libraries/LogBox/LogBox.js
@@ -12,8 +12,6 @@
 
 import Platform from '../Utilities/Platform';
 import RCTLog from '../Utilities/RCTLog';
-import * as LogBoxData from './Data/LogBoxData';
-import {parseLogBoxLog, parseInterpolation} from './Data/parseLogBoxLog';
 
 import type {IgnorePattern} from './Data/LogBoxData';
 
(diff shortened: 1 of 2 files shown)


Libraries/LogBox/LogBox.js
```diff
@@ -12,8 +12,6 @@
 
 import Platform from '../Utilities/Platform';
 import RCTLog from '../Utilities/RCTLog';
-import * as LogBoxData from './Data/LogBoxData';
```
TheSavior, line 15: This should be handled by the inline requires feature. We don't like manually inlining requires if we can avoid it.

janicduplessis: Sadly this is not enabled by default in OSS so I think doing it manually in this case is fine since it is not only for perf improvements.

PR #29002, opened 2020-05-28
Remove usage of legacy context API in modal

### LINKED ISSUE #28103: Migrate off of old context APIs

satya164:
## Summary

Remove usage of the legacy context API in Modal in favor of the new context API. Closes #28103

## Test Plan

Run RNTester app and test the Modal example. Also add a `console.warn` to make sure we get the correct context value.

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -11,13 +11,14 @@
 'use strict';
 
 const AppContainer = require('../ReactNative/AppContainer');
+const {RootTagContext} = require('../ReactNative/RootTag');
 const I18nManager = require('../ReactNative/I18nManager');
-const PropTypes = require('prop-types');
 const React = require('react');
 const ScrollView = require('../Components/ScrollView/ScrollView');
 const StyleSheet = require('../StyleSheet/StyleSheet');
 const View = require('../Components/View/View');
 
+import type {RootTag} from '../ReactNative/RootTag';
 import type {ViewProps} from '../Components/View/ViewPropTypes';
 import type {DirectEventHandler} from '../Types/CodegenTypes';
 import type EmitterSubscription from '../vendor/emitter/EmitterSubscription';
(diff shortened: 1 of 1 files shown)


Libraries/ReactNative/AppContainer.js
```diff
@@ -47,18 +44,6 @@ class AppContainer extends React.Component<Props, State> {
 
   static getDerivedStateFromError: any = undefined;
 
-  static childContextTypes:
```
TheSavior, line 50: This would break what is available to children as part of the old context api. I think that would make this a breaking change forcing everyone to use the new RootTagContext. We can't make that breaking change yet at FB as we still have too many callsites using that. There is a side effort internally to get everything off the legacy context for RootTag though. This might be blocked until then.

TheSavior: Could we make the change to Modal without making this change?

satya164: Ok I'll add it back tomorrow

satya164: Updated the PR

PR #29004, opened 2020-05-28
Fail to send unless readyState is CONNECTED

JunielKatarn:
## Summary

According to the [IETF standard](https://tools.ietf.org/html/rfc6455#section-6.1) and the [JS API spec](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send) for WebSockets, the `send` method should fail if the socket's state is not `CONNECTED`.
The React Native JavaScript implementation of WebSocket only fails if the state is `CONNECTING`, allowing for `send` to be called in both `CLOSING` or `CLOSED` states.

## Changelog

[General] [Changed] - Fail on WebSocket.send unless state is CONNECTED.

## Test Plan

Run existing tests unit and integration tests.

### DIFF
--- a/Libraries/WebSocket/WebSocket.js
+++ b/Libraries/WebSocket/WebSocket.js
@@ -163,7 +163,7 @@ class WebSocket extends (EventTarget(...WEBSOCKET_EVENTS): any) {
   }
 
   send(data: string | ArrayBuffer | ArrayBufferView | Blob): void {
-    if (this.readyState === this.CONNECTING) {
+    if (this.readyState !== this.CONNECTED) {
       throw new Error('INVALID_STATE_ERR');
     }
 


rickhanlonii:
Update: Looks like there are a few tests failing (here and our version of these tests internally).

JunielKatarn:
> Update: Looks like there are a few tests failing (here and our version of these tests internally).

Noted. If it also happens in FB internal runs, this may change may either not be correct, or surface bugs, or tests need to be updated.

I'll try to look at it this week.

martinheidegger:
Should this check also be used in `.ping`? https://github.com/facebook/react-native/pull/29004/files#diff-c864b631ae0cae13be21f9aeebeea020faabfa3982bb41223478eb4c85b40ae1R193

PR #29006, opened 2020-05-29
mark 'force' as optional property of PressEvent object

Simek:
## Summary

Refs: 
* https://github.com/facebook/react-native/blob/master/React/Fabric/RCTSurfaceTouchHandler.mm#L93
* https://github.com/facebook/react-native/blob/master/RNTester/js/examples/Pressable/PressableExample.js#L25

According to the current implementation `force` will be returned only on `iOS` platform, only on the devices which have a hardware 3D touch capabilities.

## Changelog

[General] [Fixed] - mark `force` as an optional property of the PressEvent object

## Test Plan

Go CI!

### DIFF
--- a/Libraries/Types/CoreEventTypes.js
+++ b/Libraries/Types/CoreEventTypes.js
@@ -88,7 +88,7 @@ export type TextLayoutEvent = SyntheticEvent<
 export type PressEvent = ResponderSyntheticEvent<
   $ReadOnly<{|
     changedTouches: $ReadOnlyArray<$PropertyType<PressEvent, 'nativeEvent'>>,
-    force: number,
+    force: ?number,
     identifier: number,
     locationX: number,
     locationY: number,
--- a/Libraries/Types/CoreEventTypes.js
+++ b/Libraries/Types/CoreEventTypes.js
@@ -94,7 +94,7 @@ export type PressEvent = ResponderSyntheticEvent<
     locationY: number,
     pageX: number,
     pageY: number,
-    target?: number,
+    target: ?number,
     timestamp: number,
     touches: $ReadOnlyArray<$PropertyType<PressEvent, 'nativeEvent'>>,
   |}>,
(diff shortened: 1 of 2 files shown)


TheSavior:
Thanks for working on this. Can you verify at runtime that this key doesn't exist on Android? The code you linked to isn't really definitive. For example, I could imagine Android is currently providing the `force` key, but it always being `1`. 

I'd recommend getting the event and console logging the keys on iOS showing that it includes `force`, and then running the same code on Android and showing that it does not include the key. That's what I'd expect to see in your test plan.

Simek:
@TheSavior As I were writing earlier this key is not even present on iOS, on the devices without hardware 3D Touch support, so I don't think that screenshots from Android are necessary.

### iPhone SE 2nd (no `force`)
Screenshot 2020-05-29 at 23 44 46

### iPhone 11 Pro (no `force`)
Screenshot 2020-05-29 at 23 47 01

### iPhone 8 (`force`)
Screenshot 2020-05-29 at 23 49 16

Libraries/Types/CoreEventTypes.js
```diff
@@ -88,7 +88,7 @@ export type TextLayoutEvent = SyntheticEvent<
 export type PressEvent = ResponderSyntheticEvent<
   $ReadOnly<{|
     changedTouches: $ReadOnlyArray<$PropertyType<PressEvent, 'nativeEvent'>>,
-    force: number,
+    force: ?number,
```
TheSavior, line 91: Since this key might not be present the question mark should be on the left side. As you have it right now it says "force will always be there, but the value may be undefined, null, or a number." What you want it to say is "force may not be there, but if it is it will be a number". That is written as `force?: number`.

TheSavior:
Thanks for those screenshots, that's great and makes it really clear what is happening.

Libraries/Types/CoreEventTypes.js
Simek: So `target` is also wrong?

TheSavior: I'm not sure. It's possible `target` is intending the behavior it says.

Simek: I think that I had saw an event without `target` during earlier tests, but I have no repro right now. I will open another PR if I will be able to confirm this.

PR #29011, opened 2020-05-30
Test fixes related to changes in Jest's Fake Timer behavior

makitake2013:
## Summary
Fixed tests related to changing Jest's fake timer behavior. 
And some of the test code in LogBoxData-test.js itself was incorrect so I am fixing it.

LogBoxData-test.js fails when I try to build a new local development environment and fork the React Native source to run a JavaScript test.
The error message is:
runAllImmediates is not available when using modern timers

After checking, runAllImmediates could not be used with the following modification.
https://github.com/facebook/jest/commit/71631f6bf9ccf8937f02a5ecfb64b62712b86c19

## Changelog

[Internal] [Fixed] - Test fixes related to changes in Jest's Fake Timer behavior

## Test Plan
Ran JavaScript Tests locally.

### DIFF
--- a/Libraries/LogBox/Data/__tests__/LogBoxData-test.js
+++ b/Libraries/LogBox/Data/__tests__/LogBoxData-test.js
@@ -137,7 +137,6 @@ beforeEach(() => {
 });
 
 const flushLogs = () => {
-  jest.runAllImmediates();
   jest.runAllTimers();
 };
 
@@ -240,7 +239,7 @@ describe('LogBoxData', () => {
     expect(logs[3].category).toEqual('D');
 
     addLogs(['A']);
-    jest.runAllImmediates();
+    flushLogs();
 
     // Expect `A` to be added to the end of the registry.
     logs = registry();
@@ -280,7 +279,6 @@ describe('LogBoxData', () => {
 
     // Order maters for symbolication before timeout.
     flushLogs();
-    jest.runAllTimers();
 
     expect(selectedLogIndex()).toBe(0);
 
@@ -289,7 +287,6 @@ describe('LogBoxData', () => {
 
     // Order maters for symbolication before timeout.
     flushLogs();
-    jest.runAllTimers();
 
     // This should still be 0 (the first fatal exception)
     // becuase it is the most likely source of the error.
@@ -302,7 +299,6 @@ describe('LogBoxData', () => {
     addFatalErrors(['A']);
 
     // Order maters for timeout before symbolication.
-    jest.runAllTimers();
     flushLogs();
 
     expect(selectedLogIndex()).toBe(0);
@@ -311,7 +307,6 @@ describe('LogBoxData', () => {
     addFatalErrors(['C']);
 
     // Order maters for timeout before symbolication.
-    jest.runAllTimers();
     flushLogs();
 
     // This should still be 0 (the first fatal exception)
@@ -546,12 +541,12 @@ describe('LogBoxData', () => {
 
     const lastLog = Array.from(observer.mock.calls[1][0].logs)[0];
     LogBoxData.dismiss(lastLog);
-    jest.runAllImmediates();
+    flushLogs();
     expect(observer.mock.calls.length).toBe(3);
 
     // Does nothing when category does not exist.
     LogBoxData.dismiss(lastLog);
-    jest.runAllImmediates();
+    flushLogs();
     expect(observer.mock.calls.length).toBe(3);
   });
 
@@ -564,12 +559,12 @@ describe('LogBoxData', () => {
     expect(observer.mock.calls.length).toBe(2);
 
     LogBoxData.clear();
-    jest.runAllImmediates();
+    flushLogs();
     expect(observer.mock.calls.length).toBe(3);
 
     // Does nothing when already empty.
     LogBoxData.clear();
-    jest.runAllImmediates();
+    flushLogs();
     expect(observer.mock.calls.length).toBe(3);
   });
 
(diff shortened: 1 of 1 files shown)


Libraries/LogBox/Data/__tests__/LogBoxData-test.js
```diff
@@ -311,7 +307,6 @@ describe('LogBoxData', () => {
     addFatalErrors(['C']);
 
     // Order maters for timeout before symbolication.
-    jest.runAllTimers();
     flushLogs();
```
rickhanlonii, line 310: The code here is:

```
// Order maters for timeout before symbolication.
jest.runAllTimers();	
flushLogs();
```

The first line `jest.runAllTimers()` runs the first timer to expiration. This should trigger a log to get scheduled for addition, which schedules a new timer to update the logs data. This requires another `jest.runAllTimers()` to run the new timers created when the old timers ran. This is done with `flushLogs()`.

I'm not sure what changed in Jest, but this test should fail with the changes here (so either the tests are wrong, or Jest is wrong).

rickhanlonii: @thymikee in the new timer system, do we run all timers including timers that are created when a timer is run?

E.g. what should this print?

```
setTimeout(() => {
  console.log(1);
  setTimeout(() => {
      console.log(2);
  }, 0);
}, 0);

jest.runAllTimers();
```

thymikee: I'll defer that to @SimenB

SimenB: That should log `1` then `2`, for both legacy and modern timers

makitake2013: @rickhanlonii @thymikee 
Thank you for your reviewing.
I confirmed the behavior of the timer system,
2 was output by runAllTimers run once.

スクリーンショット 2020-06-22 18 28 23
スクリーンショット 2020-06-22 18 37 54

rickhanlonii: @makitake2013 if you delete this same line in the existing tests without this change, the test will fail correct?

makitake2013: @rickhanlonii Yes, the test failed.
スクリーンショット 2020-06-25 10 01 29
スクリーンショット 2020-06-25 10 02 05

makitake2013: @rickhanlonii In the original test case, running runAllImmediates gives an error.
However, if I remove runAllImmediates and run only runAllTimers, the test still fails.
So I wonder if the original test case wasn't working properly.
スクリーンショット 2020-06-25 10 11 15
スクリーンショット 2020-06-25 10 26 00

rickhanlonii: The way I think it's supposed to work is with what I said in my original comment in this thread. Running the timers schedules another timer so we have to run them twice.

makitake2013: @rickhanlonii Thank you.Maybe I understand. At the time of writing the existing test, I think it was necessary to run the Timer multiple times.
For example, in the screen capture above, 2 is the expected value in  the existing test, but currently 3 is the actual value.
In the existing test, the timer run once, so 2 was the actual value, but now 3 is the actual value because all timers will operate when run once.

rickhanlonii: >  because all timers will operate when run once

Yeah this is what I'm concerned about, this would be a breaking change in Jest and I'm not sure that the behavior should change?

makitake2013: @rickhanlonii 
Thank you for your reply.
I also have some concerns.
Is this test still working properly in your development environment?

When I forked the latest version of the React Native source and run jest, only this test case failed.
I wonder if existing developers or the CI environment have old version of Jest in , but is it also my environment configuration issue?
If possible, please reinstall node_modules and give it a try.

I think it should not be fixed it if it's a configuration issue, and if this problem will be occurred when fork the latest version and create a new development environment, it should be fixed, thought.

SimenB: There are behavioral changes between "modern" and "legacy" timers, mostly laid out in the OP here: https://github.com/facebook/jest/pull/5171.

I don't really follow this discussion - could any of you put together a snippet that changes behavior with modern timers in a breaking way you think is wrong?

rickhanlonii: @SimenB sorry for the confusion, the snippets concerned is: 

```js
setTimeout(() => {
  console.log(1);
  setTimeout(() => {
      console.log(2);
  }, 0);
}, 0);
```

After reading up, it sounds like we changed it so that `jest.runAllTimers` runs any scheduled timers as well. That seems to make sense, when you `runAllTimers` you want to flush all of them (even the new ones scheduled).

For these test though, we need a way to run only the pending timers, so we can advance a nested timer as in:

```
// Schedule a nested timer.
jest.runOnlyPendingTimers();

// Advance the newly scheduled timer.
jest.advanceTimersByTime(1001); 

// Flush remaining timers.
jest.runAllTimers(); 
```

So my new concern is from what you linked:

> [BREAKING] `runAllPendingTimers` in Jest currently do not trigger new timers scheduled while running to the last scheduled one. Lolex does (which matches what would happen in real code). 

Did you mean to say `runAllTimers` here? If we changed `runAllPendingTimers` then the docs [say the opposite](https://jestjs.io/docs/en/timer-mocks#run-pending-timers) of what it does and I don't think I can do the above.

rickhanlonii: @makitake2013 I pushed a branch with how [I recommend that we fix this here](https://github.com/facebook/react-native/compare/master...rickhanlonii:rh-fix-logbox-test-for-jest-26?expand=1).

Do you want to copy these changes over and test?

To fix, I updated the test so that:
- The flushes mostly use `runOnlyPendingTimers`
- Remove `runAllImmediates`
- Remove dependencies on `runAllTimers` behavior with scheduled timers
- Added `flushToObservers` for when we were calling `flushLogs` just to update observers
- Updated so the right number of flushes are done automatically for each log (e.g. errors need 2)
- Updated the optimistic symbolication tests to work with Jest 26 timers

This _should_ work with the Jest 26 timers but I didn't test myself.

makitake2013: @rickhanlonii Thank you so much for your reviewing. 
I confirmed that the test you modified works on Jest 26 and updated.

rickhanlonii:
Thanks @makitake2013!

makitake2013:
Thanks @rickhanlonii ! 
This was my first pull request. I learned a lot and it was fun, so I would like to try again!

rickhanlonii:
Please do!

PR #29012, opened 2020-05-31
Better error message when missing entry file 

petrbela:
## Summary

If the entry file has been renamed (e.g., to `index.ts`), the script currently fails silently and prints out that the bundle file does not exist.

This change will print the correct error message instead.

## Changelog

[iOS] [Fixed] - Better error message when missing entry file 

## Test Plan

N/A

### DIFF
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -67,9 +67,9 @@ if [[ "$ENTRY_FILE" ]]; then
   # Use ENTRY_FILE defined by user
   :
 elif [[ -s "index.ios.js" ]]; then
-   ENTRY_FILE=${1:-index.ios.js}
+  ENTRY_FILE=${1:-index.ios.js}
 else
-   ENTRY_FILE=${1:-index.js}
+  ENTRY_FILE=${1:-index.js}
 fi
 
 if [[ $DEV != true && ! -f "$ENTRY_FILE" ]]; then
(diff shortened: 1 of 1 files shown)


petrbela:
The failing tests seem to be unrelated... Is there anything I need to do about them?

thymikee:
You can rebase to latest master and see if that helps. But anyway, now is the moment we wait for a Facebook employee to import it and verify

scripts/react-native-xcode.sh
```diff
@@ -67,9 +67,14 @@ if [[ "$ENTRY_FILE" ]]; then
...
-   ENTRY_FILE=${1:-index.ios.js}
- else
-   ENTRY_FILE=${1:-index.js}
+  ENTRY_FILE=${1:-index.ios.js}
+else
+  ENTRY_FILE=${1:-index.js}
+fi
+
+if [[ $DEV != true && ! -f "$ENTRY_FILE" ]]; then
+  echo "error: Entry file $ENTRY_FILE does not exist. If you use another file as your entry point, pass ENTRY_FILE=myindex.js" >&2
```
kiranjd, line 76: I think it's a typo. 
It should have been: `"Error: entry file...`

petrbela: Not sure what should be correct but `error: File ...` case is already used towards the end of this file so I followed the same pattern.

kiranjd:
I think you don't need to capitalize the word after colon(:)

Kiran JD

PR #28170, opened 2020-02-25
Update Android Gradle plugin to 3.6.4

friederbluemle:
## Summary

Android Studio 3.6.3 is now available in the stable channel

https://androidstudio.googleblog.com/2020/04/android-studio-363-available.html

EDIT: Another patch update to `3.6.4` was released in July 2020:

> This minor update supports compatibility with new default settings and features for package visibility in Android 11.

Details: https://developer.android.com/studio/releases/gradle-plugin#3-6-0

## Changelog

[Android] [Changed] - Update Android Gradle plugin to 3.6.4

## Test Plan

Build project

PR #29013, opened 2020-05-31
Update Android Gradle plugin to 4.0.1

friederbluemle:
This is a major version update that needs to be tested thoroughly.

## Summary

Android Studio 4.0.1 is now available in the stable channel

https://androidstudio.googleblog.com/2020/05/android-studio-40-available-in-stable.html
https://developer.android.com/studio/releases/gradle-plugin#4.0.1

## Changelog

[Android] [Changed] - Update Android Gradle plugin to 4.0.1

## Test Plan

Build project

Closes #29044

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,7 +12,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:3.6.4")
+        classpath("com.android.tools.build:gradle:4.0.1")
         classpath("de.undercouch:gradle-download-task:4.0.2")
 
         // NOTE: Do not place your application dependencies here; they belong
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -13,7 +13,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:3.6.4")
+        classpath("com.android.tools.build:gradle:4.0.1")
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }


friederbluemle:
Rebased, since #28170 was merged. Ready for early review/testing, not ready for merge.

dulmandakh:
Please revert NDK change. It works fine for me on RN 0.63 app.

dulmandakh:
Let's merge this before cutting 0.64, which should be coming soon 😇

fkgozali:
> Please revert NDK change. It just works fine for me on RN 0.63 app.

waiting for this update then I'll merge

friederbluemle:
> Please revert NDK change. It just works fine for me on RN 0.63 app.

You mean the older version works fine? Did you try the newer one? The default for `4.0.1` is `21.0.6113669`, according to this page:
https://developer.android.com/studio/projects/install-ndk#default-ndk-per-agp

fkgozali:
If we keep the latest: `ndkVersion = "21.0.6113669"`, do we need to update the helloworld android template too?

friederbluemle:
The project in `template/` is the only build.gradle file that _currently_ (c015f485c2) has a `ndkVersion` specified. Two more will be added via #29987.
- If #29013 ends up getting merged _first_, then #29987 needs to be updated
- If #29987 gets merged first, then I'll rebase and update this PR here accordingly

gedeagas:
@friederbluemle do you have any reason to update the NDK version other than it is the default version for AGP 4.0.1? 

cc @dulmandakh I think we should keep notes that once this pr ( with updated NDK version ) merge we need to update our CI NDK version to `21.0.6113669`  to ensure that a specific state of the repository is always built with the same exact tools.

dulmandakh:
AFAIK, CI still use NDK 20.x. A while ago, I tried to bump it to 21, but failed. So I propose to use same NDK version for React Native and RN apps, even we don't have known issues yet. For me, it's one less NDK version to install on my system at least.

gedeagas:
I agree with you @dulmandakh, if we don't have known issues, or need specific function I think it's better to stick with the current NDK version.

fkgozali:
Yeah, let's keep all versions in the repo consistent, and if an upgrade is needed, we'll do it in separate PR. I think it's also important to put this version down in one place so that all configs depending on it can get the same exact version (perhaps we should just add a constant in react.gradle?).

fkgozali:
@friederbluemle: thoughts?

dulmandakh (approved):

friederbluemle:
Sure, as requested, I removed the alignment of the ndk from this PR again.

The main reason why I initially wanted to align the NDK with the AGP version is that the website linked above states:

> Before release, each AGP version is thoroughly tested with the latest stable NDK release at that time.

In this case, if everything works fine with the older version, I guess we can update the AGP first.

friederbluemle:
@dulmandakh 
>AFAIK, CI still use NDK 20.x. A while ago, I tried to bump it to 21, but failed. So I propose to use same NDK version for React Native and RN apps, even we don't have known issues yet. For me, it's one less NDK version to install on my system at least.

Yes, definitely agree we should use the same NDK version for React Native and RN apps. Ideally, it should also be the same (default) version that the AGP uses (and is thoroughly tested by Google) - If there is some problem with the newer NDK, then of course - we should _not_ update it yet.

@gedeagas
> if we don't have known issues, or need specific function I think it's better to stick with the current NDK version.

Yes, I also agree with that! Although the same _could_ be said about the update of the AGP as well "if we don't have an issue it's better to stick with the current version" ;)

Either way, I'm fine updating AGP first, and NDK later - just wanted to make sure this doesn't fall off the radar because everything is "working fine". Does anyone have logs/details on what issues there are with the current repo and NDK 21? Should we open an issue to at least keep track of it?

fkgozali:
Thanks, I'll merge this ASAP. Re: NDK v21, let's get this set up in CI so that we can upgrade the versions to 21 all at once. cc @hramos re: CI.

friederbluemle:
Thanks @fkgozali :+1:

AndrewMorsillo:
Android gradle plugin v4+ seems to be causing some issues. JS bundles are failing to get copied into the aab or apk output seemingly randomly (running build twice sometimes adds the js bundle to the output) https://github.com/facebook/react-native/issues/29398

Sorry to comment on a closed pr @fkgozali @friederbluemle but this change may bite a lot of people if it goes out in a release so I thought this would be the best place to alert those likely to know how to mitigate it.

fkgozali:
I'll defer to @friederbluemle and @dulmandakh for recommendation.

AndrewMorsillo:
Seems there's a change in task ordering behavior somewhere that causes bundle to get copied after assets are merged so bundle gets left out of assets. See here: https://github.com/react-native-community/upgrade-support/issues/38#issuecomment-708543814

friederbluemle:
@AndrewMorsillo - Thanks for the report and sorry to hear this update is causing some issues on your end.
Is there an easy way to set up a test project that reproducibly shows this problem? Unfortunately I won't have much in the next week or two to investigate this further. So unless @dulmandakh or someone else can come up with a solution, and this is causing issues for a sizable number of people, please feel free to revert this PR.

That said, we have been struggling with a separate (?) issue (libhermes.so not found) even _before_ this update that might be related to an obscure internal task ordering change that happened between the 3.4 and 3.5 series. There is still no clear solution until today, and I am not sure it's related.

PR #29044, opened 2020-06-03
bump Android Gradle Plugin to 4.0.0

dulmandakh:
## Summary

Bump Android Gradle Plugin to 4.0.1, part of Android Studio 4.0 release.

## Changelog

[Android] [Changed] - Bump Android Gradle Plugin to 4.0.1

## Test Plan

RNTester builds and runs as expected. And my app with RN 0.63 builds and runs as expected.


dulmandakh:
@cpojer @hramos @mdvacca  please merge 👍

friederbluemle:
The PR draft with an update to 4.0.0 is already open here: https://github.com/facebook/react-native/pull/29013

Can we merge #28170 with the update to `3.6.3` first please? It's been open since Feb 28, and has been rebase/updated multiple times. We should update in smaller incremental steps if possible.

mdvacca:
CC @fkgozali

fkgozali:
Sorry for the delay. I need to merge the Gradle upgrade to 6.6 today, then I'll merge the android tools upgrade (this one, #28170, #29013)

friederbluemle:
Hi @fkgozali, thanks for the reply!

Just double checking, can you please confirm the order:

1. Gradle Wrapper 6.6 #29613 (Let me know if you want me to update this to `6.6.1`, released yesterday)
2. Android Gradle plugin 3.6.4 #28170
3. I'll rebase #29013 for early testing with AGP 4+, and change it from draft to open PR
4. Additional testing period before upgrading to AGP 4+ (I'd suggest at least a few weeks)

We can also close this PR (#29044), since AGP 4+ will be done in #29013

fkgozali:
> (Let me know if you want me to update this to 6.6.1, released yesterday)

No, as previously discussed, let's not always take the latest without discussion.

Re: 3/4, why not just go straight to 4.0.1 instead of 4.0.0, then 4.0.1?

friederbluemle:
@mdvacca @fkgozali - Do you happen to have any information on the recent history destroying force pushes on the master branch? It happened (at least) twice already today. https://github.com/facebook/react-native/issues/29779

rickhanlonii:
The force pushes were necessary to fix the history, so the history was changed, not destroyed.

friederbluemle:
@rickhanlonii - Thanks for the reply! Maybe "destroyed" was not the right word, sorry about that. What I meant is that history was _changed_, i.e. the commits were overridden. Just curious: What kind of fix was necessary?

dulmandakh:
@fkgozali let's merge this PR, and go ahead

fkgozali:
@dulmandakh - but there's already https://github.com/facebook/react-native/pull/29013 right? Shouldn't we just merge that one instead?

dulmandakh:
I think we can settle on NDK version later on, and it might be better to move on with AGP.

fkgozali:
> I think we can settle on NDK version later on, and it might be better to move on with AGP.

Yeah but the other PR also upgrades to 4.0.1, which one should I merge for now? cc @friederbluemle

Issue #29047, opened 2020-06-04
Cannot modify the width of Switch component


HectorRicardo:
## Description

On Android, it is possible to modify the width of a Switch by using the `android:switchMinWidth` attribute

```xml
<Switch android:switchMinWidth="50dp" />
```

or by using the `.setSwitchMinWidth(int pixels)` method.

```java
Switch mySwitch = findViewById(R.id.mySwitch);
mySwitch.setSwitchMinWidth(400);
```

![image](https://user-images.githubusercontent.com/16261233/83707241-2a0ce680-a5df-11ea-9436-bf1f881259a3.png)

There is no such attribute in React-Native. Ideally, it would be nice to have a "width" attribute for the Switch component, such that we can do

```javascript
<Switch width={50} />
```
to achieve something similar as the above screenshot.

## React Native version:

System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 6.39 GB / 15.87 GB
  Binaries:
    Node: 12.16.3 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.14.4 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      Android NDK: 17.2.4988734
  IDEs:
    Android Studio: Version  3.6.0.0 AI-192.7142.36.36.6392135
  Languages:
    Python: 3.8.3 - C:\Python38\python.EXE
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0
    react-native: 0.62.1 => 0.62.1
  npmGlobalPackages:
    *react-native*: Not Found

HectorRicardo:
I think the following code changes will allow to control the Switch's width:

On ReactAndroid/src/main/java/com/facebook/react/views/switchview/ReactSwitchManager.java:
[https://www.diffchecker.com/1k7qqHI9](https://www.diffchecker.com/1k7qqHI9)

On ReactAndroid/src/main/java/com/facebook/react/views/switchview/ReactSwitch.java:
[https://www.diffchecker.com/WnCUHbRK](https://www.diffchecker.com/WnCUHbRK)

fabriziobertoglio1987:
I already found the solution and I m writing a pull request

But changing only the width is not enough 

it needs to change also the click area

safaiyeh:
[...]

Issue #28235, opened 2020-03-05
[Android] Height and Width is not fixed


cnkyrpsgl:
Description:

    In the official document it says "All dimensions in React Native are unitless, and represent density-independent pixels." [https://reactnative.dev/docs/height-and-width]

	React native gives buttons responsive size even the height is given fixed and style is not overwritten by anything. It says dimensions represents density independent pixels  but it behaves so. I expect my buttons to be same height on different devices independent of pixels and size.

React Native Info:

    System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
    Memory: 468.47 MB / 16.00 GB
    Shell: 5.0.9 - /usr/local/bin/bash
  Binaries:
    Node: 12.14.1 - /usr/local/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.12.0 => 16.12.0 
    react-native: ^0.61.4 => 0.61.4 
  npmGlobalPackages:
    react-native-cli: 2.0.1
    react-native: 0.61.5

## Expected Results

    I expect my buttons to be same height on different devices independent of pixels and [size.](url
![Screen Shot 2020-03-05 at 09 29 28](https://user-images.githubusercontent.com/31376095/75977596-ce529480-5eed-11ea-8212-a37a1c23c5be.png)
)

prasannabisen:
can I work on this?

fabriziobertoglio1987:
This issues needs to be closed

This is exactly what hi / low dpi should look like

Research about dpi online and then close this issue

It is not a big, it is a feature.. otherwise how can I make button bigger on 4k screen!

cnkyrpsgl:
@fabriziobertoglio1987 i would give percentage size like height: ‘30%’ if i want to make it look similar. Your argument doesn’t give a solution if i want to make a fixed size element.

fabriziobertoglio1987:
[...]

PR #29059, opened 2020-06-04
[Android] Adding prop android_minWidth to Switch component

fabriziobertoglio1987:
## Summary

This issue fixes https://github.com/facebook/react-native/issues/29047
Adding prop android_minWidth to Switch component.

## Changelog

[Android] [Fixed] - Adding prop android_minWidth to Switch component

## Test Plan

**<details><summary>CLICK TO OPEN TESTS RESULTS</summary>**
<p>

I will write separate pull request to update the docs on https://github.com/facebook/react-native-website

| **AFTER** |
|:-------------------------:|
|  |

</p>
</details>


ReactAndroid/src/main/java/com/facebook/react/views/switchview/ReactSwitchManager.java
```diff
@@ -175,6 +175,12 @@ public void setTrackTintColor(ReactSwitch view, @Nullable Integer color) {
     view.setTrackColor(color);
   }
 
+  @ReactProp(name = "androidMinWidth")
+  public void setSwitchMinWidth(ReactSwitch view, Integer pixels) {
+    Integer width = pixels == null ? view.getSwitchMinWidth() : pixels;
+    view.setSwitchMinWidth(width);
```
HectorRicardo, line 181: I just have a comment (or question). React-Native uses dip (density independent pixels) for the width and size of the components., as stated in the docs: [https://reactnative.dev/docs/height-and-width](https://reactnative.dev/docs/height-and-width).

Shouldn't this method expect a dip value instead of pixels ? The conversion from dip to pixels can be handled using this method:

```java
  private float dpToPixels(float dp) {
    return TypedValue.applyDimension(
      TypedValue.COMPLEX_UNIT_DIP,
      dp,
      view.getContext().getResources().getDisplayMetrics()
    );
  }
```

HectorRicardo:
Hey Fabrizio, thanks for the pull request!

I added a comment to one of the files, suggesting the use of dip (density independent pixels), instead of pixels.

ReactAndroid/src/main/java/com/facebook/react/views/switchview/ReactSwitchManager.java
fabriziobertoglio1987: @HectorRicardo 

I added https://github.com/facebook/react-native/pull/29059/commits/7c6901e72b511565bd2ce2c1b6d91de2ef2f1956 to use 

I quote the [dpi definition](https://en.wikipedia.org/wiki/Device-independent_pixel)

>A device-independent pixel (also: density-independent pixel, dip, dp) is a unit of length.

>For example, on the Android operating system a device-independent pixel is equivalent to one physical pixel on a 160 dpi screen,[1] while the Windows Presentation Foundation specifies one device-independent pixel as equivalent to 1/96th of an inch.[2]

>As **dp** is a physical unit it has an absolute value which can be measured in traditional units, e.g. for Android devices 1 dp equals 1/160 of inch or 0.15875 mm.

Increasing `density` of pixels does not change size

image

`setSwitchMinWidth` specifies the `MINUMUM WIDTH`. The actual width will be specified with the [`style` prop](https://reactnative.dev/docs/view-style-props). The `style` prop will not be applied if the `minWidth` is not respected.


| **Density 300 - switch width 50 minWidth 50** | **Density 300 - switch width 50 minWidth 100** |
|  |  |

Persisting Issues with the actual width of this component could be related to https://github.com/facebook/react-native/issues/28235
In that case I would write a separate pull request to fix that dpi issue and fix the following PixelUtil or the View Style prop.

https://github.com/facebook/react-native/blob/851d01b0aa45285bd6a52c89ab9c1b64a637258e/ReactAndroid/src/main/java/com/facebook/react/uimanager/PixelUtil.java#L13-L20

| **Density 200 - switch is 50px width** | **Density 200 - switch is 200px width** |
|  |  |

| **Density 300 - switch is 50px width** | **Density 300 - switch is 200px width** |
|  |  |

dulmandakh:
why not support minWidth in style?

ReactAndroid/src/main/java/com/facebook/react/views/switchview/ReactSwitchManager.java
HectorRicardo: Thanks for the explanation Fabrizio! It makes total sense what you're saying, that the style's width will not be applied if min width is not respected.

HectorRicardo:
@dulmandakh that could also work, but I guess it is because the minimum width is only supported on Android and not on iOS, so by naming the property `androidMinWidth` it is clear that it only applies to android

fabriziobertoglio1987:
Thanks for the support @HectorRicardo 

@dulmandakh 

Thanks and Sorry, but I do not work on iOS. I can not tell if this is possible or not.
If you request, I will move this logic to the style prop.

**TLDR**
**I believe It is possible** to build this logic on iOS. 
The iOS Contributor would have to set a check on the `width` style prop on `iOS` side and over-ride that value based on `minWidth` value.

I did a brief research, this are my **inconclusive** results on iOS:

----------

on iOS we are using [UISwitch](https://developer.apple.com/documentation/uikit/uiswitch)

https://github.com/facebook/react-native/blob/a50fa552a7e07f611d626fae99a6ef80845d9d4d/React/Fabric/Mounting/ComponentViews/Switch/RCTSwitchComponentView.mm#L22-L24

as explained [on stackoverflow](https://stackoverflow.com/a/25106983/11452286) 

`UISwitch` inherits from `UIView`. The width/height could be changed using [`UIView transform`][1] 

```c
UISwitch *aSwitch = [[UISwitch alloc] initWithFrame:CGRectMake(120, 120, 51, 31)];
aSwitch.transform = CGAffineTransformMakeScale(2.0, 2.0);
```

Seems difficult to change the width without using [`trasforms`][2] as said [here][3]

image

I personally would not use `transform` for the width/height prop. I would use `transform` for the [transforms][2]. I would not give the user a workaround to set `UISwitch` width/height on iOS using transforms, as this may not work as he expects. 

The iOS developer would have to find a way to set the `UISwitch` width/height and then change the logic to take in consideration the `minWidth` value.

I could not find existing iOS api to set the minWidth of a UISwitch, I did not make extensive research as I want to specialize on Android. 

Thanks a lot.
I wish you a good day

[1]: https://developer.apple.com/documentation/uikit/uiview/1622459-transform
[2]: https://reactnative.dev/docs/transforms
[3]: https://stackoverflow.com/a/25104771/11452286

HectorRicardo:
I also did some research, and although I don't have any Apple device to test, I am pretty sure the default UISwitch is not customizable. I suggest we leave the androidMinWidth attribute outside of the style object.

fabriziobertoglio1987:
>why not support minWidth in style?

Thanks a lot @dulmandakh for your code review. 

The main reason is that I am specializing on Android, Java and ReactNative.
Currently I don't work with ObjectiveC and iOS. 

It is harder for me to work with many different languages and tools.

Thanks a lot

Issue #29063, opened 2020-06-05
TextInput controlled selection broken on both ios and android.


Gutyn:
This issue is a continuation of the discussion:
https://github.com/facebook/react-native/commit/dff490d140010913d3209a2f3e987914b9c4eee4#commitcomment-39332764
The link to the sample project that demonstrates the issues:
https://github.com/Ginger-Labs/Input-bug

## Description
Controlled selection seems to be broken on both ios and android, to demonstrate the issues, I create a sample project (find the link above).

## React Native version:
System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 1.55 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.14.2 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.14.5 - ~/.npm-global/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 24, 25, 26, 27, 28, 29
      Build Tools: 26.0.3, 28.0.3, 29.0.0, 30.0.0
      System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom, android-29 | Google Play Intel x86 Atom_64
      Android NDK: 21.1.6352462
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6392135
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  npmPackages:
    react: ~16.9.0 => 16.9.0 
    react-native: ~0.62.2 => 0.62.2 
  npmGlobalPackages:
    create-react-native-app: 3.4.0
    react-native-app-id: 0.0.5
    react-native-cli: 2.0.1

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.
The reproduction steps are in the sample project's ReadMe file.
For simplicity purposes, I will post them here as well:

SIM - iPhone 11 (13.4.1):
1) Click on the "Click Me" Button, set the cursor in the middle, add some text, click on the button again: Notice the text is set to needed one but the selection is not 10
[...]

Issue #27449, opened 2019-12-08
Using `toUpperCase` or `textTransform: 'uppercase'` breaks on an Android controlled TextInput


nick-michael:
React Native version: 0.61.4 and lower

Trying to force capitalization of characters inside a TextInput is broken on Android.

- `autoCapitalize="characters"` doesn't seem to do anything
- Using `toUpperCase` in the onChangeText hook causes duplication of letters
- Using `textTransform: 'uppercase'` in styles block causes the same duplication of letters
- Using `textTransform: 'uppercase'` in a non-controlled TextInput does nothing

## Steps To Reproduce

1. Create a controlled TextInput and either use onTextChanged to modify the text to uppercase or use text transform in the styles block
2. Type multiple lowercase characters into the text box

Describe what you expected to happen:
Characters should be capitalized

What actually happens:
Characters are capitalized and duplicated

Snack, code example, screenshot, or link to a repository:
https://snack.expo.io/@nmi09/rn-android-capitalize-input-bug

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/9058133/70392933-c2356480-19dc-11ea-803e-7c577f6807b1.gif)

inferusvv:
Yeah. The same issue! On iOS all works fine

mgiorgi-github:
Same issue for me....

laraferroni:
Had same issue. Adding 
```secureTextEntry={Platform.OS === 'ios' ? false : true}```
seemed to fix it for me.

YeshanJay:
Any luck finding a workaround for this??

simon-davies-avtura:
`keyboardType="visible-password"` 

Seems to be the only workaround for me with Android at the moment (I've been looking for ages!). Will need some cross platform testing locally though.

emilioheinz:
Really weird, it worked but I think that should be fixed asap.

> `keyboardType={Platform.OS === 'ios' ? 'default' : 'visible-password'}`
> 
> Seems to be the only workaround for me with Android at the moment (I've been looking for ages!). Will need some cross platform testing locally though.

arthurolga:
i have the same problem. The solution by simon-davies-avtura solved it partially, but I really wanted to use `keyboardType='email-address'`. If there is any other workaround or a solution. I just want to display the font in lowerCase on the UI.

fabriziobertoglio1987:
seems similar issue to https://github.com/facebook/react-native/issues/11068#issuecomment-574013816 

[...]

PR #29070, opened 2020-06-05
[Android] Fix letters duplication when using autoCapitalize

fabOnReact:
## Summary
This issue fixes [#11068][2] fixes [#30393][3] fixes [#27449][4] fixes [#29063][5] fixes [#23578][6] fixes [#30503][7] fixes [#25265][8].

**PROBLEM 1 - TextInput autoCapitalize creates duplicated letters**
The issue is caused by [textinput/ReactEditText.java#L556][1] which replaces the text from a starting position 0 to end position length(), but causes letters duplication ([more info][20]). The solution proposed in this pr to the above issue is:
1. store the previous cursor position in variable `start`/`endPosition`
2. use `setText()` to change the TextInput value
3. restore the cursor to the previous `start`/`endPosition`

**PROBLEM 2 - setSpan (6 ... 6) ends beyond length 3**
Reported from Joshua in this [comment][9] and reported with issue [#29063][5]
```
Exception in native call java.lang.IndexOutOfBoundsException: 
setSpan (6 ... 6) ends beyond length 3
```
The fix for this issue was merged in master with commit [EditText: maintain cursor position when text changes][30]. Relevant discussions are [#29070-comment][23] and my commits [check for reactTextUpdate.length][31], [revert changes to maybeSetSelection ][32], [code review][24]).
Related:
1. discussing difference between `setText()` and `replace()` in the `ReactEditText.java` API [issues/30393#issuecomment-752676908][20] 
2. discussing how JS TextInput keeps internal states and updates it in the `ReactEditText.java` API [issues/29063#issuecomment-658189938][21]
3. [TextInput becomes slow after lots of typing][26]
4. [breaking down why TextInput becomes slow after lots of typing][27]
5. [AOSP method TextView setText][28]
6. [AOSP method TextView getText][29]
7. [AndroidTextInputShadowNode method updateStateIfNeeded][40]
8. [Editable method replace()][41]

## Changelog
[Android] [Fixed] - Fix letters duplication when using autoCapitalize

## Test Plan
This changes fix the Java API. The Java tests are disabled for this component as explained in commit [Disabled unit tests that need yoga libs when running ][33]. Detailed tests included in [pull/29070#issuecomment-1065799687][22].

## Notes
Pull request [#27757][25] from @safaiyeh fixed letters duplication, but introduced a regression (changing the cursor position). This pull request fixes letters duplication without changing the cursor position and avoid Runtimes caused by moving the cursor to invalid position.

[1]: https://github.com/facebook/react-native/blob/7b09eb54e7ecdde91cb4748dc26207dd05088382/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L556 "ReactEditText.java#L556"
[2]: https://github.com/facebook/react-native/issues/11068
[3]: https://github.com/facebook/react-native/issues/30393
[4]: https://github.com/facebook/react-native/issues/27449
[5]: https://github.com/facebook/react-native/issues/29063
[6]: https://github.com/facebook/react-native/issues/23578
[7]: https://github.com/facebook/react-native/issues/30503
[8]: https://github.com/facebook/react-native/issues/25265

[9]: https://github.com/facebook/react-native/pull/29070#issuecomment-647716585 "comment about error setSpan(5 ... 5) ends beyond length 4"

[20]: https://github.com/facebook/react-native/issues/30393#issuecomment-752676908 "comment explaining issues caused by .replace() java method instead of setText()" 
[21]: https://github.com/facebook/react-native/issues/29063#issuecomment-658189938 "summary of how setText is called from JS to JAVA api"
[22]: https://github.com/facebook/react-native/pull/29070#issuecomment-1065799687 "Test Plan"
[23]: https://github.com/facebook/react-native/pull/29070#issuecomment-683436866 "Breaking down how TextInput.js and ReactEditText.java Components update text value"
[24]: https://github.com/facebook/react-native/pull/29070#discussion_r681530300 "Core Review"
[25]: https://github.com/facebook/react-native/pull/27757
[26]: https://github.com/facebook/react-native/issues/20119 "TextInput becomes slow after lots of typing"
[27]: https://github.com/facebook/react-native/issues/20119#issuecomment-639482512 "breaking down why TextInput becomes slow after lots of typing"
[28]: https://github.com/aosp-mirror/platform_frameworks_base/blob/1ac46f932ef88a8f96d652580d8105e361ffc842/core/java/android/widget/TextView.java#L6081-L6109 "AOSP method TextView setText"
[29]: https://github.com/aosp-mirror/platform_frameworks_base/blob/1ac46f932ef88a8f96d652580d8105e361ffc842/core/java/android/widget/TextView.java#L2336-L2338 "AOSP method TextView getText"

[30]: https://github.com/facebook/react-native/commit/de44184e01d74e4da18f1cba9b01d4c259ef5d9c "commit from ShikaSD EditText: maintain cursor position when text changes"
[31]: https://github.com/facebook/react-native/pull/29070/commits/04ce594557399d149f286d88ee3008773cf9f978 "check for reactTextUpdate.length"
[32]: https://github.com/facebook/react-native/pull/29070/commits/7ecdea7b885bbb0555ae85913f8ec4fae2fd21c0 "revert changes to maybeSetSelection"
[33]: https://github.com/facebook/react-native/commit/709a441ecf54cd9465f5946af0454ee7d10d5cbe "commit that disables java tests for issues with Yoga"

[40]: https://github.com/fabriziobertoglio1987/react-native-notes/blob/1e54f76fb1fcc4676edccfa11b61a5f245c94558/ReactCommon/react/renderer/components/textinput/androidtextinput/react/renderer/components/androidtextinput/AndroidTextInputShadowNode.cpp#L119 "AndroidTextInputShadowNode method updateStateIfNeeded"
[41]: https://github.com/aosp-mirror/platform_frameworks_base/blob/1ac46f932ef88a8f96d652580d8105e361ffc842/core/java/android/text/Editable.java#L28-L52 "Editable method replace"


safaiyeh:
Yooo! Nice glad you found the solution

cc @JoshuaGross

fabriziobertoglio1987:
@safaiyeh Thanks a lot. I wish you a good weekend :smiley:

JoshuaGross:
I'll import and test on the FB codebase this week.

fabriziobertoglio1987:
@JoshuaGross thanks a lot. I would like to help troubleshoot that warning that you are experiencing on Phabricator, but I can not see the warning message. I remain available to help and do improvements. I wish you a good week. Thanks. Fabrizio

JoshuaGross:
@fabriziobertoglio1987 I tested this in FB on a few screens and I'm running into this with some controlled TextInputs:

`setSpan(5 ... 5) ends beyond length 4`

So it looks like there are some out-of-bounds issues, in particular with controlled TextInputs

fabriziobertoglio1987:
Thanks @JoshuaGross [setSpan](https://github.com/facebook/react-native/search?q=setSpan&unscoped_q=setSpan) is called in several places.

Could you check the android log with 

```
adb logcat | grep -P "your query"
```

you can replace your query with any search string parameter you want. We would need to log the full stack from that exception and understanding exactly which line of code if failing.

We can filter the Logs with the `adb logcat` cli
https://developer.android.com/studio/command-line/logcat

Do you think this could be caused from `manageSpans`?

https://github.com/facebook/react-native/blob/9263eb5d3864a42925b699343db2c09cc8934ed0/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L487-L510

The `manageSpan` method

https://github.com/facebook/react-native/blob/9263eb5d3864a42925b699343db2c09cc8934ed0/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L546-L548

The `spannableStringBuilder` is created from 

https://github.com/facebook/react-native/blob/9263eb5d3864a42925b699343db2c09cc8934ed0/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L492

https://github.com/facebook/react-native/blob/9312313c3c6701490d728f912e0b0bbd16d91ad9/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L232-L249

Seems to me that we update the `text` properties with JavaScript and sync them on the Android Native Application, but maybe there are flaws and some variables are not be updated at the correct time.

https://github.com/facebook/react-native/blob/f8bcb1063e3b7ff363203f6b1a19f7da512eaf20/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextUpdate.java#L21-L32

Thanks a lot
I wish you a good day
Fabrizio

fabriziobertoglio1987:
I'm trying to fix the following issue

This test will replace spaces with empty string

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/RNTester/js/examples/TextInput/TextInputSharedExamples.js#L135

`useEffect` hook should trigger `setTextAndSelection`

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/Libraries/Components/TextInput/TextInput.js#L895

`useEffect` will monitor for `value` change and run, but it will not run when we replace spaces with empty string.

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/Libraries/Components/TextInput/TextInput.js#L911

Javascripts calls Java `receiveCommand` with `commandId = SET_TEXT_AND_SELECTION`

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L213

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L245

https://github.com/facebook/react-native/blob/ab56ed7d96863d7b95521ce9a1e761c215ba0f98/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L473

The cursor is in the wrong position [because `setSelection` is not called when text changes](https://github.com/fabriziobertoglio1987/react-native/blob/a66718881611eb687035ff458f7f13befd26f0a3/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L511)

safaiyeh:
Rooting for ya! This is a tough bug haha.

fabriziobertoglio1987:
Thanks a lot @safaiyeh :pray: for the support

I think I know now what is the real issue.

We have 2 sides of this functionality
1) The controlled component which keeps his own state for the TextInput value and ignores the Value sent from Java
2) The Java functionality which keeps its own internal TextInput State and the value of the text

The solution introduced with commits https://github.com/facebook/react-native/pull/29070/commits/677a78923edaf589fe07f4a5b618d26fc40734b2 and https://github.com/facebook/react-native/pull/29070/commits/a66718881611eb687035ff458f7f13befd26f0a3 uses `setText` and `setPosition`, but this causes issues when we keep a different JavaScript TextInput internal state `setSpan(5 ... 5) ends beyond length 4` as experienced in https://github.com/facebook/react-native/pull/29070#issuecomment-647716585

`end` is higher then the length of the JavaScript version of the text.
```java
setSelection(start, end);
```

https://github.com/fabriziobertoglio1987/react-native/blob/a66718881611eb687035ff458f7f13befd26f0a3/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L302

https://github.com/fabriziobertoglio1987/react-native/blob/a66718881611eb687035ff458f7f13befd26f0a3/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L511

my main issue now is that Java and JavaScript do not use the same state text (for ex the input "my w o r d has spaces" which in that example in JS internal state is change to "mywordhasspaces").

2 Solutions:

1) Fix it in Javascript using `TextInput` props `selection` and `onChangeSelection` to control the selection in JavaScript instead of Java, and in Java avoid setSelection to position higher then the Java version of the word, as show in commit https://github.com/facebook/react-native/pull/29070/commits/bab02b5cae4f7239dfc02df52b46003cebbfba30

2) Fix it in Java inside `ReactEditText` methods `maybeSetText` using the JavaScript internal state passed via `reactTextUpdate`, so I added some check to avoid to set the wrong cursor position and trigger `setSpan(5 ... 5) ends beyond length 4`, commit https://github.com/facebook/react-native/pull/29070/commits/04ce594557399d149f286d88ee3008773cf9f978

https://github.com/fabriziobertoglio1987/react-native/blob/a66718881611eb687035ff458f7f13befd26f0a3/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L473

I'm saving the length of the updated JavaScript text in Java as variable `mMaximumTextLength` and I check this variable before running operations on the text. This can be a very quick solution to fix this problem

It still a draft, but I believe I can complete it next Weekend.

Thanks a lot
I wish you good beginning of the Work Week :smiley: :pray: 
Fabrizio Bertoglio

fabriziobertoglio1987:
Related issue I should check once completed this pull request https://github.com/facebook/react-native/issues/29063

fabriziobertoglio1987 (commented):
I tested this changes today and I'm thinking of publishing this solution, as changing the TextInput.js source may impact iOS functioanalities. This solution may change the current behaviour of some functionalities (instead of replacing the Text value from position 0 to position n we use setText and then restore the cursor to the previous position) and it is not clear for me the impact of this changes, additionally I updated the TextInputExamples to achieve this behaviour, so similar functionalities/applications will need to build JS logic to achieve the same behaviour. 

A TextInput JavaScript Controlled Component which manages the TextInput text in its internal state like in example below (*Android Controlled Component replace Spaces*) and deletes spaces, will have to control the cursor position (selection) in [its internal state](https://github.com/facebook/react-native/pull/29070/files?file-filters%5B%5D=.js#diff-51e729971c02a077b6aa6004042b7b61R126-R134)

An option would be including this logic in our TextInput.js useEffect callbacks but this may cause further issues on iOS and Android. I will prioritize other prs in the future (for ex. completing https://github.com/facebook/react-native/pull/29059)

I detected non perfect behaviour with the example on the right side (Android Controlled Component replace spaces), when clicking spaces too fast the cursor would position to the beginning of the line, but I believe this issue is caused by the Example JS logic.

Really happy to receive feedback or tips. Thanks a lot!

| **Android Controlled Comp replace Chars with Capitalized letters** | **Android Controlled Component replace Spaces** | 
| |  |

packages/rn-tester/js/examples/TextInput/TextInputSharedExamples.js
```diff
@@ -122,18 +120,40 @@ class RewriteExampleInvalidCharacters extends React.Component<
...
+
+  onSelectionChangeHandler = ({nativeEvent: {selection}}) => {
+    const {start, end} = selection;
+    const {text} = this.state;
+    const maxEnd = text.length;
+    const validEnd = end > maxEnd ? maxEnd : end;
+    const validStart = start > maxEnd ? maxEnd : start;
+    const newSelection = {start: validStart, end: validEnd};
+    this.setState({selection: newSelection});
+  };
```
fabriziobertoglio1987, line 134: as displayed in the below GIF, [this solution][1] is not acceptable as it would create a different behaviour between iOS and Android. Android would need to set the selection position (the cursor) in this type of use case scenario, while the same solution on iOS would present different behaviour adding further complexity to the framework. I believe that I will need to add logic to TextInput useEffect/useLayout callbacks to handle this different scenarios, without regressions or difference in behaviour between iOS and Android.

The below gifs displays the issue experienced on iOS. Thanks a lot!



[1]: https://github.com/facebook/react-native/pull/29070/files#diff-51e729971c02a077b6aa6004042b7b61R126-R134

packages/rn-tester/js/examples/TextInput/TextInputSharedExamples.js
```diff
@@ -122,18 +120,40 @@ class RewriteExampleInvalidCharacters extends React.Component<
...
     return (
       <View style={styles.rewriteContainer}>
         <TextInput
           testID="rewrite_no_sp_input"
           autoCorrect={false}
-          multiline={false}
+          multiline={true}
           onChangeText={text => {
-            this.setState({text: text.replace(/\s/g, '')});
+            const newText = text.replace(/\s/g, '');
```
fabriziobertoglio1987, line 148: as this component keeps controller version of the text in JavaScript, the text value `an example` would differ in javascript after the above substitution and become `anexample`, the onSelectionOnChangeHandler will control the cursor position and avoid setting it to a position higher then the new word length.

This behaviour is different between iOS and Android and causes some issues. I will try to remove the `onSelectionOnChangeHandler` that controls the cursor position.

emilioheinz (approved):

vonovak:
ah, good job! Textinput is super tricky in my experience (different android versions and vendors are pain). Let's hope this does not break some other stuff 🤞

fabriziobertoglio1987:
Thanks @vonovak :pray: :peace_symbol: 
>Let's hope this does not break some other stuff 

I believe we should have complete Java tests running on ReactAndroid, verifying there are no regressions and helping developers to write this type of PRs, **but**

>This changes fix the Java API for which I can not write Java tests as explained in commit 709a441
>The java TextInputTest was excluded from the test suite in commit 709a441 as they need the Yoga libraries to run

seems that I will have to do **Yoga** in the future so I can fix the above mentioned issue and somehow **re-enable** the test suite for this and many other components.

This PR fixes one of the top issues found by the community, overall the related issues received a total of 120 thumbs ups https://github.com/facebook/react-native/issues/11068 https://github.com/facebook/react-native/issues/30393 https://github.com/facebook/react-native/issues/27449 https://github.com/facebook/react-native/issues/29063 https://github.com/facebook/react-native/issues/23578 https://github.com/facebook/react-native/issues/30503 https://github.com/facebook/react-native/issues/25265

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
```diff
@@ -111,6 +111,7 @@
   private int mFontStyle = ReactTypefaceUtils.UNSET;
   private boolean mAutoFocus = false;
   private boolean mDidAttachToWindow = false;
+  private Integer mMaximumTextLength = 0;
```
JoshuaGross, line 114: Does this need to be Integer? The wrapped object Integer type has more overhead because the actual operations (less-than, etc) will first convert this to an `int`. Let's just use `int` here instead and below as well if possible?

fabriziobertoglio1987: Thanks a lot. I'm reading changes from https://github.com/facebook/react-native/commit/de44184e01d74e4da18f1cba9b01d4c259ef5d9c and re-evaluating this solution.

cristianoccazinsp:
Please merge and release asap!

fabriziobertoglio1987:
It is interesting seeing how Flutter implements this functionality. It may also help me to understand ReactNative internals.
```dart
  /// Setting this will notify all the listeners of this [TextEditingController]
  /// that they need to update (it calls [notifyListeners]). For this reason,
  /// this value should only be set between frames, e.g. in response to user
  /// actions, not during the build, layout, or paint phases.
  ///
  /// This property can be set from a listener added to this
  /// [TextEditingController]; however, one should not also set [selection]
  /// in a separate statement. To change both the [text] and the [selection]
  /// change the controller's [value].
  set text(String newText) {
    value = value.copyWith(
      text: newText,
      selection: const TextSelection.collapsed(offset: -1),
      composing: TextRange.empty,
    );
  }
```

https://github.com/flutter/flutter/blob/60d5ba25b729a994540a2b378240edb575afc6de/packages/flutter/lib/src/widgets/editable_text.dart#L161-L176
Reading their doc

jarrisoncano:
what happens with this pr? actually i have the same issue

SrBrahma:
Annoying bug that should be fixed asap

AnOptimistSeePessimist:
There seems to be a small bug that prevents you from long pressing the delete button on the virtual keyboard to delete a lot of content

fabriziobertoglio1987:
@AnOptimistSeePessimist thanks for the report. I look into it as soon as possible. :pray:

fabriziobertoglio1987:
could be connected to this logic

https://github.com/facebook/react-native/blob/19b571c4fbc19674dfa515a517bd634246369eaf/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L558-L562

`reactTextUpdate.getText().length()` is not anymore `0` with the new logic when cleared.

https://github.com/facebook/react-native/blob/19b571c4fbc19674dfa515a517bd634246369eaf/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L562-L570

I'll have to update the condition to catch this scenario or `setText()` to `null` when cleared with long press
I'll look into it soon. Thanks

fabriziobertoglio1987:
@AnOptimistSeePessimist did you build from source this pull request and did this test on my branch?

AnOptimistSeePessimist:
minimal code example with node_modules directory: https://github.com/AnOptimistSeePessimist/AwesomeProject 

step by step:

1. git clone https://github.com/facebook/react-native 
2. cd react-native
3. git checkout v0.64.3
4. **now make changes to the code，Not testing on your branch**
// build android artifacts
5. docker run --rm --name rn-build -v $PWD:/pwd -w /pwd reactnativecommunity/react-native-android /bin/sh -c "./gradlew installArchives --no-daemon"
6. npx react-native init AwesomeProject --version 0.64.3
7. paste react-native v0.64.3 to AwesomeProject node_modules

fabriziobertoglio1987:
@AnOptimistSeePessimist Thanks. Yes, I understand the issue now and I'm able to reproduce it now. Thanks for the report. I'll try to find a solution to the problem.

fabriziobertoglio1987:
## Test Plan

This changes fix the Java API for which I can not write Java tests as explained in commit https://github.com/facebook/react-native/commit/709a441ecf54cd9465f5946af0454ee7d10d5cbe
The java TextInputTest was excluded from the test suite in commit https://github.com/facebook/react-native/commit/709a441ecf54cd9465f5946af0454ee7d10d5cbe as they need the Yoga libraries to run

**<details><summary>CLICK TO OPEN TESTS RESULTS - ANDROID</summary>**
<p>

| **BEFORE (LETTER DUPLICATION)** | **PR [#27757](https://github.com/facebook/react-native/pull/27757#discussion_r434400963)** (the cursor would move to the wrong position) | 
|  |  |

| **Replacing spaces** | **change to Capitalized letters** | 
| |  |

| **AFTER** | 
|:-------------------------:|
 |

</p>
</details>

**<details><summary>CLICK TO OPEN TESTS RESULTS - IOS</summary>**
<p>

| **Live ReWrite** | **Live ReWrite No Spaces** | 
| |  |

| **Live ReWrite** |
|:-------------------------:|
| |

</p>
</details>

ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
fabriziobertoglio1987: Sorry @JoshuaGross I added all the changes requested with https://github.com/facebook/react-native/pull/29070/commits/b86f4985ca25320301fd8196505ec5ca52e017d0 and I updated the PR summary with all the relevant information :pray: Thanks a lot

fabriziobertoglio1987:
- [ ] use [`setTextKeepState`](https://github.com/facebook/react-native/issues/11068#issuecomment-1073453014)

fabriziobertoglio1987:
<details><summary>using setText instead of replace changes TextInput height</summary>
<p>

**Expected Result**:
The TextInput does not change size while typing on fabric

**Actual Result**
The TextInput **changes** size

https://github.com/facebook/react-native/blob/3b7581cd763635aad605cc175bc3cb74182f4a85/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java#L544-L547

<details><summary>click to open code used in the test</summary>
<p>

```javascript
class RewriteExample extends React.Component<$FlowFixMeProps, any> {
  constructor(props) {
    super(props);
    this.state = {text: ''};
  }
  render() {
    const limit = 20;
    const remainder = limit - this.state.text.length;
    const remainderColor = remainder > 5 ? 'blue' : 'red';
    return (
      <View style={styles.rewriteContainer}>
        <TextInput
          testID="rewrite_sp_underscore_input"
          autoCorrect={false}
          multiline={false}
          maxLength={limit}
          onChangeText={text => {
            text = text.replace(/ /g, '_').toUpperCase();
            this.setState({text});
          }}
          style={styles.default}
          value={this.state.text}
        />
        <Text style={[styles.remainder, {color: remainderColor}]}>
          {remainder}
        </Text>
      </View>
    );
  }
}
```
</p>
</details>

<details><summary>click to open recording of test case</summary>
<p>

<video src="https://user-images.githubusercontent.com/24992535/159212653-c8fa02d9-4baf-482b-88ac-5b2d61350631.mp4" width="1000" />

</p>
</details>

</p>
</details>

MaskedHawk:
Any update about the Uppercase TextInput ? 
Just test on 0.69 and still got the issue.

bombillazo:
Dang, no update on this issue? 😞

fabriziobertoglio1987:
If I have some free time, I will work on this PR and complete it. Thanks a lot for the interest

t-fritsch:
hi @fabriziobertoglio1987 

thanks for the efforts you put into this issue. 

I wonder why you closed this PR ? This bug is so annoying.. I teach React Native since a long time now and having such "stupid" bug isn't really giving confidence to newcomers. This may not have been a problem in the past, but now RN has rising competitors that give the impression to be much more stable.

fabriziobertoglio1987:
Thanks @t-fritsch.

My knowledge of the react-native ReactEditText API improved a lot in the last 3 years. 
I wanted to publish an improved solution but never had the time.

I can [prioritize](https://github.com/sponsors/fabriziobertoglio1987) this task if you want.

fabOnReact:
This PR is included in the [react-native-improved](https://github.com/fabriziobertoglio1987/react-native-improved) library:

# react-native-improved

- **Supports ONLY react-native 0.73**.
- Supports only old architechture (new architechture is WIP)

## Set-up

In package.json

```diff
 "scripts": {
+  "postinstall": "yarn run react-native-patch"
 }
```

Then

## npm

```sh
npm install react-native-improved --save-dev
```

## yarn v1

```sh
yarn add react-native-improved --dev
```

PR #29106, opened 2020-06-10
Addressing various issues with the Appearance API (#28823)

### LINKED ISSUE #28823: Appearance module does not behave correctly: addChangeListener does not react to changes, useColorScheme() does not react to changes.
## Description
The subscription of `useColorScheme()` and `Appearance.addChangeListener()` wont work. The hook returns an initial value only on component mount. Changes to the  devices color scheme on Android does not provide the hook with a new value. I therefore tried to manually replicate this hook with `useEffect` and the `Appearance` module and found that `Appearance.addChangeListener()` when called returns "undefined", not type `EventSubscription`. The listener function provided to `Appearance.addChangeListener()` never gets called. 

I have only tested this on Android so I am not aware if this is also an issue on iOS.

## React Native version:
```
System:
    OS: Windows 10 10.0.17763
    CPU: (16) x64 Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz
    Memory: 15.05 GB / 31.92 GB
  Binaries:
    Node: 10.16.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.19.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.4 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2, 29.0.3
      System Images: android-28 | Intel x86 Atom_64, android-R | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: Version  3.6.0.0 AI-192.7142.36.36.6308749
  Languages:
    Java: 1.8.0_211 - C:\Program Files\Java\jdk1.8.0_211\bin\javac.EXE
    Python: 2.7.18 - C:\Python27\python.EXE
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0
    react-native: 0.62.2 => 0.62.2
  npmGlobalPackages:
    *react-native*: Not Found

```
## Steps To Reproduce

1. run `npx react-native init TestDark --template react-native-template-typescript`
2. Import `Appearance` & `useColorScheme` from `'react-native'`
3. copy paste the code snippet below above the return in `<App />` or use the `useColorScheme` hook and console.log its output.
4. Change the color scheme of the Android device and check the console for the correct values.

## Expected Results
The `useColorScheme` hook should return the correct colorScheme when the devices colorScheme changes.
The `Appearance.addChangeListener()` should return type `EventSubscription` not `undefined`.
The listener callback provided to Appearance.addChangeListener(cb) should be called.

```jsx
const App = () => {
  const colorScheme = useColorScheme(); // Only gives initial colorScheme, does not return new colorScheme on change.
  const [mode1, setMode1] = useState('light');
  const mode = colorScheme || 'light';

  React.useEffect(() => {
    const initialColorScheme = Appearance.getColorScheme(); // Works as expected
    setMode1(initialColorScheme || 'light');

    const listener = () => {
      console.log('called'); // Never fires
      const colorScheme1 = Appearance.getColorScheme();
      setMode1(colorScheme1 || 'light');
    };

    const subscription = Appearance.addChangeListener(listener);
    console.log('App -> subscription', subscription); // logs as undefined
    return () => {
      Appearance.removeChangeListener(listener);
    };
  }, []);
  return (
   // jsx
  )
}
```
Here is a GIF of the app created with `npx react-native init NAME --template react-native-template-typescript` command. Look for Mode: dark (useColorScheme() and Mode_1 (useEffect & Appearance module). https://gyazo.com/d4ddb35bf689b3c8926be2f613714f73

mrbrentkelly:
## Summary

This PR fixes a few issues with the Appearance API (as noted here https://github.com/facebook/react-native/issues/28823).

1. For the Appearance API to work correctly on Android you need to call `AppearanceModule.onConfigurationChanged` when the current Activity goes through a configuration change. This was being called in the RNTester app but not in `ReactActivity` so it meant the Appearance API wouldn't work for Android in newly generated RN projects (or ones upgraded to the latest version of RN).

2. The Appearance API wasn't working correctly for brownfield scenarios on Android. It's possible to force an app light or dark natively on Android by calling `AppCompatDelegate.setDefaultNightMode()`. The Appearance API wasn't picking up changes from this function because it was using the Application context instead of the current Activity context.

3. The Appearance API wasn't working correctly for brownfield scenarios on iOS. Just like on Android its possible to force an app light or dark natively by setting `window.overrideUserInterfaceStyle`. The Appearance API didn't work with this override because we were overwriting `_currentColorScheme` back to default as soon as we set it.

### Fixed

#28823

* [Android] [Fixed] - Appearance API now works on Android
* [Android] [Fixed] - Appearance API now works correctly when calling `AppCompatDelegate.setDefaultNightMode()`
* [iOS] [Fixed] - Appearance API now works correctly when setting `window.overrideUserInterfaceStyle`

## Test Plan

Ran RNTester on iOS and Android and verified the Appearance examples still worked [correctly.](url)


React/CoreModules/RCTAppearance.mm
```diff
@@ -89,7 +89,9 @@ - (dispatch_queue_t)methodQueue
 
 RCT_EXPORT_SYNCHRONOUS_TYPED_METHOD(NSString *, getColorScheme)
 {
-  _currentColorScheme = RCTColorSchemePreference(nil);
+  if (_currentColorScheme == nil) {
+    _currentColorScheme = RCTColorSchemePreference(nil);
+  }
```
koenpunt, line 94: While this change probably makes sense, it doesn't solve the issues with `overrideUserInterfaceStyle` in a brownfield app for me.

mrbrentkelly: Hey @koenpunt, if you have time to throw up a repo with a brownfield project that's broken (or hit me with some repro steps) I could probably take a look at brownfield a bit deeper sometime this week 👍 .

mrbrentkelly:
FYI the `test_js` and `test_js_prev_lts` checks are failing due to a Flow version error which doesn’t appear related to the changes in this PR. 

``
Wrong version of Flow. The config specifies version ^0.131.0 but this is version 0.132.0
Flow check failed.
``

alexandersandberg:
What's the current status of this PR? On the roadmap? 🙂

mrbrentkelly:
I've been wondering the same thing :)

andresribeiro:
So.... anything new?

svbutko:
Is this issue even on roadmap for next releases?

PR #29120, opened 2020-06-11
Make platformColors test page render with out of tree platforms

acoates-ms:
## Summary

PlatformColorExample RNTester page crashes when running on platforms other than iOS/Android.
Changed some very minor logic to take into account that platforms other than iOS/Android exist.

## Changelog

[Internal] [Fixed] - PlatformColorExample RNTester page crashes when running on platforms other than iOS/Android.

### DIFF
--- a/RNTester/js/examples/PlatformColor/PlatformColorExample.js
+++ b/RNTester/js/examples/PlatformColor/PlatformColorExample.js
@@ -199,7 +199,10 @@ function FallbackColorsExample() {
       color: PlatformColor('bogus', '@color/catalyst_redbox_background'),
     };
   } else {
-    throw 'Unexpected Platform.OS: ' + Platform.OS;
+    color = {
+      label: 'Unexpected Platform.OS: ' + Platform.OS,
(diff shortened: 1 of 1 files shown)


RNTester/js/examples/PlatformColor/PlatformColorExample.js
```diff
@@ -261,17 +264,21 @@ function VariantColorsExample() {
...
+          })}
         </Text>
         <View
           style={{
             ...styles.colorCell,
             backgroundColor:
               Platform.OS === 'ios'
                 ? DynamicColorIOS({light: 'red', dark: 'blue'})
-                : PlatformColor('?attr/colorAccent'),
+                : Platform.OS === 'android'
```
sahrens, line 279: Why not a `Platform.select` here too?

acoates-ms: In cases where the Platform.select babel transform isn't running, it will attempt to run an undefined function.

PR #29131, opened 2020-06-12
Added AccessibilityInfo.setAccessibilityFocus to RNTester a11y examples

graytmatterMS:
## Summary

RNTester has Accessibility page which includes examples of many of the AccessibilityInfo methods being used. Until now there was no test for AccessibilityInfo.setAccessibilityFocus. This PR adds a button that–on click–shifts focus to a Text component.

## Changelog

[General] [Added] - Added AccessibilityInfo.SetAccessibilityFocus to RNTester

## Test Plan

1. Run RNTester
2. Go to the accessibility page
3. Scroll to Set Accessibility Focus (the last entry)
4. Toggle VoiceOver in settings.
5. Move VoiceOver Focus to "Set Accessibility Focus"
6. Hit [Space]
- You should see the focus being moved to "Move focus here on button press".
Screen Shot 2020-06-12 at 10 25 40
Screen Shot 2020-06-12 at 10 26 12

### DIFF
--- a/RNTester/js/examples/Accessibility/AccessibilityExample.js
+++ b/RNTester/js/examples/Accessibility/AccessibilityExample.js
@@ -20,6 +20,7 @@ const {
   TouchableWithoutFeedback,
   Alert,
   StyleSheet,
+  findNodeHandle,
 } = require('react-native');
 
 const RNTesterBlock = require('../../components/RNTesterBlock');
--- a/RNTester/js/examples/Accessibility/AccessibilityExample.js
+++ b/RNTester/js/examples/Accessibility/AccessibilityExample.js
@@ -751,24 +751,22 @@ class AnnounceForAccessibility extends React.Component<{}> {
   }
 }
 
-class SetAccessibilityFocus extends React.Component<{}> {
-  _handleOnPress = () => {
-    if (findNodeHandle(this.focusRef.current)) {
-      const reactTag = findNodeHandle(this.focusRef.current);
-      AccessibilityInfo.setAccessibilityFocus(reactTag);
-    }
-  };
-  render() {
-    this.focusRef = React.createRef();
-    return (
-      <View>
-        <Button onPress={this._handleOnPress} title="Set Accessibility Focus" />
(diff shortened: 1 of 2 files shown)


tido64:
@hramos: Who should we ping for review of accessibility/RNTester stuff?

RNTester/js/examples/Accessibility/AccessibilityExample.js
```diff
@@ -750,6 +751,26 @@ class AnnounceForAccessibility extends React.Component<{}> {
   }
 }
 
+class SetAccessibilityFocus extends React.Component<{}> {
```
TheSavior, line 754: Ideally you'd write this with hooks, as we are trying to have new examples use hooks from now on. It's reasonable that you wrote it with classes though since everything else in this file is using them. If that was the only change then I'd probably just pull it in as-is, but since I have another comment, can you update this to use hooks? :)

RNTester/js/examples/Accessibility/AccessibilityExample.js
```diff
@@ -750,6 +751,26 @@ class AnnounceForAccessibility extends React.Component<{}> {
...
+  _handleOnPress = () => {
+    if (findNodeHandle(this.focusRef.current)) {
+      const reactTag = findNodeHandle(this.focusRef.current);
+      AccessibilityInfo.setAccessibilityFocus(reactTag);
+    }
+  };
+  render() {
+    this.focusRef = React.createRef();
```
TheSavior, line 762: I don't think you normally want to create the ref in render, that would wipe it away every time render is called. Normally this would be done in the constructor.

Since you should change this to hooks this problem will go away on its own anyways. :)

RNTester/js/examples/Accessibility/AccessibilityExample.js
graytmatterMS: I reimplemented the component with hooks. Or at least I think I did :). Was this done correctly? And does this solve the creating the ref in the render issue?

PR #29155, opened 2020-06-17
Update the documentation of the Button component

### LINKED ISSUE #33: Open-sourcing
I'm sure you all have talked about this a lot, but I would like to see this open-sourced sooner rather than later. I appreciate wanting to do something for the conf attendees, and that is pretty cool, but the longer this is open-sourced the more the rest of the community will feel excluded (as more and more tweets & posts are made, etc).

I have a blog post I'm working on, for example, but I'm not going to post it until this is open-sourced because otherwise I feel like I'm rubbing my early access in other's faces.

ayushjainrksh:
Fixes https://github.com/MLH-Fellowship/react-native/issues/33 

## Summary
The PR is part of an effort to update the code comments to match the current documentation on the React Native website. The project is a part of MLH fellowship program and involves automatic generation of the website docs from code comments and flow types as the end result.

To learn more about the project you can visit the project wiki: 
- [Project details](https://github.com/MLH-Fellowship/0.4.x-projects/wiki/React-Native-Flowtype-API-Docs-Generator)
- [RN Docs Standards](https://github.com/MLH-Fellowship/react-native/wiki/RN-Docs-standards)

Link to the documentation(the source of truth): 
- [button.md](https://github.com/MLH-Fellowship/react-native-website/blob/master/docs/button.md)

## Changes
* Update the title and prop description from docs.
* Remove unnecessary `*` from the code comments.
*  Add default value to props.
   * Specify the default value of a prop in the code comments with `@default` annotation.
   * For multiple defaults (different devices) use `@default {@platform android/ios}`.
* Update platforms in props.
  * Use comma separated string with `@platform` decorator.
* Add Snack player example specified in the docs to the code comments with `@example` annotation.

## Changelog
[Internal]

## Test Plan
All changes are made to the code comments and thus there is no need for testing.

[Reviewed by @jevakallio](https://github.com/MLH-Fellowship/react-native/pull/18)

### DIFF
--- a/Libraries/Components/Button.js
+++ b/Libraries/Components/Button.js
@@ -25,107 +25,194 @@ import type {ColorValue} from '../StyleSheet/StyleSheet';
 
 type ButtonProps = $ReadOnly<{|
   /**
-   * Text to display inside the button
+    Text to display inside the button. On Android the given title will be converted to the uppercased form.
    */
   title: string,
 
   /**
-   * Handler to be called when the user taps the button
+    Handler to be called when the user taps the button. The first function argument is an event in form of [PressEvent](pressevent).
    */
   onPress: (event?: PressEvent) => mixed,
 
   /**
-   * If true, doesn't play system sound on touch (Android Only)
-   **/
+    If `true`, doesn't play system sound on touch.
+
+    @platform android
+
+    @default false
+   */
   touchSoundDisabled?: ?boolean,
 
   /**
-   * Color of the text (iOS), or background color of the button (Android)
+    Color of the text (iOS), or background color of the button (Android).
+
+    @default {@platform android} '#2196F3'
+    @default {@platform ios} '#007AFF'
    */
   color?: ?ColorValue,
 
   /**
-   * TV preferred focus (see documentation for the View component).
+    TV preferred focus.
+
+    @platform tv
+
+    @default false
    */
   hasTVPreferredFocus?: ?boolean,
 
   /**
-   * TV next focus down (see documentation for the View component).
-   *
-   * @platform android
+    Designates the next view to receive focus when the user navigates down. See the [Android documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusDown).
+
+    @platform android, tv
    */
   nextFocusDown?: ?number,
 
   /**
-   * TV next focus forward (see documentation for the View component).
-   *
-   * @platform android
+    Designates the next view to receive focus when the user navigates forward. See the [Android documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusForward).
+
+    @platform android, tv
    */
   nextFocusForward?: ?number,
 
   /**
-   * TV next focus left (see documentation for the View component).
-   *
-   * @platform android
+    Designates the next view to receive focus when the user navigates left. See the [Android documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusLeft).
+
+    @platform android, tv
    */
   nextFocusLeft?: ?number,
 
   /**
-   * TV next focus right (see documentation for the View component).
-   *
-   * @platform android
+    Designates the next view to receive focus when the user navigates right. See the [Android documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusRight).
+
+    @platform android, tv
    */
   nextFocusRight?: ?number,
 
   /**
-   * TV next focus up (see documentation for the View component).
-   *
-   * @platform android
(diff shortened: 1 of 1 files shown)


Libraries/Components/Button.js
```diff
@@ -25,107 +25,210 @@ import type {ColorValue} from '../StyleSheet/StyleSheet';
...
    */
   hasTVPreferredFocus?: ?boolean,
 
   /**
-   * TV next focus down (see documentation for the View component).
-   *
-   * @platform android
+    Designates the next view to receive focus when the user navigates down. See
+    the [Android
+    documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusDown).
```
ayushjainrksh, line 68: @motiz88 I have used a `vscode` extension called `rewrap` to wrap the comments to markdown syntax. Some lines won't wrap to a maximum width(because they do not have a white space). The markdown doesn't wrap these lines either so in my opinion we should keep them as it is. What's your opinion?

Libraries/Components/Button.js
```diff
@@ -25,107 +25,210 @@ import type {ColorValue} from '../StyleSheet/StyleSheet';
...
+    onPress={onPressLearnMore}
+    title="Learn More"
+    color="#841584"
+    accessibilityLabel="Learn more about this purple button"
+  />
+  ```
+
+  @example ```SnackPlayer name=Button%20Example
+  import React from 'react';
+  import { StyleSheet, Button, View, SafeAreaView, Text, Alert } from 'react-native';
```
ayushjainrksh, line 151: @motiz88 I haven't wrapped any line of code present in the comments as I think this would cause unnecessary complications when we parse them. Also, they would require manual wrapping if we decide doing this. What do you suggest?

>This line is 84 characters.

motiz88: That makes sense 👍🏻

Libraries/Components/Button.js
motiz88, line 68: Short term:

I wouldn't insert a line break inside the text of a link just because it's followed by a long URL.

```suggestion
    the [Android documentation](https://developer.android.com/reference/android/view/View.html#attr_android:nextFocusDown).
```

Long term: (e.g. we can choose to do this for future PRs)

I'd refactor these comments to use reference-style links, improving the flow of the text.

```suggestion
    the [Android documentation][android:nextFocusDown].
    
    [android:nextFocusDown]: https://developer.android.com/reference/android/view/View.html#attr_android
```

ayushjainrksh: Thanks. I'll amend the commit to incorporate the changes for a long term.

ayushjainrksh:
@motiz88 Sorry for the delay with this one. I started working on the sync scripts which gave me a broader perspective to the structure of the comments. I've fixed the links to follow reference-style. Please let me know if the PR need any more changes.

Libraries/Components/Button.js
```diff
@@ -25,107 +26,220 @@ import type {ColorValue} from '../StyleSheet/StyleSheet';
...
- * ```
- *
+  A basic button component that should render nicely on any platform. Supports a
+  minimal level of customization.
+
+  If this button doesn't look right for your app, you can build your own button
+  using [TouchableOpacity](touchableopacity) or
+  [TouchableWithoutFeedback](touchablewithoutfeedback). For inspiration, look at
+  the [source code for this button
+  component](https://github.com/facebook/react-native/blob/master/Libraries/Components/Button.js).
```
motiz88, line 147: @ayushjainrksh, sorry to block this PR again on a minor detail, but could you please go through all links and fix the ones that are causing awkward line breaks? (e.g. using references like you did elsewhere)

ayushjainrksh: @motiz88 Ahh, sorry it went unnoticed. Fixed it now :)

ayushjainrksh:
@motiz88  The most recently updated components(that follow new guidelines for documentation) are `Button.js` and `ActivityIndicator.js`. They were updated in [this PR](https://github.com/facebook/react-native-website/pull/1959). I noticed a small recent documentation change in **Button** component in [this line](https://github.com/facebook/react-native-website/pull/1959/files#diff-6c4d3b1549f3b867a874be4e1c227adeL113). I am adding a commit to update this PR to latest version of the docs.

motiz88:
@ayushjainrksh I just noticed your most recent commit, sorry. The version I previously imported is landing now (it was blocked on an internal CI config change) so you may want to resubmit that small change as its own PR.

ayushjainrksh:
@motiz88 Okay sure thing :+1:

Asif-Malik42:
how do I run this Button code on expo. I am new to the environment . Came here while reading the documentation.
Capture

When I run this on expo following error shows.

Issue #28279, opened 2020-03-10
Android ignores unicode soft hyphen (u00AD)


ccfz:
Description:
On android soft-hyphen unicode `u00AD` is ignored, neither the word break nor the hyphen are displayed correctly. Using `u200B` is a workaround for getting better word wrapping results but is still missing the hyphen. 

React Native version:

```
System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Memory: 108.98 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.10.1 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.7 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.0, 29.0.1
      System Images: android-28 | Google Play Intel x86 Atom, android-29 | Intel x86 Atom_64
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.12.0 => 16.12.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

## Steps To Reproduce

Render the following on android vs. iOS

```javascript
const testString = 'sup\u00ADer\u00ADfrag\u00ADilis\u00ADtice\u00ADxpial\u00ADidocious'

return (
  <View>
    <Text style={{ width: 100 }}>
      {testString}
    </Text>
  </View>
);
```

## Expected Results

Screenshot from iOS
![IMG_DB2643BEA609-1](https://user-images.githubusercontent.com/6050682/76325546-bb720280-62e7-11ea-98ae-aba541486e13.jpeg)

## Actual Result 
Screenshot from Android
![IMG_25E611476E59-1](https://user-images.githubusercontent.com/6050682/76325607-d04e9600-62e7-11ea-8398-0811775e5970.jpeg)

plaa:
[...]

PR #29157, opened 2020-06-17
[Android] Adding Hyphenation Frequency prop for Text component

fabriziobertoglio1987:
## Summary

This issue fixes https://github.com/facebook/react-native/issues/28279
android_hyphenationFrequency prop for Android Text component which sets the frequency of automatic hyphenation to use when determining word breaks. 

## Changelog

[Android] [Fixed] - Adding Hyphenation Frequency prop for Text component

## Test Plan

More info are available in the [android docs](https://developer.android.com/reference/android/widget/TextView#setHyphenationFrequency(int)). I will add the documentation to the docs later once the pull request is taken in consideration for merging.

| **AFTER** | 
|:-------------------------:|
|  | 

I remain available to do improvements. Thanks a lot. Fabrizio.

### DIFF
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -68,7 +68,7 @@ const viewConfig = {
     onTextLayout: true,
     onInlineViewLayout: true,
     dataDetectorType: true,
-    androidHyphenationFrequency: true,
+    android_hyphenationFrequency: true,
   },
   directEventTypes: {
     topTextLayout: {
--- a/Libraries/Text/Text.js
+++ b/Libraries/Text/Text.js
@@ -68,6 +68,7 @@ const viewConfig = {
     onTextLayout: true,
     onInlineViewLayout: true,
     dataDetectorType: true,
+    androidHyphenationFrequency: true,
   },
   directEventTypes: {
     topTextLayout: {
--- a/Libraries/Text/TextProps.js
+++ b/Libraries/Text/TextProps.js
@@ -62,7 +62,13 @@ export type TextProps = $ReadOnly<{|
    * Set hyphenation strategy on Android.
    *
    */
-  androidHyphenationFrequency?: ?('normal' | 'none' | 'full' | 'high' | 'balanced'),
+  androidHyphenationFrequency?: ?(
+    | 'normal'
+    | 'none'
+    | 'full'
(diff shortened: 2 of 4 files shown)


TheSavior:
I can't confirm that this will get approved (I don't have context on the Android side), but we've been following a pattern of the platform prefixes being `android_`, so this prop should be `android_hyphenationFrequency`. This is also seen in these props on Pressable: https://github.com/facebook/react-native/blob/master/Libraries/Components/Pressable/Pressable.js#L122-L127

fabriziobertoglio1987:
thanks a lot @TheSavior . Changes are included in https://github.com/facebook/react-native/pull/29157/commits/4500095ef33a6a95f7daa8debd6fa9f069ba7ea3. I'm fixing the same issue also in https://github.com/facebook/react-native/pull/29059. I have no prerogative for the PR to be merged, but just hope to get a better insight in the react-native project and write more relevant pr in the future.. Thanks a lot. Fabrizio

JoshuaGross:
This seems pretty reasonable to me :)

JoshuaGross:
@fabriziobertoglio1987 I'm working on testing internally and having some conversations about the prefix. While I'm doing that, could you issue a PR to the docs site to add documentation for the new prop?

https://github.com/facebook/react-native-website/tree/master/docs

fabriziobertoglio1987:
Thanks a lot @Joshua I added the docs and also added a new commit to this Pull Request https://github.com/facebook/react-native/pull/29157/commits/14709f80a3bd035fbcce69a7dc73234c3cffaf36  as [setHyphenationFrequency](https://developer.android.com/reference/android/widget/TextView#setHyphenationFrequency(int)) android method is available only from API 23+. 

I decided to write this PR as a solution of https://github.com/facebook/react-native/issues/28279 to give users more control on the Text Component, but this could be not the optimal solution. I don't work on iOS.

I tested on RNTester example. I remain available. Thanks a lot. I wish you a good day. Fabrizio

JoshuaGross:
@fabriziobertoglio1987 Thank you for following up on this! The amendment is being merged in.

fabriziobertoglio1987:
https://github.com/facebook/react-native/pull/29258

PR #29171, opened 2020-06-18
Restore Previous Behavior for StyleSheet Validation of Null/Undefined Styles

NickGerleman:
## Summary

#27264 changed stylesheet validation to avoid enumerating properties on the prototype of a style. It introduces a secondary behavior change, where null/undefined styles used to be tolerated but now lead to an exception. This is because `for in undefined` will noop where `for of Object.keys(undefined)` will throw.

This scenario of undefined/null styles seems to actually show up in practice and was previously well tolerated. E.g. `Button.js` has code that looks like this:

```jsx
  button: Platform.select({
    ios: {},
    android: {
      elevation: 4,
      // Material design blue from https://material.google.com/style/color.html#color-color-palette
      backgroundColor: '#2196F3',
      borderRadius: 2,
    },
  }),
```

For non ios/Android platforms, that creates a style object which looks like:
```js
{
  button: undefined,
  ...
}
```

This previously meant that the component would be unstyled if created, but now means out-of-tree platforms throw if the builtin Button component is required.

This change restores the previous `for in` loop but adds a `hasOwnProperty` check to avoid properties on prototypes.

## Changelog

[General] [Fixed] - Restore Previous Behavior for StyleSheet Validation of Null/Undefined Styles

## Test Plan

Validated that importing Buttons will no longer cause an exception, and that invalid properties are still caught.

### DIFF
--- a/Libraries/StyleSheet/StyleSheetValidation.js
+++ b/Libraries/StyleSheet/StyleSheetValidation.js
@@ -51,14 +51,16 @@ class StyleSheetValidation {
     if (!__DEV__ || global.__RCTProfileIsProfiling) {
       return;
     }
-    for (const prop in styles[name]) {
-      if (styles[name].hasOwnProperty(prop)) {
-        StyleSheetValidation.validateStyleProp(
-          prop,
-          styles[name],
-          'StyleSheet ' + name,
-        );
-      }
+    if (!styles[name]) {
+      return;
+    }
+    const styleProps = Object.keys(styles[name]);
+    for (const prop of styleProps) {
+      StyleSheetValidation.validateStyleProp(
+        prop,
+        styles[name],
+        'StyleSheet ' + name,
+      );
     }
   }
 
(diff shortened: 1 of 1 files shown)


akshetpandey:
Surely a null check is a better solution?
Something like
```js
const styleProps = Object.keys(styles[name] || {});
```
or
```js
if (!styles[name]) {
  return
}

NickGerleman:
> Surely a null check is a better solution?
> Something like
> 
> ```js
> const styleProps = Object.keys(styles[name] || {});
> ```
> 
> or
> 
> ```js
> if (!styles[name]) {
>   return
> }
> ```

Yeah, I was thinking about that. It's certainly more explicit, which I like. The only concern I had was that there may be other subtle behavior changes from the previous implementation, but that might just be a bit to risk averse. I'd have no issue changing it though.

Karniej:
@NickGerleman I have this issue inside RN 0.63v & can't run the app, do you know when that fix may be in the RN ?

PR #29177, opened 2020-06-19
Align multi-line TextInput onSubmitEditing behavior

tido64:
## Summary

Aligns behavior to be consistent with Android and the [documentation](https://reactnative.dev/docs/textinput#bluronsubmit). `onSubmitEditing` should not be called when `blurOnSubmit=false`.

## Changelog

[iOS] [Fixed] - Align multi-line TextInput onSubmitEditing behavior

## Test Plan

textinput-blur-align

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -365,11 +365,13 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
   // `onSubmitEditing` is called when "Submit" button
   // (the blue key on onscreen keyboard) did pressed
   // (no connection to any specific "submitting" process).
-  [_eventDispatcher sendTextEventWithType:RCTTextEventTypeSubmit
-                                 reactTag:self.reactTag
-                                     text:self.backedTextInputView.attributedText.string
-                                      key:nil
-                               eventCount:_nativeEventCount];
+  if (_blurOnSubmit) {
+    [_eventDispatcher sendTextEventWithType:RCTTextEventTypeSubmit
+                                   reactTag:self.reactTag
+                                       text:self.backedTextInputView.attributedText.string
+                                        key:nil
+                                 eventCount:_nativeEventCount];
+  }
 
   return _blurOnSubmit;
 }


shergin:
IDK. Where exactly the documentation says that? I only found this:

> blurOnSubmit
If true, the text field will blur when submitted. The default value is true for single-line fields and false for multiline fields. Note that for multiline fields, setting blurOnSubmit to true means that pressing return will blur the field and trigger the onSubmitEditing event instead of inserting a newline into the field.

To me, it does not say anything about the case where `blurOnSubmit` is `false`.
I don't really know what's the right and most useful and versatile behavior.

Do we have any motivation for this use-case? Maybe we should change Android?

tido64:
> To me, it does not say anything about the case where `blurOnSubmit` is `false`.
> I don't really know what's the right and most useful and versatile behavior.

My interpretation of "_instead of`_" in the documentation describes the opposite case, i.e.:

- `blurOnSubmit=true`: pressing return will blur the field and trigger the onSubmitEditing event
- `blurOnSubmit=false`: inserting a newline into the field

If you're asking me what _I_ think should be the correct behaviour, I think `blurOnSubmit=false` should just submit and not insert a newline or blur. I can't think of a single use case where inserting a newline in addition to submitting makes any sense.

Regardless, this change aligns iOS with Android. Right now, macOS is aligned with Android (as of [v0.61.49](https://github.com/microsoft/react-native-macos/commit/2803dc8799dfd503d5eb740b7cdd91648077ed41)). If you think we should change the behaviour to something that is more useful, I am happy to do that but I need to know what that something is 😄

shergin:
Oh, IDK, it's a tough choice.
Seems you put a lot of thoughts into that. Let's do the right thing. What should we do?

tido64:
@shergin: This is what I'm thinking:

| Props                              | Pressing Enter... | Inserts newline |
|:-----------------------------------|:------------------|:----------------|
| (none)                             | inserts newline   | yes             |
| `blurOnSubmit`                     | inserts newline   | yes             |
| `onSubmitEditing`                  | submits only      | no              |
| `onSubmitEditing` + `blurOnSubmit` | submits and blurs | no              |

In all cases, pressing Shift+Enter will always insert a newline, and will neither submit nor blur.

I think this matches most chat clients and forms that I've interacted with. What do you think?

PR #29211, opened 2020-06-24
Animated: Interpolation: add 'immutable' configuration property

RichardFevrier:
## Summary

Hello everybody,
In my case I have a complex animated svg path. 
So I am calculating all the coordinates for each step of the animation.

This PR adds a property called immutable, adding the behavior of not trying to calculate the coordinates again but just using my previously calculated values as is.

## Changelog

[General] [Added] - Add immutable prop to AnimatedInterpolation

### DIFF
--- a/Libraries/Animated/src/nodes/AnimatedInterpolation.js
+++ b/Libraries/Animated/src/nodes/AnimatedInterpolation.js
@@ -28,6 +28,7 @@ export type InterpolationConfigType = {
   extrapolate?: ExtrapolateType,
   extrapolateLeft?: ExtrapolateType,
   extrapolateRight?: ExtrapolateType,
+  immutable?: boolean,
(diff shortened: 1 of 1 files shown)


JoshuaGross:
@RichardFevrier Could you add an RNTester example?

This is just a perf optimization, right? It shouldn't change how the animation appears visually?

I'm wondering if there's a way of doing this without introducing a new flag.

RichardFevrier:
Hi @JoshuaGross, thank you for taking time to answer to me.

I have used this solution in my project for the past two years because one of my animations was broken with the default implementation. So it wasn't only a question of performance for sure. (I wrote "wasn't" because I don't know if it still the case)

If you can think about a better implementation that would be great because I believe we should provide this mechanic to everyone.

PR #29219, opened 2020-06-25
fix: docs for `reactAddControllerToClosestParent` method

troZee:
## Summary

I added small change to method description, because `reactAddControllerToClosestParent` method does not return a `boolean` value. 

## Changelog

[Internal] [Changed] - `reactAddControllerToClosestParent` method description

## Test Plan

You can check method declaration here: https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Views/UIView%2BReact.m#L241

### DIFF
--- a/React/Views/UIView+React.h
+++ b/React/Views/UIView+React.h
@@ -83,9 +83,9 @@
 
 /**
  * This method attaches the specified controller as a child of the
- * the owning view controller of this view. Returns NO if no view
- * controller is found (which may happen if the view is not currently
- * attached to the view hierarchy).
+ * the owning view controller of this view. If host view controller not found,
+ * `controller` will not be added to the parent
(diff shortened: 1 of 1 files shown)


React/Views/UIView+React.h
```diff
@@ -83,9 +83,9 @@
 
 /**
  * This method attaches the specified controller as a child of the
- * the owning view controller of this view. Returns NO if no view
- * controller is found (which may happen if the view is not currently
- * attached to the view hierarchy).
+ * the owning view controller of this view. If host view controller not found,
```
eduardopelitti, line 86: Hi 👋 ! Perhaps this should be phrased as "If a host view controller is not found..." or "If no host view controlled is found..."

PR #29230, opened 2020-06-26
babel-plugin-inline-view-configs cannot run externally

acoates-ms:
## Summary

Attempting to run `babel-plugin-inline-view-configs` in `react-native-windows` to verify our out of tree platform view managers align with core.  Hit a couple of issues.

The first one is that when trying to use the plugin with the standard react-native babel-preset.  The default preset has some plugins which use the test option (in particular the typescript plugins).  In order for babel to know if those plugins should be run or not on the view-config, we need to ensure that we pass along the filename when calling into babel parse.

The second issue is several generated requires are using haste imports - so I changed them to full paths.

Note the `babel-plugin-inline-view-configs` still needs to be published externally at some point, and the `react-native-codegen` package needs to be published without flow types in order to fully unblock this.

## Changelog

[Internal] [Fixed] - babel-plugin-inline-view-configs cannot run externally

## Test Plan

I ran a version of this locally in react-native-windows.  But there have been a few changes since the version I was working on.  And this code isn't run in OSS.  So it'll require internal validation.  But this code is all run during build, so it should be safe if the build works.

### DIFF
--- a/packages/babel-plugin-inline-view-configs/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/babel-plugin-inline-view-configs/__tests__/__snapshots__/index-test.js.snap
@@ -22,7 +22,7 @@ interface NativeCommands {
   +scrollTo: (viewRef: React.ElementRef<NativeType>, y: Int32, animated: boolean) => void,
 }
 
-const registerGeneratedViewConfig = require('registerGeneratedViewConfig');
+const registerGeneratedViewConfig = require('react-native/Libraries/Utilities/registerGeneratedViewConfig');
 
 const {
   dispatchCommand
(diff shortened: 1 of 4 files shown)


TheSavior:
I tried to land this and ran into some problems based on internal tooling due to the codegen adding new imports via the babel transform. We require those to be statically analyzable from the original source.

The recommendation was:

1) Make sure that react-native is necessarily a dependency of the file where the dynamic dependencies would be injected.
2) Change the dynamically injected imports to be part of the react-native Public API (possibly via things like `require('react-native).UNSTABLE_registerGeneratedViewConfig`).
3) Dynamically inject imports to the react-native API instead of nested modules.

TheSavior:
Closing this as I think it is sufficiently out of date, had trouble landing, and there is probably a new approach here now as the codegen code has changed.

Issue #26293, opened 2019-09-01
KeyboardAvoidingView stop working if the keyboard type changes (iOS)


iKrushYou:
I'm implementing the `KeyboardAvoidingView` inside of a modal that takes in user input as a string and also a currency value. This requires different keyboard types. It seems that when switching between the normal keyboard and the numeric keyboard, the view (and I'm using height to make sure the view is centered) stops adjusting the container based on the keyboard.

```
info 
  React Native Environment Info:
    System:
      OS: macOS 10.15
      CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
      Memory: 734.70 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 12.5.0 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
    IDEs:
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: https://github.com/expo/react-native/archive/sdk-34.0.0.tar.gz => 0.59.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7
```

![KAV-broken mov](https://user-images.githubusercontent.com/1500721/64083373-fc5fdd80-ccec-11e9-9656-d7e64459fd5f.gif)

https://snack.expo.io/SyqoJRYrB

The strange thing is: it works sometimes when on my actual device because I use Gboard but the numberpad comes from the apple keyboard. So it has to switch entirely. But the whole thing flashes and glitches out (not a very elegant animation)

KA-32:
Confirm the same in RN 0.60.5

samiacto:
I'm experiencing this too in RN 0.61.2 when switching between default keyboard and emoji keyboard on iOS

ArthurBugan:
Any fix to this problem?, I got this problem on android 10 and expo sdk 35

ramisalem:
Any fix to this problem?

bockc:
[...]

PR #29239, opened 2020-06-28
Fix KeyboardAvoidingView to keep frame height

shunwitter:
### KeyboardAvoidingView

There is an issue like below when `behavior="height"` is selected.

KeyboardAvoidingView stops working when you change input type (e.g. from English to Emoji).
This is how to reproduce the issue.

- Use `KeyboardAvoidingView behavior="height"`
- Focus input field
- Keyboard appears
- Change keyboard type
- `KeyboardAvoidingView` goes back to the original height, therefore nested elements return to the original position.

Related issue: https://github.com/facebook/react-native/issues/26293

### KeyboardAvoidingViewExample

Added `height` option example in RNTester.

## Changelog

[General] [Fixed] - Fix KeyboardAvoidingView's behavior "height"

### Before

KeyAvoidingView-demo

### After

Fixed

### DIFF
--- a/Libraries/Components/Keyboard/KeyboardAvoidingView.js
+++ b/Libraries/Components/Keyboard/KeyboardAvoidingView.js
@@ -87,7 +87,7 @@ class KeyboardAvoidingView extends React.Component<Props, State> {
 
     // Calculate the displacement needed for the view such that it
     // no longer overlaps with the keyboard
-    return Math.max(frame.y + frame.height - keyboardY, 0);
+    return Math.max(frame.y + this._initialFrameHeight - keyboardY, 0);
   }
 
   _onKeyboardChange = (event: ?KeyboardEvent) => {
--- a/RNTester/js/examples/KeyboardAvoidingView/KeyboardAvoidingViewExample.js
+++ b/RNTester/js/examples/KeyboardAvoidingView/KeyboardAvoidingViewExample.js
@@ -24,6 +24,8 @@ const {
 const RNTesterBlock = require('../../components/RNTesterBlock');
 const RNTesterPage = require('../../components/RNTesterPage');
 
+const behaviorOptions = ['padding', 'position', 'height'];
+
 type Props = $ReadOnly<{||}>;
 type State = {|
   behavior: string,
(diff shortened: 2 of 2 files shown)


vomchik:
@shunwitter Thanks! Works for me.
Could you fix conflicts?

shunwitter:
@vomchik I guess it's already fixed thanks to [this commit](https://github.com/facebook/react-native/commit/b08fff6f869e00c20c0dcdf7aca71284c2f276f0).

krskrs:
Hi @shunwitter I'm resuming this PR  from 1 year ago, because we're facing this exact issue and it looks like that so far the only reasonable fix is yours. You said that the issue had been fixed in another commit, but after checking it thoroughly it doesn't seem the case. Latest code, on master, still doesn't include the key change you proposed in this PR, and none of the changes from the commit you referenced to as the fix seem to address this particular issue.
Currently we have 'forked' this component and we're using our version, with a fix identical to yours, but we believe this issue should be resumed and taken into consideration b/c it's a quite relevant one.
Can I ask you if you have kept using your patched code? Or if you just have stopped using react native and lost the interest in this PR, as we're almost 100% sure the issue is still there.

shunwitter:
hi @krskrs, I recently noticed this issue still exists when I create a new react-native project. I have been using my version of KeyboardAvoidingView component but maybe I can create a new PR for this.

krskrs:
Hi @shunwitter ! I have been working on a small variation of this fix. The core principle is the same, but I also tried to use a more accurate height than the one simply captured on first onLayout event, because, in some cases, I have seen multiple consecutive onLayout events with slightly different height values occurring just after component instantiation. Moreover, just in case there were external causes which might cause the base frame height to change, I try to keep it updated, as, again, using the initial value won't be enough in such cases.

https://gist.github.com/krskrs/4a3d347a8a42435e74dc32b8c3abe93b

shunwitter:
Thank for sharing it, @krskrs . I'll try that out :)

krskrs:
@shunwitter thank you. When you do it, feel free to comment or ask anything, if you want. Any comment or suggestion would be appreciated. ;)  Meanwhile I will keep it updated in case we make any further changes.

PR #29263, opened 2020-07-01
Fixing CI error "Entry file RNTester/js/RNTesterApp.ios.js does not exist"

graytmatterMS:
## Summary

PRs are failing with the error "Entry file RNTester/js/RNTesterApp.ios.js does not exist", despite it existing. The if statement around the checker will always trigger because when it looks to see if RNTester/js/RNTesterApp.ios.js exists it's inside of RNTester instead of root. I've added a "../" to solve this.

## Changelog

[Internal] [Fixed] - Fixing CI error "Entry file RNTester/js/RNTesterApp.ios.js does not exist"

## Test Plan
Screen Shot 2020-07-01 at 11 25 03
Screen Shot 2020-07-01 at 11 26 52

### DIFF
--- a/scripts/react-native-xcode.sh
+++ b/scripts/react-native-xcode.sh
@@ -83,7 +83,7 @@ else
   ENTRY_FILE=${1:-index.js}
 fi
 
-if [[ $DEV != true && ! -f "$ENTRY_FILE" ]]; then
+if [[ $DEV != true && ! -f "../$ENTRY_FILE" ]]; then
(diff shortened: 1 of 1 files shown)


graytmatterMS:
@hramos @TheSavior Could you take a look at this? I'm pretty confident https://github.com/facebook/react-native/pull/29012 introduced this error and this should fix things.

janicduplessis:
I'm also getting issues with this commit with a monorepo setup where metro runs at the root of the repo. This means ENTRY_FILE will be something like apps/app/index.js and never exist under the project dir. I suggest just reverting or making this check a warning and not fail the build.

PR #29270, opened 2020-07-03
Update JSLoader.cpp

sunnylqm:
Since react-native-cli is deprecated, the correct command should be `npx react-native start`


sunnylqm:
ping @grabbou @kelset

kelset (approved):

PR #29285, opened 2020-07-06
Fix passing react native path in Podfile template

janicduplessis:
## Summary

Since https://github.com/react-native-community/cli/commit/e949e234b03fb65f8e4ed2706dfaa745aa59a14f#diff-d1800049b92343288bcbc1c484575058 the RN cli script returns an object with `:reactNativePath` instead of just JSON. Not super familiar with how objects / JSON works in ruby but using this syntax instead works.
 
## Changelog

[Fixed] [iOS] - Fix passing react native path in Podfile template

## Test Plan

Tested in a project inside a monorepo using the latest version of RN CLI that the proper react-native path is now passed.

### DIFF
--- a/template/ios/Podfile
+++ b/template/ios/Podfile
@@ -6,7 +6,7 @@ platform :ios, '10.0'
 target 'HelloWorld' do
   config = use_native_modules!
 
-  use_react_native!(:path => config["reactNativePath"])
+  use_react_native!(:path => config[:reactNativePath])
 
   target 'HelloWorldTests' do
     inherit! :complete


janicduplessis:
cc @grabbou

fabriziobertoglio1987:
https://stackoverflow.com/a/8189435/11452286

keima:
Please fix this issue... 🥺
Almost of project has nothing problem, but example project inside native module project couldn't determine react-native path.
(e.g. https://github.com/react-native-community/react-native-share : example project don't have package.json but library(root) has.)

...

I searched caused by this issue.

Ruby's hash has two key types: "symbol" and "string".

`use_native_modules!` returns hash has `:reactNativePath` **symbol**.

see: https://github.com/react-native-community/cli/blob/v4.13.0/packages/platform-ios/native_modules.rb#L125

`config[:reactNativePath]` refers `:reactNativePath` **symbol**-ed key on `config` hash, and
`config["reactNativePath"]` refers `"reactNativePath"` **string**-ed key.
This operation is NOT same.

BTW, the reason why almost of current program seems to be running fine, assigns `prefix` var to placeholder when `config["reactNativePath"]` returns nil.

see: https://github.com/facebook/react-native/blob/v0.63.2/scripts/react_native_pods.rb#L8


Now, argument of `use_react_native` is not being used.
So if we want to or have to change react-native path, fix Podfile manually.
It's a little inconvenient.

janicduplessis:
cc @fkgozali

fkgozali:
@hramos: could you help merge this one?

hramos (approved):

611161512:
I also have this problem, when the path of the node_modules file is level with the path path of the IOS project, there is a problem, can't find the corresponding subspec. Because the following path requires the IOS project file name, which must be "IOS" :
"Project_root = Pathname. New (config [" the project"] [" ios "] [" sourceDir "])"
Can you design a parameter that developers can pass in to their podfile?

PR #29287, opened 2020-07-07
add geo location patch for ios

ideabits:
## Summary

This fix is for Geolocation accuracy problem in iOS.


yungsters:
@ideabits, can you provide a bit more context on what the problem is and how this fixes it? Also, can you explain why this is being merged into the `0.59-stable` branch instead of `main`?

Issue #29282, opened 2020-07-06
StatusBar.pushStackEntry with backgroundColor causes crash unless animated is specified


iAmGhost:
## Description
on Android, StatusBar.pushStackEntry with backgroundColor causes crash unless animated is specified.

Can be avoided by passing animated prop explicitly(true/false).

## React Native version: 0.62.2
```
System:
    OS: macOS 10.15.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 2.04 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.4.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.5 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6514223
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.7 - /usr/bin/javac
    Python: 3.8.1 - /Users/daehankim/.pyenv/shims/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.11.0 => 16.11.0 
    react-native: 0.62.2 => 0.62.2 
  npmGlobalPackages:
    *react-native*: Not Found
```
## Steps To Reproduce
1. Run code:
```javascript
StatusBar.pushStackEntry({
  backgroundColor: 'white',
});
```

2. This happens:

![device-2020-07-06-185912](https://user-images.githubusercontent.com/1367707/86581253-d402d880-bfba-11ea-84f7-4135d670ff7e.png)

## Expected Results
Error should not be caused.

PR #29289, opened 2020-07-07
Fix crash on StatusBar.pushStackEntry if animated property is not provided

drmas:
## Summary

On Android, `StatusBar.pushStackEntry` with `backgroundColor` only causes crash unless animated is specified. as reported in this bug https://github.com/facebook/react-native/issues/29282

This is fixed by simply providing false as a default value.

## Changelog

[Android] [fixed] - Fix android crash when calling `StatusBar.pushStackEntry` without provide animated property value.

## Test Plan

- Create a new project and call on app start

```
StatusBar.pushStackEntry({
  backgroundColor: 'white',
});
```
It should not cause a crash.

Here is a [Expo snack](https://snack.expo.dev/@drmas/f64393) for testing the crash

### DIFF
--- a/Libraries/Components/StatusBar/StatusBar.js
+++ b/Libraries/Components/StatusBar/StatusBar.js
@@ -473,7 +473,7 @@ class StatusBar extends React.Component<Props> {
           );
           NativeStatusBarManagerAndroid.setColor(
             processedColor,
-            mergedProps.backgroundColor.animated,
+            mergedProps.backgroundColor.animated || false,
           );
         }
         if (!oldProps || oldProps.hidden.value !== mergedProps.hidden.value) {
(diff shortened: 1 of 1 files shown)


drmas:
@chrisglein any chance we can have code review on this?

chrisglein:
> @chrisglein Chris Glein FTE any chance we can have code review on this?

I only have triage permissions (labeling and such), not contributor permissions, so a review from me won't help you :(

charlesbdudley:
@drmas Thanks for the Expo snack! Very helpful. For style consistency, can we match the style of line 439? https://github.com/facebook/react-native/blob/2380aff5bba9a786ce60428db6883f37743f8932/Libraries/Components/StatusBar/StatusBar.js#L439

drmas:
@charlesbdudley it is updated to the match the code style

Issue #27526, opened 2019-12-16
Android: KeyboardAvoidingView white-space on bottom of screen when android:windowTranslucentStatus is set to true


hrastnik:
KeyboardAvoidingView leaves space on bottom of screen when the keyboard is closed.
This bug manifests when the `android:windowTranslucentStatus` is set to `true`. (Note that this is the case for project initialized with Expo)

![kav-bug](https://user-images.githubusercontent.com/3407234/70899677-62037b80-1ff7-11ea-8ff8-567285704d27.png)

React Native version:
```
System:
    OS: macOS 10.15.2
    CPU: (4) x64 Intel(R) Core(TM) i5-4308U CPU @ 2.80GHz
    Memory: 68.76 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.4.0 - ~/.nvm/versions/node/v12.4.0/bin/node
    Yarn: 1.19.1 - ~/.nvm/versions/node/v12.4.0/bin/yarn
    npm: 6.13.1 - ~/.nvm/versions/node/v12.4.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.0
      System Images: android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.2.1/11B500 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1
```
## Steps To Reproduce

1. Initialize a new RN project
2. Add `<item name="android:windowTranslucentStatus">true</item>` to `android/app/src/main/res/values/styles.xml`
3. Copy/paste the following code to `App.js`

```javascript
import React from 'react';
import {TextInput, View, KeyboardAvoidingView} from 'react-native';

export default () => (
  <KeyboardAvoidingView behavior="padding" style={{flex: 1}}>
    <View
[...]

Issue #27089, opened 2019-11-02
Keyboard.addListener('keyboardDidShow', ...) callback is invoked with inaccurate measures in Android devices with notch


jsamr:
### Incriminated API

``` js
// In Android
Keyboard.addListener('keyboardDidShow', e => {
   // On OnePlus 7, and Pixel 3XL (simulator) this value is underestimated by exactly StatusBar.currentHeight.
    e.endCoordinates.height
})
```

### Prerequisites

In `AndroidManifest.xml`, set `android:windowSoftInputMode="adjustPan"`.

### Visual tests

*The black horizontal bar should be the same width (40dp) as the red bar, in a container with a bottom padding of the keyboard's height. This component mimics KeyboardAvoidingView with behavior padding set. The height of keyboard should be the same as the height of the red bar. The issue has been reproduced with GBoard, SwiftKey...*

**Remarks**: *I would have imagined that the view height would equal keyboard height + keyboard Y. But that is only true ... on OnePlus7, which is the problematic device!*

| Emulator, Pixel 2 XL, Android 10 | OnePlus 7, Android 10, OxygenOS 10.0.1 |
|:-------------:|:-------------:|
| <img src="https://user-images.githubusercontent.com/3646758/68069734-282a2e80-fd43-11e9-80ab-891cd31aa08f.png" width="400"> | <img src="https://user-images.githubusercontent.com/3646758/68069735-2a8c8880-fd43-11e9-8805-ee96e1eaea46.jpg" width="400"> |

React Native version:
```
System:
    OS: Linux 4.19 Manjaro Linux undefined
    CPU: (8) x64 Intel(R) Core(TM) i7-8809G CPU @ 3.10GHz
    Memory: 565.35 MB / 31.33 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.13.0 - /usr/bin/node
    npm: 6.12.0 - /usr/bin/npm
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
  npmPackages:
    react: 16.8.1 => 16.8.1 
    react-native: 0.61.3 => 0.61.3
```

## Steps To Reproduce

1. Open the reproduction demo in a Pixel 3XL emulator with Android 10+.
2. Observe that `KeyboardAvoidingView` with behavior padding doesn't add enough padding.

Describe what you expected to happen: The padding should match keyboard's real height.

**Full reproduction** : [react-native-oneplus7-keyboard-height-discrepancy
[...]

PR #29292, opened 2020-07-07
[Android] keyboardDidHide wrong screenY coordinates with windowTranslucentStatus=true

fabriziobertoglio1987:
## Summary

This issue fixes https://github.com/facebook/react-native/issues/27526, related issue https://github.com/facebook/react-native/issues/27089
Avoid returning the wrong screenY coordinates on event keyboardDidHide. 
getWindowVisibleDisplayFrame retrieves the overall visible display size in which the window this view is attached to has been positioned in. This takes into account screen decorations above the window, for both cases where the window itself is being position inside of them or the window is being placed under then and covered insets are used for the window to position its content inside. In effect, this tells you the available area where content can be placed and remain visible to users, since anything below the StatusBar is not visible to the user, the method does not work with translucent StatusBar (android:windowTranslucentStatus). 

This commit fixes this issues removing the white bar at the bottom of the view when using windowTranslucentStatus.

## Changelog

[Android] [Fixed] - keyboardDidHide wrong screenY coordinates with windowTranslucentStatus=true

## Test Plan

Works in all scenarios, but **I did not change RNTester `windowTranslucentStatus`**. I would like to discuss the potential breaking changes connected to not using [getWindowVisibleDisplayFrame](https://developer.android.com/reference/android/view/View#getWindowVisibleDisplayFrame(android.graphics.Rect)) with `keyboardDidHide`. 

I would be happy to build an alternative functionality to calculate the WindowVisibleDisplayFrameHeight under request from the Facebook Team.

**WITHOUT** `android:windowTranslucentStatus`

| **BEFORE** |
|:-------------------------:|
|  | 

| **AFTER** | 
|:-------------------------:|
|  |

**WITH** `android:windowTranslucentStatus`

| **BEFORE** | 
|:-------------------------:|
|  |

| **BEFORE (log)** | 
|:-------------------------:|
|  |

| **AFTER** | 
|:-------------------------:|
|  |

| **AFTER (log)** | 
|:-------------------------:|
|  |

RNTester **WITHOUT** `android:windowTranslucentStatus`

| **BEFORE** | **AFTER** | 
|  |  |

I remain available to do improvements. Thanks a lot. Fabrizio.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
@@ -684,7 +684,7 @@ public class ReactRootView extends FrameLayout implements RootView, ReactRoot {
         sendEvent(
             "keyboardDidHide",
             createKeyboardEventPayload(
-                PixelUtil.toDIPFromPixel(mVisibleViewArea.height()),
+                PixelUtil.toDIPFromPixel(mLastHeight),
                 0,
                 PixelUtil.toDIPFromPixel(mVisibleViewArea.width()),
                 0));


mdvacca:
Importing to review

fabriziobertoglio1987:
Thanks a lot David Vacca :pray:

andres-torres-marroquin:
This fix breaks the `keyboardDidHide` event in Android by sending `screenY` a value that is not correct, it basically crops the view at about 90% of the original height when the keyboard is hidden. I've tested 0.63.3 and 0.63.2, and the 0.63.2 works as intended. I'm using `react-native-navigation` with TopBar and BottomTabs, if that helps.

I just want to be sure if this change is actually correct?

fabriziobertoglio1987:
@andres-torres-marroquin I'll check into it this weekend, could you open and issue in react-native and reference this pr.
The issues needs to have a minimum reproducible example. Thanks a lot :pray:

Issue #10471, opened 2016-10-20
iOS: UI will be blocked when show Alert while closing Modal


nihgwu:
When show Alert while closing Modal, the Alert dialog will disappear and the Modal will block the UI entirely even after reload, only on iOS.

``` js
'use strict';

import React, { Component } from 'react';

import {
  StyleSheet,
  View,
  Text,
  Modal,
  Alert,
} from 'react-native';

class demo extends Component {
  state = {
    showModal: false,
  }

  onShowModal = () => {
    this.setState({ showModal: true });
  }
  onCloseModal1 = () => {
    this.setState({ showModal: false }, () => {
      Alert.alert('Alert', 'UI will be blocked by the modal');
    });
  }
  onCloseModal2 = () => {
    this.setState({ showModal: false }, () => {
      setTimeout(() => {
        Alert.alert('Alert', 'Alert won\'t show');
      }, 200);
    });
  }
  onCloseModal3 = () => {
    this.setState({ showModal: false }, () => {
      setTimeout(() => {
        Alert.alert('Alert', 'Works fine');
      }, 510);
    });
  }
  render() {
    const { showModal } = this.state;
    return (
      <View style={styles.container}>
        <Text onPress={this.onShowModal}>Show modal</Text>
        <Modal animationType='slide' visible={showModal} onRequestClose={this.onCloseModal3} >
          <View style={styles.container}>
            <Text onPress={this.onCloseModal1}>Close modal immediately</Text>
            <Text onPress={this.onCloseModal2}>Close modal after 200ms</Text>
            <Text onPress={this.onCloseModal3}>Close modal after more then 500ms</Text>
          </View>
        </Modal>
      </View>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'space-around',
  },
});


export default demo;
```

leeight:
+1

alma-socar:
`InteractionManager.runAfterInteractions` also doesn't work in this case

ganmor:
The opposite is also true, showing an alert in the same loop just before showing a modal will prevent the modal from being displayed

nihgwu:
Yes, I guess they are the some issue, so I just make it simple to demonstrate

? 2016?10?20??20:07?Morgan Laupies <notifications@github.com<mailto:notifications@github.com>> ???


The opposite is also true, showing an alert in the same loop just before showing a modal will prevent the modal from being displayed

-

[...]

PR #29295, opened 2020-07-08
iOS: Update RCTAlertManager to use new RCTAlertController

### LINKED ISSUE #29082: iOS: UI will be blocked when show Alert while closing Modal
Issue #10471 was locked and closed but is still an issue. Only way around it is to use a timeout which is a total hack.

devon94:
## Summary

This should fix #29082 and #10471

Currently when an alert is being shown while a modal is being dismissed, it causes the alert not to show and In some cases it causes the UI to become unresponsive. I think this was caused by using RCTPresentedViewController to try and display the Alert the currently presented view. The View the Alert was going to be shown on is dismissed and the modal doesn't show. I implemented a new RCTAlertController to show the alert on top of the view, the modal being dismissed should now not interfere with the alert being shown.

## Changelog

[iOS] [Fixed] - Fixed showing Alert while closing a Modal

## Test Plan

To recreate the bug:

1. npx react-native init Test --version 0.63.0-rc.1
2. Paste the following code into App.js

```javascript
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 * @flow strict-local
 */

import React from 'react';
import {
  SafeAreaView,
  StyleSheet,
  View,
  Text,
  StatusBar,
  Modal,
  Alert
} from 'react-native';

const App: () => React$Node = () => {
  const [visible, setVisible] = React.useState(false)

  const onShowModal = () => {
    setVisible(true)
  }

  onCloseBroken = () => {
    setVisible(false)
    Alert.alert('Alert', 'Alert won\'t show')
  }

  onCloseWorking = () => {
    setVisible(false)
    setTimeout(() => Alert.alert('Alert', 'Works fine'), 10)
  }

  return (
    <>
      <StatusBar barStyle="dark-content" />
      <SafeAreaView style={styles.container}>
        <Text onPress={onShowModal}>Show modal</Text>
      </SafeAreaView>
      <Modal animationType="fade" visible={visible} onRequestClose={onCloseWorking} >
          <Text onPress={onCloseBroken}>Close modal immediately</Text>
          <Text onPress={onCloseWorking}>Close modal with delay</Text>
        </View>
      </Modal>
    </>
  )
}

  container: {
    flex: 1,
    alignItems: 'center',
    justifyContent: 'space-around',
  },
})

export default App

```

3. cd Test && npx react-native run-ios
4. Show the modal and click the `Close modal immediately` button

The first button doesn't show the alert, the second does because it gets rendered after the modal view is dismissed. After this commit, the alert always shows on top of every view properly. You can test by pointing the react native package to my branch by modifying the package json file like this

```
    "react-native": "https://github.com/devon94/react-native.git#fix-ios-modal"
```

I was unable to reproduce the case where it causes the UI to be responsive in the test app but was able to reproduce it in our react native app at work. I can provide a video later if needed but the code is too complex to simplify into a test case.

### DIFF
--- a/React/Views/RCTAlertController.m
+++ b/React/Views/RCTAlertController.m
@@ -1,0 +1,25 @@
+#import "RCTAlertController.h"
+
+@interface RCTAlertController ()
+
+@property (nonatomic, strong) UIWindow *alertWindow;
+
+@end
+
+@implementation RCTAlertController
+
+- (UIWindow *)alertWindow {
+    if (_alertWindow == nil) {
+        _alertWindow = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+        _alertWindow.rootViewController = [UIViewController new];
+        _alertWindow.windowLevel = UIWindowLevelAlert + 1;
+    }
+    return _alertWindow;
+}
+
+- (void)show:(BOOL)animated completion:(void (^)(void))completion {
+    [self.alertWindow makeKeyAndVisible];
+    [self.alertWindow.rootViewController presentViewController:self animated:animated completion:completion];
+}
+
+@end
--- a/React/Views/RCTAlertController.m
+++ b/React/Views/RCTAlertController.m
@@ -10,7 +10,7 @@
 
 - (UIWindow *)alertWindow {
     if (_alertWindow == nil) {
-        _alertWindow = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+        _alertWindow = [[UIWindow alloc] initWithFrame:[[UIApplication sharedApplication] keyWindow].bounds];
         _alertWindow.rootViewController = [UIViewController new];
         _alertWindow.windowLevel = UIWindowLevelAlert + 1;
     }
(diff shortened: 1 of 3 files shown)


PeteTheHeat:
Nice fix!

React/Views/RCTAlertController.m
```diff
@@ -0,0 +1,25 @@
...
+
+@property (nonatomic, strong) UIWindow *alertWindow;
+
+@end
+
+@implementation RCTAlertController
+
+- (UIWindow *)alertWindow {
+    if (_alertWindow == nil) {
+        _alertWindow = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
```
PeteTheHeat, line 13: @devon94  One question: Should this be the key window size? 

The scenario I'm thinking about is iPad split screen. I think this would return the entire device screen size instead of the app's window.

devon94: Ah yea, you're right! I didn't think about iPad/split screen when looking at this issue. The keyWindow bounds should fix that scenario. I can add that later today

PR #29307, opened 2020-07-08
Fix multiline TextInput crash when inserting/removing lots of text

tido64:
## Summary

Multiline `TextInput` can crash when really long texts are inserted and removed quickly. This is caused by the fact that [`-[NSAttributedString string]`](https://developer.apple.com/documentation/foundation/nsattributedstring/1412616-string?language=objc) doesn't really return a copy, and may mutate the string while it is being used by `convertIdToFollyDynamic`. See https://github.com/microsoft/react-native-macos/pull/489#discussion_r451789471 for more details on the issue.

This issue was originally reported in https://github.com/microsoft/react-native-macos/issues/486 and was fixed in https://github.com/microsoft/react-native-macos/pull/489.

## Changelog

[iOS] [Fixed] - Fix multiline TextInput crash when inserting/removing lots of text

## Test Plan

1. Open RNTester > TextInput
2. Search for a multiline example
3. Copy some large text and paste it into the text input view
4. Remove some (or all) text
5. Repeat steps 3-4

### DIFF
--- a/Libraries/Text/TextInput/RCTBaseTextInputView.m
+++ b/Libraries/Text/TextInput/RCTBaseTextInputView.m
@@ -333,7 +333,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)initWithFrame:(CGRect)frame)
 
   [_eventDispatcher sendTextEventWithType:RCTTextEventTypeFocus
                                  reactTag:self.reactTag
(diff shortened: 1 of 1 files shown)


shergin:
For the reference: https://developer.apple.com/documentation/foundation/nsattributedstring/1412616-string?language=objc

PR #29039, opened 2020-06-03
Fix to make taps on views outside parent bounds work on Android

hsource:
## Summary

By default, Views in React Native have `overflow: visible`. When a child view is outside of the parent view's boundaries, it's visible on Android, but not tappable. This behaviour is incorrect, and doesn't match iOS behaviour.

- Taps on Views outside the bounds of a parent with `overflow: visible` (or unset) should register
- Taps on Views outside the bounds of a parent with `overflow: hidden` should continue to not register

Related issues:

- fixes #21455
- fixes #27061
- fixes #27232

### Fix

- Made `findTouchTargetView` not check that the touch was in the bounds of the immediate children, but instead
  - Check that the touch is in its own bounds when returning itself
  - Check that the touch for a child is in its own bounds only when `overflow: hidden` is set
- Modified related code to adjust to this change
- Added RNTesterApp test

## Changelog

[Android] [Fixed] - Allow taps on views outside the bounds of a parent with `overflow: hidden`

## Test Plan

This can be tested with 2 examples added to the bottom of the PointerEvents page of the RNTesterApp:

| Before | After |
| --- | --- |
| Before | After |

PR #26374, opened 2019-09-08
Allow touches on overflowed children when `overflow: visible` (Android)

danielgindi:
## Summary

Currently `overflow: visible` allows showing child views that are outside the parent's bounds (well, `overflow: visible` on the grandparent allows that, not the parent).
But those non-clipped children do not receive touches.

This is an attempt to solve it.

## Changelog

[Android] [Fixed] - `overflow: visible` now allows interaction with children that are rendered outside their parents' bounds.

## Test Plan

* Added `overflow: visible`, tapped on a button.
* Added a `MapView`, tested scrolling the map while overflowed elements are visible over it, and tapping on those elements.

PR #29393, opened 2020-07-16
Added examples to test tapping on overflowed Views to RNTesterApp

hsource:
## Summary

By default, Views in React Native have overflow: visible. When a child view is outside of the parent view's boundaries, it's visible on Android, but not tappable. This behaviour is incorrect, and doesn't match iOS behaviour.

#26374 and #29039 both fix this. @shergin [suggested](https://github.com/facebook/react-native/pull/26374#issuecomment-659108411):

> Can we start from decoupling the example from this diff for quick landing? We think we solved this issue in Fabric on iOS and Android, and the test will be very helpful to verify that.

This change does that.

## Changelog

[Internal] [Added] - Added examples to test tapping on overflowed Views to RNTesterApp

## Test Plan

Added tests to RNTesterApp. Here are the current results of testing on Android:

| master | With #29039 | With #26374 |
| --- | --- | --- |
|  |  |  |

### DIFF
--- a/RNTester/js/examples/PointerEvents/PointerEventsExample.js
+++ b/RNTester/js/examples/PointerEvents/PointerEventsExample.js
@@ -14,12 +14,25 @@ const React = require('react');
 
 const {StyleSheet, Text, View} = require('react-native');
 
-class ExampleBox extends React.Component<$FlowFixMeProps, $FlowFixMeState> {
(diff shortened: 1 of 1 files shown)


mrousavy:
@shergin does this work on Fabric?

PR #29395, opened 2020-07-16
Adding delayPressIn prop to Pressable

padapada09:
## Summary

Pressable components come with a default "delayPressIn" and while we can argue about what it's default value should be, we need to be able to configure it. So I added the prop to do so.

## Changelog

The change is small, I only added the prop to the type Prop, and use it when passing it to the Pressabillity config object.
I did not alter any native code, so there should not be a problem with that. And finally, I did not set a default value cause I figured that decision is not mine to take.

### DIFF
--- a/Libraries/Components/Pressable/Pressable.js
+++ b/Libraries/Components/Pressable/Pressable.js
@@ -65,6 +65,11 @@ type Props = $ReadOnly<{|
   delayLongPress?: ?number,
 
   /**
+   * Duration to wait after press down before calling `onPressIn`.
+   */
+  delayPressIn?: ?number,
+
+  /**
    * Whether the press behavior is disabled.
    */
   disabled?: ?boolean,
@@ -143,6 +148,7 @@ function Pressable(props: Props, forwardedRef): React.Node {
     android_ripple,
     children,
     delayLongPress,
+    delayPressIn,
     disabled,
     focusable,
     onLongPress,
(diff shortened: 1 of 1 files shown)


mrousavy:
I'm still curious on why the default delay is anything longer than 0ms. That doesn't make any sense at all, the visual feedback should be instant.

radiosilence:
Why on earth would you want to add a delay? This just makes the app feel laggy and sluggish, and is also bad if you are creating something like a remote control app where you want to do an action as soon as the user has touched stuff.

JoshuaGross:
It looks like this is being passed into Pressable.js but it looks like it's not actually doing anything, is that correct? Is there something I'm missing?

likern:
> It looks like this is being passed into Pressable.js but it looks like it's not actually doing anything, is that correct? Is there something I'm missing?

It works by passing config to `usePressability > Pressability`. **Pressability** supports that config option.

babylone-star:
> I'm still curious on why the default delay is anything longer than 0ms. That doesn't make any sense at all, the visual feedback should be instant.

The reason is because otherwise the scroll responder competes with the button press responder for attention. Without a delay, while scrolling, if you start your scroll on top of a button  then the button will run its press animation while you're scrolling, which is a UI glitch.

In contrast, if you put a short delay on the button press, this allows scroll responder to 'lock focus' so to speak and ignore any button presses while the user is scrolling.

yungsters:

PR #29403, opened 2020-07-17
Add an explicit NDK version to Android template

safaiyeh:
## Summary
Added an explicit NDK version to the Android template. This allows tooling to detect which NDK version to install automatically.

## Changelog

[Android] [Added] - Add an explicit NDK version to Android template

## Test Plan

Template builds as it should. NDK version gets installed with initial Android set up.

### DIFF
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -121,6 +121,8 @@ def jscFlavor = 'org.webkit:android-jsc:+'
 def enableHermes = project.ext.react.get("enableHermes", false);
 
 android {
+    ndkVersion "21.3.6528147"
+
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     compileOptions {
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -121,7 +121,7 @@ def jscFlavor = 'org.webkit:android-jsc:+'
 def enableHermes = project.ext.react.get("enableHermes", false);
 
 android {
-    ndkVersion "21.3.6528147"
+    ndkVersion "20.1.5948944"
 
     compileSdkVersion rootProject.ext.compileSdkVersion
 
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -121,7 +121,7 @@ def jscFlavor = 'org.webkit:android-jsc:+'
 def enableHermes = project.ext.react.get("enableHermes", false);
 
 android {
-    ndkVersion "20.1.5948944"
+    ndkVersion rootProject.ext.ndkVersion
 
     compileSdkVersion rootProject.ext.compileSdkVersion
 
--- a/template/android/build.gradle
+++ b/template/android/build.gradle
@@ -6,6 +6,7 @@ buildscript {
         minSdkVersion = 16
         compileSdkVersion = 29
         targetSdkVersion = 29
+        ndkVersion = "20.1.5948944"
     }
     repositories {
         google()


thymikee:
AFAIK the NDK is not necessary for regular users who don't build from source. Can you make sure this doesn't affect the users i.e. requiring them to install NDK?

Krizzu:
cc @dulmandakh

safaiyeh:
@thymikee I was getting errors with 0.63 as so:

```
android/app/build/intermediates/merged_native_libs/debug/out/lib/x86_64/libreactnativejni.so' due to missing strip tool for ABI 'X86_64'. Packaging it as is.
```

The fix was to install the NDK

safaiyeh:
Status on this? @dulmandakh

ashishfeels:
@safaiyeh  i am trying to build app bundle with symbols and facing an issue related to NDK though i have ndk installed it keeps saying me its not installed
0.63.2
```
> Task :app:extractReleaseNativeSymbolTables FAILED

* What went wrong:
Execution failed for task ':app:extractReleaseNativeSymbolTables'.
> NDK is not installed

```
Probably an option to set NDK would be nice

dulmandakh (approved):
@fkgozali @mdvacca I couldn't build my app with RN 0.63 due to that it requires NDK. I have no idea when that requirement was introduced, but it's best to have NDK version in the template that works with RN. Currently, we use NDK 20 and it works as expected. So let's merge this PR.

ashishfeels:
@dulmandakh for the aab to build with symbol we need to specify NDK 20, i had NDK 21 installed and it gave me the above error i mentioned please have a look.

template/android/app/build.gradle
```diff
@@ -121,6 +121,8 @@ def jscFlavor = 'org.webkit:android-jsc:+'
 def enableHermes = project.ext.react.get("enableHermes", false);
 
 android {
+    ndkVersion "20.1.5948944"
```
fkgozali, line 124: can this be put in the same place as `rootProject.ext.compileSdkVersion`? That way, all versions are put in 1 place?

safaiyeh: Sounds good I'll update this in the morning!

safaiyeh:
@fkgozali updated!

fkgozali:
Hmm, thinking about this, is this approach generally correct? According to https://github.com/facebook/react-native/wiki/Building-from-source, some environment sets up `ANDROID_NDK` env var to point to specific NDK version. That environment may not have the exact NDK version set in this PR (20.1.5948944).

* What happens if ANDROID_NDK and the version specified here mismatch? Which has the priority?
* Should we guide people to instead set ANDROID_NDK var per https://github.com/facebook/react-native/wiki/Building-from-source ?

gedeagas:
@fkgozali After careful thoughts here is my opinion. 

**What happens if ANDROID_NDK and the version specified here mismatch? Which has the priority?**

I just tried to make a quick experiment about this. 
Screenshot from 2020-09-20 15-36-45

Currently, I have two versions of NDK installed. The first one is NDK **21.3.6528147**, this is the one I specified on my env var and the second one is NDK **20.1.5948944**. When I tried to run RNTester and ReactAndroid with the explicit NDK version (**20.1.5948944**) Gradle will prioritize the explicit NDK version that we set on the Gradle file not the one on the env var. 

And when I set the version to **20.0.5594570** which is not installed locally on my machine the build would not be successful but Gradle can list all of my locally installed NDK version regardless of what I specified on the env var.

Screenshot from 2020-09-19 21-13-17

**Should we guide people to instead set ANDROID_NDK var per https://github.com/facebook/react-native/wiki/Building-from-source ?** 

The guide on installing Android NDK is quite outdated. I much prefer to install NDK from Android Studio's SDK Manager. Why? For example, no Mac NDK before NDK r21 was signed or notarized. This shouldn't matter for NDK downloads performed by Android Studio's SDK Manager, but if you manually download an NDK from https://developer.android.com/ndk/downloads on macOS 10.15 you'll see warnings when you try to run the extracted tools. The NDK is not quarantined when it is downloaded via the SDK manager, so is currently allowed by Gatekeeper. For me, I think **The SDK manager is currently the most reliable way to get the NDK** and I think this is should be the way we advise people to install NDK. 

**Additionally, what happens if React Native core updates its NDK version, but the hosting app (assume one creates an app with an older release from template) still has it pinned to an older NDK version? Would that potentially cause issue due to ABI, etc? In this case, it seems like using a global ANDROID_NDK might be more consistent?**

I just tried this, and even tho I only install android NDK version **21.3.6528147** and set that up on my env var the app won't build because somehow it's searching for non-existent NDK version **20.0.5594570**. This problem can be easily fixed if we just configure a specific NDK version on the Gradle file. 

Screenshot from 2020-09-20 15-58-01

fkgozali:
The case I wasn't sure about is :ReactAndroid, where we build the C++ by calling into `ndk-build` directly. In order to find the right NDK dir, it does this: https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L249-L271

```
def findNdkBuildFullPath() {
    // we allow to provide full path to ndk-build tool
    if (hasProperty("ndk.command")) {
        return property("ndk.command")
    }
    // or just a path to the containing directory
    if (hasProperty("ndk.path")) {
        def ndkDir = property("ndk.path")
        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
    }

    if (System.getenv("ANDROID_NDK") != null) {
        def ndkDir = System.getenv("ANDROID_NDK")
        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
    }

    def ndkDir = android.ndkDirectory ? android.ndkDirectory.absolutePath : null

    if (ndkDir) {
        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
    }
    return null
}
```

Now, I'm not sure if this logic has covered the case where `ndkVersion` is set in the root project build.gradle.

> I just tried this, and even tho I only install android NDK version 21.3.6528147 and set that up on my env var the app won't build because somehow it's searching for non-existent NDK version 20.0.5594570. This problem can be easily fixed if we just configure a specific NDK version on the Gradle file.

Yeah this is true, however, the issue is not just about missing NDK installation, but also about cross-version compatibility (C++ ABI etc). For example, perhaps the different NDK versions are already installed, but what if NDK version A output is not compatible with version B output? It may cause compilation error or runtime crash.

So it feels to me that if an app wants to build from source, the same NDK version should be used to compile **all parts** of the sources to ensure compatibility.

Thoughts? cc @safaiyeh @dulmandakh @Kudo

safaiyeh:
Is it possible to expose the NDK version used by ABI to the user land template?

fkgozali:
> Is it possible to expose the NDK version used by ABI to the user land template?

Did you mean which NDK is used by the system? I think that's set by `ANDROID_NDK` path, which may or may not have the version on it. E.g. many system just symlink `/opt/android_ndk` to something in local machine.

gedeagas:
> 
> Now, I'm not sure if this logic has covered the case where `ndkVersion` is set in the root project build.gradle.
> 

Just checked, and yeah the logic doesn't cover when Gradle execute buildReactNdkLib it will use the one that people set on `ANDROID_NDK`. I only checked stripDebugSymbol yesterday.

Kudo:
Share my understandings:

1.
```
android/app/build/intermediates/merged_native_libs/debug/out/lib/x86_64/libc++_shared.so' due to missing strip tool for ABI 'X86_64'. Packaging it as is.
android/app/build/intermediates/merged_native_libs/debug/out/lib/x86_64/libreactnativejni.so' due to missing strip tool for ABI 'X86_64'. Packaging it as is.
```
These are warnings but not errors.
If there are *.so, gradle will try to use NDK `strip` tool to strip debug symbol.
If NDK is not setup correctly, gradle will give these warnings and skip the stripping process.
However, the solution @safaiyeh here is better that will prevent developers install unexpected NDK.

2.

> Yeah this is true, however, the issue is not just about missing NDK installation, but also about cross-version compatibility (C++ ABI etc). For example, perhaps the different NDK versions are already installed, but what if NDK version A output is not compatible with version B output? It may cause compilation error or runtime crash.

AFAIK, NDK maintains ABI stability really well.
The old JSC was built from [NDK r10c](https://github.com/facebookarchive/android-jsc/blob/master/.buckconfig#L4) and we ran for years.
Until we replacing with new JSC, at that time, RN was built from [NDK r17c](https://github.com/facebook/react-native/blob/0.58-stable/.circleci/config.yml#L341)
That was really amazing that we didn't face ABI incompatibilities from NDK r10c to r17c.
 
Well, relying on NDK ABI stabilities is not a good practice. Building from source will prevent potential problems.

safaiyeh:
The original goal of this is coming from a new RN Developer perspective.

Download the template -> Know the correct NDK to install

fkgozali:
OK, thanks for the context. Based on NDK ABI stability and easier starting point when using the template, then this change seems reasonable to try out. For whether ReactAndroid and RNTester should explicitly define the version as well, let's discuss more in https://github.com/facebook/react-native/pull/29987.

PR #29429, opened 2020-07-19
[Settings] Correctly update settings after key removal

koenpunt:
## Summary

Before the code didn't account for removal of keys, so if a key was removed from the user defaults, it would stay in the settings object in the javascript realm.

By replacing the full settings object in the javascript realm, the data stays in sync.

## Changelog

[iOS] [Fixed] - Correctly update settings after user defaults key removal

## Test Plan

I've implemented the changes in my project locally, and it works. Not sure how you'd like to see this tested here.


koenpunt:
Ping ✨

Libraries/Settings/Settings.ios.js
```diff
@@ -55,19 +55,33 @@ const Settings = {
...
-          if (sub.keys.indexOf(key) !== -1 && sub.callback) {
-            sub.callback();
-          }
-        });
+        changedKeys.add(key);
+      }
+    });
+
+    Object.keys(this._settings).forEach(key => {
+      if (typeof body[key] === 'undefined') {
```
sammy-SC, line 70: Isn't `body.hasOwnProperty(key)` more appropriate here?

Issue #18395, opened 2018-03-15
ReactEdit Crash NullPointerException


yuanboGeng:
in React-Native 0.45.1

com.facebook.react.views.textinput.ReactEditText.onTouchEvent
ReactEditText.java - line 163
java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable$ConstantState.newDrawable(android.content.res.Resources)' on a null object reference
Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable$ConstantState.newDrawable(android.content.res.Resources)' on a null object reference
1
android.graphics.drawable.DrawableContainer$DrawableContainerState.getChild DrawableContainer.java:884
2
android.graphics.drawable.DrawableContainer.selectDrawable DrawableContainer.java:466
3
android.graphics.drawable.StateListDrawable.onStateChange StateListDrawable.java:104
4
android.graphics.drawable.Drawable.setState Drawable.java:736
5
android.graphics.drawable.DrawableWrapper.onStateChange DrawableWrapper.java:331
6
android.graphics.drawable.Drawable.setState Drawable.java:736
7
android.view.View.drawableStateChanged View.java:19223
8
android.widget.TextView.drawableStateChanged TextView.java:4673
9
android.view.View.refreshDrawableState View.java:19292
10
android.view.View.setPressed View.java:9135
11
android.view.View.setPressed View.java:9113
12
android.view.View.onTouchEvent View.java:12357
13
android.widget.TextView.onTouchEvent TextView.java:10095
14
com.facebook.react.views.textinput.ReactEditText.onTouchEvent ReactEditText.java:163

hramos:
Not only is this report for a significantly old version of React Native, it lacks sufficient information to be actionable.

Issue #9979, opened 2016-09-19
ReactSlider crash on Android release version


ssg-luke:
### Issue Description

Hard crash randomly and unexpectedly on Android release version only (debug is fine, iOS is fine on both debug and release) when hiding and displaying sliders based on a state variable.
### Steps to Reproduce / Code Snippets

We have a number of photo filters in our app which are selected by `<TouchableOpacity>` components and some have a slider to adjust the blur or brightness etc. These sliders are completely separate from each other and are displayed based on a string stored in the components state called `filter`.

Here is one such slider from render():

`{filter === 'gamify' && (
           <View style={styles.filterSlider}>
              <Text>How retro do you want to go?</Text>
              <Slider value={this.state.pixelSize}
                      minimumValue={1}
                      maximumValue={5}
                      onValueChange={value => this.setState({pixelSize: value})} />
            </View>)}`
#### Expected Results

No crash when hiding and displaying sliders
### Stack Trace

java.lang.NullPointerException: Attempt to write to field 'java.util.ArrayList android.animation.AnimatorSet$Node.nodeDependents' on a null object reference

http://pastebin.com/270Wg6Pu
### Additional Information
- React Native version: 0.33
- Platform(s) (iOS, Android, or both?): Android (Release only)
- Operating System (macOS, Linux, or Windows?): Mac OSX

rusel1989:
Still crashes once in while in 0.40.0

timxyz:
I've been getting the same crash in AnimatorSet with a button created in a custom Android ViewManager, since at least 0.33. The crash is very rare relative to the number of times the view manager in question creates buttons, and happens in both release and debug on Android only. 

The line in question is:

`Button button = new Button(context);`

(There is also an identical line in the shadow node which causes identical issues)

I can pretty much guarantee the crash occurring by wrapping the line above in a loop. I didn't look into it deeply, but the nature of the problem and analysis of the code in AnimatorSet suggested it might be a threading issue. Sure enough, wrapping the line of code above in a block synchronized on something available to both the ViewManager and ShadowNode (e.g. the React context) eradicates the issue, although I assume the crash will still happen if code I don't have control over also creates / inflates Buttons concurrently with mine. The line has to be synchronized in *both* the ViewManager and the ShadowNode.

Hopefully that narrows down the issue with the Slider and can help identify the root cause.

jasongrishkoff:
+1 on this issue with ReactSlider on 0.40.0.  Log below.

```
[...]

PR #29452, opened 2020-07-21
Fix - TextInput Drawable to avoid Null Pointer Exception RuntimeError #17530  

fabriziobertoglio1987:
## Summary

This issue fixes https://github.com/facebook/react-native/issues/17530 fixes https://github.com/expo/expo/issues/9905 with the help of @sunnylqm https://github.com/sunnylqm

Re-rendering a large number of TextInputs with key prop on the screen will trigger the below Null Pointer Exception Runtime Error 

NullPointerException:tempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable$ConstantState.newDrawable(android.content.res.Resources)'

The error is caused by null.newDrawable(mSourceRes) at

https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L919

More info https://github.com/facebook/react-native/pull/29452#issuecomment-662616018 https://github.com/facebook/react-native/issues/17530#issuecomment-662000718

The Theme Theme.AppCompat.Light.NoActionBar defines the Drawables for AppCompatEditText in @drawable/abc_edit_text_material.xml

https://chromium.googlesource.com/android_tools/+/7200281446186c7192cb02f54dc2b38e02d705e5/sdk/extras/android/support/v7/appcompat/res/drawable/abc_edit_text_material.xml

Removing the following line from the above xml file @drawable/abc_edit_text_material.xml fixes the error https://github.com/facebook/react-native/issues/17530#issuecomment-662000718

`<item android:state_pressed="false" android:state_focused="false" android:drawable="@drawable/abc_textfield_default_mtrl_alpha"/>`

The Theme default EditText background is replaced with a custom background, which is a copy of the original background without the above item which triggers the Runtime Error. The changes are implemented in RNTester with commit (more info in the commit) https://github.com/facebook/react-native/commit/0858d4120df242196570f5394f8ae4e623533493. The new custom drawable used as default background for the TextInput is named edit_text.

`<item name="android:editTextBackground">@drawable/edit_text</item>` 

The same changes have been added to react-native default template for creating new applications with commit (more info) https://github.com/facebook/react-native/commit/f3493083df5e1b157bbdaf18f97c59b7b0ad828c, lean core moved the cli tools to https://github.com/react-native-community/cli, but the default template for creating a new application is stored in facebook/react-native/template.

New applications will be generated with this configurations and will not experience the error, existing react-native applications will fix the error by upgrading with the [upgrade-helper](https://github.com/react-native-community/upgrade-helper).

A Minimum Reproducible Example to reproduce this error is included in commit (more info in the commit) https://github.com/fabriziobertoglio1987/react-native/commit/4a414e2cc7e17068da26ef7727a04d0aa5883a17 and https://github.com/facebook/react-native/issues/17530#issuecomment-660017858

## Changelog

[Android] [Fixed] - TextInput Drawable to avoid Null Pointer Exception RuntimeError #17530  

## Test Plan

Works in all scenarios on Android.

**<details><summary>CLICK TO OPEN TESTS RESULTS - React Native</summary>**
<p>

Test Results from the Testing in RNTester. 

Minimum Reproducible Example added with commit (more info in the commit) https://github.com/fabriziobertoglio1987/react-native/commit/4a414e2cc7e17068da26ef7727a04d0aa5883a17  

The example included in commit https://github.com/fabriziobertoglio1987/react-native/commit/4a414e2cc7e17068da26ef7727a04d0aa5883a17 will cause a [NPE Runtime Error on Master Branch](https://fabriziobertoglio.s3.eu-central-1.amazonaws.com/opensource/react-native/17530/runtime.mp4), while no error is experienced in the [feature branch](https://fabriziobertoglio.s3.eu-central-1.amazonaws.com/opensource/react-native/17530/no_runtime.mp4). The links are video hosted on s3 of this tests (playable by google chrome).

| **[BEFORE](https://fabriziobertoglio.s3.eu-central-1.amazonaws.com/opensource/react-native/17530/runtime.mp4)** | **[AFTER](https://fabriziobertoglio.s3.eu-central-1.amazonaws.com/opensource/react-native/17530/no_runtime.mp4)** | 
|  |  |

The below screenshots were taken to detect any issues with the EditText Background. There is no difference between master and feature branch.

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **BEFORE** | **AFTER** | 
|  |  |

| **AFTER** |
|:-------------------------:|
|  |

</p>
</details>

**<details><summary>CLICK TO OPEN TESTS RESULTS - React Native Cli</summary>**
<p>

As lean core move cli tools to https://github.com/react-native-community/cli, I tested the changes to the template in a separate repository https://github.com/fabriziobertoglio1987/react-native-template and generated the template with the following command

```
npx react-native init ProjectName --template file:///home/fabrizio/Documents/sourcecode/opensource/react-native-template/template
```

The generated app did not experience any issues and includes all the changes in [rn_edit_text_material.xml](https://github.com/fabriziobertoglio1987/react-native-template/blob/master/ProjectName/android/app/src/main/res/drawable/rn_edit_text_material.xml) and [styles.xml](https://github.com/fabriziobertoglio1987/react-native-template/blob/master/ProjectName/android/app/src/main/res/values/styles.xml)

</p>
</details>


sunnylqm:
Thank you @fabriziobertoglio1987  for your awesome work. I really admire your great patience and endeavor on this haunting issue.

The solution looks good to me, but still the question is not answered throughly: What is the real cause. What is wrong with the drawable? 

Why remove `<item android:state_pressed="false" android:state_focused="false" android:drawable="@drawable/abc_textfield_default_mtrl_alpha"/>` could fix? Why so specific with press/focus state? Remember this exception could also be  thrown from a touch event? https://github.com/facebook/react-native/issues/18395

And why removing key props could also prevent the exception? Is it because key props will try to reuse the component and hence cached some dirty state?

I know little to dig it like you did, but it keeps me wondering if there is something wrong with react native that managing the ui state. 
If the issue lies deep in some managing mechanism, I believe this could not be the only case, and unfortunately I found a very similar one that could be highly related https://github.com/facebook/react-native/issues/9979

Anyway, well done!

fabriziobertoglio1987:
Unluckily today I don't have any answer to those questions. I decided to try to answer them anyway with my current knowledge.

This are my findings from an entire day of work on this problem, but I believe you are aware of all this as you already pointed me to the source code in the previous messages.

The Runtime is triggered in [DrawableContainer.getChild()](https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L918) because ConstantState variable [cs](https://github.com/aosp-mirror/platform_frameworks_base/blob/8857e6b645f9b91f44c8ea0ab8f441c9c0d63da6/graphics/java/android/graphics/drawable/DrawableContainer.java#L918) is `null`. Calling `null.newDrawable(mSourceRes)` triggers a Null Pointer Exception. This call is the result of the change of state of the TextInput, in this [stack trace](https://github.com/facebook/react-native/issues/17530#issue-287662728) after pressing it. 

The Background [abc_edit_text_material.xml](https://chromium.googlesource.com/android_tools/+/7200281446186c7192cb02f54dc2b38e02d705e5/sdk/extras/android/support/v7/appcompat/res/drawable/abc_edit_text_material.xml) used for AppCompatEditText defines [StateListDrawable](https://developer.android.com/reference/android/graphics/drawable/StateListDrawable) which will change the background image of the TextInput depending on the state. 

The [lines below](https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L918) are trying to replace the old Background Image with a new one, but the process fails because `mDrawableFutures` does not include the correct list of entries.

```java
final ConstantState cs = mDrawableFutures.valueAt(keyIndex);
final Drawable prepared = prepareDrawable(cs.newDrawable(mSourceRes));
```

I am still trying to understand why `mDrawableFutures` does not include the full list of items. I believe the key to understand this is:

1) Finding where `mDrawableFutures` is initially set
2) Finding where `mDrawableFutures` is updated
3) Debug the Android Source

I try to answer your questions with my current knowledge of the problem, sorry if they are not satisfing answers. Tomorrow I will keep working on this and try to find a better solution.

> What is the real cause. 

The value of `mDrawableFutures`, how it is initizialized and then updated. I believe this is not the answer you are looking for, but it is an important starting point. Tomorrow I will investigate the role of react-native in setting and updating `mDrawableFutures`.

> What is wrong with the drawable?

I believe removing this Drawable is just a workaround to solve the real issue. The real issue is how `mDrawableFutures` is set/updated. If this issue is experienced in many different use cases, removing the Drawable may not be a good solution.

> Why remove `<item android:state_pressed="false" android:state_focused="false" android:drawable="@drawable/abc_textfield_default_mtrl_alpha"/>` could fix?
>And why removing key props could also prevent the exception? Is it because key props will try to reuse the component and hence cached some dirty state?


Maybe because a `TextInput` with key props is the only condition to trigger `state_pressed=false` and `state_focused=false`, while without a `key` prop those conditions are not triggered.

Tomorrow I will keep working on this and try to understand this issue. I will be working on this issue as long as you don't decide to assign me to another issue. 

I would be happy to receive your advise and sorry if this answer is incomplete. I did not tag you for this reason, but I wanted to write an update.
Thanks a lot  
Best Regards  
Fabrizio Bertoglio

sunnylqm:
@fabriziobertoglio1987 I'd like to hire you if I am a boss 👍  

Anyway do not let this issue take too much of your time. You do not have to understand it 100% to make it a ready pr and tag some core members to review. If this change prevent the crash without regressions it's already a great valuable contribution to the community.

fabriziobertoglio1987:
Thanks a lot @sunnylqm . I don't think that I could have done this Pull Request alone. I believe the interaction was very important to find a solution, the issue was really challenging and I am very happy that we could solve it together. 

I wish you a good evening. 
Fabrizio

Skullcan:
@fabriziobertoglio1987 any chance I can apply this fix in an active project w/o waiting for this PR to be merged?

fabriziobertoglio1987:
@Skullcan yes, you can. check the files https://github.com/facebook/react-native/pull/29452/files and apply the same changes to you `/android` folder. :smiley:

hieuhlc:
Hi guys, can we merge this PR now?

nericode:
When will this be ready PR?

whoami-shubham:
which react-native version would be containing the fix of this issue?

Uzunoff:
> which react-native version would be containing the fix of this issue?

Hello, I believe it will be after 0.63.3 version, because there are people with this version experiencing the same issue.

sunnylqm:
Steps to land a fix:

1. reviewed and approved by a core member
2. tested and merged to master branch
3. wait for a new version cut or cherry pick

And now we are at stage 0

@Uzunoff  @whoami-shubham

mikejsdev:
Are we any closer to releasing this?

dan-developer:
Will PR not be accepted?

fabriziobertoglio1987:
I think that I have to add test coverage to all my react-native prs.. but I don't know if writing them in JavaScript or Java... 
I'll try to add test coverage and hopefully that will allow us to get the prs merged.. :pray:

dan-developer:
I manually implemented your code and it really worked. I tested it to the extreme.

@fabriziobertoglio1987  send me your paypal account. My e-mail: and.dan19@gmail.com

fabriziobertoglio1987:
thanks a lot @dan-developer I really appreciate the feedback.. my email is fabrizio.developer@gmail.com .. please feel free to contact me if you need help.. I'll be again available in March 2021 :peace_symbol: :pray:

SergeyKo69:
How to fix it in Expo ?

jsamr:
@fabriziobertoglio1987 You should setup a Github Sponsor button on your profile and I'll start backing you. The amount of fixes you are willing to give to React Native codebase freely is really worth recognition by the community.

fabriziobertoglio1987:
thanks a lot @jsamr and @dan-developer I implemented the Github Sponsor functionalities and you can check different types of support at https://github.com/sponsors/fabriziobertoglio1987
I'll be contributing to react-native full-time for the upcoming 2-6 months and I believe I will be able to make some valuable contributions. Thanks a lot. Fabrizio

iqbalfaisal:
> @Skullcan yes, you can. check the files https://github.com/facebook/react-native/pull/29452/files and apply the same changes to you `/android` folder. 😃

@fabriziobertoglio1987 where TextInputKeyProp.js should be create?

fabriziobertoglio1987:
>Are we any closer to releasing this?

To answer the question of when this fix will be released in the next version of React Native, I suggest you to follow the discussions at https://github.com/react-native-community/releases/issues by subscribing to those threads. The process of releasing a new version with a bug fix is the following:

- an Issue is opened in facebook/react-native
- a contributor writes a Pull Request to fix/close the react-native Issue
- once the Pull Request is merged, the issue is closed. The review and merge is performed by the Facebook Team that works on the Facebook Ads App running React Native from source. Currently we have 351 open pull requests so the review and merge may take some time.
- The fix is tested in Facebook Ads App running React Native from master for several months
- For each realease (the upcoming one is 0.64.1) there will be a separate discussion on https://github.com/react-native-community/releases/issues to discuss
In this discussion everyone can request a commit to be cherrypicked from master. The link need to be from the a **commit** from master (for ex. https://github.com/facebook/react-native/pull/29157#issuecomment-651430945) and not a Pull Request not yet merged in master :peace_symbol: :pray:

template/android/app/src/main/res/values/styles.xml
```diff
@@ -4,6 +4,7 @@
     <style name="AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
```
guess, line 4: To keep the same styling, maybe consider making `parent="Theme.ReactNative.AppCompat.Light"` to match what `AndroidManifest.xml` was?

ddmc87:
@fabriziobertoglio1987 thanks so much for your PR. In my error logs, I see that some users have this exact problem. I would like to release an update, but I do not have this file: abc_edit_text_material.xml - I only have this file within many plugins. 

My app is based on Expo, but ejected a long time ago. Do you know why I do not have this file? Is it enough / recommended changing this file within my plugins instead?

Thanks in advance!

fabriziobertoglio1987:
@ddmc87 This files are part of an Android Dependency called androidx https://github.com/androidx/androidx/blob/androidx-main/appcompat/appcompat/src/main/res/drawable/abc_edit_text_material.xml which we compile with gradle/buck. 

When you build react-native from source I will have them in `buck-out/bin/ReactAndroid/src/main/third-party/android/androidx/__unpack_appcompat-binary#aar_unzip__/res/drawable/abc_edit_text_material.xml`

You can not patch this files in node_modules, because buck generates a bin. To use the updated android binary you need to:
1) build it
2) point ReactNative to that binary by updating gradle.properties

see https://github.com/facebook/react-native/wiki/Building-from-source#publish-your-own-version-of-react-native

arjith (approved):

Kida007:
Thanks for the PR . will this be cherrypicked for 0.65 ?? we are having production crashes . @fabriziobertoglio1987 so removing item tag will patch it right ?

stains-johnson (approved):

knro:
Still suffering from this issue on 0.64.2, any chance this would get merged for 0.65.0 ?

sunnylqm:
Hi @lunaleaps I see you are looking into this. Do you guys from the core team have any insights that it could be related to https://github.com/callstack/react-native-slider/issues/9#issuecomment-783414324 ?

yungsters:
@fabriziobertoglio1987, do you know why this bug requires "re-rendering a large number of TextInputs with key prop" to trigger the exception?

> Do you guys from the core team have any insights that it could be related to [callstack/react-native-slider#9 (comment)](https://github.com/callstack/react-native-slider/issues/9#issuecomment-783414324) ?

On first blush, the bug fixed by this PR does not seem related to that issue.

sunnylqm:
@yungsters It does not have to require "a large number of TextInputs". It's about probabilistic. So the crash can happen even with only one TextInput. So it's like a rare race condition, in certain unknown scenario the drawables was modified unexpectedly, which makes me thinking there should be similar crashes with other views that have multiple state drawables.

fabriziobertoglio1987:
Thanks a lot @yungsters and @sunnylqm 

Adding TextInputs with a key prop to the page may cause siblings to loose focus.

The runtime error originates in `createInternalEditText`, which adds a new TextInput to the screen. `EditText` uses the context and the `Theme.AppCompat.DayNight.NoActionBar` to generate the TextInput. 

https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L264-L266

<image src="https://user-images.githubusercontent.com/24992535/131800364-18476534-01c2-42c9-b183-48f433740c29.png" width="250" />

Appending a certain number of TextInputs with key prop to the screen triggers a TextInput instance change in focus.

https://chromium.googlesource.com/android_tools/+/7200281446186c7192cb02f54dc2b38e02d705e5/sdk/extras/android/support/v7/appcompat/res/drawable/abc_edit_text_material.xml

The background drawable for `state_pressed=false` `state_focused=false` is `abc_textfield_default_mtrl_alpha` https://github.com/facebook/react-native/issues/17530#issuecomment-662000718. This Drawable is not available in [`mDrawableFutures`](https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L918) when the state changes and the runtime is triggered.

**<details><summary>CLICK TO OPEN DRAWABLES PREVIEWS</summary>**
<p>

| abc_textfield_activated_mtrl_alpha | abc_textfield_default_mtrl_alpha | 
|  |  |

</p>
</details>

The Test Case https://github.com/facebook/react-native/issues/17530#issuecomment-660017858 reproduces the Runtime by adding thousand TextInputs siblings with a key prop to the page.
Adding a new TextInput sibling with key prop, may cause focus to change, triggering the change in BackgroundDrawable and the error.

**<details><summary>CLICK TO OPEN SCREEN RECORDING OF RUNTIME</summary>**
<p>

From https://github.com/facebook/react-native/issues/17530#issuecomment-660458191

| **WITH RUNTIME** | 
|:-------------------------:|
|   |

| **WITHOUT RUNTIME** | 
|:-------------------------:|
|   |

</p>
</details>

Our test case reproduces the error by adding a large number of TextInputs (100+ see https://github.com/facebook/react-native/issues/17530#issuecomment-660017858) on the screen, but other use case scenario may reproduce the same error with different conditions.

I agree with @sunnylqm and I think It's a problem in the Android OpenSource project, which should handle this exceptions without triggering a runtime error. The Android Project should log the information and use the default Background Drawable instead of terminating the program execution.

>It's about probabilistic. So the crash can happen even with only one TextInput. So it's like a rare race condition, in certain unknown scenario the drawables was modified unexpectedly, which makes me thinking there should be similar crashes with other views that have multiple state drawables.

Removing the following line from the above xml file `@drawable/abc_edit_text_material.xml` fixes the error.

```
<item android:state_pressed="false" android:state_focused="false" android:drawable="@drawable/abc_textfield_default_mtrl_alpha"/>
```

I don't believe that the TextInput missing BackgroundDrawable when changing focus is a valid reason to stop the application execution. The Android OpenSource project should log the error without stopping application execution and use the default BackgroundDrawable.

**<details><summary>CLICK TO OPEN DRAWABLES PREVIEWS</summary>**
<p>

| abc_textfield_activated_mtrl_alpha | abc_textfield_default_mtrl_alpha | 
|  |  |

</p>
</details>

React Native and Facebook developers can customize this settings in their `android/app` folder.
This PR changes the default template used to generate new applications.
Expo users do not have this option and can not fix this error without this pull request. They don't have access to the `android/app` folder and this fix would avoid the runtime on many Expo Apps.

This pr adds an example to RNTester to reproduce the error https://github.com/facebook/react-native/blob/c801edc3ed2da9cca2061da88f13c3e77844b32f/packages/rn-tester/js/examples/TextInput/TextInputKeyProp.js

The error triggers also on RNTester TextInputSharedExample.
I remain available for further discussion and improvements.

fabriziobertoglio1987:
An alternative approach would be creating a `ReactDrawableContainer` that inherits his properties from `DrawableContainer` and over-ride `setBackgroundDrawable` to use this `ReactDrawableContainer`.

1) We create a new class `ReactDrawableContainer` that inherits from `DrawableContainer`. We change the logic in `getChild` to catch this exception and use the default `BackgroundDrawable` for the `ReactEditText`

```java
public class ReactDrawableContainer extends DrawableContainer
  //  enhances getChild method from DrawableContainer 
  //  catch exceptions and return default BackgroundDrawable
  //  https://bit.ly/3yJpdDf
  @Ovverride
  public final Drawable getChild(int index) {
    try {
        this.super(index);
    } catch (e) {
        // log exception e
        // return default background drawable for EditText
    }
  }
}
```
2) We enhance method `setBackgroundDrawable` or the required class (using inheritance) to use a `ReactDrawableContainer` instead of the Android OpenSource class `DrawableContainer`.

<image src="https://user-images.githubusercontent.com/24992535/131952704-a3cc345c-a809-4a55-9f2a-0443fdbb79d5.png" width="200" />

exneval:
hope this can be merge soon in next version of RN

sunnylqm:
@fabriziobertoglio1987 Do you mean the same crash would happen on a pure native app (without react native)?

fabriziobertoglio1987:
We add 100-1.000 TextInput to the page. 
One of this TextInputs triggers a runtime error.

This is the stacktrace of the Runtime error

1) STEP ONE - the ReactEditText instance is created and the context is passed as parameter. The context is used to construct the EditText

**<details><summary>method createInternalEditText</summary>**
<p>


https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputShadowNode.java#L264-L266

```java
public class EditText extends TextView {
    public EditText(Context context) {
        this(context, null);
    }
}
```
https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/core/java/android/widget/EditText.java#L74-L77

</p>
</details>

2) STEP TWO 
The EditText triggers the View constructor during its initialization.

**<details><summary>View Constructor 1</summary>**
<p>

```java
    /**
     * Constructor that is called when inflating a view from XML. This is called
     * when a view is being constructed from an XML file, supplying attributes
     * that were specified in the XML file. This version uses a default style of
     * 0, so the only attribute values applied are those in the Context's Theme
     * and the given AttributeSet.
     *
     * <p>
     * The method onFinishInflate() will be called after all children have been
     * added.
     *
     * @param context The Context the view is running in, through which it can
     *        access the current theme, resources, etc.
     * @param attrs The attributes of the XML tag that is inflating the view.
     * @see #View(Context, AttributeSet, int)
     */
    public View(Context context, @Nullable AttributeSet attrs) {
        this(context, attrs, 0);
    }
```
https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/20b012282e0c3d94b5c0aa799cdda065f2df06db/core/java/android/view/View.java

</p>
</details>

**<details><summary>View Constructor 2</summary>**
<p>

```java
    /**
     * Perform inflation from XML and apply a class-specific base style from a
     * theme attribute or style resource. This constructor of View allows
     * subclasses to use their own base style when they are inflating.
     * <p>
     * When determining the final value of a particular attribute, there are
     * four inputs that come into play:
     * <ol>
     * <li>Any attribute values in the given AttributeSet.
     * <li>The style resource specified in the AttributeSet (named "style").
     * <li>The default style specified by <var>defStyleAttr</var>.
     * <li>The default style specified by <var>defStyleRes</var>.
     * <li>The base values in this theme.
     * </ol>
     * <p>
     * Each of these inputs is considered in-order, with the first listed taking
     * precedence over the following ones. In other words, if in the
     * AttributeSet you have supplied <code>&lt;Button * textColor="#ff000000"&gt;</code>
     * , then the button's text will <em>always</em> be black, regardless of
     * what is specified in any of the styles.
     *
     * @param context The Context the view is running in, through which it can
     *        access the current theme, resources, etc.
     * @param attrs The attributes of the XML tag that is inflating the view.
     * @param defStyleAttr An attribute in the current theme that contains a
     *        reference to a style resource that supplies default values for
     *        the view. Can be 0 to not look for defaults.
     * @param defStyleRes A resource identifier of a style resource that
     *        supplies default values for the view, used only if
     *        defStyleAttr is 0 or can not be found in the theme. Can be 0
     *        to not look for defaults.
     * @see #View(Context, AttributeSet, int)
     */
    public View(Context context, @Nullable AttributeSet attrs, int defStyleAttr, int defStyleRes) {
        this(context);

        mSourceLayoutId = Resources.getAttributeSetSourceResId(attrs);

        final TypedArray a = context.obtainStyledAttributes(
                attrs, com.android.internal.R.styleable.View, defStyleAttr, defStyleRes);

        retrieveExplicitStyle(context.getTheme(), attrs);
        saveAttributeDataForStyleable(context, com.android.internal.R.styleable.View, attrs, a,
                defStyleAttr, defStyleRes);

        if (sDebugViewAttributes) {
            saveAttributeData(attrs, a);
        }

        Drawable background = null;

        int leftPadding = -1;
        int topPadding = -1;
        int rightPadding = -1;
        int bottomPadding = -1;
        int startPadding = UNDEFINED_PADDING;
        int endPadding = UNDEFINED_PADDING;

        int padding = -1;
        int paddingHorizontal = -1;
        int paddingVertical = -1;

        int viewFlagValues = 0;
        int viewFlagMasks = 0;

        boolean setScrollContainer = false;

        int x = 0;
        int y = 0;

        float tx = 0;
        float ty = 0;
        float tz = 0;
        float elevation = 0;
        float rotation = 0;
        float rotationX = 0;
        float rotationY = 0;
        float sx = 1f;
        float sy = 1f;
        boolean transformSet = false;

        int scrollbarStyle = SCROLLBARS_INSIDE_OVERLAY;
        int overScrollMode = mOverScrollMode;
        boolean initializeScrollbars = false;
        boolean initializeScrollIndicators = false;

        boolean startPaddingDefined = false;
        boolean endPaddingDefined = false;
        boolean leftPaddingDefined = false;
        boolean rightPaddingDefined = false;

        final int targetSdkVersion = context.getApplicationInfo().targetSdkVersion;

        // Set default values.
        viewFlagValues |= FOCUSABLE_AUTO;
        viewFlagMasks |= FOCUSABLE_AUTO;

        final int N = a.getIndexCount();
        for (int i = 0; i < N; i++) {
            int attr = a.getIndex(i);
            switch (attr) {
                case com.android.internal.R.styleable.View_background:
                    background = a.getDrawable(attr);
                    break;
                case com.android.internal.R.styleable.View_padding:
                    padding = a.getDimensionPixelSize(attr, -1);
                    mUserPaddingLeftInitial = padding;
                    mUserPaddingRightInitial = padding;
                    leftPaddingDefined = true;
                    rightPaddingDefined = true;
                    break;
                case com.android.internal.R.styleable.View_paddingHorizontal:
                    paddingHorizontal = a.getDimensionPixelSize(attr, -1);
                    mUserPaddingLeftInitial = paddingHorizontal;
                    mUserPaddingRightInitial = paddingHorizontal;
                    leftPaddingDefined = true;
                    rightPaddingDefined = true;
                    break;
                case com.android.internal.R.styleable.View_paddingVertical:
                    paddingVertical = a.getDimensionPixelSize(attr, -1);
                    break;
                case com.android.internal.R.styleable.View_paddingLeft:
                    leftPadding = a.getDimensionPixelSize(attr, -1);
                    mUserPaddingLeftInitial = leftPadding;
                    leftPaddingDefined = true;
                    break;
                case com.android.internal.R.styleable.View_paddingTop:
                    topPadding = a.getDimensionPixelSize(attr, -1);
                    break;
                case com.android.internal.R.styleable.View_paddingRight:
                    rightPadding = a.getDimensionPixelSize(attr, -1);
                    mUserPaddingRightInitial = rightPadding;
                    rightPaddingDefined = true;
                    break;
                case com.android.internal.R.styleable.View_paddingBottom:
                    bottomPadding = a.getDimensionPixelSize(attr, -1);
                    break;
                case com.android.internal.R.styleable.View_paddingStart:
                    startPadding = a.getDimensionPixelSize(attr, UNDEFINED_PADDING);
                    startPaddingDefined = (startPadding != UNDEFINED_PADDING);
                    break;
                case com.android.internal.R.styleable.View_paddingEnd:
                    endPadding = a.getDimensionPixelSize(attr, UNDEFINED_PADDING);
                    endPaddingDefined = (endPadding != UNDEFINED_PADDING);
                    break;
                case com.android.internal.R.styleable.View_scrollX:
                    x = a.getDimensionPixelOffset(attr, 0);
                    break;
                case com.android.internal.R.styleable.View_scrollY:
                    y = a.getDimensionPixelOffset(attr, 0);
                    break;
                case com.android.internal.R.styleable.View_alpha:
                    setAlpha(a.getFloat(attr, 1f));
                    break;
                case com.android.internal.R.styleable.View_transformPivotX:
                    setPivotX(a.getDimension(attr, 0));
                    break;
                case com.android.internal.R.styleable.View_transformPivotY:
                    setPivotY(a.getDimension(attr, 0));
                    break;
                case com.android.internal.R.styleable.View_translationX:
                    tx = a.getDimension(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_translationY:
                    ty = a.getDimension(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_translationZ:
                    tz = a.getDimension(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_elevation:
                    elevation = a.getDimension(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_rotation:
                    rotation = a.getFloat(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_rotationX:
                    rotationX = a.getFloat(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_rotationY:
                    rotationY = a.getFloat(attr, 0);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_scaleX:
                    sx = a.getFloat(attr, 1f);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_scaleY:
                    sy = a.getFloat(attr, 1f);
                    transformSet = true;
                    break;
                case com.android.internal.R.styleable.View_id:
                    mID = a.getResourceId(attr, NO_ID);
                    break;
                case com.android.internal.R.styleable.View_tag:
                    mTag = a.getText(attr);
                    break;
                case com.android.internal.R.styleable.View_fitsSystemWindows:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= FITS_SYSTEM_WINDOWS;
                        viewFlagMasks |= FITS_SYSTEM_WINDOWS;
                    }
                    break;
                case com.android.internal.R.styleable.View_focusable:
                    viewFlagValues = (viewFlagValues & ~FOCUSABLE_MASK) | getFocusableAttribute(a);
                    if ((viewFlagValues & FOCUSABLE_AUTO) == 0) {
                        viewFlagMasks |= FOCUSABLE_MASK;
                    }
                    break;
                case com.android.internal.R.styleable.View_focusableInTouchMode:
                    if (a.getBoolean(attr, false)) {
                        // unset auto focus since focusableInTouchMode implies explicit focusable
                        viewFlagValues &= ~FOCUSABLE_AUTO;
                        viewFlagValues |= FOCUSABLE_IN_TOUCH_MODE | FOCUSABLE;
                        viewFlagMasks |= FOCUSABLE_IN_TOUCH_MODE | FOCUSABLE_MASK;
                    }
                    break;
                case com.android.internal.R.styleable.View_clickable:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= CLICKABLE;
                        viewFlagMasks |= CLICKABLE;
                    }
                    break;
                case com.android.internal.R.styleable.View_longClickable:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= LONG_CLICKABLE;
                        viewFlagMasks |= LONG_CLICKABLE;
                    }
                    break;
                case com.android.internal.R.styleable.View_contextClickable:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= CONTEXT_CLICKABLE;
                        viewFlagMasks |= CONTEXT_CLICKABLE;
                    }
                    break;
                case com.android.internal.R.styleable.View_saveEnabled:
                    if (!a.getBoolean(attr, true)) {
                        viewFlagValues |= SAVE_DISABLED;
                        viewFlagMasks |= SAVE_DISABLED_MASK;
                    }
                    break;
                case com.android.internal.R.styleable.View_duplicateParentState:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= DUPLICATE_PARENT_STATE;
                        viewFlagMasks |= DUPLICATE_PARENT_STATE;
                    }
                    break;
                case com.android.internal.R.styleable.View_visibility:
                    final int visibility = a.getInt(attr, 0);
                    if (visibility != 0) {
                        viewFlagValues |= VISIBILITY_FLAGS[visibility];
                        viewFlagMasks |= VISIBILITY_MASK;
                    }
                    break;
                case com.android.internal.R.styleable.View_layoutDirection:
                    // Clear any layout direction flags (included resolved bits) already set
                    mPrivateFlags2 &=
                            ~(PFLAG2_LAYOUT_DIRECTION_MASK | PFLAG2_LAYOUT_DIRECTION_RESOLVED_MASK);
                    // Set the layout direction flags depending on the value of the attribute
                    final int layoutDirection = a.getInt(attr, -1);
                    final int value = (layoutDirection != -1) ?
                            LAYOUT_DIRECTION_FLAGS[layoutDirection] : LAYOUT_DIRECTION_DEFAULT;
                    mPrivateFlags2 |= (value << PFLAG2_LAYOUT_DIRECTION_MASK_SHIFT);
                    break;
                case com.android.internal.R.styleable.View_drawingCacheQuality:
                    final int cacheQuality = a.getInt(attr, 0);
                    if (cacheQuality != 0) {
                        viewFlagValues |= DRAWING_CACHE_QUALITY_FLAGS[cacheQuality];
                        viewFlagMasks |= DRAWING_CACHE_QUALITY_MASK;
                    }
                    break;
                case com.android.internal.R.styleable.View_contentDescription:
                    setContentDescription(a.getString(attr));
                    break;
                case com.android.internal.R.styleable.View_accessibilityTraversalBefore:
                    setAccessibilityTraversalBefore(a.getResourceId(attr, NO_ID));
                    break;
                case com.android.internal.R.styleable.View_accessibilityTraversalAfter:
                    setAccessibilityTraversalAfter(a.getResourceId(attr, NO_ID));
                    break;
                case com.android.internal.R.styleable.View_labelFor:
                    setLabelFor(a.getResourceId(attr, NO_ID));
                    break;
                case com.android.internal.R.styleable.View_soundEffectsEnabled:
                    if (!a.getBoolean(attr, true)) {
                        viewFlagValues &= ~SOUND_EFFECTS_ENABLED;
                        viewFlagMasks |= SOUND_EFFECTS_ENABLED;
                    }
                    break;
                case com.android.internal.R.styleable.View_hapticFeedbackEnabled:
                    if (!a.getBoolean(attr, true)) {
                        viewFlagValues &= ~HAPTIC_FEEDBACK_ENABLED;
                        viewFlagMasks |= HAPTIC_FEEDBACK_ENABLED;
                    }
                    break;
                case R.styleable.View_scrollbars:
                    final int scrollbars = a.getInt(attr, SCROLLBARS_NONE);
                    if (scrollbars != SCROLLBARS_NONE) {
                        viewFlagValues |= scrollbars;
                        viewFlagMasks |= SCROLLBARS_MASK;
                        initializeScrollbars = true;
                    }
                    break;
                //noinspection deprecation
                case R.styleable.View_fadingEdge:
                    if (targetSdkVersion >= Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
                        // Ignore the attribute starting with ICS
                        break;
                    }
                    // With builds < ICS, fall through and apply fading edges
                case R.styleable.View_requiresFadingEdge:
                    final int fadingEdge = a.getInt(attr, FADING_EDGE_NONE);
                    if (fadingEdge != FADING_EDGE_NONE) {
                        viewFlagValues |= fadingEdge;
                        viewFlagMasks |= FADING_EDGE_MASK;
                        initializeFadingEdgeInternal(a);
                    }
                    break;
                case R.styleable.View_scrollbarStyle:
                    scrollbarStyle = a.getInt(attr, SCROLLBARS_INSIDE_OVERLAY);
                    if (scrollbarStyle != SCROLLBARS_INSIDE_OVERLAY) {
                        viewFlagValues |= scrollbarStyle & SCROLLBARS_STYLE_MASK;
                        viewFlagMasks |= SCROLLBARS_STYLE_MASK;
                    }
                    break;
                case R.styleable.View_isScrollContainer:
                    setScrollContainer = true;
                    if (a.getBoolean(attr, false)) {
                        setScrollContainer(true);
                    }
                    break;
                case com.android.internal.R.styleable.View_keepScreenOn:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= KEEP_SCREEN_ON;
                        viewFlagMasks |= KEEP_SCREEN_ON;
                    }
                    break;
                case R.styleable.View_filterTouchesWhenObscured:
                    if (a.getBoolean(attr, false)) {
                        viewFlagValues |= FILTER_TOUCHES_WHEN_OBSCURED;
                        viewFlagMasks |= FILTER_TOUCHES_WHEN_OBSCURED;
                    }
                    break;
                case R.styleable.View_nextFocusLeft:
                    mNextFocusLeftId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_nextFocusRight:
                    mNextFocusRightId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_nextFocusUp:
                    mNextFocusUpId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_nextFocusDown:
                    mNextFocusDownId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_nextFocusForward:
                    mNextFocusForwardId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_nextClusterForward:
                    mNextClusterForwardId = a.getResourceId(attr, View.NO_ID);
                    break;
                case R.styleable.View_minWidth:
                    mMinWidth = a.getDimensionPixelSize(attr, 0);
                    break;
                case R.styleable.View_minHeight:
                    mMinHeight = a.getDimensionPixelSize(attr, 0);
                    break;
                case R.styleable.View_onClick:
                    if (context.isRestricted()) {
                        throw new IllegalStateException("The android:onClick attribute cannot "
                                + "be used within a restricted context");
                    }

                    final String handlerName = a.getString(attr);
                    if (handlerName != null) {
                        setOnClickListener(new DeclaredOnClickListener(this, handlerName));
                    }
                    break;
                case R.styleable.View_overScrollMode:
                    overScrollMode = a.getInt(attr, OVER_SCROLL_IF_CONTENT_SCROLLS);
                    break;
                case R.styleable.View_verticalScrollbarPosition:
                    mVerticalScrollbarPosition = a.getInt(attr, SCROLLBAR_POSITION_DEFAULT);
                    break;
                case R.styleable.View_layerType:
                    setLayerType(a.getInt(attr, LAYER_TYPE_NONE), null);
                    break;
                case R.styleable.View_textDirection:
                    // Clear any text direction flag already set
                    mPrivateFlags2 &= ~PFLAG2_TEXT_DIRECTION_MASK;
                    // Set the text direction flags depending on the value of the attribute
                    final int textDirection = a.getInt(attr, -1);
                    if (textDirection != -1) {
                        mPrivateFlags2 |= PFLAG2_TEXT_DIRECTION_FLAGS[textDirection];
                    }
                    break;
                case R.styleable.View_textAlignment:
                    // Clear any text alignment flag already set
                    mPrivateFlags2 &= ~PFLAG2_TEXT_ALIGNMENT_MASK;
                    // Set the text alignment flag depending on the value of the attribute
                    final int textAlignment = a.getInt(attr, TEXT_ALIGNMENT_DEFAULT);
                    mPrivateFlags2 |= PFLAG2_TEXT_ALIGNMENT_FLAGS[textAlignment];
                    break;
                case R.styleable.View_importantForAccessibility:
                    setImportantForAccessibility(a.getInt(attr,
                            IMPORTANT_FOR_ACCESSIBILITY_DEFAULT));
                    break;
                case R.styleable.View_accessibilityLiveRegion:
                    setAccessibilityLiveRegion(a.getInt(attr, ACCESSIBILITY_LIVE_REGION_DEFAULT));
                    break;
                case R.styleable.View_transitionName:
                    setTransitionName(a.getString(attr));
                    break;
                case R.styleable.View_nestedScrollingEnabled:
                    setNestedScrollingEnabled(a.getBoolean(attr, false));
                    break;
                case R.styleable.View_stateListAnimator:
                    setStateListAnimator(AnimatorInflater.loadStateListAnimator(context,
                            a.getResourceId(attr, 0)));
                    break;
                case R.styleable.View_backgroundTint:
                    // This will get applied later during setBackground().
                    if (mBackgroundTint == null) {
                        mBackgroundTint = new TintInfo();
                    }
                    mBackgroundTint.mTintList = a.getColorStateList(
                            R.styleable.View_backgroundTint);
                    mBackgroundTint.mHasTintList = true;
                    break;
                case R.styleable.View_backgroundTintMode:
                    // This will get applied later during setBackground().
                    if (mBackgroundTint == null) {
                        mBackgroundTint = new TintInfo();
                    }
                    mBackgroundTint.mBlendMode = Drawable.parseBlendMode(a.getInt(
                            R.styleable.View_backgroundTintMode, -1), null);
                    mBackgroundTint.mHasTintMode = true;
                    break;
                case R.styleable.View_outlineProvider:
                    setOutlineProviderFromAttribute(a.getInt(R.styleable.View_outlineProvider,
                            PROVIDER_BACKGROUND));
                    break;
                case R.styleable.View_foreground:
                    if (targetSdkVersion >= Build.VERSION_CODES.M || this instanceof FrameLayout) {
                        setForeground(a.getDrawable(attr));
                    }
                    break;
                case R.styleable.View_foregroundGravity:
                    if (targetSdkVersion >= Build.VERSION_CODES.M || this instanceof FrameLayout) {
                        setForegroundGravity(a.getInt(attr, Gravity.NO_GRAVITY));
                    }
                    break;
                case R.styleable.View_foregroundTintMode:
                    if (targetSdkVersion >= Build.VERSION_CODES.M || this instanceof FrameLayout) {
                        setForegroundTintBlendMode(
                                Drawable.parseBlendMode(a.getInt(attr, -1),
                                        null));
                    }
                    break;
                case R.styleable.View_foregroundTint:
                    if (targetSdkVersion >= Build.VERSION_CODES.M || this instanceof FrameLayout) {
                        setForegroundTintList(a.getColorStateList(attr));
                    }
                    break;
                case R.styleable.View_foregroundInsidePadding:
                    if (targetSdkVersion >= Build.VERSION_CODES.M || this instanceof FrameLayout) {
                        if (mForegroundInfo == null) {
                            mForegroundInfo = new ForegroundInfo();
                        }
                        mForegroundInfo.mInsidePadding = a.getBoolean(attr,
                                mForegroundInfo.mInsidePadding);
                    }
                    break;
                case R.styleable.View_scrollIndicators:
                    final int scrollIndicators =
                            (a.getInt(attr, 0) << SCROLL_INDICATORS_TO_PFLAGS3_LSHIFT)
                                    & SCROLL_INDICATORS_PFLAG3_MASK;
                    if (scrollIndicators != 0) {
                        mPrivateFlags3 |= scrollIndicators;
                        initializeScrollIndicators = true;
                    }
                    break;
                case R.styleable.View_pointerIcon:
                    final int resourceId = a.getResourceId(attr, 0);
                    if (resourceId != 0) {
                        setPointerIcon(PointerIcon.load(
                                context.getResources(), resourceId));
                    } else {
                        final int pointerType = a.getInt(attr, PointerIcon.TYPE_NOT_SPECIFIED);
                        if (pointerType != PointerIcon.TYPE_NOT_SPECIFIED) {
                            setPointerIcon(PointerIcon.getSystemIcon(context, pointerType));
                        }
                    }
                    break;
                case R.styleable.View_forceHasOverlappingRendering:
                    if (a.peekValue(attr) != null) {
                        forceHasOverlappingRendering(a.getBoolean(attr, true));
                    }
                    break;
                case R.styleable.View_tooltipText:
                    setTooltipText(a.getText(attr));
                    break;
                case R.styleable.View_keyboardNavigationCluster:
                    if (a.peekValue(attr) != null) {
                        setKeyboardNavigationCluster(a.getBoolean(attr, true));
                    }
                    break;
                case R.styleable.View_focusedByDefault:
                    if (a.peekValue(attr) != null) {
                        setFocusedByDefault(a.getBoolean(attr, true));
                    }
                    break;
                case R.styleable.View_autofillHints:
                    if (a.peekValue(attr) != null) {
                        CharSequence[] rawHints = null;
                        String rawString = null;

                        if (a.getType(attr) == TypedValue.TYPE_REFERENCE) {
                            int resId = a.getResourceId(attr, 0);

                            try {
                                rawHints = a.getTextArray(attr);
                            } catch (Resources.NotFoundException e) {
                                rawString = getResources().getString(resId);
                            }
                        } else {
                            rawString = a.getString(attr);
                        }

                        if (rawHints == null) {
                            if (rawString == null) {
                                throw new IllegalArgumentException(
                                        "Could not resolve autofillHints");
                            } else {
                                rawHints = rawString.split(",");
                            }
                        }

                        String[] hints = new String[rawHints.length];

                        int numHints = rawHints.length;
                        for (int rawHintNum = 0; rawHintNum < numHints; rawHintNum++) {
                            hints[rawHintNum] = rawHints[rawHintNum].toString().trim();
                        }
                        setAutofillHints(hints);
                    }
                    break;
                case R.styleable.View_importantForAutofill:
                    if (a.peekValue(attr) != null) {
                        setImportantForAutofill(a.getInt(attr, IMPORTANT_FOR_AUTOFILL_AUTO));
                    }
                    break;
                case R.styleable.View_defaultFocusHighlightEnabled:
                    if (a.peekValue(attr) != null) {
                        setDefaultFocusHighlightEnabled(a.getBoolean(attr, true));
                    }
                    break;
                case R.styleable.View_screenReaderFocusable:
                    if (a.peekValue(attr) != null) {
                        setScreenReaderFocusable(a.getBoolean(attr, false));
                    }
                    break;
                case R.styleable.View_accessibilityPaneTitle:
                    if (a.peekValue(attr) != null) {
                        setAccessibilityPaneTitle(a.getString(attr));
                    }
                    break;
                case R.styleable.View_outlineSpotShadowColor:
                    setOutlineSpotShadowColor(a.getColor(attr, Color.BLACK));
                    break;
                case R.styleable.View_outlineAmbientShadowColor:
                    setOutlineAmbientShadowColor(a.getColor(attr, Color.BLACK));
                    break;
                case com.android.internal.R.styleable.View_accessibilityHeading:
                    setAccessibilityHeading(a.getBoolean(attr, false));
                    break;
                case R.styleable.View_forceDarkAllowed:
                    mRenderNode.setForceDarkAllowed(a.getBoolean(attr, true));
                    break;
            }
        }

        setOverScrollMode(overScrollMode);

        // Cache start/end user padding as we cannot fully resolve padding here (we don't have yet
        // the resolved layout direction). Those cached values will be used later during padding
        // resolution.
        mUserPaddingStart = startPadding;
        mUserPaddingEnd = endPadding;

        if (background != null) {
            setBackground(background);
        }

        // setBackground above will record that padding is currently provided by the background.
        // If we have padding specified via xml, record that here instead and use it.
        mLeftPaddingDefined = leftPaddingDefined;
        mRightPaddingDefined = rightPaddingDefined;

        // Valid paddingHorizontal/paddingVertical beats leftPadding, rightPadding, topPadding,
        // bottomPadding, and padding set by background.  Valid padding beats everything.
        if (padding >= 0) {
            leftPadding = padding;
            topPadding = padding;
            rightPadding = padding;
            bottomPadding = padding;
            mUserPaddingLeftInitial = padding;
            mUserPaddingRightInitial = padding;
        } else {
            if (paddingHorizontal >= 0) {
                leftPadding = paddingHorizontal;
                rightPadding = paddingHorizontal;
                mUserPaddingLeftInitial = paddingHorizontal;
                mUserPaddingRightInitial = paddingHorizontal;
            }
            if (paddingVertical >= 0) {
                topPadding = paddingVertical;
                bottomPadding = paddingVertical;
            }
        }

        if (isRtlCompatibilityMode()) {
            // RTL compatibility mode: pre Jelly Bean MR1 case OR no RTL support case.
            // left / right padding are used if defined (meaning here nothing to do). If they are not
            // defined and start / end padding are defined (e.g. in Frameworks resources), then we use
            // start / end and resolve them as left / right (layout direction is not taken into account).
            // Padding from the background drawable is stored at this point in mUserPaddingLeftInitial
            // and mUserPaddingRightInitial) so drawable padding will be used as ultimate default if
            // defined.
            if (!mLeftPaddingDefined && startPaddingDefined) {
                leftPadding = startPadding;
            }
            mUserPaddingLeftInitial = (leftPadding >= 0) ? leftPadding : mUserPaddingLeftInitial;
            if (!mRightPaddingDefined && endPaddingDefined) {
                rightPadding = endPadding;
            }
            mUserPaddingRightInitial = (rightPadding >= 0) ? rightPadding : mUserPaddingRightInitial;
        } else {
            // Jelly Bean MR1 and after case: if start/end defined, they will override any left/right
            // values defined. Otherwise, left /right values are used.
            // Padding from the background drawable is stored at this point in mUserPaddingLeftInitial
            // and mUserPaddingRightInitial) so drawable padding will be used as ultimate default if
            // defined.
            final boolean hasRelativePadding = startPaddingDefined || endPaddingDefined;

            if (mLeftPaddingDefined && !hasRelativePadding) {
                mUserPaddingLeftInitial = leftPadding;
            }
            if (mRightPaddingDefined && !hasRelativePadding) {
                mUserPaddingRightInitial = rightPadding;
            }
        }

        // mPaddingTop and mPaddingBottom may have been set by setBackground(Drawable) so must pass
        // them on if topPadding or bottomPadding are not valid.
        internalSetPadding(
                mUserPaddingLeftInitial,
                topPadding >= 0 ? topPadding : mPaddingTop,
                mUserPaddingRightInitial,
                bottomPadding >= 0 ? bottomPadding : mPaddingBottom);

        if (viewFlagMasks != 0) {
            setFlags(viewFlagValues, viewFlagMasks);
        }

        if (initializeScrollbars) {
            initializeScrollbarsInternal(a);
        }

        if (initializeScrollIndicators) {
            initializeScrollIndicatorsInternal();
        }

        a.recycle();

        // Needs to be called after mViewFlags is set
        if (scrollbarStyle != SCROLLBARS_INSIDE_OVERLAY) {
            recomputePadding();
        }

        if (x != 0 || y != 0) {
            scrollTo(x, y);
        }

        if (transformSet) {
            setTranslationX(tx);
            setTranslationY(ty);
            setTranslationZ(tz);
            setElevation(elevation);
            setRotation(rotation);
            setRotationX(rotationX);
            setRotationY(rotationY);
            setScaleX(sx);
            setScaleY(sy);
        }

        if (!setScrollContainer && (viewFlagValues&SCROLLBARS_VERTICAL) != 0) {
            setScrollContainer(true);
        }

        computeOpaqueFlags();
    }
```

</p>
</details>


3) STEP THREE
The View constructor (init) sets the BackgroundDrawable for the EditText with `setBackgroundDrawable`

fabriziobertoglio1987:
4. STEP FOUR The DrawableContainerState does not have the Opacity saved in mOpacity and calls `createAllFutures()` to find the opacity value

**<details><summary>DrawableContainerState.getOpacity()</summary>**
<p>

```java
        public final int getOpacity() {
            if (mCheckedOpacity) {
                return mOpacity;
            }


            createAllFutures();


            final int N = mNumChildren;
            final Drawable[] drawables = mDrawables;
            int op = (N > 0) ? drawables[0].getOpacity() : PixelFormat.TRANSPARENT;
            for (int i = 1; i < N; i++) {
                op = Drawable.resolveOpacity(op, drawables[i].getOpacity());
            }


            mOpacity = op;
            mCheckedOpacity = true;
            return op;
        }
```

 https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L678-L684
 
 https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L1158-L1175

</p>
</details>

5. STEP FIVE Runtime Error because ConstantState cs is null

**<details><summary>createAllFutures</summary>**
<p>

```java
        private void createAllFutures() {
            if (mDrawableFutures != null) {
                final int futureCount = mDrawableFutures.size();
                for (int keyIndex = 0; keyIndex < futureCount; keyIndex++) {
                    final int index = mDrawableFutures.keyAt(keyIndex);
                    final ConstantState cs = mDrawableFutures.valueAt(keyIndex);
                    mDrawables[index] = prepareDrawable(cs.newDrawable(mSourceRes));
                }


                mDrawableFutures = null;
            }
        }
```

https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L874-L885
</p>
</details>

fabriziobertoglio1987:
**<details><summary>setConstantState sets the mDrawableFutures</summary>**
<p>


```java
    protected void setConstantState(DrawableContainerState state) {
        mDrawableContainerState = state;


        // The locally cached drawables may have changed.
        if (mCurIndex >= 0) {
            mCurrDrawable = state.getChild(mCurIndex);
            if (mCurrDrawable != null) {
                initializeDrawableForDisplay(mCurrDrawable);
            }
        }


        // Clear out the last drawable. We don't have enough information to
        // propagate local state from the past.
        mLastIndex = -1;
        mLastDrawable = null;
    }
```
https://github.com/aosp-mirror/platform_frameworks_base/blob/20b012282e0c3d94b5c0aa799cdda065f2df06db/graphics/java/android/graphics/drawable/DrawableContainer.java#L1229-L1244

</p>
</details>

lunaleaps:
There's some internal discussion about the approach. I'll copy over @ShikaSD's comments

packages/rn-tester/android/app/src/main/res/values/styles.xml
```diff
@@ -1,9 +1,9 @@
 <resources>
-
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <!-- Customize your theme here. -->
         <item name="android:textColor">#000000</item>
+        <item name="android:editTextBackground">@drawable/edit_text</item>
```
lunaleaps, line 6: Re: @ShikaSD 
It would be great to add documentation why this file is needed + better prefix for the resource as well

lunaleaps (changes requested):
Re: @ShikaSD 
Not sure it is the best way to resolve this issue, as I believe that it should affect other views as well. Also including a random .xml is a nice workaround but I am not sure we want to ship it with the framework

fabriziobertoglio1987:
https://issuetracker.google.com/issues/171913944

packages/rn-tester/android/app/src/main/res/values/styles.xml
fabriziobertoglio1987: >Not sure it is the best way to resolve this issue, as I believe that it should affect other views as well. Also including a random .xml is a nice workaround but I am not sure we want to ship it with the framework

@lunaleaps  Do you mean that I have to prepare an entire new approach to fix this issue (https://github.com/facebook/react-native/pull/29452#issuecomment-912257356 and https://github.com/facebook/react-native/pull/29452#issuecomment-915687725)? Could ShikaSD give me some help or feedback on how to solve this issue? https://github.com/facebook/react-native/pull/29452#issuecomment-915687725. Thanks

lunaleaps: Let's have a discussion about @ShikaSD's concerns -- he is currently away and will return next week.

ShikaSD: Hey, thanks for approaching this, do you know if this repros everywhere or only on particular Android version?

My comment was mostly about the fact that setting XML through themes will only affect the edit text with this theme applied. I don't think we can guarantee that every React Native app uses the default theme or the material background, so here the proposed solution feels like a workaround that could help only in some cases. 🙂 
I also noticed the reports in the original issue thread with slightly different traces/views, meaning that this patch might not fix the issue completely, as we cannot control which drawables people use in their applications.

I guess the proper solution would be to patch Android SDK to actually fix this issue, but as long as this is not possible, maybe we can find a more stable workaround that can at least catch the issue before the crash 😅

ShikaSD: I also checked the comments in the thread, I am not sure what is the proposed solution in both https://github.com/facebook/react-native/pull/29452#issuecomment-915687725 or https://github.com/facebook/react-native/pull/29452#issuecomment-912257356 is. The Drawables mentioned in both comments assume that we have some access to definition of those, but it seems like the original problem comes from XML defined drawables in design/appcompat libs which we cannot really influence.

I think we should be good to go with the temporary fix for this particular edit text with minor edits (@lunaleaps ported some comments above about adding some reference/better naming), but still concerned if the drawables are also crashing on other views.

lfoliveir4:
@fabriziobertoglio1987 @lunaleaps Hi! Any predictions for this pull request to be approved?

nicewahson:
Is this PR released? I have not found any changelog for it.

MaeIg:
> Is this PR released? I have not found any changelog for it.

Hello, 

You can check [here](https://github.com/facebook/react-native/commit/254493e1fb0c3a1e279e2c957e83edac6252d041) if this fix is planned for a future release.

It should be in v0.67 ! 
image

lennartkloock:
Is this already merged? I'm using version `0.68.2` and this still happens exactly as described.

devjta:
I found that thing again in our Crashlytics..

`      Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable$ConstantState.newDrawable(android.content.res.Resources)' on a null object reference
       at android.graphics.drawable.DrawableContainer$DrawableContainerState.createAllFutures(DrawableContainer.java:879)
       at android.graphics.drawable.DrawableContainer$DrawableContainerState.getOpacity(DrawableContainer.java:1162)
       at android.graphics.drawable.DrawableContainer.getOpacity(DrawableContainer.java:433)
       at android.graphics.drawable.InsetDrawable.getOpacity(InsetDrawable.java:260)
       at android.view.View.computeOpaqueFlags(View.java:20555)
       at android.view.View.setBackgroundDrawable(View.java:25924)
       at androidx.appcompat.widget.AppCompatEditText.setBackgroundDrawable(AppCompatEditText.java:183)
       at android.view.View.setBackground(View.java:25817)
       at android.view.View.<init>(View.java:6399)
       at android.widget.TextView.<init>(TextView.java:1291)
       at android.widget.EditText.<init>(EditText.java:99)
       at android.widget.EditText.<init>(EditText.java:95)
       at androidx.appcompat.widget.AppCompatEditText.<init>(AppCompatEditText.java:100)
       at androidx.appcompat.widget.AppCompatEditText.<init>(AppCompatEditText.java:95)
       at androidx.appcompat.widget.AppCompatEditText.<init>(AppCompatEditText.java:91)
       at com.facebook.react.views.textinput.ReactEditText.<init>(ReactEditText.java:137)
       at com.facebook.react.views.textinput.ReactTextInputManager.createViewInstance(ReactTextInputManager.java:191)
       at com.facebook.react.views.textinput.ReactTextInputManager.createViewInstance(ReactTextInputManager.java:86)
       at com.facebook.react.uimanager.ViewManager.createViewInstance(ViewManager.java:183)
       at com.facebook.react.uimanager.ViewManager.createView(ViewManager.java:115)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:281)
       at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute(UIViewOperationQueue.java:194)
       at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.dispatchPendingNonBatchedOperations(UIViewOperationQueue.java:1110)
       at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.doFrameGuarded(UIViewOperationQueue.java:1081)
       at com.facebook.react.modules.core.ReactChoreographer$ReactChoreographerDispatcher.doFrame(ReactChoreographer.java:175)
       at com.facebook.react.modules.core.ChoreographerCompat$FrameCallback$1.doFrame(ChoreographerCompat.java:85)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1299)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1309)
       at android.view.Choreographer.doCallbacks(Choreographer.java:923)
       at android.view.Choreographer.doFrame(Choreographer.java:847)
       at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1283)
       at android.os.Handler.handleCallback(Handler.java:942)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8772)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)
    `

Our versions from package.json
`"react": "18.2.0",
    "react-hook-form": "^7.6.3",
    "react-i18next": "^11.11.4",
    "react-native": "0.72.4",
    "react-native-blob-util": "^0.13.17",
    "react-native-contacts": "^7.0.0",
    "react-native-device-info": "^10.11.0",
    "react-native-dialog": "^9.2.2",
    "react-native-document-picker": "^8.1.3",
    "react-native-drag-sort": "^2.4.4",
    "react-native-elements": "^3.4.1",
    "react-native-gesture-handler": "^1.10.3",
    "react-native-global-props": "^1.1.5",
    "react-native-image-picker": "^4.8.3",
    "react-native-indicators": "^0.17.0",
    "react-native-modal": "^12.0.3",
    "react-native-pdf": "^6.4.0",
    "react-native-raw-bottom-sheet": "^2.2.0",
    "react-native-safe-area-context": "^4.4.1",
    "react-native-screens": "3.25.0",
    "react-native-svg": "^12.1.1",
    "react-native-toast-message": "^1.4.9",
    "react-native-vector-icons": "^10.0.0",
    "redux-object": "^1.0.1"
`

PR #29468, opened 2020-07-22
chore: remove FlowFixMe

Naturalclar:
## Summary
Removed FlowFixMe that has been fixed

## Changelog

[General] [Fixed] - Removed FlowFixMe that has been fixed

## Test Plan
flow check passes


lunaleaps:
Sorry mis-click on the close! Importing to verify

lunaleaps:
So on import internally, I'm seeing a flow error still. Here's the error I'm seeing:
```
Cannot call `ShallowRenderer` because module `react-shallow-renderer` [1] is not a function. [not-a-function]
```
I'm unsure why this is an issue internally but not for OSS.

lunaleaps:
I've changed this to be an ESM import and that seems to resolve flow issues internally

PR #29510, opened 2020-07-28
Make some RN rules no-op for arc focus

jiawei-lyu:
Differential Revision: D22495239


yungsters:
Closing because D22495239 already landed.

PR #29511, opened 2020-07-28
update README.md with more detail

Deem97:
## Changelog

[CATEGORY] [TYPE] - Message


yungsters:
This is already mentioned in the sentence above:

> With React Native, you use native UI controls and have full access to the native platform.

PR #29523, opened 2020-07-29
Update ActivityIndicator component

ani4aniket:
Fixes https://github.com/MLH-Fellowship/react-native/issues/34

## Summary
The PR is part of an effort to update the code comments to match the current documentation on the React Native website. The project is a part of the MLH Fellowship program and involves the automatic generation of the website docs from code comments and flow types as the end result.

To learn more about the project you can visit the project wiki: 
- [Project Details](https://github.com/MLH-Fellowship/0.4.x-projects/wiki/React-Native-Flowtype-API-Docs-Generator)
- [RN Docs Standards](https://github.com/MLH-Fellowship/react-native/wiki/RN-Docs-standards)

Link to the documentation(the source of truth): 
- [activityindicator.md](https://github.com/MLH-Fellowship/react-native-website/blob/master/docs/activityindicator.md)

## Changes
* Update the title and prop description from docs.
* Remove unnecessary `*` from the code comments.
* Add Snack player example specified in the docs to the code comments as JSDoc.
* Add `@type` annotation to parse supported datatype by the prop.
* Add `@platform` annotation to specify platforms supported by a prop.
* Add `@default` annotation to parse default value of prop.

## Changelog
[Internal]

## Test Plan
All changes are made to the code comments and thus there is no need for testing.

Reviewed by @jevakallio


ani4aniket:
https://github.com/MLH-Fellowship/react-native/blob/6708d6bdb9a68c5c36f017aa32f4a1980c56e4f9/Libraries/Components/ActivityIndicator/ActivityIndicator.js#L51-L52

ani4aniket:
@motiz88 Is this alright? I don't know why one test is failing while my changes are on JSDocs comment.

Libraries/Components/ActivityIndicator/ActivityIndicator.js
```diff
@@ -29,44 +30,38 @@ type IndicatorSize = number | 'small' | 'large';
...
 type IOSProps = $ReadOnly<{|
   /**
-   * Whether the indicator should hide when not animating (true by default).
-   *
-   * See https://reactnative.dev/docs/activityindicator.html#hideswhenstopped
-   */
+    Whether the indicator should hide when not animating.
+
+    @platform ios
+  */
```
motiz88, line 36: I think this is missing an annotation for the default value (`true`).

Libraries/Components/ActivityIndicator/ActivityIndicator.js
```diff
@@ -29,44 +30,38 @@ type IndicatorSize = number | 'small' | 'large';
...
 
   /**
-   * Size of the indicator (default is 'small').
-   * Passing a number to the size prop is only supported on Android.
-   *
-   * See https://reactnative.dev/docs/activityindicator.html#size
-   */
+    Size of the indicator.
+
+    @type enum(`'small'`, `'large'`)
```
motiz88, line 59: **Long term comment - no need to fix before merging**

I'm not sure we should be embedding Markdown backticks in types like this. This may be fine to ship as a starting point, but later we should consider migrating to a type syntax with no manual formatting. This type of formatting should "just work" implicitly, like syntax highlighting.

```suggestion
    @type enum('small', 'large')
```

Libraries/Components/ActivityIndicator/ActivityIndicator.js
```diff
@@ -29,44 +30,38 @@ type IndicatorSize = number | 'small' | 'large';
...
+    @type {@platform android} number
```
motiz88, line 60: **Long term comment - no need to fix before merging**

This use of multiple `@type` annotations is pretty clever!

I think in the long run we should look into annotating the Flow types themselves, e.g. maybe something like:

```js
type IndicatorSize = 'small' | 'large' | /** @platform android **/ number;
```

Or using Flow types directly:
```js
// A utility type which the docs generator will treat specially
type RN$PlatformSpecific<PlatformT, T> = T;

type IndicatorSize = 'small' | 'large' | RN$PlatformSpecific<'android', number>;
```

The latter has the added benefit that, in the future, we can reimplement `RN$PlatformSpecific` to actually enforce the correct types per platform.

Libraries/Components/ActivityIndicator/ActivityIndicator.js
ani4aniket: In this case, the default value is coming from FlowTypes via react-docgen.
Screenshot 2020-08-01 at 5 20 06 PM

Libraries/Components/ActivityIndicator/ActivityIndicator.js
ani4aniket: Yes, this needs to be addressed. To fix that we need to split and add the script for formating the values of enums. We can try that out to fix.

Libraries/Components/ActivityIndicator/ActivityIndicator.js
ani4aniket: Yes. Thanks 😃

PR #29530, opened 2020-07-30
Add invariant check to BackHandler

ospfranco:
## Summary

Right now one can register `null` or `undefined` handlers to BackHandler, no error is thrown (except maybe with ts/flow), however once the app is running everything looks fine and only when the back button is pressed a (cryptic) runtime error will be thrown

## Changelog

[Android/tvOS] [Fixed] - Fixes passing a null/undefined callback to BackHandler hardwareBackPressed not throwing an error on until it is executed

## Test Plan

Create a BackHandler and attach a listener to 'hardwareBackPress' with a `null`/`undefined` callback, it should now throw an invariant exception


dulmandakh (approved):
LGT @cpojer

cpojer:
Please use Flow/TypeScript for this sort of enforcement instead of relying on runtime typechecking.

PR #29533, opened 2020-07-30
Fix Pressables not being tappable in PanResponder Views on Android

hsource:
## Summary

This fix makes Touchables and Pressables touchable, even if they're inside another view with a PanResponder/GestureResponder that handles touches.

I have an app that had a `TouchableWithoutFeedback` inside of a `View` with a PanResponder attached. The TouchableWithoutFeedback received touches as expected on iOS, but not on Android.

### Debugging

1. Added example to Pressable
2. Compared behaviour on Android and iOS - identified the different events being emitted and set a debug in Pressability.js `onResponderTerminationRequest`
3. Looked at the root event causing the Pressable to lose control of touches ("topTouchMove")
4. Searched for that in Android code to identify the code path that was emitting it
5. Added console.logs to the Android `MotionEvent` handler

### Root cause

After debugging, I realized this is because on Android (tested on Pixel 2 device):

1. As soon as a touch starts, a `topTouchMove` event is emitted from native code, even if the user doesn't move their finger at all
2. The topTouchMove triggers the parent PanResponder
3. This causes the touch on the Pressable to be canceled

This is unintuitive. Instead, the `topTouchMove` event should only be emitted when some actual move happens, like on iOS.

### Fix

To fix this, I added code to only emit a touchMove event if the user's touch has moved at least 10 units away from where it started. This code was based on other code in Pressability that also use this threshold.

## Changelog

[Android] [Fixed] - Fix Pressables not being tappable in PanResponder Views on Android

## Test Plan

Added a test case in RNTesterApp's Pressable screen:

| iOS | Android (before) | Android (after) |
| --- | --- | --- |
| Pressable | Android before | Android after |

### DIFF
--- a/RNTester/js/examples/Pressable/PressableExample.js
+++ b/RNTester/js/examples/Pressable/PressableExample.js
@@ -261,35 +261,39 @@ function PressableInPanResponder() {
   const appendMessage = (message: string) => {
     const lastMessage =
       messages.current.length && messages.current[messages.current.length - 1];
-    if (message === lastMessage) return;
+    if (message === lastMessage) {
+      return; // Reduce duplicate touchMove messages
+    }
     messages.current = [...messages.current, message];
     forceUpdate();
   };
 
-  const panResponder = useMemo(() => {
-    return PanResponder.create({
-      onStartShouldSetPanResponder: (): boolean => {
-        appendMessage('onStartShouldSetPanResponder, returning false');
-        return false;
-      },
-      onStartShouldSetPanResponderCapture: (): boolean => {
-        appendMessage('onStartShouldSetPanResponderCapture, returning false');
-        return false;
-      },
-      onMoveShouldSetPanResponder: (): boolean => {
-        appendMessage('onMoveShouldSetPanResponder, returning true');
-        return true;
-      },
-      onMoveShouldSetPanResponderCapture: (): boolean => {
-        appendMessage('onMoveShouldSetPanResponderCapture, returning false');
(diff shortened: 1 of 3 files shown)


hsource:
On further experimenting, it looks like iOS has the same behaviour as the existing (pre-PR) Android behaviour. I'm going to close this. I think we're supposed to just put this logic in the PanResponder

seancheung:
Thanks for this!
I was experimenting every possible way to solve the "Pressable not working in PanResponder on Android" issue.

The solution is simple in js side:

```tsx
PanResponder.create({
  onMoveShouldSetPanResponder: Platform.select({
    default: () => true,
    android: (e: GestureResponderEvent, state: PanResponderGestureState) =>
      state.dx > 10 || state.dy > 10,
  }),
})
```

unit-002:
> ```tsx
> PanResponder.create({
>   onMoveShouldSetPanResponder: Platform.select({
>     default: () => true,
>     android: (e: GestureResponderEvent, state: PanResponderGestureState) =>
>       Math.abs(state.dx) > 10 || Math.abs(state.dy) > 10
>   })
> })
> ```

I would like to try this because I am having the same issue with onPress inside Animated view using pan responder but on vscode I am having error saying

`type annotation can only be used on typescript files`

Currently I am using .js file. How to change this to js format? thanks.

gaganpreet-itilite:
> 
@unit-002 
I am also facing the same issue. any luck?

PR #29538, opened 2020-08-01
Fix RNTester TurboModules loading

janicduplessis:
## Summary

It is now required to call RCTBridge.setRCTTurboModuleRegistry for turbo modules to work properly on iOS.

## Changelog

[Internal] [Fix] - Fix RNTester TurboModules loading

## Test Plan

Tested that images in RNTester now loads properly and Redbox module missing warning are gone.


janicduplessis:
cc @fkgozali

Issue #21718, opened 2018-10-11
[FlatList] Android: onMomentumScrollEnd doesn't trigger when use scrollToOffset


myl0204:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i5-5675R CPU @ 3.10GHz
      Memory: 60.77 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.1 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 5.6.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
    IDEs:
      Android Studio: 3.0 AI-171.4443003
      Xcode: 10.0/10A255 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.5.0 => 16.5.0
      react-native: 0.57.2 => 0.57.2

## Description
onMomentumScrollEnd doesn't trigger when use `scrollToOffset` in **Android**.

And in iOS it works.

also i find that other methods e.g. scrollToIndex, have the same issue.

## Reproducible Demo

```jsx
export default class App extends Component<Props> {
  renderItem = ({item: { source }}) => {
    return (
      <TouchableOpacity onPress={() => {this.scrollView.scrollToOffset({offset: 375, animated: true})}}>
        <Image
          source={{uri: source.uri}}
          style={{width: 375, height: 375}}
        />
      </TouchableOpacity>
    )
  }

  onMomentumScrollEnd = () => {
    console.log('scrollend')
  }

  render() {
    return (
      <FlatList
        style={{ flex: 1, backgroundColor: 'black' }}
        data={[
          { source: { uri: 'http://i.imgur.com/XP2BE7q.jpg' } },
          { source: { uri: 'http://i.imgur.com/5nltiUd.jpg' } },
          { source: { uri: 'http://i.imgur.com/6vOahbP.jpg' } },
          { source: { uri: 'http://i.imgur.com/kj5VXtG.jpg' } }
        ]}
[...]

PR #29550, opened 2020-08-02
WIP: (Android/ScrollView) fix onMomentumScrollEnd event calls

AntoineDoubovetzky:
## Summary

Hi!
First of all, thanks for you work 😃 

1) This PR fixes issue #21718: onMomentumScrollBegin/End not being called when calling scrollTo/scrollToEnd programmatically, while it is called on iOS.

I first called `handlePostTouchScrolling` in the `onAnimationStart` listener but then replaced it by calling the emitEvent methods directly in the `onAnimationStart` and `onAnimationEnd` directly, because we don't need the `Runnable` in this case since we have an `onAnimationEnd` listener.

2) Working on this I also noticed that onMomentumScrollEnd is being called 3 times since we wait for 3 stable frames (commit [03ef81bfa29dd1e896d58ee8a70a3df7db76d60d](https://github.com/facebook/react-native/commit/03ef81bfa29dd1e896d58ee8a70a3df7db76d60d)) so I suggested a fix.

I added a `mIsMomentumEndEventSent` to the Runnable to check if the `emitScrollMomentumEndEvent` method has already been called or not. I've thought about checking if `mPostTouchRunnable` is null or not, but I think that it is more explicit this way.

Please let me know if you need anything more 😃 

## Changelog

[ANDROID] [FIXED] - Fix onMomentumScrollBegin/End not being called when calling scrollTo/scrollToEnd programmatically
[ANDROID] [FIXED] - Avoid onMomentumScrollEnd being called 3 times

## Test Plan

To test both fixes, I initiated a new project and replaced the app with the following code:

```
import React, {useRef} from 'react';
import {
  SafeAreaView,
  ScrollView,
  Text,
  TouchableOpacity,
  View,
} from 'react-native';

const App = () => {
  const meaScrollViewRef = useRef(null);
  const scrollToFirstElement = () => {
    if (meaScrollViewRef.current) {
      meaScrollViewRef.current.scrollTo({x: 0, y: 0});
    }
  };
  const scrollToSecondElement = () => {
    if (meaScrollViewRef.current) {
      meaScrollViewRef.current.scrollToEnd();
    }
  };

  const onMomentumScrollEnd = ({nativeEvent}) => {
    console.log(
      'onMomentumScrollEnd',
      nativeEvent.contentOffset.x,
      nativeEvent.contentOffset.y,
    );
  };

  const onMomentumScrollBegin = ({nativeEvent}) => {
    console.log(
      'onMomentumScrollBegin',
      nativeEvent.contentOffset.x,
      nativeEvent.contentOffset.y,
    );
  };

  return (
    <>
      <SafeAreaView>
        <ScrollView
          horizontal={true}
          pagingEnabled
          onMomentumScrollEnd={onMomentumScrollEnd}
          onMomentumScrollBegin={onMomentumScrollBegin}
          ref={meaScrollViewRef}>
          <View style={{width: 300, height: 400, backgroundColor: 'red'}}>
            <TouchableOpacity onPress={scrollToSecondElement}>
              <Text>Scroll to second element</Text>
            </TouchableOpacity>
          </View>
          <View style={{width: 300, height: 400, backgroundColor: 'red'}}>
            <TouchableOpacity onPress={scrollToFirstElement}>
              <Text>Scroll to first element</Text>
            </TouchableOpacity>
          </View>
        </ScrollView>
      </SafeAreaView>
    </>
  );
};

export default App;
```

And then simply change `horizontal` prop to test in horizontal and vertical mode.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/config/ReactFeatureFlags.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/config/ReactFeatureFlags.java
@@ -44,12 +44,6 @@ public class ReactFeatureFlags {
   /** This feature flag enables logs for Fabric */
   public static boolean enableFabricLogs = false;
 
-  /**
-   * Temporary feature flat to control a fix in the transition to layoutOnlyViews TODO T61185028:
-   * remove this when bug is fixed
-   */
-  public static boolean enableTransitionLayoutOnlyViewCleanup = false;
-
   /** Feature flag to configure eager initialization of Fabric */
   public static boolean eagerInitializeFabric = false;
 
--- a/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/uimanager/NativeViewHierarchyOptimizer.java
@@ -13,7 +13,6 @@ import com.facebook.common.logging.FLog;
 import com.facebook.infer.annotation.Assertions;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMapKeySetIterator;
-import com.facebook.react.config.ReactFeatureFlags;
 
 /**
  * Class responsible for optimizing the native view hierarchy while still respecting the final UI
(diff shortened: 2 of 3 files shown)


AntoineDoubovetzky:
Hi @mdvacca !

Is there anything I can do about the "Facebook Internal - Linter" check that failed ?

esinanturan:
@AntoineDoubovetzky I have tried this patch with 0.64 but I am still having issue that it get triggered 4 times

AntoineDoubovetzky:
> @AntoineDoubovetzky I have tried this patch with 0.64 but I am still having issue that its either getting called after a 5-7 seconds delay or it get triggered 4 times

I rebased the branch and it's unfortunately not working anymore. I'm working on it :)

PR #29567, opened 2020-08-04
Migrating RNTester to Packages Directory

stealthanthrax:
## Changelog

This PR aims to migrate the RNTester App to `packages` directory. But is currently, open to inspect the CI issues and resolve the merge conflicts.

Currently done
 - Working on iOS
 - Working on Android
 - Detox Tests working on iOS

Need to work on
 - Errors generated by the CI builds

[General] [Changed] - Migrated the RNTester App to the packages directory.

## Test Plan

It runs on both ios and android for now and the detox iOS builds are working.


cpojer:
You cannot call it "RNTester" if it is inside of the packages folder. It'll have to be either `packages/tester` or `packages/rn-tester` or maybe `packages/test-app` or similar.

stealthanthrax:
@cpojer , okay got it. Should I proceed with `packages/rn-tester` then?

stealthanthrax:
@cpojer , I've made the suggested changes. I can squash the commits if the migration part looks good.

packages/rn-tester/android/app/src/main/gen/com/facebook/react/uiapp/BuildConfig.java
```diff
@@ -0,0 +1,8 @@
+/*___Generated_by_IDEA___*/
```
rickhanlonii, line 1: What are these files?

packages/rn-tester/Podfile.lock
```diff
@@ -501,6 +501,33 @@ SPEC CHECKSUMS:
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
   glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
   OpenSSL-Universal: 8b48cc0d10c1b2923617dfe5c178aa9ed2689355
+<<<<<<< HEAD:packages/rn-tester/Podfile.lock
```
rickhanlonii, line 504: Merge conflict still in this file, doesn't this make the lockfile invalid?

packages/rn-tester/js/RNTesterApp.android.js
```diff
@@ -18,7 +18,7 @@ const RNTesterNavigationReducer = require('./utils/RNTesterNavigationReducer');
 const React = require('react');
 const URIActionMap = require('./utils/URIActionMap');
 
-const nativeImageSource = require('../../Libraries/Image/nativeImageSource');
+const nativeImageSource = require('../../../Libraries/Image/nativeImageSource');
```
rickhanlonii, line 21: This is a good example of the next step we'll take later to decouple RNTester from React Native. We should be able to set it up so that `react-native` resolves to `'../../../`, so that all of JavaScript references react-native, even if the native dependencies are still relative. Then we would work to treat the native deps the same way.

yarn.lock
```diff
@@ -2,310 +2,312 @@
 # yarn lockfile v1
```
rickhanlonii, line 2: Why did this change?

rickhanlonii:
Going to import this just to see what's failing internally

packages/rn-tester/android/app/src/main/gen/com/facebook/react/uiapp/BuildConfig.java
stealthanthrax: These weren't supposed to be committed. Removing them!

packages/rn-tester/Podfile.lock
stealthanthrax: Resolved them 😄

yarn.lock
stealthanthrax: Don't the `yarn.lock` file change automatically? 😕

rickhanlonii: Not if the dependencies don't change.

rickhanlonii:
Importing isn't working for me, could you try @cpojer?

PR #29594, opened 2020-08-08
Misspellings

kataras:
Most of them were documentation or method's inline code misspellings. However, the `ReactCommon/jsi/jsi/test/testlib.cpp #countOccurences` name was replaced with `countOccurrences`, it's not used though.

Thanks,
Gerasimos Maropoulos.


kataras:
I agreed to the CLA 10 minutes ago. Don't know why this @facebook-github-bot is slow but I still love it :heart:

Issue #27845, opened 2020-01-23
[0.62.0-rc.3] Catalyst macOS build issue typedef redefinition with different types


mgcrea:
React Native version:
```
System:
    OS: macOS 10.15.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 52.18 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.7.0 - /var/folders/2d/y3l4yzcj0pz9r62r554t2nqr0000gn/T/yarn--1579780584922-0.8762400095093377/node
    Yarn: 1.21.1 - /var/folders/2d/y3l4yzcj0pz9r62r554t2nqr0000gn/T/yarn--1579780584922-0.8762400095093377/yarn
    npm: 6.13.6 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 25, 26, 27, 28
      Build Tools: 23.0.1, 26.0.3, 27.0.3, 28.0.3
      System Images: android-28 | Intel x86 Atom, android-28 | Intel x86 Atom_64, android-28 | Google Play Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.11.0 => 16.11.0 
    react-native: 0.62.0-rc.1 => 0.62.0-rc.1
```

## Steps To Reproduce

1. Create a new project `npx react-native init SandboxProject_62 --version 0.62.0-rc.1`
2. Enable macOS target in xcode
3. Build

Describe what you expected to happen:

Successful build for macOS target

## Errors

```
[...]/SandboxProject_62/ios/Pods/Headers/Private/Flipper-Folly/folly/portability/Time.h:51:17:
Typedef redefinition with different types ('uint8_t' (aka 'unsigned char') vs 'enum clockid_t')
```

If I comment out this line, Flipper-Folly build does seem to pass but I end up with an OpenSSL linking error:

```
[...]

PR #29609, opened 2020-08-10
[Catalyst][Text] Fix Catalyst text rendering by disabling inappropriate subpixel-antialiasing

andymatuschak:
## Summary

I noticed when porting my iOS app to macOS via Catalyst that the text rendering was somewhat different on the two platforms. Text looked blurry and over-weight on macOS, even when disabling the Catalyst scaling transform.

I hazily remembered that I'd seen this problem before in my old Cocoa development days: this kind of blurring occurs when rendering text with sub-pixel anti-aliasing into an offscreen buffer which will then be traditionally composited, because when the SPAA algorithm attempts to blend with the underlying content (i.e. in the offscreen buffer), there isn't any. SPAA is disabled on iOS, so the issue wouldn't appear there. On macOS, typical approachs to displaying text (e.g. `CATextLayer`) normally disable SPAA, since it's been incompatible with the platform's compositing strategy since the transition to layer-backed views some years ago. But React Native uses `NSLayoutManager` to rasterize text (rather than relying on the system render server via `CATextLayer`), and that class doesn't touch the context's font smoothing bit before drawing.

This change makes macOS/Catalyst text rendering consistent with iOS text rendering by disabling SPAA.

It appears that the code I've modified is in the process of being refactored (for Fabric?). It looks like [this](https://github.com/facebook/react-native/blob/8d6b41e9bcede07fb627d57cf6c11050ae590d57/ReactCommon/react/renderer/textlayoutmanager/platform/ios/RCTTextLayoutManager.mm#L111) is the corresponding place in the new code (@sammy-SC, is that right?). I'm happy to include a change to the new renderer in this patch if someone can point me at how to test that change.

## Changelog

[iOS] [Fixed] - Improved text rendering on macOS Catalyst

## Test Plan

1. Prepare RNTester for running on macOS (or apply [this patch](https://gist.github.com/andymatuschak/d0f5b4fc1a28efc4f860801aa1deddcd) to handle parts 1 and 2, but you'll still need to do part 3):
   1. Open the workspace, navigate to the `RNTester` target's configuration, and check the "Mac" checkbox under "Deployment Info.
   2. Flipper doesn't yet compile for Catalyst (#27845), so you must disable it by: a) commenting out `use_flipper!` and `flipper_post_install` in the Podfile, then running `pod install`; and b) removing the `FB_SONARKIT_ENABLED` preprocessor flags in the Xcode project.
   3. macOS has different signing rules from iOS; you must set a development team in the "Signing & Capabilities" tab of the `RNTester` target configuration pane. Unfortunately, you must also do this in the `Pods` project for the `React-Core-AccessibilityResources` target ([this is an issue which CocoaPods must fix](https://github.com/CocoaPods/CocoaPods/issues/8891)).
2. Run RNTester with and without the patch. You'll see that the font hinting is overweight without the patch; see screenshots below (incorrect rendering above, correct rendering below; note that fonts still remain slightly blurred because of Catalyst's window scaling transform, but that's removed on Big Sur).

Screen Shot 2020-08-12 at 10 03 50 AM
Screen Shot 2020-08-12 at 10 03 15 AM

### DIFF
--- a/Libraries/Text/Text/RCTTextView.m
+++ b/Libraries/Text/Text/RCTTextView.m
@@ -105,6 +105,15 @@
   NSLayoutManager *layoutManager = _textStorage.layoutManagers.firstObject;
   NSTextContainer *textContainer = layoutManager.textContainers.firstObject;
 
+#if TARGET_OS_MACCATALYST
+  CGContextRef context = UIGraphicsGetCurrentContext();
+  UIGraphicsPushContext(context);
+  // NSLayoutManager tries to draw text with sub-pixel anti-aliasing by default on
+  // macOS, but rendering SPAA onto a transparent background produces poor results.
+  // CATextLayer disables font smoothing by default now on macOS; we follow suit.
+  CGContextSetShouldSmoothFonts(context, NO);
+#endif
+  
   NSRange glyphRange = [layoutManager glyphRangeForTextContainer:textContainer];
   [layoutManager drawBackgroundForGlyphRange:glyphRange atPoint:_contentFrame.origin];
   [layoutManager drawGlyphsForGlyphRange:glyphRange atPoint:_contentFrame.origin];
@@ -148,6 +157,10 @@
     [_highlightLayer removeFromSuperlayer];
     _highlightLayer = nil;
   }
+  
+#if TARGET_OS_MACCATALYST
+  UIGraphicsPopContext();
+#endif
 }
 
 
--- a/Libraries/Text/Text/RCTTextView.m
+++ b/Libraries/Text/Text/RCTTextView.m
@@ -107,7 +107,7 @@
 
 #if TARGET_OS_MACCATALYST
   CGContextRef context = UIGraphicsGetCurrentContext();
-  UIGraphicsPushContext(context);
+  CGContextSaveGState(context);
   // NSLayoutManager tries to draw text with sub-pixel anti-aliasing by default on
   // macOS, but rendering SPAA onto a transparent background produces poor results.
   // CATextLayer disables font smoothing by default now on macOS; we follow suit.
@@ -159,7 +159,7 @@
   }
   
 #if TARGET_OS_MACCATALYST
-  UIGraphicsPopContext();
+  CGContextRestoreGState();
 #endif
 }
 
--- a/Libraries/Text/Text/RCTTextView.m
+++ b/Libraries/Text/Text/RCTTextView.m
@@ -159,7 +159,7 @@
   }
   
 #if TARGET_OS_MACCATALYST
-  CGContextRestoreGState();
+  CGContextRestoreGState(context);
 #endif
 }
 
--- a/ReactCommon/react/renderer/textlayoutmanager/platform/ios/RCTTextLayoutManager.mm
+++ b/ReactCommon/react/renderer/textlayoutmanager/platform/ios/RCTTextLayoutManager.mm
@@ -108,9 +108,19 @@ static NSLineBreakMode RCTNSLineBreakModeFromEllipsizeMode(EllipsizeMode ellipsi
   NSLayoutManager *layoutManager = textStorage.layoutManagers.firstObject;
   NSTextContainer *textContainer = layoutManager.textContainers.firstObject;
 
+#if TARGET_OS_MACCATALYST
+  CGContextRef context = UIGraphicsGetCurrentContext();
+  CGContextSaveGState(context);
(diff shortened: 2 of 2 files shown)


andymatuschak:
Changed category to "iOS" from "Catalyst" per @pull-bot.

sammy-SC:
> It appears that the code I've modified is in the process of being refactored (for Fabric?). It looks like this is the corresponding place in the new code (@sammy-SC, is that right?). 


Hello @andymatuschak,

the code you have refactored will stay in place until Fabric is completely rolled out. Fabric on iOS is a complete rewrite.

You have found the corresponding place in Fabric :).

> I'm happy to include a change to the new renderer in this patch if someone can point me at how to test that change.

If you define `RN_FABRIC_ENABLED`, RNTester will start with the new renderer.

andymatuschak:
Thanks for that, @sammy-SC! I've just spent a while trying to get Fabric up and running (with iOS in general, not even on Catalyst). I faced immediate configuration issues which suggested to me that `master` may have some out of date Fabric components: several configuration files contain paths which are clearly wrong, and a dependency on `flow-node` was introduced without being added to the package.

I was able to make some progress by making the changes:
```
diff --git a/ReactCommon/react/renderer/graphics/React-graphics.podspec b/ReactCommon/react/renderer/graphics/React-graphics.podspec
index ddd519790e..07d54f808f 100644
--- a/ReactCommon/react/renderer/graphics/React-graphics.podspec
+++ b/ReactCommon/react/renderer/graphics/React-graphics.podspec
@@ -5,7 +5,7 @@
 
 require "json"
 
-package = JSON.parse(File.read(File.join(__dir__, "..", "..", "..", "package.json")))
+package = JSON.parse(File.read(File.join(__dir__, "..", "..", "..", "..", "package.json")))
 version = package['version']
 
 source = { :git => 'https://github.com/facebook/react-native.git' }
diff --git a/package.json b/package.json
index b9d85cdd85..99f3510d0f 100644
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "eslint-plugin-react-native": "3.8.1",
     "eslint-plugin-relay": "1.7.1",
     "flow-bin": "^0.131.0",
+    "flow-remove-types": "^2.131.0",
     "jest": "^26.0.1",
     "jest-junit": "^10.0.0",
     "jscodeshift": "^0.9.0",
diff --git a/scripts/react_native_pods.rb b/scripts/react_native_pods.rb
index 73bb9ad55a..534dd6d298 100644
--- a/scripts/react_native_pods.rb
+++ b/scripts/react_native_pods.rb
@@ -52,7 +52,7 @@ def use_react_native! (options={})
 
   if fabric_enabled
     pod 'React-Fabric', :path => "#{prefix}/ReactCommon"
-    pod 'React-graphics', :path => "#{prefix}/ReactCommon/fabric/graphics"
+    pod 'React-graphics', :path => "#{prefix}/ReactCommon/react/renderer/graphics"
     pod 'React-jsi/Fabric', :path => "#{prefix}/ReactCommon/jsi"
     pod 'React-RCTFabric', :path => "#{prefix}/React"
     pod 'Folly/Fabric', :podspec => "#{prefix}/third-party-podspecs/Folly.podspec"
```

But even after these patches, the React-graphics pod appears to have some misconfiguration issues. It fails to build with this error:
```
/Users/andym/Development/External/react-native/ReactCommon/react/renderer/graphics/platform/ios/Color.h:13:10: 'react/renderer/graphics/ColorComponents.h' file not found
```

… which makes sense, because the Pod's Xcode project is missing most its the source files, including the `RCTTextLayoutManager.mm` file I've modified:

Screen Shot 2020-08-12 at 9 57 25 AM

At this point, I stopped, because these problems don't appear to be local to my machine—it looks to me more like I'm simply not running known-good sources for Fabric.

I've updated the test plan in the pull request description to include the steps for running RNTester on macOS/Catalyst, and I've pushed a commit to my branch which implements my change for Fabric. If you'd be willing to test the change yourself or to point me to where I've gone wrong in my attempt above, I'd be grateful.

sammy-SC:
Hello @andymatuschak,

Re Fabric:
Apparently it isn't ready to be run in open source yet. I will import this PR into our systems and take it for a spin in Fabric to make sure it works.

Thank you

sammy-SC (approved):

andymatuschak:
Thank you for testing and landing this!

sammy-SC:
@andymatuschak thank you for working on this :)

PR #28914, opened 2020-05-18
Upgraded Gradle wrapper to 6.5.1

SaeedZhiany:
## Summary
Welcome to Gradle 6.5.1!

Here are the highlights of this release:
 - Experimental file-system watching
 - Improved version ordering
 - New samples

For more details see https://docs.gradle.org/6.5.1/release-notes.html

## Changelog

[Android] [Changed] - Update Gradle Wrapper to 6.5.1

## Test Plan

Build project and it should be built successfully.

PR #29613, opened 2020-08-11
Update Gradle Wrapper to 6.6

friederbluemle:
## Summary

Gradle Wrapper to 6.6

https://docs.gradle.org/6.6/release-notes.html

## Changelog

[Android] [Changed] - Update Gradle Wrapper to 6.6

## Test Plan

Build project

Closes #28914


mdvacca:
CC @kgozali

fkgozali:
we upgraded to 6.5 not too long ago, any specific benefit for this upgrade to 6.6?

fkgozali:
@friederbluemle @dulmandakh, thoughts on my question?

> we upgraded to 6.5 not too long ago, any specific benefit for this upgrade to 6.6?

friederbluemle:
@fkgozali - Thanks for the review/question.

There's no "specific benefit" for this update (i.e. it does not fix any major/blocking issue). However, the same question can be asked for the previous update to `6.5` (or other updates before that). I'm only following what seems to have been standard practice in this repo to always update Gradle Wrapper to the latest version.

Gradle 6.6 comes with some [potential performance improvements](https://docs.gradle.org/6.6/release-notes.html) (hard to measure those, and maybe negligible/non-existent in the React Native repo). Perhaps more importantly, Gradle 6.6 [fixes a whitespace error](https://github.com/gradle/gradle/pull/13293) in the `gradlew` script on line 133 that was **introduced** with Gradle 6.5 - Because this whitespace error is also present in the `template/` folder, it would be carried over into every single new React Native app created with `react-native init`.

All that said, if it was me, we should simply keep the version of the Gradle Wrapper **aligned with** the version of the Android Gradle plugin (update less often), **unless** there are specific reasons for a newer, out-of-sync version. Thoughts?

friederbluemle:
Rebased onto latest master branch.

fkgozali:
OK. The reason I asked is that, each upgrade is not as simple as just landing this PR. We have a bunch of offline caching in our internal CI that needs to be updated, and a bunch of other scripts as well. I wasn't involved in the upgrades pre-6.5 so I wasn't sure what the thoughts back then was, but moving forward we do want to be intentional when upgrading the build system, not just because it's the latest and the greatest.

That said, we may just take in this 6.6 soon (maybe in a week or so). Though further upgrades should probably be discussed briefly in the future.

friederbluemle:
@fkgozali - Sounds good, and makes sense! Thanks for the additional explanation :+1:

Like I said, in the future we could also consider simply keeping the version of Gradle Wrapper aligned with the default version used by the Android Gradle plugin, _unless_ there is a specific reason to upgrade.
Yes, this might involve a one-time _downgrade_ of Gradle Wrapper, since it will probably be quite a while until AGP defaults to Gradle 6.6+. In order to avoid a major version downgrade, a good potential opportunity would be once we move from AGP 3 to AGP 4 (#29013), since AGP 4.0.1 uses Gradle Wrapper 6.1.1.

friederbluemle:
Hi @fkgozali - Quick question: I noticed the rewritten commit from this PR that ended up on the master branch (5bc67b658e581e0176deb7ed95b51a5c1cbe65c2) has the `-bin` variant of Gradle Wrapper. This PR has `-all` (like before, and also what's the default for Android Studio). Do you happen to know why?

fkgozali:
> Hi @fkgozali - Quick question: I noticed the rewritten commit from this PR that ended up on the master branch ([5bc67b6](https://github.com/facebook/react-native/commit/5bc67b658e581e0176deb7ed95b51a5c1cbe65c2)) has the `-bin` variant of Gradle Wrapper. This PR has `-all` (like before, and also what's the default for Android Studio). Do you happen to know why?

I'm not too sure, I think I needed to run gradlew wrapper (to upgrade other scripts) and it might have produced the change. Regardless, it seems reasonable to use -bin (for less build-time download)? I noticed some gradlew used -all, some used -bin.

fkgozali:
According to https://docs.gradle.org/current/userguide/gradle_wrapper.html:

> The type of Gradle distribution. By default that’s the -bin distribution containing only the runtime but no sample code and documentation.

So it seems ok to rely on -bin?

friederbluemle:
Ooh, okay, so I guess you ran `./gradlew wrapper` without the `--distribution-type all` flag. It also happened to me before, which is why I defined an alias in my shell to make this easier to update Gradle Wrapper across many repositories that involve Android.
`-all` is actually the **default** as generated by Android Studio, and it also provides **additional sources** helpful for debugging. It's also much more likely to already exist on a developer's machine (which has built other Android projects), avoiding additional downloads and saving disk space. Therefore, the `-all` variant is preferred.
I opened a PR here: #29793

fkgozali:
I see, I'll merge that tomorrow/early next week

On Aug 27, 2020, at 5:04 PM, Frieder Bluemle <notifications@github.com> wrote:

﻿

Ooh, okay, so I guess you ran ./gradlew wrapper without the --distribution-type all flag. It also happened to me before, which is why I defined an alias in my shell to make this easier to update Gradle Wrapper across many repositories that involve Android.
-all is actually the default as generated by Android Studio, and it also provides additional sources helpful for debugging. It's also much more likely to already exist on a developer's machine (which has built other Android projects), avoiding additional downloads and saving disk space. Therefore, the -all variant is preferred.
I opened a PR here: #29793<https://github.com/facebook/react-native/pull/29793>

friederbluemle:
Sounds good, thanks for the fast replies @fkgozali :+1:

mikehardy:
When my clone of react-native includes this PR, I have a persistent line ending issue with one of the gradlew.bat files, it's resistant to all my attempts to fix so far, on ubuntu 20

```

friederbluemle:
Hi @mikehardy - Yeah, unfortunately that is a problem for me as well (I'm surprised not more people have pointed this out)..

I've already opened a PR with a brief description more than a week ago: #29792
Hopefully it can get merged soon.

PR #29622, opened 2020-08-11
Fix Xcode warnings in React-Core pod

jdeff:
## Summary

With the upgrade to React Native 0.63, we started running into nullability warnings that were breaking our build. This PR fixes those nullability warnings as well as a few other warnings in React-Core.

## Changelog

[iOS] [Fixed] - Fix xcodebuild warnings in React-Core

## Test Plan

- Nullability annotations should only affect compilation, but even though RNTester compiles, I'm not fully convinced that this won't break projects downstream. It would be good to get another opinion on this.
- The change in `RCTAllocateRootViewTag` is the only real logic change in this PR. We throw an exception if the root view tag is not in the correct format, so this change seems safe after some basic manual testing in RNTester.

### DIFF
--- a/Libraries/Image/RCTImageLoaderProtocol.h
+++ b/Libraries/Image/RCTImageLoaderProtocol.h
@@ -14,6 +14,8 @@
 #import <React/RCTImageURLLoader.h>
 #import <React/RCTImageCache.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
  * If available, RCTImageRedirectProtocol is invoked before loading an asset.
  * Implementation should return either a new URL or nil when redirection is
@@ -132,3 +134,5 @@ typedef NS_ENUM(NSUInteger, RCTImageLoaderPriority) {
 - (void)setImageCache:(id<RCTImageCache>)cache;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/React/Base/RCTBridge.m
+++ b/React/Base/RCTBridge.m
@@ -207,7 +207,10 @@ RCT_NOT_IMPLEMENTED(-(instancetype)init)
 
 - (void)didReceiveReloadCommand
 {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
   [self reloadWithReason:@"Command"];
+#pragma clang diagnostic pop
 }
 
 - (NSArray<Class> *)moduleClasses
--- a/React/Base/RCTBridge.m
+++ b/React/Base/RCTBridge.m
@@ -207,10 +207,26 @@ RCT_NOT_IMPLEMENTED(-(instancetype)init)
 
 - (void)didReceiveReloadCommand
 {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-  [self reloadWithReason:@"Command"];
-#pragma clang diagnostic pop
+  #if RCT_ENABLE_INSPECTOR
+    // Disable debugger to resume the JsVM & avoid thread locks while reloading
+    [RCTInspectorDevServerHelper disableDebugger];
+  #endif
+
+  [[NSNotificationCenter defaultCenter] postNotificationName:RCTBridgeWillReloadNotification
+                                                      object:self
+                                                    userInfo:nil];
+
+  /**
+   * Any thread
+   */
+  dispatch_async(dispatch_get_main_queue(), ^{
+    // WARNING: Invalidation is async, so it may not finish before re-setting up the bridge,
+    // causing some issues. TODO: revisit this post-Fabric/TurboModule.
+    [self invalidate];
+    // Reload is a special case, do not preserve launchOptions and treat reload as a fresh start
+    self->_launchOptions = nil;
+    [self setUp];
+  });
 }
 
 - (NSArray<Class> *)moduleClasses
@@ -261,7 +277,7 @@ RCT_NOT_IMPLEMENTED(-(instancetype)init)
  */
 - (void)reload
 {
-  [self reloadWithReason:@"Unknown from bridge"];
+  RCTTriggerReloadCommandListeners(@"Unknown from bridge");
 }
 
 /**
@@ -269,24 +285,7 @@ RCT_NOT_IMPLEMENTED(-(instancetype)init)
  */
 - (void)reloadWithReason:(NSString *)reason
 {
-#if RCT_ENABLE_INSPECTOR
-  // Disable debugger to resume the JsVM & avoid thread locks while reloading
-  [RCTInspectorDevServerHelper disableDebugger];
-#endif
-
-  [[NSNotificationCenter defaultCenter] postNotificationName:RCTBridgeWillReloadNotification object:self userInfo:nil];
-
-  /**
-   * Any thread
-   */
-  dispatch_async(dispatch_get_main_queue(), ^{
-    // WARNING: Invalidation is async, so it may not finish before re-setting up the bridge,
-    // causing some issues. TODO: revisit this post-Fabric/TurboModule.
-    [self invalidate];
-    // Reload is a special case, do not preserve launchOptions and treat reload as a fresh start
-    self->_launchOptions = nil;
-    [self setUp];
-  });
+  RCTTriggerReloadCommandListeners(reason);
 }
 
 - (void)onFastRefresh
--- a/React/Base/RCTJSStackFrame.m
+++ b/React/Base/RCTJSStackFrame.m
@@ -57,7 +57,7 @@ static NSRegularExpression *RCTJSStackFrameRegex()
                               file:(NSString *)file
                         lineNumber:(NSInteger)lineNumber
                             column:(NSInteger)column
-                          collapse:(NSInteger)collapse
+                          collapse:(BOOL)collapse
 {
   if (self = [super init]) {
     _methodName = methodName;
@@ -100,7 +100,7 @@ static NSRegularExpression *RCTJSStackFrameRegex()
                                      file:file
                                lineNumber:[lineNumber integerValue]
                                    column:[column integerValue]
-                                 collapse:@NO.integerValue];
+                                 collapse:NO];
 }
 
 + (instancetype)stackFrameWithDictionary:(NSDictionary *)dict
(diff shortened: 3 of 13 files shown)


Libraries/Image/RCTImageLoaderProtocol.h
```diff
@@ -14,6 +14,8 @@
 #import <React/RCTImageURLLoader.h>
 #import <React/RCTImageCache.h>
 
+NS_ASSUME_NONNULL_BEGIN
```
jdeff, line 17: Fixes these warnings:

```
Pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified)
```

RNTester/Podfile.lock
```diff
@@ -501,31 +501,31 @@ SPEC CHECKSUMS:
...
+  React-RCTLinking: bc76da619cb7ce6a81e87118f4d952751a19c71c
+  React-RCTNetwork: d7de48eef44f09a71b501eabf7ffbdd65dcf1fee
+  React-RCTPushNotification: 204a642ed98ad336f97b524022715849d65148c9
+  React-RCTSettings: 2d7a13e63fb9b3cb4cbcbd527442af7b3f142c72
+  React-RCTTest: e40fb7d9a3762feba08f25b7dccf0bc2de8370b4
+  React-RCTText: fecc00cae8427e5869445f09c85f33efe0a3f182
+  React-RCTVibration: c6bac747411359e5ee7e590b2dc8939e91c67b5a
+  React-runtimeexecutor: 533b76c7a01b705c651b57838439c5ba96fea654
+  ReactCommon: f17474a4663f81d00462cbe9c944d63fcae3e708
+  Yoga: a98571eb8a4bf47f31675980b61342f9f16b82fe
```
jdeff, line 528: All of these changes came if after running `bundle exec pod install` on `master`.

React/Modules/RCTUIManagerUtils.m
```diff
@@ -98,6 +99,6 @@ void RCTUnsafeExecuteOnUIManagerQueueSync(dispatch_block_t block)
 NSNumber *RCTAllocateRootViewTag()
 {
   // Numbering of these tags goes from 1, 11, 21, 31, ..., 100501, ...
-  static int64_t rootViewTagCounter = -1;
-  return @(OSAtomicIncrement64(&rootViewTagCounter) * 10 + 1);
```
jdeff, line 102: Fixes this warning:

```
'OSAtomicIncrement64' is deprecated: first deprecated in iOS 10.0 - Use atomic_fetch_add_explicit(memory_order_relaxed) from <stdatomic.h> instead
```

React/Base/RCTKeyCommands.m
```diff
@@ -113,7 +113,7 @@ + (void)initialize
 - (void)handleKeyUIEventSwizzle:(UIEvent *)event
 {
   NSString *modifiedInput = nil;
-  UIKeyModifierFlags *modifierFlags = nil;
+  UIKeyModifierFlags modifierFlags = 0;
```
jdeff, line 116: Fixes this warning:

```
Incompatible integer to pointer conversion assigning to 'UIKeyModifierFlags *' (aka 'enum UIKeyModifierFlags *') from 'UIKeyModifierFlags' (aka 'enum UIKeyModifierFlags')
```

React/Base/RCTModuleData.h
```diff
@@ -97,4 +97,4 @@ typedef id<RCTBridgeModule> (^RCTBridgeModuleProvider)(void);
 @end
 
 RCT_EXTERN void RCTSetIsMainQueueExecutionOfConstantsToExportDisabled(BOOL val);
-RCT_EXTERN BOOL RCTIsMainQueueExecutionOfConstantsToExportDisabled();
+RCT_EXTERN BOOL RCTIsMainQueueExecutionOfConstantsToExportDisabled(void);
```
jdeff, line 100: Fixes this warning:

```
This function declaration is not a prototype
```

safaiyeh:
cc @alloy

RNTester/Podfile.lock
alloy: Hmm, can you check when it was last updated and what the difference between then and now are f these podspecs?

React/Base/RCTBridge.m
```diff
@@ -207,7 +207,10 @@ - (void)setRCTTurboModuleRegistry:(id<RCTTurboModuleRegistry>)turboModuleRegistr
 
 - (void)didReceiveReloadCommand
 {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
   [self reloadWithReason:@"Command"];
+#pragma clang diagnostic pop
```
alloy, line 213: Do you know if this is being tracked in an issue? Seeing as it will now go silent.

React/Base/RCTJSStackFrame.m
```diff
@@ -100,7 +100,7 @@ + (instancetype)stackFrameWithLine:(NSString *)line
                                      file:file
                                lineNumber:[lineNumber integerValue]
                                    column:[column integerValue]
-                                 collapse:@NO];
+                                 collapse:@NO.integerValue];
```
alloy, line 103: This seems a smell. Can the `collapse` var be made a `BOOL` instead, at least internally to this class? If not, then at least it seems better to just pass `0`.

React/CxxBridge/RCTCxxBridge.mm
```diff
@@ -1125,6 +1128,8 @@ - (void)reloadWithReason:(NSString *)reason
   RCTTriggerReloadCommandListeners(reason);
 }
 
+#pragma clang diagnostic pop
```
alloy, line 1131: Same as before, this should be tracked in some way. Perhaps the alternative to an issue is to leave comments pointing to these call-sites from the deprecated implementation?

React/Modules/RCTUIManagerUtils.m
```diff
@@ -8,6 +8,7 @@
 #import "RCTUIManagerUtils.h"
 
 #import <libkern/OSAtomic.h>
```
alloy, line 10: Can this import go too?

React/Modules/RCTUIManagerUtils.m
```diff
@@ -98,6 +99,6 @@ void RCTUnsafeExecuteOnUIManagerQueueSync(dispatch_block_t block)
 NSNumber *RCTAllocateRootViewTag()
 {
   // Numbering of these tags goes from 1, 11, 21, 31, ..., 100501, ...
-  static int64_t rootViewTagCounter = -1;
-  return @(OSAtomicIncrement64(&rootViewTagCounter) * 10 + 1);
+  static _Atomic int64_t rootViewTagCounter = 0;
```
alloy, line 102: From [the docs](https://en.cppreference.com/w/c/atomic/atomic_fetch_add):

> Atomically replaces the value pointed by obj with the result of addition of arg to the old value of obj, and **returns the value obj held previously**.

So that’s why this should initialise with `0` instead of `-1`, like previously was the case.

Do you know if there are tests that cover this API?

React/Profiler/RCTProfile.m
```diff
@@ -378,7 +378,10 @@ + (void)vsync:(CADisplayLink *)displayLink
 
 + (void)reload
 {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
   [RCTProfilingBridge() reloadWithReason:@"Profiling controls"];
+#pragma clang diagnostic pop
```
alloy, line 384: Ditto.

.gitignore
```diff
@@ -88,6 +88,7 @@ RNTester/build
 /template/ios/Pods/
 /template/ios/Podfile.lock
 /RNTester/Gemfile.lock
+/RNTester/vendor
```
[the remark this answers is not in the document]
jdeff: The `vendor` directory came for running `bundle install --path vendor`, which I did to avoid installing the gems as system gems. I can remove it from this PR. It would be nice to have some better instructions on how to set up the development environment for newbies.

React/Base/RCTJSStackFrame.m
jdeff: Agreed. Shouldn't be a huge change.

RNTester/Podfile.lock
jdeff: So it looks like all of these specs are local (e.g `FBLazyVector (from \`../Libraries/FBLazyVector\`)` so it would make sense that any change in these directories would cause the checksum to be different. I don't think this is a necessary change in this PR. I'll just remove it.

React/Modules/RCTUIManagerUtils.m
jdeff: Good catch. Fixed.

React/Modules/RCTUIManagerUtils.m
jdeff: I don't know of any tests per se, but the RNTester app crashed immediately when this was initialized to `-1`.

React/Base/RCTJSStackFrame.m
jdeff: I changed it to `BOOL`.

React/Base/RCTBridge.m
jdeff: Thanks for calling this out. I went ahead and took a stab at moving off these deprecated methods. Let me know if I should revert that and just file an issue instead.

React/CxxBridge/RCTCxxBridge.mm
jdeff: Since `RCTCxxBridge` inherits from `RCTBridge`, I went ahead and just removed these overrides, since `RCTBridge` now has the same implementation as this (minus the warn log).

jdeff:
Thanks for the review @alloy. I think I've resolved all of the comments.

alloy (approved):
Great, thanks for your work!

/cc @TheSavior

React/Base/Surface/SurfaceHostingView/RCTSurfaceHostingProxyRootView.h
```diff
@@ -36,6 +36,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign) NSTimeInterval loadingViewFadeDelay;
 @property (nonatomic, assign) NSTimeInterval loadingViewFadeDuration;
 
+- (instancetype)initWithSurface:(RCTSurface *)surface
+                sizeMeasureMode:(RCTSurfaceSizeMeasureMode)sizeMeasureMode NS_UNAVAILABLE;
```
appden, line 40: @jdeff Sorry, I had to remove this one before merging because it caused issues with Facebook's internal builds.

alloy: @appden Can you share details about the build failure?

Issue #29336, opened 2020-07-10
Undefined symbols for architecture arm64 - FlipperKitNetworkPlugin


ghost:
## Description
After using the upgrade helper, in order to update to the last version of RN (0.63.0) I had some build errors related to Flipper.

Upgrade helper: https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.0

## React Native version:
```
System:
    OS: macOS 10.15.5
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 1.12 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.1 - /Users/barrosc/.rvm/gems/ruby-2.7.0/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_211 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.0 => 0.63.0 
```

## Steps To Reproduce

1. ios folder: pod install
2. root folder: react-native run-ios

## Expected Results
Expected to build and successfully launch the app

## Snack, code example, screenshot, or link to a repository:

**Build Settings**

Valid Architectures:
`arm64 arm64e armv7 armv7s`

**Podfile**

```
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

platform :ios, '10.0'

target 'my_app' do
  config = use_native_modules!

  use_react_native!(:path => config["reactNativePath"])
  
  target 'app_tests' do
    inherit! :complete
    # Pods for testing
  end

  # Enables Flipper.
  #
[...]

PR #29646, opened 2020-08-14
Wrap iOS Flipper code with #if

safaiyeh:
## Summary
Fixes https://github.com/facebook/react-native/issues/29336

Flipper should only be instantiated when in DEBUG regardless if FB_SONARKIT_ENABLED is enabled.

## Changelog

[iOS] [Changed] - Make sure Flipper only instantiated in DEBUG mode

## Test Plan

Built the template app and Flipper integration still works

### DIFF
--- a/template/ios/HelloWorld/AppDelegate.m
+++ b/template/ios/HelloWorld/AppDelegate.m
@@ -4,6 +4,7 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
 
+#if DEBUG
 #ifdef FB_SONARKIT_ENABLED
 #import <FlipperKit/FlipperClient.h>
 #import <FlipperKitLayoutPlugin/FlipperKitLayoutPlugin.h>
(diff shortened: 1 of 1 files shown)


template/ios/HelloWorld/AppDelegate.m
```diff
@@ -4,6 +4,7 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
 
+#if DEBUG
```
gedeagas, line 7: I didn't release this was deleted. no wonder my react-native 63 builds failed.

safaiyeh:
@alloy I feel like the `FB_SONARKIT_ENABLED` variable is a bit unclear to user land. The only use for it is for Flipper. While `DEBUG` makes more sense, as you wouldn't want Flipper integrated to your production app.

alloy:
What about renaming the macro to eg `FLIPPER_ENABLED`?

safaiyeh:
I like that better!

alshamiri5 (approved):

PR #29683, opened 2020-08-18
Allow PlatformColor to return user-defined named asset color

danilobuerger:
## Summary

According to https://reactnative.dev/docs/platformcolor:

> You can use the PlatformColor function to access native colors on the target platform by supplying the native color’s corresponding string value.

This should also include named asset colors as they are part of the colors available on the iOS platform.

See 
https://developer.apple.com/documentation/uikit/uicolor/2877380-colornamed?language=objc
https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_ref-Asset_Catalog_Format/Named_Color.html

This PR allows using `PlatformColor` to get a user-defined named asset color in the same manner that it works on android via xml resources:

Screenshot 2020-10-05 at 12 45 33

```ts
<View style={{ backgroundColor: PlatformColor("MyCustomColor") }}>...</View>
```

would set the background color of the view to the named asset color "MyCustomColor". Currently this doesn't work.

## Changelog

[iOS] [Added] - Allow PlatformColor to return user-defined named asset color

## Test Plan

1. Define an asset color as specified here: https://developer.apple.com/documentation/xcode/supporting_dark_mode_in_your_interface?language=objc

2. Access the color via `PlatformColor("asset color name")`

### DIFF
--- a/React/Base/RCTConvert.m
+++ b/React/Base/RCTConvert.m
@@ -863,7 +863,11 @@ static NSString *RCTSemanticColorNames()
         return color;
       } else if ([value isKindOfClass:[NSArray class]]) {
         for (id name in value) {
-          UIColor *color = RCTColorFromSemanticColorName(name);
+          UIColor *color = [UIColor colorNamed:name];
+          if (color != nil) {
+            return color;
+          }
+          color = RCTColorFromSemanticColorName(name);
           if (color != nil) {
             return color;
           }
--- a/React/Base/RCTConvert.m
+++ b/React/Base/RCTConvert.m
@@ -850,7 +850,11 @@ static NSString *RCTSemanticColorNames()
     if ((value = [dictionary objectForKey:@"semantic"])) {
       if ([value isKindOfClass:[NSString class]]) {
         NSString *semanticName = value;
-        UIColor *color = RCTColorFromSemanticColorName(semanticName);
+        UIColor *color = [UIColor colorNamed:semanticName];
+        if (color != nil) {
+          return color;
+        }
+        color = RCTColorFromSemanticColorName(semanticName);
         if (color == nil) {
           RCTLogConvertError(
               json,


alshamiri5 (approved):

danilobuerger:
Hi @shergin 👋 I would love to move this PR forward. Is there anything missing in order to get it merged?

zachgibson:
@danilobuerger how do you supply custom Android colors?

TheSavior:
The examples for PlatformColor in RNTester have tests for this behavior on Android. Can you add some examples to iOS to ensure that this keeps working?

React/Base/RCTConvert.m
```diff
@@ -850,7 +850,11 @@ + (UIColor *)UIColor:(id)json
     if ((value = [dictionary objectForKey:@"semantic"])) {
       if ([value isKindOfClass:[NSString class]]) {
         NSString *semanticName = value;
-        UIColor *color = RCTColorFromSemanticColorName(semanticName);
+        UIColor *color = [UIColor colorNamed:semanticName];
```
appden, line 853: This API was officially added for iOS 11, but we still officially support iOS 10+. This should probably be guarded by `if (@available(iOS 11.0, tvOS 11.0, *))`.

appden (changes requested):

React/Base/RCTConvert.m
appden: @danilobuerger Following up to see if you can update this. Thanks!

danilobuerger: @appden This seems no longer neccessary with https://github.com/facebook/react-native/commit/a1d626739d95d6cbbb1be169b93952cdd1465486 ?

danilobuerger:
@TheSavior You asked me the same thing in another PR (#29728) which I happily complied with and invested time in. Now that PR has again not seen any progress in nearly 3 months. So I wonder if I really should invest time again to add this here if my PRs just don't get merged. I started both PRs in August, now its January. It feels disrespectful to me if I invest all that time and effort into making PRs, but then they never get merged or answered in a reasonable time frame by the maintainers.

oblador:
@danilobuerger Can you try rebasing this branch? I suspect the failing `Facebook Internal` tests are due to the iOS 10 issues.

PR #29684, opened 2020-08-18
Fix "'RCTBlobPlugins.h' file not found" on iOS

tido64:
## Summary

Fixes iOS builds failing with the following error in certain setups:

```
/~/Pods/React-RCTBlob/Libraries/Blob/RCTBlobPlugins.mm:14:9: fatal error: 'RCTBlobPlugins.h' file not found
#import "RCTBlobPlugins.h"
        ^~~~~~~~~~~~~~~~~~
1 error generated.
```

## Changelog

[iOS] [Fixed] - "'RCTBlobPlugins.h' file not found" when building iOS

## Test Plan

- Existing builds should not fail
- Tested this change internally where we are seeing this build failure

cc @hramos


alloy (approved):
Yup, this is good 👍 

/cc @appden

PR #29687, opened 2020-08-18
Align some more of the publish/pr logic on android builds

acoates-ms:
Some build logic was added to the PRs as part of the 62 merge, but not added to publish.  This aligns more of the build logic between the two builds


acoates-ms:
OOpps.. meant for react-native-macos.

PR #29711, opened 2020-08-20
[ios] Do not override decoders to RCTImageLoader

sjchmiela:
## Summary

I (actually, [we](https://github.com/expo/expo/issues/9858)) noticed GIFs are no longer animating in Expo client after [enabling TurboModules](https://github.com/expo/expo/pull/9687).

## Changelog

[iOS] [Fixed] - Fix `RCTImageLoader` not using decoders provided.

## Test Plan

cat

The cat is moving! Before applying this commit `RCTGIFDecoder` provided in the initalizer is removed from the `_decoders` array in the

```objc
_decoders = [_bridge modulesConformingToProtocol:@protocol(RCTImageDataDecoder)];
```

Also, compare 

https://github.com/facebook/react-native/blob/8f306cd66a8bc6054ee13701f02329ab5817b69a/Libraries/Image/RCTImageLoader.mm#L243-L250

and

https://github.com/facebook/react-native/blob/8f306cd66a8bc6054ee13701f02329ab5817b69a/Libraries/Image/RCTImageLoader.mm#L177-L184

This PR makes `_decoders` behave the same as `_loaders`.

### DIFF
--- a/Libraries/Image/RCTImageLoader.mm
+++ b/Libraries/Image/RCTImageLoader.mm
@@ -247,7 +247,7 @@ RCT_EXPORT_MODULE()
       _decoders = [_bridge modulesConformingToProtocol:@protocol(RCTImageDataDecoder)];
     }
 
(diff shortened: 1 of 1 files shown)


jinshin1013:
Any progress on this guys?

janicduplessis:
cc @sammy-SC @fkgozali

fkgozali:
Thanks for the fix, good catch. I just merged this.

PR #29736, opened 2020-08-23
RN(View): Remove TextAncestor provider

ecreeth:
## Summary

I think the `TextAncestor` provider is unnecessary in the `View` component as the [renderer](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js#L4039)
will throw an error if it doesn't find the value in the context anyway.

## Changelog

[JavaScript] [Changed] - Remove TextAncestor provider for `View`

## Test Plan
- yarn test passed
- RNTester `Android` app build successful

- A `View` component with text must throw error: `Text strings must be rendered within a <Text> component.`
```jsx
<View>Hello error</View>
```

Below we can see that all the providers of the `View` component have been eliminated from the component hierarchy

| Before |  After   |
|----------|:-------------:|
| image | image |

### DIFF
--- a/Libraries/Components/View/View.js
+++ b/Libraries/Components/View/View.js
@@ -14,7 +14,6 @@ import type {ViewProps} from './ViewPropTypes';
 
 const React = require('react');
 import ViewNativeComponent from './ViewNativeComponent';
-const TextAncestor = require('../../Text/TextAncestor');
 
 export type Props = ViewProps;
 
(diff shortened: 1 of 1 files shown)


TheSavior:
Resetting the text ancestor is required to change what kinds of Text components render (inline text nodes, or block level text nodes). We need to reset that in View as text elements inside of a View need to reset to be a block level text, not an inline text node.

PR #29745, opened 2020-08-24
Update RCTModalHostViewManager.m

Mi-ZAZ:
Fix a issue that RCTModalHostView can't be dismissed while being presented

## Summary

Steps To Reproduce
A native modal presented view controller is being dismissed before the first screen shows.
A RCTModalHostView is presented when the first screen shows.
The RCTModalHostView will be dismissed after an asynchronous callback.
If the callback was called before the completion of the presenting animation, the RCTModalHostView will not be dismissed.

## Changelog

[iOS] [Fixed] - Fix that RCTModalHostView can't be dismissed while being presented

### DIFF
--- a/React/Views/RCTModalHostViewManager.m
+++ b/React/Views/RCTModalHostViewManager.m
@@ -99,7 +99,7 @@ RCT_EXPORT_MODULE()
     _dismissalBlock([modalHostView reactViewController], viewController, animated, nil);
   } else {
     self.dismissWaitingBlock = ^{
-      [viewController.presentingViewController dismissViewControllerAnimated:animated completion:completionBlock];
+      [viewController.presentingViewController dismissViewControllerAnimated:animated completion:nil];
     };
     if (viewController.presentingViewController) {
       self.dismissWaitingBlock();
--- a/React/Views/RCTModalHostViewManager.m
+++ b/React/Views/RCTModalHostViewManager.m
@@ -47,6 +47,8 @@ RCT_ENUM_CONVERTER(
 @end
 
 @interface RCTModalHostViewManager () <RCTModalHostViewInteractor>
+  
+@property (nonatomic, copy) dispatch_block_t dismissWaitingBlock;
 
 @end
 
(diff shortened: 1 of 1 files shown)


sammy-SC:
Hello @Mi-ZAZ,

there are 116 files changed in this PR, is that expected?

Mi-ZAZ:
> Hello @Mi-ZAZ,
> 
> there are 116 files changed in this PR, is that expected?

Hi,
No. After rebase facebook/master, a commit that changed those files was added to the PR.
I have just rebased again. There are 1 file changed now.

sammy-SC:
@Mi-ZAZ 

The code looks good. Could you please provide a test plan? I would like to make sure it works as intended before merging it.

Thank you

Mi-ZAZ:
Sorry. After hours of trying, I still can't reproduce the issue with RNTester.

Our application uses react-native in some dynamic pages.
Before the react-native screen shows, there are several native screen and some animation.

We found that if a Modal component was called to be dismissed while it was being presented, the presentingViewController property of the RCTModalHostViewController instance will be nil.
And the property will be set after the presenting animation finished.

I have repeat the actions above, but with RNTester the presentingViewController property was not nil.

sammy-SC (approved):
@Mi-ZAZ 

right, I gave this some thought and I see how this actually can happen.

React/Views/RCTModalHostViewManager.m
```diff
@@ -92,7 +98,13 @@ - (void)dismissModalHostView:(RCTModalHostView *)modalHostView
   if (_dismissalBlock) {
     _dismissalBlock([modalHostView reactViewController], viewController, animated, nil);
   } else {
-    [viewController.presentingViewController dismissViewControllerAnimated:animated completion:nil];
+    self.dismissWaitingBlock = ^{
+      [viewController.presentingViewController dismissViewControllerAnimated:animated completion:nil];
```
sammy-SC, line 102: one idea:

Could we just call `[viewController dismissViewControllerAnimated:completion:]` instead of using `presentingViewController`? 

I can't test it myself since I can't reproduce it.

sammy-SC, line 102: Can we try calling `dismissViewControllerAnimated:completion` on viewController instead of `presentingViewController`? I would try it myself but I can't reproduce the issue.

Mi-ZAZ: I have tried calling `dismissViewControllerAnimated:completion:` on viewController, but it can't fix this issue.
As long as the `presentingViewController` of viewController was nil, dismissing will fail.

sammy-SC: thanks.

Issue #29303, opened 2020-07-08
`global.Promise` replacement in Jest setup causing problems.


chrisbobbe:
Hi! 👋

What is the purpose of replacing the global Promise in [this line](https://github.com/facebook/react-native/blob/v0.60.6/jest/setup.js#L20) of `jest/setup.js`:

```javascript
global.Promise = jest.requireActual('promise');
```

It looks like it was first introduced in https://github.com/facebook/react-native/commit/3ff39870ce776c653823e1733363be0401896294 (quite a while ago) and has remained in a similar form since then. There's no explanation of this choice specifically (if I understand correctly), so I think if it must remain, it would be good to put a code comment there. 🙂 

## Description

'promise' refers to [the NPM package with that name](https://www.npmjs.com/package/promise).

It interferes with Jest when using their new "modern" implementation of fake timers (documented [here](https://jestjs.io/docs/en/jest-object#jestusefaketimersimplementation-modern--legacy)), as I describe at https://github.com/facebook/jest/issues/10221.

One interesting note on that issue, from a Jest maintainer, is [this](https://github.com/facebook/jest/issues/10221#issuecomment-654687396):
> As an aside, one should never replace `global.Promise` [...]. E.g. when using `async-await` you will always get the native `Promise` regardless of the value of `global.Promise`.

Might it be OK and appropriate to remove the line?

## React Native version:

```
System:
    OS: macOS 10.15.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 4.84 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.20.1 - ~/.nvm/versions/node/v10.20.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v10.20.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29
      Build Tools: 27.0.3, 28.0.3, 29.0.2
      System Images: android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
[...]

PR #29754, opened 2020-08-25
Jest setup: Stop polyfilling Promise.

chrisbobbe:
Hi! 👋 I hope anyone reading this is doing OK these days. Thanks, in any case, for continuing to put your time into the awesome React Native ecosystem.

## Summary

It looks like this line was introduced in 3ff3987, in 2015, and it
has remained in a similar form since then. I haven't found any
explanation for it.

At facebook/jest#10221 [1], a core Jest maintainer says,

"""
As an aside, one should never replace `global.Promise` [...]. E.g.
when using `async-await` you will always get the native `Promise`
regardless of the value of `global.Promise`.
"""

facebook/jest#10221 is one issue this line has raised, for anyone
using the latest features of Jest to test async code in their React
Native projects.

[1] https://github.com/facebook/jest/issues/10221#issuecomment-654687396

Fixes: #29303

## Changelog

[General] [Fixed] - Fix issue with testing async code with Jest's fake timers.

## Test Plan

It looks like there's still an effort to keep up-to-date with new versions of the `promise` NPM package; see https://github.com/facebook/react-native/commit/b23efc526446cad53adc3de7465fd8be7309d84e. But that commit still leaves it a mystery why we're polyfilling `Promise` here.

I also see [Libraries/Core/polyfillPromise.js](https://github.com/facebook/react-native/blob/master/Libraries/Core/polyfillPromise.js), but I haven't run into any problems that I can trace to that file.

It's hard to say what side effects might arise from removing the line without knowing what it's for. (https://github.com/facebook/react-native/issues/29303 exists mostly to ask that question.)

### DIFF
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -21,7 +21,6 @@ global.performance = {
   now: jest.fn(Date.now),
 };
 
-global.Promise = jest.requireActual('promise');
 global.regeneratorRuntime = jest.requireActual('regenerator-runtime/runtime');
 
 global.requestAnimationFrame = function(callback) {


andreialecu:
I really hoped this would fix https://github.com/callstack/react-native-testing-library/issues/379#issuecomment-714341282 but after making this change locally everything got broken:

```
TypeError: TaskQueue: Error with task : Cannot read property 'colors' of undefined
```
via this hook (from `react-native-paper`):
```ts
  const {colors} = useTheme();
```

The error does not appear without this change, so it seems there's a legitimate reason for `Promise` to be polyfilled.

Similar breakage has also been reported here: https://github.com/callstack/react-native-testing-library/issues/379#issuecomment-716727350

chrisbobbe:
> Similar breakage has also been reported here:

Hmm, I don't see anything there related to `react-native-paper`. It does look like there's some concerning breakage, though.

> so it seems there's a legitimate reason for `Promise` to be polyfilled.

I'm not sure how legitimate the reason is; possibly the choice wasn't well-considered, and development continued in a way that wasn't compatible with the native `Promise`, by accident? And nobody realized.

andreialecu:
There are a bunch of failing tests which seem to hint at some functionality related to pausing promises. This probably can't be done with native promises, hence why a polyfill was used.

https://app.circleci.com/pipelines/github/facebook/react-native/6090/workflows/e4779cac-5fbe-41d7-9bb2-cc1afc89ebab/jobs/164645

chrisbobbe:
> This probably can't be done with native promises

Yeah, that would be a good question to answer: can the code be changed so that it works with native `Promise`, and why or why not?

The choice affects React Native's public API; app developers are [encouraged](https://github.com/facebook/react-native/blob/v0.64.0/template/package.json#L27) to use React Native's Jest preset, which activates the polyfill when the developer tests their own app (as we see with facebook/jest#10221). Restricting app developers to use a particular, non-standard `Promise` implementation in their tests (and by extension, in their app code, if they want to avoid surprises where their tests mishandle it) is something that should be documented, at least.

Also, one reason not to replace the global `Promise` seems pretty fundamental. As I quoted in my opening comment, from a maintainer of another Facebook project (from https://github.com/facebook/jest/issues/10221#issuecomment-654687396),

> """
> As an aside, one should never replace `global.Promise` [...]. E.g.
> when using `async-await` you will always get the native `Promise`
> regardless of the value of `global.Promise`.
> """

If the polyfill is sometimes used and sometimes not, depending on what syntactic sugar happens to be used, that seems like a likely cause of some pretty hard-to-track bugs—and a deterrent to writing stable async code in a sane way. If React Native has a workaround for *that*, I'd appreciate documentation for it too. 🙂

j3tan:
Just wanted to add some information that I encountered while trying to override the `global.Promise` polyfill to use native promises. There is existing code in `react-native` that depends on the non-standard promise implementation of [done()](https://github.com/then/promise#promisedoneonfulfilled-onrejected) 

Example:
https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Interaction/InteractionManager.js#L115-L124

which causes issues if you are trying to use `InteractionManager.runAfterInteractions(...PromiseTask...)` since native promises will throw the following error from `TaskQueue` if you return a native `Promise` from the `gen` function
```
TypeError: TaskQueue: Error with task task_name: task.gen(...).then(...).catch(...).done is not a function
```
Due to a chained `done()` call here:
https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Interaction/TaskQueue.js#L162-L178

andreialecu:
@robhogan I don't want to cause panic here, but I vaguely remember trying to patch RN in the past to use the native promise, and it broke various things that I don't remember right now. :)

It could've affected other libraries, like react-native-testing-library, but again, I forgot the specifics.

There are some assumptions in certain places that the polyfilled promise is being used. See the comment above: https://github.com/facebook/react-native/pull/29754#issuecomment-948959046.

robhogan:
Hi @andreialecu - thanks for the heads up. I patched up our internal tests, but you're right that there *could* be a knock-on in `react-native-testing-library`.

This is part of an effort to update Jest, where modern timers in recent versions have better Promise handling, so I hope if there are issues created by this, modern Jest might be a way out.

@thymikee, @mdjastrzebski - just FYI, let me know if you foresee an issue here.

thymikee:
@robhogan we already use node's Promise in our preset which we recommend to our users: https://github.com/callstack/react-native-testing-library/blob/main/jest-preset/index.js

PR #29770, opened 2020-08-26
Change default metro.config.js to utilize inline requires

GantMan:
## Summary
This feature has been in [experimental mode since 0.59.0](https://reactnative.dev/blog/2019/03/12/releasing-react-native-059#-faster-app-launches-with-inline-requires).  The feature seems quite beneficial.

Most people won't know to look for the inline requires flag, and it's been existing in React Native as an experiment for months.  This defaults the flag to true as the next step towards utilizing this optimization for React Native devs initiating brand new projects.

## Changelog

[General] [Breaking] - Enable `inlineRequires` by default. Gives a performance benefit but slightly different JS execution order


TheSavior:
This landed but internal tooling was broken which lost attribution. :( 

https://github.com/facebook/react-native/commit/fd691d8f5148b83c2bdf99a54bdc7d5bbeb5d593

Thanks for contributing!

mrousavy:
Is there a reason to leave `experimentalImportSupport` disabled? I thought those two features were related?

kuasha420:
> Is there a reason to leave `experimentalImportSupport` disabled? I thought those two features were related?

The whole think is quite under documented

PR #29778, opened 2020-08-26
Re-sync with internal

bigfootjon:
fbshipit-source-id: 65e2766a6a097eca0e0d0fda8dadf6871e9276c2

### DIFF
--- a/ReactCommon/jsi/jsi/test/testlib.cpp
+++ b/ReactCommon/jsi/jsi/test/testlib.cpp
@@ -394,6 +394,23 @@ TEST_P(JSITest, HostObjectTest) {
                    .getBool());
 }
 
+TEST_P(JSITest, HostObjectProtoTest) {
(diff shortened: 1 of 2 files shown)


ReactCommon/jsi/jsi/test/testlib.cpp
```diff
@@ -394,6 +394,23 @@ TEST_P(JSITest, HostObjectTest) {
...
+      return String::createFromAscii(rt, "phoprop");
+    }
+  };
+
+  rt.global().setProperty(
+      rt,
+      "pho",
+      Object::createFromHostObject(rt, std::make_shared<ProtoHostObject>()));
+
+  EXPECT_EQ(
```
shergin, line 409: It looks like this code tests that we can use HostObjects as object prototypes.
Seems legit but I don't understand what's the point to use `Symbol.toPrimitive` here, we can use any symbol here.

PR #29792, opened 2020-08-27
Remove .gitattributes and normalize line endings

friederbluemle:
## Summary

The nested `.gitattributes` file in `packages/react-native-codegen/android/` caused some confusion on Linux and macOS, causing Git to show `packages/react-native-codegen/android/gradlew.bat` as modified (CRLF removed, LF added).

Instead of relying on repo-local `.gitattributes` files to convert endings in the working directory, the files should be committed to source control with the correct line endings in the first place. There is no reason to convert LF endings in .sh and many other file to CRLF on Windows (maybe this was an issue a long time ago, but unless Notepad is used this won't be a problem for practically all modern editors).

Also fixed the line endings of `scripts/launchPackager.bat` which was incorrectly committed as LF.

## Changelog

[Internal] [Fixed] - Line endings and .gitattributes

## Test Plan

Clone repo on Linux, macOS, and Windows, and make sure no modified files show up.


mdvacca (commented):
thanks for the PR @friederbluemle, importing to review internally

pvinis:
I totally missed this PR. I think it should be reverted and/or fixed. Here's the explanation:

- We need the `*.pbproj -text` for the template, so git can handle the xcode project files as text, and show the diffs correctly.
- The files that are used on windows should stay with CRLF because that's how they come. `gradlew.bat` in a new android project comes with CRLF and we don't want new projects to show this file as "changed".
- I agree that the root `.gitattributes` doesn't need to be there and could be moved elsewhere to be more local.

I will make a PR now and see what's up.

friederbluemle:
Thanks for the comments and concerns! Like you, I am interested in getting this fixed the "right way" :) I left a comment on the PR you opened, please check.

PR #29799, opened 2020-08-28
Removed usage of legacy context API in AppContainer

gittysachin:
## Summary
1/4 of #28103 and closes it.
Remove usage of the legacy context API in AppContainer to get rid of the old React context API.

## Changelog

Run the RNTester app and test the AppContainer example. Also, add a `console.warn` to make sure we get the correct context value.


safaiyeh:
cc @TheSavior

javache:
Closing some old PR's. Seems like this was addressed in https://github.com/facebook/react-native/commit/f9ecce7440e712782f4fd760a46e54b3102914d8

Issue #25220, opened 2019-06-11
Accessibility label is not reaching the function that is handling it on iOS


LorandP:
## Description
  When adding an accessibilityLabel to a view, on iOS that accessibility label is not reaching the function that is handling the label.

We have discovered this issue while importing a .ipa into Appium Inspector which is an automation testing tool. There the accessibility labels get concatenated, you can see that in this screenshot:
https://ibb.co/Nxq4Cdh

On further inspection, I have found that on the native side, the functions that are handling the accessibility lables don't receive the accessibility labels set.

Other issues opened on this are:
https://github.com/appium/appium/issues/10654
https://github.com/facebook/react-native/issues/21830

React Native version: 0.59.5

## Steps To Reproduce

1. Add some an accessibleLabel to a view
2. Go to the following function located in:
React -> Views -> RCTView -> (NSString *)accessibilityLabel
3. When the flow of the application gets there, you will see that the function tries to return super.accessibilityLabel but every time that is nil
4. The flow then goes into the RCTRecursiveAccessibilityLabel function which receives the self parameter from which then he tries to retrieve the subviews that contain the accessibilityLabels but there is no subview when the function gets to that point.

Describe what you expected to happen:
We expect that  the `(NSString *)accessibilityLabel` function successfully retrieves the accessibilityLabel and returns it or else the `RCTRecursiveAccessibilityLabel` function receives the subviews for the current screen

Snack, code example, or link to a repository:
```
- (NSString *)accessibilityLabel
{
  NSString *label = super.accessibilityLabel;
  if (label) {
    return label;
  }
  return RCTRecursiveAccessibilityLabel(self);
}

static NSString *RCTRecursiveAccessibilityLabel(UIView *view)
{
  NSMutableString *str = [NSMutableString stringWithString:@""];
  for (UIView *subview in view.subviews) {
    NSString *label = subview.accessibilityLabel;
    if (!label) {
      label = RCTRecursiveAccessibilityLabel(subview);
    }
    if (label && label.length > 0) {
      if (str.length > 0) {
        [str appendString:@" "];
      }
      [str appendString:label];
    }
  }
  return str;
}

a7urag:
any update on this?

a7urag:
Found this, https://github.com/tipsi/react-native-testid/blob/master/index.js

May be helpful

PR #29801, opened 2020-08-28
Fix duplicate accessibilityLabels & app crash on iOS

### LINKED ISSUE #21830: Wrong translation from react-native's accessibilityLabel to iOS name/label

nfriedly:
## Summary

I came across this because one of my customer's apps was freezing and crashing when my code tries to read the `accessibilityLabel` of a view containing a chat component. Because the labels are recursively generated from the text content, any sufficiently long chat ends up with all of the parent views having a ridiculously long `accessibilityLabel`. 

After 10 seconds of non-responsiveness from trying to generate all of this on the UI thread (when backgrounded), iOS decides the app has crashed and force-closes it.

Before this change, every single view in a RN app returned true for the conditional. 

I believe this change will also improve both the accessibility and the testability of RN apps, in addition to fixing that crash.

This change also allows for correct behavior with this code:

```objective-c
if ([view respondsToSelector:@selector(accessibilityLabel)])
    return view.accessibilityLabel;
```

Before this change, any RCTView could potentially scan its entire tree and return `nil`. 
After, only RCTViews that actually have an `accessibilityLabel` set will return one, and ones that don't will be skipped by the above code.

This PR supersedes or fixes #21830, #25963, #24113, #24118, https://github.com/appium/appium/issues/10654, possibly #25220, and probably a few other tickets I haven't identified.

The changes in this PR are for `RCTView`. It's worth mentioning that there is very similar code in [`RCTViewComponentView`](https://github.com/facebook/react-native/blob/ebf2fb7868d92661772ec59d6592f14786e9673b/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L526-L552) - I haven't dug into that as much, but I suspect it should probably get the same treatment. If desired, I can include it in this PR.

I also haven't yet reviewed the Android side of things to see what happens there, but whatever the case, the same component structure doesn't seem to be leading to crashes in Android.

I might need to change the documentation too. 

## Changelog

[iOS] [Removed] - Deleted RCTView's accessibilityLabel wrapper and RCTRecursiveAccessibilityLabel, allowing for default native behavior

## Test Plan

This is where I need a little help. What's a good way to test this? 

I spent a while digging through the RN iOS tests, I can run the tests, and even step through the harness, but I still don't feel like I fully understand what's happening. I don't have a ton of iOS or RN experience, so it may be something obvious, but I'm just not sure what I'm missing.

### DIFF
--- a/React/Views/RCTView.m
+++ b/React/Views/RCTView.m
@@ -199,15 +199,6 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
 
 #pragma mark - Accessibility
 
-- (NSString *)accessibilityLabel
-{
-  NSString *label = super.accessibilityLabel;
-  if (label) {
-    return label;
-  }
-  return nil;
-}
-
 - (NSArray<UIAccessibilityCustomAction *> *)accessibilityCustomActions
 {
   if (!self.accessibilityActions.count) {
--- a/React/Views/RCTView.m
+++ b/React/Views/RCTView.m
@@ -80,24 +80,6 @@ UIAccessibilityTraits const SwitchAccessibilityTrait = 0x20000000000001;
 
 @end
 
-static NSString *RCTRecursiveAccessibilityLabel(UIView *view)
-{
-  NSMutableString *str = [NSMutableString stringWithString:@""];
-  for (UIView *subview in view.subviews) {
-    NSString *label = subview.accessibilityLabel;
-    if (!label) {
-      label = RCTRecursiveAccessibilityLabel(subview);
-    }
-    if (label && label.length > 0) {
-      if (str.length > 0) {
-        [str appendString:@" "];
-      }
-      [str appendString:label];
-    }
-  }
-  return str.length == 0 ? nil : str;
-}
-
 @implementation RCTView {
   UIColor *_backgroundColor;
   NSMutableDictionary<NSString *, NSDictionary *> *accessibilityActionsNameMap;
@@ -223,7 +205,7 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : unused)
   if (label) {
     return label;
   }
-  return RCTRecursiveAccessibilityLabel(self);
+  return nil;
 }
 
 - (NSArray<UIAccessibilityCustomAction *> *)accessibilityCustomActions


React/Views/RCTView.m
```diff
@@ -223,7 +205,7 @@ - (NSString *)accessibilityLabel
   if (label) {
     return label;
   }
-  return RCTRecursiveAccessibilityLabel(self);
+  return nil;
 }
```
nfriedly, line 209: Come to think of it, I might be able to just remove this method entirely since all it's doing is calling the super now.

nfriedly: Update: removed. This should actually make it behave better with code like this (which is what my code actually does):

```objective-c
if ([view respondsToSelector:@selector(accessibilityLabel)])
    return view.accessibilityLabel;
```

shergin:
I think @rigdern has a context why those functions are needed.

rigdern:
@shergin Unfortunately, I don't have context on those functions. They've been in `RCTView` since the first open source commit of React Native. If we can find somebody that is intimately familiar with React Native's accessibility implementation to help us, that would be great. If not, since these functions have been around for so long, here are some things I would check before making the change to ensure it doesn't cause a regression:
  - Does `RCTView` exhibit the same issues? Or did `RCTViewComponentView` introduce them or exacerbate them? For example, perhaps in `RCTViewComponentView` hierarchies some views are marked as accessible that aren't in `RCTView` hierarchies that cause heavier use of the `accessibiiltyLabel` override.
  - Check with some app developers that have been using React Native's accessibility framework for a long time. Have they encountered any problems with React Native's `accessibilityLabel` implementation? Is it doing what they want? Would they prefer the new implementation?
  - Does Android exhibit a similar behavior and do we want to change it? When I was actively contributing to React Native, I was advocating for consistency between iOS and Android for the accessibility APIs. That way, a developer can write some accessibility code and expect it'll do the right thing on iOS and Android. However, that was a couple of years ago so I'm not familiar with how the accessibility APIs have evolved or the current philosophy behind them.

nfriedly:
I spent some time testing this in the rn-tester app on a real iPhone (it took me a while to realize that VoiceOver just doesn't work in the simulator), and I have to admit that it seems to make a somewhat worse experience there. A lot of "Possible text" statements prefixing things where it would previously just read the text. And occasionally some text is omitted entirely. So this isn't the accessibility improvement I had hoped it would be.

My main goal here is to prevent the crash, which obviously impairs accessibility. I think that can be done without a regression in functionality. Here's a couple of ideas I thought of:

* Only generate `accessibilityLabels` on selectable elements, otherwise return `nil`. 
* Put a limit on the string-length, or depth, or absolute number of child nodes that `RCTRecursiveAccessibilityLabel` looks at
* Cache the results of `RCTRecursiveAccessibilityLabel` - this would prevent the exponential as `RCTRecursiveAccessibilityLabel` when my code scans every view and `RCTRecursiveAccessibilityLabel` in turn scans every sub-view. As usual, cache invalidation is the hard part there.

What do you guys think?

nfriedly:
A couple of quick updates: 

1) We ended up working around this issue by special-casing React Native views in our code - for `RCTView`s and `RCTViewComponentView`s and their subclasses, we now use the native `accessibilityLabel` implementation from the UIView class. 

2) I have an idea for how to make this better for everyone else without negatively impacting actual accessibility: I believe that the `accessibilityLabel`s are only used by VoiceOver when `accessibilityElement` is `true` - so when `RCTView.accessibiltyLabel()` is called, I'd like to make it check that and only do the recursive string building in that case. 

I feel like there might be a better way to build the string also - maybe pass in the NSMutableString instead of creating a new one for each view. 

I'll experiment with the above when I have time, but I'm not sure when that will be now that the immediate pressure is off.

nfriedly:
I'm going to close this PR, as there are other PRs up now that address the problem better than this one (e.g. https://github.com/facebook/react-native/pull/31924 improves performance _without_ changing behavior, https://github.com/facebook/react-native/pull/31924 improves the performance _and_ the behavior.)

Issue #29790, opened 2020-08-27
React Native App Bundle Version Code (Weird Number) - 


dpedneka:
React Native App Version changes when Uploaded on Play Store  
Version - 3145730 (1.0)
Version - 6291460 (2.0)

Is this because of APP Bundle?

To Release app we followed the following process.
https://reactnative.dev/docs/signed-apk-android

$ ./gradlew bundleRelease

Why this behaviour? Is it normal or is there any changes we need to do resolve this error.

gedeagas:
Hi @dpedneka this happens because https://github.com/facebook/react-native/blob/7694b32a88078278457dd8721eb61da9c4ac0f5a/template/android/app/build.gradle#L176 your versionCode is lower than the previous apk. The current calculation on versionCodeOverride is unpredictable.

V1.0 200
```versionCode 200
versionName "1.0"```

V2.0 1
```versionCode 1
versionName "2.0"```

even tho the version name changes to v2 the version code is lower than the previous build. because of this, the build will be rejected. The current implementation makes the calculated build number unpredictable and seems arbitrary. I am in the process of making a pr to change the default build calculation closer to google implementation of build number in a flutter.  `abiVersionCode * 1000 + variant.versionCode`. 

With this change,

```versionCode 4
versionName "4.0"```

With this change the result of `versionCode` 4 for `armeabi-v7a` will be 1004 (1*1000+4) instead of the seemingly arbitrary 1048580 ( 1*1048576+4).

devlopersabbir:
## there are any solutions to solve this problem??

I have tried to rebuild it as aab 3/4 times with the changed version from the `package.json` version also `app.json` versions.

**Still I got the same problem**
![image](https://github.com/facebook/react-native/assets/82939905/8345c1bb-5972-4f73-af18-111e5fabde3e)

farhan363:
try adding versionCode line in package.json
 "name": "shashlick-driver",
  "version": "2.0.0",
  "versionCode": 2,
  "main": "node_modules/expo/AppEntry.js",

PR #29808, opened 2020-08-29
Making Android versionCodeOverride for new apps using the template human-readable

gedeagas:
## Summary

The current calculation on versionCodeOverride is not human-readable. 

Imagine if we have an android app with **versionName** `4.0` and **version code** `4`.

In the current implementation, the result of **versionCode** `4` for `armeabi-v7a` will be the seemingly arbitrary **1048580**. This makes the version code to be seemingly arbitrary and hard to read for humans. This PR proposes to change this calculation closer to google implementation of build number in Flutter (`abiVersionCode * 1000 + variant.versionCode`). 
https://github.com/flutter/flutter/blob/39d7a019c150ca421b980426e85b254a0ec63ebd/packages/flutter_tools/gradle/flutter.gradle#L643-L647

With this change, our app with `versionCode 4 versionName "4.0"` for  `armeabi-v7a`  will have **1004**  as the version code instead of the seemingly arbitrary **1048580**. As you can see adopting the flutter style implementation make the version code easier to read and debug. 

**1004** 
**1** - The ABI Type `["armeabi-v7a": 1, "x86": 2, "arm64-v8a": 3, "x86_64": 4]`
**004** - Our versionCode. 

Hopefully, this can prevent future issues like this https://github.com/facebook/react-native/issues/29790.

## Changelog

[Android] [Changed] - Making Android versionCodeOverride for new apps using the template human-readable

### DIFF
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -173,7 +173,7 @@ android {
             def abi = output.getFilter(OutputFile.ABI)
             if (abi != null) {  // null for the universal-debug, universal-release variants
                 output.versionCodeOverride =
-                        versionCodes.get(abi) * 1048576 + defaultConfig.versionCode
+                        versionCodes.get(abi) * 1000 + defaultConfig.versionCode
             }
 
         }
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -173,7 +173,7 @@ android {
             def abi = output.getFilter(OutputFile.ABI)
             if (abi != null) {  // null for the universal-debug, universal-release variants
                 output.versionCodeOverride =
-                        versionCodes.get(abi) * 1000 + defaultConfig.versionCode
+                        defaultConfig.versionCode * 1000 + versionCodes.get(abi)
             }
 
         }
--- a/template/android/app/build.gradle
+++ b/template/android/app/build.gradle
@@ -169,6 +169,7 @@ android {
         variant.outputs.each { output ->
             // For each separate APK per architecture, set a unique version code as described here:
             // https://developer.android.com/studio/build/configure-apk-splits.html
+            // Example: versionCode 1 will generate 1001 for armeabi-v7a, 1002 for x86, etc.
             def versionCodes = ["armeabi-v7a": 1, "x86": 2, "arm64-v8a": 3, "x86_64": 4]
             def abi = output.getFilter(OutputFile.ABI)
             if (abi != null) {  // null for the universal-debug, universal-release variants


fkgozali:
cc @dulmandakh - thoughts? If people are already using the arbitrary version like 1048580 before, will this cause any problem?

gedeagas:
@fkgozali @dulmandakh I also have this concern but I am not familiar with the react-native CLI `upgrade` command on whether this kind of change on the Gradle template file would be applied automatically or not. Loved to hear your thoughts on this.

dulmandakh:
IIRC, Play Store requires a new release to increment versionCode. So I don't think this PR will fix the issue, but cause more.

gedeagas:
what do you mean about ` cause more` issues? can you specific. anway i used this exact configutayion on my production apps on google play store without a problem.

fkgozali:
My thoughts:
* It's good for the starter template to provide a boilerplate logic to manage version number, but it needs to produce numbers that make sense, not magical.
* Existing app who already used the old template will be stuck with the magical version numbers
* New app using the template with this change will have more reasonable version numbers moving forward.

This seems OK to me, thoughts @dulmandakh?

On the implementation note, I usually prefer the "details" to have lower digit in the version number. Your proposal is:

```
abiVersionCode * 1000 + variant.versionCode
```

so versionCode 4 will produce:
* 1004 (armeabi-v7a)
* 2004 (x86)
* ...

What if versionCode reaches 1000? That would cause problem because it doesn't fit the formula. What about reversing this?

```
variant.versionCode * 1000 + abiVersionCode
```
This way, each "release" for the versionCode always give you strictly increasing version numbers: 4001, 4002, ...

In practice, the build number should always monotonically increase, regardless of flavor (e.g. in CI environment). So I think reversing the logic as suggested would satisfy this?

gengjiawen:
tbh I want to complete remove this part, and lock versionCode to current highest value. And instruct user to modify this when they make a new release.

gedeagas:
@fkgozali That makes sense, just updated the pr with those changes. 

@gengjiawen What is your strategy to instruct the user to change the versionCode? via the CLI?.

From the DX standpoint when you `lock versionCode to current highest value` without the current implementation of `versionCodeOverride` and just relying on versionCode user will get an error message when they uploaded the split ABI to Google Play Store (the split APK per ABI will have the same version number across all of them). Because of this for me personally react-native having a boilerplate logic to manage version number is quite helpful to new users that don't necessarily understand the native platforms.

gengjiawen:
> What is your strategy to instruct the user to change the versionCode? via the CLI?.

Like any other normal android app, change it in build.gradle. If they like, they can change it to link to commit count or any other dynamic approach.

> the split APK per ABI will have the same version number across all of them

I don't think this will be a problem.

tbh current plan make it smaller only make this worse.

gedeagas:
> Like any other normal android app, change it in build.gradle. If they like, they can change it to link to commit count or any other dynamic approach.

I was talking about `the instruction` like where we should put the warning ( via code comment, etc ) not making the cli bump the version code. 

> I don't think this will be a problem.

If we remove the `versionCodeOverride` yes the app-specific ABI will have the same version number.  By default, when Gradle generates multiple APKs, each APK will have the same version information, as specified in the module-level build.gradle file. Because the Google Play Store does not allow multiple APKs for the same app that all have the same version information, you need to ensure each APK has its own unique versionCode before you upload to the Play Store. https://developer.android.com/studio/build/configure-apk-splits

gengjiawen:
> > Like any other normal android app, change it in build.gradle. If they like, they can change it to link to commit count or any other dynamic approach.
> 
> I was talking about `the instruction` like where we should put the warning ( via code comment, cli, etc ) not making the CLI bump the version code.
> 
> > I don't think this will be a problem.
> 
> If we remove the `versionCodeOverride` yes the app-specific ABI will have the same version number and you can't upload split abi that way to google play.
> 
> By default, when Gradle generates multiple APKs, each APK will have the same version information, as specified in the module-level build.gradle file. Because the Google Play Store does not allow multiple APKs for the same app that all have the same version information, you need to ensure each APK has its own unique versionCode before you upload to the Play Store. [developer.android.com/studio/build/configure-apk-splits](https://developer.android.com/studio/build/configure-apk-splits)

Thanks for the info. The number 1048580 is take from google android docs (include the whole applicationVariants part). Change it to smaller 1000 will cause more trouble.

gedeagas:
> > > Like any other normal android app, change it in build.gradle. If they like, they can change it to link to commit count or any other dynamic approach.
> > 
> > 
> > I was talking about `the instruction` like where we should put the warning ( via code comment, cli, etc ) not making the CLI bump the version code.
> > > I don't think this will be a problem.
> > 
> > 
> > If we remove the `versionCodeOverride` yes the app-specific ABI will have the same version number and you can't upload split abi that way to google play.
> > By default, when Gradle generates multiple APKs, each APK will have the same version information, as specified in the module-level build.gradle file. Because the Google Play Store does not allow multiple APKs for the same app that all have the same version information, you need to ensure each APK has its own unique versionCode before you upload to the Play Store. [developer.android.com/studio/build/configure-apk-splits](https://developer.android.com/studio/build/configure-apk-splits)
> 
> Thanks for the info. The number 1048580 is take from google android docs (include the whole applicationVariants part). Change it to smaller 1000 will cause more trouble.

What google docs are you reffering to? https://developer.android.com/studio/build/configure-apk-splits it seems to me that the docs is using 1000 not 1048580?

gengjiawen:
> What google docs are you reffering to? [developer.android.com/studio/build/configure-apk-splits](https://developer.android.com/studio/build/configure-apk-splits) it seems to me that the docs is using 1000 not 1048580?

Google probably delete the original page recently, it's a long page introduce android gradle build system exists for many years.

gedeagas:
> Thanks for the info. The number 1048580 is take from google android docs (include the whole applicationVariants part). Change it to smaller 1000 will cause more trouble.

Is the Gradle template file patched automatically to the user Gradle file when they update react-native via the CLI? my main concern is to fix this for users that generate new app using the template not patching the older implementation. and also probably put some code comments disclaimer for those that update react-native via the rndiff website.

gengjiawen:
> > Thanks for the info. The number 1048580 is take from google android docs (include the whole applicationVariants part). Change it to smaller 1000 will cause more trouble.
> 
> Is the Gradle template file patched automatically to the user Gradle file when they update react-native via the CLI? my main concern is to fix this problem **for users that generate new app using the template** not **patching the older implementation**. and also probably put some code comments disclaimer for those that update react-native via the rndiff website.

Will make defaultConfig default version code to 1048580 and make `* 1` in applicationVariants helps ?
```
output.versionCodeOverride =
                      defaultConfig.versionCode * 1 + versionCodes.get(abi)
```

gedeagas:
@gengjiawen what are you trying to achieve with that configuration? 
``` defaultConfig {
        applicationId "com.helloworld"
        minSdkVersion rootProject.ext.minSdkVersion
        targetSdkVersion rootProject.ext.targetSdkVersion
        versionCode 1048580
        versionName "1.0"
    }
```
```
output.versionCodeOverride =
                      defaultConfig.versionCode * 1 + versionCodes.get(abi)```

is not going to solve anything, it will break the current config and still produce an arbitrary number. And the versionCode should never be static, the user should maintain this.

fkgozali:
Thinking about this, I think this change is still reasonable for **new apps created via react-native init**. Existing apps won't be affected.

@gedeagas, can you update a few things?
* In your commit summary's change log, explicitly mention "for new apps using the template"
* In your code, add a short comment with example version numbers based on the new logic

gedeagas:
@fkgozali updated.

Issue #28140, opened 2020-02-20
ScrollView.scrollIndicatorInsets behavior changed in iOS 13


justinwh:
When running on iOS 13, `ScrollView`s set their `scrollIndicatorInsets` automatically to be inside the device's "safe area" (i.e. inside the "notch").  On older iOS versions this did not happen—users needed to take steps to accommodate the notch (either setting the `scrollIndicatorInsets` directly or nesting `ScrollView` inside `SafeAreaView`).

The difference is clear using a basic test app with a fullscreen `ScrollView`.  On an iOS 13 device (this is easiest to see on a phone with a notch, e.g.  an iPhone X), when you scroll the view and reveal the indicators, they start/stop inside the notch.

![ios13](https://user-images.githubusercontent.com/428831/74971779-a263e900-53ee-11ea-8807-8c614b079e96.png)

On an iOS 12 device, the indicators start/stop at the edge of the view.

![ios12](https://user-images.githubusercontent.com/428831/74971710-86f8de00-53ee-11ea-9707-d0343093e0c6.png)

(Both of these screenshots are taken just as the scrolling reaches the top of the view.)

Here's the `App.js` I used in a brand new, latest version (0.61.5) RN project to generate the above screenshots:

```
import React from 'react';
import * as RN from 'react-native';

const a20 = new Array(20).fill({}, 0, 20);

export default function App (props) {
  return (
    <RN.ScrollView automaticallyAdjustContentInsets={false}
      contentInsetAdjustmentBehavior='never'
      style={{flex: 1}}>
      {a20.map((o, i) => (
        <RN.View key={i}
          style={{backgroundColor: ((i % 2 === 0) ? 'red' : 'blue'), height: 75}}/>
      ))}
    </RN.ScrollView>
  );
}

```

You cannot set negative `scrollIndicatorInsets` to replicate the iOS 12 behavior on iOS 13.  And any `scrollIndicatorInsets` you **do** assign are added to the ones automatically chosen by the OS, so if you set them directly to a value that worked correctly on iOS 12, you'll get broken-looking results on iOS 13 (extra large insets).

This behavior is not affected by the `ScrollView.automaticallyAdjustContentInsets` or `ScrollView.contentInsetAdjustmentBehavior` properties (as far as I can tell).

React Native version:
```
System:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
[...]

Issue #28085, opened 2020-02-15
Vertical scrollbar of FlatList not correctly vertically aligned


jasperkuperus:
As reported in https://github.com/facebook/react-native/issues/26610 but now following the issue template as discussed with @safaiyeh and @hramos 

### Description:

The vertical scrollbar of a `FlatList` does not always render correctly to the right side of the screen. See all the screenshots in this issue.

For me. This happened even when I slimmed down my code into a `<FlatList />` with only `<Text />`. It only happened though when my list was longer than the height of the screen. The scrollbar then suddenly moves to the left. The longer the list, the further the scrollbar moves to the left.

Also another condition for me. Upon a fresh first render of my `<FlatList />`, this behaviour appeared. When navigating away, effectively unmounting this component, and back, the scrollbar is suddenly OK. Close and open the app: The scrollbar is at the wrong position again.

### React Native version:

```
info Fetching system and libraries information...
System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 2.26 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.12.0 - /var/folders/_9/nh0gc1bs08qd9p3kqshr3f0h0000gn/T/yarn--1581784553520-0.7678366513125479/node
    Yarn: 1.19.1 - /var/folders/_9/nh0gc1bs08qd9p3kqshr3f0h0000gn/T/yarn--1581784553520-0.7678366513125479/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v12.12.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.2.1/11B500 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5
```

###  Steps To Reproduce

1. Create a `<FlatList />` with `<Text />` items in it
2. Make sure the list is longer than the screen
3. Check the first render of this list, scroll

###  Expected Results

Scrollbar should stick to the right of the screen.

###  Snack, code example, screenshot, or link to a repository:

[...]

Issue #26610, opened 2019-09-27
about ScrollView on ios 13 . scroll bar will be center sometimes for ios 13 and can move to phone left


KarenQiao:
"react": "16.8.3",
"react-native": "0.59.9",
  Platforms: iOS 13.0, 

![image](https://user-images.githubusercontent.com/27947842/65738561-e0493380-e114-11e9-81dc-2f385362d855.png)

KarenQiao:
System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) m3-7Y32 CPU @ 1.10GHz
      Memory: 43.45 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.13.0 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
      Android SDK:
        API Levels: 23, 26, 27, 28
        Build Tools: 27.0.3, 28.0.3
        System Images: android-26 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 11.0/11A420a - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.9 => 0.59.9 
    npmGlobalPackages:
      react-native-cli: 2.0.1

fahmidme:
Same issue here

amsdamsgram:
Hey, I have the same issue with `FlatList` component.

cernadasjuan:
Same issue here

losh11:
I'm actually experiencing this with `react-native-webview` from `react-native-community`... didn't know it affected more than just the webview.

edo1493:
I can see the same on react-native 0.60.5 with FlatList.

fahmidme:
I don't know if this happened with anyone else or is that related but incase anyone experienced it: I saw my custom contact list for a second and then it changed to something like this with the scroll bar in the middle and then the thing disappeared.

mobitar:
[...]

PR #29809, opened 2020-08-29
Add ScrollView.automaticallyAdjustsScrollIndicatorInsets prop (on iOS)

justinwh:
## Summary

iOS 13 added a new property to `UIScrollView`: `automaticallyAdjustsScrollIndicatorInsets`, which is `YES` by default.  The property changes the meaning of the `scrollIndicatorInsets` property.  When `YES`, any such insets are **in addition to** whatever insets would be applied by the device's safe area.  When `NO`, the iOS <13 behavior is restored, which is for such insets to not account for safe area.

In other words, this effects ScrollViews that underlay the device's safe area (i.e. under the notch).  When `YES`, the OS "automatically" insets the scroll indicators, when `NO` it does not.

There are two problems with the default `YES` setting:

1. It means applying `scrollIndicatorInsets` to a `ScrollView` has a different effect on iOS 13 versus iOS 12.
2. It limits developers' control over `scrollIndicatorInsets`.  Since negative insets are not supported, if the insets the OS chooses are too large for your app, you cannot fix it.

Further explanation & sample code is available in issue #28140 .

This change sets the default for this property to `NO`, making the behavior consistent across iOS versions, and allowing developers full control.

## Changelog

[iOS] [Changed] - ScrollView scrollIndicatorInsets to not automatically add safe area on iOS13+

## Test Plan

Here are screenshots of the demo app (from the original bug) before (with safe area applied to insets) & after (without safe area applied to insets):

before

after

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -172,6 +172,12 @@ type IOSProps = $ReadOnly<{|
    */
   automaticallyAdjustContentInsets?: ?boolean,
   /**
+   * Controls whether iOS should automatically adjust the scroll indicator
+   * insets. The default value is true. Available on iOS 13 and later.
+   * @platform ios
+   */
+  automaticallyAdjustsScrollIndicatorInsets?: ?boolean,
+  /**
    * The amount by which the scroll view content is inset from the edges
    * of the scroll view. Defaults to `{top: 0, left: 0, bottom: 0, right: 0}`.
    * @platform ios
--- a/Libraries/Components/ScrollView/ScrollViewNativeComponent.js
+++ b/Libraries/Components/ScrollView/ScrollViewNativeComponent.js
@@ -26,6 +26,7 @@ const ScrollViewNativeComponent: HostComponent<Props> = NativeComponentRegistry.
       alwaysBounceHorizontal: true,
       alwaysBounceVertical: true,
       automaticallyAdjustContentInsets: true,
+      automaticallyAdjustsScrollIndicatorInsets: true,
       bounces: true,
       bouncesZoom: true,
       canCancelContentTouches: true,
--- a/Libraries/Components/ScrollView/ScrollViewNativeComponentType.js
+++ b/Libraries/Components/ScrollView/ScrollViewNativeComponentType.js
@@ -21,6 +21,7 @@ export type ScrollViewNativeProps = $ReadOnly<{
   alwaysBounceHorizontal?: ?boolean,
   alwaysBounceVertical?: ?boolean,
   automaticallyAdjustContentInsets?: ?boolean,
+  automaticallyAdjustsScrollIndicatorInsets?: ?boolean,
   bounces?: ?boolean,
   bouncesZoom?: ?boolean,
   canCancelContentTouches?: ?boolean,
--- a/Libraries/Components/ScrollView/ScrollViewViewConfig.js
+++ b/Libraries/Components/ScrollView/ScrollViewViewConfig.js
@@ -24,6 +24,7 @@ const ScrollViewViewConfig = {
     alwaysBounceHorizontal: true,
     alwaysBounceVertical: true,
     automaticallyAdjustContentInsets: true,
+    automaticallyAdjustsScrollIndicatorInsets: true,
     bounces: true,
     bouncesZoom: true,
     canCancelContentTouches: true,
--- a/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm
+++ b/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm
@@ -254,6 +254,12 @@ static void RCTSendPaperScrollEvent_DEPRECATED(UIScrollView *scrollView, NSInteg
     scrollView.snapToOffsets = snapToOffsets;
   }
 
+  if (@available(iOS 13.0, *)) {
+    if (oldScrollViewProps.automaticallyAdjustsScrollIndicatorInsets != newScrollViewProps.automaticallyAdjustsScrollIndicatorInsets) {
+      scrollView.automaticallyAdjustsScrollIndicatorInsets = newScrollViewProps.automaticallyAdjustsScrollIndicatorInsets;
+    }
+  }
+
   if (@available(iOS 11.0, *)) {
     if (oldScrollViewProps.contentInsetAdjustmentBehavior != newScrollViewProps.contentInsetAdjustmentBehavior) {
       auto const contentInsetAdjustmentBehavior = newScrollViewProps.contentInsetAdjustmentBehavior;
--- a/React/Views/ScrollView/RCTScrollView.m
+++ b/React/Views/ScrollView/RCTScrollView.m
@@ -928,6 +928,18 @@ RCT_SET_AND_PRESERVE_OFFSET(setShowsVerticalScrollIndicator, showsVerticalScroll
 RCT_SET_AND_PRESERVE_OFFSET(setZoomScale, zoomScale, CGFloat);
 RCT_SET_AND_PRESERVE_OFFSET(setScrollIndicatorInsets, scrollIndicatorInsets, UIEdgeInsets);
 
+#if defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000 /* __IPHONE_13_0 */
+- (void)setAutomaticallyAdjustsScrollIndicatorInsets:(BOOL)automaticallyAdjusts API_AVAILABLE(ios(13.0))
+{
+  // `automaticallyAdjustsScrollIndicatorInsets` is available since iOS 13.
+  if ([_scrollView respondsToSelector:@selector(setAutomaticallyAdjustsScrollIndicatorInsets:)]) {
+    if (@available(iOS 13.0, *)) {
+      _scrollView.automaticallyAdjustsScrollIndicatorInsets = automaticallyAdjusts;
+    }
+  }
+}
+#endif
+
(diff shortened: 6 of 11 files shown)


justinwh:
There is an extra concern with this PR: changing the default behavior will require developers who are using ScrollView in this way to re-adjust their code.

- for devs who have already added their own workarounds, they'll need to revert them

- for devs that depended on the previous iOS 13 default behavior (and aren't concerned about iOS <=12), they'll need to explicitly add appropriate insets

Based on the low interest in the initial bug, I don't think this is a lot of people, but it's not zero either.

osartun:
What are the next steps for this PR? We have a local patch for RN in our repo with pretty much the same code. I would like to see that issue fixed so that we don't need to reapply our patch with every upgrade of RN.

> There is an extra concern with this PR: changing the default behavior will require developers who are using ScrollView in this way to re-adjust their code.
> 
> * for devs who have already added their own workarounds, they'll need to revert them
> * for devs that depended on the previous iOS 13 default behavior (and aren't concerned about iOS <=12), they'll need to explicitly add appropriate insets

Can't you just put this into a new "breaking" version? Whenever we upgrade we check the changes and are expecting something to break.
Seeing this getting fixed would be a satisfying breaking change though.

jfrolich:
I'd love that someone in the core team would look at this. Without this, it is impossible to get the scroll indicators right on iOS in some scenarios. It's also a bit frustrating because I think a lot of people would fiddle with it and conclude it's not possible to get it right, or after a long time find this patch (like me 😃)

osartun:
There are also multiple issues and implemented workarounds associated to this which would get closed:

* https://github.com/facebook/react-native/issues/26610
* https://github.com/facebook/react-native/issues/28140
* https://github.com/facebook/react-native/issues/28085
* https://github.com/standardnotes/mobile/issues/192
* https://github.com/mochixuan/react-native-drag-sort/pull/92
* https://github.com/shoutem/ui/pull/493
* …

justinwh:
Yeah, crazy huh?  I'm the original submitter—I don't know what else to do to get this fixed.  I documented the issue pretty thoroughly, found the cause, and submitted a fix PR.  I'm not authorized to merge it myself.

As far as I can tell, nobody from the project who is authorized to merge it is aware it exists.

I don't know what to do to get a maintainer's attention.  Today I tried asking in the Reactiflux discord linked from the project website, but didn't get any useful answers there either.

The original issue (which is now just shy of its one-year anniversary) still displays the "Needs: Triage" label...

On the plus side, I can confirm that I've been shipping this fix in my own app for over a year and it has not caused any problems.

lunaleaps:
These changes look good to me! cc @PeteTheHeat or @p-sun for any thoughts?

lunaleaps:
@justinwh Would it be possible to add your test case (demonstrating the scrollbar positions) as an RNTester case? How were you able to display the scrollbars persistently? We're trying to see how we can prevent this from regressing in the future.

justinwh:
> @justinwh Would it be possible to add your test case (demonstrating the scrollbar positions) as an RNTester case? How were you able to display the scrollbars persistently? We're trying to see how we can prevent this from regressing in the future.

@lunaleaps Unfortunately, I captured my screenshots manually without persistent scrollbars (i.e. just kept trying until I got the timing right).

As far as I know the scrollbars cannot be made to display persistently on iOS.  And I think their animation is too quick to capture consistently in a screenshot-based automated test.

I'm not aware of any good option for setting up a test case for this.  (But I'm open to suggestions.)

lunaleaps:
> As far as I know the scrollbars cannot be made to display persistently on iOS. And I think their animation is too quick to capture consistently in a screenshot-based automated test.

Yeah, I agree. I think then a manual test example case of this might be the best we can do? cc @p-sun who had some thoughts on the changes.

justinwh:
I added a second commit with an rn-tester screen that demonstrates the behavior (called `ScrollViewIndicatorInsets`).

When the fix is applied, this shows the scrollbars passing outside the device safe area to the screen edge (as intended).  If you revert the fix, the scrollbars stay inside the device safe area.

lunaleaps:
@facebook-github-bot import

justinwh:
I updated the PR with the changes discussed in the issue:
- Made `automaticallyAdjustsScrollIndicatorInsets` a prop on `ScrollView`
- Made the default `true`

I followed the implementation of the `contentInsetAdjustmentBehavior` prop as my template.  What I pushed is sufficient to get the rn-tester example working.  But I also found places in the tree where these props seem to be defined/redefined that I'm less familiar with.  Should I add this new prop into the following files too?

https://github.com/facebook/react-native/blob/d1ab03235cb4b93304150878d2b9057ab45bba77/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm#L257-L270

https://github.com/facebook/react-native/blob/d1ab03235cb4b93304150878d2b9057ab45bba77/ReactCommon/react/renderer/components/scrollview/ScrollViewProps.cpp#L166-L170

lunaleaps:
> I updated the PR with the changes discussed in the issue:
> 
> * Made `automaticallyAdjustsScrollIndicatorInsets` a prop on `ScrollView`
> * Made the default `true`
> 
> I followed the implementation of the `contentInsetAdjustmentBehavior` prop as my template. What I pushed is sufficient to get the rn-tester example working. But I also found places in the tree where these props seem to be defined/redefined that I'm less familiar with. Should I add this new prop into the following files too?
> 
> https://github.com/facebook/react-native/blob/d1ab03235cb4b93304150878d2b9057ab45bba77/React/Fabric/Mounting/ComponentViews/ScrollView/RCTScrollViewComponentView.mm#L257-L270
> 
> https://github.com/facebook/react-native/blob/d1ab03235cb4b93304150878d2b9057ab45bba77/ReactCommon/react/renderer/components/scrollview/ScrollViewProps.cpp#L166-L170

Re: ScrollViewProps.cpp, I think this is the defaults we're using for the Fabric native view so we should probably set the default of true there as well.

Re: RCTScrollViewComponentView.mm I believe so yes, as in updating the prop value.

Let me see if someone from with more Fabric experience can confirm. cc @p-sun

p-sun:
Yes, @lunaleaps is right. Any new prop should be added to the Fabric component as well. You can add the prop underneath the contentInsetAdjustmentBehavior prop in the two places that Luna linked, add a breakpoint inside RCTScrollViewComponentView.mm, and test it inside the RN Tester app to make sure it works.

Thanks for adding a Switch in your example for this prop. 👍

fkgozali:
Btw, RNTester iOS/Android should already support Fabric, but perhaps with some minor quirks. You can enable Fabric in RNTester iOS manually by setting this env var before `pod install`: https://github.com/facebook/react-native/blob/master/packages/rn-tester/Podfile#L22-L30

```
USE_FABRIC=1 pod install
```

Then at least you can verify whether the updated .mm/C++ files compiled and ran fine.

justinwh:
I added the Fabric changes, and confirmed they compile and run when `USE_FABRIC=1`.

The rn-tester page I added requires `Modal`, which apparently isn't implemented in Fabric yet.  But I confirmed that setting the prop on other ScrollViews in rn-tester has the correct effect.

lunaleaps:
For the example, I noticed that there's no difference between true/false for the property in the case of displaying a Modal with `pageSheet` presentation style. @justinwh was there a reason to include that example? Additionally there's a typo on the `fullScreen` example which I've fixed on the imported PR. I'm going to add some explanation to what we should expect to see when you toggle this on/off. 

Additionally did we discuss adding documentation for this?

justinwh:
@lunaleaps thank you for fixing the typo.

The `pageSheet` modal *does* show different behavior for the checked/unchecked states for me.  I'm not sure why yours doesn't, but since this is an "automatic" property I guess nobody but Apple knows how it actually works.  Maybe it's device or OS version dependent?  (I tested on the Simulator with the `iOS 14.5/iPhone 12 Pro` device type and on a real iPhone 12 Mini running iOS 14.6.)

On both devices I tested, enabling it adds a top inset to the `pageSheet` modal, which is incorrect (I think) since the modal does not underlay the statusbar/notch area.  So the behavior you're getting is actually an improvement.

In any case, what it does when enabled is outside our control.  The important thing is that it leaves the scroll indicator insets alone when disabled, and it looks like that's the case in your screenshots.  So I believe this is acceptable.  If you want to remove the `pageSheet` example from rn-tester, I have no objection.

We didn't discuss documentation.  I assumed the docs were generated from the comments, but it sounds like that's not the case.  Is there a process for synchronizing the updates to the react-native-website repo with React Native releases that I should be aware of?

Here's my proposed documentation for the new prop (let me know if you have feedback on it):
```
automaticallyAdjustsScrollIndicatorInsets
Controls whether iOS should automatically adjust the scroll indicator insets. The default value is true. Available on iOS 13 and later.
@platform ios
```

lunaleaps:
@justinwh Sounds good! I'll go ahead and land.. 

Regarding docs that sounds good! I wonder if we should also link to the UIScrollView docs for the property since. Would you like to make a PR for react-native-website? It should go under the versioned_docs `next` I believe. Let me know if you need more help!

justinwh:
I added a link to the `UIScrollView` docs and sent a PR to the docs repo:

https://github.com/facebook/react-native-website/pull/2661

glenna:
hey @justinwh we are integrating this change into our app now and _finally_ removing our RN patch for this, but are confused a bit on what the default actually is. From the code, it looks like it defaults to `true` / `YES` (in iOS world), as well as in the [RN documentation](https://reactnative.dev/docs/scrollview#automaticallyadjustsscrollindicatorinsets-ios), but in this PR description and commit message, it says

> This change sets the default for this property to NO, making the behavior consistent across iOS versions, and allowing developers full control.

What is meant?

justinwh:
@glenna The docs are correct—the default is `true`.  I didn't realize the PR's merge commit msg comes from the PR's description (which reflects my original approach, but not the final fix) instead of the commit itself (which I *did* update).

Sorry for the confusion.

glenna:
@justinwh not a problem! thanks for the clarification!

PR #29810, opened 2020-08-30
Update iOS Fabric-related files to compile on OSS

empyrical:
## Summary

This pull request updates the Podspecs and associated build scripts, and some source files so they build on OSS. RNTester now compiles with `fabric_enabled` again.

The following changes have been made:

 * Various spots that were pointing to the old `ReactCommon/fabric` location have now been updated to `ReactCommon/react/renderer`
 * Files that were attempting to use internal FB header `FBRCTFabricComponentsPlugins.h` were changed to use `RCTFabricComponentsPlugins.h`
 * `RCTFabricComponentsPlugins` in OSS was updated to include the `Image` fabric component (thanks @tsapeta)
 * Replaced old `generate-rncore.sh` build script with new `generate-rncore.js` script which does not require `flow-node` and uses the `react-native-codegen` API directly, so there is no longer any need for an interim `schema-rncore.json` file.
 * Updated Yoga podspec which wasn't fully synced with changes from the main Yoga repo
 * Updated Fabric podspec with additional needed subspecs

## Changelog

[General] [Fixed] - RNTester compiles with `fabric_enabled` again

## Test Plan

RNTester now compiles and runs in the simulator again when `fabric_enabled` is set to `true`.


hramos:
Our internal CI flagged the changes to RCTFabricComponentsPlugins. Looks like the internal codegen does not include Image in the plugin map. I'll need to address that before we can merge these changes.

empyrical:
One thing I need to do before this can be merged is have the podspec reflect the recent moving of `ReactCommon/turbomodule` to `ReactCommon/react/nativemodule`. I can give you a mention once I've pushed that change.

hramos:
Sounds good. I'm putting the final touches on switching RNTester to use the react-native-codegen modules output, then I'll be looking at getting this PR merged.

empyrical:
Rebased and got it working again! Turns out I was mistaken and did not need to make any changes to `turbomodule`, but I did need to make a small tweak to header search paths in `React-graphics.podspec`.

React/Fabric/Mounting/RCTComponentViewFactory.mm
```diff
@@ -57,8 +57,7 @@ + (RCTComponentViewFactory *)standardComponentViewFactory
   [componentViewFactory registerComponentViewClass:[RCTParagraphComponentView class]];
   [componentViewFactory registerComponentViewClass:[RCTTextInputComponentView class]];
 
-  Class<RCTComponentViewProtocol> imageClass = RCTComponentViewClassWithName("Image");
```
sammy-SC, line 60: Hello,

this needs to stay as is. Is allows us to have multiple implementations of Image component view and change them based on the app.

hramos: I'll revert this on the internal diff. Thanks Sam!

empyrical: Would adding the image component to the OSS `RCTFabricComponentsPlugins.mm` break anything internally at FB?

sammy-SC: Hey @empyrical,

sorry for the slow reply.

"Image" is already in `RCTFabricComponentsPlugins.mm`. The reason why it is here, and other components above, is for preemptive view allocation. 

Please let me know if you have any other questions.

PR #29832, opened 2020-09-02
feat(alert): use support library for alert dialog

DomiR:
## Summary

Instead of using the AlertDialog from the android library we use the AlertDialog from the support library to render dialogs properly on Lollipop phones.

## Changelog

[Android] [Fixed] - Fix AlertDialog for older android devices by using AlertDialog from the androidx support library

## Test Plan

Visual inspection.
With old AlertDialog:
ddd1692b-2bcb-424a-9355-1b54515c6618 - Donald Duck (1)

With compat AlertDialog:
Screenshot_1973-06-21-21-23-28

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/BUCK
@@ -7,6 +7,7 @@ rn_android_library(
     labels = ["supermodule:xplat/default/public.react_native.infra"],
     provided_deps = [
         react_native_dep("third-party/android/androidx:annotation"),
+        react_native_dep("third-party/android/androidx:app"),
         react_native_dep("third-party/android/androidx:core"),
         react_native_dep("third-party/android/androidx:fragment"),
         react_native_dep("third-party/android/androidx:legacy-support-core-ui"),
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/BUCK
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/dialog/BUCK
@@ -6,12 +6,7 @@ rn_android_library(
     is_androidx = True,
     labels = ["supermodule:xplat/default/public.react_native.infra"],
     provided_deps = [
-        react_native_dep("third-party/android/androidx:annotation"),
-        react_native_dep("third-party/android/androidx:app"),
-        react_native_dep("third-party/android/androidx:core"),
-        react_native_dep("third-party/android/androidx:fragment"),
-        react_native_dep("third-party/android/androidx:legacy-support-core-ui"),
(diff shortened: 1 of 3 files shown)


safaiyeh:
cc @dulmandakh

dulmandakh (approved):
LGTM

DomiR:
I don't know how to fix the BUCK files... circleci doesn't seem to build properly, but I compiled it on my machine and the result can be seen above.

DomiR:
I can't wait for the next release and I'm scared to roll with my own fork for the release 😅 

Here https://github.com/DomiR/react-native-compat-alert is reimplementation of the module, that uses androidx.appcompat.app.AlertDialog. Would be nice to see this in the next release.

ReactAndroid/src/main/java/com/facebook/react/modules/dialog/BUCK
```diff
@@ -7,6 +7,7 @@ rn_android_library(
     labels = ["supermodule:xplat/default/public.react_native.infra"],
     provided_deps = [
         react_native_dep("third-party/android/androidx:annotation"),
+        react_native_dep("third-party/android/androidx:app"),
```
dulmandakh, line 10: try changing ** react_native_dep("third-party/android/androidx:app")** with ** react_native_dep("third-party/android/androidx:appcompat-binary")**

dulmandakh:
Please remove everything in provided_deps, and add **react_native_dep("third-party/android/androidx:appcompat")**

DomiR:
@dulmandakh there is still some BUCK file for the test build that needs updating... any ideas?

dulmandakh:
You find a file that fails to compile, and usually there is a BUCK file in the same folder which might require change.

dulmandakh:
I dug a little bit, and couldn't a fix the test failure. It seems like AppCompat components require themed context, ThemedReactContext in RN case, but RN passes ReactApplicationContext to modules constructor which is used by DialogModule to show AlertDialog. Thus tests are failing. Please help @mdvacca

camilossantos2809:
Any updates? I have an app that is used a lot on devices with android 5/lollipop, this improvement would help me a lot. Thanks!

DomiR:
Sucks that internal Facebook build tools are not the OS way. It builds for me, but BUCK needs to be configured differently to allow those tests to pass.

In the meantime you can use my fork, which only replaces that import: https://github.com/DomiR/react-native-compat-alert

safaiyeh:
What versions of Android does this aim to fix?

DomiR:
I think Lollipop and earlier... other components also use the androidx compact lib, so it's not something unusual. Without compat the Alert works on older phones with the limitations mentioned above (no vertical button stacking, overflow on small devices).

Issue #26019, opened 2019-08-11
URL cannot handle "localhost" domain for base url


phallguy:
URL constructor raises a TypeError when using `localhost` as the base url.

React Native version:

System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 4.55 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.6.0 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.9.0 => 16.9.0
    react-native: 0.60.4 => 0.60.4
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce

```javascript
new URL("/home", "http://localhost:3000").toString() // TypeError invalid base URL
```

Describe what you expected to happen:

```javascript
new URL("/home", "http://localhost:3000").toString() // "http://localhost:3000/home"
```

Snack, code example, screenshot, or link to a repository:

https://snack.expo.io/@phallguy/e6bc36

jeswinsimon:
The current regex in [validateBaseUrl](https://github.com/facebook/react-native/blob/0ccedf3964b1ebff43e4631d1e60b3e733096e56/Libraries/Blob/URL.js#L108) expects a base url to be of the format `domain-name.extension`.

I am not sure if the right solution is to accept all domains without domain extension or simply grant an exception for `localhost`. If it is the later then updating the regex to below will work.

[...]

PR #29834, opened 2020-09-02
Fixed missing encoding via URLSearchParams

### LINKED ISSUE #28966: Escape URI components for URLSearchParams
## Description
Currently when using `URLSearchParams` the given values are not encoded (escaped). A comment in the current implementation says that this will not work due to Unicode issues.
https://github.com/facebook/react-native/blob/cd347a7e0ed29ae1049e041fcb34588e1aac76f9/Libraries/Blob/URL.js#L55

I don't know the ins and outs of that particular issue, but I do think that at least escaping URI components (`?,=,/,&,:`) is a good idea. I understand that only escaping those characters and not UTF-8 characters can be seen as a "half" solution. However the current implementation of `URLSearchParams` is not very useful at all. At lease escaping those URI components makes it a lot more useful, and overall a lot safer to use, in situation where no UTF-8 chars are involved.

## React Native version:
System:
    OS: Linux 5.3 Ubuntu 18.04.4 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz
    Memory: 11.28 GB / 15.58 GB
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 12.16.3 - /usr/bin/node
    npm: 6.14.5 - ~/.npm-global/bin/npm
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: https://github.com/expo/react-native/archive/sdk-36.0.1.tar.gz => 0.61.4

scarlac:
Fixes #28966

## Summary

URLSearchParams was largely useless, as it did not encode the parameters at all. This PR simply encodes the parameters as per the standard.

## Changelog

[General] [Fixed] - URLSearchParams now URL encodes passed values

## Test Plan

Ran the new code using jsc (`/System/iOSSupport/System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Helpers/jsc`)
```
_searchParams=[['size', 42],['limit', false],['baz','boing+foo&bang']]; last = this._searchParams.length - 1;
_searchParams.reduce((acc, [key, value], index) => {
  return `${acc}${key}=${encodeURIComponent(value)}${index === last ? '' : '&'}`;
}, '');
// which yields: 
// 'size=42&limit=false&baz=boing%2Bfoo%26bang'
```
And compared it running similar code but with a browser-shipped URLSearchParams in the browser console of this page:
```
_searchParams=[['size', 42],['limit', false],['baz','boing+foo&bang']]; last = this._searchParams.length - 1;
new URLSearchParams(_searchParams).toString()
// which yields:
// "size=42&limit=false&baz=boing%2Bfoo%26bang"
```
Comparing the two:
```
'size=42&limit=false&baz=boing%2Bfoo%26bang' === "size=42&limit=false&baz=boing%2Bfoo%26bang"
// returns true
```

### DIFF
--- a/Libraries/Blob/URL.js
+++ b/Libraries/Blob/URL.js
@@ -99,8 +99,8 @@ export class URLSearchParams {
       return '';
     }
     const last = this._searchParams.length - 1;
-    return this._searchParams.reduce((acc, curr, index) => {
-      return acc + curr.join('=') + (index === last ? '' : '&');
+    return this._searchParams.reduce((acc, [key, value], index) => {
+      return `${acc}${key}=${encodeURIComponent(value)}${index === last ? '' : '&'}`;
     }, '');
   }
 }


charpeni:
`URLSearchParams` has the same issue as `URL`as the implementation inside React Native is all hardcoded and not following the spec.

I would highly recommend migrating over to [`react-native-url-polyfill`](https://github.com/charpeni/react-native-url-polyfill) instead.

Actually, I should deprecate React Native's `URL` and `URLSearchParam`.

scarlac:
@charpeni I'm not sure what you are saying. Nothing is hard-coded about the PR or implementation and this solves an immediate issue with have the implementation. Unless we can integrate a polyfill from react-native-url-polyfill that ships with React Native, I don't see the point of removing it. We'd be needlessly breaking a lot of code bases for no good reason. The feature is helpful it just has this 1 very important bug.

charpeni:
Hey @scarlac.

Sorry for the confusion here, your pull request is definitely the right implementation.

The hard-coded things are coming from React Native itself which is trying to define its own implementation of `URL` and `URLSearchParams`. Unfortunately, this process can't scale well and we'll always have issues with unhandled cases and corner cases. (See: https://github.com/facebook/react-native/issues/26019, https://github.com/facebook/react-native/issues/25717, https://github.com/facebook/react-native/issues/24428, https://github.com/facebook/react-native/issues/16434). As others pull requests, we can't assert that this pull request isn't breaking any existing code (relying on the bad behavior), and merging PRs that are fixing our custom implementation is a back and forth process.

What I'm advocating for, is to deprecate RN's `URL` and `URLSearchParams` in favor of [`react-native-url-polyfill`](https://github.com/charpeni/react-native-url-polyfill) (which is spec compliant). Of course, as you said we can't remove them right away, but we can display a LogBox warning saying that it is deprecated.

[Why we can't ship React Native with `react-native-url-polyfill`.](https://github.com/facebook/react-native/pull/25719#issuecomment-521154629)

How does that sound?

scarlac:
@charpeni Yeah it looks like it's ~40k extra. I don't like it but it's a reasonable trade-off. Agree with your approach then. Thanks for the thorough explanation.

scarlac:
Closing this PR. As per @charpeni 's recommendation an app-installed polyfill is preferred and we should be removing the existing implementation.

PR #29835, opened 2020-09-02
Enable usesCleartextTraffic in main manifest

sunnylqm:
This setting might cause trouble to freshmen and they won't realize it until they send an release apk to their boss which is ...

So why not make it consistent with ios, which enables NSAllowsArbitraryLoads by default.

### DIFF
--- a/template/android/app/src/main/AndroidManifest.xml
+++ b/template/android/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
 
     <application
+      android:usesCleartextTraffic="true"
       android:name=".MainApplication"
       android:label="@string/app_name"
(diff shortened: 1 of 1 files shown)


dulmandakh:
You can have non-HTTPS in debug builds, but not in release builds. It's easy to disable HTTPS requirement in release builds, so I don't like to disable it in RN, thus reduce security/protection for all apps.

sunnylqm:
@dulmandakh I understand your concert. But the fact is that a lot of devs won't realize it until they build a release apk and send to others in a quite formal scenario, which maybe a disaster. Especially when iOS enables non-https connection in both debug and release builds.

dulmandakh:
No, iOS requires HTTPS since a while. And Android requires HTTPS since a while too. All by default.

In RN, both platforms allow HTTP connections on debug builds.

PR #29847, opened 2020-09-03
fix(RCTUITextField): apply workaround for UIKit bug that works for multiline to singleline

DomiR:
## Summary

See https://github.com/facebook/react-native/commit/e020576b34fb6ca6d3f9fe38916844b78a45c0e3 for rational.

This bug in UIKit does apply for all UITextInput views. So this PR copies over the fix from multiline-view to singleline-view.

## Changelog

[iOS] [Fixed] - Make selectTextOnFocus for iOS single line text input fields work every time

## Test Plan

Visual inspection. Same bug fix for multiline does work for singleline.

### DIFF
--- a/Libraries/Text/TextInput/Singleline/RCTUITextField.m
+++ b/Libraries/Text/TextInput/Singleline/RCTUITextField.m
@@ -192,6 +192,17 @@
   _textWasPasted = YES;
 }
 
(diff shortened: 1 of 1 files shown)


DomiR:
I've rebased to current main. First PR had flaky iOS test, which now passes. Hope this could be merged at some point...

PR #29870, opened 2020-09-05
[StyleSheet]: Import validations only in development

ecreeth:
## Summary

This PR only Import style validations in development, since it is currently included in production

| Before (8.63 KB) |  After (8.31 KB  |
|----------|:-------------:|
| before | after

## Changelog

[JavaScript] [Changed] - [StyleSheet]: Import validations only in development

## Test Plan

1. yarn test passed
2. RNTester Android app build successfu
3. An invalid style property must throw an error
Screenshot_1599331090

### DIFF
--- a/Libraries/StyleSheet/StyleSheet.js
+++ b/Libraries/StyleSheet/StyleSheet.js
@@ -10,11 +10,15 @@
 
 'use strict';
 
+let StyleSheetValidation = null;
 const PixelRatio = require('../Utilities/PixelRatio');
 const ReactNativeStyleAttributes = require('../Components/View/ReactNativeStyleAttributes');
-const StyleSheetValidation = require('./StyleSheetValidation');
 
 const flatten = require('./flattenStyle');
+
+if (__DEV__) {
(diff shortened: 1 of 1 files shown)


safaiyeh:
Looks like StyleSheet still needs validate, checkout the build errors. Is ti possible to migrate?

ecreeth:
I don't know why is failing, this is only used here
https://github.com/facebook/react-native/blob/f5c246d50e46f57edc52ec06c5fba9e5ea9f8eb7/Libraries/StyleSheet/StyleSheet.js#L364-L371

ecreeth:
@safaiyeh Tests are now passing.

ikara5 (approved):

TheSavior:
You should enable inline-requires in your metro config, that will do this optimization across the board instead of one off. We avoid doing these changes inline as it can become hard to manage and reason about at scale. We've found inline requires to be a more sustainable solution to this problem. Inline requires will also be enabled by default in 0.64.

Issue #29455, opened 2020-07-21
[0.63.0][IOS] Modal onDismiss no longer working on IOS


raymochi:
## Description
On 0.62.2 onDismiss functions were called on modals being closed. After upgrading to 0.63.0, onDimiss functions are no longer called. This was working fine before upgrading from 62.2 to 63.

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.

1. Import { Modal } from react-native
2. Add modal to render with a function passed into onDismiss property

## Expected Results
onDimiss function should be called

chrisglein:
Realize that 0.63 is new enough that you can't repro in a Snack, but it would help if you pointed to a Snack that shows this working pre-0.63.

andres-torres-marroquin:
@chrisglein I made an example using Expo SDK 38.0.0, which shows that the behavior is correct in RN 0.62 https://snack.expo.io/@srpoder/excited-cheese

Also the RN Documentation specifies that onDismiss should be available on RN 0.63 https://reactnative.dev/docs/modal#ondismiss

wincjh:
+1 onDismiss not called and not open second modal when i have closed fist modal and open second modal

diolegend:
Same issue, i upgraded to 0.63 and now when i set the modal visible to false, it don't invoke onDismiss, but when i close the modal of react native navigation, it surprisingly invoke onDismiss, but i don't need when i close the modal of react native navigation. But i need as usual, invoked when i set to false the visible of react native modal.

Steps:
1. Navigation.showModal https://wix.github.io/react-native-navigation/api/modal/#showmodal
2. showModal react native
3. closeModal react native (set to false visible)
4. No invoke onDismiss
5. Navigation.dismissAllModals()
6. Surprisingly run onDismiss of react native modal

kkoudev:
This fix in RN 0.62.1 doesn't seem to be in RN 0.63.
Is this the cause?
https://github.com/facebook/react-native/commit/27a3248a3b37410b5ee6dda421ae00fa485b525c

bitcrumb:
We're seeing the same behavior. The onDismiss callback is never called when hiding the modal. I can also confirm the behavior as explained by @diolegend.

imnotjames:
In case it helps - I cannot seem to get `onDismiss` to ever fire on the docs example expo snack using Android either.

MichaelKim39:
Both onDismiss and onHideModal not working for me, tried using the expo snack and also did not work

[...]

PR #29882, opened 2020-09-07
[iOS] Fix cannot working Modal's onDismiss.

kkoudev:
Fixes: #29455

## Summary

Modal's onDismiss is not called on iOS.
This issue occurred the commit https://github.com/facebook/react-native/commit/bd2b7d6c0366b5f19de56b71cb706a0af4b0be43 and was fixed the commit https://github.com/facebook/react-native/commit/27a3248a3b37410b5ee6dda421ae00fa485b525c.
However, the master and stable-0.63 branches do not have this modified commit applied to them.

## Changelog

[iOS] [Fixed] - Modal's onDismiss prop will now be called successfully.

## Test Plan

Tested on iOS with this change:

1. Set function Modal's onDismiss prop.
1. Set Modal's visible props is true. (show Modal)
1. Set Modal's visible props is false. (close Modal)
1. The set function in onDismiss is called.

### DIFF
--- a/Libraries/Modal/Modal.js
+++ b/Libraries/Modal/Modal.js
@@ -12,6 +12,9 @@
 
 const AppContainer = require('../ReactNative/AppContainer');
 const I18nManager = require('../ReactNative/I18nManager');
+import NativeEventEmitter from '../EventEmitter/NativeEventEmitter';
+import NativeModalManager from './NativeModalManager';
+const Platform = require('../Utilities/Platform');
 const React = require('react');
 const ScrollView = require('../Components/ScrollView/ScrollView');
 const StyleSheet = require('../StyleSheet/StyleSheet');
@@ -25,6 +28,11 @@ import type {RootTag} from '../ReactNative/RootTag';
 import type {DirectEventHandler} from '../Types/CodegenTypes';
 import {type EventSubscription} from '../vendor/emitter/EventEmitter';
 import RCTModalHostView from './RCTModalHostViewNativeComponent';
+
+const ModalEventEmitter =
+  Platform.OS === 'ios' && NativeModalManager != null
+    ? new NativeEventEmitter(NativeModalManager)
+    : null;
 
 /**
  * The Modal component is a simple way to present content above an enclosing view.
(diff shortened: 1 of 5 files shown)


imnotjames:
Just to understand - this more or less reverts the code that attempted to make this feature compatible with Android, right?

kkoudev:
@imnotjames Modal's onDismiss props is only support iOS.
Also RN Documentation specifies that onDismiss should be available on RN 0.63.
https://reactnative.dev/docs/modal#ondismiss

imnotjames:
Right - before the attempt to make it work with Android it was intended to be iOS only.  It looks like the documentation was never updated, so that's great.  Documentation is in line with this PR already :)

chj-damon:
can someone merge this pr quickly?

luco:
@shergin Merge this, please.

kelset:
hey folks - for what it's worth, I've surfaced this PR to the FB team and they are aware. But it seems that the problem is that they are struggling to have time to dedicate to this and other PRs :(

sammy-SC:
Hello @kkoudev,

thank you for the PR.

Is `onDismiss` function called if the modal is removed from React rather than using visible prop? 
I don't think it will be because the event subscription will be removed before the event is fired.

kkoudev:
@sammy-SC Yes.
That case onDismiss function is not called.
The onDismiss function is called when the visible property is set to false after the modal visible property has been set.
Because this function is equivalent of the UIViewController's dismiss completion block in iOS.

sammy-SC:
@kkoudev 

It feels like we are trading one onDismiss for another.
The [documentation](https://reactnative.dev/docs/modal/) just says it is called when the modal is dismissed, to my understanding onDismiss should be called in both cases. 

Let's merge this PR until someone has time to properly fix it.

kelset:
gentle nudge to @sammy-SC, it looks like it's failing to import it :(

sammy-SC:
@kelset thanks for the remainder. There were issues with import and I was on vacation for most of the last week. I'm taking a look at it now.

cxrod:
@kkoudev this was released in 0.63.4?

kkoudev:
@cxrod
No. I checked v0.63.4, but this fix was not included.

@kelset 
In what version will this fix be released?

ecreeth:
@kkoudev in [0.64](https://github.com/facebook/react-native/commits/0.64-stable)

kkoudev:
@ecreeth thx!

cxrod:
@kkoudev @ecreeth Thanks!

PR #29903, opened 2020-09-09
Prevent TextInputExample from Redboxing

NickGerleman:
## Summary

Freestanding text in these examples cause the public dev-mode React renderer to redbox saying:
> Text strings must be rendered within a <Text> component.

 See https://snack.expo.io/@chrisglein/repro5688 to repo the same behavior on iOS or Android. Remove the offending text to prevent the example from redboxing.

Relates to https://github.com/microsoft/react-native-windows/issues/5688

## Changelog

[General] [Fixed] - Prevent TextInputExample from Redboxing

## Test Plan

Making this change stopped RNW from redboxing on the example, and in the min repo case stops Android/iOS from redboxing.

### DIFF
--- a/packages/rn-tester/js/examples/TextInput/TextInputExample.android.js
+++ b/packages/rn-tester/js/examples/TextInput/TextInputExample.android.js
@@ -339,7 +339,6 @@ exports.examples = ([
             returnKeyType="done"
             multiline={true}
             style={{maxHeight: 400, minHeight: 20, backgroundColor: '#eeeeee'}}>
-            generic generic generic
             <Text style={{fontSize: 6, color: 'red'}}>
               small small small small small small
             </Text>
@@ -347,7 +346,6 @@ exports.examples = ([
             <Text style={{fontSize: 30, color: 'green'}}>
               huge huge huge huge huge
             </Text>
-            generic generic generic
           </AutogrowingTextInputExample>
         </View>
       );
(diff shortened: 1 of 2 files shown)


appden:
@NickGerleman I notice the Android implementation (which supports children) will wrap multiple children in a `<Text>` [here](https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L1114-L1116). Is this done on Windows as well? I didn't in the repro link where it reproduces on Android.

NickGerleman:
@appden it looks like that code is for inline children inside of a TextInput component. The repro we have is for raw text outside of a Text component nested in a <View>.

NickGerleman:
Oh wait I totally forgot the details of this PR being for TextInput. Let me take a look :)

NickGerleman:
The example repro was a bit problematic since it embedded the string children inside of a view instead of forwarded to a component, but the following example seems to redbox for both Android and iOS if in an expo snack:

```js
export default function App() {
  return (
      <TextInput style={styles.container}>
      something here
      <Text style={styles.paragraph}>
        Change code in the editor and watch it change on your phone! Save to get a shareable url.
      </Text>
      something there
      </TextInput>
    </View>
  );
}
```

Looking at the above code, it looks like Android but not iOS wraps all children into a single Text component, but doesn't seem to be doing a transformation of each.

This doesn't repro in RNTester in master, but I wonder if that might be down to a difference in the renderer used maybe? I know there are different ones with some different configs. 

Note for reproing, that it looks like Snack has started to auto-close out of redboxes and show errors in its own UI. E.g. this 
image

NickGerleman:
It looks like only 8/12 renderers try to detect and throw the error. https://github.com/facebook/react-native/search?q=%22Text+strings+must+be+rendered%22

PR #29477, opened 2020-07-23
[RN v0.63.2] Fix Xcode bundler in staging and release

makarkotlov:
Revert "feat: improve monorepo support by removing redundant PROJECT_ROOT (#28354)"

This reverts commit a8e85026cfa60056b1bcbcd39cde789e4d65f9cb.

## Summary

This commit a8e85026cfa60056b1bcbcd39cde789e4d65f9cb somehow broke the bundler when making a staging or release build in Xcode that results in unresolved files and main.jsbundle nonexistance issue. I figured this out by replacing react-native-xcode.sh from RN v0.63.2 by the one from v0.62.2 where everything works just fine and then reverting the changes line by line. It seems like this pr will fix similar issues stated here https://stackoverflow.com/questions/62806319/main-jsbundle-does-not-exist-this-must-be-a-bug-with-main-jsbundle-issue-afte/62829256#62829256 and here https://github.com/facebook/react-native/issues/29205

## Changelog

[iOS] [Fixed] - fix "main.jsbundle does not exist" issue

## Test Plan

With react-native-xcode.sh from RN v0.63.2
image

With my changes
image

PR #24156, opened 2019-03-26
Fix: use NODE_BINARY when starting packager

Esemesek:
## Summary

Fix packager script to use `NODE_BINARY` env variable.

Should fix #22868 

## Changelog

[iOS] [Fixed] - Use `NODE_BINARY` env variable in `packager.sh` script

## Test Plan

A launched app in Debug and Release mode and running `packager.sh` directly from `node_modules`.

cc. @Nabellaleen - since I wasn't able to reproduce the crash, can you verify that this does work in your project?

PR #29914, opened 2020-09-09
feat: Enable Hermes to work on iOS

### LINKED ISSUE #27845: [0.62.0-rc.3] Catalyst macOS build issue typedef redefinition with different types

### LINKED ISSUE #28810: Catalyst + ReactNative: Typedef redefinition with different types ('uint8_t' (aka 'unsigned char') vs 'enum clockid_t') error when building for mac target
## Description
I'm getting the  following errors when building for mac. This is a default application, created via `react-native init`.
Error 1: `Typedef redefinition with different types ('uint8_t' (aka 'unsigned char') vs 'enum clockid_t')`
Error 2: `ld: in /Users/jh/Documents/projects/VST/Trial3/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a(cryptlib.o), building for Mac Catalyst, but linking in object file built for iOS Simulator, file '/Users/jh/Documents/projects/VST/Trial3/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a' for architecture x86_64`

## React Native version:
```
MacOS: 10.15.4
Xcode: 11.4
RN: 0.62.0-rc.5
RN Cli: 4.8.0
```

## Steps To Reproduce

1. `react-native init App`
2. `react-native run-ios` - Works fine
3. Open `App.xcworkspace`, select mac as a target and clean build.

## Expected Results
Application working on Mac too.

image

I commented out the following line in Time.h:
```
//typedef uint8_t clockid_t;
extern "C" int clock_gettime(clockid_t clk_id, struct timespec* ts);
extern "C" int clock_getres(clockid_t clk_id, struct timespec* ts);
#endif
```

Then tried to build again and the Error 1 disappeared but Error 2 still exists:
image

Should I do something for migration to build for mac?

grabbou:
## Summary

This PR makes it possible to build iOS applications with Hermes. Note that it doesn't work with `use_frameworks!` just yet.
 
Fixes #27845 (by downgrading iOS deployment target for RCT-Folly to 9.0)
Fixes #28810 (as above)

Checklist:
- [x] Adjust release scripts to create Hermes bytecode bundle
- [x] Release new Hermes npm package that includes iOS files (unreleased right now, if you want to try locally, you have to clone Hermes and `yarn link` its master to this project)
- [x] Test on a new React Native application in both Debug and Release (Device)
- [x] Test on an RNTester application in both Debug and Release (Device)
- [x] Add missing `i386` to Hermes framework and enable Bitcode
- [x] Inspect CI failures for possible regressions
- [x] Resolve Folly issue as reported #27845 and #28810
- [x] Release new Hermes and test against it that everything works

## Changelog

[IOS] [FEATURE] - Enable Hermes on iOS
[INTERNAL] - Upgrade to CocoaPods 1.10.0 to resolve Xcode 12.0 issues
[INTERNAL] - Upgrade to Xcode 12.0 on the CircleCI
[INTERNAL] - Fix building RNTester in Release mode
[INTERNAL] - Fix build-time errors of `libevent` with `use_frameworks!`
[INTERNAL] - Introduce `USE_HERMES` variable and test all RNTester configurations on the CI
[INTERNAL] - Do not fetch CocoaPods repository since we're using CDN anyway

## Test Plan

Turn on `hermes_enabled` to true in your `Podfile`, install pods, and run the iOS application. Your app should be running Hermes now. 

Preview: (note "Engine: Hermes")

Screenshot 2020-09-09 at 19 22 32


React-Core.podspec
```diff
@@ -70,14 +70,14 @@ Pod::Spec.new do |s|
...
-    ss.platforms = { :osx => "10.14" }
+    ss.platforms = { :osx => "10.14", :ios => "10.0" }
     ss.source_files = "ReactCommon/hermes/executor/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/chrome/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/detail/*.{cpp,h}"
     ss.pod_target_xcconfig = { "GCC_PREPROCESSOR_DEFINITIONS" => "HERMES_ENABLE_DEBUGGER=1" }
     ss.dependency "RCT-Folly/Futures"
-    ss.dependency "hermes", "~> 0.6.0"
+    ss.dependency "hermes"
```
grabbou, line 80: I think since `hermes` dependency is always pulled from the local `npm` package, I don't think it makes sense to a) keep version requirement in two places and b) read a version from `package.json` here. The main reason against doing the latter is that CocoaPods has slightly different (and incompatible with npm) syntax for specifying the minimum supported version.

Since this package is part of React Native where we do have control over the Hermes version (and set a requirement on it inside `package.json`), I decided to strip it entirely.

grabbou: CC: @alloy

package.json
```diff
@@ -97,6 +97,7 @@
     "event-target-shim": "^5.0.1",
     "fbjs-scripts": "^1.1.0",
     "hermes-engine": "~0.6.0",
+    "hermes-engine-darwin": "0.6.0",
```
grabbou, line 100: Not sure what's the latest versioning pattern, but I followed what we do for `hermes-engine` already. Note that `hermes-engine` could potentially be renamed to `hermes-engine-android`.

third-party-podspecs/RCT-Folly.podspec
```diff
@@ -97,20 +97,21 @@ Pod::Spec.new do |spec|
...
                            'folly/concurrency/*.{h,cpp}',
-                           'folly/system/{ThreadId,ThreadName}.{h,cpp}',
+                           'folly/system/{ThreadId,ThreadName,HardwareConcurrency}.{h,cpp}',
                            'folly/synchronization/*.{h,cpp}',
                            'folly/synchronization/detail/*.{h,cpp}',
-                           'folly/experimental/{ExecutionObserver,ReadMostlySharedPtr,TLRefCount}.{h,cpp}',
-                           'folly/io/async/{AsyncTimeout,DelayedDestruction,DelayedDestructionBase,EventBase,EventBaseManager,EventHandler,EventUtil,HHWheelTimer,NotificationQueue,Request,TimeoutManager,VirtualEventBase}.{h,cpp}',
+                           'folly/experimental/{ExecutionObserver,ReadMostlySharedPtr,SingleWriterFixedHashMap,TLRefCount}.{h,cpp}',
+                           'folly/io/async/{AsyncTimeout,DelayedDestruction,DelayedDestructionBase,EventBase,EventBaseManager,EventBaseBackendBase,EventHandler,EventUtil,HHWheelTimer,HHWheelTimer-fwd,NotificationQueue,Request,TimeoutManager,VirtualEventBase}.{h,cpp}',
```
grabbou, line 106: These were required to make `RNTester` work, I was adding one after another until all errors were gone.

React-Core.podspec
alloy: Yep, makes sense 👍

package.json
```diff
@@ -97,6 +97,7 @@
     "event-target-shim": "^5.0.1",
     "fbjs-scripts": "^1.1.0",
     "hermes-engine": "~0.6.0",
+    "hermes-engine-darwin": "~0.6.0",
```
alloy, line 100: This should be a peer dependency, because auto-linking will otherwise always try to enable it when it's installed, no?

packages/react-native-codegen/android/gradlew.bat
```diff
@@ -1,89 +1,89 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
TheSavior, line 1: Did this file change line endings?

alloy (commented):
This is amazing! Great stuff as always, @grabbou 👏 

Thus far the changes are looking good to me, but will review again when it's no longer marked as WIP.

packages/react-native-codegen/android/gradlew.bat
grabbou: Could be. I will figure out what's wrong with it tomorrow, don't want it to be part of the commit.

grabbou:
Thank you @alloy for your swift review, appreciate it! Now time to move it out of WIP stage

package.json
grabbou: Well, I will double check that tomorrow but...theoretically, it could be a peer dependency, but for a different reason.

Auto linking will only link first-level dependencies, so the ones you have defined in your `package.json` yourself, not the ones inside React Native.

Making it a `peerDependency` would make sense if this package was meant to be installed by end users *in order to enable Hermes*. 

Since we plan to a) turn this to be the default in the future, b) make it turnable by a change to a flag, having it as a part of dependencies is a natural direction.

packages/react-native-codegen/android/gradlew.bat
grabbou: Funny thing tho is, I tried to: a) copy with Github UI manually this file from `master` to this branch b) `git checkout origin master -- <path_to_this_file>`, and both scenarios yield same result - really awkward diff.

TheSavior:
This looks pretty good to me, except for the line endings (I think that's the problem) for `gradlew.bat`?

packages/react-native-codegen/android/gradlew.bat
appden: I've had this same issue with these `*.bat` files. I resolved it for myself with some combination of fiddling with `git config core.autocrlf`, `git rm --cached -r .`, and finally `git reset --hard`, but I don't remember exactly what the silver bullet was. [This StackOverflow](https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf) page is what helped me resolve it.

empyrical: The line endings in that file were likely getting affected by the `.gitattributes` files that were present in the repo. It was happening to me as well on another pull I was working on. [A recent commit](https://github.com/facebook/react-native/commit/fb354cba05a71041775a03a3a26122e01b4782a1) fixed this, so hopefully this shouldn't keep happening in the future.

TheSavior: @empyrical, are you saying that this change is fine because now line endings are normalized and it won't have this problem in future PRs? Aka, this should have been modified in the change removing gitattributes but wasn't.

Or are you saying that something else is broken here, requires fixing, and this PR shouldn't include these changes?

empyrical: The change to this file should be removed from this diff, IMO.

What I meant to say was Git kept incorrectly trying to change the line endings in this batch file because of those nested `.gitattributes` files, and like how @grabbou was experiencing, no matter how much one tried to undo the line ending changes, it would keep getting changed. This problem shouldn't happen again in the future, because those files were removed.

package.json
alloy: > Since we plan to make it turnable by a change to a flag, having it as a part of dependencies makes it a bit easier.

I see what you mean, but I don’t really like the idea of pulling in large artefacts for platforms one may not use. However, I guess the same applies to the android engine one, which is already listed as a plain dependency.

Long term I want to get rid of those flags in the Podfile and the need for a separate file that specifies all transitive dependencies, but that will take a while longer.

In the end, I’m on the fence about doing this now. Do others have opinions?

mrousavy:
I'm sorry, did I miss something? I though Apple didn't allow other JS engines than their own WebKit JS Engine to be used in iOS apps?

sunnylqm:
There will be bytecode instead of js code so i think it complies with apple policy



---Original---

packages/react-native-codegen/android/gradlew.bat
grabbou: Will update the PR

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
```diff
@@ -589,7 +589,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\nexport PROJECT_ROOT=$SRCROOT/..\nexport SOURCEMAP_FILE=../sourcemap.ios.map\n# export FORCE_BUNDLING=true\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\nexport SOURCEMAP_FILE=../sourcemap.ios.map\n# export FORCE_BUNDLING=true\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
```
grabbou, line 592: @alloy, I see that you have added this in the previous PR (that you did). 

Ultimately, you're overwriting PROJECT_ROOT in order to correctly read HERMES_PATH from root `node_modules`. I decided to slightly update the script to make this one unnecessary. 

We have this attitude inside CLI (and generally, when it comes to scripts) to avoid setting paths that are related to project roots etc., to not complicate the setup for monorepo users.

scripts/hermes-binary.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+[ -z "$HERMES_PATH" ] && export HERMES_PATH="$REACT_NATIVE_DIR/../hermes-engine-darwin"
```
grabbou, line 7: By default, Hermes engine will get hoisted and will live next to React Native root. 

Also, decided to ask users to set path to a folder, not to a Hermes executable. I believe it is easier to speak to React Native developers in terms of `npm` packages, rather than asking them to set a path to an executable (they need to know that there's "destroot/bin" folder in case things go wrong). 

Relying on a package makes it a bit easier.

grabbou: Theoretically, we could check if HERMES_PATH="$REACT_NATIVE_DIR/node_modules/hermes-engine-darwin" exists, which is true for RNTester or when hoisting is disabled (and Hermes dependency lives inside React Native own node_modules).

Shall we? Or the error message below is good enough?

scripts/react-native-xcode.sh
```diff
@@ -145,11 +146,6 @@ case "$PLATFORM_NAME" in
     ;;
 esac
 
-USE_HERMES=
-if [[ "$BUNDLE_PLATFORM" == "macos" && -f "$HERMES_PATH" ]]; then
-  USE_HERMES=true
-fi
```
grabbou, line 151: Looks like the approach here was: "If platform is Mac and Hermes exists" which assumed Hermes depedency is optional and installed on per project base (which is aligned with your previous comment @alloy).

Did a little update - USE_HERMES can be (and should) be set by a user explicitly. No platform checks.

grabbou:
Code-wise it's ready. Will perform some testing tomorrow to ensure everything works as expected from the UX perspective and will update this PR with minor tweaks.

After that, this one is ready to be reviewed / merged.

packages/react-native-codegen/android/gradlew.bat
grabbou: I have cherry-picked a commit that you have referenced @empyrical and it fixed the line endings. Resolving the conversation.

I will rebase this PR on top of master later and the diff should get cleaned up.

grabbou:
Meanwhile, we're waiting for a new Hermes release that will make these tests pass. It should be up really soon.

package.json
grabbou: > Long term I want to get rid of those flags in the Podfile and the need for a separate file that specifies all transitive dependencies, but that will take a while longer.

I like that goal. Would love to learn more on how this could be achieved, but I guess that's something we can discuss in parallel!

packages/rn-tester/Podfile
```diff
@@ -13,12 +13,12 @@ def pods()
 
   # Enable TurboModule
   prefix_path = "../.."
-  use_react_native!(path:prefix_path)
+  use_react_native!(path: prefix_path, hermes_enabled: true, hermes_path: "#{prefix_path}/../hermes")
```
grabbou, line 16: Open question: do we want to have Hermes enabled by default in the RNTester?

CC: @TheSavior

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
TheSavior: Is this going to be a script that people have to manually modify in their xcode project UI? If so, should this be a separate script that is executed from the step to make future project upgrades easier?

Huxpro:
@grabbou Heeey Mike, here is a fresh [hermes-engine-darwin-v0.7.0.tgz](https://9577-154201259-gh.circle-artifacts.com/0/tmp/hermes/output/hermes-engine-darwin-v0.7.0.tgz). Would you mind testing if it works correctly with this PR?

packages/rn-tester/Podfile
TheSavior: That's probably a good choice to make sure we don't break this

mhorowitz: Why does this override the default at https://github.com/facebook/react-native/blob/941bc0ec195716e6a505a3c3a67f97a87ea9bcdc/scripts/react_native_pods.rb#L9 ?  Is this the path to the rpm, or to the compiler?

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
alloy: Having less such variables is indeed a good goal, but please also look at https://github.com/facebook/react-native/pull/29477 and see if that can be consolidated.

packages/rn-tester/Podfile
alloy: > Open question: do we want to have Hermes enabled by default in the RNTester?

In macOS’ RNTester we have enabled it by default.

alloy: > Why does this override the default at
> 
> https://github.com/facebook/react-native/blob/941bc0ec195716e6a505a3c3a67f97a87ea9bcdc/scripts/react_native_pods.rb#L9
> 
> ? Is this the path to the rpm, or to the compiler?

@mhorowitz It’s the path to the `npm` package on disk. The reason it’s overriden is because RNTester in this repo isn’t at the default place that a RN app repo would look like.

scripts/hermes-binary.sh
```diff
@@ -0,0 +1,17 @@
...
+
+export HERMES_CLI = "$HERMES_PATH/destroot/bin/hermesc"
+
+if [[ $USE_HERMES == true && ! -d "$HERMES_PATH" ]]; then
+  echo "error: Can't find Hermes executable - directory `$HERMES_PATH` doesn't exist. " \
+       "If you have a non-standard project structure, select your project in Xcode, find " \
+       "'Build Phases' - 'Bundle React Native code and images' and set `HERMES_PATH` to an " \
+       "absolute path to a `hermes-engine-darwin` package inside your `node_modules`" >&2
```
alloy, line 15: Probably not absolute, as that’s not portable between machines. Does the code need updating to support relative paths?

scripts/hermes-binary.sh
alloy: Sure, seems a fair assumption 👍

alloy: I’m a little unclear on the need for this to be a separate script, can you elaborate on that?

scripts/react-native-xcode.sh
alloy: > Did a little update - USE_HERMES can be (and should) be set by a user explicitly.

Why have another extra option? Having the user install the engine is already an explicit choice. It’s also not clear to me in this PR _where/how_ the user should set that variable.

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
grabbou: Yes. Sounds good! I am keeping that default assingement inside the `react-native-xcode.sh`. Just removed setting it in RNTester itself, where it's not necessary.

scripts/hermes-binary.sh
grabbou: Just followed a similar pattern that we currently have for `node-binary.sh`. Not sure if that's the ideal way going forward, but I felt like this was the pattern.

scripts/react-native-xcode.sh
grabbou: > Having the user install the engine is already an explicit choice.

In React Native itself, the engine is a dependency and will be shipped just like an Android one. Users will not have to install it, just enable it by turning the flag.

I feel like relying on its presence inside `node_modules` might be a bit confusing, say - it's still in `node_modules` while somebody finished experimenting with Hermes and wanted to get back to regular setup. 

While this pattern generally makes sense to me, I don't think it's what RN users would expect. On contrary, we already use flags like that one throughout the app.

grabbou: You would set it inside `Bundle JS` script phase, where you set other variables that you want to pass to `react-native-xcode.sh`.

scripts/hermes-binary.sh
grabbou: Relative path from project root would work. Good catch!

grabbou: I think users should set it absolute to avoid issues with PWD, e.g. by using the `SRCROOT` variable, which is already incorporated inside `react-native-xcode.sh`:

```
"$SRCROOT/../../scripts/react-native-xcode.sh"
```

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
grabbou: Changes inside #29477 are not yet in this PR, but there should be no conflict between these two (eventually, we need to find what broke the fix in that PR)

scripts/react-native-xcode.sh
grabbou: Ideally, we would do something like `#if __has_include(<hermes/hermes.h>)` to detect if Hermes is actually included. Unfortunately, I am not sure whether we can detect that from bash script, that's why I ended up using USE_HERMES flag.

scripts/hermes-binary.sh
alloy: Well, I guess the same question applies, why was that necessary? (Perhaps a git blame will give us the reason.) I personally don't like adding indirection when it isn’t serving a reason, but also don’t feel strong about this; just want to understand 😄

alloy: Looks like it was introduced here, but I couldn’t find a reason for making it a separate file https://github.com/facebook/react-native/pull/24156

scripts/hermes-binary.sh
alloy: Hmm, I guess this is ambiguous language then, because I would consider the example you gave to be a relative one. I.e. before expansion it is relative to `$SRCROOT`.

scripts/react-native-xcode.sh
alloy: > I don't think it's what RN users would expect. On contrary, we already use flags like that one throughout the app.

> You would set it inside `Bundle JS` script phase, where you set other variables that you want to pass to `react-native-xcode.sh`.

Ok, consistency is worth something 👍

scripts/hermes-binary.sh
grabbou: Would you be able to help me here with a better wording? That would be greatly appreciated!

scripts/hermes-binary.sh
grabbou: Let's move to a single file.

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
grabbou: I am resolving this based on your +1

scripts/hermes-binary.sh
alloy: How about an example?

> If you have a non-standard project structure, locate the 'Bundle React Native code and images' build phase in your Xcode project and set the `HERMES_PATH` variable to the location of the `hermes-engine-darwin` package. For example: `$SRCROOT/../../node_modules/hermes-engine-darwin`

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
```diff
@@ -589,7 +589,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\nexport SOURCEMAP_FILE=../sourcemap.ios.map\n# export FORCE_BUNDLING=true\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\nexport SOURCEMAP_FILE=../sourcemap.ios.map\nexport FORCE_BUNDLING=true\nexport USE_HERMES=true\nexport PROJECT_ROOT=\"$SRCROOT/../../\"\nexport HERMES_PATH=\"$SRCROOT/../../node_modules/hermes-engine-darwin\"\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
```
grabbou, line 592: We are explicitly setting `PROJECT_ROOT` to `../../../` which is the root of the entire repository. This is how RNTester is bundled in `DEBUG` (running `yarn start` from the root) and w/o this setting, Metro will not work (at a RNTester level, inside `packages/rn-tester`, there is no `react-native` dependency declared and it can't find `ios` and `android` platforms).

grabbou: Before this quick fix, there was no `main.jsbundle` outputted from the `react-native-xcode.sh` w/o any proper error message.

scripts/react-native-xcode.sh
```diff
@@ -194,14 +194,12 @@ else
   if [[ $EMIT_SOURCEMAP == true ]]; then
     EXTRA_COMPILER_ARGS="$EXTRA_COMPILER_ARGS -output-source-map"
   fi
-  HBC_FILE="$CONFIGURATION_BUILD_DIR/$(basename $BUNDLE_FILE)"
```
grabbou, line 192: @alloy, decided to remove this variable as it's essentially `BUNDLE_FILE` at this point. 

For reference, `BUNDLE_FILE` is `CONFIGURATION_BUILD_DIR/main.jsbundle`.

scripts/react-native-xcode.sh
```diff
@@ -194,14 +194,12 @@ else
...
-  "$HERMES_CLI" -emit-binary $EXTRA_COMPILER_ARGS -out "$HBC_FILE" "$BUNDLE_FILE"
-  mv "$HBC_FILE" "$DEST/"
-  BUNDLE_FILE="$DEST/main.jsbundle"
+  "$HERMES_CLI" -emit-binary $EXTRA_COMPILER_ARGS -out "$DEST/main.jsbundle" "$BUNDLE_FILE"
```
grabbou, line 197: Decided to remove `mv` stage and write directly into `$DEST`.

scripts/react-native-xcode.sh
TheSavior: Another reason is if people want to ship two versions of their app to experiment with, one running on Hermes, and one without. Without a flag they'd have to change their package.json to build the two different variants.

grabbou:
Running RNTester in `debug` mode works. The `react-native-xcode.sh` script works as well and produces correct files.

Note: Currently, `RNTester` fails to build in `release` mode due to a link-time error. I am not sure whether this is a result of this PR, or whether `RNTester` is failing on master as well. I will double-check and let you know.

I think it likely that `RNTester` is failing on `master` too, because [the `react-native-xcode.sh` script was broken for RNTester](https://github.com/facebook/react-native/pull/29914/commits/cb01604e09403e0738311060a26375333dca5b39#r488788079) and didn't output any `main.jsbundle`. Fixed in this PR as well.

scripts/react-native-xcode.sh
```diff
@@ -194,14 +194,12 @@ else
...
+  "$HERMES_CLI" -emit-binary $EXTRA_COMPILER_ARGS -out "$DEST/main.jsbundle" "$BUNDLE_FILE" 
   if [[ $EMIT_SOURCEMAP == true ]]; then
-    HBC_SOURCEMAP_FILE="$HBC_FILE.map"
+    HBC_SOURCEMAP_FILE="$BUNDLE_FILE.map"
     "$NODE_BINARY" "$COMPOSE_SOURCEMAP_PATH" "$PACKAGER_SOURCEMAP_FILE" "$HBC_SOURCEMAP_FILE" -o "$SOURCEMAP_FILE"
   fi
+  BUNDLE_FILE="$DEST/main.jsbundle"
 fi
 
 if [[ $DEV != true && ! -f "$BUNDLE_FILE" ]]; then
```
grabbou, line 205: I don't think we need this condition BTW. It's invalid.

First of all, if the `main.jsbundle` is not generated in `$DEV = true` but with `$FORCE_BUNDLING = true`, this will not report any issues.

Also, at this point, we always want to check whether bundle exists. If we are not bundling (and there's nothing to check), we're going to exit early here: https://github.com/facebook/react-native/pull/29914/commits/cb01604e09403e0738311060a26375333dca5b39#diff-f1a28d5a04c65bf04339b2fd16a6f56fR35-R57

I believe this code is left-over from previous form of this file and the `if` statement should be removed

grabbou: I could most likely send a follow-up PR to address that, but I am not sure whether we want to do it for just a single line.

scripts/hermes-binary.sh
grabbou: Awesome!

grabbou:
> @grabbou Heeey Mike, here is a fresh built [hermes-engine-darwin-v0.7.0.tgz](https://9577-154201259-gh.circle-artifacts.com/0/tmp/hermes/output/hermes-engine-darwin-v0.7.0.tgz). Would you mind testing if it works correctly with this PR? If it does then we will publish to npm.

@Huxpro thank you for releasing this - confirmed that it worked just fine.

-

I have updated the main PR description with two remaining todo items to do before this one is ready to go.

scripts/react-native-xcode.sh
alloy: Such a maverick! 🤠 

👍

TheSavior:
Hermes 0.7.0 has been published https://www.npmjs.com/package/hermes-engine/v/0.7.0

grabbou:
Cherry-picking https://github.com/facebook/react-native/pull/29477 to fix building in `release` mode.

grabbou:
Two TODOs resolved today. Pending TODO work to happen tomorrow.

thymikee:
If #29477 is cherry-picked to fix stuff, maybe it's time to merge it to give attribution to the OP?

grabbou:
@thymikee yes, that PR will be merged separately. I am going to remove it from this PR when I am done testing.

grabbou:
Two PRs shipped to Hermes today: 
https://github.com/facebook/hermes/pull/363 and https://github.com/facebook/hermes/pull/364.

That should be everything needed to get this up and running.

mrousavy:
Are there any benchmarks/performance comparisons between JSC and Hermes available?

grabbou:
@mrousavy too early, I will do that as soon as this lands.

grabbou:
Xcode 12 is a wider issue and will need to be addressed on a framework level. For now, my workaround works here in RNTester but will have to be removed before merging to `master`.

https://github.com/facebook/react-native/issues/29984

grabbou:
Waiting for a new release of Hermes with https://github.com/facebook/hermes/pull/354, https://github.com/facebook/hermes/pull/363, https://github.com/facebook/hermes/pull/364 and https://github.com/facebook/hermes/pull/365 released.

grabbou:
While everything works with all the Hermes PRs I have submitted, we need to address https://github.com/facebook/hermes/issues/368 before moving forward.

Hope we can make a conclusion real quick and optimise the size of the Hermes artifact so that it doesn't affect installation times. Discussions will happen in that issue and PRs for now, as there's not much left as a part of this PR.

grabbou:
Note: The CI will be failing as long as the Hermes package is not available online for broader use. Right now, it loads my package locally.

grabbou:
I was able to resolve RCT-Folly issue and I think this may resolve some other issues that were reported in the past, mainly:
- https://github.com/facebook/react-native/issues/28810
- https://github.com/facebook/react-native/issues/27845

TL;DR RCT-Folly has to be build targeting iOS 9.0 release, otherwise, it will result in an error of duplicated type definitions and macros caused by `Time.h` file.

This fix has been inspired by:
- https://github.com/facebook/flipper/blob/master/FlipperKit.podspec#L20 (what Flipper uses as a deployment target)
- https://github.com/facebook/flipper/issues/834

grabbou:
This PR is ready to go. We're waiting for Hermes to be released to CocoaPods so that we can use it here (instead of my local, symlinked version).

grabbou:
Unfortunately, there turned out to be a problem with how we release Hermes. The CocoaPods package didn't include all platforms and architectures - in other words, I was unable to run this PR with such package in all testing scenarios.

I have spent some time reworking CircleCI pipeline of Hermes (especially when it comes to releasing a new package). The PR is already and it's here: https://github.com/facebook/hermes/pull/393.

I have tested it works just fine with a Hermes package released after these changes, so I am hoping this one is really the last PR we need 🙏

third-party-podspecs/RCT-Folly.podspec
```diff
@@ -123,6 +123,5 @@ Pod::Spec.new do |spec|
                           #  'folly/experimental/{ReadMostlySharedPtr,TLRefCount}.h',
   end
 
-  # Pinning to the same version as React.podspec.
```
alloy, line 125: This needs a comment saying why it is _not_ pinned to the same version.

grabbou:
Since https://github.com/facebook/hermes/pull/393 has been merged, this PR is ready for final review and shipping.

scripts/react-native-xcode.sh
```diff
@@ -107,15 +107,15 @@ fi
 # shellcheck source=/dev/null
 source "$REACT_NATIVE_DIR/scripts/node-binary.sh"
 
-[ -z "$HERMES_PATH" ] && export HERMES_PATH="$REACT_NATIVE_DIR/../hermes-engine-darwin"
+[ -z "$HERMES_PATH" ] && export HERMES_PATH="$REACT_NATIVE_DIR/../hermes-engine"
```
alloy, line 110: Now that we’re switching to the pod version of this, we should load `hermesc` from there instead of this npm package. In the future one might not have this package installed if one doesn’t do Android development.

alloy: So something like `$PODS_ROOT/hermes-engine/destroot/bin/hermesc` (untested)

grabbou: Yeah, this would be the best solution to go forward.

scripts/react-native-xcode.sh
```diff
@@ -107,16 +107,11 @@ fi
...
 
-export HERMES_CLI="$HERMES_PATH/osx-bin/hermesc"
-
-if [[ $USE_HERMES == true && ! -d "$HERMES_PATH" ]]; then
-  echo "error: Can't find Hermes executable - directory $HERMES_PATH doesn't exist. " \
-       "If you have a non-standard project structure, locate the 'Bundle React Native code and images' " \
-       "build phase in your Xcode project and set the `$HERMES_PATH` variable to the location of the " \
-       "hermes-engine package. For example: `$SRCROOT/../../node_modules/hermes-engine` " >&2
-  exit 2
+if [[ -z "$USE_HERMES" && -f "$HERMES_CLI_PATH" ]]; then
```
alloy, line 112: Should we still have some of the above path validation when `USE_HERMES` is set to `true`?

grabbou: We could, I wonder what the messaging should be.

Theoretically, this will only happen IF `USE_HERMES` was set manually to `true`. Otherwise, our mechanism will never turn it on when file does not exist.

Shall the wording mention running `pod install` or just go directly into "setting a path to Hermes executable"?

alloy: Hmm, good question. What about this?

> USE_HERMES is set to `true` but the `hermesc` binary could not be found at `${HERMES_CLI_PATH}`. Perhaps you need to run `pod install` or otherwise point the `HERMES_CLI_PATH` variable to your custom location.

(`${HERMES_CLI_PATH}` should be expanded to the actual path)

grabbou: Oh, amazing! Thank you! I don't feel super comfortable in wordings, this saves me a lot of time 🙏

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
```diff
@@ -586,7 +589,58 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\nexport PROJECT_ROOT=$SRCROOT/..\nexport SOURCEMAP_FILE=../sourcemap.ios.map\n# export FORCE_BUNDLING=true\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\nexport SOURCEMAP_FILE=../sourcemap.ios.map\nexport USE_HERMES=true\nexport PROJECT_ROOT=\"$SRCROOT/../../\"\nexport HERMES_PATH=\"$SRCROOT/../../node_modules/hermes-engine\"\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
```
TheSavior, line 592: Is this going to require people manually modify their Xcode project from Xcode? Is it time to pull this into a separate script file to make upgrading easier?

grabbou: Nope. Forgot to remove `USE_HERMES` after this commit https://github.com/facebook/react-native/pull/29914/commits/0f73afb8c7b700878fd8dbd0a227f4d27fbb6a9a. Good catch!

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
```diff
@@ -586,7 +589,58 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export NODE_BINARY=node\nexport PROJECT_ROOT=$SRCROOT/..\nexport SOURCEMAP_FILE=../sourcemap.ios.map\n# export FORCE_BUNDLING=true\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
+			shellScript = "export NODE_BINARY=node\nexport SOURCEMAP_FILE=../sourcemap.ios.map\nexport PROJECT_ROOT=\"$SRCROOT/../../\"\n\"$SRCROOT/../../scripts/react-native-xcode.sh\" $SRCROOT/../../packages/rn-tester/js/RNTesterApp.ios.js\n";
```
TheSavior, line 592: Looks like this line has still changed. Are these changes unintentional? Looks like it flipped the order of the definition of PROJECT_ROOT and SOURCEMAP_FILE, and no longer defines `FORCE_BUNDLING=true`. If these changes aren't necessary it would be great to remove them to make upgrading easier.

packages/rn-tester/RNTesterPods.xcodeproj/xcshareddata/xcschemes/RNTester.xcscheme
```diff
@@ -67,7 +67,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
```
TheSavior, line 70: Is this change intentional? Why is it flipping to release mode?

grabbou: Ah, this one is not intentional. Accidentially committed while testing `release` part of `react-native-xcode.sh` script.

packages/rn-tester/RNTesterPods.xcodeproj/project.pbxproj
grabbou: These changes are for the RNTester project only. The only change was to change `PROJECT_ROOT` to fix it (with a different path). Previously, building in `release` mode was broken. Will make the diff more simple.

React-Core.podspec
```diff
@@ -63,14 +63,14 @@ Pod::Spec.new do |s|
...
-    ss.platforms = { :osx => "10.14" }
+    ss.platforms = { :osx => "10.14", :ios => "10.0" }
     ss.source_files = "ReactCommon/hermes/executor/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/chrome/*.{cpp,h}",
                       "ReactCommon/hermes/inspector/detail/*.{cpp,h}"
     ss.pod_target_xcconfig = { "GCC_PREPROCESSOR_DEFINITIONS" => "HERMES_ENABLE_DEBUGGER=1" }
     ss.dependency "RCT-Folly/Futures"
-    ss.dependency "hermes", "~> 0.6.0"
+    ss.dependency "hermes-engine"
```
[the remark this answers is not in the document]
grabbou: Do we need it here tho? I thought that this will resolve to whatever is already present. And the version is hardcoded in `scripts/react_native_pods.rb`, where e.g. Flipper version is set too.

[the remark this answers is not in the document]
alloy: Gotcha. This is normally _conceptually_ the right place, but for now having it in `scripts/react_native_pods.rb` would suffice and ease work for maintainers, so I’m fine with that 👍

grabbou:
Note: Upgrading to CocoaPods 1.10.0 in this PR to resolve https://github.com/facebook/react-native/issues/29984

alloy:
@grabbou Looking at the CI output, it doesn't look like the `pod install` step is using the Gemfile that locks the CP version, it should be `bundle exec pod install`. So perhaps it’s still using an older version of CP?

hramos:
> @grabbou Looking at the CI output, it doesn't look like the `pod install` step is using the Gemfile that locks the CP version, it should be `bundle exec pod install`. So perhaps it’s still using an older version of CP?

Correct, we should be using `bundle exec` there. I have an internal diff (external PR copy here: https://github.com/facebook/react-native/pull/30250) which makes this change, but it's currently blocked on Xcode 12.1.0 becoming available on our internal CI. Feel free to make the `bundle exec` change as part of this PR if necessary.

grabbou:
@hramos I'll do that - thank you @alloy for spotting it.

grabbou:
FYI @hramos, I bumped Xcode to 12 and it fixed the build time errors I was getting on the CI. I will rebase once your PR is merged and remove my commits.

packages/rn-tester/RNTesterIntegrationTests/RCTLoggingTests.m
```diff
@@ -90,7 +90,7 @@ - (void)testLogging
 
   XCTAssertEqual(_lastLogLevel, RCTLogLevelError);
   XCTAssertEqual(_lastLogSource, RCTLogSourceJavaScript);
-  XCTAssertEqualObjects(_lastLogMessage, @"Invariant Violation: Invariant failed");
+  XCTAssertTrue([_lastLogMessage containsString:@"Invariant Violation: Invariant failed"]);
```
grabbou, line 93: `js engine: Hermes` is added to the end of the error message / invariant when Hermes is on. Replaced with `containsString` to make it more universal.

CC: @alloy

packages/rn-tester/RNTesterIntegrationTests/RCTLoggingTests.m
```diff
@@ -114,7 +114,7 @@ - (void)testLogging
 
   XCTAssertEqual(_lastLogLevel, RCTLogLevelError);
   XCTAssertEqual(_lastLogSource, RCTLogSourceJavaScript);
-  XCTAssertEqualObjects(_lastLogMessage, @"Error: Throwing an error");
+  XCTAssertTrue([_lastLogMessage containsString:@"Error: Throwing an error"]);
```
grabbou, line 117: On that note, shall we have two targets for integration tests and two targets for unit tests and run them on both Hermes and JSC? To be honest, it sounds to me like an overkill, but would like to discuss this idea.

grabbou:
The good news for anyone who is following this PR is - the build failures on CircleCI are finally gone! This week I spent on trying to understand why everything works and is green on my laptop and breaks on the CI. 

My plan for tomorrow is to make both `test_ios_unit` and `test_ios_unit_frameworks` green and #shipit.

IntegrationTests/GlobalEvalWithSourceUrlTest.js
```diff
@@ -42,7 +42,8 @@ class GlobalEvalWithSourceUrlTest extends React.Component<{...}> {
         'Expected globalEvalWithSourceUrl to throw on a syntax error',
       );
     }
-    if (!(syntaxError instanceof SyntaxError)) {
+    // Hermes throws an Error, not a SyntaxError
+    if (syntaxError.jsEngine !== 'hermes' && !(syntaxError instanceof SyntaxError)) {
```
grabbou, line 46: The exact error is:
> Error: Exception in HostFunction: Compiling JS failed: 1:2:'}' expected at end of block

Is this a bug or a known characteristics of Hermes?

packages/rn-tester/RNTesterIntegrationTests/RCTLoggingTests.m
alloy: I wonder if we can achieve that simply by passing something like `GCC_PREPROCESSOR_DEFINITIONS ="RCT_USE_HERMES=0"` to the `xcodebuild` command, rather than adding extra targets.

Either way, it might be better to do that in a follow-up PR?

hramos:
@grabbou no need to wait for my PR to land first, don't let that block you, I can rebase after your changes land. As for the failing Android tests, we landed some fixes to master recently.

alloy:
Current status is:

1. This branch builds _when building static libraries_. When building dynamic frameworks it fails building the `libevent` framework, for this issue I have a PR up here: https://github.com/callstack/react-native/pull/26
1. After fixing that, I stumbled on the issue of RN `master`’s copy of `JSI` being out of sync with `JSI` that comes with hermes v0.7.1. Specifically [this recent RN commit](https://github.com/facebook/react-native/commit/d8b0e9d9ab401fb3a0b2486ef64c97420cd45051#diff-b400543a253655e65e5d44926bc702f09daab5d33944a34debe236424eb40667).
1. To at least try to get this PR into a mergeable state, I wanted to then switch to consumer Hermes from `master`, but ran into the issue of Hermes `master` [failing to build for iOS](https://app.circleci.com/pipelines/github/facebook/hermes/1575/workflows/2867aa08-cebc-4a23-8ab2-4ea6ca7da786/jobs/11843). I’ve tracked the issue down to [this Hermes commit](https://github.com/facebook/hermes/commit/04d7a3e24c4d70fc0af2a0000a6237af997bf202) which makes the `hermesc` binary a dependency, which we don’t build for iOS.

  [Reverting that commit](https://github.com/facebook/hermes/compare/master...alloy:revert-chicken-and-egg-build-dependency-breakage-for-ios) does fix the iOS build again, but I’m unsure of the ramifications. The proper fix would be to first build the macOS artefacts and then re-use its `hermesc` artefact during the iOS build, but that might make this PR drag on again.

----

@Huxpro @TheSavior Do you have any thoughts on the best recourse/prioritization?

Possible options I see are the following:

### Fastest turnaround time

This _could_ technically be released in a stable RN release.

1. Revert the commit mentioned above in #2 in RN `master`.
1. Revert the commit mentioned above in #3 in hermes `master`.
1. Finish this PR by using the existing Hermes v0.7.1 release.

### Medium turnaround time

This can _not_ be released in a stable RN release.

1. Revert the commit mentioned above in #3 in hermes `master`.
1. Finish this PR by _temporarily_ pointing the hermes dependency to `hermes` master.
1. Potentially make the proper fix to first build macOS artefacts and use the resulting `hermesc` binary.
1. [Soon-ish] release Hermes v0.7.2 with a JSI version matching RN `master`’s.
1. Update RN’s hermes dependency to the prebuilt version of v0.7.2.

### Longest turnaround time

This can _definitely_ be released in a stable RN release.

1. Revert the commit mentioned above in #3 in hermes `master` or make the proper fix to first build macOS artefacts and use the resulting `hermesc` binary.
1. Release Hermes v0.7.2 with a JSI version matching RN `master`’s.
1. Finish this PR by pointing RN’s hermes dependency to the prebuilt version of v0.7.2.

IntegrationTests/GlobalEvalWithSourceUrlTest.js
Huxpro: @grabbou It's kinda related to a known problem that is not going to be resolved anytime soon. It's not so much of Hermes but the way we are building and linking our apps. I think we can leave the code this way and post an issue on the Hermes repo

Huxpro:
@alloy thanks for the detailed summary! It seems like the real road blocking issue was just the JSI getting out of sync by https://github.com/facebook/react-native/commit/d8b0e9d9ab401fb3a0b2486ef64c97420cd45051 and we only need to do 2 and 3 of the "longest turnaround time" option to get this PR rolling again.

We are trying to patch https://github.com/facebook/hermes/commit/73594b5f568567c7469c88316ca698ef6df21685 and its dependencies to Hermes release-v0.7 branch to hopefully have a `v0.7.2` that is JSI compatible with RN `master`'s. By this way we don't need to revert from either RN or Hermes `master`.

> After fixing that, I stumbled on the issue of RN master’s copy of JSI being out of sync with JSI that comes with hermes v0.7.1. 

It would be helpful to know what are the steps to repro so we can verify if the patch resolve the issue. I can still build RN Tester app on Android on RN master with the hermes-engine v0.7.1 npm package. Starting heap profiler does not crash the app but disabling it does. I'm not sure if that is the crash you encountered or another.


> To at least try to get this PR into a mergeable state, I wanted to then switch to consumer Hermes from master, but ran into the issue of Hermes master failing to build for iOS. I’ve tracked the issue down to this Hermes commit which makes the hermesc binary a dependency, which we don’t build for iOS.
>
> The proper fix would be to first build the macOS artefacts and then re-use its hermesc artefact during the iOS build, but that might make this PR drag on again.

Yeah, that would be the proper fix. At some point we'll need a hermes 0.8.0 with the fix for RN 0.65, but that could be months. For now, we should probably not fix it and consider using Hermes from master for RN 0.64 an option.

grabbou:
Here's the idea/question as this PR is getting older and older and starts to have a lot of changes inside it:

> This branch builds when building static libraries. When building dynamic frameworks it fails.

Shall we merge it and release a release candidate with a note that `use_hermes` on iOS only works with static libraries (`use_frameworks!` not enabled) and that will be fixed in the next release candidate?

CC: @TheSavior

TheSavior:
I'm not sure what the all implications of that would be, but in general it seems totally reasonable to land parts of this PR to make things easier on you instead of batching them altogether. I'll let you and @alloy decide what the right call is here.

IntegrationTests/GlobalEvalWithSourceUrlTest.js
grabbou: Okay, thanks, going to do it now!

grabbou: https://github.com/facebook/hermes/issues/400

packages/rn-tester/RNTesterIntegrationTests/RCTLoggingTests.m
grabbou: Yeah, we can play around with it later. Mainly wanted to decide on a conceptual level what is it that we want to test - it seems reasonable to have a separate CircleCI job to test both JSC and Hermes at the same time.

In such case, we would have 3 out of 4 tests passing - dynamic frameworks and Hermes combination would fail.

alloy:
> @alloy thanks for the detailed summary! It seems like the real road blocking issue was just the JSI getting out of sync by [d8b0e9d](https://github.com/facebook/react-native/commit/d8b0e9d9ab401fb3a0b2486ef64c97420cd45051) and we only need to do 2 and 3 of the "longest turnaround time" option to get this PR rolling again.

Agreed.

> We are working on a [patch](https://github.com/facebook/hermes/pull/399) that include [facebook/hermes@73594b5](https://github.com/facebook/hermes/commit/73594b5f568567c7469c88316ca698ef6df21685) and its dependencies onto the Hermes `release-v0.7` branch to hopefully have a `v0.7.2` that is JSI compatible with RN `master`'s. By this way we don't need to revert from either RN or Hermes `master`. I'll keep everyone posted once it's merged.

Perfect 🙏 

> > After fixing that, I stumbled on the issue of RN master’s copy of JSI being out of sync with JSI that comes with hermes v0.7.1.
> 
> It would be helpful to know what are the steps to repro so we can verify if the patch resolve the issue. I can still build RN Tester app on Android on RN master with the `hermes-engine-v0.7.1` npm package.

The failure is a build time failure. Try building this PR with dynamic frameworks enabled and then build:

```bash
cd packages/rn-tester
env USE_FRAMEWORKS=1 pod install
```

> > The proper fix would be to first build the macOS artefacts and then re-use its hermesc artefact during the iOS build, but that might make this PR drag on again.
> 
> Yeah, that would be the proper fix. At some point we'll need a hermes `v0.8.0` with the fix for RN 0.65, but that could be months. For now, we should probably not consider using Hermes from master for RN 0.64 a very feasible option.

Gotcha 👍

alloy:
> Here's additional option […] Merge it and release a release candidate with a note that `use_hermes` on iOS only works with static libraries (which is the default setup, `use_frameworks!` not enabled) and that support for dynamic frameworks will be fixed in the next release candidate.

I’m in favour. I’d perhaps even go farther and not promise it to be fixed in a RC, just at a later time.

grabbou:
> I’m in favour. I’d perhaps even go farther and not promise it to be fixed in a RC, just at a later time.

Interestingly, Flipper doesn't work with `use_frameworks!` either right now.

https://github.com/facebook/react-native/blob/master/template/ios/Podfile#L20-L23

Could that be related?

grabbou:
Opening #30300 restarted CircleCI. Yay!

grabbou:
All tests are passing except `test_ios_unit_frameworks_hermes`, which is expected. This is the variation that doesn't work right now, [as per previous discussion](https://github.com/facebook/react-native/pull/29914#issuecomment-720335840).

This PR is ready to land. We may want to disable the failing job so that the CI remains green. Hope this can be done while landing to avoid re-running all tests. If not, I will submit another commit.

packages/rn-tester/Podfile
```diff
@@ -8,12 +8,16 @@ if ENV['USE_FRAMEWORKS'] == '1'
...
+  puts "Using Hermes engine"
+end
+
 def pods()
   project 'RNTesterPods.xcodeproj'
 
   # Enable TurboModule
   prefix_path = "../.."
-  use_react_native!(path:prefix_path)
+  use_react_native!(path: prefix_path, hermes_enabled: ENV['USE_HERMES'])
```
alloy, line 20: I think this should be:

```suggestion
  use_react_native!(path: prefix_path, hermes_enabled: ENV['USE_HERMES'] == '1')
```

alloy (approved):
This looks good to me now 👍 

Thanks for doing this work, @grabbou 👏🙌 

Crazy Hermes

yungsters:
@facebook-github-bot import

denissb:
Just out of curiosity - has anyone done any benchmarks on hermes for iOS? 
E.g what is the impact on bundle size and startup performance.

PR #29937, opened 2020-09-11
Platform: Implement Platform.isCatalyst

charpeni:
## Summary

Following up on #27484, instead of adding `Platform.isMac` this pull request is proposing `Platform.isCatalyst` to avoid a collision with [`react-native-macos`](https://github.com/microsoft/react-native-macos).

⚠️ Note that `Platform.OS` is still returning `ios`, even on macOS with Catalyst.

In the meantime, you can rely on [`react-native-is-catalyst`](https://github.com/charpeni/react-native-is-catalyst) as an external dependency.

## Changelog

[General] [Added] - Implement Platform.isCatalyst to know if we're running on macOS with Catalyst.

## Test Plan

| macOS | iOS |
| - | - |
|  |  |

### DIFF
--- a/Libraries/Utilities/Platform.ios.js
+++ b/Libraries/Utilities/Platform.ios.js
@@ -39,6 +39,7 @@ const Platform = {
       prerelease: ?number,
     |},
     systemName: string,
+    isCatalyst: boolean,
   |} {
     if (this.__constants == null) {
       this.__constants = NativePlatformConstantsIOS.getConstants();
@@ -66,6 +67,10 @@ const Platform = {
       return this.constants.isTesting;
     }
     return false;
+  },
+  // $FlowFixMe[unsafe-getters-setters]
+  get isCatalyst(): boolean {
+    return this.constants.isCatalyst;
   },
   select: <D, N, I>(spec: PlatformSelectSpec<D, N, I>): D | N | I =>
     'ios' in spec ? spec.ios : 'native' in spec ? spec.native : spec.default,
--- a/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
+++ b/Libraries/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h
@@ -2001,6 +2001,7 @@ namespace JS {
           RCTRequired<NSString *> osVersion;
           RCTRequired<NSString *> systemName;
           RCTRequired<NSString *> interfaceIdiom;
+          RCTRequired<bool> isCatalyst;
         };
 
         /** Initialize with a set of values */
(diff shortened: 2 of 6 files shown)


TheSavior:
cc @appden @alloy

alloy (approved):
I feel like the function doesn’t add much and the check could simply be inlined, but I assume the compiler will optimise it that way so it’s mostly a stylistic choice 🤷‍♂️

Libraries/Utilities/Platform.ios.js
```diff
@@ -67,6 +68,10 @@ const Platform = {
     }
     return false;
   },
+  // $FlowFixMe[unsafe-getters-setters]
+  get isCatalyst(): boolean {
```
appden, line 72: I'd prefer this to called `isMacCatalyst()` as it's more clear and more inline with how Apple [markets](https://developer.apple.com/mac-catalyst/) the technology.

My general concern is that this will force all platforms to implement a `isMacCatalyst()` method, which is very specific, so I wonder if we should instead go with a method called `isDesktop()` to be more inclusive for our other platforms. That way `react-native-macos` and `react-native-windows` can return `true` for this method as well.

Thoughts @alloy?

alloy: I like that, I’m just a little concerned about the boundaries. Would this return `true` on iPad OS or what is the definition of a desktop machine?

appden: No, the purpose is it would return `true` for macOS (AppKit or UIKit) and Windows. 😄

Although lots of people iPad as a desktop, I think the general expectation is that it is not classified as a desktop. If that's a controversial opinion, then we can consider another name. We already have `isPad()` and `isTV()` methods that describe the *type* of device rather than particular OS, so I think `isDesktop()`  would fit nicely in there.

alloy: Yeah, perhaps desktops and iPads together are ‘workstations’? In any case, seems like we have enough ways to narrow down the type, so I’m good with ‘desktop’ 👍

EhsanSarshar: why not just include it inside Platform.OS

charpeni: @EhsanSarshar because Mac Catalyst isn't a platform itself, it's running the iOS app on macOS, so its behavior should mostly follow iOS except for some cases where we want it to be desktop.

Also, `Platform.OS` is handled by the packager ([Metro](https://github.com/facebook/metro)) during build time, not here in React Native. See this PR for the implementation of `Platform.select` with `mac`: https://github.com/facebook/metro/pull/495.

alloy: @charpeni Will you update this to be `isDesktop`?

charpeni: Will do!

marcosrdz:
Any updates on this? This is a great enhancement

Issue #29929, opened 2020-09-11
Strange Behavior on Modal (Only in the ios)


pedrohcdo:
I noticed an error in Modal, the modal simply disappears a few times, the error only occurs if the "transparent" attribute is set to true. In the emulator it is easier to see it happen. 
I am using the latest version of React Native: 0.63.2

## Steps To Reproduce
1. Create an simple modal like (It is important that it is always visible and transparent):
```
<Modal
        transparent={true}
        visible={true}
      >
            <View style={{ flex: 1 }}> 
                <Text>I am the modal content!</Text> 
            </View>
</Modal>
```

2. Install in the ios emulator, check if you started with the modal open, now save the file where the modal is to cause an update in the app. Keep saving the file to force the app to update, in my emulator the Modal is disappearing and reappearing, with each save I give it disappears and in the next save it reappears.

## Expected Results
The expected results are that it will never be "gone". When I change the "transparent" attribute to "false", the error goes away.

wilkuintheair:
@pedrohcdo The same thing happens to me, but slightly different scenario.
In my case it doesn't mind if the modal is transparent or not. It happens all the time when I just closed previous modal.
So it goes like this:
1. I have 2 modals in my render function, one of them is invisible.
2. When I close the first one it gets invisible, the second is now visible.
3. The expected behavior is that the second modal appears, but it's not.

When I'll have some more time I'll try to make some repro project.

pedrohcdo:
> @pedrohcdo The same thing happens to me, but slightly different scenario.
> In my case it doesn't mind if the modal is transparent or not. It happens all the time when I just closed previous modal.
> So it goes like this:
> 
> 1. I have 2 modals in my render function, one of them is invisible.
> 2. When I close the first one it gets invisible, the second is now visible.
> 3. The expected behavior is that the second modal appears, but it's not.
> 
> When I'll have some more time I'll try to make some repro project.

Try the @miotplugin answer, 
Change this file React/Views/RCTModalHostViewManager.m  with this commit https://github.com/facebook/react-native/pull/29953/commits/47fd3489db3086d560209bd78641b45d74194200

preflower:
> @pedrohcdo The same thing happens to me, but slightly different scenario.
> In my case it doesn't mind if the modal is transparent or not. It happens all the time when I just closed previous modal.
> So it goes like this:
> 
> 1. I have 2 modals in my render function, one of them is invisible.
> 2. When I close the first one it gets invisible, the second is now visible.
[...]

PR #29953, opened 2020-09-15
update RCTModalHostViewManager

miotplugin:
In order to safe present RCTModalHostViewController to fix bug [29929](https://github.com/facebook/react-native/issues/29929)

## Summary

Steps To Reproduce
Create an simple modal like (It is important that it is always visible and transparent, and the view outside modal should not delete):
```
<View style={{flex:1}}>
        <Modal visible={true} transparent={true}>
          <View
            style={{
              width: 300,
              height: 400,
              backgroundColor: 'yellow',
            }}
          />
        </Modal>
      </View>
```
Install in the ios emulator, check if you started with the modal open, now save the file where the modal is to cause an update in the app. Keep saving the file to force the app to update, in my emulator the Modal is disappearing and reappearing, with each save I give it disappears and in the next save it reappears.

When save, presentModalHostView:withViewController:animated: will be first called and then dismissModalHostView:withViewController:animated: In the case, Parent view will first refresh then update Modal then call presentModalHostView:
But if delete the outside view or set the transparent false, the method dismissModalHostView:withViewController:animated: will be first called and then presentModalHostView:withViewController:animated:  . the problem won't happen. In this case, removeFromSuperview: first called and then add Modal View.

## Changelog

[iOS][Fixed] - In order to safe present RCTModalHostViewController to fix bug [29929](https://github.com/facebook/react-native/issues/29929)

[CATEGORY] [TYPE] - Message


ee2e:
Hello :)
This request has not yet been reflected, but do you know why?

Brumor:
What happened with this? I still get the issue after updating to latest react-native

miotplugin:
@shergin Have any time to review the code?

Issue #16321, opened 2017-10-11
InteractionManager crash when cancelling promise task before it completes


jfaris:
### Is this a bug report?

Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

Yes

### Environment

react-native 0.47.2

### Steps to Reproduce

1. Schedule a long-running promise task with InteractionManager.runAfterInteractions

```
    const promise = InteractionManager.runAfterInteractions({
      gen: () => {
        // do a long fetch and return a promise
      }
    });
```

2. After the task has started but before it is done, cancel the task
```
    promise.cancel()
```

3. When the task completes TaskQueue will throw an exception
```
    TaskQueue: Error resolving Promise in task undefined: undefined is not an object (evaluating '_this2._queueStack[stackIdx].popable = true')
```

### Expected Behavior

No error would be thrown. InteractionManager (TaskQueue) would ignore the completion of a cancelled task and start the next task.

### Actual Behavior

Error was thrown from TaskQueue line 158:

iOS: `TaskQueue: Error resolving Promise in task undefined: undefined is not an object (evaluating '_this2._queueStack[stackIdx].popable = true')`

Android: `TaskQueue: Error resolving Promise in task undefined: Attempted to assign to readonly property.`

### Reproducible Demo

https://snack.expo.io/HJivJZ33-

jfaris:
@sahrens It looks like you've worked on this code the most. Any thoughts on this issue?

My use case is starting a task in componentDidMount and canceling it in componentWillUnmount. If the cancel comes after the task is started, it throws this error.

PR #29969, opened 2020-09-17
Don't crash when promise task is cancelled before its resolved

robwalkerco:
## Summary

This pull request fixes a potential `TypeError` in TaskQueue.js, that happens if a promise is added to the task queue, which is cancelled between the promise starting and resolving.

The exact error this resolves is 
```js
TypeError: TaskQueue: Error resolving Promise in task gen1: Cannot set property ‘popable’ of undefined
      167 |           queueStackSize: this._queueStack.length,
      168 |         });
    > 169 |       this._queueStack[stackIdx].popable = true;
          |                                              ^
      170 |       this.hasTasksToProcess() && this._onMoreTasks();
      171 |     })
      172 |     .catch(ex => {
      at Libraries/Interaction/TaskQueue.js:169:46
```

This specific error was also reported in #16321

## Changelog

[General] [Fixed] - Prevent TypeError in TaskQueue when cancelling a started but not resolved promise.

## Test Plan

The added test demonstrates the error, if run without the fixed applied to TaskQueue.js.
This is a race condition error, so is difficult to replicate!

### DIFF
--- a/Libraries/Interaction/TaskQueue.js
+++ b/Libraries/Interaction/TaskQueue.js
@@ -156,6 +156,7 @@ class TaskQueue {
     // happens once it is fully processed.
     this._queueStack.push({tasks: [], popable: false});
     const stackIdx = this._queueStack.length - 1;
+    const stackItem = this._queueStack[stackIdx];
     DEBUG && infoLog('TaskQueue: push new queue: ', {stackIdx});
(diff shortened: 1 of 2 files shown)


robwalkerco:
Anything I can do to help get this merged?

PR #29050, opened 2020-06-04
CI: use react-native-android:2020-6-4 docker image

dulmandakh:
## Summary

In https://github.com/react-native-community/docker-android/commit/b9ea45f3c56dc7ecca80883923bd5cc94c8579e7, we bumped Buck to 2020.05.14.01, NDK to 20.1.5948944.

## Changelog

[Internal] [Changed] - CI: use react-native-android:2020-6-4 docker image, with Buck 2020.05.14.01 and Android NDK 20.1.5948944

## Test Plan

CI must be green

PR #29987, opened 2020-09-19
WIP: Add an explicit NDK version to RNTester and ReactAndroid

gedeagas:
## Summary

When I try to run RNTester with Gradle the RNTester Required me to use **NDK 20.0.5594570**. I can't seem to find an explicit NDK version anywhere in ReactAndroid and RNTester. This PR Aims to add an explicit NDK version to RNTester and ReactAndroid. 

Screenshot from 2020-09-19 21-13-17

## Changelog
[Android] [Added] - Add an explicit NDK version to RNTester and ReactAndroid.

## Test Plan
Build manually from RNTester

### DIFF
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -382,7 +382,7 @@ task extractJNIFiles {
 
 android {
     compileSdkVersion 29
-    ndkVersion ndkVersion
+    ndkVersion "20.1.5948944"
     compileOptions {
         sourceCompatibility(JavaVersion.VERSION_1_8)
         targetCompatibility(JavaVersion.VERSION_1_8)
@@ -393,7 +393,6 @@ android {
         targetSdkVersion(28)
         versionCode(1)
         versionName("1.0")
-        ndkVersion "20.1.5948944"
 
         consumerProguardFiles("proguard-rules.pro")
 
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -247,12 +247,28 @@ def getNdkBuildName() {
 }
 
 def findNdkBuildFullPath() {
-    if (System.getenv("ANDROID_HOME") != null) {
-        def sdkDir = System.getenv("ANDROID_HOME")
-        def ndkDir = new File(sdkDir, "ndk").getAbsolutePath()
-        def ndkVersionedDir = new File(ndkDir, project.android.ndkVersion).getAbsolutePath()
-        return new File(ndkVersionedDir, getNdkBuildName()).getAbsolutePath()
+    // android.ndkDirectory should return project.android.ndkVersion ndkDirectory
+    def ndkDir = android.ndkDirectory ? android.ndkDirectory.absolutePath : null
+    if (ndkDir) {
+        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
     }
+
+     // we allow to provide full path to ndk-build tool
+    if (hasProperty("ndk.command")) {
+        return property("ndk.command")
+    }
+    // or just a path to the containing directory
+    if (hasProperty("ndk.path")) {
+        ndkDir = property("ndk.path")
+        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+    }
+
+    // @TODO ANDROID_NDK && ndk.dir is deprecated and will be removed in the future. 
+    if (System.getenv("ANDROID_NDK") != null) {
+        ndkDir = System.getenv("ANDROID_NDK")
+        return new File(ndkDir, getNdkBuildName()).getAbsolutePath()
+    }
+
     return null
 }
 
@@ -271,14 +287,15 @@ def getNdkBuildFullPath() {
     if (ndkBuildFullPath == null) {
         throw new GradleScriptException(
                 "ndk-build binary cannot be found, check if you've set " +
-                        "\$ANDROID_HOME environment variable correctly and you have installed the correct NDK version (" +
-                       project.android.ndkVersion + ")",
+                        "\$ANDROID_NDK environment variable correctly or if ndk.dir is " +
+                        "setup in local.properties",
                 null)
     }
     if (!new File(ndkBuildFullPath).canExecute()) {
         throw new GradleScriptException(
                 "ndk-build binary " + ndkBuildFullPath + " doesn't exist or isn't executable.\n" +
-                        "Check that you have installed the correct NDK version (" + project.android.ndkVersion + ")",
+                        "Check that the \$ANDROID_NDK environment variable, or ndk.dir in local.properties, is set correctly.\n" +
+                        "(On Windows, make sure you escape backslashes in local.properties or use forward slashes, e.g. C:\\\\ndk or C:/ndk rather than C:\\ndk)",
                 null)
     }
     return ndkBuildFullPath
--- a/ReactAndroid/build.gradle
+++ b/ReactAndroid/build.gradle
@@ -380,7 +380,7 @@ task extractJNIFiles {
 
 android {
     compileSdkVersion 29
-
+    ndkVersion ndkVersion
     compileOptions {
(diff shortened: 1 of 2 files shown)


gedeagas:
cc @dulmandakh @safaiyeh 

@fkgozali what is the NDK version does FB use internally on the RNTester buckconfig?  I used the same version specified on the Gradle template file PR that @safaiyeh worked on.

safaiyeh (approved):

fkgozali:
Actually I have a question about this explicit version specification: https://github.com/facebook/react-native/pull/29403#issuecomment-695362361

> what is the NDK version does FB use internally on the RNTester buckconfig?

FB uses a different internal version of the NDK, so it will conflict with the version you specify here, hence my question linked above

fkgozali:
Based on https://github.com/facebook/react-native/pull/29403#issuecomment-695821870, I wonder if we can make https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L249-L271 a little smarter by detecting the version you specify here. If that's possible, that would reduce confusion a lot.

Side note: is there a way to put all these "blessed versions" in one place so that we minimize repetition? E.g. perhaps we could put it in the root project config?

gedeagas:
@fkgozali Kindly check I just updated the `findNdkBuildFullPath()` to only use the explicitly specified NDK. But this means that we will be dropping support for users that install the Android NDK outside of the SDK Manager. 

For me, I think this is the best approach, considering that support for `ANDROID_NDK_HOME` is deprecated and will be removed in the future.

ReactAndroid/build.gradle
```diff
@@ -247,25 +247,11 @@ def getNdkBuildName() {
 }
 
 def findNdkBuildFullPath() {
```
fkgozali, line 249: I think we should still keep all this, otherwise it may be breaking to some folks, even though some of these are being deprecated. Instead, maybe just add your new logic for detecting `project.android.ndkVersion`.

Then add TODO here to explain what should be deprecated.

gedeagas: @fkgozali Just updated the pr, after debugging this yesterday I concluded that my new logic is not needed, we just need to make `android.ndkDirectory` the priority and `ANDROID_NDK` as the last priority.

But there is another problem that might be problematic to some folks. Because we specifically stated the ndkVersion here,
```
defaultConfig {
        applicationId "com.facebook.react.uiapp"
        applicationId "com.facebook.react.uiapp"
        minSdkVersion 19
        minSdkVersion 19
        ndkVersion "20.1.5948944"
}
```
Gradle will throw an exception saying that there is no version of NDK that matched the requested version. 

What do you think? if some folks run a different version than the one specified here they will get an error.

fkgozali: Hmm, this is exactly what I was concerned with. I think to minimize surprise/breaking change atm, we should just make ANDROID_NDK the priority for now. If someone has it set up, we can assume that the machine has dedicated Android dev env set up, so let's trust the developer by using ANDROID_NDK.

gedeagas: @fkgozali So the purpose of this pr is to fix the problem I have when trying to run RNTester with Gradle. The ReactAndroid Required me to use the older NDK 20.0.5594570 even when i already installed a specific NDK version on `ANDROID_NDK`. As far as I know, this is a bug from Gradle https://github.com/gradle/gradle/issues/12440#issuecomment-606188282. 

@friederbluemle suggest that there is three workaround 
1. Install the older NDK it is looking for (side-by-side): SDK manager 'ndk;20.0.5594570'
2. Manually specify the newer version in build.gradle. ( The one that I suggested on this pr ). 
3. Uninstall the NDK (if we don't need it): sdkmanager --uninstall 'ndk-bundle' 

This issue started happening when we upgraded to `com.android.tools.build:gradle:3.6.4` and seems this has not yet been fixed.

So I don't know what the best solution is, because even if I set the `ANDROID_NDK` correctly I still have to install `20.0.5594570` NDK because without this NDKReactAndroid won't build.

fkgozali: > This issue started happening when we upgraded to com.android.tools.build:gradle:3.6.4 and seems this has not yet been fixed.

Side question: will version 4+ address this issue? Should we aim to upgrade to 4 before landing this PR?

gedeagas: > Side question: will version 4+ address this issue? Should we aim to upgrade to 4 before landing this PR?

Haven't tried it yet, let me try.

gedeagas: With Gradle 4.0 instead of `20.0.5594570 ` it asking for `21.0.6113669`. The problem still exist but the default NDK is changed to `21.0.6113669`. 

Screenshot from 2020-09-24 01-10-21

ermanzohre (approved):

ReactAndroid/build.gradle
```diff
@@ -391,6 +393,7 @@ android {
         targetSdkVersion(28)
         versionCode(1)
         versionName("1.0")
+        ndkVersion "20.1.5948944"
```
friederbluemle, line 396: Are you sure this goes here (inside of `defaultConfig`)? Same question for the other file modified.

friederbluemle:
Thanks for the PR @gedeagas! I think in general we should keep this aligned with whatever the currently used version the Android Gradle plugin **defaults to**, _unless_ there is a specific reason to use a different version.

For AGP versions 3.5 and later, [this page](https://developer.android.com/studio/projects/configure-agp-ndk) states:
> (Recommended) Use the ndkVersion property to set the NDK version.

Since React Native is currently using `3.6.4`, we would set it to `ndkVersion` to `20.1.5948944`, and update it together with the Android Gradle plugin (#29013). Would that work?

image

Also, I don't _think_ it should be inside `defaultConfig`, only directly inside `android`. Or did you have any issues?

friederbluemle:
Also (in case I missed it somewhere in the long discussion) - The point of `ndkVersion` is to **not** automatically select whatever NDK is installed locally, but to ensure that a specific state of the repository is always built with the same exact tools.

ReactAndroid/build.gradle
gedeagas: Yes, the reason I put it in the defaultConfig is yesterday I was playing around Gradle and trying to answer Kevin suggestion on whether or not we can `put all these "blessed versions" in one place so that we minimize repetition` but I can't seem to find a way because ReactAndroid, Build Gradle is a different entity. Do you have any suggestions? I will put it directly inside `android` for now. 

Thanks for the quick review btw @friederbluemle. Much appreciated

gedeagas:
Yes, @friederbluemle, you are correct this pr is to ensure we all use the same version of NDK. And prevent people from using the different NDK version than we already specified on the React NativeGradle template file

We should use the same version of NDK on Gradle or Buck to ensure that a specific state of the repository is always built with the same exact tools.

fkgozali:
Hmm looks like this broke the test_android CI? Can you take a look?

https://app.circleci.com/pipelines/github/facebook/react-native/6463/workflows/0393edd6-4185-490b-a1e1-d7366a43fe83/jobs/169076

```

* Where:
Build file '/root/react-native/ReactAndroid/build.gradle' line: 251

* What went wrong:
Could not determine the dependencies of task ':ReactAndroid:packageReactNdkLibs'.
> Could not create task ':ReactAndroid:buildReactNdkLib'.
   > NDK is not installed
```

gedeagas:
@fkgozali we probably need to wait for this one https://github.com/facebook/react-native/pull/29050. The Android dockerfile is already being migrated to use Side by Side NDK. 

Compile Native Libs is successful
https://app.circleci.com/pipelines/github/facebook/react-native/6480/workflows/d71dd86d-5783-4d2c-b826-26b202a252af/jobs/169292

fkgozali:
@gedeagas, @ShikaSD will merge this, but we need to adjust the config slightly to support FB CI.

Issue #29849, opened 2020-09-03
ExceptionsManager.handleException is not a function


eliw00d:
We recently upgraded to React Native 0.63, Jest 26, and React Native Testing Library 7. When running jest with `--runInBand` we have noticed the following error randomly in our CI:

```
/node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js:104
  ExceptionsManager.handleException(error, true);
                    ^
TypeError: ExceptionsManager.handleException is not a function
    at reportLogBoxError (/node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js:108:21)
    at Immediate.<anonymous> (/node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js:217:7)
    at processImmediate (internal/timers.js:456:21)
```

It appears that NativeExceptionsManager is mocked but handleException is not: https://github.com/facebook/react-native/blob/master/jest/setup.js#L42-L47

```
❯ npx envinfo --npmPackages react,react-native,react-test-renderer,@testing-library/react-native

  npmPackages:
    @testing-library/react-native: 7.0.2 => 7.0.2 
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.2 => 0.63.2 
    react-test-renderer: 16.13.1 => 16.13.1 
```

```
❯ react-native info
info Fetching system and libraries information...
System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 746.34 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /Users/eliw00d/.rvm/gems/ruby-2.4.5/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.6, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6392135
    Xcode: 11.6/11E708 - /usr/bin/xcodebuild
  Languages:
[...]

PR #30027, opened 2020-09-24
Add mocks to fix: handleException is not a function (#29849)

lukewlms:
## Summary

This solves two problems:
1. When an exception is thrown by RN code during testing (e.g. with react-native-testing-library), the exception message is swallowed and instead this message is shown:

> ExceptionsManager.handleException is not a function

2. The inline `require` call that is removed in this commit is also incompatible with running a test suite (at least in react-native-testing-library) because the on-the-fly require fails.  This throws the error:

> ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

See issue: https://github.com/facebook/react-native/issues/29849

## Changelog

[General] [Fix] - Fix issue where exceptions thrown during test runs could get swallowed and result in these messages:
- ExceptionsManager.handleException is not a function
- ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

## Test Plan

Before: an exception generated during our test suite is swallowed and shows the errors above
After: no longer repros.  exceptions logged during a test run output their messages to the console as expected

(I def would like verification from a code owner here, this code makes it work but I can't guarantee it's optimal.  Thanks!)

### DIFF
--- a/Libraries/LogBox/Data/LogBoxData.js
+++ b/Libraries/LogBox/Data/LogBoxData.js
@@ -23,6 +23,9 @@ import type {
 import parseErrorStack from '../../Core/Devtools/parseErrorStack';
 import type {ExtendedError} from '../../Core/Devtools/parseErrorStack';
 import NativeLogBox from '../../NativeModules/specs/NativeLogBox';
+
+const ExceptionsManager = require('../../Core/ExceptionsManager');
+
 export type LogBoxLogs = Set<LogBoxLog>;
 export type LogData = $ReadOnly<{|
   level: LogLevel,
@@ -98,8 +101,6 @@ export function reportLogBoxError(
   error: ExtendedError,
   componentStack?: string,
 ): void {
-  const ExceptionsManager = require('../../Core/ExceptionsManager');
-
   error.forceRedbox = true;
   error.message = `${LOGBOX_ERROR_MESSAGE}\n\n${error.message}`;
   if (componentStack != null) {
--- a/jest/setup.js
+++ b/jest/setup.js
@@ -45,6 +45,13 @@ jest
       reportException: jest.fn(),
     },
   }))
+  .mock('../Libraries/Core/ExceptionsManager', () => {
+    return {
+      handleException: (...args) => {
+        return;
+      },
+    };
+  })
   .mock('../Libraries/ReactNative/UIManager', () => ({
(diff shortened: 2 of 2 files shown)


Panda-ref:
I'll result in
```
Require cycle: node_modules/react-native/Libraries/Core/ExceptionsManager.js ->
 node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js ->
 node_modules/react-native/Libraries/Core/ExceptionsManager.js
```

Libraries/LogBox/Data/LogBoxData.js
```diff
@@ -23,6 +23,9 @@ import type {
 import parseErrorStack from '../../Core/Devtools/parseErrorStack';
 import type {ExtendedError} from '../../Core/Devtools/parseErrorStack';
 import NativeLogBox from '../../NativeModules/specs/NativeLogBox';
+
+const ExceptionsManager = require('../../Core/ExceptionsManager');
```
eliw00d, line 27: This was originally in reportLogBoxError to avoid a circular dependency. If you move it back into reportLogBoxError, but simply add the mock, does it not work as expected?

lukewlms: Thanks for taking a look.  In my experience this change is necessary to avoid Jest throwing:

> ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

I'm not sure how to resolve the circular dependency issue :\

wcandillon:
I can confirm that this fixes the issue, how can we follow in which version of RN this fix will land?

owencraston:
It would be great to get this error blocks a lot of apps that have testing infrastructure set up to upgrade.

safaiyeh:
@rickyc could you take a look at this?

safaiyeh:
@rickhanlonii could you take a look at this?

rickhanlonii (changes requested):
Hey @lukewlms, thanks for working on this!

Looks like the tests are failing and @eliw00d is correct that we will need some other way to resolve the circular dependency.

semcelik:
Is there any update or workaround on this bug?

lukewlms:
@semcelik Yes, [this workaround](https://github.com/facebook/react-native/issues/29849#issuecomment-690712524) is correcting the issue for us.

PR #30044, opened 2020-09-27
chore: deduplicate lock, update packages repository fields

Simek:
## Summary

This small PR includes the following changes:
* deduplicate yarn lock file using [`yarn-deduplicate`](https://github.com/atlassian/yarn-deduplicate) package
  * deduplicate script has been added as `update-lock`, let me know if you would like also to see this in [`postinstall`](https://docs.npmjs.com/misc/scripts) (to automatically optimize lock on every dependency change)
* according to the [npm docs](https://docs.npmjs.com/files/package.json#repository):
  * main `package.json` repository field has been replaced with shorthand
  * monorepo packages repository field has been extended by `directory`

The main goal of introducing deduplication script was to optimize the dependencies footprint while developing and speed up the initial installation process. Running `yarn-deduplicate` also increase the security in some way, because it enforces usage only of the latest version of the package. You can read more about the benefits in the deduplicate script [repository](https://github.com/atlassian/yarn-deduplicate#duplicated-packages).

## Changelog

[Internal] [Added] - add yarn lock deduplication script

## Test Plan

`yarn install` was successful, this also should not affect yarn workspaces in any way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,7 @@
   "bin": "./cli.js",
   "description": "A framework for building native apps using React",
   "license": "MIT",
-  "repository": {
-    "type": "git",
-    "url": "git@github.com:facebook/react-native.git"
-  },
+  "repository": "github:facebook/react-native",
   "engines": {
     "node": ">=10"
   },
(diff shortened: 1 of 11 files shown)


cpojer (approved):
Nice, thank you!

Issue #30056, opened 2020-09-28
ActivityIndicator works only on iOS


rodrigofbm:
## Description
I'm using RN  0.63.2 and the ActivityIndicator appears on iOS but not on Android

## React Native version:
```
System:
    OS: Linux 5.8 Fedora 32 (KDE Plasma) 32 (KDE Plasma)
    CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
    Memory: 3.62 GB / 11.48 GB
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 12.18.4 - /usr/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.6 - /usr/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 23, 28, 29, 30
      Build Tools: 28.0.3, 29.0.2, 30.0.2
      System Images: android-27 | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
  Languages:
    Java: Not Found
    Python: 3.8.5 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.2 => 0.63.2 
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps To Reproduce

1. call `<ActivityIndicator />`

## Expected Results
Show the ActivityIndicator also on Android

jayshah123:
You have to explicitly give `color` on Android for an ActivityIndicator, default `color` no longer works. I am raising a PR for fixing this soon.

rodrigofbm:
> You have to explicitly give `color` on Android for an ActivityIndicator, default platform color no longer works. I am raising a PR for fixing this soon.

lol it works, thks!

fabriziobertoglio1987:
PR https://github.com/facebook/react-native/pull/29830 fixes this issue

Issue #29495, opened 2020-07-26
ActivityIndicator not visible without color prop on Pixel 3a


emersonhsieh:
## Description
The ActivityIndicator is not visible without setting the color prop on Pixel 3a. Setting the animating prop to true does not work either. Running on the Expo snack displays a green ActivityIndicator without the color prop, which was how it behaved in React Native 0.62.2 on the Pixel 3a. I believe this is a device specific issue, but I'm not sure what the scope is.

## React Native version: 0.63.2

Device: Pixel 3a
Android version: 10 (QQ3A.200605.002)

`react-native info`:

```
daSystem:
    OS: macOS 10.15.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 110.00 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 14.2.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.8.4 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 26, 28, 29
      Build Tools: 26.0.2, 28.0.3, 29.0.1, 29.0.2
      Android NDK: Not Found
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.6010548
    Xcode: 11.4/11E146 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_212 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.2 => 0.63.2 
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.

1. Create a fresh React Native app with `npx react-native init test_app`
2. Replace `App.js` with the code below
3. Plug in Pixel 3a phone, run `npm run android`

## Expected Results

Displaying two green ActivityIndicators (both on Dark and Light modes)

## Snack, code example, screenshot, or link to a repository:

[...]

Issue #29378, opened 2020-07-15
RN 0.63.1: ActivityIndicator disapearing


ptgamr:
## Description
With the default usage of `<ActivityIndicator />`, the indicators are all disappearing. But when I add the color prop, `<ActivityIndicator color="red" />`, I can then see the indicator again.

## React Native version:
System:
    OS: Linux 5.4 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (6) x64 Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
    Memory: 4.74 GB / 15.59 GB
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.14.1/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 23, 25, 26, 28
      Build Tools: 28.0.3, 29.0.2
      System Images: android-24 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: 1.8.0_252 - /usr/bin/javac
    Python: 2.7.18 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.1 => 0.63.1 
  npmGlobalPackages:
    *react-native*: Not Found

## Steps To Reproduce
Any usage of `<ActivityIndicator />` inside the source code is not working. Can reproduce with the starter project: 

`npx react-native init RN063`

## Expected Results
The indicator should be visible with the default color if the `color` props is not specified

## Snack, code example, screenshot, or link to a repository:
N/A

Ashoat:
Is this for iOS, Android, or both?

ptgamr:
@Ashoat sorry I forgot to  mentioned in the description. It's happening only on Android.

geniuscva:
having the same problem with RN 0.63.1

hamaron:
Confirmed the same behavior on RN 0.63.2

vamvoukas:
Confirmed the same behavior on RN 0.63.2

mattijsf:
[...]

PR #29830, opened 2020-09-02
[Fix] `Android/ColorProps`: ColorProps with value null should be defaultColor instead of transparent

hank121314:
## Summary

This pr:
- Fixes: #30183
- Fixes: #30056 
- Fixes: #29950
- Fixes: #29717 
- Fixes: #29495
- Fixes: #29412
- Fixes: #29378

Because most of @ReactProps(name = ViewProps.COLOR) accept @ Nullable Integer.
For example:
https://github.com/facebook/react-native/blob/abb6433f506851430dffb66f0dd34c1e70a223fe/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactBaseTextShadowNode.java#L472-L479

After update to react-native 0.63.2 to make PlatformColor work, there is a new ColorPropSetter.
https://github.com/facebook/react-native/blob/abb6433f506851430dffb66f0dd34c1e70a223fe/ReactAndroid/src/main/java/com/facebook/react/uimanager/ViewManagersPropertyCache.java#L194-L215

But ColorPropSetter won't return an nullable value with getValueOrDefault, it will always return it's defaultValue which is 0.
And 0 is equal to TRANSPARENT, will cause <Text /> disappear.
## Changelog

[Android] [Fixed] - ColorProps with value null should be defaultColor instead of transparent

## Test Plan
Please initiated a new project and replaced the app with the following code:
```
import * as React from 'react';
import {Text, View, TouchableOpacity, PlatformColor} from 'react-native';

export default function App() {
  const [active, setActive] = React.useState(false);

  return (
    <View>
      <Text style={active ? {color: 'green'} : null}>Example</Text>
      <Text
        style={
          active ? {color: PlatformColor('@android:color/holo_purple')} : null
        }>
        Example2
      </Text>
      <TouchableOpacity onPress={() => setActive(!active)}>
        <Text>Toggle Active</Text>
      </TouchableOpacity>
    </View>
  );
}
```

Thanks you so much for your code review!

PR #30057, opened 2020-09-28
Default color for ActivityIndicator on Android

jayshah123:
## Summary

ActivityIndicator is blank after RN 0.63, and does not show up unless explicit `color` prop is provided.

Fixes https://github.com/facebook/react-native/issues/30056, 
https://github.com/facebook/react-native/issues/29495, https://github.com/facebook/react-native/issues/29378

Fix By using `PlatformColor` API to set default platform color which is according to https://developer.android.com/reference/android/R.attr#colorControlActivated

Before fix:
Screenshot_20200928-220924_RNTester App

After fix:
Screenshot_20200928-221153_RNTester App

## Changelog

[Android] [Fixed] - Show default platform color for Android ActivityIndicator when no color supplied in prop.

## Test Plan

Go to AppTester App > ActivityIndicator > Android Activity indicator should be seen with default platform color.

### DIFF
--- a/Libraries/Components/ActivityIndicator/ActivityIndicator.js
+++ b/Libraries/Components/ActivityIndicator/ActivityIndicator.js
@@ -14,7 +14,7 @@
 const Platform = require('../../Utilities/Platform');
 const React = require('react');
 const StyleSheet = require('../../StyleSheet/StyleSheet');
-const { PlatformColor } = require('../../StyleSheet/PlatformColorValueTypes');
+const {PlatformColor} = require('../../StyleSheet/PlatformColorValueTypes');
 const View = require('../View/View');
 import type {HostComponent} from '../../Renderer/shims/ReactNativeTypes';
 import type {ViewProps} from '../View/ViewPropTypes';
@@ -184,7 +184,8 @@ ActivityIndicatorWithRef.displayName = 'ActivityIndicator';
  * and run Flow. */
 ActivityIndicatorWithRef.defaultProps = {
   animating: true,
-  color: Platform.OS === 'ios' ? GRAY : PlatformColor('?attr/colorControlActivated'),
+  color:
+    Platform.OS === 'ios' ? GRAY : PlatformColor('?attr/colorControlActivated'),
   hidesWhenStopped: true,
   size: 'small',
 };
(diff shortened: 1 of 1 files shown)


Libraries/Components/ActivityIndicator/ActivityIndicator.js
```diff
@@ -183,7 +184,8 @@ ActivityIndicatorWithRef.displayName = 'ActivityIndicator';
  * and run Flow. */
 ActivityIndicatorWithRef.defaultProps = {
   animating: true,
-  color: Platform.OS === 'ios' ? GRAY : null,
+  color:
+    Platform.OS === 'ios' ? GRAY : PlatformColor('?attr/colorControlActivated'),
```
Simek, line 188: Hi @jayshah123, can you add a check and ensure that `colorControlActivated` platform color is only used on Android? 

It is the usual practice in RN repository (ex. [TextInput](https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L45)), because there are few out-of-tree platforms and the current code will cause an issue for them.

jayshah123: Yeah sure. Will update this PR.

jayshah123: Updated.

jayshah123:
Is there a way to request a review from someone in the RN team?

Libraries/Components/ActivityIndicator/ActivityIndicator.js
Simek: @jayshah123 I think that `default` should be a `null` to retain behavior before the Android platform fix addition. Currently this could break the out-of-tree default indicator color if they do not override the default value (if they leverage the similar behavior on theirs platform like Android on RN < 0.62, when passing `null` results in fallback to platform default color).

jayshah123: Makes sense, will default to null to retain behavior.
Was thinking in the direction that some visual indicator would be better than null.
Will update the PR with your suggestions.

hank121314:
Hi @Simek .
I think this issue is related to #29717 and should be fixed by #29830 .
All of the situation show that if color props is null. Android text or activityIndicator will disappear(transparent).
I think pr #29830 will be a universal fix.

jayshah123:
> #29717

Will try to check your fork.

taylorkline:
Thanks, I am integrating this patch with `yarn patch-package` and it works for me on 0.64.2.

facuhump2409 (approved):

PR #30059, opened 2020-09-28
Fix: Detection of host ip address for running on a device

seavan:
## Summary

If your host machine has both en0 and en1 disabled/point to local network, `scripts/react-native-xcode.sh` does not try to determine a more suitable IP.

We're making this change to be able to let the app connect to bundler even when en0 and en1 are disabled/point to local network. 

## Changelog

[iOS] [Fixed] - Fixed incorrect IP detection for host machines having both en0 and en1 point to local addresses.

## Test Plan

Disable your wifi and en0 and connect network to en2 if you have it. Run the app from XCode on a device. Confirm that the app can detect IP address correctly by verifying ip.txt.


yungsters:

Issue #28233, opened 2020-03-05
Crash in RCTView


Elfoniok:
Description:

This is one of the crashes that I experience when I am using react-native-debugger AND when I enable Inspector to highlight the components. Please see  reproduction steps.

NOTE: I am using expo on top.

```
ExceptionsManager.js:44 Invariant Violation: [339,"RCTView",{"left":"<<NaN>>","top":"<<NaN>>","width":"<<NaN>>","height":"<<NaN>>"}] is not usable as a native method argument

This error is located at:
    in RCTView (at ElementBox.js:64)
    in ElementBox (at InspectorOverlay.js:59)
    in RCTView (at InspectorOverlay.js:67)
    in InspectorOverlay (at Inspector.js:250)
    in RCTView (at Inspector.js:248)
    in Inspector (at AppContainer.js:65)
    in RCTView (at AppContainer.js:119)
    in AppContainer (at renderApplication.js:39)

```

React Native version:
System:
    OS: Linux 4.15 Ubuntu 18.04.3 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
    Memory: 263.64 MB / 15.55 GB
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 8.10.0 - /usr/bin/node
    Yarn: 1.21.1 - /usr/bin/yarn
    npm: 3.5.2 - /usr/bin/npm
  npmPackages:
    react: ~16.9.0 => 16.9.0 
    react-native: https://github.com/expo/react-native/archive/sdk-36.0.0.tar.gz => 0.61.4 
  npmGlobalPackages:
    create-react-native-app: 2.0.2

## Steps To Reproduce

1. Use code attached as an App.js.
2. Start react-native-debugger, have it listen on 19001 or whatever your metro bundler expose
3. Start app with expo start, then choose android emulator (but you can use device the same thing happens).
4. After the application loaded, open developer menu and use Debug option. Hopefully your react-native-debugger will overtake, you should see the application structure in it.
5. Once again from developer menu open inspector, it should be collapsed and you should be able to highlight the content on your target while walking through the structure. But as soon as you highlight RCTView, you will get error attached above. If I remove style from the code, inspector is working without problems.

## Expected Results

Should not crash :)

## Snack, code example, screenshot, or link to a repository:

[...]

PR #30064, opened 2020-09-29
Issue #28233 :: Ensure frame dimensions are numbers

dsauvespiria:
## Summary

I experience the following issue : https://github.com/facebook/react-native/issues/28233

I understand that we should understand why these values are passed there, but it seems that it is other components that are passing wrong values, so in order to have a bullet proof component, one should ensure that even if wrong values are passed to a component, the component should not crash.

## Changelog

[Android] [fixed] - I only added verification to ensure that the frame values are numbers

## Test Plan

When Inspecting elements, the debugger crashed but with this fix the debugger stop crashing

### DIFF
--- a/Libraries/Inspector/ElementBox.js
+++ b/Libraries/Inspector/ElementBox.js
@@ -32,6 +32,12 @@ class ElementBox extends React.Component<$FlowFixMeProps> {
     };
 
     let frameStyle = {...this.props.frame};
+
+    let contentStyle = {
+      width: this.props.frame.width,
+      height: this.props.frame.height,
+    };
+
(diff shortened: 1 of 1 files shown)


Gabrielfcs:
Excellent fix @dsauvespiria, but can't this happen inside `contentStyle` too?

dsauvespiria:
Hey @Gabrielfcs, indeed! I was to focus on fixing my issue... Just push another commit to fix that part... I believe that I would need to rebase my code, also there is a useless merge commit...

dsauvespiria:
So after rebasing the issue, for contentStyle the type was enforce since my PR...

Issue #28278, opened 2020-03-10
[iOS] Crash in RCTWebSocketModule didFailWithError


Etienne-M:
Description:

We're seeing some crashes in our Crashlytics dashboard with the following error:
```
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]
```
coming from
```
RCTWebSocketModule.m
-[RCTWebSocketModule webSocket:didFailWithError:]
```

React Native version: 0.61.5

## Steps To Reproduce

Unknown, although having a WebSocket would be a good start to reproduce.

## Expected Results

The app should not crash.

xiao99xiao:
This issue exists for a long time. The reason is the `socketID` and `error` in RCTWebSocketModule.m could sometimes be `nil` .
I patched the file like below and it never bothered me again.
```
-    @"message": error.localizedDescription,
-    @"id": socketID
+    @"message": error.localizedDescription ?: @"",
+    @"id": socketID ?: @(-1)
```

chrisglein:
Can you produce a minimal repo so we can narrow this down? (e.g. a Snack or pointer to a repo of a minimal repo)
You provided the version (thank you!), but can you run 'react-native info' to get the rest of the context?

Etienne-M:
> Can you produce a minimal repo so we can narrow this down? (e.g. a Snack or pointer to a repo of a minimal repo)
> You provided the version (thank you!), but can you run 'react-native info' to get the rest of the context?

Hello! Sorry, it took me a while to get you this info. There you:
```
System:
OS: macOS Mojave 10.14.6
CPU: (3) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
Memory: 10.96 GB / 16.00 GB
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
npm: 6.14.3 - ~/.nvm/versions/node/v12.13.0/bin/npm
SDKs:
iOS SDK:
Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
IDEs:
Xcode: 11.3.1/11C505 - /usr/bin/xcodebuild
npmPackages:
@react-native-community/cli: 3.0.4 => 3.0.4
react: 16.9.0 => 16.9.0
[...]

Issue #29525, opened 2020-07-29
[iOS] Crash in RCTWebSocketModule didFailWithError 


bertrandgressier:
I reopen this issue https://github.com/facebook/react-native/issues/28278

## Description
We're seeing some crashes in our Crashlytics dashboard with the following error:
```
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]
```
coming from

```
RCTWebSocketModule.m
-[RCTWebSocketModule webSocket:didFailWithError:]
```
## React Native version:
0.62.2

## Steps To Reproduce
WebSocket errors

## Expected Results
The app should not crash.

this person https://github.com/facebook/react-native/issues/28278#issuecomment-597461650 suggested a fix for this issue

```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x1a1571344 (Manquant)
1  libobjc.A.dylib                0x1a1286cc0 (Manquant)
2  CoreFoundation                 0x1a15ca550 (Manquant)
3  CoreFoundation                 0x1a15d44c8 (Manquant)
4  CoreFoundation                 0x1a1458944 (Manquant)
5  CoreFoundation                 0x1a144a1c4 (Manquant)
6  gotoconnect                    0x10010b874 -[RCTWebSocketModule webSocket:didFailWithError:] + 175 (RCTWebSocketModule.mm:175)
7  gotoconnect                    0x100638a3c __33-[RCTSRWebSocket _failWithError:]_block_invoke_2 + 619 (RCTSRWebSocket.m:619)
8  libdispatch.dylib              0x1a1212134 (Manquant)
9  libdispatch.dylib              0x1a12135ac (Manquant)
10 libdispatch.dylib              0x1a121f7d4 (Manquant)
11 CoreFoundation                 0x1a14ec8d4 (Manquant)
12 CoreFoundation                 0x1a14e758c (Manquant)
13 CoreFoundation                 0x1a14e6bc8 (Manquant)
14 GraphicsServices               0x1ab8c85cc (Manquant)
15 UIKitCore                      0x1a5699744 (Manquant)
16 gotoconnect                    0x1000d6d04 main + 28 (AppDelegate.swift:28)
17 libdyld.dylib                  0x1a1363384 (Manquant)
```

chrisglein:
[...]

PR #30066, opened 2020-09-29
fix: crash when insert nil value into a dictionary

marksinkovics:
## Summary

This PR attempts to fix issue https://github.com/facebook/react-native/issues/28278 and https://github.com/facebook/react-native/issues/29525

On Crashlytics, the following error occurs in file `RCTWebSocketModule.m` at method `-[RCTWebSocketModule webSocket:didFailWithError:]` when a nil value is inserted into a dictionary as a value.

```
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]
```

This PR is following the suggestion of this: https://github.com/facebook/react-native/issues/28278#issuecomment-597461650 and it replaces the values of any property if it is nil. In detail:
- it converts `error` to empty NSString if the original value is nil
- it converts `socketID` to a NSNumber object, which stores `-1` if the original value is nil

## Changelog

[iOS] [Fixed] - A crash in WebSocket module

## Test Plan

We were not able to reproduce the crash, but the report itself provided enough information to find a solution for this issue.


mastery-dhananjay-jadhav (approved):

React/CoreModules/RCTWebSocketModule.mm
```diff
@@ -168,7 +168,11 @@ - (void)webSocket:(RCTSRWebSocket *)webSocket didFailWithError:(NSError *)error
   NSNumber *socketID = [webSocket reactTag];
   _contentHandlers[socketID] = nil;
   _sockets[socketID] = nil;
-  [self sendEventWithName:@"websocketFailed" body:@{@"message" : error.localizedDescription, @"id" : socketID}];
+  NSDictionary *body = @{
+        @"message" : error.localizedDescription ?: @"",
```
sammy-SC, line 172: Should this say something like "undefined" instead of empty string?

sammy-SC:
thank you @marksinkovics

React/CoreModules/RCTWebSocketModule.mm
marksinkovics: Thank you for your comment. I updated the error message to `Undefined, error is nil`. 
Please let me know if you have any further feedback.

Issue #30055, opened 2020-09-28
HEAD requests fail on Android with 'TypeError: Network request failed'


lachenmayer:
Hi team,

## Description

On Android, all HTTP HEAD requests for which the server returns a `Content-Length` header fail with a "TypeError: Network request failed". This is reproducible in a brand-new app created using `create-react-native-app`, as well as on [Expo Snack](https://snack.expo.io/@picnic-devs/584fdb).

## React Native version:

<details>
<summary>
<strong>This affects React Native 0.63.2.</strong> Click here to expand full <code>react-native info</code> output.
</summary>

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-4308U CPU @ 2.80GHz
    Memory: 372.92 MB / 16.00 GB
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 12.16.3 - /var/folders/z8/xq1r3qcd7szbpx1_77xsgpqc0000gn/T/yarn--1601305570730-0.3968610362929228/node
    Yarn: 1.22.0 - /var/folders/z8/xq1r3qcd7szbpx1_77xsgpqc0000gn/T/yarn--1601305570730-0.3968610362929228/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.8.4 - /usr/local/lib/ruby/gems/2.5.0/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 25, 26, 28, 29
      Build Tools: 28.0.3, 29.0.2
      Android NDK: Not Found
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6200805
    Xcode: 11.3.1/11C505 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.6 - /usr/local/opt/openjdk@11/bin/javac
    Python: 2.7.16 - /usr/local/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: ~0.63.2 => 0.63.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

</details>

## Steps To Reproduce

[...]

PR #30072, opened 2020-09-30
[Fix] `Android/Fetch`: Handle IOException when using fetch with HEAD method

hank121314:
## Summary

This pr fixes: #30055 .

Since fetch with HEAD method did not have body. 
We should handle the error from Okhttp when it consuming the empty body.

## Changelog

[Android] [Fixed] - HEAD requests with fetch should not be failed.

## Test Plan

Please initiated a new project and replaced the app with the following code:

```
import React, { useEffect } from 'react';

export default function App() {
  useEffect(() => {
    fetch('https://determined-panini-990054.netlify.app/demo-image.png', {
      method: 'HEAD',
    })
      .then(console.warn)
      .catch(console.error);
  }, []);
  return (
      <Text>Open up App.js to start working on your app!</Text>
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
It should console warn the result instead of `Network request failed` .

Thanks you so much for your code review!

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobModule.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobModule.java
@@ -38,6 +38,7 @@ import java.util.Map;
 import java.util.UUID;
 import okhttp3.MediaType;
 import okhttp3.RequestBody;
+import okhttp3.Response;
 import okhttp3.ResponseBody;
 import okio.ByteString;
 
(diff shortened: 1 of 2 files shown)


ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobModule.java
```diff
@@ -130,8 +131,21 @@ public boolean supports(String responseType) {
         }
 
         @Override
-        public WritableMap toResponseData(ResponseBody body) throws IOException {
-          byte[] data = body.bytes();
+        public WritableMap toResponseData(Response response) throws IOException {
+          byte[] data = {};
+          try {
+            data = response.body().bytes();
```
LawrenceMouarkach, line 137: What if we want to process a 204 no content? this would still fail when u call bytes on a null body

ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobModule.java
```diff
@@ -130,8 +131,21 @@ public boolean supports(String responseType) {
...
+          } catch (IOException e) {
+            if (response.request().method().equalsIgnoreCase("HEAD")) {
```
LawrenceMouarkach, line 139:
```suggestion
            if (response.request().method().equalsIgnoreCase("HEAD") || response.code() === 204) {
```

ReactAndroid/src/main/java/com/facebook/react/modules/blob/BlobModule.java
hank121314: Changed! Thank you for your suggestions 😄 !

PR #30084, opened 2020-10-01
Chore: Upgrade Detox to 16.7.2

kelset:
## Summary

With Xcode 12 being the latest, Detox 15.x has issues - in particular, it means that if you try to `yarn install` the dependencies for the repo, you'll be greeted by this error:

```bash
error /<stuff>/react-native/node_modules/detox: Command failed.
Exit code: 1
Command: node scripts/postinstall.js
Arguments:
Directory: /<stuff>/react-native/node_modules/detox
Output:
/<stuff>/Library/Detox/ios/5824c837515589f21c08f09b716a6eda088aa31f was found, but could not find Detox.framework inside it. This means that the Detox framework build process was interrupted.
         deleting /<stuff>/Library/Detox/ios/5824c837515589f21c08f09b716a6eda088aa31f and trying to rebuild.
Extracting Detox sources...
Building Detox.framework from /<stuff>/Developer/OSS/react-native/node_modules/detox/ios_src...
child_process.js:637
    throw err;
    ^
```

With the 👍 of @hramos & @alloy I've prep'd up a small defensive PR that can be quickly merged before cutting 0.64, that bumps the version of Detox from 15.4.4 to the highest version available within the reach of "no breaking changes" in changelog.

The main reason why with 16.x this error doesn't happen is that from [16.0.0](https://github.com/wix/Detox/releases/tag/16.0.0):

> Detox now comes as a prebuilt framework on iOS, thus lowering npm install times and saving some build issues that happen due to unexpected Xcode setups.

It would have been better to update directly to latest (at the time of writing 17.7.1) but there are at least two versions that had changelogs that seem to involve bigger changes:

* https://github.com/wix/Detox/releases/tag/17.4.7
* https://github.com/wix/Detox/releases/tag/16.8.0

Hopefully CI will will show that the bump doesn't break any test 🤞

## Changelog

[Internal] [Changed] - Bumped Detox in the repo to 16.7.2 for Xcode 12 compatibility

## Test Plan

Running yarn in the main repo with Node 14 & Xcode 12, without this change, will cause the error copy-pasted above. After upgrading to this version, the error disappear.

### DIFF
--- a/packages/rn-tester/package.json
+++ b/packages/rn-tester/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "connect": "^3.6.5",
-    "detox": "15.4.4",
+    "detox": "16.7.2",
     "ws": "^6.1.4"
   }
 }
--- a/repo-config/package.json
+++ b/repo-config/package.json
@@ -18,7 +18,7 @@
     "clang-format": "^1.2.4",
     "connect": "^3.6.5",
     "coveralls": "^3.0.2",
-    "detox": "15.4.4",
+    "detox": "16.7.2",
     "eslint": "6.8.0",
     "eslint-config-fb-strict": "^24.9.0",
     "eslint-config-fbjs": "2.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1729,10 +1729,10 @@ big-integer@^1.6.7:
   resolved "https://registry.yarnpkg.com/big-integer/-/big-integer-1.6.36.tgz#78631076265d4ae3555c04f85e7d9d2f3a071a36"
   integrity sha512-t70bfa7HYEA1D9idDbmuv7YbsbVkQ+Hp+8KFSul4aE5e/i1bjCNIRYJZlA8Q8p0r9T8cF/RVvwUgRA//FydEyg==
 
-bluebird@3.5.x:
-  version "3.5.2"
-  resolved "https://registry.yarnpkg.com/bluebird/-/bluebird-3.5.2.tgz#1be0908e054a751754549c270489c1505d4ab15a"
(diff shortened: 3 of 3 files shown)


kelset:
Looks like some CI jobs are failing, but not at the tests step.

The task failing is "Report size of RNTester.app" (which is a task *after* the actual testing so per se the detox bump is fine), with the error:

> Missing GITHUB_PR_NUMBER. Example: 4687. PR feedback cannot be provided on GitHub without a valid pull request number.

Not sure how to fix it 😅

alloy:
It would be great to get some input on what versions work best from people that actually use Detox (and RN), which isn’t me; do you?

kelset:
> It would be great to get some input on what versions work best from people that actually use Detox (and RN), which isn’t me; do you?

I've used Detox extensively in the first half of the year, and tbh the mindset is always to stay as close as possible to latest release because of the amount of fixes they put into each version.

That said, I think it's slightly out of scope for the purpose of this PR which is literally to just get rid of the problem I detailed above.

We could probably open a "good first issue" for folks to go over the detox testing and improve it, in the spirit of Hacktoberfest.

janicduplessis:
Just hit this issue trying to yarn install the repo. This shouldn't affect people actually using detox in their app since it is only the version used for our tests.

janicduplessis:
cc @hramos

janicduplessis (approved):

tido64:
> Looks like some CI jobs are failing, but not at the tests step.
> 
> The task failing is "Report size of RNTester.app" (which is a task _after_ the actual testing so per se the detox bump is fine), with the error:
> 
> > Missing GITHUB_PR_NUMBER. Example: 4687. PR feedback cannot be provided on GitHub without a valid pull request number.
> 
> Not sure how to fix it 😅

Can you try re-running the builds? The other PRs seems to be passing this step. Tom and I hit this issue today. Would be great to get this in.

kelset:
@tido64 rerunning 🤞

PR #30104, opened 2020-10-03
Bump okhttp version to support Android 11

safaiyeh:
## Summary
Android 11 support landed in OkHttp with version 3.14.7, bumping to the most necessary version.

Issue: https://github.com/facebook/react-native/pull/30104

## Changelog

[Android] [Changed] - Bump OkHttp version to 3.14.7 to support Android 11


dulmandakh:
OkHTTP 3.13 requires Android 5 and above, but RN still supports 4.4. Please see https://cashapp.github.io/2019-02-05/okhttp-3-13-requires-android-5

Also it looks like the fix is back ported to 3.12 with https://github.com/square/okhttp/pull/5822.

dulmandakh:
I'm quoting 3.12.9 changelog from https://square.github.io/okhttp/changelog_3x/#version-3129

>  Fix: Don’t crash on Android 11 due to use of restricted methods. This prevents a crash with the exception, “Expected Android API level 21+ but was 29”.

franznkemaka:
Hi @dulmandakh, do you know what has to be done to fix the issue?

safaiyeh:
Interesting, if it is backported crashes shouldn't be occurring. The original issue mentioned API 30 maybe okhttp hasn't addressed that yet?

dulmandakh:
Okhttp was bumped just recently, so it only in the master and 0.63 contains buggy version. I requested to cherry pick it for 0.63.4.

safaiyeh:
Makes sense! Let's close this and link the PR to the original issue and close that one as well.

lorenc-tomasz (approved):

kelset:
@safaiyeh so this can be closed now?

dulmandakh:
let's close the PR 😉

Issue #27204, opened 2019-11-12
Crash on email detection in TextInput on Xiaomi devices running android 10


gbalduzzi:
Xiaomi devices running android 10 show a little popup with the Text `Frequent email` when they detect a valid email address in the text input currently selected.  This feature causes some crashes on my react native app. Sadly I can't reproduce in on an empty project, but it always happens on some inputs in my app. I have not been able to determine the difference between the textinputs that causes the problem and the textinput that does not cause any crash.

React Native version: 0.61.2

## Steps To Reproduce

1. Select a textinput
2. digit a valid email address (such as `test@test.com`). As soon as I type the last `c` (creating a valid email format), the app crashed

The error:

> java.lang.NullPointerException · Attempt to invoke direct method 'void android.widget.Editor$SelectionModifierCursorController.initDrawables()' on a null object reference

The full stack trace:
```
java.lang.NullPointerException: Attempt to invoke direct method 'void android.widget.Editor$SelectionModifierCursorController.initDrawables()' on a null object reference
        at android.widget.Editor$SelectionModifierCursorController.access$300(Editor.java:6696)
        at android.widget.Editor.getEmailPopupWindow(Editor.java:1469)
        at android.widget.Editor.showEmailPopupWindow(Editor.java:1477)
        at android.widget.Editor.handleEmailPopup(Editor.java:1456)
        at android.widget.Editor.updateCursorPosition(Editor.java:2099)
        at android.widget.TextView.getUpdatedHighlightPath(TextView.java:7813)
        at android.widget.TextView.onDraw(TextView.java:7998)
        at android.view.View.draw(View.java:21472)
        at android.view.View.updateDisplayListIfDirty(View.java:20349)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4396)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4369)
        at android.view.View.updateDisplayListIfDirty(View.java:20309)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4396)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4369)
        at android.view.View.updateDisplayListIfDirty(View.java:20309)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:4396)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:4369)
[...]

PR #30109, opened 2020-10-05
Fix Xiaomi TextInput crash: handling more manufacturers

OlivierFreyssinet:
## Summary

v0.63.3 fixed a crash happening with text inputs on Xiaomi phones (https://github.com/facebook/react-native/issues/27204) with this commit https://github.com/facebook/react-native/commit/07a597ad185c8c31ac38bdd4d022b0b880d02859

However this fix is limited to some devices as it only handles devices who have "Xiaomi" as a brand name, and Xiaomi has many more brands (Pocophone, Redmi etc.).

Here I added the list of known manufacturers that have this issue, from my app's monitoring tool but it's probably incomplete.

Maybe there's a more direct way by checking which OS is running (MiUI) ...

## Changelog

[Android] [Fixed] - Handle more Xiaomi devices to avoid a crash with `TextInput` when `keyboardType` prop is `"email-address"`

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactEditText.java
@@ -48,6 +48,7 @@ import com.facebook.react.views.text.TextAttributes;
 import com.facebook.react.views.text.TextInlineImageSpan;
 import com.facebook.react.views.view.ReactViewBackgroundManager;
 import java.util.ArrayList;
+import java.util.Arrays;
 
 /**
  * A wrapper around the EditText that lets us better control what happens when an EditText gets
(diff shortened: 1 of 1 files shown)


ravirajn22:
How does native developers handle this issue, does it even crash using direct native development? Adding all manufactures does not seem the right way to fix an issue. There should be another way or react-native has some other bug.

AndreasA:
Just wondering but does that fix actually  work because on one of my team's test devices the issue also happens if `keyboardType` is set to `default` as soon as the user enters something that resembles an email address and from what I  can see that is  basically what this fix does?

OlivierFreyssinet:
Indeed there's certainly a cleaner way to do this but I didn't find anything.

Closed because apparently it doesn't even fix the issue on some phones.

PR #30117, opened 2020-10-07
add iOS support for progressViewOffset on RefreshControl

### LINKED ISSUE #10718: Offsetting RefreshControl on iOS via progressViewOffset
### Description

An object class check for `RefreshControl` is hard-coded into `ScrollView` making it impossible to offset the pull-to-refresh loader via JS. Tinkering with the `style` prop on `RefreshControl` can cause crashes and strange visual artifacts. A native solution seems the only near-term viable option besides re-implementing everything from scratch.

I have a fix locally to make the Android-only property `progressViewOffset` on `RefreshControl` work on iOS as well.

I will submit a PR as soon as possible. Please reply here if there is any disinterest in resolving this issue by adding iOS support via `progressViewOffset`

### Reproduction

1. Start a project
2. Set up a <ListView> component
3. Add the `refreshControl={<RefreshControl progressViewOffset={-13}/>}` property.
4. Run in iOS simulator
5. Pull-down while list is at top
6. An activity indicator will slowly appear as you pull down.

The indicator does not move according to the `progressViewOffset` property on `RefreshControl`.

### Additional Information

* React Native version: 0.36
* Platform: iOS 10.0
* Operating System: macOS 10.12.1

Taylor123:
## Summary

fix #10718

Add `progressViewOffset` support to `RefreshControl` in iOS for better feature parity with Android.

## Changelog

[iOS] [Fix] - Add `progressViewOffset` support to `RefreshControl` in iOS for better feature parity with Android.

## Test Plan

Tested locally

### DIFF
--- a/React/Views/RefreshControl/RCTRefreshControl.m
+++ b/React/Views/RefreshControl/RCTRefreshControl.m
@@ -21,6 +21,7 @@
   BOOL _refreshingProgrammatically;
   NSString *_title;
   UIColor *_titleColor;
+  float _progressViewOffset;
 }
 
 - (instancetype)init
@@ -47,6 +48,8 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : (NSCoder *)aDecoder)
     [self beginRefreshingProgrammatically];
   }
   _isInitialRender = false;
+
+  self.frame = CGRectOffset(self.frame, 0, _progressViewOffset);
 }
 
 - (void)beginRefreshingProgrammatically
@@ -158,6 +161,11 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : (NSCoder *)aDecoder)
 {
   _currentRefreshingState = refreshing;
   _currentRefreshingStateTimestamp = _currentRefreshingStateClock++;
+}
+
+- (void)setProgressViewOffset:(float)offset
+{
+	_progressViewOffset = offset / UIScreen.mainScreen.scale;
 }
 
 - (void)refreshControlValueChanged
--- a/Libraries/Components/RefreshControl/RefreshControl.js
+++ b/Libraries/Components/RefreshControl/RefreshControl.js
@@ -69,10 +69,6 @@ type AndroidProps = $ReadOnly<{|
     | typeof RefreshLayoutConsts.SIZE.DEFAULT
     | typeof RefreshLayoutConsts.SIZE.LARGE
   ),
-  /**
-   * Progress view top offset
-   */
-  progressViewOffset?: ?number,
 |}>;
 
 export type RefreshControlProps = $ReadOnly<{|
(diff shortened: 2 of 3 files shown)


React/Views/RefreshControl/RCTRefreshControl.m
```diff
@@ -160,6 +163,11 @@ - (void)setCurrentRefreshingState:(BOOL)refreshing
   _currentRefreshingStateTimestamp = _currentRefreshingStateClock++;
 }
 
+- (void)setProgressViewOffset:(float)offset
+{
+	_progressViewOffset = offset / UIScreen.mainScreen.scale;
```
sammy-SC, line 168: why is this needed? Isn't `offset` already in points?

sammy-SC:
Hello @Taylor123,

thank you for the PR.

Could you post a screenshot where we can see the effect of progressViewOffset property? Is `layoutSubviews` called when `progressViewOffset` is updated? If not, could we use transformation to translate the view? 

Do you think we should add the use of `progressViewOffset` to RNTester, it currently isn't there.

Also the documentation will need to be updated here: https://github.com/facebook/react-native-website/blob/master/docs/refreshcontrol.md

React/Views/RefreshControl/RCTRefreshControl.m
Taylor123: Without the scaling, the offset using `CGRectOffset` didn't match up with the android `progressViewOffset`

Taylor123:
> Could you post a screenshot where we can see the effect of progressViewOffset property? Is `layoutSubviews` called when `progressViewOffset` is updated? If not, could we use transformation to translate the view?

`layoutSubviews` is not called when `progressViewOffset` is updated. It's called on mount and on scroll events.

I don't have much time to dedicate to this at the moment (maybe I will in the near future). I tried quickly reimplementing with transform, but can't get the transform to work. Not sure if there is some quirk with the UIControl in the scrollview or what.

Tried this as well as updating the transform in `layoutSubviews` to no effect.
```
- (void)setProgressViewOffset:(float)offset
{
	self.transform = CGAffineTransformMakeTranslation(0, offset);
}
```

If i can find the time to figure out why the transform isn't working, then i'll post a screenshot with an updated PR.

davidbiedenbach:
@Taylor123 Are you still working on this? Are you open to me taking up the mantle?

I've tried self.transform as well, and of course in this case the offset needs to be negated, but there seems to be some other view that ends up occluding or clipping the RefreshControl. I've had much more luck with adapting your original solution - I extracted the bit that sets the offset frame into a helper function, and I've identified a few places where it needs to be called (mostly because the parent class's methods occasionally re-calculate the frame).

I'm happy to either provide a patch, or open up a new PR, but I'm using it now in a production app and it was just validated by our QA.

cc @sammy-SC

Taylor123:
@davidbiedenbach feel free to contribute however is easiest to you! My requirements changed, so I no longer needed the fix for this. Glad you were able to get a better fix. If you open up a new PR, just tag me here so i can close this one

davidbiedenbach:
@Taylor123 @sammy-SC New PR is here: https://github.com/facebook/react-native/pull/30737

Thanks!

PR #30125, opened 2020-10-07
Remove 'qs' dependency, update yarn lock

Simek:
## Summary

This PR removes unused `qs` package (query string parser) from the main `package.json` file. 

I was not able to find any explicit usage of this package in the repository. Looking at the commit which introduced this package it is not clear why it has been added, query string is not processed in any way. 

I have also run ` yarn update-lock` to clean up lock after latest dependencies bumps. There are no `qs` package entry removal in yarn lock, because it is already required by `request` package.

Refs: https://github.com/facebook/react-native/commit/b4785e514430dc3ba45ed6d136ec63574be88e26

## Changelog

[Internal] [Removed] - remove unused `qs` dependency

## Test Plan

This change should not affect the code base.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,6 @@
     "pretty-format": "^26.5.2",
     "promise": "^8.0.3",
     "prop-types": "^15.7.2",
-    "qs": "^6.5.1",
     "react-devtools-core": "^4.6.0",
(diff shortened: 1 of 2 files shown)


cpojer (approved):
👍

PR #30127, opened 2020-10-07
Flow: fix few FlatList issues

Simek:
## Summary

This PR fixes the two suppressed Flow issues marked as errors. 

I'm not sure if `viewabilityConfig` fallback to empty object is the best solution, maybe the `viewabilityConfig` type should be adjusted instead. I'm looking for feedback for this change.

Introducing those fixes also exposed one more Flow issue with props spread, but after a bit of research on [Diff](https://flow.org/en/docs/types/utilities/#toc-diff) I was able to fix it too.

## Changelog

[General] [Fixed] - Flow: fix few FlatList issues

## Test Plan

Successful `yarn flow` check run.


Simek:
@lunaleaps Can you also look at this PR? 🙂

PR #30142, opened 2020-10-09
Use default for hermes es6 proxy enabled

janicduplessis:
## Summary

Proxy is now enabled by default in hermes 0.7 (https://github.com/facebook/hermes/releases/tag/v0.7.0). However we currently disable it because of the config we pass.

This removes the config so proxy is now enabled.

## Changelog

[Android] [Changed] - Use default for hermes es6 proxy enabled

## Test Plan

Tested that proxy is now enabled (typeof Proxy !== 'undefined') with hermes 0.7.


janicduplessis:
cc @mhorowitz

mhorowitz:
Why did you choose to remove the config option, instead of setting the option to true?

janicduplessis:
I figured the the option will eventually be removed from hermes as I don't see why Proxy support would be disabled if it is stable. Setting it to true explicitly works too if you prefer.

PR #30150, opened 2020-10-09
Update template to xcode 12

janicduplessis:
## Summary

This makes a few change to the template xcodeproj

- Update to recommended settings for Xcode 12
- Remove the main.jsbundle file that doesn't exists, this file is included automatically in the app bundle and doesn't need to be in xcode (it won't even be at that path).
- Add -e to the bundle images script. It seems like Xcode doesn't fail the build on script errors anymore so this makes it so the build stop. If it doesn't stop the app will launch and crash because the bundle doesn't exist (note this is only if running the app in release mode since in debug we don't generate the bundle).

## Changelog

[iOS] [Added] - Update template to xcode 12

## Test Plan

Made the same changes in an app and made sure it ran well.


anujraghuvanshi:
I've XCode 11.5.

I am trying to run react native project with version `0.64.1`. Should it run? Or XCode 12 is min requirement now if want to run with the latest react native version? Not getting any clue from anywhere so asking this here. Kindly confirm.

PR #30151, opened 2020-10-09
Pressability: fix typo in variable, follow event check pattern, small tweak

Simek:
## Summary

This small PR introduces the following changes to the `Pressability`:
* fixes typo in internal `isActivationTransiton` variable name
* assigns `onPressMove` to variable before check and potential usage (this is the common pattern in this file)
* utilizes destructuring assignment to simplify passing coordinates to `_touchActivatePosition`

## Changelog

[Internal] [Fixed] - Pressability: fix typo in variable, follow event check pattern, small tweak

## Test Plan

Successful `yarn test` run.

### DIFF
--- a/Libraries/Pressability/Pressability.js
+++ b/Libraries/Pressability/Pressability.js
@@ -491,8 +491,9 @@ export default class Pressability {
       },
 
       onResponderMove: (event: PressEvent): void => {
-        if (this._config.onPressMove != null) {
-          this._config.onPressMove(event);
+        const {onPressMove} = this._config;
+        if (onPressMove != null) {
+          onPressMove(event);
         }
 
         // Region may not have finished being measured, yet.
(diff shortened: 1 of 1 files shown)


Libraries/Pressability/Pressability.js
```diff
@@ -491,8 +491,9 @@ export default class Pressability {
       },
 
       onResponderMove: (event: PressEvent): void => {
-        if (this._config.onPressMove != null) {
-          this._config.onPressMove(event);
+        const {onPressMove} = this._config;
+        if (onPressMove != null) {
+          onPressMove(event);
         }
```
kevkevinpal, line 497:
```suggestion
        !!onPressMove && onPressMove(event)
```
We could probably use this logic as well I'm pretty sure

Simek: Yeah, but I was trying to mimic the pattern used in the rest of file, for example:
* https://github.com/facebook/react-native/blob/master/Libraries/Pressability/Pressability.js#L717
* https://github.com/facebook/react-native/blob/master/Libraries/Pressability/Pressability.js#L440
* https://github.com/facebook/react-native/blob/master/Libraries/Pressability/Pressability.js#L589

But I can rewrite/replace all the current usages to the line you have proposed in the separate commit, please @kevkevinpal let me know, what's the decision 🙂

kevkevinpal: I'm not sure if it's that way for readability sake since my proposed way is a bit harder to understand if you don't get the logic maybe someone else can answer that but otherwise the pr looks good ☺️

kevkevinpal (approved):

nadiia (approved):
@Simek This looks good. Thank you for the fix!

PR #30179, opened 2020-10-15
style: replace title prop by label prop

cl4pper:
## Summary
Following Material Design patterns, makes more sense to call a button text by label than title, as we can see on the following link, the use of **text label** expression: https://material.io/components/buttons

## Changelog

[General] [Changed] - Replace the word **title** by **label** makes more sense following material design patterns

## Test Plan
Use of
```
<Button
   label="Add item"
   onPress={() => addItem()}
/>
```
instead of
```
<Button
   title="Add item"
   onPress={() => addItem()}
/>
```

### DIFF
--- a/Libraries/Components/Button.js
+++ b/Libraries/Components/Button.js
@@ -26,10 +26,10 @@ import type {ColorValue} from '../StyleSheet/StyleSheet';
 
 type ButtonProps = $ReadOnly<{|
   /**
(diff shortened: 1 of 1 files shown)


TheSavior:
Thanks for the contribution. Unfortunately, this prop is intended to be cross-platform as it has behavior on iOS as well. It is not intended to align with the Material Design names, so I'm going to close out this PR.

PR #30185, opened 2020-10-15
Use getMethods() to include inherited @ReactMethods

### LINKED ISSUE #29893: @ReactNative methods unavailable from Interface
## Description
`JavaModuleWrapper#findMethods()` was refactored to use the method `Class#getDeclaredMethods()` which doesn't find inherited methods.  The native Bluetooth module that I wrote a while back recreated the IOS `RCTEventEmitter` logic in a default Interface which was implemented by my module.   Once React Native is updated to a version AFTER the change, the React Native nativeModule no longer contains any methods from the implemented interface.

This was commit 678679e - Decouple module callback mechanism from CatalystInstance.

Prior to this commit, the method `JavaModuleWrapper#getMethodDescriptors()` used `Class.getMethods()` which returns all public methods.  It seems odd that protected/private methods would be annotated with `@ReactMethod` and more likely that `@ReactMethod` should be public and inheritable (like the IOS version)?

I get it's a super old commit, I really have no idea why it wasn't noticed.  I created the module along side 0.59.1 and it worked great, which is confusing since that tag has the declared method change, which is confusing.  

## React Native version:
Branches: master, origin/0.44-stable, origin/0.45-stable, origin/0.46-stable, origin/0.47-stable, origin/0.48-stable, origin/0.49-stable, origin/0.50-stable, origin/0.51-stable, origin/0.52-stable, origin/0.53-stable, origin/0.54-stable, origin/0.55-stable, origin/0.56-stable, origin/0.57-stable, origin/0.58-stable, origin/0.59-stable, origin/0.60-stable, origin/0.61-stable, origin/0.62-stable, origin/0.63-stable, origin/HEAD and 14 more branches

## Steps To Reproduce

1. Create a new native module
2. Create an interface `public Interface RCTEventEmitter` with a `@ReactMethod` default method
3. Implement the interface in your module 
4. In JS, the implementation will be undefined

## Expected Results
I expect that Implemented or Inherited methods be found in the bridge native module.

https://github.com/kenjdavidson/react-native-bluetooth-classic

egorshulga:
## Summary

Current implementation of `JavaModuleWrapper.findMethods()` is unable to resolve inherited `@ReactMethod`s, because [getDeclaredMethods()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredMethods--) is used. its return excludes inherited methods.

[getMethods()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethods--) call could be used to fix this issue, as it includes all methods from superclasses.

Fixes #29893 

## Changelog

[Android] [Changed] - Use getMethods() to include inherited @ReactMethods

## Test Plan

All checks (incl. tests) passed in Circle CI (when master was not broken :)


yungsters:

PR #30201, opened 2020-10-17
bump Android Gradle Plugin to 4.1.0

dulmandakh:
## Summary

Android Studio 4.1 release with Android Gradle Plugin 4.1.0, thus I expect developers will update to newer version and expect React Native support. Thus release include many fixes and improvements, see[release notes](https://developer.android.com/studio/releases/gradle-plugin#4-1-0), but most importantly Google released it's API documentation at https://developer.android.com/reference/tools/gradle-api. We lacked the API documentation to develop proper React Gradle Plugin.

## Changelog

[Android] [Changed] - bump Android Gradle Plugin to 4.1.0

## Test Plan

RNTester builds and runs as expected, also my apps.


fkgozali:
Will this solve issues mentioned in https://github.com/facebook/react-native/pull/29013#issuecomment-708549178 ? (Or were they even related?)

dulmandakh:
@fkgozali unfortunately this doesn't resolve any issue. I'll investigate the issue within few days, and will create a PR if I can resolve it.

dulmandakh:
@fkgozali the issue if closed now, let's update tools before 0.64 branch.  Thanks

PR #30217, opened 2020-10-20
Sort index.js by alphabetical order and add category comments.

StuffByLiang:
## Summary

As of right now, index.js contains a big blob of imports that are mostly in alphabetical order, with a few exceptions. In addition, the imports aren't separated by Components / API / Plugins, so comments were added for ease of future maintainability. 

## Changelog

[Internal] [Changed] - Reordered imports into alphabetical order and added comments to divide imports into sections (components, apis, plugins, prop types)

## Test Plan

This change did not change the codebase. Building rn-tester still works.

@rickhanlonii

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@
 
 'use strict';
 
+// Components
 import typeof AccessibilityInfo from './Libraries/Components/AccessibilityInfo/AccessibilityInfo';
 import typeof ActivityIndicator from './Libraries/Components/ActivityIndicator/ActivityIndicator';
 import typeof Button from './Libraries/Components/Button';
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -45,6 +46,8 @@ import typeof TouchableWithoutFeedback from './Libraries/Components/Touchable/To
 import typeof View from './Libraries/Components/View/View';
 import typeof VirtualizedList from './Libraries/Lists/VirtualizedList';
 import typeof VirtualizedSectionList from './Libraries/Lists/VirtualizedSectionList';
+
+// APIs
```
rickhanlonii, line 50:
```suggestion
// Modules
```

index.js
```diff
@@ -86,15 +90,18 @@ import typeof useWindowDimensions from './Libraries/Utilities/useWindowDimension
 import typeof UTFSequence from './Libraries/UTFSequence';
 import typeof Vibration from './Libraries/Vibration/Vibration';
 import typeof YellowBox from './Libraries/YellowBox/YellowBoxDeprecated';
-import typeof LogBox from './Libraries/LogBox/LogBox';
-import typeof RCTDeviceEventEmitter from './Libraries/EventEmitter/RCTDeviceEventEmitter';
-import typeof RCTNativeAppEventEmitter from './Libraries/EventEmitter/RCTNativeAppEventEmitter';
+
+// Plugins
```
rickhanlonii, line 94: What's a plugin? I'd move these into `Modules`.

rickhanlonii (commented):
If we change the order of the actual exports, we should comment those sections as well.

cc @TheSavior @yungsters in case there are strong opinions about the sorting of this file.

index.js
StuffByLiang: @rickhanlonii ,

Currently on master, in the export section of index.js, there's a section titled API

Would you like us to change this line as well?

https://github.com/facebook/react-native/blob/9c881b0f9712baef7b6dd6eb530bf371b794fade/index.js#L50

rickhanlonii: I see, ok let's keep API!

index.js
StuffByLiang: Similar to above, there is a section in the module.exports for plugins, does this matter?

https://github.com/facebook/react-native/blob/9c881b0f9712baef7b6dd6eb530bf371b794fade/index.js#L448-L458

rickhanlonii:
Could you check the failing tests?

StuffByLiang:
> Could you check the failing tests?

Seems like the tests failing was caused by another commit previously, rn-tester app still builds with our commit

image

index.js
StuffByLiang: Sure things, thanks :) !

PR #30227, opened 2020-10-21
Fixed eslint warnings 

kevkevinpal:
## Summary
I noticed there were some eslint warnings and decided to make 3 changes to get rid of three of the warnings
> removed inline styling

> used different variable name cause current one is already defined in upper scope

> removed unused type import

## Changelog
[Internal] [Fixed] - reduced eslint warnings by making minor changes

## Test Plan
Running `yarn lint` and seeing fewer warnings show up

### DIFF
--- a/Libraries/Pressability/PressabilityDebug.js
+++ b/Libraries/Pressability/PressabilityDebug.js
@@ -52,22 +52,19 @@ export function PressabilityDebugView({color, hitSlop}: Props): React.Node {
       const baseColor =
         '#' + (normalizedColor ?? 0).toString(16).padStart(8, '0');
 
-      return (
-        <View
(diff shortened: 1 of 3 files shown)


dulmandakh:
Could you please move inline style out of the component render cycle, because that's the thing eslint rule wants you to do. You change silences the warning, but will have same performance impact as inline styles. Thanks

kevkevinpal:
> Could you please move inline style out of the component render cycle, because that's the thing eslint rule wants you to do. You change silences the warning, but will have same performance impact as inline styles. Thanks

Ok I think I just did it, sorry took a while to respond, just left the ones that required the `hitSlop` inside the render cycle

PR #30235, opened 2020-10-23
Fix test_docker: confirm envinfo package install during Docker container setup

hramos:
## Summary

With the release of Node 15 on October 20, 2020, the `nvm i node` command started installing Node 15 by default during the Docker setup step on Circle CI. This version of Node now requires user interaction down in `npx envinfo@latest` during the same step.

With these changes, we ensure we automatically accept the installation of any necessary packages.

## Changelog

[CATEGORY] - Internal

## Test Plan

Circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -587,7 +587,7 @@ jobs:
           name: Build Docker container for Android RNTester App
           command: |
             source ~/.bashrc
-            nvm install 'lts/*'
+            nvm i node
             npm i -g yarn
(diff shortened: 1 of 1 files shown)


hramos:
Ideally, we'd use the same Node version as the JavaScript jobs, but the `nvm` version on this Docker container seems to use a different syntax than my local `nvm` install. All of these failed on Circle:

```
nvm i lts
nvm i --lts
nvm i 'lts/*'
nvm i node --lts
```

PR #30240, opened 2020-10-24
Feature/addKotlinCodegen

SimonErm:
## Summary
This PR adds Generators for Kotlin to Codegen. 
Since Kotlin can interopt with Java seamlessly, the generation of the Kotlin Code is not necessary to use Kotlin for Module-Implementation, but i think it would be nice to have just Kotlin-Code. 

## Changelog

[Internal] [Added] - add Generators for Kotlin to Codegen


ecreeth:
cc @RSNara

hramos:
Hi @SimonErm, I'm excited to see these contributions to the codegen! I think that, for now, it would be best if Kotlin support is provided through a generator in a separate package that exists outside of the core RN repository. I do recognize that we need to provide better support for custom generators, we just haven't gotten to working on that yet.

SimonErm:
Thanks for your feedback @hramos :-) i fully agree with you, i just didn't thought about that before the implementation :-D

Issue #20011, opened 2018-07-03
FlatList item onPress not work the first time after refreshed


kientux:
- [x] Review the documentation: https://facebook.github.io/react-native
  - [x] Search for existing issues: https://github.com/facebook/react-native/issues
  - [x] Use the latest React Native release: https://github.com/facebook/react-native/releases

## Environment
```
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.5
      CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
      Memory: 25.63 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.11.2 - /usr/local/bin/node
      Yarn: 1.7.0 - /usr/local/bin/yarn
      npm: 5.6.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
    IDEs:
      Android Studio: 3.1 AI-173.4819257
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.3.1 => 16.3.1
      react-native: 0.56.0 => 0.56.0
    npmGlobalPackages:
      create-react-native-app: 1.0.0
      react-native-cli: 2.0.1
      react-native-scripts: 1.14.0
```

## Description
`FlatList` has item with `TouchableHighlight`, and a `RefreshControl`  attached.
`onPress` method of `TouchableHighlight` is not working the first time after `onRefresh` called.

![flatlist_bug_report](https://user-images.githubusercontent.com/4200743/42209493-f9f35b0c-7ed8-11e8-89ca-4f2ef52b5bec.gif)

If I scroll `FlatList` a bit after refreshed, then item `onPress` works fine.
// UPDATE: Android does not have this bug.

## Reproducible Demo
Fresh project created by `react-native init`

```js
import React, { Component } from "react";
import { Text, View, FlatList, TouchableOpacity, RefreshControl } from "react-native";

type Props = {};
export default class App extends Component<Props> {
  constructor() {
    super();
    this.state = { refreshing: true, items: [] };
  }

  componentDidMount() {
    this.refresh();
  }

  genItems = () => [0, 1, 2, 3, 4, 5];

  refresh = () => {
    this.setState({ refreshing: true, items: [] });
[...]

PR #30291, opened 2020-10-31
[iOS] fix: First press not working after pull to refresh

yum650350:
## Summary
According to #20011, the first onPress will not work after pull to refresh.

Dive into the code, found out that is because the state `isTouching` in `Libraries/Components/ScrollResponder.js` is not updated after the pull to refresh.

Update the `isTouching` state in `scrollResponderHandleResponderRelease` to fix this.

## Changelog

[iOS] [Fixed] - First press not working after pull to refresh

### Before
- The first onPress fail
ezgif-4-c6aa5383e898
### After
- The first onPress success
ezgif-4-195f9f6c302e


TheSavior:
This is a really great find. Awesome work. Thanks for contributing!

PR #30292, opened 2020-10-31
Build rn-codegen in a temporary directory

janicduplessis:
## Summary

When running yarn install from the codegen directory it will reinstall all dependencies for the react-native workspace inside the react-native package. In my case this caused issues with metro because it would now have 2 copies of it (node_modules/metro and node_modules/react-native/node_modules/metro).

To avoid this copy the react-native-codegen source in a temporary directory and yarn install from there, then copy the built files back.

## Changelog

[Internal] - Build rn-codegen in a temporary directory

## Test Plan

Tested the script in an app with codegen enabled. Fresh install with rn-codegen not built, made sure no extra modules are installed under node_modules/react-native/node_modules.

### DIFF
--- a/packages/react-native-codegen/scripts/oss/build.sh
+++ b/packages/react-native-codegen/scripts/oss/build.sh
@@ -11,9 +11,18 @@ THIS_DIR=$(cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOUR
 set -e
 set -u
 
-pushd "$THIS_DIR/../.." >/dev/null
+TMP_DIR=$(mktemp -d)
+CODEGEN_DIR="$THIS_DIR/../.."
+
+rm -rf "$CODEGEN_DIR/lib"
+
+cp -R "$CODEGEN_DIR/." "$TMP_DIR"
+
+pushd "$TMP_DIR" >/dev/null
 
 yarn install 2> >(grep -v '^warning' 1>&2)
-yarn run build
 
 popd >/dev/null
+
+mv "$TMP_DIR/lib" "$CODEGEN_DIR"
+rm -rf "$TMP_DIR"
--- a/packages/react-native-codegen/scripts/oss/build.sh
+++ b/packages/react-native-codegen/scripts/oss/build.sh
@@ -14,7 +14,7 @@ set -u
 TMP_DIR=$(mktemp -d)
 CODEGEN_DIR="$THIS_DIR/../.."
 
-rm -rf "$CODEGEN_DIR/lib"
+rm -rf "$CODEGEN_DIR/lib" "$CODEGEN_DIR/node_modules"
 
 cp -R "$CODEGEN_DIR/." "$TMP_DIR"
 
@@ -24,5 +24,5 @@ yarn install 2> >(grep -v '^warning' 1>&2)
 
 popd >/dev/null
 
-mv "$TMP_DIR/lib" "$CODEGEN_DIR"
+mv "$TMP_DIR/lib" "$TMP_DIR/node_modules" "$CODEGEN_DIR"
 rm -rf "$TMP_DIR"


packages/react-native-codegen/scripts/oss/build.sh
```diff
@@ -11,9 +11,18 @@ THIS_DIR=$(cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOUR
...
+CODEGEN_DIR="$THIS_DIR/../.."
+
+rm -rf "$CODEGEN_DIR/lib"
+
+cp -R "$CODEGEN_DIR/." "$TMP_DIR"
+
+pushd "$TMP_DIR" >/dev/null
 
 yarn install 2> >(grep -v '^warning' 1>&2)
-yarn run build
```
janicduplessis, line 17: yarn install already builds the package so avoid building it twice.

janicduplessis:
cc @fkgozali @hramos

Sifat1998 (approved):

fkgozali:
I'll take a look at this later this week - was focusing on making CI green until late last week

fkgozali:
Just wanted to also double check. With this change, would all these paths in the codegen gradle plugin still work fine?

https://github.com/facebook/react-native/blob/f023519e491d8352fb384bd1520606c12cbad1ca/packages/react-native-codegen/android/gradlePlugin-build/gradlePlugin/src/main/java/com/facebook/react/codegen/plugin/CodegenPluginExtension.java#L23-L34

Also, will building from source (not from node_modules/react-native) still work? At least for FB internal CI, we want it to run on source directly, not from node_modules

janicduplessis:
@fkgozali Yes it should not affect those, basically all it does is build the /lib folder in a temporary directory then copy it back in the original react-native-codegen package.

One thing I just realized is I probably also want to copy back the node_modules folder in case some codegen dependencies are missing. It works currently as they probably all end up being available transitively.

fkgozali:
Another thing to avoid confusing Metro is to make sure "temp dirs" are excluded from Metro resolution. I had to add one for `buck-out` in this commit: https://github.com/facebook/react-native/commit/17a8737ecb68f61a14f54bf0e2efeb618f97d326#diff-eb288b08cb881cff653283c14cb9630dc0924a1751950e1562a8ca2c0b39694b

Perhaps we could just update that list instead of moving tmp files around?

janicduplessis:
@fkgozali Hmm interesting, but I think it could cause issues since we still want to include `node_modules/react-native/node_modules` in my case it still contains a few libraries when installing normally (react-refresh and ws). The problem also with the current solution is rn-codegen is built inside the current dependency tree so it can mess it up and override whatever is inside `react-native/node_modules`.

packages/react-native-codegen/scripts/oss/build.sh
```diff
@@ -11,9 +11,18 @@ THIS_DIR=$(cd -P "$(dirname "$(readlink "${BASH_SOURCE[0]}" || echo "${BASH_SOUR
 set -e
 set -u
 
-pushd "$THIS_DIR/../.." >/dev/null
+TMP_DIR=$(mktemp -d)
+CODEGEN_DIR="$THIS_DIR/../.."
+
+rm -rf "$CODEGEN_DIR/lib" "$CODEGEN_DIR/node_modules"
+
+cp -R "$CODEGEN_DIR/." "$TMP_DIR"
```
fkgozali, line 19: just to make sure this is compatible with MacOS or Linux (circle CI etc), let's use `rsync` if we can

https://github.com/facebook/react-native/blob/master/packages/react-native-codegen/DEFS.bzl#L80

fkgozali:
Thinking more about this, ideally, any build-time artifacts should stay within a tmp directory, or a designated build dir, that is separate from the src dir. This prevents unexpected issue like what you experienced. So I think, if we can update the codegen plugin to somehow access the lib/cli (output of yarn run build) off this build/tmp dir, that would be best, so we can avoid moving node_modules around (to prevent weird unexpected issues like symlinks etc). This is how it works with Buck world today, see https://github.com/facebook/react-native/blob/master/packages/react-native-codegen/scripts/buck-oss/generate-all.sh#L13.

Do you think we can apply the same model here?

fkgozali:
> Do you think we can apply the same model here?

We'll just merge this one then iterate on my suggestion later

janicduplessis:
@fkgozali Sorry I forgot to come back to you on this. I think it would be possible, but does sound a bit complex since now we'd have to change codegen invocations to use this temp / build dir. Right now I think invoking codegen just uses require so node can find it wherever it is. This flow only happens when building from source in OSS which is a relatively rare setup so I think this might be a good compromise.

fkgozali:
Btw, it looks like this approach isn't compatible with FB internal setup, so I'm modifying your change a bit to do it only when not running within FB.

PR #30305, opened 2020-11-03
Libs v0.62.2

krmao:
## Summary

pure react-native v0.62.2 libs for ios pods

## Changelog

https://github.com/krmao/template/tree/master/mobile/react_native/tools/libsforios/.libsforios


cpojer:
I think you probably messed something up there in your PR, or sent it against the wrong repo. Please fix it.

Issue #29984, opened 2020-09-19
After updating to Xcode 12 I get The linked library 'libPods-*****.a' is missing one or more architectures required by this target: arm64.


Piero87:
## Description
After updating to Xcode 12 I get this error:

 `The linked library 'libPods-*****.a' is missing one or more architectures required by this target: arm64.`

Anyone knows how can solve it?

justinjoyn:
Running into this after I updated Xcode to v12.
```
ld: in /Users/user/Projects/example/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a(cryptlib.o), building for iOS Simulator, but linking in object file built for iOS, file '/Users/user/Projects/example/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a' for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
On `"react-native": "0.62.2"`

SConaway:
Please go back and fill out the issue template

genie09:
> Running into this after I updated Xcode to v12.
> 
> ```
> ld: in /Users/user/Projects/example/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a(cryptlib.o), building for iOS Simulator, but linking in object file built for iOS, file '/Users/user/Projects/example/ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a' for architecture arm64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> 
> On `"react-native": "0.62.2"`

Same error, here. I've met it when I run detox build ios.

env) 
RN 0.63.2
Xcode 12.0,  Build version 12A7209

command)
/Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -workspace ios/HelloDetox.xcworkspace -scheme HelloDetox -configuration Release -sdk iphonesimulator -derivedDataPath ios/build

swplzj:
> ❌  error: The linked library 'libPods-xxx.a' is missing one or more architectures required by this target: arm64. (in target 'xxx' from project 'xxx')

> error Failed to build iOS project. We ran "xcodebuild" command but it exited with error code 65. To debug build logs further, consider building your app with Xcode.app, by opening licaijia.xcworkspace. Run CLI with --verbose flag for more details.

env:
RN 0.61.5
Xcode Version 12.0 (12A7209)
macOS Big Sur Version 11.0 Beta (20A5374g)

[...]

Issue #28824, opened 2020-05-04
[SOLVED] 0.61.5 to 0.62.2 upgrade error: OpenSSL-Universal ENABLE_BITCODE


ezranbayantemur:
After updating 0.61.5 to 0.62.2 with [Upgrade Helper](https://react-native-community.github.io/upgrade-helper/?from=0.61.5&to=0.62.2), ENABLE_BITCODE error drives me crazy. Looks like it happens because of [switching](https://github.com/facebook/react-native/issues/27565#issuecomment-567330597) to `OpenSSL-Universal (1.0.2.19)` from `openssl-ios-bitcode`. Not any problem while creating fresh project but it happens on upgrading to 0.62.2.
 
It happens only <b>debug</b> mode on <b>real iOS device</b>

Trying to run on debug mode it gives error message on below.
`
ld: '... project path .../ios/Pods/OpenSSL-Universal/ios/lib/libcrypto.a(bio_lib.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture arm64
`
It tried `pod cache clear openssl-ios-bitcode` by #27565 (comment) but it couldn't find `openssl-ios-bitcode` pod.

Any idea about it?

## Pods
```ruby
platform :ios, '9.0'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

def add_flipper_pods!(versions = {})
  versions['Flipper'] ||= '~> 0.33.1'
  versions['DoubleConversion'] ||= '1.1.7'
  versions['Flipper-Folly'] ||= '~> 2.1'
  versions['Flipper-Glog'] ||= '0.3.6'
  versions['Flipper-PeerTalk'] ||= '~> 0.0.4'
	  versions['Flipper-RSocket'] ||= '~> 1.0'
	
	  pod 'FlipperKit', versions['Flipper'], :configuration => 'Debug'
	  pod 'FlipperKit/FlipperKitLayoutPlugin', versions['Flipper'], :configuration => 'Debug'
	  pod 'FlipperKit/SKIOSNetworkPlugin', versions['Flipper'], :configuration => 'Debug'
	  pod 'FlipperKit/FlipperKitUserDefaultsPlugin', versions['Flipper'], :configuration => 'Debug'
	  pod 'FlipperKit/FlipperKitReactPlugin', versions['Flipper'], :configuration => 'Debug'
	
[...]

PR #30312, opened 2020-11-03
Fix Detox Tests

utkarsh-dixit:
## Summary

Detox tests were failing because of `DisplayIfVisible` component and `YellowBox`. This PR is meant to fix these issues

## Changelog

[General] [Removed]: Remove DisplayIfVisible component and use simple if else conditions to show which component to show.
[General] [Fixed]: Disable Yellow box so that it doesn't block detox to perform any action on api's tab at bottom.
[General] [Changed]: Update detox from 16.7.2 to 17.10.0
[General][Removed]: Exclude arm64 for rn-tester iOS project
[General][Changed]: Updated pod file for cocoapods dependencies

## Test Plan
You can test this PR by,

`yarn build-ios-e2e && yarn test-ios-e2e`

All Test Passing


hramos:
Thanks for the fix! Can you re-enable the detox tests on Circle CI as part of this PR? I think you'd only need to uncomment some lines in `.circleci/config.yml`, and then we'll be able to confirm these are fixed.

utkarsh-dixit:
@hramos I enabled detox tests for iOS in circleci but they are failing because react native fails to build for iOS (mainly for arm64 and i386). I faced the same issue locally when I was building rn-tester for iOS, which I fixed by following the solution in this issue #29984, which was to exclude `arm64` and `i386` architectures when build rn tester app. 


`ld: in /Users/distiller/react-native/packages/rn-tester/Pods/OpenSSL-Universal/ios/lib/libcrypto.a(cryptlib.o), building for iOS Simulator, but linking in object file built for iOS, file '/Users/distiller/react-native/packages/rn-tester/Pods/OpenSSL-Universal/ios/lib/libcrypto.a' for architecture arm64`

Do you have any ideas about what is causing this issue?

mkcode:
There are a few issues here:
 * There is a compilation error of fsevents during the yarn install step
 * We need to `pod update` in order to get the latest Flipper-Folly/OpenSSLUniversal to solve this issue: https://github.com/facebook/react-native/issues/28824
 * We need to disable building for arm64 for now. See: https://stackoverflow.com/questions/63607158/xcode-12-building-for-ios-simulator-but-linking-in-object-file-built-for-ios

utkarsh-dixit:
@hramos Can you review this PR again, the tests are successfully passing on circle CI now

mkcode:
@utkarsh-dixit - Let's fix the analyze code step first.

packages/rn-tester/js/RNTesterAppShared.js
```diff
@@ -45,11 +45,9 @@ const APP_STATE_KEY = 'RNTesterAppState.v3';
...
 LogBox.ignoreLogs([/AsyncStorage has been extracted from react-native/]);
 
-const DisplayIfVisible = ({isVisible, children}) => (
-  <View style={[styles.container, !isVisible && styles.hidden]}>
-    {children}
-  </View>
-);
+// Disable yellowbox so that they don't cover the API's tab and make it
+// unaccessible for detox.
+LogBox.ignoreAllLogs();
```
mkcode, line 50: Let's test for jest presence here and only disable Logs in detox where jest is defined.

mkcode (changes requested):

kelset:
closing this since Detox is not in the repo anymore: https://github.com/facebook/react-native/pull/32907

PR #30323, opened 2020-11-05
fix: type for Pressable children prop

sanjaytwisk:
## Summary

According to the [documentation](https://reactnative.dev/docs/pressable#children) the `children` prop on the `Pressable` component is not required. This optional type is not reflected in the type annotations which causes a compiler error when trying to implement the `Pressable` component using flow or TypeScript.

This PR updates the type annotations to reflect the optional type of the `children` prop in the `Pressable` component

## Changelog

[General] [Fixed] - Fixed Pressable children prop type annotation

## Test Plan

Tested using `rn-tester` package by adding an example that renders a `Pressable` with a fixed width and height without children

### DIFF
--- a/Libraries/Components/Pressable/Pressable.js
+++ b/Libraries/Components/Pressable/Pressable.js
@@ -57,7 +57,7 @@ type Props = $ReadOnly<{|
    * Either children or a render prop that receives a boolean reflecting whether
    * the component is currently pressed.
    */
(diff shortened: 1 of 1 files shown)


TheSavior:
I'm not sure what a Pressable would even be without children. It sounds like maybe the documentation is wrong?

Issue #29451, opened 2020-07-21
[0.63.1][Android] Dimensions API stops working randomly after a while


cristianoccazinsp:
Please provide all the information requested. Issues that do not follow this format are likely to stall.

## Description
Using `Dimensions.get('window')` and `Dimensions.addEventListener(...)` stops working/firing events randomly after a while. 

I can't figure out exactly what triggers the issue, but after the event listener stops firing, calling `get` will also return incorrect values (e.g., portrait values even if in landscape).  Note that the actual UI still works as expected (rotates/adjusts), but the Dimensions API just completely breaks.

After testing for a couple of hours, the bug seems to happen after a WebView is loaded (with https://github.com/react-native-community/react-native-webview). I can't tell yet whether this was caused by an upgrade in the WebView library or react native itself. I will be linking this issue there as well.

## React Native version:
Run `react-native info` in your terminal and copy the results here.

```
info Fetching system and libraries information...
System:
    OS: macOS 10.15.4
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 108.54 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.3 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 26, 28, 29
      Build Tools: 28.0.3, 29.0.2
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_121 - /usr/bin/javac
    Python: 2.7.15 - /Library/Frameworks/Python.framework/Versions/2.7/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.1 => 0.63.1 
  npmGlobalPackages:
    *react-native*: Not Found
```

## Steps To Reproduce
[...]

Issue #29323, opened 2020-07-09
Dimensions.get('window') is not returning the current values on Android


pmulholland42:
## Description
The width and height returned by Dimensions.get('window') are not updating. The first time it's called, it returns the correct values. But if I change the device screen orientation and call it again, it returns the same values as before. This only happens on Android, and only started happening after I upgraded from RN 0.62.2 to 0.63.0

## React Native version:
0.63.0
Android only

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.

1. Add an onLayout like this to a View in your app: `onLayout={() => {
      console.log(Dimensions.get('window'));
    }`
2. Rotate the device and notice that the width and height printed to the console stay the same after each rotation

## Expected Results
The width and height should update to reflect the new screen layout

JamesClarkeFP:
have you tried the newer useWindowDimensions()?

https://reactnative.dev/docs/usewindowdimensions

pmulholland42:
I haven't yet. Seems handy, but our app is all class components so we can't use it without rewriting them or wrapping them in function components.

oliverdolgener:
I can confirm. On Android the event listener wont trigger on orientation change. It makes no difference if we use `Dimensions.addEventListener('change', setWindowSize);` or `useWindowDimensions()`. On iOS it still works

AndrewMorsillo:
onLayout seems to have changed behavior in 0.63 but I'm not sure exactly how. It's breaking my app on iOS only right now by failing to call onLayout when a parent element resize causes a child element to move on the screen. Seems likely related to this issue. What changed with onLayout in 0.63?

aryojamousi:
weirdly, when using [this](https://github.com/react-native-community/hooks) package the device orientation changes get detected...

NguyenKyThinh94:
> I can confirm. On Android the event listener won't trigger on orientation change. It makes no difference if we use `Dimensions.addEventListener('change', setWindowSize);` or `useWindowDimensions()`. On iOS it still works

Yes, it not work with android !!

CmerCy132:
Any updates on this issue?

R4DIC4L:
I can confirm it is not working on android, but on iOS is it still ok. Using RN 0.63.2. Probably also happening on 0.63.3, I will try to update later.

This has broken my orientation-dependent rendering, I had to check the 'screen' dimensions to get the correct orientation change and I see a lag in the orientation layout change after device rotation of about 2-3 seconds (only on android).

[...]

Issue #29105, opened 2020-06-10
Dimensions.get('screen') is returning reversed values for height and width


snigdhaDhanireddy29:
## Description
When I am using Dimensions.get('screen').width and Dimensions.get('screen').height, I am getting the values as reversed sometimes (i.e width is coming as height and height as width).
This is occurring randomly when the application is launched.
The application is supported only in landscape mode.

## React Native version:
System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 22.54 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.16.1 - /usr/local/bin/node
    npm: 6.9.0 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.4.1/11E503a - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce
Provide a detailed list of steps that reproduce the issue.

1. Use Dimensions.get('screen').width and Dimensions.get('screen').height, its populating reversed values. This is happening randomly sometimes.

## Expected Results
The width and height should be populated correctly

## Snack, code example, screenshot, or link to a repository:
The expected result is the below screen.
![image](https://user-images.githubusercontent.com/56933932/84469674-3cbe9580-ac47-11ea-9529-45fef1295d25.png)

But since the width and height are reversed, we are getting the screen as below.
![image](https://user-images.githubusercontent.com/56933932/84469826-90c97a00-ac47-11ea-8f95-ceecf4fd91dd.png)

chrisglein:
Are you seeing this on iOS or Android or both?
Related to #29290?

ejanzer:
[...]

PR #30324, opened 2020-11-05
[Android] Update the cached dimensions when orientation changes

ajpaulingalls:
## Summary

Currently the dimensions are created once, and then cached.  This change will reload the dimensions when the device orientation changes to insure that dimension update events follow orientation changed events.

this should help address the following issues, that I know of:
https://github.com/facebook/react-native/issues/29105
https://github.com/facebook/react-native/issues/29451
https://github.com/facebook/react-native/issues/29323

## Changelog

[Android] [Fixed] - Dimension update events are now properly sent following orientation change

## Test Plan

Open up RNTester app.  Select the Dimensions API list item.  Rotate the device and verify that the dimensions are correct based on orientation.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
@@ -711,7 +711,9 @@ public class ReactRootView extends FrameLayout implements RootView, ReactRoot {
         return;
       }
       checkForKeyboardEvents();
-      checkForDeviceOrientationChanges();
+      if (checkForDeviceOrientationChanges()) {
+        DisplayMetricsHolder.initDisplayMetrics(getContext().getApplicationContext());
+      }
       checkForDeviceDimensionsChanges();
     }
 
@@ -747,16 +749,17 @@ public class ReactRootView extends FrameLayout implements RootView, ReactRoot {
       }
     }
 
-    private void checkForDeviceOrientationChanges() {
+    private boolean checkForDeviceOrientationChanges() {
       final int rotation =
           ((WindowManager) getContext().getSystemService(Context.WINDOW_SERVICE))
               .getDefaultDisplay()
               .getRotation();
       if (mDeviceRotation == rotation) {
-        return;
+        return false;
       }
       mDeviceRotation = rotation;
       emitOrientationChanged(rotation);
+      return true;
     }
 
     private void checkForDeviceDimensionsChanges() {


luisnaranjo733:
Thanks for sharing this fix @ajpaulingalls. I tested against this branch and can confirm that it fixes the issue I was seeing where the Dimensions.addEventListener wasn't firing callbacks in some scenarios.

It looks like the test_ios_unit_frameworks_hermes check failed, which seems completely unrelated to your change. Are you able to re-queue that CircleCI build somehow? Maybe it's an intermittent failure.

ajpaulingalls:
> It looks like the test_ios_unit_frameworks_hermes check failed, which seems completely unrelated to your change. Are you able to re-queue that CircleCI build somehow? Maybe it's an intermittent failure.

I'm not sure how to kick it off again without making more code changes.  Hopefully a real reviewer will see its unrelated and merge it anyway..:)

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
```diff
@@ -747,16 +749,17 @@ private void checkForKeyboardEvents() {
...
 
-    private void checkForDeviceOrientationChanges() {
+    private boolean checkForDeviceOrientationChanges() {
       final int rotation =
           ((WindowManager) getContext().getSystemService(Context.WINDOW_SERVICE))
               .getDefaultDisplay()
               .getRotation();
       if (mDeviceRotation == rotation) {
-        return;
+        return false;
```
ejanzer, line 758: Is there any reason not to just call `DisplayMetricsHolder.initDisplayMetrics()` here instead of returning the bool? Does it need to be called after `emitOrientationChanged`? It seems like all these methods have lots of side effects anyway

ejanzer (approved):

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
luisnaranjo733: From my testing, it seems that it's less important if ```DisplayMetricsHolder.initDisplayMetrics()``` is called before/after ```emitOrientationChanged()```. What is the most critical is that it gets called before ```checkForDeviceDimensionsChanges()```, because that's where the business logic is currently running with stale state in some scenarios (seems to be related to navigating away from the activity and then coming back).

That being said, it's totally plausible that we _do also_ need to refresh the state before calling emitOrientationChanged for other scenarios. It's possible that there are event listeners for "namedOrientationDidChange" that assume that the dimensions state holder is up to date.

It seems like the safest low risk thing to do would be to take Emily's suggestion here. @ajpaulingalls does that sound reasonable to you?

ejanzer:
I think #29105 is actually a separate issue with `Dimensions.get('screen')` specifically - this PR doesn't seem to fix it for me. I'll look into it.

ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
ajpaulingalls: Since `checkForDeviceOrientationChanges()` is private and currently only called by `onGlobalLayout()` you could move the call to `DisplayMetricsHolder.initDisplayMetrics()` into it.  However, I thought this was cleaner as it helped `onGlobalLayout()` describe better what it does, and its important that `checkForDeviceOrientationChanges()` happen before the `checkForDeviceDimensionsChanges()` so hvagin the logic together seemed to make sense.  Plus, as you say, its less side effecty...:)

ejanzer: Hm, yeah I see. I think it makes sense to always update the display metrics if the orientation has changed, but your way makes it a little more obvious that we need to do that before updating dimensions in JS. I think the clearest way to do this would probably be to just combine checkForDeviceOrientationChanges and checkForDeviceDimensionsChanges into a single function that can enforce the order - what do you think of that? It seems like both functions are only called from this one place anyway.

I've updated the diff internally with the change I suggested above (for some reason I can't seem to sync it back to this PR so you can see it) and I've kicked off a job to merge it, since I know this is a time-sensitive fix for 0.64. If you have any more changes you can update here and I'll re-import (or we can follow up in a separate PR if it's already merged by then).

ajpaulingalls: No more changes on my side.  You are more than welcome to merge the two functions.  I agree that they work together to accomplish the shared goal, since dimensions are inexorably tied to orientation...

luisnaranjo733: Thanks @ejanzer for the expediency here. Once this commit lands in the master branch, I plan to ask for it to be cherry-picked into 0.63-stable in the [0.63.4 discussion](https://github.com/react-native-community/releases/issues/212) to support react-native@0.63 consumers who are currently impacted by this bug, and to support react-native@0.62 consumers who are blocked from upgrading to 63 because of this bug or have already rolled back to 62 to avoid it.

It looks like this PR has not been merged yet. Did the internal merge job fail? I would like to know when this commit has landed so that I can bring it up in the 0.63.4 discussion before it's too late.

ejanzer: Yeah it failed. Looking into it now.

ejanzer: It landed!

luisnaranjo733: Thank you Paul and Emily. This is great work!

thymikee:
Can we close all related issues then?

kelset:
I think that this is fixed only on master, so for 0.64 and 0.63 is still an issue; until we cherry pick this into those and release patches we should keep them open imho

mjmasn:
I don't think this is fixed in the case where you have e.g. a tablet locked to portrait orientation then open a react-native app using `react-native-orientation-locker` and do `Orientation.lockToLandscape()` (not sure about the case where you lock the orientation via AndroidManifest.xml but may be similar). The Dimensions module always returns the width and height reversed both via Dimensions.get('window') and `useWindowDimensions`. I think that may be what https://github.com/facebook/react-native/issues/29105 is referring to @ejanzer 

Using the `useOrientationChange` hook from `react-native-orientation-locker` returns `PORTRAIT` and `useDeviceOrientationChange` returns the 'real' device orientation, neither of which are what we want (as the app is locked to a fixed landscape orientation) so it's hard to even work around the issue (while still accounting for the possibility of split screen / floating window views).

Not strictly a RN issue but it feels like the Dimensions module should take this into account - maybe via `Activity.getRequestedOrientation`

Quedo5:
> Plataforma	Motor	Arco	Tamaño (bytes)	Diferencia
> androide	Hermes	arm64-v8a	7,507,277	+79
> androide	Hermes	armeabi-v7a	7.099.065	+92
> androide	Hermes	x86	7,945,421	+87
> androide	Hermes	x86_64	7.856.485	+92
> androide	jsc	arm64-v8a	8,971,996	+93
> androide	jsc	armeabi-v7a	8.546.899	+100
> androide	jsc	x86	8,971,649	+100
> androide	jsc	x86_64	9.522.956	+93
> [Confirmación](https://github.com/facebook/react-native/commit/3e77e15526def73e54af50e9991c18751bf9aed3) base: [3e77e15](https://github.com/facebook/react-native/commit/3e77e15526def73e54af50e9991c18751bf9aed3)

PR #30331, opened 2020-11-07
remove ReactFragmentActivity

dulmandakh:
## Summary

Remove ReactFragmentActivity class, which was deprecated at least since 0.58.

## Changelog

[Android] [Changed] - remove ReactFragmentActivity class.


fkgozali:
Actually, I just checked, we're still using this internally for other reasons. We could move it to just internal technically, but let's revisit a bit later?

fkgozali:
> Actually, I just checked, we're still using this internally for other reasons. We could move it to just internal technically, but let's revisit a bit later?

Nevermind, the removed file was just an empty abstract subclass of ReactActivity, so I updated our internal callsites.

Issue #30329, opened 2020-11-06
XMLHttpRequest + BlobRegistry memory leak


colbyr:
## Description

This is similar to the other leak I reported https://github.com/facebook/react-native/issues/30288

XMLHttpRequest/fetch requests [register the request body](https://github.com/facebook/react-native/blob/cb7f3f4499232797ce8e60b00270f30641923ddc/Libraries/Network/XMLHttpRequest.js#L247-L248) in [BlobRegistry](https://github.com/facebook/react-native/blob/cb7f3f4499232797ce8e60b00270f30641923ddc/Libraries/Blob/BlobRegistry.js#L11) via the [BlobManager](https://github.com/facebook/react-native/blob/cb7f3f4499232797ce8e60b00270f30641923ddc/Libraries/Blob/BlobManager.js#L114).

```js
const registry: {[key: string]: number, ...} = {};
```

Request bodies are registered whether or not they're used in the JS. Though Blobs can be released from the BlobManager, release is never called in the JS context. While the actual native blob is released via JSI when the JS reference is garbage collected, the id in the BlobRegistry is never removed.

The `registry` object grows indefinitely as the app makes network requests. Like the reporter of #29065, I work on a kiosk-style app. Over the course of days/weeks/months the size of that object becomes noticable.

## React Native version:

<details><summary>react-native@0.63.3</summary>

```sh
robin-rooms-react ➜ npx react-native info
info Fetching system and libraries information...
System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 29.98 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.0, DriverKit 19.0, macOS 10.15, tvOS 14.0, watchOS 7.0
    Android SDK:
      API Levels: 19, 28, 29
      Build Tools: 28.0.3, 29.0.2
[...]

PR #30336, opened 2020-11-07
use a WeakMap in BlobRegistry to prevent a memory leak

colbyr:
## Summary

Fixes https://github.com/facebook/react-native/issues/30329

The blobIds for network requests are never unregistered from the blob registry, so the registry object [grows indefinitely](https://github.com/colbyr/ReactNativeScratch/pull/2). 

Instead of registering blobId strings in an object, we can track the "blob collector" reference in a `WeakMap`. That way if the collector is garbage collected, the entry in the registry will be collected as well.

If this isn't sufficient, I'm happy to explore another avenue. This works in the app I work on, but we don't really use Blobs outside of the network stack.

## Changelog

[General] [Fixed] - stop memory leak caused by stale blobIds

## Test Plan

WeakMaps make this tricky to test. If you log them to the console in safari, the keys end up getting retained until you clear the console 🤪

1. Install an XCode Profile build of an app that makes subsequent network requests
2.Use safari to connect to the JSContext of the running app, take a heap snapshot and find the registry's WeakMap. Log it out to see the entries.
3. Clear the console and press the collect garbage button
4. Take a second heap snapshot and find the registry's WeakMap. Log it out to see the entries have updated, and the entries from the first snapshot are no longer in the registry.

### DIFF
--- a/Libraries/Blob/BlobTypes.js
+++ b/Libraries/Blob/BlobTypes.js
@@ -10,7 +10,22 @@
 
 'use strict';
 
-export opaque type BlobCollector = {...};
+export opaque type BlobCollector: {...} = {...};
+
+// **Temporary workaround**
+// TODO(#24654): Use turbomodules for the Blob module.
+// Blob collector is a jsi::HostObject that is used by native to know
+// when the a Blob instance is deallocated. This allows to free the
+// underlying native resources. This is a hack to workaround the fact
+// that the current bridge infra doesn't allow to track js objects
+// deallocation. Ideally the whole Blob object should be a jsi::HostObject.
+export function createBlobCollector(blobId: string): BlobCollector {
+  if (global.__blobCollectorProvider == null) {
+    return {};
+  } else {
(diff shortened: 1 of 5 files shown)


Libraries/Blob/BlobTypes.js
```diff
@@ -10,7 +10,22 @@
...
+// underlying native resources. This is a hack to workaround the fact
+// that the current bridge infra doesn't allow to track js objects
+// deallocation. Ideally the whole Blob object should be a jsi::HostObject.
+export function createBlobCollector(blobId: string): BlobCollector {
+  if (global.__blobCollectorProvider == null) {
+    return {};
+  } else {
+    return global.__blobCollectorProvider(blobId);
+  }
+}
```
colbyr, line 28: Now when `__blobCollectorProvider` is unavailable, I need to create a placeholder blob collector object. I needed I moved this function here from `BlobManager` so flow would allow me to create instances of the `BlobCollector` opaque type.

PR #30337, opened 2020-11-08
LifecycleEventListener provide default implementations

dulmandakh:
## Summary

LifecycleEventListener provide default implementations so that developers override only methods they need, thus reduce boilerplate code.

Also changed .buckconfig to be consistent with gradle, and target Java 8.

## Changelog

[Android] [Changed] - LifecycleEventListener provide default implementations

## Test Plan

CI is green.

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -12,3 +12,7 @@
 
 [alias]
   rntester = //packages/rn-tester/android/app:app
+
+[java]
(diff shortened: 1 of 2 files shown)


fkgozali:
Internally we haven't fully cut over to Java 8 yet, so this will have to wait for a while.

PR #30342, opened 2020-11-08
Use newer metro language: `blacklistRE` -> `blockList`

rh389:
## Summary

An [update to `metro`](https://github.com/facebook/metro/commit/94c0b541b4bfa17aee4efa0f1969565522ce830d#diff-1a3c1a959bb8c4e2e9743c03cb7a6d0c56648ffcfe129a11b9090bfc139622dd) which landed in metro 0.60 (RN 0.64+) deprecates the config `blacklistRE`, renaming it to `blockList`. Although the former is still supported it now generates a deprecation warning.

## Changelog

[General] [Fixed] - Update metro config language to `blockList`

## Test Plan

Confirm that the config is still respected (`/buck-out/` should be excluded), and that no deprecation warning is issued.

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -19,7 +19,7 @@ const getPolyfills = require('./rn-get-polyfills');
 module.exports = {
   resolver: {
     // $FlowFixMe[signature-verification-failure] Can't infer RegExp type.
-    blacklistRE: /buck-out/,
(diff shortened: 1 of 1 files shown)


rh389:
Thanks @motiz88 ! ;)

Issue #26661, opened 2019-10-01
Scrollview on Android does not call onMomentumScrollEnd when scrolled programmatically 


zomars:
This is a follow-up of #25672.

React Native version:

System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
    Memory: 37.81 MB / 16.00 GB
    Shell: 5.0.9 - /usr/local/bin/bash
  Binaries:
    Node: 8.15.0 - ~/.nvm/versions/node/v8.15.0/bin/node
    Yarn: 1.19.0 - /usr/local/bin/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v8.15.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK:
      API Levels: 23, 25, 26, 28
      Build Tools: 23.0.1, 25.0.1, 26.0.1, 26.0.2, 28.0.1, 28.0.3
      System Images: android-23 | Intel x86 Atom_64, android-23 | Google APIs Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 11.0/11A420a - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: 0.61.1 => 0.61.1
  npmGlobalPackages:
    react-native-cli: 2.0.1

## Steps To Reproduce

1. Call scrollTo({ x: something, y: something, animated: true})
2. Watch how on Android the `onMomentumScrollEnd` isn't being called, while in iOS works fine.

jakubmas:
I faced same issue with the newest React Native. Any solutions?

aravind1078:
same issue no workarounds as of now?

Bardiamist:
I faced with it. Not only programmatically. I use horizontal FlatList. `onMomentumScrollEnd` very often not called if stop scroll with small velosity.

jd20:
Can verify it's still happening in v0.61.5 on Android.

varshan09:
Same issue on Android, is there any solution for it?

djhr:
still happening on react-native v0.63.1

[...]

PR #30367, opened 2020-11-12
[Android] support onMomentumScrollEnd event after programmatic scroll to match iOS behavior

ajpaulingalls:
## Summary
This commit addresses two issues:
1) The recent changes to handlePostTouchScrolling introduced two extra onMomentumScrollEnd events.  Only one should be fired when the scroll is complete, not three
2) on iOS, an onMomentScrollEnd event is fired even when the scroll is caused programmatically.  These changes make Android behave in the same manner, ensuring that you get an end event, regardless of how the scroll was initiated.

This addresses this issue that was raised over a year ago:
https://github.com/facebook/react-native/issues/26661

and this issue from 2018
https://github.com/facebook/react-native/issues/21718

## Changelog

[Android] [Fixed] - Support onMomentumScrollEnd event after programmatic scroll to match iOS behavior

## Test Plan

I have made updates to the RNTester app to demonstrate the changes, specifically in FlatListExample and ScrollViewSimpleExample.  
With the FlatListExample, type into either the scrollToIndex... or the (new) scrollToOffset... text boxes and watch the onMomentumScrollEnd log message appear.  Similarly, you can manually scroll the scroll view and see both begin and end scroll events (and only one end scroll event)
With the ScrollViewSimpleExample, tap the new buttons on the top and watch the logs.

You can now run the app on both Android and iOS and see the same behavior.

### DIFF
--- a/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
+++ b/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactHorizontalScrollView.java
@@ -70,6 +70,7 @@ public class ReactHorizontalScrollView extends HorizontalScrollView
   private boolean mDragging;
   private boolean mPagingEnabled = false;
   private @Nullable Runnable mPostTouchRunnable;
+  private @Nullable Runnable mPostSmoothScrollRunnable;
   private boolean mRemoveClippedSubviews;
   private boolean mScrollEnabled = true;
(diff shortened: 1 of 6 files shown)


ajpaulingalls:
The last two failures are unrelated to my changes.  This PR should be good to merge...

dimaportenko:
Can we do anything to push this PR forward?

dimaportenko:
so sad :(

PR #30372, opened 2020-11-12
fix glog pod install with Xcode 12

dulmandakh:
## Summary

Today I created a project with RN 0.63 and it failed to **pod install**, and after some investigation I found following error. It's caused by Xcode 12 because it dropped support for 32 bit architectures, but following script sets architecture to armv7 which is 32bit.

This will change architecture to arm64, 64 bit.

```
configure:3727: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -arch armv7 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.1.sdk    conftest.c  >&5
clang: error: invalid iOS deployment version 'IPHONEOS_DEPLOYMENT_TARGET=12.0', iOS 10 is the maximum deployment target for 32-bit targets [-Winvalid-ios-deployment-target]
configure:3731: $? = 1
```

## Changelog

[IOS] [Changed] - fix glog pod install with Xcode 12

## Test Plan

Create a new project using react-native init, and it'll fail to do pod install. When this change applied, it'll do it successfully.

### DIFF
--- a/scripts/ios-configure-glog.sh
+++ b/scripts/ios-configure-glog.sh
@@ -16,7 +16,7 @@ if [ -z "$CURRENT_ARCH" ] || [ "$CURRENT_ARCH" == "undefined_arch" ]; then
     if [[ "$PLATFORM_NAME" == *"simulator"* ]]; then
         CURRENT_ARCH="x86_64"
     else
-        CURRENT_ARCH="armv7"
+        CURRENT_ARCH="arm64"
     fi
 fi
 


alloy:
@dulmandakh It's weird that it complains about `IPHONEOS_DEPLOYMENT_TARGET=12.0` whereas [the intention is to target iOS 10](https://github.com/facebook/react-native/blob/1bbcbd55ff94087c0a86b87164c223cee302fe4e/third-party-podspecs/glog.podspec#L37). Having said that, I don’t see any other configuration details in the prepare command regarding deployment target, so I wonder if explicitly adding `-mios-version-min=10.0` would help instead?

Sifat1998 (approved):

dulmandakh:
@alloy is it safe to specify architecture as arm64, which is only architecture supported by Apple? Also I seems to merge more future proof, because someday RN will drop iOS 10 too. I would like to have your recommendation. Thanks

alloy:
I'm confused by this script in general. I would expect this dependency to be build for all supported archs, not a single one. `CURRENT_ARCH` usually refers to the device that’s being build/tested on, but that on its own isn’t enough to prepare a lib for inclusion in production builds.

I’m unfamiliar with the history of this specific lib and its build setup, so I’ll have to test this locally and see what it’s actually building, but won’t get to that until at least tomorrow.

dulmandakh:
Thanks, I'm just curious too

safaiyeh:
@dulmandakh @alloy any updates on this? We have an issue here: https://github.com/facebook/react-native/issues/30690 that bumps into this problem. Is this a potential fix?

dulmandakh:
RN dropped support for iOS 10, thus 32 but ARM with https://github.com/facebook/react-native/commit/a1d626739d95d6cbbb1be169b93952cdd1465486. I think that it's reasonable to merge this change 😎 @fkgozali @PeteTheHeat

PR #30374, opened 2020-11-12
Avoid eating clicks/taps into ScrollView when using physical keyboard

NickGerleman:
## Summary
This is an extension of #29798 which was reverted due to cases where the soft keyboard could not be dismissed.

## Changelog

[General] [Fixed] - Avoid eating clicks/taps into ScrollView when using physical keyboard

## Test Plan

Validated with iOS simulator that taps on default ScrollView will dismiss soft keyboard and be eaten if open, but taps are not eaten when emulating a connected physical keyboard.


lyahdav:
I verified the following:
1. The prior PR (https://github.com/facebook/react-native/pull/29798) causes the bug mentioned there on my iOS simulator and Android device
2. This PR does not have that bug on my iOS simulator and Android device

appden:
@NickGerleman I confirmed this fixes an issue we've been having on macOS! We'd really appreciate you making the same PR there as well! cc @alloy

NickGerleman:
Just from the experience last time, it might be wise to wait a moment after this is merged before bringing into other forks 😉. I was about to bring the previous version of this change to older versions of RNW, but it was reverted within a day of being in core.

PR #30381, opened 2020-11-12
Add example for focus and blur events to AppStateExample

suminkimm:
## Summary
Added an example to test focus and blur events in AppState (Android only)

## Changelog

[Android] [Added] - Add example for focus and blur events to AppStateExample

## Test Plan
Example


suminkimm:
@rickhanlonii the only test that's failing is test_ios_unit_frameworks_hermes, which I believe is also failing on master!

PR #30406, opened 2020-11-17
Add modal example

suminkimm:
## Summary
Add examples for Modal component in RNTester for:
- onDismiss (iOS)
  - Note: there may be an issue with onDismiss (#29455 )
- onShow
- hardwareAccelerated (Android)
- statusBarTranslucent (Android)

## Changelog

[General] [Add] - Added examples for Modal component in RNTester

## Test Plan

Example GIF

### DIFF
--- a/packages/rn-tester/js/components/ExamplePage.js
+++ b/packages/rn-tester/js/components/ExamplePage.js
@@ -78,7 +78,7 @@ const styles = StyleSheet.create({
   examplesContainer: {
     width: ScreenWidth,
     flexGrow: 1,
-    flex:1
+    flex: 1,
   },
   description: {
     marginVertical: 8,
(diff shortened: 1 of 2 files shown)


suminkimm:
@rickhanlonii The only failing test is test_ios_unit_frameworks_hermes which was failing on master!

PeteTheHeat:
Thanks for working on this :) 

Just checking, did you test on iOS as well? (since the diff makes some iOS specific changes, and I don't see that in your test plan)

suminkimm:
> Thanks for working on this :)
> 
> Just checking, did you test on iOS as well? (since the diff makes some iOS specific changes, and I don't see that in your test plan)

@PeteTheHeat Yup, I've tested it on iOS as well.

PR #30417, opened 2020-11-17
Fix :ReactAndroid:androidJavadoc task

### LINKED ISSUE #30415: :ReactAndroid:androidJavadoc Gradle Task Fails Due to Excluding ReactBuildConfig
As part of the publish process, we indirectly run the `androidJavadoc ` task in https://github.com/facebook/react-native/blob/master/ReactAndroid/release.gradle. This task specifically excludes the `ReactBuildConfig` class. seemingly because it is dependent on build-time sources (see https://github.com/facebook/react-native/blob/3a0ed2e5510ec14a695195d9e8b7d9736c3417d6/ReactAndroid/release.gradle#L81). It does not seem like the gradle task ensures these sources are built.

https://github.com/facebook/react-native/commit/e41ee4296795566baf9c8df68293a0e8a1b12b2b added an import `ReactBuildConfig` which causes the task to fail with the following error:

```
error: cannot find symbol

  symbol:   variable ReactBuildConfig
  location: class ReactEditText
1 error
100 warnings
```

This blocks the ability to publish a new npm package.

A quick fix is to also exclude ReactEditText from Javadoc generation, but we should ideally fixup the task to import the build-time class to avoid usages of the class breaking doc generation.

NickGerleman:
## Summary

Fixes #30415

This is a quick and dirty fix to unblock publish, of excluding a class from Javadoc generation that is importing a class current build logic cannot handle. This is not a long-term fix for the issue.

## Changelog

[Internal] [Fixed] - Fix :ReactAndroid:androidJavadoc task

## Test Plan

Tested that the task now completes locally.

### DIFF
--- a/ReactAndroid/release.gradle
+++ b/ReactAndroid/release.gradle
@@ -79,6 +79,7 @@ afterEvaluate { project ->
         classpath += files(project.getConfigurations().getByName("compile").asList())
         include("**/*.java")
         exclude("**/ReactBuildConfig.java")
+        exclude("**/ReactEditText.java")
     }
 
     task androidJavadocJar(type: Jar, dependsOn: androidJavadoc) {
--- a/ReactAndroid/release.gradle
+++ b/ReactAndroid/release.gradle
@@ -73,13 +73,12 @@ if (JavaVersion.current().isJava8Compatible()) {
 
 afterEvaluate { project ->
 
-    task androidJavadoc(type: Javadoc) {
+    task androidJavadoc(type: Javadoc, dependsOn: generateReleaseBuildConfig) {
         source = android.sourceSets.main.java.srcDirs
         classpath += files(android.bootClasspath)
         classpath += files(project.getConfigurations().getByName("compile").asList())
+        classpath += files("$buildDir/generated/source/buildConfig/release")
         include("**/*.java")
-        exclude("**/ReactBuildConfig.java")
-        exclude("**/ReactEditText.java")
     }
 
     task androidJavadocJar(type: Jar, dependsOn: androidJavadoc) {


ReactAndroid/release.gradle
```diff
@@ -79,6 +79,7 @@ afterEvaluate { project ->
         classpath += files(project.getConfigurations().getByName("compile").asList())
         include("**/*.java")
         exclude("**/ReactBuildConfig.java")
+        exclude("**/ReactEditText.java")
```
fkgozali, line 82: can you try removing these 2 lines then adding this in before line 80?

```
classpath += files("$buildDir/generated/source/buildConfig/release")
```

Not sure if there's a better way to get the buildConfig dir...

NickGerleman: Will try this early tomorrow. Do you know what gradle task generates the BuildConfig? I noticed we seem to run the JavaDoc generation task before some other parts of the build.

```
#!/bin/bash -eo pipefail
node ./scripts/publish-npm.js --nightly
Picked up _JAVA_OPTIONS: -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: https://docs.gradle.org/6.7/userguide/gradle_daemon.html.
Daemon will be stopped at the end of the build stopping after processing
> Task :gradlePlugin-build:gradlePlugin:compileJava UP-TO-DATE
> Task :gradlePlugin-build:gradlePlugin:pluginDescriptors UP-TO-DATE
> Task :gradlePlugin-build:gradlePlugin:processResources UP-TO-DATE
> Task :gradlePlugin-build:gradlePlugin:classes UP-TO-DATE
> Task :gradlePlugin-build:gradlePlugin:jar UP-TO-DATE

> Task :ReactAndroid:androidJavadoc
```

fkgozali: > Do you know what gradle task generates the BuildConfig?

That's just part of Gradle base functionality, explained here: https://developer.android.com/studio/build/gradle-tips#share-custom-fields-and-resource-values-with-your-app-code

NickGerleman: I tested this quickly, and I think generation doesn't happen before the task is run. The task succeeded when I had an already built project, but failed if I deleted `ReactAndroid/build`. first (though other bits were populated in the build directory).

I can dig around a bit more though. It seems like this is a problem that others would have run into.

fkgozali: If you find out which task produces the BuildConfig, we can just make the javadoc task depend on that task

NickGerleman: Found it. Looks like I can declare `dependsOn:generateReleaseBuildConfig` and things work. The task seems somewhat private, so I wonder if it makes sense to do a more full Android build, but it solves the problem nicely.

PR #30434, opened 2020-11-19
Update Infinite Red's relevant partner information

kevinvangelder:
## Summary
Updates Infinite Red's relevant partner information

## Changelog

[DOCS] [PARTNER] - Updates Infinite Red's partner information


cpojer:
Thanks for your PR. The intention of the text in that document is to give an outline of who is partnering on React Native and mention some of the projects they may be involved in but not be an exhaustive list. If we merge this PR, we will have to expand the section for all partners to mention more projects and then have to take on the maintenance burden to add/remove projects over time. I would like us not to do that, so I'll be closing the PR.

If there is a factual inaccuracy in the existing text (ie, a project listed already was deprecated), then we should fix that, though.

PR #30445, opened 2020-11-20
Add methods to create ArrayBuffers from byte arrays

savv:
## Summary
This PR adds a factory method to the ArrayBuffer class, so that JSI methods can return them as results.

## Changelog
[General] [Added] - Factory method for ArrayBuffer in the JavaScriptCore implementation of JSI

## Test Plan
I have a test app where I can directly return data from LevelDB and see ArrayBuffers on the JS side! Let me know if I should add a test case to rn-tester.

---
This change is [Reviewable](https://reviewable.io/reviews/facebook/react-native/30445)

### DIFF
--- a/ReactCommon/jsi/jsi/jsi.h
+++ b/ReactCommon/jsi/jsi/jsi.h
@@ -252,12 +252,15 @@ class JSI_EXPORT Runtime {
 
   virtual String createStringFromAscii(const char* str, size_t length) = 0;
   virtual String createStringFromUtf8(const uint8_t* utf8, size_t length) = 0;
-  virtual Object createArrayBufferFromBytes(const void* bytes, size_t length) = 0;
   virtual std::string utf8(const String&) = 0;
 
   // \return a \c Value created from a utf8-encoded JSON string. The default
   // implementation creates a \c String and invokes JSON.parse.
   virtual Value createValueFromJsonUtf8(const uint8_t* json, size_t length);
+
+  virtual Object createArrayBufferFromBytes(const void* bytes, size_t length) {
+    throw std::runtime_error{"Not implemented."}; 
+  }
 
   virtual Object createObject() = 0;
   virtual Object createObject(std::shared_ptr<HostObject> ho) = 0;
--- a/ReactCommon/jsi/jsi/jsi.h
+++ b/ReactCommon/jsi/jsi/jsi.h
@@ -252,6 +252,7 @@ class JSI_EXPORT Runtime {
 
   virtual String createStringFromAscii(const char* str, size_t length) = 0;
   virtual String createStringFromUtf8(const uint8_t* utf8, size_t length) = 0;
+  virtual Object createArrayBufferFromBytes(const void* bytes, size_t length) = 0;
   virtual std::string utf8(const String&) = 0;
 
   // \return a \c Value created from a utf8-encoded JSON string. The default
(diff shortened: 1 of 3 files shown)


savv:
FYI - this quite similar to https://github.com/facebook/react-native/pull/28961 from @ryantrem 

I'm CCing the same people here in hopes of getting it merged in! :-)
 @mhorowitz, @tmikov @bghgary @ @shergin

tmikov:
You are adding new method to the base JSI interface but only implemented it for JSC. It seems like If we merge it, it will break the Hermes JSI implementation.

ReactCommon/jsi/jsi/jsi.h
```diff
@@ -252,6 +252,7 @@ class JSI_EXPORT Runtime {
 
   virtual String createStringFromAscii(const char* str, size_t length) = 0;
   virtual String createStringFromUtf8(const uint8_t* utf8, size_t length) = 0;
+  virtual Object createArrayBufferFromBytes(const void* bytes, size_t length) = 0;
```
tmikov, line 255: Nit: this declaration probably shouldn't be in the middle of the string-related methods.

ryantrem:
I believe everything under `ReactCommon/jsi/jsi` is automatically merged from `https://github.com/facebook/hermes/tree/master/API/jsi/jsi`, so I think you need to make API changes there first and wait for them to be migrated to the `react-native` repo before adding a JSC specific implementation of `createArrayBufferFromBytes`.

ReactCommon/jsi/jsi/jsi.h
ryantrem, line 255: As @tmikov mentioned, there are multiple implementations of JSI for different JS engines. In addition to JSC and Hermes, there is also V8, Chakra, and a new ABI safe JSI wrapper for Windows. Maybe we could provide a default (non-functional) implementation, and allow JS engine specific implementations to opt-in over time?

```c++
virtual Object createArrayBufferFromBytes(const void* bytes, size_t length) { throw std::runtime_error{"Not implemented."}; }
```

I don't think there is anything like this in JSI already though, so not sure what the right pattern is for expanding API surface area.

ryantrem:
> This PR adds a factory method to the ArrayBuffer class, so that JSI methods can return them as results.

It would be great to see this change make it in. This would be really helpful when dealing with large blocks of data managed by native code but accessible by JavaScript. We would definitely use it in [BabylonReactNative](https://github.com/BabylonJS/BabylonReactNative).

ReactCommon/jsi/jsi/jsi.h
savv: Hi Ryan, I went ahead and added this. There seems to be precedent for this, which makes me think it's a good suggestion:
https://github.com/facebook/react-native/blob/9c32140068463739b91874689f741ea9630d8c3b/ReactCommon/jsi/JSCRuntime.cpp#L942

I also implemented the same functionality in hermes:
https://github.com/facebook/hermes/pull/419

savv: I moved it a bit further below. Does that work?

savv:
As per the discussion on https://github.com/facebook/hermes/pull/419 this can be done by getting the ArrayBuffer constructor: https://gist.github.com/mhorowitz/4915d43b8d82ae0eb98a157e689ba23e

So I'm closing this PR.

mrousavy:
@savv this does still copy the "byte array". In my case I receive a buffer (`void *`) from an API and I would like to create an `ArrayBuffer` from that given `void *` buffer, that is currently not possible using JSI and your PR would solve this (afaik) - so why close it?

savv:
Capturing our discussion for posterity:
The implementation in this PR was still copying bytes around. You’d have to change Hermes in a few places to use std::move instead of copying strings, may be difficult (I don't understand Hermes's memory ownership model enough to comment)

PR #30448, opened 2020-11-21
[iOS] [Changed] Pass reactNativePath to codegen_pre_install

brunolemos:
## Summary 

`codegen_pre_install` needs the `path` param as much as `use_react_native`, and we pass to `use_react_native` by default.

https://github.com/facebook/react-native/blob/c0a2998387d1730d38a28cbe296520400a645b25/template/ios/Podfile#L9

https://github.com/facebook/react-native/blob/c0a2998387d1730d38a28cbe296520400a645b25/template/ios/Podfile#L16-L18

## Changelog

[iOS] [Changed] - Pass reactNativePath to codegen_pre_install

## Test Plan

Tested on my local machine, `pod install` fails if I set `reactNativePath` to an wrong path and works if it's correct:

image

ping: https://github.com/react-native-community/releases/issues/207

### DIFF
--- a/template/ios/Podfile
+++ b/template/ios/Podfile
@@ -14,7 +14,7 @@ target 'HelloWorld' do
   end
 
   pre_install do |installer|
-    codegen_pre_install(installer)
+    codegen_pre_install(installer, :path => config[:reactNativePath])
   end
 
   # Enables Flipper.


brunolemos:
Closing since codegen invocation was moved out of the podfile https://github.com/facebook/react-native/commit/c901c1fbce080c6e9b6ed56dbccaa7f2fa0571b2#diff-fcc02228077c73cb58800911f68c7de9938b9848279ee6caa73c27dd63d27212

PR #30449, opened 2020-11-21
Integrate Native Module codegen into Xcode build pipeline

hramos:
Move the codegen invocation out of Podfiles and into the FBReactNativeSpec Pod itself. With this change, developers do not need to modify their existing project's Podfiles, and yet the codegen will be integrated into their projects automatically by way of the FBReactNativeSpec Pod.

This is accomplished in part by injecting a script build phase into the Pods Xcode project that is generated by CocoaPods. The build phase will be executed if the generated FBReactNativeSpec{.h, -generated.mm} files are not present, or if the contents of the Libraries directory has changed. The codegen is also triggered by `pod install` via `prepare_command`.

The codegen will thus be invoked in these situations:

**RNTester:**
* When `pod install` is invoked in `packages/rn-tester/` (`prepare_command`)
* When `packages/rn-tester/RNTesterPods.xcworkspace` is built, if the output files are not present or if the input files have changed (`script_phase`).

**OSS React Native apps:**
* When `pod install` is invoked in `ios/` (`prepare_command`)
* When `ios/AwesomeProject.xcworkspace` is built, if the output files are not present or if the input files have changed (`script_phase`).

Changelog: [Internal] - Moved codegen invocation out of Podfile and into FBReactNativeSpec Pod

Differential Revision: D25138896

### DIFF
--- a/Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
+++ b/Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
@@ -9,12 +9,21 @@ package = JSON.parse(File.read(File.join(__dir__, "..", "..", "package.json")))
 version = package['version']
 
 source = { :git => 'https://github.com/facebook/react-native.git' }
+codegen_path_prefix = ".."
 if version == '1000.0.0'
   # This is an unpublished version, use the latest commit hash of the react-native repo, which we’re presumably in.
   source[:commit] = `git rev-parse HEAD`.strip
+  codegen_path_prefix = "packages"
 else
   source[:tag] = "v#{version}"
 end
+
+react_native_path = File.join(__dir__, "..", "..")
+srcs_dir = File.join(__dir__, "..")
+codegen_script_path = File.join(react_native_path, "scripts", "generate-native-modules-specs.sh")
+codegen_path = File.join(react_native_path, codegen_path_prefix, "react-native-codegen")
+generated_files = [File.join(__dir__, "FBReactNativeSpec", "FBReactNativeSpec.h"), File.join(__dir__, "FBReactNativeSpec", "FBReactNativeSpec-generated.mm")]
+codegen_command = "CODEGEN_PATH=#{codegen_path} sh '#{codegen_script_path}' | tee \"${SCRIPT_OUTPUT_FILE_0}\""
 
 folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
 folly_version = '2020.01.13.00'
@@ -44,4 +53,13 @@ Pod::Spec.new do |s|
   s.dependency "React-Core", version
   s.dependency "React-jsi", version
   s.dependency "ReactCommon/turbomodule/core", version
+
+  s.prepare_command = "touch #{generated_files.reduce() { |str, file| str + " " + file }}"
+  s.script_phase = {
+    :name => 'Generate Native Modules Code',
+    :input_files => [srcs_dir],
+    :output_files => ["$(DERIVED_FILE_DIR)/FBReactNativeSpec-codegen.log"],
+    :script => codegen_command,
+    :execution_position => :before_compile
+  }
 end
--- a/packages/rn-tester/Podfile
+++ b/packages/rn-tester/Podfile
@@ -60,16 +60,6 @@ def frameworks_pre_install(installer)
   end
 end
 
-pre_install do |installer|
-  frameworks_pre_install(installer) if ENV['USE_FRAMEWORKS'] == '1'
-  if ENV['USE_CODEGEN'] != '0'
-    prefix_path = "../.."
-    codegen_path = "../../packages/react-native-codegen"
-    system("./#{codegen_path}/scripts/oss/build.sh") or raise "Could not build react-native-codegen package"
-    codegen_pre_install(installer, {path:prefix_path, codegen_path:codegen_path})
-  end
-end
-
 post_install do |installer|
   flipper_post_install(installer)
 end
--- a/scripts/react_native_pods.rb
+++ b/scripts/react_native_pods.rb
@@ -108,21 +108,3 @@ def flipper_post_install(installer)
     end
   end
 end
-
-# Pre Install processing for Native Modules
-def codegen_pre_install(installer, options={})
-  # Path to React Native
-  prefix = options[:path] ||= "../node_modules/react-native"
-
-  # Path to react-native-codegen
-  codegen_path = options[:codegen_path] ||= "#{prefix}/../react-native-codegen"
-
-  # Handle Core Modules
-  Dir.mktmpdir do |dir|
-    native_module_spec_name = "FBReactNativeSpec"
-    schema_file = dir + "/schema-#{native_module_spec_name}.json"
-    srcs_dir = "#{prefix}/Libraries"
-    schema_generated = system("node #{codegen_path}/lib/cli/combine/combine-js-to-schema-cli.js #{schema_file} #{srcs_dir}") or raise "Could not generate Native Module schema"
-    specs_generated = system("node #{prefix}/scripts/generate-native-modules-specs-cli.js ios #{schema_file} #{srcs_dir}/#{native_module_spec_name}/#{native_module_spec_name}") or raise "Could not generate code for #{native_module_spec_name}"
-  end
-end
--- a/.gitignore
+++ b/.gitignore
@@ -103,7 +103,6 @@ package-lock.json
 /Libraries/FBReactNativeSpec/FBReactNativeSpec
 /packages/react-native-codegen/lib
 /ReactCommon/fabric/components/rncore/
-/schema-native-modules.json
 /schema-rncore.json
 
 # Visual studio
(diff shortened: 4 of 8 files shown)


Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
```diff
@@ -9,9 +9,11 @@ package = JSON.parse(File.read(File.join(__dir__, "..", "..", "package.json")))
 version = package['version']
 
 source = { :git => 'https://github.com/facebook/react-native.git' }
+scripts_path = "../../node_modules/react-native/scripts"
```
brunolemos, line 12: usually the files that references `node_modules/react-native` allows overriding it with a `:path` option to work with monorepo projects. not sure if this is required for this file, but it might

grabbou:
How's this one different from https://github.com/facebook/react-native/commit/e99b8bbb404f8cd1f11b6c7998083be530d7b8a4 that already landed?

alloy:
@grabbou From reading the chats leading up to this PR, I was under the impression that this would entirely move the invocation of codegen to an Xcode build phase, but looks like it’s still also done during `pod install`; so unsure now what the end goal is.

packages/rn-tester/Podfile
```diff
@@ -64,7 +64,9 @@ pre_install do |installer|
   frameworks_pre_install(installer) if ENV['USE_FRAMEWORKS'] == '1'
   if ENV['USE_CODEGEN'] != '0'
     prefix_path = "../.."
-    codegen_pre_install(installer, {path:prefix_path})
+    codegen_path = "../../packages/react-native-codegen"
+    system("./#{codegen_path}/scripts/oss/build.sh") or raise "Could not build react-native-codegen package"
+    codegen_pre_install(installer, {path:prefix_path, codegen_path:codegen_path})
```
alloy, line 69: I’m unsure why codegen is still invoked at all as part of `pod install`. Having it be part of the actual build pipeline means it will always run when necessary, right? So then it seems needlessly redundant to me to still invoke it here too.

hramos: We won't need to generate these as a pre-install hook when they are part of the build pipeline, I agree. I'll send an update soon that moves everything to the FBReactNativeSpec podspec and out of the Podfile.

hramos:
@alloy do you mind taking another pass at reviewing this? I want to make sure the changes make sense when it comes to CocoaPods.

Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
```diff
@@ -9,13 +9,20 @@ package = JSON.parse(File.read(File.join(__dir__, "..", "..", "package.json")))
...
   source[:commit] = `git rev-parse HEAD`.strip
+  codegenPathPrefix = "packages"
 else
   source[:tag] = "v#{version}"
 end
 
+reactNativePath = File.join(__dir__, "..", "..")
+codegenScriptPath = File.join(reactNativePath, "scripts", "generate_native_modules_specs.rb")
+codegenPath = File.join(reactNativePath, codegenPathPrefix, "react-native-codegen")
+codegenCommand = "CODEGEN_PATH=#{codegenPath} ruby '#{codegenScriptPath}'"
```
hramos, line 24: A little context on what is going on here. We need to address two scenarios: a contributor using RNTester within a clone of the react-native repository, and a end-user developing an iOS app with React Native as a dependency.

In the former case, the codegen is available locally in the `react-native` repository checkout. In the latter, the codegen is installed from npm and is in node_modules. We already have some logic in place to determine which one is the case: the version will be set to 1000.0.0 when working inside the react-native repository, otherwise it will match the actual open source version number (e.g. 0.64.0).

In either case, the relative path from this podspec file to the React Native root is the same.

Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
```diff
@@ -44,4 +51,13 @@ Pod::Spec.new do |s|
...
   s.dependency "ReactCommon/turbomodule/core", version
+
+  s.prepare_command = codegenCommand
+  s.script_phase = {
+    :name => 'Generate Native Modules Code',
+    :input_files => [File.join(__dir__, "..")],
+    :output_files => [File.join(__dir__, "Libraries", "FBReactNativeSpec", "FBReactNativeSpec", "FBReactNativeSpec.h"), File.join(__dir__, "Libraries", "FBReactNativeSpec", "FBReactNativeSpec", "FBReactNativeSpec-generated.mm")],
+    :script => codegenCommand,
+    :execution_position => :before_compile
+  }
```
hramos, line 62: Technically, the codegen does not need to run during `pod install` or `pod update`, because it will be invoked by Xcode as needed through the injected build phase (see `s.script_phase`). However, if the generated files do not exist when the FBReactNativeSpec Pod is built, then the sources glob in the Podspec will not match any files and the resulting Pods project will lack any references to FBReactNativeSpec.h or FBReactNativeSpec-generated.mm.

By using prepare_command, we can ensure these files exist because the codegen will run before the FBReactNativeSpec Pod is built.

We then use input_files and output_files to provide hints to Xcode as to whether running the codegen again at build time is necessary: if the output files are not present, the codegen must run; if the input files have changed, the codegen must run; if neither condition is met, the codegen can be skipped.

packages/rn-tester/Podfile
```diff
@@ -60,16 +60,6 @@ def frameworks_pre_install(installer)
...
-pre_install do |installer|
-  frameworks_pre_install(installer) if ENV['USE_FRAMEWORKS'] == '1'
-  if ENV['USE_CODEGEN'] != '0'
-    prefix_path = "../.."
-    codegen_path = "../../packages/react-native-codegen"
-    system("./#{codegen_path}/scripts/oss/build.sh") or raise "Could not build react-native-codegen package"
-    codegen_pre_install(installer, {path:prefix_path, codegen_path:codegen_path})
-  end
-end
-
```
hramos, line 72: In the end, we won't need to modify the Podfile at all.

scripts/generate_native_modules_specs.rb
```diff
@@ -0,0 +1,34 @@
...
+    system("node #{generateSpecsCliPath} ios #{schemaFilePath} #{outputDir} #{libraryName}") or raise "Could not generate code for #{libraryName}"
+  end
+end
+
+def main()
+  codegenPath = ENV["CODEGEN_PATH"] ||= File.join(reactNativePath, "..", "react-native-codegen")
+  codegen_core_modules(codegenPath)
+end
+
+main()
```
hramos, line 34: Right now, this script will only handle the core native modules in Libraries/. It could easily be extended in a future PR to handle native modules in arbitrary locations (e.g. RNTester's own native modules, third party modules...)

scripts/react_native_pods.rb
```diff
@@ -108,21 +108,3 @@ def flipper_post_install(installer)
...
-
-  # Handle Core Modules
-  Dir.mktmpdir do |dir|
-    native_module_spec_name = "FBReactNativeSpec"
-    schema_file = dir + "/schema-#{native_module_spec_name}.json"
-    srcs_dir = "#{prefix}/Libraries"
-    schema_generated = system("node #{codegen_path}/lib/cli/combine/combine-js-to-schema-cli.js #{schema_file} #{srcs_dir}") or raise "Could not generate Native Module schema"
-    specs_generated = system("node #{prefix}/scripts/generate-native-modules-specs-cli.js ios #{schema_file} #{srcs_dir}/#{native_module_spec_name}/#{native_module_spec_name}") or raise "Could not generate code for #{native_module_spec_name}"
-  end
-end
```
hramos, line 128: The Podfiles no longer handle codegen. This logic has been moved to a standalone Ruby script, `scripts/generate_native_modules_specs.rb`.

Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
alloy: This is correct, CocoaPods will only include files that exist on disk at the time of running `pod install`.

Your solution is good, but I wonder if it would be helpful to reduce the number of places from where codegen is invoked, i.e. to only run it as part of the Xcode build? We should be able to make this work by changing the prepare command to `touch` those paths so they _exist_ but are simply empty files. What do you think?

scripts/generate_native_modules_specs.rb
alloy: Also noting that [in the future] there is no reason that this script should remain in Ruby instead of JS or shell script.

alloy (approved):
Great work, @hramos 👌 

One question is wether it makes sense to remove the actual invocation of the codegen command during `prepare_command` phase.

scripts/generate_native_modules_specs.rb
hramos: Good point. I'd rather not add a new script, so I'll use the existing bash script with some minor changes to allow consuming the codegen from an arbitrary path.

Libraries/FBReactNativeSpec/FBReactNativeSpec.podspec
hramos: `touch` works for me as well. To ensure the script phase runs, I'll remove the codegen files from output_files, otherwise Xcode might skip the build phase.

hramos (commented):
Thanks for taking a look! I've made the requested changes, and the commit should sync back to GitHub soon.

alloy:
Ace 👍

PR #30461, opened 2020-11-23
Add instructions to template/ios/Podfile for enabling hermes

SConaway:
## Summary

Just thought I'd add these instructions so devs don't have to check the docs. Also, it makes iOS match Android with instructions in the configuration files
## Changelog
N/A (in my opinion)

## Test Plan
N/A (because not a code change)

### DIFF
--- a/template/ios/Podfile
+++ b/template/ios/Podfile
@@ -8,8 +8,8 @@ target 'HelloWorld' do
 
   use_react_native(
     :path => config[:reactNativePath],
-    # Or, to enable hermes on iOS, uncomment the below line:
-    # :hermes_enabled => true
+    # to enable hermes on iOS, change `false` to `true` and then install pods
+    :hermes_enabled => false
   )
 
   target 'HelloWorldTests' do
--- a/template/ios/Podfile
+++ b/template/ios/Podfile
@@ -6,9 +6,11 @@ platform :ios, '10.0'
 target 'HelloWorld' do
   config = use_native_modules!
 
-  use_react_native!(:path => config[:reactNativePath])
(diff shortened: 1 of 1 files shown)


template/ios/Podfile
```diff
@@ -7,6 +7,8 @@ target 'HelloWorld' do
   config = use_native_modules!
 
   use_react_native!(:path => config[:reactNativePath])
+  # Or, to enable hermes on iOS, comment out the above line and uncomment the below:
+  # use_react_native!(:path => config[:reactNativePath], :hermes_enabled => true)
```
alloy, line 11: What do you think about doing it like so?

```ruby
use_react_native(
  :path => config[:reactNativePath],
  # Or, to enable hermes on iOS, uncomment the below line:
  # :hermes_enabled => true
)
```

alloy (commented):
Thanks a bunch for taking the time! 🙏 

This is already great as is, but wondering if it feels more natural to not have to comment a whole duplicated line and if this better sets us up for other potential future options.

template/ios/Podfile
SConaway: I like that idea better! I just didn't think about it.

SConaway:
Done!

TheSavior:
Should the template just have `:hermes_enabled => false`? That should also help make it clear how to enable it even for people who don't enable it initially

alloy:
@TheSavior Smart, yeah that’s a good call 👍

kelset (approved):
LGTM

alloy (approved):
Thanks!

PR #30466, opened 2020-11-24
Feature: Android Pressable `foreground` ripple

andreialecu:
## Summary

`Pressable` does not support foreground Android ripples: https://reactnative.dev/docs/touchablenativefeedback#useforeground

This is needed for certain scenarios otherwise `Pressable` cannot be used, and users have to resort to `TouchableNativeFeedback` instead.

`foreground` is necessary when using `react-native-linear-gradient` for example, otherwise the ripple renders underneath the gradient and if used for buttons, it doesn't show any feedback.

https://github.com/react-native-linear-gradient/react-native-linear-gradient/issues/64

## Changelog

[Android] [Added] Pressable: Add foreground prop to android_ripple

## Test Plan

With `foreground: true`:

M2bx2dILpj

Also, here's a different example of using `foreground` `true/false`:

`foreground: false` (default):
Tqb6REqXyC

`foreground: true` (new):
4hv3UVFs1d


andreialecu:
The implementation is the same as the one here:

https://github.com/facebook/react-native/blob/0675beee11a51a346281f92cee56dbe2320c084f/Libraries/Components/Touchable/TouchableNativeFeedback.js#L312-L315

wolframkriesing:
great job @andreialecu, looks like this will solve [some shortcoming I stumbled upon](https://picostitch.com/tidbits/2020/12/android-ripple-effect-analyzed/)

chuchuva:
@andreialecu How do you make the ripple effect with round corners? This is what I get:

Animation

Snack demonstrating the problem: https://snack.expo.dev/6U8dxxzLx

Would this pull request fix this problem?

andreialecu:
You need to add a wrapping view with the correct border-radius + overflow hidden.

See: https://stackoverflow.com/questions/63048178/ripple-effect-leaking-at-corners-as-if-pressable-button-has-a-borderradius

